PR #1152, opened 2021-04-20
Fix navbar style for active page on example

ceifa:
The current navbar of svelte-kit-demo example it's not working properly since it uses the legacy `segment` prop on layout. This PR just fix it using the new page store.


Rich-Harris:
thank you — though this reminds me that we were planning to remove this example project from the repo! (the project in `packages/create-svelte/templates/default` is the canonical SvelteKit demo at this point)

PR #1160, opened 2021-04-21
chore: silence prettier unknown option warning

ignatiusmb:
Silences this wall of warnings

image

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -3,6 +3,13 @@
 	"singleQuote": true,
 	"trailingComma": "none",
 	"printWidth": 100,
-	"svelteSortOrder": "scripts-markup-styles",
-	"svelteBracketNewLine": true
+	"overrides": [
+		{
+			"files": ["*.svelte"],
+			"options": {
+				"svelteSortOrder": "scripts-markup-styles",
+				"svelteBracketNewLine": true
+			}
+		}
+	]
 }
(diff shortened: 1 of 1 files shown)


benmccann:
If I run `pnpm format` on `master` I don't get any warnings. How are you running it? On what OS?

Conduitry:
I'm seeing many of these warning, both for `pnpm format` and `pnpm lint`.

With later versions of prettier-plugin-svelte (like what's used in this package), the default for `svelteSortOrder` has changed (and the valid values have also changed, as `options` is now one of the items that can be included anywhere in the order.) The (new) default of `options-scripts-markup-styles` should be fine, so this option can just be removed from the `.prettierrc` in this repo.

However, the `svelteBracketNewLine` option that there is also a warning about _is_ a valid option. Using `overrides` to ensure that this option is only passed on files that are using a plugin that understands it makes sense.

Rich-Harris:
oh my god thank you this was driving me nuts

ignatiusmb:
> If I run `pnpm format` on `master` I don't get any warnings. How are you running it? On what OS?

It happens when we run the `lint` script, using `pnpm -r lint` from root will collapse most of them. That wall happens when executed on each packages individually. I think it happens on all OS, we can also see it on the previous ci workflows runs in its lint check before this PR.

<details>
<summary>Using <code>pnpm -r lint</code> from root</summary>
executed individually
</details>

<details>
<summary>Lint check on previous workflow run</summary>
executed individually
</details>

PR #866, opened 2021-04-04
make the --host flag provision unnecessary for production deployments

phaleth:
While the ability to avoid dev deployment of SvelteKit to not be exposed to an untrusted network. wherever the developer has to reside at the moment of dev work, such as in a mall, on a train or in a coffee shop using the available Wi-Fi network, is great. The same ability is not so great for production deployments where currently the `package.json` has to be modified before each deployment with the `--host 0.0.0.0` flag.

This PR attempts to make the `--host 0.0.0.0` the default suitable for deployments which leverage `npm start` or `npx svelte-kit start` command.

PR #1163, opened 2021-04-21
Remove the `start` command from the CLI in favor of a `--preview` flag on `build`

babichjacob:
People get confused about using `svelte-kit start` for production: https://github.com/sveltejs/kit/pull/866, and countless mistakes in the Discord server

This adds friction to the `start` functionality by always requiring a build to proceed it; this should hopefully make people ask "should this really build every time I want to start the server?" to which they hopefully conclude "no" and figure out the right way to start their app in production. There has also been some documentation changes to (hopefully) better address this.

Testing the production build locally is now easier (less friction) because `svelte-kit build` and `svelte-kit start` have been combined into a single command immediately accessible with the `preview` package script. I am not particularly attached to the `preview` name (neither for the flag nor the package script), so if anyone wants to change it, we can.

I don't believe we have CLI tests. I ran the examples locally and it works as expected. I would worry about a race condition if not all files are flushed to the filesystem before the `start` function runs, but I have no evidence to back up that this is a possibility.

### Tests

### Changesets

### DIFF
--- a/.changeset/friendly-camels-reflect.md
+++ b/.changeset/friendly-camels-reflect.md
@@ -1,0 +1,6 @@
+---
+'create-svelte': patch
+'@sveltejs/kit': patch
+---
+
+Remove the `start` command from the CLI in favor of a `--preview` flag on `build`
--- a/documentation/docs/06-service-workers.md
+++ b/documentation/docs/06-service-workers.md
@@ -10,4 +10,4 @@ In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker
 
 Inside the service worker you have access to the [`$service-worker` module](#modules-service-worker).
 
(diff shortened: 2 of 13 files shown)


benmccann:
I reformatted your PR description because having all the text in bold was a bit too hard to read :grin: 

I really like the idea of renaming `npm run start` to `npm run preview` as it makes it much clearer not to use it in production and I've also seen people hitting this frequently.

I'm not sure I like the idea of adding friction just for friction's sake. Is there some other benefit of building each time the server is run?

Conduitry:
Is there any previous discussion about this change that I can read, or any previously documented buy-in from other team members?

babichjacob:
> Is there any previous discussion about this change that I can read, or any previously documented buy-in from other team members?

No

Rich-Harris:
Sorry, I don't think this makes sense at all. I definitely don't want to be _forced_ to rebuild my app in order to preview it (e.g. I closed the server because I thought I was done then realised there was something else I needed to check).

Like Ben, I'd be on board with changing `svelte-kit start` to `svelte-kit preview`

PR #1167, opened 2021-04-22
account for the extremely helpful and intuitive npm behaviour

### LINKED ISSUE #1159: Include .gitignore in blank template
When starting out a new project using blank template it's really frustrations that the current version of cli doesn't come with a 
 .gitingore file which is really helpfully

Rich-Harris:
fixes #1159

### Tests

### Changesets

### DIFF
--- a/.changeset/two-students-melt.md
+++ b/.changeset/two-students-melt.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Add .gitignore files to new projects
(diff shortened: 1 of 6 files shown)


Conduitry:
I'm a little perplexed as to why this npm behavior matters. Shouldn't the published package not have any `.gitignore`s in it all, and instead have JSON files containing the names and contents of the files to write?

Rich-Harris:
I realised that was causing a fair bit of duplication, which was quite costly when applied to the base64-encoded assets (i.e. the welcome image), and moved stuff that isn't variant-specific (i.e. components and .ts/.js files) back to the filesystem

Conduitry:
I see, makes sense, thanks.

PR #1174, opened 2021-04-22
Add check script for create-svelte package

ignatiusmb:
This PR will
- add `check` script for `create-svelte`
- fix lint errors in `bin.js`
- reorder prettier ignore-path (fix _local_ lint/check error)


Rich-Harris:
thanks 👍

PR #1178, opened 2021-04-22
Don't swallow errors in hooks.js

### LINKED ISSUE #1170: Errors in hooks.js are silently swallowed
**Describe the bug**
Whenever there is any sort of issue with my `src/hooks.js` file it is silently swallowed and the hooks are no longer executed. But there is no notice of any sort of error.

**Logs**
None.

**To Reproduce**
Demo project: https://github.com/pago/bug-svelte-kit-hooks-silent-error

Without the last commit (https://github.com/pago/bug-svelte-kit-hooks-silent-error/commit/5b978684f824c654ba8ce613473f98b1af0a8241), which adds an import for `fs/promises` to the `src/hooks.js` file, the logging within `src/hooks.js` works. Once the import is added, the logging no longer happens and the hooks are just not executed anymore.
There is no warning and no information about why this happens.

**Expected behavior**
If executing the `hooks` fails for any reason, I would expect to get an error message in the console about it so that I know how to fix it and to be aware that is isn't working at the moment.

**Stacktraces**
None.

<details>

- The output of `npx envinfo --system --npmPackages svelte,@sveltejs/kit,vite --binaries --browsers`
```
  System:
    OS: macOS 11.2.3
    CPU: (8) x64 Apple M1
    Memory: 32.47 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.1 - ~/.volta/tools/image/node/14.16.1/bin/node
    Yarn: 1.22.10 - ~/.volta/tools/image/yarn/1.22.10/bin/yarn
    npm: 6.14.12 - ~/.volta/tools/image/node/14.16.1/bin/npm
    Watchman: 4.9.0 - /opt/homebrew/bin/watchman
  Browsers:
    Edge: 90.0.818.42
    Safari: 14.0.3
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.85 
    svelte: ^3.29.0 => 3.37.0 
    vite: ^2.1.0 => 2.2.1 
```

- Your browser
Safari 14.0.3 & Edge 90.0.818.42 (apparently)

- Your adapter (e.g. Node, static, Vercel, Begin, etc...)
`npm run dev` (I assume Node)
</details>

**Severity**
My actual primary issue is not being able to figure out how to read a file in the hooks since importing `fs` fails for some unknown reason (I have a workaround for now). But I've been bitten by this issue a few times already while working on a very small app and it consumes a lot of time trying to figure out why a piece of the functionality no longer works.
Swallowing the errors in what I imagine being a very prominent feature isn't a deal breaker but it does make me nervous about svelte-kit and what other errors I might need to spent a lot of time to find.

I would _recommend_ to treat this as a _medium severity_ case. It's not a major thing but it will cause frustration with engineers when they play around with Svelte Kit and research its suitability for their projects and thus might influence their decision making in a negative way (in case they can't figure out why their hooks are not broken - i.e. I thought hooks were not working in general for a couple days before realizing it was the import).

**Additional context**
Should I open up another issue about not being able to import `fs` in the hooks file? According to the documentation those run only on the server so I was expecting to be able to use that API there. Alternatively I would like to use the `?raw` modifier on an import but that causes the same silent error since I would expect the server side code to also be bundled by rollup/vite.

Rich-Harris:
Fixes #1170. Had to fix some unrelated stuff before I could push these changes, excuse the poor git hygiene

### Tests

### Changesets

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,13 +1,10 @@
 {
-  "$schema": "https://unpkg.com/@changesets/config@1.4.0/schema.json",
-  "changelog": "@changesets/cli/changelog",
-  "commit": false,
-  "linked": [],
-  "access": "restricted",
-  "baseBranch": "master",
-  "updateInternalDependencies": "patch",
-  "ignore": [
-    "hn.svelte.dev",
-    "svelte-kit-demo"
-  ]
+	"$schema": "https://unpkg.com/@changesets/config@1.4.0/schema.json",
+	"changelog": "@changesets/cli/changelog",
+	"commit": false,
+	"linked": [],
+	"access": "restricted",
+	"baseBranch": "master",
+	"updateInternalDependencies": "patch",
+	"ignore": ["hn.svelte.dev"]
 }
--- a/.changeset/lazy-carpets-join.md
+++ b/.changeset/lazy-carpets-join.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Report errors in hooks.js
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,4 +100,4 @@ jobs:
             ${{ runner.os }}-${{ matrix.node-version }}
       - run: npm install -g pnpm
       - run: pnpm install --frozen-lockfile
-      - run: pnpm build && pnpm build --filter="svelte-kit-demo" --filter="hn.svelte.dev"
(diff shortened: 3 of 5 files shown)


benmccann (approved):
lgtm. might be worth adding a test if you have time

Rich-Harris:
awkward thing to add a test for, we'd have to add a whole new test app, which would slow things down. not totally convinced it's worthwhile

PR #1179, opened 2021-04-23
[docs] Update 05-modules.md

leader22:
- Align link to stores
- Add notes for store usage

Does it make sense?


Conduitry:
Thanks, but I don't think we need to link again to the tutorial on readable stores one line below another link to it. I also don't think this is the place to be reminding people of how to use certain Svelte features - if we started doing that, where would it end? I think the single 'readable store' link suffices for this as well.

Issue #907, opened 2021-04-06
Build an output that is usable from local disk (file:// URL)


jzillmann:
**Is your feature request related to a problem? Please describe.**
Quite often i code tools which take some input (like a JSON file) and generate a HTML report out of it. This HTML report might be served by a server (e.g. Github Pages) but often just sits somewhere on the disk and should be openable/consumable from there (e.g. `open file:///Users/x/projects/y/build/test/index.html`).

**Describe the solution you'd like**
From initial experiments (with `adapter-static`) I found 2 things I would need for the above to work which I failed to accomplish with `svelte-kit` so far:
- Links/Includes (from `index.html`) need to be relative instead of absolute
  - `paths/assets||base` were quite resistant setting up anything relative 
- The JS field need to be in `IIFE` format (since [module script imports will raise CORS issues](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#other_differences_between_modules_and_standard_scripts))
  - `vite.build: { lib: { name: "my-app", entry: "src/main.ts", formats: ["iife"] }}` produced `UMD and IIFE output formats are not supported for code-splitting builds`

A solution would be to make the 2 things possible.
Perfect would be a simple way to achieve that. Like an specialized adapter.

**Describe alternatives you've considered**
Maybe what I'm trying to achieve is dump and using svelte-kit for such a thing doesn't make much sense? Just stick to `Svelte Classic` ?

With `Svelte Classic` and Vite I can accomplish the above by tweaking `vite.config.js` in the following way:
- Set `base` to `''` (makes the links relative)
- Configure `IIEF` format:
```
build: {
  lib: { name: "my-app", entry: "src/main.ts", formats: ["iife"] },
}
```
- Move the `index.html` to `public/` and add appropriate script and css includes

**How important is this feature to you?**
This is the first thing I usually set up for the majority of apps I'm crafting.

KonradHoeffner:
I understand your pain as I have the same problem. See also issues https://github.com/sveltejs/kit/issues/1553 and https://github.com/sveltejs/kit/issues/1154.

btakita:
@jzillmann Re your current `Svelte Classic` & `Vite` solution: how are you doing your routing?

jzillmann:
@btakita Not sure what you mean...

btakita:
[...]

PR #1181, opened 2021-04-23
SPA mode

### LINKED ISSUE #754: True SPA mode
We now have the `ssr: false` option which gives us something close to SPA mode: https://kit.svelte.dev/docs#ssr-and-javascript

It doesn't get us all the way there though, because in a typical SPA you would likely want to generate a single fallback page that handles all requests — for example [Surge](https://surge.sh/help/adding-a-200-page-for-client-side-routing) lets you add a 200.html file, while [Netlify](https://docs.netlify.com/routing/redirects/rewrites-proxies/#history-pushstate-and-single-page-apps) allows you to add something like this to `_redirects`:

```
/*    /200.html   200
```

By contrast, SvelteKit expects to generate (whether at runtime or prerender time) an HTML page that includes no content, but _does_ include information about the route that should be hydrated, since the router isn't invoked on load. To create a true SPA, SvelteKit needs to create a content-less file that _doesn't_ contain any route information, and the router needs to figure out what JS to load and execute.

I'm not certain how best to do this in a provider-agnostic way.

Rich-Harris:
Closes #754.

With this, we can specify a `fallback` page when using `adapter-static`. This page contains no content, but will load the SvelteKit app and invoke the router, making it suitable as a `200.html` with e.g. [Surge](https://surge.sh/help/adding-a-200-page-for-client-side-routing), or an `index.html` with [sirv](https://github.com/lukeed/sirv/tree/master/packages/sirv#optssingle) with `opts.single`.

If a fallback is specified, `adapter-static` _won't_ forcibly prerender every page, just the ones with `prerender = true`.

I added a test suite to `adapter-static`.

### Tests

### Changesets


packages/adapter-static/index.js
```diff
@@ -1,18 +1,2 @@
...
-			await utils.prerender({
-				force: true,
-				dest: pages
-			});
-		}
-	};
-
-	return adapter;
-};
+import commonjs from './index.cjs';
```
benmccann, line 1: it could be worth adding a comment here explaining why we need to use CommonJS with an ESM wrapper

benmccann (commented):
Going to make lots of people super happy! Big announcement for the Summit :smile: 

Do you think it's worth filing an issue about force disabling SSR as well? https://github.com/sveltejs/kit/issues/754#issuecomment-816396856

packages/adapter-static/package.json
```diff
@@ -3,11 +3,23 @@
 	"version": "1.0.0-next.4",
 	"scripts": {
 		"lint": "eslint --ignore-path .gitignore \"**/*.{ts,js,svelte}\" && npm run check-format",
+		"check": "tsc",
 		"format": "prettier --write . --config ../../.prettierrc --ignore-path .gitignore",
-		"check-format": "prettier --check . --config ../../.prettierrc --ignore-path .gitignore"
+		"check-format": "prettier --check . --config ../../.prettierrc --ignore-path .gitignore",
+		"test": "rm -rf test/apps/*/build && uvu test test.js"
```
SOVLOOKUP, line 9: #1183 this change can make test work in windows

packages/adapter-static/.gitignore
```diff
@@ -1,2 +1,4 @@
 .DS_Store
 node_modules
+.svelte
```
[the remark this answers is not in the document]
Rich-Harris: It's not really about gitignore, we're just using that file as a proxy for what shouldn't be linted/formatted. So until we rethink that a bit, I think we need to repeat it in packages

packages/adapter-static/index.js
Rich-Harris: I realised there's actually no benefit to this file — removed

Rich-Harris:
> Do you think it's worth filing an issue about force disabling SSR as well?

I'd rather wait for that to come up organically — I would rather we investigate alternatives (like static analysis) before providing an 'eject' button that is likely to be misused. People will just forcibly disable SSR as soon as they run into the smallest SSR problem if the option is available to them, rather than working through the problem and preserving SSR-ability where possible. That would be bad stewardship on our part.

packages/adapter-static/package.json
```diff
@@ -3,11 +3,21 @@
...
-		"typescript": "^4.2.3"
+		"playwright-chromium": "^1.10.0",
+		"port-authority": "^1.1.2",
+		"sirv": "^1.0.11",
+		"typescript": "^4.2.4"
+	},
+	"type": "module",
+	"main": "index.cjs",
+	"exports": {
+		"require": "./index.cjs"
```
Conduitry, line 21: Is this the shape this needs to take? It doesn't need to be `exports: { '.': { require: './index.cjs' } }`?

Rich-Harris: `.` is assumed at the top level

packages/adapter-static/index.cjs
```diff
@@ -1,4 +1,4 @@
-module.exports = function ({ pages = 'build', assets = 'build' } = {}) {
+module.exports = function ({ pages = 'build', assets = pages, fallback = null } = {}) {
```
benmccann, line 1: Why did this file need to be renamed to `.cjs`?

ivoreis: So it can be required by the config file https://github.com/sveltejs/kit/pull/1181/files#diff-ffa86c05d95319420b7940aecaf8b3f028dad6acf5e1ed7005b0955b01325801

btakita:
I have a requirement package an app using the `file://` protocol, meaning there is no web server. Is a SSR-less SPA not in the works? I could use Routify but would rather use Svelte Kit to have both a PWA & a file:// based SPA.

benmccann:
@btakita the issue to track that is https://github.com/sveltejs/kit/issues/907

PR #1183, opened 2021-04-23
windows `rm -rf` test fix

SOVLOOKUP:


Rich-Harris:
thank you, though I ended up taking a different path — using `npx` to install an additional dependency at test time is quite an expensive solution. we can just use node's `fs` module

PR #1000, opened 2021-04-13
Create-Svelte - Fixing sucrase issue while running build script

rafaelcamaram:
I was trying to run the `create-svelte` CLI locally but I was not able to do it with success. I realized that we got the #989 merged and, with this, we were not able to run `npm run build` in order to build the `dist` folder (and then run the CLI to set a project).

```
import { transform } from 'sucrase';
         ^^^^^^^^^
SyntaxError: The requested module 'sucrase' is expected to be of type CommonJS, which does not support named exports. CommonJS modules can be imported by importing the default export.
For example:
import pkg from 'sucrase';
const { transform } = pkg;
```

### Tests
- [X] Run the tests with `pnpm test` and lint the project with `pnpm lint`

PR #1192, opened 2021-04-23
fix(create-svelte): bump minimum required Svelte version

dummdidumm:
The Vite Svelte plugin uses the cssHash option which was added in 3.34

For some reason I can't build the templates locally, it says

```
import { transform } from 'sucrase';
         ^^^^^^^^^
SyntaxError: The requested module 'sucrase' is expected to be of type CommonJS, which does not support named exports. CommonJS modules can be imported by importing the default export.
```

So I hope I got the right place where to bump the version


Conduitry:
What version of Node are you on when you're trying to build this locally? I think later versions got better at CJS/ESM interop in areas like this. See also the comment https://github.com/sveltejs/kit/pull/1000#issuecomment-819051317 on that closed PR.

PR #1201, opened 2021-04-24
CI: Fix Windows Playwright Cache

GrygrFlzr:
The [last successful Windows CI run](https://github.com/sveltejs/kit/runs/2423924312#step:20:1) has unfortunately broken the CI by incorrectly caching nothing as the browser:

```powershell
Post job cleanup.
C:\Windows\System32\tar.exe --posix -z -cf cache.tgz -P -C D:/a/kit/kit --files-from manifest.txt
Cache Size: ~0 MB (30 B)
Cache saved successfully
Cache saved with key: pw-Windows-84213962f52cb541e4092ca888c1aa986057dc2ba0a2bed78a1292e3ee88eaa1
```

All CI runs are currently using this broken 30 byte cache for its Windows runs and promptly failing.

This PR works around this issue by changing the playwright key, because GitHub provides no way to manually evict caches and we'd have to wait 7 days with zero CI usage for it to automatically clear the cached browser.

### DIFF
--- a/.github/actions/env/action/main.js
+++ b/.github/actions/env/action/main.js
@@ -405,7 +405,7 @@ var core$1 = /*@__PURE__*/getDefaultExportFromCjs(core);
 
 const WINDOWS_CACHE_PATH = 'D:\\.pnpm-store\\v3';
 const NIX_CACHE_PATH = '~/.pnpm-store/v3';
-const WINDOWS_BROWSER_PATH = '%USERPROFILE%\\AppData\\Local\\ms-playwright';
+const WINDOWS_BROWSER_PATH = require$$1.join(require$$0.homedir(), 'AppData', 'Local', 'ms-playwright');
 const LINUX_BROWSER_PATH = '~/.cache/ms-playwright';
 const MACOS_BROWSER_PATH = '~/Library/Caches/ms-playwright';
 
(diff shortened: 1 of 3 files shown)


GrygrFlzr:
Okay apparently this run also did the same exact thing:

```
Post job cleanup.
C:\Windows\System32\tar.exe --posix -z -cf cache.tgz -P -C D:/a/kit/kit --files-from manifest.txt
Cache Size: ~0 MB (30 B)
Cache saved successfully
Cache saved with key: playwright-Windows-84213962f52cb541e4092ca888c1aa986057dc2ba0a2bed78a1292e3ee88eaa1
```

So I guess the caching we've set up is just straight up broken on Windows. I'm suspecting a Unix vs Windows slash issue.

pngwn:
This is where the location to cache is defined for the browser on windows: https://github.com/sveltejs/kit/blob/master/.github/actions/env/main.js

```
const WINDOWS_BROWSER_PATH = '%USERPROFILE%\\AppData\\Local\\ms-playwright';
```
This should be correct according to the docs, or at least was at the time, not sure why it would have broken after working successfully unless there was a change with playwright. 

I can take a look later if no one beats me to it.

GrygrFlzr:
Fixed by swapping `%USERPROFILE%` to `os.homedir()`. Maybe it was incorrectly resolving?

PR #1207, opened 2021-04-24
a11y improvements for template to-do list

geoffrich:

### Tests

### Changesets

This PR improves the accessibility of the to-do list in the starter template.

- Add label to to-do list input
- Add label to to-do list edit field
- Show the save button when focused. Before this change, the save button would be focused while navigating with the keyboard but would not be visible
- Make delete button focus state the same as hover state. Before this change, the delete button was hard to see when focused.
- ~Programmatically manage focus on delete. Since the focused element is removed from the DOM, we should explicitly focus an element since browsers can be inconsistent. I chose to focus the heading at the top of the list.~ EDIT: Removed this, it was adding too much complexity for not enough benefit.

### DIFF
--- a/packages/create-svelte/templates/default/src/routes/todos/index.svelte
+++ b/packages/create-svelte/templates/default/src/routes/todos/index.svelte
@@ -63,7 +63,7 @@
 		<input name="text" aria-label="Add todo" placeholder="+ tap to add a todo">
 	</form>
 
-	{#each todos as todo, i (todo.uid)}
+	{#each todos as todo (todo.uid)}
 		<div class="todo" class:done={todo.done} transition:scale|local={{start:0.7}} animate:flip={{duration:200}}>
 			<form action="/todos/{todo.uid}.json?_method=patch" method="post" use:enhance={{
 				pending: (data) => {
@@ -83,7 +83,7 @@
 			</form>
 
 			<form action="/todos/{todo.uid}.json?_method=delete" method="post" use:enhance={{
-				result: async () => {
+				result: () => {
 					todos = todos.filter(t => t.uid !== todo.uid);
 				}
 			}}>
--- a/packages/create-svelte/templates/default/src/routes/todos/index.svelte
+++ b/packages/create-svelte/templates/default/src/routes/todos/index.svelte
@@ -1,5 +1,4 @@
 <script context="module" lang="ts">
-	import { tick } from 'svelte';
 	import { enhance } from '$lib/form';
 	import type { Load } from '@sveltejs/kit';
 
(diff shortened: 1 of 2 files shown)


Rich-Harris:
Thank you! Can you explain the rationale for focusing the h1 after deleting a todo? I would have thought it was better to keep the focus in the same part of the document, so you can easily tab/shift-tab to the next/previous item

pngwn:
I think we should probably be focusing the next item in the list (the item that follows the deleted item).

geoffrich:
Thanks for the feedback--I originally focused the h1 because it was simplest. But I agree that it makes sense to keep focus in the same part of the document. I updated the logic to focus the next edit input if it exists, or the previous edit input if it's the last todo in the list, or the "add a todo" field if there are no todos remaining. I used `bind:this` in the `each` block so I could easily reference the next/previous input, not sure if there's a more idiomatic way.

I also had to wait a `tick` in the `handleFocus` function because the array of todo inputs wasn't updated right away in some cases and focus was being moved to an element that didn't exist.

packages/create-svelte/templates/default/src/routes/todos/index.svelte
```diff
@@ -78,12 +94,13 @@
...
 			}}>
-				<input type="text" name="text" value={todo.text}>
+				<input aria-label="Edit todo" type="text" name="text" value={todo.text} bind:this={todoInputs[i]}>
 				<button class="save" aria-label="Save todo"/>
 			</form>
 
 			<form action="/todos/{todo.uid}.json?_method=delete" method="post" use:enhance={{
-				result: () => {
+				result: async () => {
+					await manageFocus(i);
```
pngwn, line 103: Don't we want to manage focus _after_ the todo has been removed from the DOM?

Rich-Harris:
This is adding a _ton_ of complexity to something that is supposed to be an easy-to-follow demo. Isn't the current focusing behaviour okay? After deleting a todo, `Tab` will focus the toggle button in the next todo, `Shift-Tab` will focus the delete button in the previous todo, which is what I'd expect. Automatically focusing a text input that I hadn't tabbed to or otherwise selected feels surprising.

Could we just keep the `aria-label` additions and style changes and get rid of the focus stuff? Unless there's authoritative guidance (from WAI-ARIA or similar) that confirms focusing an adjacent text input is the correct behaviour. Thanks!

geoffrich:
That's fair -- I've removed the focus management behavior from the PR. It was definitely getting a little complex for a demo.

packages/create-svelte/templates/default/src/routes/todos/index.svelte
geoffrich: Removing explicit focus management entirely per Rich's comments.

Rich-Harris:
Thank you!

PR #1217, opened 2021-04-25
add install instructions and warning for netlify adapter

sw-yx:
hi! i used to work at netlify.

i figured i would better document the netlify adapter install instructions because it tripped me up as well as Rich Harris at Svelte Summit so it definitely needs some documenting

### DIFF
--- a/packages/adapter-netlify/README.md
+++ b/packages/adapter-netlify/README.md
@@ -18,15 +18,10 @@ You can then configure it inside of `svelte.config.cjs`:
 const adapter = require('@sveltejs/adapter-netlify');
 module.exports = {
 	kit: {
-		adapter: adapter(),
+		adapter: adapter(), // currently the adapter does not take any options
 		target: '#svelte'
 	}
 };
-```
-
-The Netlify adapter currently does not take any options.
-
-## Configuration
 
 This adapter expects to find a [netlify.toml](https://docs.netlify.com/configure-builds/file-based-configuration) file in the project root. It will determine where to write static assets and functions to based on the `build.publish` and `build.functions` settings, as per this sample configuration:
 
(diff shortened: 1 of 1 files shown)


packages/adapter-netlify/README.md
```diff
@@ -4,6 +4,28 @@ Adapter for Svelte apps that creates a Netlify app, using a function for dynamic
...
+	kit: {
+		adapter: adapter(),
+		target: '#svelte'
+	}
+};
+```
+
+The Netlify adapter currently does not take any options.
+
 ## Configuration
```
sw-yx, line 29:
```suggestion
		adapter: adapter(), // currently the adapter does not take any options
		target: '#svelte'
	}
};
```

Then, make sure you have a `netlify.toml` file as below.

## Configuration
```

**Comment by Conduitry** [comment]
All of the adapters and other packages are currently set up so that you should be using the next tag. I think these should all be so documented in one place. Currently that's in the FAQ but it could be there's a better place.

**Comment by sw-yx** [comment]
ya. just adding it where i normally look for this warning (http://npm.im/@sveltejs/adapter-netlify)

**Comment by Rich-Harris** [comment]
thanks — just curious about the inclusion of the `prerender` options, which mostly match the defaults but add `force: true` which is generally unrecommended — is this intentional? If so `crawl`, `enabled` and `pages` could be usefully omitted i think

**Comment by sw-yx** [comment]
no, not intentional - was copied from another template. i'll remove

File: packages/adapter-netlify/README.md [hunk 2]
```diff
@@ -4,9 +4,33 @@ Adapter for Svelte apps that creates a Netlify app, using a function for dynamic
...
+module.exports = {
+	kit: {
+		adapter: adapter(), // currently the adapter does not take any options
+		target: '#svelte',
+		prerender: { // recommend prerendering roots for Netlify, see https://kit.svelte.dev/docs#ssr-and-javascript-prerender
+			crawl: true,
+			enabled: true,
+			force: true,
+			pages: ['*'],
+		},
```
sw-yx, line 28:
```suggestion
```

PR #1224, opened 2021-04-26
Update Vite to 2.2.2

GrygrFlzr:
~~Make sure our tests succeed on Vite 2.2.3~~

They don't, so make sure our tests succeed on Vite 2.2.2.


GrygrFlzr:
```
 FAIL  dev:basics  "disables router if router=false [js]"
```
This seems consistent across multiple runs between the CI and my local copy, so there might be an actual problem here?

benmccann:
This PR only updated one of the templates. There are two templates, so you need to copy/paste every change to the templates across both of them: https://github.com/sveltejs/kit/blob/master/packages/create-svelte/templates/skeleton/package.template.json

Issue #639, opened 2021-03-24
Apps for each of the adapters


Rich-Harris:
The adapters don't have tests, and it's not totally clear what value they'd provide if they did, but I think we _should_ have hello world apps for each of the officially supported adapters running on the respective platforms.

Should they live in this repo? In `examples` or a new directory? Or in the package directories themselves? Or should they have their own individual repos?

Most importantly, who feels like setting it up? 😀

antony:
I think under `examples` is fine - `examples/begin-adapter` perhaps.

We could deploy each via API and ensure that the relevant CLI tool returns ok and not an error. We could also possibly ping the deployed endpoint looking for a 200 or similar. Optionally a very quick playwright test run against the deployed app. It'd be a pretty reasonable test.

benmccann:
I think most of the adapters should be in separate repos. Keeping the most used ones here is fine as long as we're going to be maintaining them. I don't know if we want to be responsible for an influx of community adapters though

For the examples, should we have separate standalone apps or can we just repurpose some of the existing ones. E.g. deploy the HN to Netlify and the Realworld example to Vercel

antony:
I'm not suggesting we should adopt all community adapters at all - I think we're talking about the basic set of us-maintained ones here. It makes sense for the example apps to live in the monorepo, so that they aren't a maintenance burden.

I think it should be the same app across all adapters - that way we can make checks for consistency, and spot misbehaving adapters early.

Rich-Harris:
Yeah, I was thinking a super basic hello world that has page with a dynamic parameter, an endpoint with a dynamic parameter, a handful of static files, something prerenderable, maybe a `setup` function to make sure things work in an env-agnostic way, etc. `svelte-kit-demo` is probably pretty close already. Maybe we could even have a single codebase, and we could deploy it by swapping out the adapter at build time according to whatever env var is currently applicable

benmccann:
Maybe we could use demo or sandbox app

The other part of this is detecting when it breaks. If we have an app for each adapter we need someone to periodically visit those apps in order to detect issues. We could use a downtime monitoring service like Pingdom

benmccann:
Although I guess if the deploy just fails there wouldn't be downtime - we'd get an old version instead. Maybe we need to do a GitHub Actions redeploy on each commit and see when it fails

Rich-Harris:
[...]

PR #1231, opened 2021-04-26
add yargs dep to allow programmatic prompt answering

afreidz:

### Tests

### Changesets

### Problem
Running `npm init svelte@next` requires user interaction to a series of prompts.  If a package wishes to leverage this scaffolding command, it would need to expose these prompts upstream.  In some cases, it may be preferable to pass the options to the command instead of using an interactive prompt.  Example: I plan to make a yeoman generator for use within our company to create scaffold sveltekit apps.  We have decided that TS,eslint,prettier are all required and not optional.  I wish not to burden our consumers with having to answer the prompts in ways that match our suggestions.

### Solution
Leveraging [yargs](https://www.npmjs.com/package/yargs) it is fairly trivial to parse CLI flags passed into `bin.js` and use those to values in place of the prompts.  This allows programmatic completion of the `npm init svelte@next` command.  In our use case, this will allow us to hard code the answers and not burden our consumers with answering them.

### DIFF
--- a/.changeset/short-mails-invent.md
+++ b/.changeset/short-mails-invent.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Optionally allow passing values to the command to auto-answer prompts
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -5,6 +5,8 @@ import { fileURLToPath } from 'url';
 import { bold, cyan, gray, green, red } from 'kleur/colors';
 import prompts from 'prompts';
 import { mkdirp, copy } from './utils.js';
+import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
 
 // prettier-ignore
 const disclaimer = `
@@ -21,18 +23,20 @@ async function main() {
 	console.log(gray(`\ncreate-svelte version ${version}`));
 	console.log(disclaimer);
 
+	const { argv } = yargs(hideBin(process.argv));
 	const cwd = process.argv[2] || '.';
+	prompts.override(argv);
 
 	if (fs.existsSync(cwd)) {
 		if (fs.readdirSync(cwd).length > 0) {
 			const response = await prompts({
 				type: 'confirm',
-				name: 'value',
+				name: 'overwrite',
 				message: 'Directory not empty. Continue?',
 				initial: false
 			});
 
-			if (!response.value) {
+			if (!response.overwrite) {
 				process.exit(1);
 			}
 		}
--- a/packages/create-svelte/package.json
+++ b/packages/create-svelte/package.json
@@ -4,13 +4,15 @@
 	"bin": "./bin.js",
 	"dependencies": {
 		"kleur": "^4.1.4",
-		"prompts": "^2.4.0"
+		"prompts": "^2.4.0",
+		"yargs": "^16.2.0"
 	},
 	"devDependencies": {
 		"@sveltejs/kit": "workspace:*",
 		"@types/gitignore-parser": "^0.0.0",
 		"@types/prettier": "^2.2.3",
 		"@types/prompts": "^2.0.10",
+		"@types/yargs": "^16.0.1",
 		"gitignore-parser": "^0.0.2",
 		"prettier": "^2.2.1",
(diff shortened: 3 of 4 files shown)


babichjacob:
👍 to the concept for sure (not tested the implementation yet). This will be useful for `svelte-add`'s initializer to wrap `create-svelte`'s (hinted at [here](https://github.com/orgs/svelte-add/projects/1#card-59592227) and [here](https://github.com/orgs/svelte-add/projects/1#card-59592260)).

Conduitry:
If we make this more programmatic, I'm concerned about that implying an API that we won't want to stick with. E.g., if we add a new question that someone hasn't yet added an option for in their script, is this a breaking change because running create-svelte now expects user input and won't just run automatically. (The alternative would be to make it explicitly fail if some but not all options were specified on the command line, which would also be a breaking change.) If this tool always requires human interaction, we don't need to worry about having an API because we can always assume a human is reading the interactive prompts.

afreidz:
I pulled up the docs for [prompts](https://www.npmjs.com/package/prompts) which is leveraged in `create-svelte` and under the "override" section (damn them for not deep linking :smile: ) they had this example:

babichjacob:
> 
> 
> > If we make this more programmatic, I'm concerned about that implying an API that we won't want to stick with. E.g., if we add a new question that someone hasn't yet added an option for in their script, is this a breaking change because running create-svelte now expects user input and won't just run automatically. (The alternative would be to make it explicitly fail if some but not all options were specified on the command line, which would also be a breaking change.) If this tool always requires human interaction, we don't need to worry about having an API because we can always assume a human is reading the interactive prompts.
> 
> i believe the impl is that any questions not passed a matching CLI param would still be prompted, rather than breaking. so it might be a little less risky

This is a different kind of break in that a previously non-interactive process requires interactivity.

afreidz:
i think the issue lies in decided whether `create-svelte` aims to be the sole entry point into sveltekit.  or if there is an appetite to have it be consumed by higher order systems.  i dont think there is a great solution to support user<>system AND system<>system interactions flawlessly.  but i think i would aim to break system<>system patterns over user<>system.  i think it would be a fair ask for someone maintaining a system that leverages `create-svelte` to make regular updates when things break.

afreidz:
the alternative solution for me is to snapshot the skeleton app and leverage that in my generator.  which additionally causes a maintenance point for me to keep in sync with `create-svelte`.  unless svelte were able to maintain a "kitchen sink" sveltekit repo that could be leveraged by `create-svelte` and/or external systems?  i wouldn't mind pruning things from a "yes to all" version of the app based on our desired config.

afreidz:
anymore thoughts on this or is it effectively NACK'd?

babichjacob:
> 
> 
> i think the issue lies in deciding whether `create-svelte` aims to be the sole entry point into sveltekit. or if there is an appetite to have it be consumed by higher order systems. i dont think there is a great solution to support user<>system AND system<>system interactions flawlessly. but i think i would aim to break system<>system patterns over user<>system. i think it would be a fair ask for someone maintaining a system that leverages `create-svelte` to make regular updates when things break.

To be extremely vague, I can imagine `create-svelte` being broken into a CLI entrypoint and a file that has the actual functionality that the CLI file calls, which lets external sources wrap the `create-svelte` process without needing to add argument parsing (this PR).

Conduitry:
Yeah I would definitely favor a purely interactive interface (what we have now) combined with a purely programmatic interface (which would be called from Node) rather than a hybrid CLI-with-command-line-arguments interface. It makes it very clear whether the intended behavior is for it to be interactive or not. And, if we have sensible defaults for all of the options, it lets us add further options without it being a breaking change and without surprising anyone about their automated script suddenly requiring manual interaction.

For the time being, I would like to reserve the right to make breaking changes at any time (until Kit reaches 1.0), but after that I think it's reasonable for the as-yet-unwritten Node API form of create-svelte to follow semver, but not for the interactive CLI version of it to.

jthegedus:
Coming from #639 - What about each option having a default and defaults being used to answer the questions when a single `--defaults` flag is present. In my use case I don't really care what the defaults are, I just want to init the template non-interactively. Alternative methods require a lot more work on my end, more work than re-raising this old discussion ;)

PR #1235, opened 2021-04-26
fix: remove ?import from svelte component id and update to vite 2.2.3 (fixes #1227)

### LINKED ISSUE #1227: Compatibility with latest version of Vite
@GrygrFlzr reported issues with Vite 2.2.3 in https://github.com/sveltejs/kit/pull/1224. The Vite version is currently locked to 2.2.2 in the template. We should fix the error and add back the `^` to the version specifier in the template

dominikg:

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/templates/default/package.json
+++ b/packages/create-svelte/templates/default/package.json
@@ -14,7 +14,7 @@
 		"svelte": "^3.37.0",
 		"svelte-preprocess": "^4.7.1",
 		"typescript": "^4.2.4",
-		"vite": "2.2.2"
+		"vite": "^2.2.3"
 	},
 	"type": "module",
 	"dependencies": {
--- a/packages/kit/src/core/dev/index.js
+++ b/packages/kit/src/core/dev/index.js
@@ -213,7 +213,8 @@ class Watcher extends EventEmitter {
 									}
 								}
 								let entry = `/${id}`;
-								if (!entry.endsWith('.svelte')) { // TODO utility method?
+								if (!entry.endsWith('.svelte')) {
+									// TODO utility method?
 									entry += '?import';
 								}
 								return {
--- a/packages/kit/src/core/dev/index.js
+++ b/packages/kit/src/core/dev/index.js
@@ -212,10 +212,13 @@ class Watcher extends EventEmitter {
 										}
 									}
 								}
-
+								let entry = `/${id}`;
+								if (!entry.endsWith('.svelte')) { // TODO utility method?
+									entry += '?import';
+								}
 								return {
 									module,
-									entry: `/${id}?import`,
+									entry,
 									css: [],
 									js: [],
 									styles: Array.from(styles)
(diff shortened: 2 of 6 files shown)


pnpm-lock.yaml
```diff
@@ -648,7 +648,7 @@ packages:
       rollup: 2.41.1
     dev: false
 
-  /@sveltejs/vite-plugin-svelte/1.0.0-next.7_301da3dca3251a36e99a1622a4d7ca38:
+  /@sveltejs/vite-plugin-svelte/1.0.0-next.7_6c43afb327f3ef976774a1a1e8559c8a:
```
dominikg, line 651: no clue what that is about, i'm using pnpm 6.2.1, but that hash should not have changed? wonder if its a local cache thing.

packages/kit/src/core/dev/index.js
```diff
@@ -212,10 +212,14 @@ class Watcher extends EventEmitter {
 										}
 									}
 								}
-
+								let entry = `/${id}`;
+								if (!entry.endsWith('.svelte')) {
+									// TODO utility method?
```
benmccann, line 217: I'm not sure the TODO is really necessary
```suggestion
```

benmccann:
Thanks!!

I think we'd also want to upgrade the `package.json` in both the templates: https://github.com/sveltejs/kit/tree/master/packages/create-svelte/templates

packages/kit/src/core/dev/index.js
dominikg: it was meant as a marker to start a discussion about a more generic solution to the "do we need to append ?import" problem. I'll try to come up with one.

dominikg: removed the todo for now. without additional information from or to vite i don't see a better way. Other than reverting the change in vite that enabled importing .svelte without ?import.

packages/create-svelte/templates/default/package.json
```diff
@@ -14,7 +14,7 @@
 		"svelte": "^3.37.0",
 		"svelte-preprocess": "^4.7.1",
 		"typescript": "^4.2.4",
-		"vite": "2.2.2"
+		"vite": "^2.2.3"
```
benmccann, line 17: I think you need to update https://github.com/sveltejs/kit/blob/master/packages/create-svelte/templates/skeleton/package.template.json as well

Conduitry:
Should this instead be using `config.extensions` https://kit.svelte.dev/docs#configuration to determine which files to exclude `?import` on?

benmccann:
That's a good question. I think if it were going to then we'd also need to update Vite: https://github.com/vitejs/vite/pull/3128

dominikg:
yes, excluding all configured extensions requires a change in vite. Plugins need to be able to configure extensions, i'll draft something to propose in vite.

dominikg:
started a PR on vite that would allow configuring the list of extensions in vite https://github.com/vitejs/vite/pull/3219

Issue #1221, opened 2021-04-26
Document global fetch


benmccann:
You can now call `fetch` directly from an endpoint (https://github.com/sveltejs/kit/pull/1066). This comes up a lot as a question, so we should document it

PR #1237, opened 2021-04-26
Remove Axios FAQ, add Firebase FAQ, and update preprocess and env FAQs

benmccann:
Axios works now, so no longer need to mention it. The vast majority of Vite issues have been fixed now between removing `noExternal` from the template and a couple bug fixes in recent releases of Vite

### DIFF
--- a/documentation/faq/80-integrations.md
+++ b/documentation/faq/80-integrations.md
@@ -6,11 +6,15 @@ question: How do I use X with SvelteKit?
 
 Please see [sveltejs/integrations](https://github.com/sveltejs/integrations#sveltekit) for examples of using many popular libraries like Tailwind, PostCSS, Firebase, GraphQL, mdsvex, and more.
 
-### How do I use Babel, CoffeeScript, Less, PostCSS / SugarSS, Pug, scss/sass, Stylus, TypeScript, `global` styles, or replace?
+### How do I use `svelte-preprocess`?
 
-Adding [svelte-preprocess](https://github.com/sveltejs/svelte-preprocess) to your [`svelte.config.cjs`](#configuration) is the first step. For many of the tools listed above, you only need to install the corresponding library such as `npm install -D sass`or `npm install -D less`. See the svelte-preprocess docs for full details.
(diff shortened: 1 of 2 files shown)


babichjacob:
How do we teach that `fetch` is globally available (now)?

benmccann:
Yeah, we need to add that to the docs. I wasn't sure the best way to do it, but had filed https://github.com/sveltejs/kit/issues/1221 to make sure we don't forget about it

babichjacob:
> 
> 
> Yeah, we need to add that to the docs. I wasn't sure the best way to do it, but had filed #1221 to make sure we don't forget about it

Missed that because it somehow didn't come up when I searched "fetch" or "global fetch" -- thanks!

documentation/faq/80-integrations.md
```diff
@@ -6,11 +6,11 @@ question: How do I use X with SvelteKit?
...
 
-### How do I use Babel, CoffeeScript, Less, PostCSS / SugarSS, Pug, scss/sass, Stylus, TypeScript, `global` styles, or replace?
+### How do I use `svelte-preprocess`?
 
-Adding [svelte-preprocess](https://github.com/sveltejs/svelte-preprocess) to your [`svelte.config.cjs`](#configuration) is the first step. For many of the tools listed above, you only need to install the corresponding library such as `npm install -D sass`or `npm install -D less`. See the svelte-preprocess docs for full details.
+`svelte-preprocess` provides support for Babel, CoffeeScript, Less, PostCSS / SugarSS, Pug, scss/sass, Stylus, TypeScript, `global` styles, and replace. Adding [svelte-preprocess](https://github.com/sveltejs/svelte-preprocess) to your [`svelte.config.cjs`](#configuration) is the first step and is provided by the the default template. For many of the tools listed above, you only need to install the corresponding library such as `npm install -D sass`or `npm install -D less`. See the [svelte-preprocess](https://github.com/sveltejs/svelte-preprocess) docs for full details.
```
GrygrFlzr, line 11: > provided by the default template

Not quite, we only include it on the TypeScript versions of the templates. Should we be including it in the JS versions for consistency and ease of adding integrations?

Conduitry: I'd vote no. I don't think it should be included in templates that don't use it.

babichjacob (approved):

PR #1238, opened 2021-04-27
fix : #1228 Invalid character

### LINKED ISSUE #1228: [ERR_INVALID_CHAR]: Invalid character in header content ["location"]
Initialization only ：<https://codesandbox.io/s/lingering-worker-bn37n>

Resulting in the wrong url : <https://bn37n-3000.sse.codesandbox.io/中文/>
 
Related Code :
https://github.com/sveltejs/kit/blob/2e1b8c6fe8b96f56751b88d4374055d359788e03/packages/kit/src/runtime/server/index.js#L18

image

```log
TypeError [ERR_INVALID_CHAR]: Invalid character in header content ["location"]
    at ServerResponse.setHeader (_http_outgoing.js:561:3)
    at ServerResponse.writeHead (_http_server.js:289:21)
    at Immediate.<anonymous> (file:///sandbox/node_modules/@sveltejs/kit/dist/chunks/index.js:3429:11)
```

2234839:
Fixes #1228 

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -15,7 +15,7 @@ export async function respond(incoming, options, state = {}) {
 		return {
 			status: 301,
 			headers: {
-				location: encodeURI(incoming.path.slice(0, -1)) + (q ? `?${q}` : ''))
+				location: encodeURI((incoming.path.slice(0, -1)) + (q ? `?${q}` : ''))
 			}
 		};
 	}
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -46,13 +46,7 @@ export async function respond(incoming, options, state = {}) {
 						page: null
 					});
 				}
-				if (request.path.includes('not_ascii')) {
-					console.log(
-						'[options.manifest.routes]',
-						request.path,
-						options.manifest.routes.map((el) => el.pattern)
-					);
-				}
+
 				for (const route of options.manifest.routes) {
 					if (!route.pattern.test(request.path)) continue;
 
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -15,7 +15,7 @@ export async function respond(incoming, options, state = {}) {
 		return {
 			status: 301,
 			headers: {
-				location: encodeURI((incoming.path.slice(0, -1)) + (q ? `?${q}` : ''))
+				location: encodeURI(incoming.path.slice(0, -1) + (q ? `?${q}` : ''))
 			}
 		};
 	}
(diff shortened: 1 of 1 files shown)


packages/kit/src/runtime/server/index.js
```diff
@@ -15,7 +15,7 @@ export async function respond(incoming, options, state = {}) {
 		return {
 			status: 301,
 			headers: {
-				location: incoming.path.slice(0, -1) + (q ? `?${q}` : '')
+				location: encodeURI(incoming.path.slice(0, -1)) + (q ? `?${q}` : ''))
```
benmccann, line 18: this code is invalid. you added one start paren and two closing parens

benmccann:
https://stackoverflow.com/questions/47687379/what-characters-are-allowed-in-http-header-values does suggest URI percent encoding, so this seems like a reasonable fix

2234839:
> 
> 
> https://stackoverflow.com/questions/47687379/what-characters-are-allowed-in-http-header-values does suggest URI percent encoding, so this seems like a reasonable fix

If you use encodeURIComponent for the path `/中文/` you will get `location:"%2F%E4%B8%AD%E6%96%87%2F"` In my attempt the browser does not handle it correctly, so `/` should not be escaped

Rich-Harris:
Thank you — could you add a test and a changeset please? For example you could add a test to [this file](https://github.com/sveltejs/kit/blob/master/packages/kit/test/apps/basics/src/routes/routing/__tests__.js) that verifies [this page](https://github.com/sveltejs/kit/blob/master/packages/kit/test/apps/basics/src/routes/routing/%5Bslug%5D.svelte) renders as expected if you visit e.g. `/routing/中文`

2234839:
> 
> 
> Thank you — could you add a test and a changeset please? For example you could add a test to [this file](https://github.com/sveltejs/kit/blob/master/packages/kit/test/apps/basics/src/routes/routing/__tests__.js) that verifies [this page](https://github.com/sveltejs/kit/blob/master/packages/kit/test/apps/basics/src/routes/routing/%5Bslug%5D.svelte) renders as expected if you visit e.g. `/routing/中文`

I found a new problem when trying to add test cases,
Using a file structure like `/routing/nonascii_path/中文.svelte`, accessing `/routing/non_ascii_path/中文` 

https://github.com/sveltejs/kit/blob/52bef766133edaeae8be76d8ce689ca048bf8dc7/packages/kit/src/runtime/server/index.js#L51

The pattern here is something like `/^\/routing\/non_ascii_path\/%E4%B8%AD%E6%96%87\/?$/` , causing it to fail to match request path:`/routing/non_ascii_path/中文`, but the logic that triggers the redirect later matches request.path:`/routing/non_ascii_path/%E4%B8%AD%E6%96%87` 

I'm not sure what to do to fix this, add `decodeURI` to the match or add support for non-ascii paths where `options.manifest.routes` is generated, or something else

benmccann:
`pnpm lint` is failing. You need to run `pnpm format`

I'm not sure what's encoding the pattern, but it sounds to me like it should be removed and that `options.manifest.routes` should support non-ascii paths

2234839:
> 
> 
> `pnpm lint` is failing. You need to run `pnpm format`

Okay, it's done. I thought it was a simple change and didn't download it locally for these steps

Rich-Harris:
closing this in favour of #1273, which includes these commits but also handles other related issues (and adds tests)

Issue #965, opened 2021-04-11
Build finish successfully but get stuck and not exit


TedMeftah:
**Describe the bug**
When running `yarn build` the build finishes successfully but does not exit. This cause an issue with automating builds. 

**To Reproduce**
I have boiled down the issue to having `setInterval` in a script.

```
<script>
  let time = 1;
  setInterval(() => {
    time += 1;
  }, 1000);
</script>

{time}
```
A solution to this is of course to use `onMount`

```
<script>
  import { onMount } from "svelte";

  let time = 1;
  onMount(() => {
    setInterval(() => {
      time += 1;
    }, 1000);
  })
</script>

{time}
```
**Expected behavior**
I'm not sure if this is expected behavior or not but, In `Sapper` this did not happen, the first snippet of code run just fine.

**Severity**
Not severe since there is a workaround but I think it maybe need to be documented.

Conduitry:
Interesting. I'm not entirely sure why it's causing a hang in SvelteKit but not in Sapper. But, regardless, having a `setInterval` without a `clearInterval` is a bad idea, because of memory leaks. And that's especially a problem on the server, where every single time someone loads a page that contains that component, it creates another `setInterval`, which then hangs around forever.

I think the path forward here is not to make it not hang, but to do something to warn people that this is a bad thing to be doing. In SSR, there's not really a reason to be creating an interval (or a timeout) at all, because it's not going to fire until after the SSR'd HTML is rendered. I don't know what's an effective way to warn about this. Static analysis won't really help. Do we override these globals with our own versions, and display a warning if we detect that one is being run during the synchronous period that the server-side rendering is happening? I'm not sure.

KubaJastrz:
I've also noticed this today. It only hangs when using `@sveltejs/adapter-static`, `@sveltejs/adapter-node` works fine.

In both cases, the `build` folder gets populated with correct data though 🤔 

I'll try to create a reproduction soon.

matt3224:
![Screenshot 2021-04-16 at 10 30 11 pm](https://user-images.githubusercontent.com/2401925/115085965-4f5ac880-9f03-11eb-8234-f0838dc1e3b7.png)

I also run into this, when using `@sveltejs/adapter-static` running `pnpm build` runs fine but never exits stays on the above screen ^^

matt3224:
this works with the new structure:
```js
adapter: {
   adapt: static
},
```

KubaJastrz:
[...]

PR #1243, opened 2021-04-27
Docs: Global `fetch`

### LINKED ISSUE #618: `fetch` in endpoints

I'm hosting a blog post on `/blog/[slug].svelte`. This blog post is powered by `/blog/[slug].json`, generated by an endpoint. In the endpoint we want to include some html which serves as the body of the blog post. This html is composed by markdown, and its location is defined by an `articleUrl` parameter.

The final output for `/blog/[slug].json` might look something like this:

```json
{
  "slug": "my-article",
  "articleUrl": "/articles/my-article.md",
  "articleHtml": "<h1>My article</h1>\n<p>This is my article</p>"
}
```

Making relative requests to fetch additional content, let's say `/articles/my-article.md`, is awkward if you use something like `node-fetch`. You have to make assumptions about the hosted environment, ergo:

```js
import fetch from "node-fetch";

export async function get(request) {

  (...)

  const url = new URL(project.articleUrl, "http://localhost:3000");
  const res = await fetch(url);
  const markdown = await res.text();

  project.articleHtml = renderMarkdown(markdown);

  return {
    headers: {
      'Content-Type': 'application/json'
    },
    body: project
  }
}
```

If `fetch` was available in endpoints, similar to how they are implemented in `load()` for pages, this could easily be solved like:

```js
export async function get(request, fetch) {

  (...)

  const res = await fetch(project.articleUrl);
  const markdown = await res.text();

  project.articleHtml = renderMarkdown(markdown);

  return {
    headers: {
      'Content-Type': 'application/json'
    },
    body: project
  }
}
```

`node-fetch` would normally be an option, but since we are dealing a relative path to a locally hosted file there are the shortcomings elaborated on above.

It's not really obvious how to solve this particular issue without this feature, so it could help prevent some headaches in the future if there was parity with `load()`.

### LINKED ISSUE #1221: Document global fetch

GrygrFlzr:
- Closes #618
- Closes #1221

One additional behavior that might not be clear here is that the global fetch is truly available *globally*. This means pages also have access to them.

```svelte
<script context="module">
  export async function load({ fetch: kitFetch }) {
    // fetch is still globally available
    // if you need to opt out of the server-side inlining
    // ...
  }
</script>
<script>
  // fetch available on both server and client, so cross-fetch isn't needed
  // this is extra convenient e.g. if you're using a GraphQL client
</script>
```

Not sure where to put this, or if it should be left out to encourage the usage of the special `fetch` from `load` instead.

### DIFF
--- a/documentation/docs/01-routing.md
+++ b/documentation/docs/01-routing.md
@@ -43,7 +43,9 @@ Dynamic parameters are encoded using `[brackets]`. For example, a blog post migh
 
 ### Endpoints
 
-Endpoints are modules written in `.js` (or `.ts`) files that export functions corresponding to HTTP methods. For example, our hypothetical blog page, `/blog/cool-article`, might request data from `/blog/cool-article.json`, which could be represented by a `src/routes/blog/[slug].json.js` endpoint:
+Endpoints are modules written in `.js` (or `.ts`) files that export functions corresponding to HTTP methods. All endpoints have access to `fetch` globally without needing to import an external dependency.
+
+For example, our hypothetical blog page, `/blog/cool-article`, might request data from `/blog/cool-article.json`, which could be represented by a `src/routes/blog/[slug].json.js` endpoint:
 
 ```ts
 type Request<Context = any> = {
@@ -80,11 +82,16 @@ export async function get({ params }) {
 	// is called [slug].json.js
 	const { slug } = params;
 
+	// global fetch without imports
+	const response = await fetch('/my-api');
+	const apiData = await response.json();
+
 	const article = await db.get(slug);
 
 	if (article) {
 		return {
 			body: {
+				apiData,
 				article
 			}
 		};
--- a/documentation/migrating/05-endpoints.md
+++ b/documentation/migrating/05-endpoints.md
@@ -5,3 +5,5 @@ title: Endpoints
 In Sapper, 'server routes' — now referred to as [endpoints](/docs#routing-endpoints) — received the `req` and `res` objects exposed by Node's `http` module (or the augmented versions provided by frameworks like Polka and Express).
 
 SvelteKit is designed to be agnostic as to where the app is running — it could be running on a Node server, but could equally be running on a serverless platform or in a Cloudflare Worker. For that reason, you no longer interact directly with `req` and `res`. Your endpoints will need to be updated to match the new signature.
+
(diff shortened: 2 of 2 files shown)


Conduitry:
Kicking off any sort of async thing during SSR should be discouraged (whether it's a fetch or timeout or something else) because it'll just use up resources and won't affect the rendered component. I don't know whether this is a good opportunity to mention this.

Conduitry:
See also #965. We could probably be doing more in dev mode by overriding some common globals that are bad to run during synchronous SSR and switch them on and off as appropriate when entering SSR code.

benmccann:
> Kicking off any sort of async thing during SSR should be discouraged (whether it's a fetch or timeout or something else) because it'll just use up resources and won't affect the rendered component.

What? Why is that? I'd certainly expect it to affect the rendered component

documentation/docs/01-routing.md
```diff
@@ -43,7 +43,9 @@ Dynamic parameters are encoded using `[brackets]`. For example, a blog post migh
 
 ### Endpoints
 
-Endpoints are modules written in `.js` (or `.ts`) files that export functions corresponding to HTTP methods. For example, our hypothetical blog page, `/blog/cool-article`, might request data from `/blog/cool-article.json`, which could be represented by a `src/routes/blog/[slug].json.js` endpoint:
+Endpoints are modules written in `.js` (or `.ts`) files that export functions corresponding to HTTP methods. All endpoints have access to `fetch` globally without needing to import an external dependency.
```
benmccann, line 46: It's not just endpoints with access to `fetch`, but the whole application, right? Hooks are the other place I could possibly see using it (e.g. to send metrics to an external server)

Conduitry:
Any async stuff during `load` is of course fine, but once you've determined the props to pass to the component and told it to return HTML, that's all synchronous.

Rich-Harris:
I'm not sure about these changes to be honest — we're cramming an awful lot into that `blog/[slug].json.js` code sample. Adding a random `fetch` that's unrelated to the surrounding prose reduces clarity, I think.

What if we just mentioned the `fetch` stuff in an aside? I've opened an alternative PR: #1252

GrygrFlzr:
Closing in favor of #1252

Issue #568, opened 2021-03-18
Minify prerendered HTML (and app.html?)


Rich-Harris:
`sapper export` minifies the generated HTML. Perhaps we should do the same thing here?

```js
// svelte.config.cjs
module.exports = {
  kit: {
    prerender: {
      minify: true
    }
  }
};
```

```js
// svelte.config.cjs
module.exports = {
  kit: {
    prerender: {
      minify: {
        // manually specify options (https://www.npmjs.com/package/html-minifier)
        collapseBooleanAttributes: true
        // ...
      }
    }
  }
};
```

We could potentially do the same thing for `src/app.html`, though the benefits are less pronounced.

benmccann:
I wonder if this config should live elsewhere because presumably you could minify the html on a non prerendered page

Rich-Harris:
It's not cheap, you probably wouldn't want to do it on the fly. But yeah, if it lived in a separate place (`kit.htmlMinifierOptions`?) then it could at least apply to `src/app.html` as well as prerendered pages

Conduitry:
Being able to control the minification options (or disabling it entirely) sounds good, as this is one of the requested Sapper features that we deferred on because of a lack of a sane way to handle configuration.

I have mixed feelings about whether this should be enabled by default for prerendered sites. It's probably a good idea, but then it's also been the subject of [a](https://github.com/sveltejs/sapper/issues/1061) [whole](https://github.com/sveltejs/sapper/issues/1161) [lot](https://github.com/sveltejs/sapper/issues/1171) [of](https://github.com/sveltejs/sapper/issues/1455) [invalid](https://github.com/sveltejs/sapper/issues/1663) [Sapper](https://github.com/sveltejs/sapper/issues/1702) [bugs](https://github.com/sveltejs/sapper/issues/1747).

I agree that it's probably only a good idea to include this on HTML that can be minified during build time and not at runtime.

benmccann:
I thought about it being expensive, but it might still be worth it for `public` cached pages (i.e. no `session` and `maxage`)

Conduitry:
As much as people should probably be deploying things behind a CDN, it sounds like a dangerous precedent to have default settings that will give people a bad time unless they're behind a CDN.

benmccann:
Agreed. I wouldn't turn it on by default

benmccann:
I wonder if people could implement this on their own using hooks now? Perhaps that's enough to close this issue?

Rich-Harris:
[...]

PR #1246, opened 2021-04-27
Replace outdated TODO with migration docs

benmccann:
I think we decided in https://github.com/sveltejs/kit/issues/568 that we're not going to do this

### DIFF
--- a/documentation/migrating/99-integrations.md
+++ b/documentation/migrating/99-integrations.md
@@ -2,4 +2,42 @@
 title: Integrations
 ---
 
-See [sveltejs/integrations](https://github.com/sveltejs/integrations#sveltekit) for examples of setting up popular tools like PostCSS, Tailwind CSS, mdsvex, Firebase, GraphQL, and Bulma.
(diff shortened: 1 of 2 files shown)


Conduitry:
Telling people to manually do this during the `handle` hook makes sense I think, but this might be good to document somewhere as part of the work of removing the TODO. Maybe in an FAQ? in a migration guide?

Separately, is there a way for code in `handle` to know whether pre-rendering is currently happening or whether we are at runtime? (We could maybe do something with env vars, but that sounds ugly.) The idea being that people probably don't want to be minifying HTML on the fly at runtime, just at build time.

Rich-Harris:
> is there a way for code in `handle` to know whether pre-rendering is currently happening or whether we are at runtime?

Yep: https://kit.svelte.dev/docs#modules-$app-env

```js
import { prerendering } from '$app/env';

export async function handle({ request, render }) {
  const response = await render(request);

  if (prerendering && response.headers['content-type'] === 'text/html') {
    response.body = minify_html(response.body);
  }

  return response;
}
```

PR #1264, opened 2021-04-28
Link session sections in docs

benmccann:
A couple people on Discord mentioned having difficultly putting the pieces together here and that it would be helpful to reference the section describing how to use the result of `getSession`

### DIFF
--- a/documentation/docs/04-hooks.md
+++ b/documentation/docs/04-hooks.md
@@ -43,7 +43,7 @@ export async function getContext({ headers }) {
 
 ### getSession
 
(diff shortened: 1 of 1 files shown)


Conduitry:
I'm wondering whether it's worth preserving some sort of note here (perhaps in a separate sentence) that the return value from here should only contain things that it's safe for the client to know. I realize that with this it still does tell you that this is accessible on the client, but it's hard to be too careful about explaining the ramifications of this.

Conduitry (approved):

PR #1267, opened 2021-04-28
disable FLoC

### LINKED ISSUE #1263: Disable FLoC by default
[FLoC](https://github.com/WICG/floc) [is](https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea) [bad](https://make.wordpress.org/core/2021/04/18/proposal-treat-floc-as-a-security-concern/), and I think we should join the [many](https://www.theverge.com/2021/4/16/22387492/google-floc-ad-tech-privacy-browsers-brave-vivaldi-edge-mozilla-chrome-safari) [organisations](https://github.blog/changelog/2021-04-27-github-pages-permissions-policy-interest-cohort-header-added-to-all-pages-sites/) that treat it as a security/privacy threat and disable it by default.

Add a new `kit.floc` config option that defaults to `false`, and which must be set to `true` to opt in to FLoC:

```js
module.exports = {
  kit: {
    floc: true
  }
};
```

If `false`, the following header gets added to every response:

```
Permissions-Policy: interest-cohort=()
```

* Doing nothing and leaving it to app developers. This places a significant burden on them; while the userland implementation is straightforward, we can't reasonably expect most people to be aware of this issue
* Making it opt out rather than opt in. This reduces the implementation burden, but not the awareness burden

SvelteKit is, among other things, a statement about what kind of web we want to exist. I think it's entirely appropriate to treat this as being within the framework's purview, as much as providing tools to mitigate against things like CSRF attacks.

In practical terms, not very. In terms of principle, I think it's important for the reasons given above.

Rich-Harris:
Closes #1263. 

I realised this won't apply to prerendered pages; I thought we might be able to solve that with `http-equiv` but alas, only certain headers can be represented that way: https://html.spec.whatwg.org/multipage/semantics.html#pragma-directives

### Tests

### Changesets


packages/kit/src/runtime/server/page/render.js
```diff
@@ -169,6 +169,10 @@ export async function render_response({
 		headers['cache-control'] = `${is_private ? 'private' : 'public'}, max-age=${maxage}`;
 	}
 
+	if (!options.floc) {
+		headers['permissions-policy'] = 'interest-cohort=()';
```
Conduitry, line 173: Are we worried about this being unsafe? The example PHP code in the Wordpress blog post about this had a bit more logic in place for dealing with an existing `permissions-policy` header.

Rich-Harris: We don't currently expose a mechanism for returning custom headers from pages, so this is safe — the only caveat being that user code in `handle` needs to be aware of existing headers

Conduitry: If someone returned a `permissions-policy` header from an _endpoint_ though, this would clobber it, right? Are you concerned about that?

Rich-Harris: This doesn't touch endpoint headers at all, only pages (AFAICT FLoC is only relevant there?)

Conduitry: Oh okay. Well I suppose someone could still be returning HTML from an endpoint that gets navigated to at some point, but that's probably an edge case. Adding this header to all page responses by default sounds reasonable to me.

documentation/docs/13-configuration.md
```diff
@@ -72,6 +73,18 @@ An object containing zero or more of the following `string` values:
 - `hooks` — the location of your hooks module (see [Hooks](#hooks))
 - `template` — the location of the template for HTML responses
 
+### floc
+
+Google's [FLoC](https://github.com/WICG/floc) is a technology for targeted advertising that the [Electronic Frontier Foundation](https://www.eff.org/) has deemed [harmful](https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea) to user privacy. [Browsers other than Chrome](https://www.theverge.com/2021/4/16/22387492/google-floc-ad-tech-privacy-browsers-brave-vivaldi-edge-mozilla-chrome-safari) have declined to implement it.
+
+In common with services like [GitHub Pages](https://github.blog/changelog/2021-04-27-github-pages-permissions-policy-interest-cohort-header-added-to-all-pages-sites/), SvelteKit protects your users by automatically opting out of FLoC. It adds the following header to responses unless `floc` is `true`:
+
+```
```
Conduitry, line 82: Does the current implementation of syntax highlighting correctly handle syntax-less code blocks? I recall that at some point this was an issue on the Svelte site.

Rich-Harris: this suggests it's optional: https://github.com/sveltejs/action-deploy-docs/blob/42d39f22a94353de238f9e2382645458e1ebb99f/src/format/code.ts#L29-L30

Conduitry: Ah! Okay. I was poking around for where that code lived now but wasn't finding it.

Conduitry (approved):

orta (approved):

PR #1272, opened 2021-04-29
fix parsing endpoint json body

### LINKED ISSUE #1226: Endpoint should not call JSON.stringify(body) if content-type is json
**Describe the bug**
According to the doc, endpoint's body should only be converted to JSON string when the `content-type` header doesn't exist.

Also converting to JSON string when the `content-type` header is `application/json` prevents the endpoint from functioning as a relay, that is, it can't fetch an API that returns a JSON string and then forwards the headers and the body as they are, because svelte kit would call JSON.stringify on the string again, and `content-length` would also be incorrect.

**Logs**

**To Reproduce**
```ts
export async get() {
  return {
    headers: { 'content-type': 'application/json' },
    body: '{}',
  }
}
```

browser now receives`'"{}"'`

**Expected behavior**
browser should receive `'{}'`

**Stacktraces**

**Severity**
Very, being able to make endpoint function as a relay is pretty important.

**Additional context**

I believe this is the code

https://github.com/sveltejs/kit/blob/e616da248f51584e58bece9d0e906638dba43739/packages/kit/src/runtime/server/endpoint.js#L34

I'd be happy to send a PR if svelte kit team think it's ok to remove this condition check.

hgl:
Converting to JSON string also when the 'content-type' header is 'application/json'
prevents the endpoint from functioning as a relay, that is, it can't fetch
an API that returns a JSON string and then forwards the headers and the body
as they are, because svelte kit would call JSON.stringify on the string again.

Closes #1226

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/endpoint.js
+++ b/packages/kit/src/runtime/server/endpoint.js
@@ -19,7 +19,7 @@ export default async function render_route(request, route) {
 			if (typeof response !== 'object') {
 				return {
 					status: 500,
-					body: `Invalid response from route ${request.path}; 
+					body: `Invalid response from route ${request.path};
 						 expected an object, got ${typeof response}`,
 					headers: {}
 				};
@@ -29,10 +29,7 @@ export default async function render_route(request, route) {
 
 			headers = lowercase_keys(headers);
 
-			if (
-				(typeof body === 'object' && !('content-type' in headers)) ||
-				headers['content-type'] === 'application/json'
-			) {
+			if (typeof body === 'object' && !('content-type' in headers)) {
 				headers = { ...headers, 'content-type': 'application/json' };
 				body = JSON.stringify(body);
 			}
--- a/packages/kit/src/runtime/server/endpoint.js
+++ b/packages/kit/src/runtime/server/endpoint.js
@@ -29,7 +29,10 @@ export default async function render_route(request, route) {
 
 			headers = lowercase_keys(headers);
 
-			if (typeof body === 'object' && !('content-type' in headers)) {
+			if (
+				typeof body === 'object' &&
+				(!('content-type' in headers) || headers['content-type'] === 'application/json')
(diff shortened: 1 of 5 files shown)


packages/kit/src/runtime/server/endpoint.js
```diff
@@ -29,10 +29,7 @@ export default async function render_route(request, route) {
 
 			headers = lowercase_keys(headers);
 
-			if (
-				(typeof body === 'object' && !('content-type' in headers)) ||
-				headers['content-type'] === 'application/json'
-			) {
+			if (typeof body === 'object' && !('content-type' in headers)) {
```
benmccann, line 32: should this keep the old condition, but with the parens in a different place? I.e.
```
if (typeof body === 'object' && (!('content-type' in headers) || headers['content-type'] === 'application/json'))
```

hgl: done.

benmccann (approved):

Rich-Harris:
thank you!

AaronDDM:
Hey, @hgl @benmccann @Rich-Harris 

Any way we could revisit this? This seems inconsistent and requires unnecessary additional logic to handle. 
If we are going to automatically `JSON.stringify()` we should do it for all the data and not just some of the data.

A body response of `astring` and `"astring"` are totally different if the content type is expected to be JSON.

hgl:
> requires unnecessary additional logic to handle on part of the developer

> a developer can be expected to convert their JSON response to a JavaScript object via ...

Seems contradictory.

Calling `JSON.stringify()` is only skipped if you explicitly set the `content-type: application/json`, and the doc specifies

> If the returned body is an object, and no content-type header is returned, it will automatically be turned into a JSON response.

What's you're suggesting is actually a reversion of what's in the doc.

Maybe you could expand on the use case?

PR #1280, opened 2021-04-30
invalidate resource

### LINKED ISSUE #1277: Invalidate fetched URLs
Suppose you were building an e-commerce site. You might have a shopping cart icon in your root `src/routes/$layout.svelte` component. Inside `src/routes/products/[category]/[id].svelte` you might have an 'add to cart' button. When the user clicks the button, the cart icon should update.

One perfectly valid way to model this would be to put the user's cart on the `session` store, and then update the store after the POST triggered by 'add to cart' succeeds:

```js
async function submit(e) {
  e.preventDefault(); // because we're progressively enhancing the 'add to cart' <form>

  const product = await fetch(this.action, {
    method: this.method,
    body: new FormData(this)
  });

  $session.cart.products = [...$session.cart.products, product];
}
```

But you might have valid reasons for not wanting to put that data in `session` — perhaps the cart is only visible on some parts of the app and you don't want to always pay the cost of serializing the data. Instead, you might do something like this:

```svelte
<!-- src/routes/$layout.svelte -->
<script context="module">
    const cart = await fetch('/cart.json').then(r => r.json());
    return {
      props: { cart }
    };
  }
</script>
```

This creates a problem. POSTing to `/cart.json` will add the product(s) to the cart, but will no longer cause the cart icon to update, because the layout component has no way of knowing that the result of fetching `/cart.json` has changed.

Since we're using the passed-in `fetch`, we know which URLs the layout component depends on. We could therefore expose an API like this:

```diff
+// may not be the best place for this functionality, but we can bikeshed that
+import { invalidate } from '$app/navigation';
+
async function submit(e) {
  e.preventDefault(); // because we're progressively enhancing the 'add to cart' <form>

  const product = await fetch(this.action, {
    method: this.method,
    body: new FormData(this)
  });

-  $session.cart.products = [...$session.cart.products, product];
+  invalidate('/cart.json');
}
```

SvelteKit could straightforwardly keep track of which `load` functions fetched the invalidated URL, and rerun them (assuming the components they belong to are still mounted), much as we already do for `load` functions that use `session` when that store changes value.

This is a somewhat low-tech solution. A more common approach might be to use a full-blown fat-client state management system that has a copy of the model in memory (see e.g. Firebase). One possibility would be to integrate more closely with such systems or even build one ourselves.

Personally I'm more inclined towards simple, explicit mechanisms that are easy to implement and understand; in any case this would be complementary towards the more full-blown solutions (or at the very least, not in conflict).

Other possibilities that I'm not keen on:

* Intercepting all non-GET fetches and invalidating all `load` functions automatically
* Offering a way to invalidate everything in one go, rather than invalidating specific URLs

It's a nice-to-have — it solves a problem that can already be solved in various ways, but arguably in a more elegant way. And it's the one feature I've seen in [Remix](https://remix.run/) that we don't already have a (more elegant, IMHO 💅 ) equivalent of that I'm aware of :)

(Their version re-runs all loaders indiscriminately whenever an 'action' — their word for POST/PUT/PATCH/DELETE etc requests — takes place. AFAICT it wouldn't work with external APIs, just the app's own endpoint. But it does have the nice characteristic that it runs automatically.)

tanhauhau:
Fixes #1277

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/types.d.ts
+++ b/packages/kit/src/runtime/client/types.d.ts
@@ -2,40 +2,41 @@ import { Page, LoadOutput } from '../../../types/page';
 import { CSRComponent, CSRRoute } from '../../../types/internal';
 
 export type NavigationInfo = {
-	id: string;
-	routes: CSRRoute[];
-	path: string;
-	query: URLSearchParams;
+  id: string;
+  routes: CSRRoute[];
+  path: string;
+  query: URLSearchParams;
 };
 
 export type NavigationCandidate = {
-	route: CSRRoute;
-	path: string;
-	query: URLSearchParams;
+  route: CSRRoute;
+  path: string;
+  query: URLSearchParams;
 };
 
 export type NavigationResult = {
-	reload?: boolean;
-	redirect?: string;
-	state?: NavigationState;
-	props?: Record<string, any>;
+  reload?: boolean;
+  redirect?: string;
+  state?: NavigationState;
+  props?: Record<string, any>;
 };
 
 export type BranchNode = {
-	module: CSRComponent;
-	loaded: LoadOutput;
-	uses: {
-		params: Set<string>;
-		path: boolean;
-		query: boolean;
-		session: boolean;
-		context: boolean;
-	};
-	context: Record<string, any>;
+  module: CSRComponent;
+  loaded: LoadOutput;
+  uses: {
+    params: Set<string>;
+    path: boolean;
+    query: boolean;
+    session: boolean;
+    context: boolean;
+    invalidates: Set<string>;
+  };
+  context: Record<string, any>;
 };
 
 export type NavigationState = {
-	page: Page;
-	branch: BranchNode[];
-	session_id: number;
+  page: Page;
+  branch: BranchNode[];
+  session_id: number;
 };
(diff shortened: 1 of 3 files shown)


packages/kit/src/runtime/client/types.d.ts
```diff
@@ -2,40 +2,41 @@ import { Page, LoadOutput } from '../../../types/page';
 import { CSRComponent, CSRRoute } from '../../../types/internal';
 
 export type NavigationInfo = {
-	id: string;
-	routes: CSRRoute[];
-	path: string;
-	query: URLSearchParams;
+  id: string;
```
benmccann, line 5: looks like tabs got converted to spaces in this file

Rich-Harris:
working on this locally — at the moment it doesn't play well with caching (if you invalidate a URL within a page's `maxage` window, it doesn't do anything). am also wondering if we should batch invalidations, or if `invalidate(...urls)` is a sufficiently good API that batching is unnecessary

Conduitry:
Scheduled-for-next-microtask batching, like a bunch of other things in Svelte core, might be nice. Were you thinking of larger batches than that?

Rich-Harris:
no, just microtask

Rich-Harris:
closing in favour of #1303 — thanks

PR #1284, opened 2021-04-30
Swap CI from Node 15 to Node 16

GrygrFlzr:
Do we still want to test against Node 15?


Conduitry:
It's not EOL for another month, but yeah I think we can drop this.

PR #1293, opened 2021-05-01
Remove pnpm lockfile

GrygrFlzr:
Forces CI to test against the latest versions of our dependencies.


benmccann:
This seems like it could cause a lot of confusion. People could have tests that are passing then suddenly start failing on the CI without explanation and it's because someone introduced a new version of a library. The point of the lockfile is so that we have control over when new versions of dependencies are used in development and tests to avoid those sorts of issues

pngwn:
Yeah. I don’t know what this is solving specifically but removing the lock file is not the solution. 

Even if it is does address some limitation we lose everything else the lock file brings, for all of our other dependencies, which isn’t something we want to do.

Rich-Harris:
maybe we use dependabot?

pngwn:
I'm not sure if dependabot works with `pnpm` we had some issues at work and are using [renovate](https://github.com/renovatebot/renovate) instead. 

I'm not sure if it will detect when those symlinked dependencies have changed (the version is `workspace:*`), but running `pnpm i --lockfile-only` should always update the lock file to reflect the latest kit version. We could set up a cron workflow for this if needed.

Conduitry:
Luke's suggestion was to pin devdeps to particular versions in package.json (and to have no lock file). I think dependabot noise would be a bit oppressive though.

benmccann:
I agree dependabot is noisy. Personally I'm not really sure what problem is being solved here and am perfectly fine with the status quo

PR #1294, opened 2021-05-01
fix(template): typehint default config file

lukeed:

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/templates/skeleton/svelte.config.cjs
+++ b/packages/create-svelte/templates/skeleton/svelte.config.cjs
@@ -2,7 +2,7 @@ const node = require('@sveltejs/adapter-node');
 const pkg = require('./package.json');
 
 /** @type {import('@sveltejs/kit').Config} */
-module.exports = {
+const config = {
 	kit: {
 		// By default, `npm run build` will create a standard Node app.
 		// You can create optimized builds for different platforms by
@@ -13,3 +13,5 @@ module.exports = {
 		target: '#svelte'
 	}
 };
+
+module.exports = config;


benmccann:
Is this for VS Code support? Seems like a bug there if this is required, but I'm fine with changing it here to workaround it anyway. It looks like we already do this in [the default template](https://github.com/sveltejs/kit/blob/master/packages/create-svelte/templates/default/svelte.config.cjs)

lukeed:
Yeah, it's the only correct way to actually apply JSDoc typehints – has to be applied to a variable/function

benmccann (approved):
You'll have to add a changeset with `pnpx changeset`

Rich-Harris:
Does this become a non-issue when we switch to ESM configs? Can you type an `export default`?

lukeed:
Nope, still variables and functions only

lukeed:
There is an `@exports` directive but I've never got that to work. Even then, due to the age, it was probably CommonJS-specific

Rich-Harris:
actually just realised that this PR won't change anything, the edited file gets replaced by either [this one](https://github.com/sveltejs/kit/blob/master/packages/create-svelte/shared/-typescript/svelte.config.cjs) or [this one](https://github.com/sveltejs/kit/blob/master/packages/create-svelte/shared/%2Btypescript/svelte.config.cjs). 

But now that I look at those files, I see this —

image

— in other words I'm pretty sure this only applies to ESM? Annoying that it fails with `export default` but works fine with `export const`, feels like something that could be fixed in TS. (Better still, let us type modules!)

PR #1302, opened 2021-05-02
Added warning in SSR for missing replacers

### LINKED ISSUE #1301: Add warning in SSR when %svelte.head% or %svelte.body% is missing in template file
I had this issue a moment ago. I removed the 

```html
<div id="svelte"></div>
```
from the app.html because i wanted to target body. After some development steps I found out, that my SSR isnt working. @GrygrFlzr gave me the hint thats it is missing.

In my mind, there should be warning

Kapsonfire-DE:
Fixes #1301


Rich-Harris:
Thank you — I'm going to close this in favour of #1304 which solves the same problem but at an earlier stage (which therefore also applies to `build`)

PR #1312, opened 2021-05-03
Improve promts, turn confirms into toggle

TheWolfNL:
  >This change will make it obvious what the default values are, and what the alternatives are.
  >I found myself having to run the init multiple times since the default was not obvious or I 'entered' past it 😅 
  >The resulting values remain the same as before the change.
  Example of the changed promts
  >Use arrow keys/tab/space to switch between options, enter to select.

### Tests
- [ ] Run the tests with `pnpm test` 
- [x] Lint the project with `pnpm lint`

### Changesets

### DIFF
--- a/.changeset/healthy-moons-attend.md
+++ b/.changeset/healthy-moons-attend.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Improved install prompts, turn confirms into toggle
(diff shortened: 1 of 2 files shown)


benmccann (approved):

Rich-Harris:
thanks!

PR #1316, opened 2021-05-03
Add `prerendering` to `$app/env` typings

TravisSpomer:
This change makes TypeScript stop yelling at you if you use `prerendering` from `src/hooks.ts` or wherever.

Since this doesn't affect anything at runtime I wasn't sure if it should have a changeset from `pnpx changeset`; should I add one?

### DIFF
--- a/.changeset/strong-cooks-turn.md
+++ b/.changeset/strong-cooks-turn.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Add prerendering to \$app/env typings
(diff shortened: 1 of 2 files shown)


Rich-Harris:
Thanks! It should have a changeset, yes, because that's what triggers the release workflow

TravisSpomer:
> 
> 
> Thanks! It should have a changeset, yes, because that's what triggers the release workflow

Okay, done! I think!

benmccann:
Is there anything we could have done to catch this error like type checking the tests or examples? It seems like we've had a few reports of issues like this and I'd like to avoid this keep happening if possible

Rich-Harris:
Thanks @TravisSpomer 

> Is there anything we could have done to catch this error like type checking the tests or examples?

opened a draft PR for that: https://github.com/sveltejs/kit/pull/1319

PR #1317, opened 2021-05-03
Mention `prerendering` during `svelte-kit preview`

TravisSpomer:
The documentation for `svelte-kit preview` does mention that it functions irrespective of your current adapter, but it took a while for it to click for me that that meant any changes made in `handle` wouldn't be visible when previewing. Hopefully this clarification will help someone like me in the future. :)


Rich-Harris:
thanks!

PR #1319, opened 2021-05-03
typecheck svelte files in tests

Rich-Harris:
Adds `svelte-check` to the `check` script in packages/kit, which will hopefully help us catch more type errors early (https://github.com/sveltejs/kit/pull/1316#issuecomment-831491962).

This doesn't actually fix any of the surfaced errors, but it's low-hanging fruit if anyone wants to work on that

### Tests

### Changesets


Rich-Harris:
Made two changes to types:

* Added an internal `NormalizedLoadOutput` that mirrors the public `LoadOutput`, and changed `LoadOutput` to allow `string | Error` for `loaded.error`. An alternative would be to only allow `Error`
* Referenced Vite/Svelte types inside the Kit types. This means that people only need to reference Kit types in their own apps

packages/kit/src/runtime/client/types.d.ts
```diff
@@ -23,7 +23,7 @@ export type NavigationResult = {
 
 export type BranchNode = {
 	module: CSRComponent;
-	loaded: LoadOutput;
+	loaded: NormalizedLoadOutput;
```
dummdidumm, line 26: If `NormalizedLoadOutput` is exported from the public types I wouldn't consider it an internal type and therefore move it into a public type, possibly `page`.

Rich-Harris: where is it being exported from the public types?

dummdidumm: Ok don't mind me, I'm stupid. I misread the file name and thought this was one of the public d.ts files.

Issue #984, opened 2021-04-12
Rename `context` to `stuff`


benmccann:
There are a couple different places we use the word `context`
- `load` has a `context` parameter
- `getContext` sets `request.context`

Adding this here to track and have a discussion at some point since a couple maintainers have mentioned this as being something we might want to considers

Also, it looks like perhaps we should fix this TODO in the process:
https://github.com/sveltejs/kit/blob/08ebcb518c71cd58f4717784a5144cb195b49a68/packages/kit/src/runtime/server/page/respond.js#L178

c-bandy:
Yes please!  I frequently see confusion on why`getContext()` is not what is expected in a `load()` function. It probably doesn't help that [the documentation on the matter is a bit sparse](https://kit.svelte.dev/docs#loading).

rmunn:
Should one be called `serverContext` and one be called `clientContext`? Except that `load` runs on the server *and* the client, so that's no good. How about `endpointContext` vs `loadContext`? Or just leave it called `context` when it's the thing returned by `getContext`, but call it `loadContext` in the `load` functions? I'm not strongly in favor of any of these, I'm just throwing out ideas here.

Rich-Harris:
I'm less concerned about the difference between `context` in `load` and `getContext` than I am about the overlap between `context` in `load` and `request.context`. If we could rename the latter, I think this would be less of an issue.

My previous suggestion was `request.metadata`/`getMetadata(...)` but people didn't love it.

`getRequestInfo(...)` -> `request.info`?

`getRequestState(...)`/`getState(...)` -> `request.state`?

benmccann:
`getState(...)` -> `request.state` sounds good to me

Leftium:
The overloaded use of `context` definitely made SvelteKit more confusing and difficult for me to learn. Documenting what I see to be sources/inspiration for this conflict:

- [`getContext() -> request.context`][5] resembles serverless lambda functions. (See [Netlify][2] or [AWS][1] API documentation.)
- [`load({ context })`][4] resembles [React's context API][3] for sharing data between components.

---

I propose changing `load({ context })` to `load({ info })`:
- Keep `getContext() -> request.context` because of parallels with the `context` in serverless lambda functions. (AWS/Netlify/etc...)
- Many React context API tutorials often refer to managing "global component state".  (React may have avoided using `state` because it already has a special meaning.)
- However "state" seems to imply changes will be persisted/propagated back. Especially for those coming from React. So I prefer "info."
[...]

PR #1332, opened 2021-05-04
Remove getContext in favour of request.locals

Rich-Harris:
Implements the first part of https://github.com/sveltejs/kit/issues/984#issuecomment-830656946

### Tests

### Changesets

### DIFF
--- a/documentation/docs/04-hooks.md
+++ b/documentation/docs/04-hooks.md
@@ -2,74 +2,9 @@
 title: Hooks
 ---
 
-An optional `src/hooks.js` (or `src/hooks.ts`, or `src/hooks/index.js`) file exports three functions, all optional, that run on the server — **getContext**, **getSession** and **handle**.
+An optional `src/hooks.js` (or `src/hooks.ts`, or `src/hooks/index.js`) file exports two functions, both optional, that run on the server — **handle** and **getSession**.
 
 > The location of this file can be [configured](#configuration) as `config.kit.files.hooks`
-
-### getContext
-
-This function runs on every incoming request. It generates the `context` object that is available to [endpoint handlers](#routing-endpoints) as `request.context` and that is used to derive the [`session`](#hooks-getsession) object available in the browser.
-
-If unimplemented, context is `{}`.
-
-```ts
-type Incoming = {
-	method: string;
-	host: string;
-	headers: Headers;
-	path: string;
-	query: URLSearchParams;
-	body: string | Buffer | ReadOnlyFormData;
-};
-
-type GetContext<Context = any> = {
-	(incoming: Incoming): Context;
-};
-```
-
-```js
-import * as cookie from 'cookie';
-import db from '$lib/db';
-
-/** @type {import('@sveltejs/kit').GetContext} */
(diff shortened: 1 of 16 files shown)


documentation/docs/04-hooks.md
```diff
@@ -2,117 +2,86 @@
...
-	({ context }: { context: Context }): Session | Promise<Session>;
+type GetSession<Locals = Record<string, any>, Session = any> = {
+	(request: Request<Locals>): Session | Promise<Session>;
 };
 ```
 
 ```js
 /** @type {import('@sveltejs/kit').GetSession} */
-export function getSession({ context }) {
+export function getSession(request) {
```
Conduitry, line 73: Are we confident enough about this that we want to forgo the (single) named parameter here? I have no idea what else in particular might arise, but that doesn't seem entirely watertight.

packages/create-svelte/templates/default/src/hooks.ts
```diff
@@ -1,33 +1,25 @@
...
 	// TODO https://github.com/sveltejs/kit/issues/1046
 	const response = await render({
 		...request,
 		method: (request.query.get('_method') || request.method).toUpperCase()
 	});
 
-	const { is_new, userid } = request.context;
-
-	if (is_new) {
+	if (!cookies.userid) {
```
Conduitry, line 15: Not really related to this change, but is there a reason not to just mutate the `response` object before returning it, which would be a bit simpler?

documentation/docs/04-hooks.md
Rich-Harris: I'm reasonably confident about the single parameter — I can't imagine what else we would want to derive `session` from other than `request`. I'm less confident that we want to pass the whole request in rather than just `locals` but I couldn't quite articulate why, so I figured this was probably fine

packages/create-svelte/templates/default/src/hooks.ts
Rich-Harris: nope! updated it. feels much nicer.

Rich-Harris:
This is another PR that will cause checks to fail until we release a new version — in this case because https://netlify.demo.svelte.dev/ won't build and deploy itself until `@sveltejs/kit@next` uses `request.locals` instead of `request.context`

Conduitry (approved):

packages/create-svelte/templates/default/svelte.config.js
```diff
@@ -16,7 +16,7 @@ const config = {
 };
 
 if (adapter) {
-	config.kit.adapter = require(adapter)(options);
+	config.kit.adapter = (await import(adapter)).default(options);
```
benmccann, line 19: If we need the `.default` here then do we need to do that in the adapter READMEs as well? None of them have that right now

Conduitry: I believe this is only like that because it is dynamically importing the adapter. Normally you'd just do a static import, and you'd get the default export by default.

Rich-Harris: correct

PR #1349, opened 2021-05-05
use TypedArray.set as the "simpler way" to copy

asos-tomp:

This PR addresses the TODO:

> // TODO maybe there's a simpler way to copy data between buffers?

...by replacing with [TypedArray.prototype.set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/set). 

This is a [built-in, in current v8](https://github.com/v8/v8/blob/master/src/builtins/typed-array-set.tq), seems to be [at least as performant as the loop](http://jsbench.github.io/#0ff5b7c3f39925b96849380f01d5c7e6), with gains increasing as the size of the buffer increases. 

### Tests

### Changesets

### DIFF
--- a/.changeset/lucky-bottles-kick.md
+++ b/.changeset/lucky-bottles-kick.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Use TypedArray.set to copy from Uint8Array when getting raw body in core/http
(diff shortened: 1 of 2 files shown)


Rich-Harris:
excellent, thank you!

Issue #1161, opened 2021-04-21
load function does not respect returned error status codes


kamholz:
According to the docs, a page's `load` function can return an error via `{ status: 4xx }`. However, the following test page component fails to work:

```js
<script context="module">
  export async function load({ session }) {
    if (!session.user) {
      return { status: 401 };
    }
    return {};
  }
</script>
```

Instead of returning status 401, the page returns status 200 and renders anyway. If I change it to something like `return { status: 302, redirect: 'https://google.com' }` then it does redirect, so this seems to just be an issue with error codes?

danawoodman:
Can confirm this on `1.0.0-next.84`:

```html
<script context="module">
  export async function load() {
    return { status: 401 }
  }
</script>
```

Returns `200`

lukasIO:
have you tried setting the `error` object additionally to the status code?

kamholz:
> have you tried setting the `error` object additionally to the status code?
> 
> > If something goes wrong during load, return an Error object or a string describing the error alongside a 4xx or 5xx status code.

Yes. It doesn't make a difference.

kamholz:
> Yes. It doesn't make a difference. Also, setting the `error` is not required in endpoints, so it should be consistent.

Correction, it does work now when I return `error` as well (not sure why it wasn't before). This should definitely be standardized between pages and endpoints if possible, because I'm sure other people will be confused by this at some point.

Rich-Harris:
The place to append an error for `4xx`/`5xx` status codes is here, if anyone would like to open a PR that fixes this: https://github.com/sveltejs/kit/blob/32b3fcde3c206347d0ca58ea2f8ecec649f5b7f5/packages/kit/src/runtime/load.js#L5-L8

bedewilson:
Failing test has been developed in #1355 but implementation of the fix is still pending. I closed the PR until this is done.

damain:
Looking at taking up this issue, but what is the intended behavior?

Should you be able to able to send a 4xx status without an error?
Should a 4xx status without an error also trigger the appropriate error to be generated?

kamholz:
Yes, you should be able to send a 4xx status without specifying the `error`. SvelteKit should set it to a default value in that case (whatever the standard message is for each error, I suppose).

damain:
I did the following statuses are there any others that should be addressed

```
switch (status) {
				case 400:
					errorMessage = 'Bad Request';
					break;
				case 401:
[...]

PR #1355, opened 2021-05-06
Add failing test for load status code bug

bedewilson:
Failing test for #1161. When the `load()` function returns a 4XX/5XX status code, but no `error`, the response resolves with status `200`. The specified status code should be passed through instead.

As this relates to an instance where no error (apart from the status code) is provided, I have assumed that the error message should default to the standard http response string (e.g. specifying a `500` status code should result in an `error.message` of `Internal Server Error`.

### Tests
- [X] Run the tests with `pnpm test` and lint the project with `pnpm lint`

### Changesets

### DIFF
--- a/packages/kit/test/apps/basics/src/routes/errors/__tests__.js
+++ b/packages/kit/test/apps/basics/src/routes/errors/__tests__.js
@@ -112,6 +112,34 @@ export default function (test, is_dev) {
 	);
 
 	test(
+		'server-side error from load() is a 4XX/5XX status code',
+		'/errors/load-error-code-server',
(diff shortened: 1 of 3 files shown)


benmccann:
Thanks. We can't really check this PR in with the tests failing. Do you plan to fix the underlying issue as well? If not, then this is still helpful, but I think the best course of action in that case would be to close this to help keep the review queue clean and then we can leave a comment on the issue referencing this PR or your branch so that whoever does fix it can use your test

bedewilson:
Thanks - sorry did intend to but have been caught up. Happy to close for now and I will get to it over the next few days.

Issue #1100, opened 2021-04-18
Imported/Declared CSS from $layout.svelte is not ignored with $layout.reset.svelte. Dev and Build include different CSS <link> in different order


DhyeyMoliya:
**Describe the bug**
1. I have the main `$layout.svelte` and a `$layout.reset.svelte` for `/admin` path
2. In `$layout.svelte`, we declare some `<style>` or import some stylesheet `import ../../../*.css or ../../../*.scss`.
3. In `$layout.reset.svelte`, we declare same `<style>` or import some stylesheet `import ../../../*.css or ../../../*.scss`.
4. When we open `http://localhost:3000/admin`,
5. In `svelte-kit dev`, both `<style>` from `$layout.svelte` and `<style>` from `$layout.reset.svelte` are applied.
6. In `svelte-kit build` output, both `<link href="layout.css">` and `<link href="$layout.reset.css">` are applied.
7. Main Issue is, both Styles are applied even if we have `reset` the layout.
8. Another Issue is, in `build` output's `<head>` tag, `<link href="$layout.reset.css">` is preceded by `<link href="layout.css">`, resulting in an output where style from `$layout.svelte` are applied instead of `$layout.reset.svelte` after build. Because of the declaration order of CSS. 

`svelte-kit dev` output :

<img width="539" alt="Screenshot 2021-04-18 at 5 57 29 PM" src="https://user-images.githubusercontent.com/6321840/115145615-13914180-a070-11eb-8a33-529e340ddf80.png">

`svelte-kit build` output :

<img width="704" alt="Screenshot 2021-04-18 at 5 59 23 PM" src="https://user-images.githubusercontent.com/6321840/115145628-20ae3080-a070-11eb-866d-3838b829e50a.png">

**Logs**
No Logs.

**To Reproduce**
1. Create new SvelteKit project.
2. Create a `$layout.svelte` at the root of `routes` folder
3. Create a `$layout.reset.svelte` at any sub route (e.g /admin). 
4. Add <style> or import some style file in both `$layout.svelte` and `$layout.reset.svelte`.
5. Check different behaviours of `svelte-kit dev` and `svelte-kit build`.

or

Reproduction Repo : https://github.com/DhyeyMoliya/sveltekit-css-issue
1. `main` branch for `<style>` example
2. `style-import` branch from `import ../../*.css` example

**Expected behavior**
[...]

PR #1356, opened 2021-05-06
Fix $layout.svelte styles being imported in routes having $layout.reset.svelte

DhyeyMoliya:

### Tests

### Changesets

This PR fixes https://github.com/sveltejs/kit/issues/1100 , More Details :
- The imported CSS inside `$layout.svelte` was applying in pages with `$layout.reset.svelte` as well.
- This was happening because of `manifest.js` file having an export `fallback = [c[0](), c[1]()]`. 
- Because of `c[0]()` the `$layout.svelte` was being executed at the start without it being required in a page or not.
- This PR fixes this by changing the code to `fallback = () => [c[0](), c[1]()]`. So, the code does not execute at the start.

### DIFF
--- a/.changeset/chilly-cars-protect.md
+++ b/.changeset/chilly-cars-protect.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Changed manifest.js fallback from array to a function returning the array
--- a/packages/kit/src/core/create_app/index.js
+++ b/packages/kit/src/core/create_app/index.js
@@ -103,7 +103,7 @@ function generate_client_manifest(manifest_data, base) {
 
 		export const routes = ${routes};
 
-		export const fallback = [c[0](), c[1]()];
+		export const fallback = () => [c[0](), c[1]()];
 	`);
 }
 
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -51,7 +51,7 @@ function initial_fetch(resource, opts) {
 export class Renderer {
 	/** @param {{
 	 *   Root: CSRComponent;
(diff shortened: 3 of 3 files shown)


Rich-Harris:
Thank you. The current behaviour _is_ deliberate though — the idea is that if a navigation fails, you need to have the root layout/error components available in memory (rather than a network request away, since a common reason for said failure is a lack of connectivity) so that things don't fall apart. This does cause a wasted request in the case where you initially hit a page with a layout reset, but on balance I think it's preferable.

DhyeyMoliya:
Also, to reference some future issues similar to this. 

The current code works like this :

When we load one page with main layout, the main layout css is appened to the head.

Then we go to another page with layout reset.svelte, then layout.reset's css is appended to head.

When we come back to the page with main layout.reset, the css of layout.reset is not removed by vite.

This PR is not able to solve that issue. Since it will require more deeper changes inside vite or hmr stuff.

If you say, I can discard this PR. 


I have an alternative solution that completely works without any errors. Using <link> inside <svelte:head>. So the css is only added when that component is active. Without relating to layout and it's reset logic.

Rich-Harris:
Correct, links aren't removed. As long as you're not using ambiguous selectors (`:global` in components, or selectors without appropriate namespacing elsewhere) it shouldn't have an observable effect.

> I have an alternative solution that completely works without any errors. Using `<link>` inside` <svelte:head>`. So the css is only added when that component is active. Without relating to layout and it's reset logic.

This will cause FOUC — there'll be a gap between the component rendering and the `<link>` loading — and is therefore unrecommended.

DhyeyMoliya:
> This will cause FOUC — there'll be a gap between the component rendering and the `<link>` loading — and is therefore unrecommended.

Then I'm left with no solution to this problem.

PR #1360, opened 2021-05-06
fix: Prettier not formatting .svelte files

### LINKED ISSUE #1358: Prettier does not format svelte files with default config
**Describe the bug**
The default npm `format` script in the `create-svelte` templates does not apply Prettier to `.svelte` files, even though `prettier-plugin-svelte` is installed.

To make Prettier use `prettier-plugin-svelte` and format `.svelte` files, it is necessary to use the argument `--plugin-search-dir=.`.

**To Reproduce**
```shell
pnpm init svelte@next sveltekit-test
```
Select *SvelteKit demo app* template
Enable Prettier (and optionally ESlint and Typescript)
```
cd sveltekit-test
pnpm install
pnpm run format
```
Observe that Prettier does not format `.svelte` files.

**Expected behavior**
The default npm `format` script should use Prettier to format `.svelte` files as well as `.js`, `.ts`, etc.

<details>

  System:
    OS: macOS 11.3
    CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Memory: 94.60 MB / 16.00 GB
    Shell: 5.8 - /usr/local/bin/zsh
  Binaries:
    Node: 16.0.0 - /usr/local/bin/node
    npm: 7.11.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 90.0.4430.93
    Firefox: 89.0
    Safari: 14.1
</details>

**Severity**
Fairly minor, since once you know what the issue is it can easily be resolved by modifying the `format` script in `package.json`.

**Additional context**
To fix this issue, change `package.json` as follows:
```diff
- 		"format": "prettier --write .",
+		"format": "prettier --write --plugin-search-dir=. .",
```

caesar:
Adds `--plugin-search-dir=.` param to Prettier call in `package.json` scripts, so that Prettier will format `.svelte` files as well as `.ts`, `.js`, etc.

I assume Prettier is intended to be applied to `.svelte` files, since `prettier-plugin-svelte` is installed.

Fixes #1358.

### Tests

### Changesets

### DIFF
--- a/.changeset/wicked-spies-invite.md
+++ b/.changeset/wicked-spies-invite.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+fix: Prettier not formatting .svelte files
--- a/packages/create-svelte/shared/+eslint+prettier/package.json
+++ b/packages/create-svelte/shared/+eslint+prettier/package.json
@@ -3,7 +3,7 @@
 		"eslint-config-prettier": "^8.1.0"
 	},
 	"scripts": {
-		"lint": "prettier --check . && eslint --ignore-path .gitignore .",
-		"format": "prettier --write ."
+		"lint": "prettier --check --plugin-search-dir=. . && eslint --ignore-path .gitignore .",
+		"format": "prettier --write --plugin-search-dir=. ."
 	}
 }
--- a/packages/create-svelte/shared/-eslint+prettier/package.json
+++ b/packages/create-svelte/shared/-eslint+prettier/package.json
@@ -1,6 +1,6 @@
 {
 	"scripts": {
-		"lint": "prettier --check .",
-		"format": "prettier --write ."
+		"lint": "prettier --check --plugin-search-dir=. .",
+		"format": "prettier --write --plugin-search-dir=. ."
 	}
 }


Rich-Harris:
Thanks, but I can't reproduce this — if I run `npm run format` in a fresh project, my .svelte files are formatted correctly. What am I missing?

caesar:
That's very odd. As described in #1358 this is reproducible for me with a fresh project.
The only thing I can think of is pnpm vs npm (I used pnpm), but I can't see why it would make a difference.
I'll see if I can work it out…

caesar:
As a side note, https://github.com/sveltejs/prettier-plugin-svelte#how-to-use-cli does suggest using `--plugin-search-dir=.` so it's not just me 🙂

Conduitry:
I can reproduce this issue when installing dependencies with pnpm, but not when installing them with npm. With npm. `npm run format` also gets the `.svelte` files.

caesar:
@Conduitry I was just writing exactly the same message 😀

So this must be to do with the different layout of `node_modules` used by pnpm.

It would be great if the default config supported pnpm (especially since the instructions say `npm install (or pnpm install, etc)`. So I suppose the question is whether it is appropriate to fix (work around?) it in `create-svelte` as I have here, or whether it is something that should be fixed in Prettier.

I'd lean towards the former in the short term, but I guess it might be something to look at from Prettier's side too.

Rich-Harris:
Ah, interesting. In that case the workaround probably makes sense, yeah

PR #1363, opened 2021-05-06
exclude deploy artifacts from create-svelte

Rich-Harris:
not a huge deal since we release via an action, but when i built this package locally it included the contents of `.cloudflare`. No harm in guarding against that happening during a manual release

### Tests

### Changesets

### DIFF
--- a/.changeset/tame-kangaroos-rush.md
+++ b/.changeset/tame-kangaroos-rush.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Exclude deploy artifacts from create-svelte package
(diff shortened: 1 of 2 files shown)


Conduitry:
Is this the `.gitignore` that ends up in users' projects, or is it used in deciding what gets published in the create-svelte package itself? I know there was something weird about `npm publish` renaming a file transparently.

If it's the latter, I'd probably prefer an explicit whitelist approach (with `pkg.files`) for what ends up in the package.

Rich-Harris:
It's the latter. Whitelist is probably smarter, but would require some implementing, and I figured this quick fix is better than nothing in the meantime

Conduitry (approved):
:+1: That's fair. Sounds good to me.

Issue #1365, opened 2021-05-06
Make the Node adapter respect the HOST environment variable


tsanth:
**Is your feature request related to a problem? Please describe.**
I would like to be able to direct the Node adapter to serve on a specific hostname or IP.

**Describe the solution you'd like**
I would like to be able to set the `HOST` environment variable and have that respected by the Node adapter.

**Describe alternatives you've considered**
I studiously try to avoid hacking on things in `node-modules`, so a PR makes much more sense to me (and probably everyone else).

**How important is this feature to you?**
This feels like a feature which many people would find useful, since I doubt people want to be restricted to having the Node adapter only listen on 0.0.0.0.

**Additional context**
The required changes amount to roughly 2 lines in `.../packages/adapter-node/src/index.js`.

PR #1366, opened 2021-05-06
Makes Node adapter respect HOST (#1365)

tsanth:
By default, the HOST variable is set to 127.0.0.1, for safety.

I lack the experience to set up a test runnable via `pnpm test`, but I've tested this as follows:
- I add in my PR.
- `npm run build`
- `HOST=192.168.1.64 npm start`
- Verify that I can reach my server only from its IPv4 address
- Back out my PR
- `npm run build`
- `HOST=192.168.1.64 npm start`
- Verify that I can reach my server from all of 1) localhost, 2) IPv4,  and 3) IPv6

### Tests

### Changesets

### DIFF
--- a/packages/adapter-node/src/index.js
+++ b/packages/adapter-node/src/index.js
@@ -2,9 +2,9 @@ import { createServer } from './server';
 /*eslint import/no-unresolved: [2, { ignore: ['\.\/app\.js$'] }]*/
 import * as app from './app.js';
 
-const { PORT = 3000 } = process.env; // TODO configure via svelte.config.js
+const { HOST='127.0.0.1', PORT = 3000 } = process.env; // TODO configure via svelte.config.js
 
-const instance = createServer({ render: app.render }).listen(PORT, (err) => {
+const instance = createServer({ render: app.render }).listen(PORT, HOST, (err) => {
 	if (err) {
 		console.log('error', err);
 	} else {


Rich-Harris:
Thank you — is there a reason to default to `127.0.0.1`? Given that this is intended for production servers, surely we _want_ the address to be unspecified?

tsanth:
My typical workflow is to use a local nginx instance to reverse proxy requests to Node (to simplify SSL setup, among things), so I set the default assuming that use case.

Since this is for production servers, I've modifed the default to be `0.0.0.0` instead, so as to capture all interfaces.

Rich-Harris:
Thank you — I took the liberty of adding a changeset and docs

packages/adapter-node/src/index.js
```diff
@@ -2,9 +2,9 @@ import { createServer } from './server';
 /*eslint import/no-unresolved: [2, { ignore: ['\.\/app\.js$'] }]*/
 import * as app from './app.js';
 
-const { PORT = 3000 } = process.env; // TODO configure via svelte.config.js
+const { HOST = '0.0.0.0', PORT = 3000 } = process.env; // TODO configure via svelte.config.js
```
Conduitry, line 5: Is this TODO still a TODO? Would the env vars take precedence?

Rich-Harris: it's still a TODO insofar as you can't specify a default port/host via the adapter config. Is that something worth having, if we have environment variables?

Conduitry: Well yeah what I meant is: is this something we still want to do? I don't have especially strong opinions about that. I assume it would be additional args you pass to the adapter factory function and not somewhere else in the config.

Rich-Harris: it would be, yeah. tbh i'm inclined to just remove the TODO comment — it's less confusing if there's only one way to set non-default values, and we need environment variables regardless so that it's configurable at runtime

Conduitry (approved):

PR #1370, opened 2021-05-06
Rename $layout to __layout, etc

### LINKED ISSUE #1149: Change $layout.svelte to @layout.svelte and [param].svelte to {param}.svelte?
Ok hear me out.

I think we closed https://github.com/sveltejs/kit/issues/894 prematurely; it _does_ make sense to use a character that doesn't require special handling, if possible. It's awkward to type `src/routes/$layout.svelte` on the command line — you get to the `$` and either have to reach for the backslash, or curse your lack of forethought in not typing `'src` (single quote only, no doubles!) instead of `src`.

 Even if you don't find yourself typing/tab-completing route filenames on the command line frequently, you might find yourself attempting to cmd-click on the filename in your editor's integrated terminal (e.g. in the output from `git diff` or `tsc` or whatever) in order to edit the file. At least in VSCode, that doesn't work for filenames containing `$layout`, probably because something somewhere is replacing it with the `layout` environment variable.

These problems go away with `@`.

Related to the cmd-click problem: VSCode doesn't recognise this as a file — if you hover it, it will only highlight the file up to the `[` character:

image

image

This, on the other hand...

image

...works perfectly:

image

I should note that this behaviour appears to be specific to VSCode; iTerm reverses the treatment of `[bar]` and `{bar}`. But I think it's probably more important that cmd-click works in integrated editor terminals, and VSCode is the most popular one of the bunch.

Aesthetically, `@layout.svelte` and `@error.svelte` seem fine, and I actually prefer `{param}` to `[param]`, not least because it's more aligned with the syntax of Svelte itself.

Rich-Harris:
Closes #1149.

* [x] disallow files/directories with leading `__`, so that we can add more special files in future without breakage
* [x] error on `$layout` etc
* [ ] ~~create a utility for renaming files? `npx @sveltejs/kit-rename-files` or something?~~

### Tests

### Changesets


Conduitry (changes requested):
Amusingly, I think the `__tests__.js` files are now tripping the rule about not being able to start a file with two underscores unless it's one of the known ones.

Rich-Harris:
whoops

AFK for a bit, will fix later if no-one beats me to it

documentation/migrating/04-pages.md
```diff
@@ -4,7 +4,7 @@ title: Pages and layouts
 
 ### Renamed files
 
-Your custom error page component should be renamed from `_error.svelte` to `$error.svelte`. Any `_layout.svelte` files should likewise be renamed `$layout.svelte`. The `_` prefix now exclusively denotes your _own_ 'private' components and modules, as opposed to those with a special meaning to SvelteKit.
+Your custom error page component should be renamed from `_error.svelte` to `__error.svelte`. Any `_layout.svelte` files should likewise be renamed `__layout.svelte`. The `_` prefix now exclusively denotes your _own_ 'private' components and modules, as opposed to those with a special meaning to SvelteKit.
```
Conduitry, line 7: This reads a little bit weirdly now. It should probably indicate that _single_-`_` prefixes are for your own private files.

Rich-Harris: good catch, tweaked

Conduitry (approved):

frederikhors:
This change makes working with Go a nightmare because of this: https://github.com/golang/go/issues/43854.

Can we add a config option for this, @Rich-Harris?

feedmeapples:
@Rich-Harris any updates on the config option?

rohfle:
The `all:` path prefix for go:embed has also been recently merged in Go (https://go-review.googlesource.com/c/go/+/359413/)

```
cmd/go: add //go:embed all:pattern

[...]

After long discussions on #42328 and #43854, we decided to keep
the behavior of excluding . and _ files by default, but to allow the pattern
prefix 'all:' to override this default. This CL implements that change.

Note that paths like .git and com1 are still excluded, as they must be,
since they will never be packed into a module.
```

frederikhors:
@rohfle I think it will be available from 1.18, right?

rohfle:
@frederikhors the `all:` path prefix is merged in 1.18beta1

Issue #1136, opened 2021-04-19
ETag generation throws on binary responses since 1.0.0-next.82


intrikate:
**Describe the bug**
Since `1.0.0-next.82`, attempting to set a `Content-Type` header for certain binary responses in endpoint handlers will throw an exception: `str.charCodeAt is not a function`. Responding with no type or with explicit `application/json` won’t throw. Downgrading to `next.81` fixes the issue. Upgrading to `next.83` does not.

Update: This appears to be a regression introduced with the new ETag generation in #1117.

**To Reproduce**
Make an endpoint that responds with e.g. a JPEG with the `Content-Type` (or `content-type`) header set to `image/jpeg`.

See repo: https://github.com/intrikate/sveltekit-jpeg-endpoint-issue

**Expected behavior**
`Content-Type` set, the body sent, no exceptions thrown. Essentially what `1.0.0-next.81` does.

**Stacktraces**
<details>
  <summary>Stack trace</summary>

```
str.charCodeAt is not a function
TypeError: str.charCodeAt is not a function
    at hash (file:///Users/leek/Code/Playground/sveltekit-jpeg-issue/node_modules/@sveltejs/kit/dist/ssr.js:1509:37)
    at render (file:///Users/leek/Code/Playground/sveltekit-jpeg-issue/node_modules/@sveltejs/kit/dist/ssr.js:1473:26)
    at async ssr (file:///Users/leek/Code/Playground/sveltekit-jpeg-issue/node_modules/@sveltejs/kit/dist/ssr.js:1454:10)
    at async Immediate.<anonymous> (file:///Users/leek/Code/Playground/sveltekit-jpeg-issue/node_modules/@sveltejs/kit/dist/chunks/index.js:3297:23)
```
</details>

**Information about your SvelteKit Installation:**

<details>
  <summary>Diagnostics</summary>

```
System:
    OS: macOS 11.2.3
    CPU: (4) x64 Intel(R) Core(TM) i5-4288U CPU @ 2.60GHz
    Memory: 46.50 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.0 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 6.14.11 - /usr/local/bin/npm
  Browsers:
    Chrome: 89.0.4389.128
    Firefox Developer Edition: 88.0
    Safari: 14.0.3
  npmPackages:
[...]

PR #1381, opened 2021-05-07
generate etags for binary responses

Rich-Harris:
#1136. seeing some test failures that I can't explain so I'm not going to merge this, instead I'm going to reimplement it more carefully. Just leaving this here for reference.

it involves a few changes:

* separating `ServerResponse` from the new `EndpointOutput` type (both types are public)
* `ServerResponse` body can be `string` or `Uint8Array` (the latter only/always when `content-type` is `application/octet-stream`)
* `EndpointOutput` body can also be a `JSONValue`
* `handle` response must be `ServerResponse` (i.e. `body` must be `string` or `Uint8Array` here, not `JSONValue`)
* `rawBody` can be `string` or `Uint8Array` (less sure about this bit)

Aside: brilliantly, in Node, when you call `res.end(uint8Array)` you get the following error:

> The "chunk" argument must be of type string or an instance of Buffer or Uint8Array. Received an instance of Uint8Array

Slow clap. It turns out you _can_ do `res.write(uint8Array), res.end()` though.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -98,7 +98,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 		if (seen.has(path)) return;
 		seen.add(path);
 
-		/** @type {Map<string, import('types/endpoint').ServerResponse>} */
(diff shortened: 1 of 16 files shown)


Rich-Harris:
closing in favour of #1382

PR #1385, opened 2021-05-08
account for POST bodies when serializing fetches

### LINKED ISSUE #633: Post request via fetch gets overriden/replaced in `load()`
**Describe the bug**
Two parallel (?) fetch POST requests (one from $layout.svelte, and one from index.svelte) with different contents end up having the same response on the client side.

**Logs**
nothing interesting happening in the logs

**To Reproduce**
I set up a repro repository here https://github.com/neesh-studio/sk-fetch-override-repro which loads two festival objects from a graphQL api.
One Festival is called "First Festival", the other "Another Festival". 

The SSR implementation seems to work as there is a short flash of the "Another Festival" title on page refresh. But it get's replaced on the client with "First Festival".
What i get to see: 
grafik

**Expected behavior**
what I expect to see:
grafik

I expect the fetch api to work the same on server and client and the client not overriding another (POST) request. (A quick test shows the behaviour seems fine for GET requests)

The problem goes away if I uncomment line 21 in api.ts
```js
  if (typeof window !== "undefined") fetchApi = fetch;
```

- The output of `npx envinfo --system --npmPackages svelte,@sveltejs/kit --binaries --browsers`
```bash
System:
    OS: Linux 5.4 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i5-3320M CPU @ 2.60GHz
    Memory: 10.16 GB / 12.35 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 14.15.5 - ~/.nvm/versions/node/v14.15.5/bin/node
    npm: 6.14.11 - ~/.nvm/versions/node/v14.15.5/bin/npm
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.58 
    svelte: ^3.29.0 => 3.35.0 
```

- Your browser
Chrome: Version 89.0.4389.90 (64-Bit)

- Your adapter (e.g. Node, static, Vercel, Begin, etc...)
dev server

**Severity**
Minor, there is a workaround (as described) but I wanted to post the issue early in case this is somehow connected to #591

Rich-Harris:
fixes #633. Figured the pragmatic choice is to ignore requests with a non-string body (as in, they don't get serialized; those fetches are repeated in the client) since other kinds are unlikely in `load` (presumably the main situation where this is useful is with GraphQL, which uses a JSON payload) and it would be undesirable to buffer streams, etc. 

Handling other kinds of bodies is a TODO in any case: https://github.com/sveltejs/kit/blob/7679399ff965760bdf7f44a100087b597d753d9d/packages/kit/src/runtime/server/page/load_node.js#L148-L159

In fact, I think for now it might be best to throw an error if non-string bodies are sent in `fetch` in `load`. Better than silently failing. Though either way it should affect roughly zero people.

### Tests

### Changesets

### DIFF
--- a/.changeset/young-apricots-dance.md
+++ b/.changeset/young-apricots-dance.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Account for POST bodies when serializing fetches
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -1,5 +1,6 @@
 import { writable } from 'svelte/store';
-import { normalize } from '../load';
+import { hash } from '../hash.js';
+import { normalize } from '../load.js';
 
 /** @param {any} value */
 function page_store(value) {
@@ -37,7 +38,14 @@ function page_store(value) {
  */
 function initial_fetch(resource, opts) {
 	const url = typeof resource === 'string' ? resource : resource.url;
-	const script = document.querySelector(`script[type="svelte-data"][url="${url}"]`);
+
+	let selector = `script[type="svelte-data"][url="${url}"]`;
+
+	if (opts && typeof opts.body === 'string') {
+		selector += `[body="${hash(opts.body)}"]`;
+	}
+
+	const script = document.querySelector(selector);
 	if (script) {
 		const { body, ...init } = JSON.parse(script.textContent);
 		return Promise.resolve(new Response(body, init));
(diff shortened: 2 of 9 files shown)


benmccann:
Additionally supporting JSON doesn't seem like it should be too hard and might be worth it if it helps add GraphQL support. Though presumably those users could just serialize the JSON to a string themselves before calling `fetch`

Rich-Harris:
JSON _is_ a string. You don't post `body: {...}`, you post `body: JSON.stringify({...})`

ion1:
I wonder if hash collisions will end up being a problem? They will not happen often, but that might just make it more of a pain to debug once it eventually happens.

For example, both `Hello there` and `General Kenevefiyimubi` hash to `1xhevwp` using the function.

PR #1406, opened 2021-05-10
add serverUrl to fetch options

stalkerg:
Hello, a very long time of people including me asked about support different URLs for the backend and frontend for the same data.  
It's very common if server-side code should communicate with the internal backend directly by the internal path. 
Hopefully, it's now possible with SvelteKit but in the current implementation, it produces double load on the frontend side.
This PR adds the `serverUrl` property to `fetch` options, it helps SvelteKit find a resource on the browser and avoid double load. 
Maybe, we can add this as the third argument for the `fetch` method but I suppose is too radical. 
Thanks.

### DIFF
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -34,10 +34,15 @@ function page_store(value) {
 
 /**
  * @param {RequestInfo} resource
- * @param {RequestInit} opts
+ * @param {RequestInit & {serverUrl: string}} opts
  */
 function initial_fetch(resource, opts) {
-	const url = typeof resource === 'string' ? resource : resource.url;
+	let url = resource;
+	if (opts && opts.serverUrl) {
+		url = opts.serverUrl;
+	} else if (typeof resource !== 'string') {
+		url = resource.url;
+	}
 
 	let selector = `script[type="svelte-data"][url="${url}"]`;
 
--- a/packages/kit/test/apps/basics/src/routes/load/fetch-server-url-server.json.js
+++ b/packages/kit/test/apps/basics/src/routes/load/fetch-server-url-server.json.js
@@ -1,0 +1,7 @@
+export function get() {
+	return {
+		body: {
+			answer: 42
+		}
+	};
+}
--- a/packages/kit/test/apps/basics/src/routes/load/fetch-server-url.json.js
+++ b/packages/kit/test/apps/basics/src/routes/load/fetch-server-url.json.js
@@ -1,0 +1,7 @@
+export function get() {
+	return {
+		body: {
+			answer: 42
+		}
+	};
+}
--- a/packages/kit/test/apps/basics/src/routes/load/fetch-server-url.svelte
+++ b/packages/kit/test/apps/basics/src/routes/load/fetch-server-url.svelte
@@ -1,0 +1,23 @@
+<script context="module">
+	import { browser } from '$app/env';
+
+	/** @type {import('@sveltejs/kit').Load} */
+	export async function load({ fetch }) {
+		const res = await fetch(
+			browser ? '/load/fetch-server-url.json' : '/load/fetch-server-url-server.json',
+			{
+				serverUrl: '/load/fetch-server-url-server.json'
+			}
+		);
+		return {
+			props: await res.json()
+		};
+	}
+</script>
+
+<script>
+	/** @type {number} */
+	export let answer;
+</script>
+
+<h1>{answer}</h1>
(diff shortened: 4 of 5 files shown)


stalkerg:
Very good explanation here https://github.com/sveltejs/sapper/issues/489 I can add only one point:
I want to avoid nginx overhead (especially if they have tons of alias and rewrites) for in-server communication.
Basically, you can have sveltekit server, backend server and nginx server, in that case, communicate inside server will be very expensive like:
`browser->nginx server(10.0.0.1)->sveltekit server(10.0.0.3)->nginx server(10.0.0.1)->backend server(10.0.0.2) `

instead just:
`browser->nginx server(10.0.0.1)->sveltekit server(10.0.0.3)->backend server(10.0.0.2) `

Kapsonfire-DE:
> @Kapsonfire-DE sure but it will produce a second fetch on browser-side what I trying to avoid by this PR.

but is using fetch('/my/end/point') really creating a full network request? since it could just call the get/post whatever internally

stalkerg:
> but is using fetch('/my/end/point') really creating a full network request? since it could just call the get/post whatever internally

not sure I understand what do you mean. Also just using relative paths not an option because the frontend endpoint usually has some prefixes. Like frontend is `fetch('my.app/api/add')` but on server side is `fetch('localhost:9999:/add')`

Kapsonfire-DE:
if this is really the case, the wrapper around the fetch should be changed to not make a network request and call the endpoint internally by import

Kapsonfire-DE:
ah you are talking about a external api endpoint server? then you could use the svelte-kit endpoints as proxy handler

stalkerg:
@Kapsonfire-DE I don't want to use an extra proxy, it's the wrong architecture design that killed performance. And this issue mostly solved in SvelteKit. I and other people asked multiple times about it.

Kapsonfire-DE:
in which case it kills the performance?
the server has to call the http request on SSR anyway, on CSR it connects to the svelte-kit endpoint which just use a localhost connection.

and to optimize this you can use unix sockets in your endpoints to connect to backend server

stalkerg:
no, it's a different case, CSR asking my Python backed directly (over nginx).

stalkerg:
Because I have already had public API, I don't want to provide an extra "proxy" API by SvelkteKit. Because:
1. It's still overhead and extra hope
2. Can't find an easy way to make a wildcard proxy.
3. Security (I don't want to worry about API invocation for SvelteKit proxy)
4. I have analytics for my API, and the browser should use these specific URLs things only.

Rich-Harris:
If the problem is 'we want to intercept requests to an external API server inside `fetch` on the server and modify them' then that seems like a solvable problem. We certainly wouldn't want to add non-standard appendages to the `fetch` API, partly because it's confusing but mostly because it would be repetitious; you would need to include that logic in every `load` function that used the API in question.

Something like this (name subject to bikeshedding):

```js
const PUBLIC_API = 'https://api.example.com/';
const PRIVATE_API = 'https://10.0.0.2:4567/';

/**
 * @param {RequestInit & { url: string }} opts
 * @returns {RequestInit & { url: string }}
 **/
export function mutateFetch(opts) {
  if (opts.url.startsWith(PUBLIC_API) {
    opts.url = opts.url.replace(PUBLIC_API, PRIVATE_API);
  }
}
```

Worth noting that in the case where you're proxying `/api/` requests to an external server in nginx you can easily do this in `handle` today:

```js
export async function handle({ request, render }) {
  if (request.path.startsWith('/api/') {
    const res = await fetch(`${PRIVATE_API}/${request.path.slice(5)}`);
    return {
      status: res.status,
      headers: to_object(res.headers),
      body: await res.text()
    };
  }

  return render(request);
}
```

(`load` functions call `handle` directly, there's no intermediate network requests.)

stalkerg:
> Worth noting that in the case where you're proxying /api/ requests to an external server in nginx you can easily do this in handle today

Thanks, I didn't think about handle in that key. 

> If the problem is 'we want to intercept requests to an external API server inside fetch on the server and modify them' then that seems like a solvable problem. 

I suppose, no (but probably we looking from different points of view on it). We need 'provide resource id for fetch separately from URL'.  Currently, we are using URLs as a resource ID strongly. Thanks, @Kapsonfire-DE fro better understanding the original issue. 
I agree, extend the default `fetch` interface is a confusing and bad practice but it's the most simple solution here. Also, serverUrl probably we can replace by ressourceId or ressourceUrl. 

> Something like this (name subject to bikeshedding):

hmm it's can work but it's shouldn't be a just URL it should be for recourceId to avoid second load in browser for hydration.

Conduitry:
> it's shouldn't be a just URL it should be for recourceId to avoid second load in browser for hydration

You'd call `this.fetch` with the URL that should be used on the browser (which would also be the resource id that it would be cached with) and then on the server only you would transform the URL into what should be used on the server.

Rich-Harris:
I think the first step would be to open an issue, so that we can figure out together what the solution looks like before implementing it

PR #1408, opened 2021-05-10
Use rendered CSS for AMP pages, rather than emitted CSS

### LINKED ISSUE #808: Use rendered rather than emitted CSS in AMP mode
AMP pages have a strict (and somewhat arbitrary) limit on the amount of CSS a page can include. They also provide a guarantee that doesn't exist in non-AMP mode, namely that if `condition` is `false` at SSR time...

```svelte
<script>
  import Foo from '$lib/components/Foo.svelte';

  export let condition;
</script>

{#if condition}
  <Foo/>
{/if}
```

...we can safely omit the styles emitted by `Foo.svelte`.

It would be nice if we could take advantage of that and only include styles for components that were rendered, to increase the likelihood of staying beneath the 75kb ceiling.

* Configure the Svelte plugin to not emit CSS in AMP mode
* Use `rendered.css` instead of inspecting the module graph

It depends on the outcome of the research we plan to conduct on how much traffic our app would lose if we told AMP to bugger off once and for all.

Rich-Harris:
Fixes #808. Will need a release of `vite-plugin-svelte` with [this change](https://github.com/sveltejs/vite-plugin-svelte/pull/41) before the tests will pass.

### Tests

### Changesets

### DIFF
--- a/.changeset/happy-nails-draw.md
+++ b/.changeset/happy-nails-draw.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Use rendered CSS for AMP pages
--- a/packages/kit/package.json
+++ b/packages/kit/package.json
@@ -3,6 +3,7 @@
 	"version": "1.0.0-next.104",
 	"type": "module",
 	"dependencies": {
+		"@sveltejs/vite-plugin-svelte": "^1.0.0-next.10",
 		"cheap-watch": "^1.0.3",
 		"sade": "^1.7.4",
 		"vite": "^2.2.4"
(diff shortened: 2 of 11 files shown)


Rich-Harris:
excellent, thanks so much!

Rich-Harris:
That wasn't intentional — reverted. The basic thinking is that if something is bundleable, it gets bundled, since that way Kit is faster to install. But I'm not sure if it's been applied particularly rigorously or if we should even be thinking that way

benmccann:
I don't suppose it's possible that AMP support will no longer be needed now that Google is showing non-AMP pages in the search carousel? :grin: https://www.lafoo.com/the-end-of-amp/

Rich-Harris:
maybe for SvelteKit 2. time will tell

PR #1412, opened 2021-05-10
Use a consistent version of node-fetch

benmccann:
I noticed that `adapter-node` was using a different version of `node-fetch` than `kit` is and figured it probably makes sense for them to use a consistent version


Rich-Harris:
`adapter-node` isn't actually using `node-fetch` at all, except in its tests, so it doesn't really matter what version it's using. but bumping it is harmless

Conduitry:
Is there a pnpm-workspaces-y way to make it so that this doesn't need to be updated in two places? So that both Kit and the Node adapter (and anything else that needs node-fetch) all always use the same copy?

PR #1416, opened 2021-05-11
Only cache content in /_app

bjon:
Without this change, everything in the static folder will get the Cache-Control of `public,max-age=31536000,immutable`.
This works well if you use the assets in html/svelte files (cache busting), but not so good for favicon.ico, robots.txt, etc.

### Tests
- [X] Run the tests with `pnpm test` and lint the project with `pnpm lint`

### Changesets
- [X] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts


bjon:
@Rich-Harris Also Sirv options would be great to configure. For example `brotli` and `gzip`, to support precompressed files

packages/adapter-node/src/server.js
```diff
@@ -27,6 +28,12 @@ export function createServer({ render }) {
 		? mutable(paths.prerendered)
 		: noop_handler;
 
+	const static_handler = fs.existsSync(paths.static)
+		? sirv(paths.static, {
+				maxAge: 3600
```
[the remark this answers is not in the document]
bjon: Yes, without maxAge would be fine as well. But without maxAge (or s-maxage) many CDNs wont cache.

bjon:
@Rich-Harris Should I closed this? Or should I revert the static folder and fix cache in assets folder instead?

bjon:
If we should see this as a patch, until a new, more configurable PR, is available.

packages/adapter-node/src/server.js
```diff
@@ -29,8 +29,11 @@ export function createServer({ render }) {
 
 	const assets_handler = fs.existsSync(paths.assets)
 		? sirv(paths.assets, {
-				maxAge: 31536000,
-				immutable: true
+				setHeaders: (res, pathname, stats) => {
+					if (pathname.startsWith('/_app/')) {
```
benmccann, line 33: I think this should use the configured value instead of it being hardcoded. See Rich's earlier comment where `appDir` is provided by the main configuration: https://kit.svelte.dev/docs#configuration-appdir

benmccann (commented):
lgtm except for the directory being hardcoded

I think I actually preferred the earlier iteration where there were two separate directories, but I will defer to Rich and am fine with this version. I do wonder if this also occurs on any of the other adapters or if they are free of this issue

packages/adapter-node/src/server.js
bjon: @benmccann  Is there a variable I can use instead?

benmccann: It looks like this is basically the request made in https://github.com/sveltejs/kit/issues/1435. Perhaps @Rich-Harris has thoughts on how to accomplish this?

benmccann: There's a PR for this now: https://github.com/sveltejs/kit/pull/1559

benmccann: @bjon https://github.com/sveltejs/kit/pull/1559 has been merged. If you rebase this PR hopefully you can you the new variable to address this

bjon: @benmccann I'm not sure how I should access the config object in `adapt`, from `src/server.js`?
Should I store it to disk, and read that from `src/server.js`?

benmccann: Hmm. Can you use esbuild defines? https://esbuild.github.io/api/#define

bjon: Never used esbuild, but I can look into that.

bjon: If I'm not mistaken, adapter-node isn't using esbuild at all

bjon: Ok, now I think I get it. `adapter-node` builds `files/server.js` with Rollup. Then `adapt()` copies it to `/build`. I don't really have a good solution for this, with the current build system.

benmccann: I think that @Conduitry was looking into converting it to use esbuild

Conduitry: I was, but I sort of ran into a wall because I don't know how to use esbuild. In particular, I was having trouble configuring which dependencies get bundled and which remain external dependencies. So if other folks are thinking about taking a look at this, it would be appreciated. Don't worry about duplicated effort, because I'm not on the verge of opening a PR for that or anything.

benmccann: The PR for esbuild has been opened at https://github.com/sveltejs/kit/pull/1648

bjon: @benmccann this should be fixed now

bjon:
There are errors when building with latest adapter-node.
See https://github.com/sveltejs/kit/issues/1667

Should i add `target: 'es2020'` to esbuild? 
Other adapters are doing that, and it works for adapter-node as well.

packages/adapter-node/src/server.js
```diff
@@ -29,8 +29,12 @@ export function createServer({ render }) {
 
 	const assets_handler = fs.existsSync(paths.assets)
 		? sirv(paths.assets, {
-				maxAge: 31536000,
-				immutable: true,
+				setHeaders: (res, pathname, stats) => {
+					// eslint-disable-next-line no-undef
+					if (pathname.startsWith(`/${esbuild_app_dir}/`)) {
```
benmccann, line 34: This will require that the user not prefix or suffix `config.kit.appDir` with `/`. I think it'd be good to add either some sanitation or validation. E.g. we could automatically strip that prefix / suffix character. Or we could assert its value as we're doing for the base path in https://github.com/sveltejs/kit/commit/dc56d3cef4ec0aca3d074757c4a15d1f640c5c56

benmccann (commented):
thanks for sticking with this one. really close now!

packages/adapter-node/src/server.js
bjon: @benmccann added check for leading slash + app_dir as root

bjon: and app_dir can not be empty string

bjon: But maybe these checks are unnecessary? SvelteKit doesn't work with `app_dir: "/"` or `app_dir: "/my_app_dir"` anyway, in `svelte.config.js`.

bjon: That should probably be fixed in another PR.

benmccann: It still looks like it will not work if someone specifies `config.kit.appDir` with a trailing slash

bjon: aaa stupid me...

bjon: @benmccann fixed

PR #1430, opened 2021-05-12
docs: sync up types

ignatiusmb:
Some might still be missed or not as intended, especially `body` types.

Also, is `type Request` intended to be identical? seems the same to me, perhaps we could just put it in either `routing` or `hooks` to remove duplication and additional maintenance.

### DIFF
--- a/documentation/docs/01-routing.md
+++ b/documentation/docs/01-routing.md
@@ -48,27 +48,27 @@ A file or directory can have multiple dynamic parts, like `[id]-[category].svelt
 Endpoints are modules written in `.js` (or `.ts`) files that export functions corresponding to HTTP methods. For example, our hypothetical blog page, `/blog/cool-article`, might request data from `/blog/cool-article.json`, which could be represented by a `src/routes/blog/[slug].json.js` endpoint:
 
 ```ts
-type Request<Context = any> = {
-	host: string;
+type Request<Locals = Record<string, any>> = {
 	method: 'GET';
-	headers: Record<string, string>;
+	host: string;
+	headers: Headers;
 	path: string;
-	params: Record<string, string | string[]>;
+	params: Record<string, string>;
 	query: URLSearchParams;
 	rawBody: string | Uint8Array;
 	body: string | Uint8Array | JSONValue;
-	locals: Record<string, any>; // see below
+	locals: Locals; // see hooks handle
 };
 
-type Response = {
+type EndpointOutput = {
 	status?: number;
-	headers?: Record<string, string>;
-	body?: any;
+	headers?: Headers;
+	body?: string | Uint8Array | JSONValue;
 };
 
-type RequestHandler<Context = any> = {
-	(request: Request<Context>) => Response | Promise<Response>;
-}
+type RequestHandler<Locals = Record<string, any>> = (
+	request: Request<Context>
+) => void | EndpointOutput | Promise<EndpointOutput>;
 ```
 
 ```js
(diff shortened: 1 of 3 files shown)


documentation/docs/01-routing.md
```diff
@@ -48,27 +48,27 @@ A file or directory can have multiple dynamic parts, like `[id]-[category].svelt
...
-	locals: Record<string, any>; // see below
+	locals: Locals; // see hooks handle
 };
 
-type Response = {
+type EndpointOutput = {
 	status?: number;
-	headers?: Record<string, string>;
-	body?: any;
+	headers?: Headers;
```
benmccann, line 65: it might be useful to include the type of `Header`

documentation/docs/01-routing.md
```diff
@@ -48,27 +48,27 @@ A file or directory can have multiple dynamic parts, like `[id]-[category].svelt
 Endpoints are modules written in `.js` (or `.ts`) files that export functions corresponding to HTTP methods. For example, our hypothetical blog page, `/blog/cool-article`, might request data from `/blog/cool-article.json`, which could be represented by a `src/routes/blog/[slug].json.js` endpoint:
 
 ```ts
-type Request<Context = any> = {
-	host: string;
+type Request<Locals = Record<string, any>> = {
 	method: 'GET';
```
benmccann, line 52: I think this should be `string`

documentation/docs/01-routing.md
```diff
@@ -48,27 +48,27 @@ A file or directory can have multiple dynamic parts, like `[id]-[category].svelt
...
-	headers: Record<string, string>;
+	host: string;
+	headers: Headers;
 	path: string;
-	params: Record<string, string | string[]>;
+	params: Record<string, string>;
 	query: URLSearchParams;
 	rawBody: string | Uint8Array;
 	body: string | Uint8Array | JSONValue;
```
benmccann, line 59: perhaps `ParameterizedBody<Body>`

documentation/docs/01-routing.md
```diff
@@ -48,27 +48,27 @@ A file or directory can have multiple dynamic parts, like `[id]-[category].svelt
...
-	locals: Record<string, any>; // see below
+	locals: Locals; // see hooks handle
```
benmccann, line 60: "see hooks handle" isn't very clear. see it for info about what? maybe something more like "populated by hooks handle"

documentation/docs/04-hooks.md
```diff
@@ -23,19 +23,19 @@ type Request<Locals = Record<string, any>> = {
 	params: Record<string, string>;
 	query: URLSearchParams;
 	rawBody: string | Uint8Array;
-	body: string | Uint8Array | ReadOnlyFormData | JSONValue;
-	locals: Locals;
+	body: string | Uint8Array | JSONValue;
```
benmccann, line 26: same comment for this line and the next as above

benmccann:
I guess I overlooked this PR. I sent a PR to do the same thing earlier this morning: https://github.com/sveltejs/kit/pull/1495. I left a few comments about some of the things that are different between ours

documentation/docs/01-routing.md
ignatiusmb: In that case, do we also include `ParameterizedBody` and `JSONValue` types?

documentation/docs/01-routing.md
ignatiusmb: Since this is initially specified as `'GET'`, should we just limit it to only methods in https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods? Will there be a case where `method` isn't any of these or possibly a custom one?

Though, this is probably better implemented in another PR together with the types changes.

documentation/docs/01-routing.md
ignatiusmb: That might be better 👍

documentation/docs/01-routing.md
benmccann: I didn't because those values were more complicated. If we include `ParameterizedBody` then we'd need to include `ReadOnlyFormData`, `BaseBody`, and so on. That didn't seem practical

documentation/docs/01-routing.md
benmccann: that sounds somewhat reasonable, but yeah we'd have to change the implementation as well and agree that'd be better done as a separate PR

ignatiusmb:
Thanks for the review @benmccann 👍

Issue #1049, opened 2021-04-16
Routing: Spread segments can match in the middle of another segment


rmunn:
**Describe the bug**
Routes with `[...spread]` segments can match in the middle of another segment, without needing a `/` to match. In my repro linked below, I have a `src/routes/foo/[...rest]/index.svelte` route. This matches `/foo/one/two` as expected, with `rest` having the value `'one/two'`. But it *also* matches `/food/one/two`, with `rest` getting assigned the value `'d/one/two'`. I believe that the latter match is a bug.

The other buggy route match in my repro is the route `src/routes/[dynamic]-bar/[...rest].svelte`. This correctly matches `one-bar/two/three`, with `dynamic` being assigned `'one'` and `rest` being assigned `'two/three'`. But it *also* matches `one-bard/two/three`, with `dynamic` being assigned `'one'` and `rest` being assigned `'d/two/three'`. The latter match is, IMHO, a bug.

**Logs**
Server-side console logs produced by `pnpm run dev` and loading the four routes in my repro:
```js
{
  host: 'localhost:3000',
  path: '/foo/one/two',
  query: URLSearchParams {},
  params: { rest: 'one/two' }
}
{
  host: 'localhost:3000',
  path: '/food/one/two',
  query: URLSearchParams {},
  params: { rest: 'd/one/two' }
}
{
  host: 'localhost:3000',
  path: '/one-bar/two/three',
  query: URLSearchParams {},
  params: { dynamic: 'one', rest: 'two/three' }
}
{
  host: 'localhost:3000',
  path: '/one-bard/two/three',
  query: URLSearchParams {},
  params: { dynamic: 'one', rest: 'd/two/three' }
}
```

Browser console logs contain exactly the same info.

**To Reproduce**
Clone https://github.com/rmunn/bug-repro-svelte-kit-routing-spread-segments, run `pnpm i` then `pnpm run dev`, then open http://localhost:3000/ and follow the links to see the issue.

**Expected behavior**
The `/food/one/two` and `/one-bard/two/three` routes should have not matched; instead of matching, they should have returned a 404.

**Information about your SvelteKit Installation:**

<details>
  <summary>Diagnostics</summary>

- The output of `pnpx envinfo --system --npmPackages svelte,@sveltejs/kit,vite --binaries --browsers`

```yaml
  System:
    OS: Linux 5.4 Linux Mint 19.1 (Tessa)
[...]

Issue #918, opened 2021-04-07
Rest Route Params "Stealing" Previous Dynamic Segment Param


CaptainCodeman:
**Describe the bug**
Adding rest route params appears to steal some of previous dynamic segment parameter

This would prevent [the route shown in the docs](https://kit.svelte.dev/docs#routing-advanced-rest-parameters) from working

`/[org]/[repo]/tree/[branch]/[...file]` would only get the first character for the `[branch]` param.

Example:

Given a route structure containing:

/fixed/index.svelte

```
<script>
  import { page } from '$app/stores'
</script>

<p>/fixed</p>
<pre>{JSON.stringify($page, null, '  ')}</pre>
```

/fixed/[dynamic]/index.svelte

```
<script>
  import { page } from '$app/stores'
</script>

<p>/fixed/[dynamic]</p>
<pre>{JSON.stringify($page, null, '  ')}</pre>
```

/fixed/[dynamic]/[...rest].svelte

```
<script>
  import { page } from '$app/stores'
</script>

<p>/fixed/[dynamic]/[...rest]</p>
<pre>{JSON.stringify($page, null, '  ')}</pre>
```

... works as expected until the rest route is reached, for example:

http://localhost:3000/fixed
```
/fixed

{
  "host": "localhost:3000",
  "path": "/fixed",
  "query": {},
  "params": {}
}
```

http://localhost:3000/fixed/one
```
/fixed/[dynamic]

{
  "host": "localhost:3000",
  "path": "/fixed/one",
  "query": {},
  "params": {
    "dynamic": "one"
  }
}
```

http://localhost:3000/fixed/one/two/three
```
/fixed/[dynamic]/[...rest]

{
  "host": "localhost:3000",
  "path": "/fixed/one/two/three",
  "query": {},
  "params": {
    "dynamic": "o",
    "rest": "ne/two/three"
  }
}
```

Note the "dynamic" parameter which is now `o` with the rest parameter stealing the `ne`

I also tried using `/[...rest]/index.svelte` as well (same result)

**To Reproduce**
Example repo: https://github.com/CaptainCodeman/svelte-kit-rest-routing

**Expected behavior**
I expected the final route to produce:

```
{
  "host": "localhost:3000",
  "path": "/fixed/one/two/three",
  "query": {},
  "params": {
    "dynamic": "one",
    "rest": "/two/three"
  }
}
```

**Information about your SvelteKit Installation:**

<details>
  <summary>Diagnostics</summary>

  System:
[...]

PR #1446, opened 2021-05-14
fix: try to address an issue with client-side routing due to regex changes

andyburke:
I've been seeing issues with client-side routing, where it would not match paths. It looks like there have been some recent changes in this area, and if I make the leading `/` in this regex optional, things start working again.

I can't claim to totally understand what's going on here, but this seems to fix the weird issues I was seeing.


benmccann:
The tests seem to be failing with this PR

rmunn:
The failing tests (at least the ones I've looked at) are failing on a test that compares the route output to an expected value. The route output contains, among other things, the pattern used to match the route. So if you change the pattern from `'\\/'` to `'\\/?'`, the tests are going to fail because the expected value of the route pattern still contains `'\\/'`. You'll need to adjust the relevant tests to expect a pattern containing `'\\/?'` before the unit tests will pass.

Also, it would help if you included, either in #1454 or better yer in this PR (written as a set of tests), a list of routes that fail with the code as-is, but pass with your suggested change.

rmunn:
See also #918 and #1049 for some previous routing issues that were fixed by changing the route-matching regex pattern. Please make sure that your change doesn't cause those issues to come back.

andyburke:
@rmunn can you offer any guidance? For me, *every* route client-side is broken without this fix. I can get to the first page through SSR, but then all subsequent links are broken.

I'm not sure how this is working for other people. Maybe I have something screwed up in my setup.

I assume there are regression tests for the other issues mentioned? If I get the tests passing, would that tend to indicate that everything is working as expected?

andyburke:
For clarity: before the change to this line, my setup was working flawlessly. Updating svelte broke things for me. Maybe there was some change in a changelog that I missed that outlined what I needed to do to keep things working with a new version?

rmunn:
@andyburke - Everything's working for me, so I can't really help you without specific examples. Can you create a reproduction with specific examples, please? I'm not going to be able to help unless you show me something that used to work and doesn't, and preferably make it a GitHub repo so I can clone it and reproduce it.

As I said last week:

> Also, it would help if you included, either in #1454 or better yer in this PR (written as a set of tests), a list of routes that fail with the code as-is, but pass with your suggested change.

andyburke:
See https://github.com/sveltejs/kit/issues/1454#issuecomment-847542444

Issue #647, opened 2021-03-24
Generated types for pages/endpoints


Rich-Harris:
Ways that types in SvelteKit apps could be improved:

## Implicit `params` and `props` for `load` functions (update: done)

```svelte
<script context="module">
  /** @type {import('@sveltejs/kit').Load */
  export async function load({ params, fetch }) {
    // `params` automatically typed from filename (e.g. `src/routes/blog/[slug]`)
    const res = await fetch(`/blog/${params.slug}.json`);
    const { post } = await res.json();

    return {
      props: {
        // `post` type automatically inferred from component props
        post
      }
    };
  }
</script>

<script>
  /** @type {BlogPost} */
  export let post;
</script>
```

Similarly, with [shadow endpoints](#3532), it would be good to type `body` based on component props (though this could be tricky since component props combine e.g. `post` and `get` bodies), and also type the `props` input to `load` in cases where it's used.

It might be possible to do something clever with [rootDirs](https://www.typescriptlang.org/tsconfig#rootDirs), or with preprocessors?

## Typed `goto` and `fetch`

As mentioned [below](https://github.com/sveltejs/kit/issues/647#issuecomment-1000085227), it might be possible to type functions like `goto` based on the available routes. It probably gets tricky with relative routes, but that could be a bailout.

## Typed links

This is a little tricky for us, since we use `<a>` instead of `<Link>`, but it would be neat if it was possible to typecheck links somehow.

Rich-Harris:
Relatedly, it would be nice if types for request `context` and load `context` and `session` could be declared in a single place and accounted for in generated types. (Is it confusing that request `context` and load `context` are both called 'context' but are different things?)

Conduitry:
I think it is a bit confusing yeah that load context is a Svelte context and request context is a completely unrelated thing. I don't have a better name for it right now.

Rich-Harris:
> load context is a Svelte context

actually it isn't... it's only available to `load`, not the components themselves. that's a completely different type of context

Rich-Harris:
perhaps `getMetadata(incoming)` and `request.metadata`?

Conduitry:
Oh whoops, yeah. Uh. Maybe they both should be called something other than 'context'? Kit having two concepts with the same name that are different from the Svelte concept with also the same name is definitely confusing.

Rich-Harris:
Suggestions welcome!

Rich-Harris:
```js
export async function load({ milieu }) {
  return {
[...]

PR #1447, opened 2021-05-14
Make it possible to type context and props for LoadInput and LoadOutput

### LINKED ISSUE #1442: Type Context and Props
The inability to define types for context and props is less than ideal for the development process.

Ideally, would love to see the ability to define types for context and props.

We have considered and currently declare types during the variable declaration within the route.

This feature would help our development with potential bugs, autocomplete and type hints, and we think it would be beneficial for many throughout the Svelte ecosystem.

**Additional Context**
https://github.com/sveltejs/kit/blob/master/packages/kit/types/page.d.ts
```ts
export type LoadOutput = {
	status?: number;
	error?: string | Error;
	redirect?: string;
	props?: Record<string, any> | Promise<Record<string, any>>;
	context?: Record<string, any>;
	maxage?: number;
};
```

dummdidumm:
- added generics for Context and Props
- added a helper type "MaybePromise" to not write T | Promise<T> all the time

Fixes #1442

@puchesjr is this what you had in mind? Usage would look like this:

```svelte
<script context="module">
  import { LoadInput, LoadOutput } from '@sveltejs/kit';
  import { YourInputContextTypeDefinition, YourOutputContextTypeDefinition, YourPropsTypeDefinition } from './somewhere'

  export function load(input: LoadInput<YourInputContextTypeDefinition>): Promise<LoadOutput<YourPropsTypeDefinition, YourOutputContextTypeDefinition> {
    // ...
  }
</script>
```

We could bikeshed if the output context generic should appear before the props generic.


packages/kit/types/page.d.ts
```diff
@@ -1,27 +1,42 @@
...
+> = LoadInput<Context> & {
 	status: number;
 	error: Error;
 };
 
-export type LoadOutput = {
+export type LoadOutput<
+	Props extends MaybePromise<Record<string, any>> = MaybePromise<Record<string, any>>,
+	Context extends Record<string, any> = Record<string, any>
+> = {
```
ignatiusmb, line 20: Perhaps the `extends` could be omitted while still keeping the default values here, the users will only interact with `Load` anyway?

packages/kit/types/page.d.ts
```diff
@@ -1,27 +1,42 @@
...
+export type Load<
+	InputContext extends Record<string, any> = Record<string, any>,
+	OutputContext extends Record<string, any> = Record<string, any>,
+	Props extends MaybePromise<Record<string, any>> = MaybePromise<Record<string, any>>
+> = (input: LoadInput<InputContext>) => MaybePromise<LoadOutput<Props, OutputContext>>;
+export type ErrorLoad<
+	InputContext extends Record<string, any> = Record<string, any>,
+	OutputContext extends Record<string, any> = Record<string, any>,
+	Props extends MaybePromise<Record<string, any>> = MaybePromise<Record<string, any>>
+> = (input: ErrorLoadInput<InputContext>) => MaybePromise<LoadOutput<Props, OutputContext>>;
```
ignatiusmb, line 39: I'm not sure about putting `Props` as the last one, it's arguably the most frequently used one, but it might not even matter too since we usually type them in `<script>` tag itself. Still, this might be useful when (and if possible) the exported props from `<script>` could read and check the types from return object of `load` and vice-versa.

3 positional type arguments might also be a handful to type every time, especially when they have no connections to each other. We could also just replace it altogether for named type arguments. That way, any of the types can be passed and the others can be skipped altogether.

ignatiusmb:
Since only `Load` is exposed publicly, perhaps it would look something like

```svelte
<script context="module">
  import { Load } from '@sveltejs/kit';
  import { InputContextType, OutputContextType, PropsType } from './somewhere'

  export const load: Load<InputContextType, OutputContextType, PropsType> = async ({ context }) => {
    // ...
    return {
      props: {},
      context: {}
    }
  }
</script>
```

packages/kit/types/page.d.ts
dummdidumm: To guard against breaking changes it's probably the easiest to only have one argument. I think it was more in the past but was changed to one. I'm also split on what the most frequent used here is. If people don't want to type it all, they can always create custom types like `type MyLoad<Props> = Load<Record<string, any>, Record<string, any>, Props>` .

packages/kit/types/page.d.ts
dummdidumm: I didn't realize these are not exported. I argue since they are indirectly public, they should be exposed, too.

packages/kit/types/page.d.ts
ignatiusmb: If what we want is to only have one argument, then named type argument might be the choice for it. I've tried around with this and works just the same, but this is arguably unconventional in some cases.

```ts
export type Load<
  T extends { input?: object; output?: object; props?: MaybePromise<object> } = any
> = (input: LoadInput<T['input']>) => MaybePromise<LoadOutput<T['props'], T['output']>>;

/* all of this works */
export const load: Load<{
  input: { entry: string };
  output: { hello: string };
  props: { data: { name: string } };
}> = async ({ fetch, context }) => {
  // ...
  return {
    props: {
      data: { name: '' }
    },
    context: { hello: 'world' },
  };
};
// only input
export const load: Load<{ input: { entry: string } }> = async ({ fetch, context }) => {
  // ...
};
// only output
export const load: Load<{ output: { hello: string } }> = async ({ fetch, context }) => {
  // ...
};
// only props
export const load: Load<{ props: { data: { name: string } } }> = async ({ fetch, context }) => {
  // ...
};
```

utkarshkukreti:
Might be slightly off-topic: Are there any thoughts on using `unknown` more in place of `any`, e.g. `Record<string, unknown>`? That way the user has to explicitly specify or cast the type when in strict mode, adding a little more type safety.

packages/kit/types/page.d.ts
ignatiusmb: I can't think of any reason as to why anyone would need/want to use `LoadInput` and `LoadOutput` separately or individually, yet. But, I suppose it's better than having them hack it on their own using `Parameters` and `ReturnType`.

ignatiusmb:
Using `unknown` might cause inconvenience for existing users that don't pass any type arguments. I would expect those using strict mode would just pass in their types to get the type safety.

benmccann:
Is this sort of necessary as a pre-requisite to https://github.com/sveltejs/kit/issues/647? As an end-user, I think what's described there is less verbose. We should think about how these efforts would interact. If the user is typing out the types we might come to a different conclusion on named versus positional arguments than if they're autogenerated

dummdidumm:
It's a prerequisite, yes, and it means we need even more generics for input, namely one for `params`, which makes me lean even more towards named arguments. I think for the autogenerated types it doesn't really matter if they are named or positional, they can deal with both.

puchesjr:
> * added generics for Context and Props
> * added a helper type "MaybePromise" to not write T | Promise all the time
> 
> Fixes #1442
> 
> @puchesjr is this what you had in mind? Usage would look like this:
> 
> ```svelte
> <script context="module">
>   import { LoadInput, LoadOutput } from '@sveltejs/kit';
>   import { YourInputContextTypeDefinition, YourOutputContextTypeDefinition, YourPropsTypeDefinition } from './somewhere'
> 
>   export function load(input: LoadInput<YourInputContextTypeDefinition>): Promise<LoadOutput<YourPropsTypeDefinition, YourOutputContextTypeDefinition> {
>     // ...
>   }
> </script>
> ```
> 
> We could bikeshed if the output context generic should appear before the props generic.

Yes, that is in essence what I was eluding to. Not sure why @benmccann closed my issue so didn't initially see this reply/tag...

dummdidumm:
I changed this to using two generic parameters - one for the input, one for the output - which each have named arguments. I think this feels more natural than having one big generic for both inputs and outputs. It also avoids the "how to name input/output context" question. I had to add a type helper to infer the correct default value in case a parameter of a named generic is not set. It may make sense to add some type tests later on.

Once #984  is decided we have to rename the named arguments introduce in here.

ignatiusmb (approved):
👍 Looks good! I like the type helpers.

benmccann:
I think the test code is type checked since https://github.com/sveltejs/kit/pull/1319 so if we just add types to a couple of the existing tests that should cover it

ignatiusmb:
On second thought, can we wait a bit for this. Quoting from a certain moderator in the TypeScript server

> if you're building an API around manually specified generic params, that's _usually_ a red flag and you might be misusing generics.

I was also wondering why IntelliSense didn't pick up correctly inside the `Load<{ /* here */ }, { /* and here */ }>`

ignatiusmb:
As for the type tests using `svelte-check`, maybe @dummdidumm can chime in on this since I'm uncertain if it can do stuff like `ExpectType` or `ExpectAssignable` tests and such.

dummdidumm:
On `svelte-check` - think of it as `tsc` for Svelte files, so no. That's why I also think dedicated type tests would be good.
On generics and named arguments: I tend to agree with that statement, but the problem remains - there's so much to type at once, so if we don't want 4 arguments, I don't see another way.

ignatiusmb:
After multiple shower thoughts, I think it's best if we export `LoadInput` and `LoadOutput` here, each with their own 2 type arguments, and keep `Load` as is. This way, anyone can choose whether to use just `Load` with named type arguments, or use `LoadInput` and `LoadOutput` individually with the expected positional type argument. Best of both worlds?

benmccann:
@dummdidumm it looks like this one will need to be rebased

> On svelte-check - think of it as tsc for Svelte files, so no.

I'm not sure I understand why that wouldn't help. If we have Svelte files with `load` functions in them then we'd know if they adhere to the types or not. Is there something else you want to test?

dummdidumm:
The current tests are testing the type correctness in a very rough way. I'd like to be more granular about it, and adding e2e tests for that purpose is wrong. Also I can't test that I expect a specific kind of error. That's why there's value in specific type tests.

ignatiusmb:
Can we also add `void` in the return type of `Load` to fix #1606, or at least merge this so others can use the `MaybePromise` helper types.

janosh:
Excuse me if this is a dumb question: Aren't the current `LoadInput`/`LoadOutput` incompatible with `__error.svelte`? Specifically `LoadInput` seems to be missing `error` and `status` and `LoadOutput` needs the ability to return nothing to fall through to the error page.

dummdidumm:
There's `ErrorLoadInput` for the input part. For the output part - isn't the empty object also working? Either way you may be right if `undefined` is a valid return for this, since that's not typed on `Load`

janosh:
> There's `ErrorLoadInput` for the input part.

My bad, I missed that.

A bare `return` statement, I think, better communicates the intent of falling through to the error page than `return {}` so adding `undefined` to the return type would be great.

dummdidumm:
Just checked, this was already done in #1617 , you are allowed to do `return;` when using the `Load` type.

janosh:
@dummdidumm Could you show a usage example?

Issue #1417, opened 2021-05-11
Direct communication with external API 


stalkerg:
The main idea is getting a shorter path to API from SSR and CSR without extra hops and proxies. 
For that, we should have different server URLs and client URLs for the same resource ID. Even now you can have different URLs for `fetch` between server and client but our serialization system detect resource by the whole URL and produce double request in that situation. 
This feature request is not something new and several times accrued in Sapper tracker for example https://github.com/sveltejs/sapper/issues/489 
Also, I did some rouge PR for this here https://github.com/sveltejs/kit/pull/1406, and you can find some extra details there. 
It looks like we should more detail describe this feature and cases to make a clean solution. 

It's a little confusing theme, but I want to hear your opinions and ideas.

Kapsonfire-DE:
I didnt checked in detail - but i liked the initial idea of defining a hash in the fetch method instead of using the url as identifier
The argument not to change the default fetch parameters are strong and I agree to this.

So we have different ways to accomplish this i can think of
1) give the fetch a prototype function which returns a wrapped fetch method with defining the hash method
```javascript
fetch.customHash('myCustomHash')('/my/endpoint/url', {...})
```
probably one of the ugliest ways

2) Like 2, but with import
```javascript
import {hashedFetch} from "svelte";
hashedFetch('/my/endpoint/url', {...}, 'myCustomHash')
```

3) Create nasty handle/endpoint hacks which creat additional network request on CSR.
While this works out of the box - other approaches are much cleaner

4) Define in config a host, which get automatically prepended on fetch requests
At current state the initialFetch uses current svelte kit host variable. I dont know if its used in other functions too.
But using a config like this could work:
```javascript
{
   kit: {
        fetch: {
            prependHost: 'http://10.0.10.3:1337/'
         }
   }
}
```
While this could be a good solution, whats about when different Hosts should be prepended. 

5) Add an additional hook in hooks.ts/hooks.js which allows to intercept the server fetch method
```javascript
export function interceptServerFetch(url, params, fetch) {
   if(url.startsWith('/')) url = 'http://10.0.10.3:1337' + url;
   return fetch(url, params);
}```

While fetch is often used in load, option 1 and 2 will leak this information to the browser.
4 feels good, but is very static


version 5 feels superior IMO

stalkerg:
I like 5 also but it has two issues:
1. Not so flexible as 1 and 2 
[...]

Issue #696, opened 2021-03-26
Server-side fetch request should have headers


ebeloded:
Fetch call made from the server should have the same headers, as the one made from the client. 

```html
<!-- index.svelte -->
<script context="module">

export async function load({ fetch }) {
  const { headers }  = await fetch('/content.json').then(res => res.json())
  return { 
    props: { headers }
  }
};

</script>
```

```ts
// content.json.ts
export const get = (request) => ({
  body: {
    headers: request.headers // <- headers: {} if requested during SSR, but populated otherwise
  }
});
```

In the above case the `request.headers` are empty (`{}`) if requested during SSR, but populated when requested from the client on navigation or if requested directly.

The expected behavior is that in both cases the headers are populated and requests look the same.

**envinfo:**

```
  npmPackages:
    @sveltejs/adapter-node: next => 1.0.0-next.10 
    @sveltejs/kit: next => 1.0.0-next.60 
    svelte: ^3.29.0 => 3.35.0 
```

GrygrFlzr:
Is this the same issue as #672? Although I suppose it covers a wider range (all headers) instead of just cookies.

Rich-Harris:
This is an interesting and tricky problem. The browser's `fetch` adds a number of implicit headers:

```js
{
  host: 'localhost:3000',
  connection: 'keep-alive',
  'sec-ch-ua': '',
  'sec-ch-ua-mobile': '?0',
  'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/90.0.4421.0 Safari/537.36',
  accept: '*/*',
  'sec-fetch-site': 'same-origin',
  'sec-fetch-mode': 'cors',
  'sec-fetch-dest': 'empty',
  referer: 'http://localhost:3000/load/headers',
  'accept-encoding': 'gzip, deflate, br'
}
```

Meanwhile, the `fetch` function available to `load` in SSR can read the *page's* request headers, which look something like this:

```js
{
  host: 'localhost:3000',
  connection: 'keep-alive',
  'upgrade-insecure-requests': '1',
  'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/90.0.4421.0 Safari/537.36',
[...]

PR #1465, opened 2021-05-16
Add serverFetch hook

stalkerg:
This PR base on https://github.com/sveltejs/kit/issues/1417#issuecomment-840074099 and on all discussion for direct external API requests (#1417) and headers manipulation from #696. 

Basically, now we provide developers full control of how to retrieve data on the server-side and even UNIX sockets can be an option. 
The test implementation is a little tricky but looks like we have already a similar thing for handles large responses.

Thanks.


Kapsonfire-DE (changes requested):
the type ServerFetch have to be capitalized to fit the other namings
```
export { Incoming, GetSession, Handle, ServerResponse as Response, serverFetch } from './hooks';
```

pixelmund:
This looks cool but right now, let's say i have an external api which depends on users cookies, the cookies only gets send through internal sk endpoints while ssr even if its the same domain. Couldn't we pass the 'server' request to the serverFetch hook? I would currently have to patch package svelte kit to pass request headers to the external api or create an sk endpoint which proxies the request.

stalkerg:
> Couldn't we pass the 'server' request to the serverFetch hook?

I also thought about it... it will be greater to have, but I am not sure about @Rich-Harris opinion. In that case, we will break the `fetch` API.

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -98,8 +98,12 @@ export async function load_node({
 				let response;
 
 				if (/^[a-zA-Z]+:/.test(url)) {
-					// external fetch
-					response = await fetch(url, /** @type {RequestInit} */ (opts));
+					const request = new Request(url, /** @type {RequestInit} */ (opts));
```
dummdidumm, line 101: Is a `Request` object really present in all environments and not only node?

stalkerg:
@Kapsonfire-DE thanks! Fixed.

packages/kit/src/runtime/server/page/load_node.js
stalkerg: This is `load_node.js` should be represented server-side only. (I hope NodeJS only)

stalkerg: https://developer.mozilla.org/en-US/docs/Web/API/Request should be supported very well anyway.

benmccann:
I haven't gotten a chance to take a look at this one to think about whether I think this is the best solution or not yet, but it would probably need the documentation to be updated as well

stalkerg:
@benmccann ok, I will prepare it. Also, my project has already used it with good results.

stalkerg:
I added some documentation but I think changelog will be better to add after final approves of the concept.

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -98,8 +98,12 @@ export async function load_node({
 				let response;
 
 				if (/^[a-zA-Z]+:/.test(url)) {
-					// external fetch
```
benmccann, line 101: this comment shouldn't be removed

also, I would expect the user-provided function to be used in all cases and not just the external case

documentation/docs/04-hooks.md
```diff
@@ -85,3 +85,19 @@ export function getSession(request) {
...
+This function provides possibilities to replace `fetch` on the server-side for the external API requests. It's beneficial if you want to retrieve resources differently from what it is doing on the browser. For example, if you want to change URL or headers for server-side only.
+
+```ts
+type ServerFetch = (req: Request) => Promise<Response>;
+```
+
+```js
+/** @type {import('@sveltejs/kit').ServerFetch} */
+export async function serverFetch(request) {
+	const data = await getDataMethod(request);
```
benmccann, line 100: you could do this in one line:
```
return await getDataMethod(request);
```

or did you mean for the next line to be `return new Response(data)` as in Rich's example? https://github.com/sveltejs/kit/issues/1417#issuecomment-840037007

it's unclear what `getDataMethod` is here though. maybe we should show a more realistic example such as changing the URL

packages/kit/src/runtime/server/page/load_node.js
stalkerg: > this comment shouldn't be removed

sure

> also, I would expect the user-provided function to be used in all cases and not just the external case

for internal case you can have endpoint call instead true request. Basically, it's has sense only for external API.

documentation/docs/04-hooks.md
stalkerg: > maybe we should show a more realistic example such as changing the URL

ok, I also thought about it.

packages/kit/src/core/dev/index.js
```diff
@@ -187,7 +187,8 @@ class Watcher extends EventEmitter {
 							},
 							hooks: {
 								getSession: hooks.getSession || (() => ({})),
-								handle: hooks.handle || (({ request, render }) => render(request))
+								handle: hooks.handle || (({ request, render }) => render(request)),
+								serverFetch: hooks.serverFetch
```
benmccann, line 191: should this also provide a default implementation like `serverFetch: hooks.serverFetch || fetch`? then you can just call it directly without having to check if it's defined

stalkerg: hmm maybe it's a good idea, I can check it.

stalkerg: Thanks! During these changes, I found and fix the serious issue.

SGarno:
Kudos for getting this going.   I was struggling the logic flow with how SK handles requests and getting that to work with an app that only connects with external APIs (no frontend page protection).

Question with this new serverFetch method....what is the pattern for capturing cookies? 

I have a backend API that operates in this way:

1)  Call an endpoint to identify the type of authentication that is configured on the backend server.   This returns a session_id.
2)  Depending on the response value, it may require username/password, some kind of existing authentication token (i.e. SAML), auto-negotiate, or OAuth2.
3)  If OAuth2, we have to display their authentication page to capture the token.

I am building a wrapper class library for this API that will take care of all the backend api communication (not using routes).

Will headers be available to the serverFetch without having to put code in the frontend (i.e hooks.js and routes)?

stalkerg:
> Will headers be available to the serverFetch without having to put code in the frontend (i.e hooks.js and routes)?

The serverFetch reuses the Request object from your fetch call in `load` function and you can copy your headers into a new request.  Probably you are talking about @pixelmund idea?

SGarno:
I realize @pixelmund was also asking about headers, but wasn't sure if this enhancement addressed that or not from the comments I read above (which is why I was asking).    

In my use case, the application connects to multiple external APIs (i.e. Salesforce, Snowflake, etc.) not for access to the application itself, but for usage of the systems it interacts with.    For this to go through the current hooks and /route endpoints makes writing a wrapper API class awkward in that it has to depend on a tight frontend integration (which is what I am trying to avoid by creating a wrapper class in the first place).

If this serverFetch is not going to address this use case, should I open an enhancement request?

My original thought was to propose something like the following:

```javascript
export default class MyAPI {

    #baseurl;
    #data;

    async Init (url) {
        this.#baseurl = url + "/api/v1/";

        let url = this.#baseurl + 'Server'
        let resp = await fetch(url, {
            method: 'GET',
            render: 'client',    // (could also be 'server' but headers will be undefined)
            headers: {
                Accept: 'application/json',
                'Content-Type': 'application/json',
            },
            credentials: 'same-origin',
        })

        if (resp.status != 200) {
            console.log(`${url} => ${resp.status}:${resp.statusText}`)
            throw new Error(`Unable to connect to server at specified URL [${this.#baseurl}]`)
        }

        // Available in this wrapper class when render = 'client'
        let sessionID = this.headers.cookie('session_id')

        this.#data = await resp.json()
    }
}
```


When the render option of 'client' is received by SK, it will force the fetch request to be executed client side so that the cookies can be returned back to the caller without requiring boilerplate-y code in hooks and routes.   If the value is 'server' (default) then it would operate the same as it does today.

Thoughts?

packages/kit/src/runtime/server/page/load_node.js
Rich-Harris: @dummdidumm Adapters are required to make `Request` globally available if it isn't already part of the target platform. It's part of the (currently undocumented) contract — see e.g. adapter-node https://github.com/sveltejs/kit/blob/babe515f9f5c816fd507920f1f16012263e7a71f/packages/adapter-node/src/server.js#L8

stalkerg:
ok, I will back soon

packages/kit/src/runtime/server/page/load_node.js
Conduitry: > also, I would expect the user-provided function to be used in all cases and not just the external case

This confused me for a while too. But I think the idea is that, in `load`, you would continue to call `fetch` with a path-only URI, which would then call your `handle` hook (without it being an additional network request), and then _there_ in your `handle` you would manually call `fetch` with whatever the appropriate full URL would be - and that you wouldn't really need the `serverFetch` hook at all in that case.

Conduitry: Put another way, the existing `handle` hook is essentially already the `fetch`/`serverFetch` hook for same-origin requests. We don't need another one for that.

benmccann: It seems a bit inconsistent though that you'd use `serverFetch` to alter the request URL for external requests and `handle` to alter the request URL for internal requests. I think it'd be a clearer API for `serverFetch` to be available regardless

Also, imagine that `/api/endpoint` is on the same domain, but on another server running in another language. In that case you may want to use `serverFetch` to do something like hit the service directly instead of the external load balancer. In that case you don't necessarily want to write more complicated code to have `handle` proxy to the other server compared when it'd be much simpler to have `serverFetch` simply change the URL

Conduitry: How would `handle` and `serverFetch` interact in that case? Without calling `handle` first, how would Kit know whether this request is going to be handled by an internal page/endpoint or an external one?  And if it does call `handle` first, how does the handoff to `serverFetch` happen?

benmccann: `handle` wouldn't be called at all because you wouldn't be hitting a SvelteKit endpoint. If `serverFetch` changes the URL from `/api/endpoint` to `http://backend-service/api/endpoint` it should be pretty clear that it's external since it's a different host

stalkerg: or even request by unix socket

stalkerg: > I think it'd be a clearer API for serverFetch to be available regardless

hmm, I suppose serverFetch has sense only for external API, because for internal API SvelteKit carries very well. 
Also, in that case, users should provide this check inside serverFetch somehow in any case.

stalkerg:
> If this serverFetch is not going to address this use case, should I open an enhancement request?

`serverFetch` can help here in your case, because you can write your communication with external API directly without proxy on the SvelteKit side.

SGarno:
> `serverFetch` can help here in your case, because you can write your communication with external API directly without proxy on the SvelteKit side.

Excellent!  👏    I am curious to see how it operates with an OAuth2 credentialed API.    If you have any examples of how we would render (the "SK way") the authentication url  to the client using serverFetch, that would be really helpful.

benmccann:
@SGarno this PR has not been checked in or released yet, which is indicated by its open status. I'm not sure what you're talking about, but the feature is not available

Rich-Harris:
finally got round to reviewing/merging this. thank you!

istarkov:
Related https://github.com/sveltejs/kit/issues/1784
This breaks the node-adapter sveltekit builds

PR #1471, opened 2021-05-17
Add link to http-status-codes

### LINKED ISSUE #1439: Add HTTP Status Codes ENUM
When giving proper error codes in response of endpoints/loads sometimes I have to google for the correct code, like PAYMENT_REQUIRED, BAD_REQUEST and some other things. Most Codes are well known, others not. To make it easier implementing a ENUM is a good solution.

I'd like to get the Status Code ENUMs integrated in svelte kit. 

I started using https://www.npmjs.com/package/http-status-codes package.
```javascript
export async function post(request) {
    let ct = request.headers['content-type'] ?? '';
    if(ct !== 'application/json') {
        return {
            status: StatusCodes.BAD_REQUEST,
            body: ReasonPhrases.BAD_REQUEST
        };
    }
    let data = request.body;
    return {
        body: data
    }
}
```

A decision has to be made. 
1)
I can create a PR and add it to svelte kit. 
2)
A already written npm package could be put into the package.json.
3)
Leave it to the user like current state.

dummdidumm:
Closes #1439

Not all people might be familiar with status codes. Add a hint for them that they can use a library for that if they want to.

### DIFF
--- a/documentation/docs/01-routing.md
+++ b/documentation/docs/01-routing.md
@@ -104,6 +104,7 @@ The job of this function is to return a `{ status, headers, body }` object repre
 - `5xx` — server error
 
 > For successful responses, SvelteKit will generate 304s automatically.
+> If you are unsure about status codes or can't remember them, a package like [http-status-codes](https://www.npmjs.com/package/http-status-codes) might help.
 
 If the returned `body` is an object, and no `content-type` header is returned, it will automatically be turned into a JSON response. (Don't worry about `$lib`, we'll get to that [later](#modules-$lib).)
 
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -105,6 +105,8 @@ If you return a Promise from `load`, SvelteKit will delay rendering until the pr
 
 The HTTP status code for the page. If returning an `error` this must be a `4xx` or `5xx` response; if returning a `redirect` it must be a `3xx` response. The default is `200`.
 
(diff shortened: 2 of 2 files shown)


Conduitry:
How treeshakable is this library? Do we want to just link people to MDN instead? Or Wikipedia even?

dummdidumm:
I don't know. I had this in the back of my head, too. The idea of referencing a library was that in the referenced issue the problem was precisely that "I have to google this all the time", and using a library which provides autocompletion and meaningful constants would help. If we don't want to endorse a specific library or are too concerned with the additional kilobytes then I would just not merge this and close the issue. The docs already link to a list of status codes.

benmccann:
It looks like tree shaking is a work-in-progress. There's an issue tracking it https://github.com/prettymuchbryce/http-status-codes/issues/55 and a PR that's been open since November https://github.com/prettymuchbryce/http-status-codes/pull/64 to further improve it. Version 2.1.5-beta.1 is a lot better than the latest stable version 2.1.4. No idea when 2.1.5 might be released though

benmccann:
@Conduitry @dummdidumm any thoughts about whether we want to move forward with this PR?

Conduitry:
I still don't think we want to be even semi-blessing a random third-party library for something like this, even a popular one - especially if we have doubts about the treeshakability of the library or about what direction it might be going in.

We're already linking to an MDN page on this in the docs. If someone wants more than that, they can get it themselves, along with whatever extra bytes it comes with, which would then be their responsibility.

dummdidumm:
I'm indifferent. Feel free to close (and also the related issue with a short explanation why we won't do it).

Issue #942, opened 2021-04-10
Netlify Adapter: Netlify Forms don't work


kvn-shn:
**Describe the bug**
[Netlify Forms](https://www.netlify.com/products/forms/) don't work in a Svelte component if setup with [this approach](https://www.netlify.com/blog/2017/07/20/how-to-integrate-netlifys-form-handling-in-a-react-app/) and using the [Netlify adapter](https://github.com/sveltejs/kit/tree/master/packages/adapter-netlify).

_Edit: For a workaround/fix see the comment below https://github.com/sveltejs/kit/issues/942#issuecomment-828726791_

**To Reproduce**
1. Setup a Netlify Form as described [here](https://docs.netlify.com/forms/setup/) and more specifically [here](https://www.netlify.com/blog/2017/07/20/how-to-integrate-netlifys-form-handling-in-a-react-app/):
    * Create a Netlify "form detection helper". Something like:
        ```html
        <!-- "/static/netlify-form-helper.html" -->
        <form name="test" netlify netlify-honeypot="bot-field" hidden>
            <input type="text" name="name" />
            <input type="email" name="email" />
            <textarea name="message"></textarea>
        </form>
        ```
    * Create the Svelte form component:
        ```svelte
        <!-- "/src/routes/test-form.svelte" -->
        <form name="test" method="post" action="/success">
            <input type="hidden" name="form-name" value="test" />
            <input type="text" name="bot-field" />
            <p>
                <label>Your Name: <input type="text" name="name" /></label>
            </p>
            <p>
                <label>Your Email: <input type="email" name="email" /></label>
            </p>
            <p>
                <label>Message: <textarea name="message" /></label>
            </p>
            <p>
                <button type="submit">Send</button>
            </p>
        </form>
        ```
    * Ensure Netlify's form processing is enabled.
2. If everything is setup correctly, Netlify should tell you that it detected a form named "test" for your site.
3. Now navigate to your https://example.tld/test-form and send a submission.
4. Even when taking care of all troubleshooting tips from [here](https://docs.netlify.com/forms/troubleshooting-tips/) and [here](https://answers.netlify.com/t/support-guide-form-problems-form-debugging-404-when-submitting/92), neither verified nor spam submissions show up in Netlify's form overview.

**Expected behavior**
A form submission should be shown in Netlify's form overview if submitted via my SvelteKit webapp.

**Information about your SvelteKit Installation:**
* @sveltejs/kit: next => 1.0.0-next.64
[...]

PR #1481, opened 2021-05-17
docs: explain form handling using Netlify Adapter

kvn-shn:
Extended the Netlify Adapter readme to (better) explain how Netlitfy forms can be used in SvelteKit: https://github.com/sveltejs/kit/issues/942#issuecomment-828726791

### DIFF
--- a/packages/adapter-netlify/README.md
+++ b/packages/adapter-netlify/README.md
@@ -36,7 +36,9 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
 
 It's recommended that you add the `build` and `functions` folders (or whichever other folders you specify) to your `.gitignore`.
 
-## Customizing Setup
+## Netlify alternatives to SvelteKit functionality
+
+You may build your app using functionality provided directly by SvelteKit without relying on any Netlify functionality. Using the SvelteKit versions of these features will allow them to be used in dev mode, tested with integration tests, and to work with other adapters should you ever decide to switch away from Netlify. However, in some scenarios you may find it beneficial to use the Netlify versions of these features. One example would be if you're migrating an app that's already hosted on Netlify to SvelteKit.
 
 ### Using Netlify Redirect Rules
 
@@ -45,10 +47,8 @@ During compilation a required "catch all" redirect rule is automatically appende
 - `[[redirects]]` in `netlify.toml` will never match as `_redirects` has a [higher priority](https://docs.netlify.com/routing/redirects/#rule-processing-order). So always put your rules in the [`_redirects` file](https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file).
 - `_redirects` shouldn't have any custom "catch all" rules such as `/* /foobar/:splat`. Otherwise the automatically appended rule will never be applied as Netlify is only processing [the first matching rule](https://docs.netlify.com/routing/redirects/#rule-processing-order).
 
-### Defining Netlify Forms
+### Using Netlify Forms
 
 1. Create your Netlify HTML form as described [here](https://docs.netlify.com/forms/setup/#html-forms), e.g. as `/routes/contact.svelte`. (Don't forget to add the hidden `form-name` input element!)
 2. Netlify's build bot parses your HTML files at deploy time, which means your form must be [prerendered](https://kit.svelte.dev/docs#ssr-and-javascript-prerender) as HTML. You can either add `export const prerender = true` to your `contact.svelte` to prerender just that page or set the `kit.prerender.force: true` option to prerender all pages.
-3. If your Netlify form has a [custom success message](https://docs.netlify.com/forms/setup/#success-messages), e.g. by using something like `<form netlify ... action="/success">`, then:
-   - Ensure `/routes/success.svelte` exists
-   - In it add the same prerender hint as in step 2
+3. If your Netlify form has a [custom success message](https://docs.netlify.com/forms/setup/#success-messages) like `<form netlify ... action="/success">` then ensure the corresponding `/routes/success.svelte` exists and is prerendered.
--- a/packages/adapter-netlify/README.md
+++ b/packages/adapter-netlify/README.md
@@ -36,6 +36,19 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
 
 It's recommended that you add the `build` and `functions` folders (or whichever other folders you specify) to your `.gitignore`.
 
-## Extra Resources
+## Customizing Setup
 
-Guide for SvelteKit + Netlify Forms users: https://dev.to/swyx/how-to-use-sveltekit-with-netlify-forms-5gmj
+### Using Netlify Redirect Rules
+
+During compilation a required "catch all" redirect rule is automatically appended to your `_redirects` file. (If it doesn't exist yet, it will be created.) That means:
+
+- `[[redirects]]` in `netlify.toml` will never match as `_redirects` has a [higher priority](https://docs.netlify.com/routing/redirects/#rule-processing-order). So always put your rules in the [`_redirects` file](https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file).
+- `_redirects` shouldn't have any custom "catch all" rules such as `/* /foobar/:splat`. Otherwise the automatically appended rule will never be applied as Netlify is only processing [the first matching rule](https://docs.netlify.com/routing/redirects/#rule-processing-order).
+
+### Defining Netlify Forms
+
+1. Create your Netlify HTML form as described [here](https://docs.netlify.com/forms/setup/#html-forms), e.g. as `/routes/contact.svelte`. (Don't forget to add the hidden `form-name` input element!)
+2. Netlify's build bot parses your HTML files at deploy time, which means your form must be [prerendered](https://kit.svelte.dev/docs#ssr-and-javascript-prerender) as HTML. You can either add `export const prerender = true` to your `contact.svelte` to prerender just that page or set the `kit.prerender.force: true` option to prerender all pages.
+3. If your Netlify form has a [custom success message](https://docs.netlify.com/forms/setup/#success-messages), e.g. by using something like `<form netlify ... action="/success">`, then:
+   - Ensure `/routes/success.svelte` exists
+   - In it add the same prerender hint as in step 2


sw-yx:
i dont have a say bc im not a maintainer but i'm a fan of dropping all the related netlify tips on to the readme here 👍🏽

packages/adapter-netlify/README.md
```diff
@@ -36,6 +36,27 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
...
 It's recommended that you add the `build` and `functions` folders (or whichever other folders you specify) to your `.gitignore`.
 
-## Extra Resources
-
-Guide for SvelteKit + Netlify Forms users: https://dev.to/swyx/how-to-use-sveltekit-with-netlify-forms-5gmj
+## Troubleshooting
+
+### How to make Netlify forms work?
+
+1. Add the following [prerendering configuration](https://kit.svelte.dev/docs#configuration-prerender) to `svelte.config.cjs`:
```
benmccann, line 43: `cjs` was only used in old versions. If you're on the latest version of SvelteKit it will be `js`

```suggestion
1. Add the following [prerendering configuration](https://kit.svelte.dev/docs#configuration-prerender) to `svelte.config.js`:
```

benmccann: Ah, it looks like that might have been taken directly from the article. @sw-yx maybe you'd want to update that

benmccann:
I'm not quite sure what the best path forward is here. I think that this is adding a lot of steps without explanation as to why you're following those steps. I like that [the original article](https://dev.to/swyx/how-to-use-sveltekit-with-netlify-forms-5gmj) has more info. Though we don't usually link to unofficial sources from the official readmes or docs. I'd be curious if the other maintainers have suggestions

kvn-shn:
@benmccann I thought of it less like a tutorial and more like a "Netlify forms are not working in SvelteKit what do I have to do?" troubleshooting section. In my opinion it makes sense to put it right there (when I was first encountering this issue I was actually checking the readme first if it's already explained ;) )

sw-yx:
ya exactly. @ben - i appreciate that policy is ill defined where external integrations with svelte are concerned. 

i'm happy to take on this maintenance work inside of the sveltesociety repo if you just wanna send people there instead... but either way we're all just trying to help future folks get up and running by placing hints in the first places we'd look.

benmccann:
yeah, definitely agree we need to figure out how to make this discoverable. sveltesociety might not be a bad place to put it. the nice thing about putting it there vs an article is that if something's outdated then anyone can send a PR to update

pngwn:
I think the core of this issue is that we have an official package that is for a proprietary service. I think the readme probably _is_ the right place for these docs, no-one made us support netlify directly. If we do support netlify then there is an expectation that the netlify ecosystem will be usable in some way and we need to inform users one way or the other, if the netlify docs will not cover those features in the context of kit.

The trick is how to we get those services to maintain the adapters? They have a vested interest and more capacity to keep things useful, relevant and up-to-date.

sw-yx:
the straight answer is that netlify never will do that, because realistically svelte doesnt register as a high enough priority for them... this is where we just have to be realistic that we can do more on our end than they can do on theirs.

benmccann:
I don't mind having more docs here in the README. It is a very reasonable place to put them. The main concern I had is that it was just taking the steps from the article without providing any context around why you'd be doing these things, in what cases, etc.

sw-yx:
ok, got you, yeah the balance of how much context to offer is tricky. i do like leaving links for further investigation: https://github.com/sveltejs/kit/pull/1481/files#diff-3eebbc3ad3070317b85447ad5cf803ef738eea1f2ad774f1d5506300efeb0388R62

kvn-shn:
> I don't mind having more docs here in the README. It is a very reasonable place to put them. The main concern I had is that it was just taking the steps from the article without providing any context around why you'd be doing these things, in what cases, etc.

@benmccann I guess you didn't mean it like that but believe it or not I've really put some time into it and didn't simply copy it over from the article. (btw, thanks again @sw-yx for pointing me in the right direction.)

As I said, my idea was to simply help out others who are facing the same issues I've encountered regarding SvelteKit and Netlify forms. Thus I'm assuming that someone who's troubleshooting that, already has ~90% setup and is just looking for the necessary minor tweak. For a broader overview we already have Netlify docs etc.

packages/adapter-netlify/README.md
```diff
@@ -36,6 +36,27 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
...
-
-Guide for SvelteKit + Netlify Forms users: https://dev.to/swyx/how-to-use-sveltekit-with-netlify-forms-5gmj
+## Troubleshooting
+
+### How to make Netlify forms work?
+
+1. Add the following [prerendering configuration](https://kit.svelte.dev/docs#configuration-prerender) to `svelte.config.js`:
+   ```js
+   prerender: {
+       crawl: true,
```
benmccann, line 46: This is a bit confusing because this is mostly specifying the defaults. I think `force: true` is the only one here that has an effect. And is that one actually required to be set? I don't understand why a Netlify form could be used only if prerendering. Is there also an alternative of just setting `prerender = true` on the page you have a form on or can we just remove this section altogether?

packages/adapter-netlify/README.md
```diff
@@ -36,6 +36,27 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
...
+       enabled: true,
+       force: true,
+       pages: ['*']
+   }
+   ```
+2. Create your Netlify HTML form as described [here](https://docs.netlify.com/forms/setup/#html-forms), e.g. as `/routes/contact.svelte`. (Don't forget to add the hidden "form-name" input element!)
+3. Tell SvelteKit to prerender your `contact.svelte` page component by adding:
```
benmccann, line 53: If we're specifying `force: true` you shouldn't need to do this. I'm not sure why either is necessary though

packages/adapter-netlify/README.md
```diff
@@ -36,6 +36,27 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
...
+   ```svelte
+   <script context="module">
+       export const prerender = true;
+   </script>
+   ```
+4. If your Netlify form has a [custom success message](https://docs.netlify.com/forms/setup/#success-messages), e.g. something like `<form name="contact" netlify method="POST" action="/success">`, then:
+   - Ensure `/success` points to a proper Svelte page component
```
benmccann, line 60: what is meant by "Svelte page component"? could this say URL instead?

packages/adapter-netlify/README.md
```diff
@@ -36,6 +36,27 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
...
+   - In it add the same prerender hint as in step 3
+5. That's it! _(You can find some more details in the related [GitHub discussion](https://github.com/sveltejs/kit/issues/942#issuecomment-828726791).)_
```
benmccann, line 62: let's not link to the issue. everything you need to know should be explained here. if there's something that people might want to know that's currently in the issue then we can move it here

packages/adapter-netlify/README.md
kvn-shn: I'm pretty sure that it was needed when I tested it initially. But doesn't seem to be the case anymore. Good catch!

kvn-shn:
As suggested in https://github.com/sveltejs/kit/pull/1484#discussion_r634711470 I also added readme changes regarding the redirect rules here.

packages/adapter-netlify/README.md
```diff
@@ -36,6 +36,24 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
...
+### Using Netlify Redirect Rules
+
+During compilation a required "catch all" redirect rule is automatically appended to your `_redirects` file. (If it doesn't exist yet, it will be created.) That means:
+
+- `[[redirects]]` in `netlify.toml` will never match as `_redirects` has a [higher priority](https://docs.netlify.com/routing/redirects/#rule-processing-order). So always put your rules in the [`_redirects` file](https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file).
+- `_redirects` shouldn't have any custom "catch all" rules such as `/* /foobar/:splat`. Otherwise the automatically appended rule will never be applied as Netlify is only processing [the first matching rule](https://docs.netlify.com/routing/redirects/#rule-processing-order).
+
+### Defining Netlify Forms
+
+1. Create your Netlify HTML form as described [here](https://docs.netlify.com/forms/setup/#html-forms), e.g. as `/routes/contact.svelte`. (Don't forget to add the hidden "form-name" input element!)
```
benmccann, line 50:
```suggestion
1. Create your Netlify HTML form as described [here](https://docs.netlify.com/forms/setup/#html-forms), e.g. as `/routes/contact.svelte`. (Don't forget to add the hidden `form-name` input element!)
```

packages/adapter-netlify/README.md
```diff
@@ -36,6 +36,24 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
...
+2. Tell SvelteKit to [prerender](https://kit.svelte.dev/docs#ssr-and-javascript-prerender) your `contact.svelte` component by adding:
```
benmccann, line 51: Is this necessary? If so, it would be useful to explain why

benmccann: I think this could be rewritten as:

Netlify's build bot parses your HTML files as deploy time, which means your form must be [prerendered](https://kit.svelte.dev/docs#ssr-and-javascript-prerender) as HTML. You can either add `export const prerender = true` to your `contact.svelte` to prerender just that page or set the `kit.prerender.force: true` option to prerender all pages.

packages/adapter-netlify/README.md
```diff
@@ -36,6 +36,24 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
...
+   ```svelte
+   <script context="module">
+       export const prerender = true;
+   </script>
+   ```
+3. If your Netlify form has a [custom success message](https://docs.netlify.com/forms/setup/#success-messages), e.g. by using something like `<form netlify ... action="/success">`, then:
```
benmccann, line 57: I think this could be made much shorter like:

> If your Netlify form has a [custom success message](https://docs.netlify.com/forms/setup/#success-messages) like `<form netlify ... action="/success">` then ensure the corresponding `/routes/success.svelte` exists 

I don't think prerendering has to be enabled here

kvn-shn: I won't be able to reevaluate it until the weekend but it is (was?) definitely needed for whatever reason (see top of https://github.com/sveltejs/kit/issues/942#issuecomment-828726791).

benmccann: Ok. It's not mentioned in their docs, but if you tested it then that sounds sufficient. What do you think about making this section a single line like below

> If your Netlify form has a [custom success message](https://docs.netlify.com/forms/setup/#success-messages) like `<form netlify ... action="/success">` then ensure the corresponding `/routes/success.svelte` exists and is prerendered

benmccann:
thanks for this!

Issue #930, opened 2021-04-08
Netlify Adapter: Custom redirects not working


kvn-shn:
**Describe the bug**
Custom Netlify redirects don't work. Neither using `netlify.toml` nor `static/_redirects`.

**To Reproduce**
1. Setup a minimal SvelteKit app using the [Netlify adapter](https://github.com/sveltejs/kit/tree/master/packages/adapter-netlify)
2. Specify a [redirection rule](https://docs.netlify.com/routing/redirects/), e.g. put `/old /new` in `static/_redirects`
3. Deploy the build output on Netlify (or test it locally using [Netlify Dev](https://cli.netlify.com/netlify-dev/))
4. Navigate to https://YOUR-TEST-PAGE.example/old
5. Now you'll get a Svelte 404 error

**Expected behavior**
* Instead of the 404, my website should actually redirect to https://YOUR-TEST-PAGE.example/new
* Document that redirections in `netlify.toml` aren't supported (for now?) because the adapter itself defines a "catch all" in `_redirects` [making all in `netlify.toml` unusable](https://cli.netlify.com/netlify-dev#redirects)

**Information about your SvelteKit Installation:**
* @sveltejs/kit: next => 1.0.0-next.64
* @sveltejs/adapter-netlify => 1.0.0-next.4
* svelte: ^3.29.0 => 3.36.0
* vite: ^2.1.0 => 2.1.4

**Severity**
For me personally? High, as I'm planning to replace an old website with SvelteKit and have to gracefully deal with old URLs. At least the `_redirects` definitions should work.

**Additional context**
* I already noticed that the adapter is simply overwriting my own `_redirects` file with its own content because [`writeFileSync()` is used](https://github.com/sveltejs/kit/blob/bb2d97d9c590b52b23c8a4de68794ab0b7fb5441/packages/adapter-netlify/index.js#L54).
* When investigating the issue locally I tried to change it to `appendFileSync()` (which makes more sense in my opinion). 
* Then however I'm getting the following:
  ```
  Must use import to load ES Module: MY-PROJECT/functions/render/index.js
  require() of ES modules is not supported.
  require() of MY-PROJECT/functions/render/index.js from /npm/node_modules/netlify-cli/node_modules/lambda-local/build/lambdalocal.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
[...]

PR #932, opened 2021-04-08
Support existing Netlify rules defined in _redirects.

mikenikles:
This PR appends to rather than overwrites Netlify's `_redirects` file so that individual projects can configure custom redirects.

### Before submitting the PR, please make sure you do the following
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
   - Resolves #930
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts

PR #1484, opened 2021-05-17
fix: allow custom redirects for Netlify Adapter

kvn-shn:
Fixes https://github.com/sveltejs/kit/issues/930

Kind of similar to https://github.com/sveltejs/kit/pull/932 but with a little more explanation in the readme and some validation.

On my manual tests everything was working fine:
* No `_redirects` available? A new one is created having the required rule.
* `_redirects` already exists? Required rule is appended into a new line.
* "catch all" rule is in `netlify.toml`? Compilation will fail with an error pointing to the readme.
* Custom "catch all" already is in `_redirects` (thus overriding the adapter's)? Pointed out in the readme's troubleshooting section.

### DIFF
--- a/packages/adapter-netlify/index.js
+++ b/packages/adapter-netlify/index.js
@@ -1,4 +1,4 @@
-import { existsSync, readFileSync, writeFileSync } from 'fs';
+import { appendFileSync, existsSync, readFileSync, writeFileSync } from 'fs';
 import { join } from 'path';
 import { fileURLToPath } from 'url';
 import esbuild from 'esbuild';
@@ -39,7 +39,7 @@ export default function () {
 			utils.copy_client_files(publish);
 
 			utils.log.minor('Writing redirects...');
-			utils.copy(`${files}/_redirects`, `${publish}/_redirects`);
+			appendFileSync(`${publish}/_redirects`, '\n\n/* /.netlify/functions/render 200');
 		}
 	};
 
@@ -60,6 +60,12 @@ function validate_config() {
 		if (!netlify_config.build || !netlify_config.build.publish || !netlify_config.build.functions) {
 			throw new Error(
 				'You must specify build.publish and build.functions in netlify.toml. Consult https://github.com/sveltejs/kit/tree/master/packages/adapter-netlify#configuration'
+			);
+		}
+
+		if (netlify_config.redirects) {
+			throw new Error(
+				"Redirects are not supported in netlify.toml. Use _redirects instead. For more details consult the readme's troubleshooting section."
 			);
 		}
 
--- a/packages/adapter-netlify/files/_redirects
+++ b/packages/adapter-netlify/files/_redirects
@@ -1,1 +1,0 @@
-/* /.netlify/functions/render 200


packages/adapter-netlify/index.js
```diff
@@ -63,6 +63,12 @@ function validate_config() {
 			);
 		}
 
+		if (netlify_config.redirects && netlify_config.redirects.from === '/*') {
```
caesar, line 66: I haven't tested this, but I think `netlify_config.redirects` is an array, so you might need a `forEach`?

packages/adapter-netlify/README.md
```diff
@@ -36,6 +36,15 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
 
 It's recommended that you add the `build` and `functions` folders (or whichever other folders you specify) to your `.gitignore`.
 
-## Extra Resources
+## Troubleshooting
```
benmccann, line 39: "Troubleshooting" to me means something like debugging, which doesn't really seem accurate. Maybe something more like "Customizing Setup"

packages/adapter-netlify/README.md
```diff
@@ -36,6 +36,15 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
...
 
-Guide for SvelteKit + Netlify Forms users: https://dev.to/swyx/how-to-use-sveltekit-with-netlify-forms-5gmj
+### Custom Netlify redirect rules
+
+During compilation a required "catch all" redirect rule is automatically appended to your `_redirects` file. (If it doesn't exist yet, it will be created.) That means:
+
+- If you have a `from = "/*"` rule in `netlify.toml`, it will never match, as `_redirects` has a [higher priority](https://docs.netlify.com/routing/redirects/#rule-processing-order).
```
benmccann, line 45: If I'm reading those docs correctly, it's not just `/*` that won't match but any redirect defined in `netlify.toml`

packages/adapter-netlify/index.js
```diff
@@ -38,8 +38,8 @@ export default function () {
 			utils.copy_static_files(publish);
 			utils.copy_client_files(publish);
 
-			utils.log.minor('Writing redirects...');
-			utils.copy(`${files}/_redirects`, `${publish}/_redirects`);
+			utils.log.minor('Creating/updating _redirects file...');
```
benmccann, line 41: "Writing" is more succinct than "Creating/updating" and covers both cases

packages/adapter-netlify/index.js
benmccann: I think the second half of this check should be removed. see my comment above

caesar: Makes sense. In which case just checking if it's defined is enough.

That said, would a better approach be to define the adapter's catch-all in `netlify.toml` instead of in `_redirects`? That way user-defined redirects in both locations would continue to work (so long as they were not catch-all).

kvn-shn: I played around with modifying the netlify.toml but then realized that it would never go into the final build directory (and modifying the original config file is out of question).
So either we instruct the users to set it up themselfes or we use this automatic approach but then disallowing redirects in the toml. For now I'd say supporting at least `_redirects` is better than nothing. What do you think?

caesar: Definitely better than nothing, yes.

packages/adapter-netlify/README.md
kvn-shn: But isn't it debugging?! I'm thinking of "hey, my Netlify redirect rule doesn't work. Why???" ...I reworded it a little to give a hint on what to do if something isn't working.

packages/adapter-netlify/README.md
```diff
@@ -36,6 +36,15 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
...
 
-Guide for SvelteKit + Netlify Forms users: https://dev.to/swyx/how-to-use-sveltekit-with-netlify-forms-5gmj
+### Custom Netlify redirect rules
+
+During compilation a required "catch all" redirect rule is automatically appended to your `_redirects` file. (If it doesn't exist yet, it will be created.) That means:
+
+- `[[redirects]]` in `netlify.toml` will never match as `_redirects` has a [higher priority](https://docs.netlify.com/routing/redirects/#rule-processing-order). So instead put your rules in the [`_redirects` file](https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file).
+- `_redirects` cannot have any custom "catch all" rules such as `/* /foobar/:splat`. Else the one appended by the Netlify Adapter will never be applied because only [the first matching rule](https://docs.netlify.com/routing/redirects/#rule-processing-order) is processed.
+
+### Guide for SvelteKit + Netlify Forms users
```
benmccann, line 48: let's leave this out of this PR and we can handle separately in https://github.com/sveltejs/kit/pull/1481

packages/adapter-netlify/README.md
benmccann: I guess any piece of info could be framed as either documentation or debugging, but I'd prefer to frame it as documentation. It's better that you read this stuff up front rather than waiting until something's broken and then have to debug

benmccann:
thanks!

packages/adapter-netlify/README.md
kvn-shn: Thanks for the explanation. I removed the changes as suggest below.

Issue #518, opened 2021-03-14
svelte-kit package command for building components


Rich-Harris:
Now that #511 is merged, it makes sense to have an issue for this, if only to bat around ideas. It doesn't need to be part of the public beta, but ideally would be part of 1.0.

To recap:

Almost every app has an internal library of some sort. In a SvelteKit app, the library lives in `src/lib` and is available as `$lib`. Meanwhile, almost every component library has a demo/documentation site, or at the very least something Storybook-esque during development.

Therefore an app and a component library are the same thing, but with a different focus: in SvelteKit, an app is a project where the content of `src/routes` is the important thing, while a component library is a project where the content of `src/lib` is the important thing. Other than that, they're basically the same, and can benefit from the same development workflow.

We already have a way of building apps — `svelte-kit build`. Component libraries also need to be built for distribution. We should add a `svelte-kit package` command (open to alternative name suggestions) that does the following:

* Finds components/modules that should be public. I'd suggest the same logic as in `src/routes` — any file that doesn't have a leading `_` in its filename (or that of a parent directory) is considered public
* Builds them for distribution. In the case of `.svelte` files, this would mean applying preprocessors, but otherwise leaving components uncompiled. In the case of everything else, it would presumably mean running everything through the same Vite config as would normally apply to an app, except on a 1:1 transform basis since bundling is probably impossible to do in the general case given the requirement to leave `.svelte` files uncompiled
* Creates a `package.json` with an `"exports"` field that exposes the public files (with similar name handling as routes — for example a `src/lib/Foo/index.svelte` might become `import Foo from 'my-lib/Foo'`)
* Probably some other stuff

The output of this would be a dist folder with everything the component library needs neatly packaged up, ready for `npm publish`.

dummdidumm:
I like this idea!

I think the trickiest part is going to be preprocessing. There's [an issue](https://github.com/sveltejs/component-template/issues/8) in `component-template` repo asking for this. There's also [a exploratory PR](https://github.com/sveltejs/component-template/pull/31) implementing this (but only for Svelte files) and also a [POC specifically for TypeScript](https://github.com/mattjennings/svelte-typescript-component-template). The problem with preprocessing in general is that you can't just get away with preprocessing Svelte files only, you need the same for all your `.ts` / `.scss` /etc files. In the case of TypeScript, you also want to create type definitions which can be used, and ideally you don't need to write them yourself.

josdejong:
[...]

PR #1499, opened 2021-05-20
svelte-kit package

Rich-Harris:
This is an early and incomplete attempt at `svelte-kit package` (#518). It adds a command that packages up the contents of the `src/lib` folder for publishing to npm, and some corresponding config:

```js
// svelte.config.js
export default {
  kit: {
    package: {
      // default values are shown
      dir: 'package',
      exports: {
        include: ['**'],
        exclude: []
      },
      files: {
        include: ['**'],
        exclude: []
      }
    }
  }
};
```

Specifically, it:

* copies all the files in `src/lib` that are specified by `config.kit.package.files`
* preprocesses any Svelte files
* transforms TypeScript files (TODO)
* adds a package.json that copies most values from the root package.json, but also adds an `"exports"` map that includes all the files specified by `config.kit.package.exports`

It _doesn't_ apply other transforms, either Vite built-in transforms or those specified by user plugins. I'm open to persuasion on that front but I think it's better to keep everything as simple as possible.

I'd like it to emit `.d.ts` files for `.ts` files and, if possible (if it even makes sense) `.svelte` files that contain TypeScript. Not familiar enough with the TypeScript API to know how to go about that.

The expected publishing workflow looks something like this:

```
npx svelte-kit package && npm publish package
```

Other possibilities mentioned in #518:

* exclude files with a `_` prefix from `exports` automatically
* `src/lib/Foo/index.svelte` -> `import Foo from 'my-lib/Foo'`
* expose compiled components for consumption as regular modules
* expose formats other than ESM

### Tests

### Changesets


dummdidumm:
Re TypeScript declarations: 
https://github.com/micha-lmxt/svelte-types-writer
And
https://github.com/firefish5000/svelte2dts
Both implement something like this for Svelte files. Basically they transform Svelte (with unpreprocessed TypeScript) to tsx using svelte2tsx (part of language tools) and run tsc on it to emit declaration files.

Rich-Harris:
Oh, that's good to know. svelte2dts looks particularly promising, especially if it can be coaxed into generating declaration files for Svelte components with JSDoc types. May take a stab at integrating that later if I get time, though I suppose an alternative could be to just support vanilla JS for now and worry about types in a follow-up

benmccann:
We'll have to add some docs for this as well

Rich-Harris:
I had a crack at integrating `svelte2dts` but got nowhere, so in the interests of shipping the MVP version of this command I think it probably makes sense to tackle types in a follow-up.

@dummdidumm @halfnelson I did notice that `svelte2tsx` returns an object with `exportedNames` and `events` properties. I didn't study it too closely but it almost feels like we could generate `.d.ts` files from that information? I'm glossing over slots and inferred types here, and probably many other things, but is there a future in which we can generate declarations directly from `svelte2tsx` or have I grossly underestimated the problem?

dummdidumm:
I don't see a near future where this is happening, precisely because of inferred types. I also think that it's better to just let TypeScript handle this stuff, especially since the wrapper to get this working is that small (the svelte2dts code is only a couple of hundred lines).
Happy to take a crack at this after this is merged.

ryanatkn:
For generating Svelte type definitions there's also sveld: <https://github.com/IBM/sveld>

dummdidumm:
Two things that come to my mind:
- what about css files that are referenced?
- when writing a package I often want to define the public API myself. That means I have some index.js/ts from which I re-export the public stuff. This makes for both better imports (only one location) and I can also freely reorganize my files internally without risking breaking changes. Is that possible with the current solution?

Rich-Harris:
> what about css files that are referenced?

any non `.svelte` files are copied across as-is, so a module can do `import './foo.css'` and it will continue to work. they will also have exports generated for them by default, so a consumer can do `import 'some-lib/base.css'` or whatever

> when writing a package I often want to define the public API myself

yes, any `src/lib/index.js` or `src/lib/index.svelte` becomes the main entry point. you can disable all other entry points by doing this...

```js
// svelte.config.js
export default {
  kit: {
    package: {
      exports: ['index.js']
    }
  }
};
```

...and defining the API in `src/lib/index.js`. Consumers can then do this:

```js
import { X, Y, Z } from 'your-lib';
```

TheComputerM:
> For generating Svelte type definitions there's also https://github.com/IBM/sveld ([no `lang="ts"` support yet if ever](https://github.com/IBM/sveld/issues/12)) and `svelte-type-generator` https://github.com/material-svelte/material-svelte/tree/main/tools/svelte-type-generator (uses svelte2tsx)

Just a warning, pls try to not use sveld as it [cannot parse custom css such as sass](https://github.com/IBM/sveld/issues/24) and cannot detect props which are exported as `export {localName as name}`. If you need jsdoc-like syntax for generating declaration files, please check out [what I made](https://github.com/TheComputerM/sveltyper). But I think it is better to provide an official way to generate types instead of looking into third part ones.

packages/create-svelte/templates/default/.gitignore
```diff
@@ -1,5 +1,4 @@
 .DS_Store
 node_modules
 /.svelte-kit
-/build
```
benmccann, line 4: @Rich-Harris did you mean to remove both `/build` and `/functions` here?

manuel3108:
Any chance we can get a new kit release for this? https://github.com/sveltejs/kit/pull/1543
That would create the possibility to easily test it out.

janosh:
How is this supposed to work for single-component packages? If I add an index file that simply reexports the Svelte component,

```js
// src/lib/index.js
export { default } from './MyComponent.svelte'
```

and then install the package and try

```svelte
import MyComponent from 'my-package'
```

I get

> Must use import to load ES Module: my-package/index.js
> require() of ES modules is not supported.
> require() of my-package/index.js from repo/node_modules/vite/dist/node/chunks/dep-0ed4fbc0.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
> Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from my-package/package.json.

The easiest solution to this in the past was to use the `"svelte"` field in `package.json` but `sveltekit package` does not copy that field into its output...

benmccann:
I'm not super familiar with `svelte-kit package`, but I think you can just have `Index.svelte` and don't need an `index.js` if all you're doing is exporting a single component

janosh:
I don't think using a `Index.svelte` would allow me to import as

```js
import MyComponent from 'my-package'
```

Would't I have to do this?

```js
import MyComponent from 'my-package/Index.svelte'
```

If so, I'd say it's better to keep `MyComponent.svelte` and import it as

```js
import MyComponent from 'my-package/MyComponent.svelte'
```

PR #1507, opened 2021-05-20
fix: append trailing slash in manifest

### LINKED ISSUE #1486: Invalid output and infinite redirect loop with `adapter-static` and `trailingSlash: "always"`
**Describe the bug**
Only `<meta http-equiv="refresh" content="0;url=/test/">` is written to `build/test/index.html` when using static adapter and always trailing slash.

**Logs**
None, build runs successfully, maybe just that `301 /test -> /test/` is printed during build.

**To Reproduce**

Use `npm init svelte@next test`, choose skeleton project.

Create a page `src/routes/test/index.svelte` with any content (`<p>hello</p>`). Have `svelte.config.js` like this:

```js
import adapter from "@sveltejs/adapter-static";

const config = {
  kit: {
    target: "#svelte",
    adapter: adapter(),
    trailingSlash: "always",
  },
};

export default config;
```

Reproduction repo with individual setup steps in https://github.com/mvolfik/probable-disco

**Expected behavior**
No redirects should be needed at all. `build/test/index.html` should contain normal html page with `<p>hello</p>`.

**Stacktraces**
N/A

<details>

```
$ npx envinfo --system --npmPackages svelte,@sveltejs/kit,vite,@sveltejs/adapter-static --binaries --browsers

  System:
    OS: Linux 5.4 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i3-3110M CPU @ 2.40GHz
    Memory: 292.40 MB / 7.35 GB
    Container: Yes
    Shell: 3.1.0 - /usr/bin/fish
  Binaries:
    Node: 14.16.0 - ~/.local/bin/node
    Yarn: 1.22.10 - ~/.local/bin/yarn
    npm: 7.6.3 - ~/.local/bin/npm
  Browsers:
    Chromium: 90.0.4430.212
    Firefox: 88.0.1
  npmPackages:
    @sveltejs/adapter-static: next => 1.0.0-next.9 
    @sveltejs/kit: next => 1.0.0-next.107
    svelte: ^3.34.0 => 3.38.2
```
</details>

**Severity**
Bad bad. Makes `adapter-static` unusable with `trailingSlash: "always"`.

**Additional context**
There is one similar issue #1346, but that is apparently caused by Netlify redirects, while this is only sveltekit.

pz-mxu:
Fixes https://github.com/sveltejs/kit/issues/1486

This will fix #1486 by adding a trailing slash to the pages' path if _config.kit.trailingSlash_ is "always".

### DIFF
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -92,9 +92,24 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 
 	/**
 	 * @param {string} path
+	 */
+	function normalizeTrailingSlash(path) {
+		if (config.kit.trailingSlash === 'always') {
+			return path.endsWith('/') ? path : `${path}/`;
+		}
+		if (config.kit.trailingSlash === 'never') {
+			return !path.endsWith('/') || path === '/' ? path : path.slice(0, -1);
+		}
+		return path;
+	}
+
+	/**
+	 * @param {string} path
 	 * @param {string} parent
 	 */
 	async function visit(path, parent) {
+		path = normalizeTrailingSlash(path);
+
 		if (seen.has(path)) return;
 		seen.add(path);
 
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -93,13 +93,13 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 	/**
 	 * @param {string} path
 	 */
-	function normalizeTrailingSlash(path) {
+	function normalize(path) {
 		if (config.kit.trailingSlash === 'always') {
 			return path.endsWith('/') ? path : `${path}/`;
-		}
-		if (config.kit.trailingSlash === 'never') {
+		} else if (config.kit.trailingSlash === 'never') {
 			return !path.endsWith('/') || path === '/' ? path : path.slice(0, -1);
 		}
+
 		return path;
 	}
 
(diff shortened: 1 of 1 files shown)


packages/kit/src/core/adapt/prerender.js
```diff
@@ -90,6 +90,19 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 		}
 	});
 
+	/**
+	 * @param {string} path
+	 */
+	function normalizeTrailingSlash(path) {
+		if (config.kit.trailingSlash === 'always') {
+			return path.endsWith('/') || path === '/' ? path : `${path}/`;
```
benmccann, line 98: you don't need `|| path === '/'` do you? I think that `path.endsWith('/')` would return `true` for it

packages/kit/src/core/adapt/prerender.js
```diff
@@ -217,10 +233,10 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 	for (const entry of config.kit.prerender.pages) {
 		if (entry === '*') {
 			for (const entry of build_data.entries) {
-				await visit(entry, null);
+				await visit(normalizeTrailingSlash(entry), null);
```
benmccann, line 236: it seems like a lot to keep track of to need to call `normalizeTrailingSlash` wherever `visit` is called. could you just call it inside `visit` as the first line of that method instead?

packages/kit/src/core/adapt/prerender.js
pz-mxu: True 🙈

packages/kit/src/core/adapt/prerender.js
pz-mxu: Yes, that's better. Thanks. Simplified and rebased the commit.

benmccann (approved):

Rich-Harris:
thanks 👍

PR #1508, opened 2021-05-20
Documentation only : Clarified use cases for pre-rendering, explained how pre-rendering can support personalized sites

SumitBando:
The primary goal of this PR is to drop the misleading sentence 'In other words, any app that involves user sessions or authentication is not a candidate for adapter-static, even if individual pages within an app are suitable for prerendering.'

With that misdirection gone, everything else can be left to developers' imagination.

Next goal is to add a sentence similar to:
Personalized pages can be prerendered, with some blank or placeholders parts, to be replaced on the client with personalized data fetched in onMount() from backend API endpoints, either your own, or provided by 3rd parties.

Rest of the verbiage just describes the characteristics of the approach.
Those can be shortened or eliminated.

### Tests

### Changesets


dummdidumm:
I like the start of the first paragraph, but the rest is too tailored towards jamstack. It feels off that this specific stack is mentioned in detail when the rest of the docs are very neutral.

documentation/docs/11-ssr-and-javascript.md
```diff
@@ -64,14 +64,15 @@ In many cases, you'll only want to prerender specific pages in your app. You'll
...
+
 #### When not to prerender
 
 The basic rule is this: for a page to be prerenderable, any two users hitting it directly must get the same content from the server.
 
-> In other words, any app that involves user sessions or authentication is _not_ a candidate for `adapter-static`, even if individual pages within an app _are_ suitable for prerendering.
-
 Note that you can still prerender pages that load data based on the page's parameters, like our `src/routes/blog/[slug].svelte` example from earlier. The prerenderer will intercept requests made inside `load`, so the data served from `src/routes/blog/[slug].json.js` will also be captured.
 
+Since the dynamic JAMStack sites require client side rendering, dynamic pages may fail on old devices. If your site needs to support older devices, a page server can understand and adapt to device specific needs, e.g. fully render HTML server side, or transpile JS, before shipping.
```
benmccann, line 75: this doesn't need to mention JAMStack. we've typically written "client-side" instead of "client side".

"page server" seems somewhat like a bit of invented terminology. can we just say SSR instead? also, transpiling JS seems completely separate from the idea of "page server" / SSR

SumitBando: agreed. will rewrite.

documentation/docs/11-ssr-and-javascript.md
```diff
@@ -64,14 +64,15 @@ In many cases, you'll only want to prerender specific pages in your app. You'll
...
 
+The primary benefit of fully prerendering a site is that you do not need to maintain / pay for NodeJS servers to perform SSR to serve pages. The SSR happens at deploy time, after which the site can be served from CDNs, leading to great performance. This delivery model is so powerful that it has a nickname: JAMStack. While static sites are a clear candidate for this model, JAMStack sites are not limited to static content. You can build fully authenticaled, dynamic, personalized sites with this model, just make sure all user specific data is fetched and rendered client side. Parts of pages may be pre-rendered, possibly with some placeholders, then replaced on the client with user specific data fetched from API endpoints in onMount(). You need API servers, instead of page servers.
```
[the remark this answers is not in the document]
SumitBando: 'vs just doing ssr' - that is the core of this change.

The API servers are backend servers, running Python, Java, Ruby, Node or whatever.
They can be your own, or increasingly services like Auth0, Stripe etc.
All dynamic behavior is between those backend servers and Javascript on the client, no need for a middle SSR tier.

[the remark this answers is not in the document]
benmccann: Maybe instead of "You need API servers, instead of page servers" we could say something about the pros and cons like "This can reduce hosting costs by removing the need to do server-side rendering, but also means that changes to your site may be picked up by search engines more slowly."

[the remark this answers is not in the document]
benmccann: > The SSR happens at deploy time, after which the site can be served from CDNs, leading to great performance. This delivery model is so powerful that it has a nickname: JAMStack

I'd prefer to make this sound a bit more neutral and sounding less like we're pushing any one particular solution. I'd rather discuss the pros and cons and let the user decide if it's the best fit for them. We could just remove the more flowery language to say:

>The SSR happens at deploy time, after which the site can be served from CDNs. This delivery model has a nickname: JAMStack

Part of the reason I suggest this is that it's often a good fit, but may not always be. While it will usually improve time-to-first byte, it can make time-to-interactive and other metrics further down the processing pipeline worse especially for more complicated pages. Instead of making a single round trip over the internet to the backend server, you are more likely to have to make multiple round trips over the internet to render the page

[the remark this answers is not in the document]
SumitBando: I don't think the language is suggesting 'this is the way'.
It is just trying to explain characteristics of this particular way.

SumitBando:
> I like the start of the first paragraph, but the rest is too tailored towards jamstack. It feels off that this specific stack is mentioned in detail when the rest of the docs are very neutral.

Jamstack is not a stack. It is just a name for a pattern.
It just happens to describe exactly this.
Can remove the reference entirely, main content does not change much.
Reduced the reference to just once, since I believe it helps.

documentation/docs/11-ssr-and-javascript.md
```diff
@@ -64,12 +64,16 @@ In many cases, you'll only want to prerender specific pages in your app. You'll
...
+
+This delivery model has a nickname: JAMStack. While static sites are clear candidates for this model, prerendered sites are not limited to static content. You can build authenticated, dynamic, personalized sites in this model, just make sure all user specific data is fetched and rendered client-side. Pages can be prerendered, with blanks or placeholders, to be replaced on the client with personalized data fetched in onMount() from backend API endpoints, either your own, or provided by 3rd parties. This also works well for PWAs, prerendered pages can be cached on user's device, and as user navigates from page to page, the app can fetch() data needed to personalize the page.
```
dummdidumm, line 69: I still feel that this and the next paragraph are not matching the flow of the rest of the docs. It's a short dive into what JAMstack is, why it's good, what's possible etc. I don't think this belongs in here.

SumitBando:
Purged all references to JAMStack

documentation/docs/11-ssr-and-javascript.md
```diff
@@ -64,12 +64,16 @@ In many cases, you'll only want to prerender specific pages in your app. You'll
 
 The prerenderer will start at the root of your app and generate HTML for any prerenderable pages it finds. Each page is scanned for `<a>` elements that point to other pages that are candidates for prerendering — because of this, you generally don't need to specify which pages should be accessed. If you _do_ need to specify which pages should be accessed by the prerenderer, you can do so with the `pages` option in the [prerender configuration](#configuration-prerender).
 
+One benefit of fully prerendering a site is that you do not need to maintain / pay for NodeJS servers to perform SSR to serve personalized pages. SSR happens only at deploy time, some people call this Server Side Generatin (SSG), to differentiate from per user SSR. Once generated, these sites are typically served from CDNs, leading to great "Time to first byte" performance. Search engines can crawl the prerendered static pages on CDNs, which can lead to better rankings. Make sure to not prerender pages you do not want search engines to see.
```
babichjacob, line 67: `Node.js` or just `Node`

babichjacob: Static Site Generation (SSG)

documentation/docs/11-ssr-and-javascript.md
```diff
@@ -64,12 +64,16 @@ In many cases, you'll only want to prerender specific pages in your app. You'll
...
+
+While static sites are clear candidates for this model, prerendered sites are not limited to static content. You can build authenticated, dynamic, personalized sites in this model, just make sure all user specific data is fetched and rendered client-side. Pages can be prerendered, with blanks or placeholders, to be replaced on the client with personalized data fetched in onMount() from backend API endpoints, either your own, or provided by 3rd parties. This also works well for PWAs, prerendered pages can be cached on user's device, and as user navigates from page to page, the app can fetch() data needed to personalize the page.
```
babichjacob, line 69: I would probably need to test to see if `load` still works as expected in a statically built site (with `hydrate` and `router` on), but it seems suspicious that it's recommended to switch away from `load` to `onMount` for loading data onto the page. (I would greatly appreciate someone else commenting on this)

babichjacob:
This is coming along nicely (I think), by the way.

benmccann:
Thanks for putting this PR together. I do think there's possibly some room to expand on some of these topics. As another idea for covering this material, I put together a PR showing what it could look like to cover it in the intro instead https://github.com/sveltejs/kit/pull/1525. I'm curious what the other maintainers think about the two approaches

documentation/docs/11-ssr-and-javascript.md
SumitBando: fixed these

dummdidumm:
Tough call, as I now like both variants 😄 I tend a little more towards the introductionary one though because it's at the top and some of these terms are not only used in the prerendering docs.

SumitBando:
> Tough call, as I now like both variants 😄 I tend a little more towards the introductionary one though because it's at the top and some of these terms are not only used in the prerendering docs.

As described in the summary, I can reduce this PR to one line deletion and one line addition.

Rich-Harris:
Thank you but I'm going to close this in favour of #1773, as this isn't really the place for long discussions about different techniques. The docs need to be as short as possible, and no shorter. #1773 briefly articulates the benefits of prerendering, and mentions that you can fetch personalized data in `onMount` but should consider whether it's the right approach.

I've omitted all the SEO stuff, since that's a red herring — googlebot doesn't care whether a page was pre-rendered or not.

Issue #669, opened 2021-03-25
$page.query.get(..) returns null for prerendered pages


falsetru:
**Describe the bug**
`$page.query.get(...)` returns `null`.

**To Reproduce**
- Repository: https://github.com/falsetru/sveltekit-test/tree/issue-1
```
git clone --branch issue-1 https://github.com/falsetru/sveltekit-test.git issue-1
cd issue-1
npm install
npm run build
npx sirv build/  # `npm run start` will hide the issue.
```
- Run above commands
- Visit http://localhost:5000/?name=abc

**Expected behavior**
`abc` printed, but got `null`.

**Information about your SvelteKit Installation:**

<details>
  <summary>Diagnostics</summary>

- The output of `npx envinfo --system --npmPackages svelte,@sveltejs/kit,vite --binaries --browsers`
```
  System:
    OS: Linux 5.8 Ubuntu 20.10 (Groovy Gorilla)
    CPU: (4) x64 Intel(R) Core(TM) i7-7600U CPU @ 2.80GHz
    Memory: 763.02 MB / 15.39 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 12.18.2 - /usr/bin/node
    npm: 7.7.6 - ~/.npm-packages/bin/npm
  Browsers:
    Chrome: 89.0.4389.114
    Firefox: 87.0
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.67 
    svelte: ^3.29.0 => 3.37.0 
    vite: ^2.1.0 => 2.1.5 
```

- Your browser
Google Chrome Version 89.0.4389.114 (Official Build) (64-bit)

- Your adapter
@sveltejs/adapter-static
</details>

**Additional context**
- As a workaround, I am using `new URLSearchParams(document.location.search).get('name')`.
- If I modify https://github.com/sveltejs/kit/blob/4a1c04a53897ac5f4831b54254b81fa0501656b4/packages/kit/src/runtime/server/page.js#L398 to
    ```javascript
	query: new URLSearchParams(document.location.search),
    ```
  the issue is gone. Not sure whether it's proper way.

benmccann:
> @benmccann Your solution should only work within the browser because document is not defined outside the browser (i.e. when adapter-static is prerendering everything)

I have no idea what you're referring to as being my solution because I have not commented on this thread at all

andreasnuesslein:
pff ^^ my bad.

[...]

PR #1511, opened 2021-05-20
fix: construct query params from location.search on rehydrate

pz-mxu:
Fixes https://github.com/sveltejs/kit/issues/669

When rehydrating a page, the query params were set to URLSearchParams(s(page.query.toString())). This change uses location.search when page.query is empty.

### DIFF
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -136,7 +136,7 @@ export async function render_response({
 					page: {
 						host: ${page.host ? s(page.host) : 'location.host'}, // TODO this is redundant
 						path: ${s(page.path)},
-						query: new URLSearchParams(${page.query.toString().length > 0 ? s(page.query.toString()) : 'location.search'}),
+						query: new URLSearchParams(location.search),
 						params: ${s(page.params)}
 					}
 				}` : 'null'}
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -136,7 +136,7 @@ export async function render_response({
 					page: {
 						host: ${page.host ? s(page.host) : 'location.host'}, // TODO this is redundant
 						path: ${s(page.path)},
-						query: new URLSearchParams(${s(page.query.toString())}),
+						query: new URLSearchParams(${page.query.toString().length > 0 ? s(page.query.toString()) : 'location.search'}),
 						params: ${s(page.params)}
 					}
 				}` : 'null'}


benmccann:
This isn't a section of code I'm super familiar with. Can you explain under what scenarios `page.query.toString()` would be different from `location.search`?

pz-mxu:
Actually I just didn't want to prevent what was happening before if page.query is set. We could probably just use location.search.

benmccann:
Yes, this seems simpler. Still, I'm wondering what exactly the bug was. E.g. there are other places we use `page` and I wouldn't want there to be issues there as well. Can you explain why or when `page.query.toString()` is different from `location.search`?

pz-mxu:
The bug was that when rehydrating pages on static sites, the query string from the url was not used, only the query prop of the page object which was empty.

falsetru:
This will fix #669

benmccann:
> the query prop of the page object which was empty

That's what concerns me about this fix. It seems like a bit of a workaround. Why is the query prop empty? It seems like the query prop is still going to be broken for anyone who tries to use it

benmccann:
I added an explanation for what's happening on the ticket: https://github.com/sveltejs/kit/issues/669#issuecomment-848988844. I'm not sure what the other maintainers think, but at the moment I'd lean towards disallowing usage of `$page.query` on prerendered pages instead

johnnysprinkles:
I think we need to address the issue of static rendered pages that want per-request page store data at runtime, in a wholistic way. If #669 is about query parameters, and #1561 is about route parameters, they're two sides of the same coin and we should address them both in one pass.

johnnysprinkles:
I'd be curious if @benmccann or @pz-mxu has anything thoughts about this RFC I submittted today, making the argument this that PR is a special case of a general issue. https://github.com/sveltejs/rfcs/pull/52

Rich-Harris:
Using `page.query` is deliberate — the idea is to guarantee consistency between server and client renders. Breaking that guarantee feels like a dangerous road to go down. If you don't know `location.search` at SSR time (because you're prerendering) then it feels more appropriate to disable SSR for that page than to render an incorrect page, surely?

benmccann:
Thanks for taking the time to investigate and propose this fix. I'm going to go ahead and close this given that Rich has the same opinion as I do that this isn't the correct way to address the issue

johnnysprinkles:
So this issue seems to be fixed now... if I serve a staticly built page and put this code in there:

		if (!prerendering) {
			console.log($page.url.searchParams.get('foo'));
		}

I get the actual runtime per-request value of "foo" as seen by the client, instead of the value as was snapshotted at prerender time. I'd like to find out which commit change all this around!

PR #1514, opened 2021-05-21
Add favicon to skeleton template

sidharthv96:
build was failing for the default skeleton template as favicon was missing.

I chose to add the favicon inside static rather than removing the reference from app.html as it would be helpful to know where to place static files and reference them.

### Tests

### Changesets


benmccann:
seems reasonable like a reasonable fix. could you add a changeset as well so that this will be able to be released?

sidharthv96:
@benmccann Done!

benmccann (approved):

packages/kit/src/runtime/client/router.js
```diff
@@ -242,7 +242,7 @@ export class Router {
 
 			if (incorrect) {
 				info.path = has_trailing_slash ? info.path.slice(0, -1) : info.path + '/';
-				history.replaceState({}, '', `${info.path}${location.search}`);
+				history.replaceState({}, '', `${this.base}${info.path}${location.search}`);
```
benmccann, line 245: Can you drop this change from this PR? It's unrelated so should be its own PR

sidharthv96: Oops, I was fixing another issue, my bad.

Issue #844, opened 2021-04-03
--https mode causes issues in dev by trying to connect to wss://localhost:24678


Conduitry:
**Describe the bug**
In `--https` mode, the site is served at `https://localhost:3000` and so the Vite websocket connection that is attempted fails, because it tries to make it at `wss://localhost:24678` when it exists at `ws://localhost:24678`. This results in the page being periodically reloaded.

**Logs**
```
Firefox can’t establish a connection to the server at wss://localhost:24678/.
[vite] server connection lost. polling for restart... client:340:13
[vite] connecting...
```

**To Reproduce**
`npx svelte-kit dev --https`

Open console and/or network logs and wait. Turn on timestamps on the console to help see what's happening. The page refresh rate seems to vary widely, and I haven't figured out what causes that to happen.

**Expected behavior**
I don't think we can force an HTTPS page to load a non-secured websocket, so we'll have to figure out how to serve a secure websocket over that port. This will also still probably cause issues because it will be a self-signed cert. I don't know how to resolve this.

**Severity**
High. HTTPS mode is not very usable if the browser refreshes periodically.

**Additional context**
I told you HTTPS mode was a mistake and you didn't believe me.

mindrones:
I just got this, using `npm run dev -- -H` on `SvelteKit v1.0.0-next.70`. On Chrome it refreshes about every 4 secs, no logs apart from `[vite] connecting...`; in Firefox it refreshes about every sec and I got this log just once:

<details>
<summary>Details</summary>

![sveltekit_https_periodic_refreshes](https://user-images.githubusercontent.com/1309648/113483380-06365d80-949b-11eb-8e06-a48ba2a004f6.png)

</details>

...then for some unknown reason it seems to start working (no refreshes).

> I told you HTTPS mode was a mistake and you didn't believe me.

Well, in my case I'm testing the Web Bluetooth API in Chrome which requires HTTPS so I'm glad for the `-H` option! :)

Conduitry:
This is just part of a larger issue with not being able to reach the hmr (?) server on that port. Rich had suggested trying to disable it when in https mode, but really it would be better if the port were configurable, if you could disable there feature separately from enabling https mode, and if its existence was documented for people who are e.g. running Kit inside Docker. Also ideally the whole page wouldn't reload when the server can't be reached.

Conduitry:
[...]

PR #1517, opened 2021-05-22
#844 fix for SSL HMR websockets

JBusillo:
This is a proposed fix to Issue #844

Currently, when the dev server is invoked using the https option (-H or --https) using an unsigned certificate, the Vite HMR web client cannot connect to the server using its 'wss' url.  It retries incessantly.  Browsers will reject the connection, as the user can't 'approve' the possibly unsafe site.

To correct this, the web socket should 'piggy-back' onto the https server.  Thus, when the user selects 'Proceed' from the 'Your connection is not private' dialog, both https and wss connections will be deemed valid.  Vite manages the wss connection upgrade from this point forward.

The get_server function must be called before vite.createServer, since the server reference must be available to vite.createServer.

Perhaps I'm being too cautious, but I didn't want to switch the order of the current vite.createServer block (of code) with the get_server block.  The dev server isn't ready to process requests until vite.createServer is complete.  Thus, you'll see that I coded a dummy handler that is eventually replaced once vite.createServer has completed.

I ran the tests within kit (pnpm run test), and ran tests on both 'dev' and 'preview' with all combinations of the --https and --host options.

This is my first pull request ever -- please don't be too hard on me.


packages/kit/src/core/dev/index.js
```diff
@@ -104,7 +114,11 @@ class Watcher extends EventEmitter {
 			publicDir: this.config.kit.files.assets,
 			server: {
 				...user_config.server,
-				middlewareMode: true
+				middlewareMode: true,
+				hmr: {
+					...user_config.server?.hmr,
```
benmccann, line 119: The CI is failing and I think it's because of `?.`

JBusillo: Yes, you're right.  Optional chaining was implemented in Node 14, so the Node 12 test failed.  I learn something new every day.

I haven't generated the changeset yet -- I'm a little uncertain about the options.  I think I'd only select the 'kit' package, and a 'patch' version, but I'd appreciate your confirmation.

benmccann: Yes, that's correct

JBusillo:
Is it normal if the netlify tests were cancelled?  There were no changes in the output build, since this change only affects dev and preview, when the -H or --host parameters are supplied.

JBusillo:
Good idea.  Code changed, tested locally in dev & preview using demo app.  I think I was just being too cautious -- not wanting to reorder code.

packages/kit/src/core/dev/index.js
```diff
@@ -311,6 +274,47 @@ class Watcher extends EventEmitter {
...
+					emitCss: !this.config.kit.amp
+				})
+			],
+			publicDir: this.config.kit.files.assets,
+			server: {
+				...user_config.server,
+				middlewareMode: true,
+				hmr: {
+					...(user_config.server && user_config.server.hmr),
+					...(this.https ? { server: this.server, port: this.port } : {})
```
benmccann, line 306: oh, I see the issue now. I do think that `this.vite` should continue to be defined before `this.server` (sorry to somewhat contradict my last message. I wasn't totally following before)

what I don't understand here though is why you have a `server` attribute. that's not a valid Vite option as far as I can tell: https://vitejs.dev/config/#server-hmr. If you need to provide it. can you instead provide the `protocol`, `host`, and `port` without providing `this.server` in order to break the circular dependency?

JBusillo: The hmr.server option is referenced in [Vite's web socket setup](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/ws.ts).    The first few lines (23-35) is the applicable code.  It's true that it isn't in the documentation -- it may be an oversight, and perhaps we should verify that with the Vite team.

SvelteKit uses Vite's "middlewareMode" (e.g., where the server is managed by SvelteKit, and Vite is considered as middleware).  Thus, the websocket must use the SvelteKit's https server to ensure that the certificate is valid, i.e., the certificate is validated through a Certificate Authority, or the user opts to 'proceed' when using an unsigned certificate or otherwise suspicious certificate.

When connecting to a secure websocket from the browser, the browser automatically changes the request from 'wss' to 'https'.  If the certificate is valid, the connection will allowed.  Vite will then intercept this request and complete the wss connection.  The request is identified by the header "sec-websocket-protocol: vite-hmr".

'hmr.server' tells Vite which server will receive the connection request -- it needs the actual server object, as it's adding the 'upgrade' event handler to that object to complete the connection and to accept subsequent wss requests.

Regarding the dummy (interim) handler -- My concern is that some kind of error (possibly fatal?) would occur if the client requests the website before it's initialized.

In summary:

- We need to verify if the lack of documentation for the hmr.server option was an oversight, and that it will continue to be supported (how? who to contact?)
- SvelteKit's server object must be passed to hmr.server, as this is how Vite expects it.
- Should I revert the last commit?  Should the dummy (interim) handler be kept?  Should I wait?

JBusillo: I discovered another work-around that doesn't require any code changes, but it doesn't support Android (Chrome, at least).  I don't know about IOS.  It's kind of ugly:

1. Add vite.server.https option to svelte.config.js:

```
	kit: {
		target: '#svelte',
		vite: {
			server: {
				https: true
			}
		}
	}
```
2. Start the dev server - npx svelte-kit dev -H --host 192.168.2.50   (or whichever ip points to your dev machine)
3. Point browser to https://192.168.2.50:24678
4. In the "Your connection is not private" dialog, click "Advanced" then "Proceed to ..."
5. The browser displays "upgrade required".  Ignore this.
6. Now point your browser to https://192.168.2.50:3000
7. In the "Your connection is not private" dialog, click "Advanced" then "Proceed to ..."
8. You're on your way!

You'll need to follow this procedure upon any server restart.  

Pros:   No code changes
Cons: A bit kludgy, doesn't support all devices.

benmccann: Hmm, yeah, none of these options seem great. I think it would be nice if the Vite folks could document how this should be done.

I looked through their issue tracker and found a recently resolved issue, which makes me wonder if this is possible now? Take a look at: https://github.com/vitejs/vite/pull/1992#issuecomment-843939223

JBusillo: I looked at the comment -- they use DevilBox, which generates a Certificate Authority which you import into your browser.  It then generates signed certificates using the trusted CA.  I've created my own trusted CA in the past, and it's quite an involved process.  Additionally, some companies disallow adding Trusted CA's (e.g., via WIndows Group Policy).

I also tested with the latest release of Vite (2.3.4), which includes the changes that you mentioned.  Actually, that was the version I was testing with when I discovered the latest workaround.

I've reached out to the author of that PR [in this comment](https://github.com/vitejs/vite/pull/1992#issuecomment-850731007).  So, I'll wait for his response before proceeding further.

JBusillo: I received this [this reply](https://github.com/vitejs/vite/pull/1992#issuecomment-850841408.)  It appears that this PR didn't specifically address self-signed certificates.
I opened [this issue](https://github.com/vitejs/vite/issues/3587) on Vite.  Hopefully someone will review it, and confirm that the option should be documented.

JBusillo: I received confirmation from the Vite maintainers that the server.hmr.server option is an exposed (public) option.  They asked that I open a PR to update the documentation - https://github.com/vitejs/vite/pull/3590.  It's been approved and merged into Vite's main branch, awaiting the next patch release.

So, we now need to settle on an approach.  Whichever approach, we need to have SvelteKit's server created via `get_server()` before `vite.createServer()`, as the server handle must be passed to `vite.createServer()`.

Here are two approaches:

a. Switch the order of `this.vite = await vite.createServer()` block with `this.server = await get_server` block.  This was my latest commit: 50a05c2.

b. Move only the call to `get_server()` before the `this.vite = await vite.createServer()` block, and attach the handler `this.vite.middlewares(req, res, async () => {...}` in its current location.

For the latter option, I see a few ways of doing this:

1. Add a dummy handler, returning a 204 response (No Content) until the middlewares handler is attached.  This was my commit: b5779a9.
2. Add a dummy handler that retries every 'n' milliseconds until the middlewares handler is attached.
3. take the 'listen' out of `server/index.js` and issue the listen after the middlewares handler is attached.  This would also necessitate a change in the preview code (`start/index.js`)

My concern with (a) and (b.1) is: What happens when an http(s) request is received before everything is initialized?  Although rare, this could happen if a browser tab already points to the URL, and the vite client does its 'ping' during the server's uninitialized interval, or if the user is fast enough to reload the browser during this interval.

With option (b.1), the user would only need to refresh.  I'm not sure what would happen using option (a).

Personally, I like option (b.2), but I'd appreciate your unbiased opinion.

I will await your guidance before proceeding further.

benmccann: I would say that b.3. is definitely the best approach in my mind

JBusillo: OK.  I committed the changes, but I had a problem running 'pnpm test' -- It succeeds, but, looking at the console log, I get this message:
```
(node:931474) UnhandledPromiseRejectionWarning: Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/jbusillo/Projects/kit-pull/packages/kit/test/apps/custom-extension/svelte.config.cjs' imported from /home/jbusillo/Projects/kit-pull/packages/kit/src/core/load_config/index.js
    at finalizeResolution (internal/modules/esm/resolve.js:276:11)
    at moduleResolve (internal/modules/esm/resolve.js:699:10)
    at Loader.defaultResolve [as _resolve] (internal/modules/esm/resolve.js:810:11)
    at Loader.resolve (internal/modules/esm/loader.js:88:40)
    at Loader.getModuleJob (internal/modules/esm/loader.js:241:28)
    at Loader.import (internal/modules/esm/loader.js:176:28)
    at importModuleDynamically (internal/modules/esm/translators.js:114:35)
    at exports.importModuleDynamicallyCallback (internal/process/esm_loader.js:30:14)
    at load_config (file:///home/jbusillo/Projects/kit-pull/packages/kit/src/core/load_config/index.js:109:60)
    at main (file:///home/jbusillo/Projects/kit-pull/packages/kit/test/test.js:376:24)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:931474) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:931474) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
I see that there was a merge performed for #1595, which might be affecting this.  Is it possible that no one had noticed this unhighlighted console message while testing?  Or is my repository messed up?

Earlier today, I had done a "Fetch & Merge" on my pulled repository, since I had seen that Rich Harris had changed /dev/index.js after my initial PR, and wanted to include those changes in my repo.

packages/kit/src/core/server/index.js
```diff
@@ -27,13 +25,23 @@ export async function get_server(port, host, use_https, user_config, handler) {
...
+}
 
+/**
+ * @param {http.Server | https.Server} server
+ * @param {number} port
+ * @param {string} host
+ * @returns {Promise<void>}
+ */
+export async function set_listener(server, port, host) {
+	return new Promise((fulfil) => {
```
benmccann, line 42: I might be missing something, but why not just do `server.listen(port, host || '0.0.0.0')`?

packages/kit/src/core/dev/index.js
benmccann: Hmmm. Weird. I'm getting this locally too. I don't know why we didn't see it on the CI

benmccann: Oh, nevermind. My error was because I hadn't run `pnpm build` before `pnpm test`. If I do that then the tests pass locally for me on `master`

packages/kit/src/core/server/index.js
JBusillo: Listen is asynchronous -- the callback is for waiting until the server is actually listening on that port: [Node Doc](https://nodejs.org/api/net.html#net_server_listen).  The original source used the callback.  I thought that I should, too.

packages/kit/src/core/dev/index.js
JBusillo: I created a new VSCode project by re-cloning from my pulled repository.  Tests are running fine.  I must have messed up my old project somehow.

packages/kit/src/core/server/index.js
benmccann: It looks like the callback is a no-op if I'm reading it correctly (async isn't my forte, so I could be wrong). If it is indeed a no-op, we should go ahead and remove it regardless of what the code was previously doing

packages/kit/src/core/dev/index.js
```diff
@@ -264,36 +273,36 @@ class Watcher extends EventEmitter {
...
-												<h2>${error.severity}</h2>
-												<p>Line ${error.line}, column ${error.col}: ${error.message} (<a href="${error.specUrl}">${
-														error.code
-													}</a>)</p>
-												<pre>${escape(lines[error.line - 1])}</pre>
-											`
-												)
-												.join('\n\n')}
-										`;
+										<head>
```
benmccann, line 276: I don't think there was any need to remove a level of indentation here. If `pnpm format` automatically did it for some reason then we can just ignore it, but if you manually did it then I'd probably revert the indentation change

packages/kit/src/core/dev/index.js
```diff
@@ -317,7 +326,8 @@ class Watcher extends EventEmitter {
 					res.end(e.stack);
 				}
 			});
-		});
+
+		await set_listener(this.server, this.port, this.host);
```
benmccann, line 330: `set_listener` is a weird name for this since the listener isn't passed in. can we just call it `listen` instead?

packages/kit/src/core/server/index.js
JBusillo: Summary: I agree with you.  Change made.

Detail:

The empty callback just ensures that the server is actually listening.  The former `set_listener()` function didn't return until it receives this confirmation.  The new `listener()` function will immediately return to the callee.

Every example of a native node http(s) server or an express server merely displays a console message in the callback, such as 'Server running at http://hostname:port'.

In any web application, browser requests can made at any point in time -- before the server.listen() call, before the server.listen() call is complete, or after completion.  Waiting for the listener to be "ready" doesn't really buy anything, since we have no control over the user's browsing behavior.

Hence, I agree with your suggested change.

packages/kit/src/core/dev/index.js
JBusillo: I did run pnpm run format and pnpm run lint.  I'll do it again with the upcoming commit.  I think it shouldn't be changing the indentation.  In the upcoming commit, I'm adding a pair braces.  They're optional, since there's just one very long statement in the handler - `this.vite.middlewares(...)`, however the original code did have the braces.

Hopefully in my upcoming commit, it will improve the indentation.

JBusillo: It's still there.  I can manually get rid of the extra unindents, and rerun format.  This could have happened accidentally on my part.

benmccann (approved):
thanks for being patient with this one and following up with the Vite team about getting it documented

benaltair:
@benmccann a few people on discord are reporting identical errors after this update. It's been reported that switching back to 113 has fixed it. Wondering if this PR was the culprit? Here are my Netlify build logs with 115 for instance:
```
8:27:12 PM: > The requested module 'svelte-hmr' does not provide an export named 'createMakeHot'
8:27:12 PM: file:///opt/build/repo/node_modules/@sveltejs/vite-plugin-svelte/dist/index.js:214
8:27:12 PM: import { createMakeHot } from "svelte-hmr";
8:27:12 PM:          ^^^^^^^^^^^^^
8:27:12 PM: SyntaxError: The requested module 'svelte-hmr' does not provide an export named 'createMakeHot'
8:27:12 PM:     at ModuleJob._instantiate (internal/modules/esm/module_job.js:92:21)
8:27:12 PM:     at async ModuleJob.run (internal/modules/esm/module_job.js:107:20)
8:27:12 PM:     at async Loader.import (internal/modules/esm/loader.js:179:24)
8:27:12 PM:     at async file:///opt/build/repo/node_modules/@sveltejs/kit/dist/cli.js:647:22
```

Messages:
- https://discord.com/channels/457912077277855764/819723698415599626/854618018502934548
- https://discord.com/channels/457912077277855764/819723698415599626/854627141244551198
- https://discord.com/channels/457912077277855764/819723698415599626/854781180318187560

Here's the update branch on my repo that was failing builds on Netlify:
https://github.com/bahaistudies/conference/tree/update-115

Happy to provide more details or open an issue if you think it's needed.

benaltair:
I tried downgrading to Kit 113 on its own, and still got the error. The error when away when I downgraded Vite to 2.3.6
Here's my `package.json` for reference, now with it working: https://github.com/bahaistudies/conference/blob/6004f98314d4bbeca399f45f2880190c559f1fee/package.json

benaltair:
I tried downgrading to Kit 113 on its own, and still got the error. The error went away when I downgraded Vite to 2.3.6
Here's my `package.json` for reference, now with it working: https://github.com/bahaistudies/conference/blob/6004f98314d4bbeca399f45f2880190c559f1fee/package.json

benmccann:
Hmm. Perhaps @rixo or @dominikg know something about `svelte-hmr`'s involvement or @GrygrFlzr knows anything about recent changes in Vite

JBusillo:
I cloned your repro -- dev, build and preview work with SvelteKit 1.0.0-next.115 and Vite 2.3.7.  However "start" has been renamed to "preview", which tells me that you haven't updated your package.json, and possibly some other configuration that have breaking changes since your previous SvelteKit version.

You might want to create a new SvelteKit project, and compare package.json, svelte.config.js, and any other config files, along with your library structure.  

#844 only affects dev and preview modes.  Also, Netlify deployments shouldn't even reference HMR.  To be honest, I don't know anything about Netlify.

benaltair:
I did check my `package-lock` and found higher versions there. I've since removed/added them back. I wonder what could be causes the error. Maybe this would be better handled with an Issue

benaltair:
Thanks for the tips @JBusillo. I've gotten it to run by first adopting all the config file changes, then I needed to add `.node-version` in the root with the contents `14.13.1` because otherwise I was getting an error from Netlify that the default version of Node they use is not compatible with the requirements in Kit.
```
11:20:20 PM: npm ERR! code ENOTSUP
11:20:20 PM: npm ERR! notsup Unsupported engine for conference@0.0.1: wanted: {"node":"^14.13.1"} (current: {"node":"12.18.0","npm":"6.14.4"})
11:20:20 PM: npm ERR! notsup Not compatible with your version of node/npm: conference@0.0.1
11:20:20 PM: npm ERR! notsup Not compatible with your version of node/npm: conference@0.0.1
11:20:20 PM: npm ERR! notsup Required: {"node":"^14.13.1"}
11:20:20 PM: npm ERR! notsup Actual:   {"npm":"6.14.4","node":"12.18.0"}
```
If this isn't an isolated issue, it might be worth adding a FAQ to the Netlify adapter that this file is needed, or even just add it to the skeleton project if it doesn't cause any harm perhaps?

benmccann:
The adapter should be able to spit out a `.node-version` or `.nvmrc` file as well. I'm not sure whether the better approach is adding it there or to the skeleton

Issue #1397, opened 2021-05-10
Prerendering documentation clarification


benmccann:
The docs say:

>In other words, any app that involves user sessions or authentication is not a candidate for adapter-static, even if individual pages within an app are suitable for prerendering.

This is perhaps not quite precise enough and was causing some confusion on Discord. E.g. one user said they wanted to do anything involving auth in `onMount`. There was also a question of whether this restriction still exists with ssr and hydration disabled

johnnysprinkles:
Ah that was me! Yes I wanted to represent the runtime-agnostic, serve from non-Node, static SSR camp, where all upstream service calls are made from the client. In this scenario per-rendering is per route and a parameterized route would just have nothing for the parameters from the server's point of view.

The other issue is how to pre-render each route multiple times with varying parameters (not request-specific parameters obviously, but other more static dimension like language).

Quick summary of the angle I'd be coming from on this, which I also posted in the discord channel. https://docs.google.com/document/d/1gqleERm2L6zD6HR4VcBCn19Q9Q6t9hdb3b-7_--08iI/edit

SumitBando:
The "When not to export" verbiage is carried over from Sapper days: https://sapper.svelte.dev/docs#When_not_to_export
Undaunted by the undersell by @Rich-Harris , we went ahead and built a multi hundred page, authenticated, personalized mental health web app, exported and served out of CDN , in production.

My motivation was not having to deploy / maintain / secure NodeJS servers and pay AWS more for SSR.
Our SSR happens during build/deployment on Netlify or Github servers, I see some people refer to this as Static site generation (SSG).
The fact that the CDN servers, physically close to my users, start delivering pages, instead of SSR servers in Oregon means Time-to-first-byte is phenomenally better.

I submitted the PR without knowing the existence of this open ticket.
I will refine the PR to address the review comments.

johnnysprinkles:
That's cool to hear @SumitBando. I think this issue relates to the new issue I filed, #1533. Your motivation is not paying for Node servers, mine is that at my workplace they literally don't allow us to run Node at runtime in production. Too many bad stories about malicious NPMs I guess.

[...]

PR #1525, opened 2021-05-23
Add appendix regarding different rendering methods

benmccann:
This is one piece of background that a lot of new users seem not to have a good understanding of. While we assume a background in HTML, CSS, and Svelte some discussion of rendering might be warranted since SvelteKit at its core is a rendering engine. It helps explain what SvelteKit is, how it's unique, and the value it provides.

This closes https://github.com/sveltejs/kit/pull/1508 which inspired this PR as a possible alternative way of covering this info. Also closes https://github.com/sveltejs/kit/issues/1397


documentation/docs/00-introduction.md
```diff
@@ -19,6 +19,38 @@ Building an app with all the modern best practices — code-splitting, offline s
...
+
+Pre-rendered pages are not limited to static content. You can build personalized pages if user-specific data is fetched and rendered client-side. This is subject to the caveat that you will experience the downsides of not doing SSR for that content as discussed above.
+
+#### SSG
+
+Static Site Generation (SSG) is a term that refers to a site where every page is prerendered. This is what SvelteKit's `adapter-static` does. SvelteKit was not built to do only static site generation like some tools and so may not scale as well to efficiently render a very large number of pages as tools built specifically for that purpose. However, in contrast to most purpose-built SSGs, SvelteKit does nicely allow for mixing and matching different rendering types on different pages. One benefit of fully prerendering a site is that you do not need to maintain or pay for servers to perform SSR. Once generated, the site can be served from CDNs, leading to great "time to first byte" performance. This delivery model is often referred to as JAMstack.
```
SumitBando, line 44: This intro text explaining the key concepts will be very valuable for new users.

A reference to PWA somewhere may be useful.
The SSG style was critical for building our app as PWA, since the device may not be able to reach the SSR servers when needed. Pre-rendered pages are cached on device, app is mostly functional offline, and syncs with backend when connected.

documentation/docs/11-ssr-and-javascript.md
```diff
@@ -64,13 +62,7 @@ In many cases, you'll only want to prerender specific pages in your app. You'll
...
 
-#### When not to prerender
-
-The basic rule is this: for a page to be prerenderable, any two users hitting it directly must get the same content from the server.
-
-> In other words, any app that involves user sessions or authentication is _not_ a candidate for `adapter-static`, even if individual pages within an app _are_ suitable for prerendering.
-
-Note that you can still prerender pages that load data based on the page's parameters, like our `src/routes/blog/[slug].svelte` example from earlier. The prerenderer will intercept requests made inside `load`, so the data served from `src/routes/blog/[slug].json.js` will also be captured.
+> Not all pages are suitable for prerendering. Any content that is prerendered will be seen by all users and will not be personalized. For a more in-depth discussion, see [the prerendering section of the introduction](#introduction-prerendering).
```
SumitBando, line 65: Still sounds limiting. Consider adding:

Any content that is prerendered will be seen by all users and will not be personalized , any personalization will need to performed in the browser.

documentation/docs/00-introduction.md
```diff
@@ -19,6 +19,38 @@ Building an app with all the modern best practices — code-splitting, offline s
 
 You don't need to know Svelte to understand the rest of this guide, but it will help. In short, it's a UI framework that compiles your components to highly optimized vanilla JavaScript. Read the [introductory blog post](https://svelte.dev/blog/svelte-3-rethinking-reactivity) and the [tutorial](https://svelte.dev/tutorial) to learn more.
 
+### Rendering on the web
+
+The core of SvelteKit provides a highly configurable rendering engine. This section describes some of the terms used when discussing that rendering engine. A reference for setting these options is provided later in the documentation.
+
+#### SSR
+
+Server-side rendering (SSR) is the generation of the page contents on the server. SSR is generally better for SEO. While some search engines can index content that is dynamically generated on the client-side it may take longer even in these cases. It also tends to improve perceived performance and makes your app accessible to users if JavaScript fails or is disabled (which happens [more often than you probably think](https://kryogenix.org/code/browser/everyonehasjs.html)).
```
SumitBando, line 28: better ... requires something to be better than
you have not yet mentioned CSR, which you are trying to dis

SumitBando:
The intro text explaining the key concepts will be very valuable for new users.

The intro text reflects a bias towards SSR being generally superior.

SSR is clearly superior / only choice for devices without modern JavaScript.
Other than that, it depends.

On https://github.com/sveltejs/kit/pull/1508 you wrote:  'Instead of making a single round trip over the internet to the backend server, you are more likely to have to make multiple round trips over the internet to render the page'.
For many years complex applications use a backend-for-frontend pattern, so that each page / refresh uses a single purpose-built API call. With the widespread adoption of GraphQL, everybody who needs it, does so.

If the prerendered pages are already cached on the client, APIs have to do less work and transmits fewer bytes.
So SSG+CSR wins.

The SSG style was critical for building our app as PWA, since the device may not be able to reach the SSR servers when needed. Pre-rendered pages are cached on device, app is mostly functional offline, and syncs with backend when connected.
A reference to PWA somewhere may be useful.

documentation/docs/00-introduction.md
benmccann: I'll add it under prerendering since it applies there as well and SSG is just a special case of that

documentation/docs/11-ssr-and-javascript.md
benmccann: I'll drop the "will not be personalized" instead to keep this from getting too long and repeating the intro content

documentation/docs/00-introduction.md
benmccann: changed "better" to "preferred". not trying to dis any approach but just point out the pros and cons of each

benmccann:
> On #1508 you wrote: 'Instead of making a single round trip over the internet to the backend server, you are more likely to have to make multiple round trips over the internet to render the page'. For many years complex applications use a backend-for-frontend pattern, so that each page / refresh uses a single purpose-built API call. With the widespread adoption of GraphQL, everybody who needs it, does so.

Not sure if it really matters unless it's being added in the text here, but even with GraphQL you'd probably still end up making more round trips than a traditionally SSR'd page: once to get the HTML, once to get the JS, once to make the API call

> If the prerendered pages are already cached on the client, APIs have to do less work and transmits fewer bytes. So SSG+CSR wins.

It's rare that you could cache the whole page such that you'd avoid API calls, but in the case that you could do this you could cache the page with any rendering solution so it's not a point in favor of one approach over another.

> So the correct statement is 'SSG is best for SEO'

It would be basically equivalent to SSR, but I don't expect it'd necessarily be much better unless there's something I'm overlooking

SumitBando:
> Not sure if it really matters unless it's being added in the text here, but even with GraphQL you'd probably still end up making more round trips than a traditionally SSR'd page: once to get the HTML, once to get the JS, once to make the API call
> 
> It's rare that you could cache the whole page such that you'd avoid API calls, but in the case that you could do this you could cache the page with any rendering solution so it's not a point in favor of one approach over another.

My observations (the submitted PR) were not theoretical, but based on happy results building a PWA with Svelte, designed to work offline.
The statically generated html+js are cached by ServiceWorker.
If online, each page typically needs a single API call, the payload is small compared to html/css.
If offline, works off of localStorage.

Rich-Harris:
This is a lot of content to have before the 'getting started' block — the introduction should really be just that, people shouldn't have to scroll through multiple screens before they see `npm init svelte@next my-app`.

I'd argue it's also a little confusing to have

> The core of SvelteKit provides a highly configurable rendering engine

so soon after

> SvelteKit is a framework for building extremely high-performance web apps

Perhaps it would make sense to have a separate 'terminology' or 'glossary' section at the end?

benmccann:
Thanks for the review @Rich-Harris. I've updated it if you've got a chance to take another look

Issue #1523, opened 2021-05-22
Denial of service vulnerability in adapter-node


lovasoa:
**Describe the bug**

There is a remote denial of service vulnerability that allows attackers to crash the default web server for adapter-node. 

**Logs**

The only logs I have are this:

```
May 22 22:36:08 sanipasse node[117756]: node:internal/process/promises:246
May 22 22:36:08 sanipasse node[117756]:           triggerUncaughtException(err, true /* fromPromise */);
May 22 22:36:08 sanipasse node[117756]:           ^
May 22 22:36:08 sanipasse node[117756]: Error: aborted
May 22 22:36:08 sanipasse node[117756]:     at connResetException (node:internal/errors:683:14)
May 22 22:36:08 sanipasse node[117756]:     at abortIncoming (node:_http_server:592:17)
May 22 22:36:08 sanipasse node[117756]:     at socketOnClose (node:_http_server:586:3)
May 22 22:36:08 sanipasse node[117756]:     at Socket.emit (node:events:377:35)
May 22 22:36:08 sanipasse node[117756]:     at TCP.<anonymous> (node:net:661:12) {
May 22 22:36:08 sanipasse node[117756]:   code: 'ECONNRESET'
May 22 22:36:08 sanipasse node[117756]: }
May 22 22:36:09 sanipasse systemd[1]: sanipasse.service: Main process exited, code=exited, status=1/FAILURE
May 22 22:36:09 sanipasse systemd[1]: sanipasse.service: Failed with result 'exit-code'.
```

**To Reproduce**

Run the default template sveltekit application with adapter-node on port 3000, then, from a terminal : 

```
$ nc localhost 3000
POST /x HTTP/1.1
Content-Length: 1000
Content-Type: application/json

^C
```

**Expected behavior**

No one should be able to remotely crash the server.
Whatever their payload, the sveltekit server should not crash, especially before entering user-defined code.

**Stacktraces**

```
node:internal/process/promises:246
           triggerUncaughtException(err, true /* fromPromise */);
          ^
 Error: aborted
    at connResetException (node:internal/errors:683:14)
    at abortIncoming (node:_http_server:592:17)
    at socketOnClose (node:_http_server:586:3)
    at Socket.emit (node:events:377:35)
    at TCP.<anonymous> (node:net:661:12) {
```

**Information about your SvelteKit Installation:**

- kit 1.0.0-next.107
- node v16.1.0
- Your adapter : node

**Severity**

[...]

PR #1528, opened 2021-05-23
fix: improve `getRawBody` parsing & handle error(s)

lukeed:
### Tests

### Changesets

### Summary

* The `getRawBody` utility did not handle the case where the supplied body actually exceeds the `Content-Length` header. Added that logic. (This is impossible to spoof on Playwright/browser clients. Can add a test once a unit testing system is in place.)
* fixed case where `octect-stream` request bodies were double-fulfilled
* fixed `rawBody` type interfaces (missing `null` possibility)
* Handle `getRawBody` possible error(s) in consumers/adapters

---

Related to (but does not close) #1523

---

### DIFF
--- a/packages/adapter-node/src/server.js
+++ b/packages/adapter-node/src/server.js
@@ -40,12 +40,22 @@ export function createServer({ render }) {
 		prerendered_handler,
 		async (req, res) => {
 			const parsed = new URL(req.url || '', 'http://localhost');
+
+			let body;
+
+			try {
+				body = await getRawBody(req);
+			} catch (err) {
+				res.statusCode = err.status || 400;
+				return res.end(err.reason || 'Invalid request body');
+			}
+
 			const rendered = await render({
 				method: req.method,
 				headers: req.headers, // TODO: what about repeated headers, i.e. string[]
 				path: parsed.pathname,
-				rawBody: await getRawBody(req),
-				query: parsed.searchParams
+				query: parsed.searchParams,
+				rawBody: body
 			});
 
 			if (rendered) {
--- a/packages/kit/src/core/dev/index.js
+++ b/packages/kit/src/core/dev/index.js
@@ -153,7 +153,14 @@ class Watcher extends EventEmitter {
 					const root = (await this.vite.ssrLoadModule(`/${this.dir}/generated/root.svelte`))
 						.default;
 
-					const rawBody = await getRawBody(req);
+					let body;
+
+					try {
+						body = await getRawBody(req);
+					} catch (err) {
+						res.statusCode = err.status || 400;
+						return res.end(err.reason || 'Invalid request body');
+					}
 
 					const host = /** @type {string} */ (this.config.kit.host ||
 						req.headers[this.config.kit.hostHeader || 'host']);
@@ -165,7 +172,7 @@ class Watcher extends EventEmitter {
 							host,
 							path: parsed.pathname,
 							query: new URLSearchParams(parsed.query),
-							rawBody
+							rawBody: body
 						},
 						{
 							amp: this.config.kit.amp,
--- a/packages/kit/src/core/node/index.js
+++ b/packages/kit/src/core/node/index.js
@@ -1,44 +1,45 @@
 /**
  * @param {import('http').IncomingMessage} req
- * @returns {Promise<import('types/hooks').StrictBody>}
+ * @returns {Promise<import('types/hooks').StrictBody | null>}
  */
 export function getRawBody(req) {
 	return new Promise((fulfil, reject) => {
 		const h = req.headers;
 
 		if (!h['content-type']) {
-			fulfil(null);
-			return;
+			return fulfil(null);
 		}
 
 		req.on('error', reject);
 
 		const length = Number(h['content-length']);
 
-		/** @type {Uint8Array} */
-		let data;
+		// https://github.com/jshttp/type-is/blob/c1f4388c71c8a01f79934e68f630ca4a15fffcd6/index.js#L81-L95
+		if (isNaN(length) && h['transfer-encoding'] == null) {
+			return fulfil(null);
+		}
 
-		if (!isNaN(length)) {
-			data = new Uint8Array(length);
+		let data = new Uint8Array(length || 0);
 
-			let i = 0;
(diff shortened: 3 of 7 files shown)


packages/adapter-node/src/server.js
```diff
@@ -29,9 +29,9 @@ export function createServer({ render }) {
 
 	const assets_handler = fs.existsSync(paths.assets)
 		? sirv(paths.assets, {
-				maxAge: 31536000,
```
lukeed, line 32: I think `yarn format` and/or my editor did this

packages/kit/src/core/dev/index.js
```diff
@@ -279,9 +284,8 @@ class Watcher extends EventEmitter {
 												.map(
 													(error) => `
 												<h2>${error.severity}</h2>
-												<p>Line ${error.line}, column ${error.col}: ${error.message} (<a href="${error.specUrl}">${
-														error.code
-													}</a>)</p>
+												<p>Line ${error.line}, column ${error.col}: ${error.message} (<a href="${error.specUrl}">${error.code
+														}</a>)</p>
```
lukeed, line 288: linter and/or editor

packages/kit/src/core/node/index.js
```diff
@@ -1,44 +1,46 @@
...
-				data.set(chunk, i);
-				i += chunk.length;
-			});
-		} else {
-			// https://github.com/jshttp/type-is/blob/c1f4388c71c8a01f79934e68f630ca4a15fffcd6/index.js#L81-L95
-			if (h['transfer-encoding'] === undefined) {
-				fulfil(null);
-				return;
-			}
+				const new_len = offset + Buffer.byteLength(chunk);
```
lukeed, line 28: Using the `chunk.length` was a bug. Incoming `chunks` can be strings, but even if they're still Buffer instances, `length` is not always the same as `byteLength`, which is what `content-length` tracks

packages/kit/src/core/node/index.js
```diff
@@ -48,11 +50,11 @@ export function getRawBody(req) {
 			const [type] = h['content-type'].split(/;\s*/);
 
 			if (type === 'application/octet-stream') {
-				fulfil(data);
+				return fulfil(data);
```
lukeed, line 53: Without this `return`, then decoding continued (below)

packages/kit/types/endpoint.d.ts
```diff
@@ -19,7 +19,7 @@ type JSONValue =
 	| null
 	| Date
 	| JSONValue[]
-	| { [key: string]: JSONValue };
+	| { [ key: string ]: JSONValue };
```
lukeed, line 22: linter / editor, sorry

packages/adapter-node/src/server.js
```diff
@@ -40,12 +40,20 @@ export function createServer({ render }) {
 		prerendered_handler,
 		async (req, res) => {
 			const parsed = new URL(req.url || '', 'http://localhost');
+
+			try {
+				var body = await getRawBody(req); // eslint-disable-line no-var
```
benmccann, line 45: the cleaner way to right this without needing to silence eslint is probably:
```
let body;
try {
  body = ...
} catch (err) {
    ...
}

```

lukeed: Style preference, var is much cleaner to me

lukeed: Is also exactly what var is made for 🙂

Zerotask: You should avoid using `var` though and better stick with `let` in general. For that case it's even better to use `const`, since body isn't changing afterwards.

lukeed: You cannot use `const` here, because you have to hoist the value outside of the `try/catch` scope. The only (viable) options are to define `let` early & reassign, or to use `var` – which is it's designed & intended purpose. 

The eslint rule exists to prevent you from using `var` inline / in the same scope. In other words, it was made to prevent this kind of `var` usage:

```js
function hello(name, prefix) {
  var fullname = prefix + ' ' + name;
  console.log('Hello', fullname);
}
```

benmccann: I don't really care and am not going to nitpick on it too much, but I think a lot of what you're saying isn't true. `var` wasn't made specifically for the purpose of allowing access outside block scope. It preceded `let` and `const` which were introduced due to safety and readability issues with `var`. The eslint rule is meant to discourage use of `var` in its entirety: https://eslint.org/docs/rules/no-var#rule-details. If eslint wanted to allow the usage you were suggesting the rule certainly could have been written to do so.

But what's the better way to write this doesn't really matter as much as following what eslint is suggesting. If everyone committed code with their own individual style preferences the project would be a mess. The point of eslint and prettier is to enforce a consistent style and even if some people might disagree with aspects of it, it's still better that we have consistency than things written a bunch of different ways with `eslint-ignore` scattered throughout the code

lukeed: Agree it's an aside, but `var` has always behaved this way, by design. When `let` & `const` were introduced, it was a design choice to have both of them be block-scoped instead of being function-scoped.

benmccann: thanks for changing!

packages/kit/src/core/node/index.js
```diff
@@ -1,44 +1,46 @@
...
 		req.on('error', reject);
 
 		const length = Number(h['content-length']);
 
-		/** @type {Uint8Array} */
-		let data;
-
-		if (!isNaN(length)) {
-			data = new Uint8Array(length);
+		// https://github.com/lukeed/polka/blob/next/packages/parse/index.js#L17
```
benmccann, line 17: do we need both of these links? I would just include the second one since it explains it whereas the first one doesn't really add any additional context

benmccann:
lgtm. I wonder if `getRawBody` should have a couple unit tests

packages/kit/src/core/node/index.js
benmccann, line 17:
```suggestion
```

benmccann:
I rebased this and am going to go ahead and merge it. We can always add more tests later

PR #1556, opened 2021-05-25
create-svelte: add svelte-check for TS

### LINKED ISSUE #1536: svelte-kit build swallows typescript errors
**Describe the bug**
Errors in .ts files are silently ignored when running `svelte-kit build`.

**Logs**
N/A

**To Reproduce**
1. Enable some feature in `tsconfig.json`, such as `strictNullChecks` or `noUnusedLocals`, then write some code that breaks that rule. For example, setting `noUnusedLocals: true` then writing `let unusedVar = 42;`.
2. Run `svelte-kit build`

**Expected behavior**
I expect that `svelte-kit build` would fail (exit != 0) and give a helpful error message. Instead it just silently succeeds, even though compiling the exact same code with `tsc` correctly errors.

**Stacktraces**
N/A

N/A

**Severity**
Very annoying, means that I can't rely on the build process to catch errors so I have to spend extra time checking for them manually and also worry about an avoidable bug slipping through the cracks.

**Additional context**
N/A

dummdidumm:
This add svelte-check along with two validate scripts to the package.json if the user selects TS. This uncovered missing type definitions for the cookie package in the default template, which are now added as well if the user selects TS.

Closes #1536 , if noone has objections to doing it this way. Alternatives:
- add this to `svelte-kit`. This would add needless dependencies for those who don't want to use it, and does not bring any benefit to me as it only hides the invocation and cannot add any deeper integration
- adjust the TS templates so that `svelte-check` is always invoked prior to `svelte-kit`. I'm slightly in favor of this, but it slows down the build a little obviously

### DIFF
--- a/packages/create-svelte/shared/+typescript/package.json
+++ b/packages/create-svelte/shared/+typescript/package.json
@@ -1,7 +1,12 @@
 {
+	"scripts": {
+		"validate": "svelte-check --tsconfig ./tsconfig.json",
+		"validate:watch": "svelte-check --tsconfig ./tsconfig.json --watch"
+	},
 	"devDependencies": {
 		"typescript": "^4.0.0",
 		"tslib": "^2.0.0",
+		"svelte-check": "^1.6.0",
 		"svelte-preprocess": "^4.0.0"
 	}
 }
--- a/packages/create-svelte/shared/+typescript/package.json
+++ b/packages/create-svelte/shared/+typescript/package.json
@@ -6,7 +6,7 @@
 	"devDependencies": {
 		"typescript": "^4.0.0",
 		"tslib": "^2.0.0",
-		"svelte-check": "^1.6.0",
+		"svelte-check": "^2.0.0",
 		"svelte-preprocess": "^4.0.0"
 	}
 }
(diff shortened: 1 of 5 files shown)


packages/create-svelte/shared/+typescript/package.json
```diff
@@ -1,7 +1,12 @@
 {
+	"scripts": {
+		"validate": "svelte-check --tsconfig ./tsconfig.json",
```
benmccann, line 3: I'd probably lean towards calling it `check` instead of `validate`

packages/create-svelte/shared/+typescript/package.json
```diff
@@ -1,7 +1,12 @@
 {
+	"scripts": {
+		"check": "svelte-check --tsconfig ./tsconfig.json",
```
ignatiusmb, line 3: Does `svelte-check` already supports `.ts` files, or should we also add `tsc --noEmit` here?

dummdidumm: Since version 1.6.0 it does also check ts/js files when using the `--tsconfig` flag

ignatiusmb (approved):
Really nice addition, especially for TS users

benmccann (approved):
lgtm. only reason I'm not merging is in case you wanted more opinions on how to setup the `package.json` commands first. but if you're happy with how it is then I'd say go ahead and hit the green button

dummdidumm:
I'll wait a little more, also because there's likely a new major version of `svelte-check` very soon which I'd use here then.

PR #1559, opened 2021-05-26
feat: pass validated svelte config to adapters

### LINKED ISSUE #1435: [adapter api]: pass svelte.config.js config object to adapter

The [Firebase adapter](https://github.com/jthegedus/svelte-adapter-firebase) reads a provided `firebase.json` configuration to determine destination dirs for `utils.copy_{static/client}_files` etc.

If the destination dir defined in `firebase.json` is the same as the `svelte.config.js:kit.files.assets` dir then we end up with a mess of a dir similar to the issue I am experiencing from https://github.com/sveltejs/kit/issues/587

Either:

- **preferred solution** `utils.copy_{static/client}_files` error when the destination dir is the same as the source dir (notably this misses `utils.prerender`, but calling prerender after the other copies would be enough)
- the adapter receives the SvelteKit config object so these types of checks can be made within the adapter without having to read in the config file itself

It will help reduce noise from known causes in the adapter workflow if I can nudge my users properly

**Additional context**

https://github.com/jthegedus/svelte-adapter-firebase/issues/56 is the tracking issue for identifying this conflict of a matching `source:dest` for static assets.

[update] I have temporarily worked around this by reading the `svelte.config.js` in the adapter, though obviously not ideal as it's the source config and not the computed final config.

jthegedus:
Pass validated Svelte config to Adapters.

As vaguely discussed in #1435 and #1436 adapters may need the svelte config settings, but currently must parse the files themselves, which is brittle and unnecessary.

An example of this need comes from adapters reading specific cloud provider config to determine input and output dirs. The Firebase adapter reads the `firebase.json` file to determine the output dir for the static assets, but by default this is `public` and so conflicts with the default of `kit.files.assets`. The adapter checks that these are different, but must first read the svelte config to determine the actual value of `svelte.config.js:kit.files.assets` and deal with it not being listed because we are reading the object before SvelteKit defaults are applied etc.

Closes #1435 

Following the contribution guide, `pnpm i` changed the lockfile :shrug: 

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/adapt/index.js
+++ b/packages/kit/src/core/adapt/index.js
@@ -7,7 +7,6 @@ import { get_utils } from './utils.js';
  * @param {import('types/internal').BuildData} build_data
  * @param {{ cwd?: string, verbose: boolean }} opts
  */
-
 export async function adapt(config, build_data, { cwd = process.cwd(), verbose }) {
 	const { name, adapt } = config.kit.adapter;
 
--- a/packages/kit/src/core/adapt/index.js
+++ b/packages/kit/src/core/adapt/index.js
@@ -7,6 +7,7 @@ import { get_utils } from './utils.js';
  * @param {import('types/internal').BuildData} build_data
  * @param {{ cwd?: string, verbose: boolean }} opts
  */
+
 export async function adapt(config, build_data, { cwd = process.cwd(), verbose }) {
 	const { name, adapt } = config.kit.adapter;
 
@@ -14,7 +15,7 @@ export async function adapt(config, build_data, { cwd = process.cwd(), verbose }
 
 	const log = logger({ verbose });
 	const utils = get_utils({ cwd, config, build_data, log });
-	await adapt(utils);
+	await adapt({ utils, svelteConfig: config });
 
 	log.success('done');
 }
--- a/packages/kit/src/core/adapt/index.js
+++ b/packages/kit/src/core/adapt/index.js
@@ -15,7 +15,7 @@ export async function adapt(config, build_data, { cwd = process.cwd(), verbose }
 
 	const log = logger({ verbose });
 	const utils = get_utils({ cwd, config, build_data, log });
-	await adapt({ utils, svelteConfig: config });
+	await adapt({ utils, config });
 
 	log.success('done');
 }
--- a/packages/kit/types/config.d.ts
+++ b/packages/kit/types/config.d.ts
@@ -22,13 +22,7 @@ export type AdapterUtils = {
 
 export type Adapter = {
 	name: string;
-	adapt: ({
-		utils,
-		svelteConfig
-	}: {
-		utils: AdapterUtils;
-		svelteConfig: ValidatedConfig;
-	}) => Promise<void>;
+	adapt: ({ utils, config }: { utils: AdapterUtils; config: ValidatedConfig }) => Promise<void>;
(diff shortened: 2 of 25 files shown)


packages/kit/src/core/adapt/index.js
```diff
@@ -7,14 +7,15 @@ import { get_utils } from './utils.js';
...
+
 export async function adapt(config, build_data, { cwd = process.cwd(), verbose }) {
 	const { name, adapt } = config.kit.adapter;
 
 	console.log(colors.bold().cyan(`\n> Using ${name}`));
 
 	const log = logger({ verbose });
 	const utils = get_utils({ cwd, config, build_data, log });
-	await adapt(utils);
+	await adapt({ utils, svelteConfig: config });
```
jthegedus, line 18: trying to communicate this is not the user's config for the adapter, but in fact, the `svelte.config.js` config.

dummdidumm:
The changelock change may have happened because you are on a different version of pnpm (not latest major perhaps)

jthegedus:
> The changelock change may have happened because you are on a different version of pnpm (not latest major perhaps)

I am on `6.3.0`. I updated my fork aswell before creating branch. I can remove the file from PR as it seems unnecessary.

packages/kit/types/config.d.ts
```diff
@@ -22,7 +22,13 @@ export type AdapterUtils = {
 
 export type Adapter = {
 	name: string;
-	adapt: (utils: AdapterUtils) => Promise<void>;
+	adapt: ({
+		utils,
+		svelteConfig
+	}: {
+		utils: AdapterUtils;
+		svelteConfig: ValidatedConfig;
```
benmccann, line 30: maybe if we add some documentation here explaining that it's the `svelte.config.js` file then we can use the shorter `config` name?

jthegedus: Docs as in JSDoc in the type definition of in the Adapter Author docs which are yet to exist?

benmccann: I'm not quite sure I understood, but I was thinking something like:

```
export type AdapterInput = {
	utils: AdapterUtils;
	/**
	 * The configuration provided by the user in svelte.config.js
	 */
	config: ValidatedConfig;
};

export type Adapter = {
	name: string;
	adapt: (input: AdapterInput) => Promise<void>;
};
```

Although now I'm wondering if that's even necessary because we have documented the type as `ValidatedConfig` which should be sufficient

jthegedus: I was asking if that type of comment you suggest will be recognised and thus show in editor highlighting alongside other type information.

>Although now I'm wondering if that's even necessary because we have documented the type as `ValidatedConfig` which should be sufficient

With the adapter getting the `kit.adapter` func params and `utils`, I feel `config` is not enough information, even with the ValidatedConfig type. Will rename so we can discuss.

packages/kit/src/core/adapt/index.js
```diff
@@ -7,14 +7,15 @@ import { get_utils } from './utils.js';
  * @param {import('types/internal').BuildData} build_data
  * @param {{ cwd?: string, verbose: boolean }} opts
  */
+
```
benmccann, line 10: should probably remove this blank line that got inserted

benmccann:
thanks!

PR #1572, opened 2021-05-27
Passing the Rollup output options through to Vite

### LINKED ISSUE #1571: Config for "manualChunks" not making it to Vite
**Describe the bug**
If you put something like this in your svelte config file, the manualChunks never make it to Vite because `rollupOptions.output` isn't spread in.

    kit: {
      vite: {
        build: {
          rollupOptions: {
            output: {
              manualChunks: {}

Why does that matter?

The vendor chunk logic comes from [Vite here](https://github.com/vitejs/vite/blob/4112c5d103673b83c50d446096086617dfaac5a3/packages/vite/src/node/build.ts#L564), and that might not always be what you want. That logic just takes the NPM dependencies for all your entry points and throws them into one vendor chunk. In my case I'm using D3 to make some graphs on my index route, but I'd like to not have all the other pages pay that JS download price on initial load.

There's an easy fix mentioned in https://github.com/vitejs/vite/issues/1683 but that doesn't work for SvelteKit at the moment because the output options are just overridden.

If we can set the `manualChunks` option we have a lot of flexibility. We can set to empty object to simply delete the vendor chunk concept, or if I had multiple pages using D3 and wanted shared cacheability I could make a manual D3 chunk like:

    manualChunks: {
      d3: [
        'd3-array',
        'd3-shape',
        'd3-scale',
      ]
    }

Pull request to come shortly on this, it's a simple one-liner.

johnnysprinkles:
This allows us to have more control of the rollup chunking when doing a production build. In particular, it allows us to specify `manualChunks` so we can decide what goes in the vendor chunk, or to not have a vendor chunk at all and simply have large page-specific bundles.

`vite.build()` does occur in three places in that file, I only updated in the one place, and verified it took effect when doing an "npm run build". Not sure if I should also update in the other places.

Fixes #1571

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/build/index.js
+++ b/packages/kit/src/core/build/index.js
@@ -149,6 +149,9 @@ async function build_client({
 				...(user_config.build && user_config.build.rollupOptions),
 				input,
 				output: {
+					...(user_config.build &&
+						user_config.build.rollupOptions &&
+						user_config.build.rollupOptions.output),
 					entryFileNames: '[name]-[hash].js',
 					chunkFileNames: 'chunks/[name]-[hash].js',
 					assetFileNames: 'assets/[name]-[hash][extname]'


johnnysprinkles:
Actually, wouldn't it make sense to do the same change in `build_server` and `build_service_worker` as well? For people not providing any rollupOptions.output config, it has no effect on them. I can go ahead and update the PR if so.

johnnysprinkles:
Would it actually make more sense to fix this in a general way, making a simple `deepMerge()` function for example and using that to recursively merge the user config with SvelteKit's own Vite config? Definitely not suggesting we take on a new dependency just for a merge function, we can simply write one.

Rich-Harris:
Huh, that seems like a weird default on Vite's part. I'm sure there's a rationale I'm not aware of. I suspect we should be overriding that by default, regardless of whether we respect the user's configuration here.

I think the general `deepMerge` solution probably makes sense, yeah. I think it would be good if we took that opportunity to explicitly warn if the user is setting leaf properties that are getting overridden by SvelteKit — e.g. in a case like this...

```js
// svelte.config.js
export default {
  vite: {
    build: {
      outDir: 'custom-out-dir'
    }
  }
};
```

...we might print a loud warning saying something like 

> The value for `vite.build.outDir` specified in `svelte.config.js` has been ignored, as this is controlled by SvelteKit

We might need some way to indicate that a property shouldn't be merged. For example while we want to merge plugins...

```js
plugins: [
  ...(user_config.plugins || []),
  svelte({
    emitCss: !config.kit.amp
  })
]
```

...we _don't_ want to respect the user's `optimizeDeps.entries` since there's no sensible value the user could supply there. Not 100% sure what that looks like.

> Actually, wouldn't it make sense to do the same change in `build_server` and `build_service_worker` as well?

Generally speaking it's very possible that some options make sense in one context but not in another. We sort of anticipated this by making `config.kit.vite` a function (objects get turned into functions)...

```js
const user_config = config.kit.vite();
```

...which could in theory receive options...

```js
const user_config = config.kit.vite({ type: 'service-worker' });
```

...but that's a bit of a digression and we probably don't need to worry about that right now.

johnnysprinkles:
I think Vite just got that from the [Rollup docs](https://rollupjs.org/guide/en/#outputmanualchunks) where they give "vendor chunk" as an example.

But so added a `deep_merge` function to core/utils.js. Not sure if it should go there, or maybe it would be better as a separate "merge.js" file in the same location.

I was just going to do objects and let everything else be atomic, but the way the `vite.plugins` merge together as an array made me think it should be 1. Objects interleave together, 2. Arrays combine, 3. Anything else just overwrites.

For the error message, I started with `new Error` but I guess it shouldn't be a fatal error, so I did `console.error` but that wasn't colorfully highlighted, so I ended up with `console.error(colors.bold().red(...))`. I'll drag a couple screenshots in.

johnnysprinkles:
Also not sure where the tests should go for deep_merge, I just put along side.

johnnysprinkles:
Here's with "new Error":

Screen Shot 2021-05-30 at 8 11 46 PM

johnnysprinkles:
And here's with `console.error(colors.bold().red())`. Not sure if we need the prefix saying which function it happened in, but each build() function does have different SvelteKit config for Vite.

Screen Shot 2021-05-30 at 8 36 27 PM

johnnysprinkles:
Maybe it should show the whole path, `kit.vite` on down, and maybe format that part of the error somehow? Here's in italic:

Screen Shot 2021-05-30 at 9 26 48 PM

johnnysprinkles:
And, this is a pretty significant change and somewhat of a regression risk. We can always just do that first commit instead, mixing it in surgically.

johnnysprinkles:
So I was concerned about this recursive merge because it recurses down through objects that might not be "plain old JS objects" but may have stuff in closure, or rely on `this`, etc. But after a bit of investigation I'm less concerned now. If they rely on values in closure that will still be there if you copy all the properties. There could be `this` issues, but that would be easy to fix by checking for "plain old JS objects" i.e. objects where the constructor === Object. So instead of:

    const is_object = (x) => typeof x === 'object' && !Array.isArray(x);

Might be more like:

    const is_plain_object (x) => typeof x === 'object' && !Array.isArray(x) && x.constructor === Object;

Happy to make this change if desired. Plugins are certain to be fine already, recursion stops at arrays so they never get deep cloned.

packages/kit/src/core/build/index.js
```diff
@@ -157,22 +156,33 @@ async function build_client({
...
 	});
 
+	conflicts.forEach((conflict) => {
+		console.error(
+			colors
+				.bold()
+				.red(
+					`build_client: The value for ${colors.italic(
+						`kit.vite.${conflict}`
+					)} specified in svelte.config.js has been ignored, as this is controlled by SvelteKit`
```
benmccann, line 179:
```suggestion
					)} specified in svelte.config.js has been ignored. This option is controlled by SvelteKit.`
```

benmccann:
@johnnysprinkles thanks for this! as a heads up, there's quite a few errors being thrown by `pnpm check`

johnnysprinkles:
Thanks for taking a look @benmccann, yep getting up to speed on the process here and fumbling through the jsdoc typing. I'll have an update later today addressing your comments and passing all the checks.

packages/kit/src/core/build/index.js
```diff
@@ -77,6 +78,25 @@ export async function build(config, { cwd = process.cwd(), runtime = '@sveltejs/
...
+function print_vite_config_conflicts(scope, conflicts) {
+	const prefix = scope ? scope + ': ' : '';
+	conflicts.forEach((conflict) => {
+		console.error(
+			colors
+				.bold()
+				.red(
+					`${prefix}The value for ${colors.italic(
+						`kit.vite.${conflict}`
+					)} specified in svelte.config.js has been ignored. This is controlled by SvelteKit.`
```
benmccann, line 94:
```suggestion
					)} specified in svelte.config.js has been ignored. This option is controlled by SvelteKit.`
```

packages/kit/src/core/build/index.js
```diff
@@ -77,6 +78,25 @@ export async function build(config, { cwd = process.cwd(), runtime = '@sveltejs/
 	};
 }
 
+/**
+ * @param {string?} scope used to prefix the error message
+ * @param {string[]} conflicts array of conflicts in dotted notation
+ */
+function print_vite_config_conflicts(scope, conflicts) {
```
benmccann, line 85: do we want this to happen only on `build` or should it happen for `dev` as well?

johnnysprinkles: Oh yeah, I was focused on the rollup option, which aren't applicable to dev. I updated to also do dev.

johnnysprinkles:
I was pretty out of date, synced my fork and all should be good now. The dev part looks like this in iTerm2 (annoyingly it won't say what color scheme I have enabled but it's a standard one).

Screen Shot 2021-06-04 at 8 49 52 PM

johnnysprinkles:
Oh, that's "The Hulk" from https://iterm2colorschemes.com/, one of my favorites.

benmccann:
@johnnysprinkles it looks like this one will need to be rebased

packages/kit/src/core/dev/index.js
```diff
@@ -84,47 +84,54 @@ class Watcher extends EventEmitter {
...
-				...user_config.optimizeDeps,
 				entries: []
 			},
 			ssr: {
-				...user_config.ssr,
 				noExternal: get_no_external(this.cwd, user_config.ssr && user_config.ssr.noExternal)
 			}
 		});
 
+		conflicts.forEach((conflict) => {
```
benmccann, line 118: could we put the `print_vite_config_conflicts` method somewhere shared and use it here?

johnnysprinkles: Cool, how about a generic "print_config_conflicts" that might as well use the logger() I suppose. See update.

benmccann:
Cool, this is looking pretty good! It will still need to be rebased

The other thing I was thinking is that `utils` is kind of becoming a dumping ground and maybe we can organize it a little better. One idea would be to rename the `load_config` directory to `config` and have it export a `load` function as well as the new functions that you're adding in `utils`

johnnysprinkles:
A little messy but makes sense if you view the diff of all commits as a whole.

johnnysprinkles:
And yeah, I agree about load_config => config, makes sense to me.

benmccann:
thanks for all your work on this one!

johnnysprinkles:
Thanks for the review! I’m on a road trip from Seattle to Palm Springs and
did bring my laptop for more revisions but also cool to just get it in
there.

PR #1578, opened 2021-05-28
type fixes for adapter-node and adapter-static

quentin-fox:
The adapter-node and adapter-static do not actually require any arguments, which is shown in the documentation here: https://github.com/quentin-fox/kit/blob/3f82e80f14b8ebfe0d004c004b5109d70f902a47/documentation/docs/10-adapters.md#L9-L18

The Typescript types for the adapters do not indicate that the options object are optional, even though they are not required.

I also added the null type to the fallback param, since the default value is null.

### DIFF
--- a/packages/adapter-static/index.d.ts
+++ b/packages/adapter-static/index.d.ts
@@ -1,7 +1,7 @@
 declare function plugin(options?: {
 	pages?: string;
 	assets?: string;
-	fallback?: string | null;
+	fallback?: string;
 }): import('@sveltejs/kit').Adapter;
 
 export = plugin;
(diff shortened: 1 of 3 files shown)


packages/adapter-static/index.d.ts
```diff
@@ -1,7 +1,7 @@
-declare function plugin(options: {
+declare function plugin(options?: {
 	pages?: string;
 	assets?: string;
-	fallback?: string;
+	fallback?: string | null;
```
benmccann, line 4: I wonder if we should make the default `undefined` to simplify the types?

quentin-fox: From what I can tell that'd be fine, I can't see any places where fallback couldn't be `undefined` instead of `null`, and lots of the other adapters don't specify a `fallback` key in `utils.prerender`, so that seems pretty safe. Will push that change now!

PR #1580, opened 2021-05-29
Upgrade to Vite 2.3.3

benmccann:
Only upgrading to 2.3.3 instead of the latest 2.3.4 because the tests fail otherwise


johnnysprinkles:
I was excited that my change made it into Vite 2.3.5 but now I see SvelteKit won't pick that up, how much work is involved in updating the tests?

benmccann:
No idea. I didn't try. I don't know if the issue is in the tests or in Vite

PR #1586, opened 2021-05-29
fix(adapter-netlify): Append to _redirects

roschaefer:
This was easier to fix than expected. The fix should allow users to have their own `_redirects` file in the project root. If this file exists, it gets copied to `build/` before `adapter-netlify` appends the catchall rule to it.

This behaviour is [already documented](https://github.com/sveltejs/kit/tree/master/packages/adapter-netlify#using-netlify-redirect-rules), but [currently broken](https://github.com/sveltejs/kit/issues/1585).

How can I write an automated test for this adapter?

I did the following to test this manually:

1. Use `yarn link "@sveltejs/kit"` and `yarn link "@sveltejs/adapter-netlify"` in my application repository to point the two node modules to my modified versions.
2. Run `svelte-kit build` with a `_redirects` file in the project root.
3. Check if content of `build/_redirects` is expected.
4. Run `svelte-kit build` *without* a `_redirects` file in the project root.
5. Check if content of `build/_redirects` is expected.

Any guidance or help how to match the desired code quality to get this PR merged is appreciated.

close #1585

### Before submitting the PR, please make sure you do the following
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts

PR #1596, opened 2021-05-30
netlify adapter docs: add warning on redirects file placement

sw-yx:

addresses https://github.com/sveltejs/kit/pull/1586 to hopefully save others from the same confusion

### DIFF
--- a/packages/adapter-netlify/README.md
+++ b/packages/adapter-netlify/README.md
@@ -44,7 +44,7 @@ You may build your app using functionality provided directly by SvelteKit withou
 
 > [Netlify docs here](https://docs.netlify.com/routing/redirects/)
 
-Netlify supports declarative redirects via a `_redirects` file placed in your `publish` directory (usually "`build`", as configured in your `netlify.toml` above. NOT in your project root next to `netlify.toml`).
+Netlify supports declarative redirects via a `_redirects` file. Place it in the `static/` folder, NOT in your project root next to `netlify.toml`.
 
 During compilation a required "catch all" redirect rule is automatically appended to your `_redirects` file. (If it doesn't exist yet, it will be created.) That means:
 
--- a/packages/adapter-netlify/README.md
+++ b/packages/adapter-netlify/README.md
@@ -42,12 +42,18 @@ You may build your app using functionality provided directly by SvelteKit withou
 
 ### Using Netlify Redirect Rules
 
+> [Netlify docs here](https://docs.netlify.com/routing/redirects/)
+
+Netlify supports declarative redirects via a `_redirects` file placed in your `publish` directory (usually "`build`", as configured in your `netlify.toml` above. NOT in your project root next to `netlify.toml`).
+
 During compilation a required "catch all" redirect rule is automatically appended to your `_redirects` file. (If it doesn't exist yet, it will be created.) That means:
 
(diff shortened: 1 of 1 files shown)


packages/adapter-netlify/README.md
```diff
@@ -42,13 +42,19 @@ You may build your app using functionality provided directly by SvelteKit withou
 
 ### Using Netlify Redirect Rules
 
+> [Netlify docs here](https://docs.netlify.com/routing/redirects/)
+
+Netlify supports declarative redirects via a `_redirects` file placed in your `publish` directory (usually "`build`", as configured in your `netlify.toml` above. NOT in your project root next to `netlify.toml`).
```
benmccann, line 47: the `build` directory is generated by SvelteKit, so I wouldn't think you could put it there. I thought on the last issue it was said to go in `static`?

roschaefer: Yes! It is `static/` unless you configured `kit.files.assets` to something else in the [project configuration](https://kit.svelte.dev/docs#configuration-files).

sw-yx: Yeah my bad... 

```suggestion
Netlify supports declarative redirects via a `_redirects` file. Place it in the `static/` folder, NOT in your project root next to `netlify.toml`.
```

benmccann:
This feels quite wrong to me. `_redirects` isn't a static asset. There are three different directories that get copied into the `publish` directory. We'd just be relying on the fact that static assets end up in the `publish` directory, but it's quite an unintuitive place to have to put the file.

I'd rather put it at the root of the project. In neither case would we be putting it directly in the `publish` directory, but rather relying on the adapter to copy it to the right place. It's just a question of where would you most expect it to live. Living next to `netlify.toml` makes more sense than living next to your image files, etc. in my opinion

sw-yx:
i see. so perhaps @roschaefer should reopen #1586 ?

kvn-shn:
At least to me it feels quite intuitive putting it into `static/` (e.g. next to robots.txt) because it's more like an oldschool `.htaccess` file that belongs in the _"public_html" of the server_... After all it _is_, in the context of Svelte, a static asset that Svelte can't and won't do anything with.

So in my opinion this PR would do the trick.

benmccann:
I went ahead and merged https://github.com/sveltejs/kit/pull/1586. It shouldn't stop you from putting it in `static/` if you'd like, but I think supporting it in the root directory next to `netlify.toml` makes a lot of sense

packages/adapter-netlify/README.md
```diff
@@ -42,13 +42,19 @@ You may build your app using functionality provided directly by SvelteKit withou
 
 ### Using Netlify Redirect Rules
 
+> [Netlify docs here](https://docs.netlify.com/routing/redirects/)
+
+Netlify supports declarative redirects via a `_redirects` file. Place it in the `static/` folder, NOT in your project root next to `netlify.toml`.
+
 During compilation a required "catch all" redirect rule is automatically appended to your `_redirects` file. (If it doesn't exist yet, it will be created.) That means:
 
 - `[[redirects]]` in `netlify.toml` will never match as `_redirects` has a [higher priority](https://docs.netlify.com/routing/redirects/#rule-processing-order). So always put your rules in the [`_redirects` file](https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file).
```
sw-yx, line 51: just fyi - i think this is inaccurate - the doc actually says "Rules in the _redirects file are always processed first, followed by rules in the Netlify configuration file."

in other words you CAN put stuff in netlify.toml, AND in _redirects, its just that the latter overrides the former in cases of conflict.

benmccann: I believe it's true when you consider the catch-all route that gets added

PR #1584, opened 2021-05-29
Fix Redirect issue when using baseURL

sidharthv96:
Closes  #1583
Fixes #1476

### Before submitting the PR, please make sure you do the following
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts

PR #1608, opened 2021-05-31
Fix redirect URLs when removing/adding trailing slash with paths.base set

### LINKED ISSUE #1476: navigate always discard base path
I am building an adapter for ipfs (see https://github.com/wighawag/sveltejs-adapter-ipfs)
and currently I have to inject and modify svelte generated code.

My plan is to continue documenting the process so that svelte kit can accomodate it without requiring such code injection, which is obviously very brittle

On recent version of svelte kit I noticed this line :
https://github.com/sveltejs/kit/blob/9a2cc0afd3433134045dec0550172ef194fc23d4/packages/kit/src/runtime/client/router.js#L232

It use the parsed path (which remove the base path) for the url to navigate. This is not intended as the bas path will always be ignored. It should use `url.pathname` or simply url

sidharthv96:
Fixes #1476

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/load_config/index.js
+++ b/packages/kit/src/core/load_config/index.js
@@ -135,8 +135,13 @@ export function validate_config(config) {
 	// resolve paths
 	const { paths } = validated.kit;
 
-	if (paths.base !== '' && !paths.base.startsWith('/')) {
-		throw new Error('config.kit.paths.base must be a root-relative path');
+	if (paths.base !== '') {
+		if (!paths.base.startsWith('/')) {
+			throw new Error('config.kit.paths.base must be a root-relative path');
+		}
+		if (paths.base.endsWith('/')) {
+			throw new Error('config.kit.paths.base must not end with /');
+		}
 	}
 
 	paths.assets = resolve(paths.base, paths.assets);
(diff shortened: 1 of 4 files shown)


benmccann:
This is a subset of the changes from https://github.com/sveltejs/kit/pull/1584

packages/kit/src/core/load_config/index.js
```diff
@@ -135,8 +135,13 @@ export function validate_config(config) {
...
 	const { paths } = validated.kit;
 
-	if (paths.base !== '' && !paths.base.startsWith('/')) {
-		throw new Error('config.kit.paths.base must be a root-relative path');
+	if (paths.base !== '') {
+		if (!paths.base.startsWith('/')) {
+			throw new Error('config.kit.paths.base must be a root-relative path');
+		}
+		if (paths.base.endsWith('/')) {
+			throw new Error('config.kit.paths.base must not end with /');
```
benmccann, line 143: I wonder if rather than throwing an error we should just drop the trailing `/`

sidharthv96: Isn't it better that the user is conscious of this as it's a one time config?

PR #1609, opened 2021-05-31
docs: community adapters section in docs

jthegedus:
- Add section for community adapters to the docs
- Include [Firebase adapter](https://github.com/jthegedus/svelte-adapter-firebase) in community seciton
- Move all official adapters to a single section, the distinction between serverless platforms and otherwise isn't useful to know here, especially when linking out to official products

### Tests

### Changesets

### DIFF
--- a/documentation/docs/10-adapters.md
+++ b/documentation/docs/10-adapters.md
@@ -31,16 +31,16 @@ export default {
 };
 ```
 
-A variety of official adapters exist for serverless platforms...
+A variety of official adapters exist:
 
-- [`adapter-begin`](https://github.com/sveltejs/kit/tree/master/packages/adapter-begin) — for [Begin](https://begin.com)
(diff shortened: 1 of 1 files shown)


dummdidumm:
While I think discoverability is good for community adapters, I'm not sure this is the right place to put it. Where do we start, where do we end with linking to community adapters?

ignatiusmb:
Was just about to comment the same, perhaps Svelte Society can host the community adapters and we just link it here?

jthegedus:
Linking to a 3rd party site, which then requires searching within the site to find an item (could be improved with query params for search/filtering), as well as being displayed  alongside other pre-kit libs with "adapter" in their names isn't very discoverable. You need to already want to be searching for the item to discover it, in which case Google or other search engines are likely to be used. I see it as unlikely to have any impact on discorverability, other than maybe the SEO of the linked GitHub repo.. maybe.

The static list in official docs which is linked to every user who inits an app would improve discoverability for those who didn't know it was an option.

PR #1613, opened 2021-05-31
Clarify the use of svelte-kit preview and adapters

stephane-vanraes:
This PR simple adds an extra comment to `svelte-kit preview` noting that serving the final app is dependent on the adapter used and refers the developers to the documentation of the adapter to learn how to actually serve their app in production.

Reason for this change is that is a recurring question, the documentation tells that you should **not** use the preview command but does little in explaining where to get the information on how to do it.

### DIFF
--- a/documentation/docs/10-adapters.md
+++ b/documentation/docs/10-adapters.md
@@ -2,7 +2,7 @@
 title: Adapters
 ---
 
-Before you can deploy your SvelteKit app, you need to _adapt_ it for your deployment target. Adapters are small plugins that take the built app as input and generate output that is optimised for a specific platform.
+Before you can deploy your SvelteKit app, you need to _adapt_ it for your deployment target. Adapters are small plugins that take the built app as input and generate output that is optimised for a specific platform. How the resulting app is served depends on the adapter and further instructions can be found in the documentation for each adapter.
 
 For example, if you want to run your app as a simple Node server, you would use the `@sveltejs/adapter-node` package:
 
--- a/documentation/docs/10-adapters.md
+++ b/documentation/docs/10-adapters.md
@@ -2,7 +2,7 @@
 title: Adapters
 ---
 
-Before you can deploy your SvelteKit app, you need to _adapt_ it for your deployment target. Adapters are small plugins that take the built app as input and generate output that is optimised for a specific platform. How the resulting app is served depends on the adapter and further instructions can be found in the documentation for each adapter.
+Before you can deploy your SvelteKit app, you need to _adapt_ it for your deployment target. Adapters are small plugins that take the built app as input and generate output for deployment. Many adapters are optimised for a specific hosting provider, and you can generally find information about deployment in your adapter's documentation. However, some adapters, like `adapter-static`, build output that can be hosted on numerous hosting providers, so it may also be helpful to reference the documentation of your hosting provider in these cases.
 
 For example, if you want to run your app as a simple Node server, you would use the `@sveltejs/adapter-node` package:
 
--- a/documentation/docs/13-cli.md
+++ b/documentation/docs/13-cli.md
@@ -27,7 +27,7 @@ Builds a production version of your app, and runs your adapter if you have one s
 
 ### svelte-kit preview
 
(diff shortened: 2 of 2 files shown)


documentation/docs/13-cli.md
```diff
@@ -27,7 +27,7 @@ Builds a production version of your app, and runs your adapter if you have one s
 
 ### svelte-kit preview
 
-After you've built your app with `svelte-kit build`, you can start the production version (irrespective of any adapter that has been applied) locally with `svelte-kit preview`. This is intended for testing the production build locally, **not for serving your app**, for which you should always use an adapter.
+After you've built your app with `svelte-kit build`, you can start the production version (irrespective of any adapter that has been applied) locally with `svelte-kit preview`. This is intended for testing the production build locally, **not for serving your app**, for which you should always use an adapter. The way your app will be served is dependent on the adapter used, see the documentation of your chosen adapter for specific instructions on how to run your app.
```
benmccann, line 30: the new sentence seems out of place to me because running your app in production with the adapter is unrelated to `svelte-kit preview`. it seems like it might belong in the `svelte-kit build` section or something instead

stephane-vanraes: I placed it here because the preceding sentence mentions it is **not for serving your app** leaving several users wonder what the correct command is for serving the app.
Ideally a comment like this (I can have another look later) would be somehow integrated in the sections about `adapters` themselves.

I feel the documentation is lacking here as this is a question that comes back every other day on the Discord in some form or another, but clearly some more discussion can be had here.

I would propose to move the wording indeed to `svelte-kit build` with something along the lines of

> After building you can run your app, see the chosen adapter for specific instructions on how to do this.

In preview remove the new lines, but add a link to the adapter section.
And in the adapter section another note that you should check the documentation of the adapter for serving instructions.

jthegedus: I think the following line in the `preview` docs still needs to be more clear:

```diff
-This is intended for testing the production build locally, **not for serving your app**, , for which you should always use an adapter
+This is intended for testing the production build locally, **not for serving your app in the adapter target environment locally** (for locally testing a prod build with the adapter output see the adapter docs).
```

Or something of the sort. The distinctions I see that need to be made are:

1. dev: `svelte-kit dev`
2. build prod (used for all subsequent steps): `svelte-kit build`
3. test prod build locally without adapter environment: `svelte-kit preview`
4. test prod build locally with adapter environment: see adapter docs
5. deploy prod build to adapter environment: see adapter docs

with steps 3 and 4 being the confusing pieces. 3 relates to `preview` so should be there. 4 relates to `build` so should be there.

The docs don't make these 5 steps clear in a concise manner. You have to read a lot of documentation to gather the information on these 5 steps.

documentation/docs/13-cli.md
```diff
@@ -25,9 +25,11 @@ Builds a production version of your app, and runs your adapter if you have one s
 
 - `--verbose` — log more detail
 
+After building the app you can see the documentation of your chosen adapter for specific instructions on how to serve your app.
```
benmccann, line 28:
```suggestion
After building the app, you can see the documentation of your chosen [adapter](#adapters) for specific instructions on how to serve your app.
```

documentation/docs/10-adapters.md
```diff
@@ -2,7 +2,7 @@
 title: Adapters
 ---
 
-Before you can deploy your SvelteKit app, you need to _adapt_ it for your deployment target. Adapters are small plugins that take the built app as input and generate output that is optimised for a specific platform.
+Before you can deploy your SvelteKit app, you need to _adapt_ it for your deployment target. Adapters are small plugins that take the built app as input and generate output that is optimised for a specific platform. How the resulting app is served depends on the adapter and further instructions can be found in the documentation for each adapter.
```
benmccann, line 5:
```suggestion
Before you can deploy your SvelteKit app, you need to _adapt_ it for your deployment target. Adapters are small plugins that take the built app as input and generate output for deployment. Many adapters are optimised for a specific hosting provider, and you can generally find information about deployment in your adapter's documentation. However, some adapters, like `adapter-static`, build output that can be hosted on numerous hosting providers, so it may also be helpful to reference the documentation of your hosting provider in these cases.
```

PR #1617, opened 2021-05-31
types: allow void return in `LoadOutput`

### LINKED ISSUE #1606: `LoadOutput` types incorrect
**Describe the bug**
From what I understand `load` should be able to return `undefined` in order to trigger a fallthrough to a 404 page. Currently that results in a type error because `LoadOutput` expects an object, which in my testing _doesn't_ trigger a fallthrough, even if empty.

**Logs**
N/A

**To Reproduce**
Return nothing from a typed `load` function, inspect the type error given.

**Expected behavior**
`load` should be able to return `undefined` or `null`.

ignatiusmb:
Fixes #1606, and cleans up some more types

### DIFF
--- a/packages/kit/src/core/node/index.js
+++ b/packages/kit/src/core/node/index.js
@@ -1,6 +1,6 @@
 /**
  * @param {import('http').IncomingMessage} req
- * @returns {Promise<import('types/hooks').StrictBody | null>}
+ * @returns {Promise<import('types/hooks').StrictBody>}
  */
 export function getRawBody(req) {
 	return new Promise((fulfil, reject) => {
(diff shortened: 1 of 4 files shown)


dummdidumm:
Thanks!

janosh:
Shouldn't the `void |` in

```ts
MaybePromise<void | LoadOutput<
	InferValue<Output, 'props', Record<string, any>>,
	InferValue<Output, 'context', Record<string, any>>
>>;
```

be part of `LoadOutput`? Otherwise a `Load` function like

```svelte
async function load(input: LoadInput): Promise<LoadOutput> {
  // ...
}
```

will still raise a type error when not returning.

PR #1622, opened 2021-06-01
update vite

dominikg:

### Tests

### Changesets

This PR is going to fail at tests. I'm expecting that and the fail will be used to file a regression at vite

### DIFF
--- a/packages/kit/package.json
+++ b/packages/kit/package.json
@@ -6,7 +6,7 @@
 		"@sveltejs/vite-plugin-svelte": "^1.0.0-next.10",
 		"cheap-watch": "^1.0.3",
 		"sade": "^1.7.4",
-		"vite": "2.3.3"
+		"vite": "^2.3.5"
 	},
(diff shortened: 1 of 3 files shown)


benmccann:
Looks like the tests are failing. That's why I'd pinned it at 2.3.3. I never investigated what the issue was

packages/kit/package.json
```diff
@@ -6,7 +6,7 @@
 		"@sveltejs/vite-plugin-svelte": "^1.0.0-next.10",
 		"cheap-watch": "^1.0.3",
 		"sade": "^1.7.4",
-		"vite": "2.3.3"
+		"vite": "^2.3.5"
```
dummdidumm, line 9: I think we want to pin versions for now

```suggestion
		"vite": "2.3.5"
```

dominikg:
damnit, shouldn't have merged that from github... going to start over in a new one

PR #1633, opened 2021-06-03
(feat) svelte package: transpile TS

dummdidumm:
This adds transpilation functionality for TypeScript files when using `svelte-kit package`.

It requires the user to have TypeScript installed, but that should be the case if he uses TS.

### DIFF
--- a/packages/kit/src/core/make_package/index.js
+++ b/packages/kit/src/core/make_package/index.js
@@ -63,9 +63,8 @@ export async function make_package(config, cwd = process.cwd()) {
 				? (await preprocess(source, config.preprocess, { filename })).code
 				: source;
 		} else if (ext === '.ts' && !file.endsWith('.d.ts')) {
-			// TODO transpile TS file and emit types
-			// also, we want to emit types from JSDoc annotations in .js files
-			throw new Error('svelte-kit package does not yet support TypeScript');
+			out_file = file.slice(0, -'.ts'.length) + '.js';
+			out_contents = await transpileTs(file, source);
 		} else {
 			out_file = file;
 			out_contents = source;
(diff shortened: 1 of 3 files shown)


packages/kit/src/core/make_package/index.js
```diff
@@ -63,9 +63,8 @@ export async function make_package(config, cwd = process.cwd()) {
 				? (await preprocess(source, config.preprocess, { filename })).code
 				: source;
 		} else if (ext === '.ts' && !file.endsWith('.d.ts')) {
-			// TODO transpile TS file and emit types
-			// also, we want to emit types from JSDoc annotations in .js files
-			throw new Error('svelte-kit package does not yet support TypeScript');
+			out_file = file.slice(0, -'.ts'.length) + '.js';
```
ignatiusmb, line 66: Can't we just do this?
```suggestion
			out_file = `${file.slice(0, -3)}.js`;
```

ignatiusmb: or
```suggestion
			out_file = file.replace(/ts$/, 'js');
```

packages/kit/src/core/make_package/index.js
```diff
@@ -98,6 +97,63 @@ export async function make_package(config, cwd = process.cwd()) {
...
+
+/**
+ *
+ * @param {string} filename
+ * @param {import('typescript')} ts
+ */
+function loadTsconfig(filename, ts) {
+	let basePath = process.cwd();
+	const fileDirectory = path.dirname(filename);
+	let tsconfigFile = ts.findConfigFile(fileDirectory, ts.sys.fileExists);
```
ignatiusmb, line 130: Should we also consider a possibility of a separate `tsconfig` in `src/lib` exclusive for `lib` stuff?

ignatiusmb (commented):
You beat me to it! Super excited for everything `svelte-kit`

packages/kit/src/core/make_package/index.js
dummdidumm: We could, but for me ` -'.ts'.length` is the most readable (I know what is substracted); and it's consistent with the approach taken a few lines above.

packages/kit/src/core/make_package/index.js
dummdidumm: This should work already, as the nearest tsconfig for each file is looked up (which could be optimized, but I don't think it's a bottleneck right now).

Issue #1641, opened 2021-06-05
Location State store


dkzlv:
**Is your feature request related to a problem? Please describe.**

Currently Kit's router does not give access to location state and always pushes an empty object. It's only used internally to restore scroll position.
It seems that the router should not abstract away this feature, as it can be useful for many different cases (animation restoration, saving some screen state, etc.).

**Describe the solution you'd like**

I would say, the API surface consists of 3 parts:

1. pass state object in `goto` in the second argument
2. allow for passing state in the `<a>` tag in some property, like `sveltekit:state={}`
3. adding a new readable store in `$app/stores` that updates the value as you go back and forth the history and excludes internal properties along the way

**Describe alternatives you've considered**

n/a

**How important is this feature to you?**
I wouldn't say this feature is a deal-breaker, but I also think Kit shouldn't abstract away basic browser features.

**Additional context**

My only concern is that it doesn't play nice with SSR and with TypeScript. Other than that it can be the only elegant solution sometimes.

benmccann:
There was a PR for this in Sapper: https://github.com/sveltejs/sapper/pull/1341. Unfortunately, we didn't get a chance to get it merged before switching development over to SvelteKit

gustavopch:
I've explained how I use history state to control modals in https://github.com/sveltejs/kit/issues/3236#issuecomment-1011436921. I'm linking here because it may be useful for this discussion too.

Rich-Harris:
Closing this in favour of #5478

PR #1643, opened 2021-06-06
Add initial location state to router

lukasIO:
Related to https://github.com/sveltejs/kit/issues/1641, but only adds the functionality to the `goto` function. 

An additional optional `state` parameter for the `goto` function that sets the initial state when the router calls `pushState`.

### Tests

### Changesets


dummdidumm (approved):

dkzlv:
It can be strange for users to get internal framework data in location state. Currently you can get `sveltekit:scroll` key. I give 99.9% it won't be overwritten, but it can break something if you, say, send `location.state` to some API endpoint.

Also it would be better to have a store of a kind. You may want to access location state in a layout component, it would be awkward to do stuff like
```js
let state;
$: $page.path, state = location.state;
```

benmccann:
@dkzlv I agree it's somewhat awkward to have internal framework data in the location state. Your comment will probably get lost just being on a merged PR though. It'd be better to file a new issue for this possibly with a bit more details about your proposal

Issue #1022, opened 2021-04-14
sveltekit:keepfocus


lukaszpolowczyk:
**Is your feature request related to a problem? Please describe.**

1. Refresh page
2. Press `Tab key` on the keyboard
3. The button are focused
4. Press `Right Arrow` on the keyboard
5. The button loses focus

GIF with this:
![lost-focus](https://user-images.githubusercontent.com/16800535/114312453-0534aa00-9af3-11eb-8127-fa5c146092fb.gif)

**Describe the solution you'd like**
Maybe similar to `sveltekit:noscroll` and `goto(href, { replaceState, noscroll })` , e.g. `sveltekit:keepfocus` and `goto(href, { replaceState, keepfocus })`?

It looks like something easy to do. [L245 of `router.js`](https://github.com/sveltejs/kit/blob/39b69677635a59b40675d375c64bbc4507ea0ea5/packages/kit/src/runtime/client/router.js#L245)  would have to give in a similar condition as  [L254 of `router.js`](https://github.com/sveltejs/kit/blob/39b69677635a59b40675d375c64bbc4507ea0ea5/packages/kit/src/runtime/client/router.js#L254).

**Describe alternatives you've considered**
Described by @ahtcx in https://github.com/sveltejs/kit/issues/968#issuecomment-817378602
> as a workaround I've been using [`patch-package`](https://www.npmjs.com/package/patch-package) on [L245 of `router.js`](https://github.com/sveltejs/kit/blob/39b69677635a59b40675d375c64bbc4507ea0ea5/packages/kit/src/runtime/client/router.js#L245) but this isn't ideal and I imagine `router.js` will change a lot so you'll have to update it at every change.

**How important is this feature to you?**
Using the keyboard and keeping the application focused is a matter of accessibility in general.

c-bandy:
What's the normal browser behavior when navigating between pages without a router? Does it preserve the tab state?

Conduitry:
If it's a regular old full-page load, then no. Preserving the tab state doesn't make sense, because it's generally a completely unrelated document.

c-bandy:
OK. Given that, I'm not too sure how I feel about adding a helper like this into Kit. My gut feeling says this type of thing would better be solved using client JavaScript.

lukaszpolowczyk:
@c-bandy Rather it's worth comparing to pure svelte behavior, with no routers.
Therefore, `sveltekit:keepfocus` is optional, as is `sveltekit:noscroll`.

[...]

PR #1647, opened 2021-06-07
Implement keepfocus for goto.

mohe2015:
This allows you to navigate without loosing focus of the current element.
This may be useful if you want to e.g store the current filter or sorting order of a list in the url to allow sharing it.

Closes https://github.com/sveltejs/kit/issues/1022

### Tests
- [ ] Run the tests with `pnpm test` (**doesn't run on two different systems**) and lint the project with `pnpm lint` (success) and `pnpm check` (success)

### Changesets


dummdidumm:
This would only be part of the solution, right? What if you click on a `<a>` tag but want to keep focus?

mohe2015:
> This would only be part of the solution, right? What if you click on a `<a>` tag but want to keep focus?

That's correct but for my use case (goto with replaceState, only changing query parameters) this is enough (I'm combining this with a custom store for the query parameters and the goto is in that store implementation). I wanted to start with the minimal changes so the chances of getting them in is higher.

@lukaszpolowczyk
@ahtcx @MatthiasGrandl (they seem to have the same usecase like me)
Which additional changes would you need and what do you think about the current ones?

mohe2015:
@dummdidumm Do you think this has a chance of getting merged if the tests pass?

Rich-Harris:
Thanks, I tweaked the docs and the types now that they're growing beyond what makes sense to have in a single block of text

Issue #1428, opened 2021-05-12
kit and svelte are listed a devDependency in demos but needed for running build


iOiurson:
Once you built a project with `adapter-node`, `node build` does not work without `@sveltejs/kit` and `svelte` in `node_modules`.

```bash
Cannot find package '@sveltejs/kit' imported from /Users/Romain/Code/tutos/sveltekit-app/build/app.js
```

On the kit template projects (demo or skeleton), `@sveltejs/kit` and `svelte` are listed as `devDependencies`.
My understanding was that you should be able to run a project in production without any `devDependencies` installed.

Plus i thought that all prod `dependencies` were packaged by `rollup` (via `vite`) on build, which does not seem to be the case since we still have import such as
 
```js
// in build/app.js for DEMO APP
import cookie from "cookie";
```

I'm not sure why we should not keep `@sveltejs/kit` and `svelte` as `devDependencies`, but at the same time we currently need to have the corresponding folders in `node_modules`...

Am i missing something here ?

Conduitry:
Closing this as a duplicate of #1257. This should also be fixed by #1648.

PR #1648, opened 2021-06-07
Re-bundle server app in adapter-node

### LINKED ISSUE #1257: Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@sveltejs/kit' on production build
**Describe the bug**
When I create a new SvelteKit project with the `node` adapter and try to use the files in the `build` directory, I get the following error:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@sveltejs/kit' imported from [...]
```

**Logs**
```
internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@sveltejs/kit' imported from [REDACTED]/app.js
    at packageResolve (internal/modules/esm/resolve.js:655:9)
    at moduleResolve (internal/modules/esm/resolve.js:696:18)
    at Loader.defaultResolve [as _resolve] (internal/modules/esm/resolve.js:810:11)
    at Loader.resolve (internal/modules/esm/loader.js:86:40)
    at Loader.getModuleJob (internal/modules/esm/loader.js:230:28)
    at ModuleWrap.<anonymous> (internal/modules/esm/module_job.js:56:40)
    at link (internal/modules/esm/module_job.js:55:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```

**To Reproduce**
- create a new SvelteKit skeleton project
- add `adapter-node` to it
- make sure all the deps have `"next"` as version
- `yarn && yarn build`
- isolate the `build` directory (I only want the production files in my production environment) by moving it somewhere else
- add a `package.json` file to the directory with the contents: `{ "type": "module" }` (for ESM)
- `node .` <- This is what triggers the bug

**Expected behavior**

I expect the `build` directory to contain *everything* that is needed to run the app in isolation. Adding the small `package.json` file is another thing that I do not want to do myself.

When running `node .` the server should just run.

<details>

```
  System:
    OS: macOS 11.2.3
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
    Memory: 5.37 GB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.0 - ~/.volta/tools/image/node/14.16.0/bin/node
    Yarn: 1.22.10 - ~/.volta/tools/image/yarn/1.22.10/bin/yarn
    npm: 7.7.0 - ~/.volta/tools/image/npm/7.7.0/bin/npm
  Browsers:
    Chrome: 90.0.4430.85
    Firefox: 87.0
    Safari: 14.0.3
  npmPackages:
    @sveltejs/kit: ^1.0.0-next.74 => 1.0.0-next.91 
    svelte: ^3.29.0 => 3.37.0 
    vite: ^2.2.2 => 2.2.3 
```

- Your adapter: Node
</details>

**Severity**
It's kind of annoying as I really want my production setup to be ready, but it doesn't stop me from development.

**Additional context**
All the above steps are done in a Dockerfile for production but I was able to reproduce it outside as well

<details>
  <summary>Dockerfile</summary>

```
FROM node:14-alpine as build

ARG VITE_API_URL

RUN mkdir /svelte-build
WORKDIR /svelte-build

ENV VITE_API_URL=${VITE_API_URL}

COPY . .
RUN yarn && yarn build

FROM node:14-alpine as production

RUN addgroup -S app && adduser -S app -G app

RUN mkdir /app && chown app:app /app

USER app

WORKDIR /app

RUN echo '{ "type": "module" }' > /app/package.json
COPY --chown=app:app --from=build /svelte-build/build/ /app

EXPOSE 3000

CMD node index.js
```
</details>

Conduitry:
Fixes #1257. This makes adapter-node run the server-side app through esbuild as part of the build process, treating anything in (prod) `dependencies` as external, and bundling everything else. This means that, for example, `@sveltejs/kit` will no longer need to be installed when deploying the production app - and any other dependencies that the app adds as `devDependencies` will also be bundled in. This means that, in production, you should be able to copy over your `package.json` and `build/`, run `npm install --production`, and run `node build` - which is what you'd expect to be able to do.

These changes seem to behave as expected on the project I have at work (which adds a number of dependencies, some of which should be bundled on the server and some of which should not) - but I'd still appreciate some more eyes on this PR.

### Tests

### Changesets

### DIFF
--- a/.changeset/tiny-socks-sparkle.md
+++ b/.changeset/tiny-socks-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-node': patch
+---
+
+Bundle server-side app during adapt phase
--- a/packages/adapter-node/index.js
+++ b/packages/adapter-node/index.js
@@ -1,6 +1,7 @@
-import { copyFileSync } from 'fs';
+import { readFileSync } from 'fs';
 import { join } from 'path';
 import { fileURLToPath } from 'url';
+import esbuild from 'esbuild';
 
 /**
  * @param {{
@@ -18,11 +19,17 @@ export default function ({ out = 'build' } = {}) {
 			utils.copy_client_files(static_directory);
 			utils.copy_static_files(static_directory);
 
-			utils.log.minor('Copying server');
-			utils.copy_server_files(out);
-
+			utils.log.minor('Building server');
 			const files = fileURLToPath(new URL('./files', import.meta.url));
-			copyFileSync(`${files}/server.js`, `${out}/index.js`);
+			utils.copy(files, '.svelte-kit/node');
+			await esbuild.build({
+				entryPoints: ['.svelte-kit/node/index.js'],
+				outfile: join(out, 'index.js'),
+				bundle: true,
+				external: Object.keys(JSON.parse(readFileSync('package.json', 'utf8')).dependencies || {}),
+				format: 'esm',
+				platform: 'node'
+			});
 
 			utils.log.minor('Prerendering static pages');
 			await utils.prerender({
--- a/packages/adapter-node/package.json
+++ b/packages/adapter-node/package.json
@@ -20,6 +20,9 @@
 		"check-format": "prettier --check . --config ../../.prettierrc --ignore-path .gitignore",
 		"prepublishOnly": "npm run build"
 	},
+	"dependencies": {
+		"esbuild": "^0.12.5"
+	},
 	"devDependencies": {
 		"@rollup/plugin-json": "^4.1.0",
 		"@sveltejs/kit": "workspace:*",
--- a/packages/adapter-node/rollup.config.js
+++ b/packages/adapter-node/rollup.config.js
@@ -5,10 +5,10 @@ import json from '@rollup/plugin-json';
 export default {
 	input: 'src/index.js',
 	output: {
-		file: 'files/server.js',
+		file: 'files/index.js',
 		format: 'esm',
 		sourcemap: true
 	},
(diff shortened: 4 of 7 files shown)


Conduitry:
I'm not sure whether this is something that we'd expect esbuild to be able to do, but currently with this PR if you `import 'foo'` where `foo` is a CommonJS external dependency which has `require('util')` (or some other Node builtin), then this essentially remains as a `require('util')` in the output bundle, and is not transformed into an `import 'util'`.

The CJS helper that esbuild is using and injecting is:

```js
var __require = (x) => {
  if (typeof require !== "undefined")
    return require(x);
  throw new Error('Dynamic require of "' + x + '" is not supported');
};
```

(and then later on the bundle has `__require('util')`).

I'm not sure what can be done to make this work other than to have a `global.require = createRequire(import.meta.url)` at the top of the bundle.

Conduitry:
I've just push a `require()` shim of the sort I mentioned above. It's an ugly solution, but I don't know what else we can do without needing changes in esbuild itself. I think this issue is serious enough that we shouldn't wait for upstream changes to more nicely support this, although it would still be a good idea to look through esbuild's open issues and see whether this is on their roadmap.

iOiurson:
Will this also fix #1428 ?

Conduitry:
It should. That looks like a duplicate of #1257.

Rich-Harris (approved):

PR #1649, opened 2021-06-07
fix: match up ambient modules for goto

ignatiusmb:
The recent addition from #1643 missed to update ambient types, and with more proposed additions coming up in the PRs, it can be quite a handful to manage this manually. In the meantime, this PR should make it easier to make changes by only needing to update JSDoc and markdown docs.


dummdidumm:
Thank you for catching that!

PR #1148, opened 2021-04-20
force Vite to bundle component libraries in SSR

Rich-Harris:
fixes #904, by scanning all the `package.json` files of dependencies/devDependencies and adding any that have a `"svelte"` field to `ssr.noExternal`. The user's `noExternal` is still respected, so it's not a problem if this heuristic fails — it should just make these sorts of errors rarer.

This will fail in cases where a new component library is added a dependency while the dev server is running; in those cases turning it off and on again will fix it.

Finicky to add a test for this, but I verified it works with a site that uses external component libraries.

### Before submitting the PR, please make sure you do the following
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts

PR #1665, opened 2021-06-10
Add faq docs about limitation in vite-plugin-svelte

bluwy:
Add paragraph under FAQ package section about `@sveltejs/vite-plugin-svelte`'s docs regarding the `optimizeDeps.exclude` [trick](https://github.com/sveltejs/vite-plugin-svelte/tree/main/packages/vite-plugin-svelte#importing-third-party-svelte-libraries).

### Tests

### Changesets

### DIFF
--- a/documentation/faq/70-packages.md
+++ b/documentation/faq/70-packages.md
@@ -12,4 +12,6 @@ Old beta versions of the SvelteKit template included the configuration value `no
 
 The second most commonly-encountered issue is having a Svelte component that imports a CommonJS library. In this case, you should try to work with the library authors to distribute an ESM version of the dependency. However, in the meantime, you can workaround this issue by adding the dependency to `vite.optimizeDeps.include` in `svelte.config.js`.
 
+Also, some Svelte libraries doesn't work nicely with Vite's pre-bundling process, do check out `@sveltejs/vite-plugin-svelte`'s docs on its current [limitation and workaround](https://github.com/sveltejs/vite-plugin-svelte/tree/main/packages/vite-plugin-svelte#importing-third-party-svelte-libraries).
+
 If you are still encountering issues we recommend searching both [the Vite issue tracker](https://github.com/vitejs/vite/issues) and the issue tracker of the library in question. Sometimes issues can be worked around by fiddling with the [`optimizeDeps`](https://vitejs.dev/config/#dep-optimization-options) or [`ssr`](https://vitejs.dev/config/#ssr-options) config values.


Conduitry:
While we're here touching this, I think the paragraph about Vite could use some reworking. When that was written, Vite was a direct dependency of the user's app. It's now a prod dep of SvelteKit instead. (And, currently, is actually pinned to a specific version, so there's no straightforward way for users to upgrade it anyway.) I don't have anything specific in mind about what the Vite paragraph should say instead.

bluwy:
I've updated the Vite part. Didn't include the explanation of Vite being a prop dependency though, not sure if including that would help with troubleshooting the issue.

documentation/faq/70-packages.md
```diff
@@ -12,4 +12,6 @@ Old beta versions of the SvelteKit template included the configuration value `no
...
 
-Finally, Vite has had some issues that have been fixed, so we recommend upgrading to the latest version of Vite. If you are still encountering issues we recommend searching both [the Vite issue tracker](https://github.com/vitejs/vite/issues) and the issue tracker of the library in question. Sometimes issues can be worked around by fiddling with the [`optimizeDeps`](https://vitejs.dev/config/#dep-optimization-options) or [`ssr`](https://vitejs.dev/config/#ssr-options) config values.
+Some issues may also be specific to Vite, so we recommend searching both [the Vite issue tracker](https://github.com/vitejs/vite/issues) and the issue tracker of the library in question. Sometimes issues can be worked around by fiddling with the [`optimizeDeps`](https://vitejs.dev/config/#dep-optimization-options) or [`ssr`](https://vitejs.dev/config/#ssr-options) config values. 
+
+Finally, do check out `@sveltejs/vite-plugin-svelte`'s docs on the [limitation of importing some Svelte libraries](https://github.com/sveltejs/vite-plugin-svelte/tree/main/packages/vite-plugin-svelte#importing-third-party-svelte-libraries).
```
benmccann, line 17: We avoid bundling Svelte components automatically https://github.com/sveltejs/kit/pull/1148. I wonder if we should also automatically avoid optimizing them as well?

bluwy: There's a [discussion in discord](https://discord.com/channels/457912077277855764/819723698415599626/850050004538228766) from @dominikg regarding auto excluding optimizations of Svelte libraries. Looks like the exact heuristics are already implemented in that PR ([this code specifically](https://github.com/sveltejs/kit/blob/17bf30143c14d3b45f7d116487bfd74f9e271236/packages/kit/src/core/utils.js#L106-L108)), but only for `ssr.noExternals`.

I think the specific logic could live in vite-plugin-svelte instead, though I'm not sure if @dominikg has plans for that, or to fix the underlying prebundling issue directly.

bluwy: FYI I've dug into the prebundling issue last week and wrote my findings [here](https://github.com/bluwy/vite-svelte-dedupe#readme). It should explain the reason why `optimizeDeps.exclude` work as it is for Svelte.

@dominikg had raised some concerns in discord regarding the auto-exclude feature being directly in vite-plugin-svelte. I'll let him explain further, but perhaps we could handle it in SvelteKit for now.

bluwy:
Is it possible to merge this in soon? Hoping to get this tip out in the public. Let me know if there's any other changes needed!

benmccann:
@bluwy it looks like this PR will need to be rebased

documentation/faq/70-packages.md
benmccann: Wow, that's some great investigation. I have to be honest I'm not totally following everything you've written as my understanding of the Vite internals is minimal. @dominikg, @GrygrFlzr, and @Rich-Harris probably know more about it than I do. It would be nice to file an issue that we can link to here or on the `vite-plugin-svelte` readme that explain the reason for this restriction and so that people can follow along in case it is resolved at some point

bluwy:
Rebased. I also moved it to on top of the "Vite issue tracker" part as I think the issue tracker should be the last resort if vite-plugin-svelte's workaround doesn't work.

documentation/faq/70-packages.md
bluwy: @benmccann Sorry for the long follow-up :sweat_smile: I've finally opened an issue at https://github.com/vitejs/vite/issues/3910. Will start linking the issue to kit or vite-plugin-svelte if it proves to be a hard problem to tackle in Vite side.

benmccann: I sent a PR for that here: https://github.com/sveltejs/vite-plugin-svelte/pull/69

PR #1666, opened 2021-06-10
Fix navigation when `base` path is set and validate that option's value

### LINKED ISSUE #1359: Unable to navigate to non-root routes by links
**Describe the bug**
When navigating to a page that isn't `/`(e.g `/blog` and `/works`) using links, the page being rendered is the `$error` one. But you can navigate to `/` perfectly well.

Also, if this error shows up and you reload the page, the page renders fine, but then if you navigate to any non-`/` page and come back, its again `$error`

**Logs**
Screenshot 
image

Text:
```txt
<$error> received an unexpected slot "default".
$error.svelte? [sm]:6 <$error> was created with unknown prop 'blogsList'
```

**To Reproduce**
Repro: https://github.com/PuruVJ/sveltekit-error-repro

Run `npm run dev`
Then visit `/` on localhost
Then click on the link.

You should see this

image

Reload: You'll see the page perfectly and the data will be logged to console. Go to Home by clicking on the link, then come back to `/blog` by clicking the link on home. Again the error will be there

**Expected behavior**
When I should navigate by links to any of these routes, the correct component should render and I should see the page, irrespective of which page I'm navigating from and how many times

<details>

```txt
 System:
    OS: Windows 10 10.0.19042
    CPU: (8) x64 Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
    Memory: 1000.34 MB / 7.86 GB
  Binaries:
    Node: 14.16.1 - C:\Program Files\nodejs\node.EXE
    npm: 6.14.12 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    @sveltejs/kit: ^1.0.0-next.100 => 1.0.0-next.100
    svelte: ^3.38.2 => 3.38.2
    vite: ^2.2.4 => 2.2.4

```

- Your browser - Brave

- Your adapter- @sveltejs/adapter-static
</details>

**Severity**
High severity. Unable to run Dev properly, plus it is breaking the prod build too(My assumption)

### LINKED ISSUE #1476: navigate always discard base path

ignatiusmb:
Closes #1359 
Fixes #1476 
Closes #1608 

Updating the docs in this case would be better than manipulating the paths at runtime. Changing the regex or url slicing solves the problem for dev, but still throws when adapters starts building (I tried). Rather than adding checks for all possible uses at runtime or (worse) changing the config passed from the user, it would be clearer for the user to just make sure that the config is correct.

### DIFF
--- a/.changeset/soft-bikes-bake.md
+++ b/.changeset/soft-bikes-bake.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Fix redirect URLs when removing/adding trailing slash when `base` path is set
--- a/.changeset/soft-bikes-bake.md
+++ b/.changeset/soft-bikes-bake.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Fix redirect URLs when removing/adding trailing slash when `base` path is set
+Fix navigation when `base` path is set and validate that option's value
--- a/documentation/docs/14-configuration.md
+++ b/documentation/docs/14-configuration.md
@@ -116,7 +116,7 @@ Whether to [hydrate](#ssr-and-javascript-hydrate) the server-rendered HTML with 
 An object containing zero or more of the following `string` values:
 
 - `assets` — an absolute path, or a path relative to `base`, where your app's files are served from. This is useful if your files are served from a storage bucket of some kind
-- `base` — a root-relative (i.e. starts with `/`) path that specifies where your app is served from. This allows the app to live on a non-root path
+- `base` — a root-relative path that must be prefixed with `/`. This specifies where your app is served from and allows the app to live on a non-root path
 
 ### prerender
 
--- a/documentation/docs/14-configuration.md
+++ b/documentation/docs/14-configuration.md
@@ -116,7 +116,7 @@ Whether to [hydrate](#ssr-and-javascript-hydrate) the server-rendered HTML with 
 An object containing zero or more of the following `string` values:
 
 - `assets` — an absolute path, or a path relative to `base`, where your app's files are served from. This is useful if your files are served from a storage bucket of some kind
-- `base` — a root-relative path that must starts, but doesn't end with `/` (e.g. `/base-path`). This specifies where your app is served from and allows the app to live on a non-root path
+- `base` — a root-relative path that must start, but not end with `/` (e.g. `/base-path`). This specifies where your app is served from and allows the app to live on a non-root path
 
 ### prerender
 
--- a/documentation/docs/14-configuration.md
+++ b/documentation/docs/14-configuration.md
@@ -116,7 +116,7 @@ Whether to [hydrate](#ssr-and-javascript-hydrate) the server-rendered HTML with 
 An object containing zero or more of the following `string` values:
 
 - `assets` — an absolute path, or a path relative to `base`, where your app's files are served from. This is useful if your files are served from a storage bucket of some kind
-- `base` — a root-relative path that must be prefixed with `/`. This specifies where your app is served from and allows the app to live on a non-root path
+- `base` — a root-relative path that must starts, but doesn't end with `/` (e.g. `/base-path`). This specifies where your app is served from and allows the app to live on a non-root path
 
 ### prerender
 
--- a/packages/kit/src/core/load_config/index.js
+++ b/packages/kit/src/core/load_config/index.js
@@ -135,9 +135,9 @@ export function validate_config(config) {
 	// resolve paths
 	const { paths } = validated.kit;
 
-	if (paths.base !== '' && (paths.base === '/' || !paths.base.startsWith('/'))) {
+	if (paths.base !== '' && paths.base.endsWith('/') && !paths.base.startsWith('/')) {
 		throw new Error(
-			'config.kit.paths.base must be a root-relative path. See https://kit.svelte.dev/docs#configuration-paths'
+			"config.kit.paths.base must be a root-relative path and does not ends with '/'. See https://kit.svelte.dev/docs#configuration-paths"
 		);
 	}
 
--- a/packages/kit/src/core/load_config/index.js
+++ b/packages/kit/src/core/load_config/index.js
@@ -137,7 +137,7 @@ export function validate_config(config) {
 
 	if (paths.base !== '' && (paths.base.endsWith('/') || !paths.base.startsWith('/'))) {
 		throw new Error(
-			"config.kit.paths.base must be a root-relative path and does not ends with '/'. See https://kit.svelte.dev/docs#configuration-paths"
+			"config.kit.paths.base must be a root-relative path and doesn't end with '/'. See https://kit.svelte.dev/docs#configuration-paths"
 		);
 	}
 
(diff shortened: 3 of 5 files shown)


benmccann:
Have you seen https://github.com/sveltejs/kit/pull/1608 ? This PR seems to be editing the same line of code. I'm wondering what you think about the difference between the two PRs

ignatiusmb:
I wasn't aware of that PR, it also seems to solve different, yet similar issues. Checking for a trailing slash would cover this issue and other cases, but I also think it's better to update the docs and mention it in the error message.

ignatiusmb:
I've merged the other PR here, they should appear as a co-author. Also updated this PR's description.

documentation/docs/14-configuration.md
```diff
@@ -116,7 +116,7 @@ Whether to [hydrate](#ssr-and-javascript-hydrate) the server-rendered HTML with
 An object containing zero or more of the following `string` values:
 
 - `assets` — an absolute path, or a path relative to `base`, where your app's files are served from. This is useful if your files are served from a storage bucket of some kind
-- `base` — a root-relative (i.e. starts with `/`) path that specifies where your app is served from. This allows the app to live on a non-root path
+- `base` — a root-relative path that must starts, but doesn't end with `/` (e.g. `/base-path`). This specifies where your app is served from and allows the app to live on a non-root path
```
benmccann, line 119:
```suggestion
- `base` — a root-relative path that must start, but not end with `/` (e.g. `/base-path`). This specifies where your app is served from and allows the app to live on a non-root path
```

packages/kit/src/core/load_config/index.js
```diff
@@ -135,8 +135,10 @@ export function validate_config(config) {
 	// resolve paths
 	const { paths } = validated.kit;
 
-	if (paths.base !== '' && !paths.base.startsWith('/')) {
-		throw new Error('config.kit.paths.base must be a root-relative path');
+	if (paths.base !== '' && (paths.base.endsWith('/') || !paths.base.startsWith('/'))) {
+		throw new Error(
+			"config.kit.paths.base must be a root-relative path and doesn't end with '/'. See https://kit.svelte.dev/docs#configuration-paths"
```
benmccann, line 140:
```suggestion
			"config.kit.paths.base must be a root-relative path that starts but doesn't end with '/'. See https://kit.svelte.dev/docs#configuration-paths"
```

packages/kit/src/core/load_config/index.spec.js
```diff
@@ -171,7 +171,19 @@ test('fails if paths.base is not root-relative', () => {
 				}
 			}
 		});
-	}, /^config\.kit\.paths\.base must be a root-relative path$/);
+	}, /^config\.kit\.paths\.base must be a root-relative path and doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs#configuration-paths$/);
```
benmccann, line 174:
```suggestion
	}, /^config\.kit\.paths\.base must be a root-relative path that starts but doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs#configuration-paths$/);
```

packages/kit/src/core/load_config/index.spec.js
```diff
@@ -171,7 +171,19 @@ test('fails if paths.base is not root-relative', () => {
...
+test("fails if paths.base ends with '/'", () => {
+	assert.throws(() => {
+		validate_config({
+			kit: {
+				paths: {
+					base: '/github-pages/'
+				}
+			}
+		});
+	}, /^config\.kit\.paths\.base must be a root-relative path and doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs#configuration-paths$/);
```
benmccann, line 186:
```suggestion
	}, /^config\.kit\.paths\.base must be a root-relative path that starts but doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs#configuration-paths$/);
```

packages/kit/src/core/load_config/index.js
```diff
@@ -135,8 +135,10 @@ export function validate_config(config) {
 	// resolve paths
 	const { paths } = validated.kit;
 
-	if (paths.base !== '' && !paths.base.startsWith('/')) {
-		throw new Error('config.kit.paths.base must be a root-relative path');
+	if (paths.base !== '' && (paths.base.endsWith('/') || !paths.base.startsWith('/'))) {
+		throw new Error(
+			"config.kit.paths.base must be a root-relative path that starts but doesn't end with '/'. See https://kit.svelte.dev/docs#configuration-paths"
```
benmccann, line 140:
```suggestion
			"kit.paths.base option must be a root-relative path that starts but doesn't end with '/'. See https://kit.svelte.dev/docs#configuration-paths"
```

packages/kit/src/core/load_config/index.spec.js
```diff
@@ -171,7 +171,19 @@ test('fails if paths.base is not root-relative', () => {
 				}
 			}
 		});
-	}, /^config\.kit\.paths\.base must be a root-relative path$/);
+	}, /^config\.kit\.paths\.base must be a root-relative path that starts but doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs#configuration-paths$/);
```
benmccann, line 174:
```suggestion
	}, /^kit\.paths\.base option must be a root-relative path that starts but doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs#configuration-paths$/);
```

packages/kit/src/core/load_config/index.spec.js
```diff
@@ -171,7 +171,19 @@ test('fails if paths.base is not root-relative', () => {
...
+test("fails if paths.base ends with '/'", () => {
+	assert.throws(() => {
+		validate_config({
+			kit: {
+				paths: {
+					base: '/github-pages/'
+				}
+			}
+		});
+	}, /^config\.kit\.paths\.base must be a root-relative path that starts but doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs#configuration-paths$/);
```
benmccann, line 186:
```suggestion
	}, /^kit\.paths\.base option must be a root-relative path that starts but doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs#configuration-paths$/);
```

.changeset/soft-bikes-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Fix redirect URLs when removing/adding trailing slash when `base` path is set
```
benmccann, line 5:
```suggestion
Fix navigation when `base` path is set and validate that option's value
```

benmccann:
updated some of the text slightly and merged. thank you!

ignatiusmb:
Thank _you_, Ben! That must've been a chore

benmccann:
Nope. It was easy enough

PR #1669, opened 2021-06-11
Include `charset=utf-8` by default in `application/json` endpoints.

coreh:
Hey there 👋 

I was getting [Mojibake](https://en.wikipedia.org/wiki/Mojibake) when returning unicode data from my endpoints, so I checked the headers and noticed the `charset=utf-8` option was missing from `Content-Type`. This in turn was causing the response to incorrectly be interpreted as Windows-1252.

This PR changes `src/runtime/server/endpoint.js` to include that by default for `typeof body === 'object'`.

I couldn't find any tickets about this, believe this change is simple enough that it's okay for me to open a PR without filing a ticket first.

### Tests

### Changesets

### DIFF
--- a/.changeset/fuzzy-kids-turn.md
+++ b/.changeset/fuzzy-kids-turn.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Use UTF-8 encoding for JSON endpoint responses by default
--- a/packages/kit/src/runtime/server/endpoint.js
+++ b/packages/kit/src/runtime/server/endpoint.js
@@ -54,8 +54,8 @@ export default async function render_route(request, route) {
 			/** @type {import('types/hooks').StrictBody} */
 			let normalized_body;
 
-			if (typeof body === 'object' && (!type || type === 'application/json')) {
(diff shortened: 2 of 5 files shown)


benmccann:
It looks like there are some CI failures

Conduitry:
Yep, this seems reasonable to do for responses that SvelteKit is automatically serializing.

The test failures look like just some `content-type` values in other tests that need to be updated to include the `charset`.

coreh:
Sorry, it was quite late and my brain was foggy, I think I ran tests locally on the wrong directory. 🥶

I'll fix these after work, later today

coreh:
@benmccann @Conduitry Fixed linting and the broken test.

I also removed the value check for the response body in c8cced1b0f8beda571a5a9632b10fed927a7a825, since upon further digging it looks like the tests don't actually encode/decode the strings as byte buffers like a real HTTP implementation would.

coreh:
Hey 👋

Let me know if you need me to make any additional changes or add more tests.

I can add back the test for UTF-8 encoding/decoding if there's a sanctioned way to make the HTTP response binary during the tests (or I can also try to figure out a mechanism for this if not present)

PR #1679, opened 2021-06-12
Don't check external links on prerender

v1ack:
Check if `a` element has attribute `rel` and if it's equal to `external` don't visit it

### DIFF
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -33,9 +33,9 @@ function get_src(attrs) {
 }
 
 /** @param {string} attrs */
-function get_rel(attrs) {
-	const match = /rel\s*=\s*(?:"(.*?)"|'(.*?)'|([^\s>]*))/.exec(attrs);
-	return match && (match[1] || match[2] || match[3]);
+function is_rel_external(attrs) {
+	const match = /rel\s*=\s*(?:["'][^>]*(external)[^>]*["']|(external))/.exec(attrs);
+	return !!match;
 }
 
 /** @param {string} attrs */
@@ -231,8 +231,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 					const attrs = match[2];
 
 					if (element === 'a' || element === 'link') {
-						const rel = get_rel(attrs);
-						if (rel === 'external') continue;
+						if (is_rel_external(attrs)) continue;
 
 						hrefs.push(get_href(attrs));
 					} else {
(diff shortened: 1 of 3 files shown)


packages/kit/src/core/adapt/prerender.js
```diff
@@ -26,6 +26,12 @@ function get_src(attrs) {
 	return match && (match[1] || match[2] || match[3]);
 }
 
+/** @param {string} attrs */
+function get_rel(attrs) {
```
benmccann, line 30: Can't we just call `element.getAttribute('rel')`?

v1ack: No, because we have `a` as a string, not a node

packages/kit/src/core/adapt/prerender.js
```diff
@@ -199,6 +205,9 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 					const attrs = match[2];
 
 					if (element === 'a' || element === 'link') {
+						const rel = get_rel(attrs);
+						if (rel === 'external') continue;
```
benmccann, line 209: this should probably check if it contains external rather than equaling external in case there are multiple values set in `rel`

v1ack: Yes, you are right, i haven't thought about multiple values. Maybe it's better to check only `external` value in function, because we don't need any other values in `rel`?

benmccann: sure. that sounds reasonable

packages/kit/src/core/adapt/prerender.js
```diff
@@ -26,6 +26,12 @@ function get_src(attrs) {
 	return match && (match[1] || match[2] || match[3]);
 }
 
+/** @param {string} attrs */
+function is_rel_external(attrs) {
+	const match = /rel\s*=\s*(?:["'][^>]*(external)[^>]*["']|(external))/.exec(attrs);
```
benmccann, line 31: this is the kind of thing it'd be really nice to have a unit test for

benmccann: @v1ack can you add a test for this?

benmccann: I added a similar test in https://github.com/sveltejs/kit/pull/1984. That should make it easy for you to get started

v1ack: Sorry for the long wait

benmccann:
```suggestion
	return /rel\s*=\s*(?:["'][^>]*(external)[^>]*["']|(external))/.exec(attrs);
```

PR #1382, opened 2021-05-07
Stricter body types, etags for binary responses

Rich-Harris:
supersedes #1381, fixes #1136. Updates request/response body types, and generates etags for binary bodies

### Before submitting the PR, please make sure you do the following
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [x] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts

PR #1687, opened 2021-06-14
[fix] Expand list of allowed mime types for binary bodies 

### LINKED ISSUE #1438: Serving images from endpoint routes no longer possible
**Describe the bug**

Prior to `1.0.0-next.82` it was possible to serve binary data from endpoint routes. Etag generation appears to cause an exception (`str.charCodeAt is not a function`) on subsequent versions (see #1136) and the fix in #1382, while preventing the exception also prevents serving binary data outside the narrow confines of a `Content-Type: application/octet-stream` http header.

**To Reproduce**

A minimum example which reproduces this issue (as provided in #1136 by @intrikate) is at https://github.com/intrikate/sveltekit-jpeg-endpoint-issue/blob/master/src/routes/index.jpg.js

Results:
* this example works in `next.81` and earlier
* from `next.82` it resulted in an exception `str.charCodeAt is not a function`
* after #1382 landed the exception is avoided but a 500 is served at https://github.com/sveltejs/kit/blob/11e7840c67c2cae17940b0d9664ca0fb3da91eb9/packages/kit/src/runtime/server/endpoint.js#L50

**Expected behaviour**

The ability to serve binary data from endpoints without an exception or error. Obvious use case (as per the example above) is serving images, but more use cases are likely to exist. 

<details>
  System:
    OS: macOS 10.15.7
    Memory: 3.37 GB / 32.00 GB
    Shell: 5.0.11 - /usr/local/bin/bash
  Binaries:
    Node: 14.15.1 - ~/.nave/installed/default/bin/node
    Yarn: 1.22.10 - ~/.nave/installed/default/bin/yarn
    npm: 7.6.3 - ~/.nave/installed/default/bin/npm
  Browsers:
    Chrome: 92.0.4503.0
    Edge: 90.0.818.56
    Firefox: 88.0.1
    Firefox Developer Edition: 79.0
    Safari: 14.1
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.104
    svelte: ^3.34.0 => 3.38.2

Using Firefox and the node adapter, but that's not relevant to this issue.

</details>

**Severity**
This makes it impossible to serve images from an endpoint route. For my use case (a utility for generating static fallback imagery for @abcnews interactive content) that will make it impossible to sveltekit for the task.

So I'd rate the severity high.

JeanJPNM:
Fixes #1438 by expanding the list of allowed mime types for binary bodies.
This allows an endpoint to return a response body with the most commonly used binary mime types.

### Tests

### Changesets


rmunn:
I think you need to look at https://github.com/sveltejs/kit/pull/1382/files, which added those checks, and make sure you've thought about all the other changes that PR made and whether they will still work if you remove those checks. For example, #1382 changes `packages/adapter-cloudflare-workers/files/entry.js` to return a `Uint8Array` if (and only if) the content type includes "application/octet-stream", and otherwise it returns `request.text()`. What's going to happen in that function if the content-type is "image/jpeg" and the request body was an image? Is `request.text()` going to be the right return value, or not? There are many other places where the text "application/octet-stream" appears in #1382's changes. Some of them may not need to be changed if you remove those checks, but some of them might. They definitely all need to be considered.

JeanJPNM:
I can't think of a better solution, what do you think?
```js
/** @param {Request} request */
function isRequestBodyBuffer(request) {
	const type = request.headers.get('content-type');
	return (
		type.startsWith('image') ||
		type.startsWith('audio') ||
		type.startsWith('video') ||
		type.includes('application/octet-stream')
	);
}

```

JeanJPNM:
Is there a way to share core logic with the adapters?
I was thinking of making a function to make updating the constraints for binary content easier

packages/kit/src/runtime/server/endpoint.js
```diff
@@ -63,9 +63,9 @@ export default async function render_route(request, route) {
 			if (
 				typeof body === 'object' &&
 				!(body instanceof Uint8Array) &&
-				(!type || type === 'application/json')
+				(!type || type === 'application/json' || type === 'application/json; charset=utf-8')
```
JeanJPNM, line 66: I would use `type.startsWith('application/json')` since the presence (or absence of the encoding does not change the behavior of the code

JeanJPNM:
Would this even work in a long term perspective? What if someone has a different problem with the mime types, what do we do? We expand the list or tell them to change their choice?

What if instead we let them choose which mime types get parsed as strings and which ones get parsed as buffers?

But then how are we going to exposed that api? And how are we going to access it in the rendering functions? What about the adapters, they barely touch the project configuration, how will they access that information at runtime?

What about the hooks? Don't they run at runtime and would be easy to integrate? We can a hook to let the user determine how to identify text bodies and raw bodies.

No, from what i saw, hooks are not accessible inside the framework itself so you can't just call them inside the enpoint rendered

benmccann:
> Is there a way to share core logic with the adapters?

Here are some examples where we do that:

https://github.com/sveltejs/kit/blob/06e1c25a147aeeb6d67c2daff5a4494882827bbd/packages/adapter-node/src/server.js#L1
https://github.com/sveltejs/kit/blob/06e1c25a147aeeb6d67c2daff5a4494882827bbd/packages/adapter-node/src/shims.js#L2

packages/kit/src/core/utils.js
```diff
@@ -106,3 +106,20 @@ function find_svelte_packages(cwd) {
...
+
+/**
+ * Decides how the body should be parsed based on its mime type.
+ *
+ * This is intended to be used with both requests and responses, to have a consistent body parsing across adapters.
+ *
+ * @param {string} content_type The `content-type` header of a request/response.
+ * @returns {"string"|"buffer"}
+ */
+export function get_body_type(content_type) {
```
benmccann, line 118: We use underscores in names for internal methods and camelCase for public APIs. I think this should be camelCase just as `getRawBody` is since it's exposed to all adapter authors.

How about making it `isBodyBinary` and just returning a `boolean`? `buffer` vs `string` sounds slightly odd to me

JeanJPNM: Ok, I'll make camelCase.
Yep it feels weird, I just made it an enum because it may be expanded on the future

benmccann: yeah, that's what I figured. I'm not too worried about expanding. Worst case we can just make a new method and deprecate the old one

packages/kit/rollup.config.js
```diff
@@ -48,7 +48,8 @@ export default [
 			cli: 'src/cli.js',
 			ssr: 'src/runtime/server/index.js',
 			node: 'src/core/node/index.js',
-			'install-fetch': 'src/install-fetch.js'
+			'install-fetch': 'src/install-fetch.js',
+			utils: 'src/core/utils.js'
```
benmccann, line 52: I wonder if we should make this `adapter-utils` with just the one new method rather than publishing all of our utils

JeanJPNM: Yeah it makes more sense I just didn't know which name to use until now

packages/adapter-cloudflare-workers/files/entry.js
```diff
@@ -57,9 +58,8 @@ async function handle(event) {
 /** @param {Request} request */
 async function read(request) {
 	const type = request.headers.get('content-type') || '';
-	if (type.includes('application/octet-stream')) {
+	if (isContentTypeBinary(type) === 'buffer') {
```
benmccann, line 61: this line will need to be updated

packages/kit/src/core/adapter-utils.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Decides how the body should be parsed based on its mime type.
+ *
+ * This is intended to be used with both requests and responses, to have a consistent body parsing across adapters.
+ *
+ * @param {string} content_type The `content-type` header of a request/response.
+ * @returns {boolean}
+ */
+
```
benmccann, line 9: should remove this blank line

packages/kit/src/runtime/server/endpoint.js
```diff
@@ -37,24 +38,27 @@ export default async function render_route(request, route) {
 			headers = lowercase_keys(headers);
 			const type = headers['content-type'];
 
+			const is_type_binary = isContentTypeBinary(type);
+			/** @type {import('types/hooks').StrictBody} */
+			let normalized_body;
```
benmccann, line 43: I would move this line back down where it was originally. No reason to declare it higher up since it's not used yet

benmccann (approved):

benmccann:
lgtm once the tests are passing. can you add a changeset as well?

packages/kit/src/runtime/server/endpoint.js
```diff
@@ -37,24 +38,28 @@ export default async function render_route(request, route) {
...
 			const type = headers['content-type'];
 
+			const is_type_binary = type && isContentTypeBinary(type);
+
 			// validation
-			if (type === 'application/octet-stream' && !(body instanceof Uint8Array)) {
+			if (is_type_binary && !(body instanceof Uint8Array)) {
 				return error(
-					`${preface}: body must be an instance of Uint8Array if content type is application/octet-stream`
+					`${preface}: body must be an instance of Uint8Array if content type is image/*, audio/*, video/* or application/octet-stream `
```
benmccann, line 46:
```suggestion
					`${preface}: body must be an instance of Uint8Array if content type is image/*, audio/*, video/* or application/octet-stream`
```

packages/kit/src/core/adapter-utils.js
```diff
@@ -0,0 +1,16 @@
...
+ * Decides how the body should be parsed based on its mime type.
+ *
+ * This is intended to be used with both requests and responses, to have a consistent body parsing across adapters.
+ *
+ * @param {string} content_type The `content-type` header of a request/response.
+ * @returns {boolean}
+ */
+export function isContentTypeBinary(content_type) {
+	return (
+		content_type.startsWith('image') ||
```
benmccann, line 11:
```suggestion
		content_type.startsWith('image/') ||
```

packages/kit/src/core/adapter-utils.js
```diff
@@ -0,0 +1,16 @@
...
+		content_type.startsWith('audio') ||
```
benmccann, line 12:
```suggestion
		content_type.startsWith('audio/') ||
```

packages/kit/src/core/adapter-utils.js
```diff
@@ -0,0 +1,16 @@
...
+ * This is intended to be used with both requests and responses, to have a consistent body parsing across adapters.
+ *
+ * @param {string} content_type The `content-type` header of a request/response.
+ * @returns {boolean}
+ */
+export function isContentTypeBinary(content_type) {
+	return (
+		content_type.startsWith('image/') ||
+		content_type.startsWith('audio') ||
+		content_type.startsWith('video') ||
```
benmccann, line 13:
```suggestion
		content_type.startsWith('video/') ||
```

packages/kit/src/core/adapter-utils.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Decides how the body should be parsed based on its mime type.
+ *
+ * This is intended to be used with both requests and responses, to have a consistent body parsing across adapters.
+ *
+ * @param {string} content_type The `content-type` header of a request/response.
+ * @returns {boolean}
+ */
+export function isContentTypeBinary(content_type) {
```
drzax, line 9: The logic is possibly still too blunt. There are at least a few content types I'd consider common for which I think this would return the wrong result:
* `image/svg+xml`
* `application/zip`
* `application/pdf`

An escape hatch for the developer here might be useful too given all the potential edge cases that the wonderful world of mime types involves.

JeanJPNM: I know that's why I made #1829

benmccann: I sent https://github.com/sveltejs/kit/pull/1890 to help with this

PR #1695, opened 2021-06-16
Fail if config.kit.appDir starts or ends with a slash

bjon:
Everything in [https://github.com/sveltejs/kit/blob/master/packages/kit/src/core/build/index.js](https://github.com/sveltejs/kit/blob/master/packages/kit/src/core/build/index.js) expects `config.kit.appDir` to be without leading and trailing slashes. This fix fails at `validate_config`, instead of runtime.

### Tests

### Changesets


Rich-Harris:
thank you!

PR #1696, opened 2021-06-16
chore(kit): correct `engines` constraint

lukeed:
Maintain fix from #1470, but loosen other constraints which had no business being there.


benmccann:
Hmm. I'm not sure about this. The motivation for that version number came from: https://github.com/sveltejs/kit/pull/1000#issuecomment-841808940

benmccann (approved):
there's not really a reason anyone should be using Node 15 and I do think people have hit some issues with certain versions of it, but I guess I don't have an objection to this either

lukeed:
15 is irrelevant, shouldn't be used but will still work. This was mostly about lowering to the true values

Conduitry:
I believe `14.13.1` was used because it sounded like `14.13.0` had a bug that would cause problems for us.

It looks like the net result of this PR is just to allow `14.13.0` and to allow `15.x`, and I doubt whether either of those are good things to be doing.

dominikg:
hmm, does this mean vite-plugin-svelte should also revise its constraints? given 15 is about to be EOL and 14.13.1 is a patch level i don't really see the value behind this and think the previous constraint was ok.

if 15 is allowed, should it be included in tests?

benmccann:
I don't think we should test against 15. I'm maybe okay saying we shouldn't prevent people from using it, but it definitely shouldn't be supported in any way

PR #1662, opened 2021-06-10
Use the same option source for SSL (certs and keys) in Preview and Dev

JBusillo:
Fixes #1659.  

The dev server passes the vite options (config.kit.vite) when creating the http(s) server.  This is functioning correctly.
The preview server passes the kit options (config.kit) when creating the http(s) server.  This is not functioning correctly.

In preview mode, when the user specifies a cert/key pair, the cert/key pair is ignored, since the logic is looking for the option in the wrong place.  Instead, Kit will generate its own cert/key pair.  You can observe this by examining the certificate details within the browser.

This change aligns where the dev and preview modes look for a user-specified cert/key pair.

For reference, here is an example of a user-specified cert/key pair in `svelte.config.js`
```
/** @type {import('@sveltejs/kit').Config} */
import fs from "fs";
const config = {
  kit: {
    // hydrate the <div id="svelte"> element in src/app.html
    target: "#svelte",
    vite: {
      server: {
        https: {
          key: fs.readFileSync("example.key"),
          cert: fs.readFileSync("example.crt"),
        },
      },
    },
  },
};
export default config;
```
### Before submitting the PR, please make sure you do the following
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts

PR #1698, opened 2021-06-17
fix: deep_merge should stop recursing at class instances

johnnysprinkles:
In #1662 @JBusillo notes that my #1572 is causing an issue when you pass key and cert values in `kit.vite.server.https`.

I see that he's passing buffers like this:

```
const config = {
  kit: {
    target: "#svelte",
    vite: {
      server: {
        https: {
          key: fs.readFileSync("example.key"),
          cert: fs.readFileSync("example.crt"),
        },
      },
    },
  },
};
```

which my deep_merge function tries to deeply clone. Those Buffer objects have a ton of methods on them. Not sure why this doesn't work (need to look more deeply into it), but it seems like we'd want to not clone those rich object and only clone "plain old JS objects" who's constructor is Object. Things that aren't class instances or built-ins.

This is a quick fix. I need to add tests but I'm on the road, won't get to it until this weekend.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/config/index.js
+++ b/packages/kit/src/core/config/index.js
@@ -157,13 +157,15 @@ export function validate_config(config) {
  */
 function merge_into(a, b, conflicts = [], path = []) {
 	/**
+	 * Checks for "plain old Javascript object", typically made as an object
+	 * literal. Excludes Arrays and built-in types like Buffer.
 	 * @param {any} x
 	 */
-	const is_object = (x) => typeof x === 'object' && !Array.isArray(x);
+	const is_plain_object = (x) => typeof x === 'object' && x.constructor === Object;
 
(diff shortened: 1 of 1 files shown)


johnnysprinkles:
Oh I see what's going on, those two methods `toString` and `toLocaleString`, those are enumerable properties of Buffer, but aren't enumerable properties of a plain object. This is an issue I didn't forsee. This fix will fix that though, treating Buffer as opaque.

johnnysprinkles:
Thanks for merging Ben... note that there is still a minor issue that if you had a config like:

```
vite: {
  server: {
    toString: () => {}
  }
}

It would still fail due to that conflicting with the Object's own method. I'll have another small fix for that later.

Issue #1697, opened 2021-06-16
prerendering tries to crawl data-ci-src attribute in addition to src and fails


Shackless:
**Describe the bug**

I'm using a responsive image library where you set `<img data-ci-src="myimage.png"/>` which is then later translated to `src` in the client based on viewport size and other params. This work fine in dev mode but apparently, adapter-static is validating the `data-ci-src` attribute on build which always fails with a 404 because the src path is relative and will be something like `https://cloudprovider.com/_client_/myimage.png` after the translation. The files referenced in `data-ci-src` don't actually exist in my sveltekit project. 

If I comment out the `data-ci-src` attribute in the `<img>` tag, it works.

Sounds like a regex matching the image tag and an attribute ending in src instead of matching just "src" or something like that (wild guess). 

**Logs**

```bash
> Using @sveltejs/adapter-static
> 404 /myproj/myproj-screenshot-1.png (linked from /work/myproj)
```
**To Reproduce**
Create a new project, add `<img data-ci-src="xyz.png" />` where xyz.png doesn't exist. Build with adapter-static.

**Expected behavior**
Only `<img src="...">` should be validated but not `<img data-src="...">` or `<img data-ci-src="...">`.

**Stacktraces**
<details>
  <summary>Stack trace</summary>

```
Error: 404 /myproj/myproj-screenshot-1.png (linked from /work/myproj)
    at error (file:///Users/[...]/node_modules/@sveltejs/kit/dist/chunks/index7.js:94:11)
    at visit (file:///Users/[...]/node_modules/@sveltejs/kit/dist/chunks/index7.js:176:5)
    at async visit (file:///Users/[...]/node_modules/@sveltejs/kit/dist/chunks/index7.js:236:6)
    at async visit (file:///Users/[...]/node_modules/@sveltejs/kit/dist/chunks/index7.js:236:6)
    at async visit (file:///Users/[...]/node_modules/@sveltejs/kit/dist/chunks/index7.js:236:6)
    at async visit (file:///Users/[...]/node_modules/@sveltejs/kit/dist/chunks/index7.js:236:6)
    at async visit (file:///Users/[...]/node_modules/@sveltejs/kit/dist/chunks/index7.js:236:6)
    at async visit (file:///Users/[...]/node_modules/@sveltejs/kit/dist/chunks/index7.js:236:6)
    at async prerender (file:///Users/[...]/node_modules/@sveltejs/kit/dist/chunks/index7.js:245:5)
[...]

PR #1700, opened 2021-06-17
Stricter regex for getting element attributes during prerendering

benmccann:
Fixes https://github.com/sveltejs/kit/issues/1697

### DIFF
--- a/.changeset/seven-planes-fly.md
+++ b/.changeset/seven-planes-fly.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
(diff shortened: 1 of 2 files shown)


dummdidumm:
Using `\s` instead is safer I think since one could have tabs or a newline before `src` (although that's probably very rare)

benmccann:
good catch!

Conduitry (changes requested):
Do the regexes also need to support quotes right before `href`/`src`? What happens if, for some reason, someone has `whatever="foo"href="something"` without a space between the attributes? This is valid HTML, and looks like it would be missed by this.

PR #1724, opened 2021-06-21
Register service worker even with no client-side router or hydration

jpaquim:
When testing the behavior of the different page config options in a minimal PWA, I was caught up by the fact that if `router` and `hydrate` are both false (leading to no client-side bundle being added to the page), the service worker is implicitly not registered.

From what I understand, client-side routing and hydration should be orthogonal to service worker registration, so I tried to adapt the server rendering code to support simultaneously using a service worker and disabling other client-side code.

I would appreciate some discussion around the general use-case of using service workers in this context, as well as feedback on the implementation (there is currently a bit of duplication between the `service_worker` SSR render option and the `VITE_SVELTEKIT_SERVICE_WORKER` build environment variable, as well as in the output `service-worker.js` location itself)

### DIFF
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -142,6 +142,14 @@ export async function render_response({
 				}` : 'null'}
 			});
 		</script>`;
+	} else if (options.service_worker) {
+		// a service worker will usually be registered inside the call to `start()` above, but not when
+		// `include_js` is false (both router = false and hydrate = false)
+		init = `<script type="module">
+			if ('serviceWorker' in navigator) {
+				navigator.serviceWorker.register('${options.service_worker}');
+			}
+		</script>`;
 	}
 
 	const head = [
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -142,10 +142,10 @@ export async function render_response({
 				}` : 'null'}
 			});
 		</script>`;
-	} else if (options.service_worker) {
-		// a service worker will usually be registered inside the call to `start()` above, but not when
-		// `include_js` is false (both router = false and hydrate = false)
-		init = `<script type="module">
+	}
+
+	if (options.service_worker) {
+		init += `<script>
 			if ('serviceWorker' in navigator) {
 				navigator.serviceWorker.register('${options.service_worker}');
 			}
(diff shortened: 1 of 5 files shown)


Rich-Harris:
Thank you. I actually wonder if the service worker registration should _always_ be decoupled from `start.js` — i.e. if we should generate something like this:

```html
<!-- omitted if router/hydrate are both false -->
<script type="module">
  import { start } from '...';
  start({...});
</script>

<!-- always included if service worker exists -->
<script>
  if ('serviceWorker' in navigator) {
    navigator.serviceWorker.register('/service-worker.js');
  }
</script>
```

Doing so would mean we could get rid of the `VITE_SVELTEKIT_SERVICE_WORKER` stuff altogether, and would allow the service worker to be registered even if the app failed to boot for whatever reason. What do you think?

Also, mind adding a changeset to this PR? `pnpx changeset` and follow the prompts

jpaquim:
Makes complete sense to me, I was actually starting to refactor in that direction of removing the service-worker related code from `start.js`, and make the concerns effectively orthogonal, but decided to stop and gather feedback before progressing any further.

I hadn't considered the scenario of loading the SW even if the app bundle fails to load, from what I understand it should be relatively harmless and even desired in some situations, but I wonder if it can also be a footgun... In any case, service worker registration being disabled in development should at least mitigate it.

Another observable difference is whether or not the service worker registration code is minified by rollup (at svelte-kit build-time), but I guess there's always the possibility of minifying the HTML at response-handling/static rendering time as described in the docs https://kit.svelte.dev/migrating#integrations-html-minifier

packages/kit/src/runtime/server/page/render.js
```diff
@@ -144,6 +144,14 @@ export async function render_response({
 		</script>`;
 	}
 
+	if (options.service_worker) {
```
jpaquim, line 147: As an alternative, I considered adding it below to the `head` array, something like:
```js
	const head = [
		rendered.head,
		styles.size && !options.amp
			? `<style data-svelte>${Array.from(styles).join('\n')}</style>`
			: '',
		links,
		init,
		options.service_worker
			? `<script>
			if ('serviceWorker' in navigator) {
				navigator.serviceWorker.register('${options.service_worker}');
			}
		</script>`
			: ''
	].join('\n\n\t\t');
```
Which would be preferable here?

Rich-Harris: i think it's great as-is

Rich-Harris:
thank you!

Issue #1262, opened 2021-04-28
Browser back button stops working on pages with router disabled


ghost:
**Describe the bug**
Browser back button stops working on pages with router disabled.

**To Reproduce**
```javascript
<script context="module">
  export const router = false
</script>
```

**Severity**
Blocking.

erwin-k:
If someone can point me to the code where this problem might come from I'd be happy to fix it.

More detail:
On a page with router disabled, when back button is clicked the url changes but rendering doesn't take place.

ignatiusmb:
I can't reproduce with the latest version using the demo app, tested on both Chromium-based browser and Firefox

PR #1730, opened 2021-06-21
fix browser forward/backward nav when router is disabled

erwin-k:
Refers to #1262
The page wouldn't reload or render when navigating back from a page of which the router was disabled on.
I believe this is the only way to achieve desired behaviour.

### Tests


erwin-k:
I have no idea what I'm doing :) 
I'll come back to this when I know how to use git...

Issue #1706, opened 2021-06-18
Respect kit.vite.esbuild config when building with Vercel adapter


mattjennings:
**Is your feature request related to a problem? Please describe.**
I have a dependency that needs to be excluded from the bundle, so I mark it as external in svelte.config.js like so:

```
const config = {
  preprocess: preprocess(),

  kit: {
    adapter: vercel(),
    target: '#svelte',
    vite: {
      esbuild: {
        exclude: ['external-dependency']
      }
    }
  }
}
```

This correctly gets excluded in the .svelte-kit build, but it gets included in the adapter build's output.

**Describe the solution you'd like**
Use the `kit.vite.esbuild` settings from svelte.config.js in the adapter build 

**Describe alternatives you've considered**
I noticed the node adapter excludes pkg.dependencies - maybe we just need to do the same for Vercel adapters? Is there a reason for doing that in one adapter and not another?

**How important is this feature to you?**
Important, I'm unable to deploy until I can exclude this dependency.

PR #1732, opened 2021-06-21
adapter-vercel should respect esbuild configs

thmsmlr:
This patch attempts to solve https://github.com/sveltejs/kit/issues/1706
We pass through the esbuild configurations in svelte.config.cjs into the
esbuild step of the vercel adapter.

This is primarily to solve the case where we have external packages that
are skipped by esbuild in adapater-node, but not adpater-vercel.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-vercel/index.js
+++ b/packages/adapter-vercel/index.js
@@ -8,7 +8,7 @@ export default function () {
 	const adapter = {
 		name: '@sveltejs/adapter-vercel',
 
-		async adapt({ utils }) {
+		async adapt({ utils, config }) {
 			const dir = '.vercel_build_output';
 			utils.rimraf(dir);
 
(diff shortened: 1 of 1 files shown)


benmccann:
If we're going to make this change, I assume we should also make it in the other adapters that use `esbuild` as well?

benmccann:
There's a related PR that's attempting to pass esbuild options into `adapter-node` via a new option in https://github.com/sveltejs/kit/pull/1692

packages/adapter-vercel/index.js
```diff
@@ -31,7 +31,8 @@ export default function () {
 				entryPoints: ['.svelte-kit/vercel/entry.js'],
 				outfile: join(dirs.lambda, 'index.js'),
 				bundle: true,
-				platform: 'node'
+				platform: 'node',
+				...config?.kit?.vite()?.esbuild
```
benmccann, line 35: I think that reusing `vite`'s `esbuild` config here is not the best way. People way want to configure the two separately. E.g. see https://github.com/sveltejs/kit/pull/1692#issuecomment-865615019

7antra:
@Nick-Mazuk any news ?

PR #1735, opened 2021-06-22
[fix] preserve user defined config and files on `svelte-kit package`

### LINKED ISSUE #1734: svelte-kit package retain original casing in file names

`svelte-kit package` should not change case of copied files. In my case, I'm getting `readme.md -> package/README.md`.

Project files should be treated as faithfully as possible.

Nice to have, not essential.

### LINKED ISSUE #1856: default export from svelte-kit package
How do I create a `default` export work in a `svelte-kit package`? Would be cool if [the docs](https://kit.svelte.dev/docs#packaging) included some guidance on that. If I add an index file that re-exports a Svelte component as `default`,

```js
// src/lib/index.js
export { default } from './MyComponent.svelte'
```

and then install the package and try

```svelte
import MyComponent from 'my-package'
```

I get

> Must use import to load ES Module: my-package/index.js
> require() of ES modules is not supported.
> require() of my-package/index.js from repo/node_modules/vite/dist/node/chunks/dep-0ed4fbc0.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
> Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from my-package/package.json.

The easiest solution to this in the past was to use the `"svelte"` field in `package.json` but `sveltekit package` does not copy that field into its output.

[I asked this question here as well](https://github.com/sveltejs/kit/pull/1499#issuecomment-874549564).

ignatiusmb:
This PR fixes multiple stuff regarding `package` cli command

- Retain original `README.md` filename casing from project root.
- Preserve initial `package.json` and only adds `exports` map when it's not defined by the user.
- Force forward slash paths for `exports` map values

Fixes #1734 
Fixes #1856


benmccann (approved):
this generally seems like a good idea to me, but I might leave this open for a little while to see if Rich had any other concerns in mind that I'm not aware of when he wrote it

ignatiusmb:
Aha, I found an exception to `scripts` in `package.json` keys that we may want to exclude from the final output. It may run some of the lifecycle scripts if defined and most likely ruin the build or publish process. Other than `scripts`, I think everything else should be safe and have no special stuff.

***

Sure, I'm also curious for the reasons behind those specific keys. While we're here, I think the only odd one from the changes listed in the description was the "adds `exports` map when it's not defined by the user", so I'll explain my reasoning in case anyone's curious.

Published npm packages can now "lock" their imports (or `exports` in this case) to just a certain paths or modules, so there may be a `package/internal` used only for the modules inside and cannot be directly accessed by the end-user. There's also a reason for authors to manually specify the exports, be it only exposing some of them or adding `{ "browser": ... }` and such. `svelte-kit package` forcefully overwrites this and authors would need to manually update the generated `package.json` again. I think the easiest thing to do without adding another config option is to just check the existence and use the built-in exports map when there's none defined.

dummdidumm:
In #1646 I added logic for adding `types` to the `package.json` . If this was merged, should I revert the part about "define in your `svelte-config.js` what the `types` value is? I guess it would make sense to remove it.

ignatiusmb:
I think it's a good addition for the same reason some authors might not manually specify them. Perhaps only add `types` and use the defined config when none is defined in the original `package.json` too, just like `exports` here?

packages/kit/src/core/make_package/index.js
```diff
@@ -89,27 +78,27 @@ export async function make_package(config, cwd = process.cwd()) {
 
 		write(path.join(cwd, config.kit.package.dir, out_file), out_contents);
 
-		if (exports_filter(file)) {
-			const entry = `./${out_file}`;
-			package_pkg.exports[entry] = entry;
+		if (!user_defined_exports && exports_filter(file)) {
+			const entry = `./${out_file.replace(/\\/g, '/')}`;
```
benmccann, line 82: What's the `replace` thing that had to be added here? Is that for Windows support? I wouldn't expect it's really necessary

ignatiusmb: Spot on, sadly it's necessary because the generated keys for `exports` map would also look like `"./module\\foo\\bar"` and results in a `Missing "./module/foo/bar" export` error.

benmccann (approved):

dummdidumm (commented):
Looks good to me. Could you add a test verifying the new behavior? I recently did set up the infrastructure to test this area.

ignatiusmb:
Added and adjusted the tests to also handle nested files.

I haven't added the test for the `README` copying, I think I prefer #1747 more. But, in case this is merged before the other, it'll still solve the initial problem in #1734, so I won't revert the changes for now

dummdidumm:
Let's first try to get #1747 in and then adjust this one.

ignatiusmb:
Thanks @dummdidumm 😄, finally got it working now

benmccann:
I merged #1747, so you can rebase this one now

dummdidumm (approved):

PR #1739, opened 2021-06-22
fix: deep_merge shows false errors when using key names that happen to be Object prototype methods

johnnysprinkles:
#1572 introduced a couple of minor issues. One I fixed in #1698 but there was still a minor issue where, if your config has keys which happen to be Object prototypes like `toString`, `valueOf`, etc, you'd get false conflict errors reported. It's a simple one-line fix. Not super high priority and unlikely anyone would every hit that issue but good to be robust.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/config/index.js
+++ b/packages/kit/src/core/config/index.js
@@ -181,7 +181,13 @@ function merge_into(a, b, conflicts = [], path = []) {
 			}
 			a[prop].push(...b[prop]);
 		} else {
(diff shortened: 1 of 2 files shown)


johnnysprinkles:
Maybe I don't need all those block comments, I can remove if you like.

benmccann:
I think the code comment is good. I'd probably remove the test comments since they refer to things that aren't issues anymore

PR #1743, opened 2021-06-23
Fix regex for getting links to crawl during prerendering

mou-t:
crawl first attribute during prerendering

### DIFF
--- a/.changeset/eighty-candles-grow.md
+++ b/.changeset/eighty-candles-grow.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Fix regex for getting links to crawl during prerendering
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -16,13 +16,13 @@ function clean_html(html) {
 
 /** @param {string} attrs */
 function get_href(attrs) {
-	const match = /[\s'"]href\s*=\s*(?:"(.*?)"|'(.*?)'|([^\s>]*))/.exec(attrs);
+	const match = /([\s'"]|^)href\s*=\s*(?:"(.*?)"|'(.*?)'|([^\s>]*))/.exec(attrs);
 	return match && (match[1] || match[2] || match[3]);
 }
 
 /** @param {string} attrs */
 function get_src(attrs) {
-	const match = /[\s'"]src\s*=\s*(?:"(.*?)"|'(.*?)'|([^\s>]*))/.exec(attrs);
+	const match = /([\s'"]|^)src\s*=\s*(?:"(.*?)"|'(.*?)'|([^\s>]*))/.exec(attrs);
 	return match && (match[1] || match[2] || match[3]);
 }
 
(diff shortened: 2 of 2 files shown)


benmccann:
I don't quite understand what issue this is fixing. Can you explain in more detail?

mou-t:
Sorry. I'm not good at English.

Currently not crawl

```
//svelte components
<a href="example">example</a>
```

```href="example"``` is not match regex

benmccann:
I don't understand what your PR changes. `\s` matches a space and there's a space before `href` in your example. Can you explain why the existing code wouldn't work for your example and the updated code will?

mou-t:
This example 

pattern.exec function result

match[1] = ```a```
match[2] = ```href="example/aaa"``` (not space and , ' and , ")

https://github.com/sveltejs/kit/blob/d259bca84add777a8269c5279a0ca86f9869e72e/packages/kit/src/core/adapt/prerender.js#L189-L203

ignatiusmb:
Oh! It took me a while, but I think I got it. 

The recent changes from #1700 might actually broke prerendering instead in some (and perhaps most) cases. All `get_href`, `get_src`, and `get_srcset_urls` are only used locally in `kit/packages/kit/src/core/adapt/prerender.js`. They're also passed a value from a regexp pattern exec here (linked from above, but the full scope)

https://github.com/sveltejs/kit/blob/d259bca84add777a8269c5279a0ca86f9869e72e/packages/kit/src/core/adapt/prerender.js#L189-L209

If an anchor tag or image tag have their `href` or `src` as the first attribute, this `get_*` functions would fail. I guess it's better explained in a [RegExr REPL](https://regexr.com/60lsj), look at the tools below and click on the details tab for the captured groups. Transferring those to the current `get_*` regexp patterns, we can see that (change to the tests tab) [tag 1 and 3 fails](https://regexr.com/60lsv).

ignatiusmb (approved):
This is a really good catch! Maybe also add a changeset so we can quickly publish a new release for this fix.

benmccann:
Oh, I see. Because it's just getting the attributes as input and not the whole tag

benmccann (approved):

truumahn:
I think this might have resulted in [Rest parameters as directory names](https://kit.svelte.dev/docs#routing-advanced-rest-parameters) not picked up correctly when prerendering dynamic routes (with `adapter-static` at least).

I'm using this [template project for rendering markdown](https://github.com/jakobbouchard/sveltekit-netlifycms) and since it's using an older version of SvelteKit prerender works properly.

After some debugging I've found that the issue is with this regex, once you update to 118+ the pages are no longer prerendered on dynamic routes.

Steps to reproduce:
1. `git clone https://github.com/jakobbouchard/sveltekit-netlifycms`
2. `npm install`
3. `npm run build`
4. `cat build/blog/test-file/index.html` ~ pages on dynamic routes are prerendered in `build` directory
5. `rm -rf .svelte-kit build`
6. `npm upgrade`
7. `npm run build`
8. `cat build/blog/test-file/index.html` ~ no file = no prerender

After directly editing `node_modules/@sveltejs/kit/dist/chunks/index6.js` and removing the regex parts added in this PR prerender works again.

PR #1747, opened 2021-06-23
[fix] copy essentials files from root on packaging

### LINKED ISSUE #1734: svelte-kit package retain original casing in file names

ignatiusmb:
Alternative PR, only to #1735 point number one. This will copy all (_and only_) files in the root folder to package directory output, rather than having an explicit whitelist of files to copy like `README.md`, `LICENSE`, and so on.

This should also cover copying the `.npmignore` so it should be the same and predictable for authors to blacklist files there. To keep it consistent with the ecosystem, `files` key is added to the `package.json` output as well for those who prefer to whitelist instead.

Fixes #1734


ignatiusmb:
On second thought, I think it's better to include just the "essentials" from the root folder to the final package output. One could argue that authors could just place these files in `src/lib` directly, but it would be weird &mdash; let's say someone's looking at the source code to read about the `CHANGELOG` which is usually placed at the root, but now needs to go to `src/lib` first to do so &mdash; using symlinks could be an option, but it's unconventional, to say the least.

Now, what counts as "essentials" could be subjective to each authors accordingly. In the meantime, I've added what npm counts as their "essentials" in which these files would always be included, regardless of the `files` or ignore values. I've also placed them in a separate file for easier editing in the future. Another option is to provide a key in the config file, this could be done later on, not sure where to put it but perhaps something around

- `kit.package.files.essentials`
- `kit.package.files.root`
- `kit.package.essentials`
- `kit.package.root`

benmccann (approved):
this seems better than what we currently have though I agree that there might be additional files that the author could want to include. another thought - I wonder if we could get those from the `package.json` `files` field

packages/kit/src/core/make_package/index.js
```diff
@@ -89,11 +91,16 @@ export async function make_package(config, cwd = process.cwd()) {
 		JSON.stringify(package_pkg, null, '  ')
 	);
 
-	const project_readme = path.join(cwd, 'README.md');
-	const package_readme = path.join(cwd, config.kit.package.dir, 'README.md');
+	const whitelist = fs.readdirSync(cwd).filter((file) => {
+		return essential_files.some((name) => file.startsWith(name));
```
dummdidumm, line 112: I think letter case allignment needs to happen here. Else `readme.md` is filtered out, `startsWith` is case sensitive

packages/kit/src/core/make_package/essential_files.js
```diff
@@ -0,0 +1 @@
+export default ['README', 'LICENSE', 'CHANGELOG', '.gitignore', '.npmignore'];
```
dummdidumm, line 1: Not 100% sure if we need an extra file for this

packages/kit/src/core/make_package/index.js
ignatiusmb: You're too quick! Was just about to push an update for this too

packages/kit/src/core/make_package/essential_files.js
ignatiusmb: Me too, I explained a bit as to why on the second paragraph of https://github.com/sveltejs/kit/pull/1747/#issuecomment-867452394. Hoping to get some feedback first and if it's just these files for now, I can merge it back

dummdidumm (approved):

packages/kit/src/core/make_package/essential_files.js
benmccann: @ignatiusmb did you link to the right PR / issue in your response here? It links to this PR itself and I don't really see an explanation

benmccann: I combined the files for now and will go ahead and merge this. We can separate it back out when there's more to put there

PR #1751, opened 2021-06-24
Specify actual Svelte version requirement

benmccann:
I wanted to check if an issue I'm facing is a regression in Svelte, but can't use an older version. It seems Rich frequently bumps the peerDependency to be the latest ([example](https://github.com/sveltejs/kit/pull/1374/files#diff-0ac36eb0a5cc3505c582d3959e4bb51093770685d0de4a864f7cd9b68bd50b8fL40)), but I don't think there's actually a reason the latest is required. This PR sets it equal to the version required by Sapper

### DIFF
--- a/.changeset/rich-pumpkins-applaud.md
+++ b/.changeset/rich-pumpkins-applaud.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Specify actual Svelte version requirement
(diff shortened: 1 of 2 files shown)


dominikg:
vite-plugin-svelte uses cssHash added in 3.34 https://github.com/sveltejs/svelte/blob/master/CHANGELOG.md#3340

PR #1754, opened 2021-06-24
allow customisation of HOST and PORT env vars in adapter-node

### LINKED ISSUE #1752: Ability to customize which environment variables are used for HOST/PORT in Node adapter
In deployments at work, all of our environment variables are prefixed with unique identifiers indicating what service they are intended for. For SvelteKit apps using the Node adapter, this is impossible, as they have hardcoded into them special behavior for the `HOST` and `PORT` environment variables.

Some configuration passed to the Node adapter function that lets you specify which environment variable names to use instead of `HOST` and `PORT` to determine at runtime what the server listens on.

I haven't tried putting e.g. `process.env['PORT'] = process.env['FOO_PORT'];` at the top of the hooks file, but I suspect that it will already be too late by then. If that doesn't work, I don't really have any other ideas besides using `sed` or something to hack the files after they're built.

This is going to depend on how upset the devops guy at work gets when I tell him he has to use the very generic `HOST` and `PORT` env vars for SvelteKit apps, as opposed to the specific prefixed env vars that every other service uses. I'm not clear on how environment variable configuration is getting handled, but this might actually create a conflict when another SvelteKit app needs to use different values.

**Additional context**
This was possible in Sapper because you could directly edit `src/server.js` to make it use the other environment variables instead.

Conduitry:
Resolves #1752.

I'm definitely open to other names for this configuration, because I'm not thrilled on what I currently have. If someone has a better idea of what to say in the adapter's readme, I'd also like to hear about that.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-node/README.md
+++ b/packages/adapter-node/README.md
@@ -15,7 +15,11 @@ export default {
 		adapter: adapter({
 			// default options are shown
 			out: 'build',
-			precompress: false
+			precompress: false,
+			env: {
+				host: 'HOST',
+				port: 'PORT'
+			}
 		})
 	}
 };
@@ -31,13 +35,15 @@ The directory to build the server to. It defaults to `build` — i.e. `node buil
 
 Enables precompressing using gzip and brotli for assets and prerendered pages. It defaults to `false`.
 
-## Environment variables
+### env
 
 By default, the server will accept connections on `0.0.0.0` using port 3000. These can be customised with the `PORT` and `HOST` environment variables:
 
 ```
 HOST=127.0.0.1 PORT=4000 node build
 ```
+
+However, which environment variables are used can also be customised by setting the `env` option. If specified, TKTKTK. It defaults to `{ host: 'HOST', port: 'PORT' }`.
 
 ## License
 
--- a/packages/adapter-node/README.md
+++ b/packages/adapter-node/README.md
@@ -43,7 +43,7 @@ By default, the server will accept connections on `0.0.0.0` using port 3000. The
 HOST=127.0.0.1 PORT=4000 node build
 ```
 
-However, which environment variables are used can also be customised by setting the `env` option. If specified, TKTKTK. It defaults to `{ host: 'HOST', port: 'PORT' }`.
+You can specify different environment variables if necessary using the `env` option.
 
 ## License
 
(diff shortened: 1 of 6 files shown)


packages/adapter-node/README.md
```diff
@@ -15,7 +15,11 @@ export default {
 		adapter: adapter({
 			// default options are shown
 			out: 'build',
-			precompress: false
+			precompress: false,
+			env: {
```
dominikg, line 19: maybe call it `listen` instead of env so it's more obvious whats happening and add default values here?

arxpoetica (approved):
👍

packages/adapter-node/index.js
```diff
@@ -12,10 +12,18 @@ const pipe = promisify(pipeline);
...
+ *     host?: string;
+ *     port?: string;
+ *   };
  * }} options
  */
-export default function ({ out = 'build', precompress } = {}) {
+export default function ({
+	out = 'build',
+	precompress,
+	env: { host: hostEnv = 'HOST', port: portEnv = 'PORT' } = {}
```
Rich-Harris, line 25: _ahem_ i think you mean `host_env` and `port_env`?

packages/adapter-node/README.md
```diff
@@ -31,14 +35,16 @@ The directory to build the server to. It defaults to `build` — i.e. `node buil
...
-## Environment variables
+### env
 
 By default, the server will accept connections on `0.0.0.0` using port 3000. These can be customised with the `PORT` and `HOST` environment variables:
 
 ```
 HOST=127.0.0.1 PORT=4000 node build
 ```
 
+However, which environment variables are used can also be customised by setting the `env` option. If specified, TKTKTK. It defaults to `{ host: 'HOST', port: 'PORT' }`.
```
Rich-Harris, line 46: maybe just this?

```suggestion
You can specify different environment variables if necessary using the `env` option.
```

packages/adapter-node/index.js
Conduitry: Crap, whoops, haha, I do indeed.

Conduitry: Fixed in e4d97acbf88533551014cf1a03d19ac700b24098.

packages/adapter-node/README.md
Conduitry: Sounds good. Done in 0185aaae7e8460da5aedb3b64d8a598fda47e223.

Rich-Harris (approved):

Issue #1767, opened 2021-06-27
HTML validation error for generated `<script type="svelte-data" url="/data.json">`


kvn-shn:
**Describe the bug**
This is a minor complaint/bug but when validating my SvelteKit website with https://validator.w3.org/ I noticed that pages with a [`fetch()` in the `load()` function](https://kit.svelte.dev/docs#loading) create incorrect HTML.

**Logs**
W3 Validator's output:
> Error: Bad value svelte-data for attribute type on element script: Subtype missing.
> From line 239, column 2; to line 239, column 70
> [...]>↩↩↩↩	<script type="svelte-data" url="/my-endpoint.json">{"[...]
> Error: Attribute url not allowed on element script at this point.
> From line 239, column 2; to line 239, column 70
> [...]>↩↩↩↩	<script type="svelte-data" url="/my-endpoint.json">{"[...]

**To Reproduce**
1. Create a page with a [`fetch()` in its `load()` function](https://kit.svelte.dev/docs#loading), e.g.:
    ```svelte
    <script lang="ts" context="module">
        import type { LoadInput, LoadOutput } from '@sveltejs/kit/types/page';

        export async function load({ fetch }: LoadInput): Promise<LoadOutput> {
            const url = `/my-endpoint.json`;
            const res = await fetch(url);

            if (res.ok) {
                return { props: { ...(await res.json()) } };
            }
        }
    </script>
    ```
2. Build it and then validate it with https://validator.w3.org/
3. If nothing else is wrong, only the two errors from above should appear.

**Expected behavior**
I'm pretty satisfied that there are no other issues (e.g. https://vuejs.org has 8 and https://reactjs.org/ has 3 errors). So, to position SvelteKit in the "we do it better" area, I'd expect to have no validation errors/warnings at all.

**Information about your SvelteKit Installation:**
* `"@sveltejs/adapter-netlify": "1.0.0-next.14"`
* `"@sveltejs/kit": "1.0.0-next.109"`

**Severity**
Nothing is broken, so not that high. Still, it itches my *this can be done better* nerve. ;) (Especially since I believe the fix would be pretty simple, no?!)

**Additional context**
Suggestion: Wouldn't it already suffice to replace SvelteKit's special script attributes? I'm thinking of
* `type` -> `data-type`
* `url` -> `data-url`

Edit: I decided to make a quick PR doing just that, see https://github.com/sveltejs/kit/pull/1768

PR #1768, opened 2021-06-27
fix: attribute validation for generated script tag

kvn-shn:
fixes https://github.com/sveltejs/kit/issues/1767

Strictly speaking the attributes `type`, `url` and `body` in the generated script tag are invalid. This PR replaces them with `data-type`, `data-url` and `data-body`.

### Tests
Adjusted the existing test cases.

### Changesets
Generated a minor changeset.

### DIFF
--- a/.changeset/olive-maps-join.md
+++ b/.changeset/olive-maps-join.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': minor
+---
+
+fix attribute validation in generated script tag
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -159,9 +159,10 @@ export async function render_response({
 
 			${serialized_data
 				.map(({ url, body, json }) => {
-					return body
-						? `<script type="svelte-data" url="${url}" body="${hash(body)}">${json}</script>`
-						: `<script type="svelte-data" url="${url}">${json}</script>`;
+					const attrTag = 'data-type="svelte-data"';
+					const attrUrl = `data-url="${url}"`;
+					const attrBody = body ? `data-body="${hash(body)}"` : '';
+					return `<script type="application/json" ${attrTag} ${attrUrl} ${attrBody}>${json}</script>`;
 				})
 				.join('\n\n\t\t\t')}
 		`.replace(/^\t{2}/gm, '');
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -159,10 +159,10 @@ export async function render_response({
 
 			${serialized_data
 				.map(({ url, body, json }) => {
-					const attrTag = 'data-type="svelte-data"';
-					const attrUrl = `data-url="${url}"`;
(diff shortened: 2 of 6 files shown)


Conduitry:
This could potentially be a lot of data, so we're trying to avoid having the browser parse this as javascript. Are there any `type=` values that the validator won't complain about but which the browser won't parse and execute as javascript?

kvn-shn:
> This could potentially be a lot of data, so we're trying to avoid having the browser parse this as javascript. Are there any `type=` values that the validator won't complain about but which the browser won't parse and execute as javascript?

Okay, I understand. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-type) every MIME type other than JS should do the trick. I chose `text/plain` and adjusted the PR accordingly.

(Furthermore I noticed that there is also the invalid `body` attribute used. I renamed it to `data-body`  as well.)

packages/kit/src/runtime/server/page/render.js
```diff
@@ -159,9 +159,10 @@ export async function render_response({
 
 			${serialized_data
 				.map(({ url, body, json }) => {
-					return body
-						? `<script type="svelte-data" url="${url}" body="${hash(body)}">${json}</script>`
-						: `<script type="svelte-data" url="${url}">${json}</script>`;
+					const attrTag = 'data-type="svelte-data"';
```
benmccann, line 162: you could just inline these first two variables. for the third I would try to keep the extra space from making it to the output even if it makes it a bit uglier

```
const attrBody = body ? ` data-body="${hash(body)}"` : '';
return `<script type="application/json" data-type="svelte-data" data-url="${url}"${attrBody}>${json}</script>`;
```

.changeset/olive-maps-join.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': minor
+---
+
+fix incorrect attributes in generated script tag
```
benmccann, line 5:
```suggestion
fix attribute validation in generated script tag
```

packages/kit/src/runtime/server/page/render.js
kvn-shn: I tried to prioritize readability vs terseness. imho this is cleaner especially since Prettier would otherwise add a very awkward linebreak, see https://github.com/sveltejs/kit/blob/a604f8c2369055b75aabef0029b2d0fbeb16385a/packages/kit/src/runtime/server/page/render.js#L163-L165 But if you insist I'll do it the other way.

Rich-Harris:
thanks!

.changeset/olive-maps-join.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': minor
```
benmccann, line 2: this should have been `patch`

kvn-shn: thanks for the info. :) btw, this is not clear reading from the readme: https://github.com/sveltejs/kit#changelogs

benmccann: Added!

PR #1771, opened 2021-06-27
[feat] Add configurable cache to worker responses

lukaswelinder:
### New breaking change here

- **Who does this affect**: Anyone using the cloudflare adapter.
- **How to migrate**: Users that want to maintain the existing behavior will have to set all cache options to `null`. Otherwise defaults will apply. Any static assets without file extensions will break.
- **Why make this breaking change**: The existing implementation of the cloudflare worker adapter has slow page response and does not return cache-control headers for static assets like the default does, so browser cache is not utilized for static assets.
- **Severity (number of people affected x effort)**: Fairly low impact. Very simple to revert to original behavior.

I have been using svelte-kit to rebuild my personal site/blog and noticed a few areas for improvement.

### Slow page response

The problem here is having to wait for asset fetch to fail before moving on to try to render. The solution I put in place is more of a stop-gap. It is just checking for a file extension at the end of the request path.

By adding this gate, page request time dropped by 60-80%. However, static assets without file extensions will break. I'd like to improve upon my approach and would appreciate feedback here.

### Static assets are not returned with cache-control header

The default svelte-kit adapter returns this header so that browser cache can be utilized.

Refer to the updated README documenting the properties and their purpose. 

### Testing

I did not see tests in any of the adapters, don't mind writing some, but would appreciate input here.

That aside. I'm currently using this adapter code on a production site without issues.

### DIFF
--- a/packages/adapter-cloudflare-workers/README.md
+++ b/packages/adapter-cloudflare-workers/README.md
@@ -22,6 +22,13 @@ bucket = "./build"
 entry-point = "./workers-site"
 ```
 
+The adapter accepts the following configuration properties:
+| property | default | description |
+|---|---|---|
+| `pageCacheTTL` | `null` | Sets `cache-control` `max-age` for rendered responses if not defined. WARNING: this will apply to all page/request routes, disabled by default for security reasons. |
+| `staticCacheTTL` | `60 * 60 * 24 * 90` (90 days) | Sets `cache-control` `max-age` for static assets. |
+| `edgeCacheTTL` | `60 * 60 * 24 * 90` (90 days) | Sets CloudFlare Edge Cache TTL for static assets. |
+
 It's recommended that you add the `build` and `workers-site` folders (or whichever other folders you specify) to your `.gitignore`.
 
 More info on configuring a cloudflare worker site can be found [here](https://developers.cloudflare.com/workers/platform/sites/start-from-existing)
--- a/packages/adapter-cloudflare-workers/files/entry.js
+++ b/packages/adapter-cloudflare-workers/files/entry.js
@@ -23,8 +23,10 @@ async function handle(event) {
 			// use the asset manifest to see if it exists
 			return await getAssetFromKV(event, {
 				cacheControl: {
+					// eslint-disable-next-line no-undef
 					browserTTL: STATIC_CACHE_TTL,
-					edgeTTL: EDGE_CACHE_TTL,
+					// eslint-disable-next-line no-undef
+					edgeTTL: EDGE_CACHE_TTL
 				}
 			});
 		} catch (e) {
@@ -49,7 +51,9 @@ async function handle(event) {
 		if (rendered) {
 			const { headers } = rendered;
 			// inject cache-control header
+			// eslint-disable-next-line no-undef
 			if (!!PAGE_CACHE_TTL && !(headers['Cache-Control'] || headers['cache-control'])) {
+				// eslint-disable-next-line no-undef
 				rendered.headers['cache-control'] = `max-age=${PAGE_CACHE_TTL}`;
 			}
 			return new Response(rendered.body, {
--- a/packages/adapter-cloudflare-workers/files/entry.js
+++ b/packages/adapter-cloudflare-workers/files/entry.js
@@ -7,12 +7,26 @@ addEventListener('fetch', (event) => {
 });
 
 async function handle(event) {
+	// fall back to an app route
+	const request = event.request;
+	const request_url = new URL(request.url);
+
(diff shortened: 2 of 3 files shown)


lukaswelinder:
I'd like to point out a couple things regarding the current implementation of detecting static asset requests.

My implementation [here](https://github.com/sveltejs/kit/pull/1771/commits/4c8fa3449ffb27156b61e6f4917ab41b412bbd27) just checks for GET requests with a file extension. This means a couple things:

* Files without extensions will be broken (as mentioned above). Ideally you shouldn't do this in the first place, but I digress.
* Request routes with `.json` extension will still experience latency. Not ideal, but we're still in a better place overall.

Per the TODO comment, I investigated using the manifest to identify requests for static assets, but found this to be problematic. Perhaps it is an issue with my individual build tooling or configuration, but the manifest only includes js/css assets.

I think a better approach would be to recursively scan the `_app` output directory. Maybe there is already a way to do this via the adapter API, but I'm just not familiar enough with it. I can explore this further when I have spare time later this week, but input is welcome.

benmccann:
I think we should look for `_app` rather than file extension. Take a look at https://github.com/sveltejs/kit/pull/1416 which does that for `adapter-node`

lukaswelinder:
@benmccann Agreed. I can implement this over the weekend so that we can get this in.

I have a question about another feature enhancement though. I'd like to add support for a cloudflare worker hook. Should I implement that in this PR or would you prefer that I open another once this has been merged? Thanks.

tylerbrostrom:
With regards to detecting static assets, couldn’t this be a use-case for setting a non-empty `config.kit.paths.assets`? Requests starting with `${config.kit.paths.base}/${config.kit.paths.assets}` could be safely assumed to be an immutable (i.e. content-hashed) static asset.

Regarding assets from the `static` folder, that information currently isn’t exposed to adapters, but I believe this makes a strong case for passing a `BuildData` object to adapters’ `adapt` method. That would unlock the ability to do set fine-grained caching rules (e.g. different cache-controls for different files in the `static` folder) for this adapter and those implementing their own adapters.

benmccann:
>  I'd like to add support for a cloudflare worker hook. Should I implement that in this PR or would you prefer that I open another once this has been merged? 

It's probably easier to review as separate PRs

Rich-Harris:
Thank you!

Pages probably shouldn't have their own option here. There's already a place to set headers globally, and that's [`handle`](https://kit.svelte.dev/docs#hooks-handle). I don't think we should have another Cloudflare-specific way to do that.

The regex stuff is definitely a no-go :) Could you remove that so that this PR is purely about caching? The way to check if a request is for an asset would be to check the path against a list of them. We have such a thing here... https://github.com/sveltejs/kit/blob/3e30a1b7d3f47f1c37221e21be50819a06fe5ead/packages/kit/src/core/build/index.js#L74 ...we'd just need to figure out the right way to expose it along with a) the list of generated `_app` files and b) the list of prerendered pages, if any. That could happen in a separate PR. 

Speaking of a), since those generated files are hashed, they can be treated as immutable, and should probably automatically get a `public, max-age=31536000, immutable` cache-control header. 

Speaking of b), we might need a way to differentiate between prerendered pages and other assets in the KV store. (We might want to distinguish between different assets altogether — e.g. a longer TTL for `/videos/*` than `/other-stuff/*`, but that's probably overkill.)

So in summary, perhaps an API like this makes sense?

```js
import adapter from '@sveltejs/adapter-cloudflare-workers';

export default {
  kit: {
    adapter: adapter({
      browserTTL: {
        prerendered: 60,
        static: 24 * 60 * 60
      },
      edgeTTL: {
        prerendered: 60,
        static: 24 * 60 * 60
      }
    })
  }
};
```

tylerbrostrom:
Related: #1826

Defman:
I'm happy to take a stab at this.

I'm going to simply `export` the `manifest` const from the `../output/server/app.js` file and add `_app`, which will contain the same content as `client` from
https://github.com/sveltejs/kit/blob/3e30a1b7d3f47f1c37221e21be50819a06fe5ead/packages/kit/src/core/build/index.js#L72

`_app` field should ofc be a set.

```js
import { init, render, manifest } from '../output/server/app.js';

if (manifest._app[path])
    // Should cache immutable
else if (manifest.assets[path])
    // Should cache but not immutable
...
```

I'm not sure how I should handle prerendered pages.

Rich-Harris:
Reflecting further on this, many adapters can and should set `cache-control` headers for assets, so there should be a universal way to configure it and an adapter API for using that configuration. I've opened #3194, so I'll close this — thanks

PR #1778, opened 2021-06-29
Improve RequestHandler and EndpointOutput type declarations

dannyvelas:
If I wanted my `post` function to return a `number` as a response body and I accidentally return a `string`, there is no way I can let the type-checker prevent situations like these. There is no way I can tell it "hey, make sure the body property is always a number". So, the following typechecks:
```
export const post: RequestHandler<Locals, MyBody> = async (request) => {
  return {
    headers: { "set-cookie": "jwt=token; Path=/; HttpOnly" },
    body: "wrong type"
  };
};
```

This PR changes `RequestHandler` and `EndpointOutput` so that users can specify their response body type as `number` by making `number` the third generic argument to `RequestHandler`. As a result, this won't typecheck:
```
export const post: RequestHandler<Locals, MyReqBody, number> = async (request) => {
  return {
    headers: { "set-cookie": "jwt=token; Path=/; HttpOnly" },
    body: "wrong type",
  };
};
```
I wanted to make the second generic parameter to `RequestHander`, `ReqBody` be of default type `DefaultBody`, but this causes an error in `packages/kit/src/runtime/server/endpoint.js` on line 27, because the `request.body: BaseBody` is not type compatible with `BaseBody & DefaultBody`, which is the type it would end up with (per the type definitions of `ServerRequest` and `ParametrizedBody`).

There might be a better way to avoid the error and give `ReqBody` stronger typing than unknown, but I don't currently know enough about the codebase to do so.

Background for this PR: In the discord channel `contribute-to-svelte-kit`, I proposed this idea a few messages from the last.

### Tests

### Changesets

### DIFF
--- a/packages/kit/types/endpoint.d.ts
+++ b/packages/kit/types/endpoint.d.ts
@@ -12,14 +12,14 @@ type JSONValue =
 
 type DefaultBody = JSONValue | Uint8Array;
 
-export type EndpointOutput<ResBody = DefaultBody> = {
+export type EndpointOutput<Body extends DefaultBody = DefaultBody> = {
 	status?: number;
 	headers?: Partial<Headers>;
-	body?: ResBody;
+	body?: Body;
 };
 
 export type RequestHandler<
 	Locals = Record<string, any>,
-	ReqBody = unknown,
-	ResBody = DefaultBody
-> = (request: ServerRequest<Locals, ReqBody>) => MaybePromise<void | EndpointOutput<ResBody>>;
+	Input = unknown,
+	Output extends DefaultBody = DefaultBody
+> = (request: ServerRequest<Locals, Input>) => MaybePromise<void | EndpointOutput<Output>>;
--- a/packages/kit/types/endpoint.d.ts
+++ b/packages/kit/types/endpoint.d.ts
@@ -10,12 +10,16 @@ type JSONValue =
 	| JSONValue[]
 	| { [key: string]: JSONValue };
 
-export type EndpointOutput = {
+type DefaultBody = JSONValue | Uint8Array;
+
+export type EndpointOutput<ResBody = DefaultBody> = {
 	status?: number;
 	headers?: Partial<Headers>;
-	body?: JSONValue | Uint8Array;
+	body?: ResBody;
 };
 
-export type RequestHandler<Locals = Record<string, any>, Body = unknown> = (
-	request: ServerRequest<Locals, Body>
-) => MaybePromise<void | EndpointOutput>;
+export type RequestHandler<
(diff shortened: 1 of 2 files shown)


packages/kit/types/endpoint.d.ts
```diff
@@ -10,12 +10,16 @@ type JSONValue =
 	| JSONValue[]
 	| { [key: string]: JSONValue };
 
-export type EndpointOutput = {
+type DefaultBody = JSONValue | Uint8Array;
+
+export type EndpointOutput<ResBody = DefaultBody> = {
```
ignatiusmb, line 15: This should `extends DefaultBody` as well to make sure it's still in scope

packages/kit/types/endpoint.d.ts
```diff
@@ -10,12 +10,16 @@ type JSONValue =
...
 	status?: number;
 	headers?: Partial<Headers>;
-	body?: JSONValue | Uint8Array;
+	body?: ResBody;
```
ignatiusmb, line 18: Since there's no other body, we can probably just name it `Body`

packages/kit/types/endpoint.d.ts
```diff
@@ -10,12 +10,16 @@ type JSONValue =
...
 };
 
-export type RequestHandler<Locals = Record<string, any>, Body = unknown> = (
-	request: ServerRequest<Locals, Body>
-) => MaybePromise<void | EndpointOutput>;
+export type RequestHandler<
+	Locals = Record<string, any>,
+	ReqBody = unknown,
+	ResBody = DefaultBody
```
ignatiusmb, line 24: Line 24 should extends as well

Also, `ReqBody` does gets passed to `ServerRequest`, but `ResBody` to `EndpointOutput` isn't really a response is it. How about renaming them to just `Input` and `Output`?

ignatiusmb (commented):
This looks good, just a couple of comments. A changeset would also be nice.

dannyvelas:
Cool. I just made the changes:

1. Adding `extends DefaultBody` to the generic parameter of `EndpointOutput` and the last generic parameter of `RequestHandler`.
2. Changed generic type names `ReqBody` and `ResBody` to `Input` and `Output`, respectively, in `RequestHandler`.
3. Changed generic type name `ResBody` to `Body` in `EndpointOutput`.
4. Added patch changeset, `tall-hotels-punch`.

Let me know if I misunderstood any of your suggestions - I'd be happy to go back and amend any mistakes.

dummdidumm (approved):

ignatiusmb (approved):
Thanks! Everything else looks good to me.

Issue #1781, opened 2021-06-29
Ignore node_modules when walking files configured in kit.files.{routes,etc}


xrd:
## Description

If you use `{ kit: { files: { routes: '.' } } }` inside your `svelte.config.js`, kit processes files inside your node_modules directory. This means you cannot have your routes be inside the root directory.

For example, you will see this error if anything in the node_modules directory has a double underscore:

```
$ yarn dev
yarn run v1.22.10
warning package.json: No license field
$ svelte-kit dev
> Files and directories prefixed with __ are reserved (saw node_modules/postcss-selector-parser/dist/__tests__)
```

## envinfo

$ npx envinfo --system --npmPackages svelte,@sveltejs/kit,@sveltejs/adapter-node,@sveltejs/adapter-static,@sveltejs/adapter-begin,@sveltejs/adapter-netlify,@sveltejs/adapter-vercel vite --binaries --browsers
npx: installed 1 in 1.132s

  System:
    OS: Linux 5.11 Pop!_OS 20.10
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
    Memory: 11.39 GB / 62.66 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 14.17.0 - ~/.nvm/versions/node/v14.17.0/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v14.17.0/bin/yarn
    npm: 6.14.13 - ~/.nvm/versions/node/v14.17.0/bin/npm
  Browsers:
    Chrome: 91.0.4472.101
    Firefox: 88.0.1
  npmPackages:
    @sveltejs/adapter-static: ^1.0.0-next.13 => 1.0.0-next.13 
    @sveltejs/kit: next => 1.0.0-next.119 
    svelte: ^3.34.0 => 3.38.3 

## Browser

Firefox

## Severity

This does prevent me from using SvelteKit with my current repository where I want things all at the root.

Conduitry:
I don't know why `node_modules` would be the only thing we'd want to exclude - but more importantly, I don't think this is something we want to support or enable people to do. If you're pointing `routes` at a directory, everything in that directory should be routes (unless it's already being excluded by the `_` rules explicitly specified in the docs).

xrd:
[...]

PR #1782, opened 2021-06-29
Permit exclusion of arbitrary directory prefixes when using kit.files.{routes,etc.} in config

xrd:
Changes:

* Add function `ignoreFilePrefixes` which takes a comma-separated list of prefixes to ignore, and a filename; if the filename starts with any of those prefixes, return false so filter removes that item.
* Add tests to prove functionality
* Add documentation on usage of SVELTE_KIT_FILES_IGNORE_PREFIX variable.

This only works if you then run your build command with `SVELTE_KIT_FILES_IGNORE_PREFIXES=node_modules pnpm build` which will then filter out files starting with `node_modules` in the filename. If the variable is not present, the list of files is unchanged.

Re: https://github.com/sveltejs/kit/issues/1781

### DIFF
--- a/.changeset/dry-parrots-punch.md
+++ b/.changeset/dry-parrots-punch.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Ignore node_modules when configuring routes and lib paths, permitting paths at root of project
+Provide escape hatch to ignore node_modules (or any user defined prefixes) when configuring routes and lib paths, permitting paths at root of project
--- a/.changeset/dry-parrots-punch.md
+++ b/.changeset/dry-parrots-punch.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Ignore node_modules when configuring routes and lib paths, permitting paths at root of project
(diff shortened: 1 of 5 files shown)


Conduitry:
`(f) => !f.startsWith('node_modules')` isn't the right thing to be filtering by. And as mentioned in #1781, I don't think this is a feature we want anyway.

benmccann:
I haven't read the rest of the conversation or issue, but I don't understand why you'd use an environment variable for this instead of putting this piece of configuration in the config file where all the other configuration lives (if this is even a feature to be added)

xrd:
@benmccann Right now you can use `routes: '.'` in the config to specify the routes path as the root of your directory. Kit will then read all files in that directory, which includes node_modules. It looks to be impossible to configure node_modules with a different directory name (say `_node_modules` so that kit ignores it, please read this comment, it's so great: https://github.com/npm/npm/issues/775#issuecomment-68649430), there is no way to ignore files currently.

I get what you are saying about having this be something in the config file, but it seems like a much larger change to add that to the config file. Perhaps that is the right way. But, this is a quick escape hatch and if not present is ignored completely. And, the code to implement it (after several iterations), IMHO is very easy to read, whereas a new config option would require much more discussion.

What I want to do is permit using kit as a blog, and have the file structure identical to Jekyll. This means I want posts to be in the root directory, and I don't see another way to achieve that right now.

Rich-Harris:
Thanks but this definitely shouldn't be an environment variable. If this feature is added following further discussion on #1781 then it should be done as config

Issue #1784, opened 2021-06-30
ReferenceError: fetch is not defined


istarkov:
This [change](https://github.com/sveltejs/kit/commit/d729b0848f2f45586c5e19c7b689e6eb2eaee96e#diff-ed48923f0655f935bee505cdcbf40212129ff49a89c2c9d752bd63aa1e6282c4R191
) cause fetch is not defined error with sveltekit + node-adapter.

**Describe the bug**

Build svelte-kit project with node-adapter, 
execute with node v14.16.0, it fails with error "ReferenceError: fetch is not defined"

**Logs**
```
file:///root/realadvisor/netimmo.web/.build/index.js:1836
  serverFetch: hooks.serverFetch || fetch
                                    ^

ReferenceError: fetch is not defined
    at get_hooks (file:///root/realadvisor/netimmo.web/.build/index.js:1836:37)
    at init (file:///root/realadvisor/netimmo.web/.build/index.js:1681:12)
    at file:///root/realadvisor/netimmo.web/.build/index.js:1900:1
    at ModuleJob.run (internal/modules/esm/module_job.js:152:23)
    at async Loader.import (internal/modules/esm/loader.js:166:24)
    at async Object.loadESM (internal/process/esm_loader.js:68:5)
```

**To Reproduce**

```bash
npm init svelte@next my-app # use defaults
cd my-app
npm install @sveltejs/adapter-node@1.0.0-next.29
npm install
```

change svelte.config.js on

```javascript
import node from '@sveltejs/adapter-node';

/** @type {import('@sveltejs/kit').Config} */
const config = {
  kit: {
    // hydrate the <div id="svelte"> element in src/app.html
    target: '#svelte',
    adapter: node({
      out: '.build',
    }),
  },
};

export default config;
```

```bash
npm run build
node build
```

stalkerg:
This `fetch` was in packages/kit/src/runtime/server/page/load_node.js  but now in packages/kit/src/core/dev/index.js hmm.

stalkerg:
This file contain `import '../../install-fetch.js';` and inside we do `globalThis.fetch = fetch;` it's should be enough.

stalkerg:
ok, reproduced and I know how to fix it

vfilatov:
There is a work around while the hotfix above has not been committed yet.

src/hooks.js:
```javascript
/** @type {import('@sveltejs/kit').ServerFetch} */
export async function serverFetch(request) {
  /*
[...]

PR #1786, opened 2021-06-30
hotfix: with nodeAdapter we should get fetch from globalThis

stalkerg:
Base on #1784 issue. Basically, under real nodejs we have no direct access to `fetch` even if it's set to globalThis.

### DIFF
--- a/packages/kit/src/core/dev/index.js
+++ b/packages/kit/src/core/dev/index.js
@@ -215,7 +215,7 @@ class Watcher extends EventEmitter {
 							hooks: {
 								getSession: hooks.getSession || (() => ({})),
 								handle: hooks.handle || (({ request, resolve }) => resolve(request)),
-								serverFetch: hooks.serverFetch || fetch
+								serverFetch: hooks.serverFetch || globalThis.fetch
 							},
 							hydrate: this.config.kit.hydrate,
 							paths: this.config.kit.paths,


packages/kit/src/core/dev/index.js
```diff
@@ -215,7 +215,7 @@ class Watcher extends EventEmitter {
 							hooks: {
 								getSession: hooks.getSession || (() => ({})),
 								handle: hooks.handle || (({ request, resolve }) => resolve(request)),
-								serverFetch: hooks.serverFetch || fetch
+								serverFetch: hooks.serverFetch || globalThis.fetch
```
GrygrFlzr, line 218: What about the fetch call inside `serverFetch`? Does that also need to be scoped with `globalThis.`, or does this only affect the default case of not having `serverFetch` defined?

stalkerg: no, it's fallback, `serverFetch` is optional.

stalkerg: > or does this only affect the default case of not having serverFetch defined?

yes.

Clee681:
In case this is helpful, I'm seeing the following error in my CI pipeline:

```
file:///usr/src/app/build/index.js:1553
  serverFetch: hooks.serverFetch || fetch
                                    ^
ReferenceError: fetch is not defined
    at get_hooks (file:///usr/src/app/build/index.js:1553:37)
    at init (file:///usr/src/app/build/index.js:1513:12)
    at file:///usr/src/app/build/index.js:1575:1
    at ModuleJob.run (internal/modules/esm/module_job.js:169:25)
    at async Loader.import (internal/modules/esm/loader.js:177:24)
    at async Object.loadESM (internal/process/esm_loader.js:68:5)
```

For additional context, I'm using the node adapter (i.e. `adapter: node()`).

benmccann:
I see globalThis.fetch = fetch2; on line ~3k of the output, but there's an init function on line ~1.7k that ends up calling fetch, so I guess they've gotten put out of order somehow

It seems to be this init: https://github.com/sveltejs/kit/blob/6ace879521cff9125a5ee748103a710b8f558786/packages/kit/src/core/build/index.js#L290

I don't quite understand how all these build pieces fit together, but in my limited understanding this PR seems to be papering over the issue2, which is that `globalThis.fetch` is not being set early enough. I think we may need to somehow reorder how these lines are getting put in the final code. Perhaps @Rich-Harris or @Conduitry or someone who understands this code better than I do can chime in

benmccann:
I expect we may be hitting an incorrect assumption in esbuild: https://github.com/evanw/esbuild/issues/399#issuecomment-695967781

Rich-Harris:
Oof, that's a surprising issue. This does indeed seem to be an ordering issue — anything added to `globalThis` can be read as a global value:

```js
// foo.js
globalThis.foo = 42;
```

```js
// index.js
import './foo.js';
console.log(foo); // 42
```

Is this something that broke recently?

benmccann:
> Is this something that broke recently?

@Rich-Harris, yes, it started after https://github.com/sveltejs/kit/commit/d729b0848f2f45586c5e19c7b689e6eb2eaee96e#diff-ed48923f0655f935bee505cdcbf40212129ff49a89c2c9d752bd63aa1e6282c4R191

stalkerg:
@benmccann, yes, I totally agree, it's just hiding some order issue, but at the same time, it can be a temporal solution while we are trying to find and fix it. 
Anyway, up to you.

GrygrFlzr:
I just noticed, but turns out this PR actually changes the dev environment, not the build one, so it's actually modifying the wrong files to fix the issue.

Anyway, opened #1804 to try and deal with the root ordering issue rather than patch the symptoms.

stalkerg:
> I just noticed, but turns out this PR actually changes the dev environment, not the build one, so it's actually modifying the wrong files to fix the issue.

strange, at least I can't reproduce this issue anymore from #1784 after this PR.

Issue #1258, opened 2021-04-28
Inherit Vite modes behaviour via --mode CLI argument


schibrikov:
**Is your feature request related to a problem? Please describe.**
I want to have different environments/modes like staging/production/testing and pass some environment-related variables to the application.

**Describe the solution you'd like**
Vite already provides such functionality through "modes" and accepts --mode argument in CLI. Depending on the mode different .env files may be loaded and their contents passes as `import.meta.VITE_*` variables.
I'd like Sveltekit to inherit this behaviour and pass --mode arg from CLI to Vite, instead of only limiting to development/production as it is now.

docs: https://vitejs.dev/guide/env-and-mode.html

**Describe alternatives you've considered**

- Setting the mode in Vite section of svelte.config.js using env variables
- SvelteKit could provide it's own mechanism for loading environment-related variables
- Manually replacing .env.production file during build step on CI

**How important is this feature to you?**

Due to having some workarounds I don't think it's critical, but it would be convinient to use CLI arg instead of ENV vars + reading them in config.

petrkoutnycz:
Have you managed to make it work via first option of setting mode in Vite section? I'm still a bit struggling with it. Thank you.

jthegedus:
Until something like this gets adopted, the solution I have been using is:

```shell
# env.development - Vite.js defaults to "mode=development"
pnpm run dev

# env.staging
export MODE=staging && pnpm run build

# env.production - explicit override of svelte.config.js development default
export MODE=production && pnpm run build
```

```js
// svelte.config.js
import firebaseAdapter from 'svelte-adapter-firebase';

/** @type {import('@sveltejs/kit').Config} */
const config = {
	kit: {
		adapter: firebaseAdapter({ firebaseJson: '../firebase.json' }),
		target: '#svelte',
		vite: {
			// default to development as a guard
			mode: process.env.MODE || 'development',
		},
	},
};
```

benjamin-kirkbride:
Is there not a way to explicitly set the mode you want? If not, how do you specify which of a number of `env` files to use?

akvadrako:
You can also use `NODE_ENV`. When run with `svelte-kit dev`, `process.env.NODE_ENV == 'development'`.

Rich-Harris:
[...]

PR #1789, opened 2021-07-01
[fix] expose Vite.js `mode` from `$app/env`

jthegedus:

### Tests

### Changesets

---

**Summary**

`dev` utility is great, but projects often have more than two environments (`dev`, `prod`). Exposing the Vite.js [Mode](https://vitejs.dev/guide/env-and-mode.html#modes) value would simplify the code for these multi-envs projects.

- `mode` can be set in `config.kit.vite.mode`
- `svelte-kit dev` defaults this val to `development`
- `svelte-kit build` defaults this val to `production`

**Related**

This is related to https://github.com/sveltejs/kit/issues/1258 which asks for the CLI flag `--mode` to be exposed.

**Other**

Not sure how to write a test for this, maybe:

```js
import { dev, mode } from '$app/env';

if (dev && mode !== 'development') {
	throw Error('mode should equal development when dev is true');
}
```

I am unclear on how the test suite operates.

### DIFF
--- a/documentation/docs/05-modules.md
+++ b/documentation/docs/05-modules.md
@@ -13,7 +13,7 @@ import { amp, browser, dev, mode, prerendering } from '$app/env';
 - `amp` is `true` or `false` depending on the corresponding value in your [project configuration](#configuration)
 - `browser` is `true` or `false` depending on whether the app is running in the browser or on the server
 - `dev` is `true` in development mode, `false` in production
-- `mode` is `development` in development mode or `production` during build, unless otherwise configured in `config.kit.vite.mode`.
+- `mode` is the [Vite mode](https://vitejs.dev/guide/env-and-mode.html#modes), which is `development` in dev mode or `production` during build unless configured otherwise in `config.kit.vite.mode`.
 - `prerendering` is `true` when [prerendering](#ssr-and-javascript-prerender), `false` otherwise
 
 ### $app/navigation
--- a/documentation/docs/05-modules.md
+++ b/documentation/docs/05-modules.md
@@ -7,12 +7,13 @@ SvelteKit makes a number of modules available to your application.
 ### $app/env
 
 ```js
-import { amp, browser, dev, prerendering } from '$app/env';
+import { amp, browser, dev, mode, prerendering } from '$app/env';
 ```
 
 - `amp` is `true` or `false` depending on the corresponding value in your [project configuration](#configuration)
 - `browser` is `true` or `false` depending on whether the app is running in the browser or on the server
 - `dev` is `true` in development mode, `false` in production
+- `mode` is `development` in development mode or `production` during build, unless otherwise configured in `config.kit.vite.mode`.
 - `prerendering` is `true` when [prerendering](#ssr-and-javascript-prerender), `false` otherwise
 
 ### $app/navigation
--- a/packages/kit/types/ambient-modules.d.ts
+++ b/packages/kit/types/ambient-modules.d.ts
@@ -16,7 +16,9 @@ declare module '$app/env' {
 	 */
 	export const prerendering: boolean;
 	/**
-	 * The Vite.js mode the app is running in. Configure in `config.kit.vite.mode`. By default, `svelte-kit dev` runs in `development` mode and `svelte-kit build` runs in `production` mode.
+	 * The Vite.js mode the app is running in. Configure in `config.kit.vite.mode`.
+	 * Vite.js loads the dotenv file associated with the provided mode, `.env.[mode]` or `.env.[mode].local`.
+	 * By default, `svelte-kit dev` runs with `mode=development` and `svelte-kit build` runs with `mode=production`.
 	 */
 	export const mode: string;
 }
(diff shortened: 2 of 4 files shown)


packages/kit/types/ambient-modules.d.ts
```diff
@@ -15,6 +15,10 @@ declare module '$app/env' {
 	 * `true` when prerendering, `false` otherwise.
 	 */
 	export const prerendering: boolean;
+	/**
+	 * The Vite.js mode the app is running in. Configure in `config.kit.vite.mode`. By default, `svelte-kit dev` runs in `development` mode and `svelte-kit build` runs in `production` mode.
```
benmccann, line 19: it's not quite clear from these docs and the docs above for `dev` how the two are different from each other

Rich-Harris:
Thanks — could you add a changeset and update the docs please? https://github.com/sveltejs/kit/blob/master/documentation/docs/05-modules.md

packages/kit/types/ambient-modules.d.ts
Rich-Harris: It is technically redundant, but only because we'd be reflecting Vite's behaviour and Vite's behaviour is redundant https://vitejs.dev/guide/env-and-mode.html#env-variables. In practice I think it's beneficial to have a boolean `dev` value, and if people have found that it's valuable to expose `mode` then it makes sense to have both i think

jthegedus: @benmccann Yes, I should expand on these docs. Mode is used to infer `.env.[mode]` dotenv files to load. As these values are statically replaced during a build, we can utilise `mode` to load the config for our target env.

While this is similar to `dev` I'd argue not redundant. As I raised in https://github.com/sveltejs/kit/pull/1613#discussion_r642819593 there are many instances for building and running a SvelteKit app and in each instance I may want to use different **config** and conditionally execute code when `dev=false` in `prod` builds.

EG:
| command                                                           | purpose                                                                                           | envs                                     |
| ----------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- | ---------------------------------------- |
| `svelte-kit dev`                                                  | run **dev** server against **local** resources                                                    | `NODE_ENV=development`                   |
|                                                                   |                                                                                                   | `VITE_MODE=development` (default)        |
|                                                                   |                                                                                                   | `env.development`                        |
|                                                                   |                                                                                                   | `dev=true`                               |
|                                                                   |                                                                                                   |                                          |
| run **production** against **prod** resources                     | `svelte-kit build`                                                                                | `NODE_ENV=production`                    |
|                                                                   |                                                                                                   | `VITE_MODE=production` (default)         |
|                                                                   |                                                                                                   | `env.production`                         |
|                                                                   |                                                                                                   | `dev=false`                              |
|                                                                   |                                                                                                   |                                          |
| run **production** build locally against **local** resources      | `svelte-kit preview`                                                                              | `NODE_ENV=production`                    |
|                                                                   |                                                                                                   | `VITE_MODE=production` (default) :x:     |
|                                                                   | We want to load `env.development` here but default `VITE_MODE=production` was used during `build` | `env.production` :x:                     |
|                                                                   | not useful to resolve issue                                                                       | `dev=false`                              |
|                                                                   |                                                                                                   |                                          |
| build & deploy to **staging** against **staging** resources       | `svelte-kit build`                                                                                | `NODE_ENV=production`                    |
|                                                                   |                                                                                                   | `VITE_MODE=production` (default) :x:     |
|                                                                   | We want to load `env.staging` here but default `VITE_MODE=production` was used during build       | `env.production` :x:                     |
|                                                                   | not useful to resolve issue                                                                       | `dev=false`                              |
|                                                                   |                                                                                                   |                                          |
| build & deploy to **production** against **production** resources | `svelte-kit build`                                                                                | `NODE_ENV=production`                    |
|                                                                   |                                                                                                   | `VITE_MODE=production` (default) :check: |
|                                                                   |                                                                                                   | `env.production` :check:                 |
|                                                                   |                                                                                                   | `dev=false`                              |
|                                                                   |                                                                                                   |                                          |

benmccann: That's a very long answer I don't quite know how to digest :smile:  Basically my question is are these representing different values. E.g. one is `NODE_ENV` and one is `VITE_MODE`? We should document what underlying flag these each represent

jthegedus: Yes. One is essentially `NODE_ENV`, one is `VITE_MODE`. As far as I understand `dev` cannot represent all the states that `mode` can, and therefore is not redundant.

documentation/docs/05-modules.md
```diff
@@ -7,12 +7,13 @@ SvelteKit makes a number of modules available to your application.
...
 
 ```js
-import { amp, browser, dev, prerendering } from '$app/env';
+import { amp, browser, dev, mode, prerendering } from '$app/env';
 ```
 
 - `amp` is `true` or `false` depending on the corresponding value in your [project configuration](#configuration)
 - `browser` is `true` or `false` depending on whether the app is running in the browser or on the server
 - `dev` is `true` in development mode, `false` in production
+- `mode` is `development` in development mode or `production` during build, unless otherwise configured in `config.kit.vite.mode`.
```
benmccann, line 16:
```suggestion
- `mode` is the [Vite mode](https://vitejs.dev/guide/env-and-mode.html#modes), which is `development` in dev mode or `production` during build unless configured otherwise in `config.kit.vite.mode`.
```

PR #1791, opened 2021-07-01
Add generic type for session

seanlail:
Previously, in Sapper, we could use generics to ensure our session type was used correctly:

E.g. if I had a session type that was `{ foo: string; }`

```
import { stores } from "@sapper/app";

const { session } = stores<MySessionType>();

$session.foo; // correct usage
$session.bar; // invalid, compiler shows error
```

This PR adds the same generic support in `getStores` as well as a 3rd param in the generic for `LoadInput`

SvelteKit:
```
import { getStores } from "$app/stores";

const { session } = getStores<MySessionType>();
```

### Tests

### Changesets

### DIFF
--- a/packages/kit/types/page.d.ts
+++ b/packages/kit/types/page.d.ts
@@ -2,11 +2,12 @@ import { Location as Page, MaybePromise, InferValue } from './helper';
 
 export type LoadInput<
 	PageParams extends Record<string, string> = Record<string, string>,
-	Context extends Record<string, any> = Record<string, any>
+	Context extends Record<string, any> = Record<string, any>,
+	Session = any
 > = {
 	page: Page<PageParams>;
 	fetch: (info: RequestInfo, init?: RequestInit) => Promise<Response>;
-	session: any;
+	session: Session;
 	context: Context;
 };
 
--- a/packages/kit/types/page.d.ts
+++ b/packages/kit/types/page.d.ts
@@ -13,8 +13,9 @@ export type LoadInput<
 
 export type ErrorLoadInput<
 	PageParams extends Record<string, string> = Record<string, string>,
-	Context extends Record<string, any> = Record<string, any>
-> = LoadInput<PageParams, Context> & {
+	Context extends Record<string, any> = Record<string, any>,
+	Session = any
+> = LoadInput<PageParams, Context, Session> & {
 	status: number;
 	error: Error;
 };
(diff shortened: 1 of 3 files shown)


packages/kit/types/page.d.ts
```diff
@@ -2,11 +2,12 @@ import { Location as Page, MaybePromise, InferValue } from './helper';
 
 export type LoadInput<
 	PageParams extends Record<string, string> = Record<string, string>,
-	Context extends Record<string, any> = Record<string, any>
+	Context extends Record<string, any> = Record<string, any>,
+	Session = any
```
dummdidumm, line 6: You need to adjust the `Load` and `ErrorLoad` type accordingly

seanlail: @dummdidumm Thanks, will look at that shortly.

seanlail:
@dummdidumm GrygrFlzr pointed out on Discord that this PR doesn't help with this:
```
```

I wonder if there's anything we could do here?
I had a quick search and found this https://stackoverflow.com/questions/42986950/how-to-define-import-variable-type

i.e. update your own ambient type per project, something like this:
```
declare module "$app/stores" {
    const session : MySessionType;
    export = session;
}
```

If that's the only solution, perhaps it should be documented?

dummdidumm:
That seems like a useful addition, but I wouldn't add it to the site docs, only to the jsdoc in the `ambient-modules.d.ts` file

packages/kit/types/page.d.ts
seanlail: @dummdidumm updated Load and ErrorLoad 👍

seanlail:
Odd. I just tested this in the example project.

In a component:
```
<script lang="ts">
  import { session } from "$app/stores";
</script>

<div>{$session.fo}</div>
```

This doesn't work, VSCode thinks session in the component is `Writable<any>`
```

declare module "$app/stores" {
  import type { Writable } from "svelte/store";

  export const session: Writable<{ foo: string; bar: string }>;
}
```

However, if I remove the reference it all works and VSCode picks up the error.

Does this line actually need to be included?
```
```

dummdidumm:
Yes it needs to be included, else other imports from `$..` and `.svelte` imports in TS components are no longer recognized by TypeScript (giving false positives). My guess is that the broader type wins here, since both definitions are loaded, so this approach won't work. If people really want `session` to be a certain writable, I guess they have to do something like `export const myTypedSession: Writable<SomeInterface> = session`. But I wouldn't add this to the docs then.

dummdidumm (approved):

janosh:
I'm prob missing something. Why doesn't SvelteKit propagate the return type of `getSession` from `src/hooks.ts`?

I have

```ts
// src/hooks.ts

export function getSession(): Record<string, string> {
  // ...
  return session
}
```

yet when I import

```ts
import { session } from '$app/stores'
```

it has type

```ts
(alias) const session: Writable<any>
```

PR #1798, opened 2021-07-01
[fix] Custom HTTPS files results in error:0909006C:PEM routines:get_name:no start line

ryc76:
Specifying HTTPS key pair in Vite config fails causes dev server to fail

example config
```
server: {
		ssr: true,
		https: {
			key: '/privkey.pem',
			cert: '/fullchain.pem'
		}
	},
```

Run npx svelte-kit dev --https

Results in 

error:0909006C:PEM routines:get_name:no start line

Vite was being passed the path of the files rather the file themselves.

Odd as I'm sure this used to work.

### DIFF
--- a/packages/kit/src/core/server/index.js
+++ b/packages/kit/src/core/server/index.js
@@ -1,5 +1,6 @@
 import http from 'http';
 import https from 'https';
+import fs from 'fs';
 /**
  *
  * @param {boolean} use_https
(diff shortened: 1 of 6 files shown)


packages/kit/src/core/server/index.js
```diff
@@ -1,5 +1,6 @@
 import http from 'http';
 import https from 'https';
+import fs from 'fs';
```
benmccann, line 3: it'd probably be nice to add a blank line after this one for style consistency

benmccann:
This would need a changeset

It'd be nice to know if this was always broken or broke recently. Maybe a Vite upgrade broke it? I wonder if it works with an older version of Vite

Rich-Harris:
I'm not sure I follow — Vite isn't involved in this process, the code changed in this PR is SvelteKit's own helper for starting an http/https server. Did this work in a previous version? I'd be inclined to keep the current behaviour — it's easy to do `fs.readFileSync` inside your config file, but if it's expecting a filename then you can't pass a value directly

ryc76:
> 
> 
> This would need a changeset
> 
> It'd be nice to know if this was always broken or broke recently. Maybe a Vite upgrade broke it? I wonder if it works with an older version of Vite

I believe this commit broke it

https://github.com/sveltejs/kit/commit/5aa64ab5f7fb2a7e8b4c1a67891b565ae6d36699
https://github.com/sveltejs/kit/issues/844

benmccann:
> I believe this commit broke it

@ryc76 can you explain what the breakage was? I'm not seeing that the referenced PR did anything with either of those options

ryc76:
I've revisited and can't get this working with older commits as I thought I could. I'm happy to go with @Rich-Harris  suggestion of using fs.readFileSync inside the svelte.config.js

PR #1804, opened 2021-07-02
fix: Externalize app initialization to adapters

### LINKED ISSUE #1784: ReferenceError: fetch is not defined

GrygrFlzr:
Fixes #1784. Closes #1786 as its alternative.

This PR contains a **breaking change** to the adapter API.

## Current Behavior

https://github.com/sveltejs/kit/blob/559aad5f9dae6037a2d5fc4a63de0440634f41f2/packages/kit/src/core/build/index.js#L383

The built `app.js` file contains a side effect of calling the `init` function, which ends up running before polyfills like `@sveltejs/kit/install-fetch`. Although ESM (supposedly) guarantees a top-down order for import side effects, this is currently not necessarily true with esbuild nor rollup (see https://github.com/evanw/esbuild/issues/399#issuecomment-695967781).

## Proposed Behavior

https://github.com/sveltejs/kit/blob/e2f1b3f98124bb99606dcc0016e0072ed5e0e192/packages/kit/src/core/build/index.js#L288-L292

The `init` call is removed from the built `app.js`, and instead it becomes the responsibility of the adapters to call it at least once. This lets it control when the initialization happens, allowing us to guarantee that it happens after any polyfills:

https://github.com/sveltejs/kit/blob/e2f1b3f98124bb99606dcc0016e0072ed5e0e192/packages/adapter-node/src/index.js#L4-L8

This may also be useful in the future should the `init` function need to perform other side effects.

## Non-breaking alternate solutions

1. Use `require()` and the require shim to import the app to force it to run last:
    ```js
    const { render } = require('../output/server/app.js');
    ```
    This is not a great long-term solution considering we're forging ahead with ESM anyway.
2. Call `init` from `render` the first time it's called:
    ```js
    let initialized = false;
    export function render(/* ... */) {
      if (!initialized) {
        initialized = true;
        init(/* ... */);
      }
      // ...
    }
    ```
    In environments such as `adapter-node` it won't be evaluated until the server handles the quest, which may delay the initial response. May also affect serverless environments where lambdas may live longer than a single request.

### DIFF
--- a/packages/adapter-netlify/files/entry.js
+++ b/packages/adapter-netlify/files/entry.js
@@ -4,6 +4,7 @@ import '@sveltejs/kit/install-fetch'; // eslint-disable-line import/no-unresolve
 import { init, render } from '../output/server/app.js'; // eslint-disable-line import/no-unresolved
 
 init();
+
 export async function handler(event) {
 	const { path, httpMethod, headers, rawQuery, body, isBase64Encoded } = event;
 
--- a/packages/adapter-netlify/files/entry.js
+++ b/packages/adapter-netlify/files/entry.js
@@ -1,8 +1,9 @@
 import '@sveltejs/kit/install-fetch'; // eslint-disable-line import/no-unresolved
 
 // TODO hardcoding the relative location makes this brittle
-import { render } from '../output/server/app.js'; // eslint-disable-line import/no-unresolved
+import { init, render } from '../output/server/app.js'; // eslint-disable-line import/no-unresolved
 
+init();
 export async function handler(event) {
 	const { path, httpMethod, headers, rawQuery, body, isBase64Encoded } = event;
 
--- a/packages/adapter-node/src/require_shim.js
+++ b/packages/adapter-node/src/require_shim.js
@@ -1,2 +1,2 @@
 import { createRequire } from 'module';
-global.require = createRequire(import.meta.url);
+globalThis.require = createRequire(import.meta.url);
--- a/packages/kit/src/install-fetch.js
+++ b/packages/kit/src/install-fetch.js
@@ -1,7 +1,20 @@
-// @ts-nocheck
 import fetch, { Response, Request, Headers } from 'node-fetch';
 
-globalThis.fetch = fetch;
-globalThis.Response = Response;
-globalThis.Request = Request;
-globalThis.Headers = Headers;
+Object.defineProperties(globalThis, {
+    fetch: {
+        enumerable: true,
+        value: fetch
+    },
+    Response: {
+        enumerable: true,
+        value: Response
+    },
+    Request: {
+        enumerable: true,
+        value: Request
+    },
+    Headers: {
+        enumerable: true,
+        value: Headers
+    }
+});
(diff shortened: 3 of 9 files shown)


packages/kit/src/install-fetch.js
```diff
@@ -1,7 +1,20 @@
-// @ts-nocheck
 import fetch, { Response, Request, Headers } from 'node-fetch';
 
-globalThis.fetch = fetch;
-globalThis.Response = Response;
-globalThis.Request = Request;
-globalThis.Headers = Headers;
+Object.defineProperties(globalThis, {
```
GrygrFlzr, line 3: `defineProperties` seems like a more robust solution, and also avoids having to slap `ts-nocheck` directives.

packages/adapter-node/src/require_shim.js
```diff
@@ -1,2 +1,2 @@
 import { createRequire } from 'module';
-global.require = createRequire(import.meta.url);
+globalThis.require = createRequire(import.meta.url);
```
GrygrFlzr, line 2: Consistency with the rest of the codebase.

packages/adapter-netlify/files/entry.js
```diff
@@ -1,8 +1,9 @@
 import '@sveltejs/kit/install-fetch'; // eslint-disable-line import/no-unresolved
 
 // TODO hardcoding the relative location makes this brittle
-import { render } from '../output/server/app.js'; // eslint-disable-line import/no-unresolved
+import { init, render } from '../output/server/app.js'; // eslint-disable-line import/no-unresolved
 
+init();
```
benmccann, line 6: style nit: I might put a blank line after the call to `init();` in each adapter

benmccann (approved):
Thank you!

PR #1807, opened 2021-07-02
[docs] clarify and add warning for load function

### LINKED ISSUE #1542: Documentation should clarify that server state is not automatically per-request
**Describe the bug**

External variables, stores and other states processed during Load functions will keep in memory references that will remain upon all visitors of the website that is handled by the same server node instance.

This happens with "svelte-kit dev" "svelte-kit preview" and in production, at least tested in Netlify.

I think this issue comes from the assumption than node SSR is running clean on each instance.

**To Reproduce**
Use this basic demo repo
https://github.com/Egnus/sveltekit-bug-server-XSS-repro

run `npm run dev` and  go to `localhost:3000/todos`. There is a further red text block explaining what happens there

reload the page several times, and notice how the counter stays with `1` in browser but keeps growing in the Server, which is also accessible by the visitor by viewing source code or disabling Javascript so the hydratation doesn't fix the XSS issue

Use several browsers and incognito mode to find the same values shared for all users. An experienced hacker could get access to others data by accesing those values on memory.

**Expected behavior**
SSR runs clean in every call. **(ideal safe option)** forced by the sveltekit boot.
OR
Documentation could clearly states this issues and how to protect from XSS during Load.

This is happening with the latest version of SvelteKit

**Severity**
**Very severe since this is not mentioned in the documentation**: https://kit.svelte.dev/docs#loading-input-context
and also there is no clear mention of this memory between calls anywhere.

ignatiusmb:
Closes #1542


documentation/docs/03-loading.md
```diff
@@ -4,6 +4,10 @@ title: Loading
 
 A component that defines a page or a layout can export a `load` function that runs before the component is created. This function runs both during server-side rendering and in the client, and allows you to get data for a page without (for example) showing a loading spinner and fetching data in `onMount`.
 
+> Since `load` can run on the server as well as the client browser, any state changes to a variable that happens on the server will be saved and persisted, and all clients hitting the server-rendered page will be exposed to the already mutated state instead of a clean one.
```
benmccann, line 7: I think it's a bit confusing to start this with "Since `load` can run on the server as well as the client browser" because this really only involves the server and so bringing the client into it confused me. "saved and persisted" also makes me think it will persist the value to the client, which it doesn't

documentation/docs/03-loading.md
```diff
@@ -4,6 +4,10 @@ title: Loading
...
+>
+> It is recommended to make any state changes in the `<script>` block or outside of `load`. To share states exclusive to individual clients, use [`request.locals`](#hooks-handle) or [session](#loading-input-session).
```
benmccann, line 9: > It is recommended to make any state changes in the `<script>` block or outside of `load`.

I don't know that I'd go so far as to make this a recommendation

documentation/docs/03-loading.md
ignatiusmb: Initially thought so too, but was trying to preserve the first sentence from

https://github.com/sveltejs/kit/blob/64f749db8ca1e9bb08c38f271d9cd44b31920b83/documentation/docs/03-loading.md#L71

Do you think it would be clearer if we change it back to this?

```suggestion
> It is important to note that `load` can run on the server as well as the client browser, any state changes to a variable that happens on the server will be saved and persisted, and all clients hitting the server-rendered page will be exposed to the already mutated state instead of a clean one.
```

For "saved and persisted" part, I think it was clear with the "that happens on the server" placed right before it. Although, the "server and client" thing was already mentioned in the previous paragraph as well, perhaps we could just omit the sentence before the first comma altogether?

documentation/docs/03-loading.md
ignatiusmb: I was thinking on how to _not_ recommend making state changes in `load` function, but I think that it's better to use a positive connotation and recommend something instead of discouraging without giving alternatives. Any thoughts how we should go about it?

documentation/docs/03-loading.md
benmccann: Why not just leave that sentence where it was?

documentation/docs/03-loading.md
benmccann: I don't think there's any issue with making state changes in `load` though. It seems much too broad. You should just be aware if you mutate a singleton that it will persist across calls

benmccann:
I'm not sure any of this is particular to `load`, which makes it seem a bit odd to me to put it there. E.g. it's also true for endpoints that if you mutate a singleton that it will persist across calls. And sometimes you actually want to do that - it's how we recommend you setup a database https://kit.svelte.dev/faq#integrations

documentation/docs/03-loading.md
ignatiusmb: Feels weird to mention it multiple times throughout the same section, right now twice and would be 3 times if this were added. I do think we can still improve the whole section overall.

documentation/docs/03-loading.md
ignatiusmb: Isn't the issue is exactly that, can you explain why it's too broad? I'm not aware of any other stuff on the pages that can also run on the server other than `load`.

> You should just be aware if you mutate a singleton that it will persist across calls

I think the users are aware of this, but the underlying problem is that they expect it to be persisted across calls in the client's side and not from the server directly, and really the difference is just mutating in `load` and passing it to `<script>` first before mutating, to make sure that singleton is always mutated in the client's browser.

ignatiusmb:
That's true, but most of the problems occurs in `load` function and since we're writing it in the `#loading` section, I think it should be pretty clear that the warning only applies to mutations while loading. Endpoints are a different thing (and context) since they can only run in the server and never in client.

ignatiusmb:
I think this is really a common pitfall like what https://github.com/sveltejs/kit/issues/1542#issuecomment-875620599 says, especially for users using plain SPA Svelte coming to SvelteKit. Newcomers aren't really aware of the underlying issue when using `load` and would treat it as "preparing the data of the current page for client", thus the assumption that each call was clean. One could argue that users should've read the docs more carefully, but we could also make it better and have this notes highlighted.

This is also probably why we get a lot of requests to have a separate `load` that can exclusively run on either server or client, and also a lot of other issues specifically relating to just the `load` function.

documentation/docs/03-loading.md
benmccann: We wouldn't be mentioning it here if it were moved back though, would we?

documentation/docs/03-loading.md
benmccann: A state change is just mutating some variable. It seems too broad to say you can't do that

I was under the impression that the problem discussed in the issue (https://github.com/sveltejs/kit/issues/1542) was only talking about the server-side behavior. They were surprised that the counter was incrementing on the server between different requests and users

documentation/docs/03-loading.md
ignatiusmb: That's true, I've put them back and placed this at the bottom so it doesn't feel out of place. Users will read the note about the server/client first and we can avoid duplicate mentions.

documentation/docs/03-loading.md
ignatiusmb: It's not really saying they can't, but I guess it does sounds like there's implicit restrictions and may cause more confusion. Removed in the latest change.

documentation/docs/03-loading.md
```diff
@@ -29,6 +27,8 @@ type LoadOutput = {
 };
 ```
 
+Our example blog page might contain a `load` function like the following. Note the `context="module"` — this is necessary because `load` runs before the component is rendered:
```
benmccann, line 30: I agree with moving this. I actually made this change in a few places. If you rebase against `master` it was updated a couple weeks ago in https://github.com/sveltejs/kit/commit/9701eb21b7698226ad287c7ce0f1893ffa6fb972#diff-820e98511cb87c8891f398a00ff2f70793de22e53396d0a3ac6bc2cd6cc3e56e

documentation/docs/03-loading.md
```diff
@@ -58,14 +58,16 @@ type LoadOutput = {
...
->
-> - should use the SvelteKit-provided [`fetch`](#loading-input-fetch) method for getting data in order to avoid duplicate network requests
-> - should generally run on the same domain as any upstream API servers requiring credentials
-> - should not reference `window`, `document`, or any browser-specific objects
-> - should not reference any API keys or secrets directly, which will be exposed to the client, but instead call an endpoint using any required secrets
+> `load` only applies to [pages](#routing-pages) and [layouts](#layouts), and it may run on either the server or in the client browser.
+
+Code called inside `load` blocks:
+
+- should use the SvelteKit-provided [`fetch`](#loading-input-fetch) method for getting data in order to avoid duplicate network requests
```
benmccann, line 65: do we really want to take this out of the red box? this seems like the kind of thing we want to alert users to

documentation/docs/03-loading.md
```diff
@@ -58,14 +58,16 @@ type LoadOutput = {
...
+- should generally run on the same domain as any upstream API servers requiring credentials
+- should not reference `window`, `document`, or any browser-specific objects
+- should not reference any API keys or secrets directly, which will be exposed to the client, but instead call an endpoint using any required secrets
+
+> Any state changes to a variable that happens on the server will be saved and persisted on that server as well, and all clients hitting the server-rendered page will be exposed to the already mutated state from the server. To share states exclusive to individual clients, use [`request.locals`](#hooks-handle) or [session](#loading-input-session).
```
benmccann, line 70: "state changes to a variable" should probably be "state changes to a global variable" or something more specific

the reason to use `request.locals` is to share state across hooks, request, and page on the server. And `session` allows you to share between the server and client. I think that's a bit of info that could be useful in determining which is appropriate

documentation/docs/03-loading.md
```diff
@@ -58,14 +58,16 @@ type LoadOutput = {
...
 
-> `load` only applies to components that define pages, not the components that they import.
->
-> It is important to note that `load` may run on either the server or in the client browser. Code called inside `load` blocks:
->
-> - should use the SvelteKit-provided [`fetch`](#loading-input-fetch) method for getting data in order to avoid duplicate network requests
-> - should generally run on the same domain as any upstream API servers requiring credentials
-> - should not reference `window`, `document`, or any browser-specific objects
-> - should not reference any API keys or secrets directly, which will be exposed to the client, but instead call an endpoint using any required secrets
+> `load` only applies to [pages](#routing-pages) and [layouts](#layouts), and it may run on either the server or in the client browser.
```
benmccann, line 61:
```suggestion
> `load` only applies to [page](#routing-pages) and [layout](#layouts) components, not the components that they import, and by default runs on both the server and the client browser.
```

documentation/docs/03-loading.md
ignatiusmb: Leaving it inside will make it one huge red box, and when everything is an alert, nothing is. Taking this list out of it will make it so that it's placed between two red box and reads better for list items.

documentation/docs/03-loading.md
ignatiusmb: I agree with the first one. The additional bit of info should probably be placed in `locals` and `session` sections respectively instead, since we've already linked them here.

documentation/docs/03-loading.md
ignatiusmb: > and by default runs on both the server and the client browser

Not quite sure about this bit, it can only run on either server or client but not really both at once. I think the previous one already describes it pretty nicely.

documentation/docs/03-loading.md
benmccann: One thing that I think could be a bit confusing about the current wording in this PR is that it talks about sharing state with clients and mentions `request.locals`, but `request.locals` only stays on the server and client typically refers to the web browser

documentation/docs/03-loading.md
benmccann: It doesn't run on both at the same exact instant, but it does run on both for a single request

documentation/docs/03-loading.md
benmccann: The box currently (https://kit.svelte.dev/docs#loading) is just on the edge of what I'd say is the biggest we want to go. I like that you combined the first couple sentences, which will make it smaller. I am worried that adding the new stuff would push it over the edge. But I would keep what's currently there instead of removing it and adding the new stuff

documentation/docs/03-loading.md
Rich-Harris:
```suggestion
> `load` only applies to [page](#routing-pages) and [layout](#layouts) components, and can run both on the server (unless [ssr](#ssr-and-javascript-ssr) is disabled) and in the browser.
```

documentation/docs/03-loading.md
```diff
@@ -67,14 +67,16 @@ Our example blog page might contain a `load` function like the following. Note t
...
+
+Code called inside `load` blocks:
+
+- should use the SvelteKit-provided [`fetch`](#loading-input-fetch) method for getting data in order to avoid duplicate network requests
+- should generally run on the same domain as any upstream API servers requiring credentials
+- should not reference `window`, `document`, or any browser-specific objects
+- should not reference any API keys or secrets directly, which will be exposed to the client, but instead call an endpoint using any required secrets
```
Rich-Harris, line 77:
```suggestion
- should not directly reference any API keys or secrets, which will be exposed to the client, but instead call an endpoint that uses any required secrets
```

documentation/docs/03-loading.md
Rich-Harris:
```suggestion
> Mutating any shared state on the server will affect all clients, not just the current one.
```

We probably shouldn't talk about `request.locals` here since `load` doesn't see them, it adds confusion

documentation/docs/03-loading.md
```diff
@@ -67,14 +67,16 @@ Our example blog page might contain a `load` function like the following. Note t
...
-> - should use the SvelteKit-provided [`fetch`](#loading-input-fetch) method for getting data in order to avoid duplicate network requests
-> - should generally run on the same domain as any upstream API servers requiring credentials
-> - should not reference `window`, `document`, or any browser-specific objects
-> - should not reference any API keys or secrets directly, which will be exposed to the client, but instead call an endpoint using any required secrets
+> `load` only applies to [pages](#routing-pages) and [layouts](#layouts), and it may run on either the server or in the client browser.
+
+Code called inside `load` blocks:
+
+- should use the SvelteKit-provided [`fetch`](#loading-input-fetch) method for getting data in order to avoid duplicate network requests
+- should generally run on the same domain as any upstream API servers requiring credentials
```
Rich-Harris, line 75: honestly i'm not totally sure what this is doing here. is this `load`-specific, or just a thing you have to think about when building web apps generally?

documentation/docs/03-loading.md
Rich-Harris: i realise this wording was already present, but the duplicate network requests thing is barely here nor there — the real reason to use the passed-in `fetch` is that nothing will work otherwise

```suggestion
- should use the SvelteKit-provided [`fetch`](#loading-input-fetch) wrapper rather than using the native `fetch`, which doesn't have access to cookies on the server and cannot make requests against the app's own endpoints
```

documentation/docs/03-loading.md
Rich-Harris:
```suggestion
```

Rich-Harris (changes requested):

Rich-Harris:
I realise this PR has got very involved for such a straightforward change, but this section of the docs needs some work anyway and this is as good a time as any, so I left some suggestions to that effect

ignatiusmb:
Updated.

FYI, the locals and session bit were from https://github.com/sveltejs/kit/issues/1542#issuecomment-847097268 which I thought might be worth mentioning, but if it creates more confusion then it's probably best not to.

Rich-Harris:
thanks!

PR #1808, opened 2021-07-02
[fix] handle undefined body on endpoint output

### LINKED ISSUE #1351: Error thrown when returning endpoint Response without body
**Describe the bug**
If an endpoint returns an object without the `body` key specified, an error is thrown. This only seems to happen with a 200 status code.

```javascript
// src/routes/example.json.js

export function get() {
	// does not work
	return {};
	// also does not work
	return { status: 200 };
	// works
	return { body: {} };
	// also works
	return { status: 404 };
}

```

Stack trace below.

**Expected behavior**
No error thrown, or at least a more descriptive one.

**Stacktraces**
<details>
  <summary>Stack trace</summary>

```
Cannot read property 'length' of undefined
TypeError: Cannot read property 'length' of undefined
    at hash (file:///[project_path]/node_modules/@sveltejs/kit/dist/ssr.js:1518:11)
    at render (file:///[project_path]/node_modules/@sveltejs/kit/dist/ssr.js:1483:26)
    at runMicrotasks (<anonymous>)
    at async Object.handle (/[project_path]/src/hooks.ts:18:14)
    at async respond (file:///[project_path]/node_modules/@sveltejs/kit/dist/ssr.js:1450:10)
    at async Immediate.<anonymous> (file:///[project_path]/node_modules/@sveltejs/kit/dist/chunks/index.js:3304:23)
```

</details>

Tested on `@sveltejs/kit@1.0.0-next.100`, `svelte@3.38.2`, `vite@2.2.4`.

**Severity**
This isn't too severe since it's very easy to return an empty object as the body, but the error message doesn't really help.

**Additional context**
Seems like it's trying to hash the body for the ETag header, but `body` is undefined.

ignatiusmb:
This will add an empty object as the default value to the response body.

An empty string was put in place as the first choice, but it would fail on `res.json()` where a status of 200 would usually expect. It was changed in favor of an empty object so it wouldn't throw on `res.json()` and could still be read by `res.text()` the same.

Fixes #1351


packages/kit/test/apps/basics/src/routes/endpoint-output/_tests.js
```diff
@@ -0,0 +1,26 @@
...
+/** @type {import('test').TestMaker} */
+export default function (test) {
+	test('not ok on void endpoint', null, async ({ fetch }) => {
+		const res = await fetch('/endpoint-output/empty', { method: 'DELETE' });
+		assert.equal(res.ok, false);
+	});
+	test('200 status on empty endpoint', null, async ({ fetch }) => {
+		const res = await fetch('/endpoint-output/empty');
+		assert.equal(res.status, 200);
+		assert.equal(await res.json(), {});
```
benmccann, line 12: can we also test what the other return types would show for this?

ignatiusmb: Do you mean other than `return {}` in the endpoints?

benmccann: I meant like `res.text()`, etc.

ignatiusmb: I used `res.text()` in the other tests. I think we only need to check for `.json()` and `.text()` as they're the most commonly used ones. Other than that, just making sure that it doesn't throw on other methods are probably good enough.

benmccann: I think it'd be nice to have a test checking `/endpoint-output/empty` with `.text()`. I don't see that one in this file. Not sure if it lives elsewhere already

ignatiusmb: That's true, but I don't think it's worth a separate test and `fetch` call. Response bodies can't be reused after it's been parsed and I think it's quite clear that `.text()` representation of `{}` would just be `'{}'`.

I'm just concerned that adding more `fetch` integration tests would make it even more flaky than it needs to be.

packages/kit/src/runtime/server/endpoint.js
```diff
@@ -33,7 +33,7 @@ export default async function render_route(request, route) {
 				);
 			}
 
-			let { status = 200, body, headers = {} } = response;
+			let { status = 200, body = {}, headers = {} } = response;
```
benmccann, line 36: I think we should only do this for JSON `content-type` and not for text.

Might be able to do that by leaving this as it was and tweaking lines 58 & 62 or something

ignatiusmb: That might be better, I was also considering the same recently

benmccann:
Nice. This one closes a 1.0 issue :smile:

PR #1130, opened 2021-04-19
i18n: localized / translated routes

pzerelles:
This is now just a part of what #1037 implemented and also in a different way. The goal was to make alternate routes for the same page configurable.

There is a new configuration property "alternateRoutes" in the "kit" section of svelte.config.cjs. It optionally takes a function that receives a page's Part[][] and returns Part[][][], so the routes can be multiplied and the segments in Part[][] changed. This can be used e.g. to translate the routes or add a locale prefix.

Since the router has all the routes on the client, there is an alternates() function in "$app/navigation" that will return all alternate paths of a path. This is useful for language choosers for example.

I also added a usage example to "examples/svelte-kit-demo". Currently no translation is happening. This can be dealt with using translation libraries like svelte-i18n or svelte-intl-precompile.

PR #1810, opened 2021-07-02
[feat] i18n: lang in route

dreitzner:
This implements the routing portion of #553 
Possible basis for #1274

This tries to emulate a similar behavior that is already in use in other JamStack frameworks (e.g. https://i18n.nuxtjs.org/ or https://nextjs.org/docs/advanced-features/i18n-routing)

### Implemented behavior
- new svelte option => kit.i18n => {defaultLocale, locales[] }
- svelte context module load function has optional lang param (to get CMS content, or strings from a dictionary)
- lang is injected into $app/page
- router gets props `defaultLocale` and `locales`
- router has new method `switchLocalePath`
- `switchLocalePath` is accessible through `$app/navigation`
- locales and defaultLocale are now availble throuch `$app/stores.i18n`

### Open issues

open issues that I can think of:

- [x]  tests need to be adpted
- [x] example project needs to be removed
- [x] docs are not written
- [x] packages/kit/src/core/config/options.js needs correction
- [ ] test adapters
  - [x] static
  - [ ] node
  - [ ] cloudflare
  - [ ] netlify
  - [ ] vercel
- [x] language switcher needs to be implemented
- [x] client side navigation lang injection
- [x] make locales accesible through `$app/stores.i18n` for generating language switchers
- [x] ~~do we need lang prefix for api routes too?~~
- [x] add `i18nRoute` function to prefix routes with current lang 
- [x] add changeset
- [x] fix `npm run check`


maelp:
Nice! Does it work already? Does it compile to static folders when using the static build?

dreitzner:
It works in dev environment, but adapters are not adapted yet

dreitzner:
After a little fix dev and static export work.

I guess the other adapters should also work 🤷‍♂️

packages/kit/src/runtime/client/start.js
```diff
@@ -24,8 +24,10 @@ import { set_paths } from '../paths.js';
  *     nodes: Array<Promise<import('types/internal').CSRComponent>>;
  *     page: import('types/page').Page;
  *   };
+ * 	 defaultLocale: string;
```
benmccann, line 27:
```suggestion
 *   defaultLocale: string;
```

cor:
Great work so far! Can't wait until this gets merged 😄

floratmin:
Thank you for the work, I hope that this will be available soon. I wonder if there will also a possibility to define localized route paths or exclude routes from some locales?

dreitzner:
> Thank you for the work, I hope that this will be available soon. I wonder if there will also a possibility to define localized route paths or exclude routes from some locales?

Could you elaborate with an example?
If I understand you correctly you would want something like this in svelte config
```
i18n: {
  localizedRoutes: [
    { route: '/onlyEn', lang: 'en' }
  ],
  excludedRoutes: [
    { route: '/noDeAndFr', langs: ['de', 'fr' ]}
  ]
}
```

pnpm-lock.yaml
```diff
@@ -50,6 +49,23 @@ importers:
...
+      cookie: ^0.4.1
+      svelte: ^3.34.0
+    dependencies:
+      '@fontsource/fira-mono': 4.2.2
+      '@lukeed/uuid': 2.0.0
+      cookie: 0.4.1
+    devDependencies:
+      '@sveltejs/adapter-static': link:../../packages/adapter-static
+      '@sveltejs/kit': link:../../packages/kit
+      svelte: 3.38.2
```
benmccann, line 67: The CI is complaining that you have a broken lockfile

benmccann:
It might help in reviewing this to have some documentation in the `documentation/docs` directory. It'd give a better understanding of how the feature is to be used

floratmin:
@dreitzner 
Yes kind of, I was thinking about something like:
```
const routes = {
 about: {
    en: 'about',
    de: 'ueber-uns',
  },
  impressum: {
    de: 'impressum',
  }
};
```

dominikg:
I'm not sure the comment section of a PR is the right place to discuss how route segment translation should work 
but here's my 2ct.

1) adding a potentially huge map in svelte.config.js is detrimental to it's readability. prefer an external file
2) mapping needs to work 2-way and must be deterministic / free of ambiguities 
3) keep the structure as simple as possible. both lang: 'aa' , langs: ['aa','bb'] would be confusing and make implementation more complex.

dreitzner:
> @dreitzner 
> Yes kind of, I was thinking about something like:
> ```
> const routes = {
>  about: {
>     en: 'about',
>     de: 'ueber-uns',
>   },
>   impressum: {
>     de: 'impressum',
>   }
> };
> ```

@floratmin that would be nice indeed, but I think it would be great to implement a basic structure first and then extend it at a later point.

dreitzner:
> It might help in reviewing this to have some documentation in the `documentation/docs` directory. It'd give a better understanding of how the feature is to be used

That would have been the next step, as soon as my time allows it (full-time job, three kids and a fourth almost here takes up some time 😂)

pnpm-lock.yaml
dreitzner: was part of the test project and is removed

dreitzner:
@benmccann i resolved all merge conflicts, added docs, removed the sample project, did some cleanup.

Is it ok, in you opinion to omit api routes from being i18n? If needed it can be an url param anyway, and if really needed, then it can be extended at a later point. If you do not agree, I can try to adapt the endpoints too, but I'd rather not 😉

Marr11317:
So from what I understand, no real redirection nor translation is happening here. The path is only changed to be prefixed with the current language, and the old URL is pushed to history. Is that it?

dreitzner:
> So from what I understand, no real redirection nor translation is happening here. The path is only changed to be prefixed with the current language, and the old URL is pushed to history. Is that it?

@Marr11317, yes, the "only" thing that happens is:
- extend code generation (routes, lang injection in the need places)
- utility functions to change language
- stores to access the data

Translations can build on top of this ticket. I'm not convinced that this can be generalized, because there are so many valid use cases (internal transition files in JSON, js, MD, ..., or many different CMS', fallback to default language, redirect to language, partial translation, ...). But it might be nice to have a suggested way available.

Marr11317:
Sorry for that "only", I didn’t mean it negatively. The work that has been done is amazing and is definitely a solid foundation for integrating a translation library.

dreitzner:
> Sorry for that "only", I didn’t mean it negatively. The work that has been done is amazing and is definitely a solid foundation for integrating a translation library. 

No worries, didn't understand it in a negative way. Thanks

dreitzner:
My fourth child was born today. I might not be as quick to respond in the next 3 weeks.

SpheeresX:
Kinda silly question, but is this ready for production (not really mission-critical) use? I can't wait for this to be merged.

dreitzner:
> Kinda silly question, but is this ready for production (not really mission-critical) use? I can't wait for this to be merged.

Kinda silly answer:

- yes it is fully working
- I am in the process of building a test project for adapters, but between my jobs and my family, it is not moving as fast as I'd like
- it would be great if some people could try and help with testing the adapters (and provide a link to the deployed test version)

benmccann:
Hi all, quick update on i18n. We just finished support for unicode URLs in `1.0.0-next.163`, which seemed like a prerequisite to i18n and was something I wanted to get done before looking at this. This was rather complicated because it involved changes to SvelteKit, Vite, sirv, and polka over multiple releases of each of those libraries. We should be unblocked now as a result.

However, that being said, we're prioritizing just getting things working as they are before adding new features like i18n. I understand that i18n is a blocking issue for many people and appreciate that this is an issue folks are willing to chip in on, so we'll try to balance that and not let this sit forever without some more feedback. No promises on when we might be able to get to this, but I did want to let everyone know we haven't forgotten about it!

We're also looking at ways to keep the project momentum going, so have added several new maintainers over the past year and are discussing starting to use the donations we've received to get some part-time help as well

benbender:
@dreitzner would you be able to share your test-project? If yes, I could have do a testdrive on node + vercel.

benmccann:
Is there anything this lets you do that you can't do today? I would expect you could just create a folder named `[lang]` and a store for the language and that would give you everything you need. You could update the URL when the languages changes with something like:
```
lang.subscribe(value => {
	goto(`/${value}${url.substring(url.indexOf('/', 1))}`);
});
```

dreitzner:
> Is there anything this lets you do that you can't do today? I would expect you could just create a folder named `[lang]` and a store for the language and that would give you everything you need. You could update the URL when the languages changes with something like:
> 
> ```
> lang.subscribe(value => {
> 	goto(`/${value}${url.substring(url.indexOf('/', 1))}`);
> });
> ```

The problem would be that ALL routes would have to be duplicated and kept in sync (base language and [lang]).

Also with adapter static, there would be a need to have a link to each language, which might complicate things.

Also for SSR with `context="module"` it might be a bit trickier to get the right language.

benmccann:
> The problem would be that ALL routes would have to be duplicated and kept in sync (base language and [lang]).

I don't understand what you mean by this. Are you referring to usage with a particular adapter or some more specific scenario? I was thinking you'd just move all your routes one level deeper. Create `src/[lang]` and then move all your routes into the `[lang]` folder. You wouldn't have to duplicate any routes

>Also with adapter static, there would be a need to have a link to each language, which might complicate things.

Sure, a link, button, dropdown or some way to choose the language. Wouldn't you need that with this PR as well?

>Also for SSR with context="module" it might be a bit trickier to get the right language.

How would it be tricky? You can just access `params.lang`

dreitzner:
@benmccann I hope I understood your comments correctly. 

> I don't understand what you mean by this. Are you referring to usage with a particular adapter or some more specific scenario? I was thinking you'd just move all your routes one level deeper. Create src/[lang] and then move all your routes into the [lang] folder. You wouldn't have to duplicate any routes

the usual behavior is that a default locale lives inside of `/`. If I move eveerything into `src/[lang]` I'll no longer have a base route without i18n prefix. 

> Sure, a link, button, dropdown or some way to choose the language. Wouldn't you need that with this PR as well?
my PR adds the pages to the manifest => static export should render all of them

dreitzner:
@benbender after I fixed some troubles after pulling upstream I created the test branch.

Thank you for offering to hel <3

https://github.com/dreitzner/kit/tree/553-test-branch/examples/i18n

benmccann:
> the usual behavior is that a default locale lives inside of /. If I move eveerything into src/[lang] I'll no longer have a base route without i18n prefix.

Typo on my part. I meant `src/routes/[lang]` and have updated the comment

Marr11317:
> 
> 
> > the usual behavior is that a default locale lives inside of /. If I move eveerything into src/[lang] I'll no longer have a base route without i18n prefix.
> 
> Typo on my part. I meant `src/routes/[lang]` and have updated the comment

I believe that what he meant is that if the base url is example.com, then:
```
example.com/index.html // main english page
example.com/fr/index.html //main french page
example.com/es/index.html // main spanish page
etc.
```
when no path is provided, it should default to the main english page.

dreitzner:
> > > the usual behavior is that a default locale lives inside of /. If I move eveerything into src/[lang] I'll no longer have a base route without i18n prefix.
> > 
> > 
> > Typo on my part. I meant `src/routes/[lang]` and have updated the comment
> 
> I believe that what he meant is that if the base url is example.com, then:
> 
> ```
> example.com/index.html // main english page
> example.com/fr/index.html //main french page
> example.com/es/index.html // main spanish page
> etc.
> ```
> 
> when no path is provided, it should default to the main english page.

Thank you for helping out explaining it.

This is a perfect example of what I wanted to convey.

benmccann:
Ah, so basically there needs to be a way to specify an optional route parameter. I think that might be a more generic solution than building language support directly into the routing

samuelstroschein:
> I believe that what he meant is that if the base url is example.com, then:
 > ```
 > example.com/index.html // main english page
 > example.com/fr/index.html //main french page
 > example.com/es/index.html // main spanish page
 > etc.
 > ```
 > 
 > 
>  when no path is provided, it should default to the main english page.

Wouldn't it be better to not commit the language parameter for the default language? I know most websites omit /en and so does NextJS but I think that is rooted in "we started with one language (English) and added other languages later". Having a language parameter for every language, including the default one (which often is not English btw), seems reasonable to me.   

The example above would become
 > example.com/en/index.html // main english page
 > example.com/fr/index.html //main french page
 > example.com/es/index.html // main spanish page

benbender:
Imho the question here is not only if $things can't be achieved in any other way in sveltekit, but more about providing common, basic building blocks for common tasks/ways/interfaces (like i18n) which help to build upon in a common way and to not "reinvent the wheel" over and over again in many projects by many developers. Imho that's a great way to lower the burden for devs and help them to concentrate on those business-logic-tasks they actually want to solve.

benmccann:
I agree we should make things like i18n easy to setup and use. However, if we can allow users to code it with just a few lines, that's often better than building it in. E.g. there's disagreement here about whether or not the default language should be included in the URL. We'd end up also needing an option to control that. And what if the user wants to put the language as a subdomain instead of in the route (e.g. like LinkedIn)? That would again require another option. What if instead of sub-domains the user has different domains like `google.com`, `google.co.uk`, `google.nl`? You end up with an explosion of functionality in core to handle every use case, which could have been much simpler if we just provided easy-to-use building blocks. I find this PR to be quite opinionated on a certain way to include the lang in the route, when there are multiple ways to do it

schindld:
@benmccann the two ways of doing it using subdomains you mentioned aren't really relevant here because a web proxy would have to route them to sub paths in the app anyway. I agree that the lang subpath should be optional, but otherwise sveltekit desperately needs something like this asap.

benmccann:
> a web proxy would have to route them to sub paths in the app anyway

That's one way you could implement it, but you could also have the application check the `Host` header

> sveltekit desperately needs something like this asap

You can already easily implement it in any app except for the optional routing segment

benbender:
Imho, nextjs got all the major usecases right on their try to handle this: https://nextjs.org/docs/advanced-features/i18n-routing. The usage-experience of their implementation is simply pleasant, as you don't notice it if you don't need it and simply can configure some options to get a working system for most basic usecases as a opt-in when you need it.

One alternative, if we would say we don't want it in core, as its easily achivable, could be, to have default-implementations for such kinds of problems inside a "recipe"-repo or similar. It would, at least, give the benefit of having default solutions to common problems without much overhead for the users...

benmccann:
> One alternative, if we would say we don't want it in core, as its easily achivable, could be, to have default-implementations for such kinds of problems inside a "recipe"-repo or similar. It would, at least, give the benefit of having default solutions to common problems without much overhead for the users...

The way we've been handling "recipe" type stuff is with Svelte Adders. It lets you setup all kinds of things like TailWind or Jest with a single command: https://github.com/svelte-add/svelte-add#-built-in-integration-adders

benbender:
@benmccann Yes and no. Don't think it serves the purpose on an practical level as it is not officially endorsed, referenced in the docs etc. It's there, but it's not the neccessarily the obvious solution. The default solution in the sense that you don't need to search for it or think about it, if you have a common problem.

What I had in mind is more something like the solution of blitzjs: https://github.com/blitz-js/blitz/tree/canary/recipes

benmccann:
It is referenced in the docs: https://kit.svelte.dev/faq#how-do-i-use-x-with-sveltekit-how-do-i-setup-library-x

They're also listed on https://sveltesociety.dev/templates which we point folks to for this type of info

Happy to have ideas on how to make that more discoverable if folks are missing it

benbender:
> Happy to have ideas on how to make that more discoverable if folks are missing it

I did a search for `svelte-adder` in this repo and had 0 results. This could obviously be improved. Additionally I'm unable to find any reference on https://kit.svelte.dev/docs or on https://github.com/svelte-add/svelte-add. If it's the endorsed solution, I would say its not obvious (at least not to me). And, ideally, it could be moved into this repo to make it obvious...

dreitzner:
I'll try to summarize:

- google.com/.co.uk/... is not a fair comparison, as they have many different servers that are bound to their specific URL. To have a server for each language would be possible today without any hassle
- whenever there is a REAL need to include domains/subdomains for i18n (because there is only one server that handles all), then lets talk about it then. Without a real usecase this would be an useless discussion.
- @benmccann what this PR offers is currently NOT possible with svelte-kit (I tried, but adapter static will not work as expected)
- the pattern implemented by this PR is well know and kinda expected with i18n. See NUXT, NEXT (they are what kit is for svelte) and many other frameworks
-  @benmccann an optional route parameter might be nice, but I am not sure how a static export will work, because the specific localized routes will no longer show up in the manifest
- I am not convinced if this can be solved with something like svelte adders, as there are no sufficient hooks into the build system. If there would be a way to write build plugins (that can hook into all the stages where it is needed), it might be a good idea to extract it into it's own package that can be added in. In that case I would still argue that this should be part of the kit project (as it's won package), as it is such a basic need (see original issue by Rich). 
- Having the current locale available in load and in the component is necessary, if you want to translate anything

I hope I didn't leave anything out :)

f-elix:
@dreitzner Thank you so much for working on this!

I would add one thing (which has already been mentioned earlier) to your summary: the ability to translate route params (eg: `/en/about` to `/fr/a-propos`. I think this is pretty essential.

If it helps, I am currently working on 2 different multilingual Sveltekit projects (in production). Here is a simplified version of the routes structure I use:

```
routes
  [locale]
    [...url].svelte
```

I can then grab the url param and query the data from my CMS with it. This works well but does have some limitations. I will use prerendering as an exemple here, but this applies to every situation where you need a different configuration from one page to another.

Say I have a mostly static site that I can prerender, but I also have an account page that is dynamic. I can only write `export prerender = true;` in my `[...url].svelte` file. If I want to turn off prerendering for my account page, I have to create a separate `account.svelte` file. This works, until I want to translate the slug (`/en/account` to `/fr/compte`). Now I have to duplicate not only the file (create `compte.svelte`), but also my `[locale]` folder. The file structure shown above then becomes:

```
routes
  en
    account.svelte
    [...url].svelte
  fr
    compte.svelte
    [...url].svelte
```

Even though both `[...url].svelte` and the account page files are the same in each locale folder, I have to duplicate them. It also becomes more work to add a new language, as I have to hardcode them in the file structure.

I don't know if this is possible, but would there be a way to only have one default version of my file structure (associated with the default language), and a config file where I can just specify the translated slugs of every non-dynamic routes. Something like:

```
routes
  [locale]
    account.svelte
    [...url].svelte
```

(This is just a rough exemple, this configuration format could be anything you'd like, as long as it allows translating routes)

```
{
  i18n: {
    localizedRoutes: [
      {
        slug: 'account',
        alternates: [
          {
            lang: 'fr',
            slug: 'compte'
          ]
        ]
      }
    ]
  }
}
```

dreitzner:
@f-elix I agree, that this would be an excellent feature to add on top of this PR.

I was thinking that it can be extended similar to what nuxt is doing with their i18n.

I am not sure which way it will go (page level config, svelte config), but nuxt offers them both.
My tendency would be to have this config on a page level, but there am some smart people out there who might think of some reason why something else might be better for kit 😊

benmccann:
> an optional route parameter might be nice, but I am not sure how a static export will work

Static export would work work fine with an optional route parameter. It's the same as every other dynamic parameter that we have today which all work today with static export. I'm not against adding support for i18n in core, but if we can do it in a more generic way that provides value for other use cases that seems preferable. I think supporting optional route segments would allow you to implement this feature in a more generic way rather than coding something this specific

> Having the current locale available in load and in the component is necessary, if you want to translate anything

I agree that you need a way to access the locale. Can you use a store for that?

dreitzner:
> > Having the current locale available in load and in the component is necessary, if you want to translate anything
> 
> I agree that you need a way to access the locale. Can you use a store for that?

A store works fine for the client (and should be pretty straight forward). But what about the load function in ssr when you want to fetch something from your CMS and need the lang. You would need to write some code just for that.

Concerning optional fall through routes: I think I did not communicate my use case very well concerning static export. Assume you have 10 languages (some requirement from a current project at work). If you want the crawler to know about any of those languages you would need to link to all of them, right? With this PR the manifest gets extended, which in turn allows you to crawl the site without cross linking it (with 10 languages this gets out of hand quite fast).

But if optional fall through routes would work, than this could be awesome.

benmccann:
Yes, adding a link is one option. You're quite likely to have a link somewhere in the application or else how would users change their language? You could automatically detect, but usually there's still a way to change. The other option would be to set the `prerender.entries` option: https://kit.svelte.dev/docs#configuration-prerender

lukaszpolowczyk:
What do you think about supporting something like Mozilla:
https://projectfluent.org/
?

It is about supporting variables, word variations and numeric variants for individual languages.

Marr11317:
> 
> 
> What do you think about supporting something like Mozilla: https://projectfluent.org/ ?
> 
> It is about supporting variables, word variations and numeric variants for individual languages.

I don't think this pr should make any move in regard to a "recommended library".

However, to answer you question, although having a steeper learning curve, fluent is a very nice project and its syntax makes it easy to read compared to others.

I realize that I forgot to post it here, so: 

> 
> I have been working on a template that implements a few of the features described in this thread.
> 
> [Have a look at it and tell me  what you think](https://github.com/Marr11317/sveltekit-i18n-template)

pz-mxu:
Am I right that this PR in general does multiply routes for a list of language prefixes automatically.

I also created a PR #1130 that adds a generic extension point to routes exactly for this, but is not limited to i18n. I have not looked into the code of your PR, but maybe our PRs are not so different. I took a more generic route so that i18n as a specialisation can be plugged in as a function. Translation of route segments also can be applied. Maybe we could join forces.

dreitzner:
Closed in favor of using (non-greedy rest route parameters)[https://kit.svelte.dev/docs/routing#advanced-routing-rest-parameters]

PR #1811, opened 2021-07-02
[fix] Handle 4xx and 5xx statuses without requiring `Error` instance

### LINKED ISSUE #1161: load function does not respect returned error status codes

damain:
This PR fixes #1161 

It checks if loaded.status is between 400 & 599  and if no loaded.error is sent it creates a message for each status and returns the status and message

It includes messages for the following statuses 400, 401, 403, 404, 409, 500
Any other statuses are defaulted to 500 

One failing test for a 401 status is included. 
If any other statuses or tests need to be added let me know and I'll get it done.


packages/kit/src/runtime/load.js
```diff
@@ -5,9 +5,156 @@
...
-		const error = typeof loaded.error === 'string' ? new Error(loaded.error) : loaded.error;
+	if (
+		loaded.error ||
+		(loaded.status && loaded.status >= 400 && loaded.status <= 599 && !loaded.redirect)
+	) {
 		const status = loaded.status;
+		if (!loaded.error) {
+			let errorMessage = 'not assigned';
+			switch (status) {
+				case 400:
```
benmccann, line 16: There was some discussion of whether we should include the `http-status-codes` library (https://github.com/sveltejs/kit/issues/1439). I didn't think it was worth including just to make it available to users. But if we're actually going to be using it in SvelteKit's own code, I might reconsider rather than having this list here. I'm not sure if @Conduitry or Rich have other thoughts about it

damain: In 1ccc6f33d25b56f0966a08bb2db67c749e898dea there are only these codes 400, 401, 403, 404, 409, 500 which are the most common. We could probably support this lesser list and either send a 500 or pass the correct status code with unspecified as the error text i.e. {status: 402, error:"Unspecified"}

packages/kit/src/runtime/load.js
```diff
@@ -5,9 +5,156 @@
...
+					errorMessage = 'Payment Required';
+					break;
+				case 403:
+					errorMessage = 'Forbidden';
+					break;
+				case 404:
+					errorMessage = 'Not Found';
+					break;
+				case 405:
+					errorMessage = 'Method not Allowed';
```
benmccann, line 32: should probably use a capital letter here to match the others
```suggestion
					errorMessage = 'Method Not Allowed';
```

packages/kit/src/runtime/load.js
```diff
@@ -5,9 +5,156 @@
...
+				case 429:
+					errorMessage = 'Too Many Requests';
+					break;
+				case 431:
+					errorMessage = 'Request Header Fields Too Large';
+					break;
+				case 444:
+					errorMessage = 'No Response';
+					break;
+				case 449:
```
benmccann, line 97: Some of the codes listed here such as this one are non-standard.

It actually seems invalid to send this `errorMessage`. It would need to include another header and is not a complete message
https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-wdv/1c1f4d5b-4fed-48ef-9bb6-893c2b25b74d
http://getstatuscode.com/449

packages/kit/src/runtime/load.js
```diff
@@ -5,9 +5,156 @@
...
+					errorMessage = 'Network Read Timeout Error';
+					break;
+				case 599:
+					errorMessage = 'Network Connect Timeout Error';
+					break;
+			}
+
+			let statusRecognized = errorMessage !== 'not assigned';
+			return {
+				status: statusRecognized ? status : 500,
```
benmccann, line 152: I don't think we should change the status code. We should just return it as is

```suggestion
				status,
```

packages/kit/src/runtime/load.js
```diff
@@ -5,9 +5,156 @@
...
 
-	if (loaded.error) {
-		const error = typeof loaded.error === 'string' ? new Error(loaded.error) : loaded.error;
+	if (
+		loaded.error ||
+		(loaded.status && loaded.status >= 400 && loaded.status <= 599 && !loaded.redirect)
+	) {
 		const status = loaded.status;
+		if (!loaded.error) {
+			let errorMessage = 'not assigned';
```
benmccann, line 14: There's no reason to initialize this to a string that's never used. I would just let it be undefined by default
```suggestion
			let errorMessage;
```

benmccann:
Let's remove all the default error messages (which means you can ignore most of my comments as they'll no longer be relevant). It will make the client bundle too large since they're used on the client:
> packages/kit/src/runtime/client/renderer.js:import { normalize } from '../load.js';

Also, you'll need to make Lint pass on the CI and add a changeset

damain:
> Let's remove all the default error messages (which means you can ignore most of my comments as they'll no longer be relevant). It will make the client bundle too large since they're used on the client:
> 
> > packages/kit/src/runtime/client/renderer.js:import { normalize } from '../load.js';
> 
> Also, you'll need to make Lint pass on the CI and add a changeset

Should it have a default error included 
```
return {
    status,
    error: new Error('Error Object Not Specified')
};
```

benmccann:
I'd probably not include any type of default error message

damain:
I am thinking since the status was between 4xx and 5xx an error should exist because the error is needed in order for the node not to load.
Since one was not sent we can do a warn to the console return the status and an empty error 
```
console.warn(' "status" property was returned from load() without an error ')
return {
    status,
    error: new Error()
};
```

packages/kit/test/apps/basics/src/routes/errors/_tests.js
```diff
@@ -211,4 +211,17 @@ export default function (test, is_dev) {
 			assert.ok(!has_stack_trace, 'Stack trace is visible');
 		}
 	});
+
+	test(
+		'server-side 4xx status without error from load()',
```
benmccann, line 216: this should probably have a client-side test as well. you can copy the tests from https://github.com/sveltejs/kit/pull/1355

damain:
> I am thinking since the status was between 4xx and 5xx an error should exist because the error is needed in order for the node not to load.
> Since one was not sent we can do a warn to the console return the status and an empty error
> 
> ```
> console.warn(' "status" property was returned from load() without an error ')
> return {
>     status,
>     error: new Error()
> };
> ```

Any feedback on this?

packages/kit/test/apps/basics/src/routes/errors/_tests.js
damain: will do

benmccann:
Yeah. Could either return empty or just return `Internal Server Error`

damain:
I updated the code it was working but the new client test is failing.  I will have to work on this some more to get this working.

packages/kit/src/runtime/load.js
```diff
@@ -5,9 +5,20 @@
...
 
-	if (loaded.error) {
-		const error = typeof loaded.error === 'string' ? new Error(loaded.error) : loaded.error;
+	if (
+		loaded.error ||
+		(loaded.status && loaded.status >= 400 && loaded.status <= 599 && !loaded.redirect)
+	) {
 		const status = loaded.status;
+		if (!loaded.error) {
+			console.warn(' "status" property was returned from load() without an error ');
```
benmccann, line 14: I don't think this should be a warning. it seems totally valid to me. I'd either remove it or make it `debug` (same below on line 33). If you do keep it, you should also removing the leading/trailing spaces

damain:
I am running into a sticking point with solving this. On the server it returns the correct status and loads the error template but on the client side it doesn't load the error template unless an error is passed from load. 

I also noted that even though the error is passed, response.status() on the client is always 200. However the node itself seems to have the correct status and it can be passed in as props.

At first I thought the issue could be in client/renderer.js where it only checks for error to see if it should generate the error template. I changed all the checks for error to also check status and that did not solve it.

benmccann:
I added more information on how to debug the tests here: https://github.com/sveltejs/kit/pull/1900

damain:
> I added more information on how to debug the tests here: #1900

are you sure #1900 is the correct link

benmccann:
Yes. But now that it's checked in, you can just view the updated README: https://github.com/sveltejs/kit#testing

packages/kit/src/runtime/load.js
```diff
@@ -4,10 +4,19 @@
  */
 export function normalize(loaded) {
 	// TODO should this behaviour be dev-only?
+	const hasStatusError =
```
benmccann, line 7: This should be called `has_error_status` because we use underscores for internal variables. We only use camel case for public APIs

packages/kit/test/apps/basics/src/routes/errors/_tests.js
```diff
@@ -211,4 +211,28 @@ export default function (test, is_dev) {
...
+			assert.equal(
+				await page.textContent('#message'),
+				'This is your custom error page saying: "Internal Server Error"'
+			);
+			assert.equal(response.status(), 401);
+		}
+	);
+
+	test(
+		'client-side 4xx status without error from load()',
```
benmccann, line 229: I'm curious, is there a reason this test has to be different from https://github.com/sveltejs/kit/pull/1355/files#diff-ac526ac527f7cd84aec0ca97b27f1e8cbc7ad3f259e5833fd8c3055b52f60a5eR130 ?

packages/kit/src/runtime/load.js
```diff
@@ -4,10 +4,19 @@
...
+		loaded.status && loaded.status >= 400 && loaded.status <= 599 && !loaded.redirect;
+	if (loaded.error || hasStatusError) {
+		const status = loaded.status;
+
+		if (!loaded.error && hasStatusError) {
+			return {
+				status,
+				error: new Error('Internal Server Error')
```
benmccann, line 15: I think I'd lean towards just doing `new Error()` and leaving the message empty

packages/kit/test/apps/basics/src/routes/errors/_tests.js
damain: It doesn't have to be different. I needed to change it to body in troubleshooting to figure out what was going on because the test was failing. Turned out that the tests were failing because I needed to run `pnpm build` before testing. I thought the build was automatically being done before the test was run.

I added the original tests from https://github.com/sveltejs/kit/pull/1355/files#diff-ac526ac527f7cd84aec0ca97b27f1e8cbc7ad3f259e5833fd8c3055b52f60a5eR130  with the necessary changes

packages/kit/src/runtime/load.js
damain: Changed it to be an empty message

packages/kit/src/runtime/load.js
```diff
@@ -4,10 +4,19 @@
  */
 export function normalize(loaded) {
 	// TODO should this behaviour be dev-only?
```
benmccann, line 6:
```suggestion
```

benmccann (approved):

.changeset/quiet-waves-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Handling 4xx and 5xx status messages without errors from load
```
benmccann, line 5:
```suggestion
Handle 4xx and 5xx statuses without requiring `Error` instance
```

Issue #1816, opened 2021-07-03
explicit config of adapter to null result to error `Cannot read property 'adapt' of null`


Mlocik97:
**Describe the bug**
Let say I use env variable to decide what adapter to use and as default value I want to use null (which is default value by documentation), when I set explicitly adapter to null in config, it throws error:

**Logs**
```
Cannot read property 'adapt' of null
TypeError: Cannot read property 'adapt' of null
    at Object.validate (file:///mnt/sda1/Programovanie/test/node_modules/@sveltejs/kit/dist/cli.js:181:48)
    at validate (file:///mnt/sda1/Programovanie/test/node_modules/@sveltejs/kit/dist/cli.js:583:28)
    at validate (file:///mnt/sda1/Programovanie/test/node_modules/@sveltejs/kit/dist/cli.js:581:19)
    at validate_config (file:///mnt/sda1/Programovanie/test/node_modules/@sveltejs/kit/dist/cli.js:670:20)
    at load_config (file:///mnt/sda1/Programovanie/test/node_modules/@sveltejs/kit/dist/cli.js:647:20)
    at async get_config (file:///mnt/sda1/Programovanie/test/node_modules/@sveltejs/kit/dist/cli.js:784:10)
    at async file:///mnt/sda1/Programovanie/test/node_modules/@sveltejs/kit/dist/cli.js:869:18
```

**To Reproduce**
Use config:

```js
kit: {
  adapter: null
}
```

**Expected behavior**
No error, and it should work same way as if there would be no adapter setting in config.

**Information about your SvelteKit Installation:**

<details>
  <summary>Diagnostics</summary>

System:
    OS: Linux 5.8 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (8) x64 Intel(R) Core(TM) i5-8300H CPU @ 2.30GHz
    Memory: 2.91 GB / 15.46 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Browsers:
    Chrome: 91.0.4472.114
    Firefox: 89.0.2
  npmPackages:
    @sveltejs/adapter-node: ^1.0.0-next.30 => 1.0.0-next.30 
    @sveltejs/adapter-static: ^1.0.0-next.13 => 1.0.0-next.13 
    @sveltejs/kit: next => 1.0.0-next.119 
    svelte: ^3.34.0 => 3.38.3 

</details>

**Severity**
For me it's really big problem, because I can't use for example:

```js
kit: {
[...]

PR #1819, opened 2021-07-04
Solving #1816

Mlocik97:

### Tests

### Changesets

### DIFF
--- a/documentation/docs/14-configuration.md
+++ b/documentation/docs/14-configuration.md
@@ -14,7 +14,7 @@ const config = {
 	extensions: ['.svelte'],
 
 	kit: {
-		adapter: null,
+		adapter: { adapt: () => {} },
 		amp: false,
 		appDir: '_app',
 		files: {


documentation/docs/14-configuration.md
```diff
@@ -14,7 +14,7 @@ const config = {
 	extensions: ['.svelte'],
 
 	kit: {
-		adapter: null,
+		adapter: { adapt: () => {} },
```
benmccann, line 17: this is not the actual default value, so it's wrong to say this

Mlocik97: but it can't be null as well, because it throws error,... ahhh. this is confusing.

benmccann: The default is no adapter. You must provide one to be able to build

Mlocik97: but what `no adapter` means in code,... I think, if I copy "default config" from documentation, I expect it will work, and not throw errors.

PR #1821, opened 2021-07-04
Pass along custom properties added to Error

### LINKED ISSUE #1793: error load function has different parameter prototypes depending on initial load or browser navigation
**Describe the bug**
When the error load function is called on initial load it gets passed an error object of prototype `Object`. If the same error load function is called on browser router navigation the error object has the `Error` prototype.

**Logs**
browser logs (source at https://github.com/mohe2015/sveltekit-bug-load-instanceof/blob/main/src/routes/__error.svelte#L5):
```
// page load
false __error.svelte:5:10
Error __error.svelte:6:10
Object { … }

// page router navigation
true __error.svelte:5:10
Error __error.svelte:6:10
Error.prototype { stack: "", … }
```

**To Reproduce**
```bash
git clone https://github.com/mohe2015/sveltekit-bug-load-instanceof.git
cd sveltekit-bug-load-instanceof
npm install
npm run dev -- --open
```
Open the page and see the first result.
Click on one of the navigation links and you get the second result.

**Expected behavior**
The behaviour should either be consistent if possible or the documentation should probably be updated.

<details>

- System:
    OS: Linux 5.13 NixOS 21.11 (Porcupine) 21.11.20210630.13af8f4 (Porcupine)
    CPU: (8) x64 AMD Ryzen 7 4700U with Radeon Graphics
    Memory: 1.22 GB / 15.05 GB
    Container: Yes
    Shell: 4.4.23 - /run/current-system/sw/bin/bash
  Binaries:
    Node: 16.4.0 - /nix/store/8ar86gc2f7l38kz0y9nj45mbjailkymd-nodejs-16.4.0/bin/node
    npm: 7.18.1 - /nix/store/8ar86gc2f7l38kz0y9nj45mbjailkymd-nodejs-16.4.0/bin/npm
  Browsers:
    Firefox: 89.0.2
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.119 
    svelte: ^3.38.3 => 3.38.3
- Firefox 89.0.2
</details>

**Severity**
Could work around it (and probably will) - is just weird and inconsistent.

mohe2015:
Fixes #1793

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -215,8 +215,8 @@ function serialize_error(error) {
 	if (!error) return null;
 	let serialized = try_serialize(error);
 	if (!serialized) {
-		const { name, message, stack, frame, loc } = error;
-		serialized = try_serialize({ name, message, stack, frame, loc });
+		const { name, message, stack } = error;
+		serialized = try_serialize({ ...error, name, message, stack });
 	}
 	if (!serialized) {
 		serialized = '{}';
--- a/.changeset/neat-walls-buy.md
+++ b/.changeset/neat-walls-buy.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Pass along custom properties added to Error


packages/kit/src/runtime/server/page/render.js
```diff
@@ -216,7 +216,7 @@ function serialize_error(error) {
 	let serialized = try_serialize(error);
 	if (!serialized) {
 		const { name, message, stack } = error;
-		serialized = try_serialize({ name, message, stack });
+		serialized = try_serialize({ ...error, name, message, stack });
```
benmccann, line 219: this is a weird way to write this. At this point you might as well just pass `error` instead of `{ ...error, name, message, stack }`

benmccann:
What attribute are you hoping to pass through exactly? I just passed two more in https://github.com/sveltejs/kit/pull/1827. Is there anything else that we're still missing? I don't know why this code was originally written the way that it was...

mohe2015:
> What attribute are you hoping to pass through exactly? I just passed two more in #1827. Is there anything else that we're still missing? I don't know why this code was originally written the way that it was...

I want to pass custom additional attributes of a subclass of Error through.

packages/kit/src/runtime/server/page/render.js
mohe2015: Passing error does not work because try_serialize doesn't detect it's type in development (which is your custom class prototype).  `{ ...error }` would be possible though. That how I understand it at least.

mohe2015:
My exact use case is that I want to just throw in a theoretically deep method call in `load` (not the error one) but this would create a 500 server error. So then I check in the __error `load` if it's the type of error I'm interested in (not authenticated) and return a 401 instead. I had an attribute with the status code in the error and it didn't get passed along. (The prototype doesn't neither but this seems to be unfixable as it needs to be serialized as far as I understand)

packages/kit/src/runtime/server/page/render.js
mohe2015:
```js
class MyError extends Error {
	test = "test"
}

var error = new MyError();
var { name, message, stack } = error;
console.log({ ...error, name, message, stack })
```
removing `name, message, stack` doesn't work because JavaScript is strange. Also `JSON.stringify(error)` doesn't work for the same reason AFAICT.

benmccann: Ah, it looks like it's because `Error`'s default properties are not enumerable. Though if the Svelte compiler throws an error it you can see the following properties: `stack,message,name,code,start,end,pos,filename,frame,plugin,id,pluginCode,loc`.

Anyway, your change looks good to me

Issue #1561, opened 2021-05-26
Ability to have ignored parameter in prerendered path


benmccann:
https://github.com/sveltejs/kit/issues/1533 was requesting to have an ignored parameter in the path. The parameter is unused on the server-side so they'd like to be able to prerender the page without specifying the parameter, but they'd like to use the parameter on the client-side still in `onMount`

johnnysprinkles:
Lol, my 1000 words in 20. I think the confusion is when I said "parameterized route" I meant actually parameterized, i.e. `/tshirt/[size]` when size is s, m, l, xl isn't actually parameterized, in my view, that's just 4 routes. I'm going to look into possibly making a new adapter, "adapter-static-dynamic" but a better name of course.

benmccann:
I don't think you can really do this with the existing adapter API. You need support in prerendering which is in the core SvelteKit library

johnnysprinkles:
That is true, `adapter-static` seems to just a lightweight wrapper around util/prerender.

benmccann:
It seems the issue with having an ignored parameter though would be figuring out how it should show up in the URL. You'd want SvelteKit to be able to do that. I'm not quite sure what that'd look like, so I'm going to close this until there's a good proposal for it

Issue #1533, opened 2021-05-23
Static rendering all the routes, even the parameterized ones


johnnysprinkles:
This is my exploration into the idea of static rendering all the routes such that they can be served in a totally language agnostic way. This is of course easy to do already for routes with no route params, but not so simple for the parameterized routes.

Why would I want to pre-render a parameterized route, where all the data comes from the clientside XHR call? Aren't I just prerendering a blank page? Not at all my friend, you prerender as much as you can which would typically include at least the global nav bar. Having that appear instantly instead of merely quickly upon hydration makes all the difference in giving your site an app-like feel.

And in my case it's especially important since I'm not using client-side routing, so each page load needs to be as fast-feeling as possible. Why no client-side routing? Imagine a large site with 1000 pages, an in-house project spanning many years and having many separate teams. Building this as a single app is bound to result in slow build times, so you can split it up into 10 Svelte projects where each has 100 pages (all behind a reverse proxy). But as long as you're splitting it at all it makes sense to just split it completely with full page reloads on every transition. Now you can reproportion ownership fluidly as you see fit, and most importantly, you can iteratively migrate page by page to some new framework. Perhaps whatever the hot framework of 2025 is. This a whole separate topic I could say a lot about but let's leave it at that.

So anyway, the `'*'` you pass to `kit.prerender.pages` only does the parameterless routes. A first attempt at this might be to add another route in there that has garbage for the param since it won't even be looked at from the serverside, and updating your code that runs `onMount` to only look at `window.location` instead of the page store.

I put together a repo to try out some ideas, at https://github.com/johnnysprinkles/static-all-the-things

### Setup

The first commit is just setting things up with a small constellation of services to make a more realistic prod-like setup. There's SvelteKit, there's a simple API server, and there's a web server that serves the adapter-static compiled assets off of disk. Just to ensure we have a complete break between build time and runtime the web server is written in Python. Could be anything (for me it'll probably be Java or Kotlin), but all the matters is it's not Node.

On the SvelteKit side I added two routes related to airplanes, one is a parameterless List page, and one is a detail page that has an `[id]` in its route.

![Screen Shot 2021-05-23 at 4 45 19 PM](https://user-images.githubusercontent.com/32530010/119280440-4a65f480-bbe6-11eb-83a6-5b5b7f3870fa.png)

https://github.com/johnnysprinkles/static-all-the-things/commit/b06f612a31b4009e5a44a4b05b97f7c7cffd37f1

[...]

PR #1824, opened 2021-07-04
Adding a routes.json manifest file

johnnysprinkles:
This relates to the general topic of making adapter-static more powerful and useful, discussed in various issues including:

https://github.com/sveltejs/rfcs/pull/52
https://github.com/sveltejs/kit/issues/1533
https://github.com/sveltejs/kit/issues/1561

When you use adapter-static you get a big tree of files in your `build/` folder, and it's up to you to figure out how to serve them in response to incoming requests. The simplest thing to do is check for an exact match and existence of the file on disk. Doing anything more complex would need information about what the routes are. We can currently kind of parse the manifest.js file but it would be much easier to simply open a JSON file, which any kind of runtime or language supports. The output is like this:

```
jpsimons@Johns-MacBook-Pro sv2 % cat .svelte-kit/build/generated/routes.json
[
  {
    "pattern": "^\\/$",
    "params": [],
    "componentPath": "src/routes/index.svelte"
  },
  {
    "pattern": "^\\/about\\/?$",
    "params": [],
    "componentPath": "src/routes/about.svelte"
  },
  {
    "pattern": "^\\/todos\\/?$",
    "params": [],
    "componentPath": "src/routes/todos/index.svelte"
  },
  {
    "pattern": "^\\/item\\/([^/]+?)\\/?$",
    "params": [
      "id"
    ],
    "componentPath": "src/routes/item/[id].svelte"
  }
]
```

That's the stock demo app, with the addition of a parameterized route.

My next step is I want to make an "adapter-static tester" in the form of an npx you can run to serve your static build folder. It would be fully decoupled from server-side Node, and would only be implemented in Node for convenience. It will prove that the artifacts could be served by anything.

Note: I tried to do the changelist thing, which I couldn't get to work. Also, a test unrelated to my stuff was already failing.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/create_app/index.js
+++ b/packages/kit/src/core/create_app/index.js
@@ -33,7 +33,7 @@ export function create_app({ manifest_data, output, cwd = process.cwd() }) {
 	const base = path.relative(cwd, dir);
 
 	write_if_changed(`${dir}/manifest.js`, generate_client_manifest(manifest_data, base));
-
+	write_if_changed(`${dir}/routes.json`, generate_route_manifest(manifest_data));
 	write_if_changed(`${dir}/root.svelte`, generate_app(manifest_data, base));
 }
 
@@ -105,6 +105,28 @@ function generate_client_manifest(manifest_data, base) {
 
 		export const fallback = [c[0](), c[1]()];
 	`);
+}
+
+/**
+ * Information about all page routes, in a language-agnostic JSON file.
+ *
+ * @param {ManifestData} manifest_data
+ * @returns {string}
+ */
+function generate_route_manifest(manifest_data) {
+	const page_routes = manifest_data.routes
+		.map((route) => {
+			if (route.type === 'page') {
+				const pattern = route.pattern.toString();
+				return {
+					pattern: pattern.slice(1, pattern.length - 1), // strip delimiters
+					params: route.params,
+					componentPath: route.a[route.a.length - 1]
+				};
+			}
+		})
+		.filter((route) => !!route);
+	return JSON.stringify(page_routes, undefined, 2);
 }
 
 /**


packages/kit/src/core/create_app/index.js
```diff
@@ -107,6 +107,28 @@ function generate_client_manifest(manifest_data, base) {
...
+ */
+function generate_route_manifest(manifest_data) {
+	const page_routes = manifest_data.routes
+		.map((route) => {
+			if (route.type === 'page') {
+				const pattern = route.pattern.toString();
+				return {
+					pattern: pattern.slice(1, pattern.length - 1), // strip delimiters
+					params: route.params,
+					componentPath: route.a[route.a.length - 1]
```
benmccann, line 124: there won't be any `.svelte` files in the build directory, so is this really useful?

benmccann (commented):
I don't quite understand what you're trying to do. Any static file host can already serve the output of `adapter-static` and it's not coupled to a Node serving environment. E.g. you can just run ` python3 -m http.server` in the build directory. Is there any new functionality this would allow?

packages/kit/src/core/create_app/index.js
johnnysprinkles: True but it's an easy transformation to get from that to the HTML file. Basically replacing .svelte with .html and possibly adding an /index.

johnnysprinkles:
Correct, static file hosting is covered, which is a special case of the web server general case. A couple things:

1. Actually knowing what the routes are instead of just try/catching on a file read seems like a more correct way of doing things.
2. Those parameterized routes of course, something like `/item/[id]`, no way to know how to parse the route params and find the corresponding HTML artifact without this info. It can be obtained from regex slicing the JS file but why make it hard?
3. Aside from parameterized routes, you can certainly imagine a case where there's not a 1:1 correspondence. Maybe for example your web server knows if the user is signed in, and hitting the root `/` path causes it to serve either an index_anonymous.html or an index_signedin.html. This kind of thing is easier and more robust if you know what all the routes are.

johnnysprinkles:
Also one other thing... if the web server knows the route information, it can interpolate that into the response as a global variable or whatever, and then none of the changes I suggest in https://github.com/sveltejs/rfcs/pull/52 (threading that info through into a singleton, messing around with page store, etc) would even be necessary.

johnnysprinkles:
Also just wanted to restate my argument about static SSR since I know you and Rich agree on a different opinion... my view is that static SSR (or SSG if you like) is a really awesome development, and largely obviates dynamic SSR. It's in line with Svelte philosophy of "do everything you can at compile time", it allows the flexibility to use any server runtime, and should be less resource intensive for high traffic on a given machine. I think we should be using it for everything!

And, there's really no need to be hung up on this idea that parameters from the page store need to match at build time and client runtime. Clearly, `page.host` is a counter example of that. At build time, host would be "localhost" but at runtime of course you'd want the real "in situ" host. This does mean you could be dumb and check it in load and do something with it, then have a mismatch between JS enabled/disabled, first load and client-side transition, etc.

    export async function load({ page }) {
      return {
        props: {
          message: `Hello, welcome to the website at ${page.host}`
        }
      }
    }

In this case, you know the "page.host" only makes sense at runtime, so you'd want to have any logic referencing only in client-side code, i.e. `onMount`.

My argument is the exact same holds for parameterized path params, and for query strings for that matter.

johnnysprinkles:
And it is looking like I'll need to maintain a fork of SvelteKit, which I hate to have to do, but if I do I'll document the motivation clearly and it won't be a continuously diverging fork, just "SvelteKit + like 4 commits".

johnnysprinkles:
And also, of course I'd like to get your opinion to see where you're coming from Ben. What web architectures excite and interest you? Do you think React's current direction of streaming SSR can give you all the benefits of static SSR (instant page load, at least for the shell of the page) and dynamic SSR (client never has to build up its own DOM for the complete initial render)?

johnnysprinkles:
So this might not actually be strictly necessary... if you use a wildcard convention for path params which include the param's name, while prerendering, I think you could just descend the filesystem tree much like simple static server would do, and for any path part that's not found, look for a sibling directory using said convention.

I still think it's a good idea to have the routes available in a JSON file, but I can't really argue that not having them is blocking anything.

johnnysprinkles:
Yeah, there are more important and necessary things to think about here, I think this PR is a distraction, let me close it and there will be a couple PRs coming.

benmccann:
I'm afraid I can barely keep up with all the activity on this repo. It'd help if you could organize your thoughts concisely and have a specific ask if any. There's quite a few messages here since I last took a look, so I'm not sure what to respond to if anything

johnnysprinkles:
Totally agree, I've littered the repo with lots of comments and issues. Let's consider this PR closed. I might think about resolving all my open issues and creating a new one that's more concise, which would basically say this:

"If you use static pre-rendering, the parts of the page that depend on dynamic query params or path params will of course have to be client-rendered. But SvelteKit can't support this very well as-is. Here's how we support that better."

PR #1827, opened 2021-07-05
Surface Svelte compiler errors

benmccann:
The stack trace just shows that the compiler is throwing an exception. It doesn't have the added details of what file or line number. See https://github.com/sveltejs/vite-plugin-svelte/issues/62

### DIFF
--- a/packages/kit/src/core/build/index.js
+++ b/packages/kit/src/core/build/index.js
@@ -313,7 +313,10 @@ async function build_server(
 					floc: ${config.kit.floc},
 					get_component_path: id => ${s(`${config.kit.paths.assets}/${config.kit.appDir}/`)} + entry_lookup[id],
 					get_stack: error => String(error), // for security
-					handle_error: error => {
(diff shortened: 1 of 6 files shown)


packages/kit/src/core/build/index.js
```diff
@@ -314,6 +314,10 @@ async function build_server(
 					get_component_path: id => ${s(`${config.kit.paths.assets}/${config.kit.appDir}/`)} + entry_lookup[id],
 					get_stack: error => String(error), // for security
 					handle_error: error => {
+						// frame is populated by Svelte in its CompileError and is a Rollup/Vite convention
+						if (error.frame) {
+							console.error(error.frame);
+						}
```
GrygrFlzr, line 320: Does this not need the type hint provided in the `dev` equivalent?

benmccann: No idea. None of the tests / examples do type checking so there's not an easy way to check and we're constantly braking things. I went ahead and added it

PR #1836, opened 2021-07-06
chore: Enable `vite.server.fs.strict` by default

GrygrFlzr:
- All tests (except the default config test) involving a `svelte.config.js` have been changed to enable this parameter
- `create-svelte` was updated to include this parameter by default

In the future, when Vite makes this the default behavior, we can remove it from the configs.

### Tests

### Changesets

### DIFF
--- a/.changeset/fluffy-adults-happen.md
+++ b/.changeset/fluffy-adults-happen.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Enable vite.server.fs.strict by default
--- a/examples/hn.svelte.dev/svelte.config.js
+++ b/examples/hn.svelte.dev/svelte.config.js
@@ -3,6 +3,13 @@ import netlify from '@sveltejs/adapter-netlify';
 export default {
 	kit: {
 		adapter: netlify(),
-		target: '#svelte'
+		target: '#svelte',
+		vite: {
+			server: {
+				fs: {
+					strict: true
+				}
+			}
+		}
 	}
 };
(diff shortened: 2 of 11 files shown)


dummdidumm (approved):

benmccann:
How about we make it part of the default SvelteKit config instead of overriding it in each project? IT'd be a much smaller change and then existing projects would get it without needing up update their config

GrygrFlzr:
My concern is that it'd essentially be unoverridable - kit's internal config deep merge would override the user's `vite.server.fs.strict` option.

dummdidumm:
I also prefer to not bake this into Kit. It's a vite config so it should live in that space.

benmccann:
> kit's internal config deep merge would override the user's vite.server.fs.strict option

Why would it override the user's? I believe the user-provided value is always preferred

>  It's a vite config so it should live in that space.

Yes, it'd still live at the same location. I'm just suggesting SvelteKit provide a better default. It'd still live in the `vite` config section

GrygrFlzr:
> Why would it override the user's? I believe the user-provided value is always preferred

The opposite - we have some kit configs we *don't* want the user to override, so the deep merge explicitly overrides them:

https://github.com/sveltejs/kit/blob/1b134165e6b6aa73f93a8c513adf79f4593aaef2/packages/kit/src/core/dev/index.js#L91-L92

https://github.com/sveltejs/kit/blob/6ef148df0eeeb9c29d863906068ecf1494056158/packages/kit/src/core/build/index.js#L138-L139

https://github.com/sveltejs/kit/blob/6ef148df0eeeb9c29d863906068ecf1494056158/packages/kit/src/core/config/index.js#L208-L214

benmccann:
Ah, yeah. Though perhaps we could add this first:
```
deep_merge(defaults, user_config, overrides);
```

PR #1830, opened 2021-07-05
[fix] `ReadOnlyFormData` types

JeanJPNM:
Fixes types of the `ReadOnlyFormData` interface. 
The changes allow instances of the class and its methods to be used inside a `for ... of` loop without receiving an error like:
`TS2495: Type 'ReadOnlyFormData' is not an array type or a string type.`

### Before submitting the PR, please make sure you do the following
- [ ] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [ ] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts

PR #1837, opened 2021-07-06
[fix] correct `ReadOnlyFormData` generator implementation

ignatiusmb:
#1830 prompted me to check the actual code and I found some oddities, it will now follow the written types correctly.


benmccann:
Is there some kind of test we can add for this?

ignatiusmb:
Good idea, I've added some unit tests

PR #1842, opened 2021-07-07
chore: Enable `vite.server.fs.strict` internally by default

GrygrFlzr:
This setting remains user-configurable and does not emit warnings when overridden. An FAQ entry has been added about the behavior should someone encounter a blocked import.

In the future, when Vite makes this the default behavior, we can remove it from the internal default.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/build/index.js
+++ b/packages/kit/src/core/build/index.js
@@ -134,8 +134,16 @@ async function build_client({
 	/** @type {any} */
 	const user_config = config.kit.vite();
 
+	const default_config = {
+		server: {
+			fs: {
+				strict: true
+			}
+		}
+	};
+
 	/** @type {[any, string[]]} */
-	const [merged_config, conflicts] = deep_merge(user_config, {
+	const [merged_config, conflicts] = deep_merge(default_config, user_config, {
 		configFile: false,
 		root: cwd,
 		base,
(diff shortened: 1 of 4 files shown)


packages/kit/src/core/build/index.js
```diff
@@ -134,8 +134,19 @@ async function build_client({
 	/** @type {any} */
 	const user_config = config.kit.vite();
 
+	const default_config = {
```
benmccann, line 137: Not sure if there's a ticket anywhere to track turning this on by default in Vite, but it could be good to add a link to it with a TODO to remove this after it's enabled

benmccann (approved):

benmccann:
I'm currently seeing warnings when I run the tests. I'm not sure if this PR might improve that at all?

>• • Unrestricted file system access to "/caching/private/uses-session"
• • Unrestricted file system access to "/caching/private/uses-fetch"
• • • • Unrestricted file system access to "/content-type-header"
• • • • • • • • • • • • • • • • • • Unrestricted file system access to "/encoded/endpoint"
• • • • • • • • • • • • • • • • • • • • • • • • • • Unrestricted file system access to "/errors/invalid-route-response"
• • • • • • Unrestricted file system access to "/etag/text"
• Unrestricted file system access to "/etag/binary"

GrygrFlzr:
This PR would actually get rid of those warnings. The issue stems from something related to vitejs/vite#4161 where it incorrectly attempts to load requested paths from the root of the filesystem, which the strict mode would prevent.

PR #1846, opened 2021-07-07
Docs for writing an adapter API

benmccann:
People have always written adapters by copying an existing one. We should probably provide more guidance. This can probably be fleshed out a bit more, but hopefully it's a starting point that we can add to

@jthegedus @jpaquim let me know if you have any suggestions from your work on writing the Firebase and Deno adapters

@ryanb I hope this helps as you take a look at writing a Begin adapter. Let me know as you go through the process if there are things missing

### DIFF
--- a/documentation/docs/10-adapters.md
+++ b/documentation/docs/10-adapters.md
@@ -33,7 +33,6 @@ export default {
 
 A variety of official adapters exist for serverless platforms...
 
-- [`adapter-begin`](https://github.com/sveltejs/kit/tree/master/packages/adapter-begin) — for [Begin](https://begin.com)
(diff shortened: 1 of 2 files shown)


jthegedus:
I think linking to each of the types when they're mentioned would significantly reduce time spent reading SvelteKit codebase searching for stuff, though that is in itself valuable and should probably be encouraged.

As would showing which package `init` and `render` come from.

That is if the goal of the docs is to not force people to copy/paste from another adapter.

benmccann:
Which types do you mean? Like where it says `import('@sveltejs/kit').Adapter`?

jthegedus:
For types I mean:

```js
{utils, config}
```

>Converts from the patform's request to a SvelteKit request, call `render`, convert from a SveteKit reponse to the platform's

`request` and `response` should link to those types in the SvelteKit repo

benmccann:
Thanks! Those were good suggestions. I'm going to go ahead and merge this to get something out there, but feel free to leave additional comments or send PRs against it

PR #1847, opened 2021-07-07
feat(ssr): Same-origin fetch cookie passthrough

### LINKED ISSUE #1777: Same-origin server fetch missing credentials for absolute paths
**Describe the bug**
https://github.com/sveltejs/kit/blob/5cea0c2e1b350d4f1794d5dd9572c940178c3478/packages/kit/src/runtime/server/page/load_node.js#L100-L103

Currently, all absolute fetch requests are treated as cross-origin, and cookies are not passed through. This does not adequately cover the three main cases of same-origin sites where cookies are shared:
- Different port on the same domain, e.g. `localhost:3000` and `localhost:5000`
- Subdomains, e.g. `api.domain.com` and `domain.com`
- The same domain and port, but specified as an absolute path to satisfy some library (e.g. Apollo, which sucks)

However, this is not a trivial fix:
- Passing in cookies for all external requests would lead to indiscriminately sending cookies to cross-origin domains, which is wasted bandwidth at best and a security violation at worst.
- **Browsers do not send the domain associated with cookies**. A SvelteKit server hosted at `sub.domain.com` can receive cookies from `domain.com` OR `sub.domain.com` and cannot distinguish the two.

The lack of domain information presents various issues and limitations for potential solutions.

**[Loose]** If we assume that received cookies all belong to the second-level domain:
- We must identify what a "second-level domain" actually is (e.g. `domain.co.uk` exists, so we cannot blindly assume the level of the domain based on the number of separators)
- This is a security issue for subdomains that are shared with others, as is common in shared hosting (e.g. vercel, github pages, js.org) as it can happily send *all* `vercel.app` cookies to `malicious.vercel.app`
- This assumption is invalid for any pages hosted on direct IP addresses (eg. https://1.1.1.1 from Cloudflare)

**[Strict]** If we pick the strictest domain and assume received cookies belong to the current domain of the application:
- A SvelteKit app hosted at `sub.domain.com` would have to assume `domain.com` cookies as `sub.domain.com` cookies, and so `api.domain.com` would not be eligible to receive the cookies from server-side fetch using this algorithm, despite the fact that they would normally be passed by the browser clientside.
- The above behavior would require extensive documentation, as it would be a difference between the client and the server fetch.

The above also assumes we can adequately identify the current host. The `Host` HTTP header can be manipulated by the client, which seems questionable. A configurable setting may lead unaware devs to emulate the loose mode to "make things work" and turn the feature into a footgun.

**To Reproduce**
https://github.com/GrygrFlzr/sk-cookie-issue

**Expected behavior**
Cookies passed through to the API endpoint for the "Same Domain and Port" and "Same Domain, Different Port" cases.

<details><summary>Diagnostics</summary>

```
  System:
    OS: Windows 10 10.0.22000
    CPU: (16) x64 AMD Ryzen 7 3700X 8-Core Processor
    Memory: 15.96 GB / 31.93 GB
  Binaries:
    Node: 14.16.0 - C:\program files\nodejs\node.EXE
    npm: 6.14.13 - C:\program files\nodejs\npm.CMD
  Browsers:
    Chrome: 91.0.4472.124
    Edge: Spartan (44.22000.1.0), Chromium (91.0.864.59)
    Internet Explorer: 11.0.22000.1
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.119
    svelte: ^3.34.0 => 3.38.3
```
</details>

**Severity**
Depends:
- Annoyance for situations where proxying the API in SvelteKit is possible - it's an extra network trip.
- Blocking for situations involving third party libraries.

GrygrFlzr:
Closes #1777. Mostly.

Proper subdomain testing would need fudging with the `hosts` file, which would be unreasonable to include in automated tests.

### Tests

### Changesets

### DIFF
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -98,6 +98,8 @@ So if the example above was `src/routes/blog/[slug].svelte` and the URL was `htt
 
 > When `fetch` runs on the server, the resulting response will be serialized and inlined into the rendered HTML. This allows the subsequent client-side `load` to access identical data immediately without an additional network request.
 
+> Cookies will only be passed through if the target host is the same as the SvelteKit application or a higher subdomain of it.
+
 #### session
 
 `session` can be used to pass data from the server related to the current request, e.g. the current user. By default it is `undefined`. See [`getSession`](#hooks-getsession) to learn how to use it.
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -98,7 +98,7 @@ So if the example above was `src/routes/blog/[slug].svelte` and the URL was `htt
 
 > When `fetch` runs on the server, the resulting response will be serialized and inlined into the rendered HTML. This allows the subsequent client-side `load` to access identical data immediately without an additional network request.
 
-> Cookies will only be passed through if the target host is the same as the SvelteKit application or a higher subdomain of it.
+> Cookies will only be passed through if the target host is the same as the SvelteKit application or a more specific subdomain of it.
 
 #### session
 
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -98,9 +98,32 @@ export async function load_node({
 				let response;
 
 				if (/^[a-zA-Z]+:/.test(url)) {
-					// external fetch
-					const request = new Request(url, /** @type {RequestInit} */ (opts));
-					response = await options.hooks.serverFetch.call(null, request);
+					// possibly external fetch
+					if (typeof request.host !== 'undefined') {
+						const { hostname: fetchHostname } = new URL(url);
+						const [serverHostname] = request.host.split(':');
+
+						// allow cookie passthrough for "same-origin"
+						// if SvelteKit is serving my.domain.com:
+						// -        domain.com WILL NOT receive cookies
+						// -     my.domain.com WILL receive cookies
+						// -    amy.domain.com WILL NOT receive cookies
+						// -    api.domain.dom WILL NOT receive cookies
+						// - sub.my.domain.com WILL receive cookies
+						// ports do not affect the resolution
+						// leading dot prevents mydomain.com matching domain.com
+						if (`.${fetchHostname}`.endsWith(`.${serverHostname}`) && opts.credentials !== 'omit') {
+							uses_credentials = true;
+
+							opts.headers = {
+								...opts.headers,
+								cookie: request.headers.cookie
+							};
+						}
+					}
+
+					const externalRequest = new Request(url, /** @type {RequestInit} */ (opts));
+					response = await options.hooks.serverFetch.call(null, externalRequest);
 				} else {
 					const [path, search] = url.split('?');
 
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -4,6 +4,8 @@ import { resolve } from './resolve.js';
 
 const s = JSON.stringify;
 
+const hasScheme = (/** @type {string} */ url) => /^[a-zA-Z]+:/.test(url);
+
 /**
  *
  * @param {{
@@ -97,7 +99,7 @@ export async function load_node({
 
 				let response;
 
-				if (/^[a-zA-Z]+:/.test(url)) {
+				if (hasScheme(url)) {
 					// possibly external fetch
 					if (typeof request.host !== 'undefined') {
 						const { hostname: fetchHostname } = new URL(url);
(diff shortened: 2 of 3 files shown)


packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -98,9 +98,32 @@ export async function load_node({
...
+					// possibly external fetch
+					if (typeof request.host !== 'undefined') {
+						const { hostname: fetchHostname } = new URL(url);
+						const [serverHostname] = request.host.split(':');
+
+						// allow cookie passthrough for "same-origin"
+						// if SvelteKit is serving my.domain.com:
+						// -        domain.com WILL NOT receive cookies
+						// -     my.domain.com WILL receive cookies
+						// -    amy.domain.com WILL NOT receive cookies
```
benmccann, line 110: this line and the next line seem a bit duplicative. we probably don't need both. I wonder if this line is a typo since it's so similar to "my"?

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -98,9 +98,32 @@ export async function load_node({
...
 				if (/^[a-zA-Z]+:/.test(url)) {
-					// external fetch
-					const request = new Request(url, /** @type {RequestInit} */ (opts));
-					response = await options.hooks.serverFetch.call(null, request);
+					// possibly external fetch
+					if (typeof request.host !== 'undefined') {
+						const { hostname: fetchHostname } = new URL(url);
+						const [serverHostname] = request.host.split(':');
+
+						// allow cookie passthrough for "same-origin"
```
benmccann, line 106: maybe we should also go ahead and handle the `include` case as well?

packages/kit/src/runtime/server/page/load_node.js
GrygrFlzr: Not a typo - it's showing a case where `fetchHostname.endsWith(serverHostname)` without prepending a dot would incorrectly pass cookies.

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -98,9 +98,32 @@ export async function load_node({
 				let response;
 
 				if (/^[a-zA-Z]+:/.test(url)) {
```
benmccann, line 100: It's fairly hard to tell what this regex is doing now. Could we change it to `/^https?:/` which I think would be equivalent?

packages/kit/src/runtime/server/page/load_node.js
benmccann: To clarify, if the user has specified `credentials: 'include'` I believe that we should also send cookies in that case? https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters

packages/kit/src/runtime/server/page/load_node.js
GrygrFlzr: `/^https?:/` only covers the majority use case and would break more esoteric use cases such as `chrome-extension:`, `ipfs:`, `data:`, `blob:`, which are all technically valid.
image

packages/kit/src/runtime/server/page/load_node.js
GrygrFlzr: We can't deal with cross-origin cookies, e.g. fetch from `domain.com` to `otherthing.com` would normally send cookies for `otherthing.com` to `otherthing.com`, but this never goes through `domain.com`

For same-origin targets it's treated the same as `credentials: same-origin`, though, since we're just checking that the fetch isn't `credentials: omit`.

packages/kit/src/runtime/server/page/load_node.js
benmccann: @antony would suggest the solution to this would be to refactor out a variable or function to document it. What do you think about putting a function at the top of the file like:
```
const hasScheme = url => /^[a-zA-Z]+:/.test(url);
```

Then here we could call:
```
if (hasScheme(url)) {
```

I think that'd make it more readable. I know it was already like this and you're not touching this line, but it was just something I noticed when trying to understand this section of code that would have helped me come up-to-speed more quickly
What do you think about that? Obviously it could be inlined, but I think the extra

documentation/docs/03-loading.md
```diff
@@ -98,6 +98,8 @@ So if the example above was `src/routes/blog/[slug].svelte` and the URL was `htt
 
 > When `fetch` runs on the server, the resulting response will be serialized and inlined into the rendered HTML. This allows the subsequent client-side `load` to access identical data immediately without an additional network request.
 
+> Cookies will only be passed through if the target host is the same as the SvelteKit application or a higher subdomain of it.
```
benmccann, line 101: To me "domain.com" is higher than "my.domain.com". Maybe we can use another word?

```suggestion
> Cookies will only be passed through if the target host is the same as the SvelteKit application or a more specific subdomain of it.
```

packages/kit/src/runtime/server/page/load_node.js
benmccann: I think I'd remove this line since it's already covered below where you say:
> leading dot prevents mydomain.com matching domain.com

PR #1853, opened 2021-07-08
Parse error occurs when application/x-www-form-urlencoded

Baedonghee:

### Tests

### Changesets

If the content-type is invalid encodeURI when application/x-www-form-urlencoded, the page fails to load due to an Uncaught URIError: URL malformed error.

Developers need to decodeUrlComponent themselves. I think.

test
Postman 
Method: Post 
Header
Content-Type: application/x-www-form-urlencoded
Body: raw
test=%2Furl%3Ftest%D111

result: 
URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at file:///Users/dongheebae/Desktop/svelte-kit/my-app/node_modules/@sveltejs/kit/dist/ssr.js:1420:36
    at Array.forEach (<anonymous>)
    at get_urlencoded (file:///Users/dongheebae/Desktop/svelte-kit/my-app/node_modules/@sveltejs/kit/dist/ssr.js:1418:4)
    at parse_body (file:///Users/dongheebae/Desktop/svelte-kit/my-app/node_modules/@sveltejs/kit/dist/ssr.js:1396:12)
    at respond (file:///Users/dongheebae/Desktop/svelte-kit/my-app/node_modules/@sveltejs/kit/dist/ssr.js:1518:11)
    at Immediate.<anonymous> (file:///Users/dongheebae/Desktop/svelte-kit/my-app/node_modules/@sveltejs/kit/dist/chunks/index.js:3336:29)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)


dummdidumm:
What other implications does this have? I guess this was added for a reason, so other things might break if it's omitted.

Baedonghee (commented):
난 그렇게 생각한다.
x-www-form-urlencoded 일때 hook.ts에서 예외처리를 해주면 된다.
만약 POST로 page를 로딩 하는 경우 페이지가 열리지 않는 버그가 있다.
실제로 Test를 해봤지만 문제가 나오지 않았다.
나와 같은 경우 외부에서 x-www-form-urlencoded로 body에 그들만의 암호화로 데이터를 보내면 decodeUrlComponent에서 에러가 나서 페이지가 열리지 않는 경우다.

Baedonghee (commented):
How about this if I say it's a problem?
`text
		.replace(/\+/g, ' ')
		.split('&')
		.forEach((str) => {
			const [key, value] = str.split('=');
			let decodeKey = key;
			let decodeValue = value;
			try {
				decodeKey = decodeURIComponent(key);
				decodeValue = decodeURIComponent(value);
			} catch(err) {
				decodeKey = key;
				decodeValue = value;
			}
			append(decodeKey, decodeValue);
		});`

Baedonghee:
> What other implications does this have? I guess this was added for a reason, so other things might break if it's omitted.

I think so.
When x-www-form-urlencoded, hook.ts can handle exceptions.
If you load a page at POST, there is a bug that does not open the page.
I actually tested it, but there was no problem.
In the same case as me, when externally sending data to the body in x-www-form-urlencoded, the codeUrl, the codeUrlComponent fails to open the page.

I think Sveltekit is being too kind.

Baedonghee:
> What other implications does this have? I guess this was added for a reason, so other things might break if it's omitted.

How about this if I say it's a problem?
text .replace(/\+/g, ' ') .split('&') .forEach((str) => { const [key, value] = str.split('='); let decodeKey = key; let decodeValue = value; try { decodeKey = decodeURIComponent(key); decodeValue = decodeURIComponent(value); } catch(err) { decodeKey = key; decodeValue = value; } append(decodeKey, decodeValue); });

JeanJPNM:
Why not just do this:
```js
/** @param {string} text */
function get_urlencoded(text) {
	const { data, append } = read_only_form_data();
	const params = new URLSearchParams(text);
	params.forEach((value, key) => append(value, key));

	return data;
}
```
It works exactly the same as the previous version, is faster (https://jsbench.me/xdkr18bycg/1), and doesn't break in your test case.

Baedonghee:
> Why not just do this:
> 
> ```js
> 
> /** @param {string} text */
> 
> function get_urlencoded(text) {
> 
> 	const { data, append } = read_only_form_data();
> 
> 	const params = new URLSearchParams(text);
> 
> 	params.forEach((value, key) => append(key, value));
> 
> 
> 
> 	return data;
> 
> }
> 
> ```
> 
> It works exactly the same as the previous version, is faster (https://jsbench.me/xdkr18bycg/1), and doesn't break in your test case.

Thank you for sharing the code with me. I'll send you a pr with that code. 👍

Baedonghee:
What should I do to approve this review? I'm waiting to use that function. Everyone, give me an answer.

JeanJPNM:
Maybe you should add a test for it

Baedonghee:
> ### PR을 제출하기 전에 다음 사항을 확인하십시오.
> * [ ] PR이 미리 논의된 문제를 언급하는 경우 정말 유용합니다. 많은 경우에 기능이 없는 이유가 있습니다. 큰 변경 사항의 경우 RFC를 생성하십시오: https://github.com/sveltejs/rfcs
> * [x]  이 메시지 본문은 어떤 문제를 해결하는지 명확하게 보여야 합니다.
> * [x]  이상적으로는 이 PR 없이는 실패하지만 함께 통과하는 테스트를 포함합니다.
> 
> ### 테스트
> * [x] 함께 검사를 실행 `pnpm test`과 함께 프로젝트를 보풀 `pnpm lint`및`pnpm check`
> 
> ### 변경 세트
> * [ ] PR이 하나 이상의 패키지 변경 로그에 기록해야 하는 변경을 `pnpx changeset`수행하는 경우 프롬프트를 실행 하고 따라 변경 집합을 생성합니다.
> 
> application/x-www-form-urlencoded 시 content-type이 잘못된 encodeURI인 경우 Uncaught URIError: URL malformed 오류로 인해 페이지가 로드되지 않습니다.
> 
> 개발자는 자체적으로 UrlComponent를 디코딩해야 합니다. 나는 생각한다.
> 
> test
> Postman
> 메서드: Post
> Header
> Content-Type: application/x-www-form-urlencoded
> 본문: raw
> test=%2Furl%3Ftest%D111
> 
> 결과 :
> URIError의 : URI에서 조작
> decodeURIComponent ()에서
> 36 : ///Users/dongheebae/Desktop/svelte-kit/my-app/node_modules/@sveltejs/kit/dist/ssr.js : 1,420 파일에
> Array.forEach에서 ()
> get_urlencoded(file:///Users/dongheebae/Desktop/svelte-kit/my-app/node_modules/@sveltejs/kit/dist/ssr.js:1418:4)
> parse_body(file:///Users /dongheebae/Desktop/svelte-kit/my-app/node_modules/@sveltejs/kit/dist/ssr.js:1396:12)
> 응답 시(file:///Users/dongheebae/Desktop/svelte-kit/my- app/node_modules/@sveltejs/kit/dist/ssr.js:1518:11)
> 즉시. (file:///Users/dongheebae/Desktop/svelte-kit/my-app/node_modules/@sveltejs/kit/dist/chunks/index.js:3336:29)
> processTicksAndRejections(internal/process/task_queues.js: 93:5)



> Why not just do this:
> 
> ```js
> /** @param {string} text */
> function get_urlencoded(text) {
> 	const { data, append } = read_only_form_data();
> 	const params = new URLSearchParams(text);
> 	params.forEach((value, key) => append(key, value));
> 
> 	return data;
> }
> ```
> 
> It works exactly the same as the previous version, is faster (https://jsbench.me/xdkr18bycg/1), and doesn't break in your test case.



> Why not just do this:
> 
> ```js
> /** @param {string} text */
> function get_urlencoded(text) {
> 	const { data, append } = read_only_form_data();
> 	const params = new URLSearchParams(text);
> 	params.forEach((value, key) => append(key, value));
> 
> 	return data;
> }
> ```
> 
> It works exactly the same as the previous version, is faster (https://jsbench.me/xdkr18bycg/1), and doesn't break in your test case.

It is slower as a result of trying that logic. I think I left out the forEach sentence at the end.

JeanJPNM:
This whole thing doesn't make sense, not just the pr itself but the original code too:

```js
text
     .replace(/\+/g, ' ')
```
There is no need to make this replacement since `decodeURIComponent` already does that.

Delegating the decoding to the developer doesn't fix the issue, and by default other frameworks decode the data before giving it to the developer, why should this case be different?

This issue doesn't even happen because its the framework's fault. It happens because someone made a mistake that resulted in incorrectly encoded urls. In that case, debug the code and find the root cause.

If the root cause is not on your code, report the bug to the developer of that code.

Baedonghee:
> This whole thing doesn't make sense, not just the pr itself but the original code too:
> 
> ```js
> text
>      .replace(/\+/g, ' ')
> ```
> 
> There is no need to make this replacement since `decodeURIComponent` already does that.
> 
> Delegating the decoding to the developer doesn't fix the issue, and by default other frameworks decode the data before giving it to the developer, why should this case be different?
> 
> This issue doesn't even happen because its the framework's fault. It happens because someone made a mistake that resulted in incorrectly encoded urls. In that case, debug the code and find the root cause.
> 
> If the root cause is not on your code, report the bug to the developer of that code.



> This whole thing doesn't make sense, not just the pr itself but the original code too:
> 
> ```js
> text
>      .replace(/\+/g, ' ')
> ```
> 
> There is no need to make this replacement since `decodeURIComponent` already does that.
> 
> Delegating the decoding to the developer doesn't fix the issue, and by default other frameworks decode the data before giving it to the developer, why should this case be different?
> 
> This issue doesn't even happen because its the framework's fault. It happens because someone made a mistake that resulted in incorrectly encoded urls. In that case, debug the code and find the root cause.
> 
> If the root cause is not on your code, report the bug to the developer of that code.

Okay, pass the encrypted value from the external framework to the website by post. The content-type passes to application/x-www-form-urlencoded, where the body value receives data as an encrypted value in their framework. So the website won't open and URIError: URL malformed error. Error.
Then I'll close this PR first.

PR #1871, opened 2021-07-10
Ability to imperatively parse the path params, at clientside runtime

johnnysprinkles:
Normally, client-side code would get the path params from `$page.params`. For pre-rendered pages however, those values are the values as they were at pre-render time, which isn't useful if you're using a single pre-rendered HTML artifact to handle an entire parameterized route with path params which vary at runtime.

In such case, you'd want look at the details of the particular request, by inspecting `window.location`. This is fine for path and query string, but not currently possible for path params because we don't have the information about its pattern or param names.

This simply adds that ability, and should be considered in conjunction with just-now submitted https://github.com/sveltejs/kit/pull/1870.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/app/navigation.js
+++ b/packages/kit/src/runtime/app/navigation.js
@@ -14,6 +14,7 @@ export const goto = import.meta.env.SSR ? guard('goto') : goto_;
 export const invalidate = import.meta.env.SSR ? guard('invalidate') : invalidate_;
 export const prefetch = import.meta.env.SSR ? guard('prefetch') : prefetch_;
 export const prefetchRoutes = import.meta.env.SSR ? guard('prefetchRoutes') : prefetchRoutes_;
+export const routeParams = import.meta.env.SSR ? guard('routeParams') : routeParams_;
 
 /**
  * @type {import('$app/navigation').goto}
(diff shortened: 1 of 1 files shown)


packages/kit/src/runtime/app/navigation.js
```diff
@@ -49,3 +50,13 @@ async function prefetchRoutes_(pathnames) {
...
 	await Promise.all(promises);
 }
+
+/**
+ * @returns {Record<string, string>}
+ */
+function routeParams_() {
+	const { routes } = router.parse(new URL(location.href));
+	if (routes.length === 0) return {};
+	const match = routes[0][0].exec(location.pathname);
```
benmccann, line 60: This could use a comment as to why this line and the next are `routes[0]`. It seems like it's only checking the first route? Why is that?

johnnysprinkles: That's a really good comment. I think if there's multiple ambiguous matching routes it gives them all, we'd probably want the first or the last, I'll need to investigate. The order comes from whatever order they're in generated/manifest.js.

Also I don't love the magic indices after that. I should cast it as a CSRPage type and destructure it instead.

I'll have an update soon.

johnnysprinkles:
I should probably close this, since the main reason to do this is in support of general-case static prerendering, which we're not doing now or perhaps ever. See https://github.com/sveltejs/kit/pull/1870.

PR #1876, opened 2021-07-11
[fix] correct types mismatch

### LINKED ISSUE #1494: Possible incorrect types
I believe `Buffer` was replaced with `Uint8Array` in https://github.com/sveltejs/kit/pull/1382, so perhaps the type below is wrong

https://github.com/sveltejs/kit/blob/11e7840c67c2cae17940b0d9664ca0fb3da91eb9/packages/kit/types/helper.d.ts#L10

Also, `Incoming` is unused in `hooks`:

https://github.com/sveltejs/kit/blob/586785da75a3d0d6eb27a2ecd171208bbdb935a3/packages/kit/types/hooks.d.ts#L4

It's exported and used in [`index`](https://github.com/sveltejs/kit/blob/586785da75a3d0d6eb27a2ecd171208bbdb935a3/packages/kit/types/index.d.ts) and [`internal`](https://github.com/sveltejs/kit/blob/586785da75a3d0d6eb27a2ecd171208bbdb935a3/packages/kit/types/internal.d.ts), but I wonder if `ServerRequest` should be used instead

ignatiusmb:
Resolves #1494

### DIFF
--- a/.changeset/short-buttons-mix.md
+++ b/.changeset/short-buttons-mix.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Replace return type of Buffer with Uint8Array
--- a/.changeset/spicy-jeans-rest.md
+++ b/.changeset/spicy-jeans-rest.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Remove Incoming from public types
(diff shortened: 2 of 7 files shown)


dummdidumm (approved):

benmccann (approved):
thank you!!

PR #1878, opened 2021-07-11
[feat] make template filenames clearer

### LINKED ISSUE #1874: Too many index.svelte files in SvelteKit demo app
### Describe the problem

- "index" does not describe the component
- Showing the filename in the editor tab is no longer enough
- Automatic importing gets unnecessary complicated (does't work)
- command+p to open a component doesn't works as good
- git: Reading the Changes is slightly harder

Rename the "src/lib/Counter/_**index**_.svelte" to "src/lib/Counter/_**Counter**_.svelte"

Keep the `src/routes/todos/index.svelte` as-is, that filename is used for routing.
(I'd personally would split the rendering of the form of into a component, but for example-code it's easier to digest if it remains in a single file )

### Alternatives considered

I've worked with vue and react codebases that worked with lots of index.ts files. When I've renamed these files to have meaningful and predictable filenames the reaction has been universally positive.

An alternative is to do nothing, the filename convention in the example does not affect how we setup a svelte-kit project.

### Importance

nice to have

ignatiusmb:
Resolves #1874


dummdidumm (approved):

benmccann:
I totally agree with the Counter rename. `src/lib/svelte-logo.svg` doesn't quite seem like where I'd put that file. It made sense that it was grouped with Header in the Header folder before. In my own project I've created `src/images` with an `$img` alias

dummdidumm:
I prefer not to introduce another alias in the default template, so I prefer to keep it as proposed here.

benmccann:
Or should we just keep it as in `master`? I don't think it's a good pattern to encourage people to stick all their images directly in `$lib`

dummdidumm:
How about leaving the header folder as is and only rename the file in it? So it would be `import Counter from "$lib/counter/Counter.svelte"`, which is fine for me

benmccann:
I think `lib/Counter.svelte` as it is in this PR makes sense. No need for a sub-directory when there's only a single file in it. You could use that naming scheme for `Header` though. The sub-directory makes more sense there since there are two files in it

ignatiusmb:
That is probably the sweet spot, we'll provide both ways and users can see and choose for themselves which one they prefer later.

PR #1886, opened 2021-07-12
Fix returning null from endpoints

utkarshkukreti:
https://github.com/sveltejs/kit/pull/1808 introduced a bug which converts `null` to `{}` (empty object) when returned from endpoints. This PR fixes it and includes a test for the fix.

### Tests

### Changesets

### DIFF
--- a/.changeset/heavy-turkeys-report.md
+++ b/.changeset/heavy-turkeys-report.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+fix returning null from endpoints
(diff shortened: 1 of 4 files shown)


utkarshkukreti:
I used `typeof body === 'undefined'` because the code above used it, but I think `body === undefined` would have identical behavior here? Any reason to prefer the former?

dummdidumm (approved):

ignatiusmb:
Good catch! Would it make sense to check for other types as well outside of `"object"`?

> Any reason to prefer the former?

`typeof` is safer and used it out of habit, but since we know `body` is already defined, we can probably omit it.

PR #1890, opened 2021-07-12
[fix] Default body parsing to binary

benmccann:
In response to @drzax's comment (https://github.com/sveltejs/kit/pull/1687/files#r667643869) it occurs to me that the way we're treating body content type is a bit backwards. We should treat is as bytes unless it's a known type. The user can always convert bytes into whatever they're expecting. But if we try to treat it as text unless we detect a known `content-type`, we're doomed to failure since it would be hard to create a super exhaustive list of mime types to ensure we're handling them all correctly

(This is PR aligns with [how Play Framework works](https://www.playframework.com/documentation/2.8.x/JavaBodyParsers), which is one of the web frameworks I'm most familiar with. It also aligns with how our own `parse_body` works)

### DIFF
--- a/packages/kit/src/core/adapter-utils.js
+++ b/packages/kit/src/core/adapter-utils.js
@@ -3,12 +3,13 @@
  *
  * This is intended to be used with both requests and responses, to have a consistent body parsing across adapters.
  *
- * @param {string} content_type The `content-type` header of a request/response.
+ * @param {string?} content_type The `content-type` header of a request/response.
  * @returns {boolean}
  */
 export function isContentTypeTextual(content_type) {
+	if (!content_type) return true; // defaults to json
+	const [type] = content_type.split(';'); // get the mime type
 	return (
-		!content_type || // defaults to json
 		content_type === 'text/plain' ||
 		content_type === 'application/json' ||
 		content_type === 'application/x-www-form-urlencoded' ||
--- a/packages/kit/src/core/adapter-utils.js
+++ b/packages/kit/src/core/adapter-utils.js
@@ -1,16 +1,17 @@
 /**
- * Decides how the body should be parsed based on its mime type.
+ * Decides how the body should be parsed based on its mime type. Should match what's in parse_body
  *
  * This is intended to be used with both requests and responses, to have a consistent body parsing across adapters.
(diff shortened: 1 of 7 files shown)


JeanJPNM:
This will prevent it from failing if the header has encoding
```js
	if (!content_type) return true; // defaults to json
	const [type] = content_type.split(';'); // get the mime type
	return (
		type === 'text/plain' ||
		type === 'application/json' ||
		type === 'application/x-www-form-urlencoded' ||
		type === 'multipart/form-data'
	);
```

packages/kit/src/core/adapter-utils.js
```diff
@@ -1,16 +1,17 @@
 /**
- * Decides how the body should be parsed based on its mime type.
+ * Decides how the body should be parsed based on its mime type. Should match what's in parse_body
  *
  * This is intended to be used with both requests and responses, to have a consistent body parsing across adapters.
  *
  * @param {string} content_type The `content-type` header of a request/response.
```
JeanJPNM, line 6:
```suggestion
 * @param {string | undefined} content_type The `content-type` header of a request/response.
```

benmccann:
Thanks for taking a look @JeanJPNM ! Both good suggestions!

utkarshkukreti:
@benmccann, with this change, an endpoint in my app which returns `application/xml` broke.

```
export const get = () => {
    const body = '<foo />';
    return { status: 200, headers: { 'content-type': 'application/xml' }, body }
};
```

```
Invalid response from route /foo.xml: body must be an instance of Uint8Array if content-type is not a supported textual content-type
```

Should `application/xml` be added to the list of textual content types? or am I missing something and there's a better way?

JeanJPNM:
Well currently you can use a workaround:
```js
export const get = () => {
  const body = "<foo />";
  return {
    status: 200,
    headers: { "content-type": "application/xml" },
    body: new TextEncoder().encode(body),
  };
};
```

benmccann:
I sent a fix here: https://github.com/sveltejs/kit/pull/1900

utkarshkukreti:
@JeanJPNM @benmccann thanks!

PR #1894, opened 2021-07-13
[fix] fulfil rawBody data once

ignatiusmb:
Previous early `return` were to stop further fulfilments, the intent should be clearer by using else block here.

Related #1890, #1528


benmccann:
oops. I missed that the first one didn't have a `return`. thanks for catching that!

PR #1899, opened 2021-07-13
Update 01-routing.md

dcouto:
Made the word "route" plural on line 9.


benmccann:
This has come up numerous times. I agree with you because I'm American. But the original author of this project is from somewhere where "route" is plural

dcouto:
Thanks for the clarification, @benmccann.

PR #1900, opened 2021-07-13
Fix endpoint validation to allow returning string for all content types

benmccann:
Allows you to do:
```
export const get = () => {
    const body = '<foo />';
    return { status: 200, headers: { 'content-type': 'application/xml' }, body }
};
```

### DIFF
--- a/packages/kit/src/runtime/server/endpoint.js
+++ b/packages/kit/src/runtime/server/endpoint.js
@@ -10,6 +10,7 @@ function error(body) {
 	};
 }
 
+/** @param {string} s */
 function is_string(s) {
 	return typeof s === 'string' || s instanceof String;
 }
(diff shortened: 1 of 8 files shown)


packages/kit/src/runtime/server/endpoint.js
```diff
@@ -10,6 +10,10 @@ function error(body) {
 	};
 }
```
JeanJPNM, line 12: Adding a jsdoc here (`/** @param {string} s */`) removes the linting error

benmccann: thanks

packages/kit/src/runtime/server/endpoint.js
```diff
@@ -10,6 +10,11 @@ function error(body) {
...
 
+/** @param {string} s */
```
JeanJPNM, line 13: Sorry, I messed up, `unknown` is a more appropriate type than `string` here.

benmccann: guess I should have ran it locally :smile:

PR #1906, opened 2021-07-14
[fix] adapter-netlify: support for multivalue headers

### LINKED ISSUE #1741: adapter-netlify: Support `multiValueHeaders` key in returned object from handler function
Netlify Functions allow a `multiValueHeaders` key in the returned object of the handler function. This key is specifically made for array values.

For example, as the Svelte-kit docs say, in order to set multiple cookies in a response, you can do the following in an endpoint:
```javascript
return {
  headers: {
    'set-cookie': [cookie1, cookie2]
  }
};
```

However, this fails when deployed to Netlify. In order to set multiple cookies in a Netlify function, you have to do this:
```javascript
return {
  multiValueHeaders: {
    'set-cookie': [cookie1, cookie2]
  }
}
```

This is documented here: https://docs.netlify.com/functions/build-with-javascript/#synchronous-function-format.

A way to set multiple cookies with the Netlify adapter, ideally keeping the same API as described above. The adapter should check for array values in the rendered headers object and transfer them in a `multiValueHeaders` key in the returned response.

There isn't really any alternatives to this, as the only way to set multiple cookies in a response with Netlify functions is through this key.

Very! Setting multiple cookies is supposed to be a supported feature in Svelte-kit. Also, most auth systems are going to be able to do this.

JeanJPNM:

### Tests

### Changesets

Fixes #1741 by moving `set-cookie` to `multiValueHeaders` whenever it sets more than one cookie.


packages/adapter-netlify/files/entry.js
```diff
@@ -39,3 +42,21 @@ export async function handler(event) {
...
 }
+
+/**
+ * Separates headers that hold arrays from the others, and removes them from the original headers
+ *
+ * @param {Record<string, string | string[]>} headers
+ * @param  {...string} fields
+ * @returns {Record<string, string[]>}
+ */
+function getMultivalueHeaders(headers, ...fields) {
```
benmccann, line 53: I wouldn't expect a function called `getMultivalueHeaders` to modify the input. Can we write this without the side-effect? E.g. if could return an object containing the new `headers` and `multiValueHeaders`

Also, I'd expect the `V` in this function to be capitalized to match the field in the object we're returning

JeanJPNM: Should I leave the fields parameter where it is?

benmccann:
This looks pretty good to me. The one thing I wonder though is if it's the case that `Set-Cookie` is the only multi-value header that will ever exist. I might be safer to iterate though and check which headers have multiple values rather than specifying a list of headers to check

benmccann:
thanks!!

PR #1926, opened 2021-07-16
[netlify adapter] add note on Netlify node version

sw-yx:
this is just a tweak to readme, to address confusion from users https://twitter.com/BrittneyPostma/status/1416101106648813572

### DIFF
--- a/packages/adapter-netlify/README.md
+++ b/packages/adapter-netlify/README.md
@@ -8,6 +8,8 @@ This is very experimental; the adapter API isn't at all fleshed out, and things 
 
 > ⚠️ For the time being, the latest version of adapter-netlify is at the @next tag. If you get the error `config.kit.adapter should be an object with an "adapt" method.`, this is a sign that you are using the wrong version (eg `1.0.0-next.0` instead of `1.0.0-next.9`).
 
+> ⚠️ For the time being, Netlify's default Node version is 12. SvelteKit requires Node v14 to build. you can pin the Node version with a [`.nvmrc`](https://github.com/nvm-sh/nvm#nvmrc) file: `echo "14" > .nvmrc` or [set the `NODE_ENV` environment variable](https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript).
+
 ```bash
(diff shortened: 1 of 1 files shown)


dominikg:
I think the minimum version is 12.20, not 14. For some odd reason netlify defaults to 12.16. 

And maybe `.node-version` instead of `.nvmrc` would be a more suiting filename

packages/adapter-netlify/README.md
```diff
@@ -8,6 +8,8 @@ This is very experimental; the adapter API isn't at all fleshed out, and things
 
 > ⚠️ For the time being, the latest version of adapter-netlify is at the @next tag. If you get the error `config.kit.adapter should be an object with an "adapt" method.`, this is a sign that you are using the wrong version (eg `1.0.0-next.0` instead of `1.0.0-next.9`).
 
+> ⚠️ For the time being, Netlify's default Node version is 12. SvelteKit requires Node v14 to build. you can pin the Node version with a [`.nvmrc`](https://github.com/nvm-sh/nvm#nvmrc) file: `echo "14" > .nvmrc` or [set the `NODE_ENV` environment variable](https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript).
```
benmccann, line 11:
```suggestion
> ⚠️ Netlify defaults to Node 12.16. SvelteKit requires Node v12.20 to build. You can pin the Node version with a `.node-version` or [`.nvmrc`](https://github.com/nvm-sh/nvm#nvmrc) file: `echo "14" > .nvmrc` or [set the `NODE_ENV` environment variable](https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript).
```

benmccann:
I tweaked the docs a bit based on the comments above. I wonder instead of documenting if we should just automatically create these files like https://github.com/sveltejs/kit/pull/1924 is suggesting?

sw-yx:
@ben sorry i took a haitus from github notifications and only just saw this. i think a simple fix like the readme warning works well here since I don't expect Netlify to stick to v12 too much longer. can leave the ignore discussion to that other PR

Issue #1930, opened 2021-07-17
How do I make a script run only the first time the `layout` is instantiated and not all subsequent pages (which inherit that `layout`)?


frederikhors:
### Describe the bug

In a new Svelte Kit project I have a `<script context="module">` in `__layout.svelte`.

The `load` function in it is called on each page I change.

I thought instead it was only called when the `layout` is instantiated for the first time.

I'm doing it wrong?

**How do I make a script run only the first time the `layout` is instantiated and not all subsequent pages (which inherit that `layout`)?**

### Reproduction

https://github.com/frederikhors/issue-svelte-kit-layout-context-module-load-always-called

### System Info

```shell
System:
    OS: Windows 10 10.0.19043
  Binaries:
    Node: 14.17.3 - C:\Program Files\nodejs\node.EXE
    npm: 7.19.1 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    @sveltejs/kit: 1.0.0-next.117 => 1.0.0-next.131
    svelte: 3.38.3 => 3.38.3
```

### Severity

annoyance

dummdidumm:
Yes, that's the way it works, although that may need better documentation. What's the use case for running it only the first time?

frederikhors:
To detect if the user is loggedIn otherwise it redirects to `/login`.

I prefer this way over `onMount` because I can avoid to load everything `import`ed in `__layout.svelte`.

But I just found that it repeats the actions on every page change and honestly it doesn't suit me, so I'll use `onMount` with an `isLoaded` boolean.

Do you have better ideas?

But when you think about it, why repeat the `load` operation on every page change?

kevmodrome:
The way it works feels backwards to me. I wouldn't expect the layout to be re-mounted when I change pages - I guess that's the case?

Kapsonfire-DE:
For me the layout is called only once using onMount ... it will just rerun on errors

kevmodrome:
If onMount only runs once I would definitely not expect load to run more than once. Just my 2 cents though.

frederikhors:
> For me the layout is called only once using onMount ... it will just rerun on errors

Yep. We're talking about `load` in `context="module"`.

Kapsonfire-DE:
Your load is everytime called because you destructuring page
page is changed on navigation, so load is recalled everytime a destructuring variable changes

frederikhors:
> Your load is everytime called because you destructuring page
[...]

PR #1934, opened 2021-07-17
Document load reactivity

Kapsonfire-DE:
Clarify reactive in load function

See discussion: https://github.com/sveltejs/kit/issues/1930

### DIFF
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -77,7 +77,7 @@ If `load` returns nothing, SvelteKit will [fall through](#routing-advanced-fallt
 
 ### Input
 
-The `load` function receives an object containing four fields — `page`, `fetch`, `session` and `context`.
+The `load` function receives an object containing four fields — `page`, `fetch`, `session` and `context`. The `load` function is reactive to the parameters defined in the method signature. If the function uses the `LoadInput` variable, it will re-run everytime `page`, `session` or `context` changes. If you use a destructuring assignment, it will only re-run for the destructured properties of `LoadInput`.
 
 #### page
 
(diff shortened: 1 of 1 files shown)


kevmodrome (approved):

documentation/docs/03-loading.md
```diff
@@ -77,7 +77,7 @@ If `load` returns nothing, SvelteKit will [fall through](#routing-advanced-fallt
 
 ### Input
 
-The `load` function receives an object containing four fields — `page`, `fetch`, `session` and `context`.
+The `load` function receives an object containing four fields — `page`, `fetch`, `session` and `context`. The `load` function is reactive to the parameters defined in the method signature. If the function uses the `LoadInput` variable, it will re-run everytime `page`, `session` or `context` changes. If you use a destructuring assignment, it will only re-run for the destructured properties of `LoadInput`.
```
benmccann, line 80:
```suggestion
The `load` function receives an object containing four fields — `page`, `fetch`, `session` and `context`. The `load` function is reactive to the parameters defined in the method signature. If the function is written as taking a single parameter, it will re-run everytime `page`, `session` or `context` changes. If the parameter is declared as a destructured object, it will only re-run when the declared properties change.
```

benmccann:
thanks!!

rmunn:
This line was not actually true: "If the function is written as taking a single parameter, it will re-run everytime `page`, `session` or `context` changes." You can see [here](https://github.com/sveltejs/kit/blob/781f7bec5880b8fbeec02167afed3171488c8ff4/packages/kit/src/runtime/client/renderer.js#L466-L503) that the `load()` function is only reactive to the properties actually *accessed* from the parameter. (Destructuring does count as accessing). I've created #1949 to correct the documentation.

PR #1939, opened 2021-07-17
[feat] render routes ending with `.html.svelte` to html files

### LINKED ISSUE #1443: Ability to prerender to `about.html` instead of `about/index.html` like Sapper
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Severity**

honai:

### Tests

### Changesets

Resolves #1443 by changing the prerendering behavior.  If a route ends with `.html` , write it as a file name instead of creating a folder.

### Additional Info
I have to add `flush` option to glob to exactly check the dir structure.

### DIFF
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -97,7 +97,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 	 */
 	function normalize(path) {
 		if (config.kit.trailingSlash === 'always') {
-			return path.endsWith('/') ? path : `${path}/`;
+			return path.endsWith('/') || path.endsWith('.html') ? path : `${path}/`;
 		} else if (config.kit.trailingSlash === 'never') {
 			return !path.endsWith('/') || path === '/' ? path : path.slice(0, -1);
 		}
@@ -143,7 +143,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 			const is_html = response_type === REDIRECT || type === 'text/html';
 
 			const parts = path.split('/');
-			if (is_html && parts[parts.length - 1] !== 'index.html') {
+			if (is_html && !parts[parts.length - 1].endsWith('.html')) {
 				parts.push('index.html');
 			}
 
@@ -172,7 +172,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 				const is_html = result.headers['content-type'] === 'text/html';
 
 				const parts = dependency_path.split('/');
-				if (is_html && parts[parts.length - 1] !== 'index.html') {
+				if (is_html && !parts[parts.length - 1].endsWith('.html')) {
 					parts.push('index.html');
 				}
 
--- a/packages/kit/src/core/adapt/test/index.js
+++ b/packages/kit/src/core/adapt/test/index.js
@@ -91,7 +91,7 @@ suite('prerender', async () => {
 	};
 
 	/** @type {import('types/internal').BuildData} */
-	const build_data = { client: [], server: [], static: [], entries: ['/nested'] };
+	const build_data = { client: [], server: [], static: [], entries: ['/nested', '/noindex.html'] };
 
 	const utils = get_utils({ cwd, config, build_data, log });
 
@@ -103,7 +103,7 @@ suite('prerender', async () => {
 		dest
 	});
 
-	assert.equal(glob('**', { cwd: `${prerendered_files}` }), glob('**', { cwd: dest }));
+	assert.equal(glob('**', { cwd: `${prerendered_files}` }), glob('**', { cwd: dest, flush: true }));
 
 	rimraf.sync(dest);
 });
--- a/packages/kit/src/core/adapt/test/index.js
+++ b/packages/kit/src/core/adapt/test/index.js
@@ -91,7 +91,7 @@ suite('prerender', async () => {
 	};
 
 	/** @type {import('types/internal').BuildData} */
-	const build_data = { client: [], server: [], static: [], entries: ['/nested', '/noindex.html'] };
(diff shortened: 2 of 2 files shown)


benmccann:
Is there some documentation that should be added with this? Possibly in https://kit.svelte.dev/docs#ssr-and-javascript-prerender-route-conflicts ?

honai:
I think we need some documentation about creating `*.html.svelte` .
The behavior of creating `**/index.html.svelte` is ambiguous and should be avoided (regardless of this PR).
This PR will cause users to create files like `about.html.svelte`. So it may be better to warn them that they should not create `index.html.svelte` ?

As for route conflicts caused by this PR are very edge case, as follows:
- `routes/about.html.svelte` writes to `build/about.html` (after this PR)
- `routes/about.html/index.svelte`writes to `build/about.html/index.html`

As users would not include extensions in routing directory,  I don't think we need to warn users about this.

packages/kit/src/core/adapt/prerender.js
```diff
@@ -97,7 +97,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 	 */
 	function normalize(path) {
 		if (config.kit.trailingSlash === 'always') {
-			return path.endsWith('/') ? path : `${path}/`;
+			return path.endsWith('/') || path.endsWith('.html') ? path : `${path}/`;
```
benmccann, line 100: This no longer does what I would expect for an option named `always`, which would be to always do it.

This sort of feels like it should accept a function to let the user decide like https://github.com/sveltejs/kit/pull/2008 and https://github.com/sveltejs/kit/pull/2007. Then you could have it do whatever you want

packages/kit/src/core/adapt/prerender.js
```diff
@@ -143,7 +143,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 			const is_html = response_type === REDIRECT || type === 'text/html';
 
 			const parts = path.split('/');
-			if (is_html && parts[parts.length - 1] !== 'index.html') {
+			if (is_html && !parts[parts.length - 1].endsWith('.html')) {
```
benmccann, line 146: why did this need to change? if you encounter `about.html` why are you pushing `index.html`?

packages/kit/src/core/adapt/test/index.js
```diff
@@ -91,7 +91,7 @@ suite('prerender', async () => {
 	};
 
 	/** @type {import('types/internal').BuildData} */
-	const build_data = { client: [], server: [], static: [], entries: ['/nested'] };
+	const build_data = { client: [], server: [], static: [], entries: ['/nested', '/noindex.html'] };
```
benmccann, line 94: `noindex.html` is kind of a confusing name in my opinion. I'd rather just call it something unrelated like `about.html`

benmccann:
> index.html.svelte is ambiguous and should be avoided

What's ambiguous about it? I would think that it should create a file named `index.html`

dominikg:
Is the title backwards and wanted to say `.html.svelte` instead of `.svelte.html` ? 
iirc using .html as an extension for svelte does not work in vite because vites own html plugin is greedy.

benmccann:
@honai will you be able to take a look at the questions / comments on this PR?

packages/kit/src/core/adapt/prerender.js
honai: Before this change, paths like `**/about.html` are modified to `**/about.html/index.html` so I changed not to do so.

packages/kit/src/core/adapt/test/index.js
honai: Ok, I'll change to `about.html`

packages/kit/src/core/adapt/prerender.js
honai: Svelte files are designed to compose components made with HTML/CSS/JS so users won't generate files other than `.html` file, I think.

But endpoints like `sitemap.xml.js` are considered, so I think this option should be changed, as you say.

Can I change this option to something like `never | ignore | (path: string) => boolean` , in this PR?  Or do you think it would be better to change it as a separate PR?

honai:
I'm sorry for the late response.  I replied to your comments.

> What's ambiguous about it? I would think that it should create a file named `index.html`

"ambiguous" was an inappropriate description.
There are 3 possible cases in which routes are pre-rendered to `build/about/index.html`;
- `routes/about.svelte`
- `routes/about/index.svelte`
- `routes/about/index.html.svelte` (new in this PR)

The documentation already says that [routes are collated alphabetically](https://kit.svelte.dev/docs#routing-advanced-fallthrough-routes) .  I don't think any additional explanation is necessary.  What do you think?

packages/kit/src/core/adapt/prerender.js
benmccann: Yes, you're right that you'd use an endpoint

This probably shouldn't be a function after all. It gets stringified during build and could be confusing to users

I think a better logic here might be to do it only if the last path segment does not contain `.`, but let me ask the other maintainers what they think

benmccann: Hmm, that might not be a great solution either. E.g. that wouldn't work with the URL https://bundlephobia.com/package/svelte@3.42.1

I think we just leave this as is and say you should not use the option `always` if you're generating `.html` files

benmccann:
I talked to Rich about this feature briefly. His suggestion was that we create an option for generating `about.html` as opposed to `about/index.html` and that we not try to accomplish it by changing the file extension. I think that's nicer in a lot of ways because if you move web hosts you can just change the option in your config file rather than renaming all your files

benmccann:
I'm going to go ahead and close this PR given the different approach that Rich suggested, but please feel free to send a PR with the updated approach and I'll take a look

PR #1949, opened 2021-07-19
Better documentation for load reactivity

rmunn:
#1934 was good, but it was incorrect on one point: if `load` is written as taking a single parameter, it will *not* re-run whenever `session` or `context` changes. So I've updated the documentation to reflect [how the code actually works](https://github.com/sveltejs/kit/blob/781f7bec5880b8fbeec02167afed3171488c8ff4/packages/kit/src/runtime/client/renderer.js#L466-L503).

No tests or changesets needed since this is only a documentation PR.

### DIFF
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -77,7 +77,7 @@ If `load` returns nothing, SvelteKit will [fall through](#routing-advanced-fallt
 
 ### Input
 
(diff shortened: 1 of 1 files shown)


dummdidumm (approved):

Kapsonfire-DE:
Nice, thanks for making it more precise

PR #1951, opened 2021-07-19
[feat] allow chaining of `handle` hooks

SomaticIT:
Hi svelte kit team,

This is my first PR on this wonderful project.

## Problem

I tried svelte kit on some internal projects and I noticed something that was missing. It was not possible to compose the `handle` hook with middlewares.

Middlewares allow to better compose and reuse handlers and ensure Single Responsibility Principle. I noticed that some contributors already built some package to handle cookies, session, auth... (eg: `svelte-kit-session`) but it is hard to make different modules works together.

## Solution

That's why I would like to introduce, in this PR, the concept of middleware.
Now you can write the `handle` hook like this:

```javascript
export const handle = [
    ({ request, resolve }) => {
        request.locals.first = true;
        return resolve(request);
    },
    ({ request, resolve }) => {
        request.locals.second = true;
        return resolve(request);
    },
    ({ request, resolve }) => {
        console.log(request.locals); // print { first: true, second: true }
        return resolve(request);
    }
];
```

Of course, this is an example, a better way would be to separate middlewares in modules like this:

```javascript
import session from "./middlewares/session";
import auth from "./middlewares/auth";
import i18n from "./middlewares/i18n";

export const handle = [session, auth, i18n];
```

You can notice that the signature of a middleware is exactly the same as the signature of the `handle` hook. It allows to use a middleware directly as the `handle` hook if you do not need to compose it with others middlewares.

Any middleware can respond directly by returning a `ServerResponse` object. eg:

```javascript
export default function auth_middleware({ request, resolve }) {
    if (isAuthenticated(request)) {
        return resolve(request);
    }

    return {
        status: 401,
        body: ""
    };
}
```

#### UPDATE 1

After some discussion on this PR, the API changed. Now, a `sequence` method is exported by `@sveltejs/kit/hooks`. It allows to chain `Handle` functions.

```javascript
import { sequence } from '@sveltejs/kit/hooks';

export const handle = sequence(
  ({ request, resolve }) => {
    request.locals.first = true;
    return resolve(request);
  },
  ({ request, resolve }) => {
    request.locals.second = true;
    return resolve(request);
  },
  ({ request, resolve }) => {
    console.log(request.locals); // print { first: true, second: true }
    return resolve(request);
  }
);
```
## Technical details

For now, this implementation has an issue, the middleware code is duplicated in `src/runtime/hooks.js` and `src/build/index.js`.
Do you have any idea to avoid this duplication?

Also, I don't have a good understanding of the testing framework in the project. Can you help me writting a test suite?

---

### Tests

### Changesets

_Note: I will add the changeset after changes are reviewed to ensure that the changeset is targetting the good commit_

### DIFF
--- a/packages/kit/src/runtime/hooks.js
+++ b/packages/kit/src/runtime/hooks.js
@@ -1,0 +1,23 @@
+/**
+ * @param {...import('types/hooks').Handle} handles
+ * @returns {import('types/hooks').Handle}
+ */
+export function sequence(...handles) {
+	const length = handles.length;
+	if (!length) return ({ request, resolve }) => resolve(request);
+
+	return ({ request, resolve }) => {
+		return apply_handle(0);
+
+		/**
+		 * @param {number} i
+		 * @returns {import('types/helper').MaybePromise<import('types/hooks').ServerResponse>}
+		 */
+		function apply_handle(i) {
+			const handle = handles[i];
+			const next = i >= length - 1 ? resolve : () => apply_handle(i + 1);
+
+			return handle({ request, resolve: next });
+		}
+	};
+}
--- a/packages/kit/src/runtime/hooks.js
+++ b/packages/kit/src/runtime/hooks.js
@@ -7,17 +7,20 @@ export function sequence(...handles) {
 	if (!length) return ({ request, resolve }) => resolve(request);
 
 	return ({ request, resolve }) => {
-		return apply_handle(0);
+		return apply_handle(0, request);
 
 		/**
 		 * @param {number} i
+		 * @param {import('types/hooks').ServerRequest} request
 		 * @returns {import('types/helper').MaybePromise<import('types/hooks').ServerResponse>}
 		 */
-		function apply_handle(i) {
+		function apply_handle(i, request) {
 			const handle = handles[i];
-			const next = i >= length - 1 ? resolve : () => apply_handle(i + 1);
 
-			return handle({ request, resolve: next });
+			return handle({
+				request,
+				resolve: i < length - 1 ? (request) => apply_handle(i + 1, request) : resolve
+			});
 		}
 	};
 }
--- a/packages/kit/test/apps/hooks/src/hooks.js
+++ b/packages/kit/test/apps/hooks/src/hooks.js
@@ -1,6 +1,6 @@
 // The following line does not work with vite resolution. Why?
 // import { sequence } from '@sveltejs/kit/hooks';
-import { sequence } from '../../../../dist/hooks';
+import { sequence } from '../../../../src/runtime/hooks';
 
 /** @type {import('@sveltejs/kit').Handle} */
 export const handle = sequence(
--- a/packages/kit/test/apps/hooks/src/hooks.js
+++ b/packages/kit/test/apps/hooks/src/hooks.js
@@ -1,0 +1,39 @@
+// The following line does not work with vite resolution. Why?
+// import { sequence } from '@sveltejs/kit/hooks';
+import { sequence } from '../../../../dist/hooks';
+
+/** @type {import('@sveltejs/kit').Handle} */
+export const handle = sequence(
+	({ request, resolve }) => {
+		request.locals.first = true;
+		return resolve(request);
+	},
+	({ request, resolve }) => {
+		request.locals.second = 'str';
+		return resolve(request);
+	},
+	async ({ request, resolve }) => {
+		request.locals.third = 3;
+
+		const response = await resolve(request);
+
+		return {
+			...response,
+			headers: {
+				...response.headers,
+				'X-Handle-Header': 'hello'
+			}
+		};
+	},
+	async ({ request, resolve }) => {
+		const response = await resolve(request);
+
+		return {
+			...response,
+			headers: {
+				...response.headers,
+				'X-Request-Locals': JSON.stringify(request.locals)
+			}
+		};
+	}
+);
--- a/packages/kit/package.json
+++ b/packages/kit/package.json
@@ -71,6 +71,9 @@
 		"./node": {
 			"import": "./dist/node.js"
 		},
+		"./hooks": {
+			"import": "./dist/hooks.js"
+		},
 		"./install-fetch": {
 			"import": "./dist/install-fetch.js"
 		},
--- a/packages/kit/rollup.config.js
+++ b/packages/kit/rollup.config.js
@@ -48,6 +48,7 @@ export default [
 			cli: 'src/cli.js',
 			ssr: 'src/runtime/server/index.js',
 			node: 'src/core/node/index.js',
+			hooks: 'src/runtime/hooks.js',
 			'install-fetch': 'src/install-fetch.js',
 			'adapter-utils': 'src/core/adapter-utils.js'
 		},
--- a/packages/kit/test/apps/hooks/package.json
+++ b/packages/kit/test/apps/hooks/package.json
@@ -1,0 +1,16 @@
+{
+	"name": "test-hooks",
+	"private": true,
+	"version": "0.0.1",
+	"scripts": {
+		"dev": "../../../svelte-kit.js dev",
+		"build": "../../../svelte-kit.js build",
+		"preview": "../../../svelte-kit.js preview"
+	},
+	"devDependencies": {
+		"@sveltejs/adapter-node": "workspace:*",
+		"@sveltejs/kit": "workspace:*",
+		"svelte": "^3.40.0"
+	},
+	"type": "module"
+}
--- a/packages/kit/test/apps/hooks/src/app.html
+++ b/packages/kit/test/apps/hooks/src/app.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="utf-8" />
+		<meta name="viewport" content="width=device-width, initial-scale=1" />
+		%svelte.head%
+	</head>
+	<body>
+		%svelte.body%
+	</body>
+</html>
--- a/packages/kit/test/apps/hooks/src/routes/_tests.js
+++ b/packages/kit/test/apps/hooks/src/routes/_tests.js
@@ -7,10 +7,10 @@ export default function (test) {
 
 		const headers = await response.headers();
 		assert.equal(
-			headers['X-Request-Locals'],
+			headers['x-request-locals'],
 			JSON.stringify({ first: true, second: 'str', third: 3 })
 		);
 
-		assert.equal(headers['X-Handle-Header'], 'hello');
+		assert.equal(headers['x-handle-header'], 'hello');
 	});
 }
--- a/packages/kit/test/apps/hooks/src/routes/_tests.js
+++ b/packages/kit/test/apps/hooks/src/routes/_tests.js
@@ -1,0 +1,16 @@
+import * as assert from 'uvu/assert';
+
+/** @type {import('test').TestMaker} */
+export default function (test) {
+	test('serves /', '/', async ({ page, response }) => {
+		assert.equal(await page.textContent('h1'), 'Hello world!');
+
+		const headers = await response.headers();
+		assert.equal(
+			headers['X-Request-Locals'],
+			JSON.stringify({ first: true, second: 'str', third: 3 })
+		);
+
+		assert.equal(headers['X-Handle-Header'], 'hello');
+	});
+}
--- a/packages/kit/test/apps/hooks/src/routes/index.svelte
+++ b/packages/kit/test/apps/hooks/src/routes/index.svelte
@@ -1,0 +1,5 @@
+<script>
+	let message = 'Hello world!';
+</script>
+
+<h1>{message}</h1>
--- a/packages/kit/test/apps/hooks/svelte.config.js
+++ b/packages/kit/test/apps/hooks/svelte.config.js
@@ -1,0 +1,6 @@
+/** @type {import('@sveltejs/kit').Config} */
+const config = {
+	kit: {}
+};
+
+export default config;
--- a/packages/kit/types/ambient-modules.d.ts
+++ b/packages/kit/types/ambient-modules.d.ts
@@ -129,3 +129,15 @@ declare module '$service-worker' {
 	 */
 	export const timestamp: number;
 }
+
+declare module '@sveltejs/kit/hooks' {
+	type Handle = import('@sveltejs/kit').Handle;
+
+	/**
+	 * Utility function that allows chaining `hooks` `Handle` functions in a
+	 * middleware-like manner.
+	 *
+	 * @param handles The chain of `Handle` functions
+	 */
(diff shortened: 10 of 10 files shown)


JeanJPNM:
Isn't this the same as calling those functions inside `handle`?

benmccann:
> Isn't this the same as calling those functions inside handle?

That was my initial reaction as well. But I think that it's not quite the same. E.g. what if you want one of the functions to sometimes redirect and sometimes continue on to the next function. I'm not sure you'd be able to easily do that by calling the functions yourself inside `handle`?

SomaticIT:
Thank you for your first comments.

@JeanJPNM, This is not the same because a middleware always return a response. How can you ensure that you should respond or continue with the next middleware?

```javascript
  const response = middleware({ request, resolve });
  // Sould I respond or go to the next middleware?
}
//or
  const next1 = (request) => middleware2({ request, resolve: next2 });
  const next2 = (request) => middleware3({ request, resolve: next3 });
  const next3 = (request) => resolve(request);

  const response = await middleware1({ request, resolve: next1 });
  return response;
}
```

---

@benmccann, you're right, this is different and not so easy to handle it manually as shown above.

---

@rmunn, you're right, `svelte-kit-cookie-session` allows you to pass a function to control the `next` behavior. However, it has to do this because there is no middleware handling in `svelte-kit`. Moreover, if we assume that all *svelte-kit-middleware-like-packages* are following the same architecture, it will become very hard to compose them (*famous-callback-hell*):

```javascript
export const handle = middleware1({ secret: "s3kr3t" },
  middleware2({ ...options },
    middleware3({ ...options },
      async function ({ request, resolve }) {
        if (response.headers["content-type"] === "text/html") {
          let theme = request.locals.session.data?.theme ?? "light";
          response.body = (response.body as string).replace("%session.theme%", theme);
        }
        return response;
  }
)));
```

I think something like this would be more apropriate:

```javascript
export const handle = [
  middleware1,
  middleware2,
  middleware3,
  async function ({ request, resolve }) {
    // ...
  }
];

// Your exemple would become

import { handleSession } from "svelte-kit-cookie-session";
import { handleAuth } from "some-other-library";

export const handle = [
  handleSession({ secret: "s3kr3t" }),
  handleAuth({ authParams: "foo" })
];
```

Rich-Harris:
Agree that it makes sense to be able to chain handlers. I'm not personally a huge fan of APIs like `Handle | Handle[]` though — it's better to have a simple, consistent API. We can have our cake and eat it too with something like this:

```js
import { sequence } from 'somewhere';

export const handle = sequence(
  ({ request, resolve }) => {
    request.locals.first = true;
    return resolve(request);
  },
  ({ request, resolve }) => {
    request.locals.second = true;
    return resolve(request);
  },
  ({ request, resolve }) => {
    console.log(request.locals); // print { first: true, second: true }
    return resolve(request);
  }
);
```

Or maybe the API looks slightly different:

```js
import { sequence } from 'somewhere';

export const handle = sequence(
  ({ request }, next) => {
    request.locals.first = true;
    return next();
  },
  ({ request }, next) => {
    request.locals.second = true;
    return next();
  },
  ({ request, resolve }) => {
    console.log(request.locals); // print { first: true, second: true }
    return resolve(request);
  }
);
```

Perhaps returning nothing from something in the middle is equivalent to calling `next()`:

```js
import { sequence } from 'somewhere';

export const handle = sequence(
  ({ request }) => {
    request.locals.first = true;
  },
  ({ request }) => {
    request.locals.second = true;
  },
  ({ request, resolve }) => {
    console.log(request.locals); // print { first: true, second: true }
    return resolve(request);
  }
);
```

In other words this can easily be solved in userland today. The real question is whether this is something worth including in Kit itself (`import { sequence } from '@sveltejs/kit/hooks'`?), and if so what the ideal API is.

benmccann:
The last API suggestion makes the most sense to me

SomaticIT:
@Rich-Harris, @benmccann, Thank you for your suggestions.

I like the idea of having an utility function exported as part of `@sveltejs/kit`. It avoids the weird `Handle | Handle[]` configuration and it avoids duplicating the function. I will try to make changes for this.

I also like the Rich's first and third suggestions but I prefer staying on the first. I think middlewares should have the same signature as a classic `Handle` to allow them to be used standalone or as part of a sequence depending on your use case.

I will update my PR soon to apply this changes.

Do you agree?

benmccann:
I don't understand the first suggestion. `resolve` computes the entire page HTML. Why would you be doing that three times? Or does it mean `next()` the first two and then `resolve()` the last one? I really wouldn't like that API at all where it means different things at different times

rmunn:
> I don't understand the first suggestion. `resolve` computes the entire page HTML. Why would you be doing that three times? Or does it mean `next()` the first two and then `resolve()` the last one? I really wouldn't like that API at all where it means different things at different times

If I'm understanding it correctly, the first suggestion would rely on `sequence` to set things up so that `resolve` would essentially mean `next()` the first two times and `resolve()` the last time, but the user code wouldn't have to know anything about that. And that, I think, is important, because the fact that it means different things at different times becomes an **internal implementation detail**. User code just needs to know "Hey, we're handing you a function called `resolve`, call it when you're done setting things up, and then when it's done you can do things to the result". The user code doesn't need to know, or care, whether that `resolve` is Svelte-Kit's `resolve` function, or a middleware function from another layer wrapped around Svelte-Kit's `resolve`.

Then each layer of the middleware can do its own operation without caring what other layers are doing, and (crucially) **there's no way to skip layers** by returning early. That way you can't have `maliciousMiddleware()` doing an early return and skipping the part where `securityMiddleware` handles the JSON output and strips out any passwords that were missed. (Contrived example, but you get the point). If the API is set up so that you have to either call `next()` or call `resolve(request)` depending on where you are in the middleware stack, then there's too much room for error. Even a non-malicious middleware could be written to not call `next()` on certain classes of errors, but then a higher-level middleware (that could have handled those errors gracefully) gets skipped. So having the API be consistent no matter where you happen to be sitting in the middleware stack is important, IMHO, and I prefer Rich's first API suggestion.

SomaticIT:
I just updated this PR with two major changes:

First, a `sequence` function is now exported by the (new) module `@sveltejs/kit/hooks`. To create a `Handle` chain, you have to import the module:

```javascript
import { sequence } from '@sveltejs/kit/hooks';

export const handle = sequence(
  ({ request, resolve }) => {
    request.locals.first = true;
    return resolve(request);
  },
  ({ request, resolve }) => {
    request.locals.second = true;
    return resolve(request);
  },
  ({ request, resolve }) => {
    console.log(request.locals); // print { first: true, second: true }
    return resolve(request);
  }
);
```

Next, I added a test case `packages/kit/test/apps/hooks`. For some reason, I can't run it through `pnpm test` on my computer because of some incompatibility with `playwright`. I ran it manually and it seems to work well.

*Note: I also have an issue with import in my test. It seems that `vite` is not able to resolve `@sveltejs/kit/hooks`. I had to import it from the `dist` folder. Do you have an idea about this?*

SomaticIT:
> I don't understand the first suggestion. `resolve` computes the entire page HTML. Why would you be doing that three times? Or does it mean `next()` the first two and then `resolve()` the last one? I really wouldn't like that API at all where it means different things at different times

I think that `Handle` functions does not have to know if they are executed in a `sequence` or standalone. They should be interchangeable.

benbender:
> I don't understand the first suggestion. `resolve` computes the entire page HTML. Why would you be doing that three times? Or does it mean `next()` the first two and then `resolve()` the last one?

Yes, `resolve()` is actually `next()` in the first two cases. 

> I really wouldn't like that API at all where it means different things at different times
Simple solution: alias next() to resolve() as a third argument in the sequence. So you would have "best of both world". @SomaticIT what do you think?

Btw, I extracted `sequence()` for test-purposes and it works like charm! Well done, @SomaticIT!

SomaticIT:
@benbender 

Thank you! 

I like your suggestion of having a `next` method that is a simple alias to `resolve(request)`.
I just made a little improvement of this PR to allow this.

Now, when using the `sequence` utility, you can:
```javascript
import { sequence } from '@sveltejs/kit/hooks';

export const handle = sequence(
  ({ request, next }) => {
    request.locals.first = true;
    return next();
  },
  // or
  ({ request, resolve }) => {
    request.locals.second = true;
    return resolve(request);
  },
  // both are same
  ({ request, resolve }) => {
    console.log(request.locals); // print { first: true, second: true }
    return resolve(request);
  }
);
```

It allows best of both worlds: you can make a `Handle` that can work independantly or in a sequence by keeping the `resolve(request)` syntax or you can use the `next` shortcut when you known you are working in a sequence.

What do you think?

benbender:
@SomaticIT simply perfect! Exactly what I had in mind. 
Will update my "ripoff" of your code later. If there are any issues, I will report them here (even if I don't think so).

packages/kit/src/runtime/hooks.js
```diff
@@ -0,0 +1,24 @@
...
+		 */
+		function apply_handle(i) {
+			const handle = handles[i];
+			const resolve = i >= length - 1 ? super_resolve : () => apply_handle(i + 1);
+			const next = () => resolve(request);
+
+			return handle({ request, resolve, next });
+		}
+	};
+}
```
Rich-Harris, line 24: getting rid of `next` (see comment). also, the `resolve` function should take a `request` argument

```suggestion
/**
 * @param {...import('types/hooks').Handle} handlers
 * @returns {import('types/hooks').Handle}
 */
export function sequence(...handlers) {
	const length = handlers.length;
	if (!length) return ({ request, resolve }) => resolve(request);

	return ({ request, resolve }) => {
		return apply_handle(0, request);

		/**
		 * @param {number} i
		 * @param {import('types/hooks').ServerRequest} request
		 * @returns {import('types/helper').MaybePromise<import('types/hooks').ServerResponse>}
		 */
		function apply_handle(i, request) {
			const handle = handlers[i];

			return handle({
				request,
				resolve: i < length - 1 ? (request) => apply_handle(i + 1, request) : resolve
			});
		}
	};
}
```

packages/kit/test/apps/hooks/src/hooks.js
```diff
@@ -0,0 +1,39 @@
+// The following line does not work with vite resolution. Why?
+// import { sequence } from '@sveltejs/kit/hooks';
+import { sequence } from '../../../../dist/hooks';
```
Rich-Harris, line 3: better to test source code than build output, where possible

```suggestion
import { sequence } from '../../../../src/runtime/hooks';
```

packages/kit/test/apps/hooks/src/hooks.js
```diff
@@ -0,0 +1,39 @@
...
+
+/** @type {import('@sveltejs/kit').Handle} */
+export const handle = sequence(
+	({ request, next }) => {
+		request.locals.first = true;
+		return next();
```
Rich-Harris, line 9:
```suggestion
	({ request, resolve }) => {
		request.locals.first = true;
		return resolve(request);
```

packages/kit/types/hooks.d.ts
```diff
@@ -20,11 +20,21 @@ export interface GetSession<Locals = Record<string, any>, Session = any> {
 	(request: ServerRequest<Locals>): MaybePromise<Session>;
 }
 
+export interface HandleInput<Locals = Record<string, any>> {
+	request: ServerRequest<Locals>;
+	resolve: (request: ServerRequest<Locals>) => MaybePromise<ServerResponse>;
+}
+
```
Rich-Harris, line 27: we don't need separate `SequenceHandle` types if we get rid of `next` — let's just revert all this

```suggestion
```

packages/kit/types/hooks.d.ts
```diff
@@ -20,11 +20,21 @@ export interface GetSession<Locals = Record<string, any>, Session = any> {
...
 export interface Handle<Locals = Record<string, any>> {
-	(input: {
-		request: ServerRequest<Locals>;
-		resolve(request: ServerRequest<Locals>): MaybePromise<ServerResponse>;
-	}): MaybePromise<ServerResponse>;
+	(input: HandleInput<Locals>): MaybePromise<ServerResponse>;
```
Rich-Harris, line 29:
```suggestion
	(input: {
		request: ServerRequest<Locals>;
		resolve: (request: ServerRequest<Locals>) => MaybePromise<ServerResponse>;
	}): MaybePromise<ServerResponse>;
```

packages/kit/types/hooks.d.ts
```diff
@@ -20,11 +20,21 @@ export interface GetSession<Locals = Record<string, any>, Session = any> {
...
+}
+
+export interface SequenceHandleInput<Locals = Record<string, any>> extends HandleInput<Locals> {
+	next: () => MaybePromise<ServerResponse>;
+}
+
+export interface SequenceHandle<Locals = Record<string, any>> {
+	(input: SequenceHandleInput<Locals>): MaybePromise<ServerResponse>;
```
Rich-Harris, line 37:
```suggestion
```

Rich-Harris (changes requested):
Thank you! 

> I think middlewares should have the same signature as a classic `Handle` to allow them to be used standalone or as part of a sequence depending on your use case.

This logic makes sense to me. By extension, I'm not sure we need the `next` function — let's just use `resolve(request)` everywhere. The `resolve` functions should take a `request` argument, we shouldn't just rely on the original `request` being mutated

packages/kit/types/hooks.d.ts
SomaticIT: Deleted the commit which added the `next` function.

packages/kit/types/hooks.d.ts
SomaticIT: Deleted the commit which added the `next` function.

packages/kit/types/hooks.d.ts
SomaticIT: Deleted the commit which added the `next` function.

packages/kit/test/apps/hooks/src/hooks.js
SomaticIT: Deleted the commit which added the `next` function.

packages/kit/test/apps/hooks/src/hooks.js
SomaticIT: Applied

packages/kit/src/runtime/hooks.js
SomaticIT: You're totally right, `request` should always be passed to the `resolve` function. The `request` may be mutated or totally cloned by a `Handle`.

I made the change in a new commit.

SomaticIT:
@Rich-Harris, thank you for your review, you're totally right. `request` should always be passed to `resolve`.
I also removed the `next` shortcut.

Rich-Harris:
thank you! merged via #2222

SomaticIT:
Thank you for merging this PR.
I'm happy to have contributed to this wonderful project!

PR #1954, opened 2021-07-19
fix: include esbuild config in adapter type definition

GrygrFlzr:
- Added missing type definitions to adapters.
- Fix adapter-node requiring esbuild to be defined if any options were specified.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-cloudflare-workers/index.js
+++ b/packages/adapter-cloudflare-workers/index.js
@@ -11,9 +11,9 @@ import { fileURLToPath } from 'url';
 /**
  * @param {{
  *   esbuild?: (defaultOptions: BuildOptions) => Promise<BuildOptions> | BuildOptions;
- * }} options
+ * }} [options]
  **/
-export default function (options = { esbuild: (opts) => opts }) {
+export default function (options) {
 	/** @type {import('@sveltejs/kit').Adapter} */
 	const adapter = {
 		name: '@sveltejs/adapter-cloudflare-workers',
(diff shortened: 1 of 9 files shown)


dummdidumm:
Fixes #1958 I think

packages/adapter-cloudflare-workers/index.js
```diff
@@ -11,9 +11,9 @@ import { fileURLToPath } from 'url';
 /**
  * @param {{
  *   esbuild?: (defaultOptions: BuildOptions) => Promise<BuildOptions> | BuildOptions;
- * }} options
+ * }} [options]
```
jthegedus, line 14: Is this an intentional change? Why only on some adapters?

GrygrFlzr: - `adapter-cloudflare-workers`, `adapter-vercel`, `adapter-netlify`, and `adapter-node` all use esbuild, so this option should be available but not made mandatory.
- `adapter-begin` is unmaintained
- `adapter-static` doesn't use esbuild, so it wasn't affected.

If you're referring to the specific line, that's JSDoc's way of specifying an optional parameter so that you can do something like `adapter: adapterCfw()` without having to pass in an empty object.

jthegedus: >If you're referring to the specific line, that's JSDoc's way of specifying an optional parameter so that you can do something like `adapter: adapterCfw()` without having to pass in an empty object.

I was referring to the specific line, thanks for clarifying

PR #1992, opened 2021-07-22
fix: update vite-plugin-svelte and svelte to fix hmr issue with svelte 3.40

dominikg:
svelte 3.40.0 introduced a new internal helper `insert_hydration` that has to be used by svelte-hmr instead of `insert`.
The [fix](https://github.com/sveltejs/svelte-hmr/pull/37) was released in svelte-hmr 0.14.7, which vite-plugin-svelte 1.0.0-next.14 depends on.

### Tests

### Changesets

### DIFF
--- a/.changeset/pretty-flowers-repair.md
+++ b/.changeset/pretty-flowers-repair.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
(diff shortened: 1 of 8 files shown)


dominikg:
Note svelte version is synced to 3.40 on all packages except for the peer dep of kit and the template in create-svelte to allow using older versions

PR #1999, opened 2021-07-23
Use @ts-expect-error over @ts-ignore

utkarshkukreti:
Rationale: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#ts-ignore-or-ts-expect-error

> Pick ts-expect-error if:
> * you’re writing test code where you actually want the type system to error on an operation
> * you expect a fix to be coming in fairly quickly and you just need a quick workaround
> * you’re in a reasonably-sized project with a proactive team that wants to remove suppression comments as soon affected code is valid again

There were a few unnecessary @ts-ignore which this change pointed out (which is why the diff has more removals than insertions).

cc @benmccann, #1998

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/adapt/test/index.js
+++ b/packages/kit/src/core/adapt/test/index.js
@@ -30,7 +30,7 @@ suite('copy files', () => {
 	/** @type {import('types/config').ValidatedConfig} */
 	const config = {
 		kit: {
-			// @ts-ignore
+			// @ts-expect-error
 			files: {
 				assets: join(__dirname, 'fixtures/basic/static')
 			},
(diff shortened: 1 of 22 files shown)


benmccann:
Thanks! I see there's still another 30 `@ts-ignore` comments (`grep -r ts-ignore packages/kit/src/`). Should we switch them all over?

utkarshkukreti:
`grep -r ts-ignore packages/kit/src/` prints nothing for me and neither does `git grep ts-ignore`. Maybe `grep` is looking into some build artifacts present on your system?

benmccann:
Ooops. I was on the wrong branch :smile:

PR #2006, opened 2021-07-24
[docs] minor wording improvement in migration guide

Ncookiez:
Simple typo fixed.


benmccann:
I don't think this is necessarily wrong or a typo, but agree it reads better this way, so I'll go ahead and merge it

Issue #1940, opened 2021-07-18
Configurable error handling for prerender


happycollision:
### Describe the problem

You can reference #1935 for some context. I was pretty frustrated that my build was failing on the very first 404 it encountered once the adapter took over. I had no idea that there was an option to change that behavior. Yes, I eventually found it, but I looked several times in the wrong location.

There are several reasons I didn't find the option I was looking for:

1. I assumed that since the failure was during the adapter phase of the build that I should be looking at the adapter instead of the kit config.
2. I had no idea that the failure was happening during _prerender_, because I just wasn't familiar enough with the steps involved with building. 
    1. And `--verbose` does not indicate you are prerendering either
    2. Upon closer inspection of the stack trace, "prerender" appears, but so does "adapt" and "visit", with "visit" being the most prominent in reference to the error. 
3. `prerender.force` seems—to me—to suggest an override, not of a _failure_, but of _some other option_ that perhaps a plugin is trying to set or something.

When initially looking through the docs, I saw the section describing `prerender` options. I even read `crawl` and `enabled` carefully but completely skipped over `force` for the reason I mentioned above.

Eventually I thought to search for the word "fail" in the docs (luckily only 6 occurrences) and found that the description I _glanced right past_ had the answer.

Yes, It is in the docs. Yes, it clearly states what the option does. But its name threw me off the trail.

### More log information
I think that, in general, more indication of what is happening during the build might have led me to pay closer attention to the option group I needed. If there was more indication of what was happening in the terminal, perhaps I'd have seen that, "Oh, we are in the prerendering phase" when it failed.

I really thought that the adapter was totally separate from the config for the rest of the app, but apparently not!

### Different naming
In addition to that, a different name for the `force` option might be better. One of these?

- `continueOnError`
- `continueOn404`
- `bailOnError` (inverse)
- `bailOn404` (inverse)

#### Many toggles
Perhaps—if someone wants to get crazy—more granular control? I assume that there are other errors besides a missing image link that can fail a prerender. It might be nice if you could specify...

```
{
  prerender: {
    failBuildOn: {
      missingMediaLink: false,
      missingInternalPageLink: true,
      missingExternalPageLink: false,
      // ... etc
    },
  },
}
```

But I am guessing that is WAY overboard. And probably presents a fairly large surface area for bugs.

#### Escape hatch
Though you could instead do something like this
```tsx
interface KitOptions {
  prerender: {
    onError: 'fail' | 'continue' | (err: Reason) => boolean
[...]

PR #2007, opened 2021-07-25
Change config `prerender.force` to `prerender.onError`

happycollision:
fixes sveltejs/kit#1940

Essentially, we are doing this:

```diff
kit: {
  prerender: {
    crawl: boolean;
    enabled: boolean;
-   force: boolean;
+   onError: 'fail' | 'continue' | ((details: RequestDetails) => void | never);
    pages: string[];
  }
}
```

- Changing from `force` to `onError` gives a somewhat more descriptive
  name to the option.
- Using the strings "fail" and "continue" further clarifies what the
  option does.
- Providing your own function to deal with an error allows for
  fine-tuning which errors fail the build and which do not.

The original ticket suggested that the custom function return a boolean,
but after seeing the implementation of the error handler in svelteKit, I
thought it more fitting to just allow it the exact same API: throw if
you want the build to fail.

## Teaching the new option via error message:

This PR also adds a failing option check for using the old `force` property, and gives you the proper value to use in its place as well. This should help when people version bump to the latest build.

CleanShot 2021-07-24 at 20 53 16@2x

CleanShot 2021-07-24 at 22 43 13@2x

#### Throw to fail

CleanShot 2021-07-26 at 20 53 51@2x

#### Anything but throwing will continue the build

CleanShot 2021-07-26 at 21 00 43@2x

#### fail
CleanShot 2021-07-24 at 20 59 05@2x

#### continue
CleanShot 2021-07-24 at 20 58 22@2x

### Tests

### Changesets

### DIFF
--- a/documentation/docs/14-configuration.md
+++ b/documentation/docs/14-configuration.md
@@ -48,7 +48,7 @@ const config = {
 		prerender: {
 			crawl: true,
 			enabled: true,
-			force: false,
+			onError: 'fail',
 			pages: ['*']
 		},
 		router: true,
@@ -148,7 +148,30 @@ See [Prerendering](#ssr-and-javascript-prerender). An object containing zero or 
 
 - `crawl` — determines whether SvelteKit should find pages to prerender by following links from the seed page(s)
 - `enabled` — set to `false` to disable prerendering altogether
-- `force` — if `true`, a page that fails to render will _not_ cause the entire build to fail
+- `onError`
+
+  - `'fail'` — (default) fails the build when a routing error is encountered when following a link
+  - `'continue'` — allows the build to continue, despite routing errors
+  - `function` — custom error handler allowing you to log, `throw` and fail the build, or take other action of your choosing based on the details of the crawl
+
+    ```ts
+    /** @type {import('@sveltejs/kit').PrerenderErrorHandler} */
+    const handleError = ({ status, path, referrer, referenceType }) => {
+    	if (path.startsWith('/blog')) throw new Error('Missing a blog page!');
+    	console.warn(`${status} ${path}${referrer ? ` (${referenceType} from ${referrer})` : ''}`);
+    };
+
+    export default {
+    	kit: {
+    		adapter: static(),
+    		target: '#svelte',
+    		prerender: {
+    			onError: handleError
+    		}
+    	}
+    };
+    ```
+
 - `pages` — an array of pages to prerender, or start crawling from (if `crawl: true`). The `*` string includes all non-dynamic routes (i.e. pages with no `[parameters]` )
 
 ### router
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -5,6 +5,12 @@ import { mkdirp } from '../filesystem/index.js';
 import { __fetch_polyfill } from '../../install-fetch.js';
 import { SVELTE_KIT } from '../constants.js';
 
+/**
+ * @typedef {import('types/config').PrerenderErrorHandler} PrerenderErrorHandler
+ * @typedef {import('types/config').PrerenderOnErrorValue} OnError
+ * @typedef {import('types/internal').Logger} Logger
+ */
+
 /** @param {string} html */
 function clean_html(html) {
 	return html
@@ -45,13 +51,34 @@ function get_srcset_urls(attrs) {
 	return results;
 }
 
+/** @type {(errorDetails: Parameters<PrerenderErrorHandler>[0] ) => string} */
+function errorDetailsToString({ status, path, referrer, referenceType }) {
+	return `${status} ${path}${referrer ? ` (${referenceType} from ${referrer})` : ''}`;
+}
+
+/** @type {(log: Logger, onError: OnError) => PrerenderErrorHandler} */
+function chooseErrorHandler(log, onError) {
+	switch (onError) {
+		case 'continue':
+			return (errorDetails) => {
+				log.error(errorDetailsToString(errorDetails));
+			};
+		case 'fail':
+			return (errorDetails) => {
+				throw new Error(errorDetailsToString(errorDetails));
+			};
+		default:
+			return onError;
+	}
+}
+
 const OK = 2;
 const REDIRECT = 3;
 
 /** @param {{
  *   cwd: string;
  *   out: string;
- *   log: import('types/internal').Logger;
+ *   log: Logger;
  *   config: import('types/config').ValidatedConfig;
  *   build_data: import('types/internal').BuildData;
  *   fallback?: string;
@@ -75,14 +102,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 		read: (file) => readFileSync(join(config.kit.files.assets, file))
 	});
 
-	/** @type {(status: number, path: string, parent: string | null, verb: string) => void} */
-	const error = config.kit.prerender.force
-		? (status, path, parent, verb) => {
-				log.error(`${status} ${path}${parent ? ` (${verb} from ${parent})` : ''}`);
-		  }
-		: (status, path, parent, verb) => {
-				throw new Error(`${status} ${path}${parent ? ` (${verb} from ${parent})` : ''}`);
-		  };
+	const error = chooseErrorHandler(log, config.kit.prerender.onError);
 
 	const files = new Set([...build_data.static, ...build_data.client]);
 
@@ -107,9 +127,9 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 
 	/**
 	 * @param {string} path
-	 * @param {string?} parent
+	 * @param {string?} referrer
 	 */
-	async function visit(path, parent) {
+	async function visit(path, referrer) {
 		path = normalize(path);
 
 		if (seen.has(path)) return;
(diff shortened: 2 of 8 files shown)


packages/kit/src/core/adapt/prerender.js
```diff
@@ -75,14 +75,31 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
...
-	const error = config.kit.prerender.force
-		? (status, path, parent, verb) => {
-				log.error(`${status} ${path}${parent ? ` (${verb} from ${parent})` : ''}`);
-		  }
-		: (status, path, parent, verb) => {
-				throw new Error(`${status} ${path}${parent ? ` (${verb} from ${parent})` : ''}`);
-		  };
+	/** @typedef {((status: number, path: string, parent: string | null, verb: string) => void | never)} InternalErrorHandler */
+	/** @type {InternalErrorHandler} */
+	const error = (function () {
```
happycollision, line 80: I am happy to rewrite this as something other than an IIFE, but it seemed a decent way to colocate everything.

packages/kit/src/core/config/index.spec.js
```diff
@@ -50,7 +50,9 @@ test('fills in defaults', () => {
 			prerender: {
 				crawl: true,
 				enabled: true,
-				force: false,
+				// This can be removed entirely after we remove the validation error associated with using `force`
```
benmccann, line 53:
```suggestion
				// TODO: remove this for 1.0 release
```

packages/kit/src/core/config/options.js
```diff
@@ -121,7 +121,32 @@ const options = {
 				children: {
 					crawl: expect_boolean(true),
 					enabled: expect_boolean(true),
-					force: expect_boolean(false),
+					force: {
```
benmccann, line 124:
```suggestion
				// TODO: remove this for 1.0 release
				force: {
```

packages/kit/types/config.d.ts
```diff
@@ -78,6 +78,13 @@ export type Config = {
 	preprocess?: any;
 };
 
+export type PrerenderErrorHandler = (errorDetails: {
+	status: number;
+	path: string;
+	parent: string | null;
```
benmccann, line 84: it's not clear to me from name alone - what is `parent`? should we rename to `referer`? would that be the same? or if the current name is the best, let's add some docs to explain what it is

packages/kit/src/core/adapt/prerender.js
benmccann: could we just make it a function that lives outside the `prerender` function?

happycollision: Happy to.

packages/kit/types/config.d.ts
happycollision: I think `referrer` is better, yeah.

happycollision:
All set. Thanks @benmccann!

documentation/docs/14-configuration.md
```diff
@@ -148,7 +148,30 @@ See [Prerendering](#ssr-and-javascript-prerender). An object containing zero or
 
 - `crawl` — determines whether SvelteKit should find pages to prerender by following links from the seed page(s)
 - `enabled` — set to `false` to disable prerendering altogether
-- `force` — if `true`, a page that fails to render will _not_ cause the entire build to fail
+- `onError`
+
+  - `"fail"` — (default) fails the build when a routing error is encountered when following a link
```
benmccann, line 153:
```suggestion
  - `'fail'` — (default) fails the build when a routing error is encountered when following a link
```

documentation/docs/14-configuration.md
```diff
@@ -148,7 +148,30 @@ See [Prerendering](#ssr-and-javascript-prerender). An object containing zero or
...
+  - `"continue"` — allows the build to continue, despite routing errors
```
benmccann, line 154:
```suggestion
  - `'continue'` — logs the error and allows the build to continue despite any routing errors
```

documentation/docs/14-configuration.md
```diff
@@ -148,7 +148,30 @@ See [Prerendering](#ssr-and-javascript-prerender). An object containing zero or
...
+  - function — you can use the `errorDetails` to determine whether or not to `throw`, thereby failing the build.
+
+    ```ts
+    /** @type {import('@sveltejs/kit').PrerenderErrorHandler} */
+    const handleError = ({ status, path, parent, verb }) => {
```
benmccann, line 159: need to change `parent` to `referer` here and below

documentation/docs/14-configuration.md
```diff
@@ -148,7 +148,30 @@ See [Prerendering](#ssr-and-javascript-prerender). An object containing zero or
 
 - `crawl` — determines whether SvelteKit should find pages to prerender by following links from the seed page(s)
 - `enabled` — set to `false` to disable prerendering altogether
-- `force` — if `true`, a page that fails to render will _not_ cause the entire build to fail
+- `onError`
+
+  - `"fail"` — (default) fails the build when a routing error is encountered when following a link
+  - `"continue"` — allows the build to continue, despite routing errors
+  - function — you can use the `errorDetails` to determine whether or not to `throw`, thereby failing the build.
```
benmccann, line 155: I don't see anything defining `errorDetails`, so let's remove the reference to it
```suggestion
  - `function` — custom error handler allowing you to log, `throw` and fail the build, or take other action of your choosing based on the details of the crawl
```

documentation/docs/14-configuration.md
benmccann: Also, I think that the ending brace is missing

packages/kit/src/core/adapt/prerender.js
```diff
@@ -45,6 +45,32 @@ function get_srcset_urls(attrs) {
 	return results;
 }
 
+/** @typedef {((status: number, path: string, parent: string | null, verb: string) => void | never)} InternalErrorHandler */
```
benmccann, line 48: Can we call this `PrerenderErrorHandler` or `ErrorHandler` instead? It's not just internal in my mind since the API is exposed for the user to provide it

benmccann: Oh, actually, I see `PrerenderErrorHandler` already defined in `types/config`. Can we just reuse that existing type definition?

benmccann:
Lgtm. I left a handful of minor comments. I really like the idea of using a function to give the user more control and stole the idea for another PR https://github.com/sveltejs/kit/pull/2008

packages/kit/src/core/adapt/prerender.js
happycollision: Actually, it is a slightly different type. The one defined in config is the public version, which gathers all the arguments into a single object, instead of four separate parameters. I made the distinction so that it matched the aesthetic established for the `load` function.

I’ll just make all references to an error-like function in this file just use the single argument version.

happycollision:
@benmccann , in this final pass, I took a little more liberty with the names of those properties (`verb` became `referenceType`) and also with the call sites for the `error` function (as I mentioned I would in my response to your request about types.

Additionally, on the types front, I created local references for types that are used more than once (or are especially long) near the top of the file. They can be inlined if that is more desirable, but I feel that things are more readable this way.

packages/kit/src/core/adapt/prerender.js
```diff
@@ -5,6 +5,12 @@ import { mkdirp } from '../filesystem/index.js';
 import { __fetch_polyfill } from '../../install-fetch.js';
 import { SVELTE_KIT } from '../constants.js';
 
+/**
+ * @typedef {import('types/config').PrerenderErrorHandler} PrerenderErrorHandler
+ * @typedef {import('types/config').ValidatedConfig['kit']['prerender']['onError']} OnErrorConfig
+ * @typedef {import('types/internal').Logger} Logger
+ */
+
```
happycollision, line 13: Logger is used twice, PrerenderErrorHandler is used a few times, and OnErrorConfig is just a really long line which muddied up the definition it was being used in.

packages/kit/src/core/adapt/prerender.js
```diff
@@ -183,7 +203,12 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 				if (response_type === OK) {
 					log.info(`${result.status} ${dependency_path}`);
 				} else {
-					error(result.status, dependency_path, path, 'fetched');
+					error({
+						status: result.status,
+						path: dependency_path,
+						referrer: path,
+						referenceType: 'fetched'
+					});
```
happycollision, line 211: Changing this and the one at line 185 simplifies how many function signatures we are juggling for dealing with errors (from 2 very similar down to 1). Now, all of them take one argument.

packages/kit/src/core/adapt/prerender.js
```diff
@@ -5,6 +5,12 @@ import { mkdirp } from '../filesystem/index.js';
 import { __fetch_polyfill } from '../../install-fetch.js';
 import { SVELTE_KIT } from '../constants.js';
 
+/**
+ * @typedef {import('types/config').PrerenderErrorHandler} PrerenderErrorHandler
+ * @typedef {import('types/config').ValidatedConfig['kit']['prerender']['onError']} OnErrorConfig
```
benmccann, line 10: I think a nicer way to do this line would be to create a value in the `.d.ts` file that's exported from there and then imported here:

`export type OnErrorValue = 'fail' | 'continue' | PrerenderErrorHandler`

happycollision: Agreed. Basically every time I try to not presume anything, and just work with what is available, I find I just should have made the slightly bigger change. 😆

happycollision:
@benmccann Thanks for all the attention on this PR. I am getting a sense of how much you all are willing to see changed from outside contributors now.

And I am glad that you like the "escape hatch" pattern as I have come to call it. More often than not, I find it cuts the tension between simple abstraction and edge cases.

PR #2010, opened 2021-07-25
[chore] prefer interfaces to types

benmccann:
I remember we had done this in Sapper. Makes sense to do it here too probably

### DIFF
--- a/.changeset/curvy-dots-join.md
+++ b/.changeset/curvy-dots-join.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[chore] prefer interfaces to types
--- a/packages/kit/test/types.d.ts
+++ b/packages/kit/test/types.d.ts
@@ -7,7 +7,7 @@ import { RequestInfo, RequestInit, Response as NodeFetchResponse } from 'node-fe
 // seems like that's no longer an issue? in which case we don't need
 // to wrap all these methods
 
-export type TestContext = {
+export interface TestContext {
 	base: string;
 	page: Page;
 	pages: {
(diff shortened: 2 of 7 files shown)


benbender:
@benmccann The big difference here is that interfaces can be extended while types can't. So mostly user-customizable interfaces and internal types is mostly the right rule of thumb!

benmccann:
@ignatiusmb you might be interested in this PR too given all the TypeScript stuff you've done. If it looks good to you or I don't hear anything I'll probably merge it in the next day or so

ignatiusmb (commented):
LGTM 👍 

Following a heuristic, we can prefer to use `interface` until we need to use features from `type`. Extendibility of an interfaces

jthegedus:
I'm certainly no TS expert, but as I understand the only remaining difference in latest TS is that `interface` supports implicit open typing and merging whereas `type` will error - https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

Seems to me `type` would then be less error prone as extending the type def would always need to be explicit.

Again, no TS expert.

JeanJPNM:
I think interfaces give some flexibility when it comes to customization, being able to extend types without having to create new ones can be useful in some situations.

jthegedus:
> I think interfaces give some flexibility when it comes to customization, being able to extend types without having to create new ones can be useful in some situations.

For sure, in some situations. Just seems like a huge footgun. I see this as changing from `const` to `var` because extending a `const` value is useful sometimes. Seems overkill to make all `const` values `var`s.

JeanJPNM:
Yeah, but you can only extend an interface by redeclaring it (sort of), and if people take the time to redeclare an interface they usually know what they are doing

Issue #1997, opened 2021-07-22
Cannot use interfaces to type endpoint outputs


JeanJPNM:
### Describe the bug

When setting up and endpoint with typescript, it is not possible to have interfaces as endpoint outputs

### Reproduction

```ts
// can be any endpoint
import type { RequestHandler } from '@sveltejs/kit';
type GetOutput = {
  message: string;
};
interface PostOutput {
  message: string;
}
type PutOutput = PostOutput;

// works
export const get: RequestHandler<unknown, unknown, GetOutput> = async () => {
  return {
    status: 200,
    body: {
      message: 'hello world',
    },
  };
};

// complains
export const post: RequestHandler<unknown, unknown, PostOutput> = async () => {
  return {
    status: 200,
    body: {
      message: 'hello world',
    },
  };
};

// complains
export const put: RequestHandler<unknown, unknown, PutOutput> = async () => {
  return {
    status: 200,
    body: {
      message: 'hello world',
    },
  };
};
```

### Logs

```shell
src/routes/demo.ts:20:53 - error TS2344: Type 'PostOutput' does not satisfy the constraint 'DefaultBody'.
  Type 'PostOutput' is not assignable to type '{ [x: string]: JSONValue; }'.
    Index signature is missing in type 'PostOutput'.

20 export const post: RequestHandler<unknown, unknown, PostOutput> = async () => {
                                                       ~~~~~~~~~~

src/routes/demo.ts:30:52 - error TS2344: Type 'PostOutput' does not satisfy the constraint 'DefaultBody'.

30 export const put: RequestHandler<unknown, unknown, PutOutput> = async () => {
                                                      ~~~~~~~~~


Found 2 errors.
```

### System Info

```shell
System:
    OS: Windows 10 10.0.19043
    CPU: (4) x64 Intel(R) Core(TM) i5-3230M CPU @ 2.60GHz   
    Memory: 375.01 MB / 3.89 GB
  Binaries:
    Node: 14.14.0 - C:\Program Files\nodejs\node.EXE        
    npm: 6.14.8 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 92.0.4515.107
    Edge: Spartan (44.19041.1023.0), Chromium (91.0.864.71) 
    Internet Explorer: 11.0.19041.1
  npmPackages:
    @sveltejs/adapter-node: ^1.0.0-next.35 => 1.0.0-next.35
[...]

PR #2018, opened 2021-07-26
[chore] convert remaining type aliases

ignatiusmb:
Convert remaining public type aliases from #2010

### DIFF
--- a/packages/kit/types/helper.d.ts
+++ b/packages/kit/types/helper.d.ts
@@ -26,7 +26,8 @@ export type Location<Params extends Record<string, string> = Record<string, stri
 	query: URLSearchParams;
 };
 
-export type MaybePromise<T> = T | Promise<T>;
+export type Index<T = any> = Record<string, T>;
 export type InferValue<T, Key extends keyof T, Default> = T extends Record<Key, infer Val>
 	? Val
 	: Default;
+export type MaybePromise<T> = T | Promise<T>;
--- a/packages/kit/types/helper.d.ts
+++ b/packages/kit/types/helper.d.ts
@@ -26,8 +26,8 @@ export type Location<Params extends Record<string, string> = Record<string, stri
 	query: URLSearchParams;
 };
 
-export type Index<T = any> = Record<string, T>;
 export type InferValue<T, Key extends keyof T, Default> = T extends Record<Key, infer Val>
 	? Val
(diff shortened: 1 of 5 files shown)


ignatiusmb:
It doesn't work with JSDoc for some reason, I'll come back tomorrow and try take a look again

benmccann:
Cool. Btw, not sure if you have any ideas on https://github.com/sveltejs/kit/issues/1997. I was a bit stumped on that one (also TypeScript-related)

ignatiusmb:
It looks like JSDoc parser isn't as advanced as using TypeScript directly (who would've thought), resulting in some loses of its inferencing ability, most likely on the complex and nested types. Filling in the default values seems to do the trick.

packages/kit/types/helper.d.ts
```diff
@@ -26,7 +26,8 @@ export type Location<Params extends Record<string, string> = Record<string, stri
 	query: URLSearchParams;
 };
 
-export type MaybePromise<T> = T | Promise<T>;
+export type Index<T = any> = Record<string, T>;
```
benmccann, line 29: I'm not totally in love with the name `Index` though having a slightly hard time coming up with a better name. I think `Indexed` sounds a bit better but might be confused with something like [indexed access types](https://www.typescriptlang.org/docs/handbook/2/indexed-access-types.html). I sort of wonder if it's even worth adding `Index`. I do like that it's shorter, but on the other hand I think it's less clear what it is and it adds a layer of indirection in reading the code

ignatiusmb: That's what `Record<K, T>` is for, and I think it's worth it, especially when used multiple times like in `page.d.ts`. It does abstracts the code a bit, but anyone reading would only need to see it once to know what it means.

We can rename to make it clearer too. In that case, maybe `Indexable`, `Indexed`, or `IndexSig`?

benmccann: Maybe `Rec`? It makes it a bit clearer that it's related to a record. Things with index sound a bit wrong to me also because you can create, for example, a database index on non-string fields. So if index doesn't imply `string`, why use a term unrelated to `Record` in that case?

ignatiusmb: Sure, it's just a helper type after all

PR #2019, opened 2021-07-26
[chore] typecheck example

benmccann:
This fixes some bugs. It's not actually enforced yet because I think it was trying to look at the unpackaged types from the monorepo, which was causing some issues. Maybe someone can figure that out

FYI @ignatiusmb, another TypeScript one :smile:


ignatiusmb (approved):
Typing looks good, minor nit on spacings between code

benmccann:
> minor nit on spacings between code

I don't see any other comments. Maybe you forgot to hit send on them?

ignatiusmb (commented):
It was pretty late at night yesterday, here you go :)

PR #2027, opened 2021-07-28
Read file for SSL (certs and keys) and add pfx support

### LINKED ISSUE #1659: Expose `config.kit.server.https` so that `svelte-kit preview --https` can use user-defined cert
Expose `config.kit.server.https` so that `svelte-kit preview --https` can use user-defined cert.

Not tested, but suggested fix:
```patch
diff --git a/packages/kit/src/core/load_config/options.js b/packages/kit/src/core/load_config/options.js
index c763465..ba8eb88 100644
--- a/packages/kit/src/core/load_config/options.js
+++ b/packages/kit/src/core/load_config/options.js
@@ -139,6 +139,19 @@ const options = {
 
                        router: expect_boolean(true),
 
+                       server: {
+                               type: 'leaf',
+                               default: null,
+                               validate: (option, keypath) => {
+
+                                       if (!(option && option.https && option.https.cert && option.https.key)) {
+                                               throw new Error(`${keypath} server https cert & key not defined`);
+                                       }
+
+                                       return option;
+                               }
+                       },
+
                        ssr: expect_boolean(true),
 
                        target: expect_string(null),
```

However I've tested the patched built version:
```patch
diff --git a/node_modules/@sveltejs/kit/dist/cli.js b/node_modules/@sveltejs/kit/dist/cli.js
index 9914595..e829be3 100644
--- a/node_modules/@sveltejs/kit/dist/cli.js
+++ b/node_modules/@sveltejs/kit/dist/cli.js
@@ -277,6 +277,19 @@ const options = {
 
 			router: expect_boolean(true),
 
+                       server: {
+                               type: 'leaf',
+                               default: null,
+                               validate: (option, keypath) => {
+
+                                       if (!(option && option.https && option.https.cert && option.https.key)) {
+                                               throw new Error(`${keypath} server https cert & key not defined`);
+                                       }
+
+                                       return option;
+                               }
+                       },
+
 			ssr: expect_boolean(true),
 
 			target: expect_string(null),
```

N/A

Good to have, to be able to use existing self-signed certs during development.

JBusillo:
Fixes #1659.  

This replaces PR #1662 due to my inexperience with Github.  

The dev server passes the vite options (config.kit.vite) when creating the http(s) server.  This is functioning correctly.
The preview server passes the kit options (config.kit) when creating the http(s) server.  This is not functioning correctly.

In preview mode, when the user specifies a cert/key pair, the cert/key pair is ignored, since the logic is looking for the option in the wrong place.  Instead, Kit will generate its own cert/key pair.  You can observe this by examining the certificate details within the browser.

This change aligns where the dev and preview modes look for a user-specified cert/key pair.

For reference, here is an example of a user-specified cert/key pair in `svelte.config.js`
```
import fs from "fs";
const config = {
  kit: {
    target: "#svelte",
    vite: {
      server: {
        https: {
          key: fs.readFileSync("example.key"),
          cert: fs.readFileSync("example.crt"),
        },
      },
    },
  },
};
export default config;
```

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/dev/index.js
+++ b/packages/kit/src/core/dev/index.js
@@ -91,8 +91,8 @@ class Watcher extends EventEmitter {
 	async init_server() {
 		if (!this.manifest) throw new Error('Must call init() before init_server()');
 
-		/** @type {any} */
-		const user_config = (this.config.kit.vite && this.config.kit.vite()) || {};
+		/** @type {import('types/config').ViteConfig} ViteConfig */
+		const vite_config = (this.config.kit.vite && this.config.kit.vite()) || {};
 
 		const default_config = {
 			server: {
@@ -105,12 +105,12 @@ class Watcher extends EventEmitter {
 		/** @type {(req: import("http").IncomingMessage, res: import("http").ServerResponse) => void} */
 		let handler = (req, res) => {};
 
-		this.server = await get_server(this.https, user_config, (req, res) => handler(req, res));
+		this.server = await get_server(this.https, vite_config, (req, res) => handler(req, res));
 
-		const alias = user_config.resolve && user_config.resolve.alias;
+		const alias = vite_config.resolve && vite_config.resolve.alias;
 
 		// don't warn on overriding defaults
-		const [modified_user_config] = deep_merge(default_config, user_config);
+		const [modified_user_config] = deep_merge(default_config, vite_config);
 
 		/** @type {[any, string[]]} */
 		const [merged_config, conflicts] = deep_merge(modified_user_config, {
@@ -153,7 +153,8 @@ class Watcher extends EventEmitter {
 				entries: []
 			},
 			ssr: {
-				noExternal: get_no_external(this.cwd, user_config.ssr && user_config.ssr.noExternal)
+				// @ts-ignore // ssr not defined in vite config.ts
+				noExternal: get_no_external(this.cwd, vite_config.ssr && vite_config.ssr.noExternal)
 			}
 		});
 
--- a/.changeset/big-countries-pump.md
+++ b/.changeset/big-countries-pump.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Add public API to let adapters update .gitignore
--- a/.changeset/rare-pots-invite.md
+++ b/.changeset/rare-pots-invite.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/kit": patch
+---
+
+Don't check external links on prerender
--- a/.changeset/shy-forks-grin.md
+++ b/.changeset/shy-forks-grin.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Honor HTTPS options in Preview and Dev
--- a/documentation/docs/10-adapters.md
+++ b/documentation/docs/10-adapters.md
@@ -31,6 +31,15 @@ export default {
 };
 ```
 
+Some adapters may modify your project's `.gitignore` to include their build output. In case you don't want those patterns included you can comment them out: 
+
+```diff
+.svelte-kit
+.env
+
+- build
++ # build
+```
 A variety of official adapters exist for serverless platforms...
 
 - [`adapter-cloudflare-workers`](https://github.com/sveltejs/kit/tree/master/packages/adapter-cloudflare-workers) — for [Cloudflare Workers](https://developers.cloudflare.com/workers/)
--- a/documentation/docs/80-adapter-api.md
+++ b/documentation/docs/80-adapter-api.md
@@ -26,6 +26,7 @@ The types for `Adapter` and its parameters are available in [types/config.d.ts](
 Within the `adapt` method, there are a number of things that an adapter should do:
 
 - Clear out the build directory
+- Call `utils.update_ignores` to ignore build output in existing `.gitignore` files at the location of `svelte.config.js`
 - Output code that:
   - Calls `init`
   - Converts from the platform's request to a [SvelteKit request](#hooks-handle), calls `render`, and converts from a [SvelteKit response](#hooks-handle) to the platform's
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
 		"eslint-plugin-svelte3": "^3.2.0",
 		"playwright-chromium": "^1.10.0",
 		"prettier": "2.2.1",
-		"rollup": "^2.47.0",
+		"rollup": "^2.55.0",
 		"typescript": "^4.3.5"
 	},
 	"type": "module"
--- a/packages/adapter-begin/package.json
+++ b/packages/adapter-begin/package.json
@@ -20,7 +20,7 @@
 	},
 	"devDependencies": {
 		"@sveltejs/kit": "workspace:*",
-		"rollup": "^2.47.0",
+		"rollup": "^2.55.0",
 		"sirv": "^1.0.12",
 		"typescript": "^4.3.5"
 	},
--- a/packages/adapter-cloudflare-workers/index.js
+++ b/packages/adapter-cloudflare-workers/index.js
@@ -25,6 +25,8 @@ export default function (options) {
 
 			const files = fileURLToPath(new URL('./files', import.meta.url));
 
+			utils.update_ignores({ patterns: [bucket, entrypoint] });
+
 			utils.rimraf(bucket);
 			utils.rimraf(entrypoint);
 
--- a/packages/adapter-netlify/index.js
+++ b/packages/adapter-netlify/index.js
@@ -26,6 +26,8 @@ export default function (options) {
 
 			const files = fileURLToPath(new URL('./files', import.meta.url));
 
+			utils.update_ignores({ patterns: [publish, functions] });
+
 			utils.log.minor('Generating serverless function...');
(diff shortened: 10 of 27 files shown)


JBusillo:
To assist in the review, the following document describes the files modified, changes made, strategy, etc.

<details>
  <summary>Change Summary</summary>

# SvelteKit PR 2027

## dev/index.js

* renamed user_config to vite_config
* added type checking for vite_config
* added a @ts-ignore for vite.ssr -- Hopefully Vite will add SSR typedefs in the near future, which will resolve this.

## server/index.js

The goal of the `get_server` and underling functions is to:
* If the `-H` (or `--https`) option is specified as a command line parameter and a cert/key pair (or pfx) is not specified, a self-signed certificate is created.
* If the `-H`  (or `--https`) option is specified as a command line parameter and a cert/key pair (or pfx) is specified, use the specified cert/key pair or pfx.
* Create the http(s) server, honoring SvelteKit CLI options and Vite options.

Note: Specifying custom self-signed certificates never worked in preview mode.  Now it does.

The Vite option `vite.server.https` allows any child option that is defined in `node:https.ServerOptions`.  'ca' and 'pfx' are options that are processed similarly to the 'cert' and 'key' options.  'ca' overrides the trusted CA certificates, and 'pfx' is a binary certificate format commonly used in Windows (which requires an addition 'passphrase' option).  Correctly handling these two options might prevent future issues/enhancements to be raised.

Vite allows the key, cert, ca, and pfx options to be specified as a filename, string, or buffer[].  For example, the "cert" can be specified as any of the following:
```
    cert: "./my-certificate.crt"   // filename          
    cert: fs.readFileSync("my-certificate.crt")   // returns a buffer[]
    cert: fs.readFileSync("my-certificate.crt", {encoding: 'utf-8'} )   // returns a string
    cert: "-----BEGIN CERTIFICATE-----\nMIIGEzCCA/ugAwIB ... -----END CERTIFICATE-----"  // string
```
I copied/adapted Vite's code in function readFileIfExists (see below).  

### function get_server(...)
`user_config` was renamed to `vite_config`.  Type checking was added.

If the `-H` cli flag is specified
* load http_options from vite.server.https
* resolve the ca/cert/key/pfx options -- i.e., if a filename was specified, set the option value to the file's content as a buffer[].  Otherwise, just return the passed value, which can be a string or a buffer[].
* if the 'pfx' option was specified, ignore the cert and key options. Pfx certificates already contain both the key and cert.
* if pfx/key/cert options aren't present, create a self-signed certificate.

If the `-H` cli flag is NOT specified, and if vite.server.https options are specified, Vite's HMR server gets confused, causing constant reloads from the client.  To resolve this, I just delete the vite.server.https leaf.

### function readFileIfExists(...)

This is Vite's code --- It's kludgy, but it works:

If a string is passed as a ca/cert/key/pfx option, it can either be a filename, or a string that holds the certificate contents.  A readFileSync is issued, and if successful, the resulting buffer[] is returned.  If it fails, the string is assumed to be the actual certificate contents, and is returned as-is.

If a non-string is passed, it's assumed that it's a buffer[] containing the actual certificate contents, and is returned as-is.

Yes, the @param type specification has an `any[]`.  The types for 'ca' and 'pfx' have different signatures, and want to include the crypto module, among other things.  The native Vite code has it specified as this.

## preview/index.js

* pass the vite configuration into get_server.  It was (incorrectly) passing kit configuration.  This is the 'crux' of the original issue #1659.

## types/config.d.ts

* re-export the imported ViteConfig type to support type checking in the above modules.

## Testing

My testing strategy included:

HTTPS testing:  All three variations of HTTPS options for cert/key/pfx -- filename, string and buffer[].  For "ca", I ensured (via debugger) that all three variations were formatted correctly when passed to `https.createServer`.

HTTP testing: With and without vite.server.https options (ensures that HMR doesn't reload constantly)

Verify the above in both dev and preview modes.  For dev mode, ensure that HMR continues to work properly. Also ensure that the proper certificate is used in all test cases (i.e., Kit's generated certificate or the user-specified certificate).

Cursory testing to ensure that --host and --port options are honored.

And, of course, pnpm format, pnpm lint, pnpm build and pnpm test.

</details>

packages/kit/src/core/dev/index.js
```diff
@@ -153,7 +153,8 @@ class Watcher extends EventEmitter {
 				entries: []
 			},
 			ssr: {
-				noExternal: get_no_external(this.cwd, user_config.ssr && user_config.ssr.noExternal)
+				// @ts-ignore // ssr not defined in vite config.ts
```
benmccann, line 156:
```suggestion
				// @ts-expect-error ssr is considered in beta, so not exposed by Vite
```

benmccann:
Thanks! There's kind of three different things being done here:
- Making sure the options get passed in preview
- Reading the file
- Adding support for `pfx`

To help separate the discussion, I cleaned up and committed the first part for you in https://github.com/sveltejs/kit/pull/2036

Regarding the second part, Rich had suggested maybe it'd be better to leave it the way it is? https://github.com/sveltejs/kit/pull/1798#issuecomment-872656524

I haven't looked at the `pfx` part yet, but we could do just a smaller PR for that

JBusillo:
If I understand correctly, you opened PR #2036 and only included code for the original issue -- preview using Kit options instead of (correctly) using Vite options.  These changes have been merged into master.

As far as 'reading the file: In non middleware mode, Vite allows the cert and key (as well as pfx and ca) to be specified using the filename, a string or a buffer.  I wanted to duplicate the functionality that Vite had so that middleware mode could use this same flexible (but somewhat ambiguous) specification.  However, going forward, it seems that Rich wants the cert and key values to be strings (such as the string returned by fs.readFileSync().  That's OK by me.

For the 'pfx' option:  If the key/cert options are not specified, Kit will populate the key/cert options with a self-signed certificate.  I'm not sure what node will do, e.g., throw an error or let one option take precedence over the other.  I'll test this out later today.  If the 'pfx' option takes precedence, no  changes may be needed.  Otherwise I'll open an issue for that.

So, is this PR 'dead'?  Should I close it?  Thanks.

benmccann:
>If I understand correctly, you opened PR #2036 and only included code for the original issue -- preview using Kit options instead of (correctly) using Vite options. These changes have been merged into master.

Yes

> As far as 'reading the file: In non middleware mode, Vite allows the cert and key (as well as pfx and ca) to be specified using the filename, a string or a buffer.

How does Vite differentiate between a filename and a string? I wonder if we should try to steal their logic for this. As far as I can see they only take a filename, but I could be missing something: https://github.com/vitejs/vite/blob/514e124b29597203cb90aac9d838cd9322f24451/packages/vite/src/node/server/http.ts#L43

> So, is this PR 'dead'? Should I close it?

I think we got the most important part in, but I'm happy to discuss if you want to pursue the other parts of this such as pfx support

JBusillo:
> How does Vite differentiate between a filename and a string? I wonder if we should try to steal their logic for this. As far as I can see they only take a filename, but I could be missing something: https://github.com/vitejs/vite/blob/514e124b29597203cb90aac9d838cd9322f24451/packages/vite/src/node/server/http.ts#L43

Vite's differentiation logic is contained within [this function](https://github.com/vitejs/vite/blob/514e124b29597203cb90aac9d838cd9322f24451/packages/vite/src/node/server/http.ts#L53):

If the option ('value' parameter) is not a string, it assumes that it's a buffer.  It returns "value" as is.
If the option is a string:
- Vite tries to treat it as a filename, and reads it via fs.readFileSync().  If it's successful, the value returned is the file contents as a buffer[], since there's no {encoding: 'utf-8') specified.
- If the read was not successful, it assumes that the string value specifies the actual contents of the cert/key/ca/pfx.  It returns the string as is.
 
The Node options cert, key, ca and pfx can be passed as a string or a buffer[] -- [Create Secure Context Options](https://nodejs.org/dist/latest-v14.x/docs/api/tls.html#tls_tls_createsecurecontext_options)

Using Vite's strategy might be beneficial.  The user need not specify the {encoding: 'utf-8') option for fs.readFileSync, or can eliminate the fs.readFileSync call by merely specifying the filename.

> > So, is this PR 'dead'? Should I close it?

I believe that you prefer implementing PR's that resolve current open issues -- which is good for tracking/documentation/search purposes.  Depending on what you'd like me to implement, I'd be happy to open the related issue(s) and open a separate PR to implement them.  For me, it would be safest to open a new PR, avoiding the debacle of a couple days ago :)

BTW, I tested what happens when only the pfx option is specified -- Kit creates a self-signed certificate, and Node uses this certificate and ignores the pfx certificate.  Therefore, I think we can classify this as a issue (bug).

benmccann:
Using Vite's strategy sounds like a good plan to me

No need to file an issue. We've already discussed it here. You can just go ahead and create a PR if you'd like

benmccann:
Btw, I see there's a PR pending to fix pfx support in Vite now: https://github.com/vitejs/vite/pull/4496

benmccann:
> For me, it would be safest to open a new PR, avoiding the debacle of a couple days ago 

Since you said you'd prefer opening a new PR to updating this one, I'll go ahead and close it. If you change your mind about that we can go ahead and reopen this one

JBusillo:
Thanks for closing this.  I've been a bit busy with personal items, but I'll open up the new PR  and make the changes soon.
Also, you had mentioned the Vite PR for pfx support -- You might already be aware that the PR doesn't affect us, as SvelteKit is using 'middlewares' mode, and is responsible for setting up its own HTTP/S server.

PR #2030, opened 2021-07-29
[chore] deduplicate config types

ignatiusmb:
It's getting harder to track changes that happens to the config options, this will make it so that we derive `ValidatedConfig` from `Config` and only need to change stuff in one place.


benmccann:
looks like there's a merge conflict

packages/kit/types/helper.d.ts
```diff
@@ -31,3 +31,10 @@ export type InferValue<T, Key extends keyof T, Default> = T extends Record<Key,
 	: Default;
 export type MaybePromise<T> = T | Promise<T>;
 export type Rec<T = any> = Record<string, T>;
+export type RecursiveRequired<T> = {
```
benmccann, line 34: I can't understand what this is doing. It probably deserves a code comment at least

As an alternative, should we replace `Config` with `Partial<ValidatedConfig>`?

ignatiusmb: This is a recursive implementation of [Required](https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredtype), with a few modifications to handle some of our differences between user config and validated config. I'll add a code comment for that one.

Replacing with `Partial<ValidatedConfig>` as an alternative would still need a recursive implementation of `Partial` as it only sets properties one-level deep, `ValidatedConfig` also has `kit: { files: { setup: string } }` which is only used after validation and not in the user config.

packages/kit/types/helper.d.ts
```diff
@@ -31,3 +31,14 @@ export type InferValue<T, Key extends keyof T, Default> = T extends Record<Key,
...
+	// Recursive implementation of Required utility type.
```
benmccann, line 35:
```suggestion
	// Recursive implementation of TypeScript's Required utility type.
```

PR #2037, opened 2021-07-29
[docs] update adapter doc examples

smj-edison:
This PR updates the examples in the adapter docs to use `adapter()` instead of `node()`.


smj-edison:
Never mind, this isn't necessary.

PR #2038, opened 2021-07-29
Remove invalid `margin-bottom: 0 0 1em 0`

Amerlander:
`margin-bottom: 0 0 1em 0;` seems invalid, so I removed it.

### DIFF
--- a/packages/create-svelte/templates/default/src/app.css
+++ b/packages/create-svelte/templates/default/src/app.css
@@ -72,7 +72,6 @@ a:hover {
 
 h1 {
 	font-size: 2rem;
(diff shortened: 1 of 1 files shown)


Conduitry:
I would imagine the intention was `margin: 0 0 1em 0`. I think we want to override the bottom margin to be 1em, but we also want to remove the top margin.

Amerlander:
I first thought the same, but when I did this the layout changes visually. And then I thought the layout got checked visually like a thousand times and therefore it might be better to just remove this line and keep the appearance like it is.

benmccann:
I can barely tell the difference, but like it better without

Issue #1610, opened 2021-05-31
Make SvelteKit be compatible with server middlewares


hgl:
**Is your feature request related to a problem? Please describe.**
There are a lot of issues that either ask for server side features like [SSE](https://github.com/sveltejs/kit/issues/887), [websocket](https://github.com/sveltejs/kit/issues/1491), etc or directly for server side [middlewares](https://github.com/sveltejs/kit/issues/334).

I think the root issue is that the `handle` hook exposes its own type of request and response, which are not compatible with express middlewares that is the de facto standard in Node.

I get the rationale behind this design, as @Rich-Harris said in #334

> it means that the answer to various problems becomes 'use adapter-node, because you can use arbitrary middleware'

However, the current design feel more like making adapter-node a second-class citizen in order to treat every adapter the same.

I recently came across https://github.com/vendia/serverless-express, which seems to contain a solution to this problem.

**Describe the solution you'd like**

This code converts a request-like object into `http.IncomingMessage` (there is similar code for `http.ServerResponse`)
https://github.com/vendia/serverless-express/blob/5cd87c8e7ca3a5e971d26d471cb70a6ac15cc61f/src/request.js#L7-L16

And it's the cornerstone for using middlewares in a serverless environment.

I wonder if it's a good idea to make the `handle` hook (maybe also endpoints) expose `http.IncomingMessage` and `http.ServerResponse` instead? For `adaptor-node`, the corresponding objects from polka are directly returned, for other adaptors, the aforementioned polyfill is applied.

This should open Svelte Kit to the vast server middlewares that are available.

**Describe alternatives you've considered**

**How important is this feature to you?**
Very, not being able use existing middlewares is a deal breaker for many people from the referenced issues.
But it's actually possible to polyfill in userland to use the middlewares.

**Additional context**

benmccann:
The idea that Rich had suggested was to build a converter between the SvelteKit and Express request and response objects

hgl:
That works, but if they're convertible, I wonder what's the appeal of exposing the custom types in the first place, and for adapter-node, that seems to mean converting back and forth?

jthegedus:
@benmccann Is that suggestion tracked in this repo somewhere? It would make the Firebase adapter (which uses Express APIs) much easier to integrate.

benmccann:
> if they're convertible, I wonder what's the appeal of exposing the custom types in the first place

We didn't want to use Express types because not all adapters provide an Express request / response the way adapter-node does.

[...]

PR #2048, opened 2021-07-31
[feat] allow node adapter to configure listen path

matths:
The build and run process using the current @sveltejs/adapter-node only allows to run the final node project listening to either host/port or 0.0.0.0/port specified by an environment variable.
I would expect it to allow listening to a path instead as specified by node, check documentation https://nodejs.org/api/net.html#net_server_listen_path_backlog_callback

**Use case:**
When using a reverse proxy that expects the node application to listen for a path (domain socket file) instead of a host and port, this was currently not available in the node-adapter.
With this pull request I want to make an offer for this missing feature.

With this feature branch applied yo can now do `SERVER_PATH=/tmp/socket node build` to have your server listen to a domain socket file instead of a port.

I think this is somewhat related to the #1610 as this would be the other attempt to solve this problem. If the node-apapter would lead to a middleware instead of a whole node server, the developer would be free to have his own node server and have it listen to a domain socket file instead of host/port.

**Tests**
I did clone the existing smoke test for this use case.
And I used the changed adapter-node within my project:
* in `packages/adapter-node` run `pnpn build`
* copy files to your project `cp packages/adapter-node/index.js packages/adapter-node/index.d.ts packages/adapter-node/files/* ~/myproject/src/lib/`
* do an `import adapter from './src/lib/adapter-node/index.js';`instead of `import adapter from '@sveltejs/adapter-node';` in your projects svelte.config.js
* run npm build in your project
* to run your server use `SERVER_PATH=/tmp/socket node build`

**Changeset**
I tried to include a changeset, but I am not sure which files to add.


benmccann:
@matths can you update the PR so that the Lint job passes?

benmccann:
> I think this is somewhat related to the #1610 as this would be the other attempt to solve this problem. If the node-apapter would lead to a middleware instead of a whole node server, the developer would be free to have his own node server and have it listen to a domain socket file instead of host/port.

If you got https://github.com/sveltejs/kit/pull/2051 merged would you no longer need this PR then? Because I think the more we can expose the underlying to allow users to configure it directly that's better vs trying to add all the options here individually

matths:
yes, I agree, we can skip and close this PR as I'll focus on #2051.

packages/adapter-node/src/index.js
```diff
@@ -1,12 +1,20 @@
...
 import { createServer } from './server';
 
 init();
 
-const instance = createServer({ render }).listen(port, host, () => {
-	console.log(`Listening on ${host}:${port}`);
-});
+const instance = createServer({ render });
+
+if (path) {
```
benmccann, line 10: You might be able to simplify this by changing it to:
```
	const listenOpts = { path, host, port };
	instance.listen(listenOpts, () => {
		console.log(`Listening on ${path ? path : host + ':' + port}`);
	});
```

benmccann:
@matths the tests for this PR are failing on Windows. Could you take a look?

Issue #334, opened 2021-01-20
Expose a way to inject middleware in the server pipeline for adapt-node


samccone:
Similar to #333 

It would be valuable to be able to insert additional handlers in the adapt-node server pipeline to customize the serving behavior based on deployment env. 
https://github.com/sveltejs/kit/blob/master/packages/adapter-node/src/server.js#L30

In sapper we directly exposed the server making this possible but now in sveltejs + adapt-node we do not expose this dial. It would be nice to have an escape hatch

Rich-Harris:
I feel like we need to be specific about the use cases here. The downside of a catch-all escape hatch is that it means that the answer to various problems becomes 'use `adapter-node`, because you can use arbitrary middleware' which is the opposite of what we want out of a serverless framework. I think it would be better to be able to satisfy those uses across different adapters, but it's hard to know what the relevant APIs look like unless we have a clear handle on what those uses are

benmccann:
People are going to have use cases we haven't imagined. I'm not sure we should block them from being able to use SvelteKit. I really think this is something that's going to be needed

Just a few minutes ago someone was complaining on Discord that they can't use passport. I've already been frustrated by not being able to use Multer or Formidable (https://github.com/sveltejs/kit/issues/70#issuecomment-800749792). There may be solutions to these that we can build that would be better, but there's a whole ecosystem of solutions out there that we're stopping people from using and it seems impossible that we'd reimplement all of them

Rich-Harris:
File uploads should be handled out of the box, and serverless apps need auth too. I really don't think multer, formidable and passport are the right answers in 2021

dummdidumm:
I think those were just some specific examples, showing that in general it's not really predictable what the whole JS ecosystem does, and that it may need support for middleware from SvelteKit. Also I don't think we should prevent people from using libraries which are battle-tested and which they want to use. If this was supported and people would reach more for `adapter-node`, I don't think this would be a problem. It's the user's decision to do this after all. It also would help us understand the use cases better and then maybe provide a solution which works in serverless environments, too, later on.

GrygrFlzr:
I'm crazy, so the way I've been doing this for a production app is:
- Dev mode runs an Express + GraphQL + Passport server which proxies requests to SvelteKit's dev server with Axios
- Prod mode imports the optimized build from `svelte-kit build` (with a blank adapter) into the express server and uses it as a request -> response black box

[...]

Issue #1779, opened 2021-06-29
[Request] Allow adapters to modify dev mode functionality


sw-yx:
**Is your feature request related to a problem? Please describe.**

many function platforms like begin and netlify extend the serverless function interface to offer more built in capability, eg. auth and storage features.

right now, adapters are only applied at build time. this causes a mismatch between the dev and production capabilities available to the developer.

**Describe the solution you'd like**

Inside adapters, offer a new `devFunction()` hook that allows local mocking of production platform functionality, hopefully piggybacking on the ones that each platform already offers (eg [Netlify Dev](https://docs.netlify.com/cli/get-started/#netlify-dev), Begin's [Sandbox](https://github.com/architect/sandbox)).

devFunction would be injected [here](https://github.com/sveltejs/kit/blob/1c78dec3b65d407db70a555ff54cc11442dcdc07/packages/kit/src/cli.js#L88) and it would modify [this handler](https://github.com/sveltejs/kit/blob/1c78dec3b65d407db70a555ff54cc11442dcdc07/packages/kit/src/core/dev/index.js#L84) as middleware

**Describe alternatives you've considered**
not doing it

**How important is this feature to you?**
only minor, but it will help increase the value proposition of adapters as that is a unique feature of us as a metaframework compared to alternatives.

**Additional context**

Example confusion caused https://github.com/sveltejs/kit/issues/1249

sw-yx:
i'm happy to implement if you approve design/scope

benmccann:
That sounds reasonable to me. I've added it to the 1.0 milestone so that Rich will hopefully see it at some point and chime in with his thoughts as well

Rich-Harris:
Yeah, this seems necessary, though i don't have a good grasp of what it would look like in practice. Leaving the implementation aside, can you articulate exactly what the intended dev experience would be? (i.e. do you run `netlify dev`/`sandbox`, or does `svelte-kit dev` do that behind the scenes via the adapter? etc)

JeanJPNM:
I think each adapter should handle cli commands of their target platform, letting their users add arguments as they need.
But how will be the interaction between the framework and the adapter dev mode? Should the code for the backend be processed by svelte kit? If it is what will happen with frameworks that have some "custom compilation" (such as nestjs)?

jthegedus:
I see this feature as providing little value compared to the maintenance work for Adapter maintainers. In my experience these platform CLIs don't take well to loading dev builds every few seconds.

[...]

Issue #887, opened 2021-04-05
Support Server-Sent Events (SSE) with adapter-node


babeard:
**Describe the bug**
While migrating a sapper app to svelte-kit / adapter-node, I came across a feature that seems to be missing. Namely, the ability to create custom middleware that can handle [Server-Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events) or other long-lived connections that send data to the response stream without closing the connection. Since the handle hook requires **returning** `Response | Promise<Response>`, thus closing the connection, it does not support this. Nor have I found any other way to `res.write();` using endpoints.

**Information about your SvelteKit Installation:**

<details>
  <summary>Diagnostics</summary>

```
System:
    OS: Linux 4.4 Ubuntu 20.04.1 LTS (Focal Fossa)
    CPU: (8) x64 Intel(R) Core(TM) i7-4770 CPU @ 3.40GHz
    Memory: 7.97 GB / 15.94 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
Binaries:
    Node: 14.5.0 - ~/.nvm/versions/node/v14.5.0/bin/node
    npm: 6.14.5 - ~/.nvm/versions/node/v14.5.0/bin/npm
npmPackages:
    @sveltejs/kit: ^1.0.0-next.68 => 1.0.0-next.68
    svelte: ^3.37.0 => 3.37.0
    vite: ^2.1.5 => 2.1.5
```
- Using node-adapter
</details>

**Severity**
Adds a bit of friction for migrating from Sapper, reduces feature parity, and will require me to either 
a) regress to polling the server from the client, or 
b) write a custom adapter for every project that needs SSEs. (which I'm not sure that this even helps for the development environment anyways)

**Additional context**
[Issue #334](https://github.com/sveltejs/kit/issues/334) seems relevant. 

Also originally asked within [discord chat](https://discord.com/channels/457912077277855764/819723698415599626/827550800003465216) but no one seems to know the answer, which is why I am submitting this issue.

cayter:
@babeard After going through https://vitejs.dev/guide/api-plugin.html#configureserver, I believe we can try the below:

1. Come up with a middleware with the method signature `function (req, res, next) => {}`.
2. Configure the middleware in `svelte.config.cjs`'s `vite.plugins`. (this is for `svelte-kit dev`)
3. Come up with a forked `adapter-node` that allows polka to use the same middleware. (this is for `node build/index.js`)

[...]

Issue #70, opened 2020-10-27
Handle file uploads


Rich-Harris:
Currently, the body parser bails on multipart forms that include files

benmccann:
It looks like the way you'd do this in Sapper is to add something like [Multer](https://www.npmjs.com/package/multer) or [Formidable](https://www.npmjs.com/package/formidable) as a middleware which would parse the incoming request and stick the files on the request object. Probably the way to do this would be to expose a way to add middleware, so that users can utilize these packages, so I wonder if this is essentially a duplicate of https://github.com/sveltejs/kit/issues/334. (side note, I wonder if we should have `adapter-polka` and `adapter-express` instead of `adapter-node`)

I was trying to figure out how Netlify, Lambda, etc. handle this. All the examples I see directly upload to S3 from the browser. I couldn't find how you'd handle a form upload inside a Lambda, so I guess that's something that's generally not supported

thgh:
Here is an example of file uploads to lambda: https://github.com/mpxr/lambda-file-upload/blob/master/handler.js

I may be biased, but I think serverless APIs typically don't use global middleware and instead have several awaits (one per middleware) at the start of the function. See https://nextjs.org/docs/api-routes/api-middlewares#connectexpress-middleware-support

asos-tomp:
For what it's worth, confirmed that this was working fine with sapper + [Formidable](https://www.npmjs.com/package/formidable) but not working with sveltekit.

axmad386:
I think hooks is enough for handling Middleware, except the request (ServerRequest) is little bit different from polka or express. Multer require request.pipe and unpipe to handle File Upload. Formidable require request.on which is missing from request object on handle hooks. Cmiiw

benmccann:
You'd have to convert from SvelteKit request to Express request first before calling the middleware

axmad386:
> You'd have to convert from SvelteKit request to Express request first before calling the middleware

That's exactly I want, but how. Any references?

vonadz:
Would it be possible to comment out this section in kit/src/runtime/server/parse_body/index.js:1413
`
if (directives.filename) {
  //TODO we probably don't want to do this automatically
  throw new Error('File upload is not yet implemented');
}`
for now? This will give users the chance to implement their own solutions in the handle hook.

MTyson:
I just did this in Sapper, and I'll throw out my thoughts in case it might help.

[...]

PR #2051, opened 2021-07-31
Let polka from adapter-node be just another requestHandler

matths:
I stumbled across a problem similar to the ones in #334.

Currently the build process using adapter-node exports a Polka instance listening to a configurable host/port.
This way the SvelteKit app is the only requestHandler, so to say.

So my proposal is to export the app as a requestHandler instead.
A requestHander (`(req, res) => {}`) is very similar to an express like middleware (`(req, res, next) => {}`), just missing the `next()`callback.
This way you could combine the SvelteKit app with other requestHandlers and/or middleware.

**Example**
Provided using adapter-node from this branch, you could have the option `asRequestHandler` in your svelte.config.js
```js
kit: {
	adapter: adapter({
		out: 'build',
		asRequestHandler: true,
		env: {
		}
	}),
	target: '#svelte'
}
```

When calling `npm run-script build` in your project, you can no longer run your build using `node build`.
Instead you need to implement your own server (or express application) and import the build instead.

```js
import http from 'http';
import { requestHandler } from './build/index.js';

const server = http.createServer(requestHandler);
server.listen(5000, '0.0.0.0');
```

I haven't tested with express yet, but it should look similar to this

```js
import { requestHandler } from './build/index.js';
import express from 'express';

const app = express();

var myLogger = function (req, res, next) {
  console.log('LOGGED');
  next();
};

app.use(myLogger);

app.get('/no-svelte', (req, res) => {
  res.send('Hello World!')
});

app.get('*', requestHandler); // SvelteKit

app.listen(5000)
```

**Tests**
As the added code is mostly in files that are object of the build process, I haven't found a good example on how to come up with a test. I did some manual testing of the functionality of course. 

**Changesets**
no changeset provided, yer.


Rich-Harris:
What if instead of `asRequestHandler`, the adapter outputted two files — a `server.js` (similar to the current `index.js`) but also an `index.js` that exposed the bits you need, and which `server.js` imported from?

```js
// index.js
export const paths = {
  assets: join(__dirname, '/assets'),
  prerendered: join(__dirname, '/prerendered')
};

export const handler = (req, res, next) => {...};
```

That way you could either run the 'reference implementation' server-in-a-box — `node build/server` — or you could import the handler and do whatever you want with it. We'd need to figure out the details, since the exports from that file would become API rather than implementation detail, but does that feel sufficiently flexible/future-proof? A thing about this approach that's potentially quite nice is that we can add new stuff (e.g. we could later decide to expose the `sirv` instances) without making the choice of name of `asRequestHandler` (or whatever else we might come up with) becoming outdated as needs evolve.

matths:
Yes, that sounds legit and possible.
You're talking about two targets in rollup.config.js from the adapter-node package?

So also the 'reference implementation' would then use the fake server object to trick listen() from polka, just to make a plain node httpServer on its own.
https://github.com/lukeed/polka/blob/2ce10df97921126ffcc43e23aa14af18f93f49e1/packages/polka/index.js#L65

(I forgot to mention, that having your own node and/or express server using this approach would make #2048 obsolete as this could be achieved in 'userland'.)

benbender:
I like this approach very much as this would render most of the open issues asking for middleware's, server-side-events, websockets etc obsolete, as all of this could be easily achieved in userland with your custom server. All of that without compromising on sveltekit's (serverless) core and while giving enough freedom for future technologies. \o/ 💯

jsprog:
@benbender, I really like this approach hoping that it'll land very soon. At the same time, we need to understand the needs of many developers to receive the same experience when coding (dev) without resorting to other complexities (like proxies), complex and non stable hacks.

Please, see my comment here: https://github.com/sveltejs/kit/issues/887#issuecomment-891454620

matths:
I am able to split inside packages/adapter-node (files/ directory) into the two portions mentioned by @Rich-Harris (an index.js and a server.js), but I am struggeling what I need to change in the the later step, when running `node build/server` in userland project is called.
So I am not yet pushing my recent commits as they would break the status quo.
Some help would be appreciated. Feel free to contact me and/or have a chat/teams call.

benmccann:
I think I'd override the `esbuild` entrypoint to point at your own file instead of the default: https://github.com/sveltejs/kit/tree/master/packages/adapter-node#esbuild

Your own file could look pretty similar to the default, but would need a different file path to the app:
```
import { init, render } from '../.svelte-kit/output/server/app.js'; // assuming this file lives in src this would be the path to app.js
import { host, port } from './env.js'; // eslint-disable-line import/no-unresolved
import { createServer } from './server';

init();

const instance = createServer({ render }).listen(port, host, () => {
	console.log(`Listening on ${host}:${port}`);
});

export { instance };
```

benmccann:
Btw, I'm on the Svelte Discord under this username if you want to ping me with any questions

matths:
@benmccann  I've managed to come up with my interpretation of @Rich-Harris comments somehow. Feel free to have a look and give feedback. I had some struggle with externalising the handler from the reference server. I also did not expose anything other than the handler/middleware itself from the whole polka/sirv world, yet, so no new API. I hope I didn't break existing API though.
It's working perfectly for my use case. ;)

packages/adapter-node/README.md
```diff
@@ -40,11 +40,44 @@ Enables precompressing using gzip and brotli for assets and prerendered pages. I
 By default, the server will accept connections on `0.0.0.0` using port 3000. These can be customised with the `PORT` and `HOST` environment variables:
 
 ```
-HOST=127.0.0.1 PORT=4000 node build
+HOST=127.0.0.1 PORT=4000 node build/server
```
benmccann, line 43: I would expect the default server to be exposed as `index` still so that this line wouldn't change

packages/adapter-node/README.md
```diff
@@ -40,11 +40,44 @@ Enables precompressing using gzip and brotli for assets and prerendered pages. I
...
 ```
 
 You can specify different environment variables if necessary using the `env` option.
 
+## Middleware
+
+Adapter generates a [connect](https://github.com/senchalabs/connect) compatible middleware `(req, res, next) => {}`.
```
benmccann, line 50: My background isn't in Node. Does Express use Connect? I wonder if this could say `Express / Polka` instead

packages/adapter-node/README.md
```diff
@@ -40,11 +40,44 @@ Enables precompressing using gzip and brotli for assets and prerendered pages. I
...
+
+You can have a look at `build/server.js` to study a simple reference implementation
+on how to use this with a plain node http server.
+
+You can of course your favorite server framework to combine with other and/or custom middleware, e.g. [express](https://github.com/expressjs/expressjs.com).
+
+```
```
benmccann, line 57: Other questions the docs should answer: Where do you put this file? Is there some option you need to specify to make it use this vs the default?

packages/adapter-node/index.js
```diff
@@ -78,6 +79,28 @@ export default function ({
 			const buildOptions = esbuildConfig ? await esbuildConfig(defaultOptions) : defaultOptions;
 			await esbuild.build(buildOptions);
 
+			utils.log.minor('Building SvelteKit servers reference server');
```
benmccann, line 82: don't need server in the message twice
```suggestion
			utils.log.minor('Building SvelteKit reference server');
```

packages/adapter-node/README.md
matths: To quote @Rich-Harris here:
> What if instead of `asRequestHandler`, the adapter outputted two files — a `server.js` (similar to the current `index.js`) but also an `index.js` that exposed the bits you need, and which `server.js` imported from?
> 
> That way you could either run the 'reference implementation' server-in-a-box — `node build/server` — or you could import the handler and do whatever you want with it.

benmccann: Doing that would be a backwards incompatible change. That'd be okay since we're in beta if there's some benefit to it, but it doesn't seem like there's any real benefit here that would justify breaking the deployment process of everyone who already has a project running with `adapter-node`

packages/adapter-node/index.js
matths: fixed

packages/adapter-node/README.md
matths: Not sure what you mean. Using it this way is kind of advanced stuff for experienced node users. So I am not sure how to verbose to document. I'm not a native speaker, so feel free to come up with better documentations.

packages/adapter-node/README.md
matths: changed.

packages/adapter-node/README.md
matths: I switched the filenames. `index.js` ist now `handler.js` again,
and `server.js` is now `index.js`.
So in the end, using the final adapter-node you can call `node build` and get the reference server implementation.
To use the handler/middleware, you have to import `build/handler.js`into you own server implementation.

benmccann:
I know that @Rich-Harris suggested this approach earlier, but I've realized ([as was pointed out to me](https://github.com/sveltejs/kit/issues/70#issuecomment-894307612)) that it has a very serious drawback, which is that your middleware won't work in dev mode. We do hope to allow adapters to hook into dev mode (https://github.com/sveltejs/kit/issues/1779) and I don't see how you could accomplish that with the suggested approach. Having middleware only in production and not in development would make development impossible and not satisfy the vast majority of use cases. I think we'll need to go back to the drawing board and come up with a plan for how we'll accomplish this in dev and prod

matths:
Sure, I understand that point. I was driven by a 'for production only' use case, where I want to have my own node server listen to a path (domain socket file) instead of host/port and maybe to put a access-log kind of middleware upfront. Both things I wouldn't expect to work in development mode.
I also can imagine use cases where a custom middleware does manipulations 'before' falling through into the SvelteKit handler, but I did not focus on that. I also imagine it to be difficult because the other adapters wouldn't work with such changes either.

>  I think we'll need to go back to the drawing board and come up with a plan for how we'll accomplish this in dev and prod.

I currently do not see me coming up with a solution that satisfies this goal. Thanks for considering my approach as a solution for so many demands, I am sure someone else comes up with something working for all of these. What happens to this PR, will it be closed or live until you found a solution?

benmccann:
This PR might still be a good idea to merge (with some updated documentation) since it will cover your use case of using a socket domain file instead of host/port. Though I think it's a pretty cumbersome way to accomplish that and I might rather close this and reopen https://github.com/sveltejs/kit/pull/2048 instead. The thing I initially really liked about this PR was the power it gives to the user, but it turns out it's not actually very powerful if you're still limited by it being only in production mode. There may be use cases where this PR is still a good idea and still helpful, but my inclination would be to try to go with https://github.com/sveltejs/kit/pull/2048 and see how far that can take us. If there are more use cases that come up where this PR is helpful then maybe we should revisit it

benbender:
The problem is that this pr doesn't really allow middlewares **in sveltekit** but turns sveltekit itself into a middleware.

That said, I would like to vote in favor for merging this anyways, as it opens new possibilities and usage-scenarios - even if they might be a bit clunky to implement (f.e. using a proxy in dev-mode). 
Imho this pr isn't exclusive to #2048 but more like a complementary as both PRs try to open up the configuration of polka/adapter-node for a more flexible configuration by the dev.

benmccann:
I think the best course of action with this PR is to leave it on ice for a bit and then revisit in another month or so. There will be a lot of change to Vite SSR support and I'd like to see how that changes things as [it may affect the story around the dev server](https://github.com/vitejs/vite/discussions/4230#discussioncomment-1162698). This PR could still be a decent idea, but it's probably worth considering things holistically and making sure we have a plan for dev server support at the same time

packages/adapter-node/src/polka-handler.js
```diff
@@ -15,7 +15,7 @@ const paths = {
 	prerendered: join(__dirname, '/prerendered')
 };
 
-export function createServer({ render }) {
+export function createPolkaHandler({ render }) {
```
benmccann, line 18: would end users call this? If not, we should use underscores and call it `create_polka_handler`

packages/adapter-node/README.md
```diff
@@ -45,6 +45,35 @@ HOST=127.0.0.1 PORT=4000 node build
...
+
+But you can use your favorite server framework to combine it with other middleware and server logic. You can import `createHandler()`, your ready-to-use SvelteKit bundle as middleware, from `./build/handler.js`.
+
+```
+import { createHandler } from './build/handler.js';
+import express from 'express';
+
+const app = express();
+
+var myMiddleware = function (req, res, next) {
```
benmccann, line 60:
```suggestion
const myMiddleware = function (req, res, next) {
```

packages/adapter-node/README.md
benmccann: I mean if you want to create your own server using the code below does it matter what the file path or name is?

packages/adapter-node/src/polka-handler.js
```diff
@@ -38,7 +38,7 @@ export function createServer({ render }) {
 		  })
 		: noop_handler;
 
-	const server = polka().use(
+	const polka_instance = polka().use(
```
benmccann, line 41: I think I might expect at least `compression` to be added in the reference server implementation rather than here. That way if someone wants to change it they can.

Similarly, I might export `prerenderedHandler` and `assetsHandler` and add them to the reference implementation rather than adding them automatically here in cause someone wants to provide a different implementation

benmccann (commented):
I figured out that you can actually add middleware in dev mode today without any additional changes. I documented how to do that in the FAQ. So let's move ahead with this PR

@matths I rebased this PR and made a few minor cleanups. I'll leave some comments for you as well if you'd be able to take a look. Thanks!

evdama:
@matths and @benmccann, thanks a lot for your work on this subject 👍

https://discord.com/channels/457912077277855764/943462063261511680 discibes in detail the question how you could **pass information from an express session onto sveltekit** when you use `src/server.js` which imports `handler()` from `build/handler.js` like so

```js
// src/server.js - here I use express middleware to read a users IP address before I hand over to svk using handler()

import { handler } from '../build/handler.js'
import express from 'express'
import session from 'express-session'
import parser from 'ua-parser-js'

const app = express()

if (user-from-outside-Europe) {
// add a route that lives separately from the SvelteKit app
app.get('/non-svk-route', (request, response) => { response.end('non-svk route...') })
}

// let SvelteKit handle everything else, including serving prerendered pages and static assets
app.use(handler)
app.listen(3000, () => { console.log('hey...listening on port 3000') })
```

So how would one pass data from an express session onto sveltekit so it could be used with `event.locals` inside `hooks.js` for example?

benmccann:
You can't right now, but it would be a very simple change to SvelteKit to enable it. There's a `platform` field that can be populated by the request ([e.g. adapter-cloudflare uses it](https://github.com/sveltejs/kit/tree/master/packages/adapter-cloudflare#environment-variables)). However, the team would like to see a usecase that can't be solved any other way before adding this in `adapter-node`. Based on another thread in Discord it sounded like your usecase could probably be solved by `handle`, but if that's incorrect please file a feature request explaining in detail what you're trying to accomplish and why it can't be done with the existing APIs

PR #2055, opened 2021-08-01
[fix] revert adapters automatically updating .gitignore (#1924)

Rich-Harris:
Apologies for missing this when #1924 was still open. It violates something I've learned the hard way to be a sacrosanct rule: a given file in a project should be edited by humans, or by machines, but never by both after the project has been created (`package.json` might seem like an obvious counter-example, but appending dependencies is something the package manager does as a direct consequence of your command; it doesn't get unexpected edits anywhere else). We really can't go around editing people's code, even in trivial-seeming ways in ignore files. 

Adding ignored directories _seems_ simple and uncontroversial enough to be a no-brainer. But even in #1924 _itself_ it introduced unwanted changes:

```diff
/.svelte-kit
/build
/functions

+build
```

Here, the `/build` line already excluded the `build` directory from version control. The inserted line is duplicative. Worse than that, it will cover any folder called `build` anywhere in your project, so if your app had a route whose name coincided with that fairly common English word, it would now be excluded as well. As someone who has lost work because of incorrectly-excluded folders in the past, I would be pretty irate if it happened again because my framework took it upon itself to 'fix' my ignore file. We could robustify it, but there'll always be some edge case. 

Beyond that, someone might actually want the build artifacts to be checked in. In fact I work on a team where that's the norm! (It's a practice I've railed against, but there _are_ reasons that it's done that way, and the fact that it's a thing that people do, however unconventional, outweighs any arguments about why they shouldn't.) While the documentation tells you that you can comment the line out, most people would simply delete the line, only for it to be reinserted. Fighting your tools is a bad feeling.

Here's what I think we should do instead:

* [ ] Where it makes sense, adapters should place their output inside `.svelte-kit` (i.e. `.svelte-kit/netlify`, etc) so that they are already ignored by ignore files
* [ ] Check ignore files to see if we're writing to locations that aren't already excluded, and print a loud warning if they're not. We're using `gitignore-parser` in create-svelte to do this robustly: https://github.com/sveltejs/kit/blob/3e30a1b7d3f47f1c37221e21be50819a06fe5ead/packages/create-svelte/scripts/build-templates.js#L49-L50

### Tests

### Changesets

### DIFF
--- a/documentation/docs/80-adapter-api.md
+++ b/documentation/docs/80-adapter-api.md
@@ -26,7 +26,6 @@ The types for `Adapter` and its parameters are available in [types/config.d.ts](
 Within the `adapt` method, there are a number of things that an adapter should do:
 
 - Clear out the build directory
-- Call `utils.update_ignores` to ignore build output in existing `.gitignore` files at the location of `svelte.config.js`
 - Output code that:
   - Calls `init`
   - Converts from the platform's request to a [SvelteKit request](#hooks-handle), calls `render`, and converts from a [SvelteKit response](#hooks-handle) to the platform's
@@ -34,5 +33,7 @@ Within the `adapt` method, there are a number of things that an adapter should d
 - Bundle the output to avoid needing to install dependencies on the target platform, if desired
 - Call `utils.prerender`
 - Put the user's static files and the generated JS/CSS in the correct location for the target platform
+
+If possible, we recommend putting the adapter output under `'.svelte-kit/' + adapterName` with any intermediate output under `'.svelte-kit/' + adapterName + '/intermediate'`.
 
 > The adapter API may change before 1.0.
--- a/.changeset/six-comics-report.md
+++ b/.changeset/six-comics-report.md
@@ -1,0 +1,10 @@
+---
+'@sveltejs/adapter-cloudflare-workers': patch
+'@sveltejs/adapter-netlify': patch
+'@sveltejs/adapter-node': patch
+'@sveltejs/adapter-static': patch
+'@sveltejs/adapter-vercel': patch
+'@sveltejs/kit': patch
+---
+
+revert adapters automatically updating .gitignore (#1924)
--- a/documentation/docs/10-adapters.md
+++ b/documentation/docs/10-adapters.md
@@ -31,15 +31,6 @@ export default {
 };
 ```
 
-Some adapters may modify your project's `.gitignore` to include their build output. In case you don't want those patterns included you can comment them out: 
-
-```diff
-.svelte-kit
-.env
-
-- build
-+ # build
-```
 A variety of official adapters exist for serverless platforms...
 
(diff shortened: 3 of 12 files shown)


benmccann (commented):
I'm not sure you should revert the changeset from the original PR and this one will need it's own changeset. Other than that, LGTM

> Where it makes sense, adapters should place their output inside .svelte-kit (i.e. .svelte-kit/netlify, etc) so that they are already ignored by ignore files

So obvious in retrospect. I'm embarrassed none of us thought of it in. I started us off by updating the Netlify adapter here: https://github.com/sveltejs/kit/pull/2058

documentation/docs/80-adapter-api.md
```diff
@@ -26,7 +26,6 @@ The types for `Adapter` and its parameters are available in [types/config.d.ts](
 Within the `adapt` method, there are a number of things that an adapter should do:
 
 - Clear out the build directory
-- Call `utils.update_ignores` to ignore build output in existing `.gitignore` files at the location of `svelte.config.js`
```
benmccann, line 29: It may be worth mentioning somewhere in here that we recommend putting the adapter output under `'.svelte-kit/' + adapterName`

benmccann:
FYI @JeanJPNM, @jthegedus

pngwn:
Like @benmccann we should fix forward (even though this is a revert) and add a new changeset for this.

JeanJPNM:
Seems reasonable, but I see this more as a bug, since the code can't duplicates paths relative to the root. But implementing that could make the process a lot slower. And manually stating that the path is relative to the root of the project on the adapter configuration feels like a waste of time. Maybe reverting the changes its the best option.

packages/create-svelte/shared/+eslint+prettier/package.json
```diff
@@ -3,7 +3,7 @@
 		"eslint-config-prettier": "^8.1.0"
 	},
 	"scripts": {
-		"lint": "prettier --ignore-path .gitignore  --check --plugin-search-dir=. . && eslint --ignore-path .gitignore .",
```
benmccann, line 6: this is also a good independent change we should keep

benmccann:
> Like @benmccann we should fix forward (even though this is a revert) and add a new changeset for this.

While forward fixing is nice, I think it's more important to get the change out before folks have their `.gitignore` files updated. We can go back and implement the new way afterwards

I pushed some changes and may merge this pretty soon to get it out

benmccann (approved):

jthegedus:
Should we expect to see `gitignore-parser` provided via the Adapter API in the near future?

Rich-Harris:
Yes, I think we should — we could have an API like `utils.warn_if_unignored([opts.output])`

jthegedus:
Sounds like a good soln. Hopefully loud CLI output is enough.

PR #2058, opened 2021-08-01
[chore] update Netlify build output locations

benmccann:
This removes the need for calling `update_ignores`. I did not remove that call to avoid conflicting with https://github.com/sveltejs/kit/pull/2055

I only updated this one adapter for now to get a consensus on the pattern and because it's the one where I have an app setup and felt safe changing it since I have an app deployed there

CC @sw-yx since this affects Netlify


sw-yx:
@benmccann sorry for dropping off the radar but yes acknowledged and thank you for the update

PR #2080, opened 2021-08-03
chore: remove 'shamefully-hoist = true' setting from .npmrc

dominikg:
and fix package.json problems

### Tests

### Changesets

shamefully-hoist allows omitting package.json dependencies in public packages, shadowing potential issues. Without hoisting, problems are caught and packages declarations better reflect the actual usage,


packages/adapter-static/package.json
```diff
@@ -18,7 +18,9 @@
 		"@sveltejs/kit": "workspace:*",
 		"playwright-chromium": "^1.10.0",
 		"port-authority": "^1.1.2",
+		"svelte": "^3.40.0",
 		"sirv": "^1.0.12",
-		"typescript": "^4.3.5"
+		"typescript": "^4.3.5",
```
benmccann, line 23: I notice we have typescript declared in project root `package.json`. Is there a way to use that version here? Otherwise I'm not sure the point of declaring it in the main `package.json`

dominikg: i think that line could be removed as the only reference to typescript in this package is in the `"check": "tsc"`  script. 

didn't search for extraneous declarations for this PR, just added the ones required to keep everything (build, test, format, lint) working.

benmccann: It looks like every package declares a dependency on TypeScript, but I don't think any of them use it other than `tsc`

benmccann: I'd say go ahead and remove it assuming `pnpm check` still runs and passes, but we should remove it from all of them

dominikg: got ahead of myself and pushed the typescript removal directly on master, added here now too.

packages/create-svelte/package.json
```diff
@@ -15,15 +15,16 @@
...
 		"svelte-preprocess": "^4.7.3",
 		"tiny-glob": "^0.2.8"
 	},
 	"scripts": {
 		"build": "npm run format && node scripts/build-templates",
 		"check": "tsc",
 		"lint": "eslint --ignore-path .gitignore --ignore-path ../../.gitignore \"./*.js\" && npm run check-format",
-		"format": "prettier --write . --config ../../.prettierrc --ignore-path .gitignore --ignore-path ../../.gitignore --plugin prettier-plugin-svelte",
-		"check-format": "prettier --check . --config ../../.prettierrc --ignore-path ../../.gitignore --ignore-path .gitignore --plugin prettier-plugin-svelte",
+		"format": "npm run check-format -- --write",
```
benmccann, line 26: this seems like a good change, but we should make it in all the other packages for consistency

PR #2087, opened 2021-08-04
[chore] get netlify build passing again

benmccann:


Conduitry:
Will this affect the starter projects that `create-svelte` creates? I don't think we'd want to do that. What ends up happening with this file you're editing - does it end up in the published package in some way? I'm _guessing_ not, because I don't recall a way to init a Netlify project with the CLI, but I wanted to make sure.

benmccann:
it shouldn't end up in published packages. plus, I only edited existing files, so it wouldn't be related to this PR if it were an issue

PR #2094, opened 2021-08-04
[fix] JSONValue can't be a Date

benmccann:
@NoelAbrahams, FYI your example on https://github.com/microsoft/TypeScript/issues/1897#issue-56296923 seems incorrect to me


NoelAbrahams:
We had some special handling for `Date` at the time of writing. See `toJSON`. But, yes, you would probably want to omit date.

Issue #1825, opened 2021-07-05
kit.prerender.pages can't handles paths with query strings


johnnysprinkles:
**Describe the bug**
When using adapter-static, you can list out the pages you want to prerender in kit.prerender.pages, but query strings confuse the system. They get appended to the directory name, and the index.html that should be inside is missing.

Why would you want to? Imagine you have a page with 3 tabs and want to pre-render each of them, like:

/page?tab=1
/page?tab=2
/page?tab=3

**To Reproduce**
Use a config like this, then `npm run build` with adapter-static.

```
		prerender: {
			pages: [
				'*',
				'/about?foo=bar',
			]
                }
```

**Expected behavior**
The most reasonable thing would be to put the query string on the index.html file, either before the extension or after, not sure what is better. You can see the code where URLSearchParams is just made empty all the time.

https://github.com/sveltejs/kit/blob/5418254808fd7d10cfd74d1e8c753bae68c831cc/packages/kit/src/core/adapt/prerender.js#L126

In the above example the resulting artifacts might be:

/page/index.html?page=1
/page/index.html?page=2
/page/index.html?page=3

or

/page/index?page=1.html
/page/index?page=2.html
/page/index?page=3.html (keeps the file extensions the same for OS file open support maybe?)

Seems like an easy fix, I'll probably fix myself.

**Severity**
Not severe.

johnnysprinkles:
On second thought, it's clear that the core SvelteKit team wants to steer people away from using adapter-static as a component in a larger frontend system. I think it could be super useful, but it might be kind of esoteric, and I do still think a fork of SvelteKit for that use case might be warranted at some point. Given this, let me resolve this as will not fix.

PR #2104, opened 2021-08-05
[feat] detect if app tries to access query with prerender enabled

benmccann:
Closes https://github.com/sveltejs/kit/issues/669

### DIFF
--- a/.changeset/serious-fireants-drum.md
+++ b/.changeset/serious-fireants-drum.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[feat] detect if app tries to access query with prerender enabled
--- a/packages/kit/src/core/build/index.js
+++ b/packages/kit/src/core/build/index.js
@@ -340,6 +340,7 @@ async function build_server(
 					load_component,
 					manifest,
 					paths: settings.paths,
+					prerender: ${config.kit.prerender.enabled},
 					read: settings.read,
 					root,
 					service_worker: ${service_worker_entry_file ? "'/service-worker.js'" : 'null'},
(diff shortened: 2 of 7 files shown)


johnnysprinkles:
Sorry, why prevent them from accessing query while prerendering? You could have fixed options for query string in the pages array, like:

    kit: {
      prerender: {
        pages: [
          '/dashboard?tab=1',
          '/dashboard?tab=2',
          '/dashboard?tab=3',
        ],
      },
    }

i.e. nonparameterized query string options. This change breaks that pattern right?

johnnysprinkles:
Also, do you guys not require another pair of eyes to review each change going in?

benmccann:
>You could have fixed options for query string in the pages array, like...

I don't think you could even before this change. `?` is a forbidden character on some file systems and ([prerender serving doesn't look for query string when deciding an asset to server](https://github.com/sveltejs/kit/blob/9d1f6f6f1df1712f38fc155337160f4302d0da76/packages/adapter-node/src/server.js#L19)). I'm not saying we'd never support it in the future, but historically we haven't and I'd much rather let users know it's not supported up front than only finding out after much debugging.

But I'm not sure whether it is something we should support. E.g. what would happen if someone requests `/dashboard?tab=4`? Do you return 404, do you return the non-prerendered page? Query strings in Svelte today are not considered in routing and you'd be breaking that model, which makes it unclear how it should behave

We ask for reviews on things that are a bit larger such as changes in APIs, etc. This PR was open for almost a week, so the community could comment on it and the proposed plan was mentioned months ago in the associated issue, so didn't rise to that level

johnnysprinkles:
Got it.

Ah yes, I totally forgot about my https://github.com/sveltejs/kit/issues/1825, query strings don't even work at the moment anyway. This actually makes it so they probably won't ever, which is of course fine for new websites. The only issue might be if you're porting an existing website to SvelteKit and want to match the URL form precisely, it could be a problem. Maybe any discussion can continue in #1825.

PR #2109, opened 2021-08-05
[breaking] standardize final output dir as /build (vs /.svelte-kit)

benmccann:
**Background**

We changed the output location to `.svelte-kit` yesterday because that directory was in the default `.gitignore` and we had gotten numerous requests that an adapter's output should be ignored without need to take user action. It turned out that standardizing under `.svelte-kit` has been an unpopular change. Instead we will add `build` to the default `.gitignore` alongside `.svelte-kit` and standardize on writing the final output there with intermediate output going to `.svelte-kit`

**Multi-cloud**

For the past 24 hours output has been written in a namespaced manner. The consensus on this seems to be that a more deeply nested output location is a worse experience for the majority of users. Adapters will now be standardized to writing in a non-namespaced manner. If users would like two build with two or more adapters, they can either clear out the `build` directory between builds or change the default output location using adapter options to set namespaced build directories

**Apologies**

We're very sorry for the churn this has caused. Changing the output location is a breaking change and breaking changes are a bad user experience, but two breaking changes in a row on the same feature is especially bad. We have added a new "breaking change" label on Github and made it hot pink to stand out. We will label breaking PRs with this in the future and leave them open for a longer comment period to gather more feedback. However, this PR will be merged quickly as we're aiming to get this PR in before too many people update their projects so that most people will be unaffected. For people that updated their projects in the past 24 hours, we're very sorry for the two breaking changes in a row. This will not change again. We've got only a couple dozen open issues left before 1.0 and at that point will be able to follow semver

### DIFF
--- a/.changeset/funny-trainers-hammer.md
+++ b/.changeset/funny-trainers-hammer.md
@@ -1,0 +1,8 @@
+---
+'@sveltejs/adapter-netlify': patch
+'@sveltejs/adapter-node': patch
+'@sveltejs/adapter-static': patch
+'create-svelte': patch
+---
+
+[breaking] standardize final output dir as /build (vs /.svelte-kit)
--- a/documentation/docs/10-adapters.md
+++ b/documentation/docs/10-adapters.md
@@ -17,7 +17,7 @@ export default {
 };
 ```
 
-With this, [svelte-kit build](#command-line-interface-svelte-kit-build) will generate a self-contained Node app inside `.svelte-kit/node/build`. You can pass options to adapters, such as customising the output directory in `adapter-node`:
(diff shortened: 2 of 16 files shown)


benmccann:
CC adapter maintainers @jthegedus @ryanbethel @sw-yx

ryanbethel:
@benmccann Thanks for the heads up. I will try to update the begin/architect adapter ([https://github.com/architect/sveltekit-adapter](https://github.com/architect/sveltekit-adapter))probably tomorrow.

jthegedus:
I'll update the Firebase adapter once this is released.

benmccann:
it should already be released

jthegedus:
> it should already be released

I guess this goes more to my point with #2116 as after reading the Changelog of kit I concluded this was not released.

Issue #2101, opened 2021-08-04
Consider renaming `serverFetch` to `externalFetch`


benmccann:
### Describe the problem

`serverFetch` name is unclear. That the docs need to say in bold that it's external is a bit of a code smell. Even still, we've had people get confused (https://github.com/sveltejs/kit/issues/1823)

### Describe the proposed solution

Rename to `externalFetch`. That it runs on the server is already implied by it being located in `hooks`

### Alternatives considered

status quo

### Importance

nice to have

stalkerg:
As, author of the original idea and PR I agree with you, this name will be better. Sorry for what I didn't find this name.

PR #2110, opened 2021-08-05
rename serverFetch to externalFetch

Mlocik97:
Closes https://github.com/sveltejs/kit/issues/2101

---

### Tests

### Changesets


packages/kit/src/core/dev/index.js
```diff
@@ -346,7 +346,7 @@ async function create_handler(vite, config, dir, cwd, manifest) {
 						hooks: {
 							getSession: hooks.getSession || (() => ({})),
 							handle: hooks.handle || (({ request, resolve }) => resolve(request)),
-							serverFetch: hooks.serverFetch || fetch
+							externalFetch: hooks.externalFetch || fetch
```
benmccann, line 349: copying a comment over from the old PR

on line ~293 we have a check for old hooks. we should add a check for `serverFetch` as well

```
				if (/** @type {any} */ (hooks).getContext) {
					// TODO remove this for 1.0
					throw new Error(
						'The getContext hook has been removed. See https://kit.svelte.dev/docs#hooks'
					);
				}
```

Mlocik97:
```js
if (/** @type {any} */ (serverFetch).getContext) {
					// TODO remove this for 1.0
					throw new Error(
						'The serverFetch was renamed to externalFetch. See https://kit.svelte.dev/docs#hooks-externalfetch'
					);
				}
```

so something like this?

Mlocik97: I tried, but I'm totally lost in this. I don't know how to define value of if condition.

benmccann: I would think that on line 300, you can add:
```
				if (/** @type {any} */ (hooks).serverFetch) {
					// TODO remove this for 1.0
					throw new Error(
						'The serverFetch hook has been renamed to externalFetch'
					);
				}
```

Mlocik97: oh, ofc... I'm dumb.

benmccann (approved):

benmccann:
oh, it looks like the Lint job is failing. can you fix that?

Mlocik97:
oh, I forgot to run `pnpm format`, should be fine now.

stalkerg:
ok, looks fine for me

packages/kit/types/index.d.ts
```diff
@@ -9,7 +9,7 @@ export { ErrorLoad, ErrorLoadInput, Load, LoadInput, LoadOutput, Page } from './
 export {
 	GetSession,
 	Handle,
-	ServerFetch,
+	ExternalFetch,
```
ignatiusmb, line 12: Please put this before `GetSession` for sorting consistency

packages/kit/types/internal.d.ts
```diff
@@ -1,6 +1,6 @@
 import { RequestHandler } from './endpoint';
 import { Headers, Location, ParameterizedBody } from './helper';
-import { GetSession, Handle, ServerResponse, ServerFetch, StrictBody } from './hooks';
+import { GetSession, Handle, ServerResponse, ExternalFetch, StrictBody } from './hooks';
```
ignatiusmb, line 3: Same as above

documentation/docs/04-hooks.md
```diff
@@ -92,19 +92,19 @@ export function getSession(request) {
 
 > `session` must be serializable, which means it must not contain things like functions or custom classes, just built-in JavaScript data types
 
-### serverFetch
+### externalFetch
 
 This function allows you to modify (or replace) a `fetch` request for an **external resource** that happens inside a `load` function that runs on the server (or during pre-rendering).
```
ignatiusmb, line 97: We should also update this while we're at it

> That the docs need to say in bold that it's external is a bit of a code smell.

Kapsonfire-DE:
@benmccann @stalkerg 
I guess we should rename the function before 1.0 and make externalFetch an alias with deprecation warning
as early people change their code the less makes it breaking

benmccann:
I think the current check for old `serverFetch` name should be sufficient. The first time someone runs their app they'll be told to update their code. I actually think this is safer in a lot of ways

Rich-Harris:
thank you!

PR #2112, opened 2021-08-05
[docs] numerous FAQ updates

benmccann:
- added links to various open issues
- moved info about reading `package.json` into its own FAQ. it seems unrelated to the question it was included in
- remove file system access faq since i'm not sure it is frequently asked. i've never seen someone hit it
- updated client-side FAQ to account for SSR lifecycle no-ops in newest version of Svelte
- updated links to point to new sveltesociety website


documentation/faq/80-integrations.md
```diff
@@ -23,28 +23,35 @@ Vite will attempt to process all imported libraries and may fail when encounteri
 If you need access to the `document` or `window` variables or otherwise need it to run only on the client-side you can wrap it in a `browser` check:
 
 ```js
```
Mlocik97, line 25: this should be html/svelte type?

documentation/faq/80-integrations.md
```diff
@@ -23,28 +23,35 @@ Vite will attempt to process all imported libraries and may fail when encounteri
...
+	if (browser) {
+		// client-only code here
+	}
+</script>
 ```
 
-You can also run code in `onMount` if you'd like to run it after the component has been first rendered to the DOM. In this case, you may still find a benefit of including a `browser` check as shown below because Vite may otherwise attempt to optimize the dependency and fail on it. [We hope to make this unnecessary in the future](https://github.com/sveltejs/svelte/issues/6372).
+You can also run code in `onMount` if you'd like to run it after the component has been first rendered to the DOM:
 
 ```html
```
Mlocik97, line 37: js?

documentation/faq/80-integrations.md
```diff
@@ -23,28 +23,35 @@ Vite will attempt to process all imported libraries and may fail when encounteri
...
 
-	if (browser) {
-		onMount(async () => {
-			lib = (await import('some-browser-only-library')).default;
-		});
-	}
-</script>
+If the library you'd like to use is side-effect free you can also statically import it and it will be tree-shaken out in the server-side build where `onMount` will be automatically replaced with a no-op:
+
+```html
```
Mlocik97, line 48: js?

documentation/faq/80-integrations.md
benmccann: done. thanks for pointing these out

documentation/faq/80-integrations.md
```diff
@@ -30,21 +30,26 @@ if (browser) {
...
+	method('hello world');
+});
+```
 
-	let lib;
+If the library you'd like to use is side-effect free you can also statically import it and it will be tree-shaken out in the server-side build where `onMount` will be automatically replaced with a no-op:
+
+```js
+import { onMount } from 'svelte';
+import { method } from 'some-browser-only-library');
```
bluwy, line 48:
```suggestion
import { method } from 'some-browser-only-library';
```

documentation/faq/80-integrations.md
```diff
@@ -30,21 +30,26 @@ if (browser) {
...
 
-	if (browser) {
-		onMount(async () => {
-			lib = (await import('some-browser-only-library')).default;
-		});
-	}
-</script>
+onMount(async () => {
```
bluwy, line 50:
```suggestion
onMount(() => {
```

there's no awaits in this function

documentation/faq/80-integrations.md
benmccann: thanks for catching that!

documentation/faq/80-integrations.md
benmccann: done

PR #2113, opened 2021-08-05
[feat] use Netlify internal functions directory for generated functions

### LINKED ISSUE #1249: Is it possible to add custom netlify functions
Hey there 👋

I'm trying to deploy my svelte kit app to netlify, which seems to be working perfectly, when using the adapter. 

However, I also wanted to deploy a custom function of mine and I noticed that this function is never deployed to netlify. Only the function called `render`. 

I also noticed that when I add my custom function to the specified `functions` directory and then run `npm run build`, my custom function is removed from the `functions` folder and only the `render` function is still there. 

I'm wondering am I missing something here or is it currently not possible to add a custom function to netlify when using the adapter ? 

Thanks in advance 🙇‍♂️

ascorbic:
## Description
This changes the location of generated Netlify render function to the internal functions directory (`.netlify/functions-internal`), rather than using the user functions directory. This means the user can use the normal functions directory for their own Netlify functions without them being deleted at build time.
Fixes #1249

### Tests

### Changesets

### DIFF
--- a/examples/hn.svelte.dev/.gitignore
+++ b/examples/hn.svelte.dev/.gitignore
@@ -2,5 +2,4 @@
 node_modules
 /.svelte
 /build
-/functions
 /.netlify/functions*
--- a/examples/hn.svelte.dev/.gitignore
+++ b/examples/hn.svelte.dev/.gitignore
@@ -3,3 +3,4 @@ node_modules
 /.svelte
 /build
 /functions
+/.netlify/functions*
--- a/examples/hn.svelte.dev/.netlify/netlify-plugin-pnpm/index.js
+++ b/examples/hn.svelte.dev/.netlify/netlify-plugin-pnpm/index.js
@@ -1,8 +1,9 @@
-export default {
+module.exports = {
   onPreBuild: async ({ utils: { build, run } }) => {
     try {
       await run.command("npm install -g pnpm")
-      await run.command("pnpm install")
+      await run.command("pnpm -w install")
+      await run.command("pnpm -w build")
     } catch (error) {
       return build.failBuild(error)
     }
--- a/examples/hn.svelte.dev/netlify.toml
+++ b/examples/hn.svelte.dev/netlify.toml
@@ -7,7 +7,3 @@
 
 [[plugins]]
   package = "/.netlify/netlify-plugin-pnpm"
-
-[dev]
-# Disable svelte-kit dev so we can test functions
-framework = "#static"
--- a/examples/hn.svelte.dev/netlify.toml
+++ b/examples/hn.svelte.dev/netlify.toml
@@ -1,10 +1,13 @@
 [build]
   command = "npm run build"
   publish = ".svelte-kit/netlify/build/"
-  functions = ".svelte-kit/netlify/functions/"
 
 [build.environment]
   NPM_FLAGS="--prefix=/dev/null"
 
 [[plugins]]
   package = "/.netlify/netlify-plugin-pnpm"
+
+[dev]
+# Disable svelte-kit dev so we can test functions
+framework = "#static"
--- a/examples/hn.svelte.dev/netlify.toml
+++ b/examples/hn.svelte.dev/netlify.toml
@@ -1,6 +1,6 @@
 [build]
   command = "npm run build"
-  publish = ".svelte-kit/netlify/build/"
+  publish = "build"
 
 [build.environment]
   NPM_FLAGS="--prefix=/dev/null"
--- a/packages/adapter-netlify/README.md
+++ b/packages/adapter-netlify/README.md
@@ -27,13 +27,15 @@ export default {
 };
 ```
 
-Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-builds/file-based-configuration) file in the project root. This will determine where to write static assets and functions to based on the `build.publish` settings, as per this sample configuration:
+Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-builds/file-based-configuration) file in the project root. This will determine where to write static assets based on the `build.publish` settings, as per this sample configuration:
 
 ```toml
 [build]
   command = "npm run build"
-  publish = "build/publish/"
+  publish = "build"
 ```
+
+If the `netlify.toml` file or the `build.publish` value is missing, a default value of `"build"` will be used. Note that if you have set the publish directory in the Netlify UI to something else then you will need to set it in `netlify.toml` too, or use the default value of `"build"`.
 
 ## Netlify alternatives to SvelteKit functionality
 
@@ -51,6 +53,20 @@ During compilation a required "catch all" redirect rule is automatically appende
 1. Create your Netlify HTML form as described [here](https://docs.netlify.com/forms/setup/#html-forms), e.g. as `/routes/contact.svelte`. (Don't forget to add the hidden `form-name` input element!)
 2. Netlify's build bot parses your HTML files at deploy time, which means your form must be [prerendered](https://kit.svelte.dev/docs#ssr-and-javascript-prerender) as HTML. You can either add `export const prerender = true` to your `contact.svelte` to prerender just that page or set the `kit.prerender.force: true` option to prerender all pages.
 3. If your Netlify form has a [custom success message](https://docs.netlify.com/forms/setup/#success-messages) like `<form netlify ... action="/success">` then ensure the corresponding `/routes/success.svelte` exists and is prerendered.
+
+### Using Netlify Functions
+
+[Netlify Functions](https://docs.netlify.com/functions/overview/) can be used alongside your SvelteKit routes. If you would like to add them to your site, you should create a directory for them and add the configuration to your `netlify.toml` file. For example:
+
+```toml
+[build]
+  command = "npm run build"
+  publish = "build"
+
+[functions]
+  directory = "functions"
+  node_bundler = "esbuild"
+```
 
 ## Advanced Configuration
 
@@ -76,7 +92,8 @@ The default options for this version are as follows:
 ```js
 {
 	entryPoints: ['.svelte-kit/netlify/entry.js'],
-	outfile: `pathToFunctionsFolder/render/index.js`,
+	// This is Netlify's internal functions directory, not the one for user functions.
+	outfile: '.netlify/functions-internal/__render.js',
 	bundle: true,
 	inject: ['pathTo/shims.js'],
 	platform: 'node'
--- a/packages/adapter-netlify/README.md
+++ b/packages/adapter-netlify/README.md
@@ -27,16 +27,15 @@ export default {
 };
 ```
 
-Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-builds/file-based-configuration) file in the project root. This will determine where to write static assets and functions to based on the `build.publish` and `build.functions` settings, as per this sample configuration:
+Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-builds/file-based-configuration) file in the project root. This will determine where to write static assets and functions to based on the `build.publish` settings, as per this sample configuration:
 
 ```toml
 [build]
   command = "npm run build"
   publish = ".svelte-kit/netlify/build/"
-  functions = ".svelte-kit/netlify/functions/"
 ```
 
-In this example, we placed the `build` and `functions` folders under `.svelte-kit/netlify`. If you specify another location, you will probably also want to add them to your `.gitignore`.
+There should be no need to change the publish location, but if you specify another location, you will probably also want to add it to your `.gitignore`.
 
 ## Netlify alternatives to SvelteKit functionality
 
--- a/packages/adapter-netlify/index.js
+++ b/packages/adapter-netlify/index.js
@@ -32,6 +32,7 @@ export default function (options) {
 			const defaultOptions = {
 				entryPoints: ['.svelte-kit/netlify/entry.js'],
 				// Any functions in ".netlify/functions-internal" are bundled in addition to user-defined Netlify functions.
+				// See https://github.com/netlify/build/pull/3213 for more details
 				outfile: '.netlify/functions-internal/__render.js',
 				bundle: true,
 				inject: [join(files, 'shims.js')],
--- a/packages/adapter-netlify/index.js
+++ b/packages/adapter-netlify/index.js
@@ -95,7 +95,7 @@ function get_publish_directory(utils) {
 		}
 		if (resolve(netlify_config.build.publish) === process.cwd()) {
 			throw new Error(
-				'The publish directory cannot be set to the site root. Please change it to "build" in netlify.toml.'
+				'The publish directory cannot be set to the site root. Please change it to another value such as "build" in netlify.toml.'
 			);
 		}
 		return netlify_config.build.publish;
--- a/packages/adapter-netlify/index.js
+++ b/packages/adapter-netlify/index.js
@@ -1,5 +1,5 @@
 import { appendFileSync, existsSync, readFileSync, writeFileSync } from 'fs';
-import { join } from 'path';
+import { join, resolve } from 'path';
 import { fileURLToPath } from 'url';
 import esbuild from 'esbuild';
 import toml from '@iarna/toml';
@@ -19,7 +19,10 @@ export default function (options) {
 		name: '@sveltejs/adapter-netlify',
 
 		async adapt({ utils }) {
-			const { publish } = validate_config().build;
+			// "build" is the default publish directory when Netlify detects SvelteKit
+			const publish = get_publish_directory(utils) || 'build';
+
+			utils.log.minor(`Publishing to "${publish}"`);
 
 			utils.rimraf(publish);
 
@@ -46,7 +49,7 @@ export default function (options) {
 
 			writeFileSync(join('.netlify', 'package.json'), JSON.stringify({ type: 'commonjs' }));
 
-			utils.log.info('Prerendering static pages...');
+			utils.log.minor('Prerendering static pages...');
 			await utils.prerender({
 				dest: publish
 			});
@@ -65,9 +68,12 @@ export default function (options) {
 
 	return adapter;
 }
-
-function validate_config() {
+/**
+ * @param {import('@sveltejs/kit').AdapterUtils} utils
+ **/
+function get_publish_directory(utils) {
 	if (existsSync('netlify.toml')) {
+		/** @type {{ build?: { publish?: string }} & toml.JsonMap } */
 		let netlify_config;
 
 		try {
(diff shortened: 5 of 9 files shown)


examples/hn.svelte.dev/.netlify/netlify-plugin-pnpm/index.js
```diff
@@ -1,8 +1,9 @@
-export default {
+module.exports = {
```
benmccann, line 1: Why is this being changed to CJS?

examples/hn.svelte.dev/.gitignore
```diff
@@ -2,4 +2,5 @@
 node_modules
 /.svelte
 /build
-/functions
\ No newline at end of file
+/functions
+/.netlify/functions*
```
benmccann, line 6: can we keep these paths under the `build` directory so that users don't have to modify their `.gitignore` from the default `.gitignore` provided with new projects?

ascorbic: These are the directories created by the Netlify CLI whenever it is run, so we can't change where they are. The CLI adds a `.gitignore` line when it's run if there isn't one already. By default it will ignore the whole directory, but as the plugin is in there now in the demo site, I added a line to just ignore the generated functions directories. Users won't need to add this.

examples/hn.svelte.dev/.netlify/netlify-plugin-pnpm/index.js
ascorbic: The Netlify build system doesn't accept ESM for plugins because it `require()`s them from CJS: https://github.com/netlify/build/blob/main/packages/build/src/plugins/child/logic.js#L14

examples/hn.svelte.dev/netlify.toml
```diff
@@ -1,10 +1,13 @@
...
-  functions = ".svelte-kit/netlify/functions/"
 
 [build.environment]
   NPM_FLAGS="--prefix=/dev/null"
 
 [[plugins]]
   package = "/.netlify/netlify-plugin-pnpm"
+
+[dev]
+# Disable svelte-kit dev so we can test functions
```
benmccann, line 12: how does this disable `svelte-kit dev`?

packages/adapter-netlify/index.js
```diff
@@ -55,8 +54,10 @@ export default function (options) {
 			utils.copy_client_files(publish);
 
 			utils.log.minor('Writing redirects...');
-			utils.copy('_redirects', `${publish}/_redirects`);
-			appendFileSync(`${publish}/_redirects`, '\n\n/* /.netlify/functions/render 200');
+
+			const redirectPath = join(publish, '_redirects');
+			utils.copy('_redirects', redirectPath);
+			appendFileSync(redirectPath, '\n\n/* /.netlify/functions/__render 200');
```
benmccann, line 60: does this need to be `functions-internal` instead of `functions`?

packages/adapter-netlify/index.js
```diff
@@ -32,7 +31,7 @@ export default function (options) {
 			/** @type {BuildOptions} */
 			const defaultOptions = {
 				entryPoints: ['.svelte-kit/netlify/entry.js'],
-				outfile: join(functions, 'render/index.js'),
+				outfile: '.netlify/functions-internal/__render.js',
```
benmccann, line 34: Is there some documentation on the Netlify site that talks about the `.netlify` directory? What would go in it, how it's used, etc.?

ascorbic: I'm off this week so can't check, but I don't _think_ the internal functions directory is documented yet. It was only just added a few weeks ago. Generally the `.netlify` directory is for internal use and is used for compiled functions and caches. The new internal functions directory is meant for auto-generated  functions, mainly those created by Netlify build plugins

packages/adapter-netlify/index.js
ascorbic: No, the public path is still `functions`

examples/hn.svelte.dev/netlify.toml
benmccann: I'm having a hard time understanding how gatsby is involved in this or how you'd use gatsby and sveltekit together (they seem like alternatives to one another). 

The changes in this file seem like maybe they should be just temporary changes in your own project for testing purposes and not something we'd want to check in

ascorbic: Sorry, I mean the **Netlify** cli, not Gatsby!! This is what I get for replying to GitHub notifications when I'm on holiday. But you make a fair point. We can remove this but. I can update it next week when I'm back.

ascorbic: I've removed this. Sorry for the confusion

packages/adapter-netlify/README.md
```diff
@@ -27,13 +27,12 @@ export default {
...
 ```
 
-Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-builds/file-based-configuration) file in the project root. This will determine where to write static assets and functions to based on the `build.publish` and `build.functions` settings, as per this sample configuration:
+Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-builds/file-based-configuration) file in the project root. This will determine where to write static assets and functions to based on the `build.publish` settings, as per this sample configuration:
 
 ```toml
 [build]
   command = "npm run build"
   publish = "build/publish/"
-  functions = "build/functions/"
```
benmccann, line 36: I don't know if we can just remove this. Where will it put functions then? We put this here on purpose so that all output would show up under `build` and be automatically gitignored without the user having to edit their `.gitignore` at all. It would be nice if that continued to work for people who are not using the Netlify CLI

packages/adapter-netlify/index.js
benmccann: Is `functions-internal` something we just made up or some directory the Netlify CLI expects and uses in some way? This could use a code comment explaining what this directory is

packages/adapter-netlify/README.md
benmccann: Or at least if we're going to remove this line, should we also remove the `publish` line so that ends up under `.netlify/publish` or something? It seems weird to treat these inconsistently.

Perhaps what we could do is maintain the old behavior and use the directories in these files if they are present and if they are not then we use the directory you're specifying as a default

examples/hn.svelte.dev/.gitignore
benmccann: Do we need both `/functions` and `/netlify/functions` in the `.gitignore` in that case?

packages/adapter-netlify/index.js
ascorbic: It's a special directory that the Netlify CLI looks in for extra functions. It is not yet documented. It's meant to be used for auto-generated functions like these which are expected to be `.gitignore`d. It's normally used by build plugins, but that's effectively what this is. See https://github.com/netlify/build/pull/3213 I'll add a comment to this effect.

packages/adapter-netlify/README.md
ascorbic: The functions will be in `.netlify/functions-internal`. The `functions` directory isn't needed anymore, because we're using the internal functions directory instead, which is hard-coded. The idea is that the `functions` value is for user functions, while `.netlify/functions-internal` is for auto-generated functions. If we remove the `publish` dir it will default to `static` because that's [the value specified for SvelteKit](https://github.com/netlify/framework-info/blob/main/src/frameworks/svelte-kit.json#L17). Should that be changed?

ascorbic: If they user isn't using the Netlify CLI then presumably this won't be an issue anyway, as it wouldn't be using the Netlify preset when building locally as the auto-detection won't trigger it.

examples/hn.svelte.dev/.gitignore
ascorbic: True. There's no need for `/functions` anymore.

packages/adapter-netlify/README.md
ascorbic: Though I suppose the user could have manually specified the preset.

packages/adapter-netlify/index.js
```diff
@@ -31,6 +31,7 @@ export default function (options) {
 			/** @type {BuildOptions} */
 			const defaultOptions = {
 				entryPoints: ['.svelte-kit/netlify/entry.js'],
+				// Any functions in ".netlify/functions-internal" are bundled in addition to user-defined Netlify functions.
```
benmccann, line 34:
```suggestion
				// Any functions in ".netlify/functions-internal" are bundled in addition to user-defined Netlify functions.
				// See https://github.com/netlify/build/pull/3213 for more details
```

packages/adapter-netlify/README.md
benmccann: `static` seems a very odd place to put it. I would recommend we change https://github.com/netlify/framework-info/blob/16d5b7ea82c5c1d783e3711d2f8f1adf2da7eecb/src/frameworks/svelte-kit.json#L17 to something else. Either something under `build` or `.netlify` would probably make the most sense

I don't know what "preset" refers to

Up until now, we'd advised to put the following in `netlify.toml`:
```
  publish = "build/publish/"
  functions = "build/functions/"
```

The idea was that all adapters should put their output under `build` so that it was located in a standard place and already included in the `.gitignore` which includes `build` by default. I'm not going to use the Netlify CLI since I don't want any Netlify-specific functionality in my project, so I don't necessarily want to update my project and have it start putting the output someplace new and lose the ability to control where the output is located

ascorbic: I'm not sure why framework-info has `static` as the default. Was it previously the output location? It is supposed to reflect each framework's default build location. We should certainly change it to whatever is the default.

I understand the desire to have everything under `build`, but the problem with having a user-specified functions directory is as #1249 shows, it means that there's no way for users to create their own functions. This need to separate user-generated and framework-generated functions was what led us to create the internal functions directory. 

Is there a scenario where somebody would be manually specifying the `netlify` adapter, but running `svelte-kit build` locally without the Netlify CLI? I'd assume that anybody using the Netlify adapter would either be running the build in CI, or if building locally they would be using the Netlify CLI to deploy.

benmccann: > Was it previously the output location?

`static` was never the output location. It's typically where you'd put your images, etc. When you do `npm init svelte@next` it will create a new project with a `static` folder that has a few image files in it used by the templates that live in `src/routes`. We should change it to be `build` or something.

> I understand the desire to have everything under build, but the problem with having a user-specified functions directory is as #1249 shows, it means that there's no way for users to create their own functions

I tried reading the Netlify docs on this and just got more confused. The `adapter-netlify` docs currently show:
```
[build]
  functions = "build/functions/"
```

How does Netlify know that the functions are in `build/functions`? Does it read that config value? If so, it appears to be undocumented

[What Netlify shows](https://docs.netlify.com/configure-builds/file-based-configuration/#functions) is:
```
[functions]
  directory =
```

> Is there a scenario where somebody would be manually specifying the netlify adapter, but running svelte-kit build locally without the Netlify CLI?

You have to do `npm run build` in order to do `npm run preview`, which is useful locally

ascorbic: > > Was it previously the output location?
> 
> `static` was never the output location. It's typically where you'd put your images, etc. When you do `npm init svelte@next` it will create a new project with a `static` folder that has a few image files in it used by the templates that live in `src/routes`. We should change it to be `build` or something: [netlify/framework-info#380](https://github.com/netlify/framework-info/pull/380)

Wouldn't `build/publish` be the correct value?

> I tried reading the Netlify docs on this and just got more confused. The `adapter-netlify` docs currently show:
> 
> ```
> [build]
>   functions = "build/functions/"
> ```
> 
> How does Netlify know that the functions are in `build/functions`? Does it read that config value? If so, it appears to be undocumented

`build.functions` is the deprecated config syntax. The preferred syntax now is `functions.directory`, though the old syntax still works.
> 
> > Is there a scenario where somebody would be manually specifying the netlify adapter, but running svelte-kit build locally without the Netlify CLI?
> 
> You have to do `npm run build` in order to do `npm run preview`, which is useful locally

ascorbic: How about this: we keep the default output in build, but remove the doc specifying the functions config. Then we detect if the build command is being run via the netlify cli, and if it is then either change the output to `.netlify/functions-internal` or copy the built files over. That way local builds without the CLI don't generate non-gitignored files, but builds in CI or via the cli get put in the correct place. Meanwhile users are free to create their own Netlify functions and specify whichever functions directory they want.

ascorbic: OK, I've been thinking about this more, and my suggestion is:

- Set `publish` in `netlify.toml` to `build`, and put all the static and client assets there
- Do not set `functions` in `netlify.toml`. If the user has their own functions let them set it
- Set the esbuild outfile to `.svelte-kit/netlify/__render.js`
- When building, if we either detect that we're running via the Netlify CLI (we can do this via env vars), or if `.netlify/state.json` exists (which indicates they've already used the Netlify CLI locally), then copy `.svelte-kit/netlify/__render.js` to `.netlify/functions-internal/__render.js`

If you're happy with this then I can update the PR.

benmccann: What environment variable does the Netlify CLI set?

ascorbic: In CI it sets `NETLIFY` and locally it sets `NETLIFY_LOCAL`

benmccann: ok, that plan basically sounds good to me. we could probably just look for the environment variable and don't need to look for `.netlify/state.json` since it might not always exist

maybe if the environment variable is not present then we can keep it working the same as it does today where we read the locations from the `netlify.toml` so folks not using the netlify CLI can put their build wherever they want

ascorbic: > ok, that plan basically sounds good to me. we could probably just look for the environment variable and don't need to look for `.netlify/state.json` since it might not always exist

The scenario I'm thinking about when checking for `.netlify/state.json` as well as the env vars is where somebody who is using the Netlify CLI to deploy (i.e. just to do the upload), but is calling the build command directly rather than via `netlify build`. In that scenario we wouldn't move the functions into the right place, so when they deploy they'd be missing. If we check for the `state.json` we'd catch those cases and build them to the correct place for upload, and anyone who has that will already have the directory in their `.gitignore`.

> maybe if the environment variable is not present then we can keep it working the same as it does today where we read the locations from the `netlify.toml` so folks not using the netlify CLI can put their build wherever they want

This would mean it wouldn't fix #1249 for those people. It could also catch people who've created custom Netlify functions, specified a directory, and then run a build without the Netlify CLI. In that scenario it would delete all of their custom functions. 😬 

I think it's more consistent to just put them in the `.svelte-kit` subdirectory, where there's no chance of it overwriting user code.

benmccann: > The scenario I'm thinking about when checking for .netlify/state.json as well as the env vars is where somebody who is using the Netlify CLI to deploy (i.e. just to do the upload), but is calling the build command directly rather than via netlify build

I don't understand this scenario. If you built with `npm run build` then no `.netlify/state.json` will be created. How can we know that file will exist? Does that file get checked in? (I was assuming the whole `.netlify` directory is `.gitignore`d)

> It could also catch people who've created custom Netlify functions, specified a directory, and then run a build without the Netlify CLI. 

I don't understand this scenario either. They'd have had to have specify the output directory as the location where their source code resides. Why would anyone do that?

>I think it's more consistent to just put them in the .svelte-kit subdirectory, where there's no chance of it overwriting user code.

What is "them" in this sentence? The generated functions?

ascorbic: > I don't understand this scenario. If you built with `npm run build` then no `.netlify/state.json` will be created. How can we know that file will exist? Does that file get checked in? (I was assuming the whole `.netlify` directory is `.gitignore`d)
> 

If you want to use the Netlify CLI in a local directory you have to first run `netlify link` to associate it with a site. This generates `.netlify/state.json` and adds `.netlify` to `.gitignore`.


> I don't understand this scenario either. They'd have had to have specify the output directory as the location where their source code resides. Why would anyone do that?

I thought that's what you were referring to: i.e. the functions directory? For most Netlify sites that directory contains source code for Netlify functions, but the current SvelteKit behavior (that we're changing here) is to write the generated functions there. Or are you referring to the `publish` directory? If so, then yes I think we'd read that from the toml in all cases, and default to `build` if it's not specified.

> 
> > I think it's more consistent to just put them in the .svelte-kit subdirectory, where there's no chance of it overwriting user code.
> 
> What is "them" in this sentence? The generated functions?

Yes

benmccann: Ok, thanks for clarifying and being patient as I come up-to-speed on Netlify stuff. I'm on board with your plan. Lets always output to `.netlify/functions-internal`

ascorbic: OK, so skip the CLI detection entirely?

benmccann: What would be the benefit of CLI detection? If we're not using the Netlify CLI, I'm assuming that any code `.netlify/functions-internal` will still be used?

ascorbic: Correct. I'll update the PR.

ascorbic: OK, I've updated the PR as follows:

- Always write out the generated functions to `.netlify/functions-internal`
- Read `"publish"` from `netlify.toml`, but if it's missing then default to `"build"`
- Add more warnings and checks, such as ensuring that publish isn't set to the site root (that would delete the site!)
- Updated the README

One thing I'd like to suggest for a follow-up PR is that `cwd` is passed to adapter functions. It's already available in `adapt`, and it would ensure that we resolve all the paths correctly inside adapters. As it stands I'm not sure if these will all work if `cwd` isn't `process.cwd()`

ascorbic:
The build is failing because the netlify.toml has been updated to the new format, but the package hasn't.

packages/adapter-netlify/index.js
```diff
@@ -74,23 +83,25 @@ function validate_config() {
...
 		if (netlify_config.redirects) {
 			throw new Error(
 				"Redirects are not supported in netlify.toml. Use _redirects instead. For more details consult the readme's troubleshooting section."
 			);
 		}
-
-		return netlify_config;
+		if (resolve(netlify_config.build.publish) === process.cwd()) {
+			throw new Error(
+				'The publish directory cannot be set to the site root. Please change it to "build" in netlify.toml.'
```
benmccann, line 98:
```suggestion
				'The publish directory cannot be set to the site root. Please change it to another value such as "build" in netlify.toml.'
```

benmccann:
Thanks for this! Appreciate your knowledge of the Netlify internals and explaining it all to me

PR #2122, opened 2021-08-06
Custom error handler

### LINKED ISSUE #1857: Propagate SSR errors
If an error occurs during SSR, `resolve` returns an object with a `500` status code: https://github.com/sveltejs/kit/blob/babe515f9f5c816fd507920f1f16012263e7a71f/packages/kit/src/runtime/server/index.js#L95-L99

In the case where you're using an error reporting tool like Sentry, you want to be able to get at the original error.

Not 100% sure what the best solution is here. Adding an `error` property to the response object? (Seems icky.) Having a separate hook for handling errors? (The relationship between that and `handle` could be confusing). Not catching the error in the first place? (What happens if there's no `handle` hook, or the custom `handle` hook doesn't try-catch?)

Still brainstorming.

It's important to my coworkers, so it's important to me

andrew-womeldorf:

### Tests

### Changesets

---

This PR replaces `handle_error` with a new hook `handleError`. Closes #1857. This will, by default, print errors as was happening before, or allow the user to define their own error handler.


documentation/docs/04-hooks.md
```diff
@@ -116,3 +116,23 @@ export async function serverFetch(request) {
 	return fetch(request);
 }
 ```
+
+### handleError
+
+This function allows you to handle errors with your own error handler.
+
+For example, you may have a custom logging scheme that you send to an external
+system, and you want to format the error in a manner consistent with your other
```
benmccann, line 125: I would avoid manually wrapping the text by inserting line breaks

documentation/docs/04-hooks.md
```diff
@@ -116,3 +116,23 @@ export async function serverFetch(request) {
...
+logs.
+
+```js
+/** @type {import('@sveltejs/kit').HandleError} */
+export async function handleError(error) {
```
benmccann, line 130: I'd consider making the input an object with a field named `error`. We've done that as a general rule for backwards compatibility. E.g. it might be helpful to later add the request since someone might want to log what URL the error occurred on

```suggestion
export async function handleError({ error }) {
```

benmccann (commented):
awesome. thanks!

documentation/docs/04-hooks.md
andrew-womeldorf: Makes sense, I like that.

packages/kit/src/core/build/index.js
```diff
@@ -392,6 +385,15 @@ async function build_server(
...
 				getSession: hooks.getSession || (() => ({})),
 				handle: hooks.handle || (({ request, resolve }) => resolve(request)),
+				handleError:
+					hooks.handleError ||
+					(({error}) => {
+						if (error.frame) {
+							console.error(error.frame);
+						}
+						console.error(error.stack);
+						error.stack = options.get_stack(error);
```
benmccann, line 395: We should probably do this even if the user provides a `handleError` method

benmccann:
@Rich-Harris will you have access to global `fetch` to send the error off to an error reporting service as you desired with this solution?

packages/kit/src/core/build/index.js
andrew-womeldorf: I'm not really even sure what this line is doing, since `error` isn't returned anywhere. I'm not super familiar with JS - does this modify the reference to `error`, where the caller's `error` object will now have the `error.stack` defined here?

I pasted this from the old `handle_error()` to preserve however that was working before, so I don't know what this is doing here.

andrew-womeldorf:
> @Rich-Harris will you have access to global `fetch` to send the error off to an error reporting service as you desired with this solution?

This may be naive of me, but any logging aggregation tool that I've used, I have an agent running on the server which periodically ships the logs from stdout back to the aggregator. The code in this PR, as I've observed, is only running server-side, so I don't think the client should ever have to be pushing logs back to the aggregator. So I think logging to stdout should be sufficient.

documentation/docs/04-hooks.md
```diff
@@ -116,3 +116,21 @@ export async function serverFetch(request) {
...
+
+### handleError
+
+This function allows you to handle errors with your own error handler.
+
+For example, you may have a custom logging scheme that you send to an external system, and you want to format the error in a manner consistent with your other logs.
+
+```js
+/** @type {import('@sveltejs/kit').HandleError} */
+export async function handleError({error}) {
```
benmccann, line 128: can we go ahead and add the `request` as a parameter here as well? I want to make sure we have access to it everywhere `handleError` is called and there wouldn't be any issues with this design. I think it should be straight-forward, but just want to confirm

packages/kit/src/core/build/index.js
benmccann: > does this modify the reference to error, where the caller's error object will now have the error.stack defined here?

yep!

andrew-womeldorf: Cool, I will look into adding that everywhere.

documentation/docs/04-hooks.md
andrew-womeldorf: I like that idea. I'll give it a shot

packages/kit/src/core/build/index.js
andrew-womeldorf: I don't actually think this is a good idea.

`options.get_stack(error)` is just doing `String(error)`, which I'm seeing actually drops the stack trace that I'm currently receiving.

```js
// before
  err: {
    name: 'FetchError',
    message: 'request to http://doesnotexist.foo/ failed, reason: getaddrinfo ENOTFOUND doesnotexist.foo',
    code: 'ENOTFOUND',
    stack: 'FetchError: request to http://doesnotexist.foo/ failed, reason: getaddrinfo ENOTFOUND doesnotexist.foo\n' +
      '    at ClientRequest.<anonymous> (file:///home/andrew/Work/src/github.com/andrew-womeldorf/kit/packages/kit/dist/install-fetch.js:1510:11)\n' +
      '    at ClientRequest.emit (events.js:315:20)\n' +
      '    at ClientRequest.EventEmitter.emit (domain.js:467:12)\n' +
      '    at Socket.socketErrorListener (_http_client.js:469:9)\n' +
      '    at Socket.emit (events.js:315:20)\n' +
      '    at Socket.EventEmitter.emit (domain.js:467:12)\n' +
      '    at emitErrorNT (internal/streams/destroy.js:106:8)\n' +
      '    at emitErrorCloseNT (internal/streams/destroy.js:74:3)\n' +
      '    at processTicksAndRejections (internal/process/task_queues.js:80:21)'
  }
```

```js
// after
  err: {
    name: 'FetchError',
    message: 'request to http://doesnotexist.foo/ failed, reason: getaddrinfo ENOTFOUND doesnotexist.foo',
    code: 'ENOTFOUND',
    stack: 'FetchError: request to http://doesnotexist.foo/ failed, reason: getaddrinfo ENOTFOUND doesnotexist.foo'
  }
```

benmccann: That's in `build`. In `dev` it does something much different. Also, the build code says it does that for security, so I'm not sure we'd just want to drop it. I don't think `new String` should drop the successive lines

documentation/docs/04-hooks.md
benmccann: we should add `request` to the docs

packages/kit/src/core/dev/index.js
```diff
@@ -333,19 +333,21 @@ async function create_handler(vite, config, dir, cwd, manifest) {
...
-								console.error(colors.gray(error.stack));
-							}
-						},
 						hooks: {
 							getSession: hooks.getSession || (() => ({})),
 							handle: hooks.handle || (({ request, resolve }) => resolve(request)),
+							handleError:
+								hooks.handleError ||
+								(({ error }) => {
+									vite.ssrFixStacktrace(error);
```
benmccann, line 342: this should always be done as well. we don't want this to get dropped if the user provides an implementation

packages/kit/src/core/build/index.js
andrew-womeldorf: Okay, I've added a solution that I think should be fine. Let me know if this is what you had in mind

packages/kit/src/core/dev/index.js
andrew-womeldorf: Also check here, if the solution I created matches what you were thinking

packages/kit/src/core/build/index.js
```diff
@@ -392,6 +385,18 @@ async function build_server(
 			const get_hooks = hooks => ({
 				getSession: hooks.getSession || (() => ({})),
 				handle: hooks.handle || (({ request, resolve }) => resolve(request)),
+				handleError: ((args) => {
```
benmccann, line 388: nit: `args` is so much like `arguments` that I got a little confused at first: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments

I might call it `opts` instead

benmccann (approved):

andrew-womeldorf:
I'm not sure about the failing test from the last workflow run. all of the tests pass locally

benmccann:
don't worry about it. most of the jobs passed. if it's just one failing then it's a flaky test

Rich-Harris:
Thank you — made some tweaks in #2193, will close this in favour of that

Issue #2116, opened 2021-08-06
PR numbers as links in changelogs


jthegedus:
### Describe the problem

It's difficult to determine which version a merged PR is in.

I believe the changeset title is the PR title, though I am not sure this is strictly always the case.

As a maintainer of an external adapter, when I see a change to the Adapter API I often have to check the changelog of Kit (which navigating to is annoying to begin with) and read for the changesets since the last version my adapter was specifically built for. I'd rather just ctrl+f for the PR number.

### Describe the proposed solution

Include PR numbers as a link to the PR in the changelog for each changeset

### Importance

would make my life easier

### Additional Information

I imagine 1.0 will resolve most of this, though I wish to contribute to the Adapter API discussion before 1.0.

benmccann:
We should copy https://github.com/sveltejs/vite-plugin-svelte/blob/main/scripts/changelog-github-custom.js which would fix this. Happy to accept a PR to do that

jthegedus:
I would also add the request to this PR to add the packages with changelog links in the root README :point_down: 

> From the [sveltejs/vite-plugin-svelte](https://github.com/sveltejs/vite-plugin-svelte) README

## Packages

| Package                                                     | changelog                                             |
| ----------------------------------------------------------- | ----------------------------------------------------- |
| [@sveltejs/vite-plugin-svelte](packages/vite-plugin-svelte) | [changelog](packages/vite-plugin-svelte/CHANGELOG.md) |

benmccann:
yeah, that's not a bad idea

rmunn:
I've been able to backtrack changelog entries to PRs via commit hashes. By looking up, for example, https://github.com/sveltejs/kit/commit/5911b1c7 from the 1.0.0-next.140 changelog entry, I see the commit message is "[fix] consider protocol-relative URLs as external (#2062)" and then I can look up #2062 and see the PR discussion and the linked issue.

It should be trivial to write a script to take the existing changelog with commit hashes, grab the first line of the log message for that hash, and extract a PR number from the end of that first line. As long as the commit was created using GitHub's "Squash and merge" button, the end of the commit message will be `(#NNNN)` where NNNN is the PR number. It's possible that not every changelog entry was made from a commit that used the "Squash and merge" button, but that method should allow backfilling PR numbers into most of the existing changelog entries, if desired.

benmccann:
[...]

PR #2126, opened 2021-08-07
[docs] list packages, link repos & changelogs

jthegedus:
Addresses comment from #2116 

### Tests
NA

### Changesets
NA


jthegedus:
I've no idea why that test failed given the code I changed :/

ignatiusmb (approved):
Don't worry about the tests, some of them are flaky. Lint is still passing, so it's fine.

Issue #1576, opened 2021-05-28
Nested layouts are rendering inside parent layouts for error pages


moisesbites:
**Describe the bug**

I have a __layout.svelte (1)  in the root of route directory. I have another __layout.svelte (2) more specific in a internal route. So, when I click in a link to a route that not exists yet, the error default error page is renderized in a blank page sometimes, or in the `<slot>` of __layout.svelte (2) in another times. But, if after that I click in a link to a valid route, the new page is renderized in a new `<slot>`, as if it were being rendered in the slot (1) , but still maintaining the rendering of slot (2), rendering twice .

**Logs**

Don't have server logs for that.
There is no significant log in console.

**To Reproduce**

![image](https://user-images.githubusercontent.com/8934577/120000292-ab077f80-bfa0-11eb-9104-19ff7c3be7ad.png)
![image](https://user-images.githubusercontent.com/8934577/120000461-d4c0a680-bfa0-11eb-9c49-bdb84a862bf7.png)
![image](https://user-images.githubusercontent.com/8934577/119999300-9676b780-bf9f-11eb-9d23-38cabb66fb1b.png)

```JSON
  "devDependencies": {
    "@sveltejs/adapter-node": "^1.0.0-next.20",
    "@sveltejs/kit": "next",
    "eslint": "^7.27.0",
    "eslint-config-prettier": "^8.3.0",
    "eslint-plugin-svelte3": "^3.2.0",
    "prettier": "~2.3.0",
    "prettier-plugin-svelte": "^2.3.0",
    "svelte": "^3.38.2",
    "svelte-windicss-preprocess": "^4.0.7",
    "windicss": "^3.0.12"
  },
```
**Expected behavior**
I click in a link. The page not exist. The error page is renderized. I click in a new link, to a page that exists. The page is new renderized.

**Information about your SvelteKit Installation:**

<details>
  <summary>Diagnostics</summary>

```
System:
    OS: Linux 5.8 Linux Mint 20.1 (Ulyssa)
    CPU: (4) x64 Intel(R) Core(TM) i5-2400 CPU @ 3.10GHz
    Memory: 12.14 GB / 23.38 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 16.2.0 - ~/.nvm/versions/node/v16.2.0/bin/node
    npm: 7.13.0 - ~/.nvm/versions/node/v16.2.0/bin/npm
  Browsers:
[...]

PR #2127, opened 2021-08-07
Fix/error redirect

### LINKED ISSUE #1574: Why is redirect not supported from __error.svelte cmpt?
I have a small [reproducible](https://github.com/alexbjorlig/sveltekit-redirect-question) example here.

In the `__error.svelte` cmpt I have thrown in a redirect like this in the load function:

```ts
 return {
                status: 302,
                redirect: '/login'
            }
```

But this is not working. Is this a bug, not allowed - or do I misunderstand something here?

iOiurson:
Fixes #1574

- [x] server redirect on errors
- [x] server redirect on 404
- [x] client redirect on errors
- [x] client redirect on 404
- [x] ~~client redirect only on browser (edgy)~~

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -724,16 +724,23 @@ export class Renderer {
 			context: {}
 		});
 
-		const branch = [
-			node,
-			await this._load_node({
-				status,
-				error,
-				module: await this.fallback[1],
-				page,
-				context: (node && node.loaded && node.loaded.context) || {}
-			})
-		];
+		const errorNode = await this._load_node({
+			status,
+			error,
+			module: await this.fallback[1],
+			page,
+			context: (node && node.loaded && node.loaded.context) || {}
+		});
+
+		if (errorNode && errorNode.loaded && errorNode.loaded.redirect) {
+			return {
+				redirect: errorNode.loaded.redirect,
+				props: {},
+				state: this.current
+			};
+		}
+
+		const branch = [node, errorNode];
 
 		return await this._get_navigation_result_from_branch({ page, branch });
 	}
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -661,6 +661,14 @@ export class Renderer {
 								context: node_loaded.context
 							});
 
+							if (error_loaded && error_loaded.loaded && error_loaded.loaded.redirect) {
+								return {
+									redirect: error_loaded.loaded.redirect,
+									props: {},
+									state: this.current
+								};
+							}
+
 							if (error_loaded && error_loaded.loaded && error_loaded.loaded.error) {
 								continue;
 							}
--- a/packages/kit/src/runtime/server/page/respond_with_error.js
+++ b/packages/kit/src/runtime/server/page/respond_with_error.js
@@ -18,7 +18,9 @@ import { coalesce_to_error } from '../../utils.js';
  *   status: number;
  *   error: Error;
  * }} opts
+ * @returns {Promise<import('types/hooks').ServerResponse>}
  */
+
 export async function respond_with_error({ request, options, state, $session, status, error }) {
 	const default_layout = await options.load_component(options.manifest.layout);
 	const default_error = await options.load_component(options.manifest.error);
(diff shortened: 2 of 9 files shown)


packages/kit/src/runtime/client/renderer.js
```diff
@@ -658,6 +658,14 @@ export class Renderer {
 								context: node_loaded.context
 							});
 
+							if (error_loaded?.loaded?.redirect) {
```
benmccann, line 661: `?.` isn't supported in Node 12, which is a platform we still support, so I'd write it the longhand way for now

benmccann: oh, I missed that this is client code. still iOS 13 doesn't support it and I don't see it used elsewhere in `runtime/client`, so I think I'd still avoid it

packages/kit/src/runtime/server/page/respond_with_error.js
```diff
@@ -11,7 +11,9 @@ import { coalesce_to_error } from '../../utils.js';
  *   status: number;
  *   error: Error;
  * }} opts
+ * @returns {Promise<import('types/hooks').ServerResponse>}
  */
+
```
benmccann, line 16: no need to insert a new line here

iOiurson:
I'm trying to handle the "edgy" case, which is client redirection on `start`, when no SSR is active.

It happens here:
https://github.com/sveltejs/kit/blob/c3ca3fbb05b208af4a30c8b311074fd95d161304/packages/kit/src/runtime/client/renderer.js#L191-L196

But in the end, i don't believe this is a "real" case in this PR, since by that point, the node is already loaded correctly with the work done in the previous commits.

benmccann:
yeah, it looks like you've probably covered all the cases. basically just need to rebase and add tests at this point

iOiurson:
> The nested `__error.svelte` page i added for redirections is not taken into account, `basics/routes/__error.svelte` is instead, which makes my simple test fail, obviously...

Actually, #1576 is the reason why my test does not work :/

packages/kit/test/apps/basics/src/routes/__error.svelte
```diff
@@ -1,6 +1,13 @@
 <script context="module">
 	/** @type {import('@sveltejs/kit').ErrorLoad} */
 	export function load({ status, error }) {
+		if (error?.message.includes('/redirect/nowhere')) {
```
benmccann, line 4: I expect `?.` will cause this test to fail on the CI on Node 12

packages/kit/test/apps/basics/src/routes/redirect/crashing/__error.svelte
```diff
@@ -0,0 +1,11 @@
+<script context="module">
+	/** @type {import('@sveltejs/kit').ErrorLoad} */
+	export async function load() {
+		return {
+			status: 307,
+			redirect: './c'
```
iOiurson, line 6: This line suggests a redirection to `redirect/crashing/c`, but it actually redirects to `redirect/c`.

It might creation confusion.

benmccann: Yeah, I'd have expected it to redirect to `/redirect/crashing/c`. Is it expected that it doesn't or does it seem like a bug? Maybe this is an instance of https://github.com/sveltejs/kit/issues/2140

iOiurson: It seems to be an instance of #2140 indeed.

The suggested fix of setting `trailingSlash` to `'always'` fixes it for the `crashing/__error` redirection.
But then `redirect` tests are now failing, since `redirect/a` tries to redirect to `redirect/a/b` instead of `redirect/a`. This seems to be the case for any non-`__error` redirections.

My concern is then that relative redirection behaviour is not consistent between an `__error` redirection and a regular page redirection.

I find it weird, because i don't get why the redirection path would be resolved differently if originated from an `__error` page...

benmccann:
One thing I realized is that people can actually handle the 404 case themselves today. You can just create a file like `src/routes/[zzz_catchall].svelte`:
```
<script context="module">
	/**
	 * @type {import('@sveltejs/kit').Load}
	 */
		return {
			status: 307,
			redirect: '/'
		};
	}
</script>
```

iOiurson:
> One thing I realized is that people can actually handle the 404 case themselves today. You can just create a file like `src/routes/[zzz_catchall].svelte`:

That's indeed an alternative. But I think in my case I'd rather handle it the `__error` page.

packages/kit/test/apps/basics/src/routes/redirect/crashing/__error.svelte
iOiurson: Maybe it'd be best to redirect to `../c` in `redirect/crashing/__error`, and let this test fail (and PR open) until the inconsistency is fixed.

iOiurson: I need to investigate more, but i think that redirection path resolution is broken when using relative URLs only on `__layout`, `__error`, and `index` pages.

iOiurson: See #2235

iOiurson: Ok, following discussion in #2235, the `redirect` of `crashing/__error` should be `./c`.

Tests should all pass now.

iOiurson:
All tests should be passing now. I'm not sure why deploys are failing.

benmccann:
Deploy failures are unrelated to your change. We need to fix. I think Netlify is caching old versions of the Netlify adapter

benmccann:
Thank you so much for all the effort you put into this PR. I talked about it with the rest of the maintainers at today's maintainer's meeting and the consensus was that isn't something folks wanted to add. I put more details about that on the issue here: https://github.com/sveltejs/kit/issues/1574#issuecomment-919336290

I'm terribly sorry as I know this must be a disappointing outcome given all the effort you've put into this PR and it would have been a lot nicer to have clarified that up front. I really appreciate everything you did to move this forward and help us get to a resolution on it though! Even though this didn't get merged it is nice that we now have an answer for folks on how they can accomplish this and we were able to knock an item off the 1.0 list thanks to the discussion this spurred

iOiurson:
I'd be lying if i said i'm not a little bit disappointed, but I'll get over it :)

Happy to help, for sure I'll try to find another issue to knock down, maybe something more consensual :D

benmccann:
that would be awesome. I went through and labeled the issues with "help wanted" where I think we're fairly agreed on a solution and it just needs to be implemented

thanks again for all the help in looking at this one!

PR #2128, opened 2021-08-07
[fix] fallback should still be generated when prerender is disabled

### LINKED ISSUE #1588: adapter-static `fallback` page not generated when `prerender.enabled` is `false`
**Describe the bug**
Using the adapter static in "SPA" configuration doesn't output any html entry point even though the build completes without errors. All other bundled assets are created as far as i can tell. `/build` contains

```
/_app
  |--/assets
  |--/chunks
  |--/pages
  |--...misc.js
_redirects
```

**Logs**
Build completes successfully with logs ending in 
```
vite v2.3.3 building SSR bundle for production...
✓ 38 modules transformed.
.svelte-kit/output/server/app.js   396.41kb

  ✔ done
✨  Done in 3.72s.
```

**To Reproduce**
svelte.config.js

```js
import preprocess from 'svelte-preprocess';

const config = {
  // Consult https://github.com/sveltejs/svelte-preprocess
  // for more information about preprocessors
  preprocess: preprocess(),
  kit: {
    target: '#svelte',
    adapter: adapter({
      fallback: "index.html"
    }),
    prerender: {
      enabled: false
    },
    ssr: false,
    vite: {
      ssr: {
        noExternal: ["chart.js"]
      },
    }
  }
};

export default config;
```

To help us help you, if you've found a bug please consider the following:

- If possible, we recommend creating a small repo that illustrates the problem.
- Reproductions should be small, self-contained, correct examples – http://sscce.org.

Occasionally, this won't be possible, and that's fine – we still appreciate you raising the issue. But please understand that Svelte is run by unpaid volunteers in their free time, and issues that follow these instructions will get fixed faster.

**Expected behavior**
Entrypoint index html should be generated in the root of the build dir

**Stacktraces**
N/A

<details>

  System:
    OS: macOS 11.1
    CPU: (8) arm64 Apple M1
    Memory: 520.03 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.1.0 - ~/.nvm/versions/node/v16.1.0/bin/node
    Yarn: 1.22.10 - /opt/homebrew/bin/yarn
    npm: 7.11.2 - ~/.nvm/versions/node/v16.1.0/bin/npm
  Browsers:
    Brave Browser: 91.1.25.68
    Firefox: 88.0.1
    Safari: 14.0.2
  npmPackages:
    @sveltejs/kit: ^1.0.0-next.110 => 1.0.0-next.110
    svelte: ^3.34.0 => 3.38.2

- Adapter-Static
</details>

**Severity**
How severe an issue is this bug to you? Is this annoying, blocking some users, blocking an upgrade or blocking your usage of SvelteKit entirely?

This is medium high severity, we are unable to deploy our svelte kit SPA, although the dev server works fine

Theo-Steiner:
## Fixes #1588 

The bug described in #1588 is about the adapter-static in SPA-mode (-> fallback set) not outputting the {fallback}.html file into the build directory when config.kit.prerender.enabled is set to ``false``.

This PR is an easy fix for this behavior, as config.kit.prerender.enabled can be ignored in the SPA mode. It bypasses the check for ``config.kit.prerender.enabled`` in the getter for the prerender function by modifying the expression to be true by default when a fallback is defined.

### You can see that this fixes the bug on the branch ``bug-demonstration`` on my fork of sveltejs/kit.

For this, 
1. clone the repo ``git clone https://github.com/Theo-Steiner/kit.git``
2. run ``checkout bug-demonstration``, ``pnpm i``, ``pnpm build``
3. navigate to `` cd examples/kit-bug`` 
4. execute ``pnpm build`` and do ``ls build`` and you will see that index.html was in fact created!

I did not include a test, since I honestly don't know how I would test for this behavior. 

### Tests

### Changesets

### DIFF
--- a/.changeset/grumpy-onions-applaud.md
+++ b/.changeset/grumpy-onions-applaud.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/kit": patch
+---
+
+[fix] fallback should still be generated when prerender is disabled
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -225,7 +225,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 			});
 
 			if (is_html && config.kit.prerender.crawl) {
-				const cleaned = clean_html(/** @type {string} */(rendered.body));
+				const cleaned = clean_html(/** @type {string} */ (rendered.body));
 
 				let match;
 				const pattern = /<(a|img|link|source)\s+([\s\S]+?)>/gm;
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -81,7 +81,8 @@ function chooseErrorHandler(log, onError) {
 const OK = 2;
 const REDIRECT = 3;
 
-/** @param {{
+/**
+ * @param {{
  *   cwd: string;
  *   out: string;
  *   log: Logger;
@@ -89,8 +90,13 @@ const REDIRECT = 3;
  *   build_data: import('types/internal').BuildData;
  *   fallback?: string;
  *   all: boolean; // disregard `export const prerender = true`
- * }} opts */
+ * }} opts
+ */
 export async function prerender({ cwd, out, log, config, build_data, fallback, all }) {
+	if (!config.kit.prerender.enabled && !fallback) {
+		return;
+	}
+
 	__fetch_polyfill();
 
 	const dir = resolve_path(cwd, `${SVELTE_KIT}/output`);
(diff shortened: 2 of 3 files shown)


benmccann:
Aha! Thank you. It's tremendously helpful to know what the issue here is. I don't think this is quite the right fix though because it will cause your whole site to be crawled for pre-rendering, which isn't quite what you want if you set `prerender.enabled: false`

Perhaps this would also need to change `prerender.js` to:
```
if (fallback) {
  // don't change this 
} else {
        // move the crawling `for` loop from above fallback to an else block
  	for (const entry of config.kit.prerender.pages) {
             ....
	}
}
```

Theo-Steiner:
@benmccann I made the crawling for loop conditional as you suggested! Thank's for your input =)

benmccann (approved):

benmccann:
@Theo-Steiner lgtm, but the tests are failing and will need to be updated. I added a changeset, so you might want to do a `git pull` first to pull that change down locally before updating the code

PR #2137, opened 2021-08-09
[fix] enable Vite pre-bundling except for Svelte packages

benmccann:
This is part of why you can't use any Svelte packages with CJS dependencies (https://github.com/vitejs/vite/issues/3024), which is probably the top issue we encounter using Vite.

We need to exclude Svelte packages due to https://github.com/vitejs/vite/issues/3910. Pnpm users may need to either (1) set pnpm shamefully-hoist=true or (2) install that dependency in package.json so we can import it.

This PR alone doesn't actually fix much due to https://github.com/vitejs/vite/issues/3910. However, it does allow me to make more progress in working on Vite issues because it allows me to use the `kit.vite.optimizeDeps` options. When `optimizeDeps` is set to `[]` then none of the other options work and there's no way to unset that from the config. Since it shouldn't actually be set anyway, let's go ahead and remove it

### DIFF
--- a/.changeset/cuddly-files-decide.md
+++ b/.changeset/cuddly-files-decide.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] enable Vite pre-bundling except for Svelte packages
--- a/packages/kit/src/core/build/index.js
+++ b/packages/kit/src/core/build/index.js
@@ -2,7 +2,7 @@ import fs from 'fs';
 import path from 'path';
 import { rimraf } from '../filesystem/index.js';
 import create_manifest_data from '../../core/create_manifest_data/index.js';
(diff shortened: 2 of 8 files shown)


dominikg:
as mentioned in https://github.com/sveltejs/vite-plugin-svelte/issues/125  i think this is not something limited to sveltekit and having it in vite-plugin-svelte would make it more useful.

benmccann:
I'm totally happy to have it live there. I'll go ahead and merge this for now since we'd also want to move `ssr.noExternal`. When we get the code into `vite-plugin-svelte` we can remove it from here

benmccann:
See https://github.com/sveltejs/svelte-preprocess/issues/362#issuecomment-896142299 for more details about that issue

Issue #2108, opened 2021-08-05
kit.paths.base config setting seems like doesnt work


istarkov:
### Describe the bug

I tried to set `kit.paths.base`  option at `svelte.config.js`, but seems like it has zero affect.

Documentation says that.
```markdown
base — a root-relative path that must start, but not end with / (e.g. /base-path). 
This specifies where your app is served from and allows the app to live on a non-root path
```

i.e. having svelte.config.js like
```javascript
{
  kit: {
    target: '#svelte',
    paths: {
      base: '/base',
      assets: '',
    },
  },
};
```

 I'm expecting to access all routes paths like `/base/about` etc
but getting __Not found: /base/about__ 

_but /about is working independent of kit.paths.base_

I thought it must be something like as https://nextjs.org/docs/api-reference/next.config.js/basepath but or its some other feature or its not working as expected

### Reproduction

https://github.com/realadvisor/kit-paths-base-issue

```bash
npm init svelte@next my-app
# use defaults
```

then change config on 

```javascript
/** @type {import('@sveltejs/kit').Config} */
const config = {
  kit: {
    // hydrate the <div id="svelte"> element in src/app.html
    target: '#svelte',
    paths: {
      base: '/base',
      assets: '',
    },
  },
};
export default config;
```

```bash
npm install
npm run dev
```

open browser at http://localhost:3000/base/about and see 404

### System Info

```shell
System:
    OS: Linux 5.4 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (8) x64 AMD EPYC Processor
    Memory: 3.50 GB / 15.27 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 14.16.0 - ~/.volta/tools/image/node/14.16.0/bin/node
    Yarn: 1.22.10 - ~/.volta/tools/image/yarn/1.22.10/bin/yarn
    npm: 6.14.11 - ~/.volta/tools/image/node/14.16.0/bin/npm
  Browsers:
    Chrome: 91.0.4472.114
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.142 
    svelte: ^3.34.0 => 3.42.1 
```
```


### Severity

serious, but I can work around it

### Additional Information

workaround is moving routes at build time into subdirectory but that is so so behaviour.

benmccann:
[...]

Issue #326, opened 2021-01-14
Some way to get advance warning of incorrect asset paths


Rich-Harris:
Scenario: you have a file, like `static/images/potato.jpg`, referenced like so:

```html
<img alt="potato" src="/images/potato.jpg">
```

But you also have a `paths` config:

```js
// svelte.config.js
module.exports = {
  kit: {
    paths: {
      assets: 'https://cdn.example.com/my/bucket'
    }
  }
};
```

When the app is deployed, the image will be broken, because it should have been written like so:

```html
<script>
  import { assets } from '$app/paths';
</script>

<img alt="potato" src="${assets}/images/potato.jpg">
```

But because `assets === '/.'` during development, the wrong version is indistinguishable from the right version.

Not totally sure what the right solution is. We don't want to force people to always use `${assets}` in the case where they don't use the paths config, but it causes confusing problems if it's absent when there _is_ a paths config. Maybe if there's a paths config, `assets` should be set during development to something like `/_static_assets` to force you to use the prefix, but otherwise it should behave as it currently does?

Rich-Harris:
This can be closed — in dev, `assets` is replaced with `_svelte_kit_assets`

PR #2145, opened 2021-08-09
Fix $app/paths import not working at dev mode inside context="module"

istarkov:
ref #2108

I've started playing with #2108 and found that 

```svelte
<script context="module">
import { base, assets } from '$app/paths';
...
```

Always return default values at dev and preview at serverside.

This is attempt to fix above.

Steps.
- [x] add failing test (created standalone for now as setting paths.base at options etc breaks everything)
- [x] add fix 

Without above its hard to proceed with full paths.base support. As { base } import at module level would be needed for tests.


istarkov:
@benmccann I can't reproduce locally node 12 error above. What can it be?

packages/kit/src/core/preview/index.js
```diff
@@ -63,7 +60,14 @@ export async function preview({
 	const vite_config = (config.kit.vite && config.kit.vite()) || {};
 
 	const server = await get_server(use_https, vite_config, (req, res) => {
-		const parsed = parse(req.url || '');
+		const initial_url = req.url;
+		const { assets } = config.kit.paths;
+
+		// Emulate app.use(`${assets}/`, sirv(...))
+		req.url =
+			assets.length > 1 && assets !== '/.' && req.url?.startsWith(`${assets}/`)
```
benmccann, line 68: Node 12 error is probably `?.`

packages/kit/test/apps/options-paths-base/svelte.config.js
```diff
@@ -0,0 +1,29 @@
...
+		host: 'example.com',
+		trailingSlash: 'always',
+		vite: {
+			build: {
+				minify: false
+			},
+			clearScreen: false
+		},
+		paths: {
+			base: '/path-base',
```
benmccann, line 23: Would it be possible to just set this in the existing options test? I think Rich had some concern about adding tests for new apps because it might slow things down

istarkov: Done

istarkov:
ps: locally atop of this PR, I have fix for #2108 in dev and preview modes

In short I changed manifest routes pattern generation to include basePath, and cleared few places where basePath were spliced from pathname.

Even such solution doesnt allow dynamic basePath now, but it can be extended to be dynamic by changing pattern match logic a little bit.

Dsupreme (approved):

packages/kit/test/apps/options/source/pages/host/index.svelte
```diff
@@ -1,7 +1,9 @@
 <script context="module">
+	import { base } from '$app/paths';
+
 	/** @type {import('@sveltejs/kit').Load} */
 	export async function load({ page, fetch }) {
-		const res = await fetch('/host.json');
+		const res = await fetch(`${base}/host.json`);
```
benmccann, line 6: Is this something you're going to want to do for every URL within the app or should we add a `base` html tag or something?

istarkov: It's same as we should do with links. 
Otherwise all `<a href="/any-path'` without `${base}` would point to wrong url. 

Without $base in url even we can intercept clicks and route such urls well, 
for crawlers, for open in new window that url would be wrong.

So every 'a' tag whould be like `<a href="{base}/path"`. _Not a big deal as Link component can be created_ 

`fetch` just same to be consistent.

istarkov: btw I havent knew about base tag. :-) so probably its much better. 

Would check tomorrow

Rich-Harris:
Thank you — the current behaviour was _semi_-intentional, but a little bit confusing (and error-prone, since you could easily forget to prefix URLs with `base` or `assets`), and this is a good opportunity to clarify it.

I think the `base` stuff makes sense. For `assets` it's a little trickier — a common use case for this option is to have your static assets hosted on a CDN far from your actual app, and as such it doesn't really make sense in dev or preview (since up-to-date assets won't have been uploaded to the CDN, and since it prevents things like offline development).

So I wonder if it makes sense to have a fake `assets` path during dev and preview? I wrote this up in #326 but then never got round to doing anything about it. We'd probably need to avoid changing `req.url`, since IIUC `static_handler` would respond to requests for `/photo.jpg` as _well_ as `${assets}/photo.jpg`, which we want to avoid.

istarkov:
Probably one of options would be to simplify a little bit.

In case of relative `assets` what the use case for asset to be not equal to `base` ?

For example our use case is PR deployments, because of 3rd party lib referer restrictions (no wildcards supported etc) we can't deploy each PR on own sub domain. So what we really need that same code work on any base path depending on env variable. 

With our usecase we can set `<base href="//{host}/{base}">` (thanks @benmccann  I totally forgot about  it) and everything like `<a href='/blabla'` or `` and no need in fake path (as base tag would be enough).

In case of non relative assets fake path can be an option.

Rich-Harris:
> In case of relative `assets` what the use case for asset to be not equal to `base` ?

🤯 this is... a great point. I don't think there is a good use case. We should probably just prohibit relative asset paths and use a fake one during development and preview. I have a bit of time today so will work on this now.

I've found `<base>` to be a pain in the neck, so in the long run I'd love for us to support dynamic basepaths. There's an issue for that — https://github.com/sveltejs/kit/issues/595 — but it can be addressed separately from this

Rich-Harris:
closing in favour of #2189 — thanks

PR #2158, opened 2021-08-10
[chore] type functions as methods for semantics

JeanJPNM:

### Tests

### Changesets

This is a small change, replaces all occurences of function properties by methods. 

Before:
before

After:
after

Edit: images modified to make example more clear

### DIFF
--- a/.changeset/cuddly-pens-buy.md
+++ b/.changeset/cuddly-pens-buy.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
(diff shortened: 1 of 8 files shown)


benmccann (approved):
I was staring at the screenshot for a really long time trying to figure out the difference - it looks like cube vs rectangle icon in front of `handler`. Neither mean anything to me, but this change seems fine to me :smile:

ignatiusmb (approved):
Thanks! This was in one of my todo list for a while.

PR #2160, opened 2021-08-10
[fix] Import fallback components when they are actually required

### LINKED ISSUE #2130: __layout.svelte may run twice after browser reload in dev
### Describe the bug

This issue only affects the context script for `src/routes/__layout.svelte`. Other layout files are not affected, and #1214 is different.

At first, everything works fine across browser reloads, updating some other files isn't a problem at all. But updating the top level layout file, or any of it's imported dependencies (components or modules) will trigger this issue, to cause the context script to run twice when I reload the browser's tab. Clearing the browser's cache won't help until I reload the development server.

**Before the update**
before update

**After the update (notice the abnormal additional log changing positions)**
update   reload 1
update   reload 2
update   reload 3

### Reproduction

- sveltekit skeleton project: `npm init svelte@next __layout-ctx-run-twice`
- add the top level layout file containing:
```svelte
<!-- src/routes/__layout.svelte -->
<script context="module">
    console.log('%c module: this may runs twice.', 'color: purple; font-weight: bold;')

    export function load () {
        console.log('%c load: this runs once.', 'color: magenta; font-weight: bold;')
        return {}
    }
</script>

<script>
    console.log('%c instance: this runs once.', 'color: blue; font-weight: bold;')
</script>

<slot></slot>
```
- start the development server: `npm run dev`
- navigate to `localhost:3000`, all the logs are logged once respecting the same order across reloads, and they're all coming from **`__layout.svelte? [sm]`**
- force update `src/routes/__layout.svelte`, then reload the browser tab.
- inspect the logs again to notice the first log in code (purple) is logged twice. One is coming from **`__layout.svelte? [sm]`** as the other logs, and the abnormal one is coming from **`__layout.svelte`**
- reload as much as you like and you'll notice that the abnormal log may not respect the same log order.
- finally, reload the development server, and the issue will disappear.

### Logs

_No response_

### System Info

```shell
System:
    Memory: 3.28 GB / 15.56 GB
    Container: Yes
    Shell: 5.0.3 - /bin/bash
  Binaries:
    Node: 14.17.1
    npm: 6.14.13
  Browsers:
    Chrome: 87.0.4280.141
    Firefox: 68.8.0esr
```

### Severity

annoyance

### Additional Information

Tested using both Chrome and Firefox.

### LINKED ISSUE #2154: Layout Stack not Fully Reset for src/routes/__layout.svelte
### Describe the bug

According to the official sveltekit documentation, adding __layout.reset.svelte inside a nested folder should reset layout stack and prevent inheriting from upper level layouts. Unfortunately, the browser is receiving some logs from the **module context** of `__src/routes/__layout.svelte`. ssr logs are clean and not affected.

### Reproduction

- [https://github.com/jsprog/sveltekit-issue--layout-stack-not-fully-reset](https://github.com/jsprog/sveltekit-issue--layout-stack-not-fully-reset)
- the expected behavior is: `src/routes/login/__layout.reset.svelte` should reset `src/routes/__layout.svelte`.
- after visiting `localhost:3000/login`, you'll notice the log from the module context of `src/routes/__layout.svelte` is still there.

### Logs

_No response_

### System Info

```shell
System:
    Memory: 10.01 GB / 15.56 GB
    Container: Yes
    Shell: 5.0.3 - /bin/bash
  Binaries:
    Node: 14.17.1 -
    npm: 6.14.13 -
  Browsers:
    Chrome: 92.0.4515.131
    Firefox: 68.8.0esr
```

### Severity

annoyance

### Additional Information

- ssr logs are not affected by this.
- other nested layout files are not affected by this, and their children `__layout.reset.svelte` are fully resetting them.

anudeepreddy:
Fixes #2154 and fixes #2130

The root cause for the above two issues is in the manifest file which exports the routes and the fallback. The fallback components (`__layout.svelte` and `error.svelte`) are always imported because of the way they are exported from the manifest. Always importing the components causes module-context code to run even when the layout is reset.
```js
const c = [
	() => import("..\\..\\..\\src\\routes\\__layout.svelte"),
	() => import("..\\components\\error.svelte"),
	() => import("..\\..\\..\\src\\routes\\index.svelte"),
	() => import("..\\..\\..\\src\\routes\\login\\__layout.reset.svelte"),
	() => import("..\\..\\..\\src\\routes\\login\\index.svelte")
];

const d = decodeURIComponent;

export const routes = [
	// src/routes/index.svelte
	[/^\/$/, [c[0], c[2]], [c[1]]],

	// src/routes/login/index.svelte
	[/^\/login\/?$/, [c[3], c[4]], []]
];

export const fallback = [c[0](), c[1]()];
```
## Proposed Changes
This PR changes the way the fallback components are exported and utilized in the Renderer. The fallback array now is of type `[() => CSRComponent, () => CSRComponent]` which was previously `[CSRComponent, CSRComponent]`. And the function to import these components are called only when required.

Now the generated manifest file would look something like below.
```js
const c = [
	() => import("..\\..\\..\\src\\routes\\__layout.svelte"),
	() => import("..\\components\\error.svelte"),
	() => import("..\\..\\..\\src\\routes\\index.svelte"),
	() => import("..\\..\\..\\src\\routes\\login\\__layout.reset.svelte"),
	() => import("..\\..\\..\\src\\routes\\login\\index.svelte")
];

const d = decodeURIComponent;

export const routes = [
	// src/routes/index.svelte
	[/^\/$/, [c[0], c[2]], [c[1]]],

	// src/routes/login/index.svelte
	[/^\/login\/?$/, [c[3], c[4]], []]
];

export const fallback = [c[0], c[1]];
```

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/apps/basics/src/routes/__layout.svelte
+++ b/packages/kit/test/apps/basics/src/routes/__layout.svelte
@@ -13,10 +13,13 @@
 	}
 
 	if (browser) {
-		let h3 = document.createElement('h3');
-		let text = document.createTextNode('You cannot reset me');
+		const h3 = document.createElement('h3');
+		const text = document.createTextNode('I should not be rendered');
 		h3.appendChild(text);
-		document.getElementById('nested-layout-reset-test')?.appendChild(h3);
+		const parentDiv = document.getElementById('nested-layout-reset-test');
+		if (parentDiv) {
+			parentDiv.appendChild(h3);
+		}
 	}
 </script>
 
@@ -31,7 +34,7 @@
 	export let foo;
 </script>
 
-<slot></slot>
+<slot />
 
 <footer>{foo.bar}</footer>
 
--- a/packages/kit/test/apps/basics/src/routes/__layout.svelte
+++ b/packages/kit/test/apps/basics/src/routes/__layout.svelte
@@ -1,4 +1,6 @@
 <script context="module">
+	import { browser } from '$app/env';
+
 	/** @type {import('@sveltejs/kit').Load} */
 	export async function load() {
 		return {
@@ -8,6 +10,13 @@
 				}
 			}
 		};
+	}
+
+	if (browser) {
+		let h3 = document.createElement('h3');
+		let text = document.createTextNode('You cannot reset me');
+		h3.appendChild(text);
+		document.getElementById('nested-layout-reset-test')?.appendChild(h3);
 	}
 </script>
 
--- a/packages/kit/src/core/create_app/index.js
+++ b/packages/kit/src/core/create_app/index.js
@@ -102,7 +102,7 @@ function generate_client_manifest(manifest_data, base) {
 
 		export const routes = ${routes};
 
-		export const fallback = [c[0](), c[1]()];
+		export const fallback = [c[0], c[1]];
 	`);
 }
 
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -60,7 +60,7 @@ function initial_fetch(resource, opts) {
 export class Renderer {
 	/** @param {{
 	 *   Root: CSRComponent;
-	 *   fallback: [CSRComponent, CSRComponent];
+	 *   fallback: [() => CSRComponent, () => CSRComponent];
 	 *   target: Node;
 	 *   session: any;
 	 *   host: string;
@@ -708,7 +708,7 @@ export class Renderer {
 		};
 
 		const node = await this._load_node({
-			module: await this.fallback[0],
+			module: await this.fallback[0](),
 			page,
 			context: {}
 		});
@@ -718,7 +718,7 @@ export class Renderer {
 			await this._load_node({
 				status,
 				error,
-				module: await this.fallback[1],
+				module: await this.fallback[1](),
 				page,
 				context: (node && node.loaded && node.loaded.context) || {}
 			})
--- a/packages/kit/test/apps/basics/src/routes/nested-layout/reset/_tests.js
+++ b/packages/kit/test/apps/basics/src/routes/nested-layout/reset/_tests.js
@@ -8,4 +8,8 @@ export default function (test) {
 		assert.equal(await page.textContent('h1'), 'Layout reset');
 		assert.equal(await page.textContent('h2'), 'Hello');
 	});
+
+	test('context script reset', '/nested-layout/reset', async ({ page }) => {
+		assert.ok(await page.evaluate(() => !document.querySelector('h3')));
+	});
 }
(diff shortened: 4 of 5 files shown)


packages/kit/test/apps/basics/src/routes/__layout.svelte
```diff
@@ -9,6 +11,13 @@
 			}
 		};
 	}
+
+	if (browser) {
+		let h3 = document.createElement('h3');
```
benmccann, line 16: nit: this line and the next could both be `const`

packages/kit/test/apps/basics/src/routes/__layout.svelte
```diff
@@ -9,6 +11,13 @@
...
+		let text = document.createTextNode('You cannot reset me');
+		h3.appendChild(text);
+		document.getElementById('nested-layout-reset-test')?.appendChild(h3);
```
benmccann, line 19: I would expect that `?.` fails on Node 12 on the CI. For some very strange reason, the CI did not appear to run on this PR. Github was having issues earlier today. Perhaps if you push any change (such as changing to `const` as suggested above) it will run the CI

packages/kit/test/apps/basics/src/routes/__layout.svelte
```diff
@@ -9,6 +11,13 @@
 			}
 		};
 	}
+
+	if (browser) {
+		let h3 = document.createElement('h3');
+		let text = document.createTextNode('You cannot reset me');
```
benmccann, line 17: I don't understand this message "You cannot reset me". It looks like the test is checking that the h3 is not appended because the layout is reset. In that case, would a better message be "I should not be rendered" or something? Or maybe I'm misunderstanding how the test is actually working?

DhyeyMoliya:
I had earlier sent a PR(#1356) doing the same, but was not merged because of the reasons mentioned in the comments given by Rich.

For @benmccann

packages/kit/test/apps/basics/src/routes/__layout.svelte
anudeepreddy: You understood it completely correct. I have changed the message.

packages/kit/test/apps/basics/src/routes/__layout.svelte
anudeepreddy: Thanks, I have changed it

Rich-Harris:
Yep, the behaviour is intentional. I've opened a PR that would hopefully prevent this confusion in future: https://github.com/sveltejs/kit/pull/2183

Rich-Harris:
Thanks for the PR. I'm going to close this for the reason noted above — this is intended behaviour (loading the root layout and error components in the background regardless of whether we expect to need them for the initial view). Not set in stone, but we'd need to discuss alternatives in an issue and figure out what the ideal behaviour would be, if it's not the status quo

PR #2167, opened 2021-08-11
[fix] decode path only once to fix loading unicode routes

### LINKED ISSUE #2166: Routes with umlauts are not found in node builds
### Describe the bug

When you create a route that has an umlaut in it, for example "über-uns" and you reload the page, you get a 404 Not found. This only happens when you run the build created by the adapter-node, preview and dev work as expected.

### Reproduction

https://github.com/noahsalvi/kit-umlaut-repro

install deps; npm run build; node build/

### Logs

```shell
http://localhost:3000/hell%C3%B6 404 (Not Found)
```

### System Info

```shell
System:
    OS: macOS 11.5
    CPU: (8) x64 Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
    Memory: 58.57 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.4.0 - /usr/local/bin/node
    npm: 7.18.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 92.0.4515.131
    Edge: 91.0.864.71
    Firefox: 82.0.3
    Safari: 14.1.2
  npmPackages:
    @sveltejs/adapter-node: next => 1.0.0-next.39 
    @sveltejs/kit: next => 1.0.0-next.146 
    svelte: ^3.34.0 => 3.42.1
```

### Severity

### Additional Information

The problem is that the adapter-node parses the url which encodes the umlaut characters, resulting in the route not being found.

Meaning `const parsed = new URL(......)` should be removed and replaced with just using `req.path` and `new URLSearchParams(req.query)`

I could make a pull request if needed.

noahsalvi:
Fixes #2166 

### Tests
running check returns this.. but i don't think thats due to my changes
```
Scope: all 12 workspace projects
packages/kit check$ tsc && svelte-check --ignore "src/core/make_package/test"
[3 lines collapsed]
│ Error while loading config
│ TypeError: Cannot read property 'compilerOptions' of undefined
│     at ConfigLoader.loadConfig (/Users/noahsalvi/workspace/kit/node_modules/.pnpm/svelte-check@2.2.0_svelte@3.40.0/node_m
│     at async ConfigLoader.loadAndCacheConfig (/Users/noahsalvi/workspace/kit/node_modules/.pnpm/svelte-check@2.2.0_svelte
│     at async /Users/noahsalvi/workspace/kit/node_modules/.pnpm/svelte-check@2.2.0_svelte@3.40.0/node_modules/svelte-check
│     at async Promise.all (index 2)
│     at async ConfigLoader.loadConfigs (/Users/noahsalvi/workspace/kit/node_modules/.pnpm/svelte-check@2.2.0_svelte@3.40.0
│     at async createLanguageService (/Users/noahsalvi/workspace/kit/node_modules/.pnpm/svelte-check@2.2.0_svelte@3.40.0/no
│ ====================================
│ svelte-check found 0 errors, 0 warnings, and 0 hints
└─ Done in 11.1s
packages/adapter-static check$ tsc
└─ Done in 1.7s
packages/create-svelte check$ tsc
└─ Done in 2.6s
```

### Changesets
-  [X] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

### DIFF
--- a/.changeset/perfect-zoos-fix.md
+++ b/.changeset/perfect-zoos-fix.md
@@ -1,5 +1,7 @@
 ---
 '@sveltejs/adapter-node': patch
+'@sveltejs/adapter-vercel': patch
+'@sveltejs/adapter-cloudflare-workers': patch
 ---
 
-[fix] routes with umlauts not being found
+[fix] decode path only once to fix loading unicode routes
--- a/packages/adapter-cloudflare-workers/files/entry.js
+++ b/packages/adapter-cloudflare-workers/files/entry.js
@@ -10,8 +10,9 @@ addEventListener('fetch', (event) => {
 });
 
 async function handle(event) {
+	const { request } = event;
 	// try static files first
-	if (event.request.method == 'GET') {
+	if (request.method == 'GET') {
 		try {
 			// TODO rather than attempting to get an asset,
 			// use the asset manifest to see if it exists
@@ -25,18 +26,19 @@ async function handle(event) {
 		}
 	}
 
-	// fall back to an app route
-	const request = event.request;
-	const request_url = new URL(request.url);
+	const parsedURL = new URL(request.url);
+	const path = decodeURIComponent(parsedURL.pathname);
+	const rawBody = request.body ? await read(request) : null;
+	const headers = Object.fromEntries(request.headers);
 
 	try {
 		const rendered = await render({
-			host: request_url.host,
-			path: request_url.pathname,
-			query: request_url.searchParams,
-			rawBody: request.body ? await read(request) : null,
-			headers: Object.fromEntries(request.headers),
-			method: request.method
+			host: parsedURL.host,
+			query: parsedURL.searchParams,
+			method: request.method,
+			path,
+			rawBody,
+			headers
 		});
 
 		if (rendered) {


bluwy (commented):
I think the changes in `pnpm-lock.yaml` should be removed in this PR.

noahsalvi:
done.

benmccann:
It looks like this is handled inconsistently today. `adapter-vercel`, `adapter-cloudflare-workers` would also need to be updated as they're currently parsing the path. (`adapter-netlify` and `architect/sveltekit-adapter` do not while it's not relevant for `adapter-static`)

CC @pluvial for `svelte-adapter-deno` and @jthegedus for `svelte-adapter-firebase` which also parse the URL today

We should also figure out how this interacts with `handle`. I think we are returning the parsed version there today and perhaps we should not be to be consistent with the change you are making here. Express does not parse the path so returning the unparsed path would make it more consistent and easier to interact with existing middlewares https://github.com/expressjs/expressjs.com/issues/1281.

I'm not sure whether it's better to parse in the adapter or core code. Clearly we don't want to do it twice. I'd lean towards not parsing to give the user the request as it was received and to be consistent with Express. Most important is probably that we do it consistently and test and document the expectations.

noahsalvi:
> @noahsalvi would you be able to update `adapter-vercel` and `adapter-cloudflare-workers` as you have here?

Sure 👍

noahsalvi:
btw i thought that the node adapter is using polka. Shouldn't that be the [place](https://github.com/lukeed/polka#handlerreq-res-parsed) to look? Or does it not matter due to express and polka sharing the same api?
@benmccann

benmccann:
Yeah, I just mentioned Express because it is more popular. But I just tested and confirmed that Polka works the same as Express in this manner

.changeset/perfect-zoos-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/adapter-node': patch
```
benmccann, line 2: the changeset will need a new line:
```
'@sveltejs/adapter-vercel': patch
```

.changeset/perfect-zoos-fix.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+[fix] routes with umlauts not being found
```
benmccann, line 5:
```suggestion
[fix] decode path only once to fix loading unicode routes
```

packages/adapter-cloudflare-workers/files/entry.js
```diff
@@ -25,18 +26,19 @@ async function handle(event) {
 		}
 	}
 
-	// fall back to an app route
-	const request = event.request;
-	const request_url = new URL(request.url);
+	const parsedURL = new URL(request.url);
```
noahsalvi, line 29: CFW `request` doesn't expose the usual props. That's why I did it differently for this adapter. [Doc](https://developers.cloudflare.com/workers/runtime-apis/request#properties)

packages/adapter-node/src/server.js
```diff
@@ -57,9 +56,9 @@ export function createServer({ render }) {
 			const rendered = await render({
 				method: req.method,
 				headers: req.headers, // TODO: what about repeated headers, i.e. string[]
-				path: parsed.pathname,
-				query: parsed.searchParams,
-				rawBody: body
+				path: req.path,
```
benmccann, line 59: It's not clear to me that this PR is actually changing anything

I just created an `index.js` with the source code:
`console.log(new URL('http://localhost/über-uns').pathname)`

It printed:
>/%C3%BCber-uns

If I change the code to `console.log(new URL('http://localhost/%C3%BCber-uns').pathname)` it still prints the same thing

So it's actually not clear to me that this PR is changing anything. Am I missing something? Or can you explain what was happening before vs what's happening now?

noahsalvi: Well at the moment the render function is receiving, as the `path` value, the encoded string `/%C3%BCber-uns`.

And now after changing it to take the original path from the `req` object without transforming and encoding it with  URL class, the value that it receives is `/über-uns`

I think that the thing you're missing is, that my change for the adapter node, does not wrap the path in a url object, that encodes the string.

benmccann:
Gahhh!!! This is Polka. I tested with `@latest` but it changed behavior in `@next`. I sent https://github.com/sveltejs/kit/pull/2171 as a better fix. I think this should only fix `adapter-node` after all. It was a bit of a round-about path to figuring out what was going on. Thanks for the hlpe with this one

noahsalvi:
are you sure that your fix works?
Maybe we aren't talking about the same problem, because I tried your fix and it did not work for my repro project.

Just to clarify, routes with unicode characters work on my computer only when the render function gets a none URI encoded path. Using the polka parser or the URL class to get pathname leads to it not working.
``` javascript
await render({
	method: req.method,
	headers: req.headers, // TODO: what about repeated headers, i.e. string[]
	path: '/über-uns' works but '/%C3%BCber-uns' doesn't,
	query: parsed.searchParams,
	rawBody: body
});
```

benmccann:
Thanks for testing @noahsalvi. I updated the PR and added a new test for `adapter-node`. Can you test again?

benmccann:
I just tried linking my PR against your repo and it did work

You unit test isn't correct. The browser will send an encoded path so you can't check the encoded `incoming.path` against a decoded path. But the router will decode the path when looking for a matching rout and will find the correct path and so I don't get a 404 when running against the full server

noahsalvi:
hmmm weird....

image

Just to confirm, what was the result you got from running my test?

PR #2171, opened 2021-08-11
[fix] handle paths consistently between dev and various production adapters

benmccann:
fixes https://github.com/sveltejs/kit/issues/2166

dev and `adapter-node` were passing decoded paths while the other adapters were passing the encoded path sent by the browser

The way to handle the path is to split it on forward slashes and then call `decodeURIComponent` on each path component. That allows you to handle edge cases like `%2f`. Luckily, that's what `build` is currently doing:

https://github.com/sveltejs/kit/blob/c3ca3fbb05b208af4a30c8b311074fd95d161304/packages/kit/src/core/build/index.js#L614

Otherwise, we shouldn't touch the path. We should just leave it alone and pass along exactly what we received, which is exactly how Express behaves

Previously there was another bug because we were doing `decodeURI` followed by `decodeURIComponent` which would cause each component to be double-decoded. By leaving the path alone we don't run into this issue

### DIFF
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -170,7 +170,8 @@ export class Router {
 		if (this.owns(url)) {
 			const path = url.pathname.slice(this.base.length) || '/';
 
-			const routes = this.routes.filter(([pattern]) => pattern.test(decodeURI(path)));
+			const decoded = decodeURI(path);
+			const routes = this.routes.filter(([pattern]) => pattern.test(decoded));
 
 			const query = new URLSearchParams(url.search);
 			const id = `${path}?${query}`;
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -168,9 +168,9 @@ export class Router {
 	 */
 	parse(url) {
 		if (this.owns(url)) {
-			const path = decodeURIComponent(url.pathname.slice(this.base.length) || '/');
+			const path = url.pathname.slice(this.base.length) || '/';
 
-			const routes = this.routes.filter(([pattern]) => pattern.test(path));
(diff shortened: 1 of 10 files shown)


packages/kit/src/runtime/client/router.js
```diff
@@ -168,9 +168,9 @@ export class Router {
 	 */
 	parse(url) {
 		if (this.owns(url)) {
-			const path = decodeURIComponent(url.pathname.slice(this.base.length) || '/');
+			const path = url.pathname.slice(this.base.length) || '/';
 
-			const routes = this.routes.filter(([pattern]) => pattern.test(path));
+			const routes = this.routes.filter(([pattern]) => pattern.test(decodeURI(path)));
```
Rich-Harris, line 173: shouldn't we avoid decoding multiple times?

```suggestion
			const decoded = decodeURI(path);
			const routes = this.routes.filter(([pattern]) => pattern.test(decoded));
```

Rich-Harris (commented):
I'm not totally sure I understand the scope of this change. Before, we had this:

image

With this PR, we now have this:

image

Is that intentional?

packages/kit/src/runtime/client/router.js
benmccann: done

benmccann:
Yeah, I think it's the better behavior to return the path as it was received. There are some edge cases that are impossible to handle otherwise. E.g. see this post where Netlify started out giving users the decoded path and had to switch: https://answers.netlify.com/t/bug-fix-url-encoding-preserved-in-function-event/27080. This is also inline with how Express and most other libraries work, so I think it's generally more expected as well

Rich-Harris:
Ah, gotcha. That makes sense

Issue #1564, opened 2021-05-26
SvelteKit demo app 'todos' failing on delete


JBusillo:
**Describe the bug**
The SvelteKit demo app seems to have a bug when deleting 'todos'.  It occurs when multiple todos are added and subsequently deleted.

**Logs**
```
--- Browser log
Failed to load resource: the server responded with a status of 500 (Internal Server Error)
{"status":500,"message":"Error deleting todo"}
```
Server log has no error messages

<details>
  <summary>To Reproduce</summary>

- Create a new SvelteKit demo app (npm init svelte@next app-dir), accepting all the defaults
- cd app-dir
- npm i
- npm run dev -- --open
- Open devtools

Perform the following steps (You may experience the error prior to completing all steps):

-   Navigate to TODOS, enter 3 TODOS.
-   Delete first TODO
-   Delete last TODO
- Delete remaining TODO
- Enter 3 TODOs again
- Delete first TODO
- Delete last TODO

</details>

**Expected behavior**
TODO's are deleted when requested

<details>
  <summary>Diagnostics</summary>

```
System:
    OS: Linux 5.8 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (6) x64 Intel(R) Core(TM) i5-8400 CPU @ 2.80GHz
    Memory: 25.99 GB / 31.04 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 14.17.0 - /usr/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.12.0 - /usr/local/bin/npm
  Browsers:
    Brave Browser: 90.1.24.85
    Chrome: 90.0.4430.212
    Firefox: 88.0.1
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.109 
    svelte: ^3.34.0 => 3.38.2 
```
using Chrome (also fails using Firefox)

no adapter specified
</details>

**Severity**
Not severe.  It may affect user perception of stability when evaluating SvelteKit.

**Additional context**
Failure is also present when running on Windows.

jesslawnz:
Had some difficulty reproducing this bug until I double clicked the delete icon. Able to reproduce it with a single item. 

- Create todo Item
- double click delete icon 
- error occurs.

Found another similar bug in the TODO page: 

- Focus on input '+tap to add a todo'
[...]

Issue #2100, opened 2021-08-04
Help getting to 1.0


benmccann:
### Describe the problem

We're actively working towards getting SvelteKit 1.0 out! But there's a lot to do. Just reviewing incoming PRs, prioritizing issues for 1.0, suggesting solutions for important issues, etc. is nearly a fulltime job.

We've knocked out over 500 issues that were on the 1.0 milestone, but there's still a couple dozen left

### Describe the proposed solution

If you'd like to help, please consider working on any of these issues:

https://github.com/sveltejs/kit/issues?q=is%3Aopen+is%3Aissue+milestone%3A1.0

The most straightforward items have been labeled with "help wanted"

For those, or any other tickets, we're happy to help get to a consensus on what the solution should be if it's not clear. Feel free to ping the ticket or find me on Discord under this same username. Take a look at [the contributor's guide](https://github.com/sveltejs/kit/blob/master/CONTRIBUTING.md) to get started, but feel free to ask if additional pointers would be helpful

### Alternatives considered

The maintainers do everything and 1.0 never happens :smile: 

### Importance

blocking an upgrade

kjaonline:
How can a noob like me contribute?

denny64:
Is there a backlog of things to do and their priorities? Happy to contribute wherever I can 😊

benmccann:
> How can a noob like me contribute?

Helping to answer people's questions on Discord or StackOverflow can be a good way to contribute as a beginner since you'll often learn a lot in the process

> Is there a backlog of things to do and their priorities?

https://github.com/sveltejs/kit/issues basically. A few ways to find important things to work on:
- filter by the 1.0 milestone
- filter by p1 label
- sort by thumbs up

joglomedia:
Willing to help, but I am still n00b :grin:

lamualfa:
For anyone who feel **noob**, there are many other things you can do. For example like give a little donation to Svelte team in https://opencollective.com/svelte. Whatever donation you give, it will help the team to maintain many awesome Svelte projects now and future. 

❤️❤️❤️

antony:
n00bs - we all benefit from writing tests. Good tests are always welcome.

If you think you are too n00b to contribute (you're not), then add tests, or write tests for the feature you want to add, before you add it! Start small, and learn the codebase that way.

[...]

PR #2172, opened 2021-08-12
Fix Todo delete button not disabling while waiting for API response

ric2b:
Fixes https://github.com/sveltejs/kit/issues/1564, currently the todo's in the demo app suffer from a bug where you can click delete on a todo while the first delete is still pending. 

This allows multiple delete requests to be sent, HTTP 500 errors to be received (https://github.com/sveltejs/api.svelte.dev/blob/master/src/routes/todos.ts#L46) and the todo to never visually disappear.

image

### DIFF
--- a/packages/create-svelte/templates/default/src/routes/todos/index.svelte
+++ b/packages/create-svelte/templates/default/src/routes/todos/index.svelte
@@ -59,7 +59,6 @@
 		use:enhance={{
 			result: async (res, form) => {
 				const created = await res.json();
-				created.pending_delete = false;
 				todos = [...todos, created];
 
 				form.reset();
@@ -112,7 +111,7 @@
 					}
 				}}
 			>
-				<button class="delete" aria-label="Delete todo" disabled={todo.pending_delete} />
+				<button class="delete" aria-label="Delete todo" disabled={!!todo.pending_delete} />
 			</form>
 		</div>
 	{/each}
--- a/packages/create-svelte/templates/default/src/routes/todos/index.svelte
+++ b/packages/create-svelte/templates/default/src/routes/todos/index.svelte
@@ -59,6 +59,7 @@
 		use:enhance={{
 			result: async (res, form) => {
 				const created = await res.json();
+				created.pending_delete = false;
 				todos = [...todos, created];
 
 				form.reset();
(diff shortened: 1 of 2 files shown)


benmccann (approved):
awesome. thank you! I really wanted to get this fixed before announcing 1.0 :smile:

packages/create-svelte/templates/default/src/routes/todos/index.svelte
```diff
@@ -105,12 +105,13 @@
 				action="/todos/{todo.uid}.json?_method=delete"
 				method="post"
 				use:enhance={{
+					pending: () => todo.pending_delete = true,
```
ignatiusmb, line 108:
```suggestion
					pending: () => (todo.pending_delete = true),
```

packages/create-svelte/templates/default/src/routes/todos/index.svelte
```diff
@@ -105,12 +105,13 @@
...
 					result: () => {
 						todos = todos.filter((t) => t.uid !== todo.uid);
 					}
 				}}
 			>
-				<button class="delete" aria-label="Delete todo" />
+				<button class="delete" aria-label="Delete todo" disabled={!!todo.pending_delete} />
```
ignatiusmb, line 114:
```suggestion
				<button class="delete" aria-label="Delete todo" disabled={todo.pending_delete} />
```

You'll need to update the types above as well for TS users

ignatiusmb (approved):
Thanks for the PR!

ignatiusmb:
Fixes #1564

ric2b:
@benmccann  Thank you for taking the time to open https://github.com/sveltejs/kit/issues/2100, that's where I found it :)

@ignatiusmb Thanks for the suggestions! From the description I thought it wouldn't need to be noted in a changelog since it's just the demo app, so I didn't create the changeset. And I've never used TS so I wasn't sure if an attribute that isn't always present should be in the type definition.

benmccann:
No worries. I don't normally add a changelog for stuff in the `/examples` directory but this demo app gets created for users when you run `npm init svelte@next` and I think I can't release a new version of that package with your fix without a changelog

PR #2176, opened 2021-08-12
[docs] improve Introduction

vwkd:
This PR aims to make the Introduction doc more approachable especially for new users. Current ideas but open to suggestions:

- Describe what SvelteKit is before everything else
- List of most important features
- Move specifics into later docs, e.g. pages
- Move beta disclaimer to install section

Also not sure if my understanding of how SvelteKit works is correct. I reused bits and pieces I found throughout the docs, from the [landing page](https://kit.svelte.dev), and from the [introductory blog post](https://svelte.dev/blog/whats-the-deal-with-sveltekit). Please correct if it isn't.

If desired, I would like to visit further parts of the docs in future PRs with the aim to make it more approachable to new users.

### Tests

### Changesets

### DIFF
--- a/documentation/docs/00-introduction.md
+++ b/documentation/docs/00-introduction.md
@@ -2,90 +2,46 @@
 title: Introduction
 ---
 
-SvelteKit is an application framework for building web apps using Svelte components. It's the most modern way to build web apps, with full support for serverless platforms, and a great development experience.
+SvelteKit is an application framework for building web apps using Svelte components.
 
+It's the future of building modern web apps, fully embracing serverless platforms, while providing a great development experience.
 
-
-SvelteKit is the future of Svelte apps
-SvelteKit is the most modern way to build web apps.
-
-
-SvelteKit fully embraces the serverless paradigm
-support multiple platforms
-adapters for different serverless platforms
-
+It's a *kit* of multiple things: a SSR engine, a client-side runtime, and a development toolchain.
 
 ### Features
 
-SvelteKit is a kit of multiple things: a powerful SSR engine, a performant client-side runtime, and a delightful development server.
+_If you need a quick explainer on those concepts, check out the appendix._
 
-- [server-side rendering](#appendix-ssr) and [client-side routing](#appendix-routing)
+- [server-side rendering](#appendix-ssr)
 
-By default, SvelteKit has the best settings enabled.
+For the initial page the user is served a server-side rendered version of the page
 
-SSR for the initial page
-client-side routing and CSR for subsequent pages
+- [prerendering](#appendix-prerendering) or dynamic SSR
 
-By default, the initial page is rendered on the server (SSR).
+The server-side rendered version of the page is prerendered at build-time or rendered dynamically at request-time depending on the [adapter](#adapters) used. It can be configured on a per-page level.
 
-By default, pages are rendered on both the client and server
+- [client-side routing](#appendix-routing)
 
-By default, on initial page load, users will be served a server-side rendered version of the page
+After the initial page load of the server-side rendered page, a client-side runtime takes over. Further navigation to subsequent pages is handled by the client-side router serving a client-side rendered version of the pages.
 
-Pages will be server-side rendered so that a user's initial page load is as fast as possible.
+- [highly configurable](#rendering)
 
-Ordinarily, SvelteKit [hydrates](#appendix-hydration) your server-rendered HTML into an interactive page.
-Some pages don't require JavaScript at all — many blog posts and 'about' pages fall into this category.
+Rendering can be configured on a per-page or per-app basis.
 
+- code-splitting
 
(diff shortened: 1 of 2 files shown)


documentation/docs/00-introduction.md
```diff
@@ -2,25 +2,54 @@
 title: Introduction
 ---
 
-### Before we begin
+SvelteKit is an application framework for building web apps using Svelte components.
 
-> SvelteKit is in early development, and some things may change before we hit version 1.0. This document is a work-in-progress. If you get stuck, reach out for help in the [Discord chatroom](https://svelte.dev/chat).
->
-> See the [migration guides](/migrating) for help upgrading from Sapper.
+Build modern state-of-the-art web apps that fully embrace the serverless paradigm with a great development experience.
```
vwkd, line 7: Maybe this doesn't fit here or should be rephrased.

documentation/docs/00-introduction.md
```diff
@@ -2,25 +2,54 @@
...
+
+Under the hood SvelteKit is made up of multiple things, hence the name "kit". It's a rendering engine for SSR and CSR, a client-side runtime for client-side routing and CSR, and a toolchain for development.
+
+(If you need a quick reference on these words, check out the [glossary](#appendix) in the appendix.)
+
+### Features
+
+- [server-side rendering](#appendix-ssr)
+
+For the initial page, the user is served a server-side rendered version of the page.
```
benmccann, line 17: I'm not 100% sure without testing it, but I think that these bullet points will not render correctly. This looks to me like it will create a series of one bullet point lists interleaved with new paragraphs. I think the description needs to be on the same line as the bullet point

pnpm-lock.yaml
```diff
@@ -189,7 +189,7 @@ importers:
       '@sveltejs/adapter-cloudflare-workers': link:../../../adapter-cloudflare-workers
       '@sveltejs/adapter-netlify': link:../../../adapter-netlify
       '@sveltejs/adapter-vercel': link:../../../adapter-vercel
-      '@sveltejs/kit': 1.0.0-next.144_svelte@3.40.0
+      '@sveltejs/kit': link:../../../kit
```
benmccann, line 192: a documentation PR shouldn't update the lockfile

vwkd: Fixed

documentation/docs/00-introduction.md
benmccann: I'm not sure, because the docs are now loaded from an API. @pngwn do you have ideas on testing docs rendering?

pngwn: I started work on this and dropped the ball a bit, so we could test locally. For now we'll just have to make a best judgement and go from there. I'll finish off the test setup I had in mind this weekend but it will probably need publishing for it to be useful.

benmccann:
Thanks for this!

This changes quite a few things. I'm not sure they're all necessary. E.g. the note about Discord and Sapper migration was put up front to make sure users see it. I'm not sure whether it makes sense to move, but it just seems pretty low priority to change even if it were especially since most of that note will change or go away when we hit 1.0, so I don't want to spend too much time thinking about it now when we'll be changing it again.

I think it's nice that a few more of the features are mentioned. The maintainers I asked about this were lukewarm about the bullet point format though. I also think we get a bit too far into defining terms in the intro

There was a sentence about Svelte itself that was removed that I think could have been helpful to have for people who haven't used Svelte before

Moving the thing about pages to later makes sense, but it looks like it might have been removed rather than moved

I took another stab at this based off your ideas here: https://github.com/sveltejs/kit/pull/2184

vwkd:
Thanks for the feedback!

My idea was to help introduce new users to SvelteKit by explaining what SvelteKit as early as possible.

I see how the technical description of what SvelteKit is using SSR and CSR can look intimidating. Would writing the abbreviations out instead of shortening them help? The reason I think it belongs in the Introduction is that a user unfamiliar with SvelteKit can learn what SvelteKit is before diving into the docs about the specifics. Currently, some of that high-level explanation lives burried in the appendix. I think it adds value answering "What is SvelteKit?" first.

I'm open to suggestion about the bullet point format. It was the easiest thing that came to mind, certainly not the best.

Putting the beta disclaimer back makes sense to me.

Great catch. The page stuff isn't quite lost, it lives on my local machine, I was planning to put it into the next PR about Routes if this one goes through.

vwkd:
Thanks for the review!

I see how the technical terms seem too much. I toned them down. Maybe you can take another look?

Putting the beta disclaimer back up makes sense to me.

I'm open to suggestion about the bullet point format. It was the easiest thing that came to mind, certainly not the best.

Great catch! The page stuff isn't lost, it lives in my local repo. Forgot to mention that! I was planning to look into the Routes docs next which would include it, if this PR goes through.

benmccann:
thanks for the suggestions! I merged this in https://github.com/sveltejs/kit/pull/2184 without the bullet point format

PR #2189, opened 2021-08-13
dev/preview paths

Rich-Harris:
following up https://github.com/sveltejs/kit/pull/2145#issuecomment-898452235

### Tests

### Changesets


Rich-Harris:
Also, need to update the docs

Rich-Harris:
One thing I missed somehow is that we're still not actually serving the app on the basepath during dev and preview. Working on that

packages/kit/src/core/build/index.js
```diff
@@ -39,10 +39,11 @@ export async function build(config, { cwd = process.cwd(), runtime = '@sveltejs/
...
 		build_dir,
-		base:
-			config.kit.paths.assets === '/.'
-				? `/${config.kit.appDir}/`
-				: `${config.kit.paths.assets}/${config.kit.appDir}/`,
+		// TODO this is so that Vite's preloading works. Unfortunately, it fails
+		// during `svelte-kit preview`, because we use a local asset path. If Vite
+		// used relative paths, I _think_ this could get fixed. Issue here:
+		// https://github.com/vitejs/vite/issues/2009
+		base: `${config.kit.paths.assets || config.kit.paths.base}/${config.kit.appDir}/`,
```
benmccann, line 46: I might not understand what `base` is, but I'm surprised that `assets` affects it

packages/kit/src/core/build/index.js
```diff
@@ -323,13 +322,13 @@ async function build_server(
 					amp: ${config.kit.amp},
 					dev: false,
 					entry: {
-						file: ${s(prefix + client_manifest[client_entry_file].file)},
-						css: ${s(Array.from(entry_css).map(dep => prefix + dep))},
-						js: ${s(Array.from(entry_js).map(dep => prefix + dep))}
+						file: assets + ${s(prefix + client_manifest[client_entry_file].file)},
```
benmccann, line 325: `assets` is not in quotes on this line, but is on the next line. is that on purpose?

Rich-Harris:
Ok, this is basically done apart from docs, I think. I don't think https://github.com/vitejs/vite/issues/2009 should block us, since it's not a regression — it just means that if you're using `assets` you will experience errors in `svelte-kit preview`.

packages/kit/src/core/build/index.js
Rich-Harris: if `assets` is supplied, it's an absolute URL (like `https://cdn.example.com/stuff`). If not, assets are served from the same place as the rest of the app, which is either a specified base path like `/app-goes-here` or the empty string

packages/kit/src/core/build/index.js
Rich-Harris: yep! we're just creating expressions that will resolve to the runtime (not prebaked) value of `assets`

Rich-Harris:
will do docs after lunch

packages/kit/src/core/build/index.js
benmccann: I thought by `base` that this would be the base URL for the application. Should we call it `assets_base` or something along those lines?

Rich-Harris: makes sense, updated

.changeset/clean-chicken-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Use /\_svelte_kit_assets when serving apps with specified paths.assets locally
```
benmccann, line 5: does the underscore need to be escaped here?

.changeset/honest-jars-report.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
```
benmccann, line 2: it's interesting that there are three changesets for this PR. I've never seen a PR with multiple, so don't quite know what will happen when we squash this

packages/kit/src/core/dev/index.js
```diff
@@ -325,114 +333,119 @@ async function create_handler(vite, config, dir, cwd, manifest) {
...
-							if (error.stack) {
-								console.error(colors.gray(error.stack));
-							}
-						},
-						hooks: {
-							getSession: hooks.getSession || (() => ({})),
-							handle: hooks.handle || (({ request, resolve }) => resolve(request)),
-							serverFetch: hooks.serverFetch || fetch
+				const rendered =
+					parsed.pathname.startsWith(config.kit.paths.base) &&
```
benmccann, line 337: I wonder if you could create a variable for `parsed.pathname.startsWith(config.kit.paths.base)` to avoid prettier indenting all the subsequent lines

packages/kit/test/apps/options/svelte.config.js
```diff
@@ -17,7 +17,16 @@ const config = {
 			build: {
 				minify: false
 			},
-			clearScreen: false
+			clearScreen: false,
+			server: {
+				hmr: {
```
benmccann, line 22: I don't think I've seen these options set in any of the other tests. worth a comment explaining?

benmccann:
Nice! This looks like it cleans some stuff up a bit

.changeset/clean-chicken-walk.md
Rich-Harris: that's just something `pnpx changeset` does. it's harmless

.changeset/honest-jars-report.md
Rich-Harris: three changelog entries

packages/kit/test/apps/options/svelte.config.js
Rich-Harris: left over from #2145, i just assumed it was there for a reason. we can probably remove

packages/kit/src/core/dev/index.js
Rich-Harris: it won't make any difference, it could be `x &&` and prettier would still put them on separate lines

benmccann: Ah. You could change it like this to avoid the extra indentation:
```
				if (!parsed.pathname.startsWith(config.kit.paths.base)) {
					res.statusCode = 404;
					res.end('Not found');
					return;
				}
				const rendered = await respond(
```

Rich-Harris: we could, i just don't love the duplication of 

```
res.statusCode = 404;
res.end('Not found');
```

maybe it's the lesser of two evils. not sure

benmccann: could refactor that out:
```
function not_found(res) {
  res.statusCode = 404;
  res.end('Not found');
}
if (!parsed.pathname.startsWith(config.kit.paths.base)) {
  return not_found(res);
}
```

PR #2191, opened 2021-08-13
[fix] upgrade polka and sirv. fixes handling of URLs with unicode characters

benmccann:
Fixes https://github.com/sveltejs/kit/issues/2166


Rich-Harris:
is @lukeed aware of this work? am wondering if it would be better to wait for an upstream fix

lukeed:
This won't be needed. And it's @benmccann preparing as a result of https://github.com/lukeed/polka/pull/172 which hasn't been merged or published yet

benmccann:
@lukeed why do you say it won't be needed? Is it because sirv will be updated? Kit is currently failing on unicode routes with `adapter-node`, but we can hold this and bump polka and sirv instead if you think you'll get a chance to cut a release with the fixes there

lukeed:
Yes of course it will be updated. Your polka PR is a breaking change.

lukeed:
Workaround is not needed w/ recent `sirv` check. The `req._decoded` check should have always been in there, since it was sirv's way of preventing duplicate `decodeURIComponent` calls... but that was only true when it receives a request from a `polka@next` app, since Polka _was_ writing the decoded value to `req.path`

Now that the latest `polka@next` (and express) doesn't decode automatically anymore, `req.path` isn't trustworthy on its own. It needs `req._decoded` to be there too in order to trust it.

This combo-check is backwards compatible for `polka@next` users who don't upgrade.

benmccann:
Thanks @lukeed!! I appreciate the help!

lukeed:
No prob, thanks for the jumpstart & reminder 👍

PR #2193, opened 2021-08-13
Handle error

Rich-Harris:
Closes https://github.com/sveltejs/kit/issues/1857

#2122 with a few tweaks — updated docs, and reinstated `handle_error`. Despite the naming similarity, they're separate things — `handleError` is the user's code, `handle_error` is what the framework does (which includes calling the user's code). It's better to avoid conflating them, even if the framework isn't doing much beyond what the user does (i.e. just manipulating `error.stack`)

### Tests

### Changesets

### DIFF
--- a/.changeset/chilled-laws-battle.md
+++ b/.changeset/chilled-laws-battle.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Add hook to handle errors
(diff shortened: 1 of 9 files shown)


benmccann:
This PR looks to be accidentally reverting https://github.com/sveltejs/kit/pull/2178

Rich-Harris:
yeah, there was a conflict. i was going to tweak it anyway, 'This function runs on every rendering' reads to me like a typo — sorry!

benmccann:
looks good to me other than that

just FYI, there was also a highlight that was removed besides the rendering / request wording

> `handle` runs when a page is rendered. If a page is prerendered, `handle` will not run again when the user requests the page. If you would like to run some functionality on each request, see the documentation for your [adapter](#adapters).

PR #2197, opened 2021-08-14
[chore] clean up fixable `ts-expect-error` comments

ignatiusmb:
This will reduce `ts-expect-error` comments to only those that we actually _expect_ to throw error. Also changed some to `ts-ignore` as we don't really care or need to for it. Added proper comments for the ones we can't remove.


packages/kit/src/runtime/client/renderer.js
```diff
@@ -100,9 +104,9 @@ export class Renderer {
 		};
 
 		this.stores = {
-			page: page_store({}),
-			navigating: writable(null),
-			session: writable(session)
+			page: /** @type {import('$app/stores').page} */ page_store({}),
```
benmccann, line 107: could we set the `return` type on `page_store` instead?

packages/kit/src/core/adapt/test/index.js
```diff
@@ -72,28 +81,33 @@ suite('prerender', async () => {
...
+		preprocess: null
 	};
 
 	/** @type {import('types/internal').BuildData} */
 	const build_data = { client: [], server: [], static: [], entries: ['/nested'] };
 
-	const utils = get_utils({ cwd, config, build_data, log });
+	const utils = get_utils({
+		cwd,
+		config: /** @type {import('types/config').ValidatedConfig} */ (mocked),
```
benmccann, line 107: why don't we just set the type on `mocked` to `ValidatedConfig` originally?

benmccann:
nice! looks like this will need to be rebased

PR #2198, opened 2021-08-14
[fix] discover new routes in dev server

### LINKED ISSUE #2046: Dev server don't autodiscover new routes
### Describe the bug

If the dev server is running, SvelteKit don't see newly added routes, and result in 404 when navigating to them.

### Reproduction

- Create a new project (`npm init svelte@next my-app` + select the `Skeleton project`, should also be true for demo app)
- Install the project (`npm install`)
- Start the dev server (`npm run dev`)
- Add a new page in `src/routes` (ex: `src/routes/test.svelte`)
- Try to navigate to the route (with the previous example: `http://localhost:3000/test`)

### Logs

```shell
// Server console side
09:50:09 [vite] page reload .svelte-kit/dev/generated/manifest.js

// Browser side
404
Not found: /test
Error: Not found: /test
    at render_page (file:///private/tmp/sk/my-app/node_modules/@sveltejs/kit/dist/ssr.js:1318:11)
    at async resolve (file:///private/tmp/sk/my-app/node_modules/@sveltejs/kit/dist/ssr.js:1663:12)
    at async respond (file:///private/tmp/sk/my-app/node_modules/@sveltejs/kit/dist/ssr.js:1614:10)
    at async Immediate.<anonymous> (file:///private/tmp/sk/my-app/node_modules/@sveltejs/kit/dist/chunks/index.js:3458:22)
```

### System Info

```shell
System:
    OS: macOS 10.15.7
    CPU: (8) x64 Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
    Memory: 78.90 MB / 8.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 14.16.1 - /opt/local/bin/node
    Yarn: 1.22.10 - /opt/local/bin/yarn
    npm: 6.14.8 - /opt/local/bin/npm
  Browsers:
    Safari: 14.1.2
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.139
    svelte: ^3.34.0 => 3.41.0
```

### Severity

### Additional Information

The file `.svelte-kit/dev/generated/manifest.js` is correctly generated.

But SvelteKit don't seem to reload it (it is only load at here: `.svelte-kit//dev/runtime/internal/start.js`)

A simple workaround is to stop and restart server every times a new route is created, in DX point of view it's not ideal

----

_Initial Discord message: https://discord.com/channels/457912077277855764/819723698415599626/870925145554952222_

jasonlyu123:

### Tests

### Changesets

Fixes #2046. The manifest data passed into the server handler it's the initial manifest.  And the manifest won't be mutated by reference but resigned. This PR changes to pass a get function to the handler so it'll get the up-to-date manifest.

### DIFF
--- a/.changeset/tall-jeans-hear.md
+++ b/.changeset/tall-jeans-hear.md
@@ -1,5 +1,5 @@
 ---
-'@sveltejs/kit': minor
+'@sveltejs/kit': patch
 ---
 
 fix new route discovery in dev server
--- a/.changeset/tall-jeans-hear.md
+++ b/.changeset/tall-jeans-hear.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': minor
+---
+
+fix new route discovery in dev server
--- a/packages/kit/src/core/dev/index.js
+++ b/packages/kit/src/core/dev/index.js
@@ -177,7 +177,7 @@ class Watcher extends EventEmitter {
 
 		this.vite = await vite.createServer(merged_config);
 
-		const getManifest = () => {
+		const get_manifest = () => {
 			if (!this.manifest) {
 				throw new Error("Shouldn't clear manifest after server initialized");
 			}
@@ -185,7 +185,7 @@ class Watcher extends EventEmitter {
 			return this.manifest;
 		};
 
-		handler = await create_handler(this.vite, this.config, this.dir, this.cwd, getManifest);
+		handler = await create_handler(this.vite, this.config, this.dir, this.cwd, get_manifest);
 
 		this.server.listen(this.port, this.host || '0.0.0.0');
 	}
@@ -275,9 +275,9 @@ function get_params(array) {
  * @param {import('types/config').ValidatedConfig} config
  * @param {string} dir
  * @param {string} cwd
- * @param {() => import('types/internal').SSRManifest} getManifest
+ * @param {() => import('types/internal').SSRManifest} get_manifest
  */
-async function create_handler(vite, config, dir, cwd, getManifest) {
+async function create_handler(vite, config, dir, cwd, get_manifest) {
 	/**
 	 * @type {amp_validator.Validator?}
 	 */
@@ -437,7 +437,7 @@ async function create_handler(vite, config, dir, cwd, getManifest) {
 								styles: Array.from(styles)
 							};
 						},
-						manifest: getManifest(),
+						manifest: get_manifest(),
 						prerender: config.kit.prerender.enabled,
 						read: (file) => fs.readFileSync(path.join(config.kit.files.assets, file)),
 						root,
--- a/packages/kit/src/core/dev/index.js
+++ b/packages/kit/src/core/dev/index.js
@@ -177,7 +177,15 @@ class Watcher extends EventEmitter {
 
 		this.vite = await vite.createServer(merged_config);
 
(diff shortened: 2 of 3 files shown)


packages/kit/test/apps/basics/src/routes/routing/_tests.js
```diff
@@ -238,4 +241,27 @@ export default function (test) {
...
+		// hash the filename so that it won't conflict with
+		// future test file that it has the same name
+		const route = 'bar' + new Date().valueOf();
+		const content = 'Hello new route';
+		const __dirname = path.dirname(fileURLToPath(import.meta.url));
+		const filePath = path.join(__dirname, `${route}.svelte`);
+
+		try {
+			fs.writeFileSync(filePath, `<h1>${content}</h1>`);
+			await Promise.resolve();
```
jasonlyu123, line 259: Don't know if there is any way to wait for the manifest update. Maybe I should just wait for a few seconds. It might also let tests pass in js. I currently disable the test in js because it has to wait for the dev server to send updates to the client.

.changeset/tall-jeans-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': minor
```
ignatiusmb, line 2:
```suggestion
'@sveltejs/kit': patch
```

packages/kit/src/core/dev/index.js
```diff
@@ -177,7 +177,15 @@ class Watcher extends EventEmitter {
 
 		this.vite = await vite.createServer(merged_config);
 
-		handler = await create_handler(this.vite, this.config, this.dir, this.cwd, this.manifest);
+		const getManifest = () => {
```
ignatiusmb, line 180: This should be snake_cased

packages/kit/src/core/dev/index.js
```diff
@@ -177,7 +177,15 @@ class Watcher extends EventEmitter {
...
+			if (!this.manifest) {
+				throw new Error("Shouldn't clear manifest after server initialized");
+			}
```
ignatiusmb, line 183: Will there be a situation where this will eventually throw?

ignatiusmb (commented):
Looks like the tests are finally passing, this would be a really nice improvement for the dev server!

packages/kit/src/core/dev/index.js
jasonlyu123: Doesn't see a current path that it will throw. It should only be undefined before the init method call. It's more likely to throw in line 92. It's just because init is async so we can't enforce it to be not nullable unless we have an empty object in the constructor.

packages/kit/src/core/dev/index.js
```diff
@@ -177,7 +177,15 @@ class Watcher extends EventEmitter {
 
 		this.vite = await vite.createServer(merged_config);
 
-		handler = await create_handler(this.vite, this.config, this.dir, this.cwd, this.manifest);
+		const get_manifest = () => {
+			if (!this.manifest) {
+				throw new Error("Shouldn't clear manifest after server initialized");
```
benmccann, line 182: We use single quotes for all strings
```suggestion
				throw new Error('Manifest is not available');
```

benmccann (approved):
awesome!

Issue #1485, opened 2021-05-18
Shim SvelteKit runtime import aliases / Importing `$app/*` fails


madeleineostoja:
**Describe the bug**
As far as I can tell there's no way to use sveltekit runtime imports (eg: `$app/navigation`) outside of sveltekit dev/build. Which makes testing virtually impossible. If there is a way to shim these imports outside of the main sveltekit context I haven't found it, so perhaps documentation is needed.

My particular use-case is with [Storybook](https://storybook.js.org/), where UI components that rely on any sveltekit modules break the whole setup. I tried aliasing them with webpack (pointing to `.svelte-kit/dev/...`) but that didn't work either.

Another use-case is publishing components for sveltekit that would need to rely on those imports.

**To Reproduce**
1. Setup storybook with Sveltekit
2. Create a component that imports a runtime module (eg: `$app/env`)
3. Run storybook and see if fail (`cannot resolve module $app/env`)

**Severity**
Not blocking, but makes building a component library with Storybook or other development/testing frameworks impossible. So, severe annoyance?

benmccann:
This sounds loosely related to allowing SvelteKit to build components as suggested in https://github.com/sveltejs/kit/issues/518

madeleineostoja:
Yep that would solve the second use case (distributing components/actions that rely on sveltekit), but I don’t think it would address the first? Testing etc components of a sveltekit app outside the main sveltekit dev/build runtime? Storybook being a very common use case

alexkornitzer:
TBH I am hitting this issue trying to just do basic testing with `uvu` and typescript. I have ts files that import `$app/env` and `uvu` fails to resolve this using `ts-node` but I know you guys have set testing as a post 1.0. But any hacks or workarounds would be greatly appreciated. https://github.com/sveltejs/kit/issues/19

josephspurrier:
Are we able to use something like https://github.com/eirslett/storybook-builder-vite to help with this - or is this a problem outside of Vite?

josephspurrier:
@madeleineostoja for Storybook, in the .storybook/main.js/cjs, can you try using `preprocess.replace` like this for each of your aliases:

```javascript
const path = require('path');
const sveltePreprocess = require('svelte-preprocess');

module.exports = {
	stories: ['../src/**/*.stories.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx|svelte)'],
[...]

PR #2210, opened 2021-08-15
[chore] resolve outdated todos

ignatiusmb:
Resolve some outdated TODOs that we can safely remove as we're approaching 1.0

### DIFF
--- a/packages/kit/src/runtime/app/stores.js
+++ b/packages/kit/src/runtime/app/stores.js
@@ -1,7 +1,6 @@
 import { getContext } from 'svelte';
 
-// const ssr = (import.meta as any).env.SSR;
-const ssr = typeof window === 'undefined'; // TODO why doesn't previous line work in build?
+const ssr = import.meta.env.SSR;
 
 // TODO remove this (for 1.0? after 1.0?)
 let warned = false;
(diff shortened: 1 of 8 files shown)


benmccann (approved):
nice cleanup!

packages/kit/src/runtime/app/stores.js
```diff
@@ -1,7 +1,6 @@
 import { getContext } from 'svelte';
 
-// const ssr = (import.meta as any).env.SSR;
-const ssr = typeof window === 'undefined'; // TODO why doesn't previous line work in build?
+const ssr = import.meta.env.SSR;
```
wallw-bits, line 3: This broke some test mocks on our end. What's the best way to get `import.meta.env` to be defined in a non-vite context such as Jest or another test runner?

benmccann: there's an issue about this. it's been one of the things to figure out before 1.0: https://github.com/sveltejs/kit/issues/1485

patrickleet: @wallw-bits @benmccann 

I noticed this breaking tests for me as well - I think something like 

```js
import { browser } from './env.js'

const ssr = !browser
```

would do the trick, as then you can just mock `$app/env.js` with

```
jest.mock('$app/env.js', () => ({
  amp: false,
  browser: true, // or false for testing ssr
  dev: true,
  mode: 'test'
}))
```

patrickleet: just tried it - making that change locally did work

image

I think it also makes the code more DRY - import.meta.env.SSR is already being dealt with in env.

I'll make a PR.

patrickleet: PR made: https://github.com/sveltejs/kit/pull/2353

Issue #1563, opened 2021-05-26
Ability to stream data (for non-serverless adapters like adapter-node)


jesseskinner:
**Is your feature request related to a problem? Please describe.**
I have APIs built with Sapper that I'd like to migrate to SvelteKit, but some endpoints rely heavily on using Node streams to pipe data to the response, as there can be gigabytes of data that are too large to serialize to a string server-side in memory. Some of this data is coming from a database, other data is coming from streaming and processing large files that live on AWS S3.

**Describe the solution you'd like**
I'd like to be able to return an async iterator in the body attribute of the response. I realise streaming isn't possible with most (all?) serverless adapters or `adapter-static`, but would certainly be possible with `adapter-node` and possibly other adapters that don't yet exist.

**Describe alternatives you've considered**
`adapter-node` and the development server could expose the Node `res` object to the request so I could pipe directly to the res object inside the request handler.
Otherwise, I will have to split off the API into a standalone Node server that does not live within SvelteKit, or continue to use Sapper and have a much slower build process during development.

**How important is this feature to you?**
I would love to keep my API in SvelteKit as I currently do in Sapper. It's really nice having the API endpoints live in the same router. To split it off into a separate server will incur extra complexity, such as dealing with sharing authentication, keeping the servers in sync, multiple code bases, etc. I would really rather not stay with Sapper long term.

**Additional context**

EDITED: changed the preferred solution from stream object to async iterator, because [Node Streams in v10+ are also async iterators](https://nodejs.org/dist/latest-v10.x/docs/api/stream.html#stream_readable_symbol_asynciterator).

babichjacob:
Heavily related to https://github.com/sveltejs/kit/issues/1491 and https://github.com/sveltejs/kit/issues/887.

jesseskinner:
If we wanted to ensure that returning a stream in the body remains compatible with all adapters, so that this isn't a special `adapter-node`-only feature, it is possible to collect the contents of a stream in a string for compatibility with adapters that will never support streaming. See [stream-to-string](https://github.com/jasonpincin/stream-to-string) for one implementation of this. (Obviously this would have memory limitations but would work with serverless adapters otherwise.)

babichjacob:
I've thought about this problem before, and my first idea for the API design for this in endpoints is something along these lines:

```js
export async function* get() {
    yield {
      status: 200,
      headers: {
        "Connection": "keep-alive",
        "Content-Type": "multipart/mixed",
[...]

PR #2212, opened 2021-08-15
[feat] support for streaming via Async Iterator endpoints

kjmph:
This commit demonstrates the ability to easily add Async Iterator
SvelteKit Endpoints. By having an endpoint return an async iterator as
the body, SvelteKit will keep emitting writes for as long as the
endpoint yields values. Or, for as long as the client holds the
request open; which ever ends first.

Most of the discussion is occurring over on #1563; opening pull request to be illustrative. I believe there are a few type errors that will need to be cleaned up before merging.


packages/adapter-node/package.json
```diff
@@ -29,6 +29,7 @@
 		"@sveltejs/kit": "workspace:*",
 		"c8": "^7.7.2",
 		"compression": "^1.7.4",
+		"compressible": "^2.0.18",
```
benmccann, line 32: you will need to run `pnpm install` and push the changes to get the CI to run

jesseskinner:
Today I've added some test coverage for async iterator support, including for the dev server (basics), adapter-static and adapter-node, and uncovered & fixed some bugs as well as added support for adapter-static in the process. I also rebased. I think the PR is now in good shape - the tests have definitely added confidence that things should be working well. Looking forward to any feedback whether any additional changes are needed.

packages/adapter-cloudflare-workers/files/entry.js
```diff
@@ -39,10 +39,31 @@ async function handle(event) {
...
+				const body_stream = new ReadableStream({
+					async pull(controller) {
+						const { value, done } = await rendered.body.next();
+						if (done) {
+							controller.close();
+						} else {
+							controller.enqueue(value);
+						}
+					}
+				});
```
mrkishi, line 56: Wouldn't we need something like `cancel() { return rendered.body.return(); },` here?

packages/adapter-node/src/server.js
```diff
@@ -64,8 +66,21 @@ export function createServer({ render }) {
...
+					typeof rendered.body === 'object' &&
+					typeof rendered.body[Symbol.asyncIterator] === 'function'
+				) {
+					const flush = rendered.headers && compressible(rendered.headers['content-type']) ? res.flush : null;
+					const data = Readable.from(rendered.body);
+					data.on('error', () => res.end());
+					if (flush) {
+						data.on('data', () => flush());
+					}
+					data.pipe(res);
```
mrkishi, line 80: I don't think `pipe` closes the readable stream, does it? We might need `res.on('close', () => data.destroy());`.

mrkishi (commented):
Your PR looks good but could you check my notes? Try running the following stream server and serving `{ body: await fetch('/stream').body }` from SvelteKit. It looks to keep streaming after disconnection, unless I'm messing something up?

```js
function handle_request(request, response) {
	if (request.url === '/stream') {
		console.log('request opened');

		let interval = 0;
		const stream = new Readable({
			construct(callback) {
				console.log('stream opened');

				const send = () => {
					const chunk = Math.random().toString();
					console.log('stream push', chunk);
					this.push(chunk);
				}

				interval = setInterval(send, 1000);
				callback();
			},

			read() {},

			destroy(error, callback) {
				console.log(id, 'stream closed');
				clearInterval(interval);
				callback(error);
			},
		});

		response.on('close', () => {
			console.log(id, 'request closed');
			stream.destroy();
		});

		response.writeHead(200);
		stream.pipe(response);
	}
}

const server = http.createServer(handle_request);
server.listen(8080);
```

packages/adapter-cloudflare-workers/files/entry.js
kjmph: Yes, this section needs more hardening. Will make a pass at all edge cases.

packages/kit/src/runtime/server/endpoint.js
```diff
@@ -27,7 +32,8 @@ function is_content_type_textual(content_type) {
 		type === 'text/plain' ||
 		type === 'application/json' ||
 		type === 'application/x-www-form-urlencoded' ||
-		type === 'multipart/form-data'
+		type === 'multipart/form-data' ||
+		type === 'text/event-stream'
```
jesseskinner, line 36: @kjmph I wasn't sure if this is necessary here - I was just reproducing what you previously had in adapter-utils.js - maybe it's not anymore since we check for async iterators in render_endpoint anyway? Could you have a look?

AndreasHald:
@jesseskinner anything I can do to help get this PR over the finish line?

jesseskinner:
@AndreasHald I think it's done, just waiting for a review. @mrkishi had some comments for @kjmph but I'm not sure how critical those bits are for getting this merged. 🤷‍♂️

kjmph:
I'll make the relevant changes, my apologies on my delay. I think @mrkishi is correct, we should handle the cancel of the ReadableStream for Cloudflare support as well as the closure of the pipe for Node.js support. I'll make the change when free (today or so).

packages/kit/src/runtime/server/endpoint.js
kjmph: Hello @jesseskinner, I don't believe this is necessary, for three reasons. A) you added the check that is the content type specified by the endpoint is not recognized, it will allow async iterators to progress in the response processing. B) the `is_content_type_textual` function is shared with the request processing on the POSTed body, this is undefined to have an `text/event_stream` being sent to the server. C) the `compressible` dependency will handle the actual question if there is a flush needed to get the compressor to emit the response body back to the client.

I am rebasing right now, and will remove this line.

kjmph:
Forgive my whack a mole, I see the bar for pull requests has been raised on this project. Good on ya. Anyway, I cleaned up everything I can find, I clearly don't understand how to run the tests on my own machine, yet I was able to fix everything I can find short of the timeout. I'm uncertain if the timeout is due to these changes. Note, there was an Either/Or type added to calm down TypeScript checking; perhaps that causes tests to take longer?

kjmph:
Oh would you like documentation updated in this pull request? Or, a new nav item added to demonstrate a streaming endpoint? Now that the preview is auto-deployed (Thanks Vercel), it could be helpful for maintainers.

Rich-Harris:
Thank you for putting all this together. A lot has changed since this PR began — per discussion elsewhere we decided in favour of using `ReadableStream` rather than async iterables; between that and the many changes in #3384 it probably makes sense to close this in favour of #3419

kjmph:
Thanks @Rich-Harris!

Issue #1911, opened 2021-07-15
Cannot read property 'split' of undefined - Netlify adapter


Bradley-H:
### Describe the bug

When I upload my project I keep getting cannot read property of "split".

I suspect this has something to do with the adapter and not of Svelte itself since I can run a dev server just fine. 

For whatever it might be worth, here is my project in question - https://github.com/Bradley-H/portfolio

### Reproduction

upload to netlify

### System Info

```shell
It uploads to netlify just fine, no problem
```

### Severity

blocking an upgrade

here is my website on netlify if you need it - https://bradley-portfolio.netlify.app/

Bradley-H:
here is my website on netlify if you need it - https://bradley-portfolio.netlify.app/

0xPiyush:
Hello, I am facing the same issue, I tried deploying my Sveltekit app on netlify, on visiting the app, 
[https://sveltekit-pokedex.netlify.app/](https://sveltekit-pokedex.netlify.app/)
it shows:
`Cannot read property 'split' of undefined`

Here's the netlify function logs:

```
4:12:50 PM: a210dba2 ERROR  TypeError: Cannot read property 'split' of undefined
    at parse_body (/var/task/index.js:2082:57)
    at respond (/var/task/index.js:2171:15)
    at render (/var/task/index.js:2495:10)
    at Runtime.handler (/var/task/index.js:2722:26)
    at Runtime.handleOnce (/var/runtime/Runtime.js:66:25)
```

The build succeeds and the website gets deployed but shows the above error, when testing locally using npm run build / npm run preview, the site works as intended, but shows the above error when deployed.
If you want to see the web app's code here it is:
[https://github.com/0xPiyush/sveltekit-pokedex](https://github.com/0xPiyush/sveltekit-pokedex)

using svelte netlify adapter. [https://www.npmjs.com/package/@sveltejs/adapter-netlify](https://www.npmjs.com/package/@sveltejs/adapter-netlify)

node: v14.17.2
Dev Dependencies: 
    
```
"@sveltejs/adapter-netlify": "^1.0.0-next.22",
"@sveltejs/kit": "next",
"autoprefixer": "^10.3.1",
"svelte": "^3.34.0",
"tailwindcss": "^2.2.4"
```

netlify.toml file:
```
[build]
    command = "npm run build"
    publish = "build/"
    functions = "functions/"
[build.environment]
    NODE_VERSION = "14.17.2"
```

JeanJPNM:
[...]

PR #2215, opened 2021-08-16
[chore] Make `rawBody` be the bytes

### LINKED ISSUE #1893: Consider changing `rawBody` to be the bytes
### Describe the problem

`rawBody` sounds to me like its the bytes. Sometimes we put bytes in there and sometimes we put text in there depending on the content type. The biggest issue with this in my mind is that we're making each adapter responsible for that logic, but I think it could live in `kit`

Move body parsing from the adapter to `kit`

### Alternatives considered

status quo

### Importance

nice to have

### Additional Information

_No response_

JeanJPNM:

### Tests

### Changesets

Closes #1893 by ensuring `rawBody` is an `Uint8Array` before it is passed down to request handlers.


packages/kit/src/runtime/server/index.js
```diff
@@ -39,6 +39,10 @@ export async function respond(incoming, options, state = {}) {
 		...incoming,
 		headers,
 		body: parse_body(incoming.rawBody, headers),
+		rawBody:
+			typeof incoming.rawBody === 'string'
+				? new TextEncoder().encode(incoming.rawBody)
+				: incoming.rawBody,
```
Rich-Harris, line 45: is there a case where `rawBody` could be a string (other than an out-of-date adapter, in which case we could/should throw an error), or could we simplify this?

```suggestion
		rawBody: incoming.rawBody,
```

packages/kit/src/runtime/server/parse_body/index.js
```diff
@@ -2,30 +2,37 @@ import { read_only_form_data } from './read_only_form_data.js';
...
 
-	const [type, ...directives] = headers['content-type'].split(/;\s*/);
+	const content_type = headers['content-type'];
+	const [type, ...directives] = content_type ? content_type.split(/;\s*/) : [];
+
+	const text = () =>
+		typeof raw === 'string'
+			? raw
+			: new TextDecoder(headers['content-encoding'] || 'utf-8').decode(raw);
+	const bytes = () => (typeof raw === 'string' ? new TextEncoder().encode(raw) : raw);
```
Rich-Harris, line 17: assuming `raw` is always a `Uint8Array` can we simplify this?

```suggestion
	const text = () => new TextDecoder(headers['content-encoding'] || 'utf-8').decode(raw);
```

packages/kit/src/runtime/server/parse_body/index.js
```diff
@@ -2,30 +2,37 @@ import { read_only_form_data } from './read_only_form_data.js';
...
 
 		case 'multipart/form-data': {
 			const boundary = directives.find((directive) => directive.startsWith('boundary='));
 			if (!boundary) throw new Error('Missing boundary');
-			return get_multipart(raw, boundary.slice('boundary='.length));
+			return get_multipart(text(), boundary.slice('boundary='.length));
 		}
 		default:
-			throw new Error(`Invalid Content-Type ${type}`);
+			return bytes();
```
Rich-Harris, line 35:
```suggestion
			return raw;
```

Rich-Harris (commented):
thank you! i may not have fully understood everything, but it feels like we can simplify even further? comments inline

packages/kit/src/runtime/server/index.js
JeanJPNM: Right now these are the places where changes are necessary:

https://github.com/sveltejs/kit/blob/a24aba9120aa97e87d20b74ef7146ce737fb3b15/packages/kit/src/core/adapt/prerender.js#L153-L161
https://github.com/sveltejs/kit/blob/a24aba9120aa97e87d20b74ef7146ce737fb3b15/packages/kit/src/core/adapt/prerender.js#L285-L294
https://github.com/sveltejs/kit/blob/a24aba9120aa97e87d20b74ef7146ce737fb3b15/packages/kit/src/core/dev/index.js#L363-L371
https://github.com/sveltejs/kit/blob/a24aba9120aa97e87d20b74ef7146ce737fb3b15/packages/kit/src/runtime/server/page/load_node.js#L146-L154

Rich-Harris (commented):
thanks for those changes. i had one final question (see comments) then i think this is probably good to go

packages/kit/src/runtime/server/parse_body/index.js
```diff
@@ -1,31 +1,34 @@
 import { read_only_form_data } from './read_only_form_data.js';
 
 /**
- * @param {import('types/hooks').StrictBody} raw
- * @param {import('types/helper').Headers} headers
+ * @param {Uint8Array} raw
+ * @param {Partial<import('types/helper').Headers>} headers
```
ignatiusmb, line 5: Perhaps we can revisit #1611 if this is indeed the case for incoming headers, @benmccann?

packages/kit/src/core/node/index.js
```diff
@@ -1,15 +1,13 @@
...
- * @returns {Promise<import('types/hooks').StrictBody>}
+ * @returns {Promise<Uint8Array>}
  */
 export function getRawBody(req) {
 	return new Promise((fulfil, reject) => {
 		const h = req.headers;
 
 		if (!h['content-type']) {
-			return fulfil('');
+			return fulfil(new Uint8Array());
```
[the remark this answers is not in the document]
JeanJPNM: I think it's better to pass empty arrays instead of `null` for the following reasons:
- Typescript tells the user that `rawBody` will always be an `Uint8Array`.
- It requires an explicit cast to `any` which I consider unsafe (`/** @type {any} */ (null)`).
- It avoids needing to check if `rawBody` is `null` on other parts of the code

packages/kit/src/runtime/server/parse_body/index.js
JeanJPNM: I thought it would be safer to assume headers were optional, since that could have prevented #1911 (`headers['content-type']` was `undefined`)

Rich-Harris: What does `Partial<Record<string, string>>` even mean? Surely it's the same as `Record<string, string>`?

ignatiusmb: It means the same as `Record<string, string | undefined>`

packages/kit/src/core/node/index.js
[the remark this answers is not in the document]
Rich-Harris: Ah, of course — `strictNullChecks`. I do feel like `null` is a more correct representation of 'no body was provided' than an empty array. It also means we're not unnecessarily creating those typed arrays — the cost may be negligible, but I'd argue it's almost always better to avoid doing extra work at runtime for the sake of neater types. Tried it in 86c3d94409b78b791b7fe5fb6d8ea927726751d4, the tests still pass etc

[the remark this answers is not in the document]
JeanJPNM: Looks good to me, but I have one more question, should we update the `BaseBody` type to include `null`?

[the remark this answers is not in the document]
Rich-Harris: oh, right — yeah, we probably should

packages/kit/src/runtime/server/parse_body/index.js
Rich-Harris: Interesting, thanks. I think we probably just want the `Headers` type in that case, no? If a property exists on that object, it should be a string (if it's undefined, it's a bug)

JeanJPNM: Ok

Rich-Harris:
this change is a huge improvement — thank you!

PR #2218, opened 2021-08-16
[chore] separate packaging module

ignatiusmb:
Unit tests are low level, close to the source, and are supposed to run very quickly, which the `make_package` tests aren't. The packaging cli are also completely different than the other cli commands (and modules), which deserves its own place.

In trying to work on the `svelte-kit package` issues, I feel there are some things that can be improved to make development easier. This PR will
- move out and separate the `src/core/make_package` module into its own `src/packaging` that we can flesh out more along with its tests (the include/exclude files might not be working properly, and there aren't tests for it yet to confirm)
- separate the test scripts for packaging away from the other unit tests
- separate and create `src/utils` to contain some of the reusable modules (currently only filesystem)


dummdidumm (approved):
I like this. I was always searching a little before finding that part of the source code, simply because I thought it was a somewhat wrong place. The only slight oddity for me with this change is that `config` is inside `core` and used from `packaging`. I'm not sure if moving `config` to the top level is a good solution, though.

ignatiusmb:
I thought about moving them here too, but that would result in a huge files changed tab.

Arguably, `config` is also used by others besides kit, currently for packaging, maybe also for adapters in the future where they can adjust a finer output according the specified config, or even apps that want to build on top of kit (like what the community did for Sapper back then).

So maybe it makes sense to put it separately outside of core, we can then export a couple of helper functions to read the config file. Probably needs some discussion, but I hope this one gets through first as it's a blocker (for me) right now to work on the packaging issues.

PR #2222, opened 2021-08-16
Sequence

Rich-Harris:
Follow-up to #1951 — no need to create a separate app to test this, we can put it in `basics`. Also, I'm going to add some docs here

### Tests

### Changesets


Rich-Harris:
Hmm — there isn't really a place to document this at the moment, without documenting all the other `@sveltejs/kit/*` modules, which is out of the immediate scope. Going to merge this in then raise an issue for those doc changes

Rich-Harris:
whoops, i clean forgot to add a changeset for this. will add separately

DhyeyMoliya:
Could there be a case when we skip following handlers from an earlier handler and directly send the response to client

Rich-Harris:
Yes, absolutely — if an earlier handler doesn't call `resolve`, later handlers are not called

PR #2225, opened 2021-08-17
skip watcher test

Rich-Harris:
for reasons i don't understand, this test is causing merry havoc. just gonna skip it

### Tests

### Changesets


benmccann:
FYI @jasonlyu123 Rich and I were both having a test fail on our local machines since https://github.com/sveltejs/kit/pull/2198

Issue #1923, opened 2021-07-16
emit_dts in svelte-kit package should respect the file excludes config


truumahn:
### Describe the problem

It's fantastic that automatic type generation is included with the `package` command, however it seems `svelte2tsx` [doesn't accept any other configuration besides the lib root directory](https://github.com/sveltejs/language-tools/blob/master/packages/svelte2tsx/src/emitDts.ts#L8) so when the [types are generated](https://github.com/sveltejs/kit/blob/1f8abd9aa3f244c1011f2b65a5d20d489989c319/packages/kit/src/core/make_package/index.js#L17) with `svelte-kit package`, all files get generated types.

This is not desired when `.stories.svelte`, `.test.ts` or `.test.svelte` files (files that [help test slots as a workaround](https://svelte-recipes.netlify.app/testing/#testing-slots)) are next to the `.svelte` files in the `lib` directory.

### Describe the proposed solution

It would be great to have the file excludes respected during type generation. I understand that it might be `svelte2tsx` that needs to be altered to make this work.

### Alternatives considered

For now I think I'll include a `post-package` script that cleans up files that match the glob in `excludes`. It's a hacky workaround, why generate files that need to be cleaned up, right?

### Importance

would make my life easier

dummdidumm:
`svelte2tsx` is basically a wrapper around `tsc` which then does all the heavy lifting of generating types. I don't think TypeScript has options regarding what to generate and what not to generate. What could work though if you add an `excludes` property to your `tsconfig.json`  where you exclude all these `.test.svelte` files. Unless `.test.svelte` files are not imported from some other file in the `includes` array, they should be ignored.

truumahn:
But I do want these type checked, and the files you are referring to are one thing. I don't think it's uncommon to store component specific test files (in this case, `jest`) next to the components (e.g. `Component.test.ts`).

What you are saying is, if I want to keep type checking these files I should move them out of the `lib` directory so they don't get "packaged"?

dummdidumm:
Telling tsc what to compile and what not to compile is exactly what `includes` and `excludes` is for in their `tsconfig.json` . If you still want that type-checked, you can create a `tsconfig.test.json` and refer to that config when running type checks for these files.

I guess what I'm saying is
- `tsc` is not very configurable in that regard
- working arond that with multiple tsconfigs is cumbersome but probably the most common way to do this

[...]

PR #2236, opened 2021-08-18
[chore] more packaging tests fortification

ignatiusmb:
In working towards packaging issues and preparing some more tests for the uncovered settings, I found (another) hidden regressions within the tests itself (most likely from one of my early PRs back then when setting up the infrastructure, oops). This PR will

- [x] refactor `walk` into `utils/filesystem`
- [x] ~~add files include/exclude test suite~~ delegated to another PR
- [x] fix regression of nested folders equality testing
- [x] rename test folders to describe better what is being tested
- [x] print out more descriptive assertion error messages
- [x] remove unnecessary individual suites that can be merged into one test (now lives under the "standard javascript package" test suite)

~~Hopefully I'll be able to finish the checklist in the next couple of hours. Although, there might be a bug for the files include/exclude as well, so I might probably open a separate PR to add the tests and fix it if that's the case.~~

Opening another PR for the files include/exclude after this one's merged.


ignatiusmb:
Ugh, windows specific problem again, no wonder it keeps failing in my machine.

The new test fails on windows because globrex output for `**/exclude.js` is `/^.*\/exclude\.js$/`, but because windows separator uses backslash, it doesn't register properly. Expected files should be 5, and actual files are 7.

On the other hand, the other machines should've filter it properly. It did, but the generated `.d.ts` are still copied/generated across. Expected files should be 5, and actual files are 6.

dummdidumm:
Sounds like #1923

ignatiusmb:
Aha, exactly that!

ignatiusmb:
I guess https://github.com/sveltejs/kit/pull/2236/commits/752d9a23a96d9db7479902683d01b18a20beb2e0 would've passed for the other machines if `emitTypes` was set to false

dummdidumm (commented):
`emitTypes` should be renamed to `noEmitTypes` or `emitTypesFalse` because `emitTypes` sounds like the opposite of what it's testing. Other than that this looks good.

dummdidumm (approved):

benmccann:
I'm not that familiar with the packaging stuff, so don't wait on me for a review. I'd say go ahead and merge this whenever you're ready since you got one approval - especially since this is mostly tests and not a big API change or anything

PR #2247, opened 2021-08-19
[fix] exclude emitted declarations on packaging

### LINKED ISSUE #1923: emit_dts in svelte-kit package should respect the file excludes config

ignatiusmb:
This PR contains fixes for a couple of things, including but not limited to

- properly excluding files as well as its generated declarations
- improvements for the globbing pattern users can provide
- nested files not included/excluded on windows machines

Changing our globbing package allows end-users to provide more complex but easier (more familiar, bash-like) to write patterns. It's not as tiny in size as the previous one, but we're only using it for the packaging cli, and the patterns it allows our users to write also relieves a lot of normalizing and handling on our side as well.

Fixes #1923, waiting #2236 for this to be reviewed


packages/kit/src/packaging/index.js
```diff
@@ -21,8 +21,8 @@ export async function make_package(config, cwd = process.cwd()) {
 
 	const files_filter = create_filter(config.kit.package.files);
 	const exports_filter = create_filter({
-		...config.kit.package.exports,
-		exclude: [...config.kit.package.exports.exclude, '*.d.ts']
+		include: config.kit.package.exports.include,
```
dummdidumm, line 24: Why does this need to be set explicitely? Isn't it part of the spread already?

dummdidumm (approved):

packages/kit/src/packaging/index.js
ignatiusmb: Always found myself taking a while to read and understand when I get to this part. Explicitly setting the include option makes it clearer as it now has the same indentation, and avoids spreading and immediately overriding the exclude values.

dummdidumm: Makes sense 👍

ignatiusmb:
This may also remove handwritten `.d.ts` that are not part of the exclude pattern, but then again, it doesn't really make sense to publish the typings without the logic. Except for a different case when it only contains `export type`, in which case users should make sure it is isolated properly and won't be caught by their glob pattern, or rely on `.npmignore`.

I wanted to add some disclaimer in the docs that users are fully responsible for their glob patterns when they're using the include/exclude options. But, those who are using these options should have known better themselves. I guess we'll have to see if there's any complaint later on.

PR #2259, opened 2021-08-21
[chore] export App types

benmccann:
I *think* that exporting `App` will be enough to use `init` and `request`, but if not we can tweak a bit further

NOTE: If you open `adapter-node/src/server.js` in `master`, the import `@sveltejs/kit/node` cannot be resolved. I don't understand why. It'd be nice to use these new types in our own adapters, but I can't figure out why they're not available

### DIFF
--- a/.changeset/nice-ways-search.md
+++ b/.changeset/nice-ways-search.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[chore] export App types
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -105,7 +105,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 
 	const server_root = resolve_path(dir);
 
-	/** @type {import('types/internal').App} */
+	/** @type {import('types/app').App} */
 	const app = await import(pathToFileURL(`${server_root}/server/app.js`).href);
 
 	app.init({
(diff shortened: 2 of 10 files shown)


jthegedus:
>NOTE: If you open adapter-node/src/server.js in master, the import `@sveltejs/kit/node` cannot be resolved. I don't understand why. It'd be nice to use these new types in our own adapters, but I can't figure out why they're not available

In VSCode,`@sveltejs/kit` when 'clicked to follow' takes me to `@sveltejs/kit/types/index.d.ts`

With `@sveltejs/kit/node` I guess VSCode is looking for `@sveltejs/kit/types/node` but cannot find anything. There is a `@sveltejs/kit/dist/node.js` file which is perhaps what is meant to be imported.

Esbuild doesn't seem to have a problem resolving `@sveltejs/kit/node` during build.

jthegedus:
It seems the `init` type needs updating, or all the adapters. It requires 1 param, but currently each adapter calls init without a param.

https://github.com/sveltejs/kit/blob/de736f9350fe7d5952f99fc78d790004a9f759cd/packages/adapter-vercel/files/entry.js#L6

https://github.com/sveltejs/kit/blob/de736f9350fe7d5952f99fc78d790004a9f759cd/packages/adapter-node/src/index.js#L6

https://github.com/sveltejs/kit/blob/de736f9350fe7d5952f99fc78d790004a9f759cd/packages/adapter-netlify/files/entry.js#L4

benmccann:
Thanks for taking a look! You're right - that parameter needed to be optional

As for my issue using `@sveltejs/kit/node` it seems to be because TypeScript lacks support for `exports`: https://github.com/microsoft/TypeScript/issues/33079. I'll look at some of the workarounds on that thread

PR #2260, opened 2021-08-21
[fix] clear adapter-static output

### LINKED ISSUE #1044: `svelte-kit build` should respect `vite.build.emptyOutDir`
Building using `adapter-static`, files from the previous build are still in the build folder after a subsequent build. Vite has a [`build.emptyOutDir: true`](https://vitejs.dev/config/#build-emptyoutdir) default that svelte-kit seems to ignore.

It would be nice if svelte-kit cleared the build directory, to avoid the build folder ballooning in size when doing multiple local builds.

Running `rm -rf ./build && svelte-kit build` is a good enough workaround for now.

ignatiusmb:
Fixes #1044

### DIFF
--- a/packages/adapter-static/index.js
+++ b/packages/adapter-static/index.js
@@ -4,8 +4,9 @@ export default function ({ pages = 'build', assets = pages, fallback } = {}) {
 		name: '@sveltejs/adapter-static',
 
 		async adapt({ utils }) {
-			if (assets !== pages) utils.rimraf(assets);
+			utils.rimraf(assets);
 			utils.rimraf(pages);
+
 			utils.copy_static_files(assets);
 			utils.copy_client_files(assets);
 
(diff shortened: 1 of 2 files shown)


packages/adapter-static/index.js
```diff
@@ -11,6 +11,8 @@ export default function ({ pages = 'build', assets = pages, fallback } = {}) {
 		name: '@sveltejs/adapter-static',
 
 		async adapt({ utils }) {
+			if (assets !== pages) utils.rimraf(assets);
```
benmccann, line 14: I read this line and it didn't make any sense to me. I was sitting there puzzling over it for a few minutes before I read the next line and understood. It might make sense to reverse the order of the two lines. Or maybe better yet we don't even need the `if` since I'd expect it to have no negative effect or have much performance impact if we run it twice

benmccann (approved):

PR #2265, opened 2021-08-22
[fix] register custom service worker for AMP

### LINKED ISSUE #2264: Service worker breaks AMP validation
### Describe the bug

Issue: Creating `src/service-worker.ts` breaks AMP validation in Google Webmaster because svelte-kit injects custom `<script>` tag to install the service-worker.
webmaster error

Expected: svelte-kit should inject [amp-install-serviceworker](https://amp.dev/documentation/components/amp-install-serviceworker/) if `config.kit.amp` is `true` or let users to customize it.

PR: https://github.com/sveltejs/kit/pull/2265

### Reproduction

1. Create `src/service-worker.ts`
2. Set `config.kit.amp = true` in `svelte.config.js`
3. There will be no AMP validation error during development, but it will result on AMP entries being removed from google with the following error in Google webmaster dashboard: `Custom JavaScript is not allowed.`
Example: https://github.com/asendia/salmonfit/tree/f5aea01869760c83d57403a8cad3e1bfdee8e7aa

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 11.5.1
    CPU: (8) arm64 Apple M1
    Memory: 228.31 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.1.0 - ~/.nvm/versions/node/v16.1.0/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v16.1.0/bin/yarn
    npm: 7.11.2 - ~/.nvm/versions/node/v16.1.0/bin/npm
  Browsers:
    Chrome: 92.0.4515.159
    Firefox: 91.0.1
    Safari: 14.1.2
  npmPackages:
    @sveltejs/adapter-netlify: ^1.0.0-next.29 => 1.0.0-next.29 
    @sveltejs/kit: next => 1.0.0-next.154 
    svelte: ^3.34.0 => 3.42.1
```

### Severity

### Additional Information

Workaround that I did: ugly postinstall script to modify `ssr.js`: https://github.com/asendia/salmonfit/blob/3d9b0316a85c6164db01946529bc8b15d0ec31b8/scripts/replace-sw-amp.js

asendia:
Fixes #2264

An attempt to fix invalid service-worker script tag in AMP as described in the issue
I am still not sure what is the best way to set `data-iframe-src` since it must include the full domain name.

### DIFF
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -111,6 +111,7 @@ export async function render_response({
 		<style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style>
 		<noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
 		<script async src="https://cdn.ampproject.org/v0.js"></script>`;
+		init += options.service_worker ? '<script async custom-element="amp-install-serviceworker" src="https://cdn.ampproject.org/v0/amp-install-serviceworker-0.1.js"></script>' : '';
 	} else if (include_js) {
 		// prettier-ignore
 		init = `<script type="module">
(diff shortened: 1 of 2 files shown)


packages/kit/src/runtime/server/page/render.js
```diff
@@ -145,7 +148,11 @@ export async function render_response({
 	}
 
 	if (options.service_worker) {
-		init += `<script>
+		init += options.amp
+			? `<amp-install-serviceworker
+			src="${options.service_worker}"
+			layout="nodisplay"></amp-install-serviceworker>`
```
ignatiusmb, line 154: Can we keep this in one line?

packages/kit/src/runtime/server/page/render.js
```diff
@@ -145,7 +148,11 @@ export async function render_response({
 	}
 
 	if (options.service_worker) {
-		init += `<script>
+		init += options.amp
+			? `<amp-install-serviceworker
+			src="${options.service_worker}"
```
ignatiusmb, line 153: According to their [docs](https://amp.dev/documentation/components/amp-install-serviceworker/#src-(required)), src requires link with `https` protocol, which this option does not provide.

ignatiusmb (commented):
Thanks for the PR, but I don't think this is quite right

packages/kit/src/runtime/server/page/render.js
asendia: CMIW but I think it is fine to use relative path assuming that in production it will be deployed in `https` protocol. Like in this official example page: https://amp.dev/documentation/examples/components/amp-install-serviceworker/
```html
<amp-install-serviceworker
    src="/serviceworker.js"
    data-iframe-src="https://amp.dev/serviceworker.html"
    layout="nodisplay">
</amp-install-serviceworker>
```

ignatiusmb (approved):
Looks good enough, not sure about the actual usage, hoping others with more experience on amp would chime in

asendia:
> Looks good enough, not sure about the actual usage, hoping others with more experience on amp would chime in

Sorry if this is a stupid question since I am new to this, it seems like the test is failing on node 12 ubuntu-latest. Is this due to flaky test or something? 🙇🏼

ignatiusmb:
Yes, it should be fine

Rich-Harris:
(belatedly) this is great, thank you so much!

Issue #625, opened 2021-03-24
Type checking in adapters


benmccann:
# Issue

The adapters don't have TypeScript. This leads two two big issues.

1. we introduce bugs which would be caught with type checking https://github.com/sveltejs/kit/pull/600/files#r600142752
2. VS Code is painful to use because everything is red squigglies

![Screenshot from 2021-03-23 20-41-29](https://user-images.githubusercontent.com/322311/112252862-0886f500-8c1b-11eb-9b30-e58d96ac735c.png)

# Possible solutions

https://github.com/microsoft/TypeScript/issues/33079#issuecomment-738033267

Issue #2249, opened 2021-08-20
Expose adapter `init` & `render` types for Adapter Authors


jthegedus:
### Describe the bug

Updating the [Firebase adapter](https://github.com/jthegedus/svelte-adapter-firebase/) to the changes from https://github.com/sveltejs/kit/pull/2215 was made easier by typing `render`, however finding the exact types for this was difficult. Once I found the type my refactoring was much easier and I even discovered a bug in my adapter for the Header types.

Expose the types.

Not sure if this constitutes a bug or feature request. I went bug because I think this is expected but not the current case.

### Reproduction

You can see I have had to reference internal types `@sveltejs/kit/types/internal` to type this function:

```js
/**
 * @param {import('firebase-functions').https.Request} request
 * @return {import('@sveltejs/kit/types/internal').Incoming}
 */
function toSvelteKitRequest(request) {
```

https://github.com/jthegedus/svelte-adapter-firebase/blob/c9c355eed0b26225ca4c549f7966aa695bccba37/src/files/handler.js#L29

I am not even sure I got the correct type.

### System Info

```shell
svelte-adapter-firebase on  fix/sync-kit-rawbody-changes [$⇕] is 📦 v0.10.4 via ⬢ v14.17.5 on ☁️  
➜ npx envinfo --system --binaries --browsers --npmPackages "{svelte,@sveltejs/*,vite}"
npx: installed 1 in 1.319s

  System:
    OS: Linux 5.11 Pop!_OS 21.04
    CPU: (8) x64 AMD Ryzen 5 2500U with Radeon Vega Mobile Gfx
    Memory: 6.73 GB / 15.32 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 14.17.5 - ~/.asdf/installs/nodejs/14.17.5/bin/node
    npm: 6.14.14 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 92.0.4515.131
    Firefox: 91.0
  npmPackages:
    @sveltejs/kit: ^1.0.0-next.152 => 1.0.0-next.152
```

### Severity

serious, but I can work around it

### Additional Information

I believe exposing these types will make adapter development and refactoring more robust.

`init` is probably not necessary, though I do not know if it takes params and what they may be, so perhaps for visibility sake.

It would be good to just be able to force a type on 

```js
// TODO: hardcoding the relative location makes this brittle
import {init, render} from '../output/server/app.js';
[...]

Issue #2223, opened 2021-08-16
Missing `@sveltejs/kit/*` docs


Rich-Harris:
### Describe the bug

The docs include a section for the generated modules (`$app/*`, `$lib` and `$service-worker`) but not the `@sveltejs/kit/*` ones: https://kit.svelte.dev/docs#modules

We should add some. Not sure where the best place for it is. We can probably omit `adapter-utils` since that will presumably become obsolete with #2215.

### Reproduction

visit https://kit.svelte.dev/docs, look for the docs, find no docs

### System Info

N/A

### Severity

annoyance

benmccann:
I think this is mainly just `import { sequence } from '@sveltejs/kit/hooks';`, so we should probably put that with the `handle` docs

PR #2266, opened 2021-08-22
[chore] ambient modules and imports

ignatiusmb:
A stepping stone for #2223 and a groundwork for #2249


ignatiusmb:
We can bikeshed `ambient-imports.d.ts` and `ambient-modules.d.ts`, but maybe we'll just have it in one file for this PR now to prevent the scrambled diffing on ambient-modules.

packages/adapter-cloudflare-workers/index.js
```diff
@@ -8,15 +8,11 @@ import { fileURLToPath } from 'url';
...
  */
 
-/**
- * @param {{
- *   esbuild?: (defaultOptions: BuildOptions) => Promise<BuildOptions> | BuildOptions;
- * }} [options]
- **/
+/** @type {import('.')} */
 export default function (options) {
-	/** @type {import('@sveltejs/kit').Adapter} */
```
benmccann, line 17: I'm not sure about removing this. When I open up `adapter-static` in VS Code in `master` it complains: 
>Binding element 'utils' implicitly has an 'any' type

packages/adapter-node/src/server.js
```diff
@@ -28,9 +30,9 @@ export function createServer({ render }) {
 
 	const assets_handler = fs.existsSync(paths.assets)
 		? sirv(paths.assets, {
-				setHeaders: (res, pathname, stats) => {
-					// eslint-disable-next-line no-undef
-					if (pathname.startsWith(APP_DIR)) {
+				setHeaders: (res, pathname) => {
+					// @ts-ignore - dynamically replaced with define
```
benmccann, line 34: we should probably use `@ts-expect-error`

packages/adapter-node/src/index.js
```diff
@@ -1,6 +1,8 @@
 // TODO hardcoding the relative location makes this brittle
-import { init, render } from '../output/server/app.js'; // eslint-disable-line import/no-unresolved
-import { path, host, port } from './env.js'; // eslint-disable-line import/no-unresolved
+// @ts-ignore
```
benmccann, line 2: any idea why this is needed here, but not in `adapter-cloudflare-workers/files/entry.js`?

ignatiusmb: I was also scratching my head for a while trying to figure out this and ended up just ignoring this one, turns out the tsconfig for all the other serverless adapters weren't including their "source" files as well

This should just be a reminder though and will hopefully be resolved in the upcoming PR

ignatiusmb:
Reminder for the next PR to also resolve #625

packages/adapter-node/src/server.js
```diff
@@ -40,6 +42,7 @@ export function createServer({ render }) {
 		: noop_handler;
 
 	const server = polka().use(
+		// @ts-ignore TODO - compression return doesn't play well with polka.use
```
benmccann, line 45: see https://github.com/lukeed/polka/issues/173. maybe the solution is to cast it to `unknown` and then to `polka.Middleware`?  either that or I guess someone would need to dig into to understand what is different about the Polka types

packages/kit/src/runtime/client/start.js
```diff
@@ -1,7 +1,7 @@
 // @ts-ignore - value will be replaced on build step
-import Root from 'ROOT'; // eslint-disable-line import/no-unresolved
+import Root from 'ROOT';
```
benmccann, line 2: we should probably switch these two to `@ts-expect-error`

ignatiusmb:
Updated the comments. For the polka issue we'll just have to ignore them for now, casting the types would make sense if we're actually writing TS, but ignoring them makes more sense here, plus we've also put our reasons for every ignore and expect errors so it's clearer than casting (which we should try to avoid if possible).

dummdidumm (commented):
I don't know the reasons for adding `@sveltejs/kit/node` etc, so I can't say much about that. Type-wise, this looks good. If I remember correctly, the `type Page = import('@sveltejs/kit').Page` were needed because else they were not picked up during development on Kit itself. Did that change?

ignatiusmb:
Thanks, and yes! Other than having to disable eslint duplicate imports, it seems to be working fine now.

dummdidumm (approved):

benmccann (approved):

Issue #2185, opened 2021-08-13
Consider renaming `prerender.pages` option


benmccann:
### Describe the problem

`prerender.pages` config name is confusing. E.g. https://github.com/sveltejs/kit/issues/1683

### Describe the proposed solution

rename to something like:
- `prerender.entry`
- `prerender.entryPoints`
- `prerender.start`
- `prerender.startPages`

### Importance

nice to have

Mlocik97:
prerender.start is really dumb name... I think prerender.entry sound good.

MarioDena:
+1 on prerender.entry

PR #2276, opened 2021-08-24
[feat] customizable output file names for prerendering

### LINKED ISSUE #1443: Ability to prerender to `about.html` instead of `about/index.html` like Sapper

dishuostec:

### Tests

### Changesets

Fixes #1443 

This PR provide an optional callback to custom entry names for adapter-static.

Example:

- `blog/hello-world.html.svelte` => `blog/hello-world.html`
- `about.svelte` => `about.html`

```javascript
// svelte.config.js
import { format, parse } from 'path';

const config = {
	kit: {
		adapter: adapter({
			fallback: '200.html',
			outputFileName: ({ path, is_html }) => {
				if (!is_html) return path;

				let { root, dir, name, ext } = parse(path);

				if (!ext) ext = '.html';

				return format({ root, dir, name, ext });
			}
		}),
		target: '#svelte'
	}
};

export default config;
```

### DIFF
--- a/packages/adapter-static/README.md
+++ b/packages/adapter-static/README.md
@@ -42,19 +42,29 @@ Specify a fallback page for SPA mode, e.g. `index.html` or `200.html` or `404.ht
 
 A optional function that is called per entry to return the destination to write to.
 
-Default value:
+By default, prerendered html pages are output as index.html files in the directory matching the URL path.
+
+| URL       | output file          |
+| --------- | -------------------- |
+| /         | /index.html          |
+| /foo      | /foo/index.html      |
+| /bar.html | /bar.html/index.html |
+| /foo/bar  | /foo/bar/index.html  |
+
+For example, you may prefer `/foo.html` to `/foo/index.html`:
 
 ```javascript
-/** @type {(opts: {path: string, is_html: boolean}) => string} */
-function default_output_file_name({ path, is_html }) {
+import { format, parse } from 'path';
+
+outputFileName: ({ path, is_html }) => {
 	if (!is_html) return path;
 
-	const parts = path.split('/');
-	if (parts[parts.length - 1] === 'index.html') return path;
+	let { root, dir, name, ext } = parse(path);
 
-	parts.push('index.html');
-	return parts.join('/');
-}
+	if (!ext) ext = '.html';
+
+	return format({ root, dir, name, ext });
+};
 ```
 
 ## SPA mode
--- a/packages/adapter-static/README.md
+++ b/packages/adapter-static/README.md
@@ -38,6 +38,25 @@ The directory to write static assets (the contents of `static`, plus client-side
 
 Specify a fallback page for SPA mode, e.g. `index.html` or `200.html` or `404.html`.
 
+### outputFileName
+
+A optional function that is called per entry to return the destination to write to.
+
+Default value:
+
+```javascript
+/** @type {(opts: {path: string, is_html: boolean}) => string} */
+function default_output_file_name({ path, is_html }) {
+	if (!is_html) return path;
+
+	const parts = path.split('/');
+	if (parts[parts.length - 1] === 'index.html') return path;
+
+	parts.push('index.html');
+	return parts.join('/');
+}
+```
+
 ## SPA mode
 
 You can use `adapter-static` to create a single-page app or SPA by specifying a **fallback page**.
--- a/packages/adapter-static/index.js
+++ b/packages/adapter-static/index.js
@@ -1,8 +1,17 @@
-/** @type {import('.')} */
-export default function ({ pages = 'build', assets = pages, fallback } = {}) {
+/**
+ * @param {{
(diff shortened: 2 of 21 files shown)


packages/adapter-static/README.md
```diff
@@ -38,6 +38,25 @@ The directory to write static assets (the contents of `static`, plus client-side
 
 Specify a fallback page for SPA mode, e.g. `index.html` or `200.html` or `404.html`.
 
+### entry_file_names
```
benmccann, line 41: we use underscores for internal variable names and `camelCase` for public APIs

I think that using "entry" in the name here could be confused with the crawl entry points especially after https://github.com/sveltejs/kit/issues/2185 is implemented

I think a name like `outputFileName` might be a bit less likely to cause confusion

packages/adapter-static/README.md
```diff
@@ -38,6 +38,25 @@ The directory to write static assets (the contents of `static`, plus client-side
...
+
+A optional function that is called per entry to return the destination to write to.
+
+Default value:
+
+```javascript
+/** @type {(path: string, is_html: boolean) => string} */
+function defaultEntryFileNames(path, is_html) {
```
benmccann, line 49: We generally pass just a single argument to public APIs for backwards compatibility that way if we want to add a new argument or remove an argument it's easier to do with less impact on users. You can then destructure it for almost the same usage:
```
function defaultEntryFileNames({path, is_html}) {
```

packages/adapter-static/test/apps/spa-entry-name/README.md
```diff
@@ -0,0 +1,38 @@
+# create-svelte
```
benmccann, line 1: can we test this feature using an existing test project? we try to minimize the number of test projects we create in order to avoid time starting / stopping the server, etc.

dishuostec:
According to the coding style guidelines, renamed API option name and use deconstruction of object as argument.
Added changeset.

packages/adapter-static/test/apps/spa-entry-name/README.md
dishuostec: I have no idea how to test in one project using two different adapter options. The structure of build dir is totally different with or without this option. May be we can use subdir to split different strategies?

```text
routes
├── default
│   ├── index.svelte
│   └── sub
│       └── page.svelte
└── strip-index
    ├── index.svelte
    └── sub
        └── page.svelte
```

```javascript
outputFileName: ({ path, is_html }) => {
	if (!is_html) return path;

	if (path.startsWith('/strip-index/')) {
		// return custom name
	}

	// return default name
}
```

packages/adapter-static/README.md
```diff
@@ -38,6 +38,25 @@ The directory to write static assets (the contents of `static`, plus client-side
 
 Specify a fallback page for SPA mode, e.g. `index.html` or `200.html` or `404.html`.
 
+### outputFileName
+
+A optional function that is called per entry to return the destination to write to.
+
+Default value:
```
benmccann, line 45: Rather than include the code here, I would describe it textually like:

>By default, prerendered html pages are output as `index.html` files in the directory matching the URL path.

benmccann:
It looks like `pnpm lint` is failing. Please run `pnpm format` to fix it

packages/adapter-static/README.md
```diff
@@ -38,6 +38,25 @@ The directory to write static assets (the contents of `static`, plus client-side
 
 Specify a fallback page for SPA mode, e.g. `index.html` or `200.html` or `404.html`.
 
+### outputFileName
```
benmccann, line 41: I'd expect a new option to go with the pre-rendering options. There's nothing specific about `adapter-static` to it

https://kit.svelte.dev/docs#configuration-prerender

packages/adapter-static/index.js
```diff
@@ -1,8 +1,17 @@
...
+/**
+ * @param {{
+ *   pages?: string;
+ *   assets?: string;
+ *   fallback?: string;
+ *   outputFileName?: (opts: {path: string, is_html: boolean}) => string;
+ * }} [opts]
+ */
+export default function({ pages = 'build', assets = pages, fallback, outputFileName } = {}) {
+	/** @type {import('@sveltejs/kit').Adapter} */
```
ignatiusmb, line 10: Please revert these JSDoc changes

benmccann:
Thank you for this PR. I discussed it with the rest of the maintainers at today's maintainer's meeting and the thought was that we should just add a `boolean` to add the `.html` extension. Folks thought the need for the option was fairly specific and that adding a `function` was too broad of a solution and would be harder to use for the more narrow need that has been expressed

I really appreciate all the work that you've done on this. I'm going to go ahead and close it though given the feedback from the maintainers group. I'd be happy to consider a PR for the `boolean` option though

PR #2281, opened 2021-08-24
[fix] service worker scope and registration

### LINKED ISSUE #922: Service worker have a scope fixed at build time
**Describe the bug**
this line https://github.com/sveltejs/kit/blob/ccb8825fd97d7f23ee961b30077e1a2bf1a53b42/packages/kit/src/runtime/client/start.js#L53

will always register the service worker on a fixed url, which prevent it to be used on arbitrary sub folder, a requirement for ipfs webiste.

**Expected behavior**

svelte kit should allow us to provide a mechanism to define the registration behavior, including the service worker url at runtime

While it is possible to create the service worker separatly from svelte kit and then register it manually, this lose the benefit of the `$service-worker` modules.

One solution would be to separate the 2 : one option to generate the service worker and another option to decide whether to use the builtin service worker registration

This way I can still build my service worker with the benefit of the `$server-worker` module and register as I wish

userquin:
This PR fixes the service worker `scope` when registering it on navigator:
- add option to configure the sw scope
- add option to not register the sw: this will be usefull when using prompt for update or periodic service worker updates or another logic that requires using the service worker registration
- updated the example to include a sw with auto update behavior
- added `workbox` deps to build the sw on the example
- `.gitignore`: excluded IntelliJ/Webstorm stuff

closes #922

### DIFF
--- a/documentation/docs/06-service-workers.md
+++ b/documentation/docs/06-service-workers.md
@@ -6,12 +6,7 @@ Service workers act as proxy servers that handle network requests inside your ap
 
 In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker.ts`, or `src/service-worker/index.js`, etc) it will be built with Vite and automatically registered.
 
-> You can change the location of your service worker in your [project configuration](#configuration-files).
-You can also provide the `scope` for the service worker, by default `/`. You **MUST** modify you `manifest.json` file
-to include also the `scope`: if you are using the default `scope` value, then you need to configure `scope: '/'`.
-**REMEMBER**: both, the option and the `manifest.json` entry must be the same.
-You can remove the service worker registration to allow register it on your logic, for example to allow `prompt for update`
-or configure periodic service worker updates.
+> You can change the location of your service worker in your [project configuration](#configuration-files). You can also provide the `scope` for the service worker, by default `/`. You **MUST** modify your `manifest.json` file to include also the `scope`: if you are using the default `scope` value, then you need to configure `scope: '/'`. **REMEMBER**: both, the option and the `manifest.json` entry must be the same. You can remove the service worker registration to allow register it on your logic, for example to allow `prompt for update`, configure periodic service worker updates or some logic that requires using the service worker registration.
 
 Inside the service worker you have access to the [`$service-worker` module](#modules-$service-worker).
 
--- a/documentation/docs/06-service-workers.md
+++ b/documentation/docs/06-service-workers.md
@@ -7,11 +7,9 @@ Service workers act as proxy servers that handle network requests inside your ap
 In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker.ts`, or `src/service-worker/index.js`, etc) it will be built with Vite and automatically registered.
 
 > You can change the location of your service worker in your [project configuration](#configuration-files).
-<br />
 You can also provide the `scope` for the service worker, by default `/`. You **MUST** modify you `manifest.json` file
 to include also the `scope`: if you are using the default `scope` value, then you need to configure `scope: '/'`.
 **REMEMBER**: both, the option and the `manifest.json` entry must be the same.
-<br />
 You can remove the service worker registration to allow register it on your logic, for example to allow `prompt for update`
 or configure periodic service worker updates.
 
--- a/documentation/docs/06-service-workers.md
+++ b/documentation/docs/06-service-workers.md
@@ -7,6 +7,13 @@ Service workers act as proxy servers that handle network requests inside your ap
 In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker.ts`, or `src/service-worker/index.js`, etc) it will be built with Vite and automatically registered.
 
 > You can change the location of your service worker in your [project configuration](#configuration-files).
+<br />
+You can also provide the `scope` for the service worker, by default `/`. You **MUST** modify you `manifest.json` file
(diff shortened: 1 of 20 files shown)


userquin:
**NOTE**: since the `manifest.json` file is not generated by `svelte-kit`, if configured on options, it will be also required to add on the `manifest.json` file, must be in sync, the option and the manifest entry.

userquin:
Maybe we can use `base.path` configiration option, and also manipulate the `manifest.json` file to include also the `scope` (it is required and MUST MATCH `scope` on service worker registration.

userquin:
Don't merge yet, I'm fixing validate options for `serviceWorker`.

Also including auto update logic for sw.

documentation/docs/06-service-workers.md
```diff
@@ -7,6 +7,13 @@ Service workers act as proxy servers that handle network requests inside your ap
 In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker.ts`, or `src/service-worker/index.js`, etc) it will be built with Vite and automatically registered.
 
 > You can change the location of your service worker in your [project configuration](#configuration-files).
+<br />
+You can also provide the `scope` for the service worker, by default `/`. You **MUST** modify you `manifest.json` file
+to include also the `scope`: if you are using the default `scope` value, then you need to configure `scope: '/'`.
```
benmccann, line 12: please don't manually wrap the lines. just let it be one long line and then the text editor and browser can do word wrap

userquin: And how we can solve this?

packages/kit/src/core/build/index.js
```diff
@@ -51,6 +51,9 @@ export async function build(config, { cwd = process.cwd(), runtime = '@sveltejs/
 		}),
 		output_dir,
 		client_entry_file: `${SVELTE_KIT}/build/runtime/internal/start.js`,
+		service_worker_register:
+			config.kit.serviceWorker?.register === undefined ? true : config.kit.serviceWorker.register,
```
benmccann, line 55: The CI is failing and it's probably because of `?.`, which doesn't work on older versions of Node

userquin:
I fix the problem on `packages/kit/src/core/dev/index.js`, changing the internal types for `SSRRenderOptions`, seems there is some error (passing all test before pushing changes):

```shell
pnpm run check

> kit@0.0.1 check F:\work\projects\quini\sveltejs\svelte-kit
> pnpm -r check

Scope: all 12 workspace projects
packages/kit check$ tsc && svelte-check --ignore "src/packaging/test"
[3 lines collapsed]
│ Error while loading config at  f:\work\projects\quini\sveltejs\svelte-kit\packages\kit\src\core\config\test\fixtures\export-missing\svelte.config.js
│ Error: Missing exports in the config. Make sure to include "export default config" or "module.exports = config"
│     at ConfigLoader.loadConfig (F:\work\projects\quini\sveltejs\svelte-kit\node_modules\.pnpm\svelte-check@2.2.4_svelte@3.40.0\node_modules\svelte-check\dist\src\index.js:26666:23)
│     at async ConfigLoader.loadAndCacheConfig (F:\work\projects\quini\sveltejs\svelte-kit\node_modules\.pnpm\svelte-check@2.2.4_svelte@3.40.0\node_modules\svelte-check\dist\src\index.js:26656:46)
│     at async F:\work\projects\quini\sveltejs\svelte-kit\node_modules\.pnpm\svelte-check@2.2.4_svelte@3.40.0\node_modules\svelte-check\dist\src\index.js:26618:17
│     at async Promise.all (index 2)
│     at async ConfigLoader.loadConfigs (F:\work\projects\quini\sveltejs\svelte-kit\node_modules\.pnpm\svelte-check@2.2.4_svelte@3.40.0\node_modules\svelte-check\dist\src\index.js:26620:13)
│     at async createLanguageService (F:\work\projects\quini\sveltejs\svelte-kit\node_modules\.pnpm\svelte-check@2.2.4_svelte@3.40.0\node_modules\svelte-check\dist\src\index.js:79526:5)
│ ====================================
│ svelte-check found 0 errors, 0 warnings, and 0 hints
└─ Done in 16.1s
packages/adapter-static check$ tsc
└─ Done in 1.5s
packages/create-svelte check$ tsc
└─ Done in 2.5s
```

bluwy:
The error is fixed in the master branch. Should be gone once you rebase the branch.

ignatiusmb:
Can we please put the new config in alphabetical order, along with its usage

userquin:
> 
> 
> Can we please put the new config in alphabetical order, along with its usage

Done

documentation/docs/14-configuration.md
```diff
@@ -53,6 +53,8 @@ const config = {
 		},
 		router: true,
 		serviceWorker: {
+			register: boolean,
+			scope: string,
```
benmccann, line 57: It might be better to make this be the whole options object so that users can set `type` and `updateViaCache` with this option as well https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/register

benmccann:
The `register` option was added in https://github.com/sveltejs/kit/pull/2988

I don't think we'd want to update the example to use workbox

The one thing we'd possibly be open to is an option for the service worker scope. Anytime we add new API we need agreement from multiple maintainers, so I can't speak to it alone. Personally I'd recommend just setting `register: false` and providing your own registration with a custom scope

Given that there are quite a few merge conflict and that we wouldn't want to add workbox, I think the best thing to do is to close this PR. If you still wanted to raise the scope question it's probably cleaner to just do that in a new PR than try to update this one at this point

PR #2286, opened 2021-08-25
[fix] anchor tag inside svg

### LINKED ISSUE #2256: Anchor hrefs inside SVG elements not getting handled properly by router
### Describe the bug

Anchor tags inside SVG elements cause a TypeError in the router, and fall through to browser navigation.  This results in a full page load.  The issue seems to be that SVG elements are SVGAnimatedStrings, and the href attribute is returned as an object.  I have resolved this issue locally by checking if the element is an SVG, and if so, accessing a.href.baseVal, and including document.baseURI in the URL constructor.  This has to be done inside the "click" event handler and the trigger_prefetch function.

I'm happy to submit a PR if this solution is acceptable.

### Reproduction

- Initialize Demo App
- Add SVG inline with anchor tag to index.svelte
- Click on SVG element

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19042
    CPU: (8) x64 Intel(R) Core(TM) i7-8705G CPU @ 3.10GHz
    Memory: 2.08 GB / 15.82 GB
  Binaries:
    Node: 14.17.0 - C:\Program Files\nodejs\node.EXE
    npm: 7.14.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 92.0.4515.131
    Edge: Spartan (44.19041.1023.0), Chromium (92.0.902.73)
    Internet Explorer: 11.0.19041.906
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.154
    svelte: ^3.34.0 => 3.42.1
```

### Severity

### Additional Information

_No response_

ptizzy:
With anchor tags inside SVG elements, the href attribute is an object.  The href value is at a.href.baseVal.  However, this value differs from a normal anchor tag, in that it is not the full resolved path, so in order to pass it to a URL constructor, you have to include a base URL.  This change handles SVG anchor tags correctly, including prefetch, and in the click event handler function.

Fixes #2256


packages/kit/src/runtime/client/router.js
```diff
@@ -83,7 +83,11 @@ export class Router {
 		const trigger_prefetch = (event) => {
 			const a = find_anchor(/** @type {Node} */ (event.target));
 			if (a && a.href && a.hasAttribute('sveltekit:prefetch')) {
-				this.prefetch(new URL(/** @type {string} */ (a.href)));
+        const svg = typeof a.href === 'object' && a.href.constructor.name === 'SVGAnimatedString';
+        const href = String(svg ? /** @type {SVGAElement} */ (a).href.baseVal : a.href);
+        // Include baseURI if link is inside an svg
+        const base = svg ? document.baseURI : undefined
```
benmccann, line 89: `pnpm lint` is failing because of missing semi-colon
```suggestion
        const base = svg ? document.baseURI : undefined;
```

packages/kit/src/runtime/client/router.js
```diff
@@ -138,7 +142,10 @@ export class Router {
 			// Ignore if <a> has a target
 			if (svg ? /** @type {SVGAElement} */ (a).target.baseVal : a.target) return;
 
-			const url = new URL(href);
+
+      // Include baseURI if link is inside an svg
+      const base = svg ? document.baseURI : undefined
```
benmccann, line 147:
```suggestion
      const base = svg ? document.baseURI : undefined;
```

bluwy (commented):
Besides the formatting issues Ben pointed out, code lgtm.

packages/kit/src/runtime/client/router.js
```diff
@@ -83,7 +83,11 @@ export class Router {
 		const trigger_prefetch = (event) => {
 			const a = find_anchor(/** @type {Node} */ (event.target));
 			if (a && a.href && a.hasAttribute('sveltekit:prefetch')) {
-				this.prefetch(new URL(/** @type {string} */ (a.href)));
+        const svg = typeof a.href === 'object' && a.href.constructor.name === 'SVGAnimatedString';
```
benmccann, line 86: Maybe you should refactor out a method like:
```
function get_href(a) {
    const is_svg = typeof a.href === 'object' && a.href.constructor.name === 'SVGAnimatedString';
    const href = String(is_svg ? /** @type {SVGAElement} */ (a).href.baseVal : a.href);
    const base = is_svg ? document.baseURI : undefined;
 	return new URL(/** @type {string} */ (href), base);
}
```

these lines look to basically appear below as well. then you could share them

ignatiusmb:
It would be nice if these had some accompanying tests

bluwy:
Looks like spaces and tabs are being mixed too. we'll need to normalize/format them as well

ptizzy:
I can clean up the formatting easily enough.  I'll need some guidance regarding tests, though.  I'm not sure where to begin with that.

packages/kit/src/runtime/client/router.js
```diff
@@ -143,9 +152,8 @@ export class Router {
 			if (svg ? /** @type {SVGAElement} */ (a).target.baseVal : a.target) return;
 
 
-      // Include baseURI if link is inside an svg
-      const base = svg ? document.baseURI : undefined
-			const url = new URL(href, base);
+      const full_href = get_href(/** @type {HTMLAnchorElement | SVGAElement} */ (a));
+			const url = full_href;
```
benmccann, line 156: instead of creating two constants that are equal to each other you could combine these lines:
```
      const url = get_href(/** @type {HTMLAnchorElement | SVGAElement} */ (a));
```

packages/kit/src/runtime/client/router.js
```diff
@@ -16,6 +16,18 @@ function find_anchor(node) {
 	return /** @type {HTMLAnchorElement | SVGAElement} */ (node);
 }
 
+/**
+ * @param {HTMLAnchorElement | SVGAElement } node
+ * @returns {URL}
+ */
+function get_href(node) {
+	const svg = typeof node.href === 'object' && node.href.constructor.name === 'SVGAnimatedString';
+	const href = String(svg ? /** @type {SVGAElement} */ (node).href.baseVal : node.href);
```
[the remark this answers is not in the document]
benmccann: I wouldn't do any conversion. It looks like it's being done to satisfy TypeScript because `node.href` might not be a string if node is an svg element. However, we already know from the ternary operator that it's not. The issue then is that we're checking if the element is an SVG in a way TypeScript doesn't understand. I'd rewrite this whole function to be:
```
/**
 * @param {HTMLAnchorElement | SVGAElement} node
 * @returns {URL}
 */
 function get_href(node) {
	return node instanceof SVGAElement ? new URL(node.href.baseVal, document.baseURI) : new URL(node.href);
}
```

packages/kit/src/runtime/client/router.js
```diff
@@ -138,7 +151,7 @@ export class Router {
 			// Ignore if <a> has a target
 			if (svg ? /** @type {SVGAElement} */ (a).target.baseVal : a.target) return;
```
benmccann, line 152: We can get rid of this cast as well:
```
```

Issue #1683, opened 2021-06-13
Netlify adapter: prerender field of `svelte.config.js` basically doesn't work - must prerender per page


sw-yx:
**Describe the bug**

I built a small blog and tried using this in `svelte.config.js`:

```js
	kit: {
		target: '#svelte',
		adapter: adapter(),
		prerender: {
			crawl: true,
			force: true,
			pages: [
				'/',
				'/blog'
			],
		}
	}
```

and ran a build - none of these pages were built.

i also tried 

```js
			pages: [
				'/',
				'/*',
			],
```

and it didnt work. (you can check by running a build locally and checking the `/build` folder)

the only thing that worked was adding `export const prerender = true;` on a per page basis.

**To Reproduce**

- run build on this: https://github.com/sw-yx/swyxkit/tree/d74a0af94bbec011f58f35b8e787981f005539a6 no HTML pages built
- run build on this: https://github.com/sw-yx/swyxkit/tree/d5ad85f368f02cbf31d36a70d8fceea29b05a382 HTML pages are built

**Severity**

if this field just doesnt work for the netlify adapter yet, we should document or warn.

if this is a fixable or known issue, let me know and i can look into how to fix or document.

benmccann:
@sw-yx is this still happening with the latest versions of SvelteKit and the adapter? I did fix an issue with prerender recently that might have fixed this

sw-yx:
@benmccann i think i updated versions and see no change:
![CleanShot 2021-07-27 at 23 03 42@2x](https://user-images.githubusercontent.com/6764957/127271972-68f29b31-9966-401a-89a1-63f2407b4441.png)

benmccann:
https://kit.svelte.dev/docs#configuration-prerender

`pages` is where crawling starts. It's not a list of pages to prerender. I'd say this is working as expected

`prerender.pages` might not be the best name for the option. Maybe something like `prerender.crawlEntryPoints` would be better?

sw-yx:
@benmccann crawlEntryPoints would be more explicit, yes

benmccann:
filed a request to consider renaming

PR #2291, opened 2021-08-25
[breaking] update prerender options for providing list of pages

benmccann:
Closes https://github.com/sveltejs/kit/issues/2185

The main reason for this is to clarify the options. It was confusing to people that `pages` wasn't just a list of pages to prerender and that the option behaved differently depending on whether `crawl` was `true` or `false`

This also happens to be more powerful because you can specify some pages to crawl and some pages to pre-render without crawling, which you could not do before

For users that had set `pages` with `crawl: true` they will need to move that option value to `crawl`

### DIFF
--- a/.changeset/gorgeous-cougars-beg.md
+++ b/.changeset/gorgeous-cougars-beg.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] update prerender options for providing list of pages
--- a/documentation/docs/14-configuration.md
+++ b/documentation/docs/14-configuration.md
@@ -46,10 +46,10 @@ const config = {
 			base: ''
 		},
 		prerender: {
-			crawl: true,
+			crawl: ['*'],
 			enabled: true,
 			onError: 'fail',
-			pages: ['*']
+			pages: []
 		},
 		router: true,
 		serviceWorker: {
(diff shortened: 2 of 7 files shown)


dreitzner (changes requested):
As soon as the tests are fixed/extended this should be fine. 👏

Rich-Harris:
I don't understand this PR at all! `pages` and `crawl` are both treated identically here, no?

```js
if (config.kit.prerender.enabled) {
	if (config.kit.prerender.crawl) {
		visit_pages(config.kit.prerender.crawl);
	}
	visit_pages(config.kit.prerender.pages);
}
```

Even if that's unintentional, we're still left with a situation where `crawl` and `pages` have confusingly similar meanings. I'm open to changing things if people are confused (perhaps `pages` becomes `entries`?) but this doesn't seem like the right fix to me.

benmccann:
No, `pages` and `crawl` are not identical (unless I made a mistake somewhere). `pages` only pre-renders that exact list of pages and none of the linked pages whereas `crawl` will crawl all of the pages discoverable by links. Does it make any more sense to you now given that explanation?

Rich-Harris:
I'm afraid not, no. As far as I can see, `pages` and `crawl` are treated identically in this PR?

benmccann:
Oops. Yeah. I've fixed it so that it should behave the way I described now (I hope. I'm on my phone, so not ideal for code editing. It looks like it needs a rebase, so I can do that later today)

Rich-Harris:
Gotcha. I'm still pretty confused about why we'd want to do this — I can't really think of a situation where you'd want to crawl some pages but not _these specific pages_, which you still want to prerender. This PR wouldn't solve the problem swyx encountered in #1683, which was that he didn't understand that `pages` was a set of entry points for the crawler and that he'd still need to slap `prerender = true` on those pages unless he was using adapter-static.

I really think we should just rename `prerender.pages` to `prerender.entries` and keep `crawl` as it is.

PR #2307, opened 2021-08-27
[docs] syncs up rest of declarations

ignatiusmb:
#2299 was merged too quickly


ignatiusmb:
Oops, I missed a coupe of `=` and `;` left behind from converting type alias to interfaces

PR #2309, opened 2021-08-27
[docs] remove Server prefix from Request and Response types

benmccann:
I think this will make the docs easier to read than saying `ServerRequest`, but oh it's really called `Request`! :smile:

### DIFF
--- a/documentation/docs/01-routing.md
+++ b/documentation/docs/01-routing.md
@@ -69,8 +69,7 @@ export interface IncomingRequest {
 type ParameterizedBody<Body = unknown> = Body extends FormData
 	? ReadOnlyFormData
 	: (string | RawBody | ReadOnlyFormData) & Body;
-// ServerRequest is exported as Request
-export interface ServerRequest<Locals = Record<string, any>, Body = unknown>
+export interface Request<Locals = Record<string, any>, Body = unknown>
 	extends IncomingRequest {
 	params: Record<string, string>;
 	body: ParameterizedBody<Body>;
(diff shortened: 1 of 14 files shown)


benmccann:
They were already exported as `Request` and `Response`. It was just internally that we called them `ServerRequest` and `ServerResponse`

ignatiusmb:
There might be some more I'm missing, but these were the initial concerns and why it's prefixed with `Server`, not doing so 

- will override the global ones on files where we import our `Request` and `Response` types
- makes it harder to refactor, `Request` and `Response` is available globally and won't throw an error when checked
- makes it harder to understand on first glance when are we actually referring them to our request or the global one

We did write them as `Request` and `Response` previously while still having them prefixed with `Server` on the declaration types. But, then there isn't a clear distinction between ours and the global types used in both `LoadInput.fetch` and `ExternalFetch`.

The "it's written as ServerRequest, but really it's just Request" is trippy, but it's needed if we want users to understand our types and clarify the distinction with the global ones. Although, casual readers would probably only use the types in the examples and those who want to know more in-depth would most likely take a look at the source directly. Perhaps we should just simplify the types code blocks and put a link to the source somewhere near it?

benmccann:
Fair enough

I don't really like copy / pasting the code into the docs. It's hard to maintain. And it's hard to read because you have to look at the various referenced types to understand a single type. We could use typedoc, which would be much nicer. It's only a couple lines of code, but we'd have to put it in https://github.com/sveltejs/sites, which makes it slightly trickier because we'd have to check out this repo or something

PR #2313, opened 2021-08-28
[fix] adapter-cloudflare-workers: support setting multiple values to a header

### LINKED ISSUE #2305: [cloudflare adapter] Can't set multiple cookies
### Describe the bug

When setting multiple cookies from `hooks` in the `handle` function, I have the following code:
```js
return {
  // rest of response
  headers: {
    'set-cookie': ['cookie1', 'cookie2'],
  },
}
```

When using the node adapter, it correctly uses separate entries in the returned headers for each cookie.
```
set-cookie: cookie1
set-cookie: cookie2
```

However when running on cloudflare workers, with the cloudflare adapter, the cookies are merged into a single entry:
```
set-cookie: cookie1,cookie2
```

This results in only one cookie being set.

### Reproduction

Return multiple cookies and use the `@sveltejs/adapter-cloudflare-workers` adapter

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 11.5.2
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 407.40 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.7.0 - /usr/local/bin/node
    npm: 7.21.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 92.0.4515.159
    Safari: 14.1.2
  npmPackages:
    @sveltejs/adapter-cloudflare-workers: ^1.0.0-next.19 => 1.0.0-next.19 
    @sveltejs/adapter-node: ^1.0.0-next.43 => 1.0.0-next.43 
    @sveltejs/kit: next => 1.0.0-next.158 
    svelte: ^3.34.0 => 3.42.3
```

### Severity

### Additional Information

_No response_

JeanJPNM:

### Tests

### Changesets

Closes #2305 by using `Headers.append` on headers that have multiple values.

### DIFF
--- a/.changeset/quiet-crews-hunt.md
+++ b/.changeset/quiet-crews-hunt.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-cloudflare-workers': patch
+---
+
+Support assigning multiple values to a header
--- a/packages/adapter-cloudflare-workers/files/entry.js
+++ b/packages/adapter-cloudflare-workers/files/entry.js
@@ -41,7 +41,7 @@ async function handle(event) {
 		if (rendered) {
 			return new Response(rendered.body, {
 				status: rendered.status,
-				headers: rendered.headers
+				headers: makeHeaders(rendered.headers)
 			});
 		}
 	} catch (e) {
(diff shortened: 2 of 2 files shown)


half2me:
@JeanJPNM do you think it would be worth to check if headers is an instance of `Headers` and if so, just leave it as it is. Currently this breaks if `headers` is already a `Headers` object.

JeanJPNM:
That's supposed to be the original behavior, `headers` was supposed to be a plain object. being able to return an instance of `Headers` was a side effect of the implementation of `adapter-cloudflare-workers`.

half2me:
@JeanJPNM okay, makes sense :)

Issue #2230, opened 2021-08-17
Addding config.kit.paths.base breaks `npm run build`


marcel0ll:
### Describe the bug

On versions >1.0.0-next.147 adding a 'base' path to the project makes it no longer build.
This happens while using adapter-static

### Reproduction

1. Create new project

npm init svelte@next my-app
cd my-app
npm install
npm run dev -- --open

2. Install and config adapter-static 

Install with npm i -D @sveltejs/adapter-static@next, then add the adapter to your svelte.config.js:

```
// svelte.config.js
import adapter from '@sveltejs/adapter-static';

export default {
	kit: {
		adapter: adapter({
			// default options are shown
			pages: 'build',
			assets: 'build',
			fallback: null
		})
	}
};
```

3.  Config 'base' path
 
```
import adapter from '@sveltejs/adapter-static';

/** @type {import('@sveltejs/kit').Config} */
const config = {
	kit: {
		// hydrate the <div id="svelte"> element in src/app.html
		target: '#svelte',
		paths: {
			base: '/something'
		},
		adapter: adapter({
			// default options are shown
			pages: 'build',
			assets: 'build',
			fallback: null
		})
	}
};

export default config;
```

4.  `npm run build`

### Logs

```shell
vite v2.5.0 building SSR bundle for production...
✓ 34 modules transformed.
.svelte-kit/output/server/app.js   68.93 KiB

Run npm run preview to preview your production build locally.

> Using @sveltejs/adapter-static
> 404 /_app/start-518edf61.js (linked from /)
Error: 404 /_app/start-518edf61.js (linked from /)
...
```

### System Info

```shell
System:
    OS: Linux 4.15 Linux Mint 19.2 (Tina)
    CPU: (4) x64 Intel(R) Core(TM) i5-6300HQ CPU @ 2.30GHz
    Memory: 7.40 GB / 15.59 GB
    Container: Yes
    Shell: 4.4.20 - /bin/bash
  Binaries:
    Node: 14.17.0 - ~/.nvm/versions/node/v14.17.0/bin/node
    Yarn: 1.22.5 - /usr/bin/yarn
    npm: 6.14.13 - ~/.nvm/versions/node/v14.17.0/bin/npm
  Browsers:
    Brave Browser: 91.1.26.74
    Chrome: 91.0.4472.114
    Firefox: 90.0
  npmPackages:
    @sveltejs/adapter-static: ^1.0.0-next.16 => 1.0.0-next.16
[...]

PR #2318, opened 2021-08-29
[fix] replace kit.paths.base if it runs on localhost

baseballyama:
fix: https://github.com/sveltejs/kit/issues/2230

### Tests

### Changesets

---
## Explain What I Did

In this PR https://github.com/sveltejs/kit/pull/2189, the way of handle for `kit.paths.assets` in `svelte.config.js` was changed.
After that, the issue https://github.com/sveltejs/kit/issues/2230 comes.

I think this issue is only a problem in a dev environment like localhost.
(Please check details at [here](https://github.com/sveltejs/kit/compare/master...baseballyama:fix/2230?expand=1#diff-5ac38abd30867f8548e3fea0cb97d5f03a6cf239fca1db8ec03af61173db595bR259-R266))

So I just handled this issue as like workaround code.

As an alternative way, we can modify module bunder (like Vite).
In this case, we can create a folder which is the value of  `kit.paths.assets` as the parent of the `_app` folder. (And maybe something more.)
But in my opinion, we will have a chance to handle this issue for each module bundler.
This is just a little bit troublesome.
Therefore I choose this simple way.

So could you please review this way is reasonable?
Or should I modify Vite or somewhere else?

## MEMO

1. I don't know how to do the brawser test with `npm run dev` command. Please tell me how can I add a test for it.

2. When I execute `pnpx changeset`, it shows `If you are unsure if this is correct, contact the package's maintainers before committing this changeset`, so now I didn't create changeset. (After getting a review comment, I will create it.)

Thank you!

### DIFF
--- a/.changeset/small-pandas-search.md
+++ b/.changeset/small-pandas-search.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+To avoid 404, replace `kit.paths.base` to '' if it runs on dev server (localhost).
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -262,8 +262,13 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 					// In this case, paths between hrefs and actual bundled asserts are not matched.
 					// e.g.) `href` is `/some-basepath/_app/start-8e35c5cf.js`, and actual path is `/_app/start-8e35c5cf.js`
 					// Therefore we need to remove base path from hrefs.
-					// const resolved = resolve(path, href);
-					const resolved = resolve(path, href).replace(config.kit.paths.base, '');
+					const resolved = (() => {
+						const resolved1 = resolve(path, href);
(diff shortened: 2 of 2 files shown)


baseballyama:
I realized that just using replace() has chance to mis-replacement if middle of file path has path which is same as base path.
So I will fix it but I'm happy if someone review for this fixing policy!

dreitzner (approved):
Lgtm

benmccann:
Another PR to fix this issue has been submitted at https://github.com/sveltejs/kit/pull/2407. I'm not that familiar with this code and haven't looked closely at either PR yet to know which makes more sense

baseballyama:
I discussed about this PR with the author of https://github.com/sveltejs/kit/pull/2407.
And I think we should adopt https://github.com/sveltejs/kit/pull/2407 instead of this PR.
Therefore I will close this PR after merge https://github.com/sveltejs/kit/pull/2407 if nothing happens.

benmccann:
Ok, I'll just go ahead and close it now then

PR #2324, opened 2021-08-30
[docs] Remove comment about fixed security issue

andrewbanchich:
Looks like [the security issue](https://github.com/vitejs/vite/issues/2820) with Vite mentioned in the docs has been addressed, so this removes the comment.


benmccann:
Thanks!

PR #2327, opened 2021-08-30
[feat] merge packaging exports

### LINKED ISSUE #1944: A way append to `svelte-kit package`'s exports output.
### Describe the problem

`svelte-kit package`'s current implementation does not export components [if I've included exports in my `package.json`](https://github.com/sveltejs/kit/blob/master/packages/kit/src/core/make_package/index.js#L84).

I propose making appending exports to be the default behaviour, since if I don't want `svelte-kit package` to touch my exports I can just do `exports:{ exclude: ["**"] }`.

### Alternatives considered

Scripting to add a custom export after `svelte-kit package` is done running.

### Importance

would make my life easier

### Additional Information

_No response_

ignatiusmb:
Resolves #1944 

Still not sure if `package.exports.behavior` is a good settings, inputs are welcome. Also note to self, need to add some docs.


packages/kit/types/config.d.ts
```diff
@@ -52,6 +52,7 @@ export interface Config {
 			dir?: string;
 			emitTypes?: boolean;
 			exports?: {
+				behavior?: 'replace' | 'merge';
```
benmccann, line 55: these are probably fine defaults. I wonder if there won't be more complicated cases where you want a mixture or something in which case maybe you'd need a `function` or something along those lines. I'm trying to think of what a good name could be if we added more options here in the future. maybe `mergeStrategy`?

ignatiusmb: A mixture should be available to do if the user chooses `merge` behavior and provides the `include`/`exclude` as well.

Only case I can think of right now are these two,, maybe we should change it to a simple boolean instead if there's isn't any other cases, hoping others could chime in on this too.

packages/kit/src/packaging/index.js
```diff
@@ -86,19 +82,20 @@ export async function make_package(config, cwd = process.cwd()) {
...
-	const main = pkg.exports['./index.js'] || pkg.exports['./index.svelte'];
-
-	if (!user_defined_exports && main) {
-		pkg.exports['.'] = main;
+	if (!pkg.exports) {
+		pkg.exports = generated;
+	} else if (config.kit.package.exports.behavior === 'replace') {
+		pkg.exports['./package.json'] = './package.json';
+	} else if (config.kit.package.exports.behavior === 'merge') {
+		pkg.exports = { ...generated, ...pkg.exports };
```
dummdidumm, line 97: This _might_ not be enough for some, they could expect a deep merge. Not sure what the best default here would be.

packages/kit/src/packaging/test/fixtures/exports-replace/package.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "exports-replace",
 	"version": "1.0.0",
-	"description": "default exports settings",
+	"description": "exports.behavior settings set to replace",
 	"exports": {
 		".": {
 			"import": "./index.js"
```
dummdidumm, line 7: I would change the export here to something else to test it really is overwritten. Right now the test result is that only `package.json` is appended.

packages/kit/src/packaging/index.js
```diff
@@ -86,19 +82,20 @@ export async function make_package(config, cwd = process.cwd()) {
...
 	}
 
-	const main = pkg.exports['./index.js'] || pkg.exports['./index.svelte'];
-
-	if (!user_defined_exports && main) {
-		pkg.exports['.'] = main;
+	if (!pkg.exports) {
+		pkg.exports = generated;
+	} else if (config.kit.package.exports.behavior === 'replace') {
+		pkg.exports['./package.json'] = './package.json';
```
dummdidumm, line 95: I'm not sure this is correct. If there are exports present already, and mode is `replace`, only the `package.json` is overwritten (at least I read the code like that, maybe something above this `if` happens that changes this). Altering the replace-test (see my other comment) might uncover this.

dummdidumm (commented):
I think this is a good addition. I'm split on whether we should do a boolean (that could become a function) or do the enum. If there's more than true/false/function, then definitely the enum, but other than `deepMerge` I don't see a valid third option (and maybe `merge` should merge deep anyway).

ignatiusmb:
Come to think of it, there can't be a situation where both `behavior` is set to `replace` and `include`/`exclude` are also used. What if we set the default value of `kit.package.exports` to `null`, so exports map in package.json wouldn't get tampered by SvelteKit when user has defined it by themselves, but if it's empty, it will use the generated version.

One awkward thing to this is that users that would like to get their self-defined exports map merged but still have all other exports generated would need to pass an empty object to the config, `exports: {}`, to indicate they're not replacing the generated ones. But, this allows us to keep the config as is, any thoughts @benmccann @dummdidumm?

packages/kit/src/packaging/test/fixtures/exports-replace/package.json
dummdidumm: Ah yeah my bad, I got confused by what replaces what there 😄

documentation/docs/14-configuration.md
```diff
@@ -132,8 +129,8 @@ Options related to [creating a package](#packaging).
 
 - `dir` - output directory
 - `emitTypes` - by default, `svelte-kit package` will automatically generate types for your package in the form of `d.ts.` files. While generating types is configurable, we believe it is best for the ecosystem quality to generate types, always. Please make sure you have a good reason when setting it to `false` (for example when you want to provide handwritten type definitions instead).
-- `exports` - contains a `includes` and a `excludes` array which specifies which files to mark as exported from the `exports` field of the `package.json`
-- `files` - contains a `includes` and a `excludes` array which specifies which files to process and copy over when packaging
+- `exports` - specifies which files to mark as exported from the `exports` field of the `package.json`, can be set to `null` and will use existing `exports` field if available, or an object that contains an `includes` and an `excludes` array and will merge existing `exports`, with the value from `package.json` taking precedence
+- `files` - contains an `includes` and an `excludes` array which specifies which files to process and copy over when packaging
```
ignatiusmb, line 133: Thinking of expanding this section to use `####` headings instead

dummdidumm: I think it's not necessary yet. But the sentence could be split into multiple sentences to make reading a little easier.

dummdidumm:
I'm not completely sure I understand. Rereading #1944, there's truth in the OP's post: " if I don't want svelte-kit package to touch my exports I can just do `exports:{ exclude: ["**"] }`." So maybe the best way would be to just skip the extra layer of "is this `null`" or "add a `behavior` param" and instead request users to configure their `include`/`exclude` correctly. The new default behavior would be to merge everything present in `package.json` - why else would you add exports there if you want them overwritten?

ignatiusmb:
What would I do without you and your reviews @dummdidumm

dummdidumm (approved):
Well I didn't get to this in hindsight obvious conclusion, too, at first 😄

PR #2328, opened 2021-08-30
[fix] support using arrays for kit.vite.resolve.alias

### LINKED ISSUE #2319: How to use kit.vite.resolve.alias if SvelteKit doesn't not allow this config?
### Describe the bug

I need to use https://vitejs.dev/config/#resolve-alias in my `svelte.config.js` and I'm using it like this:

```js
resolve: {
  alias: [
    {
      find: /^custom\/node_modules\/package!(.*)/,
      replacement: "my-custom-alternative",
    },
  ],
},
```

but when I build SvelteKit writes in red in my console this text:

`build_client: The value for kit.vite.resolve.alias specified in svelte.config.js has been ignored. This option is controlled by SvelteKit.`.

How to use aliases now?

### System Info

```shell
System:
    OS: Windows 10 10.0.19043
  Binaries:
    Node: 14.17.5 - C:\Program Files\nodejs\node.EXE
    npm: 7.21.0 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    @sveltejs/adapter-static: 1.0.0-next.17 => 1.0.0-next.17
    @sveltejs/kit: 1.0.0-next.159 => 1.0.0-next.159
    svelte: 3.42.4 => 3.42.4
```

### Severity

bluwy:
Fixes #2319

Thanks to @JeanJPNM for pointing out the additional places to update.

Notes:

1. Maybe we can do the entire config merge as a Vite plugin per this PR's implementation?
2. I can't put `packages/kit/test/global.d.ts` into `packages/kit/test/ambient.d.ts` because the latter has `export {}` in it, making it a TS module.

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/apps/basics/src/routes/alias/_tests.js
+++ b/packages/kit/test/apps/basics/src/routes/alias/_tests.js
@@ -1,8 +1,0 @@
-import * as assert from 'uvu/assert';
-
-/** @type {import('test').TestMaker} */
-export default function (test) {
-	test('handles resolve.alias as array', '/alias', async ({ page }) => {
-		assert.equal(await page.textContent('p'), 'foo');
-	});
-}
--- a/packages/kit/test/apps/basics/src/routes/alias/_tests.js
+++ b/packages/kit/test/apps/basics/src/routes/alias/_tests.js
@@ -1,0 +1,8 @@
+import * as assert from 'uvu/assert';
+
+/** @type {import('test').TestMaker} */
+export default function (test) {
+	test('handles resolve.alias as array', '/alias', async ({ page }) => {
+		assert.equal(await page.textContent('p'), 'foo');
+	});
+}
--- a/.changeset/sour-waves-return.md
+++ b/.changeset/sour-waves-return.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+support using arrays for kit.vite.resolve.alias
--- a/packages/kit/src/core/build/index.js
+++ b/packages/kit/src/core/build/index.js
@@ -166,12 +166,6 @@ async function build_client({
 				preserveEntrySignatures: 'strict'
 			}
 		},
-		resolve: {
-			alias: {
-				$app: path.resolve(`${build_dir}/runtime/app`),
-				$lib: config.kit.files.lib
-			}
-		},
 		plugins: [
 			svelte({
 				extensions: config.extensions,
@@ -179,7 +173,19 @@ async function build_client({
 				compilerOptions: {
 					hydratable: !!config.kit.hydrate
 				}
-			})
+			}),
+			{
+				name: 'sveltekit:alias',
+				// leverage vite's config merge since alias can be an object or array
+				config: () => ({
+					resolve: {
+						alias: {
+							$app: path.resolve(`${build_dir}/runtime/app`),
+							$lib: config.kit.files.lib
+						}
+					}
+				})
+			}
 		]
 	});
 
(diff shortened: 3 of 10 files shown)


benmccann:
thanks for the fix! it does feel a bit weird to me to do config merging in two different ways with this PR as it is right now. I would lean towards saying we either make the existing SvelteKit config merging work with regexes or do it with a Vite plugin for the whole Vite config blob (not sure if it'd be the same plugin or not, but I was also planning on adding a Vite plugin in https://github.com/sveltejs/kit/pull/2232 to provide the SvelteKit middleware to the Vite server after https://github.com/vitejs/vite/pull/4640 is merged)

dominikg:
i'm in favor of using vite plugins to provide additional config, note that you cannot add vite-plugin-svelte via a config hook though (plugins are final before the hooks are called). But everything else can be done via one or multiple `sveltekit:config` plugins with respective hooks.

benmccann:
hmm, I wonder if it just makes things more complicated. we'll still need the sveltekit config merging for non-vite stuff. I wonder if it wouldn't just be simpler to fix the sveltekit config merging to handle regexes instead. it already handles functions, so shouldn't be much different from that

JeanJPNM:
The problem isn't because of the regexes, is because the `alias` option had an array as a value, wich cannot be merged with an object.

Here is how it should be:
https://github.com/sveltejs/kit/blob/bf3efd10fbd599d0feebdefd6cf4a53bd4dce398/packages/kit/src/core/dev/index.js#L121-L136

But this is a lot more verbose.

JeanJPNM:
I think we should create a `plugins` and make plugins for general config, dev and build, to avoid this kind of code duplication.

JeanJPNM:
The options I see are:
- Hard coded replace
- Anonymous plugins
- Named plugins
- Function specifically made to merge aliases

benmccann:
ah, thanks for clarifying. it looks like it was handled correctly in `dev`, but not in `build` :stuck_out_tongue_closed_eyes:

I see that `plugins.svelte.emitCss` is also handled differently between the two and I have no idea why - that also looks like it might be broken in some way

bluwy:
Looking back now, I think adjusting the `deep_merge` to take into account of of `resolve.alias` would be cleaner. The extra plugin do seem off, but I think it's the safer choice in the long run if we do let Vite handle the config merging for us. [resolveConfig](https://vitejs.dev/guide/api-javascript.html#resolveconfig) might be handy too. I can look into that soon, but for now I'll refactor the code to not use a Vite plugin

bluwy:
Updated to have `deep_merge` take into account of `resolve.alias`, instead of using Vite plugins.

packages/kit/test/apps/basics/src/routes/alias/_tests.js
```diff
@@ -0,0 +1,8 @@
+import * as assert from 'uvu/assert';
+
+/** @type {import('test').TestMaker} */
+export default function (test) {
+	test('handles resolve.alias as array', '/alias', async ({ page }) => {
```
benmccann, line 5: could we unit test the config merging rather than having an integration test? the browser tests are kind of flaky and this seems like something we should be able to test without a server and browser

bluwy: yeah that makes sense after the change I made. Will move it as a unit test.

bluwy: Added a new unit test

benmccann:
awesome! this looks great! thanks so much :smile:

Issue #2086, opened 2021-08-04
SvelteKit 1.0 release blockers in Vite


benmccann:
### Update

Vite 2.7 fixed most issues that users faced. If you're having issues with a library please make sure you're using Vite 2.7. If you're still having problems, please file an issue

### Describe the bug

SvelteKit apps are no different from any other Vite app from the most part. The unique things that they do are include [`vite-plugin-svelte`](https://github.com/sveltejs/vite-plugin-svelte) and [automatically add Svelte component libraries to `optimizeDeps.exclude` and `ssr.noExternal`](https://github.com/sveltejs/kit/blob/7c1d7e6ddc75fbb00b6c6b7db67342227d2c60a4/packages/kit/src/core/build/index.js#L465) (this latter functionality may be replaced with the `experimental.prebundleSvelteLibraries` option)

**Blockers**

https://github.com/vitejs/vite/issues/2579 / https://github.com/vitejs/vite/issues/3024 / https://github.com/vitejs/vite/issues/3910 (prebundling skips `.svelte`) - experimental solution. see [`prebundleSvelteLibraries`](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#prebundlesveltelibraries)
https://github.com/vitejs/vite/pull/7154 (externalizing of nested deps) - slated for 2.9
~~https://github.com/sveltejs/kit/issues/3118 / https://github.com/sveltejs/kit/issues/3053 (`package.json` issues)~~ - fixed in 2.8
~~https://github.com/vitejs/vite/issues/5890 (not externalizing dependency)~~ - fixed in 2.7
~~https://github.com/vitejs/vite/issues/3953 (resolving packages with `exports`)~~ - fixed in 2.7
~~https://github.com/vitejs/vite/issues/5446 (`context="module"` not scanned in separate scope)~~ - fixed in 2.7
~~https://github.com/vitejs/vite/issues/4425 (no main - only sub-directories exported. e.g. firebase)~~ - fixed in 2.7
~~https://github.com/vitejs/vite/pull/5197 (ESM-only deps)~~ - fixed in 2.7
~~https://github.com/vitejs/vite/issues/4306 (svelte store functions overwritten)~~ - fixed in 2.6.11
~~https://github.com/vitejs/vite/pull/4699 (can't use SvelteKit-provided libraries - https://github.com/sveltejs/kit/issues/2237)~~ - fixed in 2.5.3
[...]

PR #2343, opened 2021-09-01
Chore/update vite plugin svelte

dominikg:

### Tests

### Changesets

with kit already updated to vite-2.5.3 , users can update to vite-plugin-svelte@1.0.0-next.20 on their own but this should increase adoption-rate and visibility for a feature that should solve a lot of pain people were having with transitive cjs dependencies.

Not sure if patch or minor :man_shrugging:

### DIFF
--- a/.changeset/twenty-hairs-shave.md
+++ b/.changeset/twenty-hairs-shave.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Enable nested dependency optimization by updating to @sveltejs/vite-plugin-svelte@1.0.0-next.20
--- a/.changeset/twenty-hairs-shave.md
+++ b/.changeset/twenty-hairs-shave.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Enable nested dependency optimization by updating to @sveltejs/vite-plugin-svelte@1.0.0-next.20
+Enable nested dependency optimization by updating to @sveltejs/vite-plugin-svelte@1.0.0-next.21
--- a/documentation/faq/00-other-resources.md
+++ b/documentation/faq/00-other-resources.md
@@ -2,4 +2,4 @@
 question: Other resources
 ---
 
-Please see [the Svelte FAQ](https://svelte.dev/faq) and [`vite-plugin-svelte`](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md) FAQ as well for the answers to questions deriving from those libraries.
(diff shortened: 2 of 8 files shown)


bluwy:
Should we remove the old auto-exclude code here too? (even though it should still work without). And we could also consider removing [this faq section](https://kit.svelte.dev/faq#packages).

dummdidumm (approved):

dominikg:
yes, i think we should, good catch @bluwy . No point in sveltekit and vite-plugin-svelte doing the same thing.
Also updating the docs to mention the new automatic behavior is a good idea. Should be done on vite-plugin-svelte tho with a link from here.

dominikg:
docs change in vite-plugin-svelte https://github.com/sveltejs/vite-plugin-svelte/pull/158

bluwy:
> I'd maybe just remove one sentence of the FAQ:

@benmccann I think the first and second paragraph of that section can be removed though, as it should never happen anymore. The second paragraph also has a broken link. Removing those would also made the third paragraph redundant.

Perhaps we could replace with the ESM-only dep issue, or some other issues listed at https://github.com/sveltejs/kit/issues/2086? as they would now be more common.

benmccann:
The link in the second paragraph of the FAQ appears to be broken, so yeah we can remove the second paragraph. Perhaps we should add something at the very top of the SvelteKit FAQ that points at https://svelte.dev/faq and https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md

For the first perhaps we can rewrite it to something like:
> SSR in Vite is not yet stable. Libraries work best with Vite when they distribute both CJS and ESM in their package and you may wish to work with library authors to make this happen.

benmccann:
Oh crap. I thought I was being helpful and making some docs changes to this PR, but accidentally pushed them to `master` instead. @dominikg I just dropped your last commit from this PR to avoid the merge conflict and added a new PR to fix the lint error by formatting. I can merge this then if everything looks okay to you guys now

dominikg:
maybe wait a little, @bluwy  found an edge case where having packages without a valid export like `@types/node`  as dependency of a svelte dep (eg routify) breaks vites optimizer. We're adding an option to exclude nefarious packages or honor existing excludes.

So there will be next.21 soonish

dominikg:
next.21 is released

.changeset/twenty-hairs-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Enable nested dependency optimization by updating to @sveltejs/vite-plugin-svelte@1.0.0-next.20
```
benmccann, line 5:
```suggestion
Enable nested dependency optimization by updating to @sveltejs/vite-plugin-svelte@1.0.0-next.21
```

Issue #1891, opened 2021-07-12
Add directory imports when generating packages


Rich-Harris:
### Describe the problem

`npm run package` creates a default entry point for an `index.js` (or `index.svelte`) file directly inside `src/lib`, but not for other index files. In other words these files...

```
src/lib/index.js
src/lib/foo/index.js
src/lib/bar.svelte
```

...will result in the following `pkg.exports`:

```js
"exports": {
  "./package.json": "./package.json",
  "./index.js": "./index.js",
  "./foo/index.js": "./foo/index.js",
  "./bar.svelte": "./bar.svelte",
  ".": "./index.js"
}
```

### Describe the proposed solution

I think this would be preferable:

```diff
"exports": {
  "./package.json": "./package.json",
- "./index.js": "./index.js",
- "./foo/index.js": "./foo/index.js",
+ "./foo": "./foo/index.js",
  "./bar.svelte": "./bar.svelte",
  ".": "./index.js"
}
```

In other words, index files are never referenced directly, only by the containing folder. (A clashing `index.js` and `index.svelte` would presumably be an error in this scenario.)

### Alternatives considered

There are basically four options:

1. always be explicit (i.e. even `src/lib/index.js` doesn't get special treatment). Consistent, but ugly
2. the status quo (make an exception for the main entry point)
3. generate directory entry points in _addition_ to the existing file entry points. Consistent, but ugly, and probably a bit confusing (which do you document? both? ugh)
4. this proposal: only generate directory entry points for any `index.[whatever]` files. Slightly less consistent, but a definite aesthetic win. This sort of thing is encouraged by the design of `pkg.exports`, even though native ESM ordinarily requires you to be very explicit, and it will be familiar to Node users (i.e. basically all of us)

### Importance

would make my life easier

dummdidumm:
Option 5: set the `exports` field yourself manually. This was changed recently; that field is not generated if it already exists on your `package.json`. Things like this sound like use cases for doing this and keep the config somewhat lean.

Rich-Harris:
Sure, but we still need a sensible default. There's value in having predictable behaviour — if people are using the escape hatch of setting `exports` themselves it's because the framework itself is falling short

PR #2345, opened 2021-09-01
[feat] file imports in exports map

ignatiusmb:
Related to #1891 

We should also be able to import js files without having to specify the extension. Question is, do we still keep the the export with the extension? I don't know any use case out there that needs to explicitly do `import stuff from 'package/module/stuff.js'`, which can simply be `'package/module/stuff'`

Drafting this first to wait for test cases from #2327


ignatiusmb:
Does that also applies to imports from `node_modules`? I only encounter those errors from relative paths so far like `./foo/hello.js` or `../utils/index.js`, don't know if `my-module/utils/stuff.js` can be `my-module/utils/stuff` or needs it's extension too.

We can probably export both to be on the safe side? At the cost of increased package.json file size
```json
"exports": {
  "./foo/utils.js": "./foo/utils.js",
  "./foo/utils": "./foo/utils.js"
}
```

dummdidumm:
I think with ESM type imports you need that explicit exports map to import anything not imported through the base path (`package/subpath` vs `package`). But since, with ESM, you need to fully specify the path file, you have to do `import { x } from './foo/index.js'` inside your code, and could have a file named `./foo.js` next to the `./foo` folder.

```
foo.js
foo
  index.js
```

With the export map simplifying `foo.js` to foo, and with the "simplify `foo/index.js` to `foo` logic already present, this would clash. 

That said, I'm not sure if this would ever happen. And it would be highly confusing from a package consumer perspective to have a `foo.js` _and_ a `foo` import present, meaning different things.

ignatiusmb:
Oh, I totally forgot exports field will take higher priority for module resolutions rather than the actual relative paths 😄 

You're also totally right about the second point, this change would always overwrite the `foo/index.js` to `foo.js` if they both exists. Do we want then want `package` to also handle these kinds of overlaps and throw an error, simply prints a warning, or mention it in the docs? When do we draw the line for stuff that is out of `package` command scope? I'm leaning to just mentioning them in the docs for now.

dummdidumm (approved):
I'm good with these changes. The DX of importing these packages without `.js` surpasses the very small likelyhood of a clash.

ignatiusmb:
Do you think we're missing any odd characters in filenames with the current regex?

benmccann:
I think that for ESM that needing to import with the extension does not apply for things listed in `exports` in `package.json`

Rich-Harris:
Agree that the likelihood of collision is very small, but I do think we should error in that case — it's more likely to be an oversight than anything (an incomplete refactor or whatever) and an error would be preferable to an export no longer working

packages/kit/src/packaging/index.js
```diff
@@ -85,7 +88,15 @@ export async function make_package(config, cwd = process.cwd()) {
 		if (exports_filter(file)) {
 			const entry = `./${out_file.replace(/\\/g, '/')}`;
 			const key = entry.replace(/\/index\.js$|(\/[^/]+)\.js$/, '$1');
+
+			if (clashes[key]) {
```
ignatiusmb, line 92: I think we can probably reuse `generated` here instead?

ignatiusmb:
Makes sense that an incomplete refactor would produce this, also nice error message! Way better than writing it in the docs

ignatiusmb:
Welp, here we go again. The new tests do fail, but not for reasons I expected.

Rich-Harris:
I don't think we can reuse `generated` — I added `clashes` because otherwise the error message is incorrect in the case where the input file is TypeScript — you should get a message like

> Duplicate "./utils" export. Please remove or rename either $lib/utils/index.ts or $lib/utils.ts

but instead you get 

> Duplicate "./utils" export. Please remove or rename either $lib/utils/index.js or $lib/utils.js

because `generated` contains the output files rather than the input files

ignatiusmb:
Ah, you're right! I didn't take TS files into account, my bad.

PR #2351, opened 2021-09-02
[chore] export package.json from adapters

benmccann:
We always tell people they should export `package.json`, so we should probably follow our own advice :smile:


ignatiusmb (approved):

jthegedus:
What's the motivation for this? I apparently have never read any of the "We always tell people" conversations.

benmccann:
If you don't export `package.json` then tooling can't access it. There's some tooling like `rollup-plugin-svelte` that tries to access the `package.json` of every package (e.g. to determine if it's a Svelte package in that case) and you will get an error if it's not exported

Issue #2308, opened 2021-08-27
Change env var prefix


benmccann:
### Describe the problem

Vite will now let you change `VITE` env var prefix to be something else like `PUBLIC` or `CLIENT`

### Describe the proposed solution

https://github.com/vitejs/vite/pull/4676

### Importance

nice to have

JeanJPNM:
Do we have to do something? I mean, if it is on the vite config we just need to bump the version of vite right?

benmccann:
yeah. and I think we should change our default and document it

dreitzner:
Maybe `KIT`? Would it make sense to expose it in the config, because it would be a breaking change?

benmccann:
I sent a PR to accept both `VITE_` and `PUBLIC_` so that it won't break existing applications

Rich-Harris:
Isn't this just an extra thing people have to learn? If someone knows Vite (and to become expert in SvelteKit, they will eventually need to know their way around Vite, at least a bit) then they will know about `VITE_`, and if they see it without context then they'll likely understand that it's a Vite feature (and therefore that they can learn about it on https://vitejs.dev/).

By contrast I don't think I would immediately know what `PUBLIC_` was unless I happened to have read the relevant part of the docs (or rather the FAQs, per #2355). It feels like this could be left to user config, rather than increasing the surface area of SvelteKit-specific idioms that people need to learn.

benmccann:
It is an extra thing to learn to some extent, but if you don't know about `PUBLIC_` I don't think there's much harm that would come from it. You'd simply create all your environment variables with a `VITE_` prefix instead and you could carry on your marry way in bliss. Or if you read code in a project you weren't familiar with and you saw `PUBLIC_` you might not know that was needed to access it on the client, but by the same respect you also wouldn't know that `VITE_` was needed either unless you read the docs or had some prior Vite familiarity. The worst thing that would happen in that case is you might create one with a `VITE_` prefix when the project was using a `PUBLIC_` prefix and it'd be an extra comment in code review

Karlinator:
Is there a specific reason why you suggest `PUBLIC_`? It's a pretty well-established pattern generally to use the framework name as a prefix is it not?. `VITE_`, `REACT_APP_`, etc. My first thought would probably be to use `KIT_`, though I don't have any strong opinions on that.

I agree that, though the _reasons_ make sense, it has always been a little weird to put `VITE_` everywhere when the immediate framework is SvelteKit.

[...]

PR #2355, opened 2021-09-03
[feat] accept PUBLIC as prefix for environment variables

benmccann:
Closes https://github.com/sveltejs/kit/issues/2308

### DIFF
--- a/documentation/faq/60-env-vars.md
+++ b/documentation/faq/60-env-vars.md
@@ -2,7 +2,7 @@
 question: How do I use environment variables?
 ---
 
-Vite uses [dotenv](https://github.com/motdotla/dotenv) to load environment variables from a file named `.env` or similar. By default, Vite only exposes environment variables prefixed with `VITE_`. However, SvelteKit overrides Vite's default `envPrefix` setting to also allow `PUBLIC_` as a prefix. You would need to instantiate dotenv yourself if you want all environment variables exposed in `process.env['YOUR_ENV_VAR']`. We hope to see all environment variables exposed on the server-side [in the future](https://github.com/vitejs/vite/issues/3176).
+Vite uses [dotenv](https://github.com/motdotla/dotenv) to load environment variables from a file named `.env` or similar. SvelteKit allows you to use any environment variables prefixed with either `VITE_` or `PUBLIC_` because it adds `PUBLIC_` to [Vite's `envPrefix` setting](https://vitejs.dev/config/index.html#envprefix) by default. You would need to instantiate dotenv yourself if you want all environment variables exposed in `process.env['YOUR_ENV_VAR']`. We hope to see all environment variables exposed on the server-side [in the future](https://github.com/vitejs/vite/issues/3176).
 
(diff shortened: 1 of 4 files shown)


ignatiusmb (approved):
At least `PUBLIC_` makes it clear that it's public.

dummdidumm:
I think the changelog should make it more clear that this is a breaking change and how to fix it.

documentation/faq/60-env-vars.md
```diff
@@ -2,20 +2,20 @@
 question: How do I use environment variables?
 ---
 
-Vite uses [dotenv](https://github.com/motdotla/dotenv) to load environment variables from a file named `.env` or similar. Only environment variables prefixed with `VITE_` are exposed. You would need to instantiate dotenv yourself if you want all environment variables exposed in `process.env['YOUR_ENV_VAR']`. We hope to see all environment variables exposed on the server-side [in the future](https://github.com/vitejs/vite/issues/3176).
+Vite uses [dotenv](https://github.com/motdotla/dotenv) to load environment variables from a file named `.env` or similar. By default, Vite only exposes environment variables prefixed with `VITE_`. However, SvelteKit overrides Vite's default `envPrefix` setting to also allow `PUBLIC_` as a prefix. You would need to instantiate dotenv yourself if you want all environment variables exposed in `process.env['YOUR_ENV_VAR']`. We hope to see all environment variables exposed on the server-side [in the future](https://github.com/vitejs/vite/issues/3176).
```
ignatiusmb, line 5: @dummdidumm I don't think this is breaking, but maybe it can be phrased more friendlier here

```suggestion
Vite uses [dotenv](https://github.com/motdotla/dotenv) to load environment variables from a file named `.env` or similar. SvelteKit allows you to use any environment variables prefixed with either `VITE_` or `PUBLIC_`. You would need to instantiate dotenv yourself if you want all environment variables exposed in `process.env['YOUR_ENV_VAR']`. We hope to see all environment variables exposed on the server-side [in the future](https://github.com/vitejs/vite/issues/3176).
```

dummdidumm: Ah I misread the code - the allowed prefixes are now both vite and client, right?

ignatiusmb: Yeap

ignatiusmb: Wait no, it's vite and public

benmccann: I updated the text similar to your suggestion, but also mentioning `envPrefix` still

Correct that this allows both to avoid being breaking

antony:
It's not breaking in that we also still support VITE_. However it is breaking in that if I had an env var called PUBLIC_DATABASE_PASSWORD which was previously a private env var, it is now exposed. It's a super edge case but it could happen.

It doesn't feel like a patch version.

dummdidumm:
Everything's a patch version in `next` versions right now

antony:
Oh hah, yes it is :)

dreitzner (approved):
Looks great and can be extended with exposing it through the settings at a later point.

Rich-Harris:
Commented on https://github.com/sveltejs/kit/issues/2308#issuecomment-913291165 — tl;dr I feel like this maybe makes things more complicated?

PR #2356, opened 2021-09-03
[fix] load function should not leak props

### LINKED ISSUE #2133: Load function leaks props to other components
### Describe the bug

When returning an object with the `props` property from the load function in a page component, according to the docs they are available as props to the same page component

> If the `load` function returns a `props` object [sic!], the props will be passed to the component when it is rendered.

However currently, the props are also passed to all other page components including the error component (but not the layout component!).

Furthermore, this issue also occurs for the load function in the error component. The load function in the layout component however behaves as expected.

I noticed this when reading through the console logs, which warned that page components that didn't expect any props were passed props. Upon further investigation, I found they were passed the props that were returned in the load functions of other page components.

This isn't really a breaking issue, because for every component the props returned from its own load function take precedence. So in practice the issue can occur only if you forget to remove props (`export let data`) that you don't use, or like me follow to investigate console warnings.

### Reproduction

See https://github.com/vwkd/sveltekit-load-bug

This example has three different pages `/`, `/about`, and `/about/contact` as well as a layout and an error page. The layout component has a different colour than the page components for clarity. Click around the pages using the navigation menu.

The load function returning the props is in the index route. You can see that the props from the index's load function are passed to the about, the about/contact and the error page. They are however, correctly, not passed to the layout component.

Note, if you open the website on any other page than the index page, then the data will be undefined until you open the index page at least once, because the index page hasn't yet been prefetched.

https://user-images.githubusercontent.com/33468089/128720491-29562ed5-bee2-4c01-8a40-27b80023c755.mov

It's important to realise that there is nothing special about the index component. The load function could have been in any other of the components (about, about/contact or error). You can confirm this yourself by moving the `<script context="module">...</script>` code to another component. The `error` branch in the reproduction repo is such an example, where the load function is in the error component instead.

https://user-images.githubusercontent.com/33468089/128720748-1bc94ec4-97ab-4954-8a0c-f7c79acab550.mov

It gets interesting when you have several components with prop exporting load functions, because the other components then switch their props depending on the latest component that was loaded. The `double` branch in the reproduction repo is such an example, where in addition to the load function in the index component there is a load function in the about/contact component. You can see the about component switched depending on if the index or the about/component was the last visited.

https://user-images.githubusercontent.com/33468089/128720868-08120a0a-9d42-4025-a71d-80f59d4068d0.mov

Lastly, on the `layout` branch in the reproduction repo is an example of the load function in the layout component, which behaves correctly, and doesn't pass the props to other page components and error component.

https://user-images.githubusercontent.com/33468089/128720992-ffe74c59-4fe9-41df-b3f3-b999ef6920a7.mov

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 11.5
  Binaries:
    Node: 16.6.0 - ~/.nvm/versions/node/v16.6.0/bin/node
    npm: 7.19.1 - ~/.nvm/versions/node/v16.6.0/bin/npm
  Browsers:
    Chrome: 92.0.4515.131
    Firefox: 90.0.2
    Safari: 14.1.2
  npmPackages:
    @sveltejs/adapter-static: ^1.0.0-next.11 => 1.0.0-next.16 
    @sveltejs/kit: next => 1.0.0-next.142 
    svelte: ^3.34.0 => 3.42.1
```

### Severity

annoyance

### Additional Information

Not sure if this is actually a Svelte bug, or a SvelteKit bug.

rmunn:
Fixes #2133.

When navigating from a page with a `load()` function to a page without a `load()` function, props were not being reset, so if the second page had a prop with the same name as the prop returned by the first page's `load()` function, the second page would receive the same prop that the first page did.

### Tests

### Changesets

### DIFF
--- a/.changeset/polite-elephants-care.md
+++ b/.changeset/polite-elephants-care.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] load function should not leak props to other components
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -404,7 +404,7 @@ export class Renderer {
 
 		for (let i = 0; i < filtered.length; i += 1) {
 			const loaded = filtered[i].loaded;
-			if (loaded) result.props[`props_${i}`] = await loaded.props;
+			result.props[`props_${i}`] = loaded ? await loaded.props : null;
 		}
 
 		if (
--- a/packages/kit/test/apps/basics/src/routes/load/_tests.js
+++ b/packages/kit/test/apps/basics/src/routes/load/_tests.js
@@ -231,4 +231,12 @@ export default function (test, is_dev) {
 		assert.equal(await page.innerHTML('.parsed'), '{"oddly":{"formatted":"json"}}');
 		assert.equal(await page.innerHTML('.raw'), '{ "oddly" : { "formatted" : "json" } }');
 	});
+
+	test('does not leak props to other pages', '/load/props/about', async ({ page, clicknav }) => {
+		assert.equal(await page.textContent('p'), 'Data: undefined');
+		await clicknav('[href="/load/props/"]');
+		assert.equal(await page.textContent('p'), 'Data: Hello from Index!');
+		await clicknav('[href="/load/props/about"]');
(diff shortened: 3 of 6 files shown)


dummdidumm (approved):

benmccann:
thank you!!

PR #2358, opened 2021-09-03
[feat] package overrides

### LINKED ISSUE #2242: Change `"svelte"` path in `package.json` in root during dev and in package dir after `npm run package`
### Describe the problem

When developing a package I would like to have `"svelte": "src/lib/index.ts"` in my `package.json` (using it with `npm link`) but when I build I would like it changed to `"svelte": "index.js"` in `package` dir.

Is this already possible and I don't know?

I'm using `"@sveltejs/kit": "1.0.0-next.153"` with it skeleton project.

Transform that path in `package.json` during `npm run package`.

### Importance

would make my life easier

ignatiusmb:
Resolves #2242 

There's a lot of changes here, but it's mostly refactors and cleanups for reusability.


dummdidumm (commented):
The refactorings are good, but I'm not sure about the addition of this. What other use cases for this are there besides the super-duper-niche-case in the referenced issue? Maybe split the refactorings out into a separate PR to get these in first, and then discuss on the addition of the config.

ignatiusmb:
Good idea, I've opened up a separate PR for that. Other than the one in the issue, I can't think of any right now, tbh.

benmccann:
I think a more flexible solution might be to have a `function` which lets the user mutate the value we compute instead of a key-value map

ignatiusmb:
We're currently only modifying the `exports` field from package json, everything else are copied as-is. By using a function, do you mean exposing the entire `package.json` object for the user to modify?

benmccann:
I wasn't necessarily thinking that, but I think I like that solution even better :smile:

ignatiusmb:
Well, it was a combined efforts 😀

ignatiusmb:
But, as dummdidumm said, I'm not sure what's the use case other than the super-duper-niche referenced issue. I feel it's become a lot more flexible (still not sure about the name `override`) now that it's a function, but I can't think of an easier way for the user to handle cases like those.

packages/kit/src/packaging/test/fixtures/typescript/svelte.config.cjs
```diff
@@ -1,5 +1,10 @@
 const preprocess = require('svelte-preprocess');
 
 module.exports = {
-	preprocess: preprocess()
+	preprocess: preprocess(),
+	kit: {
+		package: {
+			override: (pkg) => ({ ...pkg, svelte: 'index.js' })
```
benmccann, line 7: I think `svelte` is being added automatically now, right? So this test seems a bit confusing that we're adding something that is normally already there (or maybe it's updating it?). Perhaps it'd be clearer to just add something `foo: 'bar'`?

benmccann:
This generally seems okay to me. I feel like there will be other use cases we haven't anticipated. But if you don't think it's necessary and prefer to close it that's fine too. What do you think @dummdidumm ?

dummdidumm:
I'd say it won't hurt to leave it in draft mode for a few more weeks/months to see if any requests come up that make this a needed feature. But I would definitely not merge it right now.

packages/kit/src/packaging/test/fixtures/typescript/svelte.config.cjs
ignatiusmb: Yes, this test is based on the issue, which needed to replace the `svelte` key (`src/lib/index.ts` -> `index.js`)

Rich-Harris:
It's been a few months, what's the status of this? Do we need it or should we close it?

ignatiusmb:
There hasn't been any activity around the issue itself as well, I guess we don't really need this (I surely don't) and also close the referenced issue as "won't do".

jacob-8:
I'm late to the party, but I have the exact same problem as #2242 in my monorepo. I would like to have:

- `"svelte": "src/lib/index.ts"` in my package A `package.json` so changes are reflected immediately in sibling packages of the monorepo that consume package A components
- However, when packaging and publishing to npm I would like it changed to `"svelte": "index.js"` in the `package` dir.

If `override: (pkg) => ({ ...pkg, svelte: 'index.js' })` is all the fix that's needed that would be great to have this feature. I think I'll manually have to adjust `index.js` to `src/lib/index.ts` for the moment when developing and just not commit that change as I publish the package via Github actions.

PR #2359, opened 2021-09-03
[feat] provide the platform request object

benmccann:
There are some things that are not possible to do today that this will enable. E.g. if you [use `adapter-node`'s Kit middleware with a custom server](https://github.com/sveltejs/kit/tree/master/packages/adapter-node#middleware) and add some other middleware that populates a custom property onto the request then you have no way to access that today.

### DIFF
--- a/.changeset/tricky-timers-boil.md
+++ b/.changeset/tricky-timers-boil.md
@@ -1,0 +1,8 @@
+---
+'@sveltejs/adapter-netlify': patch
+'@sveltejs/adapter-node': patch
+'@sveltejs/adapter-vercel': patch
+'@sveltejs/kit': patch
+---
+
+[feat] provide the platform request object
--- a/documentation/docs/04-hooks.md
+++ b/documentation/docs/04-hooks.md
@@ -24,24 +24,25 @@ type ResponseHeaders = Record<string, string | string[]>;
 type RequestHeaders = Record<string, string>;
 
 export type RawBody = null | Uint8Array;
-export interface IncomingRequest {
+export interface IncomingRequest<PlatformRequest = any> {
 	method: string;
 	host: string;
 	path: string;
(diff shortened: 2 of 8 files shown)


benbender:
Halleluja! \o/ :)

Rich-Harris:
I think we need to tread very carefully here, and consider what this looks like in non-Node-like environments, and how it affects other things. For example https://kit.svelte.dev/faq#how-do-i-use-x-with-sveltekit-how-do-i-use-middleware makes me nervous (the fact that you need to specify middleware both in your custom server and in your config); to me this feels like a symptom of a design that isn't fully baked. 

At the very least, `platformReq` probably isn't the right name given that Lambdas and Cloudflare Workers etc refer to 'events' rather than 'requests'.

benmccann:
> For example https://kit.svelte.dev/faq#how-do-i-use-x-with-sveltekit-how-do-i-use-middleware makes me nervous (the fact that you need to specify middleware both in your custom server and in your config); to me this feels like a symptom of a design that isn't fully baked.

Yeah, I agree with this. This is something I plan to cleanup shortly. However, I've been blocked on that by getting a PR into Vite to allow for something more sensible (https://github.com/vitejs/vite/pull/4640). In the meantime I figured I'd at least document what's possible at the moment to unblock folks who need the functionality even if it's not ideal.

> At the very least, platformReq probably isn't the right name given that Lambdas and Cloudflare Workers etc refer to 'events' rather than 'requests'.

Any other suggestions? Even if those platforms use the word events to refer to requests, they're still requests. I think the folks using those platforms are unlikely to be confused and they're probably pretty unlikely to need to use it anyway

benmccann:
Talked about this in today's maintainer's meeting. Rich suggested we provide a way to put stuff from the original request into `locals` instead

PR #2360, opened 2021-09-03
[chore] deep_merge refactors and clean up

ignatiusmb:
Extracted the refactors from #2358


packages/kit/src/core/build/index.js
```diff
@@ -1,14 +1,18 @@
...
-import { create_app } from '../../core/create_app/index.js';
-import vite from 'vite';
+
 import { svelte } from '@sveltejs/vite-plugin-svelte';
 import glob from 'tiny-glob/sync.js';
+import vite from 'vite';
+
+import { rimraf } from '../../utils/filesystem.js';
+import { deep_merge } from '../../utils/object.js';
+
```
dummdidumm, line 10: Is there a reason for the empty lines?

packages/kit/src/core/dev/index.js
```diff
@@ -1,22 +1,26 @@
...
+import { svelte } from '@sveltejs/vite-plugin-svelte';
 import amp_validator from 'amphtml-validator';
-import vite from 'vite';
+import CheapWatch from 'cheap-watch';
 import colors from 'kleur';
-import create_manifest_data from '../../core/create_manifest_data/index.js';
-import { create_app } from '../../core/create_app/index.js';
-import { rimraf } from '../../utils/filesystem.js';
+import vite from 'vite';
+
```
dummdidumm, line 11: Same here (empty lines)

packages/kit/src/packaging/index.js
```diff
@@ -1,8 +1,10 @@
 import * as fs from 'fs';
 import * as path from 'path';
-import micromatch from 'micromatch';
 import { createRequire } from 'module';
+
+import micromatch from 'micromatch';
 import { preprocess } from 'svelte/compiler';
+
```
dummdidumm, line 7: And here

packages/kit/types/helper.d.ts
```diff
@@ -8,6 +8,17 @@ interface ReadOnlyFormData {
 	[Symbol.iterator](): Generator<[string, string], void>;
 }
 
+type ToJSON = { toJSON(...args: any[]): JSONValue };
+type JSONValue = Exclude<JSONString, ToJSON>;
+export type JSONString =
```
dummdidumm, line 13: No sure about the wording. Json string sounds like stringified json to me, but it's JSON itself. So that may be the better name, "JSON"

dummdidumm (commented):
Nice one! Couple of nits, that's all.

packages/kit/src/core/build/index.js
ignatiusmb: Just like needing to group codes into related sectioned blocks, I find myself having a hard time reading and separating what imports a certain file relies on when it's jumbled up together, and giving these imports some "breathing space" makes the separation clearer between the used member syntax, especially when it's over 5-6 imports.

packages/kit/src/core/dev/index.js
ignatiusmb: Same reason as https://github.com/sveltejs/kit/pull/2360#discussion_r702237586

packages/kit/src/packaging/index.js
ignatiusmb: This one, I think I forgot to condense them from the previous commits, but might as well 😄

packages/kit/types/helper.d.ts
ignatiusmb: I'm glad you commented on this, we've had a couple of discussion for this too on #2170 and ended up with `JSONResponse` then because it's used as a response, but since we'll want to reuse this somewhere else, "Response" doesn't quite cover it anymore.

"JSON" sounds perfect, but I'm not sure it it will collide with the built-in JSON object. I chose `JSONString` to indicate that anything defined here will be passed through `JSON.stringify`, but yeah it does sounds like the end result of stringified JSON.

dummdidumm: Mhm yeah that's true, we don't want the name clash. JSONObject?

ignatiusmb: Object kind of restrict it since it can also be some primitives like string or number. `JSONable`? (Surprisingly, my spellchecker doesn't complain about this, which it does for `StringifiableJSON`)

dummdidumm: For me a JSON object is a JavaScript object which is valid JSON. Don't know if that's only me that has this semantic connection thoughö

ignatiusmb: That would be `JSONValue` for me, which is already defined above. Naming is hard.

ignatiusmb:
This might conflict with #2373

Issue #1198, opened 2021-04-24
Make server version of fetch() in load() copy over 'set-cookie' to the page response


hgl:
**Is your feature request related to a problem? Please describe.**
Currently, the isomorphism of the `load()` function breaks when a `fetch` response, from an internal url, contains `set-cookie`. If it's called by the client, the cookie is set in the browser, if it's called by the server, the cookie is dropped.

One use case is to create a session when a guest visits a page that should only be visible to a registered user. This session should record what page to return the guest to, once she registers.

Some people might suggest passing the return url as a query to the register page, but that means we now have to sanitize the url, prevent users from being redirected to external sites, and even if we do this, we still can't prevent users from being redirected to other irrelevant internal pages (e.g., people register via an url shared online that contains the query to redirect them to an irrelevant page)

**Describe the solution you'd like**
The server's version of `fetch` should copy over cookie headers from the fetch response, from an internal url, to the page response, (including a redirect response like 302)

**Describe alternatives you've considered**
We can redirect guests to an endpoint that creates the session and return the cookie header, but that forces a page refresh, breaking the SPA experience that svelte kit tries so hard to deliver.

**How important is this feature to you?**
Very, and I think this recording return url use case should be really common.

**Additional context**
None.

hgl:
Hmm, just realize this requires merging 'set-cookie' if there are multiple `fetch()`s, sounds messy.

Feel free to close this issue if this doesn't seem worth it.

GrygrFlzr:
Simple experimentation with default configurations shows that multiple same-origin cookies make it to the client even when SSR'd, which makes sense.

This situation you're describing seems to only occur with either:
- Hydration disabled (`export const hydrate = false;`) in which case the client is not doing any fetches and it is expected that cookies are not passed through.
- Cross-origin cookies - In the case of a `thirdparty.com` API, it is impossible for `mydomain.com` to set a `thirdparty.com` cookie. This is not unique to SvelteKit as it is a behavior enforced by all modern browsers.
    Setting a `mydomain.com` cookie with the same contents as the `thirdparty.com` cookie does not solve the issue as it will be sent to the wrong domain on the client side. You need to proxy the requests through your own API and store a same-origin cookie containing the information that needs to be sent to the third party API.

If neither of these describe your situation, please elaborate.

hgl:
Hmm, here are my findings after experimenting with a small test case:

`routes/session.ts`
```
export async function get({ params }) {
	return {
		status: 200,
[...]

PR #2362, opened 2021-09-03
[fix] pass along set-cookie headers during SSR

benmccann:
fixes https://github.com/sveltejs/kit/issues/1198

### DIFF
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -49,6 +49,11 @@ export async function load_node({
 	 */
 	const fetched = [];
 
+	/**
+	 * @type {string[]}
+	 */
+	let set_cookie_headers = [];
+
 	let loaded;
 
 	const page_proxy = new Proxy(page, {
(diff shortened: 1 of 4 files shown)


benwoodward:
Thanks Ben!

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -219,7 +224,11 @@ export async function load_node({
 								/** @type {import('types/helper').ResponseHeaders} */
 								const headers = {};
 								for (const [key, value] of response.headers) {
-									if (key !== 'etag' && key !== 'set-cookie') headers[key] = value;
+									if (key === 'set-cookie') {
+										set_cookie_headers = set_cookie_headers.concat(value);
```
dishuostec, line 228: The cookie value may included multiple k/v entries with comma joined see https://datatracker.ietf.org/doc/html/rfc2616#section-4.2. We should convert them to array of string. Here is a implementation: https://github.com/nfriedly/set-cookie-parser/blob/9cb7f099b9328bc6ab7b7c1eab31d8cfb41b2e44/lib/set-cookie.js#L126

Issue #2361, opened 2021-09-03
`query` accessible from page store during prerendering, but shouldn't be


ssssota:
### Describe the bug

Building with `sveltejs/adapter-static` does not give me the correct query.

`svelte-kit dev` and `svelte-kit preview` work fine.

However, if I build with `@sveltejs/adapter-static`, I don't get the correct query parameters.
Especially after a reload.

### Reproduction

https://github.com/ssssota/sveltekit-static-query

### System Info

```shell
❯ npx envinfo --system --binaries --browsers --npmPackages "{svelte,@sveltejs/*,vite}"

  System:
    OS: Windows 10 10.0.19042
    CPU: (12) x64 AMD Ryzen 5 2600 Six-Core Processor
    Memory: 8.18 GB / 15.93 GB
  Binaries:
    Node: 14.17.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.11 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 7.20.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 92.0.4515.159
    Edge: Spartan (44.19041.1023.0), Chromium (92.0.902.84)
    Internet Explorer: 11.0.19041.906
  npmPackages:
    @sveltejs/adapter-static: ^1.0.0-next.18 => 1.0.0-next.18
    @sveltejs/kit: next => 1.0.0-next.162
    svelte: ^3.42.4 => 3.42.4
```

### Severity

serious, but I can work around it

### Additional Information

I think the hydrate part of the initialization is suspicious.

*[/build/index.html](https://github.com/ssssota/sveltekit-static-query/blob/gh-pages/index.html#L18-L41)*
```js
import { start } from "/_app/start-19affe50.js";
start({
	target: document.querySelector("#svelte"),
	paths: {"base":"","assets":""},
	session: {},
	host: location.host,
	route: true,
	spa: false,
	trailing_slash: "never",
	hydrate: {
		status: 200,
		error: null,
		nodes: [
			import("/_app/pages/__layout.svelte-e8a46f20.js"),
			import("/_app/pages/index.svelte-9233398d.js")
		],
		page: {
			host: location.host, // TODO this is redundant
			path: "/",
			query: new URLSearchParams(""), /* 🤔 */
			params: {}
		}
	}
});
```

ssssota:
#2316 may have the same problem.

ssssota:
#669 may be the same problem.
[...]

PR #2363, opened 2021-09-03
[fix] query access check should apply to page store as well

benmccann:
Closes https://github.com/sveltejs/kit/issues/2361

### DIFF
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -56,19 +56,10 @@ export async function load_node({
 
 	let loaded;
 
-	const page_proxy = new Proxy(page, {
-		get: (target, prop, receiver) => {
-			if (prop === 'query' && prerender_enabled) {
-				throw new Error('Cannot access query on a page with prerendering enabled');
-			}
-			return Reflect.get(target, prop, receiver);
-		}
-	});
-
 	if (module.load) {
 		/** @type {import('types/page').LoadInput | import('types/page').ErrorLoadInput} */
 		const load_input = {
-			page: page_proxy,
+			page: prerender_enabled ? create_page_proxy(page) : page,
 			get session() {
 				uses_credentials = true;
 				return $session;
@@ -376,4 +367,19 @@ export function resolve(base, path) {
 	const prefix = (path_match && path_match[0]) || (base_match && base_match[0]) || '';
 
 	return `${prefix}${baseparts.join('/')}`;
+}
+
+/**
+ * @param {import('types/page').Page} page
+ * @returns
+ */
+export function create_page_proxy(page) {
+	return new Proxy(page, {
+		get: (target, prop, receiver) => {
+			if (prop === 'query') {
+				throw new Error('Cannot access query on a page with prerendering enabled');
+			}
+			return Reflect.get(target, prop, receiver);
+		}
+	});
 }
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -1,6 +1,7 @@
 import devalue from 'devalue';
 import { writable } from 'svelte/store';
 import { hash } from '../../hash.js';
+import { create_page_proxy } from './load_node.js';
 
 const s = JSON.stringify;
 
@@ -12,6 +13,7 @@ const s = JSON.stringify;
  *   options: import('types/internal').SSRRenderOptions;
  *   $session: any;
  *   page_config: { hydrate: boolean, router: boolean, ssr: boolean };
+ *   prerender_enabled: boolean,
  *   status: number;
  *   error?: Error,
  *   page?: import('types/page').Page
@@ -22,6 +24,7 @@ export async function render_response({
 	options,
 	$session,
 	page_config,
+	prerender_enabled,
 	status,
 	error,
 	page
@@ -42,6 +45,7 @@ export async function render_response({
 		error.stack = options.get_stack(error);
 	}
 
+	const query = page ? s(page.query.toString()) : '';
 	if (page_config.ssr) {
 		branch.forEach(({ node, loaded, fetched, uses_credentials }) => {
 			if (node.css) node.css.forEach((url) => css.add(url));
@@ -65,7 +69,7 @@ export async function render_response({
 				navigating: writable(null),
 				session
 			},
-			page,
+			page: prerender_enabled && page ? create_page_proxy(page) : page,
 			components: branch.map(({ node }) => node.module.default)
 		};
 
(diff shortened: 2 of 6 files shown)


packages/kit/src/runtime/server/page/render.js
```diff
@@ -136,7 +136,7 @@ export async function render_response({
 					page: {
 						host: ${page && page.host ? s(page.host) : 'location.host'}, // TODO this is redundant
 						path: ${s(page && page.path)},
-						query: new URLSearchParams(${page ? s(page.query.toString()) : ''}),
+						query: new URLSearchParams(document.location.search),
```
ignatiusmb, line 139: I think I read somewhere from Rich that we don't really want this, I forgot where I saw it but it's something about different results in server and client. Aside from that, can we also use `document` here?

benmccann: It was in https://github.com/sveltejs/kit/issues/669 or one of the linked issues. The issue was that folks wanted to set it in this way so that they could use it on pre-rendered pages, but then it'd only work on the client and the pre-rendered pages wouldn't match. However, it's okay to do as part of this PR because we're explicitly disallowing it from being used during pre-rendering so no one can get unexpected results.

I had to change it here because if I try to access `page.query` as the code was doing before it'd throw the validation error, which I could have possibly gotten around in other ways, but at the cost of messier code

Rich-Harris: This would in fact break an app I recently worked on, in which we were using SvelteKit to generate HTML embeds and using the page query to configure them. Those embeds were then placed in a context where `location.search` would have equalled something incorrect.

While it's something of a niche situation, I'd prefer that we keep the ability to generate HTML that behaves in a predictable way. Couldn't we simply address this TODO instead? https://github.com/sveltejs/kit/blob/aad27f4ed3a2a57f68c6cb55076b9d3132bcbaac/packages/kit/src/core/adapt/prerender.js#L268-L270

benmccann: I'm not quite understanding how `page.query` and `location.search` would be different or exactly what you're doing, but I've reverted the change to this line, so hopefully that addresses your concern?

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -51,19 +51,10 @@ export async function load_node({
...
-			}
-			return Reflect.get(target, prop, receiver);
-		}
-	});
-
 	if (module.load) {
 		/** @type {import('types/page').LoadInput | import('types/page').ErrorLoadInput} */
 		const load_input = {
-			page: page_proxy,
+			page: create_page_proxy(page, prerender_enabled),
```
Rich-Harris, line 57: let's change this — we don't want to incur the overhead of creating the proxy unnecessarily

```suggestion
			page: prerender_enabled ? create_page_proxy(page) : page,
```

packages/kit/src/runtime/server/page/render.js
```diff
@@ -65,7 +69,7 @@ export async function render_response({
 				navigating: writable(null),
 				session
 			},
-			page,
+			page: page ? create_page_proxy(page, prerender_enabled) : undefined,
```
Rich-Harris, line 72: ditto

```suggestion
			page: prerender_enabled ? create_page_proxy(page) : page,
```

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -367,3 +358,19 @@ export function resolve(base, path) {
...
+export function create_page_proxy(page, prerender_enabled) {
+	return new Proxy(page, {
+		get: (target, prop, receiver) => {
+			if (prop === 'query' && prerender_enabled) {
+				throw new Error('Cannot access query on a page with prerendering enabled');
+			}
+			return Reflect.get(target, prop, receiver);
+		}
+	});
+}
```
Rich-Harris, line 376:
```suggestion

/**
 * @param {import('types/page').Page} page
 * @returns
 */
export function create_page_proxy(page) {
	return page && new Proxy(page, {
		get: (target, prop, receiver) => {
			if (prop === 'query') {
				throw new Error('Cannot access query on a page with prerendering enabled');
			}
			return Reflect.get(target, prop, receiver);
		}
	});
}
```

Rich-Harris:
Is there a concern that we'd get a confusing error in edge cases like this?

```js
console.log({ ...page });
```

```js
JSON.stringify(page);
```

Also, what about cases where `query` is accessed in `onMount` or similar (i.e. it's not encountered during SSR, but will still be incorrect)? Part of me feels like we'd be better off going the simpler route and just warning (or erroring) when the crawler finds a URL with a query string, rather than when the query is read.

benmccann:
I'd be okay changing it from an error to a warning. I think warning on pages where there's a query string might cause a number of false positive warnings though. It's not invalid to have a query string. E.g. you might want to use the query string only in `onMount` via `location.search`. The thing we want to prevent is people from relying on it during prerendering which I think this more precisely targets

Rich-Harris:
The wrinkle is that these would have different behaviour if they ran on the first page the user visits...

```js

// idiomatic
onMount(await () => {
  const res = await fetch(`/data/${$page.query.get('slug')}.json`);
  // ...
});
```

```js

// non-idiomatic
onMount(await () => {
  const query = new URLSearchParams(location.search);
  const res = await fetch(`/data/${query.get('slug')}.json`);
  // ...
});
```

...and the non-idiomatic code would be the one with the expected behaviour.

If we _did_ use `query: new URLSearchParams(location.search)` to initialise the page store (but only when prerendering), then people's expectations wouldn't be violated in that regard. That would at least mean that behaviour would be consistent between first load and subsequent navigations if `$page.query` was read in `onMount`. But if the error were downgraded to a warning, that could cause the page to look different before and after hydration if it was read in `load` or init.

So the question becomes which is the lesser evil:

|                  | consistently empty query                                                         | query populated by location.search                                         |
|------------------|----------------------------------------------------------------------------------|----------------------------------------------------------------------------|
| error on read            | 1️⃣ confusing error in edge case where `page.query` is accessed accidentally         | 1️⃣ confusing error in edge case where  `page.query`  is accessed accidentally |
| warning on read  | 2️⃣ `$page.query`  empty after hydration, forcing use of non-idiomatic code          | 3️⃣ incorrect hydration, if `query` is read in load or init                    |
| warning on crawl | 4️⃣ false positives, if author uses non-idiomatic code to read query after hydration | 3️⃣ incorrect hydration, if `query` is read in load or init                    |

benmccann:
One thing I worry about with a warning is that it'd be easy to miss. Vite spit out two screens full of build logs when a user hit this in https://github.com/sveltejs/kit/issues/2444 and I think it'd be easy for one line to get lost amongst all the others. We'd probably want to make sure that it's yellow as a warning so that it would stand out. But then if we make it stand out the false positives are even more annoying - especially since there wouldn't really be a way to silence it

I'm not quite sure why the first row in the table says "confusing error". I'd hope that we could make the error message not confusing. I understand that the user might expect they could access `query`, but if the user has an incorrect mental model of how things work they could be surprised in any of the six boxes. I think the error is the least surprising in this regard though because you will be sure to see it, get the correct understanding of how things work, and move on with life. Without the error, I worry it'd be very easy to miss it and be confused for a much longer period of time

Rich-Harris:
Closing this in favour of #3314, now that we're passing the full `URL` object around

PR #2373, opened 2021-09-05
Dry out config

Rich-Harris:
Simplifies the config validation without introducing new dependencies (follow-up to https://github.com/sveltejs/kit/pull/2366#issuecomment-913187660)

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/config/index.js
+++ b/packages/kit/src/core/config/index.js
@@ -33,7 +33,23 @@ export async function load_config({ cwd = process.cwd() } = {}) {
 		? config_file_esm
 		: path.join(cwd, 'svelte.config.cjs');
 	const config = await import(url.pathToFileURL(config_file).href);
-	const validated = validate_config(config.default);
+
+	const type = typeof config;
+
+	if (type === 'undefined') {
+		throw new Error(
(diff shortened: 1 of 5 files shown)


bluwy:
Some things I found while refactoring in #2366:

1. `target`, `adapter`, `host` and `hostHeader` are nullable, but not reflected in `Config` and `ValidatedConfig` types.
  https://github.com/sveltejs/kit/blob/90874c74cf981ba730ca7f0f0164c7bbaf0f2ac6/packages/kit/types/config.d.ts#L32

2. Maybe we want to validate/coerce this together as well:
  https://github.com/sveltejs/kit/blob/90874c74cf981ba730ca7f0f0164c7bbaf0f2ac6/packages/kit/src/core/config/index.js#L39-L46

Rich-Harris:
Ah yeah, that's mildly annoying about the types, though if I amend `Config` to have `string` | `null` etc it causes more problems, so I'm okay glossing over that for the time being.

I considered making the homegrown validation logic accept options so that we could do the resolution in the validator, but when I implemented it the extra complexity outweighed those six extra lines, so I made the pragmatic choice. Can always revisit that in future if we end up with more options that need to be amended post-validation to account for things like `cwd`.

PR #2378, opened 2021-09-06
[fix] error components render with correct layout client-side as well as server-side

### LINKED ISSUE #2372: Seeing weird behavior where an error is sometimes shown in __layout slot, and sometimes not
### Describe the bug

It seems that error handling is quite different when it's done from the browser vs from SSR. To demonstrate I have two screenshots, both showing exactly the same non-existing webpage, the URL is 100% identical.

In this first one I started from an existing character page and clicked on "Fake link" on the left side, which takes me to the non-existing page, using the router in the browser:

Screen Shot 2021-09-05 at 18 41 35

It's very weird how the error is showing the _entire website_ within the website in the `slot` of `__layout.svelte`!

The second one is the result of reloading that non-existing page, or going directly to it (not via the router):

Screen Shot 2021-09-05 at 18 41 50

What I would expect to happen is that errors would always show in the same consistent manner, both from SSR and from the router (client side).

### Reproduction

I have a single top-level `__error.svelte` page that renders the error with the navigation bar and the footer and all that, basically what you see in that second screenshot.

I also have a bunch of nested `__layout.svelte` components. In this particular example, there is a `campaigns/[campaignId]/characters/__layout.svelte` file, and a `campaigns/[campaignId]/characters/[characterId].svelte` page. In the page I am simply returning `{ status: 404, error: "Character not found" }` from the `load` function if the character isn't found. 

The `__layout.svelte` shows the list of characters on the left side, and has a `<slot />` on the right side to show character details. When the error is triggered from the client side router, the error page (with the entire website around it!) is shown _inside of that slot_, but when it's triggered from SSR, it only shows the error page.

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.0
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 128.39 MB / 32.00 GB
  Binaries:
    Node: 16.6.2 - /usr/local/bin/node
    npm: 7.20.3 - /usr/local/bin/npm
  Browsers:
    Safari: 15.0
  npmPackages:
    @sveltejs/kit: ^1.0.0-next.145 => 1.0.0-next.145 
    svelte: ^3.34.0 => 3.42.1
```

### Severity

blocking an upgrade

### Additional Information

I'm not able to update past version 145 at the moment because of https://github.com/sveltejs/kit/issues/2371. I will retest with the latest SvelteKit when I can.

rmunn:
Fixes #2372.

If a deeply-nested page produced an error in its `load()` function, and that page had multiple layout components but only a top-level error component, the error component rendered differently in server-side and client-side rendering. Server-side rendering correctly chose the top-level layout component to render a top-level error component, but client-side rendering was incorrectly choosing the layout of the deeply-nested page that caused the error, instead of the layout of the folder the error component belonged to.

With this PR, client-side rendering now uses the same approach as server-side rendering to choose layouts for error components.

### Tests

### Changesets

### DIFF
--- a/.changeset/blue-teachers-allow.md
+++ b/.changeset/blue-teachers-allow.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] error components render with correct layout client-side as well as server-side
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -566,7 +566,7 @@ export class Renderer {
 		const page = { host: this.host, path, query, params };
 
 		/** @type {Array<import('./types').BranchNode | undefined>} */
-		const branch = [];
+		let branch = [];
 
 		/** @type {Record<string, any>} */
 		let context = {};
@@ -665,7 +665,7 @@ export class Renderer {
 								continue;
 							}
 
-							branch.push(error_loaded);
+							branch = branch.slice(0, j + 1).concat(error_loaded);
 							break load;
 						} catch (e) {
 							continue;
--- a/packages/kit/test/apps/basics/src/routes/nested-layout/__layout.svelte
+++ b/packages/kit/test/apps/basics/src/routes/nested-layout/__layout.svelte
@@ -1,2 +1,2 @@
 <slot></slot>
-<p>This is a nested layout component</p>
+<p id="nested">This is a nested layout component</p>
--- a/packages/kit/test/apps/basics/src/routes/nested-layout/_tests.js
+++ b/packages/kit/test/apps/basics/src/routes/nested-layout/_tests.js
@@ -7,4 +7,29 @@ export default function (test) {
 		assert.equal(await page.textContent('p'), 'This is a nested layout component');
 		assert.equal(await page.textContent('h1'), 'Hello from inside the nested layout component');
 	});
+
+	test('renders errors in the right layout', '/nested-layout/error', async ({ page }) => {
+		assert.equal(await page.textContent('footer'), 'Custom layout');
+		assert.ok(await page.evaluate(() => !document.querySelector('p#nested')));
+		assert.equal(
+			await page.textContent('#message'),
+			'This is your custom error page saying: "Error"'
+		);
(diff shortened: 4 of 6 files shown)


Rich-Harris:
thank you!

kevinrenskers:
Thanks for the quick fix, but it seems like I am now getting the error page every time I use the client side router, when I also define a lower-lever `__error.svelte` component.

My file structure, you can see that I have a toplevel `__error.svelte` and one inside of the `characters` route:

Screen Shot 2021-09-06 at 19 05 52

Now, when I directly open http://localhost:4000/campaigns/1/characters/1, I see the page just fine, as expected:

Screen Shot 2021-09-06 at 19 07 01

But if I go to that exact same page using the list on the left, I get the error page, where the error and status are both undefined:

Screen Shot 2021-09-06 at 19 07 53

(As an additional question: why is `__error.svelte` not rendered inside of the `characters/__layout.svelte` file's `slot`, but instead using the `__layout.svelte` _one level above it_?)

kevinrenskers:
If I remove `characters/__error.svelte`, everything works fine, but I want errors on these pages to be shown nested inside of the overall layout of the site, with the character list on the left. I am currently not able to do that without just copying everything from `characters/__layout.svelte` to `characters/__error.svelte` 🤔

kevinrenskers:
Okay I found a bit more (hopefully useful) info.

In my `characters/[characterId].svelte` page I have this `load` function:

```
<script context="module">
  export async function load({ page, context }) {
    const character = context.fetchedCharacters[page.params.characterId];
    if (!character) {
      return {
        status: 404,
        error: "Character not found"
      }
    }

    return { props: context };
  }
</script>
```

This works fine, it's using the context which is given by `characters/__layout.svelte`'s `load` function.

However, as soon as I introduce that `characters/__error.svelte` component, `context` is empty on client side route changes. Removing `characters/__error.svelte` makes it work again, even if no error is otherwise triggered. Very very strange.

kevinrenskers:
Turns out this behavior was already there before this PR got merged, I'll create a new ticket.

Issue #2384, opened 2021-09-07
Missing "." export in "@sveltejs/kit" package


jacobbogers:
### Describe the bug

When issue `npm run build`  get this error

### Reproduction

`npm run build` 

My svelte.config.js

```javascript
import preprocess from 'svelte-preprocess';
const config = {

	kit: {
		amp: false,
		appDir: '_app',
		files: {
			assets: 'static',
			//hooks: 'src/hooks',
			lib: 'src/lib',
			routes: 'src/routes',
			serviceWorker: 'src/service-worker.ts',
			template: 'src/app.html'
		},
		floc: false,
		hydrate: true,
		package: {
		 	files: {
		 		exclude: ['**/*.test.ts']
		 	}
		},
		prerender: {
			crawl: true,
			enabled: true,
			onError: 'continue',
			pages: ['*']
		},
		router: true,
		ssr: true,
		target: '#svelte',
		trailingSlash: 'never'
	},
	preprocess: preprocess({
		replace: [['process.env.NODE_ENV', JSON.stringify(process.env.NODE_ENV)]],
	}),
};

export default config;
```

My package.json

```json
{
  "name": "~TODO~",
  "version": "0.0.1",
  "scripts": {
    "dev": "svelte-kit dev",
    "build": "svelte-kit build",
    "preview": "svelte-kit preview",
    "check": "svelte-check --tsconfig ./tsconfig.json",
    "check:watch": "svelte-check --tsconfig ./tsconfig.json --watch",
    "lint": "prettier --check --plugin-search-dir=. . && eslint --ignore-path .gitignore .",
    "format": "prettier --write --plugin-search-dir=. .",
    "test": "jest"
  },
  "//": "force install svelte-jester on version 1.8.2, 2.1.1 is broken and being fixed",
  "devDependencies": {
    "@sveltejs/kit": "next",
    "@testing-library/jest-dom": "^5.14.1",
    "@testing-library/svelte": "^3.0.3",
    "@types/jest": "^27.0.1",
    "@typescript-eslint/eslint-plugin": "^4.19.0",
    "@typescript-eslint/parser": "^4.19.0",
    "eslint": "^7.22.0",
    "eslint-config-prettier": "^8.1.0",
    "eslint-plugin-svelte3": "^3.2.0",
    "jest": "^27.1.0",
    "jest-localstorage-mock": "^2.4.17",
    "prettier": "~2.2.1",
    "prettier-plugin-svelte": "^2.2.0",
[...]

PR #2387, opened 2021-09-08
[fix] simple dot exports for adapters

ignatiusmb:
#2384

### DIFF
--- a/.changeset/fuzzy-bottles-bake.md
+++ b/.changeset/fuzzy-bottles-bake.md
@@ -1,0 +1,9 @@
+---
+'@sveltejs/adapter-cloudflare-workers': patch
+'@sveltejs/adapter-netlify': patch
+'@sveltejs/adapter-node': patch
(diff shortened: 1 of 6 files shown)


benmccann:
I don't know that this is proper and that the old version wasn't. Is there something in the Node docs showing what we have in `master` is wrong? I think this is more likely due to a Vite bug. Anyway, they have a PR to revert the log message (https://github.com/vitejs/vite/pull/4873), so let's see what happens there before changing anything on our side

ignatiusmb:
Sorry, "proper" was not the right choice of words. The current is one in master is also correct and can even support multiple types of import styles. Though, I found this silences the warning on the adapters side and we won't be using the extra features it gives anyway since we're only using/encouraging ES modules import style.

benmccann:
Thanks for taking a look at quieting the output. They've checked in the change to silence the message in Vite, so let's just go with that

PR #2389, opened 2021-09-08
[fix] deeply-nested error components render with correct layout

### LINKED ISSUE #2383: __error.svelte renders in the wrong layout
### Describe the bug

Errors can be rendered using the wrong layout. Instead of using `__layout.svelte` on the same level as `__error.svelte`, it sometimes uses a layout of a higher level.

### Reproduction

I have a repo on Github with a minimal example: https://github.com/kevinrenskers/error-repro.

Start the dev server and go to http://localhost:3000/campaigns/1/characters/1:

Screen Shot 2021-09-07 at 19 44 46

Click on "Broken link" in the list on the left. Now an error page shows up, but the error is not shown next to the list, as I would expect:

Screen Shot 2021-09-07 at 19 45 07

It definitely has something to do with the `__layout.svelte` hierarchy, because as soon as you delete the one in `[campaignId]`, the error shows as expected:

Screen Shot 2021-09-07 at 19 47 06

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.0
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 8.23 GB / 32.00 GB
  Binaries:
    Node: 16.6.2 - /usr/local/bin/node
    npm: 7.20.3 - /usr/local/bin/npm
  Browsers:
    Safari: 15.0
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.164 
    svelte: ^3.34.0 => 3.42.4
```

### Severity

### Additional Information

_No response_

rmunn:
Under some circumstances, a deeply-nested error component could render with a layout from a parent folder instead of the layout from the error component's own folder. This commit fixes that bug, so that error components are correctly matched up with layout components in the same folder even when deeply nested.

Fixes #2383.

### Tests

### Changesets

### DIFF
--- a/.changeset/empty-poets-check.md
+++ b/.changeset/empty-poets-check.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] deeply-nested error components render with correct layout
--- a/packages/kit/src/core/create_manifest_data/index.js
+++ b/packages/kit/src/core/create_manifest_data/index.js
@@ -220,23 +220,24 @@ export default function create_manifest_data({ config, output, cwd = process.cwd
 				components.push(item.file);
 
 				const concatenated = layout_stack.concat(item.file);
+				const errors = error_stack.slice();
 
 				const pattern = get_pattern(segments, true);
 
 				let i = concatenated.length;
 				while (i--) {
-					if (!error_stack[i] && !concatenated[i]) {
-						error_stack.splice(i, 1);
+					if (!errors[i] && !concatenated[i]) {
+						errors.splice(i, 1);
 						concatenated.splice(i, 1);
 					}
 				}
 
-				i = error_stack.length;
+				i = errors.length;
 				while (i--) {
-					if (error_stack[i]) break;
+					if (errors[i]) break;
 				}
 
-				error_stack.splice(i + 1);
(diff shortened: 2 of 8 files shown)


rmunn:
The root cause was modifying `error_stack` in place in the `create_manifest_data` function. Under certain conditions, which #2383 reproduces, `error_stack` could contain `[ root_error, undefined, undefined, nested_error ]` and the correct logic would be to remove precisely one `undefined` from that array, but not two. But because `error_stack` was being modified in place rather than a copy being made and the copy modified, the first pass through the loop would correctly remove a single `undefined` to produce `[ root_error, undefined, nested_error ]`, but the second pass through the loop would again remove a single `undefined` to produce `[ root_error, nested_error ]`. The second pass through the loop would therefore produce an error stack that was out of sync with the component stack, and the error would then be rendered in the wrong layout.

By making a copy of the `error_stack` array and modifying the copy, both passes through the loop start with `[ root_error, undefined, undefined, nested_error ]` and therefore both passes correctly produce `[ root_error, undefined, nested_error ]`, which lines up the `nested_error` component correctly with the layout it's supposed to be rendered with.

benmccann:
thanks!

Issue #93, opened 2020-10-29
Content Security Policy support


benmccann:
I don't see anything in the code related to CSP support, so I'm guessing that doesn't exist yet unless I missed it.

Sapper supports it via `%sapper.cspnonce%`: https://sapper.svelte.dev/docs#Content_Security_Policy_CSP

acoyfellow:
Hi - it looks like this (mostly) can be solved with `hooks` now, in the `handle` function. 

no `nonce`'s yet but i'm working on it, maybe someone can see that solution

https://gist.github.com/acoyfellow/d8e86979c66ebea25e1643594e38be73

c-bandy:
Technically a duplicate of #93, but this issue has better info so lets go with this one?

seanlail:
@acoyfellow Great work so far!

I just want to highlight that it's really important that eval is not used in SvelteKit.
Can we ensure that there is no unsafe use, as @dominikg has suggested.

I've just been looking at an issue in Sapper [here](https://github.com/sveltejs/sapper/pull/1760) where the issue is that we have to allow `unsafe-eval`, which defeats the purpose of using CSP.

cc/ @Conduitry (as we were talking about this in the other ticket).

Rich-Harris:
Noting this, from [Twitter](https://twitter.com/captndestructo/status/1422378557108588544):

> is there any way to move the inline script for SvelteKit to an external one? I’d like to enforce a stricter CSP, but currently need unsafe-inline to accommodate. Thanks!

For server-rendered pages we could just use a `nonce`, I think? For pre-rendered pages that's no good, we would need to generate a hashed file (say `/boot/xyz123.js`) for each unique script.

dominikg:
could still hash inline content: https://content-security-policy.com/hash/

and pass the hash in a policy via meta in head https://content-security-policy.com/examples/meta/

I'm not sure if meta policies are additive though so may not be ideal.
external scripts would work best asthose are covered by self, no hash required

Karlinator:
As far as I can see it's only the init script in the head, right? Nothing else is inline? Or does this depend on build configurations/adapters?

In any case meta policies are allowed alongside HTTP policies. However, any source must pass all specified policies, meaning if you wanted to load scripts from a CDN it must be specified both places and the HTTP header must still send `unsafe-inline`. Basically, generating CSP in more than one place is kind of a mess.

See: https://www.w3.org/TR/CSP3/#multiple-policies

[...]

PR #2391, opened 2021-09-08
[feat] Add nonce placeholders in rendered pages.

Karlinator:
Related to #93 

Adds a CSP nonce placeholder to all script and style tags if `config.kit.noncePlaceholders` is true. This allows a hook to easily replace them with a nonce alongside where they are likely already inserting the CSP headers themselves.

As currently configured this supports `'strict-dynamic'` (that's why the nonce is added to all scripts, not just the inline init script).

Note that this only works on dynamic hosts. Nonces can't be used in a static environment.

This will need documentation somewhere; I'm open to doing that if someone points me to where to place it :) Particularly an example of a CSP hook would be good.

### Tests

### Changesets

### DIFF
--- a/.changeset/lovely-rats-cheer.md
+++ b/.changeset/lovely-rats-cheer.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Add option to place CSP nonce placeholders in rendered pages.
--- a/documentation/docs/15-content-security-policy.md
+++ b/documentation/docs/15-content-security-policy.md
@@ -1,0 +1,49 @@
+---
+title: Content Security Policy
+---
+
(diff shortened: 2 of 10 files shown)


benmccann:
Thanks for this! Can you add some docs in https://github.com/sveltejs/kit/tree/master/documentation/docs showing how it would be used?

Karlinator:
I've created an alternative at #2394 which generates and inserts nonces instead of just a placeholder. It seems like a better overall solution to me, but I'm keeping both PRs open so maintainers can choose the best option.

Karlinator:
I'm closing this in favour of #2394. We can come back to this concept if that one doesn't work out.

Issue #909, opened 2021-04-06
Allow to pass custom headers on load responses


jfranciscosousa:
**Is your feature request related to a problem? Please describe.**
This can be related to my previous issue, which is really just some details missing in the docs https://github.com/sveltejs/kit/issues/793

Despite `svelte-kit` having a way to define `maxage` to control caching for HTML pages, we really don't have a way of setting any other headers. The more common use cases are there: redirect and basic caching. But I feel like we need more control. On Next.js you can use the `context` object and use `setHeader`. Here on `svelte-kit` you can't even use the `context` because that's for layout only, so I need to resort to "hacky" solutions to get what I need.

My basic use case is to use `s-maxage` instead of `maxage`. I don't want to cache HTML pages on the client because JS and CSS hashes change between deploys, and if a user with an HTML page from a previous deploy refreshes the page (without clearing cache), the user gets an unstyled and unscripted page. Browsers make a request even if the assets are cached, but the 404 produced by the CDN (in this case, Vercel), results in those unstyled and unscripted pages. The only way around this is to probably use service workers and cache assets with it, but I want to avoid that.

I feel like a `headers` prop is just something that sooner or later will become handy, the same way it is on endpoints.

**Describe the solution you'd like**
Just allow us to specify the `headers` on a `load` response, just like on endpoints:

```js
  export const load = async ({ fetch }) => {
    const res = await fetch("/apistuff.json");

    if (res.ok) {
      const data = await res.json();

      return {
        props: {
          data
        },
        headers: {
          "cache-control": "public, s-maxage=3600"
        }
      };
    }

    return {
      error: new Error("error during fetch")
    };
  };
```

**Describe alternatives you've considered**
I've read about hooks and managed to work around this by doing:
```ts
import type { Handle } from "@sveltejs/kit";

/**
 * Replace all max-age cache controls with s-maxage
 */
export const handle: Handle = async (request, render) => {
  const response = await render(request);

  if (!response) return response;

  const cacheControlHeader = response?.headers?.["cache-control"];

  if (cacheControlHeader) {
    response.headers["cache-control"] = cacheControlHeader.replace("max-age", "s-maxage");
  }

  return {
    ...response
  };
};
```

Not that pretty honestly, but works for me. This replaces all `maxage` usages with `s-maxage` which is something I'm 100% ok with, but just for this project. But this just fixes the issue for **this project and this scenario**, if someone wants to pass another random header, no dice.

[...]

PR #2394, opened 2021-09-09
[feat] add CSP nonces to `script`/`style` tags

Karlinator:
Related to #93 

Alternative to #2391 (don't merge them both).

This PR adds an option, `kit.cspNonce`, which, when enabled, causes Kit to generate CSP nonces and embed them in pages when it renders them. It then makes the nonce value available to as `request.locals.nonce`, from where it can be used in a hook to be added into the Content Security Policy of the page, either using HTTP headers or HTML `<meta>` tags.

I made this because it felt a lot cleaner than doing a regex text replace, which is the alternative presented in #2391 (in fact there aren't any string searches). It also removes more of the burden from the developer.

I'm not completely confident that this will work in all environments. `node:crypto` is now an external runtime dependency, which should be fine for all environments actually running in node.

Again, this obviously doesn't work for `adapter-static`. A different strategy (externalising the init script) is required there.

The nonce is added to all scripts, even those not inline, in order to support `strict-dynamic`.  They are also added to stylesheets because it can't hurt (and even allows `style-src: 'none'` in certain cases).

### Tests

### Changesets

### DIFF
--- a/documentation/docs/14-content-security-policy.md
+++ b/documentation/docs/14-content-security-policy.md
@@ -4,7 +4,7 @@ title: Content Security Policy
 
 At the moment, SvelteKit supports adding Content Security Policy via hooks. In environments with a runtime, HTTP headers can be added to the response object.
 
-However, SvelteKit also requires some small pieces of inline JavaScript in order for hydration to work. To avoid using `'unsafe-inline'` (which, as the name suggests, should be avoided), SvelteKit can be configured to inject CSP Nonces into the html it generates.
+However, SvelteKit also requires some small pieces of inline JavaScript in order for hydration to work. To avoid using `'unsafe-inline'` (which, as the name suggests, should be avoided), SvelteKit can be configured to inject CSP nonces into the HTML it generates.
 
 The nonce value is availiable to hooks as `request.locals.nonce`. A basic CSP handler hook might then look like this:
 
@@ -42,7 +42,7 @@ export async function handle ({ request, resolve }) => {
 
 Because of the way Vite performs hot reloads of stylesheets, `'unsafe-inline'` is required in dev mode.
 
-Be warned: some other features of Svelte (in particular CSS transitions and animations) might run afoul of this Content Security Policy and require either rewriting to JS-based transitions or enabling `style-src: 'unsafe-inline'`.
+Be warned: some other features of Svelte ([in particular CSS transitions and animations](https://github.com/sveltejs/svelte/issues/6662)) might run afoul of this Content Security Policy and require either rewriting to JS-based transitions or enabling `style-src: 'unsafe-inline'`.
 
 The `'strict-dynamic'` directive is optional but supported by Kit. If not using it you must allow `'self'`.
 
--- a/documentation/docs/14-content-security-policy.md
+++ b/documentation/docs/14-content-security-policy.md
@@ -2,7 +2,7 @@
 title: Content Security Policy
 ---
 
-At the moment, SvelteKit supports adding Content Security Policy via hooks. In environments with a runtime, HTTP headers can be added to the response object.
+At the moment, SvelteKit supports adding Content Security Policy via hooks. In environments with a server-side runtime, HTTP headers can be added to the response object.
 
 However, SvelteKit also requires some small pieces of inline JavaScript in order for hydration to work. To avoid using `'unsafe-inline'` (which, as the name suggests, should be avoided), SvelteKit can be configured to inject CSP nonces into the HTML it generates.
 
--- a/documentation/docs/14-content-security-policy.md
+++ b/documentation/docs/14-content-security-policy.md
@@ -6,27 +6,27 @@ At the moment, SvelteKit supports adding Content Security Policy via hooks. In e
 
 However, SvelteKit also requires some small pieces of inline JavaScript in order for hydration to work. To avoid using `'unsafe-inline'` (which, as the name suggests, should be avoided), SvelteKit can be configured to inject CSP nonces into the HTML it generates.
 
-The nonce value is availiable to hooks as `request.locals.nonce`. A basic CSP handler hook might then look like this:
+The nonce value is available to hooks as `request.locals.nonce`. A basic CSP handler hook might then look like this:
 
 ```javascript
 export async function handle ({ request, resolve }) {
-  const directives = {
-    'default-src': ["'self'", 'static.someotherdomain.com'],
-    'script-src': ["'strict-dynamic'"],
-    'style-src': ["'self'"]
-  };
   const response = await resolve(request);
 
   if (response.headers['content-type'] !== 'text/html') {
   	return response
-	}
+  }
 
   const nonce = request.locals.nonce;
 
-  directives['script-src'].push(`'nonce-${nonce}'`);
-  directives['style-src'].push(`'nonce-${nonce}'`);
+  const directives = {
+    'default-src': ["'self'", 'static.someotherdomain.com'],
+    'script-src': ["'strict-dynamic'", `'nonce-${nonce}'`],
+    'style-src': ["'self'", `'nonce-${nonce}'`]
+	};
 
   if (process.env.NODE_ENV === 'development') {
+    // Because of the way Vite performs hot reloads of stylesheets,
+    // 'unsafe-inline' is required in dev mode.
     directives['style-src'].push('unsafe-inline')
   }
 
@@ -43,8 +43,6 @@ export async function handle ({ request, resolve }) {
   };
 };
 ```
-
-Because of the way Vite performs hot reloads of stylesheets, `'unsafe-inline'` is required in dev mode.
 
 Be warned: some other features of Svelte ([in particular CSS transitions and animations](https://github.com/sveltejs/svelte/issues/6662)) might run afoul of this Content Security Policy and require either rewriting to JS-based transitions or enabling `style-src: 'unsafe-inline'`.
 
--- a/documentation/docs/14-content-security-policy.md
+++ b/documentation/docs/14-content-security-policy.md
@@ -4,7 +4,7 @@ title: Content Security Policy
 
 At the moment, SvelteKit supports adding Content Security Policy via hooks. In environments with a server-side runtime, HTTP headers can be added to the response object.
 
-However, SvelteKit also requires some small pieces of inline JavaScript in order for hydration to work. To avoid using `'unsafe-inline'` (which, as the name suggests, should be avoided), SvelteKit can be configured to inject CSP nonces into the HTML it generates.
+However, SvelteKit also requires some small pieces of inline JavaScript for hydration. To avoid using `'unsafe-inline'` (which, as the name suggests, should be avoided), SvelteKit can be configured to inject CSP nonces into the HTML it generates.
 
 The nonce value is available to hooks as `request.locals.nonce`. A basic CSP handler hook might then look like this:
 
@@ -13,8 +13,8 @@ export async function handle ({ request, resolve }) {
   const response = await resolve(request);
 
   if (response.headers['content-type'] !== 'text/html') {
-  	return response
-  }
+  	return response;
+	}
 
   const nonce = request.locals.nonce;
 
--- a/documentation/docs/14-content-security-policy.md
+++ b/documentation/docs/14-content-security-policy.md
@@ -17,6 +17,10 @@ export async function handle ({ request, resolve }) {
   };
   const response = await resolve(request);
 
+  if (response.headers['content-type'] !== 'text/html') {
+  	return response
+	}
+
   const nonce = request.locals.nonce;
 
   directives['script-src'].push(`'nonce-${nonce}'`);
--- a/documentation/docs/14-content-security-policy.md
+++ b/documentation/docs/14-content-security-policy.md
@@ -9,43 +9,43 @@ However, SvelteKit also requires some small pieces of inline JavaScript for hydr
 The nonce value is available to hooks as `request.locals.nonce`. A basic CSP handler hook might then look like this:
 
 ```javascript
-export async function handle ({ request, resolve }) {
-  const response = await resolve(request);
+export async function handle({ request, resolve }) {
+	const response = await resolve(request);
 
-  if (response.headers['content-type'] !== 'text/html') {
-  	return response;
+	if (response.headers['content-type'] !== 'text/html') {
+		return response;
 	}
 
-  const nonce = request.locals.nonce;
+	const nonce = request.locals.nonce;
 
-  const directives = {
-    'default-src': ["'self'", 'static.someotherdomain.com'],
-    'script-src': ["'strict-dynamic'", `'nonce-${nonce}'`],
-    'style-src': ["'self'", `'nonce-${nonce}'`]
+	const directives = {
+		'default-src': ["'self'", 'static.someotherdomain.com'],
+		'script-src': ["'strict-dynamic'", `'nonce-${nonce}'`],
+		'style-src': ["'self'", `'nonce-${nonce}'`]
 	};
 
-  if (process.env.NODE_ENV === 'development') {
-    // Because of the way Vite performs hot reloads of stylesheets,
-    // 'unsafe-inline' is required in dev mode.
-    directives['style-src'].push('unsafe-inline')
-  }
+	if (process.env.NODE_ENV === 'development') {
+		// Because of the way Vite performs hot reloads of stylesheets,
+		// 'unsafe-inline' is required in dev mode.
+		directives['style-src'].push('unsafe-inline');
+	}
 
-  const csp = Object.entries(directives)
-    .map(([key, arr]) => key + ' ' + arr.join(' '))
-    .join('; ');
+	const csp = Object.entries(directives)
+		.map(([key, arr]) => key + ' ' + arr.join(' '))
+		.join('; ');
 
-  return {
-    ...response,
-    headers: {
-      ...response.headers,
-      'Content-Security-Policy': csp
-    }
-  };
-};
+	return {
+		...response,
+		headers: {
+			...response.headers,
+			'Content-Security-Policy': csp
+		}
+	};
+}
 ```
 
 Be warned: some other features of Svelte ([in particular CSS transitions and animations](https://github.com/sveltejs/svelte/issues/6662)) might run afoul of this Content Security Policy and require either rewriting to JS-based transitions or enabling `style-src: 'unsafe-inline'`.
 
 The `'strict-dynamic'` directive is optional but supported by Kit. If not using it you must allow `'self'`.
 
-The nonce placeholders can be toggled with the `kit.cspNonce` configuration option.
+The nonce placeholders can be toggled with the `kit.cspNonce` configuration option. Since nonces must be uniquely generated for each request, this also disables prerendering.
--- a/documentation/docs/14-content-security-policy.md
+++ b/documentation/docs/14-content-security-policy.md
@@ -9,7 +9,7 @@ However, SvelteKit also requires some small pieces of inline JavaScript in order
 The nonce value is availiable to hooks as `request.locals.nonce`. A basic CSP handler hook might then look like this:
 
 ```javascript
-export async function handle ({ request, resolve }) => {
+export async function handle ({ request, resolve }) {
   const directives = {
     'default-src': ["'self'", 'static.someotherdomain.com'],
     'script-src': ["'strict-dynamic'"],
--- a/documentation/docs/14-content-security-policy.md
+++ b/documentation/docs/14-content-security-policy.md
@@ -1,0 +1,49 @@
+---
+title: Content Security Policy
+---
+
+At the moment, SvelteKit supports adding Content Security Policy via hooks. In environments with a runtime, HTTP headers can be added to the response object.
+
+However, SvelteKit also requires some small pieces of inline JavaScript in order for hydration to work. To avoid using `'unsafe-inline'` (which, as the name suggests, should be avoided), SvelteKit can be configured to inject CSP Nonces into the html it generates.
+
+The nonce value is availiable to hooks as `request.locals.nonce`. A basic CSP handler hook might then look like this:
+
+```javascript
+export async function handle ({ request, resolve }) => {
+  const directives = {
+    'default-src': ["'self'", 'static.someotherdomain.com'],
+    'script-src': ["'strict-dynamic'"],
+    'style-src': ["'self'"]
+  };
+  const response = await resolve(request);
+
+  const nonce = request.locals.nonce;
+
+  directives['script-src'].push(`'nonce-${nonce}'`);
+  directives['style-src'].push(`'nonce-${nonce}'`);
+
+  if (process.env.NODE_ENV === 'development') {
+    directives['style-src'].push('unsafe-inline')
+  }
+
+  const csp = Object.entries(directives)
+    .map(([key, arr]) => key + ' ' + arr.join(' '))
+    .join('; ');
+
+  return {
+    ...response,
+    headers: {
+      ...response.headers,
+      'Content-Security-Policy': csp
+    }
+  };
+};
+```
+
+Because of the way Vite performs hot reloads of stylesheets, `'unsafe-inline'` is required in dev mode.
+
+Be warned: some other features of Svelte (in particular CSS transitions and animations) might run afoul of this Content Security Policy and require either rewriting to JS-based transitions or enabling `style-src: 'unsafe-inline'`.
+
+The `'strict-dynamic'` directive is optional but supported by Kit. If not using it you must allow `'self'`.
+
+The nonce placeholders can be toggled with the `kit.cspNonce` configuration option.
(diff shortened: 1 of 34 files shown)


Karlinator:
Hmm, looking at the tests I may have overestimated support for nodeJS builtins in serverless environments. We might get somewhere using the Web Crypto API instead? I know Cloudflare supports it, though that's only experimental in Node (and only in node 16).

Perhaps this can be injected somehow from the adapters? So `adapter-node` would use `crypto`, Cloudflare would use the global webcrypto object etc?

If there is no reasonable way to generate a random value across platforms then we might just have to leave that up to the developer. An alternative then is to take in the nonce as `request.locals.nonce` and insert it directly. In effect we would be keeping most of the good parts of this approach (no regex) and requiring that the developer provide a random value themselves based on whichever platform they are running on.

packages/kit/src/runtime/server/index.js
```diff
@@ -32,12 +33,15 @@ export async function respond(incoming, options, state = {}) {
 	}
 
 	const headers = lowercase_keys(incoming.headers);
+	const nonce = randomBytes(32).toString('base64');
```
dominikg, line 36: Is a plain base64 nonce safe to use everywhere or does it need to replace some chars to avoid escaping a context its used in (http attribute, http header value, ???)

dominikg:
I like this a lot more than the placeholder approach. Is it allowed to reuse nonce values for multiple scripts or should each one have a separate one?

Would it also be posssible to calculate hashes instead of just a random nonce?

packages/kit/src/runtime/server/index.js
Karlinator: Base64 should be safe here. In the linked file they replace `+`, `/` and `=`, none of which are problematic in this context. In the Vite plugin this is done because the result is used as filenames, where `/` in particular is problematic.

Looking at [the list of base64 symbols](https://base64.guru/learn/base64-characters) everything is safe for this context.

Karlinator:
We could _probably_ calculate hashes, but that requires data flow in the other direction instead (or we're back to regexes). I'm not sure why you want hashes for this? SSR like this is basically the poster child for CSP nonces.

Using the same nonce for every resource on a single page request is (as far as I can tell) accepted practice. There shouldn't be any security degradation from that (still impossible to guess, and all nonces after initial page load are disregarded anyway), and it avoids some unpleasant complexity for us since we don't need to know ahead of time how many nonces to generate.

The randomness is a problem in general (see my above comment on platform support). But the current version (which only works on Node) uses Node's `crypto` API, which is intended for exactly these kinds of things. I even doubled the recommended minimum nonce length from 128 bits to 256 bits just for the sake of it!

dominikg:
My concern in regards to random was that in a scenario with a high request volume and low entropy on the server, random might be become blocking/adding to the response time https://nodejs.org/api/crypto.html#crypto_crypto_randombytes_size_callback

The advantage of hashes is that they can be precalculated on build time (for anything non-dynamic at least) and reused from a manifest file, so less work at runtime and no need for a source of randomness.

Karlinator:
I see your concern now, thank you for elaborating. I don't know? I haven't measured response times, but `randopmBytes()` seems to be the common way to generate nonces in NodeJS.

I kind of consider hashing a separate feature to be honest, though I absolutely support it. As discussed in #93 some parts of the init script is unique to the request under SSR so hashing can't be done ahead-of-time for dynamic environments. However, it would be a nice inclusion for static environments or on a per-request basis if entropy is indeed a problem. I imagine hashing within the render and then providing the hash value as `request.locals.cspHashes.scripts[0]`. Static environments could insert it into a CSP `<meta>` tag.

Could we require (or encourage) adapters to shim a `generateNonce()` (and potentially a `hashScript()` ) function similar to what we do with `fetch`? `adapter-node` would map it to `randomBytes(32).toString('base64')`, `adapter-cloudflare-workers` would use the WebCrypto API, etc. It's one more thing for adapters to implement, but I think that's pretty reasonable. I would hope all environments have _some_ way of generating a random string, though I can't guarantee that. I guess if it can't then it should just emit an error at build time if the `kit.scpNonce` option is set.

Karlinator:
Marked as draft because of aforementioned platform issues.

dominikg:
crazy idea, feel free to put me on blast for this.

What if you did extract all inline items into virtual files served from memory

```html
<body>
<script>
  alert('inline');
</script> 
</body>
```
would become 
```html
<body>
<script src="some-virtual-script-uri" />
</body>
```

As the generated html is served to the client in the very next step the client is going to request the virtual files, the memory cache holding them could evict them after first response, hopefully without blowing up the server requirements.

Yes, extra roundtrips occur, and it would most certainly not be cool for stuff like critical inlined css (but that should be static and could be hashed instead).

Also i am very happy to see your efforts here, please don't take my questions or suggestions as a detriment. I want this as much as you, so if nonces turn out to be a good first step into offering baseline usable csp support for kit with SSR, i'm all for it.

Karlinator:
That would require server state would it not? I like the idea, but I'm not sure it would work well on serverless platforms? Or if running several parallel instances?

I am developing a SvelteKit application for production (living on the edge here, one of the reasons I'm keen on this feature), and in our setup (`adapter-node` running in containers) I _think_ I could accommodate that solution, but it would require setting up session-based routing which isn't the case right now since SvelteKit is stateless.

A similar-ish crazy idea is to set up a special endpoint `/init-script?session=whatever&hydrate=stuff`, and then the init script becomes `<script src="/init-script?session=whatever"></script>`, essentially feeding the information that is now embedded into the script to GET parameters instead, where they would then be embedded into the script. That avoids any server state, though at the cost of performing more dynamic file rendering.

Either way I think we should still want the option of nonces since it enables `strict-dynamic`. Hashes technically work for this too, but that would require moving some stuff around (you can nonce a script with src, but you can't hash it).

Karlinator:
I've (not without a little bit of trouble) changed this to use a `cspNonceGenerator`  which it expects to be shimmed by the adapter, and implemented such a shim in `adapter-node` (in general they should be quite easy to implement for others as well).

The lint CI job evidently doesn't like the way I did this; any suggestions @dominikg @benmccann ?

documentation/docs/14-content-security-policy.md
```diff
@@ -0,0 +1,49 @@
+---
+title: Content Security Policy
+---
+
+At the moment, SvelteKit supports adding Content Security Policy via hooks. In environments with a runtime, HTTP headers can be added to the response object.
+
+However, SvelteKit also requires some small pieces of inline JavaScript in order for hydration to work. To avoid using `'unsafe-inline'` (which, as the name suggests, should be avoided), SvelteKit can be configured to inject CSP Nonces into the html it generates.
```
benmccann, line 7:
```suggestion
However, SvelteKit also requires some small pieces of inline JavaScript in order for hydration to work. To avoid using `'unsafe-inline'` (which, as the name suggests, should be avoided), SvelteKit can be configured to inject CSP nonces into the HTML it generates.
```

documentation/docs/14-content-security-policy.md
```diff
@@ -0,0 +1,49 @@
...
+      ...response.headers,
+      'Content-Security-Policy': csp
+    }
+  };
+};
+```
+
+Because of the way Vite performs hot reloads of stylesheets, `'unsafe-inline'` is required in dev mode.
+
+Be warned: some other features of Svelte (in particular CSS transitions and animations) might run afoul of this Content Security Policy and require either rewriting to JS-based transitions or enabling `style-src: 'unsafe-inline'`.
```
benmccann, line 45:
```suggestion
Be warned: some other features of Svelte ([in particular CSS transitions and animations](https://github.com/sveltejs/svelte/issues/6662)) might run afoul of this Content Security Policy and require either rewriting to JS-based transitions or enabling `style-src: 'unsafe-inline'`.
```

Karlinator:
I think this might actually be ready now, though I don't know that I have tested the adapters well enough.

I am open to suggestions for naming things better 😄 

Third-party adapters will need to update to implement this, but users will only notice that it's missing if they turn on `kit.cspNonce`, which will generate runtime warning and promptly not do anything, so it should be safe to roll out and let adapters catch up.

Karlinator:
So to clarify the final state of this PR:

A new config option `kit.cspNonce` is added. When this is true, SvelteKit will generate nonces for each request and embed it into all relevant script and style tags. The nonce value is exposed as `request.locals.nonce`.

Kit relies on the adapters to shim a nonce generator, `generateCspNonce()`, which should return a cryptographically secure (pseudo)random string with at least 128 bits of entropy in a suitable encoding (like base64). This PR implements this into all the official adapters (as well as the dev server). Adapters which don't shim this will cause Kit to emit a runtime warning if `kit.cspNonce` is set, and will prevent Kit from inserting any nonce value at all (and `request.locals.nonce` will be `undefined`).

Implementation varies a little between the adapters. Most of them have access to Node's `crypto` API, while `adapter-cloudflare-workers` uses the WebCrypto API which is provided in the global scope there.

Nonce generation does not run when prerendering. This PR has absolutely no effect on static sites (we'll need hashing or init script externalisation for that).

documentation/docs/14-content-security-policy.md
```diff
@@ -0,0 +1,49 @@
...
+
+  const csp = Object.entries(directives)
+    .map(([key, arr]) => key + ' ' + arr.join(' '))
+    .join('; ');
+
+  return {
+    ...response,
+    headers: {
+      ...response.headers,
+      'Content-Security-Policy': csp
```
benmccann, line 37: Doing this in the `handle` hook would end up applying to endpoints as well, which probably isn't quite what we want. We should probably do this in the layout instead. Of course that would require https://github.com/sveltejs/kit/issues/909 to be implemented first. That issue is tagged to the 1.0 milestone, so hopefully we'll get to it before too too long

Karlinator: While this is true (and will be relevant later) we can't document for a feature that doesn't exist yet 😄 

Setting a CSP on an endpoint response probably won't _hurt_ (it'll most likely just be discarded), but you can also check if the response type is "text/html" first. It invokes the handler at every endpoint call, but won't add headers to them. I should maybe add that to the example.

I've been thinking a little about a more permanent home for CSP headers in a more built-in way, but that's a different discussion.

Karlinator: Actually, doing this in a layout requires more than just the ability to set headers. It also requires access to `request.locals` in order to put the nonce into the header, which is problematic I think.

Karlinator:
I've made a change to the adapter API proposed in this PR, in order to be a little bit more generic.

Instead of `generateCspNonce`, Kit now expects a function `generateRandomString(bytes: number)`, which should return a base64-encoded random string with `bytes` number of bytes. This allows us to choose the level of entropy generated, for any future needs.

Karlinator:
Testing has revealed that (in addition to not having implemented it correctly in the first place), noncing only works very inconsistently in `adapter-cloudflare-workers`. It seems to be serving prerendered pages for the most part, except when it is forbidden to by the presence of a load function? At any rate, it's often serving pages without running any of the relevant code running at all. I don't know how many environments behave like this? I also don't know a great solution at the moment.

Defman:
Instead of having svelte generate the nonce, could we delegate it to the adapter. Ie having svelte render function take a nonce , which would be added to all scripts, styles, and etc. and added to the returned headers by render.

```js
const rendered = await render({
	host: request_url.host,
	path: request_url.pathname,
	query: request_url.searchParams,
	rawBody: await read(request),
	headers: Object.fromEntries(request.headers),
	method: request.method,
        nonce: "rAnd0m",
});

return new Response(rendered.body, {
	status: rendered.status,
	headers: makeHeaders(rendered.headers)
});
```

Karlinator:
@Defman I really liked that, so I did it!

The adapter generates a nonce in it's own way and simply supplies it as `incoming.nonce`. There is _some_ added complexity related to not generating nonces if `kit.cspNonce` is false (specifically the adapters have to inject that value with esbuild's `define` property), but this feels cleaner overall, and the API is a little less implicit this way.

At a later date we probably want to offer hashing as well, which has some similar problems (platform-dependent api to access hash functions, for example) with the added complexity that hashes must be computed inside Kit as they cannot be precomputed. A similar non-global-shim solution for that (on the adapter side) would be something like this (example is adapter-cloudflare, so Web Crypto API):
```javascript
const rendered = await render(
    {
      host: request_url.host,
      path: request_url.pathname,
      query: request_url.searchParams,
      rawBody: await read(request),
      headers: Object.fromEntries(request.headers),
      method: request.method,
      nonce: generateNonces && btoa(crypto.getRandomValues(new Uint32Array(2)))
    },
    {},
    generateHashes && async (content) => btoa(await crypto.subtle.digest('SHA-256', content))
);
```

Defman:
I don't think we have any issue supporting prerendered pages, we would just have to supply a nonce at pregeneration and store it when later serving the pregenerated pages. The nonce would stay the same, but since the content of the pages can't change it would not matter.

Karlinator:
The CSP Spec does not agree:

>Nonces override the other restrictions present in the directive in which they’re delivered. It is critical, then, that they remain unguessable, as bypassing a resource’s policy is otherwise trivial.

>If a server delivers a nonce-source expression as part of a policy, the server MUST generate a unique value each time it transmits a policy. The generated value SHOULD be at least 128 bits long (before encoding), and SHOULD be generated via a cryptographically secure random number generator in order to ensure that the value is difficult for an attacker to predict.

https://www.w3.org/TR/CSP3/#security-nonces

"Each time it transmits a policy" is very clear, and does not allow storing nonces generated ahead of time. I would _strongly_ urge we don't break with spec over this (though some developers might hack their way into a system that does).

If you are generating the policy ahead of time (say, in a static environment), you should use hashes instead.

There is more discussion around this topic in general at #93, but my intention is to also implement hashing, especially for use in prerendering. We can return the hash lists as part of the ServerResponse or something, and the adapter can take it from there.

Defman:
Like render takes a `nonce` we can now provide a `hash` function, which takes bytes and spits out a hash. That way the adapter can chose what hash to use. More general we could provide some way a function which is called pr style/js block and returns an object which are added as properties for the block.

```js
const rendered = await render(
    {
      host: request_url.host,
      path: request_url.pathname,
      query: request_url.searchParams,
      rawBody: await read(request),
      headers: Object.fromEntries(request.headers),
      method: request.method,
      hooks: { script: ({}) => { nonce: "rAnd0m" }, style: script: ({}) => { nonce: "rAnd0m" }}
    }
);
```

And for prerender

```js
let hashes = [];

const rendered = await render(
    {
      host: request_url.host,
      path: request_url.pathname,
      query: request_url.searchParams,
      rawBody: await read(request),
      headers: Object.fromEntries(request.headers),
      method: request.method,
      hooks: { script: async ({content}) => { hashes.push(await crypto.subtle.digest('SHA-256', content));, ... }
    }
);
```

Just a suggestion, this way we delegate all the work to the adapter. We can ofcs, provide some utilities to make it easier and simpler.

Karlinator:
Sadly, Kit can't really be _completely_ agnostic between hashes and nonces, given that Kit has to inject the nonces (but not the hashes) into the html itself. And the `render` function is only ever really called for html documents anyway (all `js` and `css` files are treated as static assets). The original problem we're trying to solve here is the inline init script in the html `head` which governs hydration (and is dynamically generated), which makes the whole issue fairly specialised to begin with. The fact that I'm also adding the nonce to all the other tags is essentially just an afterthought which enables using `strict-dynamic` for those who are so inclined.

Defman:
Hmm the hashes could become part of the manifest, that would solve all the problems right?

Karlinator:
The init script is unique for each request for dynamic pages. It _can't_ be hashed ahead of time except for in static (prerendered) pages.

Hashing static assets into the manifest could be a good idea in general; they can still be used to enable `strict-dynamic`, or just to lock down the CSP with hash+host. The script producing the hash list would have to get most of it from the manifest but also hash the init script.

documentation/docs/14-content-security-policy.md
benmccann: Ok, yeah, let's do it in `handle`. But can we check if the `content-type` is `text/html` and only do it then?

Karlinator: Yeah that'll work. In fact I already added that to the example :)

Karlinator:
You've got to love a single seemingly-unrelated test case failing, but only in a single version of node and not on windows/macOS. _sigh_

benmccann:
The tests are flaky. I wouldn't worry about it

documentation/docs/14-content-security-policy.md
```diff
@@ -0,0 +1,53 @@
+---
+title: Content Security Policy
+---
+
+At the moment, SvelteKit supports adding Content Security Policy via hooks. In environments with a runtime, HTTP headers can be added to the response object.
```
benmccann, line 5:
```suggestion
At the moment, SvelteKit supports adding Content Security Policy via hooks. In environments with a server-side runtime, HTTP headers can be added to the response object.
```

documentation/docs/14-content-security-policy.md
```diff
@@ -0,0 +1,53 @@
...
+
+However, SvelteKit also requires some small pieces of inline JavaScript in order for hydration to work. To avoid using `'unsafe-inline'` (which, as the name suggests, should be avoided), SvelteKit can be configured to inject CSP nonces into the HTML it generates.
```
benmccann, line 7:
```suggestion
However, SvelteKit also requires some small pieces of inline JavaScript for hydration. To avoid using `'unsafe-inline'` (which, as the name suggests, should be avoided), SvelteKit can be configured to inject CSP nonces into the HTML it generates.
```

documentation/docs/14-content-security-policy.md
```diff
@@ -0,0 +1,53 @@
...
+
+The nonce value is availiable to hooks as `request.locals.nonce`. A basic CSP handler hook might then look like this:
```
benmccann, line 9:
```suggestion
The nonce value is available to hooks as `request.locals.nonce`. A basic CSP handler hook might then look like this:
```

benmccann:
It looks like this PR will need a minor rebase

documentation/docs/14-content-security-policy.md
```diff
@@ -0,0 +1,53 @@
...
+export async function handle ({ request, resolve }) {
+  const directives = {
+    'default-src': ["'self'", 'static.someotherdomain.com'],
+    'script-src': ["'strict-dynamic'"],
+    'style-src': ["'self'"]
+  };
+  const response = await resolve(request);
+
+  if (response.headers['content-type'] !== 'text/html') {
+  	return response
```
benmccann, line 21:
```suggestion
  	return response;
```

documentation/docs/14-content-security-policy.md
```diff
@@ -0,0 +1,53 @@
...
+
+At the moment, SvelteKit supports adding Content Security Policy via hooks. In environments with a runtime, HTTP headers can be added to the response object.
+
+However, SvelteKit also requires some small pieces of inline JavaScript in order for hydration to work. To avoid using `'unsafe-inline'` (which, as the name suggests, should be avoided), SvelteKit can be configured to inject CSP nonces into the HTML it generates.
+
+The nonce value is availiable to hooks as `request.locals.nonce`. A basic CSP handler hook might then look like this:
+
+```javascript
+export async function handle ({ request, resolve }) {
+  const directives = {
```
benmccann, line 13: can you move this down to line 26 where it's first used? you could also then add the nonces where you declare this instead of calling `.push`

documentation/docs/14-content-security-policy.md
```diff
@@ -0,0 +1,53 @@
...
+  if (response.headers['content-type'] !== 'text/html') {
+  	return response
+	}
+
+  const nonce = request.locals.nonce;
+
+  directives['script-src'].push(`'nonce-${nonce}'`);
+  directives['style-src'].push(`'nonce-${nonce}'`);
+
+  if (process.env.NODE_ENV === 'development') {
```
benmccann, line 29: Is this needed because of https://github.com/sveltejs/svelte/issues/6662? If so, it might be nice to add a comment here with a link to the issue. That issue doesn't mention anything about it being only for dev mode, so I'm wondering if that should be added to the issue description or we should remove this `if` check?

benmccann: Oh, I see the explanation down below on line 47. I'd move it up here and make it a comment

documentation/docs/14-content-security-policy.md
```diff
@@ -0,0 +1,53 @@
...
+    }
+  };
+};
+```
+
+Because of the way Vite performs hot reloads of stylesheets, `'unsafe-inline'` is required in dev mode.
+
+Be warned: some other features of Svelte ([in particular CSS transitions and animations](https://github.com/sveltejs/svelte/issues/6662)) might run afoul of this Content Security Policy and require either rewriting to JS-based transitions or enabling `style-src: 'unsafe-inline'`.
+
+The `'strict-dynamic'` directive is optional but supported by Kit. If not using it you must allow `'self'`.
```
benmccann, line 51:
```suggestion
The `'strict-dynamic'` directive is optional but supported by Kit. If not using it, you must allow `'self'`.
```

packages/adapter-cloudflare-workers/index.js
```diff
@@ -34,11 +34,16 @@ export default function (options) {
 			utils.log.minor('Generating worker...');
 			utils.copy(`${files}/entry.js`, '.svelte-kit/cloudflare-workers/entry.js');
 
+			writeFileSync('${files}/nofig.js', `export const generateNonces = ${config.kit.cspNonce}`);
```
benmccann, line 37: what is `nofig`? does this file get included somehow? I don't see anything referring to it. Can we give it a clearer name?

packages/adapter-cloudflare-workers/files/entry.js
```diff
@@ -1,3 +1,4 @@
+/* global GENERATE_NONCES */
```
benmccann, line 1: would it work to remove this line and do `globalThis.GENERATE_NONCES` instead?

packages/kit/src/core/build/index.js
```diff
@@ -349,7 +349,8 @@ async function build_server(
 					ssr: ${s(config.kit.ssr)},
 					target: ${s(config.kit.target)},
 					template,
-					trailing_slash: ${s(config.kit.trailingSlash)}
+					trailing_slash: ${s(config.kit.trailingSlash)},
+					cspNonce: ${s(config.kit.cspNonce)}
```
benmccann, line 353: We use snake_case for internal variables
```suggestion
					csp_nonce: ${s(config.kit.cspNonce)}
```

packages/kit/src/runtime/server/index.js
```diff
@@ -32,12 +32,26 @@ export async function respond(incoming, options, state = {}) {
 	}
 
 	const headers = lowercase_keys(incoming.headers);
+	/**
+	 * @type {string | undefined}
```
benmccann, line 36: You could make this a single line:
```
/** @type {string | undefined} */
let nonce;
```

packages/adapter-cloudflare-workers/index.js
Karlinator: I honestly don't know. Probably a typo for 'config'? I think this line was supposed to be removed actually, as the variable is passed in through esBuild.

packages/adapter-cloudflare-workers/files/entry.js
Karlinator: Maybe, I'll try it.

Karlinator:
@pzuraq well the first thing you can do is nudge me, which you've already done to great effect 😄 

I've gone through the last comments from Ben now, but I haven't had the time to test all the adapters and verify that they work, which probably should be done before this can be merged. You could do that maybe? Otherwise I'll get to it at some point, but it's been busier lately.

Karlinator:
@benmccann turns out we couldn't use `globalThis.GENERATE_NONCES` (esbuild creates variables, not properties on the global scope). I think this is cleaner though?

I ran into an issue with some of the adapters always serving prerendered pages, which meant they completely ignored the nonces. My solution right now is to disable prerendering when `config.kit.cspNonce` is set, since combining the two doesn't really make sense (unless you use a placeholder and string replacement to insert the nonce into the prerendered page, which makes it not really prerendered anymore). I know Kit likes the concept of mixing prerendered and not prerendered pages, but I'm not sure what the best way of supporting this would be. If you want a CSP without gaping holes then we would need to add hashing as well probably. For now I think this is an acceptable tradeoff. If that sounds good I'll document it and then this should be ready to merge.

I have now tested this with all the official adapters, and they all work as expected. This doesn't break anything for 3rd-party adapters (it's a new expectation, but the worst that can happen if they don't support it is that no nonces are added), so giving them a heads up should be enough?

benmccann:
Oof. I'm terribly sorry we haven't gotten this PR in yet. I just came back to take a look and I'm afraid there are quite a few conflicts now. I hope it won't be too hard to rebase

Karlinator:
Hmm, the adapter API overhaul (the Builder and stuff) breaks the way I did this. I need to look more closely at it to find a good way to tell the adapter runtime whether it should pass nonces or not.

Karlinator:
Closing in favour of #3499

PR #2399, opened 2021-09-10
[docs] better explain how to start working with the codebase

babakfp:
Hi
I wasted a bunch of time on how to start to develop SvelteKit, so I decided to write a better step-by-step guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,40 +30,41 @@ Please make sure the issue you're reporting involves SvelteKit. Many issues rela
 If an issue is caused by Vite, please report in the [Vite issue tracker](https://github.com/vitejs/vite/issues).
 
 ## Developing
+This description guides you on how to get started developing SvelteKit itself.
 
-1. First of all, you need to install [`pnpm`](https://pnpm.js.org/en/). You can't use any other package managers like: `npm` or `yarn`.
+1. First of all, you need to install [`pnpm`](https://pnpm.js.org/en/). You can't use any other package managers like `npm` or `yarn`.
 ```bash
 npm i -g pnpm
 ```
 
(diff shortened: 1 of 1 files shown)


dummdidumm:
What where the problems you had with the current guide? For me the PR seems not very different to what's already written down.

babakfp:
Hi🖐
If you know what to do, yeah there is no noticeable difference, but if don't what to do, there is a good difference. I tried to explain step by step and what to do and what not to do.
Once I used yarn and once npm but it didn't work. So you need to only use pnpm. You need to run both repo dev and the example app from the root of the repo. You don't need to install the dependencies for the example app. and...

bluwy:
I also think the previous was already well explained. If you follow the instructions there you shouldn't be hitting into issues you had mentioned.

benbender:
Imho the difference is that the edited version doesn't require the user/dev to know about monorepos. The current version has some level of knowledge about them as a somewhat implicit requirement. Additionally it may be easier to follow complete sentences and distinct steps instead of the old version. At the end, I think it's an improvement.

Mlocik97:
Sry but actual version is better than Your PR.

babakfp:
I'm not sure about the `### Git hooks, how to use them` part. I just added a title(and changed its description) to separate it from the steps.

benmccann:
Thanks for the suggestions. It seems like there's been a mixed reaction about some of these suggestions. I integrated some of these suggestions into the README in various ways. What I came up with ended up looking a bit different than this, so I ended up just doing it as a new commit and will close this one. Hope it's clearer now. Thanks!!

PR #2407, opened 2021-09-12
[fix] prerendering not working with `kit.paths.base` set.

### LINKED ISSUE #2230: Addding config.kit.paths.base breaks `npm run build`

Karlinator:
Fixes #2230 

Local assets were not being excluded from being `visit`ed during prerendering because they were not being detected as being one of the local files.

Specifically, this used to work until the semantics of `kit.paths.assets` was changed in #2189 to only be either empty or an absolute (external) path, and assets were instead prefixed with `paths.base` if `paths.assets` was not an absolute path. This line should have been updated then.

### Tests

### Changesets

### DIFF
--- a/.changeset/unlucky-dodos-greet.md
+++ b/.changeset/unlucky-dodos-greet.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Fix prerendering/adapter-static failing when `kit.paths.base` was set.
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -262,7 +262,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 					const parsed = new URL(resolved, 'http://localhost');
 					const pathname = decodeURI(parsed.pathname);
 
-					const file = pathname.replace(config.kit.paths.assets, '').slice(1);
+					const file = pathname.replace(config.kit.paths.base, '').slice(1);
 					if (files.has(file)) continue;
 
 					if (parsed.search) {
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -260,16 +260,16 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 					if (!resolved.startsWith('/') || resolved.startsWith('//')) continue;
 
 					const parsed = new URL(resolved, 'http://localhost');
-					const pathname = decodeURI(parsed.pathname);
+					const pathname = decodeURI(parsed.pathname).replace(config.kit.paths.base, '');
 
-					const file = pathname.replace(config.kit.paths.base, '').slice(1);
+					const file = pathname.slice(1);
 					if (files.has(file)) continue;
 
 					if (parsed.search) {
 						// TODO warn that query strings have no effect on statically-exported pages
 					}
 
-					await visit(pathname.replace(config.kit.paths.base, ''), path);
+					await visit(pathname, path);
 				}
 			}
 		}


benmccann:
Did you see there's also another fix pending for this issue at https://github.com/sveltejs/kit/pull/2318? I'm not that familiar with this code and haven't looked closely at either PR yet to know which makes more sense

Karlinator:
I have seen that. It actually leaves this incorrect behaviour behind and fixes the issue by doing a string replacement before that point instead. That replacement was precisely what the line I fixed is intended to do (and what it did before #2189). Either PR will work, but I suggest that if #2318 is merged it should also remove that `pathname.replace(config.kit.paths.assets, '')` which does absolutely nothing.

Overall I'd say this seems like a cleaner fix to me.

baseballyama:
Hi.

I'm the author of https://github.com/sveltejs/kit/pull/2318.
I checked your PR and I think your PR is better than mine.

--- 
(This is just follow up)

According to current spec, `config.kit.paths.assets` can be empty string or absolute URL.
So always `if (!resolved.startsWith('/') || resolved.startsWith('//'))` should be `true` if  `config.kit.paths.assets` is absolute URL.
So after the process, it doesn't need to consider 'config.kit.paths.assets'.

Therefore my PR's code can move to after the `if` statement.
And at that time, my PR is almost the same as this PR and this PR has no useless processing compare to mine.
So I think this PR is better than mine.

---
By the way, I have 2 small requests in my opinion.
So if these are reasonable, I'm happy if you will implement them.

1. DRY

Now it has `pathname.replace(config.kit.paths.base, '')` twice.
So I think `decodeURI(parsed.pathname)` can be change to `decodeURI(parsed.pathname).replace(config.kit.paths.assets, '')`.

2. Prevent mis-replacement

`pathname.replace(config.kit.paths.assets, '')` has chance to miss replacement.
For instance, if `pathname` is `aaa/bbb/ccc` and  `config.kit.paths.base` is `bbb`,
`pathname.replace(config.kit.paths.assets, '')` become `aaa//ccc`.

So it should use `startsWith`  before replacement.

Thank you!

Karlinator:
Thank you! I agree on the first point, good spot!

I don't think the `startsWith` is actually necessary though. By definition, all paths should always start with the base path unless they are external asset paths. External asset paths are already covered by `!resolved.startsWith('/') || resolved.startsWith('//')` as you mention. `String.prototype.replace` only replaces the first occurrence. The scenario where it mis-replaces should never happen, and if it does then something else has already gone wrong.

baseballyama:
Oh right.
That is may not actually be necessary.

Thank you for the discussion!

sidharthv96:
Just tested this build in my project, the build works perfectly!
I'm able to deploy the built files and they work.

The only issue is that `npm run preview` is broken when the base path is set. All assets are 404.

But that is a smaller bug that the build not working, which was keeping [mermaid.live](https://mermaid.live) on v147 till now.

@benmccann if this can be merged, that blocker would be fixed :)

Karlinator:
Yup, I had the same sequence of events as you, that's why #2409 came about 😄 Fixed this and then "wait, what's up with `preview`?"

benmccann:
thanks for testing it

nikitaiavdeev:
`kit.paths.base` doesn't applies as a prefix to 'service-worker.js'. As a result it tries to register it from the root directory and fails.
`navigator.serviceWorker.register('/service-worker.js');`
instead of
`navigator.serviceWorker.register('kit.paths.base/service-worker.js');`

sidharthv96:
@nikitaiavdeev You might have to raise this as a new issue to keep track of it better.

PR #2409, opened 2021-09-12
[fix] strip `paths.base` before resolving assets in preview server.

Karlinator:
The preview server was implicitly assuming that asset URLs would never start with the base path. In fact, they _always_ do.

Vaguely related to #2230 (but this was a new issue I discovered and diagnosed while working on #2407).

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/preview/index.js
+++ b/packages/kit/src/core/preview/index.js
@@ -86,18 +86,16 @@ export async function preview({
 
 			const parsed = new URL(initial_url, 'http://localhost/');
 
-			const rendered =
-				parsed.pathname.startsWith(config.kit.paths.base) &&
-				(await app.render({
-					host: /** @type {string} */ (
-						config.kit.host || req.headers[config.kit.hostHeader || 'host']
-					),
-					method: req.method,
-					headers: /** @type {import('types/helper').RequestHeaders} */ (req.headers),
-					path: parsed.pathname.replace(config.kit.paths.base, ''),
-					query: parsed.searchParams,
-					rawBody: body
-				}));
+			const rendered = await app.render({
+				host: /** @type {string} */ (
+					config.kit.host || req.headers[config.kit.hostHeader || 'host']
+				),
+				method: req.method,
+				headers: /** @type {import('types/helper').RequestHeaders} */ (req.headers),
+				path: parsed.pathname.replace(config.kit.paths.base, ''),
+				query: parsed.searchParams,
+				rawBody: body
+			});
 
 			if (rendered) {
 				res.writeHead(rendered.status, rendered.headers);
@@ -120,6 +118,9 @@ export async function preview({
 				render_handler();
 			}
 		} else {
+			if (initial_url.startsWith(config.kit.paths.base)) {
+				req.url = initial_url.slice(config.kit.paths.base.length);
+			}
 			assets_handler(req, res, () => {
 				static_handler(req, res, render_handler);
 			});
--- a/packages/kit/src/core/preview/index.js
+++ b/packages/kit/src/core/preview/index.js
@@ -86,16 +86,18 @@ export async function preview({
 
 			const parsed = new URL(initial_url, 'http://localhost/');
 
-			const rendered = await app.render({
-				host: /** @type {string} */ (
-					config.kit.host || req.headers[config.kit.hostHeader || 'host']
-				),
-				method: req.method,
-				headers: /** @type {import('types/helper').RequestHeaders} */ (req.headers),
-				path: parsed.pathname.replace(config.kit.paths.base, ''),
-				query: parsed.searchParams,
-				rawBody: body
-			});
+			const rendered =
+				parsed.pathname.startsWith(config.kit.paths.base) &&
+				(await app.render({
+					host: /** @type {string} */ (
+						config.kit.host || req.headers[config.kit.hostHeader || 'host']
+					),
+					method: req.method,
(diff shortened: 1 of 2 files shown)


packages/kit/src/core/preview/index.js
```diff
@@ -120,6 +118,9 @@ export async function preview({
 				render_handler();
 			}
 		} else {
+			if (initial_url.startsWith(config.kit.paths.base)) {
+				req.url = initial_url.slice(config.kit.paths.base.length);
+			}
```
benmccann, line 123: should the `assets_handler` call on the list line be inside this `if` block to be consistent with the `if (initial_url.startsWith(SVELTE_KIT_ASSETS))` just above this?

packages/kit/src/core/preview/index.js
```diff
@@ -86,18 +86,16 @@ export async function preview({
 
 			const parsed = new URL(initial_url, 'http://localhost/');
 
-			const rendered =
-				parsed.pathname.startsWith(config.kit.paths.base) &&
```
benmccann, line 90: why remove this check?

Karlinator:
Sure thing!

Prettier is being weird for me right now. It doesn't agree with itself between the CI Lint and running `lint` or `format` locally, but I figured it out eventually.

packages/kit/src/core/preview/index.js
Karlinator: Ah, moving the response into the if just makes it never respond when a resource isn't found, instead of responding with a 404 (because there just is no handler given). 404s are handled inside the handlers.

packages/kit/src/core/preview/index.js
Karlinator: I honestly don't remember. Without this check, the app is available both at `/` _and_ at the base path, which probably isn't intended. I'll put it back.

Karlinator:
I've noticed static assets are avaliable both at `/base-path/_app/whatever` _and_ at `/_app/whatever`. I can add a check to reject with 404 all requests that don't start with the base path, but looking at this I suspect there are more errors in the if/else stuff around 110? Or at the very least it's confusing as it is now. I'll have a little more of a think.

packages/kit/src/core/preview/index.js
```diff
@@ -120,6 +118,9 @@ export async function preview({
 				render_handler();
```
benmccann, line 118: It's unrelated to your change, but I wonder what is happening here. it looks like it'd just go in an infinite loop? maybe I'm misreading something

benmccann:
Hmm. Honestly, I'd love to just get rid of all this stuff and call Vite's preview functionality. It looks like their `preview` method is not exposed though, so that would be the first step

packages/kit/src/core/preview/index.js
Karlinator: I've been thrown by the structure of this file before. At this point we are outside the declaration of `render_handler`, and back into the callback to `get_server` (which is the request handler). This section is just deciding what paths to strip in what ways in order to handle the base and asset paths correctly. In this case it calls `render_handler` without calling `assets_handler` and `static_handler` first.

No loops, but I don't particularly like it.

Karlinator:
I mean I don't disagree. But I think the base path stuff would still be a problem (we apparently have bugs in the dev server about this, see #2465 as you know). However, right now we are basically supporting two quite separate and different servers for preview and dev which isn't great.

benmccann:
Yeah, if there are issues in the dev server that's probably the more important one to fix since it's more used and will stick around longer. This maybe wouldn't be worth as much effort if the code were all about to be rewritten, but that would depend on https://github.com/vitejs/vite/pull/5014 and I don't know how long that'd take or if it will even be accepted, so probably worth getting this in still

packages/kit/src/core/preview/index.js
benmccann: nevermind. I missed some of the braces when reading the code the first time around to see where the methods started and ended

packages/kit/src/core/preview/index.js
benmccann: I just pushed a change to add it back

Karlinator:
I won't spend any more time trying to improve/clean up this until we know the outcome of vitejs/vite#5014 then. Hopefully we can just rip this all out soon-ish.

raduab:
I get the following error starting with version 171. I am using adapter-node:

```
> Using @sveltejs/adapter-node
 > .svelte-kit/output/server/chunks/app-0d4ff97d.js:1703:25: error: Could not resolve "./global-61357231.js"
    1703 │       load: () => import("./global-61357231.js")
         ╵                          ~~~~~~~~~~~~~~~~~~~~~~

 > .svelte-kit/output/server/chunks/__layout-048eba31.js:377:7: error: Could not resolve "./global-61357231.js"
    377 │ import("./global-61357231.js");
        ╵        ~~~~~~~~~~~~~~~~~~~~~~

> Build failed with 2 errors:
.svelte-kit/output/server/chunks/__layout-048eba31.js:377:7: error: Could not resolve "./global-61357231.js"
.svelte-kit/output/server/chunks/app-0d4ff97d.js:1703:25: error: Could not resolve "./global-61357231.js"
Error: Build failed with 2 errors:
.svelte-kit/output/server/chunks/__layout-048eba31.js:377:7: error: Could not resolve "./global-61357231.js"
.svelte-kit/output/server/chunks/app-0d4ff97d.js:1703:25: error: Could not resolve "./global-61357231.js"
```

PR #2413, opened 2021-09-12
[chore] expose Body generic to hook functions

### LINKED ISSUE #1215: Should ReadOnlyFormData type be exported?
Currently in the demo app, typescript complains about this method access:

https://github.com/sveltejs/kit/blob/399220dc71d88ff9f90bed27ac7ec6c9ad485579/packages/create-svelte/templates/default/src/routes/todos/index.json.ts#L25

```
Property 'get' does not exist on type 'BaseBody'.
  Property 'get' does not exist on type 'string'.
```

The `export` statement should probably start with

```typescript
export const post: RequestHandler<any, ReadOnlyFormData>
```

But that requires `ReadOnlyFormData` being able to be imported

**Additional context**

george-lim:
Closes #1215

[PR #1256](https://github.com/sveltejs/kit/pull/1256) allows us to avoid importing internal type `ReadOnlyFormData` for RequestHandler, since it has the `Body` generic. However, the hook functions don't expose the `Body` generic so we currently still need to import `ReadOnlyFormData` type if we want to process form data for post requests in our handle hooks.

This will enable users to pass in `FormData` for `Body` generic in GetSession, Handle, and HandleError.

image

### DIFF
--- a/packages/kit/types/hooks.d.ts
+++ b/packages/kit/types/hooks.d.ts
@@ -16,19 +16,19 @@ export interface ServerResponse {
 	body?: StrictBody;
 }
 
-export interface GetSession<Locals = Record<string, any>, Session = any> {
-	(request: ServerRequest<Locals>): MaybePromise<Session>;
+export interface GetSession<Locals = Record<string, any>, Body = unknown, Session = any> {
+	(request: ServerRequest<Locals, Body>): MaybePromise<Session>;
 }
 
-export interface Handle<Locals = Record<string, any>> {
+export interface Handle<Locals = Record<string, any>, Body = unknown> {
 	(input: {
-		request: ServerRequest<Locals>;
-		resolve(request: ServerRequest<Locals>): MaybePromise<ServerResponse>;
+		request: ServerRequest<Locals, Body>;
+		resolve(request: ServerRequest<Locals, Body>): MaybePromise<ServerResponse>;
 	}): MaybePromise<ServerResponse>;
 }
 
-export interface HandleError<Locals = Record<string, any>> {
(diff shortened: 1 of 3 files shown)


packages/kit/types/hooks.d.ts
```diff
@@ -16,19 +16,19 @@ export interface ServerResponse {
 	body?: StrictBody;
 }
 
-export interface GetSession<Locals = Record<string, any>, Session = any> {
-	(request: ServerRequest<Locals>): MaybePromise<Session>;
+export interface GetSession<Locals = Record<string, any>, Body = unknown, Session = any> {
```
ignatiusmb, line 19: Inserting the argument in the middle will be a breaking change, but it does make sense considering the usage order and keeps it consistent with the order from other types like `Load`. Let's make sure to also update the docs accordingly so users can crosscheck properly.

ignatiusmb (changes requested):
Thank you, can we also update them in the docs

packages/kit/types/hooks.d.ts
benbender: Alternatively, one could use variadic tuples if BC should be kept: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html

This would result in more work and bloat in the implementation, so I would personally vote against it. Just wanted to point out that it would be possible.

george-lim: @ignatiusmb I updated the docs to reflect the changes.
@benmccann I agree with you there, it's cool that it's possible but I don't think it's worth the bloat.

ignatiusmb (approved):

ignatiusmb:
thanks!

Issue #1491, opened 2021-05-19
Native support for web sockets


AndreasHald:
We've been attempting to migrate our app from sapper to sveltekit and ran into the problem of web sockets, in sapper we could simply attach the WebSocket server directly to the express server and be done with it, but as per Rich Harris' comments [here](https://github.com/sveltejs/kit/issues/334#issuecomment-804987028) and here [here](https://github.com/sveltejs/kit/issues/334#issuecomment-806289525) in sveltekit we don't want to expose direct access to the middlwares thus promoting the node adapter over the serverless ones.

However seeing as some of the serverless providers are starting to support web sockets ([cloudflare](https://blog.cloudflare.com/introducing-websockets-in-workers/)) ([begin](https://blog.begin.com/introducing-architect-5-0-fully-serverless-websockets-20bb97673c4e)) ([AWS Lambda](https://aws.amazon.com/blogs/compute/announcing-websocket-apis-in-amazon-api-gateway/)) should sveltekit provide a way to handle these connections directly? 

A native approach to this could also potentially provide a tool to solve this [issue](https://github.com/sveltejs/kit/issues/87) and [this](https://github.com/sveltejs/kit/issues/887)

**Describe the solution you'd like**
An interesting solution could be to expose functions in endpoints in the same way we handle http requests, and then let the adapters handle the rest

```routes/api/index.ts```
```js
// Handle post requests
export const post: RequestHandler = async (request) => {....}
// Handle web socket requests
export const ws: WebSocketRequestHandler = async (request) => {....}
```

A concern with this approach could be mixing protocols, within an endpoint could be confusing. Another could be if this becomes too complicated since as far as I know the initial [handshake](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#the_websocket_handshake) to start a WebSocket connection is handled over http, would this then be handled directly by sveltekit which could be considered too much "magic" since it would have to intercept a get request that would then not make it to the get handler in the endpoint. Or should we let the user implement the get endpoint and return the WebSocket pair and upgrade header.

**Describe alternatives you've considered**
The way we go about it today is a small script that manipulates the build output of sveltekit as such however this does not work in dev mode which is a pain point.
```js
const instance = createServer({ render: app.render }).listen(PORT, HOST, (err) => {
	if (err) {
		console.log('error', err);
	} else {
		console.log(`Listening on port ${PORT}`);
	}
});
app.server.installSubscriptionHandlers(instance);
[...]

PR #2414, opened 2021-09-13
[feat] adapter-node entryPoint option

benmccann:
@Xananax can you test and see if this solves all the problems you had creating a custom server? (https://github.com/sveltejs/kit/issues/1491#issuecomment-915495629)

### DIFF
--- a/.changeset/young-cougars-poke.md
+++ b/.changeset/young-cougars-poke.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-node': patch
+---
+
+[feat] add entryPoint option for custom servers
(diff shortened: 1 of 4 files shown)


Xananax:
That's definitely more elegant, thanks! Is there any example of accessing session details from that middleware? I can probably figure it out if not

benmccann:
If you put your own middleware before the SvelteKit middleware you wouldn't be able to access the SvelteKit session there because it wouldn't exist yet. If you put it after, you might be able to do it by copying info into the adapter-node `req.locals` or something, but that would depend on something like https://github.com/sveltejs/kit/pull/2359 being merged first

PR #2425, opened 2021-09-14
Add homepage to package.json

frederikhors:

### Tests

### Changesets


Conduitry:
Any changes to this file should be indented with tabs like the rest of the file, not with spaces.

I'm not sure how valuable of a change this is, but I would argue that, if we did add a homepage, it should be https://kit.svelte.dev/ - There is a separate `package.json` field for the repository URL.

frederikhors:
> Any changes to this file should be indented with tabs like the rest of the file, not with spaces.
> 
> I'm not sure how valuable of a change this is, but I would argue that, if we did add a homepage, it should be https://kit.svelte.dev/ - There is a separate `package.json` field for the repository URL.

Done.

packages/kit/package.json
```diff
@@ -1,6 +1,11 @@
 {
 	"name": "@sveltejs/kit",
 	"version": "1.0.0-next.166",
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/sveltejs/kit"
+	},
```
bluwy, line 7: We should also set `repository.directory` since this is a monorepo, so that relative links in a README are properly resolved. [Example](https://github.com/sveltejs/vite-plugin-svelte/blob/786c180f851b12042f53daef3c2bb2de4c744ad3/packages/vite-plugin-svelte/package.json#L35)

frederikhors: Done!

antony:
Where does this information appear? We have this info in the root package, and if we wanted to put it in sub-packages, we'd want to put it into all of them rather than just one.

dummdidumm:
This will appear on npm, there will be links to the homepage/github repo then, which I agree should be done for all packages. To me it's always a little strange to see a package on npm without any link to github, which makes me think it's closed source or something like that.

benmccann:
this change seems like a good idea to me, but agree that it should be done for everything under `packages/`

frederikhors:
> this change seems like a good idea to me, but agree that it should be done for everything under `packages/`

Done!

bluwy (approved):

packages/adapter-begin/package.json
```diff
@@ -1,6 +1,12 @@
 {
 	"name": "@sveltejs/adapter-begin",
 	"version": "1.0.0-next.13",
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/sveltejs/kit",
+		"directory": "packages/adapter-begin"
+	},
+	"homepage": "https://kit.svelte.dev",
```
benmccann, line 9:
```suggestion
	"homepage": "https://github.com/sveltejs/kit/tree/master/packages/adapter-begin",
```

packages/adapter-cloudflare-workers/package.json
```diff
@@ -1,6 +1,12 @@
 {
 	"name": "@sveltejs/adapter-cloudflare-workers",
 	"version": "1.0.0-next.21",
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/sveltejs/kit",
+		"directory": "packages/adapter-cloudflare-workers"
+	},
+	"homepage": "https://kit.svelte.dev",
```
benmccann, line 9:
```suggestion
	"homepage": "https://github.com/sveltejs/kit/tree/master/packages/adapter-cloudflare-workers",
```

packages/adapter-netlify/package.json
```diff
@@ -1,6 +1,12 @@
 {
 	"name": "@sveltejs/adapter-netlify",
 	"version": "1.0.0-next.30",
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/sveltejs/kit",
+		"directory": "packages/adapter-netlify"
+	},
+	"homepage": "https://kit.svelte.dev",
```
benmccann, line 9:
```suggestion
	"homepage": "https://github.com/sveltejs/kit/tree/master/packages/adapter-netlify",
```

benmccann:
I was starting to suggest that the `homepage` for the adapters should be their directory in GitHub since that's where the docs for them are, but I think I changed my mind on that because the README already gets published on npm, so there's no value in clicking through to GitHub to see it

Lgtm!

benmccann:
Is this supposed to be visible on npm? I don't see it https://www.npmjs.com/package/@sveltejs/adapter-static/v/1.0.0-next.19

bluwy:
It should be, but maybe it's because npm only parses the links based on the `latest` tag, not the `next` tag.

PR #2430, opened 2021-09-15
[feat] prefer functions to globs

### LINKED ISSUE #2374: Figure out a consistent globbing strategy
### Describe the problem

In #2247, we switched to using `micromatch` (from `globrex`) to determine which files are included/excluded when creating packages. But we're still using `globrex` via `tiny-glob` here: https://github.com/sveltejs/kit/blob/aad27f4ed3a2a57f68c6cb55076b9d3132bcbaac/packages/kit/src/core/create_manifest_data/index.js#L50-L58

As a result, our inclusion/exclusion logic is inconsistent, and we're bundling two libraries with overlapping functionality.

Personally I'd love to not use micromatch. It's a very big library for constructing regexes. If we do need to provide the extra flexibility, then should we take a leaf out of uvu's book and just accept regexes in the first place?

```js
export default {
  kit: {
    package: {
      exports: {
        exclude: /^internal/
      }
    }
  }
};
```

### Alternatives considered

Instead of regexes, a function?

```js
export default {
  kit: {
    package: {
      exports: file => !file.startsWith('internal')
    }
  }
};
```

### Importance

nice to have

### Additional Information

_No response_

ignatiusmb:
Resolves #2374


packages/kit/package.json
```diff
@@ -25,7 +25,6 @@
 		"kleur": "^4.1.4",
 		"locate-character": "^2.0.5",
 		"marked": "^2.0.3",
-		"micromatch": "^4.0.4",
```
dummdidumm, line 28: `@types/micromatch` can be removed, too.

dummdidumm (approved):
It's cool to see how these slight differences in handling files now all are harmonized into "this is a function and it returns whether or not to include the file".

packages/kit/package.json
ignatiusmb: Nice catch!

packages/kit/src/core/config/options.js
```diff
@@ -75,14 +75,8 @@ const options = object(
...
 				dir: string('package'),
-				exports: object({
-					include: array_of_strings(['**']),
-					exclude: array_of_strings(['**/_*'])
-				}),
-				files: object({
-					include: array_of_strings(['**']),
-					exclude: array_of_strings([])
-				}),
+				exports: fun((filepath) => !/^_|\/_|\.d\.ts$/.test(filepath)),
```
benmccann, line 78: should probably put the same comment I suggested above here as well

documentation/docs/14-configuration.md
```diff
@@ -132,8 +126,25 @@ Options related to [creating a package](#packaging).
...
-- `exports` - contains an `includes` and an `excludes` array which specifies which files to mark as exported from the `exports` field of the `package.json`. Will merge existing values if available with values from `package.json` taking precedence
-- `files` - contains an `includes` and an `excludes` array which specifies which files to process and copy over when packaging
+- `exports` - a function with the type of `(filepath: string) => boolean`. When `true`, the following export will be included in the `exports` field of the `package.json`. Any existing values in the `package.json` source will be merged with values from the original `exports` field taking precedence
+- `files` - a function with the type of `(filepath: string) => boolean`. When `true`, the following file will be processed and copied over to the final output folder, specified in `dir`
```
benmccann, line 130: this seems weird. `micromatch` won't necessarily be installed by the user. what's the reason for this?

documentation/docs/14-configuration.md
```diff
@@ -132,8 +126,25 @@ Options related to [creating a package](#packaging).
 
 - `dir` - output directory
 - `emitTypes` - by default, `svelte-kit package` will automatically generate types for your package in the form of `d.ts.` files. While generating types is configurable, we believe it is best for the ecosystem quality to generate types, always. Please make sure you have a good reason when setting it to `false` (for example when you want to provide handwritten type definitions instead)
-- `exports` - contains an `includes` and an `excludes` array which specifies which files to mark as exported from the `exports` field of the `package.json`. Will merge existing values if available with values from `package.json` taking precedence
-- `files` - contains an `includes` and an `excludes` array which specifies which files to process and copy over when packaging
+- `exports` - a function with the type of `(filepath: string) => boolean`. When `true`, the following export will be included in the `exports` field of the `package.json`. Any existing values in the `package.json` source will be merged with values from the original `exports` field taking precedence
```
benmccann, line 129: what does "the following export" refer to?

benmccann (commented):
looks like this will need a minor rebase

ignatiusmb:
The micromatch example is a reference from https://github.com/sveltejs/kit/issues/2374#issuecomment-914416495

benmccann:
@ignatiusmb it looks like this PR will need a rebase

documentation/docs/14-configuration.md
```diff
@@ -132,8 +127,26 @@ Options related to [creating a package](#packaging).
...
-- `exports` - contains an `includes` and an `excludes` array which specifies which files to mark as exported from the `exports` field of the `package.json`. Will merge existing values if available with values from `package.json` taking precedence
-- `files` - contains an `includes` and an `excludes` array which specifies which files to process and copy over when packaging
+- `exports` - a function with the type of `(filepath: string) => boolean`. When `true`, the filepath will be included in the `exports` field of the `package.json`. Any existing values in the `package.json` source will be merged with values from the original `exports` field taking precedence
+- `files` - a function with the type of `(filepath: string) => boolean`. When `true`, the file will be processed and copied over to the final output folder, specified in `dir`
+
+```js
+// svelte.config.js
+// example usage with a globbing library
```
benmccann, line 135: I think I missed this comment earlier and I thought this was demonstrating a default or something. I somewhat question if we need this example, but if we want to keep it, I'd probably remove this comment and instead put it in the prose above like `For advanced file matching you can use these options in conjunction with a globbing library:`

documentation/docs/14-configuration.md
```diff
@@ -32,14 +32,8 @@ const config = {
...
 			emitTypes: true,
-			exports: {
-				include: ['**'],
-				exclude: ['**/_*']
-			},
-			files: {
-				include: ['**'],
-				exclude: []
-			}
+			exports: (filepath) => !/^_|\/_|\.d\.ts$/.test(filepath),
```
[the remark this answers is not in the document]
jiv-e: Should we still include the word "exclude" or "excludes" in the property name?
Just using exports doesn't make sense in my opinion. I would prefer more explicit naming.

What about "exportExcludes"?

Or for future additions to exports related configuration maybe still use the nested property structure like this:
```
package: {
  ...
  exports: {
    // Function which excludes all .d.ts and files starting with _ as the name from the exports.
    exclude: (filepath) => !/^_|\/_|\.d\.ts$/.test(filepath),
```

The second option would also feel natural to read and understand in the code:
```
if (config.kit.package.exports.exclude(normalized))
```

jiv-e (changes requested):

documentation/docs/14-configuration.md
```diff
@@ -181,7 +195,7 @@ Enables or disables the client-side [router](#ssr-and-javascript-router) app-wid
 
 An object containing zero or more of the following values:
 
-- `exclude` - an array of glob patterns relative to `files.assets` dir. Files matching any of these would not be available in `$service-worker.files` e.g. if `files.assets` has value `static` then ['og-tags-images/**/*'] would match all files under `static/og-tags-images` dir.
+- `files` - a function with the type of `(filepath: string) => boolean`. When `true`, the following file will be available in `$service-worker.files`, otherwise it will be excluded.
```
benmccann, line 198:
```suggestion
- `files` - a function with the type of `(filepath: string) => boolean`. When `true`, the given file will be available in `$service-worker.files`, otherwise it will be excluded.
```

Conduitry:
This missed one of the references to `config.kit.serviceWorker.exclude` in https://kit.svelte.dev/docs#modules-$service-worker

We should get that fixed and get this released quickly, because anytime we merge PRs that update the docs, the site is redeployed right away, not when we release the feature to npm.

benmccann:
Thanks for catching that. PR here: https://github.com/sveltejs/kit/pull/2485

PR #2431, opened 2021-09-15
[fix] add svelte field when packaging

### LINKED ISSUE #1959: Add svelte exports field to package json when running package command
### Describe the problem

SvelteKit currently has a heuristic in place which tells Vite which packages are Svelte packages (for noExternal stuff) by looking up the `package.json` content and seeing if there's a `svelte` property.

The `svelte-kit package` command does not currently set this field.

Use the same logic that is used for the `main` entry to produce a `svelte` entry.

### Alternatives considered

A workaround for now is for users to specify this field in their `package.json` themselves as it's copied over as of lately.

### Importance

would make my life easier

### Additional Information

The `svelte` key is likely more of a historical thing at this point, and deserves some more thought regarding the new developments, but for now the quick fix is to just add this field

dummdidumm:
Several heuristics in Kit/vite-plugin-svelte to tell Vite to mark Svelte packages rely on the "svelte" property. Vite/Rollup/Webpack plugin can all deal with it.

Fixes #1959

@dominikg @bluwy  I didn't quite follow through your discussion over at `vite-plugin-svelte`, but I think this change is needed for Svelte packages to properly get detected, right?

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/packaging/index.js
+++ b/packages/kit/src/packaging/index.js
@@ -112,19 +112,24 @@ export async function make_package(config, cwd = process.cwd()) {
 		// rely on the "svelte" property. Vite/Rollup/Webpack plugin can all deal with it.
 		// See https://github.com/sveltejs/kit/issues/1959 for more info and related threads.
 		if (pkg.exports['.']) {
-			const svelte_export = typeof pkg.exports['.'] === 'string' ? pkg.exports['.'] : undefined;
+			const svelte_export =
+				typeof pkg.exports['.'] === 'string'
+					? pkg.exports['.']
+					: pkg.exports['.'].import || pkg.exports['.'].default;
 			if (svelte_export) {
 				pkg.svelte = svelte_export;
 			} else {
 				console.warn(
-					'The "." entry in "exports" is not a string. ' +
+					'Cannot generate a "svelte" entry point because ' +
+						'the "." entry in "exports" is not a string. ' +
 						'If you set it by hand, please also set one of the options as a "svelte" entry point'
 				);
 			}
 		} else {
 			console.warn(
-				'The "." entry in "exports" is not a string. ' +
-					'If you set it by hand, please also set one of the options as a "svelte" entry point'
+				'Cannot generate a "svelte" entry point because ' +
+					'the "." entry in "exports" is missing. ' +
+					'Please specify one or set a "svelte" entry point yourself'
 			);
 		}
 	}
--- a/packages/kit/src/packaging/index.js
+++ b/packages/kit/src/packaging/index.js
@@ -104,6 +104,27 @@ export async function make_package(config, cwd = process.cwd()) {
 	}
 
 	pkg.exports = { ...generated, ...pkg.exports };
+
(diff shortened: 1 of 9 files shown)


dominikg:
yes, thats right. the svelte field is relied upon. We should also make sure that exports contains
`"./package.json":"./package.json"`, this is important for both rollup-plugin-svelte and vite-plugin-svelte to be able to read the svelte field.

packages/kit/src/packaging/index.js
```diff
@@ -104,6 +104,27 @@ export async function make_package(config, cwd = process.cwd()) {
...
+		} else {
+			console.warn(
+				'The "." entry in "exports" is not a string. ' +
+					'If you set it by hand, please also set one of the options as a "svelte" entry point'
+			);
+		}
+	} else {
+		console.warn(
+			'The "." entry in "exports" is not a string. ' +
+				'If you set it by hand, please also set one of the options as a "svelte" entry point'
```
ignatiusmb, line 124: We might want to throw an error for this one

```suggestion
			'The "." entry in "exports" does not exists. ' +
				'Please specify one or set a "svelte" entry point yourself'
```

dummdidumm: Why error in this case? It could be 

```
"." {  "import": "./index.js" }
```

dummdidumm:
> yes, thats right. the svelte field is relied upon. We should also make sure that exports contains
> `"./package.json":"./package.json"`, this is important for both rollup-plugin-svelte and vite-plugin-svelte to be able to read the svelte field.

This is already the case 👍

dominikg:
actually the "svelte" field is only needed packages that export ".svelte" components. Utilities that come without any components, eg just some fancy store or context utility can still ship without a "svelte" field.
So maybe a warning to add it would be better, or somehow find out if any .svelte is exported?

Or is it implied that package is always used for ui, as utilities generally would not need to go through preprocess + compile ?

packages/kit/src/packaging/index.js
ignatiusmb: Looks like it isn't that clear from this comment view, but this is for the outer `else` block where there's no `"."` export as well. Also, yes, never mind about throwing the error.

bluwy:
Yeah @dominikg's comment are spot on. The `svelte` field is used to workaround Vite's prebundling process, and is only needed if they export Svelte components. I don't think Kit reads the `svelte` field anymore though, the logic has been moved to `vite-plugin-svelte`.

packages/kit/src/packaging/index.js
```diff
@@ -104,6 +106,29 @@ export async function make_package(config, cwd = process.cwd()) {
 	}
 
 	pkg.exports = { ...generated, ...pkg.exports };
+
+	if (contains_svelte_files) {
```
bluwy, line 110: Should we ignore if the user already set the `pkg.svelte` field?
```suggestion
	if (!pkg.svelte && contains_svelte_files) {
```

dominikg (approved):

bluwy (approved):
LGTM. I think we should apply @ignatiusmb's suggestion too so we have two different warnings for the two cases.

Issue #2401, opened 2021-09-11
@ not working in route paths


SleepingRobot:
### Describe the bug

Routes containing `@` appear to have been broken in `@sveltejs/kit 1.0.0-next.163` and result in a 404 error. 162 is working as expected, 163-165 reproduce the issue.

### Reproduction

```
npm init svelte@next
npm install
```

create file `src/routes/@test.svelte` with the following contents:
```
<h1>@test</h1>
```

run `npm run dev`

browsing to this url `http://localhost:3000/@test` displays `404 Not found: /@test` error page.

changing `@sveltejs/kit` from `next` to `1.0.0-next.162` in `package.json`, and then running `npm i && npm run dev` results in the `@test` page properly loading.

### System Info

```shell
System:
    OS: Linux 5.11 Ubuntu 21.04 (Hirsute Hippo)
    CPU: (4) x64 Intel(R) Core(TM) i5-4690K CPU @ 3.50GHz
    Memory: 7.19 GB / 15.47 GB
    Container: Yes
    Shell: 5.1.4 - /bin/bash
  Binaries:
    Node: 16.9.1 - ~/.n/bin/node
    Yarn: 1.22.5 - /usr/bin/yarn
    npm: 7.23.0 - ~/.npm-global/bin/npm
  Browsers:
    Chrome: 93.0.4577.63
    Firefox: 92.0
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.165 
    svelte: ^3.34.0 => 3.42.5
```

### Severity

serious, but I can work around it

### Additional Information

I am using a `src/routes/@[username]` path in my application for user profile pages, and they are all resulting in 404 errors now. I can technically work around the issue by refactoring my app to use a different user profile path, but I'd really like to avoid that if possible.

benbender:
Seeing the same and should be related to #2354. My investigation in dev reveals the following:

**route with "@":**

page-object: 
```
{
  host: 'localhost:3000',
  path: '/users/@bb',
  query: URLSearchParams {},
  params: {}
}
```

manifest-entry:
```
// src/routes/users/@[username]/index.svelte
[/^\/users\/%40([^/]+?)\/?$/, [c[0], c[7]], [c[1]], (m) => ({ username: d(m[1])})],
```


**Without "@" in the path:**

page-object:
```
{
  host: 'localhost:3000',
  path: '/users/bb',
  query: URLSearchParams {},
  params: { username: 'bb' }
}
```

manifest-entry:
```
[...]

Issue #1746, opened 2021-06-23
% cannot be escaped in url params


andykais:
**Describe the bug**
parameterized routes do not properly parse the `%` sign in a url.

If I have a url that contains a percent that needs to be escaped, I would use the function `encodeUri`. For example:
```js
encodeURI('/sample/test%20me') // returns "/sample/test%2520me"
```
What actually happens, is svelte kit reads this url as `test me`

**Logs**

**To Reproduce**
```bash
npm init svelte@next
npm install
```
create the file `src/routes/sample/[param].svelte` with the following contents:
```js
<script context="module">
  export function load({ page }) {
    const { param } = page.params
    return {
      props: { param }
    }
  }
</script>

<script>
  export let param
</script>
param is: "{param}"
```

opening this url: `http://localhost:3000/sample/test%2520me` displays `param is "test me"`

**Expected behavior**
the page should display `param is test%20me`.

**Stacktraces**

**Information about your SvelteKit Installation:**

<details>
  <summary>Diagnostics</summary>
System:
    OS: Linux 5.12 Arch Linux
    CPU: (4) x64 Intel(R) Core(TM) i5-6300U CPU @ 2.40GHz
    Memory: 5.48 GB / 18.96 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 16.3.0 - /usr/local/bin/node
    Yarn: 1.22.4 - ~/.npm-packages/bin/yarn
    npm: 7.11.2 - ~/.npm-packages/bin/npm
  Browsers:
    Firefox: 89.0.1
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.116 
    svelte: ^3.34.0 => 3.38.3 

- Your browser: `Google Chrome 91.0.4472.114 `
</details>

**Severity**
I have a local web app that displays system files using a route like `src/routes/file/[...filepath].svelte`. For whatever reason, some files happen to contain %20 inside them (likely because they were encoded weirdly once upon a time. With this bug as it stands, I cannot display _any_ files that contain percents inside them, because I have no way of telling the server that a percent is part of the filepath.

**Additional context**
I found a possibly related issue in the polka server repo: https://github.com/lukeed/polka/issues/119 (I assume polka is still the internal server)

benmccann:
[...]

PR #2435, opened 2021-09-15
[fix] revert #2354 and fix double character decoding a different way

benmccann:
Fixes https://github.com/sveltejs/kit/issues/2401

Reverts most of #2354, but leaves the test for https://github.com/sveltejs/kit/issues/1746

### DIFF
--- a/packages/kit/src/core/dev/index.js
+++ b/packages/kit/src/core/dev/index.js
@@ -231,15 +231,30 @@ function get_params(array) {
 	// src/routes/[x]/[y]/[z]/svelte, create a function
 	// that turns a RegExpExecArray into ({ x, y, z })
 
+	// input has already been decoded by decodeURI
+	// now handle the rest that decodeURIComponent would do
+	const d = /** @param {string} s */ s => s
+		.replace(/%23/g, '#')
+		.replace(/%3[Bb]/g, ';')
+		.replace(/%2[Cc]/g, ',')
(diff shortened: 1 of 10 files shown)


packages/kit/src/core/dev/index.js
```diff
@@ -231,15 +231,31 @@ function get_params(array) {
 	// src/routes/[x]/[y]/[z]/svelte, create a function
 	// that turns a RegExpExecArray into ({ x, y, z })
 
+	// input has already been decoded by decodeURI
+	// now handle the rest that decodeURIComponent would do
+	const d = /** @param {string} s */ (s) =>
```
dummdidumm, line 236: This looks the same like in `build` - maybe put it into a common file for both to use? `utils/encoding.js` or something like that.

benmccann: Yeah, it is the same, but I'm not sure it's that easy to share code between the two because the one in `build` is a string that gets written out to a file and isn't normal code. I think I could still do it, but would probably need to setup a new file to be output in the Rollup config as well. I'm just not sure it's worth it at the moment given all the other stuff there is to do though I do agree it'd be nicer

benmccann:
I'm going to go ahead and merge this since we just got another report of it in https://github.com/sveltejs/kit/issues/2401

benmccann:
I'm going to go ahead and merge this since we just got another report of it in https://github.com/sveltejs/kit/issues/2451

PR #2437, opened 2021-09-16
[fix] provide valid value for `letter-spacing` CSS property in template

Grawl:


bluwy:
Context: `%` isn't valid for `letter-spacing`. Perhaps we could remove that line altogether so the style would be the same as before? A before and after screenshots would help too.

Grawl:
why not

bluwy:
Your fix might work too, but it's hard to tell how it would look like unless we spin up the template locally, so a screenshot could help speed up the review process.

Grawl:
I set `0.1em` because I guess it was planned with some space between letters in header nav.

With `letter-spacing: 0.1em`:

localhost_27565_

With `letter-spacing: 1%`:

localhost_27565_ (1)

(not working `letter-spacing` rule)

bluwy:
Thanks for the screenshots! I think the new letter-spacing indeed looks nicer, :+1: from me

bluwy (approved):

benmccann:
Thanks @Grawl for the change and @bluwy for the review!

PR #2439, opened 2021-09-16
[breaking] rename context param of load to stuff

### LINKED ISSUE #984: Rename `context` to `stuff`

dummdidumm:
Closes #984

### Tests

### Changesets


Conduitry:
Do we have the ability to do that in dev mode only?

Alternatively, is there a reasonable way to check for that only during SSR? It _is_ a nice warning to be able to give people, but I don't want to be sending extra bytes to browsers.

Also alternatively, should it always be a warning/error to return _any_ unknown keys from the `load` function, long-term, not just this one that we're renaming?

benmccann:
Yeah, we can do it dev mode only. Check the mode from `$env` or `import.meta.env.DEV` and Rollup/Vite will remove it from production build

packages/kit/src/runtime/load.js
```diff
@@ -52,5 +54,12 @@ export function normalize(loaded) {
 		}
 	}
 
+	if (dev && /** @type {any} */ (loaded).context) {
```
dummdidumm, line 57: Added the warning. Using `dev` from `./app/env` here, which wasn't used anywhere else yet (only for user-facing-code), hope that doesn't have any bad implications.

dummdidumm: Well, it does apperently, tests are failing now because `env` is not set in tests. What now?

dummdidumm:
Can we safely use `import.meta.env` in all Kit code that is run by the user? If so, is the failure of the tests only a hint for us to provide these `import.meta.env` definitions?

packages/kit/src/runtime/load.js
bluwy: Would `context` still work in prod? If not I think we should make a hard error here.

But regarding the error, I believe it's because `import.meta.env` is non-standard in nodejs so it would resolve to `undefined`, hence the error. A workaround is to shim it with `import.meta.env = { /* ... */ }` at the top of this file, though we need to make sure we do that for tests only. Using `process.env.NODE_ENV` may be fine for this as well.

benmccann: I'd expect it to be defined in the integration tests. I think the issue is around this being unit tested. Rather than modify this to shim it, I wonder if we could move the check to another location that's called in integration tests, but not unit tests.

dummdidumm: This sounds wrong to me. I picked this method as it seems the central location to validating the load response. Also we probably don't have this check in for long, we could remove it at 1.0 and before that shimming it sounds like a pragmatic solution in the meantime

benmccann: Ok. That makes sense

dummdidumm:
I failed to come up with a way to set this that is both tree shakeable and doesn't require us to introduce esm import mocking into the unit tests, so I just left the error in - it will be shipped now every time (sorry @Conduitry ). I think this is okay since we will throw it out in a couple of weeks anyway.

packages/kit/src/runtime/server/page/respond.js
```diff
@@ -185,11 +185,11 @@ export async function respond(opts) {
 				}
 			}
 
-			if (loaded && loaded.loaded.context) {
+			if (loaded && loaded.loaded.stuff) {
 				// TODO come up with better names for stuff
```
Conduitry, line 189: Can we get rid of this TODO finally? :)

Were there other things that we still wanted to find better names for pre-1.0?

benmccann: Yeah, I think we should get rid of the TODO :smile:

Issue #948, opened 2021-04-10
`svelte-kit dev` does not recognize occupied ports on Windows


emme1444:
**Describe the bug**
When running `svelte-kit dev`, on Windows, SvelteKit does not throw an error for an already occupied port. Instead what happens is SvelteKit prints the "welcome" message, when trying to reach the started dev server it is of course shadowed by the process already occupying that port.

**To Reproduce**

On a Windows machine:
1. Make sure some service is running on port `3000`.
2. Make a new directory and `cd` into it.
2. `npm init svelte@next`
3. `npm i`
4. `npm run dev` ⬅ This should throw, but doesn't!

**Expected behavior**

An error should be thrown to indicate that the port the dev server is trying to launch on is already in use by another process.

**Information about your SvelteKit Installation:**

<details>
  <summary>Diagnostics</summary>

- The output of `npx envinfo --system --npmPackages svelte,@sveltejs/kit,vite --binaries --browsers`
```
  System:
    OS: Windows 10 10.0.19042
    CPU: (4) x64 Intel(R) Core(TM) i5-4460  CPU @ 3.20GHz 
    Memory: 16.23 GB / 23.92 GB
  Binaries:
    Node: 14.15.4 - C:\Program Files\nodejs\node.EXE      
    Yarn: 1.22.10 - ~\AppData\Roaming\npm\yarn.CMD        
    npm: 6.14.10 - C:\Program Files\nodejs\npm.CMD        
  Browsers:
    Chrome: 89.0.4389.114
    Edge: Spartan (44.19041.423.0), Chromium (89.0.774.75)
    Internet Explorer: 11.0.19041.1
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.71
    svelte: ^3.29.0 => 3.37.0
    vite: ^2.1.0 => 2.1.5
```
</details>

**Severity**

Mildly annoying, but for many users nonetheless.

**Additional context**
I poked around in the source for the `svelte-kit` cli. I noticed that you guys use the `port-authority` package's `blame` function. According to [the `port-authority` documentation](https://www.npmjs.com/package/port-authority#additional-functions), under _Additional functions_, this function only supports Unix-like operating systems.

Maybe a better fix is for this, would be to add support for this in `port-authority`, just thought this should be tracked.

Thanks for this awesome framework!

Conduitry:
[...]

PR #2441, opened 2021-09-16
[feat] Use dynamic port if specified port is already in use

baseballyama:
This is related to https://github.com/sveltejs/kit/issues/948

## Summary of the PR

At least on Windows, when I execute `npm run dev` like the above issue's reproduction steps, I got the below error if 3000 port is already in use.
```
Error: listen EADDRINUSE: address already in use 127.0.0.1:3000
```

I think this is correct behavior.
But on the other hand, [webpack-dev-server](https://github.com/webpack/webpack-dev-server) has the feature that dev server finds free port if a default port is already in use.
https://github.com/webpack/webpack-dev-server/pull/685/files

So I implemented almost the same feature.
If a user specifies a specific port, dev server will check only the port.
But if not, dev server searches usable port from 3000.

### Questions

- Is there any issue with this PR?
- Should I add some tests? (But honestly, I don't know how to write tests for this PR...)
- Should I add/edit some documents?

Thank you!!

--- 

### Tests

### Changesets

### DIFF
--- a/.changeset/tiny-keys-eat.md
+++ b/.changeset/tiny-keys-eat.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Use dinamic port if specified port is already in use
--- a/packages/kit/package.json
+++ b/packages/kit/package.json
@@ -11,6 +11,7 @@
 	"dependencies": {
 		"@sveltejs/vite-plugin-svelte": "^1.0.0-next.24",
 		"cheap-watch": "^1.0.3",
+		"portfinder": "^1.0.28",
 		"sade": "^1.7.4",
 		"vite": "^2.5.7"
 	},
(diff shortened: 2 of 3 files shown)


bluwy:
We should be able to easily craft our own port finding logic too with the current `port-authority` package. Don't think we need the `portfinder` package either.

baseballyama:
@Conduitry 

Thank you for telling the background.
So I felt that the existing implementation has intention.
So @bluwy gave me the alternative idea but should I spend more time on it?
Or should I close it at least for now?

> adding dependencies is something that should always be looked at very carefully.

Agree with you. and sorry for that. 
(Maybe it was a little bit casual...)

benmccann:
This functionality is already available in Vite with the `strictPort: false` option: https://vitejs.dev/config/#server-strictport

I think a better solution to this problem would be to merge https://github.com/sveltejs/kit/pull/2232, so that the Vite server options take effect, which would remove a lot of confusion that people have about those options not doing anything today

PR #2446, opened 2021-09-16
[docs] rename troubleshooting docs section and link more resources

benmccann:
"Troubleshooting" is something you'd only look at if something is going wrong. But we should help people who just don't know how to do something without experiencing a specific error message, etc. so I think "Additional Resources" is a better name that more people will look at. It also allows us to add more types of links to the section to provide more pointers to helpful resources


dummdidumm (commented):
I think @babichjacob expressed concerns about making Svelte Adders too public yet as it's not stable yet. Don't know if that information is up to date.

babichjacob:
> I think @babichjacob expressed concerns about making Svelte Adders too public yet as it's not stable yet. Don't know if that information is up to date.

You're close! My concern is about it being printed out in the project initializer output (`npm init svelte@next`) rather than the official docs or official FAQ being pointed to.

Here, it *may* be enough to only have the FAQ section because `svelte-add` is discoverable from there.

documentation/docs/90-additional-resources.md
```diff
@@ -0,0 +1,17 @@
...
+
+### FAQs
+
+Please see the [SvelteKit FAQ](/faq) for solutions to common issues and helpful tips and tricks.
+
+The [Svelte FAQ](https://svelte.dev/faq) and [`vite-plugin-svelte` FAQ](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md) may also be helpful for questions deriving from those libraries.
+
+### Integrations
+
+[Svelte Adders](https://sveltesociety.dev/templates#category-Svelte%20Add) allow you to setup many different integrations with a single command. Please see [sveltesociety.dev](https://sveltesociety.dev/) for a full listing of templates, components, and tools available for use with Svelte and SvelteKit. The SvelteKit FAQ also has a [section on integrations](faq#integrations), which may be helpful if you run into any issues.
```
bluwy, line 13:
```suggestion
[Svelte Adders](https://sveltesociety.dev/templates#category-Svelte%20Add) allow you to setup many different integrations with a single command. Please see [sveltesociety.dev](https://sveltesociety.dev/) for a full listing of templates, components, and tools available for use with Svelte and SvelteKit. The SvelteKit FAQ also has a [section on integrations](/faq#integrations), which may be helpful if you run into any issues.
```

Maybe `/faq` (with a leading slash) just in case,

benmccann:
I removed it from the FAQ to remove any possible duplication. I think setting up integrations is common enough that we should give some basic pointers in the main docs like mentioning `svelte-preprocess` there

bluwy (approved):

Issue #2434, opened 2021-09-15
Upgrade to `node-fetch` 3.0 final


benmccann:
### Describe the problem

we're on a beta version

### Describe the proposed solution

I tried briefly in https://github.com/sveltejs/kit/pull/2422, but ran into some issue. I'd love if someone were able to take a closer look and figure out what's going on

### Importance

nice to have

dominikg:
I tried to look into this today:
node-fetch updated to pure esm with 3.0.0. But their [package.json](https://github.com/node-fetch/node-fetch/blob/main/package.json)  looks a bit strange. 
It only has a "main" field pointing at src/index.js   no "exports", no "module". 
According to [node docs](https://nodejs.org/api/packages.html#packages_main), main is for require, so cjs. 

I'm not entirely sure if/how that trips up kits build, but its very odd.

dominikg:
tried it with a local node-fetch where i added "module" and "exports", no luck. There is a strange import to "streams/web" in `kit/dist/chunks/index5.js` which is experimental and only available in node16. And what can i say, when i use the edited node-fetch and node16, i can build the hn example ...

If someone has an idea whats causing that import, i'd be glad to hear it.

benmccann:
It looks like they tried to pollyfill it:
https://github.com/sveltejs/kit/pull/2422/files#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR4090

I'm not sure why that isn't working

dominikg:
the code that tries it is in `fetch-blob` , streams.cjs

```js
if (!globalThis.ReadableStream) {
  try {
    Object.assign(globalThis, require('stream/web'))
  } catch (error) {
		// TODO: Remove when only supporting node >= 16.5.0
    Object.assign(globalThis, require('web-streams-polyfill/dist/ponyfill.es2018.js'))
  }
}
```

dominikg:
aaaand there is a bug report on fetch-blob already https://github.com/node-fetch/fetch-blob/issues/117

winston0410:
> tried it with a local node-fetch where i added "module" and "exports", no luck. There is a strange import to "streams/web" in `kit/dist/chunks/index5.js` which is experimental and only available in node16. And what can i say, when i use the edited node-fetch and node16, i can build the hn example ...
> 
> If someone has an idea whats causing that import, i'd be glad to hear it.

I tried building in node16 in local and node14.x in vercel, still the same error for not finding stream.

PR #2447, opened 2021-09-16
chore: update dependencies

dominikg:
This PR updates all dependencies in all packages to latest available versions, with the exception of `node-fetch` which is pinned to 3.0.0-beta.9 to avoid the current issue with 3.0.0 final (#2434)

Notable updates are rollup plugins: commonjs  17 -> 20 and nodeResolve  11 -> 13
Also updated are all versions in templates of create-svelte

The changeset includes adapter-node, as that uses rollup plugin commonjs in its own build

### Tests

### Changesets


dominikg:
triggered another testrun as windows fail might have been a fluke (timeout)

benmccann (approved):

dummdidumm (approved):

PR #2449, opened 2021-09-17
Fix typo in appendix

learnitmyway:


dummdidumm:
The sentence is correct as is. It explains that SSG has the advantage that you _don't_ need to do SSR (Server Side Rendering) on every page request.

learnitmyway:
Ah I see. In that case I would suggest re-wording the sentence incase I'm not the only one who gets confused, but then again I might be the only one 😅 .

Issue #1950, opened 2021-07-19
Resolve aliases while running `svelte-kit package`


kwangure:
### Describe the problem

Aliases are currently not resolved when packaging since components are only preprocessed. 

For example, importing `$lib/css/shared.css` from `$lib/components/Component/SubComponent` involves a lot of `../` without aliases. This is very brittle especially considering something like `$lib/css/shared.css` is something that might be shared across many components at different nested levels.

### Describe the proposed solution

It would be helpful if `svelte-kit` resolved these aliases.

My proposal around Svelte files would be to create a built-in preprocessor that uses `plugin-alias` since this is already `svelte-kit` dependency.

It's not immediately clear what the approach to resolving aliases in js files should be because `svelte-kit` currently avoids any bundling and just copies them. I haven't used Typescript, so I don't know that `tsc` aliases are compatible with `plugin-alias`.

### Alternatives considered

Not using aliases in any files that will be packaged (my current approach).

### Importance

would make my life easier

dummdidumm:
Since TS does not do the resolving itself, we have to implement/add it ourselves somehow.
Options:
- use https://www.npmjs.com/package/tsconfig-paths somehow. This would read the paths from the tsconfig and resolve them. This would add another dependency required to install (~90kb). This would also be somewhat inconsistent because it would deal with aliases by reading the tsconfig, not by reading the Vite config.
- access the `resolve.alias` config of Vite to get the aliases. When preprocessing TS, it might be possible to write a transformer which will resolve these. When not preprocessing Svelte, I'm not sure how to do this other than either also running it through TS or with a hacky regexp.

jonathangreenemeier-vizio:
I wasn't super happy with the potential workaround of removing all the `$lib/` imports so wrote a little script to rewrite these imports in the packaged code.  I thought I would share this in case anyone else finds it useful while waiting for `svelte-kit package` to become less experimental/new: https://github.com/6eDesign/svelte-calendar/blob/master/scripts/postpackage.js

I'm running this as a `postpackage` npm script (where a `package` script simply invokes `svelte-kit package`).  If you are using a non-default output directory (something other than `package/`) you can specify that folder when invoking the script, eg: `PACKAGE_FOLDER=my-package-folder node scripts/postpackage`

[...]

PR #2453, opened 2021-09-17
[fix] resolve $lib alias when packaging

dummdidumm:
Fixes the 90% case of #1950

### Tests

### Changesets


packages/kit/src/packaging/index.js
```diff
@@ -144,11 +153,44 @@ export async function make_package(config, cwd = process.cwd()) {
...
+	 * @param {string} import_path
+	 */
+	const replace_import_path = (match, _, import_path) => {
+		if (!import_path.startsWith('$lib/')) {
+			return match;
+		}
+
+		const full_path = path.join(config.kit.files.lib, file);
+		const full_import_path = path.join(config.kit.files.lib, import_path.slice('$lib/'.length));
+		let resolved = path.relative(path.dirname(full_path), full_import_path).replace(/\\/g, '/');
```
ignatiusmb, line 185: What happens when `full_import_path` doesn't actually exists, for whatever reasons it may be (most likely a bad copy-paste or refactor)?

`path.relative` won't throw as long as it's a string, should we throw an error, leave it with the original alias in place, or use the non-existent resolved path (currently the case)?

packages/kit/src/packaging/index.js
```diff
@@ -144,11 +153,44 @@ export async function make_package(config, cwd = process.cwd()) {
...
+		resolved = resolved.startsWith('.') ? resolved : './' + resolved;
```
ignatiusmb, line 186: Maybe inline this ternary below to keep `resolved` as `const`?

ignatiusmb (commented):
I'm a bit concerned about using string replacement for resolving module aliases, I know bundlers like rollup or esbuild handle these stuff differently, there must be a reason a separate plugin for alias and replace exists in rollup, or separate options in esbuild. One of the case I can think of is when a user have a code block with an import path that uses `$lib`, it will match the pattern and replace the path when it's not supposed to do so.

With that being said, I agree this should be sufficient for now while we try to figure out more ways to tackle this, and _possibly_ in an asynchronous way as the "pause" is getting longer as the lib folder gets bigger. I know @dominikg has a prototype with using esbuild, but I think it's only for transpiling TS files and not Svelte.

packages/kit/src/packaging/index.js
dummdidumm: I think that's on the user then, the also things like `svelte-check` should be able find out. We don't check other file paths for correctness, so in my mind it would be inconsistent to check the aliases.

dummdidumm:
Fully agree that this is a hacky solution and I'd very much would like to leverage vite instead. There's a [library mode](https://vitejs.dev/guide/build.html#library-mode) but it works too different from what we want/need, as it bundles everything and will transform Svelte to JS, which we actively want to avoid.
If we do not want to use string replacements, we could also parse the script parts of the Svelte files - don't know which parsers are available to us though without adding additional dependencies.

dominikg:
not entirely sure if it's helpful in this case but vite-plugin-svelte has 2 different features that may be combined to solve it.

1) there is experimental code that demonstrates how to use vite plugins as svelte preprocessors, this could be used to try and run vite's alias resolving during preprocess
2) vite plugins can contribute additional preprocessors via `api.sveltePreprocess`

So in theory you could build a vite plugin that searches for the other vite plugin it wants to run during svelte-preprocess, save that as local state in configResolved hook and then offer api.sveltePreprocess which uses the other vite-plugin and is picked up by vite-plugin-svelte.

I agree that a simple string replace of aliases is too dangerous. For the time being it may be best to recommend not using aliases for libraries build with sveltekit package command

packages/kit/src/packaging/index.js
```diff
@@ -144,11 +153,44 @@ export async function make_package(config, cwd = process.cwd()) {
...
-		const package_path = path.join(cwd, config.kit.package.dir, pathname);
+		const package_path = path.join(abs_package_dir, pathname);
 		if (!fs.existsSync(package_path)) fs.copyFileSync(full_path, package_path);
 	}
 }
 
+/**
+ * Resolves the `$lib` alias.
+ *
+ * TODO: make this more generic to also handle
```
benmccann, line 164: this is a bit of an unexpected place to wrap the line. should probably have more like 80 or 100 characters on this line before wrapping

benmccann:
I think we'd probably want to handle `$app` as well:

https://github.com/sveltejs/kit/blob/6bd21c75d99730e78cd1e5221497a41d6ea3c7fe/packages/kit/src/core/build/index.js#L172

It probably wouldn't be too hard to refactor out the code to set the config defaults then look at the actual value of `kit.vite.resolve.aliases`

I agree this is a bit dangerous, but it also seems better than not supporting it, so I'm a bit torn

I wonder if we need to do something as advanced as what @dominikg is suggesting. Could we just have a svelte-preprocessor that does a replacement using the AST? I don't think it would necessarily need to invoke Vite, which would make things quite a bit easier

bluwy:
If we are able to use Vite to preprocess the components, we could take advantage of [useVitePreprocess](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#usevitepreprocess) which removes the need for `svelte-preprocess` (for most cases). The caveat mentioned would be addressed soon (https://github.com/sveltejs/vite-plugin-svelte/pull/173). Hopefully it could be a default for `create-svelte` one day.

benmccann:
Ohhh. I did not understand from Domenik's comment or the API names that these methods were not calling `svelte-preprocess`. I thought it would be too complex to have both involved.

dummdidumm:
I'm a little confused by the different solutions outlined in this PR. What we need is something that resolves any alias for `.js`, `d.ts` and `.svelte` files, with the additional difficulty that the generation of `d.ts` files is separate (vite or `svelte-preprocess` therefore can't hook into it). So the most robust solution to me sounds like "do alias resolving after having done everything else as a separate step", which means we need some parser that understands both TS and JS. In case of `.svelte` files, we can use Svelte's `parse` which includes parsing the JS part. In case of `d.ts` files, we can hopefully use TypeScript somehow. In case of `.js` files -> I don't know, since we can't ensure that people have TypeScript installed (they only need it in case of `d.ts` generation) - that we could change though to always require TypeScript when people want to build libraries.

bluwy:
I think it's fine to merge the PR as is with the current approach, we could look into other long-term solutions so we don't block others who need this feature. A documentation update noting the caveat would be nice too. Unless we want to ship a more robust experience to the end-users.

Issue #2424, opened 2021-09-14
browser-only redirect during load not working


WHenderson:
### Describe the bug

Browser-only redirects during the load function are being ignored.

### Reproduction

https://github.com/WHenderson/sveltekit-bug-browser-redirect

Code such as the following fails to redirect:
```html
<script context="module">
    import {browser} from "$app/env";

    export async function load() {
        if (browser) {
            // Note there would normally be some condition logic here
            return {
                status: 303,
                redirect: '/redirected',
            };
        }

        return {};
    }
</script>
```

I've been looking at the sveltekit code that should be responding to the redirect and I can't see how it can be hit. Perhaps it has become a dead code branch?

`/runtime/internal/start.js:564`
```javascript
if (result.redirect) {
	// this is a real edge case — `load` would need to return
	// a redirect but only in the browser
	location.href = new URL(result.redirect, location.href).href;
	return;
}
```

### Logs

```shell
N/A. No errors or logs are generated.
```

### System Info

```shell
Chrome
```

### Severity

serious, but I can work around it

### Additional Information

Browser-only redirects make sense when using things such as client-side authentication. In my current example, I wish to navigate away from the sign-in page once the user is authenticated and the authentication check will _always_ fail on the server since the authentication information isn't sent to the server.

dummdidumm:
Fixing the issue will maybe also solve (part of?) #1575

PR #2462, opened 2021-09-20
[fix] browser only redirect during load

baseballyama:
FIX: https://github.com/sveltejs/kit/issues/2424

I simply implemented it according to the comments of https://github.com/sveltejs/kit/issues/2424.
And I think it works well.
But I'd like to know if there's anything I haven't considered.

**Question**

If there are multiple redirects, what should I do?
`redirect: (redirect && redirect.loaded) ? redirect.loaded.redirect : undefined,`

=>
I think mainly this is a programming mistake.
So it doesn't need to handle it.
But it might be kind to throw an error if multiple redirects are there.

Thank you!

### Tests

### Changesets


packages/kit/src/runtime/client/renderer.js
```diff
@@ -392,6 +392,7 @@ export class Renderer {
 
 		/** @type {import('./types').NavigationResult} */
 		const result = {
+			redirect: branch.find((b) => b?.loaded?.redirect)?.loaded?.redirect,
```
dummdidumm, line 395: We can't use optional chainning unfortunately because we don't transpile the code and Node 14 does not support it in all versions.

baseballyama: Thank you for reviewing.

OK. Should I use Node12 for kit?

dummdidumm: You don't need to do that, you just have to remember this caveat - and the CI will catch this, too, in case it's overlooked.

baseballyama: Thank you for telling me!
And I updated it.

baseballyama:
Sorry I forgot to run 'pnpm lint' after commit https://github.com/sveltejs/kit/pull/2462/commits/b99330a330b15f831f85e817b1f76cefa0f5837d
So I updated.

packages/kit/test/apps/basics/src/routes/redirect-on-load/_tests.js
```diff
@@ -0,0 +1,9 @@
+import * as assert from 'uvu/assert';
+
+/** @type {import('test').TestMaker} */
+export default function (test, is_dev) {
+	test('redirect-on-load', '/redirect-on-load', async ({ base, page, clicknav }) => {
```
benmccann, line 5: I don't think this is testing the new functionality you added. I think it's testing server-side redirect. If you look at some of the other tests there should be examples of testing both

baseballyama: Oops. sorry!
I updated!
(This time, again I checked tests that fails without this PR but passes with it.)

And during the update, I faced puppeteer issue due to redirect.
So I edited `packages/kit/test/test.js` in the last commit.

packages/kit/test/apps/basics/src/routes/redirect-on-load/_tests.js
```diff
@@ -0,0 +1,14 @@
+import * as assert from 'uvu/assert';
+
+/** @type {import('test').TestMaker} */
+export default function (test, is_dev) {
+	test('redirect-on-load', '/redirect-on-load', async ({ base, page, js }) => {
+		if (js) {
+			await page.waitForTimeout(50);
```
benmccann, line 7: I would not use this. Per the docs for that method:
>Note that page.waitForTimeout() should only be used for debugging. Tests using the timer in production are going to be flaky. Use signals such as network events, selectors becoming visible and others instead.

baseballyama: Yes. Totally agree with you.
But already few exists tests use `waitForTimeout` so I followed that.

Then at least I will update this PR.
But should I refactor exists `waitForTimeout`?
I think we should do it to avoid random test failures.
If yes, I will create one more PR which is for existing tests with `waitForTimeout`.

benmccann: Yes, that would be great to fix the other tests so that they don't use `waitForTimeout`. Thank you! I filed an issue to track it: https://github.com/sveltejs/kit/issues/2473

baseballyama: Oh, Thank you for creating a new issue. I will handle it.

baseballyama: I have added a function to avoid the use of  `waitForTimeout`.
I hope to get your review.

And I will create a new PR for removing all `waitForTimeout` based on this branch.

benmccann:
What is the new test code for `get_elements` / iframe? It doesn't look related to redirects, so I'm not sure why it's included in this PR

baseballyama:
> What is the new test code for `get_elements` / iframe? It doesn't look related to redirects, so I'm not sure why it's included in this PR

Thank you for your review!

I want to write a test for `get_elements` because this is a new code,
But I didn't find proper place, so I added it into `routes` folder.

I can understand this is a little bit weird.

Then I have 3 opinions and I prefer No3 but what do you think?
1. Write test in `route` folder (My current way)
 --> This is reasonable to write test, but this is not test for Kit's spec. So it's little bit weird.
2. create new test file which is like `test.spec.js` and write test for `get_elements`.
 --> It relatively takes time and effort.
3. No test
--> `get_elements` will not update frequently and already I checked that it works by existing test cases. So there is no risk even there is no test.

benmccann:
I don't think we should need to add a `get_elements` method. We just need to check that we loaded the URL which was redirected to

baseballyama:
> I don't think we should need to add a `get_elements` method. We just need to check that we loaded the URL which was redirected to

Hum....
I re-checked that the waiting process like `waitForTimeout ` or `get_elements` was necessary or not.
But now we don't need it. (I'm a bit confused because I remember checking this out yesterday, and thought I needed it.)
(Could it be dependent on CPU usage or some machine resource??)
(But I run `pnpm:test` many times but always passed.)

Anyway now this PR doesn't need any waiting process except redirect process.

---
And at next PR (https://github.com/sveltejs/kit/pull/2488),
I want to solve this your comment.

> What is the new test code for get_elements / iframe? It doesn't look related to redirects, so I'm not sure why it's included in this PR

I hope I will get your opinion about it.
(Personally, I prefer No3)
> No test

packages/kit/test/apps/basics/src/routes/redirect-on-load/_tests.js
```diff
@@ -0,0 +1,13 @@
+import * as assert from 'uvu/assert';
+
+/** @type {import('test').TestMaker} */
+export default function (test, is_dev) {
```
benmccann, line 4: `is_dev` is unused

```suggestion
export default function (test) {
```

baseballyama: removed!

BTW, Some existing tests also declared unused `is_dev`.
But test files have been excluded from `pnpm lint` and `pnpm check`.
So it has a chance to have a mistake like this.

benmccann: linting is enabled now: https://github.com/sveltejs/kit/commit/a2c9a15ec5604a5216fa3940768421bf838fa71c

it's not catching unused variables in other tests though, so perhaps there's some other check that would need to be enabled for that

baseballyama: I think because `.eslintrc.json` extends `@sveltejs` but `no-unused-vars` is off.

Should I add it? (If yes I will create one more PR.)
```
"rules": {
    "no-unused-vars": "error"
}
```

https://github.com/sveltejs/kit/blob/a2c9a15ec5604a5216fa3940768421bf838fa71c/packages/kit/.eslintrc.json#L3

[sveltejs/eslint-config]
https://github.com/sveltejs/eslint-config/blob/31fd4faeea88990069502460b023698b1c9c2d13/index.js#L15

benmccann: We don't use that rule because it's the JavaScript rule and won't work on TypeScript files. Instead we've turned on the TypeScript version of the rule: https://github.com/sveltejs/eslint-config/blob/31fd4faeea88990069502460b023698b1c9c2d13/index.js#L48

It looks like that rule is essentially disabled because it is set to `none`. It would need to be changed to something like `after-used` to enable it.

We try to keep the rules for this repo in sync with the `sveltejs/svelte` repo, so any changes would need to happen in the following order:
- Enable the check locally in `sveltejs/svelte` and fix all errors, but don't check in the change to `.eslintrc.json`
- Modify the `sveltejs/eslint-config` package to update the rule after the PR to `sveltejs/svelte` is merged so that we know people agree to the code changes resulting from changing the eslint config value
- Update `sveltejs/svelte` and `sveltejs/kit` with the new version of `sveltejs/eslint-config` and fix any remaining errors

baseballyama: Thank you for tell me how to change lint settings for svelte.

> It looks like that rule is essentially disabled because it is set to none. It would need to be changed to something like after-used to enable it.

I enabled below rule locally in `sveltejs/svelte`.
But 246 errors come. lol

```
rules: {
  '@typescript-eslint/no-unused-vars': ['error', { args: 'after-used' }]
}
```

Of cause I can fix these errors.
But I'm afraid that it will have a significant impact on other projects.
(I can look around at all the projects but should we?)

So Is there a way to get prior approval from the core team?
Or can we consider this chat as an approval?

benmccann: No need to worry about other projects. But the change would need approval from the maintainers besides myself. You could possibly try sending a PR fixing a good handful of the errors for discussion so that people can see what the changes would be without you having to do the work of fixing all of them.

baseballyama: Okay! I will try to send a PR.
Thank you for supporting me!

packages/kit/test/test.js
```diff
@@ -197,7 +197,20 @@ function duplicate(test_fn, config, is_build) {
 
 				if (start) {
 					response = await context.pages.js.goto(context.base + start);
-					await context.pages.js.evaluate(() => window.started);
+
+					// If redirecting, sometimes below error comes. Therefore wait until redirect.
+					// "page.evaluate: Execution context was destroyed, most likely because of a navigation."
+					// MOTE: waitForNavigation is extremely slow so doesn't use.
```
benmccann, line 203:
```suggestion
					// NOTE: waitForNavigation is extremely slow so doesn't use.
```

benmccann: although I don't think this NOTE makes sense and I'd probably just remove it. `goto` on the previous line waits for the page `load` event, which is the same thing `waitForNavigation` would do. So it can't be any slower. Maybe the issue is that it's already been awaited for, but then the issue is waiting twice for the same event and not that the method is too slow

I think I'd just remove this line since it adds more confusion than it helps with

benmccann: Although now I'm confused because I don't know exactly where you were going to call `waitForNavigation`. The other option might be to move that comment to be right above the line where you might use it

packages/kit/test/test.js
```diff
@@ -197,7 +197,20 @@ function duplicate(test_fn, config, is_build) {
...
+					const max = 100;
+					for (let i = 0; i <= max; i++) {
+						try {
+							await context.pages.js.evaluate(() => window.started);
+							break;
+						} catch (e) {
+							if (!(e instanceof Error) || i === max) throw e;
+							await new Promise((resolve) => setTimeout(resolve, 10));
```
benmccann, line 211: I didn't understand in the NOTE above where you were suggesting the `waitForNavigation` might be called. I think that this would be a good place to do it though as an alternative to calling `setTimeout`. I'd prefer not to call `setTimeout` if possible

packages/kit/test/test.js
baseballyama: But In most cases, the test failed due to the bellow error without this change.

```
.svelte-kit/output/client/service-worker.js   9.25 KiB / brotli: 2.08 KiB
• • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • ✘ • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • • •   (232 / 233)

   FAIL  build:basics  "redirect-on-load [js]"
    page.evaluate: Execution context was destroyed, most likely because of a navigation.

    at Object.handler (file:///C:/Users/baseballyama/Desktop/git/svelte-contributes/kit-2424/kit/packages/kit/test/test.js:201:29)
    at async Number.runner (file:///C:/Users/baseballyama/Desktop/git/svelte-contributes/kit-2424/kit/node_modules/.pnpm/uvu@0.5.1/node_modules/uvu/dist/index.mjs:77:5)
    at async Module.exec (file:///C:/Users/baseballyama/Desktop/git/svelte-contributes/kit-2424/kit/node_modules/.pnpm/uvu@0.5.1/node_modules/uvu/dist/index.mjs:131:33)
    at async main (file:///C:/Users/baseballyama/Desktop/git/svelte-contributes/kit-2424/kit/packages/kit/test/test.js:394:2)
```

I think this is similar to https://github.com/microsoft/playwright/issues/4192.

And I can't move this waiting process to `redirect-on-load` test file.
Because this error comes before called `callback` function.

But I have modified the code in response to the comment below that you gave me.
Could you please check the updated code?

packages/kit/test/test.js
baseballyama: Ah Definitely! I should update it!


---
[This is just additional explanation]

If error comes during `await context.pages.js.evaluate(() => window.started);`,
it means already navigation was changed.
So after call `await context.pages.js.waitForNavigation();`,
`await context.pages.js.evaluate(() => window.started);` should be succeed.
(But run a for loop in case of multiple redirects.)

packages/kit/test/test.js
```diff
@@ -197,7 +197,15 @@ function duplicate(test_fn, config, is_build) {
 
 				if (start) {
 					response = await context.pages.js.goto(context.base + start);
-					await context.pages.js.evaluate(() => window.started);
+					for (let i = 0; i <= 9; i++) {
```
benmccann, line 200:
```suggestion
					// Evaluate will throw an exception if the page is navigated away from while the
					// script is executing, which may happen during redirect tests. In that case, the
					// start event will be fired after the redirect
					for (let i = 0; i <= 9; i++) {
```

benmccann:
thank you!!

PR #2476, opened 2021-09-23
disabled router fixes

erwin-k:
When the router is disabled for a particular page the back button doesn't respond on said page.

### Tests

### DIFF
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -154,7 +154,9 @@ export class Router {
 		});
 
 		addEventListener('popstate', (event) => {
-			if (event.state && this.enabled) {
+			if (!this.enabled) return (window.location.href = location.href);
+
+			if (event.state) {
 				const url = new URL(location.href);
 				this._navigate(url, event.state['sveltekit:scroll'], false, []);
 			}
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -143,6 +143,10 @@ export async function render_response({
 				}` : 'null'}
 			});
 		</script>`;
+	} else if (!page_config.router) {
+		init = `<script>
+			addEventListener('popstate', () => location.reload());
+		</script>`;
 	}
 
 	if (options.service_worker) {


packages/kit/src/runtime/client/router.js
```diff
@@ -154,7 +154,9 @@ export class Router {
 		});
 
 		addEventListener('popstate', (event) => {
-			if (event.state && this.enabled) {
+			if (!this.enabled) return (window.location.href = location.href);
```
benmccann, line 157: what is the reason for this? it looks like it reloads the page? that doesn't seem like the right behavior to me, but maybe i'm misunderstanding

erwin-k: Yes it reloads the page when the back button is used, this is expected behaviour to me on a page with the router disabled.
Without this it still changes the url but doesn't rerender the page.

benmccann: Hmm. That doesn't seem like quite the right fix to me. Rather than re-implementing back-button behavior, is there a way we can make the back button work? E.g. is there a call to `event.preventDefault()` somewhere we can remove or something similar happening?

erwin-k: The reason the back button stops working is because clicking links on a page with no router causes a refresh, which means when you navigate back to this page it causes a refresh, landing on a page with no router means there wont be a router or renderer initialized, this is where the back button stops working because it's working with a history without refreshes but there is no renderer to respond to that.

Maybe the router and renderer should always be created? That would make a way to solving this.

benmccann: > clicking links on a page with no router causes a refresh

Huh? Clicking links on a page with no router should cause a server-side navigation that loads a new page. Refresh means the existing page is reloaded, which is not what I'd expect when clicking a link

> this is where the back button stops working because it's working with a history without refreshes but there is no renderer to respond to that

Can you explain more what you mean by this? I believe the history is just a set of URLs and that when you navigate to a new URL is should get pushed onto the history stack whether it's a client-side or server-side navigation. Am I wrong about this? I don't really understand what you're saying

erwin-k: > Huh? Clicking links on a page with no router should cause a server-side navigation that loads a new page. Refresh means the existing page is reloaded, which is not what I'd expect when clicking a link

Sorry, what I meant with refresh was as you stated, server-side navigation.

> Can you explain more what you mean by this? I believe the history is just a set of URLs and that when you navigate to a new URL is should get pushed onto the history stack whether it's a client-side or server-side navigation. Am I wrong about this? I don't really understand what you're saying

What I mean is that the history keeps track of which URLs cause a server side request, using the navigation buttons will load the router-less pages from the server because that's how they were put in to the history. '

When navigating **back** to a page with no router, it gets loaded from the server but with no scripts for the router and renderer, navigating back again from this page won't load the page from the server, because you visited this router-less page from a page with a router. 
That's where this problem is, you can visit a router-less page without the "refresh effect" and because you can end up with no router/renderer on router-less pages the pages with a router in the history don't render.
I hope that clears it up.

These changes attempt to give that page refresh effect when it's expected.

erwin-k: Here is a [reproducible example](https://github.com/erwin-k/kit/tree/router-fixes-example) of this bug, run the hn.svelte.dev example app and go to page "1".

erwin-k: @benmccann I opened an issue a while ago (#2500) to continue this discussion, I would like it if you could take a look.

erwin-k: @benmccann under the hood the browser history keeps track wether navigation was *preventDefaulted*, so the browser will follow history and *prevent the default* navigation when coming back from these router-less pages. (url changes but nothing is happening, because there is no renderer)

PR #2479, opened 2021-09-23
[feat] create-svelte: Add prompt for choosing adapter

### LINKED ISSUE #2459: Include a default adapter for the demo project
### Describe the problem

I installed the demo project and tried to build, but got an error mentioning that no adapter was specified, which seems somewhat confusing as a default experience.

Default the demo project to the static adapter/node adapter (not sure which is more appropriate), or add a question to the setup (along with whether to enable Typescript etc...) that defaults to including an adapter.

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

supercoww:
Fixes #2459 

Add a new prompt for adapter. The default option is `"I'll add it later or use a community-provided adapter"`.

Updated:
image

image

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -232,7 +232,7 @@ function write_common_files(cwd, options, name) {
 	});
 
 	if (options.adapter) {
-		pkg.devDependencies[options.adapter.npm] = 'next';
+		pkg.devDependencies[options.adapter.npm] = '^1.0.0-next.1';
 	}
 	pkg.dependencies = sort_keys(pkg.dependencies);
 	pkg.devDependencies = sort_keys(pkg.devDependencies);
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -5,6 +5,7 @@ import { fileURLToPath } from 'url';
 import { bold, cyan, gray, green, red } from 'kleur/colors';
 import prompts from 'prompts';
 import { mkdirp, copy } from './utils.js';
+import adapters from './adapters.js';
 
 // prettier-ignore
 const disclaimer = `
@@ -79,6 +80,18 @@ async function main() {
 				initial: false,
 				active: 'Yes',
 				inactive: 'No'
+			},
+			{
+				type: 'select',
+				name: 'adapter',
+				message: 'Add an adapter',
+				choices: [
+					{
+						title: "I'll add it later or use a community-provided adapter (https://bit.ly/3kvqIRy)",
+						value: null
+					},
+					...adapters.map((adapter) => ({ title: adapter.name, value: adapter }))
+				]
 			}
 		])
 	);
@@ -121,6 +134,12 @@ async function main() {
 						'Svelte-specific formatting options: https://github.com/sveltejs/prettier-plugin-svelte#options'
 				)
 			)
+		);
+	}
+
+	if (options.adapter) {
+		console.log(
+			bold(green(`✔ Added ${options.adapter.name}.\n` + `Documentation: ${options.adapter.url}`))
 		);
 	}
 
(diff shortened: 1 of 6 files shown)


packages/create-svelte/bin.js
```diff
@@ -79,6 +80,18 @@ async function main() {
...
 				active: 'Yes',
 				inactive: 'No'
+			},
+			{
+				type: 'select',
+				name: 'adapter',
+				message: 'Add an adapter',
+				choices: [
+					{
+						title: "I'll add it later or use a community-provided adapter (https://bit.ly/3kvqIRy)",
```
dominikg, line 90: do we really want to drive traffic through bit.ly here? who controls that link.
Would rather want to see an npm keyword search or a link to a svelte community page

packages/create-svelte/bin.js
```diff
@@ -196,12 +215,23 @@ function write_common_files(cwd, options, name) {
...
+					.replace(/\t+\/\/ insert:adapter/, options.adapter ? '\t\tadapter: adapter(),' : '');
+			}
 			const dest = path.join(cwd, file.name);
 			mkdirp(path.dirname(dest));
 			fs.writeFileSync(dest, file.contents);
 		}
 	});
 
+	if (options.adapter) {
+		pkg.devDependencies[options.adapter.npm] = 'next';
```
dominikg, line 233: Should the version be the actual value of the latest release instead of the tag? 
People installing like this would get left behind once adapters are released as 1.0 which switches to latest tag.

As a workaround for having to pull every adapter version for now we could use "^1.0.0-next.0"  which should match all releases of the adapters including "1.x.y". It won't match new majors after 1 eg "2.x.y" or new prereleases after 1.0.0 eg "1.1.2-next.1"

packages/create-svelte/bin.js
supercoww: I agree on not using bitly. The full link (https://sveltesociety.dev/components/#category-SvelteKit%20Adapters) will make the option longer but I think its still the better option. I copied the link from the related issue shared by Ben.

packages/create-svelte/bin.js
supercoww: @dummdidumm I tried using the shared files for this but would end up making 10 different svelte.config.js and that would make future changes more difficult. Can use it for adding adapter to devDependencies but the current logic will still be needed if we want to include community adapters as discussed here https://github.com/sveltejs/kit/issues/2459#issuecomment-925236207

packages/create-svelte/bin.js
benmccann: https://sveltesociety.dev/components#category-SvelteKit%20Adapters seemed kind of long to me. Maybe we can updates sveltesociety.dev first to make it something shorter like https://sveltesociety.dev/components#adapters

supercoww: I moved the link to the description on next line .

image

If that is not good enough, I can make the change on sveltesociety.dev.

benmccann: I think making the change on sveltesociety.dev would be really nice. It's a pretty long description right now

dummdidumm:
I approve the change but not its current implementation, at least not yet. It's very different to what is present and having two different styles to add prompts is confusing. Honestly I was never a big fan of the refactoring to use these folders, which now shows as it would result in many more folders needed to be created. I'm okay to merge this after we agree on how to refactor this into a consistent approach.

supercoww:
If we want to provide an option to choose from a list of community adapters fetched from an API, we will still need some runtime logic outside the folders similar to the current implementation.
I guess I can modify the `svelte.config.js` file without the `// insert:` placeholders if that is desirable.

Rich-Harris:
We ended up going in a different direction — #2867 — so I'm going to close this. Thank you!

PR #2482, opened 2021-09-23
[fix] externalize middlewares.js when using adapter-node's entryPoint

### LINKED ISSUE #2481: Error with custom entryPoint in adapter-node
### Describe the bug

When setting up adapter-node I'm configuring a custom entry point copied from the example here: https://github.com/sveltejs/kit/tree/master/packages/adapter-node#middleware but I run into this error when I try to start the server with `node build`: `Dynamic require of "buffer" is not supported`.

The error message points to the `jsonwebtoken` package which I use on some node only endpoints.
But, this is **not** specific to this package, it's the same with `dotenv` and other node packages.

If I skip the entryPoint config in the adapter setup, it works.

### Reproduction

Minimal repro repo: https://github.com/oskarhane/sveltekit-adapter-node-bug-repro

```
npm ci
```

All I've done there is to `npm init svelte@next`, installing `@sveltejs/adapter-node`, `jsonwebtoken` and `polka` and configured the adapter after creating a single request handler that uses the jwt package.

```
npm run build
node build
```

Edit (toggle the comments) the adapter config in `svelte.config.js` and try again to have it work / not work.

### Logs

```shell
▶ node build   
file:///Users/oskarhane/Development/tmp/sk-bug/build/index.js:13
  throw new Error('Dynamic require of "' + x + '" is not supported');
        ^

Error: Dynamic require of "buffer" is not supported
    at file:///Users/oskarhane/Development/tmp/sk-bug/build/index.js:13:9
    at node_modules/safe-buffer/index.js (file:///Users/oskarhane/Development/tmp/sk-bug/build/index.js:33:18)
    at __require2 (file:///Users/oskarhane/Development/tmp/sk-bug/build/index.js:16:44)
    at node_modules/jws/lib/sign-stream.js (file:///Users/oskarhane/Development/tmp/sk-bug/build/index.js:564:19)
    at __require2 (file:///Users/oskarhane/Development/tmp/sk-bug/build/index.js:16:44)
    at node_modules/jws/index.js (file:///Users/oskarhane/Development/tmp/sk-bug/build/index.js:744:22)
    at __require2 (file:///Users/oskarhane/Development/tmp/sk-bug/build/index.js:16:44)
    at node_modules/jsonwebtoken/decode.js (file:///Users/oskarhane/Development/tmp/sk-bug/build/index.js:777:15)
    at __require2 (file:///Users/oskarhane/Development/tmp/sk-bug/build/index.js:16:44)
    at node_modules/jsonwebtoken/index.js (file:///Users/oskarhane/Development/tmp/sk-bug/build/index.js:2835:15)
```

### System Info

```shell
System:
    OS: macOS 11.5.2
    Memory: 3.38 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.17.4 - ~/.nvm/versions/node/v14.17.4/bin/node
    Yarn: 1.22.11 - ~/.nvm/versions/node/v14.17.4/bin/yarn
    npm: 6.14.14 - ~/.nvm/versions/node/v14.17.4/bin/npm
  npmPackages:
    @sveltejs/adapter-node: ^1.0.0-next.49 => 1.0.0-next.49 
    @sveltejs/kit: next => 1.0.0-next.171 
    svelte: ^3.34.0 => 3.38.3 
```

### Severity

### Additional Information

_No response_

Conduitry:
Fixes #2481, hopefully without breaking anything else. This is almost the same as my suggested solution in my comment on that issue - except it also includes an override for the path in the `onResolve` hook to avoid having `import { assetsMiddleware, prerenderedMiddleware, kitMiddleware } from "../build/middlewares.js";` in the final file.

### Tests

### Changesets

### DIFF
--- a/.changeset/mighty-ears-divide.md
+++ b/.changeset/mighty-ears-divide.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-node': patch
+---
+
+[fix] Correctly treat `middlewares.js` as external when using `entryPoint` option
--- a/packages/adapter-node/index.js
+++ b/packages/adapter-node/index.js
@@ -7,7 +7,7 @@ import {
 	statSync,
 	writeFileSync
 } from 'fs';
-import { join } from 'path';
+import { join, resolve } from 'path';
 import { pipeline } from 'stream';
 import glob from 'tiny-glob';
 import { fileURLToPath } from 'url';
@@ -81,7 +81,6 @@ export default function ({
 				entryPoints: [entryPoint],
 				outfile: join(out, 'index.js'),
 				bundle: true,
-				external: ['./middlewares.js'], // does not work, eslint does not exclude middlewares from target
 				format: 'esm',
 				platform: 'node',
 				target: 'node12',
@@ -90,8 +89,12 @@ export default function ({
 					{
 						name: 'fix-middlewares-exclude',
 						setup(build) {
-							// Match an import called "./middlewares.js" and mark it as external
-							build.onResolve({ filter: /^\.\/middlewares\.js$/ }, () => ({ external: true }));
+							// Match an import of "middlewares.js" and mark it as external
+							build.onResolve({ filter: /\/middlewares\.js$/ }, ({ path, resolveDir }) => {
+								if (resolve(resolveDir, path) === resolve(out, 'middlewares.js')) {
+									return { path: './middlewares.js', external: true };
+								}
+							});
 						}
 					}
 				]


benmccann (approved):
awesome. thanks!  I'm very happy you took a look at this one as esbuild is not my forte.

I do wonder if there might be a better way to set this up. Rather than making the user import `'../build/middlewares.js'`, perhaps we could have them do an import from `@sveltejs/adapter-node`? And I wonder if that might even allow us to do just a single esbuild compilation instead of the two that `adapter-node` is doing now

Conduitry:
And we would then tell esbuild to treat that specially when resolving it, and instead point it to `.svelte-kit/node/middlewares.js`? That's possible. It would probably be tidier, but would it be too magical? That would be a separate discussion, though, and we should fix what we have in the meantime. I think I'll open an issue shortly so we can chat about that.

Conduitry:
Hm. I'm having second thoughts somewhat. IIRC part of the reason for the `middlewares.js` thing - even before there was `entryPoint`- was that we wanted a way to expose middleware that could then be used as part of an entirely different build process afterwards. Does this sound familiar? We could add a nice alias for `../build/middlewares.js` later as a non-breaking change, but I don't think we should stop producing that as its own build artifact if we want to give people the option to consume that however they want.

benmccann:
> And we would then tell esbuild to treat that specially when resolving it, and instead point it to .svelte-kit/node/middlewares.js?

I was thinking it could be just a regular import from a regular node module. We might have to refactor slightly to pass in `render`

> we wanted a way to expose middleware that could then be used as part of an entirely different build process afterwards. Does this sound familiar?

Folks wanted to be able to create a custom server and this was the solution that came out of it. I don't think anyone would miss the separate build artifact as long as they can still build a custom server

Conduitry:
Oops. This made `middlewares.js` no longer external when you're _not_ using `entryPoint`, because the path to the imported file is different - it's in `.svelte-kit/node/middlewares.js` for regular builds. There's only an import from `build/middlewares.js` in custom `entryPoint` builds where you're pointing to that file. I'll submit a PR shortly.

PR #2498, opened 2021-09-26
Fix a typo in the README.md

Peyman-Borhani:
Edit :   code that's run   ->   code that runs


Conduitry:
This isn't a typo. It's a stylistic choice. "Code that's run" ("code that is run") is the passive voice, and is grammatical.

PR #772, opened 2021-03-30
Set target to es2019 in default tsconfig.json

benmccann:
Should at least make it 2020. Someone on Discord ran into an issue with it being an older version

I hadn't set it in `jsconfig.json` because that file is pretty minimal and doesn't have any of the other `tsconfig.json` contents. I'm not sure which settings go in both vs just one or the other

PR #2502, opened 2021-09-27
create-svelte: Use target es2020 as default in tsconfig.json

### LINKED ISSUE #2457: Use `target es2020` as default in tsconfig.json
### Describe the problem

I think we as world are ready for `"target": "es2020"` as default when we generate new `SvelteKit` project.

Aren't we?

Use `"target": "es2020"`.

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

supercoww:
fixes #2457 

### Tests

### Changesets

### DIFF
--- a/.changeset/chilled-buses-develop.md
+++ b/.changeset/chilled-buses-develop.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Use target es2020 as default in tsconfig.json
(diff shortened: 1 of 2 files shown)


dummdidumm:
We have not done this in the past because some we support Node 12.x, and that doesn't support es2020 features, so it would be confusing for new users if they run into a situations where their deploy target does not support the newer features which are not transpiled if the target is es2020. More info in #772 .
I'm therefore slightly in favor of not merging this.

Conduitry:
Yup, agreed, I don't think we want to do this.

Issue #2467, opened 2021-09-21
Showing port number are different between Vite and Kit if 3000 is already in use


baseballyama:
### Describe the bug

When I run `npm run dev` in kit project,
Different port shows if 3000 is already in use as you can see from the log.

### Reproduction

After use 3000 port by some process, run this commands.

```
yes "" | npm init svelte my-app
cd my-app
npm install
npm run dev
```

### Logs

```shell
vite v2.5.7 dev server running at:

  > Local: http://localhost:3001/
  > Network: use `--host` to expose

  SvelteKit v1.0.0-next.169

  network: not exposed
  network: not exposed
  local:   http://localhost:3000

  Use --host to expose server to other devices on this network
```

### System Info

```shell
System:
    OS: Windows 10 10.0.19043
    CPU: (8) x64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz  
    Memory: 2.55 GB / 15.86 GB
  Binaries:
    Node: 14.17.0 - C:\Program Files\nodejs\node.EXE       
    npm: 6.14.13 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1023.0), Chromium (93.0.961.52)
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.170
    svelte: ^3.34.0 => 3.42.6
```

### Severity

annoyance

### Additional Information

Maybe `packages\kit\src\cli.js#welcome` should be removed most of the code??

https://github.com/sveltejs/kit/blob/master/packages/kit/src/cli.js#L217

And the output should be like this?

```
  vite v2.5.7 dev server running at:

  > Local: http://localhost:3000/
  > Network: use `--host` to expose

  SvelteKit v1.0.0-next.169

  Use --host to expose server to other devices on this network
```

Thank you.

bluwy:
The duplicated welcome messages likely came from #2232, I think the best case is if we can replace `vite v2.5.7 dev server` with `SvelteKit v1.0.0-next.169`, though that's tricky. Otherwise having a `SvelteKit v1.0.0-next.169` message below vite's might be fine too.

benmccann:
I thought kit would exit if the port were already in use:

https://github.com/sveltejs/kit/blob/d8efaa2faa9c7a93c9b7e7cf234bc485de330d28/packages/kit/src/cli.js#L206

[...]

PR #2507, opened 2021-09-27
[feat] allow using Vite's `strict.port` option

benmccann:
Closes https://github.com/sveltejs/kit/issues/2467

This delegates port checking to Vite and allows you to set `strict.port: false`. I set the `strict.port` default to `true`, which is different than Vite's default value for backwards compatibility

The server startup message is printed twice, but that will be fixed when we upgrade to Vite 2.6. That upgrade will remove their startup message and leave only ours

This doesn't touch `preview` mode for now. It may be possible to use the Vite server in preview mode as well, but that would be blocked by an upgrade to Vite 2.6

### DIFF
--- a/.changeset/beige-trees-confess.md
+++ b/.changeset/beige-trees-confess.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[feat] allow using Vite's `strict.port: false` option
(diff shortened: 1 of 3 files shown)


Conduitry:
Is there any cleanup we can do now - or would stuff like removing `port-authority` as a dependency happen after we also (hypothetically) switch to using Vite's server for `preview` as well?

benmccann:
Yeah, I don't think we could remove `port-authority` yet since it's still used in `preview` mode. Not sure if you had any other ideas for cleanup, but I don't think there's a whole lot more we could do at the moment

Conduitry (approved):

Issue #2510, opened 2021-09-28
Unable to use own certificates in dev mode


taylorzane:
### Describe the bug

When using SvelteKit in HTTPS mode, it is no longer possible (as of #2232) to specify a collection of HTTPS certificates in the `svelte.config.js` file.

### Reproduction

MVP Repository to cause the error (doesn't actually have certificates, but they're not important at this point)

https://github.com/taylorzane/svelte-kit-https-repro

```
git clone https://github.com/taylorzane/svelte-kit-https-repro
cd svelte-kit-https-repro
npm install
npm run dev # throws a warning, doesn't try to use empty certs object
```

`svelte.config.js`
```js
import fs from 'fs'
import path from 'path'

const keys = []
const certs = []
const cas = []

keys.push(fs.readFileSync(path.join('/path/to/certs', 'key.pem')))
certs.push(fs.readFileSync(path.join('/path/to/certs', 'cert.pem')))
cas.push(fs.readFileSync(path.join('/path/to/certs', 'ca.pem')))

/** @type {import('@sveltejs/kit').Config} */
const config = {
  kit: {
    target: '#app',
    vite: {
      server: {
        https: {
          key: keys,
          cert: certs,
          ca: cas
        }
      }
    }
  }
}

export default config
```

### Logs

```shell
(full logs are sensitive, but this is the relevant snippet)

$ pnpm run dev
The value for kit.vite.server.https specified in svelte.config.js has been ignored. This option is controlled by SvelteKit.

$ pnpm run dev -- --https
The value for kit.vite.server.https specified in svelte.config.js has been ignored. This option is controlled by SvelteKit.
```

### System Info

```shell
System:
    OS: macOS 11.4
    CPU: (12) x64 Intel(R) Core(TM) i7-8700B CPU @ 3.20GHz
    Memory: 1.93 GB / 32.00 GB
    Shell: 5.8 - /usr/local/bin/zsh
  Binaries:
    Node: 16.2.0 - ~/.nodenv/versions/16.2.0/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.13.0 - ~/.nodenv/versions/16.2.0/bin/npm
  Browsers:
    Chrome: 93.0.4577.82
    Firefox: 91.0.1
    Firefox Developer Edition: 79.0
    Safari: 14.1.1
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.173 
    svelte: ^3.41.0 => 3.42.5
[...]

PR #2512, opened 2021-09-28
[fix] allow passing in https certs again

benmccann:
fixes https://github.com/sveltejs/kit/issues/2510


benmccann:
Yes, this has been rolled out. You can see the changelog here: https://github.com/sveltejs/kit/blob/master/packages/kit/CHANGELOG.md

Issue #2521, opened 2021-09-29
ERR_CONNECTION_REFUSED can not connect by ip (--host) in 1.0.0-next.176


thanet-s:
### Describe the bug

ERR_CONNECTION_REFUSED can not connect by ip (--host) in `1.0.0-next.176`
and can not connect form outside docker
but no problem in `1.0.0-next.175` by change version in package.json and re install deps

### Reproduction

-

### System Info

```shell
System:
    OS: macOS 11.5
    CPU: (8) arm64 Apple M1
    Memory: 217.02 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.10.0 - ~/.nvm/versions/node/v16.10.0/bin/node
    Yarn: 1.22.11 - ~/.nvm/versions/node/v16.10.0/bin/yarn
    npm: 7.24.0 - ~/.nvm/versions/node/v16.10.0/bin/npm
  Browsers:
    Chrome: 94.0.4606.61
    Firefox: 91.0.2
    Safari: 14.1.2
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.176 
    svelte: ^3.42.6 => 3.43.0
```

### Severity

blocking an upgrade

whataboutpereira:
Same here.

```
  vite v2.5.10 dev server running at:

  > Local: http://localhost:3000/
  > Network: use `--host` to expose

  SvelteKit v1.0.0-next.176

  local:   http://localhost:3000
  network: http://192.168.1.185:3000
```

```
$ curl http://192.168.1.185:3000
curl: (7) Failed to connect to 192.168.1.185 port 3000: Connection refused
```

benmccann:
Must be related to https://github.com/sveltejs/kit/pull/2512

Conduitry:
Could it have been introduced in #2232? Should https://github.com/sveltejs/kit/pull/2232/files#diff-ed48923f0655f935bee505cdcbf40212129ff49a89c2c9d752bd63aa1e6282c4R148 have passed along `'0.0.0.0'` if the user specified `--host` without an explicit address?

benmccann:
That change was included in `1.0.0-next.169`, but the issue report said it worked in `1.0.0-next.175` and broke in `1.0.0-next.176`

PR #2525, opened 2021-09-29
[fix] restore functioning of --host CLI flag with no arg

benmccann:
Fixes https://github.com/sveltejs/kit/issues/2521. Also lets you specify these values in `svelte.config.js`, which wasn't working either


benmccann:
I'm going to go ahead and merge this and push it out since the bug seems to be affecting a fair number of people

PR #2529, opened 2021-09-30
[feat] SSR never option

### LINKED ISSUE #1650: Tru-er SPA mode

JeanJPNM:
This pull request aims to close #1650 by adding a `never` value to the `ssr` option, preventing svelte kit from importing the files on the server altogether.

Other alternatives: add `spa` option to config.

Todo list:

- [x] Build simple approach
- [x] Add tests
- [x] Document feature
- [x] Add changeset


benmccann:
Thanks for this! I rebased the PR against `master` so that the lint job would pass

JeanJPNM:
Hey @benmccann, do you how can I create a new test app?

.changeset/bright-tips-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Added a "never" value to the config.kit.ssr option, it prevents pages from being evaluated on the server on both ssr and prerendering,n
```
dummdidumm, line 5: The end of the commit message seems wrong/cropped

documentation/docs/11-ssr-and-javascript.md
```diff
@@ -22,6 +22,8 @@ You can disable SSR app-wide with the [`ssr` config option](#configuration-ssr),
 </script>
 ```
 
+Note: `"never"` is not supported on page level `ssr` exports.
```
rmunn, line 25: I suggest adding a sentence to explain that if the `ssr` config option is set to `"never"` then page-level `ssr` exports are ignored.

documentation/docs/14-configuration.md
```diff
@@ -199,7 +199,8 @@ An object containing zero or more of the following values:
 
 ### ssr
 
-Enables or disables [server-side rendering](#ssr-and-javascript-ssr) app-wide.
+- `true` or `false` — Enables or disables [server-side rendering](#ssr-and-javascript-ssr) app-wide.
+- `"never"` — Prevents all pages from being evaluated on the server, on both [server-side rendering](#ssr-and-javascript-ssr) and [prerendering](#ssr-and-javascript-prerender). Using this option is recommended when building an SPA.
```
rmunn, line 203: Might be worth explaining the difference between `false` and `"never"` here: `ssr: false` in the config still allows individual pages to select `export const ssr = true`, but `ssr: "never"` in the config will completely ignore `export const ssr = whatever` in pages. Perhaps in the "`true` or `false`" bullet point you could add the sentence "Page-level `ssr` exports can override this value", and in the `"never"` bullet point you can explain that page-level `ssr` configs are ignored if `"never"` is chosen at the config level.

JeanJPNM:
Do you think the docs are clear enough?

abalmos:
Is there any known workaround while this PR bakes? As far as I can tell, this issue completely blocks final building our SPA that uses web workers extensively.

JeanJPNM:
Here are the [current workarounds](https://kit.svelte.dev/faq#how-do-i-use-x-with-sveltekit-how-do-i-use-a-client-side-only-library-that-depends-on-document-or-window)

dummdidumm (approved):

bluwy (approved):

Rich-Harris:
Thanks for the PR. I'm -1 on this — this has been discussed in the maintainer's chatroom but not here, I think, so let me explain:

In general, we want to encourage the use of SSR as widely as possible. We therefore _don't_ want people to solve the 'I can't SSR this particular page' problem by disabling SSR for the entire app. While there are situations where disabling SSR app-wide is the correct thing to do, in the many cases where that's not the case, the existence of this option would encourage bad practices.

I think something like the following is preferable:

```js
  return resolve(request, { ssr: false });
}
```

That way it can easily be scoped to specific parts of the app, for example, or even conditionally enabled for some UAs. It requires some discussion (not sure if adding options to `resolve` makes the `sequence` helper too complicated, for example) but I think that's the right direction to go in.

Conduitry:
I might suggest `resolve({ request, ssr: false })` while we can still sneak in breaking changes. We had opted for positional arguments at the time because there was only one and we only foresaw there ever being one, but once we have two, I think it makes sense to make them _all_ named, not just the new options.

NathanaelA:
@Rich-Harris  - I'm working on a Native binding (kinda like Node-Gui & NativeScript Svelte) but mine connects using GTK from node.    Having a easy method to completely disable SSR would be a god send.    I've already had to start "rewriting" .svelte files after parsing into js to change any `import x from "svelte"` to `import x from "svelte/internal"` because the svelte package.json file is set to hand me the SSR.js file when you include svelte from node.  Which then of course break the events like `onMount`.      I don't believe this PR deals with this specific issue, but before you discount adding a method to completely disable it -- there are valid use-cases where SSR is completely unwanted and there needs to be some way to disable it either during compiling or via some setting which is the point I am wanting to bring up.   :grinning:

benmccann:
Rich's suggestion would allow you to completely disable SSR if you'd like. It'd be more powerful though - it would let you decide whether to disable on a per-page / per-request basis. If you really have a need to always set it to `false` you could do that still, but it would also give you far more power to do things like enable it only for GoogleBot, disable only for `/admin` route, or whatever else you'd like to do. I think making it more flexible in that manner is a good idea

fabiot21:
I agree that Rich's suggestion is a more powerful option, but is it possible to have both options? I mean accepting "SSR never option" now (thinking on those cases where disabling SSR app-wide is the correct thing to do) and build the other better option in the future for more flexibility.

I think having the ssr never option in the short period will increase a lot Sveltekit (and Svelte) popularity among full SPA users, because they are going to have a framework with standard routing system, etc... I often see SPA users having problems with sveltekit and finally opting to use other alternatives (like routify).

benmccann:
No, we're not going to implement short-term solutions. I don't think it'd be terribly difficult to have the option passed from `handle` instead of `svelte.config.js`, so we should just do it right the first time around

abalmos:
> I think having the ssr never option in the short term will increase a lot Sveltekit (and Svelte) popularity among full SPA users, because they are going to have a popular svelte framework with standard routing system, etc... I often see SPA users having problems with sveltekit and finally opting to use other alternatives (like routify or going back to Angular / React).

@fabiot21 This is exactly my current situation. I have been hopeful this PR would land in time for the next release. We have a total "local-first" PWA which does not communicate with a backend for anything except to synchronize an IndexedDB database from a  `Worker` . I get considerable trouble from SSR even thought its of no use to me for both `Worker` and IndexedDB objects, even with the static adapter. I much enjoy everything else about Sveltekit

JeanJPNM:
> I think something like the following is preferable:
> 
> ```js
> export function handle({ request, resolve }) {
>   return resolve(request, { ssr: false });
> }
> ```
> 
> That way it can easily be scoped to specific parts of the app, for example, or even conditionally enabled for some UAs.

That's a great solution, but it doesn't solve one problem: the final app bundle. All sveltekit pages and their dependencies are bundled into `app.js` regardless of whether they actually need to be. Moving `ssr` evaluation to handle makes it too flexible, because we can't make it clear which pages will be SSRd at compile time, so every page will get bundled and that would result on the same `x is not defined` errors for side effect imports.

abalmos:
I don't understand why I'm am being marked as off topic for simply acknowledging that @Rich-Harris is right:

> ... While there are situations where disabling SSR app-wide is the correct thing to do ...

I understand that I disagree with some of the dev's poosition, but pretending that how I'm actively using SvelteKit is noise, is pretty toxic thing to do.

I honestly don't understand how disabling SSR app wide for a PWA that depends on browser specific API is bad practice. I also fail to see why that usecase is not important  enough to support just becuase some users _could_ use a feature wrong.

Maybe (ikely) I just don't understand, but I don't see how `handle` replaces `ssr: never` in my case. I need SvelteKit to _never_ run the final JS during the static build. For one example: loading data from IndexedDB in component `load` function. I don't have an API to `fetch` from, just IndexedDB to load from.

benmccann:
@abalmos the intention with Rich's suggestion was that you be able to set `ssr: false` on every page (i.e. app-wide), but also have finer-grained control. It seemed like that would solve your needs and so just saying again that you care about the feature wouldn't really change anything - we were already trying to provide a way to do what you want. The comments were hidden because they didn't add any technical details. We do know that there's a real need here - https://github.com/sveltejs/kit/issues/1650 has 49 thumbs up, which is how we track user demand for feature requests. We try to keep PRs purely focused to the implementation details. If all 49 people who wanted the feature were chiming in about it, we'd not be able to figure out how to get this out the door because the real discussion would get lost. Apologies in advance - I'm going to hide this comment and the last one as they don't relate to the PR. But I do understand the very real frustration in dealing with SSR errors when you're just trying to build a client-side app and we're not ignoring that - just trying to keep the discussion focused on how to solve it

abalmos:
But `ssr: false` isn't sufficient, unless your also going redefine that to mean what `ssr: never` means in this PR?

benmccann:
@JeanJPNM that's a great point.

I think there are two reasons things might get bundled right now
1. Kit sets `inlineDynamicImports` to `true` by default: https://github.com/sveltejs/kit/blob/d15e369bc80c7b521f1d86d6be9c6d547a0e6526/packages/kit/src/core/build/index.js#L445
2. The adapters run the Vite output through `esbuild`

If we can merge https://github.com/sveltejs/kit/pull/2580 then hopefully it would address both those points and allow Rich's suggestion on how to implement this to work still. I haven't tested to make certain, so maybe you'd like to take a peek at that branch and see if it would help here?

packages/kit/src/core/build/index.js
```diff
@@ -259,32 +259,35 @@ async function build_server(
...
-			: [];
-
-		metadata_lookup[file] = {
-			entry: client_manifest[file].file,
-			css,
-			js,
-			styles
-		};
-	});
+	if (allow_ssr) {
```
benmccann, line 266: is there any harm in having `metadata_lookup` populated even if it's not used? I wonder if it might be simpler to just leave the code as is and not add an extra condition, but want to make sure I'm not missing anything

benmccann:
> But `ssr: false` isn't sufficient, unless your also going redefine that to mean what `ssr: never` means in this PR?

Ah, I think you didn't quite understand @Rich-Harris's comment the same way I did. If `ssr: false` is passed in via `handle` then I interpreted that to mean that would be the final decision about whether the page was server-rendered. That means we'd not have to load the page to read the `ssr` config value embedded in the page and, by not loading the page on the server, we'd avoid all the pain that people experience. (Whether that means the `ssr` option still exists on the page or not I don't know. We'd have to discuss that. But I think it solves the issue either way as long as the `handle` value takes precedence)

That would provide a way to say "just shut this damn thing off", but you have an option to do it on a page or set of pages and don't have to go straight to disabling it for the whole app - which I think was the objection to having a global `never` flag with no finer-grained control.

packages/kit/src/core/build/index.js
JeanJPNM: Not really, I just thought that it would be better to not populate `metadata_lookup` when it isn't going to be used.

abalmos:
@benmccann I appreciate the conversation. That is how I understood the proposal as well, but wouldn't that scheme still have problems when an import has something browser only, like in this comment: https://github.com/sveltejs/kit/issues/1650#issuecomment-886037938 ?

My concern boils down to (and currently is) that the module context has browser only imports because of code in `load`.

Or is SvelteKite somehow able import just the `handle` function without evaluating the module context's imports?

benmccann:
> Or is SvelteKite somehow able import just the handle function without evaluating the module context's imports?

`handle` does not require the imports in your module to be evaluated. Today when `handle` calls `resolve`, `resolve` ends up loading the page on the server, but that would no longer happen with this new proposal.

abalmos:
I see. Hmm...  🤔 I'll have to dig back into the code. Clearly I have missed understood something. Thanks @benmccann, if it can work that way then it does indeed seem sufficient

benmccann:
@JeanJPNM I tested and the only thing we need is indeed to set `inlineDynamicImports: false` in the config. I sent a PR to do that by default: https://github.com/sveltejs/kit/pull/2753. In the meantime, you should be able to set it manually in the `svelte.config.js` to unblock development

JeanJPNM:
With that problem out of the way, we can implement @Rich-Harris's solution without any drawbacks (though I think that option shold be renamed to something like `loadPage`)

benmccann:
I much prefer `ssr` to `loadPage` as a name

JeanJPNM:
Closing in favor of #2804.

PR #2534, opened 2021-10-01
fix: introduce escape utility and use it to escape html attribute 

### LINKED ISSUE #2530: XSS using data-url when doing SSR
### Describe the bug

It is possible to inject a script tag in `data-url` in the `svelte-data` script tag.

E.g. if we have the search term in the query params https://www.google.com/search?q=Svelte we can create a malicious term that inject a script tag while SSR.

Happens on Firefox, Brave and Chrome, but Safari seems to protect against my POC, however it should be possible to circumvent it in other scenarios.

##### term.js
```js
export const term = '/asdf?search="></script><script>alert(1)</script><script type="application/json';
```

##### index.svelte
```html
<script context="module">
  import { term } from './term';

    const res = await fetch(term);

    return {
      props: {
        results: await res.json()
      }
    };
  }
</script>

<h1>Supersearch</h1>

```

#### Deployed POC using query params

POC Deployed: https://debug-app.vkrae.workers.dev/?search=%22%3E%3C/script%3E%3Cscript%3Ealert(1)%3C/script%3E%3Cscript%20type=%22application/json

POC Code: https://github.com/valterkraemer/debug-app/commit/1cc04c89d033bd31c0fb3ccb88ef8bb039128507

### Logs

_No response_

### System Info

```shell
System:
  OS: macOS 11.6
  CPU: (8) arm64 Apple M1
  Memory: 144.72 MB / 8.00 GB
  Shell: 5.8 - /bin/zsh
Binaries:
  Node: 14.17.6 - ~/.nvm/versions/node/v14.17.6/bin/node
  npm: 6.14.15 - ~/.nvm/versions/node/v14.17.6/bin/npm
Browsers:
  Brave Browser: 93.1.29.79
  Chrome: 94.0.4606.61
  Firefox: 92.0.1
  Safari: 14.1.2
npmPackages:
  @sveltejs/adapter-cloudflare-workers: ^1.0.0-next.21 => 1.0.0-next.23 
  @sveltejs/kit: next => 1.0.0-next.178 
  svelte: ^3.42.6 => 3.43.1
```

### Severity

### Additional Information

_No response_

dominikg:
and body of injected data script block

possibly fixes #2530 

full json escape is a proto implementation and tests fail locally for me. Unfortunately i wasn't able to tell if it's just bc fixtures need updating or there's an issue with the way it's encoded.

cc @Conduitry 

### Tests

### Changesets

### DIFF
--- a/.changeset/cyan-knives-mix.md
+++ b/.changeset/cyan-knives-mix.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Fix escaping of URLs of endpoint responses serialized into SSR response
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -47,7 +47,7 @@ function page_store(value) {
 function initial_fetch(resource, opts) {
 	const url = typeof resource === 'string' ? resource : resource.url;
 
-	let selector = `script[data-type="svelte-data"][data-url="${url}"]`;
+	let selector = `script[data-type="svelte-data"][data-url=${JSON.stringify(url)}]`;
 
 	if (opts && typeof opts.body === 'string') {
 		selector += `[data-body="${hash(opts.body)}"]`;
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -1,5 +1,6 @@
 import { normalize } from '../../load.js';
 import { respond } from '../index.js';
+import { escape_json_in_html } from '../../../utils/escape.js';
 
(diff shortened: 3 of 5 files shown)


Conduitry:
Thanks! I've tidied a couple of function names, fixed the escaping of the selector on the other end where we retrieve the serialized endpoint response, and added a changeset. I'm going to watch the tests and then I think we're ready to go.

Conduitry:
Oh, I should also say that I've pulled down the repro in #2530 and run it locally with this branch, and things seem to be working fine. The attribute value in the HTML gets escaped properly, and then it's successfully retrieved later with `document.querySelector`.

PR #2536, opened 2021-10-01
[chore] add test for malformed urls

ntchou:
Update adapter-node test. Add a test for malformed url

### DIFF
--- a/.changeset/pretty-melons-jog.md
+++ b/.changeset/pretty-melons-jog.md
@@ -1,5 +1,0 @@
----
-'@sveltejs/adapter-node': patch
----
-
-Add test for malformed url
--- a/.changeset/pretty-melons-jog.md
+++ b/.changeset/pretty-melons-jog.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-node': patch
+---
+
+Add test for malformed url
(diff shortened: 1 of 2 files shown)


packages/adapter-node/tests/smoke.js
```diff
@@ -61,4 +61,11 @@ test('passes through umlaut as encoded path', async () => {
 	server.server.close();
 });
 
+test('serve a 400 when we have a malformed url', async () => {
+	const server = await startServer();
+	const res = await fetch(`http://localhost:${PORT}//`);
```
benmccann, line 66: it's the double slash at the end that's invalid I'm assuming? I was staring at this forever trying to figure it out. My eyes just glazed over that since it was pretty subtle. Maybe a comment would be helpful
```suggestion
	// url has a double slash at the end making it invalid
	const res = await fetch(`http://localhost:${PORT}//`);
```

ntchou: Yes you're right. I'm going to apply your suggestion

.changeset/pretty-melons-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/adapter-node': patch
+---
+
+Add test for malformed url
```
dummdidumm, line 5: A changelog is not needed since this doesn't affect the "production code"

ntchou: Done

benmccann (approved):

PR #2543, opened 2021-10-01
[fix] don't expose prerender options

benmccann:
Further work on https://github.com/sveltejs/kit/issues/2249. I had some more improvements I wanted to work on, but wanted to break this off as a separate PR to make it more reviewable

CC @jthegedus

### DIFF
--- a/packages/kit/types/internal.d.ts
+++ b/packages/kit/types/internal.d.ts
@@ -1,4 +1,5 @@
 import { RequestHandler } from './endpoint';
+import { App as PublicApp, IncomingRequest } from './app';
 import {
 	ExternalFetch,
 	GetSession,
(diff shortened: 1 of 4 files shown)


packages/kit/types/internal.d.ts
```diff
@@ -139,12 +155,7 @@ export interface SSRRenderOptions {
 export interface SSRRenderState {
 	fetched?: string;
 	initiator?: SSRPage | null;
-	prerender?: {
-		fallback: string;
-		all: boolean;
-		dependencies: Map<string, ServerResponse>;
-		error: Error;
```
jthegedus, line 146: Given tests pass it is probably no longer required, but this error type is lost.

jthegedus (approved):

packages/kit/types/internal.d.ts
benmccann: Yeah, I removed it since it looked unused

benmccann:
I'm going to go ahead and merge this so I can base the next bit of work on top of it

PR #2544, opened 2021-10-02
[fix] expose init and render for adapters

benmccann:
Closes https://github.com/sveltejs/kit/issues/2249. There's still some desire to fix the TODO about whether there might be a better way to get the file path to the application, but that's outside the scope of https://github.com/sveltejs/kit/issues/2249, so I think we can close the issue and consider the TODO separately

It turns out that `init` also doesn't need to expose so many options. They're just used internally for `prerender` and `preview`

@jthegedus I'm not sure if it's possible for you to test this to see if it allows you to finally use the types? I haven't really tried doing an npm/pnpm `link` to an external project


jthegedus:
I can test this out today with my adapter, cheers for the work and keeping me updated with progress :smile:

jthegedus:
I still can't apply these types on import, I have to type a re-assignment of App for VSCode to work properly.

```js
import * as App from '../output/server/app.js';
import {toSvelteKitRequest} from './firebase-to-svelte-kit.js';

/** @type {import('@sveltejs/kit').App} */
const app = App;

app.init();
```

typing the import like this does not work for me:

```js
/** @type {import('@sveltejs/kit').App} */
import * as App from '../output/server/app.js';
```

nor this:

```js
import {init, render} from '../output/server/app.js';

/** @type {import('@sveltejs/kit').Init} */
init();
```

This reassignment is nothing though compared to the public API being correct, which I believe this change fixes, so thanks for that :)

benmccann:
I wouldn't have expected the last two to change, but I'm surprised the last one doesn't work. What do you see with the last one that isn't working?

jthegedus:
All the tooltips and go to definition just point to the import statement, not the type defined on the line above. I reloaded vscode a couple times to ensure it was restarting and ingesting the types correctly. Tooltip reads "import init". go to definition worked with clicking through the type itself. :shrug:

jthegedus:
This is a small api used by few people so having the reassignment as a "requirement" for proper typing of `init` and `render`. Can be resolved if/when the TODO is resolved

PR #2546, opened 2021-10-02
Chore/6782 lint config improvement

baseballyama:
This PR is submitted according to https://github.com/sveltejs/svelte/issues/6782.

I revised few production codes.
Therefore could you please check this revise is OK?

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -218,7 +218,7 @@ export async function load_node({
 
 				if (response) {
 					const proxy = new Proxy(response, {
-						get(response, key, receiver) {
+						get(response, key, _receiver) {
 							async function text() {
 								const body = await response.text();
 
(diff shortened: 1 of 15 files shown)


packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -218,7 +218,7 @@ export async function load_node({
 
 				if (response) {
 					const proxy = new Proxy(response, {
-						get(response, key, receiver) {
+						get(response, key, _receiver) {
```
benmccann, line 221: Why not just remove this one?

baseballyama: Type of `ProxyHandler#get` is `get?(target: T, p: string | symbol, receiver: any): any;`.
This is defined by ES2015.
Personally, in this case, I prefer to notate args properly. (In this case, actually, there are 3 args, so I notate 3 args.)
(Maybe because I like typed languages.)

But this is just personal preference.
Therefore if you like to remove `_receiver`, of cause I will remove it.

benmccann:
Pushed those changes in https://github.com/sveltejs/kit/commit/63303efe20e7a354d6777fba49403e47221d617e

baseballyama:
Thank you!

Issue #2545, opened 2021-10-02
Support Dockerfile/Kubernetes spec generation on project bootstrap


c16a:
### Describe the problem

Svelte developers need to go through the extra effort of writing Dockerfiles or Kubernetes specifications to deploy the assets.

### Describe the proposed solution

It would be a better developer experience if the project starter also generates a `Dockerfile` and/or a Kubernetes `Deployment` YAML.

This can be an opt-in feature in the interactive prompt.

### Importance

would make my life easier

c16a:
@dominikg This merely adds two optional files during `npm init svelte@next my-app` by prompting the user if they want to enable support for Docker and Kubernetes.

This doesn't imply the devserver via k8s, but offers a template for `Deployment` and `Service`. Ideally, you'd need to add the `sveltjs-static` adapter and have the Docker image expose the static files via httpd or nginx. But doing this would need us to add the adapter which seemed restrictive. Would be open to doing that, if it makes more sense.

dominikg:
What exactly are you looking for?

I don't think create-svelte is the right place to add docker support. Ideally this could be offered via community supported packages, eg a svelte-add or an adapter

c16a:
Alright, will close it.

dominikg:
I do think supporting docker/k8s in some form is worth talking about and implementing. 
But there are different ways to get there and asking users first thing in create svelte and offering suboptimal images that use npm run dev isn't helpful. Novice users would end up with a dangerous setup and experienced users don't save time either as the have to edit it.

PR #2550, opened 2021-10-03
Docker and Kubernetes support during npm init

c16a:
This commit adds support for adding Docker and Kubernetes support during
project creation using npm init. The following files are additionally
created
* Dockerfile and .dockerignore (when Docker support is enabled)
* k8s.yaml with Deployment and Service (when Kubernetes support is
enabled)

Signed-off-by: Chaitanya Munukutla <chaitanya.m61292@gmail.com>

### Tests

### Changesets

### DIFF
--- a/.changeset/yellow-terms-lay.md
+++ b/.changeset/yellow-terms-lay.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': minor
+---
+
+Add Docker and Kubernetes support during npm init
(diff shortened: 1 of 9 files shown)


c16a:
Fixes #2545

dummdidumm:
This to me looks a little too specific to me (although I'm fine with it if other members feel it is a nice addition). For me this might be better suited to be added to the `svelte-adders` package

dominikg:
what exactly is the goal of this?  i'd like to have a clear picture first and discuss how this can be achieved outside kit core.

why would you run the devserver via k8s ?

benmccann:
Thanks! It will be great to have Docker / k8s support. But agree with the others that this would be better as a Svelte Adders project, so will close this. Please do pursue it as an Adder though and let us know if we can help with that

Issue #2517, opened 2021-09-28
Source maps in adapter-node


alastaircoote:
### Describe the problem

Because adapter-node transpiles the original source the error stacks you see in production aren't always that useful in tracking down an issue. I thought it might be as simple as adding source maps to the esbuild config:

```
esbuild: function (config) {
    config.sourcemap = 'inline';
    return config;
}
```

which _sort of_ works, but now the source lines it gives me are in `.svelte-kit/build/app.js`. If I add:

```
build: {
 	sourcemap: 'inline'
}
```

to Sveltekit's vite config then `.svelte-kit/build/app.js` _does_ have source maps but [esbuild doesn't merge source maps](https://github.com/evanw/esbuild/issues/933). I've played a little with the [merge-source-maps](https://www.npmjs.com/package/merge-source-maps) module but it definitely isn't working with SvelteKit out of the box. 

### Describe the proposed solution

I'm not entirely sure what the answer is here but wanted to flag it as a desired feature. I'm happy to dig further into it myself but might benefit from the wisdom of someone more familiar with SvelteKit's internals. merge-module-maps fails because it tries to load `${PROJECT_DIR}/kit/dist/node.js` which is presumably `@svelte/kit/dist/node.js` but I'm not totally sure where that's being imported.

### Importance

nice to have

benmccann:
The easy solution would probably be to make `esbuild` bundling optional. It was only added in https://github.com/sveltejs/kit/pull/1648, so shouldn't be required since we always got away without doing it before.

The better solution is probably to see if we can make Vite responsible for all bundling. That would also solve the problem that @ignatiusmb is trying to figure out of how to have the adapter reference the application output file without hardcoding it - the adapter would no longer need to because it could be kept internal to `build/index.js`. Instead, the adapter could just be a library and SvelteKit could do all the building instead of the adapter being in charge of the build process.

Conduitry:
I'd hesitate to say we got away without doing it before that PR. `@sveltejs/kit` was previously a _production_ dependency of all apps using SvelteKit and the Node adapter, which is pretty bad.

Issue #798, opened 2021-03-31
Type reflection issues


MirrorBytes:
**Describe the bug**
Typescript metadata reflection does not seem to work at all. At first I thought it was related to the graphql adder, but it turns out that type-graphql works perfectly fine outside of sveltekit in an api folder served by `vercel dev`. This is not a permanent solution though considering there are existing issues pertaining to MIME types: #781.

**Logs**
This is produced in both the response and the dev console (not browser).

```bug
Error when evaluating SSR module /src/resolvers/user.ts:
Error: Unable to infer GraphQL type from TypeScript reflection system. You need to provide explicit type for argument named 'data' of 'register' of 'UserResolver' class.
    at Object.findType (/Users/bobby/Documents/phorm-kit-vercel/node_modules/type-graphql/dist/helpers/findType.js:19:15)
    at Object.getParamInfo (/Users/bobby/Documents/phorm-kit-vercel/node_modules/type-graphql/dist/helpers/params.js:10:49)
    at /Users/bobby/Documents/phorm-kit-vercel/node_modules/type-graphql/dist/decorators/Arg.js:14:25
    at eval (/src/resolvers/user.ts:14:54)
    at __decorate (/src/resolvers/user.ts:9:24)
    at eval (/src/resolvers/user.ts:74:1)
    at instantiateModule (/Users/bobby/Documents/phorm-kit-vercel/node_modules/vite/dist/node/chunks/dep-66eb515d.js:69030:166) (x18)
Error: Unable to infer GraphQL type from TypeScript reflection system. You need to provide explicit type for argument named 'data' of 'register' of 'UserResolver' class.
    at Object.findType (/Users/bobby/Documents/phorm-kit-vercel/node_modules/type-graphql/dist/helpers/findType.js:19:15)
    at Object.getParamInfo (/Users/bobby/Documents/phorm-kit-vercel/node_modules/type-graphql/dist/helpers/params.js:10:49)
    at /Users/bobby/Documents/phorm-kit-vercel/node_modules/type-graphql/dist/decorators/Arg.js:14:25
    at eval (/src/resolvers/user.ts:14:54)
    at __decorate (/src/resolvers/user.ts:9:24)
    at eval (/src/resolvers/user.ts:74:1)
    at instantiateModule (/Users/bobby/Documents/phorm-kit-vercel/node_modules/vite/dist/node/chunks/dep-66eb515d.js:69030:166)
```

**To Reproduce**
[...]

Issue #2523, opened 2021-09-29
[Sveltekit v next.175+ with Netlify adapter] Cannot find module 'stream/package.json'


f-elix:
### Describe the bug

Builds now fail on Netlify since version next.175. Local builds still work, and Netlify builds work with version next.174. I have another Sveltekit project that runs on Vercel and I have no build problems there.

I would like to provide more detail but I have no idea what's going on here. I provided the error log below.

### Reproduction

This is a reproduction repo: https://github.com/f-elix/sveltekit-netlify-build-fail

It's essentially the Sveltekit demo app with netlify-adapter. I tried to deploy it on Netlify and got the same error as my main project.

### Logs

```shell
11:36:32 AM: ────────────────────────────────────────────────────────────────
11:36:32 AM:   Dependencies installation error                               
11:36:32 AM: ────────────────────────────────────────────────────────────────
11:36:32 AM: ​
11:36:32 AM:   Error message
11:36:32 AM:   A Netlify Function failed to require one of its dependencies.
11:36:32 AM:   Please make sure it is present in the site's top-level "package.json".
​
11:36:32 AM:   In file "/opt/build/repo/.netlify/functions-internal/__render.js"
11:36:32 AM:   Cannot find module 'stream/package.json'
11:36:32 AM:   Require stack:
11:36:32 AM:   - /opt/buildhome/.netlify-build-nvm/versions/node/v12.22.6/lib/node_modules/@netlify/build/node_modules/@netlify/zip-it-and-ship-it/src/node_dependencies/resolve.js
11:36:32 AM:   - /opt/buildhome/.netlify-build-nvm/versions/node/v12.22.6/lib/node_modules/@netlify/build/node_modules/@netlify/zip-it-and-ship-it/src/node_dependencies/index.js
11:36:32 AM:   - /opt/buildhome/.netlify-build-nvm/versions/node/v12.22.6/lib/node_modules/@netlify/build/node_modules/@netlify/zip-it-and-ship-it/src/main.js
11:36:32 AM:   - /opt/buildhome/.netlify-build-nvm/versions/node/v12.22.6/lib/node_modules/@netlify/build/src/plugins_core/functions/index.js
11:36:32 AM:   - /opt/buildhome/.netlify-build-nvm/versions/node/v12.22.6/lib/node_modules/@netlify/build/src/commands/get.js
[...]

Issue #1547, opened 2021-05-24
Cannot build with .node files and Vercel adapter


Nick-Mazuk:
**Describe the bug**

Cannot build Sveltekit while using Sharp and the Vercel adapter. Sharp works in development. I suspect this issue applies to all dependencies with `.node` files.

This seems unrelated to the ["error when including packages"](https://kit.svelte.dev/faq#packages) question in the FAQ or https://github.com/sveltejs/kit/issues/737

**Logs**
```bash
> Using @sveltejs/adapter-vercel
 > node_modules/.pnpm/sharp@0.28.2/node_modules/sharp/lib/output.js:4:22: error: No loader is configured for ".node" files: node_modules/.pnpm/sharp@0.28.2/node_modules/sharp/build/Release/sharp.node
    4 │ const sharp = require('../build/Release/sharp.node');
```

**To Reproduce**

1. Initialize new Sveltekit project
2. Install sharp `pnpm i sharp`
3. Create a new endpoint and import sharp
4. Add Vercel adapter
5. `pnpm build`

Reproduction repo: https://github.com/Nick-Mazuk/sveltekit-sharp

**Expected behavior**

Site would build without issues.

**Stacktraces**
If you have a stack trace to include, we recommend putting inside a `<details>` block for the sake of the thread's readability:

<details>
  <summary>
Stack trace
</summary>

```bash
pnpm build

> ~TODO~@0.0.1 build /Users/nickmazuk/Documents/Github/sveltekit-sharp
> svelte-kit build

vite v2.3.3 building for production...
✓ 14 modules transformed.
.svelte-kit/output/client/_app/manifest.json                    1.21kb
.svelte-kit/output/client/_app/layout.svelte-39a4acb4.js        0.49kb / brotli: 0.29kb
.svelte-kit/output/client/_app/pages/index.svelte-af73b2f3.js   0.79kb / brotli: 0.42kb
.svelte-kit/output/client/_app/error.svelte-8eafbf20.js         1.19kb / brotli: 0.57kb
.svelte-kit/output/client/_app/assets/start-a8cd1609.css        0.16kb / brotli: 0.10kb
.svelte-kit/output/client/_app/chunks/vendor-3681a316.js        5.05kb / brotli: 1.98kb
.svelte-kit/output/client/_app/start-22b096ce.js                17.18kb / brotli: 5.50kb
vite v2.3.3 building SSR bundle for production...
[...]

PR #2580, opened 2021-10-10
[breaking] do a single bundling with Vite removing esbuild

benmccann:
## Motivation

The second bundling step is a pain. It leads to a whole class of new bugs where users are now exposed to both all Vite bugs and all esbuild bugs (or sometimes not bugs, but purposely omitted features such as lack of TypeScript reflection). It also makes it substantially more difficult to understand difficult edge cases because you need to learn two build systems.

## Description

Removes `esbuild` from the `adapter-node` and `netlify` adapters. Rich suggested that Vercel and Cloudflare require everything to be in a single file, so I've left the `esbuild` step for those adapters

## Related issues

- Fixes https://github.com/sveltejs/kit/issues/2517
- Fixes https://github.com/sveltejs/kit/issues/798
- Closes https://github.com/sveltejs/kit/issues/2581
- May either fix or at least help with https://github.com/sveltejs/kit/issues/1547
- It might fix this bug though we also have another fix out for it: https://github.com/sveltejs/kit/issues/2523#issuecomment-941743994
- I think there were a couple others I forgot as well

# Breaking changes

**Adapter authors**

- Import app via `$server-build` alias instead of hard-coded relative path
- Need to specify `serverEntryPoint`
- Simply import `@sveltejs/kit/install-fetch` directly instead of more complicated shim

**Adapter users**

- I removed the options for setting environment variable names in `adapter-node`. Users can create a custom entry point if they'd like to create environment variables with their own naming scheme
- Users of `adapter-node`'s `entryPoint` option should switch to the new `serverEntryPoint` option

# Open questions

- Do we like the name `$server-build` for the alias name?
- Is there some way we can allow typing the app for adapter authors? `$server-build` doesn't seem to be pulling in types at the moment. I'll leave this for another PR since it's not a regression, but it would be nice to think about


benmccann:
Closing for now. Need to get the pre-req PR working first

documentation/docs/14-configuration.md
```diff
@@ -195,6 +196,10 @@ See [Prerendering](#ssr-and-javascript-prerender). An object containing zero or
 
 Enables or disables the client-side [router](#ssr-and-javascript-router) app-wide.
 
+### serverEntryPoint
+
+A file path to a custom entry point for the server. Passed to `vite.build.rollupOptions.input`. See [the Rollup docs](https://rollupjs.org/guide/en/#input) for more info.
+
```
Rich-Harris, line 202: Not sure I understand this — why is this being documented as part of `config.kit`? isn't it purely an adapter concern?

benmccann: It's not really an adapter concern after this change because the code for it is now contained in `build/index.js`. Vite is now doing the bundling and it needs to know what the entry point is. Most of the time that will be an adapter-provided file. But the user may want to override it. Not sure if you have any other ideas for how we might specify this and allow it to be overriden?

benmccann: I've updated it to be an adapter option

antony:
Just to confirm that I have tested my own app against the netlify adapter here to determine if this fixes the very strange missing transient dependency missing issues that occur when trying to `adapt` (where some-lib is a transient dependency of an installed package):

```
> Using @sveltejs/adapter-vercel
 > .svelte-kit/output/server/chunks/__layout-158d2ad5.js:4:7: error: Could not resolve "some-lib"
```

and can confirm that this PR resolves it.

tylerbrostrom:
>…Cloudflare require[s] everything to be in a single file, so I've left the esbuild step for those adapters.

It should also be possible to remove the `esbuild` step `adapter-cloudflare` and `adapter-cloudflare-workers` given that [JavaScript modules are now supported on Cloudflare Workers](https://blog.cloudflare.com/workers-javascript-modules/) (and are the prefered format moving forward).

Granted, this _would_ be a breaking change for `adapter-cloudflare-workers`, as it uses the older, alternative “service-worker” format.

Rich-Harris:
closing in favour of #2931

Issue #2588, opened 2021-10-12
Hashchange event does not trigger.


whataboutpereira:
### Describe the bug

Hashchange event doesn't want to work with SvelteKit v1.0.0-next.180.

Neither `<svelte:window on:hashchange>` nor adding the event listener directly with `window.addEventListener` has any effect.

There are no visible errors anywhere.

If I use the same in a plain HTML file in a browser, there are no problems triggering the event.

Browsers tested: Chrome, Firefox, Safari.

### Reproduction

```
<script>
    import { onMount } from 'svelte';

    const log = () => console.log('hashchange', location.hash);

    onMount( ()	=> window.addEventListener('hashchange', log ) );

</script>

<svelte:window on:hashchange={log}/>

<a href="#test1">test 1</a>
<a href="#test2">test 2</a>

<p id="test1">test1</p>
<p id="test2">test2</p>
```

### System Info

```shell
System:
    OS: Linux 5.4 CentOS Linux 8
    CPU: (1) x64 Intel(R) Xeon(R) CPU E3-1220 v5 @ 3.00GHz
    Memory: 1.85 GB / 2.00 GB
    Container: Yes
    Shell: 4.4.20 - /bin/bash
  Binaries:
    Node: 12.22.3 - /usr/bin/node
    npm: 7.24.1 - /usr/local/bin/npm
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.180 
    svelte: ^3.42.6 => 3.43.1
```

### Severity

serious, but I can work around it

whataboutpereira:
Wonderful, thanks! Surprised it has gone under the radar this long. :)

PR #2591, opened 2021-10-12
[fix] fixes for firing of hashchange event

benmccann:
fixes for `hashchange` (https://github.com/sveltejs/kit/issues/2588) which was implemented in the last PR (https://github.com/sveltejs/kit/pull/2590)


Co1eCas3:
I upgraded yesterday evening after being guided to this PR in an attempt to resolve issues relating to detecting hashchange events. Since upgrading, my issue is not resolved -- not only are hashchange events still not being detected but $page.url.hash is only ever an empty string.

For clarity, I have `$: console.log($page.url.hash)` in my component for the following images:

This is when I navigate directly to the hash:
image

This is navigating via an <a> tag (get the log in my click but not from the reactive statement)
image

Issue #1859, opened 2021-07-09
Netlify deploy failing when trying to bundle functions


felixakiragreen:
**Describe the bug**
When trying to deploy to Netlify the build fails at the step: "2. Functions bundling"
`yarn dev`, `yarn build`, `yarn run preview`, `netlify dev` all work with no issues, it only happens when I try to deploy.

**Logs**
```
 2. Functions bundling                                         
8:27:53 PM: ────────────────────────────────────────────────────────────────
8:27:53 PM: ​
8:27:54 PM: Packaging Functions from functions directory:
8:27:54 PM:  - render/index.js
8:27:54 PM: ​
8:27:57 PM: ​
8:27:57 PM: ────────────────────────────────────────────────────────────────
8:27:57 PM:   Dependencies installation error                               
8:27:57 PM: ────────────────────────────────────────────────────────────────
8:27:57 PM: ​
8:27:57 PM:   Error message
8:27:57 PM:   A Netlify Function is using "bufferutil" but that dependency has not been installed yet.
8:27:57 PM: ​
8:27:57 PM:   By default, dependencies inside a Netlify Function's "package.json" are not automatically installed.
  There are several ways to fix this problem:
8:27:57 PM:     - Removing your Function's "package.json" and adding the dependencies to the project's top-level "package.json" instead. This is the fastest and safest solution.
8:27:57 PM:     - Running "npm install" or "yarn" inside your Netlify Function in your build command.
8:27:57 PM:     - Adding the following plugin to your "netlify.toml":
8:27:57 PM: ​
8:27:57 PM:   [[plugins]]
8:27:57 PM:   package = "@netlify/plugin-functions-install-core"
8:27:57 PM: ​​
8:27:57 PM:   In file "/opt/build/repo/frontend/functions/render/index.js"
8:27:57 PM:   Cannot find module 'bufferutil'
```

**To Reproduce**
I am using the latest versions: (hardcoded into package.json to prevent issues)
```
"@sveltejs/adapter-netlify": "^1.0.0-next.19",
"@sveltejs/kit": "^1.0.0-next.19",
```

**Expected behavior**
I expect it to deploy. I have deployed 3 svelte-kit  to netlify, so this isn't my first rodeo. However, they were with previous versions: `"@sveltejs/adapter-netlify": "1.0.0-next.12"` & "@sveltejs/adapter-netlify": "1.0.0-next.14"`

I tried adding the plugin, but it didn't help.

**Stacktraces**
[...]

PR #2605, opened 2021-10-14
[docs] use Netlify's esbuild node_bundler command

brewhousedigital:
Netlify is trying out a new node_bundler system and you can enable it through the flag 'esbuild'. Without this flag, users may have to install additional packages into their package.json file that don't relate to their project. Using the new node_bundler, you don't need to do this anymore.

### Related to this issue:
#1859


benmccann:
I'm going to wait and see how that issue plays out. I'll leave this open in the meantime

PR #2609, opened 2021-10-15
[fix] Invalidating a route now fires sveltekit:navigation-start event

### LINKED ISSUE #2528: `sveltekit:invalidate-start` and `sveltekit:invalidate-end` events
### Describe the problem

My use case for firing `invalidate-start` is to reload data on same path link and show loading indicator.
(And also to show loading indicator on navigation to different path).

I also think that it would be consistent to have both `start` and `end` events.

Dispatch `sveltekit:invalidate-start` on invalidate()

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

henninggross:
Closes #2528 

By now also calling `router.notify()` when invalidating an URL it is made sure that the `sveltekit:navigation-start` event is fired upon invalidation.
This way one can for example show a loading indicator when trying to reload the same page.

I am calling `router.notify()` instead of just dispatching the `sveltekit:navigation-start` event to also update `navigating`.

Other than that it is just more consistent to have a `start` event when also having an `end` event as one could expect both to be fired the same way they are when navigating normally.

### Tests

### Changesets

### DIFF
--- a/.changeset/modern-berries-kick.md
+++ b/.changeset/modern-berries-kick.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Invalidating a route now fires sveltekit:navigation-start event


benmccann:
This doesn't seem quite right to me because, if the page URL doesn't change, I'm not sure we should be firing `navigation-start` and `navigation-end`

Issue #2602, opened 2021-10-13
Adapter Static stops `svelte-kit dev` without routes, no error shown


benaltair:
### Describe the bug

If you have no routes in your project, `npm run dev` will start `svelte-kit dev` but will hang with no errors or output. Surveying the file structure, it's clear that adding the missing route will solve the issue. It would be good to show an error in the terminal warning the user that no routes are detected.

### Reproduction

Add `@sveltejs/adapter-static@next` to a skeleton Kit project and run `npm run dev` once the project is set up.

### Logs

```shell
PS D:\Github\sev-website> npm run dev
> sev-website@0.0.1 dev D:\Github\sev-website
> svelte-kit dev
```
As you can see above, no errors, it just hangs indefinitely.
```


### System Info

```shell
System:
    OS: Windows 10 10.0.19043
    CPU: (16) x64 AMD Ryzen 7 3800X 8-Core Processor
    Memory: 17.12 GB / 31.92 GB
  Binaries:
    Node: 14.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.5 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.14.11 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 94.0.4606.81
    Edge: Spartan (44.19041.1266.0), Chromium (94.0.992.47)
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    @sveltejs/adapter-static: ^1.0.0-next.21 => 1.0.0-next.21
    @sveltejs/kit: next => 1.0.0-next.183
    svelte: ^3.42.6 => 3.43.2
```

### Severity

annoyance

### Additional Information

This is a nice to have and would increase the accessibility of Svelte Kit to those new with the framework.

PR #2614, opened 2021-10-16
[fix] avoid infinite loop if no routes found

baseballyama:
fix: https://github.com/sveltejs/kit/issues/2602

Please refer [My Comment](https://github.com/sveltejs/kit/pull/2614#issuecomment-945034814) for details.
~~I handled only `src/routes/index.svelte` because if another file is not found, we can notice it on browser.~~
~~(We can see a 404 message on browser.)~~

~~(Is it correct that always `src/routes/index.svelte` exists?🤔)~~
-> According to the comment, may not have `src/routes/index.svelte`.

### Tests

### Changesets

### DIFF
--- a/.changeset/serious-bears-listen.md
+++ b/.changeset/serious-bears-listen.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-[fix] avoid infinite loop and log warning if no routes found
+[fix] avoid infinite loop if no routes found
--- a/packages/kit/src/cli.js
+++ b/packages/kit/src/cli.js
@@ -92,10 +92,6 @@ prog
 		try {
 			const watcher = await dev({ port, host, https, config });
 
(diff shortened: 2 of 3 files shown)


Conduitry:
This isn't the right solution. The root path might be served by an endpoint, or it might not exist at all (especially if the SvelteKit app is behind something that only sends certain URLs to it), or the user may simply be [using a different file extension](https://kit.svelte.dev/docs#configuration) for their page routes.

baseballyama:
@Conduitry 

Thank you for telling point what I'm not aware of.
And I updated code.

The direct cause is an infinity loop here.
https://github.com/sveltejs/kit/blob/26e81e8b62192d1b9e3a958742a6d1f249cbca7b/packages/kit/src/core/create_app/index.js#L132-L142

So I updated it to prevent infinity loop.

And if none of the route components are present, this infinity loop comes.
But I'm not sure that this is a possible use case for kit users.

So right now I added warn message if there is no route component.

Issue #1386, opened 2021-05-08
The navigation announcer's CSS could be inlined


ebeloded:
**Describe the bug**
Production build generates a `start-[hash].css` file with the following content:
```css
#svelte-announcer.svelte-1j55zn5{position:absolute;left:0;top:0;clip:rect(0 0 0 0);clip-path:inset(50%);overflow:hidden;white-space:nowrap;width:1px;height:1px}
```
This file is included in built HTML, but is not required by the application.

**To Reproduce**

1. Init svelte kit skeleton
2. Build

**Expected behavior**

Only the files needed for the application should be in the resulting build.

<details>
  <summary>npmPackages</summary>
    @sveltejs/kit: next => 1.0.0-next.102 
    svelte: ^3.38.2 => 3.38.2 
</details>

**Severity**
Not a big issue, but requires an extra request to get the css, and negatievely reflects framework tidiness. 

**Other questions**
Is there some unused javascript shipped with the bundle?

babichjacob:
This CSS is used to visually hide [the announcer text on page navigation](https://github.com/sveltejs/kit/issues/975#issuecomment-818109880) such that screen readers can still read it.

ebeloded:
Thank you for linking the explanation about the announcer, @babichjacob. 

However, I think that my issue still stands: the announcer div can have its styles inlined, which wouldn't require the extra CSS file.

My issue with the CSS file is that it needs to be generated, and the client needs to request it, which increases request count.

babichjacob:
Relevant lines of code: 

https://github.com/sveltejs/kit/blob/5023e98f73b87a849b680255f9474f27e649cc82/packages/kit/src/core/create_app/index.js#L181-L200

I'm not sure, without another maintainer commenting, if a PR for this would be accepted, especially if it only covers this specific case *manually* rather than being a general solution to https://github.com/sveltejs/kit/issues/962

Rich-Harris:
This is indeed a special case of #962 so I'll close this

PR #2620, opened 2021-10-17
[feat] `inlineCss` option

### LINKED ISSUE #962: Can CSS be inlined in the server-rendered pages ?
Currently, HTML is rendered in the response by the server, but the CSS is linked through <link> attributes, thus delaying proper page rendering until the stylesheet request is done.

Is it possible to inline CSS (or at least the 'start' stylesheet) like Next.JS does? Or is it planned?

untp:
fixes #962

### Tests

### Changesets


Rich-Harris:
Thanks — have updated this PR so that it's once again in a mergeable state.

In principle I'm in favour of this change (though I wonder if it should be a threshold rather than a toggle — if one page had a very large amount of CSS, you might want that page to use `<link>` rather than `<style>`.

Unfortunately there's one major problem. If you visit a page that has styles from components A, B and C, you'll have this...

```html
<style>
  /* A */
  /* B */
  /* C */
</style>
```

...then if you navigate to a page that has components B, C and D you'll get this:

```html
<style>
  /* A */
  /* B */
  /* C */
</style>

<link rel="stylesheet" href="/assets/B.css">
<link rel="stylesheet" href="/assets/C.css">
<link rel="stylesheet" href="/assets/D.css">
```

If you then navigate back to the first page, you'll load A again.

In contrast, when styles are always included as links, the first navigation will only load D, and the second navigation won't load anything at all. Aside from resulting in faster navigation and less wasted bandwidth, this also prevents a situation where the CSSOM is cluttered with duplicated selectors.

The CSS loading is done by Vite here — [here's where it checks to see if styles are already in the page](https://github.com/vitejs/vite/blob/71868579058512b51991718655e089a78b99d39c/packages/vite/src/node/plugins/importAnalysisBuild.ts#L60-L61).

I'm not sure what the best solution to this is. Perhaps we include the links as well as the `<style>`, to fool Vite, but add the [`disabled`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-disabled) attribute so that they remain inert?

Rich-Harris:
Implemented `<link disabled>`, and took the opportunity to tidy up the page rendering code a bit since it was starting to get somewhat hairy

untp:
Thanks for the commits.

I think using disabled attribute is the best option.

It is better to using threshold. Because there are some situations that it is needed. For example, when using cloudflare workers adapter, it is better to keeping server script size small, because cloudflare workers script can be up to 1 MB after compression. Also developer may want to reduce bandwidth with only inlining small css files.

The test is failing because of [svelte-announcer](https://github.com/sveltejs/kit/blob/master/packages/kit/src/core/create_app/index.js) style. We can inline styles of svelte-announcer as discussed in #1386. Whether inlineCss enabled or not, svelte-announcer's style should be inlined.

.changeset/tiny-files-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[feat] inlineCss option
```
Rich-Harris, line 5:
```suggestion
add inlineStyleThreshold option, below which stylesheets are inlined into the page
```

gustavopch:
What's the unit in `inlineStyleThreshold`? Bytes? BTW, it may be too late, but it could have been named `minBytesToInlineStyle` to make that self-evident.

PR #2632, opened 2021-10-18
[feat] add prerender.subfolders setting

### LINKED ISSUE #1443: Ability to prerender to `about.html` instead of `about/index.html` like Sapper

bfanger:
Fixes #1443

Usage in svelte.config:

```js
kit: {
  prerender: {
    subfolders: false,
  },
   ...
```
Setting the `kit.prerender.subfolders` setting to **false** (default is **true**) will change the filename generation  from "/about/index.html"  to "/about.html"

Inspiration for the `subfolders` name came from nuxt:
https://nuxtjs.org/docs/configuration-glossary/configuration-generate/#subfolders


Rich-Harris:
Thank you, and apologies for not getting round to this PR until now. One thing I'm not 100% sure about is the `subfolders` name — something about it feels a little non-descriptive. What about something like `createIndexFiles`?

bfanger:
Apology is appreciated.
I agree that the naming could be better, but don't think `createIndexFiles ` is it, that name makes sense from the kit codebase standpoint, but from a user perspective it's about is not generating folders, the index files are a byproduct..

if we change the config from a boolean to an enum, that show the intent better:

`routeOutput: 'file' | 'folder'`

What do you think?

Rich-Harris:
I think a boolean makes more sense to be honest. We discussed in the maintainers chatroom and decided to go with `createIndexFiles` — it's the sort of thing that's easy to find in the docs when you start looking for information about `index.html`

Rich-Harris (approved):

isapir:
Is `kit.prerender.subfolders: false` still a thing?  

I get an error:
> Error: Unexpected option config.kit.prerender.subfolders

I have this in `svelte.config.js`: 

```
/**
 * from https://kit.svelte.dev/docs/adapter-static
 */

export default {
	kit: {
		adapter: adapter({
			// default options are shown. On some platforms
			// these options are set automatically — see below
			pages: 'build',
			assets: 'build',
			fallback: null,
			precompress: false,
			strict: true
		}),
		prerender: {
			subfolders: false,
		},
	}
};
```

bfanger:
```ts
// src/routes/+layout.js
export const trailingSlash = 'always';
```


https://kit.svelte.dev/docs/page-options#trailingslash

Issue #2611, opened 2021-10-15
Modularize SvelteKit


benmccann:
### Describe the problem

A popular point of feedback from the other maintainers is that SvelteKit is not as modular as we had hoped we might make it at the beginning of the project

### Describe the proposed solution

We could possible split out a few different pieces:
- The core router without file system support
- The file system router depending on the core router
- Kit itself with Vite integration. We've already split out `vite-plugin-svelte` and I'm not sure there's much more that makes sense to split out there
- Put optional features in plugins (e.g. AMP https://github.com/sveltejs/kit/issues/2603)

I'd written a bit about this in https://github.com/sveltejs/rfcs/pull/36 originally for Sapper and tried to update the proposal a bit for SvelteKit though further updates may be needed

I don't know whether it'd be possible to make the router implementation pluggable, but I think it'd be a cool goal. I also think it'd be nice to remove any globals which would help for testing (right now you can mostly only write tests if you use Jest and mock the globals) and allow you to have multiple router instances (@pngwn has long asked for microfrontend support and while that's not something I'm overly familiar with I could imagine multiple routers getting pulled in)

### Importance

nice to have

jarda-svoboda:
A pluggable router might be a way how to possibly create something like this https://github.com/sveltejs/kit/issues/1251 🤔

bertybot:
Do you think we could put in the option to override the default start.js file? The start.js file makes the system a whole lot less modular. We force an entry into the client with a bunch overhead even if the user doesn't use most of the features, and it also doesn't allow any configuration of the router.  

It would be nice if we could write our own start.js or Sveltkit constructor like other JS frameworks. Vue allows you to set the entry with a Vue constructor and configure the features you want al a carte like a Router. We could fall back to the default one if the user don't supply one.  It would also save on bundle size since right now the router is still loaded even if the user doesn't use it and ends up as a lot of dead code.

PR #2656, opened 2021-10-21
[breaking] only route pages on the client-side

benmccann:
Right now we have to send a list of all endpoints to the client, but 99% of the time they aren't used for anything and so it's generally unnecessary bandwidth overhead.

Removing this means that developers need to put `rel="external"` on any link to an endpoint. This seems like a reasonable trade-off since it is very rare to link to endpoints. Endpoints are primarily gotten via `fetch` which doesn't utilize the router.

In addition to reducing the size of the client JS, it also will make it slightly easier to refactor out the router (https://github.com/sveltejs/kit/issues/2611) since it will have one less concern to deal with

### DIFF
--- a/documentation/docs/01-routing.md
+++ b/documentation/docs/01-routing.md
@@ -130,8 +130,6 @@ The job of this function is to return a `{ status, headers, body }` object repre
 - `4xx` — client error
 - `5xx` — server error
 
-> For successful responses, SvelteKit will generate 304s automatically.
-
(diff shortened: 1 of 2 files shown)


Conduitry:
Discussions about `sveltekit:external` aside, I do think this is probably a reasonable change to make to how routing is handled if it means less stuff getting sent to the client. Rich should probably weigh in, though, and if we're going to be making this change, we should make sure the docs are updated to make the new behavior clear.

benquan:
This reasoning is completely ommited from the documentation. This post is the only reference I have found to do this.

PR #2663, opened 2021-10-22
[feat] pass `ComponentLoader` to `handle`

benmccann:
This is needed if you want to load any `.svelte` file from `handle`

One potential usecase would be allowing you to use any other router with SvelteKit allowing you to leverage SvelteKit adapters, HMR, etc. without using our router. I'm sure there'd be other use cases as well.

I tried using this with Routify and it basically worked except that Routify doesn't seem to support SSR.

Even if we don't expose this yet, separating out `ComponentLoader` from the routing options might be a nice change.

I have purposefully not documented this as I consider it an experimental API still. We can add it to the docs after we flesh it out.

### DIFF
--- a/.changeset/three-steaks-ring.md
+++ b/.changeset/three-steaks-ring.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[feat] experimental: pass ComponentLoader to handle
--- a/packages/kit/src/core/build/index.js
+++ b/packages/kit/src/core/build/index.js
@@ -301,6 +301,11 @@ async function build_server(
 				.replace('%svelte.head%', '" + head + "')
 				.replace('%svelte.body%', '" + body + "')};
 
+			const loader = {
+				fixStackTrace: ({ error }) => {},
+				loadComponent
+			};
+
 			let options = null;
 
 			const default_settings = { paths: ${s(config.kit.paths)} };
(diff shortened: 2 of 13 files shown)


jakobrosenberg:
Been having a blast experimenting with this today.

It seems there are two approaches for adding third party routers (sorry if I'm stating the obvious):

### Option A - loadComponent

**hooks.js**
```javascript
import { Router, createRouter } from '@roxi/routify'
import routes from '../.routify/routes.default.js'

export async function handle({ request, loader }) {

    // create a router instance
    const router = createRouter({ routes })

    // let router run async functions, eg. fetch dynamic imports, run preloads, guards etc.
    await router.url.replace(request.path) 

    // create SSR node    
    const ssrNode = await loader.loadComponent({ id: 'src/routes/root.svelte' });
    
    // inject the router into  `src/routes/root.svelte`
    const rendered = ssrNode.module.default.render({ router });    
...
```

**src/routes/root.svelte**
```html
<script>
    import { Router } from '@roxi/routify'
    export let router
</script>

<Router {router} />
```


### Option B - plain SSR

**hooks.js**
```javascript
import { Router, createRouter } from '@roxi/routify'
import routes from '../.routify/routes.default.js'

export async function handle({ request, loader }) {

    const router = createRouter({ routes })
    await router.url.replace(request.path)
    const rendered = Router.render({router})
```


Option B doesn't require a `root.svelte` file, but I'm not sure about pros and cons of using `loadComponent`

benmccann:
Option B will not return the list of JS and CSS files needed to populate the page's `head` the way that `loadComponent` will

Rich-Harris:
I'm afraid I think this is a bit of a non-starter, aside from whether it's a desirable API — it makes server-side code-splitting completely impossible. 

Supporting other routers is a non-goal as far as I'm concerned — the router is a large part of what SvelteKit _is_. So my inclination is to close this, unless there's another use case that we want to try and enable?

benmccann:
I do think there are cases where you might want to use a different router as @pngwn frequently points out. And I don't think that SSR code splitting would necessarily be a blocker since not all adapters support it and it's an optional feature on those that do. That being said, this PR is out-of-date and I'm not tremendously motivated to update it since the Routify folks never did much with it

PR #2708, opened 2021-10-29
[fix] remove unnecessary body

mkorobkov:
Empty string on return body was introduced on this [commit](https://github.com/sveltejs/kit/commit/2a1e9795ad9773e04669a7b9d0234caa0b688ade#diff-70bbf390c5a53b49a9f23557426bbd1bee9188a247ce1ef34e5f4f217036c7f6) I think that happened because ServerResponse.body type did not support nulls.

Because of that, there are problems when using miniflare(v2) which [uses](https://github.com/cloudflare/miniflare/blob/76cfa71fee062125fafe68b24aadee8709977193/packages/core/src/standards/http.ts#L345) nodejs/undici package to create Response instance. Undici [throws](https://github.com/nodejs/undici/blob/d8b3f35b8e46d3eb471e06ddcf1727b6aa01eefa/lib/fetch/response.js#L160) [TypeError](https://fetch.spec.whatwg.org/#ref-for-null-body-status%E2%91%A0) if we pass non-null body with 304(or any other null body status from [fetch specs](https://fetch.spec.whatwg.org/#null-body-status)).

This pull request removes unnecessary body returned on 304 and 204 statuses.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-netlify/files/entry.js
+++ b/packages/adapter-netlify/files/entry.js
@@ -32,9 +32,9 @@ export async function handler(event) {
 	};
 
 	if (rendered.body instanceof Uint8Array) {
-		// Function responses should always be strings, and responses with binary
+		// Function responses should be strings (or undefined), and responses with binary
 		// content should be base64 encoded and set isBase64Encoded to true.
-		// https://github.com/netlify/functions/blob/main/src/function/response.d.ts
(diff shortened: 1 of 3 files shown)


benmccann:
I'm not entirely sure this is a bug in SvelteKit vs Undici, but it doesn't hurt either

PR #2709, opened 2021-10-30
[fix] support etag W/ prefix

mkorobkov:
Remove [W/ prefix](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match#directives) from if-none-match header value before comparing with body hash.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -71,7 +71,7 @@ export async function respond(incoming, options, state = {}) {
 							if (!cache_control || !/(no-store|immutable)/.test(cache_control)) {
 								let if_none_match_value = request.headers['if-none-match'];
 								// ignore W/ prefix https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match#directives
-								if (if_none_match_value?.startsWith('W/')) {
+								if (if_none_match_value?.indexOf('W/"') === 0) {
 									if_none_match_value = if_none_match_value.substring(2);
 								}
 
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -71,7 +71,7 @@ export async function respond(incoming, options, state = {}) {
 							if (!cache_control || !/(no-store|immutable)/.test(cache_control)) {
 								let if_none_match_value = request.headers['if-none-match'];
 								// ignore W/ prefix https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match#directives
-								if (if_none_match_value?.indexOf('W/"') === 0) {
+								if (if_none_match_value?.startsWith('W/"')) {
 									if_none_match_value = if_none_match_value.substring(2);
 								}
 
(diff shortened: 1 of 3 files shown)


Rich-Harris:
Thanks — just to clarify, is this something you've encountered? From the MDN docs it seems like `W/` is redundant, are you saying that there are user agents that add it anyway?

Also, mind adding a changeset to the PR so that it causes a version bump? Thank you

mkorobkov:
> Thanks — just to clarify, is this something you've encountered? From the MDN docs it seems like `W/` is redundant, are you saying that there are user agents that add it anyway?
> 
> Also, mind adding a changeset to the PR so that it causes a version bump? Thank you

Hi, Richard. Sorry for the delay in reply.
As I understood, it depends on server response, not user agent.

I have encountered it while using cloudflare workers. It returns weak etag if request contains "Accept-Encoding: gzip" header(I have not investigated the reason why they do this).

`curl -v "https://postify.online/_app/error.svelte-315d5cf3.js" 2>&1 | grep etag`
`curl -v -H "Accept-Encoding: gzip" "https://postify.online/_app/error.svelte-315d5cf3.js" 2>&1 | grep etag`

Fedeorlandau:
> > Thanks — just to clarify, is this something you've encountered? From the MDN docs it seems like `W/` is redundant, are you saying that there are user agents that add it anyway?
> > Also, mind adding a changeset to the PR so that it causes a version bump? Thank you
> 
> Hi, Richard. Sorry for the delay in reply. As I understood, it depends on server response, not user agent.
> 
> I have encountered it while using cloudflare workers. It returns weak etag if request contains "Accept-Encoding: gzip" header(I have not investigated the reason why they do this).
> 
> `curl -v "https://postify.online/_app/error.svelte-315d5cf3.js" 2>&1 | grep etag` prints: `< etag: "_app/error.svelte-315d5cf3.dcb826ff86.js"`
> 
> `curl -v -H "Accept-Encoding: gzip" "https://postify.online/_app/error.svelte-315d5cf3.js" 2>&1 | grep etag` prints: `< etag: W/"_app/error.svelte-315d5cf3.dcb826ff86.js"`

Hi, @mkorobkov looks like Nginx converts your Etags into weak ones if you ask it to gzip your content. 
https://github.com/nginx/nginx/blob/3334585539168947650a37d74dd32973ab451d70/src/http/modules/ngx_http_gzip_filter_module.c#L291 and https://stackoverflow.com/a/51975550

packages/kit/src/runtime/server/index.js
```diff
@@ -69,9 +69,15 @@ export async function respond(incoming, options, state = {}) {
 						if (response.status === 200) {
 							const cache_control = get_single_valued_header(response.headers, 'cache-control');
 							if (!cache_control || !/(no-store|immutable)/.test(cache_control)) {
+								let if_none_match_value = request.headers['if-none-match'];
+								// ignore W/ prefix https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match#directives
+								if (if_none_match_value?.indexOf('W/"') === 0) {
```
Rich-Harris, line 74:
```suggestion
								if (if_none_match_value?.startsWith('W/"')) {
```

Rich-Harris (approved):

Rich-Harris:
got it. thank you!

Issue #2645, opened 2021-10-20
adapter-static data:image URL error 404


leighwilliams:
### Describe the bug

Hi,

I'm trying to use an emoji favicon as described here: https://css-tricks.com/emojis-as-favicons/

``` html
<link rel="icon" href="data:image/svg+xml,<svg xmlns=%22http://www.w3.org/2000/svg%22 viewBox=%220 0 100 100%22><text y=%22.9em%22 font-size=%2290%22>💥</text></svg>">
```

When building this with adapter-static it fails to handle the data:image/svg URL:

```
Error: 404 /%22data:image/svg+xml,%3Csvg (linked from /)
```

### Reproduction

https://github.com/leighwilliams/svelte-kit-repro

1. npm run build

### Logs

```shell
...
> Using @sveltejs/adapter-static
> 404 /%22data:image/svg+xml,%3Csvg (linked from /)
Error: 404 /%22data:image/svg+xml,%3Csvg (linked from /)
    at file:///project/node_modules/@sveltejs/kit/dist/chunks/index4.js:86:11
    at visit (file:///project/node_modules/@sveltejs/kit/dist/chunks/index4.js:213:5)
    at async visit (file:///project/node_modules/@sveltejs/kit/dist/chunks/index4.js:281:6)
    at async prerender (file:///project/node_modules/@sveltejs/kit/dist/chunks/index4.js:291:6)
    at async Object.prerender (file:///project/node_modules/@sveltejs/kit/dist/chunks/index4.js:353:4)
    at async adapt (file:///project/node_modules/@sveltejs/adapter-static/index.js:13:4)
    at async adapt (file:///project/node_modules/@sveltejs/kit/dist/chunks/index4.js:378:2)
    at async file:///project/node_modules/@sveltejs/kit/dist/cli.js:878:5
npm ERR! code ELIFECYCLE
npm ERR! errno 1
...
```

### System Info

```shell
System:
    OS: macOS 11.4
    CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Memory: 142.95 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.17.0 - ~/.volta/tools/image/node/14.17.0/6.14.13/bin/node
    Yarn: 1.22.0 - ~/.volta/tools/image/yarn/1.22.0/bin/yarn
[...]

PR #2713, opened 2021-10-31
[fix] support data:image URLs in prerendering

gg187on:
fixes: https://github.com/sveltejs/kit/issues/2645

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -261,7 +261,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 				}
 
 				for (const href of hrefs) {
-					if (!href) continue;
+					if (!href || href.startsWith('"data:')) continue;
 
 					const resolved = resolve(path, href);
 					if (!resolved.startsWith('/') || resolved.startsWith('//')) continue;
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -21,6 +21,12 @@ function clean_html(html) {
 		.replace(/<!--[\s\S]*?-->/gm, '');
 }
 
+/** @param {string} html */
+export function get_tag(html) {
+	const match = /<(a|img|link|source)\s+([\s\S]+?)>/.exec(html);
+	return match;
+}
+
 /** @param {string} attrs */
 export function get_href(attrs) {
 	const match = /(?:[\s'"]|^)href\s*=\s*(?:"(.*?)"|'(.*?)'|([^\s>]*))/.exec(attrs);
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -261,7 +261,8 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 				}
 
 				for (const href of hrefs) {
-					if (!href || href.startsWith('"data:')) continue;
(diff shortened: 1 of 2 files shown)


bluwy:
Would nice to add a test case for this as well

benmccann:
Discussion on Discord about whether we might need to do HTML parsing to make this work: https://discord.com/channels/457912077277855764/781266015601295360/904411015083606106

gg187on:
there is long discusion https://discord.com/channels/457912077277855764/781266015601295360/906016984267886682 about using combination of more regexes with JS logic...

I liked idea of using something like @syberag suggested

```js
function extractData(source) {
    let getTag = /<(a|link|img|source)\b/g
    let getAttributes = /(?<name>\w+)\s*(?:=(?:"([^"]*)"|'([^']*)|([^>]+)\b))?/g
    let tags = source.matchAll(getTag)

    return Array.from(tags).map(tag => {
        let start = tag.index
        let subSource = source.substr(start + tag[0].length) // Issue is probably here, because I take the document until the end instead of just the tag
        let attributes = Array.from(subSource.matchAll(getAttributes))
        return {
            tag: tag[1],
            attributes: attributes.map(attr => ({name: attr.groups.name, value: attr[2]||attr[3]||attr[4]}))}
    })
}
```

even tho, this doesn't work correctly, but with this approach I think, we can make performance fast, dependency less and less buggy solution.

Rich-Harris:
My understanding from the Discord thread is that the issue here is attributes containing the `>` character. It feels like the proper solution here is for attribute values to be properly escaped by Svelte, so that we _are_ able to use regexes safely — parsing the HTML would be a hugely expensive change. Does that feel like a good approach?

gg187on:
> My understanding from the Discord thread is that the issue here is attributes containing the `>` character. It feels like the proper solution here is for attribute values to be properly escaped by Svelte, so that we _are_ able to use regexes safely — parsing the HTML would be a hugely expensive change. Does that feel like a good approach?

In input, we get string, that contains HTML code... but in this string how we can detect which `>` character to escape, and which not? Resp. do we escape all of them? If so, how will that regex match, what is tag, what attribute, and where tag ends?

Now imagine You have this crazy input:

```html
<img data-something=" src=" crossorigin src="dog>cat.png"> crazy link </a>
```

It means, that `" src="` is not attribute name, but value. Regex is chomsky type 3 grammar, while HTML is type 2.
We write regex, that will find something between `"` and `"`... what we will get in match:

1. ` src=`
2. ` crossorigin src=`
3. `dog>cat.png`

And that's when there is not another tag before or after.

If we match anything between `<` and `>` we get:

`img data-something=" src=" crossorigin src="dog` -> our issue now.

When we want to read src value, we get:

1. ` crossorigin src=`
2. `dog`

those both values are incorrect, we want `dog>cat.png`

I'm not sure how we would be able to fix it with escaping, when we still needs to find where to escape `>` character and where not. Also there is still a lot of other issues related to how we parse it with regex. For example: https://github.com/sveltejs/kit/issues/2742

where @Conduitry answered https://github.com/sveltejs/kit/issues/2742#issuecomment-962531568

I think, because regex can't cover all cases of HTML language parsing, I'm sure it would be good to rewrite it, and use parse5 parser instead. It would fix a lot of bugs, and even those, we don't know about today.

gg187on:
I see there is good solution with parse5 in PR https://github.com/sveltejs/kit/pull/2923 ... that fix this problem.

PR #2714, opened 2021-10-31
[fix] upgrade minor deps. `fetch-blob` 3.1.3 needed for Netlify deploys

benmccann:
Upgrade to `fetch-blob` 3.1.3 to make sure our deploys to Netlify succeed. This probably won't affect user projects which will need to do their own upgrades, but should help our own deploys of the examples

This is only half of the fix. The other half is for Netlify to bump a dependency on their side. PR pending for that: https://github.com/netlify/cli/pull/3533

Ref https://github.com/sveltejs/kit/issues/1859

### DIFF
--- a/.changeset/flat-eels-hear.md
+++ b/.changeset/flat-eels-hear.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] upgrade minor deps. fetch-blob 3.1.3 needed for Netlify deploys
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -24,21 +24,21 @@ importers:
     devDependencies:
       '@changesets/cli': 2.17.0
       '@changesets/get-github-info': 0.5.0
-      '@rollup/plugin-commonjs': 21.0.0_rollup@2.58.0
-      '@rollup/plugin-json': 4.1.0_rollup@2.58.0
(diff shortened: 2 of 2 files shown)


bluwy (commented):
Do we want to update `package.json` too? That usually happens when running `pnpm update -r`

benmccann:
I don't really think it's necessary or possibly even desired as I think it'd make it slightly harder for users to pick older versions of libraries. It also would cause more merge conflicts with outstanding PRs.

bluwy:
I'm still not clear of your point :sweat_smile: but I think updating to the latest would be better in the long term. Merge conflicts in outstanding PRs would already be affected by the lock file change, so having the `package.json` updated as well could give them a hint of the dependencies that changed when resolving conflicts.

Nevertheless it's just a question I've thought of, I'm fine with the PR as is too.

bluwy (approved):

benmccann:
With sem ver it's just as good to say `^3.0.0` as `^3.2.1`. Or it's actually even better. It lets the user use `3.1.0` if they need to for some reason. Updating `package.json` would just update the minimum version allowed and not the actual version used

bluwy:
Yeah good point too. It was that I find updating to the latest version cleaner, and would imply being the version we officially support (especially if we use a new feature from a minor version). But being able to easily resolve to a lower version is a plus too.

PR #2715, opened 2021-10-31
[docs] updated Netlify adapter documentation to include AWS_LAMBDA_JS_RUNTIME (closes #2687)

### LINKED ISSUE #2687: Netlify adapter: Optional chaining "?." not supported by default in netlify
### Describe the bug

After successfully deploying a very simple Svelte Kit app (one page, one json endpoint) with netlify adapter, my deploys would succeed but when visiting the deployed production website netlify would fail with error:

```
Runtime.UserCodeSyntaxError: SyntaxError: Unexpected token '.'
    at _loadUserApp (/var/runtime/UserFunction.js:98:13) (...)
```

Not being able to debug netlify, it took me quite some time of googling to realize this was caused by the optional chaining operator `?.` being used in my code and not being supported by the default Netlify build (and not being transpiled during build).  

I worked around the issue by adding extra configuration to vite so that the js code was lowered to es6:

```
vite: {
            build: {
              target: [ 'es6' ]
            },
         ...
```

I took the default demo svelte kit app, added `?.`, followed https://github.com/sveltejs/kit/tree/master/packages/adapter-netlify steps and deployed and reproduced the error. If others are experiencing this onboarding to svelte kit and netlify adapter It would be important to change the adapter to handle this by default more gracefully or improve documentation? Am I missing something obvious here?

### Reproduction

Repo reproducing this: https://github.com/jsilveira/bug-sveltekit-optional-chaining-netlify (svelte kit default app modifying `src/routes/todos/index.json.js:8` to use `?.`). Deployed in netlify, visit: https://distracted-wozniak-9548a1.netlify.app/todos to see the netlify error in action.

### System Info

```shell
Binaries:
    Node: 14.15.3
    npm: 6.14.9
  npmPackages:
    @sveltejs/adapter-netlify: 1.0.0-next.33 => 1.0.0-next.33
    @sveltejs/kit: next => 1.0.0-next.188
    svelte: ^3.44 => 3.44.0
```

### Severity

accuser:
Netlify defaults to using Node.js 12 as the runtime for JavaScript functions, but this version is no longer supported by SvelteKit, as noted in #2687. I have added the `AWS_LAMBDA_JS_RUNTIME` setting to the documentation for the Netlify adapter.

### DIFF
--- a/packages/adapter-netlify/README.md
+++ b/packages/adapter-netlify/README.md
@@ -31,9 +31,14 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
 [build]
   command = "npm run build"
   publish = "build"
+
+[build.environment]
+  AWS_LAMBDA_JS_RUNTIME = "nodejs14.x"
 ```
 
 If the `netlify.toml` file or the `build.publish` value is missing, a default value of `"build"` will be used. Note that if you have set the publish directory in the Netlify UI to something else then you will need to set it in `netlify.toml` too, or use the default value of `"build"`.
+
+The `build.environment.AWS_LAMBDA_JS_RUNTIME` value specifies the Node.js runtime for JavaScript functions running on Netlify. Netlify uses Node.js 12 by [default](https://docs.netlify.com/functions/build-with-javascript/?#runtime-settings) which is no longer supported (#2604).
 
 ## Netlify alternatives to SvelteKit functionality
 
--- a/packages/adapter-netlify/README.md
+++ b/packages/adapter-netlify/README.md
@@ -40,8 +40,6 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
 
 If the `netlify.toml` file or the `build.publish` value is missing, a default value of `"build"` will be used. Note that if you have set the publish directory in the Netlify UI to something else then you will need to set it in `netlify.toml` too, or use the default value of `"build"`.
 
-The `build.environment.AWS_LAMBDA_JS_RUNTIME` value specifies the Node.js runtime for JavaScript functions running on Netlify. Netlify uses Node.js 12 by [default](https://docs.netlify.com/functions/build-with-javascript/?#runtime-settings) which is no longer supported (#2604).
-
 ## Netlify alternatives to SvelteKit functionality
 
 You may build your app using functionality provided directly by SvelteKit without relying on any Netlify functionality. Using the SvelteKit versions of these features will allow them to be used in dev mode, tested with integration tests, and to work with other adapters should you ever decide to switch away from Netlify. However, in some scenarios you may find it beneficial to use the Netlify versions of these features. One example would be if you're migrating an app that's already hosted on Netlify to SvelteKit.
--- a/packages/adapter-netlify/README.md
+++ b/packages/adapter-netlify/README.md
@@ -33,6 +33,8 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
   publish = "build"
 
 [build.environment]
+  # SvelteKit requires Node 14+, but Netlify uses Node 12 by default
+  # https://docs.netlify.com/functions/build-with-javascript/?#runtime-settings
   AWS_LAMBDA_JS_RUNTIME = "nodejs14.x"
 ```
 


bluwy (approved):

packages/adapter-netlify/README.md
```diff
@@ -31,10 +31,15 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
...
   command = "npm run build"
   publish = "build"
+
+[build.environment]
+  AWS_LAMBDA_JS_RUNTIME = "nodejs14.x"
 ```
 
 If the `netlify.toml` file or the `build.publish` value is missing, a default value of `"build"` will be used. Note that if you have set the publish directory in the Netlify UI to something else then you will need to set it in `netlify.toml` too, or use the default value of `"build"`.
 
+The `build.environment.AWS_LAMBDA_JS_RUNTIME` value specifies the Node.js runtime for JavaScript functions running on Netlify. Netlify uses Node.js 12 by [default](https://docs.netlify.com/functions/build-with-javascript/?#runtime-settings) which is no longer supported (#2604).
```
benmccann, line 41:
```suggestion
The `build.environment.AWS_LAMBDA_JS_RUNTIME` value specifies the Node.js runtime for JavaScript functions running on Netlify. [Netlify uses Node.js 12 by default](https://docs.netlify.com/functions/build-with-javascript/?#runtime-settings), which is no longer supported by SvelteKit (#2604).
```

benmccann:
I wonder if rather than updating the docs, we can use Netlify's framework detector to specify Node 14 if not otherwise specified?

https://github.com/netlify/framework-info/blob/main/src/frameworks/svelte-kit.json

benmccann:
We could also set the `NODE_VERSION` like Nuxt does to make it so that you don't have to set that in an `.nvmrc`, etc.

https://github.com/netlify/framework-info/blob/39f9e3fff4370a23f1d62f76cc07bfda3b20e59b/src/frameworks/nuxt3.json#L19
https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript

accuser:
> PR for `framework-info` here: [netlify/framework-info#553](https://github.com/netlify/framework-info/pull/553)
> 
> Let's leave this one open for the time being until we get a response on that one and test it out

This makes sense until the PR for the Netlify framework detector is accepted.

ascorbic:
Unfortunately this won't work right now, but we're working on it, and nor will the framework detection code. I made a mistake when I added the Nuxt env vars so you can't use that approach. They're actually only used when running `netlify dev`, which I misunderstood. The good news is that we will very shortly be switching the default Lambda runtime to 14, and this issue has moved that up the priority list. 

The current correct solution is to set `AWS_LAMBDA_JS_RUNTIME` in the site UI, not in the toml.

packages/adapter-netlify/README.md
```diff
@@ -31,10 +31,15 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
 [build]
   command = "npm run build"
   publish = "build"
+
+[build.environment]
+  AWS_LAMBDA_JS_RUNTIME = "nodejs14.x"
```
benmccann, line 36: I think it might be nicer to co-locate the documentation with the setting
```suggestion
  # SvelteKit requires Node 14+, but Netlify uses Node 12 by default
  # https://docs.netlify.com/functions/build-with-javascript/?#runtime-settings
  AWS_LAMBDA_JS_RUNTIME = "nodejs14.x"
```

packages/adapter-netlify/README.md
```diff
@@ -31,10 +31,15 @@ Then, make sure you have a [netlify.toml](https://docs.netlify.com/configure-bui
...
 ```
 
 If the `netlify.toml` file or the `build.publish` value is missing, a default value of `"build"` will be used. Note that if you have set the publish directory in the Netlify UI to something else then you will need to set it in `netlify.toml` too, or use the default value of `"build"`.
 
+The `build.environment.AWS_LAMBDA_JS_RUNTIME` value specifies the Node.js runtime for JavaScript functions running on Netlify. Netlify uses Node.js 12 by [default](https://docs.netlify.com/functions/build-with-javascript/?#runtime-settings) which is no longer supported (#2604).
+
```
benmccann, line 44:
```suggestion
```

benmccann:
Oops, I missed the comment that this won't work: https://github.com/sveltejs/kit/pull/2715#issuecomment-957681594

Closing as a result of that comment

Issue #2105, opened 2021-08-05
Prerender crawls URLs external to app on same domain


benmccann:
### Describe the bug

I haven't tested this, but think I just noticed a bug in the code

We should use consistent logic for determining if a URL is external

This seems right:
https://github.com/sveltejs/kit/blob/be30a938c42619d0af61ba98d540c103ffade216/packages/kit/src/runtime/server/page/load_node.js#L113

This does not:
https://github.com/sveltejs/kit/blob/be30a938c42619d0af61ba98d540c103ffade216/packages/kit/src/core/adapt/prerender.js#L249

### Reproduction

todo

### System Info

```shell
`master`
```

### Severity

serious, but I can work around it

si3nloong:
@benmccann I can work on this

benmccann:
Sure!

Btw, no need to comment about taking an issue. If it's been open for awhile with no PRs, it's quite unlikely anyone is working on it. Thanks for the help!

si3nloong:
@benmccann I think this issue can be close since Rich already solved it https://github.com/sveltejs/kit/commit/f12bbcc686829243f111db1b463d4a261fb67c23

Issue #1567, opened 2021-05-27
prerendering does not crawl imported images outside of pages


PingTouG:
**Describe the bug**
`routes/todos/index.json.ts` imports an image. When using `@sveltejs/adapter-static`, `npm run build` produces a 404 error

**Logs**
> 404 /_app/assets/industry_1.166d527b.png (linked from /todos)

**To Reproduce**
Checkout https://github.com/PingTouG/svelte-kit-bug. An image was added to the todos tab as shown below

`src/routes/todos/index.json.ts`
```ts
import welcomeImg from '$lib/images/home/industry_1.png';
....
export const get: RequestHandler<Locals> = async (request) => {
....
return { body: { welcomeImg  } }
```
`src/routes/todos/index.svelte`
```svelte
....
<img src={welcomeImg} />
.....
```

**Expected behavior**
`npm run build` is ok

**Information about your SvelteKit Installation:**

<details>
  <summary>Diagnostics</summary>

- @sveltejs/kit : next
- @sveltejs/adapter-static : ^1.0.0-next.11
- svelte : ^3.34.0

- chorme
</details>

**Severity**
The website cannot run

ignatiusmb:
This only happens for prerendered pages. Images imported images outside of pages, in this case an endpoint, does not get crawled. The error would also appear if we use adapter-node and set `export const prerender = true;`

bluwy:
This is because assets are only generated for the client side builds (which doesn't bundle endpoints). SSR builds (which bundles endpoints) don't generate assets. To workaround this, you can specify all the possible links from SSR in the client bundle [like this](https://github.com/bluwy/website/blob/7fec5d757692b26c32cd82335bc9f4e84d70aba2/src/data/assets.js), which shouldn't impact bundle size, but not the most ergonomic way of handling it.

I'm not sure how we should handle this though, perhaps SvelteKit can have Vite crawl for asset imports in endpoints. I  also think I've seen a similar issue in Vite but I can't find it.

m1212e:
Converting images to a blob worked for me as a workaround. The +server.ts returns this:
```
const result = fetch(...)
return new Response(await result.blob());
```

Rich-Harris:
closing as a dupe of #5240 (even though this issue was first) — the underlying issue is https://github.com/vitejs/vite/issues/11429, and there's a fix here https://github.com/vitejs/vite/pull/11430

PR #2725, opened 2021-11-02
[chore] refactor `is_rel_external` regex

si3nloong:
- ✅ add test cases for `clean_html`, `get_src`, `is_rel_external` and `get_srcset_urls`
- ✅ export function `clean_html`, `get_src` and `get_srcset_urls` for testing purpose
- 🐛 fix `get_srcset_urls` bug
- 🐛 fix `is_rel_external` bug when value is not an URI


si3nloong:
@benmccann can you review and merge it

benmccann:
There's a related PR where it looks like we might have to do HTML parsing to fix the bug: https://github.com/sveltejs/kit/pull/2713#issuecomment-959088636. I'd like to get Rich to weigh in before we move forward with either

benmccann:
Would you be able to file an issue for the bugs you found so that we don't lose track of them? There's not enough detail from the PR description to really understand what they are

si3nloong:
Hi @benmccann, I actually refactoring the pre-render function using `parse5` (https://github.com/si3nloong/kit/tree/prerender), it will cover the bug fixes I mentioned earlier as well as #2742 , #2105 and #1567 

It still WIP, will definitely update here once I done

PR #2729, opened 2021-11-02
[fix] make HN example RSS feed work

benmccann:
no changeset since this is just an example


Conduitry:
Out of curiosity, what did this fix? I'm having a hard time seeing what changed beyond rewriting this to use `await`.

benmccann:
I don't fully know :man_shrugging:  But it seems to be cleaner code in any case in addition to working better

ignatiusmb:
It's probably because the previous one isn't `return`-ing the `fetch`, causing the function to return `void` instead of the resolved `.then`

Conduitry:
Oh whoops, duh, yeah I missed that.

Issue #2664, opened 2021-10-22
Anchor links broken when navigating across pages


mikenikles:
### Describe the bug

I upgraded to next.188 yesterday and notice anchor links no longer work across pages.

**Steps to reproduce**
1. Navigate to https://deploy-preview-1131--gitpod-kumquat.netlify.app/blog
2. Click "Sign up" at the top right

**Actual result**
The homepage loads, scrolled to the top

**Expected result**
The homepage loads, scrolled down to the Get Started Now section.

Note, if you click "Sign Up" again after step 2 (i.e. when you're on the homepage), the anchor link works.

Has anyone else seen this? I've been looking through Discord and GitHub, not much luck so far but will continue to search.

### Reproduction

1. `npm init svelte@next` & follow the instructions
2. In `src/routes/index.html` add a bunch of `<br />` elements to make the page scrollable, then add `<h1 id="jump-here">Here is a page anchor</h1>` at the bottom of the page
3. In `src/routes/about.html` add a link: `<a href="/#jump-here">Jump to section on homepage</a>`

### System Info

```shell
System:
    OS: macOS 11.2.3
    CPU: (8) x64 Apple M1
    Memory: 136.04 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.17.4 - /usr/local/bin/node
    npm: 7.7.6 - /opt/homebrew/bin/npm
  Browsers:
    Brave Browser: 94.1.30.87
    Chrome: 95.0.4638.54
    Edge: 93.0.961.47
    Firefox: 92.0
    Firefox Developer Edition: 88.0
    Safari: 14.0.3
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.188 
    svelte: ^3.34.0 => 3.44.0
```

### Severity

blocking an upgrade

mquandalle:
I have updated my app to use the latest version of kit (`next.193`), and I still experience this issue.

mikenikles:
With `next.193`, here's what I experience...

Setup
* `a.svelte` contains my anchor
* `b.svelte` links to `a.svelte`

❌ Does not work
1. Navigate to `/b`
2. Click on link to `/a`
3. `/a` is scrolled to the top, not the anchor

✅ Does work
1. Navigate to `/b`
2. Reload the page **twice** (yep, 2 times)
3. Click on link to `/a`
4. `/a` scrolls to the anchor as expected

[...]

PR #2735, opened 2021-11-03
[fix] reset scroll when navigated from scrolled page

### LINKED ISSUE #2733: Page scroll position not reset to top on navigation (regression)
### Describe the bug

When navigating to a new page, if you are already scrolled partway down the page, the next page load will keep the current scroll position and not scroll back to the top.

I believe this broke with next-193. When I install next-192, this does not happen -- navigating to a new page always scrolls to the top, just like regular browser navigation.

### Reproduction

Init the SvelteKit demo project. Add the following paragraph in `about.svelte` after the last paragraph.

```html
<p>Go <a href="/">home</a></p>
```

Make the browser window small enough so that the page scrolls and scroll to the bottom of the page. Click the Home link. The Home page loads, but does not scroll to the top (i.e. the nav bar is not visible).

scroll-bug.

Tested in Chrome and Firefox on Windows 10.

### Logs

_No response_

### System Info

```shell
System:
    OS: Linux 4.19 Ubuntu 20.04.1 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz
    Memory: 7.79 GB / 12.40 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Yarn: 1.22.5 - /usr/bin/yarn
    npm: 8.1.1 - ~/.nvm/versions/node/v14.17.0/bin/npm
  Browsers:
    Chrome: 89.0.4389.90
  npmPackages:
    @sveltejs/kit: ^1.0.0-next.193 => 1.0.0-next.193
    svelte: ^3.34.0 => 3.44.1
```

### Severity

### Additional Information

_No response_

bluwy:
Fixes #2733 and maybe #2664

Move the scroll to top logic before the new route mounts so that we can properly compare the `pageYOffset === 0`. This works because scroll position is preserved between route changes by default.

I also updated our `isVisble()` tests to `is_intersecting_viewport()`, since `isVisble` checks for visibility anywhere in the document. `is_intersecting_viewport` makes sure they are in the viewport.

Nonetheless, @mquandalle and @mikenikles I'd appreciate if y'all can try to see if this branch fixes your issue as well.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -275,33 +275,34 @@ export class Renderer {
 			this._init(navigation_result);
 		}
 
-		if (opts) {
-			if (!opts.keepfocus) {
+		if (!opts) {
+			await 0;
+		} else {
+			const { hash, scroll, keepfocus } = opts;
+
+			if (!keepfocus) {
 				document.body.focus();
 			}
 
-			// Scroll to top so we can compare the `pageYOffset` below. We cannot
-			// compare by recording the `pageYOffset` here as there is a possibility
-			// it will change, e.g. different page heights
-			scrollTo(0, 0);
-		}
+			const oldPageYOffset = pageYOffset;
+			await 0;
+			const maxPageYOffset = document.body.scrollHeight - innerHeight;
 
-		await 0;
-
-		// After `await 0`, the `onMount()` function in the component executed.
-		// If there was no scrolling happening (checked via `pageYOffset`),
-		// continue on our custom scroll handling
-		if (pageYOffset === 0 && opts) {
-			const { hash, scroll } = opts;
-
-			const deep_linked = hash && document.getElementById(hash.slice(1));
-			if (scroll) {
-				scrollTo(scroll.x, scroll.y);
-			} else if (deep_linked) {
-				// Here we use `scrollIntoView` on the element instead of `scrollTo`
-				// because it natively supports the `scroll-margin` and `scroll-behavior`
-				// CSS properties.
-				deep_linked.scrollIntoView();
+			// After `await 0`, the `onMount()` function in the component executed.
+			// If there was no scrolling happening (checked via `pageYOffset`),
+			// continue on our custom scroll handling
+			if (pageYOffset === Math.min(oldPageYOffset, maxPageYOffset)) {
+				const deep_linked = hash && document.getElementById(hash.slice(1));
+				if (scroll) {
+					scrollTo(scroll.x, scroll.y);
+				} else if (deep_linked) {
+					// Here we use `scrollIntoView` on the element instead of `scrollTo`
+					// because it natively supports the `scroll-margin` and `scroll-behavior`
+					// CSS properties.
+					deep_linked.scrollIntoView();
+				} else {
+					scrollTo(0, 0);
+				}
 			}
 		}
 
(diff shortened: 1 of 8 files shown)


packages/kit/src/runtime/client/renderer.js
```diff
@@ -275,8 +275,15 @@ export class Renderer {
...
-		if (!opts?.keepfocus) {
-			document.body.focus();
+		if (opts) {
+			if (!opts.keepfocus) {
+				document.body.focus();
+			}
+
+			// Scroll to top so we can compare the pageYOffset below. We cannot
+			// compare by recording the pageYOffset here as there is a possiblity
+			// it will change, e.g. different scroll heights
```
benmccann, line 285: what do you mean by different scroll heights? do you mean different page heights?

could you use the minimum of `pageYOffset` and the page height?

bluwy: Yeah page heights sounds clearer :+1: 

> could you use the minimum of pageYOffset and the page height?

I'm not too sure what you mean here, are you talking about the words to use for documentation?

bluwy:
> Could this result in a behavior of jumping to the top of the page and the down to the final scroll position?

This would technically happen from the code flow, but it happens between an `await 0` so it should be fast enough that it's not noticeable, but otherwise I don't see another way around this unfortunately, unless we can hook into an event loop where the component is mounted but not yet called `onMount()`.

packages/kit/src/runtime/client/renderer.js
bluwy: Fixed in 47ea554

packages/kit/src/runtime/client/renderer.js
```diff
@@ -275,14 +275,21 @@ export class Renderer {
...
 
-		if (!opts?.keepfocus) {
-			document.body.focus();
+		if (opts) {
+			if (!opts.keepfocus) {
+				document.body.focus();
+			}
+
+			// Scroll to top so we can compare the `pageYOffset` below. We cannot
+			// compare by recording the `pageYOffset` here as there is a possiblity
```
benmccann, line 284:
```suggestion
			// compare by recording the `pageYOffset` here as there is a possibility
```

packages/kit/src/runtime/client/renderer.js
```diff
@@ -275,14 +275,21 @@ export class Renderer {
...
 		}
 
-		if (!opts?.keepfocus) {
-			document.body.focus();
+		if (opts) {
+			if (!opts.keepfocus) {
+				document.body.focus();
+			}
+
+			// Scroll to top so we can compare the `pageYOffset` below. We cannot
```
benmccann, line 283: > We cannot compare by recording the `pageYOffset` here as there is a possiblity it will change, e.g. different page heights

I don't know if this is quite true? What I meant in my last comment about using the min of `pageYOffset` and the page height is that I wonder if you could check something like `if (pageYOffset === Math.min(oldPageYOffset, newPageHeight))` to see if the user has not scrolled

bluwy: Ah great idea. I think that might work too, I'll try that when I get the time.

bluwy: That seems to work! The code looks a lot cleaner now. Pushed the changes.

benmccann (approved):
thanks for updating these tests!

Issue #365, opened 2021-02-03
SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds


Rich-Harris:
We've been seeing this in production, via Sentry — almost exclusively on iPhones. I assume it's related to this: https://github.com/sveltejs/kit/blob/ad9551c47727b74fcc7c31614962e7bc37792633/packages/kit/src/runtime/internal/router/index.js#L49-L63

zackify:
What if you changed it to only run every 500ms? Then it would only run a max of 60 times in 30 seconds. I feel like that is still pretty frequent for restoring the scroll position later.

benmccann:
It should not be running when continuously moving the mouse or when not moving the mouse, but only when the mouse was moving and then stops. Changing it to 500s sounds like a copout instead of really understanding the issue as to why debouncing isn't working here.

zackify:
I see what you mean!

pago:
Not entirely sure if it should go here or in a separate issue. Happy to move it to a separate issue if needed.

I'm observing the same issue mentioned here but I also have some (iOS only) errors about `pushState` in my Sentry log.

```
SecurityError: Attempt to use history.pushState() more than 100 times per 30 seconds
```

Might be related somehow.

philippdormann:
seeing these issues in production on iOS 14.7.1 devices (iPhone 12 mini - "iPhone13,1")
```
SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds
```

```
SecurityError: Attempt to use history.pushState() more than 100 times per 30 seconds
```

acoyfellow:
Can confirm I'm picking this up via Sentry too:
```
SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds
```

Mobile Safari
Version:14.1.2

iOS
Version:14.7.1

iPhone

si3nloong:
May be we can refer to this solution.

https://github.com/ampproject/amphtml/pull/23938/files

MyIsaak:
Check out my PR on this issue: https://github.com/sveltejs/kit/pull/2738

bluwy:
I tested this on my iphone today and I discovered that ios emits the scroll event later than other browsers. Intervals between scroll events rapidly fluctuate between 30 - 150ms. And an easy way to aggressively trigger this bug is to quickly scroll up and down, that would consistently give 200-300ms.

[...]

PR #2738, opened 2021-11-04
[fix] persist scroll position without scroll event (#365)

MyIsaak:
## Solution

A faster approach to persisting scroll position without using 'scroll' events. When a `popstate` event is fired, which happens **after** a user navigates backwards (or forwards), we navigate in reverse to store the scroll position in the history's state of the previous page, and subsequently navigate in reverse again to arrive at the intended page.

This should also fix Sentry API abuse complaints regarding history.replaceState described in #365.

## Problem

A debounced 'scroll' event is used to store scroll events, which harms scrolling performance. Unfortunately browsers don't provide an event like 'popstate' that fires **before** a user navigates backwards (or forwards), however my trick should do it for the browsers I've tested it on.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -282,9 +282,7 @@ export class Renderer {
 		await 0;
 
 		// After `await 0`, the onMount() function in the component executed.
-		// If there was no scrolling happening (checked via pageYOffset),
-		// continue on our custom scroll handling
-		if (pageYOffset === 0 && opts) {
+		if (opts) {
 			const { hash, scroll } = opts;
 
 			const deep_linked = hash && document.getElementById(hash.slice(1));
--- a/.changeset/shaggy-ants-smell.md
+++ b/.changeset/shaggy-ants-smell.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Persist scroll position without scroll event
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -41,6 +41,8 @@ export class Router {
 		this.trailing_slash = trailing_slash;
 		/** Keeps tracks of multiple navigations caused by redirects during rendering */
 		this.navigating = 0;
+		/** A stack of routes which gets larger when navigating back in history, and smaller when navigating forward.*/
+		this.history = sessionStorage.getItem('sveltekit:history')?.split(',') || [];
 
 		/** @type {import('./renderer').Renderer} */
 		this.renderer = renderer;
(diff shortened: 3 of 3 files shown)


packages/kit/src/runtime/client/renderer.js
```diff
@@ -282,9 +282,7 @@ export class Renderer {
 		await 0;
 
 		// After `await 0`, the onMount() function in the component executed.
-		// If there was no scrolling happening (checked via pageYOffset),
-		// continue on our custom scroll handling
-		if (pageYOffset === 0 && opts) {
```
MyIsaak, line 287: Please could someone enlighten me on this. I'm not sure about the intent of limiting scroll function to when we're at the top of the page. Why not persit scroll in any scroll position?

bluwy: `pageYOffset === 0` is used to check if the dev did manual scrolling in an `onMount()` function, if they did we shouldn't apply our scrolling logic. It's not the best heuristic but I haven't found a better way around it yet.

bluwy (commented):
I'm a bit wary of this approach. Is this technique being documented or used out in the wild?

packages/kit/src/runtime/client/renderer.js
bluwy: FYI there's also another fix that touches that in #2735

MyIsaak:
Navigating in reverse to modify history's state of previous page: In practice this solution is unique because we're still able to store scroll position inside the browser's history state. In principle, when navigating forwards and back the page the 'popstate' event is called regardless of wether the new page has loaded [(See steps below taken from mdn)](https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event). 

```
1. If new-entry doesn't currently contain an existing Document, fetch the content and create its Document before continuing. This will eventually send events such as DOMContentLoaded and load to the Window containing the document, but the steps below will continue to execute in the meantime.
2-11. history.state, title, readyState are updated...
12. If the value of state changed, the popstate event is sent to the document.
```

I'd say the only risk is that a document loads before `popstate` the event, however this is unlikely because the document loads in async/parallel to the popstate event, where navigating back should cancel our document load and complex document loads take longer than state changes.

Tracking a history of links to differentiate between back and forward in "popstate" event: https://github.com/angular/angular/commit/3c6b653089837459809a370ebcaf8911c3bab9ed

benmccann:
I haven't had a chance to take a look at this PR yet, but https://github.com/sveltejs/kit/pull/2735 was merged so this one would need to be rebased

bluwy:
@MyIsaak I've sent #2749 that would fix #365 in a simpler way. I think that should work for the time being. Perhaps that would leave this PR to be an alternative of keeping the scroll state. I haven't fully test this yet though, but it seems like there's a failing test.

benmccann:
The double navigation thing here isn't super clear to me, so I merged #2749 as the fix for #365. Let's see how that plays out and if it addresses the issue. Thanks for taking a look at it!

MyIsaak:
No problem. Performance isn't always most important, compared to if it actually works.

Issue #2712, opened 2021-10-30
TypeScript load function example


benblazak:
### Describe the problem

It took me a long time to figure out how to write a TypeScript `load` function for `__error.svelte`

### Describe the proposed solution

Include an example :)

``` typescript
import type { ErrorLoadInput, LoadOutput } from "@sveltejs/kit";
export async function load({
  error,
  status,
}: ErrorLoadInput): Promise<LoadOutput> {
  return {
    props: {
      title: `${status}: ${error.message}`,
      stack: error.stack,
    },
  };
}
```

### Alternatives considered

Don't include an example :(

### Importance

nice to have

JeanJPNM:
You can import the `Load` type:

```ts
import type { Load } from "@sveltejs/kit"

export const load: Load = ({ error, status }) => ({
  props: {
    title: `${status}: ${error.message}`,
    stack: error.stack,
  }
})
```

But I agree that there should be an example for this.

benblazak:
that's much nicer -- thanks :)

tried it just now. using `Load` compiled and ran fine, but failed svelte-check.  had to use `ErrorLoad`.

``` typescript
import type { ErrorLoad } from "@sveltejs/kit";
export const load: ErrorLoad = ({ error, status }) => ({
  props: {
    title: `${status}: ${error.message}`,
    stack: error.stack,
  },
});
```

@benmccann do you guys like pull requests for things this small?

if so, is there support for code tabs in the sveltekit docs, or should i just write it after the javascript version in https://kit.svelte.dev/docs#layouts-error-pages?

bluwy:
> if so, is there support for code tabs in the sveltekit docs, or should i just write it after the javascript version in https://kit.svelte.dev/docs#layouts-error-pages?

Feel free to open a PR for the javascript version on that section! Seems to be lacking the jsdoc typings in the code block example. I'm not sure if we'd want a typescript version soon (not my call), but a javascript version is a good start.

benblazak:
So just adding this on top?

``` javascript
/**
 * @type {import('@sveltejs/kit').ErrorLoad}
 */
```

I can add that if it's helpful :)

In that case, should we also add `ErrorLoadInput` to the list of types in the "Loading" section just below?

``` typescript
export interface ErrorLoadInput<
	PageParams extends Rec<string> = Rec<string>,
	Stuff extends Rec = Rec,
	Session = any
> extends LoadInput<PageParams, Stuff, Session> {
	status?: number;
	error?: Error;
[...]

PR #2743, opened 2021-11-04
[docs] add types for error load function

benblazak:
related to https://github.com/sveltejs/kit/issues/2712

Pure doc change.

This makes the error `load` description consistent with the description for `load`.

Seems that we're not cluttering the docs with TypeScript for now, since we don't have code tabs to make it neat.  JSDoc types are helpful though.

### Tests

### Changesets

### DIFF
--- a/documentation/docs/02-layouts.md
+++ b/documentation/docs/02-layouts.md
@@ -76,6 +76,20 @@ For example, if `src/routes/settings/notifications/index.svelte` failed to load,
 
 > SvelteKit provides a default error page in case you don't supply `src/routes/__error.svelte`, but it's recommended that you bring your own.
 
+```ts
+// declaration type
+// * also see type for `LoadOutput` in the Loading section
+
+export interface ErrorLoadInput<
+	PageParams extends Rec<string> = Rec<string>,
+	Stuff extends Rec = Rec,
+	Session = any
+> extends LoadInput<PageParams, Stuff, Session> {
+	status?: number;
+	error?: Error;
+}
+```
+
 If an error component has a [`load`](#loading) function, it will be called with `error` and `status` properties:
 
 ```html
@@ -98,17 +112,3 @@ If an error component has a [`load`](#loading) function, it will be called with 
 ```
 
 > Server-side stack traces will be removed from `error` in production, to avoid exposing privileged information to users.
-
-```ts
-// declaration type
-// * also see type for `LoadOutput` in the Loading section
-
-export interface ErrorLoadInput<
-	PageParams extends Rec<string> = Rec<string>,
-	Stuff extends Rec = Rec,
-	Session = any
-> extends LoadInput<PageParams, Stuff, Session> {
-	status?: number;
-	error?: Error;
-}
-```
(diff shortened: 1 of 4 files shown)


documentation/docs/02-layouts.md
```diff
@@ -80,11 +80,14 @@ If an error component has a [`load`](#loading) function, it will be called with
...
 <script context="module">
+	/**
+	 * @type {import('@sveltejs/kit').ErrorLoad}
+	 */
 	export function load({ error, status }) {
 		return {
 			props: {
-				title: `${status}: ${error.message}`
-			}
+				title: `${status}: ${error.message}`,
```
benmccann, line 89: I'm not sure I like the addition of the trailing commas on this line and the next

documentation/docs/02-layouts.md
```diff
@@ -80,11 +80,14 @@ If an error component has a [`load`](#loading) function, it will be called with
 
 ```html
 <script context="module">
+	/**
+	 * @type {import('@sveltejs/kit').ErrorLoad}
```
benmccann, line 84: It's a bit of a mix, but usually for 1-line javadocs we write them as:
```
/** @type {import('@sveltejs/kit').ErrorLoad} */
```

We should probably change the others that aren't condensed to 1-line for consistency

documentation/docs/02-layouts.md
benblazak: removed them

thought i was using the default sveltekit prettier config, but now that i think of it i probably changed that setting

documentation/docs/02-layouts.md
benblazak: changed this one

documentation/docs/02-layouts.md
```diff
@@ -97,3 +98,17 @@ If an error component has a [`load`](#loading) function, it will be called with
 ```
 
 > Server-side stack traces will be removed from `error` in production, to avoid exposing privileged information to users.
+
+```ts
```
benmccann, line 102: everywhere else in the docs, I think we put the types first and then the example code

benblazak: ahh, you're right -- i like it much better the way i wrote it, but i'll change it for consistency

documentation/docs/02-layouts.md
```diff
@@ -76,10 +76,25 @@ For example, if `src/routes/settings/notifications/index.svelte` failed to load,
 
 > SvelteKit provides a default error page in case you don't supply `src/routes/__error.svelte`, but it's recommended that you bring your own.
 
+```ts
+// declaration type
+// * also see type for `LoadOutput` in the Loading section
+
+export interface ErrorLoadInput<
+	PageParams extends Rec<string> = Rec<string>,
+	Stuff extends Rec = Rec,
```
bluwy, line 85: I think we should expand `Rec` similarly to the `LoadInput` example in [loading](https://kit.svelte.dev/docs#loading) as it's an internal utility type.

```suggestion
	PageParams extends Record<string, string> = Record<string, string>,
	Stuff extends Record<string, any> = Record<string, any>,
```

Rich-Harris (approved):

Rich-Harris:
thank you!

PR #2749, opened 2021-11-05
[fix] increase scroll debounce timeout

### LINKED ISSUE #365: SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds

bluwy:
Fixes #365 

iOS scroll event intervals happens between 30-150ms, sometimes around 200ms. So we set the debounce timeout to 200ms to prevent most false positives.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -88,7 +88,7 @@ export class Router {
 					'sveltekit:scroll': scroll_state()
 				};
 				history.replaceState(new_state, document.title, window.location.href);
-				// iOS scroll event intervals happens between 30-150ms, sometimes around 200ms
+				// iOS scroll event intervals happen between 30-150ms, sometimes around 200ms
 			}, 200);
 		});
 
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -88,6 +88,7 @@ export class Router {
 					'sveltekit:scroll': scroll_state()
 				};
 				history.replaceState(new_state, document.title, window.location.href);
+			// iOS scroll event intervals happens between 30-150ms, sometimes around 200ms
 			}, 200);
 		});
 
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -88,7 +88,7 @@ export class Router {
 					'sveltekit:scroll': scroll_state()
 				};
 				history.replaceState(new_state, document.title, window.location.href);
-			// iOS scroll event intervals happens between 30-150ms, sometimes around 200ms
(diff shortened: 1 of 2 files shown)


packages/kit/src/runtime/client/router.js
```diff
@@ -88,7 +88,7 @@ export class Router {
 					'sveltekit:scroll': scroll_state()
 				};
 				history.replaceState(new_state, document.title, window.location.href);
-			}, 50);
+			}, 200);
```
benmccann, line 91: should we add a comment here like:
```suggestion
			// iOS scroll event intervals happens between 30-150ms, sometimes around 200ms
			}, 200);
```

packages/kit/src/runtime/client/router.js
```diff
@@ -88,7 +88,8 @@ export class Router {
 					'sveltekit:scroll': scroll_state()
 				};
 				history.replaceState(new_state, document.title, window.location.href);
-			}, 50);
+			// iOS scroll event intervals happens between 30-150ms, sometimes around 200ms
```
benmccann, line 91: I just noticed an extra s in the text I copied from the PR description :smile: 
```suggestion
			// iOS scroll event intervals happen between 30-150ms, sometimes around 200ms
```

bluwy: My grammar :see_no_evil: Fixed in the latest commit

benmccann:
Can you explain this bug in a bit more detail? I'm not sure I understand yet exactly how it occurred

Is my understanding below correct?
- On a normal device, the scroll event is firing constantly as you scroll. Each time a new event comes in, it cancels the previous timer, and after events stop coming in for 50ms it save the state
- On iOS, scroll events are also fired constantly as you scroll, but sometimes with large gaps between them. E.g. 100ms. In that case, you will be saving the state every 100ms. Which could mean 10 times per second. Or 100 times in 10 seconds. So you would have to scroll the page non-stop for 5-10 seconds to make the bug trigger? I'm kind of surprised people would be doing that very often and so I wonder if I'm missing something

bluwy:
Yep, that's how it is with iOS currently.

> So you would have to scroll the page non-stop for 5-10 seconds to make the bug trigger?

I won't say non-stop since it can happen intermittently, so as long as the user scrolls for 5-10s in a 30s window, it'll cause the error. And it happens really often as long as your thumb stays on screen, even the slightest scroll.

benmccann (approved):

benmccann:
Ah, thanks for the explanation!!

Issue #1718, opened 2021-06-20
Build fails when passing manualchunks to rollup


paul-schwendenman:
**Describe the bug**
When passing manual chunks to rollup through vite, the build falls.

**Logs**
n/a see stack trace

**To Reproduce**

To recreate, I added: 

```
		vite: {
			build: {
				rollupOptions: {
					output: {
						manualChunks: {
							lodash: ['lodash']
						}
					}
				}
			}
		}
```
to `svelte.config.js` in the hacker news example and I get the following error. 

```
The "output.manualChunks" option is not supported for "output.inlineDynamicImports".
> The "output.manualChunks" option is not supported for "output.inlineDynamicImports".
Error: The "output.manualChunks" option is not supported for "output.inlineDynamicImports".
```

Note, I get the same error with different `manualChunk` configs. I also got an error in my project after installing the local build. I tried disabling `inlineDynamicImports` but that config is protected by sveltekit.

**Expected behavior**
The build should succeed. I suppose there should be a way to specify "`output.inlineDynamicImports`." If sveltekit is manually edited to disable `inlineDynamicImports` the build will pass:

> I hacked my @sveltejs/kit locally to comment out // inlineDynamicImports: true and everything works as expected with a lodash chunk on the Sveltekit demo app

from [comment](https://github.com/sveltejs/kit/issues/1571#issuecomment-864470983) by @johnnysprinkles 

**Stacktraces**

<details>
  <summary>Stack trace</summary>

```
The "output.manualChunks" option is not supported for "output.inlineDynamicImports".
> The "output.manualChunks" option is not supported for "output.inlineDynamicImports".
Error: The "output.manualChunks" option is not supported for "output.inlineDynamicImports".
    at error (/home/paul/external/kit/node_modules/.pnpm/vite@2.3.7/node_modules/vite/node_modules/rollup/dist/shared/rollup.js:164:30)
    at getManualChunks (/home/paul/external/kit/node_modules/.pnpm/vite@2.3.7/node_modules/vite/node_modules/rollup/dist/shared/rollup.js:20695:20)
    at normalizeOutputOptions (/home/paul/external/kit/node_modules/.pnpm/vite@2.3.7/node_modules/vite/node_modules/rollup/dist/shared/rollup.js:20466:23)
    at getOutputOptions (/home/paul/external/kit/node_modules/.pnpm/vite@2.3.7/node_modules/vite/node_modules/rollup/dist/shared/rollup.js:20817:12)
[...]

PR #2753, opened 2021-11-06
[fix] do not set inlineDynamicImports

benmccann:
This would enable both https://github.com/sveltejs/kit/pull/2529 and https://github.com/sveltejs/kit/pull/2580

Rich said he set it because he thought Netlify may need it, but I deployed an app to Netlify without the setting and it worked just fine: https://eloquent-mestorf-107ac7.netlify.app/top/1. So we could possibly remove it since it doesn't really seem necessary

This is a setting people have complained about in the past (https://github.com/sveltejs/kit/issues/1571 and https://github.com/sveltejs/kit/issues/1718), so we recently changed it to allow it to be overriden. That means people have been setting it to false in production already, which should give us confidence that it works

### DIFF
--- a/.changeset/honest-singers-guess.md
+++ b/.changeset/honest-singers-guess.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] do not set inlineDynamicImports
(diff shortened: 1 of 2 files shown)


JeanJPNM (approved):

bluwy (approved):
Yes! I've been advocating for this for a while now. With this change we wouldn't need `svelte/ssr` or `svelte` node condition in the first place, though it's also good that we have that today for smaller SSR bundles.

PR #2755, opened 2021-11-06
[fix] unselect before navigation

### LINKED ISSUE #1054: Text selection is kept when switching between svelte routers - should be resetting?
Text selection is kept when switching between svelte routers(If it is in `$layout.svelte`). I don't know if this shouldn't be cleared by the router by default either?

The `router.js` file just executes `document.body.focus();` every time the router is rebooted:
https://github.com/sveltejs/kit/blob/39b69677635a59b40675d375c64bbc4507ea0ea5/packages/kit/src/runtime/client/router.js#L245
And for the scroll position (conditionally):
https://github.com/sveltejs/kit/blob/39b69677635a59b40675d375c64bbc4507ea0ea5/packages/kit/src/runtime/client/router.js#L254
Should text selection be treated the same?

`document.body.focus();` and `scrollTo(0, 0);` is to simulate a full website refresh.
Should `router.js` be run with a line that resets the text selection:
```javascript
window.getSelection().empty();
```

I emphasize: This is NOT needed directly by me, but I noticed this behavior during experiments with sveltekit.

treenoder:
fixes #1054

- *[x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- *[x] This message body should clearly illustrate what problems it solves.
- *[x] Ideally, include a test that fails without this PR but passes with it.

### Tests
- *[x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- *[x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0


packages/kit/src/runtime/client/utils.js
```diff
@@ -9,3 +9,13 @@ export function get_base_uri(doc) {
...
+
+/** @param {Window} win */
+export function unselect_all(win) {
+	if (win.getSelection) {
+		const selection = win.getSelection();
+		if (selection) {
+			selection.removeAllRanges();
+		}
+	}
+}
```
bluwy, line 21: I believe the global variable `window` is generally avoided in the router code. We could probably simplify this as:

```js
getSelection()?.removeAllRanges()
```

With this one line, maybe we don't need a function after all.

packages/kit/src/runtime/client/router.js
```diff
@@ -250,8 +250,8 @@ export class Router {
 	 * @param {string} [hash]
 	 */
 	async _navigate(url, scroll, keepfocus, chain, hash) {
+		unselect_all(window);
```
bluwy, line 253: Assuming after we inline the code, I'm not sure if placing the unselect logic here is correct. I have a feeling it should be at [this line](https://github.com/sveltejs/kit/blob/93726c38d2118b1dc136f3ef659c5ae6bb039138/packages/kit/src/runtime/client/renderer.js#L277), but @benmccann may know better where this fits.

packages/kit/test/apps/basics/src/routes/selection/_tests.js
```diff
@@ -0,0 +1,34 @@
+import * as assert from 'uvu/assert';
+
+/** @type {import('test').TestMaker} */
+export default function (test) {
+	test('reset selection', '/selection/a', async ({ page, clicknav }) => {
+		await page.waitForTimeout(50);
```
bluwy, line 6: Is there a reason to wait for timeout? We have tried to avoid this if possible. If it's really needed, we should add a comment explaining it..

packages/kit/test/apps/basics/src/routes/selection/_tests.js
```diff
@@ -0,0 +1,34 @@
...
+					selection.removeAllRanges();
+					selection.addRange(range);
+					return selection.rangeCount;
+				}
+				return 0;
+			}),
+			1
+		);
+		await clicknav('[href="/selection/b"]');
+		await page.waitForTimeout(50);
```
bluwy, line 22: And the timeout here too.

treenoder: Test was inconsistent without that timeout, on my setup at least. Maybe someone can run tests to confirm.

treenoder: Can't reproduce inconsistencies so waitForTimeout calls was removed. By the way, is there a docker file or something for contributors, to build/test project?

bluwy: > By the way, is there a docker file or something for contributors, to build/test project?

There isn't one that I know of, but I don't have much issues with local test inconsistencies in the past. Playwright ships its own chromium browser IIRC so that should eliminate most of the cross-machine quirks.

packages/kit/src/runtime/client/router.js
```diff
@@ -251,7 +251,6 @@ export class Router {
 	 */
 	async _navigate(url, scroll, keepfocus, chain, hash) {
 		const info = this.parse(url);
-
```
bluwy, line 254: Would be great to leave this file untouched.

bluwy (approved):
I have one small request below, but otherwise this looks good to me! Tests seem to pass too. I'll wait for Ben's comment on my question above before merging.

packages/kit/src/runtime/client/renderer.js
```diff
@@ -275,6 +275,8 @@ export class Renderer {
 			this._init(navigation_result);
 		}
 
+		getSelection()?.removeAllRanges();
+
```
benmccann, line 279: not exactly related to your change, but `if (!opts)` shouldn't the `document.body.focus()` line still be called? I don't quite get what that check is doing and wonder if it should be removed? maybe i'm just sleep deprived enough that I'm making a dumb mistake in reading this code...

bluwy: I'm actually quite confused of the part inside `if (!opts)` too. When I refactored that, I tried to keep it as before. That means `document.body.focus()` and scrolling only happens if `opts` exist, I thought it was intentional.

If it wasn't, maybe we should refactor that again, and always do focusing and scrolling regardless of `opts`. We can do that in a separate PR though.

Rich-Harris:
excellent, thank you!

packages/kit/test/apps/basics/src/routes/selection/_tests.js
```diff
@@ -0,0 +1,32 @@
...
+					selection.removeAllRanges();
+					selection.addRange(range);
+					return selection.rangeCount;
+				}
+				return 0;
+			}),
+			1
+		);
+		await clicknav('[href="/selection/b"]');
+		assert.equal(
```
benmccann, line 21: i'm worried this might introduce flakiness, which is why you sometimes needed the timeout. I wonder if you need to wait until the selection is present instead of simply checking that it is

rmunn:
This is probably causing #2853.

PR #2763, opened 2021-11-09
Support for Vercel beta filesystem API

Rich-Harris:
Vercel now has [edge functions](https://vercel.com/features/edge-functions), which allow us to render pages at the edge instead of inside serverless functions (which suffer from cold boots and higher latency).

This is a first pass at updating the Vercel adapter to take advantage of them, via the (not currently publicly documented) beta filesystem API.

Currently, to deploy an app with the adapter in this PR, you would need to

1. `vc deploy --prebuilt` rather than building on Vercel (which would use the currently published packages)
2. add a `ENABLE_FILE_SYSTEM_API=1` environment variable to the project via the Vercel web UI

Deployed example here: https://fs-api-kit-test-rich-harris.vercel.app/

There's some major outstanding todos:

* [ ] Vercel edge functions don't receive request bodies. Because of that, they're impractical for `POST`/`PUT` requests to endpoints. (Notice in the deployed example above you can't add a new TODO.) In these cases, the middleware will need to fall back to a serverless function, but this means that we essentially need to build two copies of the app (edge functions _and_ serverless functions) and configure the routing manifest accordingly
* [ ] Because edge functions run as 'middleware' before anything else happens, the app needs to know that it should disregard requests for static assets. This means that we need to pass an asset manifest to the adapter. In this PR I've done that by passing `build_data` through unchanged, but this probably warrants a closer look (and in fact the adapter API more generally could use a tidy up and some documentation)
* [ ] For larger apps, having a single edge function respond to every single request probably isn't ideal; ideally we would intelligently route people towards functions that handled a subtree of the app (or a single page/endpoint in the best case). This isn't specific to the Vercel adapter, and is something that needs to be supported by Kit (and the adapter API) itself, but it's arguably more acute in the case of edge functions/cloudflare workers because code needs to be evaluated before each and every request. The existence of [fallthrough routes](https://kit.svelte.dev/docs#routing-advanced-fallthrough-routes) probably makes this all a bit more complicated than it otherwise would be, I haven't looked at it in detail yet
* [ ] The API is still in beta, so things might well change

### Tests

### Changesets

### DIFF
--- a/packages/adapter-vercel/files/entry.js
+++ b/packages/adapter-vercel/files/entry.js
@@ -1,34 +1,42 @@
-import { getRawBody } from '@sveltejs/kit/node';
-
 // TODO hardcoding the relative location makes this brittle
 import { init, render } from '../output/server/app.js';
+import { assets } from './assets.js';
 
 init();
 
-export default async (req, res) => {
-	const { pathname, searchParams } = new URL(req.url || '', 'http://localhost');
+const getResult = (body, options) => ({
+	promise: Promise.resolve(),
+	waitUntil: Promise.resolve(),
+	response: new Response(body, options)
+});
 
-	let body;
+_ENTRIES = typeof _ENTRIES === 'undefined' ? {} : _ENTRIES;
 
-	try {
-		body = await getRawBody(req);
-	} catch (err) {
-		res.statusCode = err.status || 400;
-		return res.end(err.reason || 'Invalid request body');
+_ENTRIES['middleware_kit'] = {
+	async default({ request }) {
+		if (request.method === 'GET') {
+			const { pathname, searchParams } = new URL(request.url);
+
+			if (!assets.has(pathname.slice(1))) {
+				const rendered = await render({
+					method: request.method,
+					headers: request.headers,
+					path: pathname,
+					query: searchParams,
+					rawBody: new Uint8Array() // TODO
+				});
+
+				if (rendered) {
+					const { status, headers, body } = rendered;
+					return getResult(body, { status, headers });
+				}
+			}
+		}
+
+		return getResult(null, {
+			headers: {
+				'x-middleware-next': '1'
+			}
+		});
 	}
-
-	const rendered = await render({
-		method: req.method,
-		headers: req.headers,
-		path: pathname,
-		query: searchParams,
-		rawBody: body
-	});
-
-	if (rendered) {
-		const { status, headers, body } = rendered;
-		return res.writeHead(status, headers).end(body);
-	}
-
-	return res.writeHead(404).end();
 };
--- a/packages/adapter-vercel/files/entry.js
+++ b/packages/adapter-vercel/files/entry.js
@@ -10,33 +10,34 @@ const getResult = (body, options) => ({
 	response: new Response(body, options)
 });
 
-_ENTRIES = typeof _ENTRIES === 'undefined' ? {} : _ENTRIES;
+// eslint-disable-next-line
+_ENTRIES = {
+	middleware_kit: {
+		async default({ request }) {
+			if (request.method === 'GET') {
+				const { pathname, searchParams } = new URL(request.url);
 
-_ENTRIES['middleware_kit'] = {
-	async default({ request }) {
-		if (request.method === 'GET') {
-			const { pathname, searchParams } = new URL(request.url);
+				if (!assets.has(pathname.slice(1))) {
+					const rendered = await render({
+						method: request.method,
+						headers: request.headers,
+						path: pathname,
+						query: searchParams,
+						rawBody: new Uint8Array() // TODO
+					});
 
-			if (!assets.has(pathname.slice(1))) {
-				const rendered = await render({
-					method: request.method,
-					headers: request.headers,
-					path: pathname,
-					query: searchParams,
-					rawBody: new Uint8Array() // TODO
-				});
-
-				if (rendered) {
-					const { status, headers, body } = rendered;
-					return getResult(body, { status, headers });
+					if (rendered) {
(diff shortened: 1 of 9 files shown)


packages/adapter-vercel/index.js
```diff
@@ -31,12 +31,24 @@ export default function (options) {
 			utils.log.minor('Generating serverless function...');
 			utils.copy(join(files, 'entry.js'), '.svelte-kit/vercel/entry.js');
 
+			const assets = new Set([...utils.build_data.static, ...utils.build_data.client]);
+
+			utils.build_data.static.forEach((file) => {
+				if (file.endsWith('/index.html')) {
```
antony, line 37: probably pedantic but to remove the magic number -11 I would probably do either:

```js
const index_filename = '/index.html'
if (file.endsWith(index_filename)) {
	assets.add(file.slice(0, -(index_filename.length)));
}
```

or, even

```js
assets.add(path.dirname(file))
```

packages/adapter-vercel/index.js
```diff
@@ -13,14 +13,14 @@ export default function (options) {
...
-			const dir = '.vercel_build_output';
+			const dir = '.output';
 			utils.rimraf(dir);
 
 			const files = fileURLToPath(new URL('./files', import.meta.url));
 
 			const dirs = {
 				static: join(dir, 'static'),
-				lambda: join(dir, 'functions/node/render')
+				pages: join(dir, 'server/pages')
```
antony, line 23: surely `join(dir, 'server', 'pages')` would be preferable here?

packages/adapter-vercel/files/entry.js
```diff
@@ -1,34 +1,43 @@
...
+});
+
+// eslint-disable-next-line
+_ENTRIES = {
+	middleware_kit: {
+		async default({ request }) {
+			if (request.method === 'GET') {
+				const { pathname, searchParams } = new URL(request.url);
+
+				if (!assets.has(pathname.slice(1))) {
```
antony, line 20: you could probably remove a level of nesting here, and a hard-to-grok negation:

```js
if (assets.has(pathname.slice(1))) {
  return
}
```

then the rest could be a level less-nested.

antony (commented):
Just nitpicks really - came here to see what the new fs api looked like.

packages/create-svelte/templates/default/package.json
```diff
@@ -4,7 +4,8 @@
 	"scripts": {
 		"dev": "svelte-kit dev",
 		"build": "svelte-kit build --verbose",
-		"start": "svelte-kit start"
+		"start": "svelte-kit start",
+		"deploy": "ADAPTER=@sveltejs/adapter-vercel npm run build && vercel deploy --prebuilt"
```
benmccann, line 8: Was this checked in by accident? I don't think it'd work unless you also add this to the default `package.json`

packages/adapter-vercel/files/entry.js
```diff
@@ -1,34 +1,43 @@
...
-		res.statusCode = err.status || 400;
-		return res.end(err.reason || 'Invalid request body');
+const getResult = (body, options) => ({
+	promise: Promise.resolve(),
+	waitUntil: Promise.resolve(),
+	response: new Response(body, options)
+});
+
+// eslint-disable-next-line
+_ENTRIES = {
```
benmccann, line 14: a few comments / links to vercel docs would be useful here. I don't know what `_ENTRIES`/`middleware_kit`, 'x-middleware-next', `waitUntil`, etc. are

packages/create-svelte/templates/default/package.json
Rich-Harris: yes, this was just for testing

packages/adapter-vercel/files/entry.js
Rich-Harris: sadly not — you need the `x-middleware-next` thing otherwise it will fail cryptically. so we'd either need to duplicate the `return` statement later on, or construct the URL before the main `if` block and add it to the condition there

packages/adapter-vercel/files/entry.js
Rich-Harris: the docs aren't public yet, i've been sneaking a peek at some in-development stuff. agree that the API is very cryptic in its current form

packages/adapter-vercel/index.js
Rich-Harris: i tend to think these sorts of magic numbers make sense in context, but maybe that's just me. an alternative is to just comment it

```suggestion
				if (file.endsWith('/index.html')) {
					// remove /index.html suffix
```

packages/adapter-vercel/index.js
Rich-Harris: it's all the same to node, so i personally prefer `server/pages` as it's more compact and readable

packages/adapter-vercel/index.js
Conduitry: I'm +1 for just commenting this rather than eliminating the magic number entirely.

packages/adapter-vercel/files/entry.js
leo: > the docs aren't public yet

Indeed! They are available [here](https://docs-git-file-system-api-docs.vercel.sh/docs/file-system-api) at the moment. We're still working on adding details about the things mentioned above, however. And the API will evolve to be simpler over the coming months — in tandem with Next.js' output. Regardless, the API currently shown in the PR will be retained forever for backward-compatibility.

leo (commented):
I recommend using something different than the `_middleware.js` file name, since we're not dealing with Middleware here. The Edge Function doesn't run on top of something ("middleware") — instead it renders something.

Issue #2762, opened 2021-11-09
Support dynamic content of app.html


atuttle:
### Describe the problem

While I imagine the current static approach for app.html is acceptable in ~99% of use cases, sometimes it would be useful to make its contents dynamic.

I have a multi-tenant application that loads a customer-provided skin at runtime and wraps my app in it. Hypothetically, let's say the skin looks like this:

```html
<html>
<head>
<title>APP!</title>
</head>
<body>
<p>some content before the application</p>
{{app}}
<p>some content after the application</p>
</body>
</html>
```

Achieving this isn't particularly difficult.

```js
//__layout.js
<script context="module">
  //load the skin async based on things like hostname
  //and other portions of the URL, then...
  return { props: { skin: skinHTML } };
</script>

const [before, after] = skin.split('{{app}}');
// ... some regex magic to strip out the html/head/body tags
{@html before}
<slot />
{@html after}
```

Indeed, this is exactly what I'm currently doing. But there's a problem. If the skin _wraps_ the app in a tag, Svelte(Kit?) starts to misunderstand the DOM and the app gets duplicated. Given a skin that contains this: `<div>{{app}}</div>`...

The app behaves very oddly, and most of the time results in 2 copies of itself on the page. (To the best of my understanding, sometimes in dev hmr it will clean itself up to a single instance, but not always)

Here's a video demonstrating why I want to do this, and how it affects the application: https://youtu.be/voYqm2Chu1U

If the contents of app.html were dynamic, it would be trivial for me to replace `{{app}}` with `<div id="svelte"></div>`.

### Describe the proposed solution

I can see that everything is currently hard-coded to read app.html from the file system and use that value.

https://github.com/sveltejs/kit/blob/fb262e5d3ac551b369cf3f50d8a073975f8b0606/packages/kit/src/core/dev/index.js#L462-L465

It would be nice if there were some sort of hook or middleware that I could add that would be called instead that could return the HTML contents currently obtained from app.html. It would need to happen at runtime (server-side) so that I can parse the URL to get the correct skin for the current hostname.

### Alternatives considered

The only other thing I can think of is to use Vanilla Svelte and express/etc to set the page contents to the skin at runtime. I'd hate to give up the routing/etc benefits of SvelteKit, though.

I haven't been able to think of any other alternatives, but I'm open to any current/proposed solution that lets me build the features I need in my app. It doesn't have to be the solution I proposed.

### Importance

would make my life easier

Mlocik97:
[...]

Issue #2221, opened 2021-08-16
use `document.currentScript` instead of `target` option


Rich-Harris:
### Describe the problem

This is a super niche problem to have, but it's one I've faced on two recent projects. Currently, SvelteKit knows which bit of markup to hydrate because of tight coupling between the config... https://github.com/sveltejs/kit/blob/446c3469276d37190f6d783fddf12575372cae8e/packages/create-svelte/templates/skeleton/svelte.config.js#L12... and the app template: https://github.com/sveltejs/kit/blob/446c3469276d37190f6d783fddf12575372cae8e/packages/create-svelte/templates/skeleton/src/app.html#L10

The tight coupling is unfortunate in any case, but where this gets really tricky is if you're using SvelteKit to generate markup that will be embedded multiple times on a page, as in this pseudo-code:

```js
// src/app.html wraps the %svelte.head% and %svelte.body% with these tags
const head_start = '<!-- START:HEAD -->';
const head_end = '<!-- END:HEAD -->';
const body_start = '<!-- START:BODY -->';
const body_end = '<!-- END:BODY -->';

// after building app with adapter-static
for (const file of glob('build/**/index.html')) {
  const html = fs.readFileSync(file);

  const head = contents.slice(contents.indexOf(head_start) + head_start.length, contents.indexOf(head_end));
  const body = contents.slice(contents.indexOf(body_start) + body_start.length, contents.indexOf(body_end));

  const slug = file.slice('build/'.length, -'/index.html'.length);
  const metadata = get_metadata(slug);

  await cms.upload({ ...metadata, head, body });
}
```

The script ends up hydrating the wrong markup unless you use a placeholder selector instead of `#svelte` and rewrite it to be unique:

```diff
-target: '#svelte',
+target: '#PLACEHOLDER',
```

```diff
-<div id="svelte">%svelte.body%</div> 
+<div id="PLACEHOLDER">%svelte.body%</div> 
```

```diff
-const html = fs.readFileSync(file);
+const html = fs.readFileSync(file).replace(/PLACEHOLDER/g, `svelte-${Math.round(Math.random() * 1e9)}`);
```

That's pretty ugly (and brittle, since there's a non-zero chance of collisions. We're currently using page slugs instead of random numbers, which avoids collisions, but it turns out that sometimes the same embed will be used multiple times in the NYT live blog...), and involves deep knowledge about how SvelteKit works.

### Describe the proposed solution

What if `<div>%svelte.body%</div>` turned into this?

[...]

PR #2773, opened 2021-11-10
Support overwriting the default template loader via a handle hook

atuttle:
Ref #2762

[Here's a brief video explaining why I can't accomplish the same thing without this change](https://youtu.be/voYqm2Chu1U). (Apologies for the dog snoring in the background! 🐶 💤 )

Basically, our app is multi-tenant and we allow our customers to provide their own skins for different sections. But most importantly: **the skin content is not within our control.** All customers run on the same cluster of servers, and at runtime we inspect the hostname/url and look up the appropriate skin to wrap around the app.

As discussed in #2762, I tried to make this work with `{@html before}<slot />{@html after}` in a layout, and that works _as long as the skin doesn't have ANY tags wrapping the app content_ (must be a direct descendant of `<body>`). Since I'm splitting the skin content on a token, any wrapping tags become unclosed, and then `{@html ...}` goes haywire.

This change allows me to overwrite the default `template({ head, body})` method called during SSR by specifying a custom per-request template method in a `handle()` hook:

```js
// src/hooks.js
import loadSkin from '...';
	const skinHTML = await loadSkin(request.host, request.path);
	request.locals.template = ({ head, body }) => {
		return skinHTML
			.replace('</head>', `${head}</head>`)
			.replace('{{app}}', `<div id="svelte">${body}</div>`);
	};

	return resolve(request);
}
```

I'm not sure how I could write a test for this functionality, but if someone wants to offer some advice, I can give it a try.

### Tests

I also am not able to get the tests to pass locally; both before and after my change they fail at the same spot:

image

### Changesets

### DIFF
--- a/.changeset/metal-feet-run.md
+++ b/.changeset/metal-feet-run.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+added support for dynamic base template contents (app.html) during SSR
--- a/packages/kit/src/runtime/server/page/respond.js
+++ b/packages/kit/src/runtime/server/page/respond.js
@@ -28,6 +28,11 @@ export async function respond(opts) {
 	/** @type {Array<SSRNode | undefined>} */
 	let nodes;
 
+	//allow developer to overwrite default template function during SSR (e.g. via `handle` hook)
+	if (request.locals.template) {
+		options.template = request.locals.template;
+	}
+
 	try {
 		nodes = await Promise.all(route.a.map((id) => (id ? options.load_component(id) : undefined)));
 	} catch (err) {


rmunn:
Not being part of the core team, I can't approve PRs. But I can give you some feedback. I suspect that `request.locals` is not going to be the right place to put this, as Svelte-Kit tried very hard not to touch the `locals` object or require any particular properties to be in it. Otherwise this feature *will* interfere with someone's code at some point, where they want to set a `template` property in their `request.locals` for some reason of their own, and suddenly Svelte-Kit has reserved a property name in the object that is supposed to be 100% for user data.

Now, using something other than `request.locals` will mean that you'll have to change the data type of the `request` in some way, or perhaps add an optional second parameter to the `resolve` function passed to `handle` (where that second parameter would be your `({head, body}) => "template"` function). So obviously it's tempting to use `request.locals`. But I'm pretty sure the core team won't approve a PR that makes Svelte-Kit rely on `request.locals` having any specific structure.

atuttle:
Appreciate the feedback @rmunn. I'm happy to make whatever the necessary changes are, once I'm given some ("blessed") direction.

benmccann:
> Not being part of the core team, I can't approve PRs. But I can give you some feedback

@rmunn thanks so much for all your contributions! We'd love to keep you involved. Are you on Discord by chance? I sent you an email, but just wanted to mention it here as well to make sure I had the right address :smile:

Rich-Harris:
This is a fun problem to solve. I'm wondering if you could do this sort of thing...

```diff
<!-- src/app.html -->
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <link rel="icon" href="/favicon.png" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />

    %svelte.head%
  </head>
  <body>
+    <!-- BEFORE -->
    <div id="svelte">%svelte.body%</div>
+    <!-- AFTER -->
  </body>
</html>
```

...and then, in `src/hooks.js`...

```js

  if (response.headers['content-type'] === 'text/html') {
    const skin = await loadSkin(request.host, request.path);
    const [before, after] = skin.split('{{app}}');
    response.body = response.body
      .replace('<!-- BEFORE -->', before)
      .replace('<!-- AFTER -->', after || '');
  }

  return response;
}
```

...and have users provide a skin that only includes `<body>` rather than `<head>` content?

(One minor detail that shouldn't affect anything, but is worth mentioning just in case — I'd like to remove the `id="svelte"` stuff in the near future: https://github.com/sveltejs/kit/issues/2221)

@rmunn is right that `request.locals` is a non-starter, unfortunately — we'd basically need to change the `resolve` API to allow additional arguments — which is why I'm wondering if this solution would get you where you need to go.

atuttle:
@Rich-Harris excellent thinking! And that may not even require any SK mods. I'll give it a go and get back to you. (I was looking forward to having been a contributor tho! 😞  Guess I'll have to find something else I can help out with...)

Congrats on the new job, by the by.

atuttle:
As expected, that worked fantastically. I guess I have to withdraw this PR, then.

benmccann:
I'll go ahead and close this then since it sounds like you're no longer pursuing it

7antra:
Hey team ! 

Sorry, I don't want to reopen this, but how far is difficult to have a dynamic template ? for exemple if you have one template for main app : app.html and an other for newsletter for exemple newsletter.html with tags `%svelte.head%` and `%svelte.body%` inside. 

Is it very difficult for load the one we want depending URL.pathname ?

(Sorry for my English btw)

atuttle:
@7antra I documented the exact steps I ended up taking, which sounds like it can do what you want: https://adamtuttle.codes/blog/2021/sveltekit-customizing-app-html-at-runtime/

PR #2781, opened 2021-11-11
[fix] add test cases and fix `is_content_type_textual`

si3nloong:

### Tests

### Changesets

Nothing much, just add more test cases


Rich-Harris:
thank you! mind fixing the lint errors?

for reference, it's helpful if you don't delete the pull request template — it includes a checklist that includes things like linting, and it's useful for maintainers to be able to understand the scope of a PR (whether it requires a version bump, etc)

packages/kit/src/runtime/paths.spec.js
```diff
@@ -0,0 +1,14 @@
+import { test } from 'uvu';
+import * as assert from 'uvu/assert';
+import { set_paths, base, assets } from './paths.js';
+
+test('paths', () => {
```
benmccann, line 5: `set_paths` might be a clearer name

but I question if this is really a useful test. just testing that setting a variable sets it - there doesn't seem to be much room for error. and I would expect this is already covered by our other tests

packages/kit/src/utils/error.spec.js
```diff
@@ -0,0 +1,17 @@
+import { test } from 'uvu';
+import * as assert from 'uvu/assert';
+import { coalesce_to_error } from './error.js';
+
+test('coalesce_to_error', () => {
```
benmccann, line 5: this is at least a little more complicated than the last test, but still would think this functionality is likely already covered by an integration test and that there's little room to go wrong here. I and the other maintainers agree that we should rely more on unit tests than integration tests, but I think the units could be for more complicated functionality

si3nloong:
@benmccann can you review again

packages/kit/src/runtime/paths.spec.js
si3nloong: removed

packages/kit/src/utils/error.spec.js
si3nloong: test removed

Rich-Harris (approved):

Rich-Harris:
thanks!

Issue #2789, opened 2021-11-12
Add info `Port 3000 is occupied` for `npm run dev`


Zerotask:
### Describe the problem

If you run the default `npm run dev` without specifying a specific port (-p), you will get the following error, if the port is already in use:
```
> listen EACCES: permission denied 127.0.0.1:3000
Error: listen EACCES: permission denied 127.0.0.1:3000
    at Server.setupListenHandle [as _listen2] (node:net:1317:21)
    at listenInCluster (node:net:1382:12)
    at doListen (node:net:1520:7)
    at processTicksAndRejections (node:internal/process/task_queues:84:21)
```

If you run `npm run preview` and the port is already in use, you will get a more beginner friendly error message:
```
Port 3000 is occupied
Terminate the process occupying the port or specify a different port with --port
```

### Describe the proposed solution

Unify the error message. If a port is already in use, provide the same error message.

### Importance

nice to have

benmccann:
What OS are you running?

Zerotask:
Windows 10

Conduitry:
FWIW, I can't reproduce this. I tried this on Windows 10 earlier today (with both Node 14 and Node 16) and running `npm run dev` simultaneously in two different projects, and the second one displayed the more friendly error, and not the `EACCES` mentioned in the issue.

Zerotask:
I just tried to reproduce this right after I started my PC.

on the first run of `npm run dev` it started without problems (port was not occupied). Then after it was running, I started the command again (with the same project and another project) and got the user-friendly error message and not EACESS.

Then I stopped it and started an express web server on port 3000 and after it, I started `npm run dev`. The Vite dev server started without problems, but it can't be accessed. I got 404 from express. I tried the same with a Go (gin) web server and it was the same.
The expected behavior would be to see the user-friendly error message in this case.
Besides that, I could not reproduce the EACESS error. My Svelte, Node, NPM etc. versions were the same as from this test https://github.com/sveltejs/kit/pull/2792#issuecomment-968368935

The only difference was, that I left the Windows Insider Program (Release Preview Channel) a few days ago. But I can't imagine that this was the cause.

bluwy:
Thanks for the thorough investigation @Zerotask! Since the bug is in Vite, I'll close this in favour of https://github.com/vitejs/vite/issues/5801, which seems to be having similar issues. Your comment there would help too, and it seems like a workaround is also found there.

PR #2792, opened 2021-11-12
[chore] add port check for dev command

Zerotask:
Close https://github.com/sveltejs/kit/issues/2789

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/cli.js
+++ b/packages/kit/src/cli.js
@@ -85,6 +85,8 @@ prog
 	.option('-H, --https', 'Use self-signed HTTPS certificate')
 	.option('-o, --open', 'Open a browser tab')
 	.action(async ({ port, host, https, open }) => {
+		await check_port(port);
+
 		process.env.NODE_ENV = process.env.NODE_ENV || 'development';
 		const config = await get_config();
 
--- a/packages/kit/src/cli.js
+++ b/packages/kit/src/cli.js
@@ -85,10 +85,13 @@ prog
 	.option('-H, --https', 'Use self-signed HTTPS certificate')
 	.option('-o, --open', 'Open a browser tab')
 	.action(async ({ port, host, https, open }) => {
-		await check_port(port);
-
 		process.env.NODE_ENV = process.env.NODE_ENV || 'development';
 		const config = await get_config();
+		const vite_config = config.kit.vite();
+
+		if (!vite_config.server || vite_config.server.strictPort !== false) {
+			await check_port(port);
+		}
 
 		const { dev } = await import('./core/dev/index.js');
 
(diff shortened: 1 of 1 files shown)


bluwy:
I don't think it's correct to check the ports in dev mode. There is a possibility someone set `kit.vite.server.fs.strictPort: false`, which would search a different port, e.g. `3001` and run that. I think the better way to solve this is to do the checks in Vite instead.

Zerotask:
The current solution (as mentioned in the issue) is not good and not beginner friendly. This should be changed. Either this way or with another way.
Checking ports is pretty straightforward and it's the same function which is used for the preview command. Maybe it could be wrapped around an if statement which checks this property (if it has access to it).

bluwy:
Sorry, I meant `kit.vite.server.strictPort`. Yes I agree it's not beginner-friendly, however I think the best fix is in Vite itself directly, that way it'll benefit anyone using Vite too.

bluwy:
Actually I just tested in Vite and it did print a human-friendly message. Maybe it's the way we call the Vite dev server, or we're calling the port earlier before Vite? I haven't dug deep into this but I'm curious why Vite's message doesn't surface.

packages/kit/src/cli.js
```diff
@@ -85,10 +85,13 @@ prog
 	.option('-H, --https', 'Use self-signed HTTPS certificate')
 	.option('-o, --open', 'Open a browser tab')
 	.action(async ({ port, host, https, open }) => {
-		await check_port(port);
-
 		process.env.NODE_ENV = process.env.NODE_ENV || 'development';
 		const config = await get_config();
+		const vite_config = config.kit.vite();
+
+		if (!vite_config.server || vite_config.server.strictPort !== false) {
```
Zerotask, line 92: As a work-around for the moment: if server is not set or strictPort is not false, do a check port

Rich-Harris:
Can you describe the steps to reproduce this? This is what I see if I `npm run dev` while the port is occupied (with a freshly made project, no alterations to the config):

image

As an aside, I notice that we're calling `config.vite()` twice — once in `cli.js`, once in `dev.js`. This PR didn't introduce it, but if we end up changing things around then it's probably a good opportunity to clean it up. Maybe we should attach the relevant info to the `watcher`, and only call `config.vite()` inside `dev.js`?

Zerotask:
I just tested it again and still got the error:
> listen EACCES: permission denied 127.0.0.1:3000
Error: listen EACCES: permission denied 127.0.0.1:3000
    at Server.setupListenHandle [as _listen2] (node:net:1317:21)
    at listenInCluster (node:net:1382:12)
    at doListen (node:net:1520:7)
    at processTicksAndRejections (node:internal/process/task_queues:84:21)

I followed the steps from kit.svelte.dev:
> npm init svelte@next my-app
cd my-app
npm install
npm run dev -- --open

I did not change any config files. I just created the svelte demo app and played around with some svelte files, nothing special.

Rich-Harris:
Looking into this further, it seems that there's an underlying Vite issue: https://github.com/vitejs/vite/issues/5801. It's probably best that we don't add temporary code to work around that issue, so I'll close this. Thank you

PR #2810, opened 2021-11-17
[docs] update FAQ with newer info about Firebase and Vite

benmccann:
We can tweak it when Vite 2.7 is released, but thought I'd at least get some newer info out there in the meantime


benmccann:
i'm going to go ahead and merge this to kick off a redeploy of the site

Issue #2785, opened 2021-11-12
Weird behavior with `page.query` change and `goto`


csangonzo:
### Describe the bug

I wanted to create a dropdown select in my navigation header, since I'll need this selected value as a global value used on the client and on the server as well I decided that I'll use a query param (if you have a good solution to create a global state that stays the same when refreshing the page and is available on the server and on the client without any woodoo stuff, please let me know :) ). So every time someone changes this value, the page has to refresh using this newly set id on the current page. First I tried with `$page.query.set('test',paramTest);`, since the documentation says:
> The load function is reactive, and will re-run when its parameters change, but only if they are used in the function. Specifically, if page.query, page.path, session, or stuff are used in the function, they will be re-run whenever their value changes. 

However I couldn't get the load function to rerun by changing the `page.query`, so I decided to pair it with `goto`. So now if I change (set) the query param and use `goto` afterwards it's fine.. except on every second click somehow.. I have a list of let's say four ids, if I'm on the first one (1) by default and I change it to the second (2), the load function runs and everything's fine, same for the third one (3), but if I continue, the fourth (4) one won't do anything, the load function won't run again. I can carry on with this and the first will run, the second won't, the third will and the fourth won't.
I have no idea why this happens, but it's kinda weird.

### Reproduction

I created a whole new sveltekit template with:
`npm init svelte@next my-app`
I added two extra files under `routes`:
`__layout.svelte`:

```html
<script>
    import { goto } from '$app/navigation';
    import { page } from '$app/stores';

    let paramTest = $page.query.get('test');
</script>

<nav>
    
	<a href="/">Home</a>
	<a href="/test">Test</a>

    <select 
    bind:value={paramTest}
    on:change={() => {
        $page.query.set('test',paramTest);
        goto(`?${$page.query.toString()}`);
    }}>
        <option value="1">1</option>
        <option value="2">2</option>
        <option value="3">3</option>
        <option value="4">4</option>
    </select>
</nav>

<slot></slot>
```

`test.svelte`
```html
<script context="module">
	export async function load({ page}) {
        console.log('TEST LOAD',page.query.get('test'));
        const a = page.query.get('test');

        return {}
    }
</script>


<script lang="ts">
    import { page } from '$app/stores';
</script>


<p>
    This is a test page
    <br>
    Param: {$page.query.get('test')}
</p>
```

The even strangest part of this is that with the code below you can reproduce the issue:
```js
$page.query.set('test',paramTest);
[...]

Issue #969, opened 2021-04-11
Add a way to change the query params on a page easily.


davjhan:
**Is your feature request related to a problem? Please describe.**
I would like an easy way to change the query params in the address bar at runtime.

**Usecases:**
- I want to sync a text input on the page to a query param so that when the user copies the address from the address bar and shares it, the link automatically has the field pre-filled.
- I want to pass an error message to another page that I want to display temporarily. I set the error as a query param and use `goto`  to redirect to that page. If i let the user acknowledge the error banner and hide it, I want to remember that by stripping the query param from the URL so that when they refresh, the banner doesn't re-appear.

**Describe the solution you'd like**
An API that allows me to change the url, with convenient syntax for adding and removing query params. Ideally, this will not trigger a refresh of the page in itself.

Perhaps augment the `goto` API? https://kit.svelte.dev/docs#modules-app-navigation
**Describe alternatives you've considered**
There are workarounds for the usecases that I've mentioned, but this is a common enough usecase imo.

**How important is this feature to you?**
Not a deal-breaker but it would allow me to have a simpler architecture that stores state more obviously, in the url, instead of a store.
**Additional context**
As an aside, It would also be nice to have a slicker way to append query params to a `goto` call. Currently, judging by the signature from the documentation, theres no way to do it.

`goto(href, { replaceState, noscroll })`

Thanks!

mohe2015:
Which workarounds are you thinking of @davjhan ?

mohe2015:
Just realized when the router has fully loaded it seems like this already works.

davjhan:
So you can use the `$page.query` object in  `import {page} from '$app/stores'`, but this is an
[URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) object and not a svelte store. As such, there is no reactivity that comes out of the box. The workaround is you creating a custom store that auto updates this URLSearchParams instance AND manually overwrites the url.

Here is a sample component that you can use to test how this works.

```svelte
<script>
    import {page} from '$app/stores'
</script>
<main class='p-4 flex flex-col items-center'>
    <button on:click={()=> $page.query.set("weather","sunny")}>Set query param</button>
    <h1>{$page.query.get("weather")}</h1>
    <button on:click={()=> console.log($page.query.get("weather"))}>print weather to console</button>
</main>
```
[...]

PR #2811, opened 2021-11-17
[docs] Add more info on query params

mohamadHarith:
This PR adds more info on how to update the query params of a page, why we need to use SvelteKit's fetch and whether the load function will be called during navigation using goto. These informations are unavailable in the docs and has caused confusions as seen in [#2785 ](https://github.com/sveltejs/kit/issues/2785) and [#969](https://github.com/sveltejs/kit/issues/969)

- [ x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [ x] This message body should clearly illustrate what problems it solves.
- [x ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x ] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x ] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

### DIFF
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -78,7 +78,7 @@ SvelteKit's `load` receives an implemention of `fetch`, which has the following 
 
 > Code called inside `load` blocks:
 >
-> - should use the SvelteKit-provided [`fetch`](#loading-input-fetch) wrapper rather than using the native fetch as the latter will make the request twice each during SSR and hydration
+> - should use the SvelteKit-provided [`fetch`](#loading-input-fetch) wrapper rather than using the native `fetch`
 > - should not reference `window`, `document`, or any browser-specific objects
 > - should not directly reference any API keys or secrets, which will be exposed to the client, but instead call an endpoint that uses any required secrets
 
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -92,7 +92,7 @@ The `load` function receives an object containing four fields — `page`, `fetch
 
 #### page
 
-`page` is a `{ host, path, params, query }` object where `host` is the URL's host, `path` is its pathname, `params` is derived from `path` and the route filename, and `query` is an instance of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).
+`page` is a `{ host, path, params, query }` object where `host` is the URL's host, `path` is its pathname, `params` is derived from `path` and the route filename, and `query` is an instance of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Mutating `page` does not update the current URL, rather, it should be done by means of navigation using `goto`.
 
 So if the example above was `src/routes/blog/[slug].svelte` and the URL was `https://example.com/blog/some-post?foo=bar&baz&bizz=a&bizz=b`, the following would be true:
 
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -92,7 +92,7 @@ The `load` function receives an object containing four fields — `page`, `fetch
 
 #### page
 
-`page` is a `{ host, path, params, query }` object where `host` is the URL's host, `path` is its pathname, `params` is derived from `path` and the route filename, and `query` is an instance of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Mutating `page` does not update the current URL, rather, it should be done by means of navigation using `goto`.
+`page` is a `{ host, path, params, query }` object where `host` is the URL's host, `path` is its pathname, `params` is derived from `path` and the route filename, and `query` is an instance of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Mutating `page` does not update the current URL; you should instead navigate using [`goto`](#modules-$app-navigation).
 
 So if the example above was `src/routes/blog/[slug].svelte` and the URL was `https://example.com/blog/some-post?foo=bar&baz&bizz=a&bizz=b`, the following would be true:
 
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -78,7 +78,7 @@ SvelteKit's `load` receives an implemention of `fetch`, which has the following 
 
 > Code called inside `load` blocks:
 >
-> - should use the SvelteKit-provided [`fetch`](#loading-input-fetch) wrapper rather than using the native `fetch`
+> - should use the SvelteKit-provided [`fetch`](#loading-input-fetch) wrapper rather than using the native `fetch`. Although, using native `fetch` will work, it will cause the API server to be called twice - during SSR and hydration
 > - should not reference `window`, `document`, or any browser-specific objects
 > - should not directly reference any API keys or secrets, which will be exposed to the client, but instead call an endpoint that uses any required secrets
 
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -78,7 +78,7 @@ SvelteKit's `load` receives an implemention of `fetch`, which has the following 
 
 > Code called inside `load` blocks:
 >
-> - should use the SvelteKit-provided [`fetch`](#loading-input-fetch) wrapper rather than using the native `fetch`. Although, using native `fetch` will work, it will cause the API server to be called twice - during SSR and hydration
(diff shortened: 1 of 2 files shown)


documentation/docs/03-loading.md
```diff
@@ -78,7 +78,7 @@ SvelteKit's `load` receives an implemention of `fetch`, which has the following
 
 > Code called inside `load` blocks:
 >
-> - should use the SvelteKit-provided [`fetch`](#loading-input-fetch) wrapper rather than using the native `fetch`
+> - should use the SvelteKit-provided [`fetch`](#loading-input-fetch) wrapper rather than using the native `fetch`. Although, using native `fetch` will work, it will cause the API server to be called twice - during SSR and hydration
```
gtm-nayan, line 81: How about

"... using the native `fetch` as the latter will make the request twice each during SSR and hydration"

mohamadHarith: yes, this sound better

documentation/docs/03-loading.md
```diff
@@ -78,7 +78,7 @@ SvelteKit's `load` receives an implemention of `fetch`, which has the following
 
 > Code called inside `load` blocks:
 >
-> - should use the SvelteKit-provided [`fetch`](#loading-input-fetch) wrapper rather than using the native `fetch`
+> - should use the SvelteKit-provided [`fetch`](#loading-input-fetch) wrapper rather than using the native fetch as the latter will make the request twice each during SSR and hydration
```
benmccann, line 81: I like that this explains why, but it could be said a bit more clearly as it somewhat sounds like twice during SSR and twice during hydration. also not sure why you'd take fetch out of backticks. 
```suggestion
> - should use the SvelteKit-provided [`fetch`](#loading-input-fetch) wrapper rather than using the native `fetch` because SvelteKit's version has built-in hydration support to avoid making a second request on the client.
```

documentation/docs/05-modules.md
```diff
@@ -22,7 +22,7 @@ import { amp, browser, dev, mode, prerendering } from '$app/env';
 import { goto, invalidate, prefetch, prefetchRoutes } from '$app/navigation';
 ```
 
-- `goto(href, { replaceState, noscroll, keepfocus, state })` returns a `Promise` that resolves when SvelteKit navigates (or fails to navigate, in which case the promise rejects) to the specified `href`. The second argument is optional:
+- `goto(href, { replaceState, noscroll, keepfocus, state })` returns a `Promise` that resolves when SvelteKit navigates (or fails to navigate, in which case the promise rejects) to the specified `href`. Causes the `load` function of the `href` to be called. The second argument is optional:
```
benmccann, line 25: this change seems unnecessary to me. can you explain why you thought it wouldn't be called?

documentation/docs/03-loading.md
benmccann, line 81: I'm not sure about this change. the reason to do it is given right above this. now the reason is being repeated, but only a partial explanation is being given

it also could be said a bit more clearly as it somewhat sounds like twice during SSR and twice during hydration. also not sure why you'd take fetch out of backticks

documentation/docs/05-modules.md
```diff
@@ -57,7 +57,15 @@ Because of that, the stores are not free-floating objects: they must be accessed
...
 - `navigating` is a [readable store](https://svelte.dev/tutorial/readable-stores). When navigating starts, its value is `{ from, to }`, where `from` and `to` both mirror the `page` store value. When navigating finishes, its value reverts to `null`.
-- `page` is a readable store whose value reflects the object passed to `load` functions — it contains `host`, `path`, `params` and `query`. See the [`page` section](#loading-input-page) above for more details.
+- `page` is a readable store whose value reflects the object passed to `load` functions — it contains `host`, `path`, `params` and `query`. See the [`page` section](#loading-input-page) above for more details. Updating the query params of a page can be done by means of navigation by creating a new instance of `$page.query` and using the [goto](#modules-$app-navigation). For example:
+  
+  ```
+  let query = new URLSearchParams($page.query.toString());
+
+  query.set('word', word);
+      
+  goto(`?${query.toString()}`);
```
benmccann, line 67: could use `const` and don't need the double spacing
```suggestion
  const query = new URLSearchParams($page.query.toString());
  query.set('word', word);
  goto(`?${query.toString()}`);
```

documentation/docs/05-modules.md
```diff
@@ -57,7 +57,15 @@ Because of that, the stores are not free-floating objects: they must be accessed
 The stores themselves attach to the correct context at the point of subscription, which means you can import and use them directly in components without boilerplate. However, it still needs to be called synchronously on component or page initialisation when `$`-prefix isn't used. Use `getStores` to safely `.subscribe` asynchronously instead.
 
 - `navigating` is a [readable store](https://svelte.dev/tutorial/readable-stores). When navigating starts, its value is `{ from, to }`, where `from` and `to` both mirror the `page` store value. When navigating finishes, its value reverts to `null`.
-- `page` is a readable store whose value reflects the object passed to `load` functions — it contains `host`, `path`, `params` and `query`. See the [`page` section](#loading-input-page) above for more details.
+- `page` is a readable store whose value reflects the object passed to `load` functions — it contains `host`, `path`, `params` and `query`. See the [`page` section](#loading-input-page) above for more details. Updating the query params of a page can be done by means of navigation by creating a new instance of `$page.query` and using the [goto](#modules-$app-navigation). For example:
```
benmccann, line 60: this seems like an odd place to add this detail

mohamadHarith: Could you please suggest a better a place to add this detail? i

documentation/docs/05-modules.md
mohamadHarith: yes this seems unnecessary, i will remove it.

documentation/docs/03-loading.md
mohamadHarith: Ok, will remove this change.

documentation/docs/05-modules.md
bluwy: I'm not sure where this will fit either, it sounds more like a tip than documentation. Putting it in the FAQ may be too niche too. What if we word this differently, like for [page](https://kit.svelte.dev/docs#loading-input-page) we can say that "mutating page does not update the current URL".

mohamadHarith: alright, will remove this then.

documentation/docs/03-loading.md
```diff
@@ -92,7 +92,7 @@ The `load` function receives an object containing four fields — `page`, `fetch
 
 #### page
 
-`page` is a `{ host, path, params, query }` object where `host` is the URL's host, `path` is its pathname, `params` is derived from `path` and the route filename, and `query` is an instance of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).
+`page` is a `{ host, path, params, query }` object where `host` is the URL's host, `path` is its pathname, `params` is derived from `path` and the route filename, and `query` is an instance of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Mutating `page` does not update the current URL, rather, it should be done by means of navigation using `goto`.
```
mohamadHarith, line 95: @benmccann @bluwy can we add this information here?

benmccann: this sounds reasonable. I wonder if an even better solution than documenting it might be to make the store exposed to the user readonly? https://github.com/sveltejs/svelte/pull/6518

bluwy: I think `page` is referred here as an object rather than a store, but if it was imported from `$app/stores`, it's already a [readable store](https://kit.svelte.dev/docs#modules-$app-stores). So we would need to freeze the object, but I'm not sure if it's something we want to do. The change looks good to me though.

bluwy (approved):

documentation/docs/03-loading.md
Rich-Harris:
```suggestion
`page` is a `{ host, path, params, query }` object where `host` is the URL's host, `path` is its pathname, `params` is derived from `path` and the route filename, and `query` is an instance of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Mutating `page` does not update the current URL; you should instead navigate using [`goto`](#modules-$app-navigation).
```

Rich-Harris:
thanks 👍

PR #2815, opened 2021-11-17
Adapter cloudflare

pngwn:
Adds an adapter for cloudflare pages with functions.

cc @lukeed


lukeed:
Failing tests are unrelated

I ported this adapter over into a test repo and [its deployment was successful](https://github.com/lukeed/pages-fullstack/runs/4243175070). Verified it all works and so this is ready to go IMO

Rich-Harris (approved):

packages/adapter-cloudflare/README.md
```diff
@@ -0,0 +1,59 @@
+# adapter-cloudflare
+
+[Adapter](https://kit.svelte.dev/docs#adapters) for building SvelteKit
+applications on Cloudflare Pages with Workers integration.
+
+_**Comparisons**_
+
```
benmccann, line 7: It might be nice to add a bit more detail here. E.g. I think it was mentioned that this should be the preferred one most of the time and only if you need to do something very advanced you would use `adapter-cloudflare-workers`? And it'd be nice to add that to the `adapter-cloudflare-workers` readme as well

packages/adapter-cloudflare/README.md
```diff
@@ -0,0 +1,59 @@
...
+entryPoints: '< input >',
+outfile: '<output>/_worker.js',
+allowOverwrite: true,
+format: 'esm',
+bundle: true,
+```
+
+## Changelog
+
+[The Changelog for this package is available on
```
benmccann, line 58: should also update the main readme which links to all the various packages and changelogs

packages/adapter-cloudflare/files/worker.js
```diff
@@ -0,0 +1,58 @@
...
+			});
+
+			if (rendered) {
+				return new Response(rendered.body, {
+					status: rendered.status,
+					headers: makeHeaders(rendered.headers)
+				});
+			}
+		} catch (e) {
+			return new Response('Error rendering route:' + (e.message || e.toString()), { status: 500 });
```
benmccann, line 31:
```suggestion
			return new Response('Error rendering route: ' + (e.message || e.toString()), { status: 500 });
```

packages/adapter-cloudflare/README.md
lukeed: We can sunset the Workers one once the "Fullstack Pages" feature is out of this early beta phase. Should be in the next few weeks, which means I'll likely have an update for this by next meeting

lukeed:
@benmccann resolved all comments in the linked PR

Crisfole:
So close to having this available! Is there a process for performing releases? https://github.com/sveltejs/kit/pull/2818

PR #2821, opened 2021-11-17
remove ide mutations

lukeed:
This is antagonistic and serves no real benefit. Our `lint` and `format` commands have the final say in code style anyway, implemented at CI-time. It also encompasses _all_ users & their tooling preferences because there's no way around the check(s)

As it stands, VSCode auto-installing extensions and changing its behavior is very off-putting and borderline antagonistic to contributors. Let people develop using their tools they know and are used to – our only care is that the end result aligns with what we want, and that's what the scripts and CI are for.


.changeset/twenty-books-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+
+---
+
+remove `.vscode` settings/config mutations
```
benmccann, line 5: we don't need a changeset since it doesn't change any code that's distributed

.vscode/settings.json
```diff
@@ -1,5 +0,0 @@
-{
-	"cSpell.language": "en-GB",
```
benmccann, line 2: haha. Rich will love the American spellings :rofl:

benmccann:
I suppose I'm fine with removing the prettier extension since it appears to be buggy.

Do we still want to remove the eslint one and `insertFinalNewline` though? Those still seem helpful

Rich-Harris (approved):

Rich-Harris:
let's try this

PR #2820, opened 2021-11-17
chore: address #2815 feedback

lukeed:
Addresses @benmccann's feedback from #2815

PR #2823, opened 2021-11-17
[docs] update cloudflare adapter readmes

benmccann:
update the readmes without other reformatting

Closes https://github.com/sveltejs/kit/pull/2820


lukeed:
Does not close 2821

PR #2831, opened 2021-11-18
Remove empty package dirs

Rich-Harris:
Fixes a very minor annoyance — if the `config.kit.package.files` option is used to omit some files in `src/lib` from a package, an empty directory will still be created as a byproduct of types being emitted. This fixes that.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/utils/filesystem.js
+++ b/packages/kit/src/utils/filesystem.js
@@ -41,8 +41,11 @@ export function copy(from, to, filter = () => true) {
 	return files;
 }
 
-/** @param {string} cwd */
-export function walk(cwd) {
+/**
+ * @param {string} cwd
+ * @param {boolean} [dirs]
+ */
+export function walk(cwd, dirs = false) {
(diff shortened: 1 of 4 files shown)


ignatiusmb (approved):

packages/kit/src/utils/filesystem.js
```diff
@@ -41,8 +41,11 @@ export function copy(from, to, filter = () => true) {
 	return files;
 }
 
-/** @param {string} cwd */
-export function walk(cwd) {
+/**
+ * @param {string} cwd
+ * @param {boolean} [dirs]
```
dummdidumm, line 46:
```suggestion
 * @param {boolean} dirs
```
Maybe also add a little comment what this param does? In general maybe a little doc on what is returned by this function. Or better yet, adjust the name to better tell what it actually does, like `collect_filenames`

dummdidumm (approved):

PR #2832, opened 2021-11-18
follow redirects when prerendering

Rich-Harris:
When prerendering, if a page or endpoint redirects to another location, the redirect is logged but not followed. In a case I encountered just now, a `/docs` link in the nav bar was followed, but the page consisted of this...

```svelte
<script context="module">
  export function load() {
    return {
      status: 308,
      redirect: '/docs/getting-started'
    };
  }
</script>
```

...and `/docs/getting-started` was never crawled. Easily fixed by ensuring that there's a link to the correct URL, but I would expect Kit to do the right thing here regardless.

Haven't added a test yet as I'm under the cosh

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -1,10 +1,11 @@
 import { readFileSync, writeFileSync } from 'fs';
 import { dirname, join, resolve as resolve_path } from 'path';
-import { pathToFileURL, resolve, URL } from 'url';
+import { pathToFileURL, URL } from 'url';
 import { mkdirp } from '../../utils/filesystem.js';
 import { __fetch_polyfill } from '../../install-fetch.js';
 import { SVELTE_KIT } from '../constants.js';
 import { get_single_valued_header } from '../../utils/http.js';
+import { resolve } from '../../utils/url.js';
 
 /**
  * @typedef {import('types/config').PrerenderErrorHandler} PrerenderErrorHandler
@@ -191,6 +192,11 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 					log.warn(`${rendered.status} ${decoded_path} -> ${location}`);
 					writeFileSync(file, `<meta http-equiv="refresh" content="0;url=${encodeURI(location)}">`);
 					written_files.push(file);
+
+					const resolved = resolve(path, location);
+					if (resolved.startsWith('/')) {
+						await visit(resolved, path);
+					}
 				} else {
 					log.warn(`location header missing on redirect received from ${decoded_path}`);
 				}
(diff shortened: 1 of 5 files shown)


packages/kit/src/core/adapt/prerender.js
```diff
@@ -191,6 +192,11 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 					log.warn(`${rendered.status} ${decoded_path} -> ${location}`);
 					writeFileSync(file, `<meta http-equiv="refresh" content="0;url=${encodeURI(location)}">`);
 					written_files.push(file);
+
+					const resolved = resolve(path, location);
+					if (resolved.startsWith('/')) {
```
benmccann, line 197: we've had bugs with this logic before and it's in a few different places like line 129 of `load_node`. I wonder if we should make a util function for it. I think this will need to be updated to match `load_node` (and I'm pretty sure it's somewhere else too  - maybe elsewhere in prerendering):

```
resolved.startsWith('/') && !resolved.startsWith('//')
```

benmccann:
I noticed one issue, but looks good to me other than that

packages/kit/src/core/adapt/prerender.js
Rich-Harris: good catch, added a util. bonus — it's significantly faster than calling `startsWith`

packages/kit/src/utils/url.js
```diff
@@ -0,0 +1,35 @@
...
+		else baseparts.push(part);
+	}
+
+	const prefix = (path_match && path_match[0]) || (base_match && base_match[0]) || '';
+
+	return `${prefix}${baseparts.join('/')}`;
+}
+
+/** @param {string} path */
+export function is_root_relative(path) {
```
benmccann, line 33: I wonder if `is_absolute_path` might be a clearer name?

packages/kit/src/utils/url.js
```diff
@@ -0,0 +1,35 @@
...
+	return path[0] === '/' && path[1] !== '/';
```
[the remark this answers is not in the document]
Rich-Harris: that would be the correct outcome — `/` _is_ a root-relative path

packages/kit/src/utils/url.js
Rich-Harris: my understanding of the terminology is that `https://example.com/foo` is an absolute path, `//example.com/foo` is a protocol-relative path, and `/foo` is a root-relative path

packages/kit/src/utils/url.js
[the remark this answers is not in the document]
benmccann: ah, it just returns `undefined`. I was worried it might throw an exception

benmccann (approved):

PR #2854, opened 2021-11-21
chore: remove usage of pageXOffset

mquandalle:
Super tiny change : replace `pageXOffset` by `scrollX`. 
`pageXOffset` is marked as a deprecated API, which make VSCode render it as crossed :

image

`pageXOffset` is an alias for `scrollX` so this doesn't change anything except on IE8- 
https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollX

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -2,8 +2,8 @@ import { get_base_uri } from './utils';
 
 function scroll_state() {
 	return {
-		x: pageXOffset,
-		y: pageYOffset
+		x: scrollX,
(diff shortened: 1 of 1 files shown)


bluwy:
Thanks, but `pageXOffset` is specifically used for browser compatibility and I don't think we want to change that first yet. We could revisit this in the future though once we have a plan for a legacy build. Closing this at the meantime.

GauBen:
Hey @mquandalle @bluwy I guess this can be merged now, SvelteKit requires dynamic import() which has a larger support than scrollX

Issue #2844, opened 2021-11-20
Add environment to Cloudflare adapter


piperswe:
### Describe the problem

Cloudflare Workers provide an environment object containing environment variables and bindings to other Cloudflare services, such as Workers KV and Durable Objects. Currently, the Cloudflare adapter doesn't seem pass this object to endpoints, so those endpoints can't access those services.

### Describe the proposed solution

The Cloudflare adapter should pass the environment object through to the endpoint somehow, potentially through the request's locals. In addition, it would be great to have a way to export Durable Object classes.

### Importance

would make my life easier

piperswe:
Related to https://github.com/sveltejs/kit/issues/2807

tv42:
Not being able to access Cloudflare KV makes implementing sessions way too hard :-(

samuel-bach:
@tv42 yeah it is a pity. For now I'm using KV via [API](https://api.cloudflare.com/#workers-kv-namespace-read-key-value-pair) which works but is of course cumbersome

samuba:
@tv42 yeah it is a pity. For now I'm using KV via [API ](https://api.cloudflare.com/#workers-kv-namespace-read-key-value-pair) in my project. It works but is cumbersome.

denizkenan:
I had to fork existing adapter and pass env to request object. It is a pity that, this adapter even exists. Without access to ENV, Durable Objects using workers is almost pointless.

eric-naguras:
We need this, I hope it will be fixed soon.

bryanwood:
> I had to fork existing adapter and pass env to request object. It is a pity that this adapter even exists(it is pre-alpha IMHO). Without providing access to ENV & Durable Objects, using workers is almost pointless.

This no longer works since the commit that added a check for the keys being passed in from the adapter:

https://github.com/sveltejs/kit/blob/f89fb524f851b794cceba4a91864d127dc3948ee/packages/kit/src/core/build/build_server.js#L91-L94

As a hacky workaround I'm adding `env` to the headers object.

In the forked adapter:
```
const rendered = await app.render({
        url,
        rawBody: new Uint8Array(await req.arrayBuffer()),
        headers: Object.fromEntries([...req.headers, ["env", env]]),
        method: req.method,
      });
```

In hooks.js

```
export const handle = async ({ request, resolve }) => {
	request.locals.env = request.headers['env'];
	delete request.headers['env'];
	// etc ...
}
```

[...]

Issue #2807, opened 2021-11-15
Expose cloudflare specific functionality (event.waitUntil) in hooks


tomyan:
### Describe the problem

I'm trying to integrate with Sentry from my SvelteKit app that runs on cloudflare workers. Ideally I'd find a library that works when running in dev (i.e. `npm run dev` running in node) so I can try it out but also works when deployed in cloudflare workers. I don't think this is possible with any of the libraries, so I'm going to try to just get it to work when deployed in workers. I was planning on using toucan-js (https://github.com/punkeel/toucan-js). However this requires that I pass in the "workers fetch event" - the docs say that Toucan has to call the `waitUntil` method (https://developers.cloudflare.com/workers/runtime-apis/fetch-event#waituntil) - I assume so that it can reliably send the telemetry after the response has gone to the user.

I was wondering if there could be a way to expose features that are specific to the platform being run on like this - in my case I'm definitely targeting cloudflare so I'm okay being somewhat coupled to that platform? I wonder if it would make sense to make it possible for an adaptor to allow access to platform specific functionality and let the developer decide whether to couple to a specific platform or not, rather than force the lowest common denominator on everyone.

### Describe the proposed solution

I would like to see an adaptor be able to make available provider specific options to the `handle` hook. I would then also like to see the cloudflare adapter provide this, although I'm using my own out of tree adapter (https://github.com/budgetdraw/sveltekit-cloudflare-adapter) so this is less relevant to me.

### Alternatives considered

There is another sentry library targeting cloudflare workers:

https://github.com/cfworker/cfworker/blob/main/packages/sentry/README.md

This also suggests using `FetchEvent.waitUntil`.

I also noticed that the SvelteKit docs include a Sentry example:

https://kit.svelte.dev/docs#hooks-handleerror

However, I assume this is using the `@sentry/node` package (https://www.npmjs.com/package/@sentry/node), which wouldn't work (I don't think) when deployed to a worker.

### Importance

would make my life easier

tomyan:
I guess this might already be possible by adding extra key(s) to the request object outside the interface described in the documentation. If so perhaps what I'm asking for is an official way to do this

Issue #2606, opened 2021-10-14
Environment variables when using the cloudflare-worker adapter


kevmodrome:
### Describe the problem

On the latest Svelte Summit website I had to use environment variables and ran into some pretty rough DX. Cloudflare workers require environment variables to be supplied either via a [wrangler.toml file](https://developers.cloudflare.com/workers/platform/environment-variables) or manually added using the wrangler CLI or  the web interface. These environment variables are then available as top level variables in the script. 

This presents a bit of a problem when developing applications as these variables are not available. To get around it you need to use some other way of storing environment variables. I ended up using a package called `env-cmd` to run dev/build and preview with. However, since you grab the variables from different places (process.env.X vs X) you end up having to write conditionals in your code to handle this.

Here's [one](https://github.com/svelteschool/summit-fall-2021/blob/main/src/routes/subscribe.ts#L14) I wrote for an email API key:
```js
api_key: process.env['EMAIL_API_KEY'] || EMAIL_API_KEY
```
I understand I can probably write some wrapping function around this, but I still need to define the variables in two places.

### Describe the proposed solution

Ideally you would only need to keep your variables in one place and having a way to reach the variables in a consistent way would be great.

### Alternatives considered

Using another adapter.

### Importance

would make my life easier

### Additional Information

Maybe there needs to be a bigger discussion around environment variables in general here. I'm really only familiar with CF adapter but I imagine there are probably other deploy targets that also handle variables differently.

antony:
I actually have a similar issue with Vercel. I've not tried SvelteKit for comparison, but to deploy our Sapper apps I have multiple deployment configurations. It's fairly annoying. I wonder if we could do something clever with the toml/vercel config during deploy, like pulling in the required variables from `.env` and putting them into these configs (we already write some custom config for begin, for example).

Or maybe it's not for SvelteKit to solve. It's hard to tell!

Rich-Harris:
i'm going to close this in favour of #4296

PR #2855, opened 2021-11-22
Adapter-native Request and Response (enables full Cloudflare Worker functionality)

skirsten:

### Tests

### Changesets

---

Hi, this is my weekend project that allows to directly interact with the Request and Response objects (and more) used by the adapter (if implemented).
This makes most sense for the two Cloudflare adapters, as it exposes the full CF Worker functionality:
- Env vars
- KV Namespaces
- waitUntil
- the Request.cf and Response.cf property
- The ability to proxy requests through without buffering
- *basically anything a normal worker has access to*

Here is a example that tests some of the Cloudflare specific functionality exposed by the adapter:
```ts
import type { AdapterRequestHandler } from '@sveltejs/adapter-cloudflare';

export const get: AdapterRequestHandler = async (request) => {
  if (request.adapter) {
    const colo = request.adapter.request.cf?.colo;

    request.adapter.ctx.waitUntil(
      (async () => {
        // Do stuff here
        return;
      })()
    );

    return {
      adapter: {
        // No buffering in memory happens here! Its passed straight through.
        response: fetch(
          `https://fakeimg.pl/440x230/282828/eae0d0/?retina=1&text=${encodeURIComponent(
            `proxied by CF in ${colo}`
          )}`,
          { cf: { cacheEverything: true } }
        )
      }
    };
  }
  return {
    status: 302,
    headers: {
      location: `https://fakeimg.pl/440x230/282828/eae0d0/?retina=1&text=${encodeURIComponent(
        'fallback to redirect'
      )}`
    }
  };
};
```

To test it, I build a example project https://github.com/skirsten/kit/tree/master, deployed here https://svelte-kit-cf-adapter-demo.pages.dev/.

It should be 100% backwards compatible and there is no need to use this new feature in your project.
This closes a number of issues:
- https://github.com/sveltejs/kit/issues/2844
- https://github.com/sveltejs/kit/issues/2807
- https://github.com/sveltejs/kit/issues/2606
- *and probably more*

Unfortunately I don't have the time to finish this, so this is a draft. If the approach taken here is okay with the maintainers, I would be glad if somebody can take this over and finish it.

### Missing
- [ ] Mismatch between Cloudflare Worker types and other types (Request, fetch, etc.). I am stuck here, I have no idea how to force it to use the Cloudflare Worker types...
- [ ] Only tested `adapter-cloudflare` not the `adapter-cloudflare-workers` and not `adapter-netlify` (was only a proof of concept to keep it generic)
- [ ] Update docs
- [ ] Test (automated?) all the edge-case interactions e.g. pre-rendered, handle hook, etc.
- [ ] Fix one failing CI. Seems unrelated.

### DIFF
--- a/examples/hn.svelte.dev/src/routes/__error.svelte
+++ b/examples/hn.svelte.dev/src/routes/__error.svelte
@@ -1,4 +1,5 @@
 <script context="module">
+	/** @type {import('@sveltejs/kit').ErrorLoad} */
 	export function load({ error, status }) {
 		return {
 			props: { error, status }
--- a/examples/hn.svelte.dev/src/routes/index.svelte
+++ b/examples/hn.svelte.dev/src/routes/index.svelte
@@ -1,4 +1,5 @@
 <script context="module">
+	/** @type {import('@sveltejs/kit').Load} */
 	export function load({ page }) {
 		let host = page.host;
 		const i = host.indexOf(':');
--- a/examples/hn.svelte.dev/src/routes/rss.js
+++ b/examples/hn.svelte.dev/src/routes/rss.js
@@ -1,6 +1,7 @@
-import {dev} from '$app/env';
+import { dev } from '$app/env';
+
 /**
- * @type {import('@sveltejs/kit').RequestHandler}
+ * @type {import('@sveltejs/adapter-netlify').AdapterRequestHandler}
  */
 export function get() {
 	return {
--- a/examples/hn.svelte.dev/src/routes/rss.js
+++ b/examples/hn.svelte.dev/src/routes/rss.js
@@ -1,7 +1,7 @@
 import { dev } from '$app/env';
 
 /**
- * @type {import('@sveltejs/adapter-netlify').AdapterRequestHandler}
+ * @type {import('@sveltejs/kit').RequestHandler}
  */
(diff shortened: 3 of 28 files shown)


examples/hn.svelte.dev/src/routes/__error.svelte
```diff
@@ -1,4 +1,5 @@
 <script context="module">
+	/** @type {import('@sveltejs/kit').ErrorLoad} */
```
benmccann, line 2: this is kind of a big PR. what do you think about sending the changes to `hn.svelte.dev` in a separate PR to break things up a bit? we could merge that change more quickly

Rich-Harris:
As of #3384, the platform `Request` object is passed to your SvelteKit app, which IIUC enables what you need. The wrinkle, of course, is that you can't use the extra bits during development — we may in future need to expose a way for adapters to augment the request.

Anyway I'll close this — if there's more stuff that needs to happen around this, by all means raise a new issue. Thanks!

tylerbrostrom:
@Rich-Harris Heads up: #3384 doesn’t (fully) resolve the issues that would be resolved by this PR.

#3426 lays out how those issues might be resolved, taking #3384 into account.

PR #2861, opened 2021-11-22
[fix] Server-side fetch request should have headers (#969)

aolose:
Server-side fetch  in load function will include original request headers.

I'm only test it in the Page file , if server-side fetch not in load , headers will lost :(  ,   

### Tests

### Changesets


benmccann:
please fix the failing lint job

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -101,6 +101,12 @@ export async function load_node({
 					};
 				}
 
+				//merge  headers  from  request
```
benmccann, line 104: we usually put a space after `//`
```suggestion
				// merge  headers  from  request
```

bluwy: There are two spaces between words too, we should have one space only.

benmccann:
@aolose this PR would need to be rebased and my comment would need to be addressed

aolose:
I deleted the repo I forked earlier, so I fork it again.
The new pr is here: #3631.

Issue #2805, opened 2021-11-15
adapter-auto


Rich-Harris:
### Describe the problem

I've seen feedback from Vercel customers that SvelteKit 'doesn't work', because it wasn't clear to them that they needed to add an adapter. Even though it's documented reasonably well, it's definitely a source of friction. It would shock me if people using Kit with other platforms weren't experiencing the same thing.

### Describe the proposed solution

There's probably a lot of value in having Kit auto-detect common platforms and build itself accordingly. (Nuxt does this, for example.) To that end I propose something like this in the default template:

```js
import adapter from '@sveltejs/adapter-auto';

export default {
  kit: {
    adapter: adapter()
  }
};
```

TBD which platforms would be supported here, but the adapters tend to be fairly small and have overlapping dependencies, so we can probably be generous.

### Alternatives considered

Builds could fail noisily if an adapter isn't configured. This would help with diagnosis, but it's still friction.

### Importance

would make my life easier

bluwy:
Would #2479 help with this? I think it's a good start before we go with something automatic.

antony:
does it actually require an adapter module at all? couldn't the default behaviour just be to find and install the relevant adapter (and print a warning that you should probably configure your project correctly?)

TheOnlyTails:
I imagine this being integrated into the SvelteKit package itself, so users won't have to wonder what a seemingly unimportant package is doing in their project.

However, I do think it should still be required for the `adapter` field, so it won't be completely magical, and could even lead to discovery by users wondering "what's this doing here? better look it up!"

Conduitry:
@antony Going out and downloading an adapter on the fly was the behavior Rich originally suggested in the most recent Svelte maintainers call, which seemed to me to be a slightly gross mixing of concerns. I had then suggested making this be its own adapter instead, so the logic for picking an adapter can live solely in an adapter, rather than in SvelteKit core.

Rich-Harris:
#2479 probably _would_ help, but it has downsides of its own — for people who just want to try Kit out (and have no intention of deploying their experiment, or don't yet know what it should be) it's a premature decision. It also adds complexity to the project scaffolding itself (overwriting a magic placeholder, etc(.

> I had then suggested making this be its own adapter instead, so the logic for picking an adapter can live solely in an adapter, rather than in SvelteKit core.

exactly this — good to keep it separate, I think. And having it be 'just another adapter' makes it easier for people to figure out where and how to change it, if they have more niche requirements

[...]

PR #2867, opened 2021-11-22
first crack at adapter-auto

Rich-Harris:
#2805. For now this just adds adapter-vercel and adapter-netlify so that we can quickly try it out (difficult to test without publishing the package), but if it works we can immediately add other adapters (though I think for now it's really just Cloudflare Pages? not sure if it makes sense to try and detect that we should use adapter-node or adapter-static)

### Tests

### Changesets


packages/adapter-auto/package.json
```diff
@@ -0,0 +1,32 @@
...
+	"files": [
+		"files",
+		"index.d.ts"
+	],
+	"scripts": {
+		"lint": "eslint --ignore-path .gitignore \"**/*.{ts,js,svelte}\" && npm run check-format",
+		"format": "npm run check-format -- --write",
+		"check-format": "prettier --check . --config ../../.prettierrc --ignore-path .gitignore"
+	},
+	"dependencies": {
```
sjroe, line 28: Missing `"@sveltejs/adapter-cloudflare": "workspace:*"`

Rich-Harris: thanks 👍

Rich-Harris:
One alternative we might consider is keeping the `test()` logic but then only importing the underlying adapter if it's installed, otherwise prompting the user to install it. This eliminates the risk of `adapter-auto` growing crufty over time, but does introduce some additional friction.

An in-between option would be to install it automatically during the build (e.g. if `process.env.VERCEL` then we automatically install `adapter-vercel`). This adds some complexity though (need to consider different package managers etc), and probably isn't as fast as having it installed initially.

dominikg (approved):

documentation/docs/10-adapters.md
```diff
@@ -2,45 +2,98 @@
...
-- [`adapter-cloudflare`](https://github.com/sveltejs/kit/tree/master/packages/adapter-cloudflare) — for [Cloudflare Pages](https://developers.cloudflare.com/pages/)
-- [`adapter-cloudflare-workers`](https://github.com/sveltejs/kit/tree/master/packages/adapter-cloudflare-workers) — for [Cloudflare Workers](https://developers.cloudflare.com/workers/)
-- [`adapter-netlify`](https://github.com/sveltejs/kit/tree/master/packages/adapter-netlify) — for [Netlify](https://netlify.com)
-- [`adapter-vercel`](https://github.com/sveltejs/kit/tree/master/packages/adapter-vercel) — for [Vercel](https://vercel.com)
+Within the `adapt` method, there are a number of things that an adapter should do:
 
-...and traditional platforms:
+- Clear out the build directory
+- Output code that:
+  - Calls `init`
```
Conduitry, line 90: Can the `init` being referred to here be clarified?

Similarly, can it be clarified where the `render` that's supposed to be called below comes from?

Or do you think that the current instructions will suffice if someone is simultaneously following along reading an existing adapter?

documentation/docs/10-adapters.md
```diff
@@ -2,45 +2,98 @@
...
+  - Converts from the platform's request to a [SvelteKit request](#hooks-handle), calls `render`, and converts from a [SvelteKit response](#hooks-handle) to the platform's
+  - Globally shims `fetch` to work on the target platform. SvelteKit provides a `@sveltejs/kit/install-fetch` helper for platforms that can use `node-fetch`
+- Bundle the output to avoid needing to install dependencies on the target platform, if desired
+- Call `utils.prerender`
+- Put the user's static files and the generated JS/CSS in the correct location for the target platform
 
-- [`adapter-node`](https://github.com/sveltejs/kit/tree/master/packages/adapter-node) — for creating self-contained Node apps
-- [`adapter-static`](https://github.com/sveltejs/kit/tree/master/packages/adapter-static) — for prerendering your entire site as a collection of static files
+If possible, we recommend putting the adapter output under the `build/` directory with any intermediate output placed under `'.svelte-kit/' + adapterName`.
```
Conduitry, line 97: I'd probably write this as `.svelte-kit/[adapter-name]` or something similar, rather than as a string operation.

documentation/docs/10-adapters.md
Rich-Harris: yeah these docs were not all that complete. fleshing them out now

Rich-Harris:
The adapter API could definitely use far more comprehensive documentation but a) probably best it doesn't derail this PR (since it's really just moving stuff from one file into another) and b) it should probably wait until the API is more final. i addressed the points above though

Conduitry (approved):

PR #2882, opened 2021-11-23
Add package command

Rich-Harris:
If component libraries are to be a first-class use case for Kit, it stands to reason that `npm run package` should be available in new projects alongside `npm run build`.

### Tests

### Changesets

### DIFF
--- a/.changeset/sour-taxis-laugh.md
+++ b/.changeset/sour-taxis-laugh.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Add an npm run package command to templates
--- a/packages/create-svelte/templates/default/package.template.json
+++ b/packages/create-svelte/templates/default/package.template.json
@@ -4,6 +4,7 @@
 	"scripts": {
 		"dev": "svelte-kit dev",
 		"build": "svelte-kit build",
+		"package": "svelte-kit package",
 		"preview": "svelte-kit preview"
 	},
 	"devDependencies": {
--- a/packages/create-svelte/templates/skeleton/package.template.json
+++ b/packages/create-svelte/templates/skeleton/package.template.json
@@ -4,6 +4,7 @@
 	"scripts": {
 		"dev": "svelte-kit dev",
 		"build": "svelte-kit build",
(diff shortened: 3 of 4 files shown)


benmccann:
Or should it be part of `svelte-create`? An `npm run package` command wouldn't make as much sense if you're building a web app instead of a library. I wonder if it would end up confusing people because they think they're packaging their app for production deployment or something?

Conduitry:
I agree about worrying that this might be confusing for people who are just writing an app and not a reusable library.

I'm not sure what @benmccann's suggesting though. That this become one of the prompts in the `create-svelte` CLI? Something else?

dummdidumm:
Also in favor of a prompt "do you want to create a library or an app?"

Rich-Harris:
I think `npm run build` is a widely-enough used convention that if someone opts to `npm run package` instead it's a definite case of PEBKAC. The docs (https://kit.svelte.dev/docs#command-line-interface) are fairly clear in any case.

Disagree that a prompt is appropriate — the whole point of this feature is that apps and libraries are the same thing (an app has an internal library, a library has a demo/docs app, the only difference is which bit is user-facing). If the only thing the prompt changes is whether or not `"package": "svelte-kit package"` gets added to the package.json, that's a pretty meagre payoff for adding another (potentially confusing) decision at project init time. 

I think the solution here is something I've been meaning to get round to anyway — add some language when `svelte-kit package` completes along the following lines:

> Successfully built library: src/lib -> package
> To publish your package:
>  $ cd package
>  $ npm publish

Conduitry:
:+1: Adding that message when the `package` command completes would address my concerns in a tidy enough way, and does seem better than adding another prompt to the CLI.

Rich-Harris:
Alright, added that. I really think we need to revisit this —

> Cannot generate a "svelte" entry point because the "." entry in "exports" is missing. Please specify one or set a "svelte" entry point yourself

— it's very confusing and shouldn't be necessary. Will open a separate issue though as it will probably mean coordinating with the plugins

Conduitry (approved):

Issue #2865, opened 2021-11-22
Add a way to `keepfocus` when using the `invalidate` API


pzuraq:
### Describe the problem

I may have missed it, but it doesn't seem like there is a way currently to use the `invalidate` API and keep focus on the currently active element. We oftentimes use the `invalidate` API in order to poll our data for updates, and as this is essentially a background task it shouldn't really interrupt the user's current action (e.g. typing in a form field). Currently we're working around this by refocusing the element after all `invalidate` promises have resolved, but this feels like a bit of a hack.

### Describe the proposed solution

Allow options to be passed to `invalidate` as the second parameter, with `keepfocus` being the first supported option:

```js
invalidate('api.example.com', { keepfocus: true });
```

### Importance

would make my life easier

### Additional Information

Multiple `invalidate`s can run for a single transition, so it's kind of hard to know what to do if the options for each invalidate conflict.

bluwy:
Does this only happen recently? I think #2819 might've accidentally caused this.

@benmccann  Come to think of it, maybe the initial way was intended after all, I don't think we would want to handle `hash, scroll, keepfocus` for `invalidate`. Should we revert that and document it?

PR #2883, opened 2021-11-23
[fix] revert #2819 and add code comment

benmccann:
reverts https://github.com/sveltejs/kit/pull/2819, which was a bad idea. fixes https://github.com/sveltejs/kit/issues/2865 and adds a code comment to make it clearer

### DIFF
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -224,16 +224,16 @@ export class Renderer {
 			});
 		}
 
-		await this.update(info, chain, no_cache, nav_opts);
+		await this.update(info, chain, no_cache, opts);
 	}
 
 	/**
 	 * @param {import('./types').NavigationInfo} info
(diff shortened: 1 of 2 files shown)


packages/kit/src/runtime/client/renderer.js
```diff
@@ -275,43 +275,49 @@ export class Renderer {
 			this._init(navigation_result);
 		}
 
-		const { hash, scroll, keepfocus } = opts || {};
+		// opts must be passed if we're navigating
+		// they will not be supplied if we're simply invalidating
+		if (!opts) {
+			await 0;
+		} else {
+			const { hash, scroll, keepfocus } = opts || {};
```
bluwy, line 283:
```suggestion
			const { hash, scroll, keepfocus } = opts;
```
nit: We can remove the opts fallback

bluwy (approved):
Thanks for handling this!

PR #2905, opened 2021-11-24
Update .gitignore files in create-svelte package to ignore all dotfiles with a prefix of .env

### LINKED ISSUE #2904: Ignore all dotfiles with a .env prefix by default for new projects
### Describe the problem

When a developer runs `npm init svelte@next` to create a new app and selects either the demo or skeleton project, the codebase generated for them contains a `.gitignore` file that will ignore a file that is named `.env`, but will not ignore env files with other names, such as the commonly used `.env.local`.

This makes it easy for developers to add their environment variables in a `.env.local` file, work on their project, then accidentally commit all the files for their project to git without realizing that their `.env.local` (and the potentially sensitive API keys therein) was included in the commit.

Add a `.env.*` line to the `.gitignore` files of the create-svelte package templates. This way, all files with a ".env." prefix will be git ignored by default, including `.env.local`.

### Alternatives considered

n/a

### Importance

nice to have

### Additional Information

n/a

kellenmace:
Fixes #2904.

### Tests

### Changesets

### DIFF
--- a/.changeset/pretty-snakes-burn.md
+++ b/.changeset/pretty-snakes-burn.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Git ignore files with a .env. prefix
--- a/packages/create-svelte/templates/default/.gitignore
+++ b/packages/create-svelte/templates/default/.gitignore
@@ -3,6 +3,6 @@ node_modules
 /build
 /.svelte-kit
 /package
-.env
+.env*
 .vercel
 .output
--- a/packages/create-svelte/templates/default/.gitignore
+++ b/packages/create-svelte/templates/default/.gitignore
@@ -3,6 +3,7 @@ node_modules
 /build
 /.svelte-kit
 /package
-.env*
+.env
+.env.*
 .vercel
 .output
--- a/packages/create-svelte/templates/skeleton/.gitignore
+++ b/packages/create-svelte/templates/skeleton/.gitignore
@@ -3,4 +3,5 @@ node_modules
 /build
 /.svelte-kit
 /package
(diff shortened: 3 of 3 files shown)


Conduitry:
IMO this should only add `.env.*`, if anything, and not `.env*`.

bjon:
Shouldn't only env files ending with _.local_ be ignored?
We're using vite's support for modes, e.g. `.env.[stage|prod]`

Rich-Harris:
Thank you! Aside from the feedback above, can you reinstate the [deleted checklist](https://github.com/sveltejs/kit/blob/master/.github/PULL_REQUEST_TEMPLATE.md) and add a changeset?

kellenmace:
Thanks for the feedback, all! 🙌

@Conduitry - Sounds good. I updated the PR so that `.env` and `.env.*` files are git ignored. This way, on the off chance that the dev has some other file that begins with ".env" in their project, like ".envision", it won't be affected.

@bjon - Files like `.env.local`, `.env.stage` and `.env.prod` should generally not be added to version control. They often contain sensitive data like API keys. Devs should set their environment variables in their hosting provider's UI for their staging and prod sites. If a developer has a reason to want those included in git (which I wouldn't recommend), they can just tweak the `.gitignore` file to accomplish that. Next.js, for example, excludes those files in their `.gitignore` file for new projects, as well: https://github.com/vercel/next.js/blob/canary/examples/active-class-name/.gitignore#L27-L31

@Rich-Harris - Yes, checklist reinstated and changeset added. Please let me know if anything else is needed. 👍🏼

antony (approved):

Issue #2262, opened 2021-08-21
Allow page.query to be accessed in load function even when prerender is true


mattjennings:
### Describe the problem

#2104 introduced throwing an error when page.query was accessed. This makes sense to prevent usage on the client, but now you cannot use it during `load` which was beneficial for prerendering crawled pages. For example, a page displaying a list of blog posts with pagination using a `?page=X` query. As long as those can be crawled, using query is OK here.

### Describe the proposed solution

Check if page.query is being accessed on the server before throwing the error

https://github.com/sveltejs/kit/pull/2104/files#diff-7266c49c71803f38dda70345d857f13a7a2494cc649b70af6b689627199f5b79R53-R63

### Alternatives considered

I could use params instead, which is probably better in most scenarios, but in my case I'm using it on my index page and it's more awkward than using a query. I would prefer `mywebsite.com/?page=2` over `mywebsite.com/2`.

### Importance

would make my life easier

### Additional Information

Here's an example of a template I made that utilized crawled pages with queries (you'll have to set `const PAGE_SIZE = 1` in `routes/index.svelte` to get more than 1 page):

https://github.com/mattjennings/sveltekit-blog-template

The build will work with kit 1.0.0-next.144, but break on 145

benmccann:
Even on 144 your template was broken in some ways - it was just hidden to you. E.g. if you build the site and visit `/?page=2` directly it will still load page 1. That's not what I'd expect as a user. This check pointed out that what you're doing is not valid

Prerendering works by saving a static file with the name of the page. In your case `index.html`. But we can't save an `index.html` and `index.html?page=2`. `?` is an invalid character on at least some file systems and your web server wouldn't know to serve a different file based on the query string.

If you really want to use `?page=2` then be aware that it's something that only works on the client-side and use `location` in `onMount` instead.

mattjennings:
I hadn't realized it only worked on client-side navigation (and obviously that's not what I'd expect as a user either). I appreciate the explanation, as I wasn't sure how it "worked" in the first place, but that makes sense.

PR #2910, opened 2021-11-25
[docs] Clarify usage of `page.query`

adriandelgado:
This PR clarifies the usage of `page.query`. Currently if you use it in a prerendered page it throws an error unless you use it in a browser/client context only. These behavior is undocumented and people have asked related questions before #2262 #669 

### Tests

### Changesets

### DIFF
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -92,7 +92,7 @@ The `load` function receives an object containing four fields — `page`, `fetch
 
 #### page
 
-`page` is a `{ host, path, params, query }` object where `host` is the URL's host, `path` is its pathname, `params` is derived from `path` and the route filename, and `query` is an instance of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Mutating `page` does not update the current URL; you should instead navigate using [`goto`](#modules-$app-navigation).
+`page` is a `{ host, path, params, query }` object where `host` is the URL's host, `path` is its pathname, `params` is derived from `path` and the route filename, and `query` is an instance of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Using `query` on the server makes the current page not [prerenderable](#ssr-and-javascript-prerender). Mutating `page` does not update the current URL; you should instead navigate using [`goto`](#modules-$app-navigation).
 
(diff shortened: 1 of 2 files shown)


documentation/docs/03-loading.md
```diff
@@ -92,7 +92,7 @@ The `load` function receives an object containing four fields — `page`, `fetch
 
 #### page
 
-`page` is a `{ host, path, params, query }` object where `host` is the URL's host, `path` is its pathname, `params` is derived from `path` and the route filename, and `query` is an instance of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Mutating `page` does not update the current URL; you should instead navigate using [`goto`](#modules-$app-navigation).
+`page` is a `{ host, path, params, query }` object where `host` is the URL's host, `path` is its pathname, `params` is derived from `path` and the route filename, and `query` is an instance of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Using `query` on the server makes the current page not [prerenderable](#ssr-and-javascript-prerender). Mutating `page` does not update the current URL; you should instead navigate using [`goto`](#modules-$app-navigation).
```
Rich-Harris, line 95: i think we only need to mention it in one place, so we can probably revert this change

```suggestion
`page` is a `{ host, path, params, query }` object where `host` is the URL's host, `path` is its pathname, `params` is derived from `path` and the route filename, and `query` is an instance of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Mutating `page` does not update the current URL; you should instead navigate using [`goto`](#modules-$app-navigation).
```

documentation/docs/11-ssr-and-javascript.md
```diff
@@ -72,6 +72,8 @@ The basic rule is this: for a page to be prerenderable, any two users hitting it
 
 Note that you can still prerender pages that load data based on the page's parameters, like our `src/routes/blog/[slug].svelte` example from earlier. The prerenderer will intercept requests made inside `load`, so the data served from `src/routes/blog/[slug].json.js` will also be captured.
 
+> Using [`page.query`](#loading-input-page) on a server context will cause the current page to not be prerenderable. If you want to use `page.query` and keep the page prerenderable, you should only use it in a [client context](#modules-$app-env)
```
Rich-Harris, line 75:
```suggestion
Accessing [`page.query`](#loading-input-page) during prerendering is forbidden. If you need to use it, ensure you are only doing so in the browser (for example in `onMount`).
```

Rich-Harris (approved):

Rich-Harris:
thank you!

Issue #2870, opened 2021-11-23
Router changes the URL prematurely


ebeloded:
### Describe the bug

When using links within the SvelteKit app, the URL gets updated immediately, even though the underlying page is still loading. In this simple demonstration the second-page loads for 1s, but the URL is updated a immediately:

![CleanShot 2021-11-23 at 14 09 45](https://user-images.githubusercontent.com/5859177/142983691-54a6d6ce-9274-4860-a64f-7555625a04f4.gif)

This is unconventional and results in asynchrony between the URL and the page content.

For contrast, here is what happens when clicking on the second link with rel="external", which triggers normal browser behavior: 

![CleanShot 2021-11-23 at 14 10 35](https://user-images.githubusercontent.com/5859177/142984018-569e3747-4127-427d-96d1-9a41b58f9224.gif)

Here you can see that both the URL and the page change at the same time. The page represents the URL at every moment. This behavior and universal across browsers.

SvelteKit mimicks browser behavior with History API. To do it correctly, the URL needs to be changed after the page is loaded.

### Reproduction

[Stackblitz](https://stackblitz.com/edit/sveltekit-aajlpj?file=src/routes/index.svelte)
[Source Code](https://github.com/ebeloded/sveltekit-issue-2870/tree/master)

### System Info

```shell
-
```

### Severity

annoyance

harshmandan:
Can confirm it happens with me as well.

Rich-Harris:
Explanation of why the current behaviour is intentional: https://github.com/sveltejs/kit/pull/2917#issuecomment-979419299

ebeloded:
Given the split in opinions regarding the URL change behavior, could this be added as a configuration option?

rmunn:
> Explanation of why the current behaviour is intentional: [#2917 (comment)](https://github.com/sveltejs/kit/pull/2917#issuecomment-979419299)

But since the browser's built-in behavior is the opposite (see second GIF in https://github.com/sveltejs/kit/issues/2870#issue-1060911996), IMHO it should be the other way around, with the URL not updating until after `load()` completes. There's also the fact that the current behavior is causing #2945, and may be causing other issues as well.

benmccann:
I wonder, if we did want to mirror the browser behavior, if the most analogous place to update the url would be in between the network fetch and dom rendering, which I think would be here:

[...]

PR #2917, opened 2021-11-25
fix: router changes the URL prematurely

si3nloong:

### Tests

### Changesets


si3nloong:
Currently, the client-side navigation doesn't comply with the server-side navigation standard. We should change the URL path after the component or page is rendered.

This is a fix for issue #2870, but I think I need to include more test cases.

Rich-Harris:
Thank you, but the current behaviour is intentional, and changing it would require a thorough discussion. 

There's no universally accepted 'right' way to address this. [This poll](https://twitter.com/simonw/status/1434940499576561666) gives a slight edge to the behaviour in this PR...

Screen Shot 2021-11-25 at 1 35 16 PM

...but the following thread gives some very convincing reasons why changing the URL at the start of the navigation makes more sense:

Screen Shot 2021-11-25 at 1 36 38 PM
Screen Shot 2021-11-25 at 1 36 45 PM
Screen Shot 2021-11-25 at 1 36 59 PM
Screen Shot 2021-11-25 at 1 37 07 PM

Personally I think delaying the change is worse UX — it robs the user of valuable feedback that their click was registered, and makes error recovery harder since they can't simply refresh the page if client-side navigation fails for some reason. I'm also persuaded by the argument that state should follow URL, not vice versa.

This position is supported by the behaviour of widely-used SPAs like YouTube and Instagram. Delaying the update until after a successful navigation would therefore violate the expectations that users have developed through using these sites.

Rich-Harris:
Closing in favour of #3241 which has a simpler implementation, in case we were to decide to go ahead with this change

Issue #2927, opened 2021-11-26
Prevent user misuse the `page.query`


si3nloong:
### Describe the problem

There are plenty of people misunderstand the usage (#969, #2785 , #2811) of `import('$app/stores').page.query`.

### Describe the proposed solution

Instead of just updating the documentation, I suggest we update the type of `query` as below to prevent them to mutate the query value.
<img width="760" alt="Screenshot 2021-11-26 at 9 07 25 PM" src="https://user-images.githubusercontent.com/28108597/143586342-13cecc16-3387-42dc-ac48-d6c59e4df782.png">

Nevertheless, the user still can cast the type as `URLSearchParams` to set the value if they want to.
<img width="773" alt="Screenshot 2021-11-26 at 9 06 52 PM" src="https://user-images.githubusercontent.com/28108597/143586461-e121b5c4-fe9d-4754-81e7-e10e00336f3c.png">

@Rich-Harris @benmccann what do you guys think?

### Importance

would make my life easier

### Additional Information

PR is included #2928

PR #2928, opened 2021-11-26
[chore] update `URLSearchParams` data type to `ReadonlyURLSearchParams`

si3nloong:

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -745,7 +745,7 @@ export class Renderer {
 	 *   status?: number;
 	 *   error: Error;
 	 *   path: string;
-	 *   query: URLSearchParams
+	 *   query: import('types/page').ReadonlyURLSearchParams
 	 * }} opts
 	 */
 	async _load_error({ status, error, path, query }) {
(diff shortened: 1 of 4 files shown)


ignatiusmb (commented):
If this is supposed to fix #2927, then it's better to actually implement a RO class just like `ReadOnlyFormData`. Restricting the usage of certain methods through types aren't enough and misleading to say the least, especially when the actual data is still an instance of `URLSearchParams` living standard. Also, it won't have any effect for users that doesn't do any type checking, so this would only cause unnecessary errors for TS (and `checkJs`) users.

Rich-Harris:
> then it's better to actually implement a RO class just like `ReadOnlyFormData`. Restricting the usage of certain methods through types aren't enough and misleading to say the least

Very much this, except that we _don't_ want to implement this since it would add bulk to client-side JS for very limited upside. Mutating the URL is _probably_ not going to do anything good, but it's not the sort of thing you'd do by accident. If someone is doing it they probably have a good reason.

In any case this PR is out of date since there's no longer such a thing as `request.query` or `page.query`, so it would need to be `ReadonlyURL` rather than `ReadonlyURLSearchParams`. Closing — thank you

Issue #2581, opened 2021-10-10
ESBuild define option can break adapter-node assetsMiddleware


arnivuo:
### Describe the bug

I switched from another adapter to use node adapter and had **esbuild** options with `define`. That basically overwrites `esbuild_app_dir` defined by the adapter breaking assetsMiddleware. A symptom is rather odd as all assets are responding to status code 500 with no clue why.

Maybe the adapter should warn the user if they overwrite the `esbuild_app_dir` property from `define` or the documentation could mention it. Now you can figure it out but it's not that obvious.

### Reproduction

In svelte.config.js configure node adapter like:
```javascript
adapterNode({
  esbuild(defaultOptions) {
    return {
      ...defaultOptions,
      define: {
        foo: 'bar'
      }
    };
  }
});
```
Build, run `node build` and check broken http://localhost:3000 site.

Fix is to do configuration e.g.:
```javascript
adapterNode({
  esbuild(defaultOptions) {
    defaultOptions.define = {
      ...defaultOptions.define,
      foo: 'bar'
    }
    return defaultOptions;
  }
});
```

### System Info

```shell
System:
    OS: macOS 11.5.2
    CPU: (4) x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
    Memory: 63.46 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.1 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.12 - /usr/local/bin/npm
  Browsers:
    Chrome: 94.0.4606.71
    Safari: 14.1.2
  npmPackages:
    @sveltejs/adapter-node: ^1.0.0-next.53 => 1.0.0-next.53
    @sveltejs/adapter-static: ^1.0.0-next.13 => 1.0.0-next.20
    @sveltejs/kit: 1.0.0-next.180 => 1.0.0-next.180
    svelte: ^3.34.0 => 3.43.1
```

### Severity

annoyance

PR #2931, opened 2021-11-26
Overhaul adapter API

### LINKED ISSUE #1197: Exclude prerendered pages from serverless functions/workers
There's no sense in shipping the code for routes that got prerendered; they will never be called on during SSR. We can't guarantee that of routes with dynamic parameters (we might have prerendered `/blog/one` and `/blog/two` but we might need `/blog/[slug].svelte` in order to render `/blog/three` in future), but we can for other cases. It's a small win but worth having.

I have some half-baked ideas for what the API could look like but I should probably get back to writing my Svelte Summit talk before I explore all the ways in which those ideas are bad

It's a nice-to-have. The current situation is wasteful, but not egregiously so

Rich-Harris:
At present, `svelte-kit build` generates a monolithic `app.js`, via Vite. Adapters interact with it by importing the `init` and `render` functions therein and using them in the context of a platform-specific function — `adapter-node` calls `render` inside a Polka handler, `adapter-netlify` calls it inside a Lambda event handler, and so on.

This doesn't scale. Lambdas have a size limit of 50MB; Cloudflare Workers have just 1MB. Many apps won't reach these limits but many will. We need to be able to split apps into smaller chunks. Doing so has additional benefits:

* startup is faster, in the case where code is bundled into a single script (e.g. workers)
* routing should be faster, if a function only bundles a subset of the complete routing table, since we no longer need to evaluate _n_ regexes inside SvelteKit to find the ones that match

This PR implements that idea, by updating the adapter API with a method to output multiple functions.

* [x] update adapter API to generate multiple functions
  * [ ] ~~need to be able to return an array of entries, because of how Architect handles catch-alls (see below)~~
* [x] add a way to generate a manifest with all routes, for adapter-node etc
* [x] prevent functions being generated for prerendered pages (fixes #1197)
* [x] update adapter-node
* [ ] update and re-enable adapter-node tests
* [x] update adapter-vercel (this is partially blocked by some upcoming work on the filesystem API)
  * [ ] allow configuration of externals, so that lambdas can require native deps like `sharp`
* [x] update adapter-netlify
* [x] update adapter-cloudflare
  * [x] expose prerendered pages
  * [x] rethink how `ASSETS` is passed in; it's currently duplicative with the app manifest
* [x] rewrite adapter API documentation
* [x] remove sandbox from `pnpm-workspace.yaml` and lockfile
* [x] update adapter READMEs
* [x] add a utility for creating dirs inside `.svelte-kit` — hardcoding that directory feels super brittle
* [x] rename `utils` to `builder`, which still isn't a great name but is definitely better

It would also be nice to reach out to adapter authors who will be affected by these breaking changes.

Closes https://github.com/sveltejs/kit/pull/2580 as well as the issues mentioned in that PR. I.e. closes https://github.com/sveltejs/kit/issues/2517, closes https://github.com/sveltejs/kit/issues/2581

## Platform quirks

I think the API I've come up with, `createEntries`, is about as simple as it could be, though feedback is welcome. There's a fair bit of complexity inherent in this problem, because every platform does routing a little bit differently, and SvelteKit has two features in particular that are tricky to express on some of them:

* Routing syntax is more flexible than most platforms:
  * Most platforms will only allow 'catch-all' or 'splat' parameters at the end of a route, rather than the middle — `/foo/*` is okay, `/foo/*/bar` is not
  * [Rest parameters](https://kit.svelte.dev/docs#routing-advanced-rest-parameters), the SvelteKit equivalent of a splat, accept 0 or more segments. Some platforms, e.g. Architect, interpret `*` as 1 or more segments, meaning that to express `/foo/[...rest]/bar` we need both `/foo/*` and `/foo/bar` routes pointing to the same function
  * `foo/bar-[baz]` will match `/foo/bar-x` but not `/foo/qux-x`. Most platforms only allow you to express `/foo/:param`
* A page or endpoint can [fall through](https://kit.svelte.dev/docs#routing-advanced-fallthrough-routes) to other routes. Most platforms have first-match-wins semantics, which means that a function may need to be responsible for multiple routes, and a route may appear in multiple functions.

In summary, a _platform_ route needn't correspond to an _app_ route, and the way that they're grouped will differ between platforms.

The `createEntries` method generates those groups:

```js
const entries = utils.createEntries(route => {
  // for a file like `src/routes/foo/[bar].js` that exports `get` and `post`,
  // `route` would look like this:
  // 
  // {
  //   type: 'endpoint',
  //   pattern: /^foo/([^/]+)$/,
  //   segments: [
  //     { content: 'foo', dynamic: false, rest: false },
  //     { content: '[bar]', dynamic: true, rest: false }
  //   ],
  //   methods: ['get', 'post']
  // }

  return {
    // the `id` is used to uniquely identify a platform route.
    // for example, `src/routes/foo/a-[x]` and `src/routes/foo/b-[x]`
    // are both reduced to /foo/:param on some platforms, so we
    // discard the latter
    id: '...',
    
    // a function may need to handle multiple routes for two reasons:
    //   1. the platform's routing is less granular than SvelteKit's
    //   2. one route might fall through to another
    //
    // because of that, we need to know which (lower-precedence) app
    // routes should also be included in the function that handles
    // the main `route`. a route like `src/routes/[...all].svelte`
    // would be included in _every_ function
    filter: fallthrough => {
      return matches(route, fallthrough);
    },

    // once we have a group of routes that belong to an individual
    // function, we can create that function
    complete: entry => {
      const manifest = entry.generateManifest({
        relativePath: '../server'
      });

      const fn = `import { init } from '../handler.js';\n\nexport const handler = init(${manifest});\n`;

      writeFileSync(...);
      redirects.push(...);
    }
  }
});
```

It's a little complicated, but then it's doing a slightly complicated thing. Most SvelteKit users will never need to understand this or touch it.

## Notable API changes

This felt like as good a time as any to introduce some overdue changes to the adapter API (though maybe it would be kind to deprecate certain methods instead of nuking them altogether).

We use snake_case inside the SvelteKit codebase, because it's inherently superior to camelCase. Since the wider JavaScript community isn't yet in agreement, our policy is to use camelCase for public APIs. Currently, the adapter API violates that principle.

In addition to adding `createEntries`, this PR updates the following method names:

* `copy_client_files` -> `writeClient`
* `copy_server_files` -> `writeServer`
* `copy_static_files` -> `writeStatic`

The `copy` method signature has changed:

```js
// old
utils.copy(from, to, file => file[0] !== '.');

// new
utils.copy(from, to, {
  filter: file => file[0] !== '.',
  replace: {
    ANSWER: '42'
  }
});
```

The `app.js` file still exists, but whereas it previously exported `init` and `render` functions, it now exports an `App` class that expects a manifest of the form created by the `entry.generateManifest({...})` method above:

```js
const fn = `
import { App } from './server/app.js';

const app = new App(${entry.generateManifest({ relativePath: './' })});

export const handler = (req) => {
  const [path, search] = req.url.split('?');

  const { status, headers, body } = app.render({
    method: req.httpMethod,
    headers: req.headers,
    path,
    query: new URLSearchParams(search),
    rawBody: normalise_body(req.body)
  });

  return {
    statusCode: status,
    headers,
    body
  };
};
`;

fs.writeFileSync(`.platform/functions/${entry.name}.js`, fn);
```

It also exports an `override` function, but this is used by SvelteKit during prerendering and `svelte-kit preview`; adapter authors don't need to use it.

---

### Tests

### Changesets


Rich-Harris:
Hmm. I find myself wondering if it's necessary to generate functions in two steps — maybe instead of this...

```js
const entries = utils.createEntries(route => {
  return {
    id,
    data: {
      name
    },
    filter: (fallback) => matches(route, fallback)
  };
});

entries.forEach(entry => {
  fs.writeFileSync(
    `${out}/${entry.data.name}.js`,
    `blah(${entry.generateManifest({ relativePath: './' })`
  );
});
```

...it should be something like this (except with a better name than `callback`) so that carting `data` around is unnecessary:

```js
utils.createEntries(route => {
  return {
    id,
    filter: (fallback) => matches(route, fallback),
    callback: (generateManifest) => {
      fs.writeFileSync(
        `${out}/${name}.js`,
        `blah(${generateManifest({ relativePath: './' })`
      );
    }
  };
});
```

packages/kit/src/core/build/index.js
```diff
@@ -179,42 +221,42 @@ async function build_client({
 
 	print_config_conflicts(conflicts, 'kit.vite.', 'build_client');
 
-	await vite.build(merged_config);
+	const output = /** @type {import('rollup').OutputChunk[]} */ (
+		/** @type {any} */ (await vite.build(merged_config)).output
```
benmccann, line 225: can this be something more specific like `RollupOutput`?

packages/kit/src/core/build/index.js
```diff
@@ -491,15 +431,42 @@ async function build_server(
...
+		if (!chunk.facadeModuleId) return;
+		const id = chunk.facadeModuleId.slice(cwd.length + 1);
+		lookup[id] = chunk.exports;
+	});
+
+	/** @type {Record<string, import('types/internal').HttpMethod[]>} */
+	const methods = {};
+	manifest_data.routes.forEach((route) => {
+		if (route.type === 'endpoint') {
+			if (lookup[route.file]) {
```
benmccann, line 450: could combine these
```suggestion
		if (route.type === 'endpoint' && lookup[route.file]) {
```

packages/kit/src/core/build/index.js
```diff
@@ -491,15 +431,42 @@ async function build_server(
 
 	print_config_conflicts(conflicts, 'kit.vite.', 'build_server');
 
-	await vite.build(merged_config);
+	const output = /** @type {import('rollup').OutputChunk[]} */ (
+		/** @type {any} */ (await vite.build(merged_config)).output
+	);
+
+	/** @type {Record<string, string[]>} */
```
benmccann, line 438: maybe this and the next ~20 lines could be refactored out into something like a `get_methods` function?

packages/kit/src/core/build/index.js
```diff
@@ -223,179 +265,56 @@ async function build_server(
...
-		const css = Array.from(css_deps);
+	});
 
-		const styles = config.kit.amp
-			? Array.from(css_deps).map((url) => {
-					const resolved = `${output_dir}/client/${config.kit.appDir}/${url}`;
-					return fs.readFileSync(resolved, 'utf-8');
-			  })
-			: [];
+	manifest_data.components.forEach((file) => {
```
benmccann, line 285: something that could use a comment: why a separate loop here over `manifest_data.components`? i.e. why does the loop above only handle endpoints?

packages/kit/src/core/build/index.js
```diff
@@ -179,42 +221,42 @@ async function build_client({
...
+	);
 
-	const client_manifest_file = `${client_out_dir}/manifest.json`;
 	/** @type {import('vite').Manifest} */
-	const client_manifest = JSON.parse(fs.readFileSync(client_manifest_file, 'utf-8'));
-	fs.renameSync(client_manifest_file, `${output_dir}/manifest.json`); // inspectable but not shipped
+	const manifest = JSON.parse(fs.readFileSync(`${client_out_dir}/manifest.json`, 'utf-8'));
 
-	return client_manifest;
+	// TODO where does this go?
```
benmccann, line 231: I don't understand this comment. Perhaps it could be expanded upon?

packages/kit/src/core/build/index.js
Rich-Harris: endpoints are defined by a single module, but pages are defined by multiple components (plan a: [layout, layout, page], plan b: [error]). since there's component overlap between pages (many pages share layout/error components) we loop over the deduped list instead. added a comment

packages/kit/src/core/build/index.js
Rich-Harris: was just an out of date comment — removed

packages/kit/src/core/build/index.js
Rich-Harris: yep, that's way nicer

packages/adapter-cloudflare/index.js
```diff
@@ -3,24 +3,17 @@ import { fileURLToPath } from 'url';
...
 
-/**
- * @param {esbuild.BuildOptions} [options]
- */
+/** @type {import('.')} */
 export default function (options = {}) {
 	return {
 		name: '@sveltejs/adapter-cloudflare',
-		async adapt({ utils, config }) {
+		async adapt(utils) {
```
benmccann, line 10: It seems like it'd be more future-safe to keep this as `{ utils }` in case we want to add anything else in the future. I think we were trying to do that with all the public APIs. It's also slightly less for adapter authors to have to change to adapt to this PR

Rich-Harris: the way i think of it is that `utils` already _is_ the future-proof bag of non-positional arguments. this...

```js
adapt(utils) {
  utils.rimraf(out);
  // ...
}
```

...is equivalent to this:

```js
adapt({ rimraf }) {
  rimraf(out);
  // ...
}
```

it's the same as endpoints taking a `request` as a single argument:

```js
export function get({ path, params }) {
  // ...
}
```

if we need to add more stuff in future, we could do so by putting it on `utils`.

this might feel 'nicer' if it wasn't called `utils` — not sure what a better name would be though.

benmccann: I don't really have any better ideas for names. Maybe `args`? That would feel more natural if `config` were added back in the future and would clarify that it's not just utils, but perhaps makes less sense in the short-term

Rich-Harris:
> need to be able to return an array of entries, because of how Architect handles catch-alls (see below)

actually it occurs to me that the API doesn't need to handle that situation; an Architect adapter could simply write two routes for a single entry. it would probably look something like this:

```js
return {
  id: '...',
  data: {
    routes: ['/foo/*', '/foo/baz']
  },
  filter: (other) => {...}
}
```

packages/adapter-cloudflare/index.js
Rich-Harris: ended up with `builder`, per discord discussion, though remain open to alternatives

Rich-Harris:
Have marked this ready for review (CI failures notwithstanding), but intend to follow up with:

* fixing up the adapter-node tests. to be honest some of them feel a bit random, i think we could have a more joined up approach to testing adapters generally
* i'm pretty confident that getting rid of the ability to customise esbuild options is the right thing to do (in cases where we haven't removed esbuild altogether, like node). the one exception is adapter-vercel, where we're using esbuild (because it's easier than trying to use the platform's bundling mechanism) but need to opt out in cases where a native dependency like `sharp` is being used. I think in these cases we will probably need to expose an `external` option rather than exposing the esbuild guts; as well as excluding external deps from the bundled .js file, we also need to jump through some hoops to install those deps alongside the lambda, and that's easier to do if we're being explicit about what options we're exposing rather than just throwing it fully open and hoping for the best
* the docs aren't very comprehensive, but i'm wary of fleshing them out too far as they will take up a lot of space. i think we should probably move to a multi-page doc setup like https://svelte-cubed.vercel.app/docs/getting-started, but it would be good to implement some kind of search (e.g. a Cmd-K palette that uses [Stork](https://stork-search.net/) or [TinySearch](https://endler.dev/2019/tinysearch) or something) alongside that, and that's definitely out of scope for the time being

Rich-Harris:
Forgot to add changesets and ping community adapter maintainers — something for the morning

.gitignore
```diff
@@ -5,6 +5,7 @@ package-lock.json
 yarn.lock
 /examples/**/.svelte
 /examples/**/build
+/sandbox
```
benmccann, line 8: does this make sense to add?

packages/adapter-node/index.js
```diff
@@ -16,103 +7,55 @@ import zlib from 'zlib';
...
-				format: 'esm',
-				platform: 'node',
-				target: 'node14',
-				inject: [join(files, 'shims.js')],
-				define: {
-					APP_DIR: `"/${config.kit.appDir}/"`
+			builder.copy(files, out, {
+				replace: {
+					APP: './server/app.js',
+					MANIFEST: './manifest.js',
```
benmccann, line 48: should the adapter have to know about these? maybe we could use Vite aliases?

packages/kit/src/core/build/build_server.js
```diff
@@ -0,0 +1,274 @@
...
+import { deep_merge } from '../../utils/object.js';
+import { print_config_conflicts } from '../config/index.js';
+import { posixify, resolve_entry } from '../utils.js';
+import { create_build } from './utils.js';
+import { SVELTE_KIT } from '../constants.js';
+import { s } from '../../utils/misc.js';
+
+/**
+ *
+ * @param {{ runtime: string, hooks: string, config: import('types/config').ValidatedConfig }} param0
```
benmccann, line 13:
```suggestion
 * @param {{ runtime: string, hooks: string, config: import('types/config').ValidatedConfig }} param0
```

packages/kit/src/core/build/build_service_worker.js
```diff
@@ -0,0 +1,102 @@
...
+			];
+		`
+			.replace(/^\t{3}/gm, '')
+			.trim()
+	);
+
+	/** @type {import('vite').UserConfig} */
+	const vite_config = config.kit.vite();
+
+	const default_config = {
```
benmccann, line 61: you can remove this (same in server and client). it'll be the default in Vite in a couple of days: https://vitejs.dev/config/#server-fs-strict

packages/kit/src/core/dev/index.js
```diff
@@ -200,43 +200,95 @@ class Watcher extends EventEmitter {
...
 		/** @type {import('types/internal').SSRManifest} */
 		this.manifest = {
-			assets: manifest_data.assets,
-			layout: manifest_data.layout,
-			error: manifest_data.error,
-			routes: manifest_data.routes.map((route) => {
-				if (route.type === 'page') {
+			appDir: this.config.kit.appDir,
+			assets: new Set(manifest_data.assets.map((asset) => asset.file)),
+			_: {
```
benmccann, line 205: `_` is the property name or is it just a special syntax I'm unfamiliar with? Assuming it's the name, can we come up with something more descriptive?

packages/kit/src/core/dev/index.js
```diff
@@ -200,43 +200,95 @@ class Watcher extends EventEmitter {
...
+				mime: get_mime_lookup(manifest_data),
+				entry: {
+					file: `/${SVELTE_KIT}/dev/runtime/internal/start.js`,
+					css: [],
+					js: []
+				},
+				nodes: manifest_data.components.map((id) => {
+					return async () => {
```
benmccann, line 213: I think of a manifest as just being static data, so wouldn't quite expect to see a function in the manifest. We can't leave this as `load_component`?

packages/kit/src/utils/filesystem.js
```diff
@@ -17,27 +18,60 @@ export function rimraf(path) {
...
+
+	const regex = opts.replace
+		? new RegExp(`\\b(${Object.keys(opts.replace).join('|')})\\b`, 'g')
+		: null;
+
+	/**
+	 * @param {string} from
+	 * @param {string} to
+	 */
+	function go(from, to) {
```
benmccann, line 44: `go` sounds like `cd`. would something like `do_copy` be accurate?

benmccann:
@ascorbic I'd love your eyes on this as it makes big changes to SvelteKit's `adapter-netlify`

Looping in community adapter authors as well: @jthegedus @pluvial @geoffrich @MacFJA @HalfdanJ @ryanbethel

ascorbic:
This is looking great, @Rich-Harris. Thanks for the heads-up, @benmccann. I'm going to go over it in detail today, but from an initial look it seems great.

This does make me think there should be a standardised way to do adapters like these. SvelteKit, Remix and Nuxt (via Nitro) have implemented basically the same thing, and it seems if there were a shared way of doing this it would save everyone a lot of effort, letting framework authors target a common API, and deployment platforms maintain a single glue layer. What do you think?

packages/adapter-node/index.js
Rich-Harris: Vite isn't involved here

packages/kit/src/core/dev/index.js
Rich-Harris: the idea is that the manifest is considered public API so that adapters can use stuff like `assets` and `appDir` (which the cloudflare adapter uses to handle requests for static assets), but there's also a private area where we make no guarantees of stability. i figured `_` is as good a name as any since it's a commonly used convention for 'private stuff'

packages/kit/src/core/dev/index.js
Rich-Harris: it makes things considerably more complex if the manifest isn't self-contained. am open to alternative naming suggestions, haven't come up with anything thus far

packages/kit/src/utils/filesystem.js
Rich-Harris: i have an irrational aversion to the `do_foo` naming convention. in my mind 'go' is like 'we've finished setup, now we can go'

Rich-Harris:
@ascorbic thanks — I would love to see more standardisation around this. I suspect it would be a very challenging thing to do though.

For example while Vercel and Netlify both let you target lambda (albeit Vercel wraps it with a Node http-like interface) or V8 isolates (albeit with different runtimes), other platforms are either all-lambda (e.g. Begin) or all-isolates (e.g. Cloudflare Pages). And routing semantics differ in various ways (Vercel uses regexes, while Netlify uses the `_redirects` syntax, which looks similar to the syntax Architect uses but differs in e.g. how splats are handled). It seems hard to imagine a glue layer that could smooth over all that complexity and all the edge cases around bundling without a) suffering from lowest common denominator syndrome or b) adding some amount of overhead.

But it's definitely possible that you're seeing things I'm not, or that I'm being unduly pessimistic. Perhaps there's a _good enough_ interchange format that could sit between frameworks and platforms that would make more specialised adapters unnecessary for 90% of cases. Definitely worth thinking further on.

ascorbic:
@Rich-Harris I think it's true that it's not realistic right now to expect the hosting platforms to standardise their own APIs because of different support features and limitations. What I'm suggesting is a tool that can serve as an abstraction layer between them. Each platform could be implemented as a plugin or adapter, and then frameworks can use a standard API for all target platforms. So for example it could provide primitives for defining routing, and rules for bundling entrypoints. I'm wondering if Nitro could do some of this, because I know the Nuxt team are planning to separate it and make it framework-agnostic. Cc @danielroe @pi0

Rich-Harris:
We're thinking along similar lines — getting platforms to coordinate is definitely a non-starter but a glue layer could work; I'm just slightly less pessimistic than you about how hard it would be and what trade-offs it would involve! If there's a way for us to make it happen then consider me an enthusiastic supporter of the idea

danielroe:
@Rich-Harris Would love to chat more about this!

.gitignore
Rich-Harris: i dunno. it's been really helpful to have a sandbox app that's linked to the workspace, but as we saw you can't avoid that stuff showing up in the lockfile. maybe i should just do that stuff in a completely separate repo that links to the packages in here, i guess i'll delete this

dslatkin:
> * the docs aren't very comprehensive, but i'm wary of fleshing them out too far as they will take up a lot of space. i think we should probably move to a multi-page doc setup like https://svelte-cubed.vercel.app/docs/getting-started, but it would be good to implement some kind of search (e.g. a Cmd-K palette that uses [Stork](https://stork-search.net/) or [TinySearch](https://endler.dev/2019/tinysearch) or something) alongside that, and that's definitely out of scope for the time being

Would love this for the docs.

Somewhat related - if you're ever considering a dark mode for the docs, that would make it so much nicer and easier for adoption by us night owls. 🦉

packages/adapter-cloudflare-workers/index.js
```diff
@@ -1,73 +1,78 @@
...
+				`${tmp}/manifest.js`,
+				`export const manifest = ${builder.generateManifest({
+					relativePath
+				})};\n`
+			);
 
-			/** @type {BuildOptions} */
-			const default_options = {
-				entryPoints: ['.svelte-kit/cloudflare-workers/entry.js'],
+			await esbuild.build({
```
benmccann, line 53: It looks like cloudflare supports ES modules now, so maybe we can ditch `esbuild`?

https://blog.cloudflare.com/workers-javascript-modules/

packages/kit/types/app.d.ts
```diff
@@ -1,9 +1,15 @@
...
+import { PrerenderOptions, SSRManifest } from './internal';
 
-export interface App {
-	init(): void;
-	render(incoming: IncomingRequest): Promise<ServerResponse>;
+export class App {
+	constructor(manifest: SSRManifest);
+	render(
+		incoming: IncomingRequest,
+		options?: {
```
benmccann, line 9: the `options` are only ever used internally, so we had moved them to `internal.d.ts` in order to simplify the method signature for adapter authors

packages/kit/types/app.d.ts
```diff
@@ -1,9 +1,15 @@
 import { ReadOnlyFormData, RequestHeaders } from './helper';
 import { ServerResponse } from './hooks';
+import { PrerenderOptions, SSRManifest } from './internal';
```
benmccann, line 3: we probably shouldn't import things from `internal`. if they're being exposed they should be moved here

packages/kit/types/internal.d.ts
```diff
@@ -105,11 +101,21 @@ export type SSRRoute = SSREndpoint | SSRPage;
...
+export type SSRNodeLoader = () => Promise<SSRNode>;
+
 export interface SSRManifest {
-	assets: Asset[];
-	layout: string;
-	error: string;
-	routes: SSRRoute[];
+	appDir: string;
+	assets: Set<string>;
+	_: {
```
benmccann, line 109:
```suggestion
	/** private fields */
	_: {
```

packages/adapter-cloudflare-workers/index.js
Rich-Harris: This doesn't appear to work for Cloudflare Pages yet — in wrangler I get this...

> SyntaxError: Cannot use import statement outside a module

...and on trying to deploy I get this:

> Deploying to Cloudflare's global network

Hopefully it'll get ironed out at some point, and we can drop the final bundling step in a future PR

Rich-Harris:
Kicking myself as I realise there's a bit of a flaw in this whole idea — we don't know at build time which pages depend on which endpoints. 

So to take the demo app as an example, the `/todos` page depends on the `/todos.json` endpoint, but that relationship isn't expressed by the `createEntries` logic that decides which routes belong to which function. As a result, when a function tries to server-render `/todos`, which results in Kit loading the `/todos.json` endpoint from that function's manifest, it will fail.

Options:

1. Bundle every endpoint with every function. This would still be an improvement over the status quo, but less conclusively. It would also prevent us from (for example) running some endpoints on the edge and others in a lambda, where relevant (probably only Vercel), which is useful in cases where one endpoint has Node dependencies that can't be bundled
2. Fall back to making real HTTP requests during SSR when fetching from an endpoint that isn't in the function's manifest. Not 100% sure what this would look like, or how significant the request overhead would be
3. Use static analysis or heuristics to determine which endpoints belong with which pages. (This is basically impossible, and a bad idea.)

packages/kit/types/app.d.ts
Rich-Harris: done

packages/kit/types/app.d.ts
```diff
@@ -1,11 +1,21 @@
...
+import { PrerenderOptions, SSRNodeLoader, SSRRoute } from './internal';
 
-export interface App {
-	init(): void;
+export class App {
+	constructor(manifest: SSRManifest);
 	render(incoming: IncomingRequest): Promise<ServerResponse>;
 }
 
+export class InternalApp extends App {
```
benmccann, line 10: perhaps this should be in `internal.d.ts`?

Rich-Harris:
@tanhauhau points out that if HTTP is merely a fallback for the cases where the endpoint doesn't exist in the current function's manifest, then maybe a heuristic becomes tenable. Implementation-wise, this entails the following:

1. Return `null` rather than a rendered 404 page when SSR'ing a response to an internal `fetch`
2. If the internal `fetch` returns `null`, fall back to a real HTTP fetch

To construct the URL for the real HTTP fetch we need to know the protocol as well as the host. Right now the host can be derived from [`hostHeader`](https://kit.svelte.dev/docs#configuration-hostheader), but this isn't set by default because we don't want to trust `x-forwarded-host` for any old reverse proxy. There's no equivalent option for protocol. We could add one...

```js
export default {
  kit: {
    hostHeader: 'x-forwarded-host',
    protocolHeader: 'x-forwarded-proto'
  }
};
```

...or make the whole thing neater and more future-proof:

```js
export default {
  kit: {
    headers: {
      host: 'x-forwarded-host',
      protocol: 'x-forwarded-proto'
    }
  }
};
```

Either way, this will be unset in the common case, because right now people don't generally need the host or the protocol. With this change, apps deployed to serverless platforms where route splitting happens (just Netlify in this PR, but Vercel soon) _would_ need to know those. I'm therefore wondering if it ought to be possible for adapters to specify default headers — if you're using the Netlify or Vercel adapters then it's reasonable to automatically trust them:

```diff
export default function () {
  return {
    name: '@sveltejs/adapter-vercel',
+   headers: {
+     host: 'x-forwarded-host',
+     protocol: 'x-forwarded-proto'
+   },
    async adapt(builder) {
```

The heuristic I'm imagining is something like 'in a function that includes a `/foo/[bar]` page, include a `/foo/[bar].json` endpoint if such there be'. It would capture idiomatic cases, but not cases where `/foo/[bar]` fetches data from e.g. `/api/foo/[bar]` — those would fall back to HTTP. 

Maybe that's okay. Maybe it's too magical and unpredictable. Not sure. There would definitely be _some_ cases where two functions would need to be involved, and in the worst case scenario that could involve two cold starts, so adapters probably ought to provide the option to disable splitting.

packages/kit/types/app.d.ts
Rich-Harris: I don't think it matters. If something is re-exported from `types/index.d.ts`, it's public, otherwise it's internal. There are plenty of examples of other internal types that are exported from files that also contain public types (e.g. `types/config.d.ts`)

packages/kit/types/config.d.ts
```diff
@@ -1,38 +1,85 @@
...
+}
+
+export interface Builder {
 	log: Logger;
 	rimraf(dir: string): void;
 	mkdirp(dir: string): void;
+
+	/**
+	 * Create entry points that map to individual functions
+	 * @param fn TODO explain this
```
benmccann, line 29: this TODO seems easy enough to fix now

packages/adapter-netlify/index.js
```diff
@@ -1,83 +1,158 @@
...
+			// for esbuild, use ESM; for zip-it-and-ship-it, use CJS
+			const esm = netlify_config?.functions?.node_bundler === 'esbuild';
+
+			/** @type {string[]} */
+			const redirects = [];
+
+			if (esm) {
+				builder.copy(`${files}/esm`, '.netlify');
+			} else {
+				// TODO might be useful if you could specify CJS/ESM as an option to writeServer
```
benmccann, line 55: I think this TODO would imply esbuild being included inside `writeServer`, which I don't like the idea of. Could we instead have an adapter option which returns the server build format to pass to the Vite build?

packages/adapter-netlify/index.js
```diff
@@ -1,83 +1,158 @@
...
-			utils.copy_client_files(publish);
-
-			utils.log.minor('Writing redirects...');
-
-			const redirectPath = join(publish, '_redirects');
-			utils.copy('_redirects', redirectPath);
-			appendFileSync(redirectPath, '\n\n/* /.netlify/functions/__render 200');
+			builder.writeServer('.netlify/server');
+
+			// for esbuild, use ESM; for zip-it-and-ship-it, use CJS
```
benmccann, line 46:
```suggestion
			// for esbuild, use ESM
			// for zip-it-and-ship-it, use CJS until https://github.com/netlify/zip-it-and-ship-it/issues/750
```

packages/adapter-node/README.md
```diff
@@ -60,77 +56,33 @@ env: {
...
-app.listen(3000);
-```
-
-For using middleware in dev mode, [see the FAQ](https://kit.svelte.dev/faq#how-do-i-use-x-with-sveltekit-how-do-i-use-middleware).
-
-## Advanced Configuration
-
-### esbuild
+// let SvelteKit handle everything else, including serving
+// prerendered pages and static assets
```
benmccann, line 78: this seems like a short enough comment it could it on one line. or at least I'd move the break later in the sentence
```suggestion
// let SvelteKit handle everything else, including serving prerendered pages and static assets
```

packages/adapter-netlify/README.md
```diff
@@ -66,38 +68,6 @@ During compilation a required "catch all" redirect rule is automatically appende
   node_bundler = "esbuild"
 ```
 
-## Advanced Configuration
```
benmccann, line 69: I wonder if we'll catch anyone out by removing this since we haven't actually entirely gotten rid of esbuild yet in the zip-it-and-ship-it case. It looks like Vercel always uses esbuild still and so we're probably even more likely to have people there who can no longer deploy with these options removed there

packages/kit/src/cli.js
```diff
@@ -101,12 +101,12 @@ prog
 				process.stderr.write(data);
 			});
 
-			if (!watcher.vite || !watcher.vite.httpServer) {
```
benmccann, line 104: I personally think `vite` was a much clearer name. If you know it's Vite then you have a better understanding of what it is whereas `server` is much more generic. And `server.httpServer` is a little weird

documentation/docs/01-routing.md
```diff
@@ -59,7 +59,6 @@ type RequestHeaders = Record<string, string>;
 export type RawBody = null | Uint8Array;
 export interface IncomingRequest {
 	method: string;
-	host: string;
```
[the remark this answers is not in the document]
Rich-Harris: It's on the request, but not `IncomingRequest` (which is what adapters pass to Kit). Having `host` on the type here was a mistake

packages/adapter-netlify/index.js
Rich-Harris: we can cross that bridge if another adapter also needs CJS output. I don't think it's worth allowing opening the pandora's box of allowing adapters to affect build output just for this

packages/adapter-netlify/README.md
Rich-Harris: from discord, discussing the original reason for exposing `esbuild` configuration: 

> so of those issues, the node-related ones should vanish since esbuild is no longer being used there. the sharp thing is a real problem for environments where we need to bundle (vercel, cloudflare) but giving people the ability to modify the esbuild config isn't going to change that AFAICT. the type reflection issue seems unrelated to any final bundling step, it was happening during dev i think? so it feels like we can probably remove the ability to modify the config, at least for now

there's just no reason to expose it that i can see. we could reinstate it if someone comes up with a valid use case, but right now there doesn't appear to be one

packages/kit/src/cli.js
Rich-Harris: reverted

packages/kit/types/config.d.ts
Rich-Harris: done, though it's a very bare-bones description — I don't think we can fully explain this method without some documentation and examples

packages/adapter-netlify/index.js
benmccann: maybe just remove this TODO? it'll won't be an issue after Netlify supports ESM anyway

documentation/docs/01-routing.md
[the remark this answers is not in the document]
benmccann: It should be added to the request about 15 lines down from here then?

packages/kit/src/core/adapt/test/index.js
```diff
@@ -39,9 +39,16 @@ suite('copy files', () => {
 	};
 
 	/** @type {import('types/internal').BuildData} */
-	const build_data = { client: [], server: [], static: [], entries: [] };
+	const build_data = {
+		// @ts-ignore
```
benmccann, line 43: we've switched to `@ts-expect-error` if it's something that we want to permanently ignore and are using `@ts-ignore` for temporary todos

packages/kit/src/core/build/build_client.js
```diff
@@ -0,0 +1,119 @@
...
+}) {
+	create_app({
+		manifest_data,
+		output: build_dir,
+		cwd
+	});
+
+	copy_assets(build_dir);
+
+	process.env.VITE_SVELTEKIT_AMP = config.kit.amp ? 'true' : '';
```
benmccann, line 39: I don't see this used anywhere

packages/kit/src/core/build/build_client.js
```diff
@@ -0,0 +1,119 @@
...
+
+	const client_out_dir = `${output_dir}/client/${config.kit.appDir}`;
+
+	/** @type {Record<string, string>} */
+	const input = {
+		start: path.resolve(cwd, client_entry_file)
+	};
+
+	// This step is optional — Vite/Rollup will create the necessary chunks
```
benmccann, line 48: I wonder if this should be a dev only thing? I've always felt that SvelteKit file names reveal a bit too much about my internal code that I might want to obfuscate a bit more

packages/kit/src/core/build/build_client.js
benmccann: If it's meant to be exposed to user apps, it's not in the docs

packages/kit/src/core/dev/index.js
benmccann: renaming might not be a bad idea just because there's also the Vite `manifest.json` and it's sometimes confused me in the past that there are two things referred to as the manifest. Maybe `metadata`?

packages/kit/src/core/adapt/builder.js
```diff
@@ -0,0 +1,171 @@
...
+			for (let i = 0; i < routes.length; i += 1) {
+				const route = routes[i];
+				const { id, filter, complete } = fn(facades[i]);
+
+				if (seen.has(id)) continue;
+				seen.add(id);
+
+				const group = [route];
+
+				// figure out which _lower priority_ routes should be considered fallbacks
```
benmccann, line 59: does VS Code render markdown in code comments? it makes it harder to read on Github at least
```suggestion
				// figure out which lower priority routes should be considered fallbacks
```

packages/kit/types/config.d.ts
```diff
@@ -1,38 +1,83 @@
...
 
-export interface AdapterUtils {
+export interface RouteDefinition {
+	type: 'page' | 'endpoint';
+	pattern: RegExp;
+	segments: RouteSegment[];
+	methods: HttpMethod[];
+}
+
+export interface AdapterEntry {
```
benmccann, line 12: it could help to comment this one. questions I have: what's an adapter entry (is there one or multiple), what's the ID represent, what does the filter do?

documentation/docs/01-routing.md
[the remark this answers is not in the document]
Rich-Harris: good catch, added

packages/adapter-netlify/index.js
Rich-Harris: yeah, that seems reasonable. removed

packages/kit/src/core/adapt/test/index.js
Rich-Harris: fixed

packages/kit/src/core/build/build_client.js
Rich-Harris: it's used here https://github.com/sveltejs/kit/blob/1ec368a9b90d65f13a818ff348187a7160544ea7/packages/kit/src/runtime/app/env.js#L16

packages/kit/src/core/build/build_client.js
Rich-Harris: Hmm. There isn't really such a thing as a dev-only build. I don't know if this rises to the level of exposing an option. Personally I definitely err on the side of debuggability over obfuscation (which is no defense against evildoers anyway), so my preference would be to maintain the current behaviour. 

This isn't new to this PR though (it just looks that way because `build_client` was factored out into a new module) — maybe this is worth a separate issue so we don't need to decide it here?

benmccann: i'm fine punting on it since it's not new like you said

packages/kit/types/config.d.ts
Rich-Harris: I've added some docs to the best of my abilities — it's not the easiest API to describe (though I've searched in vain for ways to simplify it further)

packages/kit/src/core/dev/index.js
Rich-Harris: Definitely not `metadata` 😅 — this is the meat of the app, or at least a slice of that meat (when we're talking about route splitting rather than a full manifest). It's more like 'content' or something, but I don't think that really captures it.

packages/kit/types/config.d.ts
```diff
@@ -1,38 +1,99 @@
...
+	/**
+	 * A string that uniquely identifies a function and is used for deduplication.
```
benmccann, line 14: I was thinking JS function when I first read this, so maybe we should say serverless function here? But then I though possibly it's not even a serverless thing, so maybe call it an HTTP service?
```suggestion
	 * A string that uniquely identifies an HTTP service (e.g. serverless function) and is used for deduplication.
```

Rich-Harris: yeah, i like that

packages/kit/src/core/dev/index.js
benmccann: `inventory`? `catalog`? `catalogue`? :grin:

abdo643-HULK:
Thx for the service-worker register change. 👍

kory-smith:
Immediately after this change, I started getting this error when using the vercel adapter:

```
> builder.getBuildDirectory is not a function
TypeError: builder.getBuildDirectory is not a function
    at adapt (file:///Users/kory/Gits/sydney-kory-wedding-website/node_modules/@sveltejs/adapter-vercel/index.js:17:24)
    at adapt (file:///Users/kory/Gits/sydney-kory-wedding-website/node_modules/@sveltejs/kit/dist/chunks/index4.js:393:8)
    at file:///Users/kory/Gits/sydney-kory-wedding-website/node_modules/@sveltejs/kit/dist/cli.js:896:11
```

Rich-Harris:
@kory-smith you need to upgrade `@sveltejs/kit`

matths:
> `adapter-node` calls `render` inside a Polka handler

So how does it work now?
When do you expect to have a documentation about the node-adapter and how to customize?
Is it still possible to use the whole SvelteKit App as just one middleware in another node.js Server application either with Polka, Express or whatever? Can you point me to an example, please?

Mlocik97:
yes, Kit as middleware works fine and same way as it's written in docs.

matths:
@Mlocik97 can you point me to an example?
https://kit.svelte.dev/faq#how-do-i-use-x-with-sveltekit-how-do-i-use-middleware is an example how to use other middleware in DEV mode. No example about PROD mode.

https://kit.svelte.dev/docs#adapters-supported-environments-node-js
no more options beside `out:`

Mlocik97:
it's partially described here: https://github.com/sveltejs/kit/tree/master/packages/adapter-node#custom-server

but I see it's missing in docs (it was there before), not sure why... but when they rewrote those adapter sections, somehow, part for this was deleted.

abdo643-HULK:
> it's partially described here: https://github.com/sveltejs/kit/tree/master/packages/adapter-node#custom-server
> 
> but I see it's missing in docs (it was there before), not sure why... but when they rewrote those adapter sections, somehow, part for this was deleted.

as far as I can remember it was only in the readme, like the current version

Mlocik97:
if I remember correctly there was short description about it and link to Readme. It's not now... anyway it would be better to add it to docs.

HalfdanJ:
I'm starting porting over [svelte-adapter-appengine](https://github.com/HalfdanJ/svelte-adapter-appengine) to this new API, the serving part is pretty straight forward there (based on the adapter-node), but the routes are a bit more tricky. I need to create a routing file that correctly sends traffic either to static/prerendered content that is hosted from cloud storage (so never hitting the node server), or the dynamic traffic to the node server. 

The new createEntries function seem to be a step forward in terms of creating this routing file correctly, but one thing I'm wondering about is how to determine in that function if a page is actually prerendered, and therefore shouldn't be handled by the node server. I could check if the `.html` file exists, but would it be possible to add a `"prerendered": true` to the entries in createEntries?

geoffrich:
@HalfdanJ `builder.prerender()` returns a list of prerendered files, which you should be able to use to determine if a page was prerendered.

https://github.com/sveltejs/kit/blob/1a6adc8a7edbb19cb10762af1a236090dcce97b5/packages/kit/types/config.d.ts#L85-L87

HalfdanJ:
Aha ok that helps a bit, 
I was looking at createEntries source, and I'm unsure if its intended that the prerendered paths even show up? I see this line https://github.com/sveltejs/kit/blob/master/packages/kit/src/core/adapt/builder.js#L66 that filters out prerendered paths, but the code above (facades) doesnt filter them out (https://github.com/sveltejs/kit/blob/master/packages/kit/src/core/adapt/builder.js#L41). Is that intended?

jmsunseri:
pardon for asking here about this but how do i solve this issue not that the esbuild option is gone from the adapter config?
https://webcache.googleusercontent.com/search?q=cache%3AZPljxCZy8sYJ%3Ahttps%3A%2F%2Fwww.reddit.com%2Fr%2Fsveltejs%2Fcomments%2Frndsgu%2Fsolution_to_weird_error_when_trying_to_use_pg%2F+&cd=3&hl=en&ct=clnk&gl=tw 

```
vite: {
			ssr: {
				noExternal: ['pg-admin']
			}
		}
```

has no effect on this issue.

cryptodeal:
I second this ^^; My vercel build has also been failing since the removal of the esBuild config options in @sveltejs/adapter-vercel.

I've continued developing using the following package versions (last working):
```
"@sveltejs/adapter-vercel": "1.0.0-next.32",
"@sveltejs/kit": "1.0.0-next.206",
```

However, that's definitely less than ideal as I'm continuing to build around an outdated API for both @sveltejs/kit & @sveltejs/adapter-vercel.

The dependency causing my build issues (@napi-rs/*) is a package required indirectly as a dependency of a package in a shared monorepo used in the frontend. (Backend defines db schema, controllers, autogen type definitions, etc. that are used both in the frontend and for writing real time data scraped from various sources to the db, hence the monorepo greatly streamlines the dev process).

I'd previously built the frontend project successfully using both Vite resolve.alias and via installing the pnpm symlinked version of the built package as a dependency during the build process (setting pnpm to shamefully-hoist when building on Vercel) using the following esBuild config options:
```
adapter({
  esbuild(defaultOptions) {
    return {
      ...defaultOptions,
      plugins: [],
      external: ['@napi-rs/*']
    };
  }
})
```

Following the breaking API changes, I've tested with updated versions of @sveltejs/kit & @sveltejs/adapter-vercel using at least the following Vite config settings (both individually and in varying combinations), but to no avail. 

Vite Config Options Tested:
```
  vite.ssr.noExternal: ['@napi-rs/*']
  ssr.external:  ['@napi-rs/*']
  vite.optimizeDeps.exclude: ['@napi-rs/*']
```
(I think I'd tried setting it as a rollup external as well using Vite config options, but that also failed to resolve the issue)

Build always fails with the following error:
```
error: No loader is configured for ".node" files
```

Any solutions for getting this build to work with the latest versions of @sveltejs/kit and @sveltejs/adapter-vercel?

geoffrich:
@jmsunseri @cryptodeal comments like these on merged PRs are easily lost. If this PR removed a feature you needed, can you open an issue instead?

jmsunseri:
> @jmsunseri @cryptodeal comments like these on merged PRs are easily lost. If this PR removed a feature you needed, can you open an issue instead?

I wasn't entirely sure it was an issue and maybe just me misunderstanding the correct path to take now that it's been removed.  I had assumed there was a new better way to do it.

cryptodeal:
Good call, @geoffrich; I just opened [Issue #3588](https://github.com/sveltejs/kit/issues/3588).

PR #2932, opened 2021-11-26
Add config to exclude files from routing

### LINKED ISSUE #2930: Config to exclude files from routing
### Describe the problem

The following directory layout creates an endpoint for `post.spec.js` and `post.md`, and the only way to opt out is to make them [private modules](https://kit.svelte.dev/docs#routing-private-modules) with an underscore/dot prefix:

```
src
└── routes
    └── posts
        ├── post.md  👈 converted to endpoint!
        ├── post.spec.js  👈 converted to endpoint!
        └── post.svelte
```

Some developers might [require the exact layout/naming shown](https://stackoverflow.com/q/69457803/6277151), where:

 1. The files are **colocated**
 2. The files **have the same basename**

Add a configuration option that excludes files.

For example, a `kit.excludes` config could accept an array of `String`, `RegExp`, or `Function`.

 - `String` - The basename of the file/directory is compared against the string, and it's excluded if matches.
 - `RegExp` - The basename of the file/directory is tested against the regular expression, and it's excluded if matches.
 - `({ filepath, basename }) => boolean` - The file/directory is excluded if the function returns `true`.

The default would be `[^_.]` to match the current behavior. This also would enable users to override the current behavior, allowing for underscore/dot-prefixed files to be included in routing.

### Alternatives considered

 - *Use underscore/dot prefix to make the files private modules* - Defeats requirement 1
 - *Store the files outside of `src/routes`* - Defeats requirement 2

### Importance

nice to have

### Additional Information

_No response_

tony19:

### Tests

### Changesets

---

## Problem

The following directory layout creates an endpoint for `post.spec.js` and `post.md`, and the only way to opt out is to make them [private modules](https://kit.svelte.dev/docs#routing-private-modules) with an underscore/dot prefix:

```
src
└── routes
    └── posts
        ├── post.md  👈 converted to endpoint!
        ├── post.spec.js  👈 converted to endpoint!
        └── post.svelte
```

Some developers might [require the exact layout/naming shown](https://stackoverflow.com/q/69457803/6277151), where:

 1. The files are **colocated**
 2. The files **have the same basename**

## Solution

This PR adds `kit.excludes` config, which accepts an array of `String`, `RegExp`, or `Function`.

 - `String` - The basename of the file/directory is compared against the string, and it's excluded if matches.
 - `RegExp` - The basename of the file/directory is tested against the regular expression, and it's excluded if matches.
 - `({ filepath, basename }) => boolean` - The file/directory is excluded if the function returns `true`.

The default would be `[^_.]` to match the current behavior. This also would enable users to override the current behavior, allowing for underscore/dot-prefixed files to be included in routing.

*Example config:*

```js
// svelte.config.js
export default {
	kit: {
		excludes: [
			/^[_.]/,
			/\.(test|spec)\.js$/,
			'README.md',
			({ basename }) => basename.startsWith('@'),
			({ filepath }) => filepath.includes('demo')
		]
	}
};
```

Closes #2930

### DIFF
--- a/.changeset/hip-walls-flash.md
+++ b/.changeset/hip-walls-flash.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Add kit.excludes config
--- a/documentation/docs/01-routing.md
+++ b/documentation/docs/01-routing.md
@@ -163,7 +163,7 @@ The `body` property of the request object will be provided in the case of POST r
 
 ### Private modules
 
-A filename that has a segment with a leading underscore, such as `src/routes/foo/_Private.svelte` or `src/routes/bar/_utils/cool-util.js`, is hidden from the router, but can be imported by files that are not.
+A filename that has a segment with a leading underscore, such as `src/routes/foo/_Private.svelte` or `src/routes/bar/_utils/cool-util.js`, is hidden from the router, but can be imported by files that are not. The pattern that matches a private module can be configured with [`excludes`](#configuration-excludes).
 
 ### Advanced
 
--- a/documentation/docs/14-configuration.md
+++ b/documentation/docs/14-configuration.md
@@ -25,6 +25,9 @@ const config = {
 			serviceWorker: 'src/service-worker',
 			template: 'src/app.html'
 		},
+		excludes: [
+			/^_/,
+		],
 		floc: false,
 		host: null,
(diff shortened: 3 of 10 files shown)


Rich-Harris:
Thank you — am picking this up in #3576 (I couldn't push to this branch for some reason), so I'll close this

PR #2950, opened 2021-11-30
adapter-cloudflare: handle non-slash path separators

piperswe:
On Windows, the paths in `ASSETS` use backslashes instead of slashes, so they'll never match a request. This patch allows the adapter to handle platforms with separators other than `/`.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-cloudflare/index.js
+++ b/packages/adapter-cloudflare/index.js
@@ -16,12 +16,12 @@ export default function (options = {}) {
 
 			const static_files = utils
 				.copy(config.kit.files.assets, target_dir)
-			  .map((f) => f.replace(sep, '/'))
+				.map((f) => f.replace(sep, '/'))
 				.map((f) => f.replace(`${target_dir}/`, ''));
 
 			const client_files = utils
 				.copy(`${process.cwd()}/.svelte-kit/output/client`, target_dir)
-			  .map((f) => f.replace(sep, '/'))
+				.map((f) => f.replace(sep, '/'))
 				.map((f) => f.replace(`${target_dir}/`, ''));
 
 			// returns nothing, very sad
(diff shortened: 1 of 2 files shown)


piperswe:
I am now realizing that this isn't as useful as I was thinking since Cloudflare Pages builds sites itself on Linux.

benmccann:
> I am now realizing that this isn't as useful as I was thinking since Cloudflare Pages builds sites itself on Linux.

Is this something you still want to pursue then? Is it useful for testing and debugging locally on Windows?

piperswe:
> > I am now realizing that this isn't as useful as I was thinking since Cloudflare Pages builds sites itself on Linux.
> 
> Is this something you still want to pursue then? Is it useful for testing and debugging locally on Windows?

It could be - this patch works fine and barely adds any complexity, so it's probably worth it to merge.

packages/adapter-cloudflare/index.js
```diff
@@ -16,19 +16,19 @@ export default function (options = {}) {
...
+				.map((f) => f.replace(`${target_dir}${sep}`, ''));
 
 			// returns nothing, very sad
 			// TODO(future) get/save output
 			await utils.prerender({
 				dest: `${target_dir}/`
 			});
 
-			const static_assets = [...static_files, ...client_files];
+			const static_assets = [...static_files, ...client_files].map((f) => f.replaceAll(sep, '/'));
```
lukeed, line 31: This requires Node 15, which is higher than SvelteKit's 14 minimum. Best to use a RegExp, `/[\\\/]+/g`

lukeed (changes requested):

lukeed:
As mentioned, this is optional. Not necessary cuz Pages is always built on Linux

Conduitry (commented):
I'm personally against this added complexity, since the build will always be run on Linux anyway. As Luke noted in his comment, even innocuous changes like this one can inadvertently introduce new dependencies (like that on Node 15). There's no reason to make this more flexible than it needs to be.

lukeed:
+1, in favor of closing

pngwn:
May need to revisit if we start running adapters in dev but we should wait until we have a problem before we solve it. 

Also don't think we should add unnecessary complexity for no reason, however small.

PR #2955, opened 2021-12-01
[fix] 1714 rest params in endpoint

### LINKED ISSUE #1714: Allow rest parameters as end of endpoint route
I'm trying to organise each route in my sveltekit app with each `.svelte` route fetching their data from a corresponding `.json.js` exportint a `get()` function.

It's all working nicely, until I've had to create a route with a terminal rest parameter:

```
routes
-- concepts
---- [...conceptPath].svelte
```

Sveltekit was pretty happy with the rest path for the `.svelte` route. However, it rejected a `[...conceptPath].json.js` with:

> rest parameter must be a standalone segment

It seems happy with the rest parameter set as a folder, though:

```
routes
-- concepts
---- [...conceptPath]
------- index.json.js
------- index.svelte
```

~While this works~, it'll lead to a good deal of `index.svelte`/`index.json` files

Wrote a bit soon regarding the above structure working, the server seems to favour the `index.json.js` route over the `index.svelte`, and I'm getting JSON instead of HTML.

This was circumvented again with setting the `.json.js` to a different path than index. 

```
-- concepts
---- [...conceptPath]
------ load.json.js
------ index.svelte
```

It makes the delegation from the `.svelte` file to `.json.js` a little less generic.

It'd be great to be able to declare endpoints with rest parameters. If there's a technical limitation to it, though, having the situation highlighted in the documentation would work too.

Placing the rest parameter as a folder provides a workaround.

It's more a mild annoyance than anything as there's a workaround. Having some clarification in the documentation could help people wonder why their route is rejected though.

dreitzner:
- [x] fixes #1714 

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/create_manifest_data/index.js
+++ b/packages/kit/src/core/create_manifest_data/index.js
@@ -103,7 +103,7 @@ export default function create_manifest_data({ config, output, cwd = process.cwd
 				throw new Error(`Invalid route ${file} — brackets are unbalanced`);
 			}
 
-			if (/.+\[\.\.\.[^\]]+\]/.test(segment) || /\[\.\.\.[^\]]+\].+/.test(segment)) {
+			if (/.+\[\.\.\.[^\]]+\]/.test(segment) || /\[\.\.\.[^\]]+\].+/.test(segment) && !/\[\.\.\.[^\]]+\]\.json/.test(segment)) {
 				throw new Error(`Invalid route ${file} — rest parameter must be a standalone segment`);
 			}
 
--- a/packages/kit/src/core/create_manifest_data/index.js
+++ b/packages/kit/src/core/create_manifest_data/index.js
@@ -103,7 +103,10 @@ export default function create_manifest_data({ config, output, cwd = process.cwd
 				throw new Error(`Invalid route ${file} — brackets are unbalanced`);
 			}
 
-			if (/.+\[\.\.\.[^\]]+\]/.test(segment) || /\[\.\.\.[^\]]+\].+/.test(segment) && !/\[\.\.\.[^\]]+\]\.json/.test(segment)) {
+			if (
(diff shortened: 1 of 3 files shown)


packages/kit/src/core/create_manifest_data/index.js
```diff
@@ -103,7 +103,10 @@ export default function create_manifest_data({ config, output, cwd = process.cwd
 				throw new Error(`Invalid route ${file} — brackets are unbalanced`);
 			}
 
-			if (/.+\[\.\.\.[^\]]+\]/.test(segment) || /\[\.\.\.[^\]]+\].+/.test(segment)) {
+			if (
+				/.+\[\.\.\.[^\]]+\]/.test(segment) ||
+				(/\[\.\.\.[^\]]+\].+/.test(segment) && !/\[\.\.\.[^\]]+\]\.json$/.test(segment))
```
bluwy, line 108: I doubt we want a hard-coded check for `[...rest].json.js`. `.json` is a convention and it can be anything else by the user. Also, have you confirm with this change that a SvelteKit app still routes normally? I have a hunch there are other places to fix as well.

With that said, I'm not sure what's the path forward, or maybe we should stick with the issue workaround for now. I'll see if the others has an opinion on this.

dreitzner: It might be good to make it more flexible in the future.

But there are strong reasons to implement that fix:
- `.json` is the framework convention for endpoints
- `.json` is used in the example project
- because of that we would create unnecessary friction for new users coming to the framework

Rich-Harris:
Thank you. The answer is both simpler and more complicated than this — simpler in that we can just allow _any_ rest segment to have a prefix/suffix (as I believe was the case with Sapper), and more complex in that we need to tweak the implementation a bit. Closing this in favour of #3240.

PR #2959, opened 2021-12-01
[docs] highlight marketable features

benmccann:
I'd like to highlight HMR, editor support, and plugins on the homepage amongst other features and link to the relevant documentation sections, so they need to be mentioned in the docs :smile:  Links from homepage are being added in https://github.com/sveltejs/sites/pull/72


documentation/docs/00-introduction.md
```diff
@@ -14,7 +14,7 @@ SvelteKit is a framework for building extremely high-performance web apps.
...
 
-It uses [Vite](https://vitejs.dev/) with a [Svelte plugin](https://github.com/sveltejs/vite-plugin-svelte) to provide a lightning-fast and feature-rich development experience, where changes to your code are reflected in the browser immediately.
+It uses [Vite](https://vitejs.dev/) with a [Svelte plugin](https://github.com/sveltejs/vite-plugin-svelte) to provide a lightning-fast and feature-rich development experience with [Hot Module Reload (HMR)](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#hot), where changes to your code are reflected in the browser instantly.
```
bluwy, line 17: I think HMR stands for "Hot Module Replacement", though I've seen "Reload" being used interchangebly too. Regarding the linked URL, maybe we can link it to https://github.com/withastro/esm-hmr ?

benmccann: good catch!

I think linking there could be slightly confusing because it says Snowpack are the only ones using it: https://github.com/withastro/esm-hmr#whos-using-esm-hmr. perhaps if that gets updated then it'd make more sense to link to

PR #2961, opened 2021-12-01
[docs] adapter clarifications

benmccann:
Reading the docs now makes it sounds like Node.js and static sites aren't officially supported because they're not under "Supported Environments" and are listed after the community adapters.


Rich-Harris (approved):

Rich-Harris:
yeah, this is much better 👍

PR #2964, opened 2021-12-01
remove unused reload flag on NavigationResult

moatra:
I spent a few hours this morning debugging why a `popstate` event during navigation to one of my pages would result in the target page content being displayed at the previous page url.  Turns out I have an endpoint at the same url as the target page and the client router was performing `location.reload()`.  That code path didn't correctly handle a `popstate` while reloading.    That codepath is no longer possible after #2656 because it removes the endpoint routes from the client manifest, but I noticed the `location.reload()` logic is still present.  This PR cleans that up.

-----

### Tests

### Changesets

### DIFF
--- a/.changeset/yellow-mangos-fry.md
+++ b/.changeset/yellow-mangos-fry.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+remove unused reload flag on NavigationResult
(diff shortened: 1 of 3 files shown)


benmccann (approved):

Conduitry:
Is this fixing a specific bug or just tidying up unused code? If it's tidying up code, I don't think it needs a changeset. If it's fixing a bug, is it possible to have a test for it?

moatra:
It's just tidying up code.  Do PRs on this repo get squashed, or should I force update my branch to remove the changeset?

benmccann:
we squash

bluwy (approved):

Issue #1538, opened 2021-05-24
Hook for initialization?


hgl:
**Is your feature request related to a problem? Please describe.**
It should be pretty common for a svelt kit app to utilize external services that you need to initialize clients to talk to them. The doc demonstrates such use case:

```js
import db from '$lib/database';

export async function get({ params }) {
	const { slug } = params;
	const article = await db.get(slug);
	// ...
}
```

One issue with this code is that it assumes `db` can be initialized in the module itself either synchronously (which is not always the case) or via top-level `await` (which svelte kit doesn't seem to support).

**Describe the solution you'd like**
Offer a hook like `async func init() {}` that only runs once when it starts? 

**Describe alternatives you've considered**

**How important is this feature to you?**
Very, since it's pretty awkward to workaround when the initialization is async (e.g., force initialization to be synchronous and only call the async part when actually talk to external services).

**Additional context**

ignatiusmb:
Possibly related to #1530

hgl:
That issue is about client initialization. But yeah, the two issues should probably be considered en masse to ensure an intuitive API.

hgl:
Another way to solve this problem is that svelte kit offers a central place to initialize such clients, and pass the clients as an arguments to endpoints and hooks:


```js
export async function get({ params, svcs }) {
	const { slug } = params;
	const article = await svcs.db.get(slug);
	// ...
}
```

This seems to be more conventional than the approach that a module is directly usable as an external service.

babichjacob:
You seem to have missed the `request.locals` convention in the `handle` hook. Here's an example from the SvelteKit demo app: 
https://github.com/sveltejs/kit/blob/6372690df51257cb811185d33892b6b3146620e8/packages/create-svelte/templates/default/src/hooks.ts#L1-L7

There's no rule that you have to re-do work every request (or that the information involved has to be personalized to the user that made the request). Let me demonstrate:

```js
// src/hooks.js
import { stuff } from "db-library";

async function setupDatabaseOrAnythingReally() {
    const dbClient = await stuff({ ...options });
    return dbClient;
}

// Initialize the database *once*
const dbPromise = setupDatabaseOrAnythingReally();

/** @type {import('@sveltejs/kit').Handle} */
export async function handle({ request, render }) {
    // Wait for the db to initialize the first time
    // or instantly get the result of when it was initialized
    const db = await dbPromise;
    
    // Make the database available to the endpoint answering the request
    request.locals = { db };
    return await render(request);
[...]

PR #2965, opened 2021-12-02
Add FAQ entry about client-side init functions

rmunn:
Demonstrate how to guarantee that a client-side initialization function will be called only once per client.

After writing [this comment](https://github.com/sveltejs/kit/issues/1538#issuecomment-984241996) explaining how to call a library's init function only once on the client, I thought that that might be a generally-useful technique that should be documented. The FAQ section on integrations seems like the right place to put it.

No tests or changesets because this is a documentation-only PR. If a changeset should be added even to doc-only PRs, let me know and I'll be happy to add one.

### Tests

### Changesets

### DIFF
--- a/documentation/faq/80-integrations.md
+++ b/documentation/faq/80-integrations.md
@@ -38,6 +38,54 @@ export default config;
 
 See [Vite's `configureServer` docs](https://vitejs.dev/guide/api-plugin.html#configureserver) for more details including how to control ordering.
 
+### How do I setup client-side code with an initialization function?
+
+If you have a function that needs to run only once on the *client side*, such as an initialization function for a GraphQL query library, you could use an approach like this:
+
+```js
+// clientConnection.ts
+import { initalizeConnection } from 'some-library';
+import { readable } from 'svelte/store';
+
+let storedConnection = null;
+const noop = () => {};
+
+export const connection = readable(null, (set) => {
+	if (!storedConnection) {
+		storedConnection = initializeConnection('params');
+	}
+	set(storedConnection);
+	return noop;
+});
+```
+
+Then in the rest of your code, you can do this:
+
+```js
+import { connection } from '$lib/clientConnection.ts';
+
+function getResult(query) {
+	return $connection.get(query);
+}
+```
+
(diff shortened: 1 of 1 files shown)


rmunn:
The unit test failure is likely a flaky test that needs fixing, as it failed due to a timeout rather than an actual error. Plus, since this PR made no code changes, it's almost certainly not caused by the PR. I don't seem to have sufficient repo permissions to request a re-run of all jobs on the failing workflow, though. Someone with write access to the repo will probably have to do that.

Rich-Harris:
If no unsubscription is required, isn't this equivalent?

```js
// clientConnection.ts
import { initalizeConnection } from 'some-library';
export const connection = initializeConnection('params');
```

```js
import { connection } from '$lib/clientConnection.ts';
function getResult(query) {
	return connection.get(query);
}
```

Added a similar comment to #1538

documentation/faq/80-integrations.md
```diff
@@ -38,6 +38,54 @@ export default config;
 
 See [Vite's `configureServer` docs](https://vitejs.dev/guide/api-plugin.html#configureserver) for more details including how to control ordering.
 
+### How do I setup client-side code with an initialization function?
+
+If you have a function that needs to run only once on the *client side*, such as an initialization function for a GraphQL query library, you could use an approach like this:
+
+```js
+// clientConnection.ts
+import { initalizeConnection } from 'some-library';
```
paolodina, line 47: initalizeConnection typo

documentation/faq/80-integrations.md
```diff
@@ -38,6 +38,54 @@ export default config;
...
+
+```js
+import { connection } from '$lib/clientConnection.ts';
+
+function getResult(query) {
+	return $connection.get(query);
+}
+```
+
+This technique uses the [initalization function of Svelte stores](https://svelte.dev/docs#svelte_store), which is called when the number of subscribers goes from zero to one (but not from one to two, etc) to set up a connection on demand. And because the connection is cached in an external variable, when one page unsubscribes from the store and a second page subscribes, the store's initialization function will be called again, but it will return the cached connection instead of creating a new one.
```
paolodina, line 72: `initalization` typo

Rich-Harris:
Going to close this as I don't think there's any advantage to the store-based approach — thanks

Issue #2974, opened 2021-12-02
Ability to intercept a navigation request


zommerberg:
### Describe the problem

I was trying to build a confirmation dialog feature that appears before the user tries to leave the page.
![Screenshot_20211203_001741](https://user-images.githubusercontent.com/43446836/144512206-5c8b3c45-9bb9-46d5-badc-dcfced3779f4.png)

The problem that occurred is that normally in this case we would use `beforeunload` listener and prevent the navigation event from firing before leaving the page. However, in sveltekit we have client-side routing so this event is not fired. 

Currently, there is no good way of handling that.

### Describe the proposed solution

My proposed solution is to create a listener that exposes the navigation intent. 
Within this intent, we should be able to cancel the navigation and get the required information about the page we are trying to navigate to.

The idea is to create a `onBeforeNavigate` interceptor that is triggered when we use the following events:
- goto()
- popstate events
- anchor tags clicks

The interceptor could be used in the following way:
```
<script>
  import { browser } from '$app/env';
  import { onBeforeNavigate } from '$app/navigation';

     if (browser) {
       onBeforeNavigate((navigationIntent) => {

          //the url we are going to navigate to
	  console.log(navigationIntent.url);

          //optionally cancel the navigation
	  navigationIntent.cancel();

        });
      }
	
</script>
```

### Alternatives considered

An alternative is just to disable the client-side router

### Importance

i cannot use SvelteKit without it

zommerberg:
Closing this since there is an issue for that already #560

zommerberg:
Reopening this since it is a different request from the `onNavigate` function proposed by Rich. I think we can move the discussion about the `onBeforeNavigate` function in this request.

zommerberg:
I was looking at the current route implementation and new `onBeforeNavigate` would need to be triggered by a  synthetic event. Currently we have "navigation-start" and "navigation-end" event, we should add something like "navigation-intent" event
and add it to all of the palces that handle click, goto and browser navigations like so:
```
                        //by default the navigation is allowed
	                let startNavigation = true;

                        //once we verified that the route is valid we emit a navigation-intent event
                        //we pass in the information about the current route as well as a function that 
                        //can prevent us from executing the navigation
			dispatchEvent(
				new CustomEvent('sveltekit:navigation-intent', {
					detail: {
						url: url.href,
						cancelNavigation: () => {
							startNavigation = false;
						}
					}
				})
			);

[...]

Issue #560, opened 2021-03-17
onNavigate lifecycle function


Rich-Harris:
#552 suggested a need for a lifecycle function that gets called when navigation occurs but the destination component is the same as the one we're already on (since `onMount` only runs the first time the component is mounted, and the alternative — unmounting components upon navigation — is bad for a variety of reasons).

I propose an `onNavigate` hook that lives in `$app/navigation` and looks something like this:

```js
export function onNavigate(fn) {
  let mounted = false;

  const unsubscribe = getStores().page.subscribe(() => {
    if (mounted) fn();
  });

  onMount(() => {
    mounted = true;
    fn();

    return () => {
      unsubscribe();
      mounted = false;
    }
  });
}
```

```svelte
<script>
  import { onMount } from 'svelte';
  import { onNavigate } from '$app/navigation';

  onMount(() => {
    // runs once when the page mounts
  });

  onNavigate(() => {
    // runs when the page mounts, and also whenever SvelteKit
    // navigates to a new URL but stays on this component
  });
</script>
```

As mentioned in https://github.com/sveltejs/kit/issues/404#issuecomment-800630109:

> There might be some nuance around ensuring that the callback doesn't run immediately before the component is unmounted, but you get the gist. This would work in any component (layout or otherwise) like a normal lifecycle function, including components created after the page was initially created (e.g. lazily-loaded stuff).

lowi:
We need a hook onBeforeNavigate as well. As referenced by @benmccann in [sapper#1047](https://github.com/sveltejs/sapper/pull/1047), it's very useful to be able to cancel navigation when a user is editing something and has forgotten to click "Save".
For example, I show a dialog "You didn't save your changes. Do you want to stay on the page or lose your changes?".
A basic onNavigate hook doesn't allow this kind of behavior.

PatrickG:
> We need a hook onBeforeNavigate as well. As referenced by @benmccann in [sapper#1047](https://github.com/sveltejs/sapper/pull/1047), it's very useful to be able to cancel navigation when a user is editing something and has forgotten to click "Save".
> For example, I show a dialog "You didn't save your changes. Do you want to stay on the page or lose your changes?".
> A basic onNavigate hook doesn't allow this kind of behavior.

If this will not be handled by svelte/kit, I will provide a package for it, simliar to [this](https://github.com/PatrickG/sapper-navigation-enhancer).
But I think it would be good, if this were handled by svelte/kit.
[...]

PR #2982, opened 2021-12-04
"onBeforeNavigate" navigation interceptor and "onNavigate" lifecycle function

zommerberg:
This pr closes issues #2974 and #560

This pr introduces a `onBeforeNavigate` interceptor and a `onNavigate` lifecycle function.

## onBeforeNavigate

This is a navigation interceptor that can be used to either look up the incoming URL we are going to navigate to or even prevent the navigation from completing by cancelling the navigation request. It works only for internal app navigation.

The navigation interceptor is called `onBeforeNavigate` and is being called each time we use the following methods:
- goto()
- popstate events
- internal anchor tags clicks

The interceptor can be used in the following way:
```
<script>
  import { onBeforeNavigate } from '$app/navigation';

  onBeforeNavigate(async (url) => {
	// the URL we are going to navigate to
	console.log(url);

	// cancel or allow navigation in the return statement
	return false;
   });
	
</script>
```
It can be used with and without async, also It works without a return.

## onNavigate

Additionally, the pr adds a `onNavigate`  lifecycle function that runs when the page mounts, and also whenever SvelteKit navigates to a new URL but stays on this component.

```
<script>
  import { onNavigate } from '$app/navigation';

       onNavigate(() => {
	  console.log("Navigation compelted!");
        });
	
</script>
```
---

### Tests

### Changesets


packages/kit/src/runtime/client/router.js
```diff
@@ -169,6 +197,14 @@ export class Router {
 		addEventListener('popstate', (event) => {
 			if (event.state && this.enabled) {
 				const url = new URL(location.href);
+
+				const canNavigate = triggerNavigationIntentListener(url);
+				if (!canNavigate) {
+					//"disabling" the back/forward button click by pushing the last known location
```
benmccann, line 203:
```suggestion
					// "disabling" the back/forward button click by pushing the last known location
```

packages/kit/src/runtime/client/router.js
```diff
@@ -244,6 +283,24 @@ export class Router {
 		return this.renderer.load(info);
 	}
 
+	/** @param {(navigationIntent: import('./types').NavigationIntent) => void} fn */
+	onBeforeNavigate(fn) {
+		/** @param {Event} event*/
+		function onBeforeNavEventListener(event) {
+			/** @type {any}*/
```
benmccann, line 290: it'd probably be better to create a type for this than to use `any`
```suggestion
			/** @type {any} */
```

pnpm-lock.yaml
```diff
@@ -943,6 +943,10 @@ packages:
       '@types/node': 16.11.11
     dev: true
 
+  /@types/unist/2.0.6:
```
benmccann, line 946: I wouldn't expect new packages to be added since no `package.json` was updated

packages/kit/src/runtime/client/router.js
zommerberg: You are right, I've added a CustomEvent type cast here

packages/kit/src/runtime/client/router.js
zommerberg: Do I need to do any changes here or just the spacing?

pnpm-lock.yaml
zommerberg: Correct, I've removed all changes from pnpm-lock.yaml

zommerberg:
@benmccann  I've implemented the required changes, this can be reviewed again.

packages/kit/src/runtime/app/navigation.js
```diff
@@ -16,6 +16,7 @@ export const goto = import.meta.env.SSR ? guard('goto') : goto_;
 export const invalidate = import.meta.env.SSR ? guard('invalidate') : invalidate_;
 export const prefetch = import.meta.env.SSR ? guard('prefetch') : prefetch_;
 export const prefetchRoutes = import.meta.env.SSR ? guard('prefetchRoutes') : prefetchRoutes_;
+export const onBeforeNavigate = import.meta.env.SSR ? guard('onBeforeNavigate') : onBeforeNavigate_;
```
Theo-Steiner, line 19: If you change this line to be 
```javascript
export const onBeforeNavigate = import.meta.env.SSR ? () => {} : onBeforeNavigate_;
```
you no longer need to wrap the whole thing in ``if (browser)`` since the guard won't throw an error when SSRing.

I understand why you would like this API without a browser check, as that would make it behave similar to onMount. On the other hand, this would cause onBeforeNavigate to fail silently when users tried to run it in an SSR only context. Adding a console.warn would not work for this either, since even if used correctly, it would warn the user.

I guess if we were to go the no ``if (browser)`` route, solid documentation for this API would be needed to communicate this.

Theo-Steiner:
On testing this API with the hacker news example I noticed that it doesn't seem to work on the very first navigation after either URL-navigating to or refreshing the route this is called on. I have yet to figure out why this is the case though.

packages/kit/src/runtime/app/navigation.js
zommerberg: Yes, I think it will be easier if we remove the if (browser) check. I've added a commit that removes it, thanks!

zommerberg:
> On testing this API with the hacker news example I noticed that it doesn't seem to work on the very first navigation after either URL-navigating to or refreshing the route this is called on. I have yet to figure out why this is the case though.

@Theo-Steiner  After further testing, I found a similar issue in dev mode. The cause is that the HMR that is caused by a change inside of the `onBeforeNavigate` interceptor does not invalidate the CustomEvent listener. Meaning the new one does not get initialized and the old one is not being disposed of. 

After a first change, a full reload is triggered, after subsequential change a hot reload is triggered so keep that in mind when testing (probably a vite thing). The issue occurs during the hot reload stage.

We could solve it by re-validating the CustomEvent each time a hot reload is detected or using something instead of a CustomEvent... Or we could mark the feature as experimental until the hot reload invalidates CustomEvents.

bluwy:
> How can we trigger a full page reload each time a component that has onBeforeNavigate inside has been changed?

I don't think it would be possible (or really hard), but I think the proper way is to have HMR work in the first place for it. I'm not too sure why the behaviour happens though, maybe @rixo can give their thoughts.

PS I took a brief look at the PR, seems to mostly look good. We would want to refactor some names to snake_case though for internal variables/functions, only user-facing variables/functions are camelCase. Otherwise, we may also want to have Rich's opinion on this new API.

rixo:
I identified a HMR issue that could be the root cause of the problems you're seing here.

The problem itself is that, as soon as you have modified the __layout.svelte, each time you do a full reload, we're getting a duplicated version of the __layout component in the browser.

One is the expected `__layout.svelte?t=...`, loaded by `manifest.js`. The other one is loaded via incorrect URL `__layout.svelte` with no version timestamp (it should have the same timestamp as the other, precisely to avoid this kind of duplicated modules in the browser). The faulty one comes from the generated init code in the HTML doc.

```html
		<script type="module">
			import { start } from "/.svelte-kit/dev/runtime/internal/start.js";
			start({
				...
				hydrate: {
					...
					nodes: [
						import("/src/routes/__layout.svelte"),
						import("/src/routes/[list]/[page].svelte")
					],
					...
				}
			});
		</script>
```

The `import("/src/routes/__layout.svelte")` in there doesn't get transformed. This looks like a bug/limitation in Vite (imports in .html are not transformed to HMR URLs?).

Maybe this should be fixed in Vite?

Maybe we should generate this init block in something like `.svelte-kit/generated/init.js` instead of directly in the HTML, to avoid hitting this limitation? Quick dirty testing confirms that this should work. The only thing we need to ensure is that the `.js` files imported in the HTML never changes during the lifetime of a dev server session.

zommerberg:
@rixo I can confirm that putting the `onBeforeNavigate` interceptor outside of the `__layout.svelte` stops causing the HMR issues.

How should we proceed with the PR in this case?

zommerberg:
@bluwy  I've added snake_case to the internal functions.

bluwy:
Thanks for hitting this so quickly @rixo. It indeed looks like a kit issue here. The HTML isn't processed by Vite, hence the imports weren't timestamped. Maybe we should do that instead, though I'm not very familiar to say if it's intentional.

> How should we proceed with the PR in this case?

I think we can ignore it for now as it's outside the scope of the PR. I could create an issue to track this behaviour.

zommerberg:
Then I would guess if the PR looks good it can be merged, right? 

@bluwy  You mentioned that we may also want to have Rich's opinion on this new API.

bluwy:
Yup. I'd also like to take a deeper look at this some time soon. And having Rich's opinion would be better too.

documentation/docs/05-modules.md
```diff
@@ -30,6 +30,7 @@ import { goto, invalidate, prefetch, prefetchRoutes } from '$app/navigation';
...
 - `prefetchRoutes(routes)` — programmatically prefetches the code for routes that haven't yet been fetched. Typically, you might call this to speed up subsequent navigation. If no argument is given, all routes will be fetched; otherwise, you can specify routes by any matching pathname such as `/about` (to match `src/routes/about.svelte`) or `/blog/*` (to match `src/routes/blog/[slug].svelte`). Unlike `prefetch`, this won't call `load` for individual pages. Returns a `Promise` that resolves when the routes have been prefetched.
+- `onBeforeNavigate((navigationIntent)=>void)` — a navigation interceptor that triggers before we navigate to a new route. This is helpful if we want to conditionally prevent a navigation or lookup the upcoming url.
```
benmccann, line 33:
```suggestion
- `onBeforeNavigate((navigationIntent) => void)` — a navigation interceptor that triggers before we navigate to a new route. This is helpful if we want to conditionally prevent a navigation or lookup the upcoming URL.
```

pnpm-lock.yaml
```diff
@@ -4454,4 +4454,4 @@ packages:
       eslint-plugin-import: 2.25.3_eslint@8.3.0
       eslint-plugin-svelte3: 3.2.1_eslint@8.3.0
       typescript: 4.4.4
-    dev: true
+    dev: true
```
benmccann, line 4457: shouldn't need to touch the lock file at all

benmccann:
I feel like we'd probably want to add `onBeforeNavigate` and `onNavigate` together in the same PR

zommerberg:
> I feel like we'd probably want to add `onBeforeNavigate` and `onNavigate` together in the same PR

@benmccann  Nice idea, I've updated the pr so it now includes the proposed changes and the new `onNavigate` lifecycle method.

Amine-Smahi:
Hello team,

I hope it gets merged soon.

Good luck.

packages/kit/src/runtime/client/router.js
```diff
@@ -6,6 +8,25 @@ function scroll_state() {
...
+			detail: {
+				url,
+				cancel: () => {
+					allow_navigation = false;
+				}
+			}
+		})
+	);
+	return allow_navigation;
+}
```
bluwy, line 29: If I understand correctly, the `sveltekit:navigation-intent` event is sync only. So if the user wants to do something async in `onBeforeNavigate`, it's not possible? I believe async code would be a common usecase that we'd want to support

packages/kit/src/runtime/client/router.js
```diff
@@ -6,6 +8,25 @@ function scroll_state() {
 		y: pageYOffset
 	};
 }
+/**
```
bluwy, line 11: A space above
```suggestion

/**
```

packages/kit/src/runtime/client/router.js
```diff
@@ -249,6 +289,41 @@ export class Router {
 		return this.renderer.load(info);
 	}
 
+	/** @param {() => void} fn */
+	onNavigate(fn) {
```
bluwy, line 293: This should be snake_case following the other method names in this class

packages/kit/src/runtime/client/router.js
```diff
@@ -249,6 +289,41 @@ export class Router {
...
+
+			return () => {
+				unsubscribe();
+				mounted = false;
+			};
+		});
+	}
+
+	/** @param {(navigationIntent: import('./types').NavigationIntent) => void} fn */
+	onBeforeNavigate(fn) {
```
bluwy, line 312: snake_case as well.

packages/kit/test/ambient.d.ts
```diff
@@ -18,6 +18,8 @@ declare global {
 
 	const invalidate: (url: string) => Promise<void>;
 	const prefetch: (url: string) => Promise<void>;
+	const onBeforeNavigate: (fn: { url: URL; cancel: () => void }) => void;
```
bluwy, line 21: Would be nice to reuse the `NavigationIntent` type here

packages/kit/test/test.js
```diff
@@ -122,7 +122,24 @@ async function setup({ port }) {
...
+					(navigationIntent) => onBeforeNavigate(navigationIntent),
+					navigationIntent
+				),
+
+			/**
+			 * @param {() => void} fn
+			 * @returns {Promise<void>}
+			 */
+
+			onNavigate: () => pages.js.evaluate(() => onNavigate(() => {}))
```
bluwy, line 142: Extra line space

```suggestion
			onNavigate: () => pages.js.evaluate(() => onNavigate(() => {}))
```

packages/kit/test/types.d.ts
```diff
@@ -32,6 +32,8 @@ export interface TestContext {
 		invalidate(url: string): Promise<void>;
 		prefetch(url: string): Promise<void>;
 		prefetchRoutes(urls?: string[]): Promise<void>;
+		onBeforeNavigate(navigationIntent: { url: URL; cancel: () => void }): void;
```
bluwy, line 35: We can reuse `NavigationIntent` type here too

packages/kit/types/ambient-modules.d.ts
```diff
@@ -67,6 +67,19 @@ declare module '$app/navigation' {
 	 * Returns a Promise that resolves when the routes have been prefetched.
 	 */
 	export function prefetchRoutes(routes?: string[]): Promise<any>;
+
+	/**
+	 * A navigation interceptor that triggers before we navigate to a new route.
+	 * This is helpful if we want to conditionally prevent a navigation from completing or lookup the upcoming url.
+	 */
+	export function onBeforeNavigate(
+		fn: (navigationIntent: { url: URL; cancel: () => void }) => void
```
bluwy, line 76: And the `NavigationIntent` type here

documentation/docs/05-modules.md
```diff
@@ -30,6 +37,8 @@ import { goto, invalidate, prefetch, prefetchRoutes } from '$app/navigation';
...
 - `prefetchRoutes(routes)` — programmatically prefetches the code for routes that haven't yet been fetched. Typically, you might call this to speed up subsequent navigation. If no argument is given, all routes will be fetched; otherwise, you can specify routes by any matching pathname such as `/about` (to match `src/routes/about.svelte`) or `/blog/*` (to match `src/routes/blog/[slug].svelte`). Unlike `prefetch`, this won't call `load` for individual pages. Returns a `Promise` that resolves when the routes have been prefetched.
+- `onBeforeNavigate((navigationIntent) => void)` — a navigation interceptor that triggers before we navigate to a new route. This is helpful if we want to conditionally prevent a navigation or lookup the upcoming URL.
```
bluwy, line 40: It'll be nice to explain `navigationIntent` here (or somewhere else that fits) and show how an example of canceling a navigation.

bluwy (commented):
Thanks for keeping this PR up-to-date, and your patience 😅 I have a few comments below.

packages/kit/src/runtime/client/router.js
zommerberg: @bluwy Thanks for the detailed review! 

I thought about that when I was creating the navigator and came to the conclusion that I should not make it async.

Three reasons for  that:
- CustomEvents are sync only, there is no way for me to wait until the `sveltekit:navigation-intent` returns something. And it is becoming even trickier if we have multiple `onBeforeNavigation` intents defined on one page.  Now we need to make sure that every intent was completed correctly before we navigate. 

- Making it async might even create a bad UX for the user since he will be wondering what is happening and why it he not navigating anywhere.

- There is a workaround that I've created for async tasks.
If we have to make a complex check or an API request before leaving the page we could simply:
1) Save the URL we are going to navigate to in memory or in global storage
2) Prevent navigation from happening
3) Show a Loading Spinner to the user
4) Once a task completes we use goto(savedURL) to manually leave the page after the async task is completed.
4.1) The programmatic goto() will trigger the `onBeforeNavigate` interceptor a second time but it is up to the user to create the condition that will prevent the heavy async from firing.

Example async implementation
```
let savedUrl
let heavyStuffDone = false;

onBeforeNavigate((navigationIntent) => {
        if(heavyStuffDone) return;

	savedUrl = navigationIntent.url;
	navigationIntent.cancel();

        const canNavigate = await DoSomethingHeavy()
        
        if(canNavigate){
            heavyStuffDone = true;
            goto(savedUrl.href)
        }
});
```

packages/kit/src/runtime/client/router.js
```diff
@@ -174,6 +204,14 @@ export class Router {
 		addEventListener('popstate', (event) => {
 			if (event.state && this.enabled) {
 				const url = new URL(location.href);
+
+				const allow_navigation = dispatch_navigation_intent(url);
+				if (!allow_navigation) {
+					// "disabling" the back/forward button click by pushing the last known location
+					history.pushState({}, '', this.last_known_location);
```
PatrickG, line 211: I don't know if this is the right thing to do.
The browser history gets messed up when navigating forward or when navigating back more than one history entry.

Ideally, we would write the current history entry index into the history state and navigate back to the current history entry ([like in my implementation for sapper](https://github.com/sveltejs/sapper/pull/1047/files#diff-c0287c8e7cff79fb8144be45f68c985b7f1291adf216e532a84858d70f7b6a1eR129)).

packages/kit/src/runtime/client/router.js
PatrickG: > Example async implementation
> 
> ```
> 	let heavyStuffDone = false;
> 
> 	onBeforeNavigate(async (navigationIntent) => {
> 		if (heavyStuffDone) return;
> 		navigationIntent.cancel();
> 
> 		await new Promise((r) => setTimeout(r, 2000));
> 
> 		const canNavigate = true;
> 		if (canNavigate) {
> 			heavyStuffDone = true;
> 			goto(navigationIntent.url.pathname);
> 		}
> 	});
> ```

Pushing a history entry to the browser history is suboptimal when the user tried to navigate backwards.

packages/kit/src/runtime/client/router.js
zommerberg: Just saw your implementation, there are some nice ideas, I will adjust my PR, thanks! I believe your solutions is better

packages/kit/src/runtime/client/router.js
zommerberg: > The CustomEvent is not really needed, is it?
I saw you pr form onNavigate in sapper, I like some of the ideas, I will try to remove the CustomEvent, the main issue with CustomEvent is that we cannot have multiple callbacks that talk to the parent, cannot have it truly async and have HMR issues in dev in vite. 

> Most of the time, you want to display a popup that informs the user that there are unsaved changes in a form and ask the user if they still want to navigate or not.
Yeah it is true, I was trying to say that it is always expected to show a modal/loading spinner even during heavy async stuff before navigating, meaning we would never need to wait for async stuff to complete to show something, we always show stuff before heavy stuff happens.

>Pushing a history entry to the browser history is suboptimal when the user tried to navigate backward.
This is correct, and I will fix you as you suggested in you other comments to write the history entry index into every history state and navigate back to the current history entry, meaning once this is done the async example with the custom event will still a valid option for async stuff, but I agree that it may look like a hack

packages/kit/src/runtime/client/router.js
PatrickG: As history state key, I would suggest something similar to `sveltekit:scroll`. Maybe `sveltekit:index` or `sveltekit:entry`.

Also, due to the discontinued development of Sapper, I created a package for Sapper ([sapper-navigation-enhancer](https://github.com/PatrickG/sapper-navigation-enhancer)) (and SvelteKit (not published (yet))) that does basically the same.

zommerberg: I am stuck exactly at the state key. Currently I have no idea how to prevent the URL from changing if:
- we want to go forward
- we want to go back, when it is allowed to go forward.

Currently the only thing I managed to do is the:
history.go(1) to prevent a back button click.
However if we go forward and we use history.go(-1) to prevent the forward button click, it doesn't work as expect, the URL still changes. 

I am going to look deeper into the library you mentioned

PatrickG: I don't quite understand where you're stuck.
Are you on the Svelte Discord server?

zommerberg: Yeah, add me on Discord - pavelMilkman#4025

packages/kit/src/runtime/client/router.js
bluwy: Thanks for the explanation @zommerberg. I still think preventing async isn't the most ergonomic API though. Having it async, the function should still be able to control the loading state before & after the `await`, so UX-wise I don't think it's an issue. At least for me, I think having `async` stall the navigation would match my expectation of it, instead of recommending another `goto`.

Since you mentioned about removing `CustomEvent`s, perhaps async could work again?

zommerberg: Yes, I after think about it I agree, we should remove CustomEvent and make it async

zommerberg: What would be the most ideal approach here?

zommerberg: I like the idea on how @PatrickG  did it in https://github.com/PatrickG/sapper-navigation-enhancer

bluwy: Awesome, I think there's a few ways to tackle this. I haven't looked too deep into @PatrickG's implementation, but having some sort of list of functions that are passed to `onNavigate` might work, and then running them sequentially.

packages/kit/src/runtime/client/router.js
```diff
@@ -258,6 +269,9 @@ export class Router {
 		if (!allow_navigation) return;
 
 		if (this.enabled && this.owns(url)) {
+			if (!replaceState) {
+				state['sveltekit:index'] = ++this.current_history_index;
+			}
```
PatrickG, line 274: I forgot something.
If `replaceState` is `true`, `state['sveltekit:index']` need to be set to the current history index.

```js
    state['sveltekit:index'] = replaceState ? this.current_history_index : ++this.current_history_index;
```

zommerberg: gotcha!

packages/kit/src/runtime/client/router.js
zommerberg:
```js
async trigger_on_before_navigate_callbacks(url) {
		const callbacksNavigationResult = !(
			await Promise.all(this.on_before_navigate_callbacks.map(([callback]) => callback(url)))
		).some((result) => result === false);

		return callbacksNavigationResult;
	}
```	

This is the direction I am heading to, the onBeforeNavigate callback needs to return a boolean

zommerberg: And internally this function will be called during goto,click and popstate events:

```js
const allow_navigation = this.trigger_on_before_navigate_callbacks(url);
if (!allow_navigation) return;
```

zommerberg:
@bluwy @PatrickG  I've updated the PR so the onBeforeNavigate method is async. 
It no longer uses the CustomEvent, I made it the same way as @PatrickG  did it in his sapper library.
Internally it uses callbacks and it seems to work without any issues. This can be reviewed again.

bluwy (commented):
Thanks for the documented code. Makes it easy to review. I have some requested changes below, but the overall flow looks good to me. I also made https://github.com/sveltejs/kit/issues/560#issuecomment-1000841432 for a potential syntax alternative if you're interested. With that said, I don't have the final call of the ideal API, but this PR is shaping up well for me. (Router is getting bigger though)

Pinging @Rich-Harris and @benmccann if they have any thoughts of the syntax before we move forward and make more code changes.

bluwy:
Note: I've put this into next maintainer's meeting list to discuss the path forward of it.

packages/kit/src/runtime/client/router.js
```diff
@@ -55,9 +57,20 @@ export class Router {
...
+		this.on_before_navigate_callbacks = [];
 	}
 
 	init_listeners() {
+		if (history.state['sveltekit:index'] >= 0) {
+			this.current_history_index = history.state['sveltekit:index'];
+		} else {
+			history.replaceState({ ...history.state, 'sveltekit:index': 0 }, '', location.href);
+		}
+
```
PatrickG, line 73: It might be better, to move this to [line 59](https://github.com/sveltejs/kit/blob/c86cac6553ce9ed3f65b0b62f7159422de1be88d/packages/kit/src/runtime/client/router.js#L59).

zommerberg: Screenshot_20220102_110125

Is that how you mean it? Maybe we do not need the `if (history.state['sveltekit:index'] >= 0) {` at all if we assume the `sveltekit:index` is always 0 during initialization?

PatrickG: No, that would overwrite the index on a reload.
The if-block is needed.

I just thought, it might be better, because it would be just one `replaceState` call on the initial page.

zommerberg: Does it make sense to actually initialize the `this.current_history_index` with 0 in the constructor rather than in the if statement? And also is it ok to call `history.replaceState` conditionally rather than always? Maybe we could always call `history.replaceState` and make the `sveltekit:index` dynamic? Like that:

Screenshot_20220102_112951

zommerberg: Hm maybe it does make sense to put it in a conditional statement, during reload the .replaceState will not do anything since we already have the state, right? So putting it inside a condition might be correct.

PatrickG: The end-result would be the same.
Only that there would be always one `replaceState` call.

PatrickG: > Hm maybe it does make sense to put it in a conditional statement, during reload the .replaceState will not do anything since we already have the state, right? So putting it inside a condition might be correct. But doesn't hurt I guess?

Yea, thats what meant

zommerberg: Nice, I've pushed a new commit that will apply what we've discussed

PatrickG: > Nice, I've pushed a new commit that will apply what we've discussed

Now the if statement makes no sense 😅

I think you meant it to be like:
```js
if (this.current_history_index === 0) { ... }
```

Or 
```js
if (history.state?.['sveltekit:index'] === undefined) { ... }
```

zommerberg: correct, fixed

zommerberg: oh wait not really

zommerberg: now it is fixed

zommerberg:
Closing this in favor of #3293

PR #2986, opened 2021-12-05
feat: rework router scroll handling

### LINKED ISSUE #2794: Scroll issue when navigating between pages (follow-up)
### Describe the bug

This is a follow-up to #2760, which was closed #2761. I have updated my replication case to v`next.196` which has the fix from #2761. The issue is partially fixed but now there's an issue when using the back button.

### Reproduction

1. Go to https://gavinr.github.io/sveltekit-scroll-issue/
2. Scroll down 75% on the page, click link ~400
3. Next page loads at the top (As expected! This was fixed by #2761)
4. Scroll to the BOTTOM of this page.
5. Click "Back" button in the browser
    - Expected: page loads where you just were (about 75% down with item 400 in view)
    - Actual: page loads scrolled near the TOP of the page

### Logs

_No response_

### System Info

```shell
System:
  I have tested this on multiple browsers (Firefox, Chromium) on multple OSs (Windows 10, Linux) - it happens on all of them.
Binaries:
  Node: 14.18.1 - /usr/bin/node
  npm: 8.1.1 - /usr/bin/npm
npmPackages:
  @sveltejs/kit: next => 1.0.0-next.196 
  svelte: ^3.42.6 => 3.44.1
```

### Severity

blocking an upgrade

### Additional Information

Thank you for your time supporting this project!

bluwy:
Fixes #2794

---

The way the router currently works isn't great, the code flow when clicking a link goes:

1. Render new route and call `onMount`
2. Determine if `onMount` scrolls the page, if not, we apply kit's scrolling

The heuristic used in no2 is sketchy, and does not always determine correctly, causing many issues in the past.

---

With this PR, it goes:

1. Render new route
2. Apply kit's scroliing
3. Call `onMount`

Which would now work reliably.

---

This PR achieves that by using a preprocessor for route components, that injects an `action` which is [called immediately after mount and before `onMount`](https://svelte.dev/repl/e0e911d8e15c4646b7c1caa4e8bbc81f?version=3.44.2).

Note: This adds a new `@sveltejs/kit/router` exports, which is only used internally.

---

Alternatively, @dominikg also suggested a new `export const autoscroll = false` flag, and expose kit's scrolling logic for advanced usage. I think that would be the last resort if this PR doesn't make things work OOTB.

### Tests

### Changesets


bluwy:
Looks like there's some sort of race condition in windows (again 😢)

bluwy:
Marking this as draft until I figure out the windows issue, but it's likely an issue within the test than the source implementation. This PR can still be reviewed at the meantime.

packages/kit/src/core/route_preprocessor.js
```diff
@@ -0,0 +1,38 @@
+import path from 'path';
+import MagicString from 'magic-string';
+
+/**
+ * Injects __handle_route of `@sveltejs/init-route` to handle scrolling when component mounts
```
benmccann, line 5: I did Ctrl+f for `init-route` and don't see anything. Was it renamed or accidentally left off the PR?

bluwy: Oops yeah I left that one while refactoring some stuff. Should be `@sveltejs/kit/router`.

bluwy:
```suggestion
 * Injects `use:__handle_route` from `@sveltejs/kit/router` to handle scrolling when route component mounts
```

packages/kit/src/core/route_preprocessor.js
```diff
@@ -0,0 +1,55 @@
...
+		markup({ content, filename }) {
+			if (
+				filename &&
+				filename.startsWith(routesPath) &&
+				config.extensions.some((ext) => filename.endsWith(ext)) &&
+				!path.basename(filename).startsWith('_')
+			) {
+				const s = new MagicString(content);
+
+				const svelte_body_regex = /<svelte:body[^>]*\/>|<svelte:body.*<\/svelte:body>/;
```
benmccann, line 24: do we need to do this in a preprocessor or could we just put it in the root component?

https://github.com/sveltejs/kit/blob/76d26ecaf371c976330263c6da1464316165a05b/packages/kit/src/core/create_app/index.js#L145

bluwy: Interesting, I'm not too familiar with that but it looks like it's only mounted once throughout the app lifetime. The scroll handling would need to happen whenever a new route component mounts. Nonetheless, I'm still having trouble getting this PR done. The `svelte:body` approach isn't as robust as I thought since it would fail if the developer already used it in the markup. I'd need some more time to find alternatives.

f-elix: @bluwy Couldn't the action could return an `update` function that runs on every navigation?

bluwy: We'd need the scrolling to happen exactly between "render new route" and "onMount". I'm not sure if the `update` function guarantees that. I haven't had time to revisit this lately, and my solution for this PR is flawed, so unfortunately it'll be stalled for a while more.

Rich-Harris:
Oof. This is clever, but I _really_ don't like using a preprocessor for this. I'll admit I don't completely have my head round the problem (there's a lot of PRs and issues to follow, and I haven't closely followed all the changes to scroll behaviour) but it seems like the whole reason we're in this pickle is that we're trying to disable scrolling when the page scrolls programmatically in `onMount`, is that right?

If so, I think that's a bad reason to go to these lengths. SvelteKit being in control of scroll position is a feature, not a bug, and programmatically scrolling on navigation is generally something we should discourage anyway (it's disorienting). If someone really has a good reason to do it, and we want to accommodate it, then it should be done explicitly somehow.

Straw man based on the tests that fail when I comment out all the `no_scroll_happened` stuff (would like to come up with a better/more concise name than `disableScrollHandling`):

```svelte
<script>
  import { disableScrollHandling } from '$app/navigation';

  onMount(() => {
    disableScrollHandling();
    document.getElementById('abcde')?.scrollIntoView();

    // disableScrollHandling only applies to current navigation,
    // and errors if it is called at any other time
    setTimeout(() => {
      disableScrollHandling(); // Error: disableScrollHandling was called outside a navigation event
    });
  });
</script>
```

```svelte
<script>
  import { disableScrollHandling } from '$app/navigation';

  const focusAndScroll = /** @param {HTMLInputElement} node */ (node) => {
    disableScrollHandling();

    node.focus();
    node.scrollIntoView();
  };
</script>

<div>They (don't) see me scrollin'...</div>
<div style="background-color: peru;">
  <label for="input">Focus! </label>
  <input use:focusAndScroll id="input" type="text" />
</div>
<div style="background-color: teal;">They (not) focusin'</div>
```

For now this would presumably apply during component mounting, but it could be extended to a future `onNavigate` lifecycle function.

bluwy:
> it seems like the whole reason we're in this pickle is that we're trying to disable scrolling when the page scrolls programmatically in `onMount`, is that right?

That is correct. Ideally we won't also have to disable any scrolling if we can do the scroll before `onMount()` is called (and after component is rendered), which is what the PR tries to achieve, so the end-user doesn't have to manually apply `disableScrollHandling()` consciously and it'll just work.

However, I still haven't got around making this implementation robust, and I agree things are starting to get iffy here. Will be closing this.

> disableScrollHandling

This looks like a good compromise. It would be tricky though if the action comes from a third-party package (rare case), but the user can always manually call the function `onMount`. Which leads to another idea we had:

```svelte
<script context="module">
  export const autoscroll = false; // disables router scroll per-page basis
</script>
```

I think both ways have its pros and cons (imperative/declarative), but if we discourage scrolling on mount, perhaps the `export const` approach is still fine.

Issue #2660, opened 2021-10-21
Provide workbox integration or documentation


ninsky:
### Describe the problem

I'm trying to integrate [Google Workbox](https://developers.google.com/web/tools/workbox) which seems to be not that easy (at least for me as a newbie). I've found [some examples](https://github.com/FunMiles/SveltekitPWA/blob/main/src/service-worker.ts) but it seems that those don't integrate Workbox completely as there's no workbox logging on the console and also lighthouse audits don't run till the end.

### Describe the proposed solution

I'm pretty new to SvelteKit and have no knowledge about the framework architecture. I think it would help to have some kind of documentation how such an integration could be done. Even better would be a SK configuration option to enable Workbox. 

### Alternatives considered

According to [this comment](https://github.com/sveltejs/kit/issues/1792#issuecomment-872205124) the Workbox integration of kit-sw-workbox seems to be unintuitive but maybe it's "good enough". Maybe it would also make sense to support[ these guys](https://github.com/svelte-society/sveltesociety.dev/issues/69) to provide a recipe to integrate Workbox.

### Importance

For me using Workbox is more critical that the underlying JS framework. But I would love to use SK. 

### Additional Information

To me the importance is very high as I want to develop a PWA based on Workbox with offline capabilities.

ninsky:
@userquin I'm still a bit struggling how to import a workbox-sw object into my Service Worker. By doing some more research I've found this library ([kit-sw-workbox](https://www.npmjs.com/package/kit-sw-workbox)) which helped me to (at least) set up a Service Worker. (I've even asked the same question there [how to import a workbox object](https://github.com/Anyass3/kit-sw-workbox/issues/2)). I just want to let you know about this libray. Maybe it's a help for your work on this issue.

userquin:
@ninsky 

Are you using `vite-plugin-pwa` or `sveltekit service worker module`? If you're using latter I cannot help you, I think `sveltekit` will not support it. If you want to see how you can use `importScripts` you can take a look at this: https://github.com/antfu/vite-plugin-pwa/issues/33#issuecomment-797814153 or this https://github.com/antfu/vite-plugin-pwa/issues/35#issuecomment-796370356 (example using workbox from CDN).

If you are using `vite-plugin-pwa`:

If you want to build your own service worker, take a look at these service workers: 
- auto update behavior: https://github.com/antfu/vite-plugin-pwa/blob/main/examples/sveltekit-pwa/src/claims-sw.ts
[...]

Issue #2367, opened 2021-09-04
Disable service-worker.js selectively on page_config


acoyfellow:
### Describe the problem

I've got a use case where I'd like to disable the `service-worker.js` on only specific page levels. I want to exclude it on only 2 routes in my app.

### Describe the proposed solution

```
<script context="module">
  export const service_worker = false;
</script>
```

### Alternatives considered

Having dual svelte-kit apps, both sharing about 99% of the codebase, but one does not have a `service-worker.js`

### Importance

would make my life easier

### Additional Information

I've not yet fully understood how to submit the pull request (my understanding of the svelte-kit internals are limited), but i was able to simply modify `...node_modules/@sveltejs/kit/dist/ssr.js`: (i find it [here](https://github.com/sveltejs/kit/blob/a7f4ca1283175d20c7852b61ebfc9772bfb15b2e/packages/kit/src/runtime/server/page/render.js#L147))

find: 
`if (options.service_worker) {`

replace with:
`if (options.service_worker && !!page_config.service_worker) {`

To enable 
```
<script context="module">
  export const service_worker = false;
</script>
```

And we could just add one more section to the page-config options inside of the docs.

** edit: my proposed solution does not do what I expected it to - I obviously am not understanding the svelte-kit internals enough (yet)

benmccann:
Would https://github.com/sveltejs/kit/pull/2281 help you? You could disable the automatic service worker registration and then register yourself. I'd prefer that to adding more options

acoyfellow:
Yes, I think so, if i'm understanding that pull correctly:

I'd be able to set `serviceWorker.register` to false in my config, then just opt-in each page that i *do* want it on

acoyfellow:
Looks like https://github.com/sveltejs/kit/pull/2281 might be ready for approval.. I'm still looking for a solution here

benmccann:
https://github.com/sveltejs/kit/pull/2988 has been merged which adds an option to disable the existing service worker registration, so I think you could now add your own to address this and I'll go ahead and close this. Let me know if there's something that would still be required

Issue #1749, opened 2021-06-24
Provide the "register" object returned by navigator.serviceWorker.register


Myrmod:
**Is your feature request related to a problem? Please describe.**
I am currently trying to implement push notifications within my SvelteKit application.

Usually I would go with something like this:

```js
const registration = await navigator.serviceWorker.register('/service-worker.js')
if (!import.meta.env.VITE_VAPID_PUBLIC_KEY) {
      console.error('VITE_VAPID_PUBLIC_KEY is not defined.')

      return
}

// create subscribtion
const subscription = registration.pushManager.subscribe({
      userVisibleOnly: true,
      applicationServerKey: `${import.meta.env.VITE_VAPID_PUBLIC_KEY}`,
})

// send push notification
await fetch('/subscribe', {
      method: 'POST',
      body: JSON.stringify(subscription),
      headers: {
        'content-type': 'application/json',
      },
})
```

**Describe the solution you'd like**
Currently we have `$service-worker` available. So I would suggest adding the `registration` to this object. Not sure if this is possible though.

**Describe alternatives you've considered**
I couldn't find a way to do add push notifications so far. So this is the only way I would suggest.

**How important is this feature to you?**
Some applications need a push notification. So if I cannot add one, the application is useless.

Best regards

ignatiusmb:
The `$service-worker` module is only available inside a service worker file, would this be used outside of that too?

benmccann:
@ignatiusmb the registration of a service worker is separate and doesn't rely directly on the `$service-worker` JS. It happens here:

https://github.com/sveltejs/kit/blob/a5c058a932bf9dbf2e67d0f07f0bc2d2ab2ca28f/packages/kit/src/runtime/server/page/render.js#L147

@Myrmod would https://github.com/sveltejs/kit/pull/2281 be enough to consider this issue solved? It lets you disable the registration. I'd hope that you could then register it on your own which would let you access the return value

ignatiusmb:
Yes, so I think the proposed solution for having the `registration` in `$service-worker` wouldn't really work here.

benmccann:
https://github.com/sveltejs/kit/pull/2988 has been merged which adds an option to disable the existing service worker registration, so I think you could now add your own to address this and I'll go ahead and close this. Let me know if there's something that would still be required

alanxp:
is it possible to get the ServiceWorkerRegistration without manually registering our own file, since all i need is registration to get to pushmanager and work with push notifications

[...]

PR #2988, opened 2021-12-06
[feat] allow to disable service worker auto inject

SomaticIT:
This PR allows to disable the automatic injection of the service worker.
This allows developpers to code a custom logic to register service workers.

It is particularly useful when using `workbox` to manage the service worker.

### Workaround

As a workaround, I actually use a script that I run after `svelte-kit build` to strip the injected service worker from `index.html` but it only work for SPA builds. In SSR environments, a custom `hook` may be able to strip it.

But it's not a viable solution since it can break anytime the injected code is changed on the svelte-kit side.

### Usage

In your `svelte.config.js`, you just have to set the following option:

```js
const config = {
  // ... your config
  serviceWorker: {
    inject: false
  }
};
```

### Fixes

- #922 (you can specify the scope manually)
- #1749 (you can get the registration)
- #2367 (you can set the scope to ignore some parts of the site)
- #2660 (we can easily register workbox manually)

### Tests

### Changesets

### Todo

- [X] Write documentation
- [X] Add a test case
- [ ] Once validated, add changeset


SomaticIT:
I added documentation and test.

The CI fails on windows due to a timing issue (not related with this PR).
Maybe a contributor can rerun the windows checks.

I think this is ready for review.

benmccann:
This generally seems reasonable to me and similar to https://github.com/sveltejs/kit/pull/2281. My main question for the other maintainers would be about what we should call the option name

SomaticIT:
@benmccann, thank you for your reply.

1/ Regarding #2281, you're right, I found this PR after implementing the feature. However, this PR takes a simpler approach: if you want to customize the service worker behavior, just disable automatic injection and do what you want 😁. It means less impact on the kit code (only 3 lines impacted) and less maintenance.

2/ Regarding the option name, I think there are two possibilities:
- `inject`: to explain the fact that the code is injected in your page. *This was my first choice but I'm open to change*
- `register`: to explain that the behavior of the code which actually register the service worker.

Let met know your preference, I will adapt the PR to your choice.

SomaticIT:
Hello,

I would like to know if it's possible to review this PR?
It is a very simple modification (3 lines of code impacted). 

I'm using a custom script to workaround this issue but I'm creating new projects every week and it will become very hard to maintain in the future.

Thank you in advance

Rich-Harris (approved):

Rich-Harris:
Thank you! I renamed the option to `register`, moved the test app (`sw-inject` -> `options-2`, since it serves the same function as `options` and would have been combined if the options under test weren't incompatible) and added a changeset

SomaticIT:
Thank you for your intervention!
I totally agree with all your changes.

Issue #1046, opened 2021-04-15
Method overrides


Rich-Harris:
I could have sworn we'd already talked about this at some point, but I guess not.

**Is your feature request related to a problem? Please describe.**
Bewilderingly, the only valid `<form>` methods are GET and POST — in other words `<form method="put">` is invalid, and will result in `form.method === 'get'`.

A popular workaround is to use a method override query string parameter, like so:

```html
<form method="post" action="/todos/{id}?_method=put">
  <!-- form elements -->
</form>
```

(As evidence for the validity of this approach, there's an official Express middleware for it: https://www.npmjs.com/package/method-override).

It would be nice if such a thing existed in SvelteKit as well, since we want to encourage people to build apps that work with JavaScript disabled, and that's only really possible if you can use PUT and DELETE alongside GET and POST without having to resort to `fetch`.

**Describe the solution you'd like**
Since the `?_method=put` cowpath already exists, I propose that we bake it into SvelteKit, while making it configurable:

```js
// svelte.config.cjs
module.exports = {
  kit: {
    methodOverride: {
      key: '_method',
      allowed: ['post', 'put', 'delete']
    }
  }
};
```

Then, when any request that comes in where `query.has(override)`, we change the request method before passing it to `handle`.

**How important is this feature to you?**
It would be possible to do in userland...

```js
export function handle({ request, render }) {
  return render({
    ...request,
    method: request.query.get('_method') || request.method
  });
}
```

...but I think this is the sort of thing that belongs in the framework.

Conduitry:
Looking at the readme of that middleware, are they actually blessing that particular query parameter name? The only thing I saw blessed was a particular HTTP header. Or are you saying this would be off by default and people would need to explicitly opt in and say what query parameter they want to use?

Rich-Harris:
They're blessing it in the sense that it's in the README. I've seen it used in a couple of different places

![image](https://user-images.githubusercontent.com/1162160/115113709-a095a700-9f59-11eb-8aba-43400ad606d7.png)

I'm suggesting that `_method` would be the default, but you could change it to whatever:

```js
// svelte.config.cjs
module.exports = {
  kit: {
    methodOverride: {
      key: 'httpverb',
      allowed: ['patch']
    }
  }
};
```

Conduitry:
[...]

PR #2989, opened 2021-12-06
Add functionality to override http methods (issue #1046)

am283721:
See: https://github.com/sveltejs/kit/issues/1046

In short: Given that the only valid `<form>` methods are GET and POST, functionality is provided to override this limitation and allow the use of other HTTP verbs. This is particularly helpful when ensuring your application works even when javascript fails or is disabled.

I could see this potentially needing to be an RFC due to either the impact of the change or the differences in opinion about how it should be implemented or what the default config values should be. Let me know if this is the case.

### Tests

### Changesets


pnpm-lock.yaml
```diff
@@ -30,7 +30,7 @@ importers:
       '@sveltejs/eslint-config': github.com/sveltejs/eslint-config/9a7d728e03ac433e5856a6e06775c17ee986d641_d9525e585486a10ae6317f3f61e9597f
```
benmccann, line 30: you shouldn't need to modify the lockfile. there's a merge conflict since this file is being changed, so it'd be nice to drop it from the PR

documentation/docs/01-routing.md
```diff
@@ -161,6 +161,45 @@ The `body` property of the request object will be provided in the case of POST r
 - Form data (with content-type `application/x-www-form-urlencoded` or `multipart/form-data`) will be parsed to a read-only version of the [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) object.
 - All other data will be provided as a `Uint8Array`
 
+#### HTTP Method Overrides
+
+Given that the only valid `<form>` methods are GET and POST, functionality is provided to override this limitation and allow the use of other HTTP verbs. This is particularly helpful when ensuring your application works even when javascript fails or is disabled.
+
+- Disabled by default to prevent unintended behavior for those who don't need it
```
benmccann, line 168: this would be a breaking change. Rich should weigh in, but I would guess his desire would likely be to keep it enabled by default

documentation/docs/01-routing.md
```diff
@@ -161,6 +161,45 @@ The `body` property of the request object will be provided in the case of POST r
...
+- For security purposes, the original method on the form must be `POST` and cannot be overridden with `GET`
```
benmccann, line 169: this is a bit vague. it'd be better if we can say something more specific
```suggestion
- To protect against CSRF attacks, the original method on the form must be `POST` and cannot be overridden with `GET`
```

am283721: Agreed. Although I'm not sure if this is for CSRF protection or for protection against other attack vectors (cache poisoning was the example given: https://github.com/sveltejs/kit/issues/1046#issuecomment-949382948).

Unfortunately I lack expertise in this field. Maybe @Prinzhorn could approve your suggestion or provide a more detailed/accurate statement?

pnpm-lock.yaml
am283721: Reverted to the previous lockfile. It appears the build is failing due to the new config options.

`Unexpected option config.kit.methodOverride`

I'm assuming Netlify is building the `create-svelte` apps using the version of sveltekit provided by npm, and not this version of the kit? How is this typically handled?

documentation/docs/01-routing.md
benmccann: Ah, ok. I didn't realize it was implemented directly in the sample app. Thanks for the clarification

documentation/docs/14-configuration.md
```diff
@@ -124,6 +130,19 @@ export default {
...
+See [HTTP Method Overrides](#routing-endpoints-http-method-overrides). An object containing zero or more of the following:
+
+- `enabled` — set to `true` to enable method overriding
+- `key` — query parameter name/field name to use for passing the intended method value
+- `allowedMethods` - array of HTTP methods that can be used when overriding the original request method
+- `strategy`
+
+  - `'both'` — (default) will look for the override key in both the list of query parameters and the form fields
+  - `'url_parameter'` — only allow overriding via a query parameter
+  - `form_data` — only allow overriding via a form field
```
Prinzhorn, line 144: Missing quotes

packages/kit/src/runtime/server/index.js
```diff
@@ -40,6 +41,31 @@ export async function respond(incoming, options, state = {}) {
...
+		if (
+			['both', 'form_data'].includes(strategy) &&
+			request.body instanceof ReadOnlyFormData &&
+			request.body.has(method_key)
+		) {
+			new_request_method = (request.body.get(method_key) || request.method).toUpperCase();
+		}
+
+		if (['both', 'url_parameter'].includes(strategy) && incoming.query.has(method_key)) {
+			new_request_method = (incoming.query.get(method_key) || request.method).toUpperCase();
```
Prinzhorn, line 57: `allowedMethods` should take care of malformed (empty) methods already, so there should be no need for a fallback to `request.method` (this would just mask a bug in your code where you don't set the method correctly)

packages/kit/src/runtime/server/index.js
```diff
@@ -40,6 +41,31 @@ export async function respond(incoming, options, state = {}) {
...
+		}
+
+		if (
+			new_request_method &&
+			allowedMethods.includes(new_request_method) &&
+			new_request_method !== 'GET'
```
Prinzhorn, line 63: I think in addition to checking `GET` here it should also be validated in the `allowedMethods` validator. Because otherwise a user might want to add `GET` in `allowedMethods` and wonder why it is silently ignored. The validator would then have a descriptive error explaining that `GET` is not allowed ever.

But then it probably doesn't need to be explicitly checked here, when it will never be in the allowed list.

Maybe also throw `HEAD` in the mix? At this point I'm not sure if these cases are needed if `allowedMethods` exists.

documentation/docs/01-routing.md
Prinzhorn: I don't think we need to be that specific and it's also not just a security thing. GET has vastly different semantics from POST, PUT etc. We should focus on why the feature exist, what problem it solves. I don't see any problem being solved by changing the method from/to GET.

I'd shorten the paragraph and the two bullet points to something like:

---

In contrast to `fetch` the only valid methods for a `<form>` are `GET` and `POST`. Svelte allows you to override the `<form>` method to workaround this limitation if you need to. That way your application can transparently work when JavaScript fails or is disabled by using `fetch` and `<form>` interchangeably with the same endpoint.

documentation/docs/01-routing.md
```diff
@@ -161,6 +161,45 @@ The `body` property of the request object will be provided in the case of POST r
...
+- There are 2 different ways to specify a method override strategy
```
Prinzhorn, line 170: There is _one_ way to specify the override strategy. But there are two strategies.

documentation/docs/01-routing.md
```diff
@@ -161,6 +161,45 @@ The `body` property of the request object will be provided in the case of POST r
...
+			key: '_method',
+			allowedMethods: ['PUT', 'PATCH', 'DELETE'],
+			strategy: 'both'
+		},
+	}
+};
+```
+
+```html
+<form method="post" action="/todos/{id}?_method=put">
```
Prinzhorn, line 190: I don't have hard feelings for `put` being lower case, but I'd slightly prefer if it would just be the correct case and also remove `toUpperCase` in the implementation?

documentation/docs/14-configuration.md
am283721: Good catch

packages/kit/src/runtime/server/index.js
am283721: Agreed. I'll remove the fallback

documentation/docs/01-routing.md
am283721: I'll update the example to use upper case, however I wouldn't feel good about removing the `toUpperCase` in the implementation. Mostly because the create-svelte default template currently uses this method override technique and all of the url parameters in the template are lower case. This could throw off a lot of people when trying to upgrade Sveltekit.

documentation/docs/01-routing.md
am283721: I like your version of why this feature exists but I'm unsure about removing those 2 bullet points. It should be pointed out that the feature needs to be enabled by the user. The defaults are noted later in the configuration section but they might not notice that. And explaining that the form method must be POST and cannot be overridden with GET will prevent users from using this feature incorrectly and forcing them to figure out what is wrong when they receive build errors or their app doesn't function as expected.

documentation/docs/01-routing.md
am283721: Will update to:

> - There are 2 different strategies available for overriding form methods

documentation/docs/01-routing.md
Prinzhorn: I agree about documenting everything. But for me docs are a last resort (the actual text, anything beyond skimming through code examples) when things already went wrong and I need to figure out why. But we can do much better. During dev when we see `_method` and `methodOverride` is disabled we can _tell_ the developer that it needs to be enabled. Same if we see `_method` with something other than `POST`. Same for all other cases that are currently silently ignored. If the method is not in `allowedMethods` arguable it should even return a 400 in production. Or at the very least during dev it should _tell_ you. We have the knowledge, let's not make the user run into unexpected behavior (e.g. silently ignoring `_method` for `GET`). Instead let's fail as loud as possible so they don't need to open their browser to actually read the docs or search though `/issues`. Let them stay in the zone and be like "oh, I need to set `enabled: true`, gotcha, thanks friendly error message".

packages/kit/src/runtime/server/index.js
am283721: I've updated the validator to check for `GET` in `allowedMethods` and removed the check for `GET` in this file.

documentation/docs/01-routing.md
am283721: Agreed. Can anyone tell me how compile/build warnings are generated in Sveltekit?

And I'm on the fence regarding returning a 400 if the method isn't in allowed methods, but I agree silently ignoring it is probably not the correct action. I'd like to handle it similar to other scenarios in Sveltekit.

packages/kit/src/runtime/server/index.js
```diff
@@ -43,6 +44,39 @@ export async function respond(incoming, options, state = {}) {
 		locals: {}
 	};
 
+	if (options.method_override.enabled && request.method.toUpperCase() === 'POST') {
+		const { strategy = '', key: method_key = '', allowedMethods = [] } = options.method_override;
```
Rich-Harris, line 48: No need to provide defaults — once config has gone through validation, whatever was specified as the default in [src/core/config/options.js](https://github.com/sveltejs/kit/blob/master/packages/kit/src/core/config/options.js) is guaranteed to be present if no value was provided

Rich-Harris (changes requested):
Thanks, this is great. 

It's always a good idea to avoid things like 'both', since it's not future-proof — if we decided we wanted to support method overrides via headers, for example (which is sometimes used to get around firewalls that disallow non-GET/POST requests), then 'both' would need to become 'all', which would be a breaking change.

If we implement https://github.com/sveltejs/kit/issues/70#issuecomment-1001222642, we'll no longer be able to read form data before selecting a handler, which means that the hidden input method won't work in future. 

What I'd suggest, therefore, is something like the following defaults...

```js
methodOverride: {
  allowed: [],
  header: null,
  parameter: '_method'
}
```

...which would typically be used like so:

```js
methodOverride: {
  allowed: ['put', 'delete']
}

// or
methodOverride: {
  allowed: ['put', 'delete'],
  header: 'x-http-method-override'
}
```

`enabled` is unused here because it's arguably redundant — it's enabled if you specify allowed methods, disabled otherwise.

packages/kit/src/runtime/server/index.js
am283721: Thanks for the review. I agree regarding `enabled`, will remove.

It looks like you're also proposing changing `key` to `parameter` and `allowedMethods` to `allowed`. Those make sense, especially if we're getting rid of the hidden input method.

And do we want to include supporting method overrides via headers in this PR, or save it for a future enhancement?

Rich-Harris: For simplicity let's skip `header` for now, it'll be an easy thing to add later. Thank you!

Rich-Harris (approved):

Rich-Harris:
Wonderful, thank you!

brandonleboeuf:
I see that this change has been made in the repo, but they are not present when `npm init svelte@next` is run.

Getting these errors still 
Screen_Shot_2022-01-15_at_4 57 05_PM

am283721:
@Rich-Harris @benmccann 

The changeset for this PR includes both the @sveltejs/kit and create-svelte packages, but it looks like the changes to the create-svelte template aren't included when scaffolding a new project a thus the new config settings aren't copied over.

I'm new to using changesets, so it may have been something I did wrong. Could you help take a look at this? Thank you

geoffrich:
@am283721 Thanks for the report, I don't think this is related to the changeset. See the issue I opened above.

PR #2990, opened 2021-12-06
[fix] rename dotfiles to avoid template problems

### LINKED ISSUE #2952: [bug] `npm init svelte@next` - some files missing
### Describe the bug

I don't get any warning using npm 6 and Node 12 on Ubuntu. If I create an `.npmrc` with `engine-strict=true` then it will fail. We try to do that today:

https://github.com/sveltejs/kit/blob/master/packages/create-svelte/templates/skeleton/.npmrc
https://github.com/sveltejs/kit/blob/master/packages/create-svelte/templates/default/.npmrc

But those files don't get created for some reason

### Reproduction

If I run `npm init svelte@next my-test-app` I don't see an `.npmrc` file being created in the new app

### Logs

_No response_

### System Info

```shell
System:
    OS: Linux 5.13 Ubuntu 21.10 21.10 (Impish Indri)
  Binaries:
    Node: 16.4.0 - ~/.nvm/versions/node/v16.4.0/bin/node
    npm: 7.18.1 - ~/.nvm/versions/node/v16.4.0/bin/npm
```

### Severity

annoyance

### Additional Information

_No response_

Conduitry:
Fixes #2952.

Rather than adding another special case for `.npmrc` like there already is for `.gitignore`, this simply renames all dotfiles to instead begin with `DOT-` when the `create-svelte` package is being built, and un-renames them when bootstrapping a new project with the CLI. This doesn't currently handle dotfiles other than at the top level of a particular template, but we don't have any of those. I didn't include that for simplicity, but if other folks think that would be a good idea to handle now while we're thinking of it, I'd be amenable to that.

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -134,7 +134,7 @@ async function main() {
  */
 function write_template_files(template, typescript, name, cwd) {
 	const dir = dist(`templates/${template}`);
-	copy(`${dir}/assets`, cwd, (name) => name.replace('gitignore', '.gitignore'));
+	copy(`${dir}/assets`, cwd, (name) => name.replace('DOT-', '.'));
 	copy(`${dir}/package.json`, `${cwd}/package.json`);
 
 	const manifest = `${dir}/files.${typescript ? 'ts' : 'js'}.json`;
--- a/.changeset/odd-ligers-march.md
+++ b/.changeset/odd-ligers-march.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+include missing .npmrc in templates
--- a/packages/create-svelte/scripts/build-templates.js
+++ b/packages/create-svelte/scripts/build-templates.js
@@ -56,7 +56,7 @@ async function generate_templates(shared) {
 					contents
 				});
 			} else {
-				const dest = path.join(assets, name).replace('.gitignore', 'gitignore'); // npm does wacky stuff to gitignores
+				const dest = path.join(assets, name.replace(/^\./, 'DOT-'));
 				mkdirp(path.dirname(dest));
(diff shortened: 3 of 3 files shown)


ignatiusmb (approved):

bluwy (approved):

dummdidumm:
I don't see the renamed dot-files in this PR

Conduitry:
They're not renamed in this repo, and they shouldn't need to be. They're renamed as they're copied into the `dist` folder that gets packaged up, and then renamed back by the CLI when they're copied to the user's directory. Try this out locally and examine the contents of `npm pack` and the results of running the CLI.

dummdidumm (approved):
Ah now I understand, thanks!

packages/create-svelte/bin.js
```diff
@@ -134,7 +134,7 @@ async function main() {
  */
 function write_template_files(template, typescript, name, cwd) {
 	const dir = dist(`templates/${template}`);
-	copy(`${dir}/assets`, cwd, (name) => name.replace('gitignore', '.gitignore'));
+	copy(`${dir}/assets`, cwd, (name) => name.replace('DOT-', '.'));
```
dreitzner, line 137: Maybe use regex here too, for the off chance that there will ever be any file containing `DOT-` in it's name...
```suggestion
	copy(`${dir}/assets`, cwd, (name) => name.replace(/^DOT-/, '.'));
```

dreitzner (approved):
LGTM

packages/create-svelte/bin.js
dreitzner: `DOT-` could also be extracted as constsant into `packages/create-svelte/utils.js`

benmccann:
I didn't review, but with four approvals, this seems safe to merge :smile:

PR #2995, opened 2021-12-06
fix: update `sveltekit:prefetch` mouse detection

### LINKED ISSUE #2929: sveltekit:prefetch not working
### Describe the bug

When I move mouse pointer over a `<a sveltekit:prefetch href="/page2">`, no chunks are preloaded.

bug-demo

### Reproduction

I created a simple repository to reproduce the issue: https://github.com/ww9/sveltekit-prefetch-test

It was created with `npm init svelte@next test` and:

- type: skeleton demo app
- TypeScript: yes
- ESLint: yes
- Prettier: yes

Then added some links with `sveltekit:prefetch` but it doesn't work.

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19043
    CPU: (12) x64 Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz   
    Memory: 21.07 GB / 31.84 GB
  Binaries:
    Node: 16.6.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.17 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.1.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (96.0.1054.34)  
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.3 
    @sveltejs/kit: next => 1.0.0-next.201 
    svelte: ^3.44.0 => 3.44.2
```

### Severity

annoyance

### Additional Information

Svelte is amazing ❤

borntofrappe:
dispatches a custom event so that `event.composedPath` returns an array of objects - fixes #2929

### Tests

### Changesets

### DIFF
--- a/.changeset/lazy-ways-swim.md
+++ b/.changeset/lazy-ways-swim.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+dispatch custom event to retrieve composed path
--- a/.changeset/lazy-ways-swim.md
+++ b/.changeset/lazy-ways-swim.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-dispatch custom event to retrieve composed path
+fix `sveltekit:prefetch` mouse detection
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -109,7 +109,7 @@ export class Router {
 		const handle_mousemove = (event) => {
 			clearTimeout(mousemove_timeout);
 			mousemove_timeout = setTimeout(() => {
-				event.target.dispatchEvent(new CustomEvent('trigger_prefetch', { bubbles: true }))
+				event.target.dispatchEvent(new CustomEvent('trigger_prefetch', { bubbles: true }));
 			}, 20);
 		};
 
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -109,12 +109,13 @@ export class Router {
 		const handle_mousemove = (event) => {
 			clearTimeout(mousemove_timeout);
 			mousemove_timeout = setTimeout(() => {
-				trigger_prefetch(event);
+				event.target.dispatchEvent(new CustomEvent('trigger_prefetch', { bubbles: true }))
 			}, 20);
 		};
 
 		addEventListener('touchstart', trigger_prefetch);
 		addEventListener('mousemove', handle_mousemove);
+		addEventListener('trigger_prefetch', trigger_prefetch);
 
 		/** @param {MouseEvent} event */
 		addEventListener('click', (event) => {
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -109,9 +109,8 @@ export class Router {
 		const handle_mousemove = (event) => {
 			clearTimeout(mousemove_timeout);
 			mousemove_timeout = setTimeout(() => {
-				if (!event.target) return;
 				// ensure event.composedPath() in find_anchor(event) returns an array of EventTarget objects
-				event.target.dispatchEvent(
+				event.target?.dispatchEvent(
 					new CustomEvent('sveltekit:trigger_prefetch', { bubbles: true })
 				);
 			}, 20);
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -109,7 +109,8 @@ export class Router {
 		const handle_mousemove = (event) => {
 			clearTimeout(mousemove_timeout);
 			mousemove_timeout = setTimeout(() => {
-				// ensure event.composedPath() in find_anchor(event) returns an array of EventTarget objects
+				// event.composedPath(), which is used in find_anchor, will be empty if the event is read in a timeout
+				// add a layer of indirection to address that
 				event.target?.dispatchEvent(
 					new CustomEvent('sveltekit:trigger_prefetch', { bubbles: true })
 				);
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -109,13 +109,16 @@ export class Router {
 		const handle_mousemove = (event) => {
 			clearTimeout(mousemove_timeout);
 			mousemove_timeout = setTimeout(() => {
-				event.target.dispatchEvent(new CustomEvent('trigger_prefetch', { bubbles: true }));
+				// ensure event.composedPath() in find_anchor(event) returns an array of EventTarget objects
+				event.target.dispatchEvent(
+					new CustomEvent('sveltekit:trigger_prefetch', { bubbles: true })
+				);
 			}, 20);
 		};
 
 		addEventListener('touchstart', trigger_prefetch);
 		addEventListener('mousemove', handle_mousemove);
-		addEventListener('trigger_prefetch', trigger_prefetch);
+		addEventListener('sveltekit:trigger_prefetch', trigger_prefetch);
 
 		/** @param {MouseEvent} event */
 		addEventListener('click', (event) => {
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -94,7 +94,7 @@ export class Router {
 			}, 200);
 		});
 
-		/** @param {MouseEvent|TouchEvent} event */
+		/** @param {Event} event */
 		const trigger_prefetch = (event) => {
 			const a = find_anchor(event);
 			if (a && a.href && a.hasAttribute('sveltekit:prefetch')) {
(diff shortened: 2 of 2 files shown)


.changeset/lazy-ways-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+dispatch custom event to retrieve composed path
```
bluwy, line 5: Maybe we can rephrase this as

```suggestion
fix `sveltekit:prefetch` mouse detection
```

So that it's identifiable as a prefetch fix in the changelog.

packages/kit/src/runtime/client/router.js
```diff
@@ -109,12 +109,13 @@ export class Router {
 		const handle_mousemove = (event) => {
 			clearTimeout(mousemove_timeout);
 			mousemove_timeout = setTimeout(() => {
-				trigger_prefetch(event);
+				event.target.dispatchEvent(new CustomEvent('trigger_prefetch', { bubbles: true }));
```
bluwy, line 112: We could add a comment above here explaining the reason of emitting a custom event.

packages/kit/src/runtime/client/router.js
```diff
@@ -109,12 +109,13 @@ export class Router {
...
 			}, 20);
 		};
 
 		addEventListener('touchstart', trigger_prefetch);
 		addEventListener('mousemove', handle_mousemove);
+		addEventListener('trigger_prefetch', trigger_prefetch);
```
bluwy, line 118: I think we can namespace the event as `sveltekit:trigger_prefetch` so we know it's SvelteKit specific.

packages/kit/src/runtime/client/router.js
borntofrappe: I'm afraid of being verbose, but would somethink like the following work?

```js
// the custom event guarantees that event.composedPath() doesn't return an empty array
```

Or, in a positive form

```js
// the custom event guarantees that event.composedPath() returns an array of EventTarget objects
```

borntofrappe: Looking at the other comments a more active tone might be preferable

```js
// ensure event.composedPath() returns an array of EventTarget objects
```

bluwy: Yeah the active tone comment sounds nice. But I wouldn't worry too much about that, feel free to explain as clearly as needed, as long as it's not one giant essay. (That's a lot of as)

benmccann:
I didn't review the PR yet, but please note that the Lint CI job is failing and would need to be fixed

borntofrappe:
> I didn't review the PR yet, but please note that the Lint CI job is failing and would need to be fixed

Apologies, but I don't know how to consider the message "No overload matches this call". Is this connected to TypeScript? Unfortunately I need some guidance on this point.

bluwy:
For "No overload matches this call", TypeScript isn't happy that `handle_prefetch` requires the `MouseEvent | TouchEvent` param because of the custom `sveltekit:handle_prefetch` event. But looking at the types for `find_anchor`, we could simplify the `handle_prefetch` param as `Event`. That should fix it.

borntofrappe:
> For "No overload matches this call", TypeScript isn't happy that `handle_prefetch` requires the `MouseEvent | TouchEvent` param because of the custom `sveltekit:handle_prefetch` event. But looking at the types for `find_anchor`, we could simplify the `handle_prefetch` param as `Event`. That should fix it.

I see. Using `Event` seems to fix that particular issue, but `pnpm check` highlights another one when accessing `event.target`.

> Object is possibly 'null'.ts(2531)
>
> (property) Event.target: EventTarget | null

I've managed to remove the message by returning the function early if `event.target` is null, but you might have a better solution.

```js
if (!event.target) return;

event.target.dispatchEvent(
  new CustomEvent('sveltekit:trigger_prefetch', { bubbles: true })
);
```

packages/kit/src/runtime/client/router.js
```diff
@@ -109,12 +109,17 @@ export class Router {
 		const handle_mousemove = (event) => {
 			clearTimeout(mousemove_timeout);
 			mousemove_timeout = setTimeout(() => {
-				trigger_prefetch(event);
+				if (!event.target) return;
+				// ensure event.composedPath() in find_anchor(event) returns an array of EventTarget objects
+				event.target.dispatchEvent(
+					new CustomEvent('sveltekit:trigger_prefetch', { bubbles: true })
+				);
```
bluwy, line 116:
```suggestion
				// ensure event.composedPath() in find_anchor(event) returns an array of EventTarget objects
				event.target?.dispatchEvent(
					new CustomEvent('sveltekit:trigger_prefetch', { bubbles: true })
				);
```

Using `?.` would be a bit cleaner.

bluwy (approved):
LGTM. The failing tests seems to be caused by an issue with `amphtml-validator` (unrelated), I've made #3022 to fix this.

packages/kit/src/runtime/client/router.js
```diff
@@ -109,12 +109,16 @@ export class Router {
 		const handle_mousemove = (event) => {
 			clearTimeout(mousemove_timeout);
 			mousemove_timeout = setTimeout(() => {
-				trigger_prefetch(event);
+				// ensure event.composedPath() in find_anchor(event) returns an array of EventTarget objects
```
benmccann, line 112: I didn't understand the comment until I read through the issue thread
```suggestion
				// event.composedPath(), which is used in find_anchor, will be empty if the event is triggered in a timeout
				// add a layer of indirection to address that
```

packages/kit/src/runtime/client/router.js
```diff
@@ -109,12 +109,16 @@ export class Router {
...
+				event.target?.dispatchEvent(
+					new CustomEvent('sveltekit:trigger_prefetch', { bubbles: true })
```
benmccann, line 114: I wonder if something like `sveltekit:mousestop` might be a better name for this? What do you guys think?

bluwy: Either works well for me, but it does looks nicer when used in the `addEventListener` call below.

borntofrappe: I lean toward `sveltekit:trigger_prefetch` as well.

bluwy (approved):

packages/kit/src/runtime/client/router.js
Rich-Harris: My instinct is to look for an alternative solution altogether — it seems weird to dispatch a custom event just to indirectly call our own function, no?

Rich-Harris: Actually, I realised there's a deeper problem here. The `composedPath` stuff was [introduced](https://github.com/sveltejs/kit/pull/2769) because recursively checking `parentNode` to find an `<a>` element doesn't work if the `<a>` in question is in shadow DOM.

The solution here dispatches a custom event on `e.target`, but the `composedPath` of the custom event _won't_ include the shadow `<a>` in certain cases. [Demo here](https://stackblitz.com/edit/vite-gdntva) — basically, if you have a custom element with an `<a>` in the shadow DOM like this...

```js
customElements.define(
  'fancy-a',
  class extends HTMLElement {
    constructor() {
      super();

      let a = document.createElement('a');
      a.href = this.getAttribute('href');
      a.innerHTML = '<slot></slot>';

      let shadowRoot = this.attachShadow({ mode: 'open' });
      shadowRoot.append(a);
    }
  }
);
```

...then whether or not the `<a>` will be included in the `composedPath` depends on whether the original event happened to happen within an element in light DOM inside the custom element.

I'm not exactly sure how to solve this, or if it's even possible. We might have to just say that prefetching won't work with shadow DOM.

Rich-Harris: Ok, it looks like shadow roots have an `elementsFromPoint` method, like `document`, which means we could solve it [thusly](https://stackblitz.com/edit/vite-5cmvnt):

```js
function find_anchor(x, y, root = document) {
  for (const element of root.elementsFromPoint(x, y)) {
    if (element.tagName === 'A') return element;
    if (element.shadowRoot && element.shadowRoot !== root) {
      const found = find_anchor(x, y, element.shadowRoot);
      if (found) return found;
    }
  }
}
```

This does run the risk of false positives (e.g. an element is on top of the `<a>`, blocking clicks) but I guess nonsense like this is just the price we pay for supporting web components. Perhaps there's a neat way to use `composedPath` for `touchstart` and only use `elementsFromPoint` for the mouse events.

bluwy: That does look a bit risky and potentially more resource intensive. I wonder if it would be fine to not support `sveltekit:prefetch` in web components at all, the usecase would be rare. And also that normal link clicks should already work in web components.

benmccann: I would be totally fine dropping support for prefetch with web components

dummdidumm: I'm on the pragmatic side here. I don't think that dispatching a custom element with an explanation why this is needed is bad nor does it clutter the code.

borntofrappe: The biggest problem with the custom event is that Rich showed it doesn't work all the times. I share the sentiment from @bluwy that the issue is becoming rather cumbersome.

Rich-Harris: Very well, if we're okay only having partial support for shadow DOM then we can merge this — thanks. Very annoying that the existence of shadow DOM forces us to jump through these ridiculous hoops in the first place!

PR #3005, opened 2021-12-08
fix build_server fails when sources contain symlinks

martaver:

### Tests

### Changesets

I've been playing with building sveltekit in bazel, which symlinks sources to its sandbox for build. Vite respects these symlinks, resolving their paths using `fs.realpathSync` and indexes chunks in `client_manifest` by their relative, real path.

This causes build_server to fail because it is trying to look up chunks by their path under the configured `files` paths.

Vite offers an option `resolve.preserveSymlinks` that fixes this problem, but there are also advantages to fully resolving symlinks in source files and node_modules paths, such as being able to handle workspaces correctly.

This PR adds a fallback check to build_server that also checks the resolved relative realpath of a file if the first attempt fails.

I've crossed off the other reqs, but could use a little guidance pointing out where I should set up a test for this, since I fixed this using patch in my own installation. At least I can say that this PR doesn't break anything.

### DIFF
--- a/packages/kit/test/test.js
+++ b/packages/kit/test/test.js
@@ -9,6 +9,7 @@ import { preview } from '../src/core/preview/index.js';
 import { load_config } from '../src/core/config/index.js';
 import { fileURLToPath, pathToFileURL } from 'url';
 import { format } from 'util';
+import path from 'path';
 
 /**
  * @param {{ port: number }} opts
@@ -283,6 +284,14 @@ function duplicate(test_fn, config, is_build) {
 	};
 }
 
+/**
+ * 
+ * @param {*} app string
+ */
+ function getCwd(app) {
+	return fileURLToPath(new URL(`apps/${app}`, import.meta.url));
+}
+
 async function main() {
 	// @ts-expect-error
 	globalThis.UVU_DEFER = 1;
@@ -290,7 +299,7 @@ async function main() {
 
 	const apps = process.env.APP
 		? [process.env.APP]
-		: fs.readdirSync(new URL('apps', import.meta.url));
+		: fs.readdirSync(new URL('apps', import.meta.url));	
 
 	/**
 	 * @param {string} app
@@ -415,7 +424,7 @@ async function main() {
 	}
 
 	for (const app of apps) {
-		const cwd = fileURLToPath(new URL(`apps/${app}`, import.meta.url));
+		const cwd = getCwd(app);
 		const tests = await Promise.all(
 			glob('**/_tests.js', { cwd }).map((file) => import(pathToFileURL(`${cwd}/${file}`).href))
 		);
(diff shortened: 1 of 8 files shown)


martaver:
Seems failing test in `apps/amp` is inherited from master...

bluwy:
The amp test has been fixed in master, CI should be working properly after a rebase.

packages/kit/test/test.js
```diff
@@ -283,6 +284,14 @@ function duplicate(test_fn, config, is_build) {
 	};
 }
 
+/**
+ *
```
benmccann, line 288: blank line can be removed

packages/kit/test/test.js
```diff
@@ -283,6 +284,14 @@ function duplicate(test_fn, config, is_build) {
...
+ * @param {*} app string
```
benmccann, line 289: can a more specific type be listed here?

packages/kit/test/test.js
```diff
@@ -283,6 +284,14 @@ function duplicate(test_fn, config, is_build) {
...
+ */
+function getCwd(app) {
```
benmccann, line 291: this should be `get_cwd`. We use snake_case for internal methods

pnpm-lock.yaml
```diff
@@ -30,7 +30,7 @@ importers:
       '@sveltejs/eslint-config': github.com/sveltejs/eslint-config/9a7d728e03ac433e5856a6e06775c17ee986d641_d9525e585486a10ae6317f3f61e9597f
       '@typescript-eslint/eslint-plugin': 5.5.0_15fb0f7dd5018b02e6608eb3a323af2f
       '@typescript-eslint/parser': 5.5.0_eslint@8.3.0+typescript@4.4.4
-      action-deploy-docs: github.com/sveltejs/action-deploy-docs/773baf07ba9391eee581dcecfb672265e5fdbddb
+      action-deploy-docs: github.com/sveltejs/action-deploy-docs/e238bd8fcb7c96b683534eff14f3c7283d6a85c2
```
benmccann, line 33: you shouldn't need to change the lock file

packages/kit/test/test.js
```diff
@@ -432,7 +441,20 @@ async function main() {
...
-main();
+// sets up a symlinked route
+const basicsCwd = getCwd('basics');
+const simlinkRoutePath = path.resolve(`${basicsCwd}/src/routes/with-symlink/symlinked`);
+removeSymlinkRoute();
+fs.symlinkSync(path.resolve(`${basicsCwd}/symlink-to-routes`), simlinkRoutePath, 'dir');
+
+main().finally(() => removeSymlinkRoute());
+
+function removeSymlinkRoute() {
```
benmccann, line 456: this should be snake_case as well

packages/kit/test/test.js
Rich-Harris, line 288:
```suggestion
```

Rich-Harris:
I've tried to bring this branch up to date with `master` but they've diverged too far and I can't get the test to pass — it would be easier to start again, so I'm going to close this — sorry.

To be totally honest I don't understand what problem this is solving and why it's important, but if it is then would you mind opening an issue with a minimal reproduction? Thanks

martaver:
Hey Rich, 

I created this originally because I've been trying to get sveltekit to build in bazel, where source files are symlinked into the build sandbox. By default, vite dereferences symlinks which messed up sveltekit's build. Adding the `preserveSymlinks` option to the `vite` configuration fixed this, however I got to thinking that it might be nice if this wasn't necessary and started to look at a way to get sveltekit's build to support dereferenced symlinks.

I realised after a while though that this is a bit of a rabbit hole, and it's probably just easier to configure vite to `preserveSymlinks`. E.g. what if the original source itself also contains dirs symlinked under routes? Where does this ever end, and how does sveltekit know whether a path is under 'routes' or not?

So happy to close this PR and chalk it up to rubber ducking. I meant to clean this up, so sorry for leaving it hanging and wasting your time!

PR #3013, opened 2021-12-09
[feat] pass `render` function to `handle`

MrWaip:
This is needed if you want to render some page from handle hook.

One potential use case would allowing render custom error page if you catch an error in handle function.


Conduitry:
What does this provide that the existing `resolve` function doesn't?

Conduitry:
There's a checklist in the PR description template that begins "please don't delete this checklist". Please don't delete it.

Please also don't open PRs for features without discussing them in an issue first.

Issue #3009, opened 2021-12-08
Additional headers are lost with a server-side fetch


illright:
### Describe the bug

When a server-side `load` function makes a request to the same hostname (not necessarily the same domain) and includes additional headers with the `Headers` object instead of a regular object, these additional headers are replaced with a `Cookie` header. 

There are two issues to address:
- [ ] Overwritten headers if they are passed as a `Headers` object
- [ ] Incorrect detection of similar hostnames (ports aren't considered)

### Reproduction

https://github.com/illright/sveltekit-repro-headers-class

### Logs

```shell
N/A
```

### System Info

```shell
System:
    OS: Linux 5.10 Arch Linux
    CPU: (8) x64 Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
    Memory: 3.32 GB / 3.76 GB
    Container: Yes
    Shell: 3.3.1 - /bin/fish
  Binaries:
    Node: 17.1.0 - /usr/sbin/node
    npm: 8.1.3 - /usr/sbin/npm
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.3
    @sveltejs/kit: next => 1.0.0-next.201
    svelte: ^3.44.0 => 3.44.2
```

### Severity

blocking all usage of SvelteKit

illright:
Actually, it seems like the ports should not be considered, according to [RFC 6265](https://www.ietf.org/rfc/rfc6265.txt):

> Similarly, cookies for a given host are shared
   across all the ports on that host, even though the usual "same-origin
   policy" used by web browsers isolates content retrieved via different
   ports.

That must mean this issue can be closed.

PR #3034, opened 2021-12-13
fix: normalize the usage of headers

illright:

### Tests

### Changesets

This will normalize the usage of headers across the server-side `fetch` to ensure that `Record<string, string>`, `string[][]` and `Headers` are all treated correctly. Previosly, as described in https://github.com/sveltejs/kit/issues/3009, whenever headers in the form of `Headers` would come to the server-side fetch, the would be lost under a certain condition.

This only fixes one of the problems in #3009, the other one about the hostnames not being compared correctly still persists.

### DIFF
--- a/packages/kit/test/apps/basics/src/routes/headers/_tests.js
+++ b/packages/kit/test/apps/basics/src/routes/headers/_tests.js
@@ -11,7 +11,7 @@ export default function (test) {
 		'allows headers to be sent as a Headers class instead of a POJO',
 		'/headers/class',
 		async ({ page }) => {
-			assert.equal(await page.innerHTML('pre'), JSON.stringify({ foo: 'bar' }));
+			assert.equal(await page.innerHTML('p'), 'bar');
 		}
 	);
 }
--- a/packages/kit/test/apps/basics/src/routes/headers/_tests.js
+++ b/packages/kit/test/apps/basics/src/routes/headers/_tests.js
@@ -6,4 +6,12 @@ export default function (test) {
 		const headers = response.headers();
 		assert.equal(headers['permissions-policy'], 'interest-cohort=()');
 	});
+
+	test(
+		'allows headers to be sent as a Headers class instead of a POJO',
+		'/headers/class',
+		async ({ page }) => {
+			assert.equal(await page.innerHTML('pre'), JSON.stringify({ foo: 'bar' }));
+		}
+	);
 }
--- a/.changeset/smooth-cars-build.md
+++ b/.changeset/smooth-cars-build.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Headers from a `Headers()` class are now correctly handled by the server to avoid losing headers
--- a/.changeset/smooth-cars-build.md
+++ b/.changeset/smooth-cars-build.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Headers from a `Headers()` class are now correctly handled by the server to avoid losing headers
+Handle `Headers` instance in server-side `fetch`
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -3,11 +3,6 @@ import { Headers } from 'node-fetch';
 import { normalize } from '../../load.js';
 import { respond } from '../index.js';
 import { escape_json_string_in_html } from '../../../utils/escape.js';
-import {
-	ensure_headers_class,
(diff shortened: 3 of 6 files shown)


illright:
I couldn't figure out how to resolve an error that was preventing me from running tests, so I haven't seen this error before. I'm guessing that the build is failing due to `node-fetch` being imported somewhere where it shouldn't be imported, but I'm honestly not sure how to approach fixing this error. Help is highly appreciated.

benmccann:
PR to address the test failure here: https://github.com/sveltejs/vite-plugin-svelte/pull/225

illright:
@benmccann looks like the PR you mentioned is merged. Is there anything I can do to speed up the release of these changes? They address the issue that's blocking one of my projects that I must release soon.

bluwy:
@illright We'll cut a new release for vite-plugin-svelte soon. There's some CI issue there as well, which will be solved by https://github.com/sveltejs/vite-plugin-svelte/pull/226. Once that gets approved, we can update kit with that and rebase this PR.

illright:
For a second thought, it's not a blocker, I can simply disable SSR for now, everything works in CSR mode. But this PR still remains very desirable for me. Just letting you know that there is no rush.

In the meantime, could you please take a look at the code and let me know if everything looks alright there?

bluwy (commented):
I'm not well-versed in this area of the code so I'll leave the others to review this (Not sure if supporting the `Headers` class a common thing). But at the meantime, `master` has updated to the latest `vite-plugin-svelte`, so a rebase should fix the CI issue.

illright:
Here's the background for this issue – I'm using [`ky`](https://github.com/sindresorhus/ky) as the API client in my application. It handles header management, among other things, and passes a `Headers` instance to the underlying `fetch`, and I have no control over that. So I imagine that to all other users of `ky` this is desirable to support.

Rich-Harris:
Thank you! We can make this quite a bit simpler, I think — `new Headers(old_headers)` will work whether `old_headers` is a POJO or an instance of `Headers`, and `response.headers` is always guaranteed to be an instance of `Headers`. Because of that we don't need any new helper functions. I've updated the PR

packages/kit/test/apps/basics/src/routes/headers/_tests.js
```diff
@@ -6,4 +6,12 @@ export default function (test) {
 		const headers = response.headers();
 		assert.equal(headers['permissions-policy'], 'interest-cohort=()');
 	});
+
+	test(
+		'allows headers to be sent as a Headers class instead of a POJO',
```
benmccann, line 11: this makes it sounds like you could pass either. is that correct or only a Headers instance is accepted now?

illright:
@Rich-Harris are you sure about this? It seems to me that, according to [the Fetch Standard](https://fetch.spec.whatwg.org/#headers-class), the `Headers` class is only required to accept a POJO or an array of entries, nothing is mentioned about it accepting an instance of itself.

The reason I'm concerned is that SvelteKit doesn't ship an implementation of Headers, so for Node, we cannot guarantee that the people will use an implementation that is capable of this somewhat non-standard behaviour. But I agree that this makes the code quite a bit more bloated. Maybe we could somehow expose `Headers` from SvelteKit to allow users to ensure that they are using the correct implementation?

packages/kit/test/apps/basics/src/routes/headers/_tests.js
illright: This test was written by Rich, but I believe that both should be accepted, otherwise we'd make a breaking change

Rich-Harris: both should be accepted, because we're just mirroring the native `fetch` API (until this PR, i didn't realise you could pass a `Headers` object; turns out `fetch` accepts either)

Rich-Harris:
Yep! It'll accept any iterable that yields string pairs:

```js
stuff = {
  [Symbol.iterator]: function* () {
    yield ['problems', '99'];
    yield ['answer', '42'];
  }
};

const headers = new Headers(stuff);

headers.get('problems'); // '99'
headers.get('answer'); // '42'
```

That includes Headers, Map and so on. If anyone is calling fetch with a `Headers` object that doesn't implement the iterable protocol, their code is broken and that's very much on them.

SvelteKit exposes `Headers` from `node-fetch` during development. In production, the environment is expected to provide a standards-compliant `fetch` implementation (`adapter-node` and others do this via `node-fetch`, but many environments like Deno etc already provide it).

Rich-Harris (approved):

PR #3047, opened 2021-12-14
[adapter-cloudflare] Update static file logic to allow spaces in filenames

mcfadyeni:
**This PR makes the Worker URL-decode any received paths so that static files with spaces in the filename can be found.**

Without this change, accessing a static file that has a space in it (for example, `https://example.com/Example File.pdf`) fails. This is because:
1. The file is added to the `ASSETS` Set with a space (non-URL encoded).
2. The browser automatically URL encodes any space (` `) characters to `%20`, as it should.
3. When the request is received by the worker, it looks for a static file in the `ASSETS` Set with `%20` instead of space (` `). This fails, so the request is passed to SvelteKit which replies with a 404.

I've tested this change on my own local copy of adapter-cloudflare in production and it works.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-cloudflare/files/worker.js
+++ b/packages/adapter-cloudflare/files/worker.js
@@ -7,7 +7,7 @@ export default {
 	async fetch(req, env) {
 		const url = new URL(req.url);
 		// check generated asset_set for static files
-		if (ASSETS.has(url.pathname.substring(1))) {
+		if (ASSETS.has(decodeURIComponent(url.pathname.substring(1)))) {
 			return env.ASSETS.fetch(req);
 		}
 
--- a/packages/adapter-cloudflare/files/worker.js
+++ b/packages/adapter-cloudflare/files/worker.js
@@ -6,14 +6,15 @@ init();
 export default {
 	async fetch(req, env) {
 		const url = new URL(req.url);
+
 		// check generated asset_set for static files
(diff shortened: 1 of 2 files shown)


benmccann (approved):

pngwn (approved):
Lgtm.

packages/adapter-cloudflare/files/worker.js
```diff
@@ -7,7 +7,7 @@ export default {
 	async fetch(req, env) {
 		const url = new URL(req.url);
 		// check generated asset_set for static files
-		if (ASSETS.has(url.pathname.substring(1))) {
+		if (ASSETS.has(decodeURIComponent(url.pathname.substring(1)))) {
```
lukeed, line 10:
```suggestion
    let pathname = url.pathname.substring(1);
    try { 
      pathname = decodeURIComponent(pathname);
    } catch (err) { 
      // ignore
    }
    
		if (ASSETS.has(pathname)) {
```

mcfadyeni: Looks good; I'll make that change.

lukeed (approved):

lukeed:
the `lint` step needs to be fixed, but looks good otherwise

mcfadyeni:
Sorry! Fixed now.

PR #3064, opened 2021-12-17
[feat] use preserveValueImports flag in create-svelte

dummdidumm:
TypeScript doesn't know about import usages in the template because it only sees the script of a Svelte file. Therefore preserve all value imports. Requires TS 4.5 or higher, therefore also bump the version of TS and Prettier (the latter in order to deal with the new TS syntax).
Turning on preserveValueImports turns off the advanced but somewhat brittle import transpilation in svelte-preprocess which should give a small performance boost and more robustness.

This requires ESBuild 0.14 or higher. According to Vite's `package.json` they install a dependency of 0.13.x or higher, so new users should get the version. Pinging @bluwy / @benmccann to ensure I'm not wrong because they know more about Vite than me 😄 

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/shared/+typescript/package.json
+++ b/packages/create-svelte/shared/+typescript/package.json
@@ -4,7 +4,7 @@
 		"check:watch": "svelte-check --tsconfig ./tsconfig.json --watch"
 	},
 	"devDependencies": {
-		"typescript": "^4.5.4",
+		"typescript": "~4.5.4",
 		"tslib": "^2.3.1",
 		"svelte-check": "^2.2.6",
 		"svelte-preprocess": "^4.10.1"
--- a/packages/create-svelte/shared/+typescript/package.json
+++ b/packages/create-svelte/shared/+typescript/package.json
@@ -4,9 +4,9 @@
 		"check:watch": "svelte-check --tsconfig ./tsconfig.json --watch"
 	},
 	"devDependencies": {
-		"typescript": "^4.4.3",
+		"typescript": "^4.5.4",
 		"tslib": "^2.3.1",
 		"svelte-check": "^2.2.6",
-		"svelte-preprocess": "^4.9.4"
+		"svelte-preprocess": "^4.10.1"
 	}
 }
--- a/.changeset/metal-ears-travel.md
+++ b/.changeset/metal-ears-travel.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+use preserveValueImports flag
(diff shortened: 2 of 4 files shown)


dominikg:
^0.13 excludes 0.14, so we'll have to wait until vite updates ( should happen with 2.8 )

dummdidumm:
Oh wow, TIL, caret actually means "do not modify the left-most non-zero element". Thanks for the clarification, marking as draft then in the meantime.

Rich-Harris:
I wonder if we should be using [extends](https://www.typescriptlang.org/tsconfig#extends) to encapsulate things that are important to SvelteKit? This would also allow future versions of Kit to make further changes without requiring existing apps to update their tsconfig.json files.

I haven't used it in the past so I don't know what the caveats are but I imagine it could also be used to keep `paths` in sync with `config.kit.files.lib` (though I'm not sure what file you would be extending in that case, since it would be app-specific rather than something inside `node_modules/@sveltejs/kit`). Maybe we could even tackle https://github.com/sveltejs/kit/issues/647?

dummdidumm:
We could very well do this, yes. There's precedence to this already in the Svelte starter template where we extend from `@tsconfig/svelte` (I repo by Orta for config bases). This tsconfig should live within SvelteKit so we can update it and everyone updating Kit would get it directly.
When we want things like "keep X in sync with Y" then we are talking about another dynamically generated tsconfig though. I'm not sure how to best approach this - would there be a "syncTsconfig" command which looks at things like `config.kit.files.lib` and generates the results accordingly?

Rich-Harris:
I'm imagining something like this...

```json
```

...where `.svelte-kit/tsconfig.json` is regenerated along with the other generated files, but it has two drawbacks:

1. it's sort of ugly
2. it doesn't exist until you first run a `svelte-kit` command, so a newly-created project would be full of squigglies

I don't know if those are dealbreakers?

packages/create-svelte/shared/+typescript/package.json
```diff
@@ -4,9 +4,9 @@
 		"check:watch": "svelte-check --tsconfig ./tsconfig.json --watch"
 	},
 	"devDependencies": {
-		"typescript": "^4.4.3",
+		"typescript": "^4.5.4",
```
dominikg, line 7: ~4.5.4 instead?

dummdidumm: No, this is for the `devDependencies` of the template, not those of the Kit repo, and I don't want to update this every time a new minor version for TS is released.

dummdidumm: After the discussion in Discord I changed my mind, being more defensive and do `~` makes sense, maybe there are future TS updates that don't play nice with things in Vite/ESBuild.

benmccann (approved):

Issue #3074, opened 2021-12-20
amp-install-service worker incorrect location


asendia:
### Describe the bug

In the past I tried to fix amp service worker validation error in this PR: https://github.com/sveltejs/kit/pull/2265
Updating this:
```javascript
if (options.service_worker) {
  init += `<script>
    if ('serviceWorker' in navigator) {
      navigator.serviceWorker.register('${options.service_worker}');
    }
```
to this:
```javascript
if (options.service_worker) {
  init += options.amp
    ? `<amp-install-serviceworker src="${options.service_worker}" layout="nodisplay"></amp-install-serviceworker>`
    : `<script>
    if ('serviceWorker' in navigator) {
      navigator.serviceWorker.register('${options.service_worker}');
```

Apparently `<amp-install-service-worker>` only works inside `<body>` unlike normal service worker registration `navigator.serviceWorker.register(` this causes validation error when you enable service_worker & amp from your config.

#### Possible solutions
1. I am thinking whether it is better for the user to include `<amp-install-serviceworker src="${options.service_worker}" layout="nodisplay"></amp-install-serviceworker>` manually in the body e.g. in `__layout.svelte`?

This can be done by updating the code above to
```javascript
if (options.service_worker && !options.amp) {
  init += `<script>
    if ('serviceWorker' in navigator) {
      navigator.serviceWorker.register('${options.service_worker}');
    }
```
This won't let service worker in amp works out of the box in sveltekit.

2. Or should sveltekit inject `<amp-install-serviceworker>` in the body? https://github.com/sveltejs/kit/pull/3075. But this limits the customization of `<amp-install-serviceworker>` element.

### Reproduction

1. Open: https://test-broken-amp-sw.salmonfit.com/
2. Validate in amp validator: https://validator.ampproject.org/

#### How to reproduce in local:
1. In a sveltekit project, enable amp from you `svelte.config.js`
```javascript
import preprocess from 'svelte-preprocess';
import adapter from '@sveltejs/adapter-static';
import { resolve } from "path";

const config = {
  preprocess: preprocess(),
  kit: {
    amp: true, // enable amp
    adapter: adapter({
      pages: 'build',
      assets: 'build',
      fallback: null,
    }),
...
```
2. Enable service worker by creating `src/service-worker.ts`

3. Build the project
```sh
npm build
npm preview
```

4. Open localhost:3000
5. Validate the source in amp validator: https://validator.ampproject.org/
[...]

PR #3075, opened 2021-12-20
[fix] invalid amp-install-serviceworker

asendia:
Fixes https://github.com/sveltejs/kit/issues/3074 by injecting `<amp-install-serviceworker>` inside `<body>` instead of `<head>`. But user won't be able to customize the component.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -170,26 +170,23 @@ export async function render_response({
 		init
 	].join('\n\n\t\t');
 
-	const body = options.amp
-		? `${rendered.html}
-			${
-				options.service_worker
-					? `<amp-install-serviceworker src="${options.service_worker}" layout="nodisplay"></amp-install-serviceworker>`
-					: ''
-			}`
-		: `${rendered.html}
+	let body = rendered.html;
+	if (options.amp) {
+		if (options.service_worker) {
+			body += `<amp-install-serviceworker src="${options.service_worker}" layout="nodisplay"></amp-install-serviceworker>`;
+		}
+	} else {
+		body += `${serialized_data
+			.map(({ url, body, json }) => {
+				let attributes = `type="application/json" data-type="svelte-data" data-url=${escape_html_attr(
+					url
+				)}`;
+				if (body) attributes += ` data-body="${hash(body)}"`;
 
-			${serialized_data
-				.map(({ url, body, json }) => {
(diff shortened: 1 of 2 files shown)


asendia (commented):
Line 175 is hard to read 😓

packages/kit/src/runtime/server/page/render.js
```diff
@@ -173,7 +171,12 @@ export async function render_response({
...
 
 	const body = options.amp
-		? rendered.html
+		? `${rendered.html}
+			${
+				options.service_worker
+					? `<amp-install-serviceworker src="${options.service_worker}" layout="nodisplay"></amp-install-serviceworker>`
+					: ''
+			}`
 		: `${rendered.html}
```
bluwy, line 180: I think we can refactor this as:

```js
let body = rendered.html

if (options.amp) {
  if (options.service_worker) {
    body += ''
  }
} else {
  body += ''
}
```

to make it a bit readable.

bluwy (commented):
I'm not super familiar with amp, so I'm giving a general code quality review. I think we can add a test for this as well.

packages/kit/src/runtime/server/page/render.js
asendia: I agree with your suggestion & have modified the code, thanks 👍🏼

Rich-Harris (approved):

Rich-Harris (changes requested):
thank you! mind adding a changeset so this will trigger a new version?

packages/kit/src/runtime/server/page/render.js
```diff
@@ -172,21 +170,23 @@ export async function render_response({
...
+		body += `${serialized_data
+			.map(({ url, body, json }) => {
+				let attributes = `type="application/json" data-type="svelte-data" data-url=${escape_html_attr(
+					url
+				)}`;
+				if (body) attributes += ` data-body="${hash(body)}"`;
+
+				return `<script ${attributes}>${json}</script>`;
+			})
+			.join('\n\n\t')}`;
```
bluwy, line 188:
```suggestion
		body += serialized_data
			.map(({ url, body, json }) => {
				let attributes = `type="application/json" data-type="svelte-data" data-url=${escape_html_attr(
					url
				)}`;
				if (body) attributes += ` data-body="${hash(body)}"`;

				return `<script ${attributes}>${json}</script>`;
			})
			.join('\n\n\t');
```

We should be able to remove the outer template string interpolation.

bluwy (approved):

Rich-Harris (approved):

PR #3076, opened 2021-12-20
docs: remove vite 2.7 notes

bluwy:
Remove some FAQ notes that reference Vite 2.7 as we are now already using it.

### Tests

### Changesets

### DIFF
--- a/documentation/faq/60-env-vars.md
+++ b/documentation/faq/60-env-vars.md
@@ -4,6 +4,6 @@ question: How do I use environment variables?
 
 Vite uses [dotenv](https://github.com/motdotla/dotenv) to load environment variables from a file named `.env` or similar. Only environment variables prefixed with `VITE_` are exposed ([you can set `envPrefix` to change this](https://vitejs.dev/config/#envprefix)). Vite will use these and statically replace them at build-time.
 
-To use environment variables at runtime, you would need to instantiate dotenv yourself in your server code so that they are exposed at `process.env.YOUR_ENV_VAR`. You may also use `$session` to pass them to the client if needed.
+To use environment variables at runtime, you would need to instantiate dotenv yourself in your server-side code so that they are exposed at `process.env.YOUR_ENV_VAR`. You may also use `$session` to pass them to the client if needed.
 
 Please see [the Vite documentation](https://vitejs.dev/guide/env-and-mode.html#env-files) for more info about environment variables.
--- a/documentation/faq/60-env-vars.md
+++ b/documentation/faq/60-env-vars.md
@@ -2,6 +2,8 @@
 question: How do I use environment variables?
 ---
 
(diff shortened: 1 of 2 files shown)


ignatiusmb (approved):

documentation/faq/80-integrations.md
```diff
@@ -87,7 +87,7 @@ onMount(() => {
 
 ### How do I use Firebase?
 
-Please use Firebase SDK v9. You will also need to add Firebase dependencies to `ssr.external` ([example](https://github.com/benmccann/sveltekit-firebase/blob/9e3097fd859e4f81e4775885ecb584561f098fd3/svelte.config.js#L11)) or set `kit.ssr: false` until Vite 2.7 is released.
+Please use Firebase SDK v9. You will also need to add Firebase dependencies to `ssr.external` ([example](https://github.com/benmccann/sveltekit-firebase/blob/9e3097fd859e4f81e4775885ecb584561f098fd3/svelte.config.js#L11)).
```
bluwy, line 90: @benmccann Is the `ssr.external` thing still needed for Vite 2.7?

documentation/faq/60-env-vars.md
```diff
@@ -2,6 +2,6 @@
 question: How do I use environment variables?
 ---
 
-Vite uses [dotenv](https://github.com/motdotla/dotenv) to load environment variables from a file named `.env` or similar. Only environment variables prefixed with `VITE_` are exposed ([you can set `envPrefix` to change this](https://vitejs.dev/config/#envprefix)). For now, you would need to instantiate dotenv yourself if you want all environment variables exposed in `process.env['YOUR_ENV_VAR']`. [Vite 2.7 will expose all environment variables on the server-side](https://github.com/vitejs/vite/pull/5404).
+Vite uses [dotenv](https://github.com/motdotla/dotenv) to load environment variables from a file named `.env` or similar. Only environment variables prefixed with `VITE_` are exposed ([you can set `envPrefix` to change this](https://vitejs.dev/config/#envprefix)). For now, you would need to instantiate dotenv yourself if you want all environment variables exposed in `process.env.YOUR_ENV_VAR`.
```
benmccann, line 5: Should we update this to differentiate between client side and server side? I think the VITE_ prefix is only needed on the client side now?

documentation/faq/80-integrations.md
benmccann: No, it's no longer needed. Firebase should hopefully just work out of the box now. We could probably just remove this entire FAQ now.

documentation/faq/60-env-vars.md
Conduitry: We want to be careful about "environment variables that are baked in at build time" versus "server-side environment variables that are accessed at runtime". My understanding is that the `VITE_` prefix is specifically about the former, and that it is a build-time replacement performed both in server code and client code. You'd need to handle it yourself (say with `$session`) if you wanted to expose runtime server-side environment variables to the client.

documentation/faq/80-integrations.md
bluwy: removed the firebase section

bluwy:
I've also updated the env var section to differentiate between build-time and runtime env vars. Using `process.env.*` as the runtime example probably isn't the best, given worker environments for example where there's no `process`. But I think this would do for now until we have a general solution for that.

Issue #2697, opened 2021-10-28
Generated URLs are inconsistent to the basepath with `adapter-static`


illright:
### Describe the bug

Sapper was gracefully handling non-root basepaths by injecting a `<base>` tag into `<head>` in place of `%sapper.base%`. SvelteKit doesn't have that, but it does have `kit.paths.base`. By changing it, the auto-generated URLs of SvelteKit are rebased, but if there were any previous static URLs, for example, a favicon in `app.html`, these URLs will be broken.

Due to the fact that there are now both rebased and non-rebased URLs on a page, this can no longer be remedied by injecting a `<base>` tag with a hook, as @benmccann suggested in https://github.com/illright/attractions/issues/311#issuecomment-950333419.

### Reproduction

https://github.com/illright/sveltekit-repro-basepath

### System Info

```shell
System:
    OS: Linux 5.14 Manjaro Linux
    CPU: (8) x64 Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
    Memory: 3.16 GB / 7.62 GB
    Container: Yes
    Shell: 3.3.1 - /usr/bin/fish
  Binaries:
    Node: 16.11.0 - /usr/bin/node
    npm: 8.0.0 - /usr/bin/npm
  Browsers:
    Firefox: 93.0
  npmPackages:
    @sveltejs/adapter-static: ^1.0.0-next.21 => 1.0.0-next.21 
    @sveltejs/kit: next => 1.0.0-next.191 
    svelte: ^3.42.6 => 3.44.0
```

### Severity

blocking all usage of SvelteKit

### Additional Information

This blocks reliable deployment of SvelteKit-generated websites to GitHub Pages.

mgrubinger:
Can reproduce this in a similar setup (building for github-pages). I worked around it for now is to move the <link>s in app.html to my __layout.svelte file and wrap them in <svelte:head>. 
Would be great to see the paths in app.html respect kit.paths.base though 👍

Rich-Harris:
As of the most recent version you can do e.g. `%svelte.assets%/favicon.png` in your `src/app.html`

nstuyvesant:
Seeing `%svelte.head%`, `%svelte.body%` and `%svelte.assets%`, are these tokens for the template documented anywhere? Are there others that can be used?

PR #3088, opened 2021-12-22
2697: add base and asset transform to html template 🦎

dreitzner:

### Tests

### Changesets

### Changes
- add transforms for paths (assets, base) to html
- add info to migration guide
- add info into config docs

closes https://github.com/sveltejs/kit/issues/2697

### DIFF
--- a/.changeset/cuddly-lemons-live.md
+++ b/.changeset/cuddly-lemons-live.md
@@ -1,0 +1,8 @@
+---
+'@sveltejs/kit': patch
+---
+
+# Add base and asset transform to html template
+
+- `%svelte.assets%` is the placeholder for `config.paths.assets`
+- `%svelte.base%` is the placeholder for `config.paths.base`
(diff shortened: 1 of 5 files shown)


Rich-Harris:
Thank you. Adding `assets` definitely makes sense to me. Adding `base` could be a little bit tricky though, because we have ambitions to support dynamic basepaths — #595 — and that wouldn't work if we were writing the basepath into the HTML.

I think what we probably want to do is allow `assets` in the template...

```diff
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <meta name="description" content="" />
-    <link rel="icon" href="/favicon.png" />
+    <link rel="icon" href="%svelte.assets%/favicon.png" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    %svelte.head%
  </head>
  <body>
    <div id="svelte">%svelte.body%</div>
  </body>
</html>
```

...and then replace assets with `paths.assets` if it's set (in which case it's an absolute URL) and a relative path to the root otherwise. So `/` would look like this...

```html
<link rel="icon" href="./favicon.png" />
```

...but `/foo/bar/baz` would look like this:

```html
<link rel="icon" href="../../favicon.png" />
```

We'd also need to account for the `trailingSlash` option, since if `/foo/bar/baz` redirected to `/foo/bar/baz/` it would be `../../..` rather than `../..`.

Rich-Harris:
Closing in favour of #3234

PR #3093, opened 2021-12-23
sort rest params alphabetically

tanhauhau:
- `file` variable already `posixify` when declared: https://github.com/tanhauhau/kit/blob/3ccc049f197399eb7a0a955b3ad6919bc002703c/packages/kit/src/core/create_manifest_data/index.js#L67
- from what i read, since [...rest] params has to be a standalone segment, it is not possible to be `is_index: true`.
- and, it is plausible to have multiple routes with `[...rest]` params on the same level, since there's fallthrough routes. updated the comparator to sort it in alphabetical order

### Tests

### Changesets

### DIFF
--- a/.changeset/metal-rings-mix.md
+++ b/.changeset/metal-rings-mix.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Sort rest routes alphabetically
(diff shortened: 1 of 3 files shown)


Rich-Harris (approved):

Rich-Harris (changes requested):
Good catch, thank you. I initially approved this but realised it doesn't have a changeset — mind adding one? (I'd do it myself but I always find editing forks to be a nightmare — if you're comfortable opening PRs from the main repo in future rather than from a fork, then feel free!)

Rich-Harris (approved):

Rich-Harris:
Installed the [GitHub CLI](https://cli.github.com/) to make it easy to edit pull requests, so I took care of the changeset

Issue #598, opened 2021-03-23
Provide a way for layouts to determine if current page is an error page


benmccann:
In Sapper we did https://github.com/sveltejs/sapper/pull/1498

Suggestion from Rich for SvelteKit: 
> Given that it only applies to the root layout I think I'd prefer that we pass the error (and possibly status) as a prop. It doesn't belong on the page store

Rich-Harris:
Created a new `straightforward` label for issues like this that should be pretty easy to implement since they don't touch a bunch of different places. In this case I think we just need to add a prop here... https://github.com/sveltejs/kit/blob/64d02910d0000a4ca28af7de1a1cae8f487cc5d7/packages/kit/src/core/create_app/index.js#L191 ...and add a test.

The one downside of this approach is that we'll get `created with unknown prop` warnings in dev if the root layout doesn't expose an `error` prop, but that's not the worst thing in the world

Conduitry:
Could we partially relieve that by having the starter template define the prop in the root layout, even if it doesn't use it? That might not be worth it though, and would add bytes to the prod version just to avoid a dev time warning. I'm torn.

Rich-Harris:
I think Svelte should probably skip the warning in cases where the prop is undefined _anyway_.

Relatedly, we have a similar problem with it often warning that a default slot was provided to a leaf component

Rich-Harris:
We decided that the best way to do this would be

```js
import { status, error } from '$app/stores';
```

Rich-Harris:
closed via #3096

PR #3096, opened 2021-12-23
Add `status` and `error` to page store

tanhauhau:
related to #598

not sure how to add test for this

### Tests

### Changesets


packages/kit/src/runtime/client/renderer.js
```diff
@@ -754,6 +756,8 @@ export class Renderer {
 			params: {}
 		};
 
+		this.stores.error.set(error);
```
benmccann, line 759: The CI is failing on this line. You'll need to make sure `pnpm check` passes

Rich-Harris:
Since we've got rid of the `page` object in `load` functions in #3126, and `stores.page` now contains `url` and `params` directly, @Conduitry made the excellent observation that it should also include `error` and `status` rather than us having separate stores. This also conveniently solves any potential issues around stores getting out of sync (e.g. you're explicitly subscribing to both `error` and `status` and don't want to experience tearing)

Rich-Harris (approved):

PR #3112, opened 2021-12-27
Update wording for reproductions in issue template

benmccann:
Remove "within a reasonable time-frame". If we're up-front about the requirements then we should just go ahead and close the issue if a reproduction is not provided. Telling people to follow the instructions places the burden on us to go back and check the issue after some period of time and isn't very scalable.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -17,7 +17,7 @@ body:
     id: reproduction
     attributes:
       label: Reproduction
-      description: A link to a repository that reproduces the issue. Explaining how to reproduce is generally not enough. It pushes the burden of creating a reproduction project onto a small set of volunteer maintainers and isn't scalable. If no reproduction is provided, the issue will be closed.
+      description: A link to a repository, or a fork of https://node.new/sveltekit, that reproduces the issue. Explaining how to reproduce is generally not enough. It pushes the burden of creating a reproduction project onto a small set of volunteer maintainers and isn't scalable. If no reproduction is provided, the issue will be closed.
       placeholder: Reproduction
     validations:
       required: true
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -17,7 +17,7 @@ body:
     id: reproduction
     attributes:
       label: Reproduction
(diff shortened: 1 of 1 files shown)


bluwy:
I think giving some time to create a reproduction may still be beneficial, but to help us maintainers, we could follow Vite and create a special flow to automate closing if no reproduction given within N days? ([Example](https://github.com/vitejs/vite/issues/6279)). They have a [workflow](https://github.com/vitejs/vite/blob/eb08ec5a838e13c75f6d034137d5835b44ecf279/.github/workflows/issue-labeled.yml#L31-L40) to automate this, and another [one](https://github.com/vitejs/vite/blob/main/.github/workflows/issue-close-require.yml) to auto close it.

Rich-Harris:
I'm in favour of the wording change — if a repro is necessary, don't open the issue until you have one. I do think closing should be a matter of human discretion though — even if people didn't hate bots that auto-close issues, a human would still need to add the 'needs reproduction' label, so you're only gaining the illusion of automation and efficiency. Makes more sense to just close those issues immediately I think — people will eventually get the hint.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -17,7 +17,7 @@ body:
     id: reproduction
     attributes:
       label: Reproduction
-      description: A link to a repository that reproduces the issue. Explaining how to reproduce is generally not enough. It pushes the burden of creating a reproduction project onto a small set of volunteer maintainers and isn't scalable. If no reproduction is provided within a reasonable time-frame, the issue will be closed.
+      description: A link to a repository that reproduces the issue. Explaining how to reproduce is generally not enough. It pushes the burden of creating a reproduction project onto a small set of volunteer maintainers and isn't scalable. If no reproduction is provided, the issue will be closed.
```
Rich-Harris, line 20: how about adding this? it makes it easier to create repros, easier for us to verify them, and reduces repo clutter (both for us and for people opening issues)

```suggestion
      description: A link to a repository, or a fork of https://node.new/sveltekit, that reproduces the issue. Explaining how to reproduce is generally not enough. It pushes the burden of creating a reproduction project onto a small set of volunteer maintainers and isn't scalable. If no reproduction is provided, the issue will be closed.
```

bluwy:
I don't mind either ways. Though I feel that closing immediately would be a bit harsh, and then having a repro to re-open again. But if this is the preferred way to encourage repros, I'm all in as well. PS: +1 on the stackblitz link.

Rich-Harris (approved):

Issue #981, opened 2021-04-12
<Routes> received an unexpected slot "default".


vfilatov:
**Describe the bug**
Recently appeared in a dev tool three warnings.

**Logs**
<Routes> received an unexpected slot "default".    index.mjs?v=76284753:1656 
<Routes> was created with unknown prop 'status'    index.svelte:2
<Routes> was created with unknown prop 'error'    index.svelte:2

**To Reproduce**
npm init svelte@next
npm install
npm run dev
in browser, open dev tool and navigate to http://localhost:3000

frederikhors:
**Describe the bug**
From `1.0.0-next.71` to `1.0.0-next.72`:

now I have these warning in console:

```
<Routes> received an unexpected slot "default".
<Routes> was created with unknown prop 'status'.
<Routes> was created with unknown prop 'error'.
```

![image](https://user-images.githubusercontent.com/41120635/114468405-b3bc1600-9beb-11eb-8768-401ef01e685c.png)

The code is the same.

vfilatov:
looks fixed in "SvelteKit v1.0.0-next.74"

vfilatov:
I was wrong, it still throws a warning.
`
<Routes> received an unexpected slot "default".
`

mcdba:
1.0.0-next.74 this warning too

higoka:
message still exists in `v1.0.0-next.77`

rchrdnsh:
Still exists in 78 as well.

asv1:
80 still show message.

Rich-Harris:
People, you don't need to comment saying that the warning is still there. As long as the issue remains open you can assume it's still there — that's how issues work :)

rchrdnsh:
apologies, thought I saw someone mention somewhere that they thought the issue had gone away as of version 70 something...

Anyhoo, here is maybe some more useful information :-)

This seems to happen for me for slots and props as well...and as many props as are being added in a module script, as the following video illustrates (all the frontmatter in mdsvex files are getting added as unknown props in this case):

https://user-images.githubusercontent.com/7884263/115158638-ff374f80-a043-11eb-8582-5c56cc0df274.mp4

BUT! ...it doesn't happen in production at all, at least for me :-)

https://user-images.githubusercontent.com/7884263/115159282-44a94c00-a047-11eb-96ba-1b40ee4a84ef.mp4

[...]

PR #3115, opened 2021-12-27
fix: <Route> received an unexpected slot "default"

bfanger:
Previously:
```svelte
<Route>
  {#if subroute}
    <Subroute />
  {/if}
</Route>
```
if the subroute is falsy an empty default slot is passed which causes the warning. https://github.com/sveltejs/kit/issues/981

This change renders:

```svelte
{#if subroute}
  <Route>
    <Subroute />
  </Route>
{:else}
  <Route />
{/if}
```

**Caveat:**. 
Svelte as has no VirtualDOM, so when a route toggles between having a subroute and not having subroute the <Route> component will be remounted and internal state is lost. This is could be a non-issue, as I was not able to produce that scenario.

### DIFF
--- a/.changeset/moody-students-report.md
+++ b/.changeset/moody-students-report.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-fix <Route> received an unexpected slot "default"
+fix `<Route> received an unexpected slot "default"` warning
--- a/.changeset/moody-students-report.md
+++ b/.changeset/moody-students-report.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+fix <Route> received an unexpected slot "default"
--- a/packages/kit/src/core/create_app/index.js
+++ b/packages/kit/src/core/create_app/index.js
@@ -130,11 +130,13 @@ function generate_app(manifest_data) {
 
 	while (l--) {
 		pyramid = `
-			<svelte:component this={components[${l}]} {...(props_${l} || {})}>
-				{#if components[${l + 1}]}
+			{#if components[${l + 1}]}
+				<svelte:component this={components[${l}]} {...(props_${l} || {})}>				
 					${pyramid.replace(/\n/g, '\n\t\t\t\t\t')}
-				{/if}
-			</svelte:component>
+				</svelte:component>
+			{:else}
+				<svelte:component this={components[${l}]} {...(props_${l} || {})} />
+			{/if}
 		`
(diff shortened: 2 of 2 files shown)


packages/kit/src/core/create_app/index.js
```diff
@@ -130,11 +130,13 @@ function generate_app(manifest_data) {
...
-				{#if components[${l + 1}]}
+			{#if components[${l + 1}]}
+				<svelte:component this={components[${l}]} {...(props_${l} || {})}>				
 					${pyramid.replace(/\n/g, '\n\t\t\t\t\t')}
-				{/if}
-			</svelte:component>
+				</svelte:component>
+			{:else}
+				<svelte:component this={components[${l}]} {...(props_${l} || {})} />
+			{/if}
```
bluwy, line 139: I used to do this trick in my other projects, but a caveat I found is that this would also rerender the component if only the child components change, which is sometimes undesirable, especially for route transitions. And so, I think we should fix the warning in Svelte side, and leave the code as is for now to prevent a potential breaking change.

bfanger: When does that happen? Are there situations where the upper level component stays the same but the underlying level changes to falsy?

Rich-Harris: I'm also curious about this — if we can fix it here rather than in Svelte, that would be preferable I think, since the warning is useful.

The `components` 'branch' is an array of nodes, where every node except the leaf is a layout. Layouts always have a slotted child, leaves never do. Which means that I don't think there's ever a scenario in which a component could move between the consequent and the alternate — layouts will _always_ be in the consequent, leaves will _always_ be in the alternate.

In other words this is a brilliantly simple (and in hindsight, obvious!) fix for this issue as far as I can see — thank you @bfanger! But I'd love to better understand @bluwy's concern before merging.

bluwy: Ah thanks for the explanation. I wasn't sure what the `components` array would contain, but if the quirk I mention never happens. This looks good to me!

.changeset/moody-students-report.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+fix <Route> received an unexpected slot "default"
```
bluwy, line 5:
```suggestion
fix `<Route> received an unexpected slot "default"` warning
```

We probably want to wrap this with backticks. I don't think the changelog would render `<Route>` correctly without backticks.

bluwy (approved):

PR #3117, opened 2021-12-27
[chore] do all middleware munging in the Vite plugin

benmccann:
We should do less in the CLI and more in the Vite plugin. This makes it so that `init_server` now does nothing but build the Vite config and use it to start a server.

No changelog on this PR since it should not introduce any user-visible changes

### DIFF
--- a/packages/kit/src/core/dev/index.js
+++ b/packages/kit/src/core/dev/index.js
@@ -180,7 +180,6 @@ class Watcher extends EventEmitter {
 		}
 
 		this.vite = await vite.createServer(merged_config);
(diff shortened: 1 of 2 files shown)


bluwy (approved):
@dominikg Would this allow auto-restart on config changed for kit? Moving the middleware removal code into `configureServer` might make it work I think.

dominikg:
> @dominikg Would this allow auto-restart on config changed for kit? Moving the middleware removal code into `configureServer` might make it work I think.

Yes, might work. Maybe the config stuff needs to be moved to a config hook too

PR #3066, opened 2021-12-17
Prerender pages in parallel

kauppilainen:
#3065 

This PR enables prerendering pages in parallel. 
When enabling prerendering the entries are now batched and prerendered in parallel. Errors emerged when running more than 10 entries at once, thus the batch size. The batch size should maybe be configurable or not needed.
```
SyntaxError: Unexpected token < in JSON at position 0
  at JSON.parse (<anonymous>)
  at Response.json (file:///Users/felix/Projects/kit/packages/kit/dist/install-fetch.js:4868:15)
  at runMicrotasks (<anonymous>)
  at processTicksAndRejections (node:internal/process/task_queues:96:5)
  at async getAllEntities (file:///Users/felix/Projects/kit/examples/web/.svelte-kit/output/server/chunks/wordpress-cb2619c2.js:12:16)
  at async getAllProductsWithFilterParams (file:///Users/felix/Projects/kit/examples/web/.svelte-kit/output/server/chunks/woocommerce-a9f9aaa5.js:41:34)
  at async Promise.all (index 0)
  at async get (file:///Users/felix/Projects/kit/examples/web/.svelte-kit/output/server/chunks/_slug_.json-c5701545.js:10:40)
  at async render_endpoint (file:///Users/felix/Projects/kit/examples/web/.svelte-kit/output/server/chunks/app-52ab529a.js:96:20)
  at async resolve (file:///Users/felix/Projects/kit/examples/web/.svelte-kit/output/server/chunks/app-52ab529a.js:1259:56)
```

`dev:basics` tests are currently failing but to my understanding that's just for my setup and not any project specific. Tests regarding the default options are all passing.

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`
- [ ] Pass tests

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #3120, opened 2021-12-28
Concurrency

### LINKED ISSUE #3065: Statically generate pages in parallel at build time
### Describe the problem

It takes a long time to complete builds for sites with many statically generated pages (800 pages in 40 minutes ).

A flag to run the prerendering in parallel.

### Alternatives considered

Reach the same user experience by relying on CDN to cache server side rendered html instead of generating it at build time.

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
Supersedes #3066, closes #3065. Adds a new `config.kit.prerender.concurrency` setting, which controls how many pages can be prerendered simultaneously (default `1`). 

### Tests

### Changesets

### DIFF
--- a/documentation/docs/14-configuration.md
+++ b/documentation/docs/14-configuration.md
@@ -41,6 +41,7 @@ const config = {
 			base: ''
 		},
 		prerender: {
+			concurrency: 1,
 			crawl: true,
 			enabled: true,
 			entries: ['*'],
@@ -55,7 +56,7 @@ const config = {
 		trailingSlash: 'never',
 		vite: () => ({})
 	},
-	
+
 	// SvelteKit uses vite-plugin-svelte. Its options can be provided directly here.
 	// See the available options at https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md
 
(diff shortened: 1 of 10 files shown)


documentation/docs/14-configuration.md
```diff
@@ -163,6 +164,7 @@ An object containing zero or more of the following `string` values:
 
 See [Prerendering](#ssr-and-javascript-prerender). An object containing zero or more of the following:
 
+- `concurrency` — how many pages can be prerendered simultaneously. In cases where prerendering performance is network-bound (for example loading content from a remote CMS) this can speed things up
```
benmccann, line 167: In my Java mindset, I was thinking maybe this should be called `threads` or something more specific, but then I remembered the JS is single threaded and it's the event loop we're taking advantage of. Maybe it's worth adding a bit more detail to the docs?
```suggestion
- `concurrency` — how many pages can be prerendered simultaneously. JS is single-threaded, but in cases where prerendering performance is network-bound (for example loading content from a remote CMS) this can speed things up by processing other tasks while waiting on the network response
```

benmccann (approved):
lgtm

documentation/docs/14-configuration.md
mquandalle: Multi-threading is possible in JS with workers.

On that matter https://github.com/Aslemammad/tinypool looks like an interesting library (even if it is way more complex than the queue implemented in this PR). It is used by [vitest](https://vitest.dev/) to run tests in parallel.

Issue #269, opened 2020-12-14
Page metadata


Conduitry:
**Is your feature request related to a problem? Please describe.**
Components that are used in layouts currently don't have access to which route they're being used on. The `$page` store tells us the underlying `path` and the `params` that have been extracted from that path, but doesn't tell us anything about which route those params have meaning in.

When making a navbar in Sapper, I generally have to re-analyze `$page.path` with regexes or string comparisons or whatever in order to tell which page I'm on and what state the navbar should be in.

**Describe the solution you'd like**
Since the framework has already decided which route this request maps to, it can expose this information in the `$page` store. The most immediately obvious way to me for this to be handled would be for the relative path to the `.svelte` file that is the page route be included as a string in (say) `$page.route`.

**Describe alternatives you've considered**
If we didn't want the inelegance or fragility of exposing file paths as strings, would be to let each page component expose - in some way - a user-specified key that Kit doesn't do anything with apart from exposing in the `$page` store. This could be a separate module context export, or it could be another value in the `load` response, or maybe something else. This was also suggested in https://github.com/sveltejs/kit/issues/1459

**How important is this feature to you?**
It'd be nice to have. That my code needs to re-figure out something about what page I'm on that the framework has already determined and then kept for itself is irritating.

**Additional context**
None that I can think of right now.

Conduitry:
This would be more of an issue with route fallthrough (#548) because there's no longer necessarily a quick way to tell routes apart looking just at their paths.

alteredorange:
I think this is important, because it looks like sveltekit doesn't export segment like sapper did. (At least segment is always undefined when I export it from $layout). Currently I'm using `$page.path.substring(1)` for all my route needs, but it would be great if there was an actual route param, as I'm not sure the substring works with all edge cases.

Wolfr:
@alteredorange This code helped me implement a simple selected state in my SvelteKit demo app. Thank you.

Brawl345:
Using $page.path works, but I want to e.g. see `$error` there when I'm on an error page. Currently, there is no way to do that afaik? Next.js has `router.route` for example

Conduitry:
Exposing error information has its own issue, #598.

gterras:
Apologies if I'm missing something but from my understanding:

The ability to get `the relative path to the .svelte file` is crucial and the only way to properly generate `hrefs`  without relying on any kind of hardcoded values (like current folder depth or name). See this common use case (caps for clarity): 

```
│── index.svelte
[...]

PR #3126, opened 2021-12-29
replace request.origin/path/query with request.url

### LINKED ISSUE #3078: Full Request URL Data
### Describe the problem

I redirect users to external auth providers that accept a redirect url as a query parameter. In order to do this from a hook I need to rebuild the url by combining the various pieces by hand and some environment information to determine protocol. This is inconvenient and doesn't easily allow using the platform.

I'd like a `url: URL` property on the `ServerRequest`. It _does_ mean all the adapters need tweaking, but it feels like a reasonable need for a web server.

### Alternatives considered

Building the full URL by hand.

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
This is a substantial breaking change, but a welcome one — it simplifies things a bit. It replaces `request.origin`, `request.path` and `request.query` with a single `request.url`, which is a standard `URL` object.

Similarly, page `load` functions now take a `{ url, params, ... }` object rather than `{ page, ... }`, and the `page` store has been replaced with `url` and `params` stores.

@Conduitry suggested we take this opportunity to rethink the naming of `params`, though I haven't had any particularly great ideas for what to rename it to.

Closes #3078.

### Tests

### Changesets


examples/hn.svelte.dev/src/routes/index.svelte
```diff
@@ -1,6 +1,7 @@
 <script context="module">
-	export function load({ page }) {
-		let host = page.host;
+	/** @type {import('@sveltejs/kit').Load} */
+	export function load({ url }) {
+		let host = url.host;
```
Conduitry, line 4: This could probably be simplified to use `url.hostname`.

packages/kit/src/core/adapt/prerender.js
```diff
@@ -330,11 +329,10 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 	if (fallback) {
 		const rendered = await app.render(
 			{
+				url: `${config.kit.host || 'sveltekit'}://${config.kit.host || 'prerender'}/[fallback]`,
```
Conduitry, line 332: Is the `[fallback]` thing still necessary here? Is anything looking at that?

packages/kit/src/runtime/client/renderer.js
```diff
@@ -451,19 +444,19 @@ export class Renderer {
...
+		if (!this.current.params || JSON.stringify(params) !== JSON.stringify(this.current.params)) {
+			result.props.params = params;
 		}
 
 		const leaf = filtered[filtered.length - 1];
 		const maxage = leaf.loaded && leaf.loaded.maxage;
 
 		if (maxage) {
-			const key = `${page.path}?${page.query}`;
+			const key = url.origin + url.pathname; // omit hash
```
Conduitry, line 459: Do we want to include `+ url.search` here like we had before?

Conduitry:
I didn't have any great ideas of what to rename `params` to either, unfortunately. My concern was that, since it was no longer part of the `$page` store and no longer alongside other URL-y things, the name `params` would be a bit vague. I had half-heartedly suggested `slugs` but Rich wasn't any more thrilled with that one than I was.

examples/hn.svelte.dev/src/routes/index.svelte
Rich-Harris: good catch. i have a feeling this `url` thing will pay all sorts of dividends

packages/kit/src/core/adapt/prerender.js
Rich-Harris: the string is mostly there to keep the typechecker happy. i chose this in case someone logged it out in `__layout` for some reason — probably totally unnecessary but couldn't think of anything more suitable

packages/kit/src/runtime/client/renderer.js
Rich-Harris: whoops, that was meant to be `pathname + search` not `origin + pathname`, d'oh

Rich-Harris:
Still to do:

* [ ] fix the one remaining broken test (change detection for `load` functions is less granular than previously — need to either make it more granular, or change the test)
* [ ] docs
* [ ] add getters on the `request`/`load` arguments so that people currently using `path`, `query` and `origin` understand why their app broke

documentation/docs/03-loading.md
```diff
@@ -90,18 +86,24 @@ It is recommended that you not store pre-request state in global variables, but
...
-- `page.path === '/blog/some-post'`
-- `page.params.slug === 'some-post'`
-- `page.query.get('foo') === 'bar'`
-- `page.query.has('baz')`
-- `page.query.getAll('bizz') === ['a', 'b']`
+#### params
+
+`params` is derived from `url.pathname` and the route filename.
+
+For a route filename example like `src/routes/a/[b]/[c]` and a `url.pathname` of `/a/x/y/z`, the `params` object would look like this:
```
Conduitry, line 99: `[...c]`?

Rich-Harris: indeed

Rich-Harris:
The template app preview deployments will continue to fail until this actually gets released, which makes me think that the whole idea of preview deployments in this context is a bit... pointless. unless we can configure them all to use pnpm instead

Conduitry:
Maybe `$params` could be something like `$route.params`? So, for now it would be an object with one key, but this could eventually be expanded to things like #269 and maybe #3096? Is that too crazy?

Rich-Harris:
oh, i like that a lot. yes, making that change

Conduitry:
Is it possible/desirable to only define the getters for the old `page`/`params` stuff in dev mode?

Also (pet peeve alert) can we only use the word 'deprecate' when something does still work but we display a warning about how it won't work sometime in the future? If something just plain old doesn't work anymore, that's not a deprecation.

Rich-Harris:
done, done and done

documentation/docs/03-loading.md
```diff
@@ -13,12 +13,8 @@ export interface LoadInput<
...
 	Session = any
 > {
-	page: {
-		origin: string;
-		path: string;
-		params: PageParams;
-		query: URLSearchParams;
-	};
+	url: URL;
+	params: PageParams;
```
Conduitry, line 17: This needs updating.

documentation/docs/03-loading.md
```diff
@@ -88,20 +84,26 @@ It is recommended that you not store pre-request state in global variables, but
...
+The `load` function receives an object containing four fields — `page`, `fetch`, `session` and `stuff`. The `load` function is reactive, and will re-run when its parameters change, but only if they are used in the function. Specifically, if `url`, `session`, or `stuff` are used in the function, they will be re-run whenever their value changes. Note that destructuring parameters in the function declaration is enough to count as using them. In the example above, the `load({ params, fetch, session, stuff })` function will re-run every time `session` or `stuff` is changed, even though they are not used in the body of the function. If it was re-written as `load({ params, fetch })`, then it would only re-run when `params.slug` changes. The same reactivity applies to `params`, but only to the params actually used in the function. If `params.foo` changes, the example above would not re-run, because it did not access `params.foo`, only `params.slug`.
```
Conduitry, line 87: This was only partly updated.

Rich-Harris:
I wonder if `$route` should contain `url` as well as `params`? It would mean that `url` and `params` were back to being 'siblings' (as they are in endpoints and `load`), and would also solve an edge case I was sort of thinking about, which is that if you were manually subscribing to the stores (rather than relying on Svelte's update cycle), `$route` and `$url` would be out of sync:

```js
route.subscribe($route => {
  // this will update before `url`, or after it
});

url.subscribe($url => {
  // this will update before `route`, or after it
});
```

The downside is that you have to type `$route.url.pathname` instead of `$url.pathname`, but that's probably fine?

documentation/docs/03-loading.md
```diff
@@ -88,20 +84,26 @@ It is recommended that you not store pre-request state in global variables, but
...
-So if the example above was `src/routes/blog/[slug].svelte` and the URL was `https://example.com/blog/some-post?foo=bar&baz&bizz=a&bizz=b`, the following would be true:
+> In some environments this is derived from request headers, which you [may need to configure](#configuration-headers), during server-side rendering
 
-- `page.origin === 'https://example.com'`
-- `page.path === '/blog/some-post'`
-- `page.params.slug === 'some-post'`
-- `page.query.get('foo') === 'bar'`
-- `page.query.has('baz')`
-- `page.query.getAll('bizz') === ['a', 'b']`
+#### params
```
Conduitry, line 95: This needs updating.

documentation/docs/03-loading.md
Rich-Harris: It does? This is the `load` arguments, not the stores — looks correct to me?

documentation/docs/03-loading.md
Rich-Harris: see previous response — I think it's correct?

documentation/docs/05-modules.md
```diff
@@ -52,12 +52,13 @@ Stores are _contextual_ — they are added to the [context](https://svelte.dev/t
...
+- `getStores` is a convenience function around `getContext` that returns `{ navigating, url, route, session }`. This needs to be called at the top-level or synchronously during component or page initialisation.
 
-The stores themselves attach to the correct context at the point of subscription, which means you can import and use them directly in components without boilerplate. However, it still needs to be called synchronously on component or page initialisation when `$`-prefix isn't used. Use `getStores` to safely `.subscribe` asynchronously instead.
+The stores themselves attach to the correct context at the point of subscription, which means you can import and use them directly in components without boilerplate. However, it still needs to be called synchronously on component or page initialisation when the `$`-prefix isn't used. Use `getStores` to safely `.subscribe` asynchronously instead.
 
 - `navigating` is a [readable store](https://svelte.dev/tutorial/readable-stores). When navigating starts, its value is `{ from, to }`, where `from` and `to` both mirror the `page` store value. When navigating finishes, its value reverts to `null`.
```
Conduitry, line 59: This needs updating. From what I saw of the code, I think `from` and `to` are now both `URL` instances, correct? This would mean that you no longer have access to the `params`. Is that fine?

Rich-Harris: I actually think it's better — IIUC `params` could be incorrect in `to`, because you can't compute `params` until you've definitively selected a route

documentation/docs/03-loading.md
Conduitry: Oh! I guess I was assuming the `load` arguments would also reflect the available stores. Are they not anymore? Is that confusing? (It was confusing to me just now, anyway.)

Rich-Harris: see https://github.com/sveltejs/kit/pull/3126#issuecomment-1002753326

documentation/docs/03-loading.md
Rich-Harris: should be all good now i think. i took the opportunity to slim it down a bit as well

documentation/docs/03-loading.md
Conduitry: I see. I had skimmed that, and had intended to circle back to that afterwards, but I apparently did not skim it well enough. Hm. I guess for consistency what I would argue for is a `route` argument to `load` that contains `url` and `params`, but that might be more inconvenient than it's worth. Although, if you're not worried about `$route.url.pathname`, why be worried about it in `load`?

documentation/docs/05-modules.md
Rich-Harris: (this was updated)

pnpm-lock.yaml
```diff
@@ -222,8 +222,8 @@ importers:
       '@lukeed/uuid': 2.0.0
       cookie: 0.4.1
     devDependencies:
-      '@sveltejs/adapter-auto': 1.0.0-next.5
-      '@sveltejs/kit': 1.0.0-next.203_svelte@3.44.2
+      '@sveltejs/adapter-auto': link:../../../adapter-auto
```
Conduitry, line 225: What's this change? This looks fishy.

packages/kit/src/runtime/client/renderer.js
```diff
@@ -596,19 +591,15 @@ export class Renderer {
 		const [pattern, a, b, get_params] = route;
 		const params = get_params
 			? // the pattern is for the route which we've already matched to this path
-			  get_params(/** @type {RegExpExecArray}  */ (pattern.exec(decoded_path)))
+			  get_params(/** @type {RegExpExecArray}  */ (pattern.exec(url.pathname))) // TODO this needs to account for paths.base
```
benmccann, line 607: `pathname` is not decoded and so I think this will break when you try to do the route matching. I'm not sure you can discard `decoded_path`. I'd be hesitant to change this as it's been a big source of bugs. The server-side is pretty solid at this point, but I don't know if I wrote as many tests for the client-side

pnpm-lock.yaml
Rich-Harris: idk, i've noticed that this sort of thing just happens from time to time

packages/kit/src/runtime/client/types.d.ts
```diff
@@ -1,12 +1,10 @@
...
-import { Page } from 'types/page';
 
 export type NavigationInfo = {
 	id: string;
 	routes: CSRRoute[];
-	path: string;
-	decoded_path: string;
-	query: URLSearchParams;
+	url: URL;
+	// decoded_path: string; TODO do we need this?
```
benmccann, line 7: Without `decoded_path` to cache the result of the decoding we're having to do the decoding twice now. Not sure if that's a concern or not. I'd say we should either restore this for the performance enhancement or remove this line and the TODO

packages/kit/src/runtime/server/utils.js
```diff
@@ -15,6 +15,7 @@ export function decode_params(params) {
 	for (const key in params) {
 		// input has already been decoded by decodeURI
 		// now handle the rest that decodeURIComponent would do
+		// TODO shouldn't we just use decodeURIComponent in the first place?
```
benmccann, line 18: answered this on Discord. no, unfortunately we can't. So I think we can probably remove this TODO

packages/kit/src/runtime/client/renderer.js
Rich-Harris: fixed this and added a test

packages/kit/src/runtime/client/types.d.ts
Rich-Harris: yeah i was working on this as you were typing the comment :)

packages/kit/src/runtime/server/utils.js
Rich-Harris: removed

documentation/docs/05-modules.md
```diff
@@ -48,16 +48,16 @@ import { base, assets } from '$app/paths';
 import { getStores, navigating, page, session } from '$app/stores';
 ```
 
-Stores are _contextual_ — they are added to the [context](https://svelte.dev/tutorial/context-api) of your root component. This means that `session` and `page` are unique to each request on the server, rather than shared between multiple requests handled by the same server simultaneously, which is what makes it safe to include user-specific data in `session`.
+Stores are _contextual_ — they are added to the [context](https://svelte.dev/tutorial/context-api) of your root component. This means that `session` and `route` are unique to each request on the server, rather than shared between multiple requests handled by the same server simultaneously, which is what makes it safe to include user-specific data in `session`.
```
benmccann, line 51: I think this should be `page`?
```suggestion
Stores are _contextual_ — they are added to the [context](https://svelte.dev/tutorial/context-api) of your root component. This means that `session` and `page` are unique to each request on the server, rather than shared between multiple requests handled by the same server simultaneously, which is what makes it safe to include user-specific data in `session`.
```

documentation/migrating/04-pages.md
```diff
@@ -33,9 +33,9 @@ import { stores } from '@sapper/app';
 const { preloading, page, session } = stores();
 ```
 
-The `page` and `session` stores still exist; `preloading` has been replaced with a `navigating` store that contains `from` and `to` properties.
+The `page` and `session` stores still exist; `preloading` has been replaced with a `navigating` store that contains `from` and `to` properties. `page` now has `url` and `params` properties, but no `path` or `query`.
 
-You access them differently in SvelteKit. `stores` is now `getStores`, but in most cases it is unnecessary since you can import `navigating`, `page` and `session` directly from [`$app/stores`](/docs#modules-$app-stores).
+You access them differently in SvelteKit. `stores` is now `getStores`, but in most cases it is unnecessary since you can import `navigating`, `route` and `session` directly from [`$app/stores`](/docs#modules-$app-stores).
```
benmccann, line 38: `page` here as well?
```suggestion
You access them differently in SvelteKit. `stores` is now `getStores`, but in most cases it is unnecessary since you can import `navigating`, `page` and `session` directly from [`$app/stores`](/docs#modules-$app-stores).
```

packages/kit/types/hooks.d.ts
```diff
@@ -1,11 +1,13 @@
...
-import { MaybePromise, ResponseHeaders } from './helper';
+import { ParameterizedBody, RawBody } from './app';
+import { MaybePromise, RequestHeaders, ResponseHeaders } from './helper';
 
 export type StrictBody = string | Uint8Array;
 
-export interface ServerRequest<Locals = Record<string, any>, Body = unknown>
-	extends IncomingRequest {
-	origin: string;
+export interface ServerRequest<Locals = Record<string, any>, Body = unknown> {
```
benmccann, line 6: this could still extend `IncomingRequest` since the first four fields are shared. It makes the docs look nicer that they're not shared, but I wonder if that's a good enough reason to avoid the inheritance. I really just want the typedocs autogenerated :smile:

Rich-Harris: auto-generated docs are the worst! The types will diverge further when we tackle #70, so I figured we should do what's right for readability here rather than brevity

silentworks:
@Rich-Harris I know this is closed but am I correct in thinking the original statement of `...and the page store has been replaced with url and params stores` no longer applies since it seems `page` store still exists in the latest version?

Rich-Harris:
Yep, classic case of the comments not matching the code. Last-minute bikeshedding

Issue #2793, opened 2021-11-13
Add option to use polling instead of filesystem events when watching for filesystem changes


qwerty01:
### Describe the problem

Currently sveltekit uses `cheap-watch` to detect when files have changed while running the dev server (https://github.com/sveltejs/kit/blob/master/packages/kit/src/core/dev/index.js#L76). `cheap-watch` makes use of `fs.watch`, which uses a system-specific event callback when a file changes.

Unfortunately, when listening for events on a network drive (such as when using docker with a mapped folder), no events are emitted when files are changed. Other similar tools have an option to use filesystem polling instead of event callbacks, such as chokidar: https://github.com/paulmillr/chokidar#performance:
> usePolling (default: false). Whether to use fs.watchFile (backed by polling), or fs.watch. If polling leads to high CPU utilization, consider setting this to false. It is typically necessary to set this to true to successfully watch files over a network

### Describe the proposed solution

I believe the most flexible solution would be to switch to a different filesystem watcher that supports polling as well as filesystem events (such as chokidar). Configuring the module could then be achieved through the svelte.config.js file and a link directly to the module's documentation can be added to the current documentation for configuring watch settings.

### Alternatives considered

Another option could be to use a secondary module for polling the filesystem and have a config option in svelte.config.js to either use polling or filesystem events.

### Importance

would make my life easier

### Additional Information

The current use case for me is to enable developing in docker. Currently, filesystem watching doesn't work at all due to network shares not triggering filesystem events, so HMR does not work. Everything works when running locally so it's not a showstopper, but it would be nice to have support for dockerized development environments.

bluwy:
One possible solution is to use Vite's file watcher from the dev server as well, which uses chokidar. I'm not sure how that would affect HMR, but it may be doable.

Rich-Harris:
in https://github.com/sveltejs/kit/pull/3138 we switched to using Vite's chokidar instance, so you can configure the watcher via the Vite config: https://vitejs.dev/config/#server-watch

PR #3129, opened 2021-12-29
[refactor] create ViteDevServer in dev function instead of in Watcher

benmccann:
It doesn't really make make sense for the server to be created inside the file system watcher. This makes them both be created in `dev`. The next step might be to see if we can create the watcher inside the Vite plugin. At that point, SvelteKit dev mode would really be doing nothing except for creating Vite config and plugin.


dominikg:
could we use the watcher exposed by vite in server.watcher , accessible from configureServer hook? 

https://vitejs.dev/guide/api-javascript.html#vitedevserver

Rich-Harris:
Unfortunately this is too far out of date since #2931 and would need to be redone.

I'm not 100% sure what it's buying us, though? The dev service is comprised of two parts — a filewatcher, and a Vite server — and that's true whether they belong to a single class defined in `src/core/dev/index.js` or two separate objects both of which are returned to the caller in `src/cli.js`. Personally I prefer to have `src/cli.js` know as little as possible about the implementation of the various commands.

> It doesn't really make make sense for the server to be created inside the file system watcher

This isn't what's happening — the file system watcher _and_ the server are both being created inside the `Watcher` class; they're siblings. If it's just a case of bad naming then we could call it `DevService` or something instead?

Rich-Harris:
> could we use the watcher exposed by vite in server.watcher , accessible from configureServer hook?

yeah, that would make sense. I prefer `cheap-watch` generally, but there probably isn't anything to be gained by using it when there's already a chokidar instance lying around

Rich-Harris:
Experimenting locally with using the Vite filewatcher. One wrinkle I've found: because Vite's own handlers are invoked first, if you delete a page you'll get an error in the console telling you that an import can't be resolved, because the manifest hasn't yet been updated to remove that import. Not sure if that's something we have any control over?

Conduitry:
I won't be mortally offended if SvelteKit stops using cheap-watch - it's had way more of an exciting life and way more npm downloads already than I ever expected it to.

There was also an issue #2793 recently about allowing polling instead of filesystem events, and I think that could probably also be resolved if we were using Vite's own copy of Chokidar?

dominikg:
> There was also an issue #2793 recently about allowing polling instead of filesystem events, and I think that could probably also be resolved if we were using Vite's own copy of Chokidar?

yes: https://vitejs.dev/config/#server-watch

example: https://github.com/sveltejs/vite-plugin-svelte/blob/c92f1bf82d5fafa8c08de463879e5d2b2728583a/packages/e2e-tests/hmr/vite.config.js#L20

Rich-Harris:
> because the manifest hasn't yet been updated to remove that import. Not sure if that's something we have any control over?

Actually this seems unrelated — I can create the same error on `master`. Something to do with the client-side `.svelte-kit/dev/generated/manifest.js` file not being invalidated correctly

benmccann:
My first commit on this PR made more sense, but the second one wasn't in a great state and then I had to go to bed. Thanks Rich for taking over on this one. I'll close in favor of https://github.com/sveltejs/kit/pull/3138

PR #3133, opened 2021-12-30
Update adapters to use new app.render API

Rich-Harris:
Follow-up to #3126 — adds missing changeset and updates adapters

### Tests

### Changesets


documentation/docs/05-modules.md
```diff
@@ -48,7 +48,7 @@ import { base, assets } from '$app/paths';
 import { getStores, navigating, page, session } from '$app/stores';
 ```
 
-Stores are _contextual_ — they are added to the [context](https://svelte.dev/tutorial/context-api) of your root component. This means that `session` and `page` are unique to each request on the server, rather than shared between multiple requests handled by the same server simultaneously, which is what makes it safe to include user-specific data in `session`.
+Stores are _contextual_ — they are added to the [context](https://svelte.dev/tutorial/context-api) of your root component. This means that `session` and `route` are unique to each request on the server, rather than shared between multiple requests handled by the same server simultaneously, which is what makes it safe to include user-specific data in `session`.
```
Rich-Harris, line 51: oops, this error snuck back somehow

```suggestion
Stores are _contextual_ — they are added to the [context](https://svelte.dev/tutorial/context-api) of your root component. This means that `session` and `page` are unique to each request on the server, rather than shared between multiple requests handled by the same server simultaneously, which is what makes it safe to include user-specific data in `session`.
```

documentation/migrating/04-pages.md
```diff
@@ -35,7 +35,7 @@ const { preloading, page, session } = stores();
 
 The `page` and `session` stores still exist; `preloading` has been replaced with a `navigating` store that contains `from` and `to` properties. `page` now has `url` and `params` properties, but no `path` or `query`.
 
-You access them differently in SvelteKit. `stores` is now `getStores`, but in most cases it is unnecessary since you can import `navigating`, `page` and `session` directly from [`$app/stores`](/docs#modules-$app-stores).
+You access them differently in SvelteKit. `stores` is now `getStores`, but in most cases it is unnecessary since you can import `navigating`, `route` and `session` directly from [`$app/stores`](/docs#modules-$app-stores).
```
Rich-Harris, line 38: Ditto

```suggestion
You access them differently in SvelteKit. `stores` is now `getStores`, but in most cases it is unnecessary since you can import `navigating`, `page` and `session` directly from [`$app/stores`](/docs#modules-$app-stores).
```

PR #3138, opened 2021-12-30
[chore] dev mode refactor

Rich-Harris:
WIP. follow-up to #3129. Using Vite's watcher means we can push more logic into the plugin. Also, `Watcher` extends `EventEmitter` but we're not actually emitting any events, so we can make all that stuff way simpler

### Tests

### Changesets


packages/kit/src/core/dev/index.js
```diff
@@ -1,11 +1,9 @@
-import { EventEmitter } from 'events';
 import fs from 'fs';
 import path from 'path';
 import { URL } from 'url';
 
 import { svelte } from '@sveltejs/vite-plugin-svelte';
 import amp_validator from 'amphtml-validator';
-import CheapWatch from 'cheap-watch';
```
benmccann, line 8: this can be removed from the `package.json`

packages/kit/src/core/dev/index.js
```diff
@@ -475,43 +322,113 @@ async function create_plugin(config, dir, https, get_manifest) {
...
-		configureServer(vite) {
-			return () => {
-				remove_html_middlewares(vite.middlewares);
-				vite.middlewares.use(create_kit_middleware(vite));
-			};
-		}
+		address_info,
+		server_config: vite_config.server,
+		allow,
+		cwd,
```
benmccann, line 405: I'm not sure that `cwd` or `close` are used?

Rich-Harris: `close` is (by the tests). i'm still working on this PR, it's probably not worth reviewing until it's marked ready — more changes to come

Rich-Harris:
This will fix #2793

Rich-Harris:
Bit weirded out by these deployment failures. Investigating

Rich-Harris:
deploys all passing — issue turned out to be netlify being a bit overzealous with the build cache

benmccann (approved):

Issue #2547, opened 2021-10-02
`svelte-kit package` requires there be an app.html file


chanced:
### Describe the bug

`svelte-kit package` requires there be an app.html file.

### Reproduction

```#bash
npm init svelte@next project-name
cd project-name
npm i && npm i -D svelte2tsx typescript
rm src/app.html
echo "import preprocess from 'svelte-preprocess'; \n/** @type {import('@sveltejs/kit').Config} */ \nconst config = { preprocess: preprocess() }; \nexport default config;" > svelte.config.js
npm exec svelte-kit package
```

### Logs

```shell
▶ npm exec svelte-kit package
src/app.html does not exist
Error: src/app.html does not exist
    at validate_template (file:///Users/chance/Projects/spike/svelte-kit-without-apphtml/node_modules/@sveltejs/kit/dist/cli.js:688:9)
    at load_config (file:///Users/chance/Projects/spike/svelte-kit-without-apphtml/node_modules/@sveltejs/kit/dist/cli.js:708:2)
    at async get_config (file:///Users/chance/Projects/spike/svelte-kit-without-apphtml/node_modules/@sveltejs/kit/dist/cli.js:769:10)
    at async file:///Users/chance/Projects/spike/svelte-kit-without-apphtml/node_modules/@sveltejs/kit/dist/cli.js:924:18
```

### System Info

```shell
System:
    OS: macOS 11.5.2
    CPU: (8) x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
    Memory: 105.45 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.9.0 - ~/.nvm/versions/node/v16.9.0/bin/node
    npm: 7.21.1 - ~/.nvm/versions/node/v16.9.0/bin/npm
  Browsers:
    Brave Browser: 79.1.1.23
    Chrome: 94.0.4606.61
    Edge: 92.0.902.78
    Firefox: 91.0.2
    Safari: 14.1.2
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.179 
    svelte: ^3.42.6 => 3.43.1
```

### Severity

annoyance

Conduitry:
[...]

PR #3161, opened 2021-12-31
Add in check_template option

ghostdevv:
Currently when using `svelte-kit package` on it's own it fails due to not finding a `app.html`, this pr makes the package command skip the check for app.html and makes it default for everything else

### Tests

### Changesets


ghostdevv:
Don't believe the netlify things are failing because of this pr

.changeset/famous-donuts-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+add in a check_template option to config loaders
```
bluwy, line 5: I think this should describe from the user point of view, e.g.
```suggestion
make html template optional for `svelte-kit package`
```

Conduitry:
This would resolve #2547.

ghostdevv:
Awesome, I know you mentioned in that issue that you might not want to encourage users to use svelte kit package standalone - for all my other libs I use it all together but for this particular project the documentation makes more sense being in Routify so I want to use svelte kit package essentially standalone

Rich-Harris:
Thank you! Because of this I realised that we're actually doing the validation at the wrong time — in dev, it's possible to _start_ with a valid template, but then make it invalid without raising an error. What we really need to do is validate it at the point of use. I've added a `load_template` function that `dev` and `build` can both use — that way `load_config` doesn't need to know anything about template validity.

Rich-Harris (approved):

ghostdevv:
Amazing thank you

PR #3162, opened 2021-12-31
use `@playwright/test`

Rich-Harris:
I don't yet know if this is a great idea, or even possible, but I thought it would be worth the experiment. The integration tests are, as has been noted, somewhat baroque — we're contorting uvu in all sorts of weird and interesting ways and it feels very duct-taped together. Using Playwright ought to have a number of benefits:

* we're already using it under the hood, this is just a more idiomatic usage
* it actually has [docs](https://playwright.dev/docs/intro) (obviously uvu does too, but not for how we're [ab]using it)
* it will run tests in parallel, so _in theory_ it should be faster
* it potentially provides a model for how to set up tests in the project template
* we can get rid of all our weird duct tape code

Potential downsides:

* Oddly, the first bunch of tests I ported (the `amp` app) run _slower_ now. Remains to be seen if that will continue to be the case for an app with many more tests
* Remains to be seen if having less control over the nuts and bolts makes some things impossible to test
* Converting the tests would be a bunch of manual work

Will convert some more stuff over and see how it feels

### Tests

### Changesets


Rich-Harris:
Preliminary results — time to run `pnpm test:integration` is down to ~35s from ~1m20s. The output is easier to read too. Got a couple of tests still to fix, but on the whole I think this is a good change, even if I will always find this sort of thing completely absurd:

```js
// sensible
assert.ok(path.endsWith('/blah'));

// fucking idiotic
expect(path.endsWith('/blah')).toBe(true);
```

Rich-Harris:
seeing some infuriating flakiness I can't explain

Rich-Harris:
Alright, it looks like the tests are passing now, except for the scroll behaviour tests which I've disabled on CI for now. Will open an issue if we merge this reminding us to investigate later.

That said... it turns out we're not seeing the expected performance improvements, because in CI the tests run with a single worker. It's actually noticeably _slower_ than our current approach (in which tests run serially, but the `page` object is reused between tests). So maybe this whole thing was just an infuriating waste of time?

dominikg (approved):

PR #3171, opened 2022-01-01
[feat] allow returning redirect from `handleError`

### LINKED ISSUE #2824: Have a way to render an error page when there's an exception in a hook (e.g. handle)
### Describe the problem

When an exception occurs in handle you get the exception and stack trace in the response to the user.

It would be great if there was a way to render an error page.

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

nhe23:
Fixes #2824 by allowing the user to optionally return a redirect value for a custom error page. This way an error page can be rendered even if a hook throws an exception.

### Tests

### Changesets


packages/kit/test/apps/basics/src/hooks.js
```diff
@@ -18,6 +18,20 @@ export const handleError = ({ request, error }) => {
 		: {};
 	errors[request.url.pathname] = error.stack || error.message;
 	fs.writeFileSync('test/errors.json', JSON.stringify(errors));
+	console.error(error.message);
```
benmccann, line 21: I'm not sure we'd want to be logging to the console even in tests

packages/kit/test/apps/basics/src/hooks.js
```diff
@@ -18,6 +18,20 @@ export const handleError = ({ request, error }) => {
...
+	if (error.frame) {
+		console.error(error.frame);
+	}
+	if (error.stack) {
+		console.error(error.stack);
+	}
+
+	if (error.message === '"Testing hook exception"') {
```
benmccann, line 29: I wonder if the double quotes are needed here?

documentation/docs/04-hooks.md
```diff
@@ -96,6 +100,12 @@ export interface HandleError<Locals = Record<string, any>, Body = unknown> {
 export async function handleError({ error, request }) {
 	// example integration with https://sentry.io/
 	Sentry.captureException(error, { request });
+
+	//Optionally return a status and redirect destination
```
benmccann, line 104: we always put a space after `//`
```suggestion
	// Optionally return a status and redirect destination
```

packages/kit/test/apps/basics/src/hooks.js
nhe23: They are needed. Somehow the error that gets thrown in the handle hook is wrapped in double quotes. As my changes did not cause this I thought it was the wanted behaviour and did not look into why it is wrapped in double quotes.

packages/kit/test/apps/basics/src/hooks.js
nhe23: Prior to the recent switch to playwright there was no handleErrors hook in the tests. That's why I mimicked the default implementation (which is logging errors to console.error) as some tests were depending on the output to be exactly this way. But with the recent changes this is not needed anymore and I'll remove the console.errors

nhe23:
I have no idea why the `pnpm` check command in the ci fails. It works on my machine and the error saying `Property 'started' does not exist on type 'PlaywrightTestArgs...` does not make any sense to me, as im using 'started' in exactly the same way as other tests.

packages/kit/test/apps/basics/test/test.js
```diff
@@ -620,6 +619,21 @@ test.describe.parallel('Errors', () => {
 			expect(await page.innerHTML('h1')).toBe('401');
 		}
 	});
+
+	test('custom error page is rendered if handle hook throws an error', async ({
+		baseURL,
+		page,
+		javaScriptEnabled,
+		started
```
benmccann, line 627: if you rebase against `master` there is no `started` anymore. You should rebase and then delete this, the `await started()`, and the `waitForTimeout` lines

nhe23: Thank you for the hint. I thought that I already had rebased, but that must have been another branch I was working on ... 🤦

packages/kit/test/apps/basics/src/hooks.js
nhe23: Removed console.error from test

packages/kit/test/apps/basics/test/test.js
nhe23: Now it works. The failing test on windows is also failing in other PRs ... so this is probably not the fault of my changes, right?

Rich-Harris:
closing per https://github.com/sveltejs/kit/pull/3235#issuecomment-1007798838

PR #3178, opened 2022-01-02
fix: change url protocol for prerender

### LINKED ISSUE #3164: `$page.url.pathname` represented as "//prerender/" when building using the static adapter
### Describe the bug

Using the latest versions:

- @sveltejs/kit version 1.0.0-next.211
- @sveltejs/adapter-static version 1.0.0-next.24

In development, `$page.url.pathname` in `routes/index.svelte` will correctly be displayed as "/".

However, when building for production using the static adapter, the same route is represented as "//prerender/".

Inspecting `build/index.html`, the markup has the correct "/" but the script that hydrates the markup has the following:

```js
<script type="module">
  import { start } from "/_app/start-40ff21d5.js";
  start({
    target: document.querySelector("#svelte"),
    paths: { base: "", assets: "" },
    session: {},
    route: true,
    spa: false,
    trailing_slash: "never",
    hydrate: {
      status: 200,
      error: null,
      nodes: [
        import("/_app/layout.svelte-96b29dd0.js"),
        import("/_app/pages/index.svelte-91ca303b.js"),
      ],
      url: new URL("sveltekit://prerender/"),
      params: {},
    },
  });
</script>
```

Evaluating `new URL("sveltekit://prerender/")` shows that the `URL.pathname` is `//prerender/`.

### Workaround

A workaround is to explicitly set `kit.ssr` or `kit.hydrate` to `false`.

### Reproduction

https://github.com/metonym/sveltekit-page.url.pathname

### Logs

_No response_

### System Info

```shell
System:
  OS: macOS 11.4
  CPU: (8) x64 Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
  Memory: 19.40 MB / 16.00 GB
  Shell: 5.8 - /bin/zsh
Binaries:
  Node: 16.13.1 - ~/.nvm/versions/node/v16.13.1/bin/node
  npm: 8.1.2 - ~/.nvm/versions/node/v16.13.1/bin/npm
Browsers:
  Chrome: 96.0.4664.110
  Firefox: 89.0.2
  Safari: 14.1.1
npmPackages:
  @sveltejs/adapter-static: next => 1.0.0-next.24 
  @sveltejs/kit: next => 1.0.0-next.211 
  svelte: ^3.44.0 => 3.44.3
```

### Severity

blocking an upgrade

### Additional Information

_No response_

bluwy:
Fixes #3164

Fore prerenders, change fallback `sveltekit` protocol to `http`. (Thoughts?)

In Node and Safari, the `sveltekit` protocol is recognized and properly parsed. Firefox and Chrome however don't and incorrectly parses it. e.g. `new URL('sveltekit://prerender/')`, Firefox and Chrome reported `//prerender/` as the `pathname`.

Looking at the [spec](https://url.spec.whatwg.org/#url-representation) (the scheme table), I think Node and Safari are correct, since _presumably_ we're not using a domain or ipv4 address as the host.

### Tests

### Changesets

### DIFF
--- a/.changeset/gentle-rules-camp.md
+++ b/.changeset/gentle-rules-camp.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Fix url pathname for prerenders
--- a/packages/kit/src/core/adapt/prerender.js
+++ b/packages/kit/src/core/adapt/prerender.js
@@ -180,7 +180,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 
 		const rendered = await app.render(
 			{
-				url: `${config.kit.protocol || 'sveltekit'}://${config.kit.host || 'prerender'}${path}`,
+				url: `${config.kit.protocol || 'http'}://${config.kit.host || 'prerender'}${path}`,
 				method: 'GET',
 				headers: {},
 				rawBody: null
@@ -329,7 +329,7 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 	if (fallback) {
 		const rendered = await app.render(
 			{
-				url: `${config.kit.host || 'sveltekit'}://${config.kit.host || 'prerender'}/[fallback]`,
+				url: `${config.kit.protocol || 'http'}://${config.kit.host || 'prerender'}/[fallback]`,
 				method: 'GET',
 				headers: {},
 				rawBody: null


Rich-Harris:
I think this is indicative of a deeper bug that resulted from #3126 — before, we'd pass a `path` (which is the same whether you're prerendering or not) but now we need a full URL for hydration. (The obvious thing to do would be to just use `new URL(location)`, but that has edge cases of its own when you're using Kit to prerender snippets of HTML that are then embedded on a _different_ path.)

This seems like a workable fix for now but we should probably think of something more comprehensive

Rich-Harris (approved):

PR #3182, opened 2022-01-02
Add `disableScrollHandling` function

Rich-Harris:
Fixes https://github.com/sveltejs/kit/issues/2794 by adding a new `disableScrollHandling` function. Calling it in `onMount` or inside an action will, if a navigation is currently happening, prevent Kit from applying its normal scroll handling. (Calling it when navigation is not currently happening will cause an error to be thrown.)

See https://github.com/sveltejs/kit/pull/2986#issuecomment-1002302203 for previous discussion.

I prefer the imperative `disableScrollHandling` to a declarative `export const autoscroll = false` for various reasons:

1. This _should_ feel a bit awkward. Disabling default scroll behaviour will, in general, disorient your users. It should be a last resort. Adding an `autoscroll` page option feels like we're _blessing_ that behaviour rather than simply _tolerating_ it.
2. The implementation is more straightforward
3. If provides more flexibility — it's easier to use it inside an action (etc). Any page that could use `autoscroll` could also use `disableScrollHandling`, but the reverse isn't true

### Tests

### Changesets

### DIFF
--- a/.changeset/beige-months-obey.md
+++ b/.changeset/beige-months-obey.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Add disableScrollHandling function
+Breaking: Add disableScrollHandling function (see https://kit.svelte.dev/docs#modules-$app-navigation)
--- a/.changeset/beige-months-obey.md
+++ b/.changeset/beige-months-obey.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Add disableScrollHandling function
--- a/packages/kit/src/runtime/client/singletons.js
+++ b/packages/kit/src/runtime/client/singletons.js
@@ -4,10 +4,12 @@ export let router;
 /** @type {import('./renderer').Renderer} */
 export let renderer;
 
-/** @param {{
+/**
+ * @param {{
  *   router: import('./router').Router?;
  *   renderer: import('./renderer').Renderer;
- * }} opts */
+ * }} opts
+ */
 export function init(opts) {
 	router = opts.router;
 	renderer = opts.renderer;
--- a/packages/kit/src/runtime/client/singletons.js
+++ b/packages/kit/src/runtime/client/singletons.js
@@ -1,7 +1,14 @@
 /** @type {import('./router').Router?} */
 export let router;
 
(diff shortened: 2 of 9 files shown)


Rich-Harris:
cc @bluwy since you've been looking a lot at scroll stuff lately

packages/kit/src/runtime/client/singletons.js
```diff
@@ -1,7 +1,14 @@
...
-/** @param {import('./router').Router?} _ */
-export function init(_) {
-	router = _;
+/** @type {import('./renderer').Renderer} */
+export let renderer;
+
+/** @param {{
+ *   router: import('./router').Router?;
+ *   renderer: import('./renderer').Renderer;
+ * }} opts */
```
benmccann, line 10: VS Code's syntax highlighting won't work if you put the `@param` on the same line as the `/**` unless it's just a single-line comment
```suggestion
/**
 * @param {{
 *   router: import('./router').Router?;
 *   renderer: import('./renderer').Renderer;
 * }} opts
 */
```

benmccann:
just one minor formatting comment. otherwise lgtm

benmccann (approved):

f-elix:
Just wanna thank you so SO much for this PR! I haven't updated my Sveltekit projects since version 192 because of the navigation scroll issues. Much appreciated!

.changeset/beige-months-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Add disableScrollHandling function
```
bluwy, line 5: Since this is a breaking change, perhaps we can add some information here on how to migrate over, and a big `BREAKING CHANGE` text. I noticed on discord many still have questions regarding the recent breaking changes, even though we have provided some helpful error messages.

bluwy (approved):
LGTM. Thanks for the explanation of the imperative approach too. I agree this would work well for most cases needed for now.

.changeset/beige-months-obey.md
Rich-Harris, line 5:
```suggestion
Breaking: Add disableScrollHandling function (see https://kit.svelte.dev/docs#modules-$app-navigation)
```

PR #3183, opened 2022-01-02
Loadfunction custom headers

### LINKED ISSUE #909: Allow to pass custom headers on load responses

nhe23:
Fixes #909. The load function can now return headers, which are passed to the response headers.

### Tests

### Changesets

### DIFF
--- a/.changeset/mean-cheetahs-train.md
+++ b/.changeset/mean-cheetahs-train.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Optionally return headers from load function
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -8,6 +8,10 @@ A component that defines a page or a layout can export a `load` function that ru
 // Declaration types for Loading
 // * declarations that are not exported are for internal use
 
+// type of string[] is only for set-cookie
+// everything else must be a type of string
+type ResponseHeaders = Record<string, string | string[]>;
+
 export interface LoadInput<
 	PageParams extends Record<string, string> = Record<string, string>,
 	Stuff extends Record<string, any> = Record<string, any>,
@@ -30,6 +34,7 @@ export interface LoadOutput<
 	props?: Props;
 	stuff?: Stuff;
 	maxage?: number;
+	headers?: ResponseHeaders;
 }
 ```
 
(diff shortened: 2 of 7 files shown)


nhe23:
On my machine all tests run without errors. I don't think my changes caused that test to fail...

Rich-Harris:
Thank you — I've left comments on the issue: https://github.com/sveltejs/kit/issues/909#issuecomment-1004128975

nhe23:
I'll close this PR because setting custom headers in the load function is not the best way to solve #909.

PR #3197, opened 2022-01-04
[chore] override page.goto to run started logic

benmccann:
Overrides `page.goto` to run the `logic` the previously resided in `started`. That method could not be called from `page.goto` as it would create a cycle between `page` and `started`, so I inlined the logic into `page.goto` as it was no longer used anywhere else

### DIFF
--- a/packages/kit/test/utils.js
+++ b/packages/kit/test/utils.js
@@ -59,11 +59,9 @@ export const test = base.extend({
 	},
 
 	// @ts-expect-error
-	clicknav: async ({ page, javaScriptEnabled, started }, use) => {
+	clicknav: async ({ page, javaScriptEnabled }, use) => {
 		/** @param {string} selector */
 		async function clicknav(selector) {
-			await started();
-
 			if (javaScriptEnabled) {
 				await page.evaluate(() => {
 					window.navigated = new Promise((fulfil, reject) => {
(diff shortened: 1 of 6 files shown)


packages/kit/test/utils.js
```diff
@@ -143,11 +126,34 @@ export const test = base.extend({
...
+			 * @param {object}	opts
+			 */
+			async function (url, opts) {
+				const res = await goto.call(page, url, opts);
+				if (javaScriptEnabled) {
+					await page.waitForFunction(() => window.started);
+				}
+				return res;
+			};
+		await use(page);
```
bluwy, line 142: I'm not sure how safe this is to override `goto`, but I think this should be something like:

```js
const promises = []
if (javascriptEnabled) {
  promises.push(page.waitForFunction(() => window.started))
}
promises.push(goto.call(page, url, opts))
return Promise.all(promises)[promises.length - 1]
```

It could be refactored to look nicer, but in my experience, the `waitFor` must always be called before the goto. If we apply this pattern for the `started()` code we use to have, it should fix the flakes too.

This `Promise.all` pattern is describe [here](https://playwright.dev/docs/navigations#asynchronous-navigation).

Rich-Harris: Advisability of monkey-patching notwithstanding, I think it's fine? it shouldn't be racy (it's not like `waitForNavigation` where you need to start waiting before triggering the navigation), and you need to go to the page before checking for `window.started`, otherwise in a situation like this...

```js
await page.goto('/foo');
expect(x).toBe(y);

await page.goto('/bar'); // `window.started` is still true from before
expect(x).toBe(y);
```

...the promise could resolve prematurely, I think?

Rich-Harris (approved):
bravo dude, this is superb stuff — would never have realised you can monkey-patch `page` like that

packages/kit/test/utils.js
bluwy: Yeah, now that I think of it, since `started` is a promise we could wait on that later as well, so the race condition I'm imagining wouldn't be an issue. And the example is true too. LGTM then 👍

bluwy (approved):

PR #3218, opened 2022-01-05
[chore] explicit `tick()`

benmccann:
I think `await 0` would return to the event loop and I'm not entirely sure it's going to happen in order (I think it should, but there seem to be [exceptions](https://stackoverflow.com/questions/2734025/is-javascript-guaranteed-to-be-single-threaded/2734311#2734311)). If we call `tick()` then hopefully we explicitly know the DOM will be rendered. At the very least, I think this makes the code clearer

Hmmm. This doesn't seem to have helped the flake. Maybe it's still a good idea though?


Rich-Harris (approved):

Rich-Harris:
at the very least, it makes the purpose of the `await` more explicit

PR #3219, opened 2022-01-06
Improve error message when a Svelte config is not found

ykomatsu:
`svelte-kit` error message can be a bit confusing when both `svelte.config.js` and `svelte.config.cjs` are absent.
We need `svelte.config.js` , not `svelte.config.cjs` :

```
$ npm run dev
Cannot find module '/Users/ykoma/Repositories/kit-playground/svelte.config.cjs' imported from /Users/ykoma/Repositories/kit-playground/node_modules/@sveltejs/kit/dist/cli.js
...
```

After this PR:

```
$ npm run dev
Cannot find module '/Users/ykoma/Repositories/kit-playground/svelte.config.js' imported from /Users/ykoma/Repositories/kit-playground/node_modules/@sveltejs/kit/dist/cli.js
...
```

### Tests

### Changesets


dummdidumm (commented):
Please provide a changeset, other than that this looks good 👍

.changeset/neat-points-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Improve an error message for MODULE_NOT_FOUND
```
bluwy, line 5:
```suggestion
Improve an error message for MODULE_NOT_FOUND
```

Improve error message when a Svelte config is not found

bluwy (approved):

Rich-Harris (changes requested):
Thank you! This still isn't a great error message though — it should be a much more explicit 'Missing svelte.config.js', or perhaps

> You need to create a svelte.config.js file — see https://kit.svelte.dev/docs#configuration

In fact, there's already some logic for that, it's just a) in a [weird place](https://github.com/sveltejs/kit/blob/b4aef9635c579087466d85819cea3294af948390/packages/kit/src/cli.js#L32-L39) that means it never actually runs, and b) overly complicated. We should just check to see if `svelte.config.js` or `svelte.config.cjs` exists, and if not throw a helpful error inside `load_config`

Rich-Harris (approved):

Rich-Harris:
Thanks. Realised we can actually simplify this stuff quite a bit, because the checks for `svelte.config.cjs` and `vite.config.js` are from many months ago, and really aren't needed any more

Issue #3194, opened 2022-01-03
Adapters should add cache-control headers to static assets


Rich-Harris:
### Describe the problem

As noted in https://github.com/sveltejs/kit/issues/909#issuecomment-1004128975, most adapters don't configure cache-control headers for the three categories of static assets — hashed (Vite output), `static` directory contents, and prerendered pages.

### Describe the proposed solution

For hashed assets there's no real reason they shouldn't have `cache-control: public, immutable, max-age=31536000`. 

For the contents of the `static` directory, it really ought to be configurable. This could be done on a per-adapter basis (e.g. there's a PR for `adapter-cloudflare-workers` [here](https://github.com/sveltejs/kit/pull/1771)), but since it's a universal requirement I think it belongs in the main config. Straw man:

```js
// svelte.config.js
export default {
  kit: {
    static: {
      // cache everything in `static` for 1 hour
      cache: 3600

      // cache large images for one hour, small images for ten minutes, don't cache anything else
      cache: ({ filename, size }) => {
        if (filename.startsWith('images') {
          return size > 1_000_000 ? 3600 : 600;
        }

        return 0;
      }
    }
  }
};
```

Prerendered pages should adhere to whatever the `cache-control` header was at the time of prerendering. For some platforms it may be difficult or impossible to set cache headers for static assets, but here we can at least cache stuff in the browser with a `<meta http-equiv>` tag.

### Importance

would make my life easier

bjon:
This would be great.

bjon:
Shouldn't this return a cache-control string, instead of a max-age integer?
That way we can custom every thing. _private, no-store_, etc.

Rich-Harris:
Static assets are public by definition; there's no value in being able to mark them non-public. Perhaps it makes sense to expose `revalidate`, but what's a situation in which you would need `no-store` (rather than just setting `max-age=0`)?

Flexibility isn't _always_ a good thing — cache-control headers can be a footgun. I'd argue it's a good thing, for example, not to expose the ability to inadvertently mark prerendered pages or `static` contents as `immutable`. For the sake of both avoiding the possibility of those bugs, and exposing a more intuitive API (nothing is simpler than a single `max-age` value), it's very possible that the more restrictive API is preferable. Are there real world use cases that would be prevented by that?

khromov:
There is an edge case here if you set `export const ssr = false;` where a user can get a broken page view when returning to a tab that has been unloaded from memory. 

Repro steps:

[...]

PR #3222, opened 2022-01-06
make static assets immutable

Rich-Harris:
first part of #3194

* [x] adapter-cloudflare
* [x] adapter-cloudflare-workers
* [x] adapter-netlify
* [x] adapter-node
* [x] adapter-vercel

### Tests

### Changesets

### DIFF
--- a/.changeset/brave-weeks-allow.md
+++ b/.changeset/brave-weeks-allow.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Expose appDir to adapters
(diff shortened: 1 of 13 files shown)


Rich-Harris:
cases this doesn't (yet) address:

* `paths.assets` is set, meaning assets are hosted externally. in these cases, `manifest.assets` should be empty, and adapters probably shouldn't attempt to deploy or serve static files or write custom headers for them
* `paths.base` is set. adapters are probably doing several wrong things here

i think it's probably better to figure those out separately

PR #3223, opened 2022-01-06
type adapter-netlify

Rich-Harris:
This just gets rid of some of the red squigglies in adapter-netlify/src/handler.js. No changeset/test/etc necessary

### Tests

### Changesets

### DIFF
--- a/packages/adapter-netlify/src/handler.js
+++ b/packages/adapter-netlify/src/handler.js
@@ -3,13 +3,19 @@ import { App } from './server/app.js';
 
 __fetch_polyfill();
 
+/**
+ *
+ * @param {import('@sveltejs/kit').SSRManifest} manifest
+ * @returns {import('@netlify/functions').Handler}
+ */
 export function init(manifest) {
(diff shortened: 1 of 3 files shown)


packages/adapter-netlify/src/handler.js
```diff
@@ -3,13 +3,19 @@ import { App } from './server/app.js';
...
+ */
 export function init(manifest) {
+	/** @type {import('@sveltejs/kit').App} */
 	const app = new App(manifest);
 
 	return async (event) => {
 		const { httpMethod, headers, rawUrl, body, isBase64Encoded } = event;
 
-		const encoding = isBase64Encoded ? 'base64' : headers['content-encoding'] || 'utf-8';
+		const encoding = isBase64Encoded ? 'base64' : 'utf-8';
```
Rich-Harris, line 18: not sure what the `headers['content-encoding']` stuff was doing here — if the charset were declared it would be in the `content-type`, but i'm pretty confident that it's always 'utf-8' if it's not base 64 encoded

bluwy (approved):

Issue #3232, opened 2022-01-07
Static adapter does not honor base path for static files


Ghirigoro:
### Describe the bug

1. Create a project with a static adapter.
2. Add this to the config: `paths: { base: '/apath'}`
3. Build the project
4. Preview the project
5. Note that any files in `static` folder are resolved at the root and ignore the base path. i.e. `localhost:3000/img.png` instead of the expected  `localhost:3000/apath/img.png`

I would expect the base path to be honored across the board and not just for js/css files.

### Reproduction

NOTE: I'm not sure if this is a bug or static files intentionally must always resolve to the root. If this is expected behavior I'll change this to a feature request, if it is indeed a bug then I will provide a repository highlighting the issue.

### Logs

```shell
NA
```

### System Info

```shell
System:
    OS: macOS 12.0.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 664.50 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.2.0 - /usr/local/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 8.1.4 - /usr/local/bin/npm
  Browsers:
    Brave Browser: 93.1.29.80
    Chrome: 97.0.4692.71
    Firefox: 95.0.2
    Firefox Developer Edition: 82.0
    Safari: 15.1
  npmPackages:
    @sveltejs/adapter-static: ^1.0.0-next.23 => 1.0.0-next.24 
    @sveltejs/kit: next => 1.0.0-next.216 
    svelte: ^3.42.6 => 3.45.0
```

### Severity

blocking all usage of SvelteKit

Ghirigoro:
Someone kindly pointed out on the discord chat that I can work around this issue by manually prefixing my file paths for things in `static` with the base path. For example:
```
<script lang="ts">
  import { base } from '$app/paths';
  //...
</script>
<img src={`${base}/my-static-file.svg`}/>
```
This is a bit tedious but fine if that's the intended approach, but it breaks down in places where you don't have access to `$app/paths` - e.g. referencing font files in css, or favicons in html.

benmccann:
@Ghirigoro can you set [the `assets` option](https://kit.svelte.dev/docs#modules-$app-paths) to be your domain + base directory as an easier workaround?

benmccann:
[...]

Issue #3231, opened 2022-01-07
Favicon does not use kit `paths`


DblK:
### Describe the bug

When setting the following configuration for svelete:

```
...
kit: {
  ...
  paths: {
    base: '/admin'
  },
}
```

In the `app.html` there is the following line `<link rel="icon" href="/favicon.png" />`.
The `%svelte.head%` add the path while generating files but favicon is left untouched.

### Reproduction

Simply create a new app and change the kit config as previously stated.

### System Info

```shell
System:
    OS: macOS 10.15.7
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
    Memory: 18.91 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.17.0 - ~/.nvm/versions/node/v14.17.0/bin/node
    npm: 7.24.0 - ~/.nvm/versions/node/v14.17.0/bin/npm
  Browsers:
    Chrome: 96.0.4664.110
    Firefox: 89.0.2
    Safari: 14.0.3
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.9 
    @sveltejs/adapter-static: ^1.0.0-next.24 => 1.0.0-next.24 
    @sveltejs/kit: next => 1.0.0-next.216 
    svelte: ^3.44.0 => 3.44.3
```

### Severity

annoyance

### Additional Information

To solve the issue I have to manually change the path for the favicon.
Since I do not know yet where I will serve my app, I will need to change the svelte config BUT also the `app.html` to avoid 404 error for favicon.

ghostdevv:
Very interesting to see where this issue will go, as app.html isn't processed and talks of potentially doing that occurred in #3127

I wonder what happens when you import a asset in one of your routes such as `<img src="/favicon.png" />` if that image is also broken or if that url is processed

Conduitry:
I would suggest just putting the `<link rel=icon>` somewhere else where you do have access to the `base` information - like `__layout.svelte`, and then using https://kit.svelte.dev/docs#modules-$app-paths to construct the appropriate URL. I don't think it makes sense to make it any more complicated than that or to process `app.html` in some way.

I could go either way on whether the `create-svelte` template apps should be updated like this.

Rich-Harris:
This will be fixed by #3234 — just add `%svelte.assets%` to the template (i.e. `%svelte.assets%/favicon.png`)

PR #3234, opened 2022-01-07
replace %svelte.assets% with relative path

### LINKED ISSUE #2697: Generated URLs are inconsistent to the basepath with `adapter-static`

Rich-Harris:
Fixes #2697, supersedes #3088. Replaces `%svelte.assets%` in `src/app.html` with a relative path to the root (or to `paths.assets` in production, if specified).

In other words this...

```html
<link rel="icon" type="image/png" href="%svelte.assets%/favicon.png" />
```

...will work regardless of whether `paths.assets` is set, and will also work if you've set `paths.base` (or, in future, if we implement dynamic basepaths somehow) because a page like `/custom-basepath/foo/bar/baz` will link to `../../favicon.png`, which resolves to `/custom-basepath/favicon.png`. It will still work if `trailingSlash` is  `"always"` .

(Note that there's currently a bug around `paths.base` — static assets aren't served from there in dev mode. Needs fixing separately.)

### Tests

### Changesets


Conduitry:
Also adjacent: #3231 and #3232

Rich-Harris:
The deployments are failing because they're trying to build with the most recent released version of Kit, which doesn't know how to handle `%svelte.assets%` and is therefore resulting in a malformed URI error when prerendering. We can safely ignore

Conduitry:
Should this PR be marked as resolving #3231 now?

I didn't look closely at what was happening in #3232.

Rich-Harris:
It should, yeah. I need to look into #3232

dreitzner:
@Rich-Harris should I port the relevant docs over from #3088 ?

PR #3235, opened 2022-01-07
[feat] allow to render error pages from handleError

### LINKED ISSUE #2824: Have a way to render an error page when there's an exception in a hook (e.g. handle)

nhe23:
Fixes #2824 by optionally rendering error pages if `renderError` is returned from the `handleError` hook. This solution is better than #3171 because the right status code is returned and the user can reload the page in order to try again.

### Tests

### Changesets


Rich-Harris:
Thank you. Adding new API is unnecessary though — all we need to do is this...

```js
// pseudo-code
try {
  return handle();
} catch (e) {
  handle_error(e);
  return render_error_page();
}
```

...and wrap all the error page rendering logic in a try-catch (previously, some of it was outside the try-catch) so that that's guaranteed not to throw. I've opened #3239, so I'll close this and #3171

PR #3240, opened 2022-01-07
allow prefixes and suffixes around rest parameters

### LINKED ISSUE #1714: Allow rest parameters as end of endpoint route

Rich-Harris:
Fixes #1714. Not sure why rest segments were ever required to be standalone — probably just me being too lazy to figure out the implementation and not being able to see a valid use case. Since there clearly _is_ a valid use case, and nothing seems to break if we remove that restriction, I finally got around to figuring out the implementation. (Turns out it's not that hard.)

Prefixes probably aren't all that useful, but suffixes (i.e. `/tree/[...rest].json.js`) certainly are.

### Tests

### Changesets

### DIFF
--- a/.changeset/silly-ants-battle.md
+++ b/.changeset/silly-ants-battle.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Allow prefixes and suffixes around rest parameters
--- a/packages/kit/src/core/create_manifest_data/index.js
+++ b/packages/kit/src/core/create_manifest_data/index.js
@@ -103,10 +103,6 @@ export default function create_manifest_data({ config, output, cwd = process.cwd
 				throw new Error(`Invalid route ${file} — brackets are unbalanced`);
 			}
 
-			if (/.+\[\.\.\.[^\]]+\]/.test(segment) || /\[\.\.\.[^\]]+\].+/.test(segment)) {
-				throw new Error(`Invalid route ${file} — rest parameter must be a standalone segment`);
-			}
-
 			const parts = get_parts(segment, file);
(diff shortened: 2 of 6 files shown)


benmccann:
I didn't look at the code, but it seems like a fairly safe and desirable change and has tests, so I say ship it!

PR #3241, opened 2022-01-08
Wait until navigation ends before updating history

### LINKED ISSUE #2870: Router changes the URL prematurely

Rich-Harris:
Still not sure if this is desirable or not, but I'm opening this in order to close #2917. Fixes #2870.

It would be fairly straightforward to make this configurable, but I think something this fundamental should probably be decided by the framework.

### Tests

### Changesets


packages/kit/src/runtime/client/router.js
```diff
@@ -165,17 +165,15 @@ export class Router {
...
-			this._navigate(url, noscroll ? scroll_state() : null, false, [], url.hash);
+			this._navigate(url, noscroll ? scroll_state() : null, false, [], url.hash, {}, 'pushState');
 			event.preventDefault();
 		});
 
 		addEventListener('popstate', (event) => {
 			if (event.state && this.enabled) {
 				const url = new URL(location.href);
-				this._navigate(url, event.state['sveltekit:scroll'], false, []);
+				this._navigate(url, event.state['sveltekit:scroll'], false, [], url.hash, null, null);
```
benmccann, line 176: should we just allow these lass two parameters to be `undefined` so that you don't have to pass anything here?

benmccann:
I would say that I lean in favor of this change. It's taken me a couple days to comment on this as I really wanted to have time to read through and understand the various arguments and think about it.

However, I think there may be a slightly different implementation that I prefer even more, which would be to update the URL in between network fetching and DOM rendering (https://github.com/sveltejs/kit/issues/2870#issuecomment-1003725582). This is most analogous to the browser's behavior. It also satisfies quite a lot of the people who wanted the URL to update before navigation (https://github.com/sveltejs/kit/pull/2917#issuecomment-979419299) because a lot of those arguments seemed to be that the UI should update after the URL. It sort of seems like it could get us the best of both worlds.

It's harder to figure out how to structure the code if we do that. The separation between router and renderer gets even messier. But I still think it's the best way to do it and we can figure out a refactoring later.

Rich-Harris:
> However, I think there may be a slightly different implementation that I prefer even more, which would be to update the URL in between network fetching and DOM rendering

That feels like the worst of all worlds to me. It would be completely imperceptible to users (the difference is basically a microtask), but would mean that we couldn't rely on page state following navigation (for example during tests). What would be the advantage?

Rich-Harris:
I realised there's another compelling reason to update history at the start of the navigation rather than the end — unless the app implements some sort of loading bar UI, there's no visual indication that a navigation has been accepted other than the URL changing. That can make things feel sticky. When not using client-side navigation, the browser _does_ provide immediate UI (the reload button starts spinning), which makes this much less of a concern. (The future [`appHistory`](https://wicg.github.io/app-history/) API will align client-side navigation with the default behaviour, and I can imagine SvelteKit 2 or 3 using it.)

benmccann:
That's probably the best argument I've heard for doing it before. And it does look like the [AppHistory API will be helpful](https://web.dev/app-history-api/#transition) if/when it comes around.

Doing it after certainly would make testing easier as you pointed out. I'd prioritize users over ourselves though, so I guess I'm convinced on doing it before. Hopefully we can figure out the test flakiness one of these days

Rich-Harris:
The nice thing is that because of #3303, this is basically just a one-line change if and when we do want to do it. So there's probably no sense in leaving this PR open

Issue #2603, opened 2021-10-14
Move amp support to a plugin


benmccann:
### Describe the problem

I don't like amp. It's bad for the web

Also, it slows things down a lot:

https://github.com/sveltejs/kit/blob/6c9b73977f7cb54902c55ccc395510b7ac3eb35b/packages/kit/test/utils.js#L161

### Describe the proposed solution

Can we use hooks, vite plugins (https://www.npmjs.com/package/vite-plugin-amp), or do we need to add new extension points?

Also, we can potentially move some of the stuff to other options like the proposed `inlineCss` option (https://github.com/sveltejs/kit/pull/2620), so that it's not AMP-specific.

### Importance

nice to have

Rich-Harris:
I would love to get rid of built-in AMP support. AMP is flaming hot garbage.

SvelteKit's AMP support consists of the following elements:

1. Styles are inlined, regardless of size
2. HTML is run against the AMP validator during dev
3. `import { amp } from '$app/env'`
4. The [AMP boilerplate](https://amp.dev/documentation/guides-and-tutorials/learn/spec/amp-boilerplate/) is added
5. Instead of including all _imported_ styles, it only includes _rendered_ styles, since there's no possibility of components being dynamically added later

Most of this could be done in userland today, especially if we do #3511:

1. Set `inlineStyleThreshold` to `Infinity`
2. Do validation in `handle`
3. `VITE_AMP=true npm run dev` and `import.meta.env.VITE_AMP`
4. Boilerplate can be injected right before validation occurs

The last one is the only part that would need changes to Kit itself, I think. Proposal: if a page is rendered with `router: false` and `hydrate: false` — i.e. Kit doesn't include any JS — then we could use rendered CSS (instead of imported CSS) in all cases, rather than making it an AMP-specific feature.

We could then move the validation/boilerplate stuff to a separate package (`@sveltejs/amp-kit`?) and document the required config changes.

Rich-Harris:
Another wrinkle we'll need to accommodate: https://github.com/sveltejs/kit/pull/4073

PR #3249, opened 2022-01-08
[chore] only import AMP validator if using AMP

benmccann:
ref https://github.com/sveltejs/kit/issues/2603

this should make server startup and our tests faster


dreitzner (approved):
lgtm

Rich-Harris (approved):

PR #3276, opened 2022-01-10
try to get deploys working with pnpm

Rich-Harris:
At the moment, deploys of the demo app (which are useful as an end-to-end test) sometimes fail noisily because the app depends on the `next` version of Kit etc, rather than the current state of the repo. In theory if we could use pnpm instead on Netlify, Cloudflare and Vercel (the three platforms for which we currently do automated deploys) that would no longer be the case?

It might even fix the persistent build cache issues we experience with Netlify.

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/templates/default/netlify.toml
+++ b/packages/create-svelte/templates/default/netlify.toml
@@ -1,6 +1,3 @@
 [build]
   command = "npm run build"
   publish = "build"
-
-[build.environment]
-  NPM_FLAGS="--prefix=/dev/null"
--- a/packages/create-svelte/templates/default/netlify.toml
+++ b/packages/create-svelte/templates/default/netlify.toml
@@ -1,5 +1,5 @@
 [build]
-  command = "npx pnpm install --store=node_modules/.pnpm-store && npx pnpm build"
+  command = "npx pnpm install --store=node_modules/.pnpm-store && npx pnpm build -r"
   publish = "build"
 
 [build.environment]
--- a/packages/create-svelte/templates/default/netlify.toml
+++ b/packages/create-svelte/templates/default/netlify.toml
@@ -1,3 +1,6 @@
 [build]
   command = "npm run build"
   publish = "build"
+
+[build.environment]
+  NPM_FLAGS="--prefix=/dev/null"
--- a/packages/create-svelte/templates/default/netlify.toml
+++ b/packages/create-svelte/templates/default/netlify.toml
@@ -1,3 +1,6 @@
 [build]
-  command = "npm run build"
+  command = "npx pnpm install --store=node_modules/.pnpm-store && npx pnpm build"
   publish = "build"
+
+[build.environment]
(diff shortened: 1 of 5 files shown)


Rich-Harris:
Well, the trick suggested in https://www.seancdavis.com/posts/use-pnpm-with-netlify/ doesn't work:

> 1:32:28 PM: npm ERR! EEXIST: file already exists, mkdir '/dev/null'

Rich-Harris:
Cloudflare Pages and Netlify are both choking, because they install dependencies with `npm` immediately after the repo is cloned (this fails because of the `workspace:*` dependencies). Vercel, by contrast, exposes a configurable `install` command.

Anyone know if there's a way to disable `npm install` on Pages and Netlify?

Rich-Harris:
Alright, working on Netlify as well now. Just need to figure out a way to disable `npm install` on Cloudflare Pages

packages/create-svelte/templates/default/netlify.toml
```diff
@@ -1,3 +1,6 @@
 [build]
-  command = "npm run build"
+  command = "npx pnpm install --store=node_modules/.pnpm-store && npx pnpm build -r"
   publish = "build"
+
+[build.environment]
+  NPM_FLAGS = "--version"
```
Conduitry, line 6: It'd be nice to have a little comment here explaining that this is a workaround to make the automatic running of npm a no-op.

bluwy:
The trick for Cloudflare pages is the same as Netlify, you can set the env var `NPM_FLAGS` to `--version` and the same build script. It works when deploying my website.

Rich-Harris:
I got a bit further with the `NPM_FLAGS` hack, but now running into this:

> ERR_PNPM_NO_MATCHING_VERSION_INSIDE_WORKSPACE  In : No matching version found for @sveltejs/adapter-auto@* inside the workspace

Perhaps Cloudflare Pages discards everything outside the root directory?

bluwy:
Strange, it seems like so from the error, perhaps we need the root as `/` for now if that's possible

Rich-Harris:
that did the trick! changing the root from `/packages/create-svelte/templates/default` to `/`, and the build output directory from `/.svelte-kit/cloudflare` to `/packages/create-svelte/templates/default/.svelte-kit/cloudflare` fixed it

Conduitry:
Is there anything in this repo that could be done to represent that change in some way, rather than having the config just live hidden in Cloudflare?

Rich-Harris:
I don't think so, unfortunately — the relevant settings are only exposed via the UI AFAICT. Unless we add it to a README somewhere or something

Conduitry:
Having it live somewhere in the repo - even if it's configuration that needs to be manually applied by a person interacting with a UI - would still be worthwhile I think.

Rich-Harris:
Yeah, I added some notes in 7b2db6b7e72fe1c5ec3bfce8fbbfc5ee2c16a746 — probably best we can do

PR #3293, opened 2022-01-11
[feat] `beforeNavigate` and `afterNavigate` lifecycle functions

### LINKED ISSUE #552: Before and after hooks for routing
**Describe the bug**
`onMount` is not fired on page change, if the actual page component doesn't change. This is all routes which contain [dynamic] parameters, since the underlying components are the same.

**To Reproduce**
Reproduction here:

https://github.com/beyonk-adventures/sapper-onmount-issue

The colour pages set the value from the path parameters on load, and onMount sets the value in the page. Because onMount isn't refired, the colour never changes when clicking between pages, only on hard refresh, or when visiting the home page and then clicking a new colour.

Essentially it means you can't change the content of the page unless you visit an url which results in a completely different page component being loaded.

**Expected behavior**
Though using the same physical page component, `/different/routes` and `/different/paths` which map to `/different/[parameter].svelte` should fire onMount, since the route is changing.

**Information about your Sapper Installation:**
- Sapper └── sapper@0.27.16 

**Severity**
Extremely High. I don't actually know how to work around this.

**Additional context**
Fixing this would be a breaking change, but I think it's far more logical an approach than the current behaviour.

### LINKED ISSUE #560: onNavigate lifecycle function

### LINKED ISSUE #2974: Ability to intercept a navigation request

Rich-Harris:
Todo (from discussion below):

* [x] rename to `beforeNavigate` and `afterNavigate`
* [x] make `beforeNavigate` callback synchronous
* [x] pass full navigation object (`{from, to}`) to `beforeNavigate` (and `afterNavigate`?)
* [x] add a `cancel` function to `beforeNavigate` instead of returning `false`
* [x] `beforeNavigate` should work with external navigation as well as internal navigation
* [x] `beforeNavigate` should call [`onbeforeunload`](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload) on the user's behalf

---

For whatever reason, I can't sync my local version of #2982 with that branch, so I've duplicated it for further review. All credit belongs to @zommerberg!

Fixes #2974, fixes #560, and fixes #552.

### Tests

### Changesets


Rich-Harris:
I think this API might be more future-proof:

```js
onBeforeNavigate(async ({ from, to }) => {
  console.log(`about to navigate from ${from.href} to ${to.href}`);

  // cancel or allow the navigation in the return statement
  return false;
});
```

`{ from, to }` is the contents of the `navigating` store (https://kit.svelte.dev/docs#modules-$app-stores), which could conceivably include useful things like `type: 'back'` or whatever in future

Rich-Harris:
More thoughts, which I plan to spend some time on tomorrow:

* It would probably make sense if the `onNavigate` callback also took the `{ from, to }` argument
* For consistency with Svelte itself, which has `beforeUpdate` and `afterUpdate`, it would probably be better to call these functions `beforeNavigate` and `afterNavigate`.
* What should happen if an error is thrown in the `beforeNavigate` callback? Should it just blow up, like errors in other lifecycle function callbacks, or should it cause an error page to be rendered?
* Right now, callbacks only run for internal navigation. It feels like it should probably run for external navigation too — if someone is using `beforeNavigate` like they'd use `onbeforeonload` (i.e. to prompt the user to save unsaved data), then it should happen if someone clicks a link to https://external.com just as much as a link to `/blah`

Separately, I had been wondering about adding an `onUpdate` function to Svelte which runs at the same time as `beforeUpdate`. If the callback returns a function, it is treated as an `afterUpdate` callback. So [this example](https://svelte.dev/tutorial/update) from the tutorial would change from this...

```js
let div;
let autoscroll;

beforeUpdate(() => {
  autoscroll = div && (div.offsetHeight + div.scrollTop) > (div.scrollHeight - 20);
});

afterUpdate(() => {
  if (autoscroll) div.scrollTo(0, div.scrollHeight);
});
```

...to this:

```js
let div;

onUpdate(() => {
  const autoscroll = div && (div.offsetHeight + div.scrollTop) > (div.scrollHeight - 20);
  if (autoscroll) {
    return () => div.scrollTo(0, div.scrollHeight);
  }
});
```

This is obviously a bit speculative but _if_ we were to do that, and _if_ we were to use `beforeNavigate` and `afterNavigate` for symmetry, then we might find ourselves wanting an `onNavigate` with similar semantics:

```js
onNavigate(({ to }) => {
  const time = Date.now();
  return () => console.log(`navigated to ${to.href} in ${Date.now() - time}ms`);
});
```

In that world, we might find that 'return `false` to cancel navigation' is unduly restrictive. I therefore wonder if we should do this instead? It's more explicit and prevents gotchas. 

```js
beforeNavigate(({ to, cancel }) => {
  if (to.pathname === '/') {
    console.log("you can't go home again");
    cancel();
  }
});
```

benmccann:
> For consistency with Svelte itself, which has beforeUpdate and afterUpdate, it would probably be better to call these functions beforeNavigate and afterNavigate.

Svelte itself is semi-inconsistent. It has `afterUpdate` vs `onMount` and `onDestroy`. The tricky thing here is that if you ever wanted to create a `beforeDestroy` then we'd be in an awkward spot.

dummdidumm:
How would you redirect in a before/after navigate hook? Would it make sense to pass `go` for convenience? Would calling `go` imply calling `cancel`?

bluwy:
`beforeNavigate` and `afterNavigate` makes sense, returning a function for `onNavigate` might conflict in semantics with `onMount`, since the returned function executes in the destroy phase rather than an after mount phase.

Re `onbeforeunload`, I think it makes sense to handle that too if the API surface is sharable. We may have to have `to` as undefined though.

I think we also need a way for redirects as dummdidumm mentioned, perhaps recommending a `cancel()` and a `goto()` combo? Or another way is by passing `next()` so `next(false)` is cancel; `next('/url')` is a redirect. Similar to [vue router](https://router.vuejs.org/guide/advanced/navigation-guards.html#global-before-guards).

dominikg:
+1 to redirects and async 

Esp, async adds a lot of complexity though. You have to keep track of the navigation state and also make sure the url/history are updated at the right times.
E.g. what happens if a user presses browser-back while an aync navigation is still under way or changes the url in the address bar?

Rich-Harris:
Another thought: the `beforeNavigate` callback should probably be synchronous, for two reasons:

1. delaying navigation for any reason is a recipe for frustrated users. it's probably better to not have that footgun
2. feels racy. you could imagine work happening in `beforeNavigate` _after_ a subsequent navigation has been triggered. Could result in all sorts of interesting bugs

We could revisit that in future if someone presents a compelling use case but for now I think we nix the `await`

Rich-Harris:
Ugh, GitHub's buggy UI strikes again — didn't see the responses prior to my last.

We don't need a special way to redirect within the callback — just use `goto`:

```js
beforeNavigate(({ to, cancel }) => {
  cancel();
  goto(to.href.replace('here', 'there');
});
```

A third reason to disallow async callbacks — it wouldn't be possible to use them with `onbeforeunload`.

dummdidumm:
> Another thought: the `beforeNavigate` callback should probably be synchronous, for two reasons:
> 
>     1. delaying navigation for any reason is a recipe for frustrated users. it's probably better to not have that footgun

I don't think so, you can already make users frustrated by having a `load` function which takes a long time. And if you have something that determines whether or not redirection is ok, and it's asynchronous, you would need to do that anyway - and the other place for that is `load` right now.
 
>     2. feels racy. you could imagine work happening in `beforeNavigate` _after_ a subsequent navigation has been triggered. Could result in all sorts of interesting bugs

It's definitely not easy to implement.

> We could revisit that in future if someone presents a compelling use case but for now I think we nix the `await`

"Is the user allowed to enter this page?" Right now this would only be possible to check asynchronously through the `load` function.

Which in general begs the question, what are the similarities and differences for these navigation hooks and something running in a `load` function of a layout file? Maybe answering this in a concise way (also for the docs) would help answering how the best API looks like.

Rich-Harris:
> "Is the user allowed to enter this page?"

This is a question for the page to answer, not the things that link to the page. As far as aborting navigation goes, the only responsibility of a `beforeNavigate` callback is to determine whether the user is allowed to leave the _current_ page. That's likely to be driven by unsaved changes or incomplete forms — stuff that can easily be determined synchronously.

Rich-Harris:
Example of a 'you have unsaved changes' modal that works with internal navigations and `beforeunload`:

```svelte
<script>
  import { beforeNavigate, goto } from '$app/navigation';

  let intercepted = null;
  let unsaved = true;

  beforeNavigate(({ to, cancel }) => {
    if (!unsaved) return; // nothing to do
    if (intercepted) return; // we're here because of the `goto`

    cancel();

    // if `to` has a value, we're navigating internally and
    // should display our own modal. otherwise, we're in
    // beforeunload, which means the user will already
    // see a system dialog and doesn't need ours too
    if (to) {
      intercepted = { url: to };
    }
  });
</script>

{#if intercepted}
  <div class="modal-background" on:click={() => (intercepted = null)}>
    <div class="modal" on:click={(e) => e.stopPropagation()}>
      <p>you have unsaved changes!</p>
      <button on:click={() => (intercepted = null)}>stay on page</button>
      <button on:click={() => goto(intercepted.url)}>leave page</button>
    </div>
  </div>
{/if}
```

packages/kit/src/runtime/client/router.js
```diff
@@ -122,7 +151,7 @@ export class Router {
 		addEventListener('sveltekit:trigger_prefetch', trigger_prefetch);
 
 		/** @param {MouseEvent} event */
-		addEventListener('click', (event) => {
+		addEventListener('click', async (event) => {
```
bluwy, line 154: I think we can remove this `async`

packages/kit/src/runtime/client/router.js
```diff
@@ -155,8 +184,6 @@ export class Router {
 			// Ignore if <a> has a target
 			if (a instanceof SVGAElement ? a.target.baseVal : a.target) return;
 
-			if (!this.owns(url)) return;
-
```
bluwy, line 159: IIUC, this is removed so in `_navigate` we can execute `before_navigate` callbacks? I'm not sure if this is right since if we navigate to external links, `onbeforeunload` should be called too? (which also calls` before_navigate`)

Perhaps I'm missing another piece of this.

bluwy (commented):
A late review, but I have a question below 😬

packages/kit/src/runtime/client/router.js
benmccann: If on `beforeunload` is called before external navigation, then I would also tweak the docs and change "This includes navigation to external sites. `to` will be `null` if the user is closing the page." to "`to` will be `null` if the user is closing the page or navigating to an external site."

bluwy: Ah my bad, I just checked that `onbeforeunload` isn't called for external navigation, so I guess this is fine. I thought `before_navigate` would be called twice for external navigation.

PatrickG:
Sorry I'm late to the party.

> Example of a 'you have unsaved changes' modal that works with internal navigations and `beforeunload`:
> 
> ```svelte
> <script>
>   import { beforeNavigate, goto } from '$app/navigation';
> 
>   let intercepted = null;
>   let unsaved = true;
> 
>   beforeNavigate(({ to, cancel }) => {
>     if (!unsaved) return; // nothing to do
>     if (intercepted) return; // we're here because of the `goto`
> 
>     cancel();
> 
>     // if `to` has a value, we're navigating internally and
>     // should display our own modal. otherwise, we're in
>     // beforeunload, which means the user will already
>     // see a system dialog and doesn't need ours too
>     if (to) {
>       intercepted = { url: to };
>     }
>   });
> </script>
> 
> {#if intercepted}
>   <div class="modal-background" on:click={() => (intercepted = null)}>
>     <div class="modal" on:click={(e) => e.stopPropagation()}>
>       <p>you have unsaved changes!</p>
>       <button on:click={() => (intercepted = null)}>stay on page</button>
>       <button on:click={() => goto(intercepted.url)}>leave page</button>
>     </div>
>   </div>
> {/if}
> ```

The only problem with this is if the user navigated backwards with the browser controls.
The modal will be presented, the user clicks "leave page". If the user then navigates back one more entry, he ends up on the page that presented the modal instead of two pages before that.
That's where async blocking would be helpful.

packages/kit/src/runtime/client/router.js
```diff
@@ -169,20 +196,48 @@ export class Router {
...
+				details: {
+					state: {},
+					replaceState: false
+				},
+				accepted: () => event.preventDefault(),
+				blocked: () => event.preventDefault()
+			});
 		});
 
 		addEventListener('popstate', (event) => {
```
sabine, line 213: It is very helpful to know if we're navigating because of a `popstate` event, so we can close open menus and modals on browser back button use, before navigating away from a route.

Can we thread this "back button used" bit of info through `_navigate` all the way to `beforeNavigate`?

benmccann:
Looks like you got it figured out: https://discord.com/channels/457912077277855764/939868205869072444/946323798477656084

akshg05:
Hello, I followed the above but am and am able to block the navigation but onbeforeunload is not being called and I don't see the native popup.
Is there something missing which need to be done?

PR #3288, opened 2022-01-11
Crawl prerendered pages without regex hackery

Rich-Harris:
This is an alternative to #2923. It prevents the sort of bugs that can arise when you're extracting attributes with regex...

```html
<img data-something=" src=" crossorigin src="dog>cat.png"> crazy link </img>
```

...without the overhead of fully parsing the HTML. `parse5` is great, but it's [much too large](https://bundlephobia.com/package/parse5@6.0.1) to add on a whim, and does far more than we need. This implementation takes a tiny fraction of the size (it adds less than 100LOC) and is roughly 3x faster (I'm sure someone suitably motivated could make it faster still), which is important in this context. 

Fixes #2645 and #2742.

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [x] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #3301, opened 2022-01-12
Parse value of srcset attribute closer to specifications

### LINKED ISSUE #3299: Regression: adapter-static build fails if url contains comma in srcset attribute on image tag
### Describe the bug

When building with adapter-static, if the srcset attribute on an image tag contains a comma, the build fails trying to fetch not the external url pointing to the image but whatever follows the comma as an internal url.
This results in a 404 and the build failing.
Example: 
not ``https://res.cloudinary.com/demo/image/upload/q_auto:best,w_864/sample.jpg`` is fetched but ``/w_864/sample.jpg``

I am pretty sure this bug was introduced by #3288 , since it doesn't occur on 1.0.0-next.223 or lower and also seems to fit the changes.

### Reproduction

Super simple reproduction of the problem: https://github.com/Theo-Steiner/crawling-external-srcset

### Logs

```shell
https://app.warp.dev/block/kmoZl6fbECWcw8hKUKyzIp
```

### System Info

```shell
System:
    OS: macOS 11.6
    CPU: (8) arm64 Apple M1
    Memory: 97.00 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.5.0 - /opt/homebrew/bin/node
    npm: 7.19.1 - /opt/homebrew/bin/npm
    Watchman: 4.9.0 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 97.0.4692.71
    Firefox Developer Edition: 91.0
    Safari: 14.1.2
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.10
    @sveltejs/adapter-static: ^1.0.0-next.26 => 1.0.0-next.26
    @sveltejs/kit: next => 1.0.0-next.225
    svelte: ^3.44.0 => 3.46.0
```

### Severity

blocking upgrade

### Additional Information

Many image hosts use a comma in their URLs as an api for enumerating multiple transformations. Since commas are legal in URLs, we should fix this.

Theo-Steiner:

This PR attempts to fix #3299 which was a regression introduced by #3288 .

## Problem

The problem lies with the the specifications of the srcset attribute:

> If present, its value must consist of one or more image candidate strings, each separated from the next by a U+002C COMMA character (,). If an image candidate string contains no descriptors and no ASCII whitespace after the URL, the following image candidate string, if there is one, must begin with one or more ASCII whitespace.

Simply comma separating the value of the srcset attributes doesn't work in the case that there is a "," comma in one or more of the URLs of the srcset. Since popular image hosts like cloudinary use commas to enumerate transformations, this should be addressed, as prerendering fails when the crawler tries to crawl the resulting broken links.

## Proposed Solution

Rather than just reverting back to the previous RegularExpression, I attempted to write a primitive parser to implement the specification, as Rich made the original PR to get away from hard to understand regexes.
There are problems with my parser though: 

A. It fails in the case that there is a URL ending with a comma 
B. It is significantly slower than the previous regex solution (https://jsbench.me/mxkybbu22g/1)
C. There might be other problems I'm overlooking

I also modified the fixtures to test for this behavior

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/adapt/prerender/crawl.js
+++ b/packages/kit/src/core/adapt/prerender/crawl.js
@@ -171,7 +171,7 @@ export function crawl(html) {
 								}
 								candidates.push(value);
 								for (const candidate of candidates) {
-									const src = candidate.trim().split(WHITESPACE)[0];
+									const src = candidate.split(WHITESPACE)[0];
 									hrefs.push(src);
 								}
 							}
--- a/packages/kit/src/core/adapt/prerender/crawl.js
+++ b/packages/kit/src/core/adapt/prerender/crawl.js
@@ -157,15 +157,19 @@ export function crawl(html) {
 								hrefs.push(value);
 							} else if (name === 'srcset') {
 								const candidates = [];
-								// Parse the content of the srcset attribute.
-								// The regexp is modelled after the srcset specs (https://html.spec.whatwg.org/multipage/images.html#srcset-attribute)
-								// and should cover most reasonable cases.
-								const regex =
-									/\s*([^\s,]\S+[^\s,])\s*((?:\d+w)|(?:-?\d+(?:\.\d+)?(?:[eE]-?\d+)?x))?/gm;
-								let sub_matches;
-								while ((sub_matches = regex.exec(value))) {
-									candidates.push(sub_matches[1]);
+								let insideURL = true;
+								value = value.trim();
+								for (let i = 0; i < value.length; i++) {
+									if (value[i] === ',' && (!insideURL || (insideURL && value[i + 1] === ' '))) {
+										candidates.push(value.slice(0, i));
+										value = value.substring(i + 1).trim();
+										i = 0;
+										insideURL = true;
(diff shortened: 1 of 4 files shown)


packages/kit/src/core/adapt/prerender/crawl.js
```diff
@@ -156,7 +156,20 @@ export function crawl(html) {
...
+										value = value.substring(i + 1).trim();
+										i = 0;
+										insideURL = true;
+									} else if (value[i] === ' ') {
+										insideURL = false;
+									}
+								}
+								candidates.push(value);
 								for (const candidate of candidates) {
 									const src = candidate.trim().split(WHITESPACE)[0];
```
Rich-Harris, line 174: the `trim` is no longer necessary, since it's happening on line 165

```suggestion
									const src = candidate.split(WHITESPACE)[0];
```

Rich-Harris:
thank you!

PR #3302, opened 2022-01-12
Fix readonly form data return

tmaxmax:
The typings for `ReadOnlyFormData`'s `get` and `getAll` methods are wrong, as they both return `undefined` when there is no value for the given key. This results in a loss of type-safety and misguided programmers, as at least I have thought that they return the empty string/array when there are no values.
This PR changes the behavior of these methods so they are identical to the spec's [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) ones: `get` returns `null` if no value is present, `getAll` returns an empty array. This is the way I expected, and I think everyone coming to Svelte, it should work, especially given that for endpoints we pass `FormData` as the `Input` type parameter for a `RequestHandler`. It is also a breaking change, as it results in some errors from Typescript (even in this codebase, see commit ae40649).

EDIT: I also wonder if the types for `RequestHeaders` are also incorrect, as both Node's [`IncomingHttpHeaders`](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules__types_node_http_d_._http_.incominghttpheaders.html) and spec's [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) objects return either `undefined` or `null` when the header is not present.

### Tests

### Changesets

### DIFF
--- a/.changeset/rotten-tips-teach.md
+++ b/.changeset/rotten-tips-teach.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Change ReadOnlyFormData behavior to mimic the spec's FormData interface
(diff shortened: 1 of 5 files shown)


Rich-Harris (approved):

Rich-Harris:
Thank you!

Issue #3236, opened 2022-01-07
History state is overwritten to `{}` during navigation since `next.208`


gustavopch:
### Describe the bug

Since `@sveltejs/kit@1.0.0-next.208`, `goto()`'s `state` parameter stopped working. This line is overwriting the state to `{}` whenever navigation occurs:

https://github.com/sveltejs/kit/blob/1a6adc8a7edbb19cb10762af1a236090dcce97b5/packages/kit/src/runtime/client/router.js#L281

This is the commit that introduced the bug:  https://github.com/sveltejs/kit/commit/d17f459ef35f07f012d77ff7c8665e26380448cd.

### Reproduction

- Open the repro: https://stackblitz.com/edit/sveltekit-xfxpb3
- Click the button "Goto with state".
- It will take you to another route where the History state will be visible.
- You'll see the state will be empty. It should be what was passed to `goto()`.

### System Info

```shell
System:
    OS: macOS 12.0.1
    CPU: (8) arm64 Apple M1
    Memory: 112.81 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.1 - ~/.asdf/installs/nodejs/16.13.1/bin/node
    Yarn: 1.22.17 - ~/.asdf/installs/nodejs/16.13.1/.npm/bin/yarn
    npm: 8.2.0 - ~/.asdf/installs/nodejs/16.13.1/.npm/bin/npm
    Watchman: 2021.11.15.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 97.0.4692.71
    Safari: 15.1
  npmPackages:
    @sveltejs/adapter-auto: 1.0.0-next.9 => 1.0.0-next.9 
    @sveltejs/kit: 1.0.0-next.208 => 1.0.0-next.208 
    svelte: 3.44.3 => 3.44.3
```

### Severity

blocking all usage of SvelteKit

benmccann:
@gustavopch thanks for the investigation! care to send a PR?

gustavopch:
@benmccann This would fix the problem, but I'm not sure if it might break some other behavior. What do you think?
```diff
- history.replaceState({}, '', info.url);
+ history.replaceState(history.state, '', info.url);
```

zommerberg:
This won't be needed once #3241 is merged

benmccann:
Rich sent https://github.com/sveltejs/kit/pull/3312 to remove this feature, which would in effect close this issue.

[...]

PR #3312, opened 2022-01-12
Remove `state` option from `goto`

Rich-Harris:
This came from #1643 and references #1641 but doesn't actually do anything useful as far as I can tell — it's one part of a three part proposal that was never discussed and is useless without the other two

### Tests

### Changesets


packages/kit/src/runtime/client/router.js
```diff
@@ -225,7 +221,7 @@ export class Router {
 				keepfocus,
 				chain,
 				url.hash,
-				state,
+				{},
```
benmccann, line 224: if you're removing it, I'd think it would be removed from `_navigate` as well?

Rich-Harris: ah yep, i thought it was used in the other cases for scroll management, but evidently not. i think it's needed in #3293 but there's already going to be a merge conflict there anyway

Rich-Harris:
Going to close this as @gustavopch described a use case in https://github.com/sveltejs/kit/issues/3236#issuecomment-1011436921. I think we can probably do a better job of enabling that use case (and related ones, like the modal overlay you get when you click a photo in on an instagram.com profile, instead of being navigated to the photo page) without the drawbacks of exposing `history.state` like this (namely that `goto` triggers a useless navigation, and we're forced to mutate the state object), but that can come later.

Issue #3286, opened 2022-01-11
304 responses should contain etag


Maggi64:
### Describe the bug

Currently the correct headers are missing on 304 responses.

The IETF standards i founds say this:
`The server generating a 304 response MUST generate any of the following header fields that would have been sent in a 200 (OK) response to the same request: Cache-Control, Content-Location, Date, ETag, Expires, and Vary.`
https://datatracker.ietf.org/doc/html/rfc7232#section-4.1

https://github.com/sveltejs/kit/blob/b10ba93136b93bccb8ce3eac5508034fa0b8d8cd/packages/kit/src/runtime/server/index.js#L118-L140

For me this causes cloudflare caching to not revalidate my api responses correctly when a `cache-control: max-age=60` header was set. Cloudflare support was so kind to point me to this issue.

### Reproduction

Can create a repo if requested

### System Info

```shell
"@sveltejs/adapter-node": "1.0.0-next.61",
"@sveltejs/kit": "1.0.0-next.222",
```

### Severity

serious, but I can work around it

### Additional Information

I would suggest the following change to fix this issue.

```javascript
const etag = `"${hash(response.body || '')}"`;
response.headers['etag'] = etag
if (if_none_match_value === etag) {
  return {
    status: 304,
    headers: response.headers
  };
}

Conduitry:
Interesting, I wasn't aware of this. It does indeed sound in the spec that we should be returning an ETag here. Thanks for the report!

vicky1999:
I am interested to solve this issue.  Can I work on this issue?

bluwy:
@vicky1999 feel free if no one created a PR yet.

bluwy:
#3133 fixed this

PR #3313, opened 2022-01-12
304 response should contain etag

vicky1999:
Solves #3286 

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -127,15 +127,14 @@ export async function respond(incoming, options, state = {}) {
 								}
 
 								const etag = `"${hash(response.body || '')}"`;
+								response.headers['etag'] = etag;
 
 								if (if_none_match_value === etag) {
 									return {
 										status: 304,
-										headers: {}
+										headers: response.headers
 									};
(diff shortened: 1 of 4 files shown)


packages/kit/src/runtime/server/index.js
```diff
@@ -127,15 +127,14 @@ export async function respond(incoming, options, state = {}) {
 								}
 
 								const etag = `"${hash(response.body || '')}"`;
+								response.headers['etag'] = etag;
 
 								if (if_none_match_value === etag) {
 									return {
 										status: 304,
-										headers: {}
+										headers: response.headers
```
Conduitry, line 135: I don't think we need to include `content-type` or any of the other headers in `response.headers`. We can probably move `response.headers['etag'] = etag;` back where it was, and do `headers: { etag }` here instead.

vicky1999: okay

vicky1999:
only etag is added to the response if it is a 304 response

bluwy (approved):

Maggi64:
Yes that looks like it would fit the spec, thanks!

PR #3318, opened 2022-01-13
[fix] page store origin http

### LINKED ISSUE #3248: $page.url.origin is set to https when a page is served over http
### Describe the bug

$page.url is not setto the actual URL of the page in adapter-node when served over http 

### Reproduction

```svelte
<script>
</script>

{$page.url.origin}
```

Use adapter-node, compile and serve the application with`svelte-kit build && node build`, load `http://localhost:3000` in your browser. 

expected: `http://localhost:3000`
actual: `https://localhost:3000`

### System Info

```shell
latest sveltekit
```

### Severity

### Additional Information

_No response_

nhe23:
Fixes #3248 by setting the protocol in the start.js in the client.

### Tests

### Changesets


nhe23:
Should this include a changeset? This is just a minor fix that is probably not worth mentioning in the changelogs...

Rich-Harris:
Thank you, but I'm not sure this is desirable behaviour. If `$page.url` is incorrect on first load, then it's incorrect on the server render too — this is _masking_ the issue rather than _fixing_ it. If something relies on the protocol being correct, then the configuration options are available to make it so:

* https://kit.svelte.dev/docs#configuration-headers
* https://kit.svelte.dev/docs#configuration-protocol

I could see us maybe adding some sort of warning for when protocols don't match, pointing to the docs above (though only in dev, which perhaps isn't that valuable since you're more likely to hit it in prod), but I don't think the behaviour itself needs changing.

Re changesets — it's partly about the changelog, but the main reason to use them is that they enable release automation. If something should trigger a release (and bugfixes definitely qualify, however minor) then the PR does need a changeset.

nhe23:
I agree. However the current behaviour makes it difficult to test the the built application locally because once deployed (behind a webserver) the desired protocol is HTTPS and therefore changing the svelte config would not make sense. 
The only other way to solve this issue is to have different configuration values for building when testing locally. If the values where dependent on some environment variable then I guess only one configuration would be needed. 
The workflow would look like this: 
Build (with protocol set to http) -> test locally -> rebuild (with protocol set to https) -> deploy

The solution for the corresponding issue would then be to make this behaviour clearer in the docs an maybe point out how it could be solved with the above workflow.

Conduitry:
I have an app where I want to be testing the actual production build but doing so over HTTP (which means that test-only build-time configuration is not an option), and what I've done is put a proxy in front of the app in CI that sets an `x-forwarded-proto: http` header on all requests going to the app.

nhe23:
I think the documentation label should be added to #3248 again as changing the current behaviour is not the desired way. 
It would be probably enough to add some information about this in the `$app/stores` section that mentions the page store currently like this: 

- `page` contains an object with the current [`url`](https://developer.mozilla.org/en-US/docs/Web/API/URL), [`params`](#loading-input-params) and [`stuff`](#loading-output-stuff).

Rich-Harris:
> which means that test-only build-time configuration is not an option

would an env var make your life easier, per @nhe23's suggestion? presumably you could then do away with the proxy?

Conduitry:
It would remove _this_ need for the proxy, but in my case it was already there for a bunch of stuff (including in local development) to insert other headers and simulate some behavior of API gateway things that happen in production.

I hadn't pushed for a non-header solution when we were discussing this a little ways back because it wouldn't have really helped me - but I do see that if you don't already have a proxy as part of your setup, adding one would be a lot more of a headache than just setting an additional env var would.

benmccann:
Could we use the `:scheme` and `:authority` headers for HTTP/2 users? It looks to me like that should make it generally work without having to worry about the user setting anything like env variables, x-forwarding headers, etc.

nhe23:
Basically all Browsers only support HTTP/2 with TLS: [https://caniuse.com/http2](https://caniuse.com/http2). Unfortunately this won't work...

Rich-Harris:
I'm going to close this because as of #3384 most adapters will use the native `Request` object which will have the correct protocol. `adapter-node` is gaining the ability to set a base URL via an environment variable, which basically solves the problem. Thanks

talesmgodois:
Really buggy feature this one.

PR #3322, opened 2022-01-13
Test with `src` modules

Rich-Harris:
#3310. First step is moving stuff out of `.svelte-kit/dev` and `.svelte-kit/build` into `.svelte-kit/modules`, since it's the same stuff in both cases. (The contents of `.svelte-kit` needs a bit of a tidy-up generally, and this is a good time to do it.)

* [ ] This shouldn't apply in CI

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [ ] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [ ] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [ ] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #3327, opened 2022-01-13
shuffle things around a bit

Rich-Harris:
this is a checkpoint on the way to #3322, which I can't seem to get to work in CI. figured it'd be easier to get the reorganisation stuff done separately, then try to tackle the src/build thing

### Tests

### Changesets


packages/kit/src/core/utils.js
```diff
@@ -3,10 +3,13 @@ import path from 'path';
 import colors from 'kleur';
 import { copy } from '../utils/filesystem.js';
 import { fileURLToPath } from 'url';
+import { SVELTE_KIT } from './constants.js';
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
+export const modules = path.resolve(`${SVELTE_KIT}/modules`);
```
Conduitry, line 11:
```suggestion
export const modules = path.posix.resolve(`${SVELTE_KIT}/modules`);
```

Conduitry: I don't know why what I'm suggesting results in `import { start } from "/@fs/blah/kit/packages/kit/test/apps/basics/.svelte-kit/modules/client/start.js";` on Windows rather than `import { start } from "/.svelte-kit/dev/runtime/internal/start.js";` but I'm not especially anxious to dig into that right now.

In any case, the tests look like they're passing for me locally.

packages/kit/.gitignore
```diff
@@ -1,7 +1,7 @@
 .DS_Store
 /node_modules
 /dist
-/assets/runtime
+/assets/*
 /assets/kit.js
```
ignatiusmb, line 5: while we're here, might as well remove this

```suggestion
```

ignatiusmb (approved):

benmccann (approved):
lgtm - though I didn't look super closely since it doesn't seem like there are any huge changes

Issue #2838, opened 2021-11-18
Make dev & preview be friendlier for base!='/'


tv42:
### Describe the problem

Right now, if I set `svelte.config.js` `config.kit.paths.base = '/foo'`, and run `npm run dev` or `npm run build && npm run preview`,

- http://localhost:3000 gives an unstyled "Not found"
- http://localhost:3000/foo gives a styled 404 page from `__error.svelte` (does not use `src/routes/index.svelte`)

### Describe the proposed solution

Two things:

1. The base URL without slash should probably render `src/routes/index.svelte`, like all the things below it behave.
2. It would be nicer if SvelteKit recognized these scenarios that will never work (`/`, `/foo` if the above is not considered a bug), and simply redirected the poor developer to what works: `/${base}/`, in this case `/foo/`. If item 1 from above is a bug that can be fixed, then the only redirect needed would be `/` -> `/${base}`.

### Importance

would make my life easier

PR #3337, opened 2022-01-14
Fix svelte-kit dev/preview when config.kit.paths.base is set

andrewsoutar:
This fixes a few small issues I had when using `svelte-kit dev` and `svelte-kit preview` when `config.kit.paths.base` is set:

- The base-prefixed root path would be resolved to an empty-string path (instead of "/"), causing it not to be found; looks like #2838 is related
- Vite always serves the static files from the server's root directory, so update the `path.assets` of the build to look there (not sure if there is an issue for this)

After these changes, `svelte-kit dev` and `svelte-kit preview` seem to be working for my (small) app together with `paths.base`.

-------

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -179,7 +179,7 @@ export async function create_plugin(config, output, cwd) {
 
 						paths.set_paths({
 							base: config.kit.paths.base,
-							assets: config.kit.paths.assets ? SVELTE_KIT_ASSETS : config.kit.paths.base
+							assets: config.kit.paths.assets ? SVELTE_KIT_ASSETS : ''
 						});
 
 						let body;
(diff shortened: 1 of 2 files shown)


Rich-Harris:
Thank you! Unfortunately I don't think this covers all the cases — it fixes the case where a request is made for `/basepath` (the existing test missed this case because it coincided with `trailingSlash: 'always'`, which meant the truncated path was `/` as expected), but it doesn't appear to change how static assets are handled.

Specifically, if you have a `static/hello.txt` file, it should be served at `/basepath/hello.txt`, but instead it's served at `/hello.txt`.

This needs a test (and changeset) before it can be merged. I went ahead and added one, but since I couldn't make modifications to this branch ([here's how to make PRs editable by maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)) I opened a new PR — #3345 — so I'll close this one.

I'll open a separate PR with a failing test for static assets.

PR #3338, opened 2022-01-14
add jwt auth example

mishrasatyam:
Adding a json web token (JWT) authentication **example** in examples directory. I see a lot of people have doubts about hooks, setting cookie, auth etc and an example will be helpful.

### Tests

### Changesets


Rich-Harris:
Thank you but this isn't the place for this sort of thing. Eventually we'll create a space for this sort of documentation, but it should be small and self-contained (i.e. **just** the JWT stuff, not an entire app, certainly not one that includes tailwind etc)

uvulpos:
@Rich-Harris since you guys also recommend Svelte-Kit for People who just want to use Svelte with an external Backend (in Java, Go, PHP or whatever), I would appreciate it, if we could add this into the https://learn.svelte.dev/ course website. Maybe into a new chapter because I would agree with @mishrasatyam that many people struggle with that and it is an important topic. 

What do you think?

PR #3339, opened 2022-01-14
fix: prerendering encodes url multiple times

ivanhofer:
Currently the `prerendering` of SvelteKit will take the url from a redirect and pass it to the `encodeURI` function. This causes url parts e.g. query strings that are already encoded to be encoded a second time which may lead to incorrect urls.

This PR fixes that issue. Redirect urls only get encoded a single time.

### Tests

> linting will fail because of `src/core/adapt/test/fixtures/prerender/build/redirect-url-encoding/index.html`. I was not able to exclude the file from prettier, without removing the functionality for reading the ignore patterns from the `.gitignore` file. What should we de here?

### Changesets


packages/kit/src/core/adapt/prerender/prerender.js
```diff
@@ -167,8 +167,16 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 				if (location) {
 					mkdirp(dirname(file));
 
+					let url;
+					try {
+						url = new URL(location);
```
Rich-Harris, line 172: This would fail in the case of non-absolute URLs (e.g. `location: /other`)

packages/kit/src/core/adapt/prerender/prerender.js
```diff
@@ -167,8 +167,16 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
...
+					} catch (error) {
+						log.error(`'${location}' is not a valid URL`);
+						url = location;
+					}
+
 					log.warn(`${rendered.status} ${decoded_path} -> ${location}`);
-					writeFileSync(file, `<meta http-equiv="refresh" content="0;url=${encodeURI(location)}">`);
+					writeFileSync(file, `<meta http-equiv="refresh" content="0;url=${url.toString()}">`);
```
Rich-Harris, line 179: This should probably just be `location`?
```suggestion
					writeFileSync(file, `<meta http-equiv="refresh" content="0;url=${location}">`);
```

packages/kit/src/core/adapt/test/index.js
```diff
@@ -130,7 +130,19 @@ suite('prerender', async () => {
 		dest
 	});
 
-	assert.equal(glob('**', { cwd: prerendered_files }), glob('**', { cwd: dest }));
+	const expectedFiles = glob('**', { cwd: prerendered_files });
+	const actualFiles = glob('**', { cwd: dest });
+
+	// test if all files are present
+	assert.equal(expectedFiles, actualFiles);
```
Rich-Harris, line 137: house style — we use snake_case instead of camelCase for anything that isn't public API, i.e. `expected_files` rather than `expectedFiles` (though single-word variables are better still). `assert.equal` takes `(actual, expected)`, not `(expected, actual)`

```suggestion
	const expected = glob('**', { cwd: prerendered_files });
	const files = glob('**', { cwd: dest });

	// test if all files are present
	assert.equal(files, expected);
```

Rich-Harris: (I realise there is some adjacent camelCased code that snuck in in a previous PR; will address that separately)

Rich-Harris (changes requested):
Thank you — comments inline. Please remember to add a changeset :)

Rich-Harris:
> linting will fail because of `src/core/adapt/test/fixtures/prerender/build/redirect-url-encoding/index.html`. I was not able to exclude the file from prettier, without removing the functionality for reading the ignore patterns from the `.gitignore` file. What should we de here?

this is a deeply stupid consequence of the lack of an `ignore` option in .prettierrc files, but there's a hack that works — add `requirePragma: true` for specific files https://github.com/sveltejs/kit/blob/9de809405f13a24268da167d3c50ad894ecd14a8/.prettierrc#L13-L24

packages/kit/src/core/adapt/prerender/prerender.js
ivanhofer: I'm not sure why it was originally set to `encodeURI`. I didn't want to break it so I first convert it to an `URL` object.
Is it possible that a unencoded url reaches this part? I think yes, so we can't use `location` here.

packages/kit/src/core/adapt/prerender/prerender.js
ivanhofer: Right! So it is either a string starting with a protocol `'http://'` or a string starting with `'/'` right? Or are there also other cases?

I wasn't sure about the error handling. What exactly should happen if `location` is an invalid url?
Right now it just uses the unencoded `location` instead

ivanhofer:
I will address the comments in the next days!

packages/kit/src/core/adapt/prerender/prerender.js
Rich-Harris: Relative URLs (`./foo`) are also fine. This isn't the place to worry about whether they're valid or not — if they're not, then the next step (where we crawl `location`) will fail

packages/kit/src/core/adapt/prerender/prerender.js
Rich-Harris: Spelunking through the git history led me here: https://github.com/sveltejs/kit/pull/256

I think we need to escape the string, so that this doesn't cause havoc...

```
location: "><script>alert('pwned')</script>
```

...but it seems like `encodeURI` was probably the wrong choice.

ivanhofer: > ```
> location: "><script>alert('pwned')</script>
> ```
This will no longer be a problem because it is an invalid URL. The code will throw an error

Rich-Harris: This is a valid (unencoded) URL:

```
https://example.com/"><script>alert('pwned')</script>
```

ivanhofer: and with this implementation, this would output the url as:
`"https://example.com/%22%3E%3Cscript%3Ealert('pwned')%3C/script%3E"`

Rich-Harris:
Thank you. I think we can do this more simply by just escaping the attribute — I've opened #3456, will close this

PR #3345, opened 2022-01-14
Fix truncated basepath

Rich-Harris:
Supersedes #3337. Ensures that requests for `/basepath` (where `paths.base === '/basepath`) resolve correctly (the existing tests didn't catch this because of `trailingSlash: 'always'`).

This _doesn't_ fix static asset handling, which is a bit finicky — will open a separate PR for that.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/adapt/prerender/prerender.js
+++ b/packages/kit/src/core/adapt/prerender/prerender.js
@@ -230,8 +230,12 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 
 					const parsed = new URL(resolved, 'http://localhost');
 
-					if (!parsed.pathname.startsWith(config.kit.paths.base)) continue;
-					const pathname = decodeURI(parsed.pathname).slice(config.kit.paths.base.length) || '/';
+					let pathname = decodeURI(parsed.pathname);
+
+					if (config.kit.paths.base) {
+						if (!pathname.startsWith(config.kit.paths.base)) continue;
+						pathname = pathname.slice(config.kit.paths.base.length) || '/';
+					}
 
 					const file = pathname.slice(1);
 					if (files.has(file)) continue;
--- a/packages/kit/src/core/adapt/prerender/prerender.js
+++ b/packages/kit/src/core/adapt/prerender/prerender.js
@@ -229,7 +229,9 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 					if (!is_root_relative(resolved)) continue;
 
 					const parsed = new URL(resolved, 'http://localhost');
-					const pathname = decodeURI(parsed.pathname).replace(config.kit.paths.base, '');
+
+					if (!parsed.pathname.startsWith(config.kit.paths.base)) continue;
+					const pathname = decodeURI(parsed.pathname).slice(config.kit.paths.base.length) || '/';
 
 					const file = pathname.slice(1);
 					if (files.has(file)) continue;
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -105,7 +105,7 @@ export async function respond(incoming, options, state = {}) {
 					});
 				}
 
-				const decoded = decodeURI(request.url.pathname).replace(options.paths.base, '');
+				const decoded = decodeURI(request.url.pathname).replace(options.paths.base, '') || '/';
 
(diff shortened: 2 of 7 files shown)


packages/kit/src/runtime/server/index.js
```diff
@@ -105,7 +105,7 @@ export async function respond(incoming, options, state = {}) {
 					});
 				}
 
-				const decoded = decodeURI(request.url.pathname).replace(options.paths.base, '');
+				const decoded = decodeURI(request.url.pathname).replace(options.paths.base, '') || '/';
```
Conduitry, line 108: Should the similar line in `prerender.js` get the same treatment?

Separately, should these be adjusted to only do the replacement at the start of the string? Is it possible that `options.paths.base` won't appear at the beginning of the path but does (by chance) appear later in the string?

If we know that `request.url.pathname` will _always_ begin with `options.paths.base`, can we just do `decoder(request.url.pathname).slice(options.paths.base.length) || '/'`? I'd expect this to be very slightly faster. And if it doesn't always begin with that, then I don't think this replacement is safe anyway.

Rich-Harris: good points, will update. i don't think we can guarantee that `request.url.pathname.startsWith(options.paths.base)` — someone might be, i don't know, running several SvelteKit node apps from a single Express server

Rich-Harris: updated

packages/kit/src/core/adapt/prerender/prerender.js
```diff
@@ -229,7 +229,9 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 					if (!is_root_relative(resolved)) continue;
 
 					const parsed = new URL(resolved, 'http://localhost');
-					const pathname = decodeURI(parsed.pathname).replace(config.kit.paths.base, '');
+
+					if (!parsed.pathname.startsWith(config.kit.paths.base)) continue;
```
Conduitry, line 233: Do we need to decode first before checking this?

andrewsoutar:
Getting static assets to work with `svelte-kit dev` is indeed tricky. What I've tried:
- In my original changes, I simply set `$app/paths.assets` to be empty in the build; Vite was already serving the assets from the root path, so as long as every asset use is prefixed by `$app/paths.assets`, everything works. Of course, this breaks if asset paths are hardcoded, relative, etc. These cases were already broken.
- Change Vite's base path - right now it's hardcoded to /. Changing it to `config.kit.paths.base + '/'` seems to work (in my limited testing) if I hack the base path back in before sending it to the router for SSR - assets are correctly served from `paths.base`. But it fails if the user has a custom `paths.assets`, because the app is built with the asset path as `/_svelte_kit_assets`, but the assets are served from `paths.base/svelte_kit_assets`. We could change the asset path in the app build, but then it breaks parity with `svelte-kit preview`. We could also use a combination of that scheme and the existing one, depending on whether `paths.assets` is set, but that sounds like a mess.
- Serve the static assets "by hand", rather than having Vite do it for us. `svelte-kit preview` already does this. Honestly, I don't know enough about Vite to know what the tradeoffs are here.

packages/kit/src/core/adapt/prerender/prerender.js
andrewsoutar: In a couple of [other](https://github.com/sveltejs/kit/blob/9e482ac43b2d3f903f497ab5c9cc92691aec9906/packages/kit/src/core/preview/index.js#L94) [places](https://github.com/sveltejs/kit/blob/9e482ac43b2d3f903f497ab5c9cc92691aec9906/packages/kit/src/core/dev/plugin.js#L138), it seems to be checked without decoding. `paths.base` doesn't seem to be documented one way or the other (whether it's encoded or decoded); on the other hand, the encoded form isn't canonical, so e.g. even with a "normal"-looking base of `/foo`, someone visiting via `/%66%6f%6f` might not match, even though it's the "same" URL. (The same issue, I would imagine, applies to the other places where it's checked as well - including probably in adapters - and behavior would vary depending on whether whatever is upstream canonicalizes the URL.) Honestly, not sure how much of a problem that is.

Rich-Harris: updated the PR

bluwy (approved):

PR #3346, opened 2022-01-14
Handling of static assets with `paths.base`

Rich-Harris:
If you have a file like `static/answer.txt`, and a `paths.base` of `/basepath` (with no `paths.assets`), the file should be served at `/basepath/answer.txt`, but instead it's served at `/answer.txt` because Vite doesn't know about the basepath. This PR fixes it by bypassing Vite's `servePublicMiddleware` and implementing it ourselves with `sirv` (which Vite is already using).

### Tests

### Changesets


andrewsoutar:
(Copied over from #3345)

Getting static assets to work with `svelte-kit dev` is indeed tricky. What I've tried:
- In my original changes, I simply set `$app/paths.assets` to be empty in the build; Vite was already serving the assets from the root path, so as long as every asset use is prefixed by `$app/paths.assets`, everything works. Of course, this breaks if asset paths are hardcoded, relative, etc. These cases were already broken.
- Change Vite's base path - right now it's hardcoded to /. Changing it to `config.kit.paths.base + '/'` seems to work (in my limited testing) if I hack the base path back in before sending it to the router for SSR - assets are correctly served from `paths.base`. But it fails if the user has a custom `paths.assets`, because the app is built with the asset path as `/_svelte_kit_assets`, but the assets are served from `paths.base/svelte_kit_assets`. We could change the asset path in the app build, but then it breaks parity with `svelte-kit preview`. We could also use a combination of that scheme and the existing one, depending on whether `paths.assets` is set, but that sounds like a mess.
- Serve the static assets "by hand", rather than having Vite do it for us. `svelte-kit preview` already does this. Honestly, I don't know enough about Vite to know what the tradeoffs are here.

Issue #3305, opened 2022-01-12
When there is not title in the page(inside svelte:head in svelte page component in the src/routes folder) the browser shows title for the last page that had it


PH4NTOMiki:
### Describe the bug

So if a Svelte page doesn't have a title in svelte:head, the browser shows the title for the last page that did have it, and in "normal" sites in this case browser only shows the URL of the page as the browser tab title, and if you reload that URL in the browser it would be inconsistent, so the quick fix would be to check: 
If there is no title for the page when navigating, just set `document.title = '';` and that would fix this.

### Reproduction

[https://stackblitz.com/edit/sveltekit-7mnqkm?file=src/routes/page2.svelte](https://stackblitz.com/edit/sveltekit-7mnqkm?file=src/routes/page2.svelte)

open this link(open in new window in stackblitz), click "Go to Page2" and the browser tab title would still be "Page 1" from the previous link.

### System Info

```shell
latest sveltekit
```

### Severity

annoyance

Rich-Harris:
> just set `document.title = '';` and that would fix this.

Duplicating my comment from #3347: 

---

Unfortunately this creates a bug — if (say) you have a `<title>Foo</title>` in `foo/__layout.svelte`, then clicking between `/foo/bar` and `/foo/baz` will cause the title to be removed (unless the child pages each have a `<title>` of their own).

To my mind this bug is worse than the bug being fixed here, since having a title in a layout that applies to its child pages might be deliberate, and it's worse to interfere with deliberate behaviour than to have buggy behaviour that results from an oversight.

biltongza:
In my case - I have my title set in app.html which I want to be the "default" title, and then sub pages can modify it, but if I navigate away from the page it should revert to the "default" title or whatever title the new page has.

PH4NTOMiki:
@Rich-Harris I understand that, and I'm guessing it's really hard to do some "inheritance" for the titles, but if it is possible that would be great, to look up in the preceding layout components and if there isn't any title in them, and there is no title in the current page component to set it to empty string.

Rich-Harris:
I'm going to close this as I think it's fundamentally impossible to solve at a framework level. The `title` isn't a stack that we can rewind when a component is unmounted, and any attempt to treat it like one is going to run into edge cases. The simple answer is that you do need to specify a title. 

One alternative approach would be to have this in your root layout...

```svelte
<script>
  import { page } from '$app/stores';
</script>

<svelte:head>
[...]

PR #3347, opened 2022-01-14
solves #3305 - When there is not title in the page(inside svelte:head in svelte page component in the src/routes folder) the browser shows title for the last page that had it

vicky1999:
solves #3305 

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -279,7 +279,7 @@ export class Renderer {
 		}
 
 		this.updating = true;
-
+		document.title = '';
 		if (this.started) {
 			this.current = navigation_result.state;
 


Rich-Harris:
Thank you. Unfortunately this creates a bug — if (say) you have a `<title>Foo</title>` in `foo/__layout.svelte`, then clicking between `/foo/bar` and `/foo/baz` will cause the title to be removed (unless the child pages each have a `<title>` of their own).

To my mind this bug is worse than the bug being fixed here, since having a title in a layout that applies to its child pages might be deliberate, and it's worse to interfere with deliberate behaviour than to have buggy behaviour that results from an oversight.

vicky1999:
I have a solution.  but, I am not sure whether this will work or not.  can we set the title to empty based on a condition? for example, if the parent url is different, then make the title as empty.  will this approach solve the issue?
For example, if we have the pages `/foo/bar` and `/foo/baz`.  first, when navigating, it will check if the parent url `foo` is the same.  if so, we don't change the title to empty.

Mlocik97:
no, You would need to extract content of `<svelte:head>`'s `<title>` in `__layout.svelte` and reset it back to it.

PR #3348, opened 2022-01-14
Don't overwrite custom etag

cristovao-trevisan:
Kit should not overwrite custom etag value.

Sometimes it's useful to provide a custom etag value which should not be overwritten by sveltekit.
For example, in my particular use case, when proxying an image from S3 it's possible to retrieve the ETAG with the metadata, without downloading it.
This is also useful when the etag is based on api requests, making it possible to skip rendering the page.

### Tests

### Changesets

### DIFF
--- a/.changeset/clever-readers-turn.md
+++ b/.changeset/clever-readers-turn.md
@@ -2,6 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Don't overwrite ETAG header.
-
-Now user can provide a custom ETAG header, which should be manually checked against the 'if-none-match' header.
+Preserve explicit ETag header
--- a/.changeset/clever-readers-turn.md
+++ b/.changeset/clever-readers-turn.md
@@ -1,0 +1,7 @@
+---
+'@sveltejs/kit': patch
+---
+
+Don't overwrite ETAG header.
+
+Now user can provide a custom ETAG header, which should be manually checked against the 'if-none-match' header.
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -122,8 +122,9 @@ export async function respond(incoming, options, state = {}) {
 							: await render_page(request, route, match, options, state, ssr);
 
 					if (response) {
-						// inject ETags for 200 responses
-						if (response.status === 200) {
+						// inject ETags for 200 responses, if the endpoint
+						// doesn't have its own ETag handling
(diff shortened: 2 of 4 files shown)


Rich-Harris:
Thank you! Can you add a test and a changeset please? You could add a test to this section: https://github.com/sveltejs/kit/blob/9e482ac43b2d3f903f497ab5c9cc92691aec9906/packages/kit/test/apps/basics/test/test.js#L690-L735

Conduitry:
Should we also, then, trust an `etag` from the user's `response` object when we're checking whether we should return a 304?

cristovao-trevisan:
Yes @Rich-Harris , I can add a test and the changeset, will do soon

@Conduitry , I don't think there's a stateless (or serveless) way to check the user provided etag, I assume that if the etag is being set manually the user know what he's doing and will check the if-none-matches equality himself

.changeset/clever-readers-turn.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sveltejs/kit': patch
+---
+
+Don't overwrite ETAG header.
+
+Now user can provide a custom ETAG header, which should be manually checked against the 'if-none-match' header.
```
Rich-Harris, line 7:
```suggestion
Preserve explicit ETag header
```

packages/kit/src/runtime/server/index.js
```diff
@@ -152,7 +152,9 @@ export async function respond(incoming, options, state = {}) {
 									};
 								}
 
-								response.headers['etag'] = etag;
+								if (!response.headers['etag']) {
+									response.headers['etag'] = etag;
+								}
```
Rich-Harris, line 157: Shouldn't we be bailing out on line 121? If there's already an ETag then surely we should never get to this point?

```suggestion
								response.headers['etag'] = etag;
```

Feels like we need to make this change instead:

```diff
-if (response.status === 200) {
+if (response.status === 200 && !response.headers['etag']) {

Rich-Harris (changes requested):
Thanks — response inline

Rich-Harris (approved):

packages/kit/src/runtime/server/index.js
Rich-Harris: made this change

cristovao-trevisan:
Was going to apply your suggestion right now, sorry to be late but thanks for all the work!!

PR #3351, opened 2022-01-15
Implementation: Handling of static assets with `paths.base`

andrewsoutar:
With this PR, `svelte-kit dev` now serves static assets from the appropriate place: `/_svelte_kit_assets` if `paths.assets` is set ([as documented](https://kit.svelte.dev/docs#modules-$app-paths)), otherwise `paths.base`. It accomplishes this by serving static assets with its own `sirv` instance instead of relying on Vite. This neatly bypasses the problems discussed in #3346, with minimal additional complexity. These changes build off the failing test added in #3346 (which now passes).

I based the options for the sirv instance off what was already being used in `svelte-kit preview` and what Vite's `servePublicMiddleware` (which, prior to this change, was serving the static assets).

### Tests

### Changesets

### DIFF
--- a/.changeset/tender-actors-tease.md
+++ b/.changeset/tender-actors-tease.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Handle static assets with /basepath in svelte-kit dev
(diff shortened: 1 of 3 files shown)


Rich-Harris:
thank you! i think this is probably the right approach (bypassing Vite), as annoying as it seems

Issue #3356, opened 2022-01-15
Cannot redefine property: fetch


GauBen:
### Describe the bug

I'm trying to import a built svelte-kit site within another one, and it is not possible because importing `import { handler } from 'sub-site/build/handler'` has side-effects, namely calling `__fetch_polyfill`.

### Reproduction

1. Create a svelte-kit website that uses adapter-node
2. Build this website
3. Create another svelte-kit website with the following configuration

```js
import adapter from '@sveltejs/adapter-node'
import preprocess from 'svelte-preprocess'
import { handler } from 'sub-site/build/handler'

/** @type {import('@sveltejs/kit').Config} */
const config = {
  preprocess: preprocess(),
  kit: {
    adapter: adapter(),
    target: '#svelte',
    vite: {
      plugins: [
        {
          name: 'sub-site',
          configureServer({ middlewares }) {
            middlewares.use(handler) // Load the sub-site here
          },
        },
      ],
    },
  },
}

export default config
```

### Logs

```shell
> Cannot redefine property: fetch
    at Function.defineProperties (<anonymous>)
    at __fetch_polyfill (file:///home/.../node_modules/.pnpm/@sveltejs+kit@1.0.0-next.229_svelte@3.46.1/node_modules/@sveltejs/kit/dist/install-fetch.js:6494:9)
```

### System Info

```shell
SvelteKit v1.0.0-next.229
Adapter Node v1.0.0-next.61
```

### Severity

annoyance

### Additional Information

I'll submit a PR in a few minutes that fixes this issue, the solution is to add `configurable: true` to all definitions here:
https://github.com/sveltejs/kit/blob/f92c3e212dbb3247904fd5c3229c040a9650db41/packages/kit/src/install-fetch.js#L4-L23

GauBen:
@Rich-Harris Could you please bump adapter-node to use kit v231?

bluwy:
#3357 that fixes this has been merged. Closing this.

> Could you please bump adapter-node to use kit v231?

I'm not sure what you mean. adapter-node doesn't have a hard version on kit?

GauBen:
> I'm not sure what you mean. adapter-node doesn't have a hard version on kit?

That's right, but since adapter-node was built and published before #3357 was merged, the dist still contains the old `__fetch_polyfill`.

bluwy:
Ah I see. Looks like adapter vercel, node and netlify builds based on kit. I'll make a changeset for them so they get re-released.

PR #3357, opened 2022-01-15
[fix] make properties overwritable (#3356)

GauBen:
This PR makes import side effects less aggressive by allowing __fetch_polyfill() to run several times.
See #3356 for more details.


Rich-Harris:
Thank you! Mind adding a changeset? Would do it myself but I can't checkout this PR for whatever reason

Rich-Harris (approved):

PR #3362, opened 2022-01-15
[feat] added an option to make handler behave as a pure middleware 

GauBen:
## Updated

I committed to the same branch because I needed the fix to work, so here is a bigger PR...

**New feature:**
Added a build option, `handle404: boolean`, that, when set to false, makes `handler` behave as a pure middleware.

Please see #3363 for context

---

## Old message

Fixed the upper bound of the loop

See also
https://github.com/sveltejs/kit/blob/a79b5bf51b38b778af85ed4ade1d26bf7584c417/packages/kit/src/hooks.js#L22

### DIFF
--- a/.changeset/fresh-carrots-travel.md
+++ b/.changeset/fresh-carrots-travel.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-node': patch
+---
+
+Added an option to make `handler` invoke the next middleware on 404 errors.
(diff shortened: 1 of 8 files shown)


GauBen:
This PR is a bit of a mess, sorry for that

GauBen:
Closing because this PR has a lot of conflicting files and probably wont be merged

PR #3374, opened 2022-01-16
[fix] refactor navigation singletons to avoid storing undefined reference

### LINKED ISSUE #3269: Router (goto, invalidate etc.) not working in production
### Describe the bug

I noticed that the router (goto, invalidate etc.) doesn't seems to me work, upon reaching a reasonable size. 
The singletons module gets bundled inside vendor.js and so the router is set to undefined before the init function is invoked.

```js
/// vendor-c70aba79.js
let router$1;
function init(opts) {
  router$1 = opts.router;
}
/// ....
const router = router$1;
const goto = goto_;
const invalidate = invalidate_;
async function goto_(href, opts) {
  return router.goto(href, opts, []);
}
```
` vendor-c70aba79.js:1117 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'goto')
    at goto_ (vendor-c70aba79.js:1117)
    at HTMLButtonElement.click_handler (index.svelte-8e9d7a38.js:401)`

### Reproduction

[https://github.com/pixelmund/kit-router-not-working](https://github.com/pixelmund/kit-router-not-working)

Steps to reproduce:

-  pnpm install
-  pnpm build 
-  pnpm start
-  Click on the doesn't work button and view error logs

### Logs

```shell
vendor-c70aba79.js:1117 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'goto')
    at goto_ (vendor-c70aba79.js:1117)
    at HTMLButtonElement.click_handler (index.svelte-8e9d7a38.js:401)
goto_ @ vendor-c70aba79.js:1117
click_handler @ index.svelte-8e9d7a38.js:401
```

### System Info

```shell
System:
    OS: Windows 10 10.0.22000
    Memory: 23.29 GB / 31.93 GB
  Binaries:
    Node: 16.2.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.1.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (97.0.1072.55)
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.10
    @sveltejs/adapter-node: ^1.0.0-next.60 => 1.0.0-next.60
    @sveltejs/kit: next => 1.0.0-next.218
    svelte: ^3.44.0 => 3.45.0
```

### Severity

### Additional Information

_No response_

pixelmund:
This PR fixes #3269, once you import $app/navigation on multiple places inside your app and it's included inside chunks/vendor.js the router would be set before the init function is called and would always be undefined.

Current Bundle:
``` js
/// vendor-c70aba79.js
let router$1;
function init(opts) {
  router$1 = opts.router;
}
/// ....
const router = router$1; // Router is undefined and doesnt change on init call
const goto = goto_;
const invalidate = invalidate_;
async function goto_(href, opts) {
  return router.goto(href, opts, []);
}
```

New Bundle: 

``` js
let router;
function init(opts) {
  router = opts.router;
}
const goto = goto_;
const invalidate = invalidate_;
async function goto_(href, opts) {
  return router.goto(href, opts, []);
}
```

### Tests

### Changesets

### DIFF
--- a/.changeset/brown-mangos-tie.md
+++ b/.changeset/brown-mangos-tie.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] #3269 refactor navigation/singletons
--- a/.changeset/brown-mangos-tie.md
+++ b/.changeset/brown-mangos-tie.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-[fix] refactor navigation singletons
+[fix] refactor navigation singletons to avoid storing undefined reference
--- a/.changeset/brown-mangos-tie.md
+++ b/.changeset/brown-mangos-tie.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-[fix] #3269 refactor navigation/singletons
+[fix] refactor navigation singletons
--- a/packages/kit/src/runtime/client/singletons.js
+++ b/packages/kit/src/runtime/client/singletons.js
@@ -1,4 +1,7 @@
-/** @type {import('./router').Router} */
+/**
+ * The router is nullable, but not typed that way for ease-of-use
+ * @type {import('./router').Router}
+ */
 export let router;
 
 /** @type {import('./renderer').Renderer} */
--- a/packages/kit/src/runtime/client/singletons.js
+++ b/packages/kit/src/runtime/client/singletons.js
@@ -1,4 +1,4 @@
-/** @type {import('./router').Router?} */
+/** @type {import('./router').Router} */
 export let router;
 
 /** @type {import('./renderer').Renderer} */
@@ -11,6 +11,6 @@ export let renderer;
  * }} opts
  */
 export function init(opts) {
-	router = opts.router;
+	router = /** @type {import('../client/router').Router} */ (opts.router);
 	renderer = opts.renderer;
 }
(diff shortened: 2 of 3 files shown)


pixelmund:
Don't know why the test is failing seems not related to this change.

packages/kit/src/runtime/client/singletons.js
```diff
@@ -1,4 +1,4 @@
-/** @type {import('./router').Router?} */
+/** @type {import('./router').Router} */
```
bluwy, line 1: I think we can add a jsdoc comment here mentioning this is nullable, but is not typed as so to prevent extra type checks.

benmccann:
I don't understand how this PR changes anything. If the router is not present when `init` is called, won't that still be the case?

Also, I'm not sure if it helps, but Vite is removing the vendor chunk: https://github.com/vitejs/vite/pull/6534

pixelmund:
The router const is assigned with the undefined router$1 and the init call only reassigns the router$1 with the actual router. But the const router stays always undefined. This PR removes the const so the reassignment in init works as expected.

benmccann (approved):
Ah, I see. No still seems like a good change to make even on top of the Vite change

packages/kit/src/runtime/client/singletons.js
benmccann: I've updated it to address this

.changeset/brown-mangos-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] refactor navigation singletons
```
benmccann, line 5:
```suggestion
[fix] refactor navigation singletons to avoid storing undefined reference
```

benmccann (approved):

Issue #3363, opened 2022-01-15
adapter-node's handler should never show a 404


GauBen:
### Describe the problem

The `ssr` middleware outputted by adapter-node is not a proper middleware because it never calls `next()`, making it impossible to compose it with other middlewares.

### Describe the proposed solution

I'll submit a PR that'll introduce a new build option, `handle404?: boolean`:
- when set to true (or undefined), `handler` will handle 404 errors and display a custom error page
- when set to false, `handler` will call the next middleware on 404

### Alternatives considered

Make `handler` a pure middleware, and handle 404 in index.js, after `handler`:

https://github.com/sveltejs/kit/blob/668b0aaa2b5626e8f610f395ac551f98eafbd10b/packages/adapter-node/src/index.js#L11-L16

It represents a significant amount of work but allows small performance improvements when using only `handler`.

### Importance

i cannot use SvelteKit without it

Conduitry:
I feel like this could/should be done more flexibly by having the Node adapter look for something on the `response` object that the user could set in the `handle` hook whenever they want - and not just when the status code is 404. I'm not sure what that API ought to look like exactly.

benmccann:
Why can't you just use `handle` to return your custom 404 page?

Conduitry:
I can see how it would be nice for SvelteKit's middleware to be proper middleware and not to have it responsible for calling the code you want it to fall through to.

All I disagree with is having it solely look for HTTP 404 response codes when deciding whether to do so.

GauBen:
> I feel like this could/should be done more flexibly by having the Node adapter look for something on the `response` object that the user could set in the `handle` hook whenever they want - and not just when the status code is 404. I'm not sure what that API ought to look like exactly.

This would feel weird because the `handle` hook is made to be the end of the middleware chain

> Why can't you just use `handle` to return your custom 404 page?

I want to chain middlewares, not just handle 404 pages.
The idea is to chain of several SvelteKit apps, as can be seen here:
https://github.com/GauBen/modularized-sveltekit-app/blob/cf1f5dbb9261690a4f774f04a08959fa18d86d5f/common/index.js#L5-L14

benmccann:
I wonder if it might be cleaner and better performance to instead pass a parameter to `app.render` that says whether to render 404 pages. Then `render` can return nothing instead of rendering a 404 page that you're just throwing away

GauBen:
Yep I thought about this too so I browsed the code to see how hard it would be to implement...

[...]

Issue #831, opened 2021-04-02
Get raw request body in endpoint


max-niederman:
**Is your feature request related to a problem? Please describe.**
I have an endpoint in my SvelteKit app which handles webhook requests from Stripe. Each request is signed so that it can be verified to come from Stripe.

However, the code which verifies this signature needs the raw content of the message. This is a problem because the request has a content-type header of `application/json`, so the development server as well as a couple official adapters (notably `@sveltejs/adapter-node`) parses the request from JSON, preventing me from verifying the request. The other adapters either pass no body or pass the body unparsed.

Personally, it seems *very* weird to me that a SvelteKit application could get exactly the same request and have a **totally different body** passed to the underlying endpoint depending on what adapter was used.

**Describe the solution you'd like**
Simply put, adapters shouldn't parse the body. That code can be removed and put somewhere in the rendering process, so the raw body can be extracted by the `handle` or `getContext` hook. For ease of use, a boolean option could (and probably should) be added as either an export of the endpoint module or somewhere in the Svelte config.

**Describe alternatives you've considered**
-  Expose an additional property (e.g. `rawBody`) on the request object.
-  Remove all body parsing whatsoever.

**How important is this feature to you?**
I would say this issue is of moderate importance, since I can fairly easily make it work with a seperate server for the aforementioned endpoint. Every time I see it, I get a sudden, inexplicable urge to throw up, but it *does* work.

**Additional context**
The code which parses the body is [here](https://github.com/sveltejs/kit/blob/ba4f9b792417e4528f593e6bb6959eb1786376af/packages/kit/src/core/http/get_body/index.js#L4) and an identical function is run in the Node adapter (for example) [here](https://github.com/sveltejs/kit/blob/ba4f9b792417e4528f593e6bb6959eb1786376af/packages/adapter-node/src/server.js#L46).

Conduitry:
Can you use https://kit.svelte.dev/docs#hooks-getcontext to put the raw request body in the context?

acoyfellow:
@Conduitry in my testing: it's already parsed at that point. I couldn't get the rawBody that way - I was attempting to do the same thing @max-niederman was with Stripe

max-niederman:
> Can you use https://kit.svelte.dev/docs#hooks-getcontext to put the raw request body in the context?

@Conduitry No, the request is parsed before the hooks.

babichjacob:
[...]

PR #3384, opened 2022-01-17
Breaking: use `Request` and `Response` objects in endpoints and hooks

Rich-Harris:
This PR introduces a number of breaking changes that

1. lay the foundation for streaming request/response bodies
2. enable multipart form handling (including file uploads)
3. better align SvelteKit with modern platforms that deal with [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request) and [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) objects natively

## How to update your app

Hooks (`handle`, `handleError` and `getSession`) and endpoints previously received a proprietary `Request` object:

```ts
interface Request<Locals = Record<string, any>, Body = unknown> {
  url: URL;
  method: string;
  headers: RequestHeaders;
  rawBody: RawBody;
  params: Record<string, string>;
  body: ParameterizedBody<Body>;
  locals: Locals;
}
```

Instead, they now receive a `RequestEvent`:

```ts
export interface RequestEvent<Locals = Record<string, any>> {
  request: Request; // https://developer.mozilla.org/en-US/docs/Web/API/Request
  url: URL; // https://developer.mozilla.org/en-US/docs/Web/API/URL
  params: Record<string, string>;
  locals: Locals;
}
```

`method` and `headers` are no longer necessary as they exist on the `request` object. (`url` is still provided, since the `URL` object contains conveniences like `url.searchParams.get('foo')`, whereas `request.url` is a string.)

### Updating hooks

The `resolve` function passed to `handle` now returns a `Promise<Response>`; `handle` must do likewise.

A rewritten `handle` hook might look like this:

```diff
-export async function handle({ request, resolve }) {
-  const response = await resolve(request);
+export async function handle({ event, resolve }) {
+  const response = await resolve(event);

-  if (response.headers['content-type'].startsWith('text/html')) {
-    response.body = response.body.replace(/cloud/g, 'butt');
+  if (response.headers.get('content-type').startsWith('text/html')) {
+    const body = await response.text();
+    return new Response(body.replace(/cloud/g, 'butt'), response);
  }
  
  return response;
}
```

(This example illustrates the new APIs but also demonstrates a way in which rewriting HTML is now less efficient; we may need to add a `transformPage` option or something to satisfy that use case in a more streamlined way.)

### Updating endpoints

Similarly, handlers receive a `RequestEvent`. Most `GET` handlers will be unchanged, but any handler that needs to read the request body will need to update:

```diff
-export async function post({ params, body }) {
+export async function post({ params, request }) {
+ const body = await request.formData(); // or request.json(), etc
  await do_something_with(params, body);
  return { status: 201 };
}
```

> In future we will add helpers for handling request bodies in streaming fashion cross-platform

Handlers do _not_ need to return a `Response` object, but they can. Specifically, the `headers` property can be any form that can be passed to the [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) constructor...

```js
headers = new Headers(); // undefined is ok
headers = new Headers({ 'content-type': 'text/html' }); // so is a POJO
headers = new Headers(headers); // so is a `Headers` instance (or a `Map` etc, for that matter)
```

...which means that since a `Response` object has a compliant `status`, `headers` and `body`, you can do this sort of thing:

```js
// src/routes/api/[...path].js
export const get = ({ params }) => fetch(`https://myapi.com/${params.path}`);
```

### Updating svelte.config.js

If you're using the `headers`, `host` or `protocol` config options, you should remove them. Most adapters will automatically set the correct URL without the help of these options; in other cases like `adapter-node` the options have been moved into the adapter options.

## How to update your custom adapter

The interface for adapters has changed as well — the `app.render` function now takes a `Request` and returns a `Promise<Response>`.

On platforms that use these objects natively (like Cloudflare Workers and Deno) this means you can delete some code. On Lambda-like and Node-like platforms you must create the `Request` and handle the `Response` yourself — consult the `adapter-node` and `adapter-netlify` examples:

https://github.com/sveltejs/kit/blob/84f6136c567e9adb85005c23d90c02a87ea269d7/packages/adapter-node/src/handler.js#L37-L49

https://github.com/sveltejs/kit/blob/84f6136c567e9adb85005c23d90c02a87ea269d7/packages/adapter-netlify/src/handler.js#L17-L23

(Note that your adapter _may_ need to expose options to configure the base URL (or protocol/host header) like `adapter-node` does, following the removal of the `headers`/`host`/`protocol` config options.)

## What's next

After these API changes are in place, we can implement support for streaming request/response bodies, which will enable handling of large files and so on (at least on platforms that support streaming, i.e. not Lambda).

<details>

<summary>Original PR comment</summary>

Breaking change — changes the `app.render` signature, which adapters will need to accommodate

* [x] `app.render` should accept a `Request` as input
* [x] `app.render` should return a `Response`
* [x] allow endpoints to return a `Headers` object (enables one-line `fetch` proxying)
* [x] need to expose helpers for converting between `Request` and `Response` and node req/res
* [x] handle streaming bodies (this bit will be finicky since we realistically need to support node readable streams, at least until the standard `ReadableStream` object is supported everywhere — `node-fetch` uses node streams under the hood instead of web streams)
* [x] docs

This doesn't yet tackle file uploads (#70), that can happen in a follow-up I think. It also doesn't mean that Svelte components somehow support streaming, this is purely about simplifying the API and enabling streamed endpoint request/response bodies

Follow-ups:

* [ ] Handle large file uploads ([#issuecomment-1015569661](#issuecomment-1015569661))
* [ ] Figure out if we polyfill `ReadableStream` to enable streamed responses in a cross-platform way

### Tests

### Changesets

</details>


packages/adapter-cloudflare/files/worker.js
```diff
@@ -45,41 +45,14 @@ export default {
...
-				return new Response(rendered.body, {
-					status: rendered.status,
-					headers: make_headers(rendered.headers)
-				});
-			}
 		} catch (e) {
 			return new Response('Error rendering route: ' + (e.message || e.toString()), { status: 500 });
 		}
-
-		return new Response({
```
benmccann, line 65: i'm curious, why isn't 404 handling needed anymore in `adapter-cloudflare` but remains in `adapter-node`? you may want to weigh in on https://github.com/sveltejs/kit/issues/3363#issuecomment-1013919393 which seems like it's quite related to this change

Rich-Harris: It was a mistake that it was there in the first place — `app.render` will always return a response (unless we make the change #3363 is asking for)

Rich-Harris:
A realisation: if the signature of `app.render` changes to `(request: Request) => Promise<Response>`, then the adapter needs to construct a full URL (since `request.url` is always absolute). In the case of modern environments that implement `Request` natively, this is already taken care of; you just pass the request unchanged.

Netlify provides a `rawUrl` property on the event (I _think_ this is a non-standard extension to the Lambda signature, though I could be wrong). Vercel simulates a Node request, but the protocol is always `https` and the host is provided in the `host` header so we can reliably construct a full URL. Vercel edge functions, when we're able to support them, expose `Request` objects like Cloudflare Workers (and Deno).

So of the officially supported adapters, the only one that can actually make use of the `host` and `protocol` config (or `headers.host`/`headers.protocol`) is `adapter-node`. I suspect `adapter-node` is the odd one out, since most adapters are tailored for a specific environment where the full URL is already available or easy to construct reliably.

The conclusion is that we should remove `headers`/`host`/`protocol` from the config, and put it inside `adapter-node` instead.

Rich-Harris:
Another realisation: since request bodies can only be consumed once, and since there are valid reasons to need to control _how_ it's consumed (#831, https://github.com/sveltejs/kit/issues/70#issuecomment-1008432747), we can't realistically parse the body before calling handlers.

And the endpoint-level config idea (https://github.com/sveltejs/kit/issues/70#issuecomment-1008432747) is a non-starter unfortunately, since we would need to parse the body _before_ identifying which endpoint(s) matched the request.

So I've changed my mind a bit about the API — I think we need to expose the underlying `Request` object to handlers:

```diff
-export async function post({ url, method, headers, rawBody, body, params, locals }) {
+export async function post({ request, url, params, locals }) {
+  const body = await request.json();
  // ...
}
```

(Since `request` contains `method` and `headers`, we can ditch those as well as `rawBody` and `body`.)

And this would in fact enable file uploads, albeit in a suboptimal way:

```js
export async function post({ request }) {
  const data = await request.formData();
  await send_to_s3(data.get('video'));

  // ...
}
```

For handling large files this isn't ideal, since it buffers all the data in memory. Essentially what we'd eventually need is a way to handle uploads in streaming fashion — Remix has [parseMultiPartFormData](https://remix.run/docs/en/v1/api/remix#parsemultipartformdata-node) which is close to what I'm imagining, except that it appears to be Node-only. Something like this:

```js
import { multipart } from '$app/somewhere';

export async function post({ request }) {
  for await (const { name, value } of multipart(request)) {
    if (name === 'video') {
      await send_to_s3(value.name, value.type, value.stream());
    }
  }

  // ...
}
```

Happily, that doesn't need to be part of this PR.

documentation/docs/04-hooks.md
```diff
@@ -21,41 +21,23 @@ If unimplemented, defaults to `({ request, resolve }) => resolve(request)`.
...
 
-export type RawBody = null | Uint8Array;
-
-type ParameterizedBody<Body = unknown> = Body extends FormData
-	? ReadOnlyFormData
-	: (string | RawBody | ReadOnlyFormData) & Body;
-
-export interface Request<Locals = Record<string, any>, Body = unknown> {
+export interface RequestEvent<Locals = Record<string, any>> {
+	request: Request;
```
benmccann, line 26: it seems funny to me that we have `url` and `request.url`. would it be possible to extend the request by adding `params` and `locals` to it to avoid this duplication?

I've always hated that the serverless environments call it an event. I think the idea is that you can have serverless endpoints which receive events that are not HTTP requests, but that's not applicable here. I'd love if we could find a way to avoid renaming `request` to `event` in all our docs

.changeset/large-icons-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Expose standard Request object to endpoints
```
Conduitry, line 5: Do we want this to say 'to endpoints and to hooks'? Or something similar to that?

Rich-Harris, line 5:
```suggestion
Breaking: Expose standard Request object to endpoints and hooks
```

documentation/docs/04-hooks.md
Conduitry: `request.url` is a string, unfortunately, and not a `URL` instance. Rich and I just briefly chatted about this on Discord, and it seemed nicer to expose the parsed URL to users, since we are doing it anyway, and since it's reasonable to assume that they will want this data.

Extending official documented-in-MDN classes with additional fields doesn't sit well with me. I really don't think that's a good idea. I'm not wild about the name `event`, but am not sure what would be better.

Rich-Harris: Extending `Request` is a non-starter — it'd be very confusing, mixing and matching stuff that's documented on MDN with our own monkey-patched properties, and in any case we don't own the request object (the adapters do). 

`request.url` is a string, whereas `url` is a `URL` object. An endpoint _could_ do this if they needed it...

```diff
-const foo = url.searchParams.get('foo');
+const foo = new URL(request.url).searchParams.get('foo');
```

...but since we're already generating the `URL` object ourselves, it probably makes sense to just expose it.

I don't love `event` either, but since we _do_ need to have an object with a `request` property (since monkey-patching is a no-go), and `request.request` sounds silly, I couldn't come up with anything better. Open to suggestions though

.changeset/mighty-pandas-search.md
```diff
@@ -0,0 +1,10 @@
+---
+'@sveltejs/adapter-cloudflare': patch
+'@sveltejs/adapter-cloudflare-workers': patch
+'@sveltejs/adapter-netlify': patch
+'@sveltejs/adapter-node': patch
+'@sveltejs/adapter-vercel': patch
+'@sveltejs/kit': patch
+---
+
+change app.render signature to (request: Request) => Promise<Response>
```
Rich-Harris, line 10:
```suggestion
Breaking: change app.render signature to (request: Request) => Promise<Response>
```

.changeset/strong-schools-rule.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sveltejs/adapter-node': patch
+'@sveltejs/kit': patch
+---
+
+Remove protocol/host configuration options from Kit to adapter-node
```
Rich-Harris, line 6:
```suggestion
Breaking: Remove protocol/host configuration options from Kit to adapter-node
```

documentation/docs/04-hooks.md
```diff
@@ -8,11 +8,11 @@ An optional `src/hooks.js` (or `src/hooks.ts`, or `src/hooks/index.js`) file exp
 
 ### handle
 
-This function runs every time SvelteKit receives a request — whether that happens while the app is running, or during [prerendering](#page-options-prerender) — and determines the response. It receives the `request` object and a function called `resolve`, which invokes SvelteKit's router and generates a response (rendering a page, or invoking an endpoint) accordingly. This allows you to modify response headers or bodies, or bypass SvelteKit entirely (for implementing endpoints programmatically, for example).
+This function runs every time SvelteKit receives a request — whether that happens while the app is running, or during [prerendering](#page-options-prerender) — and determines the response. It receives an `event` object and a function called `resolve`, which invokes SvelteKit's router and generates a response (rendering a page, or invoking an endpoint) accordingly. This allows you to modify response headers or bodies, or bypass SvelteKit entirely (for implementing endpoints programmatically, for example).
```
benmccann, line 11:
```suggestion
This function runs every time SvelteKit receives a request — whether that happens while the app is running, or during [prerendering](#page-options-prerender) — and determines the response. It receives an `event` object representing the request and a function called `resolve`, which invokes SvelteKit's router and generates a response (rendering a page, or invoking an endpoint) accordingly. This allows you to modify response headers or bodies, or bypass SvelteKit entirely (for implementing endpoints programmatically, for example).
```

packages/adapter-node/README.md
```diff
@@ -43,17 +49,39 @@ By default, the server will accept connections on `0.0.0.0` using port 3000. The
 HOST=127.0.0.1 PORT=4000 node build
 ```
 
-You can specify different environment variables if necessary using the `env` option:
+Node HTTP servers don't give SvelteKit a reliable way to know the URL that is currently being requested. The simplest way to tell SvelteKit where the app is being served is to set the `BASE` environment variable:
```
benmccann, line 52: it's not really the servers, but the protocol they're implementing
```suggestion
HTTP doesn't give SvelteKit a reliable way to know the URL that is currently being requested. The simplest way to tell SvelteKit where the app is being served is to set the `BASE` environment variable:
```

packages/kit/src/node.js
```diff
@@ -47,3 +49,27 @@ export function getRawBody(req) {
 		});
 	});
 }
+
+/** @type {import('@sveltejs/kit/node').GetRequest} */
+export async function getRequest(base, req) {
+	return new Request(base + req.url, {
+		method: req.method,
+		headers: /** @type {Record<string, string>} */ (req.headers),
+		body: await getRawBody(req)
```
benmccann, line 58: doesn't this sort of prevent you from streaming by consuming the whole body?

packages/kit/src/node.js
```diff
@@ -47,3 +49,27 @@ export function getRawBody(req) {
 		});
 	});
 }
+
+/** @type {import('@sveltejs/kit/node').GetRequest} */
+export async function getRequest(base, req) {
+	return new Request(base + req.url, {
```
benmccann, line 55: should we copy the other request data such as the referrer as well?

packages/kit/src/utils/http.js
```diff
@@ -1,21 +1,22 @@
...
-			return undefined;
+/** @param {Partial<import('types/helper').ResponseHeaders> | undefined} object */
+export function to_headers(object) {
+	const headers = new Headers();
+
+	if (object) {
+		for (const key in object) {
+			const value = object[key];
+
+			if (value) {
```
benmccann, line 9: personal style preference - I generally try invert like `if (!value) continue;` to avoid too many indents. (you could do the same above with `if (object)`

benmccann (commented):
The PR title is "Streams", but there's no mention of streaming in the docs. Is it worth mentioning something about how to do streaming?

dummdidumm:
Meta comment: Since this sounds like a big breaking change, we should add some advice on how to migrate. One possibility would be to enhance the changeset message like "See the PR for more details and a migration path" and then adjust this PR's first post with these contents accordingly.

packages/kit/src/node.js
Rich-Harris: This is where things get a little bit tricky. `node-fetch` doesn't implement `ReadableStream`, it uses node streams for the request body. So in development, preview, and adapter-node apps, you can't actually stream the body by doing this sort of thing:

```js
const reader = request.body.getReader();

let chunk;
while (chunk = await reader.read()) {
  if (chunk.done) {
    return { status: 204 };
  }

  await do_something_with(chunk.value);
}
```

Because of that, there's effectively no difference between exposing the stream and exposing the buffered body — it's going to get buffered regardless, unless people consume the node stream _as_ a node stream, which we'd want to discourage. 

I'm not 100% sure what to do about this — probably something involving proxies and a `ReadableStream` polyfill, until we can drop support for Node <16 — so I figured we could start by getting the basic API in place, then figuring out those details later.

packages/kit/src/node.js
Rich-Harris: not sure I understand? we're passing through the headers, is there something else we're missing?

documentation/docs/10-adapters.md
```diff
@@ -92,7 +92,7 @@ Within the `adapt` method, there are a number of things that an adapter should d
 - Output code that:
   - Imports `App` from `${builder.getServerDirectory()}/app.js`
   - Instantiates the app with a manifest generated with `builder.generateManifest({ relativePath })`
-  - Listens for requests from the platform, converts them to a a [SvelteKit request](#hooks-handle), calls the `render` function to generate a [SvelteKit response](#hooks-handle) and responds with it
+  - Listens for requests from the platform, converts them to a standard [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request if necessary), calls the `render` function to generate a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) and responds with it
```
ignatiusmb, line 95:
```suggestion
  - Listens for requests from the platform, converts them to a standard [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) if necessary, calls the `render` function to generate a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) and responds with it
```

packages/kit/src/utils/http.js
Rich-Harris: it doesn't make sense to return early if `!object` because then you'd be duplicating `return headers`. but i added the `continue`

ignatiusmb (commented):
We should probably add (or enhance) a changeset to mention TS users that there's no more `Body` generics

packages/kit/src/node.js
benmccann: I was looking at the other [constructor options on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request) and wondering if we needed any of them?  `mode`, `credentials`, `cache`, `redirect`, `referrer`, `integrity`

Rich-Harris:
I've updated the title and PR comment to better explain what's going on, and add a migration guide. I'll update error messages to link to it

packages/kit/src/node.js
benmccann: maybe just add a TODO for now?
```
TODO: use whatwg web streams when node-fetch v4 is released
// https://github.com/node-fetch/node-fetch/issues/1452
// https://github.com/node-fetch/node-fetch/issues/1454
```

Rich-Harris:
Alright, I think the remaining comments have been addressed (unless someone has a better name than `RequestEvent`), and this is ready for review:

> The PR title is "Streams", but there's no mention of streaming in the docs

This is mentioned in the top comment as a TODO, and the PR title has been changed

> We should probably add (or enhance) a changeset to mention TS users that there's no more `Body` generics

The changed types are called out in the top comment

> Meta comment: Since this sounds like a big breaking change, we should add some advice on how to migrate. One possibility would be to enhance the changeset message like "See the PR for more details and a migration path" and then adjust this PR's first post with these contents accordingly.

Done! (error messages will link here)

packages/adapter-node/index.js
```diff
@@ -13,7 +13,16 @@ const files = fileURLToPath(new URL('./files', import.meta.url));
...
+	env: {
+		path: path_env = 'SOCKET_PATH',
+		host: host_env = 'HOST',
+		port: port_env = 'PORT',
+		base: base_env,
+		headers: {
+			protocol: protocol_header_env = 'PROTOCOL_HEADER',
+			host: host_header_env = 'HOST_HEADER'
+		}
+	} = {}
```
lovasoa, line 25: @Rich-Harris , you probably meant

```suggestion
	} = {headers:{}}
```

otherwise you get a `Cannot read properties of undefined (reading 'protocol')` when env is not specified.

Conduitry: This is one of the things I'm attempting to address in #3423.

timscodebase:
Is SvelteKit broken?  I have no idea how to fix this and I just want to build a site.  SvelteKit used to work

timscodebase:
Is Sveltekit Broken?  I have no idea how to this and I just want to build a site.  SvelteKit used to work

benmccann:
@timscodebase please view the recently opened issues to see if an issue has already been reported and file a new one if not. We can't help you if we don't know what issue you're experiencing and you haven't shared any details. We're also likely to lose track of comments in closed PRs

benmccann:
FYI community adapter authors @jthegedus @pluvial @geoffrich @MacFJA @HalfdanJ @ryanbethel. This PR will likely require adapter changes

SwaroopH:
> Is SvelteKit broken? I have no idea how to fix this and I just want to build a site. SvelteKit used to work


@timscodebase I ran into this (pointed to PR) when I deployed on a server with an older node version (v14 I think). Works well on v17.3.0

damevin:
@timscodebase Did you fix your problem ? I've the same of you 🙃

Mravuri96:
[HTML minifier](https://github.com/sveltejs/kit/blob/master/documentation/migrating/99-integrations.md) docs needs correction.

mattiasinokuchi:
Does this mean simple requests from a html form is no longer possible to use?

hyunbinseo:
Is there a way to type endpoint response (return value) using the type RequestHandler? Used to type the response by passing the response type as the second generic of RequestHandler, but it is no longer available.

```
// packages/kit/test/apps/basics/src/routes/routing/shadow/action.js

- /** @type {import('@sveltejs/kit').RequestHandler<any, FormData>} */
- export function post({ body }) {
+ /** @type {import('@sveltejs/kit').RequestHandler} */
+ export async function post({ request }) {
+ 	const body = await request.formData();
```

gressus:
**issue with cookies**

according this post, i changed 
`export async function handle({ request, resolve }) {`
to
`export async function handle({ event, resolve }) {`

but now can't get cookie like this: 
`event.request.headers?.cookie == undefined`

before this change i successfully got cookie like this:
`request.headers?.cookie`

Has anyone else experienced this, any ideas?

babichjacob:
@gressus Header values can be retrieved from `Headers` objects with the `.get` method:

```js
const cookie = headers.get("cookie");
```

lkj4:
I just did the migration which took me half a day and some headache for a medium-sized app. Most users will typically need to rewrite their entire backend/api code. My two cents: As one mentioned we  need a migration guide, migrating is a bigger issue than eg going from webpack 4 to 5. We should also stress this in the console log, so not just with this one-liner but a multi-line hash-framed sign right in the face. We should mention some advantages, otherwise users might freak out and think we went nuts. Let just one user rant on reddit how this bump broke his entire code base and we can join the Next repo haha.

One (minor) thing I didn't get: Before the endpoint function headers were clean, now you need to have something like this if you want to type locals:

```
export async function post({
  request,
  url,                         
  params,                           
  locals,                         
}: {
  request: Request 
  url: URL        
  params: Record<string, string>    
  locals: Locals                  
}) {
```

hyunbinseo:
@lkj4 https://github.com/sveltejs/kit/pull/3384 is a PR merged 18 days ago, and is released in [1.0.0-next.234](https://github.com/sveltejs/kit/blob/master/packages/kit/CHANGELOG.md#100-next234).

In that version, you are able to type `Locals` in endpoints using the `RequestHandler`.

```
export const post: RequestHandler<Locals> = async ({ locals }) => ({ status: 200 });
```

```typescript
  Locals = Record<string, any>,
  Output extends DefaultBody = DefaultBody
> {
  (request: RequestEvent<Locals>): MaybePromise<
    Either<Response | EndpointOutput<Output>, Fallthrough>
  >;
}
```

---

The current version released is [1.0.0-next.260](https://github.com/sveltejs/kit/blob/master/packages/kit/CHANGELOG.md#100-next260), in which you can type [`App.Locals`](https://kit.svelte.dev/docs#typescript-app-locals).

---

It is clearly stated in the [FAQ](https://kit.svelte.dev/faq#version-numbers) that SvelteKit is in beta and can have changes in its APIs.

> SvelteKit is not 1.0 yet. Should I use it? What about Sapper?

> SvelteKit is currently in beta while we finalize the APIs.

It is advised that you follow the [change-log](https://github.com/sveltejs/kit/blob/master/packages/kit/CHANGELOG.md) for any breaking changes.

Reference discussion on receiving updates on new versions. https://github.com/sveltejs/kit/discussions/3618

ChristianSch:
@hyunbinseo we know that there will be breaking changes. I think @lkj4 meant it more in the lines of "how can I get my stuff to work again". I also feel that the information given by the project is not enough to understand what's going on and I'm still trying to fix the stuff that was broken by this PR.

akiarostami:
@hyunbinseo, I have to agree with @ChristianSch. We understand that Sveltekit is in beta and major changes can and will happen. At the same time, Sveltekit has been promoted as a stable and reliable solution (and it is!) so many of us have started using sveltekit in our products. These changes with no clear documentation (unless one goes and reads all the details of every minor update). All the breaking changes in the last few weeks with no clear way of knowing how to change our code have been very time-consuming (and frustrating).

dummdidumm:
Could you explain where exactly the frustration is coming from? Is it not knowing what is a breaking change? Is it having to read the changelog? Is it too brief migration docs/explanations in the PR, or that you expect these docs to appear somewhere else?

akiarostami:
@dummdidumm of course. 

Most of the breaking changes are mentioned in the [change-log](https://github.com/sveltejs/kit/blob/master/packages/kit/CHANGELOG.md) page, but not all. For example, [#3576](https://github.com/sveltejs/kit/pull/3576) was a breaking change, but it's not mentioned in the change log page.
Besides that, some of these changes are not as easy as removing an entry from the config file (like #3576), and require a full understanding of the change. [#3384](https://github.com/sveltejs/kit/pull/3384) is a good example. It requires us to completely understand the new ways of handling requests in the hook and endpoints. These changes were not in the documentation when they took effect, and while my under-development app stopped working, the documentation only reflected the old methods and not the new changes. It took me a couple of days to figure out how to change my auth to work with the new changes.

Again, let me reemphasize my love for Svelte/Sveltekit and how much I appreciate the effort.

benmccann:
thanks. these are good call outs. I sent a PR to try to address the points you mentioned: https://github.com/sveltejs/kit/pull/3769

If there are other please let me know on the PR and I can update it

> For example, https://github.com/sveltejs/kit/pull/3576 was a breaking change, but it's not mentioned in the change log page.

I'm assuming you meant 3578

> These changes were not in the documentation when they took effect

Hmm. That's very surprising. The PR included the docs and the docs should get deployed as soon as they are updated

hyunbinseo:
@ChristianSch @akiarostami I agree that there were major breaking changes and that they are hard to follow.

I have given up updating `@sveltejs/kit` in a few month-old projects since migration seems rather challenging.

---

@dummdidumm Would be nice if [documentation](https://kit.svelte.dev/docs) for different `@next` versions is provided, even in the beta state.

It would be even better if documentation `diff` for two `@next` versions is provided, showing what has changed.

For example, I cannot access proper documentation for `@1.0.0-next.197` which is used in one of my projects.

dummdidumm:
Here is how I imagine doing this:
1. follow the changelog
2. only update to the version where a breaking change occurs. Read the PR and changelog for a description of the change and adjust accordingly. Repeat after migration is done with the next version that contains a breaking change as long as there are newer versions. In other words: don't bump to the latest version directly.

I didn't need to do this myself, so I'm not sure if this really works for you in practice. Did you do the same? If not, do you feel this approach would help?

ChristianSch:
@dummdidumm (and benmccan) thank you for taking the time.

>  Is it too brief migration docs/explanations in the PR, or that you expect these docs to appear somewhere else?

I think this would be my biggest problem. I don't update every version but tried to update to 260 from 201. I think I got the event changes working, but my build is broken (`> Cannot read properties of undefined (reading 'file')`), the types aren't detected for `App.locals` (which are specified in src/global.d.ts) and my hooks aren't working, breaking the auth completely. I bet most of them are due to my inexperience with svelte and typescript, but I can't google anything and don't know how to fix them because there are no resources or even questions/tickets or examples about these.

Breaking changes aren't a problem, I'm also using Ory Kratos in this project, which is also < v1. Resources to handle this are causing my "road block".

I really appreciate you folks listening! ☮️

benmccann:
Yes, I think the recommendation to upgrade gradually is a good one. 60 versions at once will be harder

Also, in case it helps, we have several examples: https://kit.svelte.dev/docs#additional-resources-examples. The `kit.svelte.dev` code is always on the latest versions. The others might lag slightly, but should all be on relatively recent versions

QubDigitalNV:
This change completely breaks SvelteKit and makes adapter-static unusable. No method of debugging has eliminated this error, which is caused by removing 'target' from the config:

Error: request in handle has been replaced with event. See https://github.com/sveltejs/kit/pull/3384 for details
at Object.get request [as request] (file:///C:/Apps/qub/.svelte-kit/output/server/app.js:1832:15)
at Object.handle (file:///C:/Apps/qub/.svelte-kit/output/server/app.js:1869:25)
at respond (file:///C:/Apps/qub/.svelte-kit/output/server/app.js:1737:42)
at App.render (file:///C:/Apps/qub/.svelte-kit/output/server/app.js:1954:12)
at visit (file:///C:/Apps/qub/node_modules/@sveltejs/kit/dist/chunks/index5.js:536:30)
at Object.fn (file:///C:/Apps/qub/node_modules/@sveltejs/kit/dist/chunks/index5.js:520:22)
at dequeue (file:///C:/Apps/qub/node_modules/@sveltejs/kit/dist/chunks/index5.js:95:42)
at Object.add (file:///C:/Apps/qub/node_modules/@sveltejs/kit/dist/chunks/index5.js:122:4)
at enqueue (file:///C:/Apps/qub/node_modules/@sveltejs/kit/dist/chunks/index5.js:520:12)
at prerender (file:///C:/Apps/qub/node_modules/@sveltejs/kit/dist/chunks/index5.js:645:6)

lkj4:
> The current version released is [1.0.0-next.260](https://github.com/sveltejs/kit/blob/master/packages/kit/CHANGELOG.md#100-next260), in which you can type [`App.Locals`](https://kit.svelte.dev/docs#typescript-app-locals).

I did now and it didn't work and `App.Locals` seems to be ignored.

> In that version, you are able to type Locals in endpoints using the RequestHandler.

Didn't work either and leads to an `Hint: 'Locals' is declared but its value is never read.`. `locals` is read in the request handler though.

(this is not a support request, it's about this PR which might need to be tested and improved further)

hyunbinseo:
@lkj4 Assuming that you are using a version later than [`1.0.0-next.257`](https://github.com/sveltejs/kit/blob/master/packages/kit/CHANGELOG.md#100-next257), try the following

(since you have confirmed that `App.Locals` is not working in your case)

---

- Create a `src/app.d.ts` file with the following content

```Typescript

// See https://kit.svelte.dev/docs/typescript
// for information about these interfaces
declare namespace App {
	interface Locals {
    sample: string;
  }
}
```

- Access `locals.sample` in a function in an `endpoint`.

```Typescript
import type { RequestHandler } from '@sveltejs/kit';

export const post: RequestHandler = async(event) => {
  event.locals.sample;

  return {
    status: 200
  };
}
```

---

Would be helpful if you specify the `@sveltejs/kit` version that you are using.

lkj4:
I use 260 and did all of that above. I also just tried to create a mini repo with the demo app which uses 269 where typing via App.Locals works in my editor (nvim). Strange...

lkj4:
Ok, found the problem, forget to type explicitly the request handler with `RequestHandler` and same type `Handle` in `hooks.js`, so my bad, pls ignore my previous posts

WeAreELIC:
So, is it possible to get the rawBody data now? I'm using Stripe's API and need the raw request body from an endpoint's post function. Help!

j2l:
I was lead here by a svelte error when trying to update packages of https://github.com/timsonner/sveltekit-system-call
Noob question: how to adapt the [module](https://github.com/timsonner/sveltekit-system-call/blob/main/src/routes/api/index.json.ts)?
Thank you svelte gurus!

benwoodward:
I'm also struggling to figure out how to get the `rawBody` value, I need it for a mux.com webhook

KTruong008:
> I'm also struggling to figure out how to get the `rawBody` value, I need it for a mux.com webhook

I'm not sure about mux.com but I was able to get the rawBody for Stripe's checkout webhook with this:

```
function toBuffer(ab: any) {
  const buf = Buffer.alloc(ab.byteLength);
  const view = new Uint8Array(ab);
  for (let i = 0; i < buf.length; ++i) {
    buf[i] = view[i];
  }
  return buf;
}

export async function post(event: RequestEvent<Record<string, string>>) {
   ...
  const preRawBody = await event.request.arrayBuffer();
  const rawBody = toBuffer(preRawBody);
  
  try {
    stripeEvent = stripe.webhooks.constructEvent(
      rawBody,
      stripeSignature,
      process.env.STRIPE_WEBHOOK_SECRET
    );
  } ...
 }
```

benwoodward:
Thanks @KTruong008, this is how I fixed my code (you may find using `.buffer()` simpler)

```js
export async function post({ request }) {
  const rawBody = await request.buffer();
  const body = JSON.parse(rawBody.toString());
  const signature = request.headers.get('mux-signature') || '';
  const eventData = body.data;

  let isValidSignature = false;

  try {
    isValidSignature = Webhooks.verifyHeader(
      rawBody,
      signature,
      MUX_WEBHOOK_SECRET
    );
```

hyunduk0206:
I need help finding right replacement.

Original `took.ts` was:
```ts

export const handle: Handle = async ({ request, resolve }) => {
	if (request.url.searchParams.has('_method')) {
		request.method = request.url.searchParams.get('_method').toUpperCase();
	}

	return response;
};
```
=> `request in handle has been replaced with event. See https://github.com/sveltejs/kit/pull/3384 for details`

So I changed
```ts

	if (event.url.searchParams.has('_method')) {
		event.request.method = event.url.searchParams.get('_method').toUpperCase();
	}

	return response;
};
```

But it is not working giving this error
```
event.method has been replaced by event.request.method. See https://github.com/sveltejs/kit/pull/3384 for details
Error: event.method has been replaced by event.request.method. See https://github.com/sveltejs/kit/pull/3384 for details
    at Object.get (file:///workspace/voicebot/.svelte-kit/runtime/server/index.js:2532:10)
    at Module.api (/workspace/voicebot/src/routes/services/todo/todos/_api.ts:10:24)
    at get (/workspace/voicebot/src/routes/services/todo/todos/index.json.ts:5:12)
    at render_endpoint (file:///workspace/voicebot/.svelte-kit/runtime/server/index.js:150:25)
    at resolve (file:///workspace/voicebot/.svelte-kit/runtime/server/index.js:2656:17)
    at async Object.handle (/workspace/voicebot/src/hooks.ts:8:24)
    at async respond (file:///workspace/voicebot/.svelte-kit/runtime/server/index.js:2564:20)
    at async fetch (file:///workspace/voicebot/.svelte-kit/runtime/server/index.js:1697:17)
    at async load (index.svelte:5:20)
Unexpected token < in JSON at position 0
SyntaxError: Unexpected token < in JSON at position 0
    at JSON.parse (<anonymous>)
    at Proxy.<anonymous> (file:///workspace/voicebot/.svelte-kit/runtime/server/index.js:1797:21)
    at async load (index.svelte:12:28)
    at async load_node (file:///workspace/voicebot/.svelte-kit/runtime/server/index.js:1826:12)
    at async respond$1 (file:///workspace/voicebot/.svelte-kit/runtime/server/index.js:2191:15)
    at async render_page (file:///workspace/voicebot/.svelte-kit/runtime/server/index.js:2392:19)
    at async resolve (file:///workspace/voicebot/.svelte-kit/runtime/server/index.js:2657:11)
    at async Object.handle (/workspace/voicebot/src/hooks.ts:8:24)
    at async respond (file:///workspace/voicebot/.svelte-kit/runtime/server/index.js:2564:20)
```

Would you please tell me how to fix the problem?
(source code: https://github.com/hyunduk0206/voicebot/blob/main/src/hooks.ts)

benwoodward:
@hyunduk0206 I recommend asking this kind of question on StackOverflow or the svelte discord. However, I would just console.log out the values to see what's contained within the parameters.

hyunduk0206:
@benwoodward Thank you for your reply. I will not ask this kind of question here, except for this question.

I have tried to print the values, but I couldn't due to error. But I can write down the part of code that is related to the value.

```ts
import type { Request } from '@sveltejs/kit';
import PrismaClient from '$lib/prisma';

const prisma = new PrismaClient();

export const api = async (request: Request, data?: Record<string, unknown>) => {
	let status = 500;
	let body = {};

	switch (request.method.toUpperCase()) {
		case 'GET':
			status = 200;
			body = await prisma.todo.findMany();
			break;
		case 'POST':
			status = 201;
			body = await prisma.todo.create({
				data: {
					created_at: data.created_at as Date,
					done: data.done as boolean,
					text: data.text as string
				}
			});
			break;
		case 'DELETE':
			status = 200;
			body = await prisma.todo.delete({
				where: {
					uid: request.params.uid
				}
			});
			break;
		case 'PATCH':
			status = 200;
			body = await prisma.todo.update({
				where: {
					uid: request.params.uid
				},
				data: {
					done: data.done,
					text: data.text
				}
			});
			break;

		default:
			break;
	}

	if (request.method.toUpperCase() !== 'GET' && request.headers.accept !== 'application/json') {
		return {
			status: 303,
			headers: {
				location: '/'
			}
		};
	}

	return {
		status,
		body
	};
};

```

hyunduk0206:
I solved this problem by following the offical doc (https://kit.svelte.dev/docs/routing#endpoints-http-method-overrides).

nmfrankel:
Adding the info of how to read request body data to [kit.svelte.com](kit.svelte.com) would be quite useful, it took me over a day to find this thread.

csjh:
> Adding the info of how to read request body data to [kit.svelte.com](kit.svelte.com) would be quite useful, it took me over a day to find this thread.

https://kit.svelte.dev/docs/web-standards#fetch-apis-request ?

nmfrankel:
Yes, my app only has ``+server.js`` routes for the db access and I couldn't find it anywhere when including +server.js in the search. There's documentation and an example for [``url.params``](https://kit.svelte.dev/docs/routing#server), but I think ``request.body`` is at least as important to require an example.

I found the answer from this thread, Svelte gave the link as a console error, but all the attempts with ``request.formData()`` (as suggested on StackOverflow), only gave the error without the link.

xpluscal:
I’m currently getting “Invalid Request Body” when deployed and uploading a file using a Named Action through a form. Any ideas why?

xpluscal:
> kit.svelte.com

Hey does this solution still work for you? I'm having issues with the arraybuffer. It seems that you can get the body via `response.text()` but the stripe verification still fails.

Artur-Galstyan:
Stripe verification doesn't work with `let requestBody = await request.test()`.

It seems that it is simply not possible to use Stripe webhooks with Sveltekit :(

Issue #3189, opened 2022-01-03
bug(netlify): Package subpath './hooks' is not defined by "exports"


Mrbeyond:
### Describe the bug

Sveltekit app having version "svelte": "^3.44.0" works perfectly locally but showing "errorMessage":"Package subpath './ssr' is not defined by \"exports\" in /var/task/node_modules/@sveltejs/kit/package.json" when deployed to production on netlify. I have tried both auto-adapter and explicit netlify adapter but the error persists.

### Reproduction

Repo link:
https://github.com/Mrbeyond/pubside

### Logs

```shell
{"errorType":"Error","errorMessage":"Package subpath './ssr' is not defined by \"exports\" in /var/task/node_modules/@sveltejs/kit/package.json","trace":["Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './ssr' is not defined by \"exports\" in /var/task/node_modules/@sveltejs/kit/package.json","    at new NodeError (internal/errors.js:322:7)","    at throwExportsNotFound (internal/modules/esm/resolve.js:322:9)","    at packageExportsResolve (internal/modules/esm/resolve.js:511:7)","    at resolveExports (internal/modules/cjs/loader.js:450:36)","    at Function.Module._findPath (internal/modules/cjs/loader.js:490:31)","    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:888:27)","    at Function.Module._load (internal/modules/cjs/loader.js:746:27)","    at Module.require (internal/modules/cjs/loader.js:974:19)","    at require (internal/modules/cjs/helpers.js:93:18)","    at Object.<anonymous> (/var/task/.netlify/server/app.js:45:29)"]}
```

### System Info

```shell
"devDependencies": {
    "@sveltejs/adapter-auto": "next",
    "@sveltejs/adapter-netlify": "^1.0.0-next.37",
    "@sveltejs/kit": "next",
    "autoprefixer": "^10.4.1",
    "cssnano": "^5.0.14",
    "eslint": "^7.32.0",
    "eslint-config-prettier": "^8.3.0",
    "eslint-plugin-svelte3": "^3.2.1",
    "node-sass": "^7.0.1",
    "postcss-load-config": "^3.1.0",
    "prettier": "^2.4.1",
    "prettier-plugin-svelte": "^2.4.0",
    "svelte": "^3.44.0",
    "svelte-preprocess": "^4.10.1",
    "tailwindcss": "^3.0.8"
  },
```

[...]

PR #3389, opened 2022-01-18
[fix] export CJS files for Netlify

benmccann:
Hopefully fixes https://github.com/sveltejs/kit/issues/3189. I'm not really sure how to test this. And my app works on Netlify anyway somehow. It should get us closer at least though. The thing I'm unsure of is if the `assets` directory also needs to be bundled as CJS

Hopefully we won't need this forever, but it seems like Netlify is still a ways off from supporting CJS. Work is in progress in https://github.com/netlify/cli/issues/3514, but there's more work that needs to be done after that

### DIFF
--- a/.changeset/dirty-walls-battle.md
+++ b/.changeset/dirty-walls-battle.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] export CJS files for Netlify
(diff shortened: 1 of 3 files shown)


Rich-Harris:
This shouldn't be necessary. I think they just have out of date versions

PR #3395, opened 2022-01-18
document serviceWorker register option

ignatiusmb:

### Tests

### Changesets

### DIFF
--- a/documentation/docs/14-configuration.md
+++ b/documentation/docs/14-configuration.md
@@ -232,6 +232,7 @@ Enables or disables the client-side [router](#page-options-router) app-wide.
 
 An object containing zero or more of the following values:
 
+- `register` - set to `false` to disable automatic service worker registration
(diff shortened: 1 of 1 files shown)


documentation/docs/14-configuration.md
```diff
@@ -232,6 +232,7 @@ Enables or disables the client-side [router](#page-options-router) app-wide.
 
 An object containing zero or more of the following values:
 
+- `register` - set to `false` to disable automatic service worker registration
```
benmccann, line 235:
```suggestion
- `register` - whether to enable automatic service worker registration. Default is `true`
```

ignatiusmb: I'm following a similar wording style to the `prerender.enabled` config, and we don't really mention the default value for the config anywhere since it's already displayed in the code block at the start, so I don't think the last part is necessary

Rich-Harris, line 235: we could do it like this — consistent with others, but unambiguous

```suggestion
- `register` - if set to `false`, will disable automatic service worker registration
```

Rich-Harris (approved):

PR #3396, opened 2022-01-18
[fix] avoid downloading CSS with highest priority when `inlineStyleThreshold` is used

### LINKED ISSUE #3307: disabled on new inlineStyleThreshold still causes stylesheet to download
### Describe the problem

Love the new CSS in-lining feature, one smallish thing, it leaves the `<link>` tag in place and adds a `disabled` attribute, i.e.
```
<link disabled rel="stylesheet" href="/_app/assets/pages/__layout.svelte-27e15f73.css">
```
This means the browser still downloads it (with high priority being a stylesheet)

Perhaps adding a media query along with the disabled attribute? This would prevent the browser from downloading the file.

Something like

```
<link disabled media="screen and (max-width: 1px)" rel="stylesheet" href="/_app/assets/pages/__layout.svelte-27e15f73.css">
```

Would achieve this.

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

dwsmart:
Fixes #3307 
`<link disabled rel="stylesheet" href="...">` still causes the browser to download the file, with the highest priority, even though it's never used.

This PR adds a media query with a 1px max screen width, this prevents the browser from downloading the CSS, but leaves the link tag in place so vite functions correctly.

`<link disabled media="screen and (max-width: 1px)" rel="stylesheet" href="...">`

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -148,8 +148,8 @@ export async function render_response({
 		}
 		// prettier-ignore
 		head += Array.from(css)
-            .map((dep) => `\n\t<link${styles.has(dep) ? ' media="print"' : ''} rel="stylesheet" href="${options.prefix + dep}">`)
-            .join('');
+			.map((dep) => `\n\t<link${styles.has(dep) ? ' disabled media="(max-width: 0)"' : ''} rel="stylesheet" href="${options.prefix + dep}">`)
+			.join('');
 
 		if (page_config.router || page_config.hydrate) {
 			head += Array.from(js)
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -148,8 +148,8 @@ export async function render_response({
 		}
 		// prettier-ignore
 		head += Array.from(css)
-				.map((dep) => `\n\t<link${styles.has(dep) ? ' disabled' : ''} rel="stylesheet" href="${options.prefix + dep}">`)
-				.join('');
+            .map((dep) => `\n\t<link${styles.has(dep) ? ' disabled media="screen and (max-width: 1px)"' : ''} rel="stylesheet" href="${options.prefix + dep}">`)
+            .join('');
 
 		if (page_config.router || page_config.hydrate) {
 			head += Array.from(js)
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -24,139 +24,139 @@ import { create_prerendering_url_proxy } from './utils.js';
  * }} opts
  */
 export async function render_response({
-	branch,
-	options,
-	state,
-	$session,
-	page_config,
-	status,
-	error,
-	url,
-	params,
-	ssr,
-	stuff
+    branch,
+    options,
+    state,
+    $session,
+    page_config,
+    status,
+    error,
+    url,
+    params,
+    ssr,
+    stuff
 }) {
-	const css = new Set(options.manifest._.entry.css);
-	const js = new Set(options.manifest._.entry.js);
-	/** @type {Map<string, string>} */
-	const styles = new Map();
+    const css = new Set(options.manifest._.entry.css);
+    const js = new Set(options.manifest._.entry.js);
+    /** @type {Map<string, string>} */
+    const styles = new Map();
 
-	/** @type {Array<{ url: string, body: string, json: string }>} */
-	const serialized_data = [];
+    /** @type {Array<{ url: string, body: string, json: string }>} */
+    const serialized_data = [];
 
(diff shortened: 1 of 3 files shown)


packages/kit/src/runtime/server/page/render.js
```diff
@@ -148,8 +148,8 @@ export async function render_response({
 		}
 		// prettier-ignore
 		head += Array.from(css)
-				.map((dep) => `\n\t<link${styles.has(dep) ? ' disabled' : ''} rel="stylesheet" href="${options.prefix + dep}">`)
-				.join('');
+            .map((dep) => `\n\t<link${styles.has(dep) ? ' disabled media="screen and (max-width: 1px)"' : ''} rel="stylesheet" href="${options.prefix + dep}">`)
```
Rich-Harris, line 151: I wonder if we can make this even simpler — any reason we can't change `media` to `none`? That way we can get rid of the `disabled` attribute too. This works for me locally, what do you reckon?

```suggestion
            .map((dep) => `\n\t<link${styles.has(dep) ? ' media="none"' : ''} rel="stylesheet" href="${options.prefix + dep}">`)
```

Rich-Harris (changes requested):
Thank you! Mind adding a changeset so this triggers a release?

Also, would be great to have a test for this — you could add one after this test, perhaps: https://github.com/sveltejs/kit/blob/49114b9a37e93335b027865e68fa4d58a0a7a2e5/packages/kit/test/apps/options/test/test.js#L45-L69

Example of capturing which network requests were made: https://github.com/sveltejs/kit/blob/49114b9a37e93335b027865e68fa4d58a0a7a2e5/packages/kit/test/apps/basics/test/test.js#L776-L778

packages/kit/src/runtime/server/page/render.js
dwsmart: Will add the changeset and tests later, sorry!

I mistakenly believed that disabled needed to be in there for Vite, don't know where I got that from!

The `media="none"` thing is a pedantic point, strictly speaking [there is no `none` value](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-media) (only `all`, `screen` and `print`), so it's relying on browsers to interpret a value they don't understand as 'don't download this'. so `screen and (max-width: 1px)` is actually "valid" and actively instructing according to spec.

That being said, I've not met a browser yet that wouldn't work as intended with `media="none"`, and I do think it's more intuitive and readable, so perhaps we should just use that?

An alternative would be `media="print"` which would prevent the browser from downloading (unless the user printed the page!), but still be spec compliant?

Rich-Harris: Fair point — I'm not much of a purist but I did wonder about that, and it would be nice to generate HTML that validates.

`print` seems good — no real harm if the user _does_ print and download the styles. `media="(max-width: 0)"` would also work I think, but `print` is shortest.

dwsmart: OK made the changes to `print`, and amended the test, however, it seems like `print` (and any non-matching media value) do still get downloaded. Just with a lowest network priority.

That's still an improvement in my book, but I retitled this PR to more accurately reflect that.

packages/kit/src/runtime/server/page/render.js
```diff
@@ -148,8 +148,8 @@ export async function render_response({
 		}
 		// prettier-ignore
 		head += Array.from(css)
-				.map((dep) => `\n\t<link${styles.has(dep) ? ' disabled' : ''} rel="stylesheet" href="${options.prefix + dep}">`)
-				.join('');
+            .map((dep) => `\n\t<link${styles.has(dep) ? ' media="print"' : ''} rel="stylesheet" href="${options.prefix + dep}">`)
+            .join('');
```
Rich-Harris, line 152: I did some testing in different browsers. Turns out `media="print"` or `media="(max-width: 0)"` doesn't disable downloading, as you noted. `disabled` _does_ disable downloading in Chrome and Safari, but not Firefox. But combining them — `disable media="..."` — disables downloading in all three.

On reflection I think `(max-width: 0)` is preferable to `print`, since it kind of says 'hello! i am a deliberate hack' to anyone who happens to look at the HTML

```suggestion
			.map((dep) => `\n\t<link${styles.has(dep) ? ' disabled media="(max-width: 0)"' : ''} rel="stylesheet" href="${options.prefix + dep}">`)
			.join('');
```

Rich-Harris (approved):

Rich-Harris:
After all that back-and-forth it turns out we need to keep `disabled` in there! I've updated the PR, per the comment above — basically landed on exactly what you said in #3307 😅 

Thanks for the help/patience

dwsmart:
Thank you for your patience and help with the PR, and thanks to y'all for SveleKit.

PR #3404, opened 2022-01-19
gh-3403: Remove encodeURI call for redirect load property

### LINKED ISSUE #3403: Document and use consistent encoding expectations for redirect output property
### Describe the problem

The current documentation for the [`redirect`](https://kit.svelte.dev/docs#loading-output-redirect) output property of the loading function just says it should be:

>  a string containing the location to which they should be redirected

From what I can tell, this string is being treated with different expectations around encoding status depending on the current runtime location:

* The [server](https://github.com/sveltejs/kit/blob/0eb435899c5bd0db87adff971db6fe99b448a4b8/packages/kit/src/runtime/server/page/respond.js#L115-L125) will pass the `redirect` string into [`encodeURI(redirect)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI)
* The [client](https://github.com/sveltejs/kit/blob/21151556cf7091ec84cd9d6820aa7e2afd58cdef/packages/kit/src/runtime/client/renderer.js#L259-L279) passes the `redirect` string into [`new URL(redirect, current_url)`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL)

If the expectation is that the developer should return an already properly encoded URI string, then the server path results in double encoding.

If the expectation is that the developer should be returning an unencoded string, the client path is relying on the URL constructor to tolerate unencoded URI strings.  For the most part this works because the javascript engine bends over backwards, but can result in unintended results.  

Example:

* Say I have a `/login` page that accepts a `previous` query parameter[1] that should contain the path of where to send the user after login.  
* Pages behind a login wall that need to redirect unauthenticated users to the `/login` page with an appropriate `previous` query parameter.
* A user visits `/app?foo=bar&fizz=buzz` while unauthenticated, and the login wall wants to send them to the login page

If the login-wall treats the `redirect` string as **pre-encoded**:
```
return {
  status: 302,
  redirect: '/login?previous=' + encodeURIComponenent(request.url.pathname + request.url.search)
}
```
Server side rendering will use a Location header of
```
encodeURI('/login?previous=' + encodeURIComponent('/app?foo=bar&fizz=buzz'))
> "/login?previous=%252Fapp%253Ffoo%253Dbar%2526fizz%253Dbuzz"  // double encoded query parameter

```
And the client will use
```
new URL('/login?previous=' + encodeURIComponent('/app?foo=bar&fizz=buzz'), 'http://example.com')
> URL { href: "http://example.com/login?previous=%2Fapp%3Ffoo%3Dbar%26fizz%3Dbuzz", origin: "http://example.com", protocol: "http:", username: "", password: "", host: "example.com", hostname: "example.com", port: "", pathname: "/login", search: "?previous=%2Fapp%3Ffoo%3Dbar%26fizz%3Dbuzz" } // single encoded query parameters.
```

If the login-wall treats the `redirect` string as **unencoded**:
```
return {
  status: 302,
  redirect: '/login?previous=' + request.url.pathname + request.url.search
}
```
Server side rendering will use a Location header of
```
encodeURI('/login?previous=/app?foo=bar&fizz=buzz')
> "/login?previous=/app?foo=bar&fizz=buzz"  // completely unencoded query parameter

```
And the client will use
```
new URL('/login?previous=/app?foo=bar&fizz=buzz', 'http://example.com')
> URL { href: "http://example.com/login?previous=/app?foo=bar&fizz=buzz", origin: "http://example.com", protocol: "http:", username: "", password: "", host: "example.com", hostname: "example.com", port: "", pathname: "/login", search: "?previous=/app?foo=bar&fizz=buzz" }} // completely unencoded query parameters.

// Especially problematic because we effectively lose the `fizz` query param on the `previous` url:
$_.searchParams.get("previous")
> "/app?foo=bar"
```

[1] Yes, I'm familiar with the [problems of open redirects](https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html).  This is a simplified example.

The `redirect` property should be treated as already encoded, and the documentation should reflect that.

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

moatra:
Fixes #3403 -- developers can now return a `redirect` property from a module's `load()` without it be double-encoded during SSR.

### Tests
- [/] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`
    - There seems to be a flaky test about a scroll bar.

### Changesets

### DIFF
--- a/packages/kit/test/apps/basics/test/test.js
+++ b/packages/kit/test/apps/basics/test/test.js
@@ -472,6 +472,22 @@ test.describe.parallel('Encoded paths', () => {
 		expect(decodeURI(await page.innerHTML('h3'))).toBe('/encoded/苗条');
 	});
 
+	test('redirects do not re-encode the redirect string', async ({ page, clicknav }) => {
+		await page.goto('/encoded');
+
+		await clicknav('[href="/encoded/redirect"]');
+
+		// check innerText instead of innerHTML because innerHTML would return the '&amp;' character reference instead of '&' character.
+		expect(await page.innerText('pre')).toBe('/苗条?foo=bar&fizz=buzz');
+	});
+
+	test('redirects do not re-encode the redirect string during ssr', async ({ page }) => {
+		await page.goto('/encoded/redirect');
+
+		// check innerText instead of innerHTML because innerHTML would return the '&amp;' character reference instead of '&' character.
+		expect(await page.innerText('pre')).toBe('/苗条?foo=bar&fizz=buzz');
+	});
+
 	test('sets charset on JSON Content-Type', async ({ request }) => {
 		const response = await request.get('/encoded/endpoint');
 		expect(response.headers()['content-type']).toBe('application/json; charset=utf-8');
--- a/.changeset/early-clocks-sneeze.md
+++ b/.changeset/early-clocks-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+The redirect property returned from a module's load function must now be a properly encoded URI string value.
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -143,6 +143,8 @@ If something goes wrong during `load`, return an `Error` object or a `string` de
 
 If the page should redirect (because the page is deprecated, or the user needs to be logged in, or whatever else) return a `string` containing the location to which they should be redirected alongside a `3xx` status code.
 
+The `redirect` string should be a [properly encoded](https://en.wikipedia.org/wiki/Percent-encoding) URI.  Both absolute and relative URIs are acceptable.
+
 #### maxage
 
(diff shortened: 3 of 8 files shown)


packages/kit/test/apps/basics/test/test.js
```diff
@@ -472,6 +472,22 @@ test.describe.parallel('Encoded paths', () => {
...
+		await page.goto('/encoded');
+
+		await clicknav('[href="/encoded/redirect"]');
+
+		// check innerText instead of innerHTML because innerHTML would return the '&amp;' character reference instead of '&' character.
+		expect(await page.innerText('pre')).toBe('/苗条?foo=bar&fizz=buzz');
+	});
+
+	test('redirects do not re-encode the redirect string during ssr', async ({ page }) => {
+		await page.goto('/encoded/redirect');
```
moatra, line 485: I don't have experience with playwright, but I made the assumption `goto` does a full page load that triggers SSR.

Rich-Harris (approved):

packages/kit/test/apps/basics/test/test.js
Rich-Harris: you assume correctly :)

Rich-Harris:
thank you!

Issue #87, opened 2020-10-29
Handle long sessions


Rich-Harris:
In [Second-guessing the modern web](https://macwright.com/2020/05/10/spa-fatigue.html), Tom MacWright makes several good points (I responded [here](https://dev.to/richharris/in-defense-of-the-modern-web-2nia)).

The biggest, to me, is this one:

> User sessions are surprisingly long: someone might have your website open in a tab for weeks at a time. I’ve seen it happen. So if they open the ‘about page’, keep the tab open for a week, and then request the ‘home page’, then the home page that they request is dictated by *the index bundle that they downloaded last week*. This is a deeply weird and under-discussed situation.

I wrote:

> It's an excellent point that isn't really being addressed, though (as Tom acknowledges) it's really just exacerbating a problem that was always there. I think there are solutions to it — we can iterate on the 'index bundle' approach, we could include the site version in a cookie and use that to show actionable feedback if there's a mismatch — but we do need to spend time on it.

It would be great if there was some mechanism for solving (or at least alleviating) this problem. I think the 'include the site version in a cookie' approach is the most promising, but I haven't thought through any of the details.

benmccann:
For posterity, here was the corresponding Sapper issue: https://github.com/sveltejs/sapper/issues/389

oodavid:
I could see users falling into various camps as to how to approach this issue.

For us, we have an "your app is out of date" message. Which is great for "dashboard" type sites, but not so good for blogs etc.

<img width="591" alt="Screenshot 2021-03-30 at 09 48 26" src="https://user-images.githubusercontent.com/46879/112961329-2e415c00-913d-11eb-9566-b5e8df30ebdf.png">

For Svelte, I've seen this handled with a `navigator.serviceWorker` `statechange` event. Example code:

https://github.com/tobinbradley/Mecklenburg-County-GeoPortal/blob/c22980c0915e48be18e292cb5f4d9cc221e41a09/src/js/registerServiceWorker.js

I'd love to see this approach well documented in the `$service-worker` module.

My biggest concern with this issue isn't the the inner-workings, but rather the learning-curve. Most users will have to experience the issue before they start to look for a solution. For many it will be a heisenbug. I'm tempted to request that the default behaviour should be a _little_ intrusive.

[...]

PR #3412, opened 2022-01-19
[feat] application versioning and update detection support

pz-mxu:
This PR will offer two options for navigation errors in the router. These errors occur for example if a new version of the site is deployed while being open in a browser. The JS files for pages might have changed and so has the filename. The previous page JS files, that the router still knows about, will give an import error 500 when trying to navigate to that page. For more detail on the problem, see #87.

The `kit.browser.router` configuration option in svelte.conf.js is now an object with `enabled` and `onError` properties.

`enabled` is just a boolean that replaces the old `kit.browser.router` boolean option.

`onError` has two possible values

- "fail": will result in exactly the same behaviour as now - import error 500
- "reload": location.href will be set to force a reload when navigation failed and the app version changed

The "reload" option should be enough for most users with the problem described. "fail" is currently the default option.

The mentioned app version is an automatically generated timestamp, that can be accessed with 

`import { version} from "$app/env"`

and can also be overridden by setting the environment variable `VITE_APP_VERSION` to a custom version string.

There is also an `updated` store that can be imported with

`import { updated } from "$app/stores"`.

The app version will be checked periodically in the background by fetching "_app/version.json" every 5 minutes. The store is set to `true` when the fetched version differs from the running version. The store has also a `check` function, that will immediately perform the check and return the result as a promise.

### Tests

### Changesets


benmccann:
> "redirect": location.href will be set to force a reload when navigation failed

I didn't look at the code, but just from the description, it sounds like "reload" would be a better name. That being said, I think that would rarely be the best behavior and not something I'd want to encourage, so I'd be much happier to remove this option and have the user implement it themselves (e.g. via the `function`)

pz-mxu:
> That being said, I think that would rarely be the best behavior and not something I'd want to encourage, so I'd be much happier to remove this option and have the user implement it themselves (e.g. via the `function`)

The implementation is "reload" is trivial, so that's fine for me.

Rich-Harris:
Thank you! Personally I disagree about reloading not being the best option — I think it's _exactly_ what you want to happen if you attempt to click a link (or `goto`) and the module can't be imported (unless it's because you're offline, in which case rendering the error page is appropriate). I'm not sure what other behaviour you'd want to implement?

The trouble, of course, is that the error might be happening for some other reason than a new deploy — perhaps the imported module immediately throws an error in the browser, or something. If we reload indiscriminately then we miss the opportunity to discover those cases. Instead, we need a cohesive idea of an app 'version' (that can be used to disable client-side routing pre-emptively, or show a toast, or assess whether API requests are up to date, etc) that is used to determine whether reloading is appropriate.

Stringifying functions is a no-go, by the way — it gets horribly confusing when suddenly closures don't work, etc. If we need that level of flexibility (I'm not sure we do, yet) then we need client-side hooks.

pz-mxu:
> ...we need a cohesive idea of an app 'version' (that can be used to disable client-side routing pre-emptively, or show a toast, or assess whether API requests are up to date, etc) that is used to determine whether reloading is appropriate.
I added an app version. The reload will only happen now on error when the version changed.

> Stringifying functions is a no-go, by the way — it gets horribly confusing when suddenly closures don't work, etc. If we need that level of flexibility (I'm not sure we do, yet) then we need client-side hooks.
Understood, so for now I removed the custom function option.

packages/kit/src/core/build/build_client.js
```diff
@@ -29,6 +29,8 @@ export async function build_client({
 	output_dir,
 	client_entry_file
 }) {
+	if (!process.env.VITE_APP_VERSION) process.env.VITE_APP_VERSION = new Date().toISOString();
```
Rich-Harris, line 32: We need a way to specify the version string via config. It should fall back to a timestamp (`Date.now()` is fine, I think — no need for an ISO string) but I've seen cases where the same app version is built multiple times, and the builds disagree. In those cases you need to be able to specify a commit hash or a `package.json` version or whatever.

packages/kit/src/core/build/build_client.js
```diff
@@ -105,6 +107,12 @@ export async function build_client({
 	const entry_css = new Set();
 	find_deps(entry, vite_manifest, entry_js, entry_css);
 
+	fs.writeFileSync(
+		`${client_out_dir}/version.json`,
+		JSON.stringify(process.env.VITE_APP_VERSION),
```
Rich-Harris, line 112: To make this future-proof it should be an object:

```suggestion
		JSON.stringify({ version: process.env.VITE_APP_VERSION })
```

packages/kit/src/core/build/build_client.js
```diff
@@ -105,6 +107,12 @@ export async function build_client({
...
+		'utf-8'
```
Rich-Harris, line 113: No need to specify encoding with `writeFileSync`
```suggestion
```

packages/kit/src/runtime/client/renderer.js
```diff
@@ -108,7 +147,17 @@ export class Renderer {
...
+			session: writable(session),
+			updated: checkable_store(
+				false,
+				(set) => {
+					return has_version_changed().then((changed) => {
+						if (changed) set(true);
+						return changed;
+					});
+				},
+				300000
```
Rich-Harris, line 159: This will need to be configurable

packages/kit/types/internal.d.ts
```diff
@@ -5,6 +5,9 @@ import { Fallthrough, RequestHandler } from './endpoint';
 import { Either } from './helper';
 import { ExternalFetch, GetSession, Handle, HandleError, RequestEvent } from './hooks';
 import { Load } from './page';
+import { RouteOnErrorValue } from './config';
+
+type PageId = string;
```
Rich-Harris, line 10: guessing this is left over from something?

packages/kit/src/runtime/client/renderer.js
```diff
@@ -39,6 +40,44 @@ function notifiable_store(value) {
...
+ */
+async function has_version_changed() {
+	if (import.meta.env.DEV) return false;
+
+	const headers = new Headers();
+	headers.append('pragma', 'no-cache');
+	headers.append('cache-control', 'no-cache');
+
+	const current_version = import.meta.env.VITE_APP_VERSION;
+	const new_version = await fetch('_app/version.json', { headers })
```
Rich-Harris, line 74: This needs to use `config.kit.appDir` somehow, it can't be hardcoded to `_app`.

Also, no need to create a `Headers` object, you can just do this:

```js
const new_version - await fetch(`${base}/${app_dir}/version.json`, {
  headers: {
    'pragma': 'no-cache',
    'cache-control': 'no-cache'
  }
})
```

We shouldn't be swallowing the error here — at the very least an explicit call to `await updated.check()` should throw if the fetch fails.

Rich-Harris (changes requested):
Thank you. Thinking on this further, I'm reaching the conclusion that there's no reason for any behaviour other than automatic reloading. This is an area where the framework needs to be a little bit opinionated and do the right thing, since most developers don't really think about long session handling until they get bitten. Since client-side routing is best thought of as a progressive enhancement, falling back to a full page refresh should be perfectly acceptable in a well-written app.

To that end, I think we should get rid of the `onError` option (though if there's a compelling reason not to, I would be interested to hear it).

As mentioned inline, we need to be able to configure both the version string and the polling interval. I propose the following:

```js
// svelte.config.js
export default {
  kit: {
    version: {
      name: Date.now(), // default to timestamp
      pollInterval: 60_000 // default to checking once a minute
    }
  }
};
```

packages/kit/src/core/build/build_client.js
pz-mxu: I thought that using VITE_APP_VERSION environment variable makes using a commit hash already very easy, but I agree that having all in one place in the config is also good. I propose that we set VITE_APP_VERSION to the value configured in the config file, but if there is nothing, it can also be set via environment. the fallback to `Date.now()`is later anyway.

packages/kit/src/runtime/client/renderer.js
pz-mxu: I added that catch because of error messages in the console in dev. Since dev is handled differently now, this is not necessary anymore.

pz-mxu:
> Thank you. Thinking on this further, I'm reaching the conclusion that there's no reason for any behaviour other than automatic reloading. This is an area where the framework needs to be a little bit opinionated and do the right thing, since most developers don't really think about long session handling until they get bitten. Since client-side routing is best thought of as a progressive enhancement, falling back to a full page refresh should be perfectly acceptable in a well-written app.
> 
> To that end, I think we should get rid of the `onError` option (though if there's a compelling reason not to, I would be interested to hear it).

I totally agree, but I read the discussion on #87 and many people requested more configurability. For me personally, the reload is all I need and using SvelteKit in 4 projects already, that solution has helped with the problems after deploy. I don't know how the final decision on this is made and if I should remove the option right away or wait.

pz-mxu:
> I didn't see any real arguments in favour of configurability in #87, just people working around the problem by making static assets outlive the deployment, which is a partial solution at best (since it ensures users will see stale content, and requires API compatibility between versions).
> 
> My inclination would be to _not_ make it configurable, and see if that causes problems. We can always add configuration in future, but removing unnecessary or problematic configuration is extremely hard.

Ok, I am nearly done with the changes - configurable version and poll interval. Will remove the configurability of reload which will make the changes much smaller, too.

packages/kit/src/core/build/build_client.js
```diff
@@ -29,6 +29,10 @@ export async function build_client({
 	output_dir,
 	client_entry_file
 }) {
+	process.env.VITE_APP_VERSION = config.kit.version?.name || Date.now().toString();
+	process.env.VITE_APP_VERSION_FILE = `${config.kit.appDir}/version.json`;
+	process.env.VITE_APP_VERSION_POLL_INTERVAL = `${config.kit.version?.pollInterval || 60_000}`;
```
Rich-Harris, line 34: Config is always fully defined — defaults should be specified in `options.js` (I realise this might make testing slightly more finicky)

```suggestion
	process.env.VITE_APP_VERSION = config.kit.version.name;
	process.env.VITE_APP_VERSION_FILE = `${config.kit.appDir}/version.json`;
	process.env.VITE_APP_VERSION_POLL_INTERVAL = `${config.kit.version.pollInterval}`;
```

pz-mxu:
I don't know if passing the version, poll interval and version file's location as internal environment variables is acceptable, but I didn't come up with any better solution without adding more parameters passed to Renderer.

packages/kit/src/core/config/options.js
```diff
@@ -255,6 +255,11 @@ const options = object(
 
 			trailingSlash: list(['never', 'always', 'ignore']),
 
+			version: object({
+				name: string(null),
```
Rich-Harris, line 259:
```suggestion
				name: string(Date.now().toString()),
```

pz-mxu: But this will break the tests, because the timestamp will not be the same.

packages/kit/src/runtime/app/env.js
```diff
@@ -14,5 +14,9 @@ export const mode = import.meta.env.MODE;
  * @type {import('$app/env').amp}
  */
 export const amp = !!import.meta.env.VITE_SVELTEKIT_AMP;
+/**
+ * @type {import('$app/env').version}
+ */
+export const version = import.meta.env.VITE_APP_VERSION?.toString();
```
Rich-Harris, line 20: it's guaranteed to exist and be a string

```suggestion
export const version = import.meta.env.VITE_APP_VERSION;
```

packages/kit/src/core/config/options.js
pz-mxu: I had it that way before.

packages/kit/src/runtime/app/env.js
Rich-Harris: ah, I see — it's not being set in dev. I think it's probably easier to set it to something dev-specific in dev, that way we have fewer points of divergence

packages/kit/src/core/config/options.js
Rich-Harris: it might require some finessing (overriding/deleting the value during testing, a la `remove_keys`, or monkey-patching `Date.now()` or something) but i think it's probably preferable to be able to make guarantees about the shape of the config

packages/kit/src/core/adapt/prerender/fixtures/large-page/input.html
```diff
@@ -1786,6 +1786,16 @@ <h3 id="modules-$app-stores">$app/stores</h3>
...
+							<li>
+								<code>updating</code> is a
+								<a
+									href="https://svelte.dev/tutorial/writable-stores"
+									target="_blank"
+									rel="noopener noreferrer"
+									>writable store</a
+								>
+								whose initial value is false. The app's version will be checked every minute for changes, setting the value to true if the version does not match the version at the time the store was created. It can be written to when custom logic is required to detect updates. There is a <code>check</code> function to force the version check immediately, returning the result.
+							</li>
```
Rich-Harris, line 1798: we don't need to keep this file up to date. in fact i think we should remove it from the repo soon (it was originally there for benchmarking purposes)

```suggestion
```

packages/kit/src/core/adapt/prerender/fixtures/large-page/output.json
```diff
@@ -90,6 +90,7 @@
 	"#loading-output-stuff",
 	"https://svelte.dev/tutorial/writable-stores",
 	"#hooks-getsession",
+	"https://svelte.dev/tutorial/writable-stores",
```
Rich-Harris, line 93: ditto

```suggestion
```

pz-mxu (commented):
This causes the type check to fail.

`export const version = import.meta.env.VITE_APP_VERSION;`

Rich-Harris:
I don't think we need to expose `version` at all, to be honest. We could add it in future if it's useful but it would keep this PR simpler if we got rid of it

documentation/docs/14-configuration.md
```diff
@@ -68,6 +68,10 @@ const config = {
 		},
 		target: null,
 		trailingSlash: 'never',
+		version: {
+			name: Date.now(),
+			pollInterval: 60_000,
```
Rich-Harris, line 73: After discussion with other maintainers, this should probably default to `0`, which would mean 'no polling'

packages/kit/src/core/adapt/prerender/fixtures/large-page/input.html
```diff
@@ -1732,7 +1732,7 @@ <h3 id="modules-$app-stores">$app/stores</h3>
 						<ul>
 							<li>
 								<code>getStores</code> is a convenience function around <code>getContext</code> that
-								returns <code>{ navigating, page, session }</code>. This needs to be called at the
+								returns <code>{ navigating, page, session, updating }</code>. This needs to be called at the
```
Rich-Harris, line 1735: removing this change (see #3661)

```suggestion
								returns <code>{ navigating, page, session }</code>. This needs to be called at the
```

packages/kit/src/runtime/client/renderer.js
```diff
@@ -104,11 +147,26 @@ export class Renderer {
 			promise: null
 		};
 
+		const versionPollInterval =
+			typeof import.meta.env.VITE_APP_VERSION_POLL_INTERVAL === 'string'
+				? parseInt(import.meta.env.VITE_APP_VERSION_POLL_INTERVAL)
+				: 60_000;
```
Rich-Harris, line 153: as long as we set `process.env.VITE_APP_VERSION_POLL_INTERVAL` in dev as well as build (albeit to 0, i.e. no polling) we can simplify this (note that it also changes `versionPollInterval` to `version_poll_interval` to be consistent with internal variable names elsewhere in the codebase):

```suggestion
		const version_poll_interval = +import.meta.env.VITE_APP_VERSION_POLL_INTERVAL;
```

packages/kit/src/runtime/client/renderer.js
```diff
@@ -104,11 +147,26 @@ export class Renderer {
...
+			session: writable(session),
+			updated: checkable_store(
+				false,
+				(set) => {
+					return has_version_changed().then((changed) => {
+						if (changed) set(true);
+						return changed;
+					});
+				},
+				versionPollInterval
```
Rich-Harris, line 168: see previous

```suggestion
				version_poll_interval
```

packages/kit/src/runtime/client/renderer.js
pz-mxu: Gives type error now, too.

packages/kit/src/runtime/client/renderer.js
```diff
@@ -39,6 +40,48 @@ function notifiable_store(value) {
...
 
+/**
+ * @param {any} value
+ * @param {(set: (new_value: any) => void) => any} fn
+ * @param {number | undefined} interval
+ */
+function checkable_store(value, fn, interval) {
+	const { set, update, subscribe } = writable(value);
+
+	setInterval(() => {
```
benmccann, line 51: perhaps this should only be done if `interval > 0`

pz-mxu: Already changed that while setting default interval to 0. Not yet pushed ;-)

documentation/docs/05-modules.md
```diff
@@ -71,6 +71,8 @@ The stores themselves attach to the correct context at the point of subscription
 - `page` contains an object with the current [`url`](https://developer.mozilla.org/en-US/docs/Web/API/URL), [`params`](#loading-input-params), [`stuff`](#loading-output-stuff), [`status`](#loading-output-status) and [`error`](#loading-output-error).
 - `session` is a [writable store](https://svelte.dev/tutorial/writable-stores) whose initial value is whatever was returned from [`getSession`](#hooks-getsession). It can be written to, but this will _not_ cause changes to persist on the server — this is something you must implement yourself.
 
+- `updated` is a [writable store](https://svelte.dev/tutorial/writable-stores) whose initial value is false. The app's version will be checked every minute for changes, setting the value to true if the version does not match the version at the time the store was created. It can be written to when custom logic is required to detect updates. There is a `check` function to force the version check immediately, returning the result.
```
Conduitry, line 74: This should be updated with the new default behavior.

`updated` should be added to the list of `import`s above from `'$app/stores'`.

For consistency with the other elements in this list, we should remove the newline before `- updated`.

Rich-Harris: I didn't actually clock that the store was writable. I don't think that's a good idea, I think it should be a readable store with a `check` method

Rich-Harris:
```suggestion
- `updated` is a [readable store](https://svelte.dev/tutorial/readable-stores) whose initial value is false. If [`version.pollInterval`](#configuration-version) is a non-zero value, SvelteKit will poll for new versions of the app and update the store value to `true` when it detects one. `updated.check()` will force an immediate check, regardless of polling.
```

documentation/docs/05-modules.md
```diff
@@ -56,20 +56,21 @@ import { base, assets } from '$app/paths';
...
 - `page` contains an object with the current [`url`](https://developer.mozilla.org/en-US/docs/Web/API/URL), [`params`](#loading-input-params), [`stuff`](#loading-output-stuff), [`status`](#loading-output-status) and [`error`](#loading-output-error).
 - `session` is a [writable store](https://svelte.dev/tutorial/writable-stores) whose initial value is whatever was returned from [`getSession`](#hooks-getsession). It can be written to, but this will _not_ cause changes to persist on the server — this is something you must implement yourself.
+- `updated` is a [writable store](https://svelte.dev/tutorial/writable-stores) whose initial value is false. The app's version, which can be specified as [`config.kit.version.name`], will optionally be checked for changes regularly, setting the value to true if the version does not match the version at the time the store was created. There is a `check` function to force the version check immediately, returning the result. The interval for regular checks can be specified in ms as [`config.kit.version.pollInterval`]. A value of `0` will disable regular checks. The store can be written to when custom logic is required to detect updates.
```
Rich-Harris, line 73:
```suggestion
- `updated` is a [readable store](https://svelte.dev/tutorial/readable-stores) whose initial value is false. If [`version.pollInterval`](#configuration-version) is a non-zero value, SvelteKit will poll for new versions of the app and update the store value to `true` when it detects one. `updated.check()` will force an immediate check, regardless of polling.
```

documentation/docs/05-modules.md
pz-mxu: The only reason I made it writable is to enable people to add custom checks while still using that store. Since the store is used when routing errors occur, that behavior can be manually triggered.

packages/kit/src/runtime/app/stores.js
```diff
@@ -77,3 +78,21 @@ export const session = {
...
 };
+
+/** @type {typeof import('$app/stores').updated} */
+export const updated = {
+	subscribe(fn) {
+		const store = getStores().updated;
+
+		if (browser) {
+			updated.set = store.set;
+			updated.update = store.update;
```
Rich-Harris, line 89: we should make this a readable store

```suggestion
```

packages/kit/src/runtime/app/stores.js
```diff
@@ -77,3 +78,21 @@ export const session = {
...
+			updated.check = store.check;
+		}
+
+		return store.subscribe(fn);
+	},
+	set: () => throw_error('set'),
+	update: () => throw_error('update'),
```
Rich-Harris, line 96:
```suggestion
```

packages/kit/src/runtime/client/renderer.js
```diff
@@ -39,6 +40,50 @@ function notifiable_store(value) {
...
+
+	if (interval > 0) {
+		setInterval(() => {
+			fn(set);
+		}, interval);
+	}
+
+	return {
+		set,
+		update,
```
Rich-Harris, line 59:
```suggestion
```

packages/kit/src/runtime/server/page/render.js
```diff
@@ -80,12 +80,23 @@ export async function render_response({
...
 
+		const updated = (() => {
+			const { set, update, subscribe } = writable(false);
+			return {
+				set,
+				update,
+				subscribe,
+				check: () => {}
+			};
+		})();
```
Rich-Harris, line 91:
```suggestion
		const updated = readable(false);
		updated.check = () => {};
```

Rich-Harris: we could in fact hoist this outside the function and reuse it, since its value will never change

packages/kit/types/ambient-modules.d.ts
```diff
@@ -110,6 +110,7 @@ declare module '$app/stores' {
 		navigating: typeof navigating;
 		page: typeof page;
 		session: Writable<Session>;
+		updated: Writable<boolean> & { check: () => boolean };
```
Rich-Harris, line 113:
```suggestion
		updated: typeof updated;
```

packages/kit/types/ambient-modules.d.ts
```diff
@@ -132,6 +133,11 @@ declare module '$app/stores' {
 	 * It can be written to, but this will not cause changes to persist on the server — this is something you must implement yourself.
 	 */
 	export const session: Writable<any>;
+	/**
+	 * A writable store indicating if the site was updated since the store was created.
+	 * It can be written to when custom logic is required to detect updates.
+	 */
+	export const updated: Writable<boolean> & { check: () => boolean };
```
Rich-Harris, line 140:
```suggestion
	export const updated: Readable<boolean> & { check: () => boolean };
```

documentation/docs/14-configuration.md
```diff
@@ -68,6 +68,10 @@ const config = {
 		},
 		target: null,
 		trailingSlash: 'never',
+		version: {
+			name: Date.now(),
```
Rich-Harris, line 72:
```suggestion
			name: Date.now().toString(),
```

documentation/docs/14-configuration.md
```diff
@@ -253,6 +257,13 @@ Whether to remove, append, or ignore trailing slashes when resolving URLs to rou
 
 > Ignoring trailing slashes is not recommended — the semantics of relative paths differ between the two cases (`./y` from `/x` is `/y`, but from `/x/` is `/x/y`), and `/x` and `/x/` are treated as separate URLs which is harmful to SEO. If you use this option, ensure that you implement logic for conditionally adding or removing trailing slashes from `request.path` inside your [`handle`](#hooks-handle) function.
 
+### version
+
+An object containing zero or more of the following values:
+
+- `name` - current app version string
+- `pollInterval` - interval in milliseconds to poll for version changes
+
```
Rich-Harris, line 266: Probably worth explaining what this config is for

```suggestion

Client-side navigation can be buggy if you deploy a new version of your app while people are using it. If the code for the new page is already loaded, it may have stale content; if it isn't, the app's route manifest may point to a JavaScript file that no longer exists. SvelteKit solves this problem by falling back to traditional full-page navigation if it detects that a new version has been deployed, using the `name` specified here (which defaults to a timestamp of the build).

If you set `pollInterval` to a non-zero value, SvelteKit will poll for new versions in the background and set the value of the [`updated`](#modules-$app-stores) store to `true` when it detects one.
```

Rich-Harris:
Made a small tweak so that the interval is always set (albeit to 0) even in dev. 

One last (I think) thing — we probably need to use the `VITE_SVELTEKIT_` namespace for consistency with existing code.

pz-mxu:
> Made a small tweak so that the interval is always set (albeit to 0) even in dev.

Great, I already looked for the place where this has to go.

documentation/docs/14-configuration.md
```diff
@@ -253,6 +257,16 @@ Whether to remove, append, or ignore trailing slashes when resolving URLs to rou
...
+Client-side navigation can be buggy if you deploy a new version of your app while people are using it. If the code for the new page is already loaded, it may have stale content; if it isn't, the app's route manifest may point to a JavaScript file that no longer exists. SvelteKit solves this problem by falling back to traditional full-page navigation if it detects that a new version has been deployed, using the `name` specified here (which defaults to a timestamp of the build).
+
+If you set `pollInterval` to a non-zero value, SvelteKit will poll for new versions in the background and set the value of the [`updated`](#modules-$app-stores) store to `true` when it detects one.
```
Rich-Harris, line 269:
```suggestion
If you set `pollInterval` to a non-zero value, SvelteKit will poll for new versions in the background and set the value of the [`updated`](#modules-$app-stores) store to `true` when it detects one.

```

Rich-Harris (approved):

Rich-Harris:
We did it! Thanks so much for this mammoth effort @pz-mxu — this is a great addition to the framework

pz-mxu:
Happy I could be of help.

jycouet:
I'm not sure it's the right way to comment here, but I'll give it a try.

I'm precisely looking at `kit.browser.router` `onError` `reload` (Because I'm deploying a lot atm 🥳)
But on `SvelteKit v1.0.0-next.295` (latest now) I can't figure out how to have this excellent reloading behavior.

Could you help me?

Antonio-Bennett:
Hey @jycouet After reading through this seems like it is a configuration on `svelte.config.js` specifically https://kit.svelte.dev/docs/configuration#version

jycouet:
I did change `pollInterval` to 5000 to see.
But it's not doing `onError` => `reload`. Actually, as I'm not using it (`updated`), it's not doing anything 😅

Now, using `updated` I'm deploying to see if it's better. 
In `__layout.svelte`
```ts
import { updated } from '$app/stores';

updated.subscribe(async (value) => {
  if (value) {
    // Hard refresh when version is not matching. 
    // I can test only on prod! (dev or preview, this is not working) https://github.com/sveltejs/kit/issues/3957
    // Todo:
    // Nicer solution here one day: https://github.com/sveltejs/kit/issues/3666
    // or move to web workers? https://github.com/sveltejs/kit/issues/3667
    window.location.reload();
  }
});
```

=> It didn't refresh my page on the next deployment! I didn't see an error as well... 🤔

It's really hard to debug as I don't know how to reproduce locally.
Do you have ideas?

frederikhors:
> It didn't refresh my page on the next deployment!

Still not working?

jycouet:
Seems like it's not working :(

frederikhors:
I'm on 1.0.0-next.295 and `updated` store works. I tried with 5000 ms too...

jycouet:
🥳🥳🥳 I found out!!!

You need: `NODE_ENV: production` _(I had an issue in my CI that this fixed now 👌)_.
Another excellent info, the **version** can be found in: `VITE_SVELTEKIT_APP_VERSION`

EskelCz:
@Rich-Harris Isn't there an issue of the app losing its state? So for example a long text the user has written. I would rather show a warning message that the app was meanwhile updated and ask the user to refresh when it's convenient to them.

Also I love that SvelteKit at least has a solution, but when using Svelte+Vite only, there's no solution to be found. :(
This is the issue thread there: https://github.com/vitejs/vite/issues/11804

Thanks for your consideration.

Issue #3426, opened 2022-01-20
Allow adapters to pass in platform specific context


fezproof:
### Describe the problem

> Related #2844 #2807

As of #3384, hooks and endpoints receive a standard Request object. This simplified and standardised the API for adapters a lot, but also means that no custom context can be passed into the app for the developer to access.

As an example, when using the `adapter-cloudflare` and deploying to Cloudflare pages, you can pass in KV stores and Durable Objects to your functions. With the current Cloudflare adapter, you cannot access the env that is being passed in.
```ts
export default {
	async fetch(req, env) { // All context needed in env
		// ...
		if (url.pathname.startsWith(prefix)) {
			/** @type {Response} */
			const res = await env.ASSETS.fetch(req); // <-- env used to get static assets
	         // ...
		try {
			return await app.render(req); // <-- env not passed to the app
		} catch (e) {
                // ..
```

With the new `Request` `Response` type pattern implemented, it is now not possible to pass it in at all.

This would also be relevant to Netlify, Vercel, and any custom adapter for larger teams using something like Azure or AWS.

### Describe the proposed solution

The most straight forward way to solve this would be to pass in the request to `app.render` as part of an object, with the adapter environment passed in as another
```ts
return await app.render({ request, env });
```
This would be another breaking change for the adapter API, but shouldn't change anything for applications, as the request is already passed into handlers as part of an object
```ts
export async function post({ params, request, env }) {
  await env.TodoList.put("Task", "Clean Kitchen")
  const out = await env.TodoList.get("Task");
```

Another generic type should also be exposed to Handler and Hook functions to allow typing for the environment.

### Alternatives considered

There are possibilities of using globals, but this could cause issues in most environments.

### Importance

would make my life easier

### Additional Information

This is my first issue here, but am really impressed with how everything is amazing with svelte kit so far. Absolutely an inspiration <3
Am happy to create a PR for this if the changes make sense.
Hello from Perth Australia!

benmccann:
Agreed that this is needed :smile: Closing as duplicate of https://github.com/sveltejs/kit/issues/2426

tylerbrostrom:
@benmccann possible to opt for this issue over #2426, as this issue better illustrates the problem and proposes a possible solution?

PR #3429, opened 2022-01-20
Allow passing platform context from adapters

### LINKED ISSUE #2426: A way to provide info from platform request to application
### Describe the problem

There are things on the original platform request that are needed in SvelteKit and can't be accessed today. E.g. some Node middlewares stick things on the request and the user may need to access them, Cloudflare KV and Durable Objects need access to env (https://github.com/sveltejs/kit/issues/3426), etc.

A way for adapters to copy info from platform request into `locals`, a new field like `meta`, some new argument to `render`, or something else

### Alternatives considered

Had considered exposing the whole request object (https://github.com/sveltejs/kit/pull/2359)

### Importance

### Additional Information

_No response_

fezproof:
This is a draft PR that implements some of the ides in #2426 and #3426. I have only currently implemented passing in a new `platform` object from `adapter/cloudflare-pages` as an example, as it is what I am familiar with. Maintainers more familiar with the other adapters and environments should probably handle them.

The main change in this PR allows this in adapters:
```ts
return await app.render({ request: req, platform: { env } });
```
The developer can now access `env` from the new `platform` object, which is passed to hooks and endpoints.

Doing this PR has raised an important question though. 
*How should a pattern like this be handled in local development?*

Perhaps each adapter could provide a default `platform` that can act as a sort of 'polyfill'. Something like KV and Durable Objects from Cloudflare, or user context from Netlify shouldn't be that hard to shim.

### Tests

### Changesets
- [X] ~~Waiting on #3427 to allow changeset generation~~

Closes #2426


Rich-Harris:
Thank you! I'd personally lean towards creating a new property on `RequestEnv` (perhaps `meta`) and allowing that to be completely controlled by the adapter, rather than muddling `locals`. What do you think?

```js
response = await app.render({ request, meta: env });
```

The local development question is a good one, and it's actually an even deeper can of worms — should the adapter try to simulate the environment completely, by e.g. adding web APIs supported by Cloudflare Workers and Deno? should it _remove_ Node built-ins? (It's useful to be able to use `fs` during prerendering, in endpoints that will never run in a worker — does that affect the decision?) It's possible that the way to solve this is to somehow integrate with platforms' own development tools, though it's far from clear what that looks like.

All of which is to say that it's probably fine to kick that can down the road for now and address it in a follow-up PR!

fezproof:
Thanks for the feedback, will update the PR and docs with this approach. I thought about this too but I was afraid adding yet another object passed into the handle hook event could become overbearing, but it also makes a lot of sense to keep them seperate, allowing the developer to make the choice of what to do with it.

Another question to raise is whether this should be passed all the way down to endpoints. Separation of concerns, and not forcing the dev to implement `handle` get up and running with an adapter to get all its features. 

Again, probably a better question for a follow up PR :)

fezproof:
RE the environment point about `fs` and such, I think the Cloudflare adapter kind of handles this. I am not exactly sure how the other adapters work (yet), but the es build that happens as part of cloudflares build system forces a *'browser'* like environment. It might be ok to allow node specific things to be used in dev, if they are going to fail the build anyway before a deploy.

Rich-Harris:
> Another question to raise is whether this should be passed all the way down to endpoints

definitely, i think — it's easiest to understand if `RequestEvent` is used consistently throughout the app. not to mention simpler to implement!

fezproof:
@Rich-Harris @benmccann PR is now ready for review. I am not sure where to put more documentation for the new object, so some guidance there would be appreciated :)

Testing with a local repo, I am thinking meta should be made Readonly? Probably want to encourage that only locals is the object that should be changed.

benmccann:
I'm not in love with the name `meta`. Two alternatives would be to rename it to `platform` or to stick the items directly onto the request object. We should check with the other maintainers on the desired API

The other suggestion I would have is to stick the platform request on the "meta" object or whatever we end up calling it. This has been requested several times in `adapter-node` especially where some middlewares will stick stuff on the request and today there's no way to access it

fezproof:
> I'm not in love with the name `meta`. Two alternatives would be to rename it to `platform` or to stick the items directly onto the request object. We should check with the other maintainers on the desired API

I think you are right, meta doesn't quite describe what it is.

> The other suggestion I would have is to stick the platform request on the "meta" object or whatever we end up calling it. This has been requested several times in `adapter-node` especially where some middlewares will stick stuff on the request and today there's no way to access it

👍  will give this a shot, but I don't know if it should be a hard or fast rule. It makes sense for adapter node, but each adapter should probably handle this differently

documentation/docs/01-routing.md
```diff
@@ -192,6 +196,10 @@ A route can have multiple dynamic parameters, for example `src/routes/[category]
 
 > `src/routes/a/[...rest]/z.svelte` will match `/a/z` as well as `/a/b/z` and `/a/b/c/z` and so on. Make sure you check that the value of the rest parameter is valid.
 
+#### Platform object
+
+Some adapters may pass in some extra context when calling your application. They do this by setting the `meta` object, that can be accessed in both route handlers and hooks ([detailed later](#hooks)). It is up to the adapter to decide what is passed in.
```
benmccann, line 201:
```suggestion
Some adapters may pass in some extra context when calling your application. They do this by setting the `platform` object, that can be accessed in both route handlers and hooks ([detailed later](#hooks)). It is up to the adapter to decide what is passed in.
```

documentation/docs/04-hooks.md
```diff
@@ -125,7 +131,7 @@ export function getSession(event) {
 					email: event.locals.user.email,
 					avatar: event.locals.user.avatar
 				}
-			}
+		  }
```
benmccann, line 134: this should be changed back to a tab

packages/kit/src/core/build/build_server.js
```diff
@@ -171,14 +171,14 @@ export async function build_server(
...
 	fs.writeFileSync(
-		input.app,
-		template({
-			cwd,
-			config,
-			hooks: app_relative(hooks_file),
-			has_service_worker: service_worker_register && !!service_worker_entry_file
-		})
-	);
+    input.app,
```
benmccann, line 174: it looks like these may have been changed to spaces? but I'd have thought `pnpm format` or `pnpm lint` would catch it

tv42:
@benmccann 

> The other suggestion I would have is to stick the platform request on the "meta" object or whatever we end up calling it. This has been requested several times in adapter-node especially where some middlewares will stick stuff on the request and today there's no way to access it

I think this part is now done as of https://github.com/sveltejs/kit/pull/3384

This PR is still needed for the things that aren't riding on the request, like Cloudflare KV.

benmccann:
> I think this part is now done as of #3384

They're different types. 3384 uses a web/fetch `Request`, but if the user wants the NodeJS / Express `Request` that it was created from then we would still need to provide it here

packages/kit/src/core/build/build_server.js
fezproof: I took a look around the app, there seems to be spaces in stead of tabs everywhere :O 

Do you want to just ignore these for now and I can do a lint blitz in a seperate PR?

fezproof: Looks like this one was caused by an existing prettier-ignore, so fixed that

documentation/docs/04-hooks.md
fezproof: Not sure what is causing this one, can't find a setting that would do it. Fixed now

fezproof:
@Rich-Harris @benmccann This PR is ready for review 👍

documentation/docs/10-adapters.md
```diff
@@ -93,6 +93,7 @@ Within the `adapt` method, there are a number of things that an adapter should d
   - Imports `App` from `${builder.getServerDirectory()}/app.js`
   - Instantiates the app with a manifest generated with `builder.generateManifest({ relativePath })`
   - Listens for requests from the platform, converts them to a standard [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) if necessary, calls the `render` function to generate a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) and responds with it
+  - `render` optionally accepts a `platform` object, which allow passing of environment specific values and functions. The `platform` object will become be accessable from the handle hook, and any API endpoints in the application
```
am283721, line 96: Just a couple minor spelling/grammar changes:

`render` optionally accepts a `platform` object, which allows passing of environment specific values and functions. The `platform` object will be accessible from the handle hook and any API endpoints in the application

documentation/docs/04-hooks.md
benmccann: the code blocks in markdown files aren't linted. If it's just a regular `.js` file it should be handled though

packages/adapter-node/src/handler.js
```diff
@@ -45,7 +45,7 @@ const ssr = async (req, res) => {
 		return res.end(err.reason || 'Invalid request body');
 	}
 
-	setResponse(res, await app.render(request));
+	setResponse(res, await app.render({ request, platform: { nodeRequest: req } }));
```
benmccann, line 48: personally, I would just call this `request` as I think `platform.request` should hopefully be clear enough
```suggestion
	setResponse(res, await app.render({ request, platform: { request: req } }));
```

we should also update the README for each adapter to document what they populate into `platform` (i.e. `request` here and `env` in `adapter-cloudflare`)

fezproof: That makes sense, I may call it req instead though as that is what most node devs are used to

benmccann: I still like `request` more than `req`

fezproof:
I have now updated the readme, but we should probably just get this in now if everything looks good. If other adaptors need to be updated, they should be in their own PRs.

packages/adapter-node/src/handler.js
fezproof: Done :)

documentation/docs/01-routing.md
```diff
@@ -51,11 +51,12 @@ Endpoints are modules written in `.js` (or `.ts`) files that export functions co
 // Declaration types for Endpoints
 // * declarations that are not exported are for internal use
 
-export interface RequestEvent<Locals = Record<string, any>> {
+export interface RequestEvent<Locals = Record<string, any>, Platform = Record<string, unknown>> {
```
benmccann, line 54: Maybe let's use `any` instead of `unknown` for now for consistency and then make any changes across the whole codebase when addressing https://github.com/sveltejs/kit/issues/3513#issuecomment-1022238157 so that we come up with a consistent set of guidelines for using `any` vs `unknown`

benmccann (approved):
looks pretty good to me. @Rich-Harris should take a look as well though

.changeset/cyan-numbers-change.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sveltejs/adapter-cloudflare': patch
+'@sveltejs/adapter-node': patch
+---
+
+Pass `platform` object to SvelteKit
```
Conduitry, line 6: Do we want to say in the changelog what it contains? If so, we're presumably want separate changesets for the Cloudflare and Node adapters.

Rich-Harris:
After discussion, we decided it makes more sense to have `app.render(request, { platform })` since `request` is the required argument and the options are optional. This also means that it's a non-breaking change. I've updated the PR accordingly.

Though now that I review it more closely, I find myself asking what problem this actually solves. In the Cloudflare Workers case, `env` is a global, so as far as I can tell there's nothing to be gained from passing it around like this. The example added to the `adapter-node` README demonstrates reading a cookie, which is trivial to do already.

Do we actually need this?

Rich-Harris:
Ah, I see there's a difference between old and new-style workers, and it isn't a global if you're using the new kind (which we are)

fezproof:
Yeah that’s right. For reference for others seeing this later.

Env is only global in the current common js cloudflare workers. I the newer ES Module workers, and cloudflare pages, it is only passed through as env in the fetch function call.

See 
https://blog.cloudflare.com/workers-javascript-modules/
https://developers.cloudflare.com/pages/platform/functions#advanced-mode
From the durable object docs
> In modules-syntax workers, bindings are delivered as a property of the environment object passed as the second parameter when an event handler or class constructor is invoked. This is new compared to pre-module workers, in which bindings show up as global variables.

documentation/docs/04-hooks.md
Conduitry: This still doesn't look fixed.

documentation/docs/04-hooks.md
```diff
@@ -125,7 +130,7 @@ export function getSession(event) {
...
```
Conduitry, line 133:
```suggestion
			}
```

fezproof:
Perhaps the node README could have an example with something like passport? I haven’t used a node server for a few years so I don’t know what the cool kids are using these days.
We could also delay updates to other adaptors for a future PR as well.

Rich-Harris:
My inclination is probably just to leave adapter-node unchanged until a use case presents itself. Otherwise we might find we're adding the wrong surface area. Updated the PR

Conduitry:
@Rich-Harris Ben had said on Discord

> The Node adapter is passing the request which is something we really needed because some middlewares stick stuff on the request and we have no way to access those at the moment

Rich-Harris (approved):

Rich-Harris:
> some middlewares stick stuff on the request and we have no way to access those at the moment

Would be curious to know what these middlewares are doing and whether there's any intrinsic reason it can't be done inside Kit instead. Presumably the middleware in question doesn't have access to anything that isn't represented on the `Request` object, it's just shoving stuff on `req` in the higgledy-piggledy way traditional of Express middlewares. If there's a valid use case then we could definitely do it in a follow-up PR, but the more we can encourage people to do things in an idiomatic way that can be used beyond just Node apps, the better

PR #3437, opened 2022-01-20
Add programmatic interface to create-svelte

### LINKED ISSUE #3421: Add a programmatic interface to create-svelte
### Describe the problem

`create-svelte` only works via the CLI, and only interactively. It would be helpful to expose a programmatic (and versioned) interface, so that we can keep a template repo up to date with `sveltejs/kit` (this template repo would be used to populate things like https://node.new/sveltekit, which will otherwise constantly be out of date)

Something like this:

```js
import { create } from 'create-svelte';

await create(dir, {
  eslint: false,
  typescript: false,
  prettier: true
});
```

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

AnatoleLucet:
This pr adds a programmatic interface to `create-svelte` so it can be called using a single function call.

Closes #3421 

Not sure if the package should be bundled, since it is now exporting an entry point. Let me know if it needs to.

### Tests

### Changesets

### DIFF
--- a/.changeset/fifty-foxes-tan.md
+++ b/.changeset/fifty-foxes-tan.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Add a programmatic interface to create-svelte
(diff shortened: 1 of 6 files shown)


Rich-Harris (approved):

Rich-Harris:
excellent, thank you!

PR #3451, opened 2022-01-20
Update template automatically

Rich-Harris:
In theory, this will update https://github.com/sveltejs/kit-template-default every time a new version of `create-svelte` is released. This will allow us to use that repo as the base for things like https://node.new/sveltekit.

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/package.json
+++ b/packages/create-svelte/package.json
@@ -34,7 +34,7 @@
 		"format": "npm run check-format -- --write",
 		"check-format": "prettier --check . --config ../../.prettierrc --ignore-path ../../.gitignore --ignore-path .gitignore --plugin prettier-plugin-svelte --plugin-search-dir=.",
 		"prepublishOnly": "npm run build",
-		"postpublish": "./scripts/update-template-repo.sh"
+		"postpublish": "bash ./scripts/update-template-repo.sh"
 	},
 	"files": [
 		"dist",
--- a/packages/create-svelte/package.json
+++ b/packages/create-svelte/package.json
@@ -33,7 +33,8 @@
 		"lint": "eslint --ignore-path .gitignore --ignore-path ../../.gitignore \"./*.js\" && npm run check-format",
 		"format": "npm run check-format -- --write",
 		"check-format": "prettier --check . --config ../../.prettierrc --ignore-path ../../.gitignore --ignore-path .gitignore --plugin prettier-plugin-svelte --plugin-search-dir=.",
-		"prepublishOnly": "npm run build"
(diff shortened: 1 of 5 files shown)


packages/create-svelte/package.json
```diff
@@ -33,7 +33,8 @@
 		"lint": "eslint --ignore-path .gitignore --ignore-path ../../.gitignore \"./*.js\" && npm run check-format",
 		"format": "npm run check-format -- --write",
 		"check-format": "prettier --check . --config ../../.prettierrc --ignore-path ../../.gitignore --ignore-path .gitignore --plugin prettier-plugin-svelte --plugin-search-dir=.",
-		"prepublishOnly": "npm run build"
+		"prepublishOnly": "npm run build",
+		"postpublish": "./scripts/update-template-repo.sh"
```
bluwy, line 37:
```suggestion
		"postpublish": "bash ./scripts/update-template-repo.sh"
```

Do we need a leading `bash` command here? IIRC it's needed in Ubuntu (CI).

packages/create-svelte/scripts/update-template-repo.sh
```diff
@@ -0,0 +1,37 @@
...
+  # $1 : relative filename
+  echo "$(cd "$(dirname "$1")" && pwd)/$(basename "$1")"
+}
+
+DIR=$(get_abs_filename $(dirname "$0"))
+TMP=$(get_abs_filename "$DIR/../node_modules/.tmp")
+
+echo "dir: $DIR"
+echo "tmp: $TMP"
+
```
Rich-Harris, line 13: this was for testing, can be removed

```suggestion
```

packages/create-svelte/tsconfig.json
```diff
@@ -4,8 +4,8 @@
 		"checkJs": true,
 		"noEmit": true,
 		"noImplicitAny": true,
-		"target": "es2020",
```
benmccann, line 7: this isn't necessarily a small change. everytime we've tweaked this it's had an impact on some platform and I wouldn't be surprised if it introduces some breakage somewhere. it'd probably be best to do in a separate PR with a changeset if it's necessary

Rich-Harris: it doesn't matter anyway, this PR is arguably redundant since #3458 and the zillion follow-ups. going to close it

benmccann: Ooh. @antony's PR made the same change though, which I don't feel great about. Is necessary or do you think we can back that part out?

PR #3465, opened 2022-01-20
[fix] Exclude keys common for T and U on Only type

JeanJPNM:

### Tests

### Changesets

This pull request fixes an issue that happened when the arguments of `Only` had a common property name, creating a key conflict and making the type resolve to `never`. The issue is fixed by excluding the keys common for T and U before iterating over them.

Here is a [reproduction](https://www.typescriptlang.org/play?#code/C4TwDgpgBA8gdgGxAHgCoBooFUB8UC8UA3gFACQA2gApQCWcUA1hCAPYBmUqAugFxfVuAbhIBfKADJi5anQbM2nLNwD8-OBABuEAE4jRIkhAAeYVjuBRQkKAFFawABa60mXAViIUGbHgA+nkjIWJioOIb0wLrsAIYAxtAAYqys0lBQMeoArgC2AEa6JOl5-ADOwDr0AOZiJJHR8dAAQjE6aVAlHSkIEDFwRVBx2fmFoiQ9luwp-PZOLsmsmC06eISk6ZlQAIzoA50ARDGlACb7teMQlnmtMw7OOsgLS62r7Z0VWRC76UPbYkA)

And [here](https://www.typescriptlang.org/play?ssl=25&ssc=2&pln=17&pc=1#code/C4TwDgpgBA8gdgGxAHgCoBooFUB8UC8UA3gFACQA2gApQCWcUA1hCAPYBmUqAugFxfVuAbhIBfKADJi5anQYBRAB4BjBAFcAJhGTM2nLJl0cuObgH5+cCADcIAJxGiRJCIrCs7wKKEhR5tYAALezRMXAJYRBQMbDwAH0ikZAMTZ3pge3YAQ2VoADFWVmkoKCzLNQBbACN7EhKq-gBnYDt6AHMxEnTMnOgAISy7YqgGkcKECCy4Oqhlcura0RIJr3ZC-n8gkILWTAG7PEJSErKoAEZ0GdGAIizGjWvO5YgvKsGNgOC7ZB29wcPhqMWmoIJcSnNzmIgA) is how the fixed version behaves

### DIFF
--- a/packages/kit/types/helper.d.ts
+++ b/packages/kit/types/helper.d.ts
@@ -13,7 +13,7 @@ export type JSONString =
 export type ResponseHeaders = Record<string, string | string[]>;
 
 // <-- Utility Types -->
-type Only<T, U> = { [P in keyof T]: T[P] } & { [P in keyof U]?: never };
+type Only<T, U> = { [P in keyof T]: T[P] } & { [P in Exclude<keyof U, keyof T>]?: never };
 
 export type Either<T, U> = Only<T, U> | Only<U, T>;
 export type InferValue<T, Key extends keyof T, Default> = T extends Record<Key, infer Val>


benmccann:
what was broken? is there some kind of test we could add for this? I'm not a TypeScript expert, so have no idea how to read this

dummdidumm:
At my phone right now so can't investigate further. I understand the reason for the change but I wonder why `Only` was created in the first place then, because now that's the same as just using `T` as is.

JeanJPNM:
I've updated the description to include demonstrations on the typescript playground

ignatiusmb:
It mostly reads as "`Only` take keys from `T` and error out if any keys from `U` are present". This change guards against possible collisions of common keys, which is nice. But, we're using this internally and exclusively for isolating `{ fallthrough: true }`, which I highly doubt would collide with anything other types.

We probably don't need this right now, and having the types be as simple as they could would probably be best here. We can always come back to this when we need more complex isolation in the future.

Rich-Harris (approved):

PR #3471, opened 2022-01-20
publish tags

antony:
WIP for now, clean up our debugging, and prepare to set version in the commit message for the template repository.

### Tests

### Changesets


antony:
Before we try to get the `npm_version_number` into this PR, we need to consider if this is the correct version number to include. Updating kit components without updating the cli obviously would not update this release number, and all of the kit component versions would be listed in the `package.json` anyway. So I wonder if this becomes more of a date, or something, or a reference to a build workflow run, instead.

.github/workflows/update_template.yml
```diff
@@ -51,9 +49,8 @@ jobs:
       - name: Publish
         working-directory: ${{ github.workspace }}/template
         run: |
-          touch "generated-$(date +%s)"
           git config user.email "noreply@svelte.dev"
           git config user.name "[bot]"
           git add -A
-          git commit -m "version $npm_package_version"
+          git commit -m "version ${{ github.ref_name }}"
```
Rich-Harris, line 55: maybe this?

```suggestion
          git commit -m "create-svelte version ${{ github.ref_name }}"
```

.github/workflows/update_template.yml
```diff
@@ -1,11 +1,9 @@
...
 
 on:
-  workflow_run:
-    workflows: [Release]
-    types:
-      - completed
-  workflow_dispatch:
+  push:
+    tags:
+      - create-svelte@*
```
pngwn, line 6: Are we actually tagging? I thought we weren't? This will never run if we aren't.

pngwn (commented):
Small comment, looks fine otherwise.

.github/workflows/update_template.yml
antony: That's the other part of this PR. According to changesets docs, they always tag on publish, but unless you push those tags, they get discarded.

pngwn: Are we going to start publishing with changesets then instead of doing it manually?

Rich-Harris: we already do publish with changesets, no?

package.json
```diff
@@ -9,7 +9,7 @@
 		"check": "pnpm -r check",
 		"lint": "pnpm -r lint",
 		"format": "pnpm -r format",
-		"release": "pnpm publish --tag=next --filter=\"@sveltejs/*\" --filter=\"create-svelte\"",
+		"release": "pnpm publish --tag=next --filter=\"@sveltejs/*\" --filter=\"create-svelte\" && git push --follow-tags",
```
Rich-Harris, line 12: is this the right change @antony / @pngwn?

```suggestion
		"release": "pnpm build -r && changesets publish && git push --follow-tags",
```

antony: I think you call `changeset` via pnpm i.e. `pnpm changeset version` rather than directly.

I'll update the PR.

antony: Ok done, one question, why the `pnpm build -r`. We have already built the monorepo at this point via `pnpm build` which is the previous step in `release.yaml`.

I think we'd be safe to remove it from here.

Rich-Harris: Pretty sure the publish step was building, so I was just copying what was already happening. If that means we've been building twice and now we only need to build once, then great

Rich-Harris (approved):

PR #3472, opened 2022-01-20
various create-svelte improvements

Rich-Harris:
Three changes:

1. if you call `npm init svelte@next` without specifying a directory, it will prompt you for one, in line with other scaffolding tools. (if you hit 'enter', it will use the current directory, as currently happens)
2. if you Ctrl-C in the middle of the prompts, it will bail out, as you'd expect
3. the project name is an explicit option to `create`, rather than being inferred from the directory (i.e. the inference has been moved back to the CLI) 

### Tests

### Changesets

### DIFF
--- a/.changeset/lemon-camels-brush.md
+++ b/.changeset/lemon-camels-brush.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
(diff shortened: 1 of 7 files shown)


antony (approved):
Great!

PR #3486, opened 2022-01-21
feat(kit): type `config.compilerOptions` as `CompileOptions` instead of any

metonym:
This types `compilerOptions` using the `CompileOptions` interface for an improved developer experience. This is a nice-to-have as it would be useful to glean which options and values are valid without looking up the docs.

`vite-plugin-svelte` also [strongly types `compilerOptions`](https://github.com/sveltejs/vite-plugin-svelte/blob/e0505412daf2be60f4adde9491332784088522a9/packages/vite-plugin-svelte/src/utils/options.ts#L378).

### Changesets

### DIFF
--- a/.changeset/young-ears-rhyme.md
+++ b/.changeset/young-ears-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Type compilerOptions as CompileOptions instead of any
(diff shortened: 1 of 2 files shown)


Rich-Harris (approved):

Rich-Harris:
thank you!

Issue #3488, opened 2022-01-21
The HTML minifier docs in the migration guide are out of date


geoffrich:
### Describe the bug

The [HTML minifier docs](https://kit.svelte.dev/migrating#integrations-html-minifier) need to be updated with the breaking Request/Response changes. Specifically, this snippet needs to read the body and headers a different way.

```html
  if (prerendering && response.headers['content-type'] === 'text/html') {
    response.body = minify(response.body, minification_options);
  }
```

### Reproduction

Go to the [migration docs](https://kit.svelte.dev/migrating#integrations-html-minifier) and notice that the HTML minifier section is out of date.

### Logs

```shell
n/a
```

### System Info

```shell
n/a
```

### Severity

annoyance

PR #3492, opened 2022-01-21
[docs] update html-minifier docs

benmccann:
closes https://github.com/sveltejs/kit/issues/3488

### DIFF
--- a/documentation/migrating/99-integrations.md
+++ b/documentation/migrating/99-integrations.md
@@ -35,7 +35,7 @@ export async function handle({ event, resolve }) {
   const response = await resolve(event);
 
   if (prerendering && response.headers.get('content-type') === 'text/html') {
-    return new Response(minify(response.text(), minification_options), {
+    return new Response(minify(await response.text(), minification_options), {
     	status: response.status,
     	headers: response.headers
     });
--- a/documentation/migrating/99-integrations.md
+++ b/documentation/migrating/99-integrations.md
@@ -34,8 +34,11 @@ const minification_options = {
 export async function handle({ event, resolve }) {
   const response = await resolve(event);
 
-  if (prerendering && response.headers['content-type'] === 'text/html') {
-    response.body = minify(response.body, minification_options);
+  if (prerendering && response.headers.get('content-type') === 'text/html') {
+    return new Response(minify(response.text(), minification_options), {
+    	status: response.status,
+    	headers: response.headers
+    });
   }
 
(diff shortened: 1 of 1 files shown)


Rich-Harris:
I kinda wonder if we should implement a `transformPage` option on `resolve` to handle this use case:

```js
const response = await resolve(request, {
  transformPage: ({ html }) => minify(html, minification_options)
});
```

Creating a duplicate response feels a bit wasteful

Conduitry:
That feels like a little bit of a confusing and niche API for this. What would it get passed besides `html`? Would it only get called for `text/html` content? Would it let you do anything to modify headers or other content besides the body? This feels like a hook within a hook.

benmccann:
It is a bit verbose currently. I also wonder if it'd be a big niche to offer a helper. I do think if we were going to offer a helper it should allow you to override more than just the body such as:
```
cloneResponse(original, { status: newStatus, body: newBody, headers: newHeaders })
```

Rich-Harris:
It would be purely for manipulating rendered pages. This line... https://github.com/sveltejs/kit/blob/0cca67063a4f38806eb4428b76548b8d948035ca/packages/kit/src/runtime/server/page/render.js#L218 ...would become something like 

```js
const html = options.transformPage(options.template({ head, body, assets }));
```

There are other reasons you might want to manipulate the HTML, like replacing `<html lang="%lang%">` with the current language (yes, Kit should provide first-class support for that anyway, but until it does...), plus probably others we haven't thought of. So it's not _just_ scenario-solving

Conduitry:
We should merge this PR as-is anyway. I don't want to rush into an API for a new feature because it's holding up a fix in the docs about how the system currently works.

Rich-Harris:
I've opened #3511

documentation/migrating/99-integrations.md
```diff
@@ -34,8 +34,11 @@ const minification_options = {
 export async function handle({ event, resolve }) {
   const response = await resolve(event);
 
-  if (prerendering && response.headers['content-type'] === 'text/html') {
-    response.body = minify(response.body, minification_options);
+  if (prerendering && response.headers.get('content-type') === 'text/html') {
+    return new Response(minify(response.text(), minification_options), {
```
Rich-Harris, line 38:
```suggestion
    return new Response(minify(await response.text(), minification_options), {
```

Rich-Harris (approved):

PR #3493, opened 2022-01-21
insert http-equiv cache-control tag when prerendering

Rich-Harris:
If a page specifies a `maxage` in its `load` result, a cache-control header is generated. Since apps don't (typically) have any control over headers with prerendered content, this is the next best thing. Mentioned briefly in https://github.com/sveltejs/kit/issues/909#issuecomment-1004128975.

This also lays the groundwork for inserting CSP headers in prerendered content.

### Tests

### Changesets

### DIFF
--- a/.changeset/modern-sloths-greet.md
+++ b/.changeset/modern-sloths-greet.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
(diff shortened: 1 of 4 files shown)


benmccann (approved):
seems reasonable to me

Issue #3491, opened 2022-01-21
Validate `handle` response


PH4NTOMiki:
### Describe the bug

I upgraded the adapter-node and kit, and changed the hooks to reflect new syntax from https://github.com/sveltejs/kit/pull/3384 and now the server crashes when started with the exit code in log

### Reproduction

https://github.com/PH4NTOMiki/sveltekit-book

### Logs

```shell
file:///home/mutic/dev/sveltekit-book/node_modules/@sveltejs/kit/dist/node.js:64
        res.writeHead(response.status, Object.fromEntries(response.headers));
                                              ^

TypeError: object is not iterable (cannot read property Symbol(Symbol.iterator))
    at Function.fromEntries (<anonymous>)
    at setResponse (file:///home/mutic/dev/sveltekit-book/node_modules/@sveltejs/kit/dist/node.js:64:40)
    at file:///home/mutic/dev/sveltekit-book/node_modules/@sveltejs/kit/dist/chunks/index.js:2083:8
    at processTicksAndRejections (node:internal/process/task_queues:96:5)

Node.js v17.2.0
```

### System Info

```shell
System:
    OS: Linux 5.13 Manjaro Linux
    CPU: (8) x64 Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
    Memory: 1.28 GB / 7.65 GB
    Container: Yes
    Shell: 5.1.12 - /bin/bash
  Binaries:
    Node: 17.2.0 - /usr/local/bin/node
    npm: 8.1.4 - /usr/bin/npm
  Browsers:
    Firefox: 95.0.1
  npmPackages:
    @sveltejs/adapter-node: ^1.0.0-next.66 => 1.0.0-next.66 
    @sveltejs/kit: ^1.0.0-next.236 => 1.0.0-next.236 
    svelte: ^3.44.3 => 3.44.3
```

### Severity

blocking an upgrade

### Additional Information

Everything was working without any issue before update.

benmccann:
I'm guessing it's this line:

https://github.com/sveltejs/kit/blob/1bce8e9eb1539a63bfdafd7a602b33062ea24b77/packages/kit/src/node.js#L64

Rich-Harris:
Can you provide more details about how to reproduce this? Your repro is working fine for me

PH4NTOMiki:
Actually for me it works sometimes, and sometimes not, (with and without the entries() fix).
[...]

PR #3494, opened 2022-01-21
[fix] correctly serialize response headers

benmccann:
fixes https://github.com/sveltejs/kit/issues/3491. Note thought that I haven't tested this

I'm surprised that TypeScript didn't catch this. I guess that the tests don't because this is only used by adapters which are largely untested?

### DIFF
--- a/.changeset/bright-pens-joke.md
+++ b/.changeset/bright-pens-joke.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] correctly serialize response headers
--- a/packages/kit/src/node.js
+++ b/packages/kit/src/node.js
@@ -61,7 +61,7 @@ export async function getRequest(base, req) {
 
 /** @type {import('@sveltejs/kit/node').SetResponse} */
 export async function setResponse(res, response) {
-	res.writeHead(response.status, Object.fromEntries(response.headers));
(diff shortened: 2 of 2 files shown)


Conduitry:
I'm surprised this is necessary. I think the `Headers` object is already an iterable, and I don't think you'd need to call `.entries()` on it to get the iterator.

Rich-Harris:
> I'm surprised this is necessary. I think the `Headers` object is already an iterable, and I don't think you'd need to call `.entries()` on it to get the iterator.

Correct. I'm not sure what's going on here

Conduitry:
In the browser console for me:

```js
a = new Headers
a.set('a', 'b')
Object.fromEntries(a)
// => Object { a: "b" }
```

I'm also using `Object.fromEntries(request.headers)` in an app, and it's working fine with whatever polyfilled version of these APIs is getting used.

Rich-Harris:
FWIW I can't reproduce it

PH4NTOMiki:
@Rich-Harris  I just tried replacing that line in my local file, and now the server works, weird

PH4NTOMiki:
@Rich-Harris and I even I think that is iterable without calling `entries()` so it's weird for me too

benmccann:
I'll close this. We can continue discussion in the issue. Thanks for the due diligence on this

PR #3496, opened 2022-01-21
error if handle returns non-Response

Rich-Harris:
this would have prevented #3491 — if you return anything other than a `Response` from `handle`, this will let you know about it

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -196,7 +196,7 @@ export async function respond(request, options, state = {}) {
 			}
 		});
 
-		// TODO remove for 1.0
+		// TODO for 1.0, change the error message to point to docs rather than PR
 		if (response && !(response instanceof Response)) {
 			throw new Error('handle must return a Response object' + details);
 		}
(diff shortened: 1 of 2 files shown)


packages/kit/src/runtime/server/index.js
```diff
@@ -195,6 +195,13 @@ export async function respond(request, options, state = {}) {
 				throw new Error('request in handle has been replaced with event' + details);
 			}
 		});
+
+		// TODO remove for 1.0
```
benmccann, line 199: Validation might still be useful after 1.0 to make sure users don't do anything stupid. Personally I would remove the todo

PH4NTOMiki: > Validation might still be useful after 1.0 to make sure users don't do anything stupid. Personally I would remove the todo

I'm obviously not on the Svelte developer team, but I agree with that, the check makes sense before and after 1.0 version

Rich-Harris: Checking might be useful, but pointing to the PR is specifically to help people migrate. After 1.0 they should just be directed to the docs. I changed the comment instead of removing it

PR #3499, opened 2022-01-21
CSP

Rich-Harris:
#93.

Lots to do here. CSP is a big topic and we might not nail it in a single PR, but I think this is a good starting point (let me know if I've overlooked or messed up anything):

* [x] add types for CSP directives
* [x] update `options.js` to expect new `csp` config option (including `mode: 'nonce' | 'hash' | 'auto'`)
* [ ] ~~generate hashes (probably easiest to just use sha256 everywhere?) from static assets~~ not necessary, it turns out
* [x] if `default-src` or `script-src` or `style-src` doesn't specify `unsafe-inline`, generate nonces or hashes (depending on the value of `mode` — defaults to 'auto', which means 'nonce for dynamically rendered stuff, hash for prerendered stuff'). Realistically, many apps will need `style-src: 'unsafe-inline'` until Svelte switches over to using WAAPI for transitions
* [ ] ~~if `default-src` or `script-src` or `style-src` specifies `strict-dynamic`, include nonces/hashes for _everything_ Kit controls~~ also not necessary
* [x] generate header, and add to response (or use http-equiv, for prerendering)
* [x] replace `%svelte.nonce%` in `src/app.html`, but not during prerendering
* [x] in dev, add `style-src 'unsafe-inline'` if necessary, skip nonce, ~~add `connect-src: ws://*`~~
* [x] always generate a nonce if `%svelte.nonce%` is present in the template (unless prerendering, obvs)

Stuff that's important but should probably stay out of scope for this PR:

* [ ] externalising scripts and styles when prerendering
* [ ] report-only 
* [ ] CSP for AMP pages
* [ ] `strict-dynamic` support

### Tests

### Changesets


Karlinator:
I've been experimenting a bit, both with hashing at compile time and general mayhem.

Conclusion: Browsers are weird and inconsistent. 

When using `strict-dynamic`, Chrome will happily load all script files as long as the inline init script is hashed (it appears Chrome sees the preloading as safe, which I suppose is accurate, as long as CSP is satisfied when they are applied). Firefox, however, refuses to load the script imported from the inline init script even though that inline script is hashed. This happens when I remove the `modulepreload` links as well, so might just be an issue with Firefox and `strict-dynamic`. Firefox has no issues when the init script is nonced, however, even if the hashes or nonces for the preload links are not present (just like Chrome).

I think this genuinely is a bug in Firefox's impllementation of `strict-dynamic`.

Without `strict-dynamic`, Chrome allows external scripts to be hashed (but only if the link has an `integrity` attribute with the hash), Firefox does not (a known deficiency). Both function with either hashes or nonces covering the init script as long as the rest are allowed with `script-src 'self'`

Other than that I was able to get Kit to hash the client js files on render and put the hashes in the manifest (though I'm not sure I'm happy with where I put it), and then the runtime fetches those hashes and insert them into `integrity` attributes. It occurs to me that there's no real reason to not include the `integrity` attribute? Even if it isn't required/useful for your specific CSP setup it's still an anti-tamper check I guess. Code: https://github.com/Karlinator/kit/commit/a74901ef864615d8f93d974aa378bcd0fe2c5eae

What's the best way for me to contribute code to this PR btw, if you want it? Should I just make PRs into the `csp` branch?

Rich-Harris:
> Conclusion: Browsers are weird and inconsistent.

Yikes. Thanks for investigating, that does indeed sound like a Firefox bug. The best we can do is probably support `strict-dynamic` to the best of our understanding, and note the bug in the documentation (and find/create the corresponding issue on bugzilla).

> What's the best way for me to contribute code to this PR btw, if you want it? Should I just make PRs into the `csp` branch?

I have some work locally that would likely conflict if we were both working in `src` (mostly moving code around in `render.js`, e.g. meta tags need to be prepended _after_ scripts have been generated) — will push it up later when I'm back at my desk. Might be easiest to hold off until then so we don't find ourselves in merge conflict hell. I haven't started thinking about tests though, so that would be super helpful if you get a chance!

Karlinator:
> Yikes. Thanks for investigating, that does indeed sound like a Firefox bug. The best we can do is probably support strict-dynamic to the best of our understanding, and note the bug in the documentation (and find/create the corresponding issue on bugzilla).

I'm trying to narrow down the exact issue with reproduction so I can report it. But yeah, we'll just have to do our best with it, and advise that using `nonce` mode seems more consistent.

>  Might be easiest to hold off until then so we don't find ourselves in merge conflict hell

Sure. I was mostly experimenting, trying to get a flow of compile-time hashing (since the chunk files are completely static).

I might try to look at tests—or I'll just make more abstract contributions (read: find bugs in browsers) some more

Karlinator:
Bugzilla report: https://bugzilla.mozilla.org/show_bug.cgi?id=1751573

Karlinator:
A couple of edge cases with the dev server: 

Vite inserts a whole bunch of inline style tags in the dev server (for perfectly good reasons). The SSR can nonce the style tag it sees as much as it wants; Vite will promptly replace it with un-nonced tags after hydration. Since this only affects dev mode it's probably fine to insert `style-src 'unsafe-inline'`? We also have to remove the nonce from `style-src` then in dev mode otherwise `unsafe-inline` is ignored.

Vite connects over WebSocket in dev mode, so if `default-src` or `connect-src` is specified, it needs to add `connect-src ws://localhost:*` (or `ws://*` if you're hosting over the network?), else Vite can't do hot reloading.

I'd be fine with treating the dev server a little differently and noting it in the docs in case anyone is confused. If you're testing your CSP then you should do it in the adapter's emulator or something anyway.

Rich-Harris:
Gah — tests are all failing because Node only added `webcrypto` in version 15. Ideally we'd use the Web Crypto API because that's what's available in modern environments (Node's `crypto` module is only available in Node itself, of course). Not sure what the best solution is — maybe we should use [SJCL](https://bitwiseshiftleft.github.io/sjcl/) or something for now, and move over to Web Crypto in the future when we can drop support for older Node.

I realised we only need to hash the CSS files for `strict-dynamic`, since the only JS on the page (that Kit controls) is the inline init block, so that's the only thing that needs to be nonced or hashed. (Unless modulepreload links need to be nonced/hashed?) Ideally we'd use a similar mechanism to the `inlineStyleThreshold` mechanism, meaning that we only serve hashes in production when `mode === 'hash'` (_not_ `auto`, since in that case we only use hashes for prerendering, in which case we should be able to read the hashes from the build directory rather than including them in the app), and only for the stylesheets needed by the current manifest (i.e. not one big object of hashes).

That, and the dev mode stuff, should be the last pieces of this PR — will look into it in the morning

Karlinator:
My solution to the crypto stuff before was to outsource (some of) it to the adapter. Any crypto that happens at runtime could be provided by the adapter, which means they can differ according to their platform needs (node <17 etc). Random nonces kind of has to be platform-dependent, right? I don't think SJCL can do those.

One option is to use SJCL for hashing, and have the adapter supply nonces for each request? And then move to WebCrypto when we can drop old node versions (and WebCrypto becomes stable in Node).

You are correct re: modulepreload. The only reason to nonce or hash them is to allow using neither `'strict-dynamic'` nor `'self'`, but as noted elsewhere Firefox only supports this with nonces, not hashes. And that use case seems... slim. We can also hash them just for the `integrity` attribute, but that's not strictly CSP related.

In fact, it seems like hashing external stylesheets is not actually supported in CSP at all, only noncing or domain allowlisting. Additionally, `strict-dynamic` has no effect for stylesheets.

We could nonce the stylesheets if `style-src 'self'` or `style-src-elem 'self'` is not specified, but I think doing that automatically might lead to confusion and CSP errors later. We certainly could support noncing our own `<link rel="stylesheet">` tags with a setting, but needing that is a kind of weird situation to be in, and breaks things like importing other stylesheets.

Of course `%svelte.nonce%` is still useful for allowing only one stylesheet and not any stylesheet on that CDN. But there's no real alternative to nonces in that situation when prerendering.

>  meaning that we only serve hashes in production when mode === 'hash' (not auto, since in that case we only use hashes for prerendering, in which case we should be able to read the hashes from the build directory rather than including them in the app), and only for the stylesheets needed by the current manifest (i.e. not one big object of hashes).

Yes, I think.

Rich-Harris:
I went to the slightly extreme step of adding a sha256 implementation, heavily based on SJCL's. Then, I made it possible to generate nonces with either Web Crypto _or_ the node `crypto` module, depending on which is available (the assumption being that every platform we care about will have one or the other). Then I realised I could have just used the sha256 stuff from web/node without spending the day modernising and customising the SJCL code. Then I decided to keep it anyway, since it turns out it's actually faster (albeit slower than node's `crypto`), and lets us do things synchronously.

This way, adapters don't need to worry about it, and there's no possibility of an adapter using an insecure random number generator for nonces.

Karlinator:
@Rich-Harris Re: windows there are some more paths that break. Doing the same thing there seems to fix it for me on Windows: https://github.com/sveltejs/kit/blob/cb0968e30dbaa0095fd45fd7c35650ff81019b7c/packages/kit/src/core/config/options.js#L95
```javascript
files: object({
	assets: string('static'),
	hooks: string(join('src', 'hooks')),
	lib: string(join('src', 'lib')),
	routes: string(join('src', 'routes')),
	serviceWorker: string(join('src', 'service-worker')),
	template: string(join('src', 'app.html'))
}),
```

Rich-Harris:
Alright, I think this is done. There's some follow-up work to do, as mentioned at the top, but I believe this covers the majority of what we need.

Karlinator:
Ooo, I've found more CSP weirdness.

`'strict-dynamic'` with hashes currently doesn't work in dev mode but does work in production builds (in Chrome at least), because when using hashes with `strict-dynamic` the imported script apparently needs to be preloaded (which it isn't in the dev server). When the inline script is nonced there is no such constraint.

Rich-Harris:
struggling to get it to work in dev _or_ prod with `strict-dynamic`, which is weird because i could have sworn it was working fine earlier. might just need to get some sleep

packages/kit/src/runtime/server/page/csp.js
```diff
@@ -0,0 +1,197 @@
...
+/** @type {Promise<void>} */
+export let csp_ready;
+
+/** @type {() => string} */
+let generate_nonce;
+
+/** @type {(input: string) => string} */
+let generate_hash;
+
+if (typeof crypto !== 'undefined') {
```
dominikg, line 13: Is it safe to assume existing crypto is web-crypto or should it do additional checks?

packages/kit/src/runtime/server/page/csp.js
```diff
@@ -0,0 +1,197 @@
...
+	const array = new Uint8Array(16);
```
dominikg, line 14: Is reusing the array safe and a perf improvement? otherwise might be cleaner to have it inside the function

packages/kit/src/runtime/server/page/csp.js
```diff
@@ -0,0 +1,197 @@
...
+
+	generate_nonce = () => {
+		crypto.getRandomValues(array);
+		return base64(array);
+	};
+
+	generate_hash = sha256;
+} else {
+	const name = 'crypto'; // store in a variable to fool esbuild when adapters bundle kit
```
dominikg, line 23: esbuild evolves and might not be fooled by this forever. could we explicitly tell it to ignore crypto?

packages/kit/src/runtime/server/page/crypto.js
```diff
@@ -0,0 +1,239 @@
+const encoder = new TextEncoder();
```
benmccann, line 1: should we add something at the top of this file like:
```
// TODO: remove this in favor of web crypto API once we no longer support Node 14
```

packages/kit/src/runtime/server/page/csp.js
```diff
@@ -0,0 +1,197 @@
...
+
+		this.script_needs_nonce = this.#script_needs_csp && !this.#use_hashes;
+		this.style_needs_nonce = this.#style_needs_csp && !this.#use_hashes;
+
+		if (this.script_needs_nonce || this.style_needs_nonce) {
+			this.nonce = generate_nonce();
+		}
+	}
+
+	// TODO would be great if these methods weren't async
```
benmccann, line 105: are they async?

packages/kit/src/runtime/server/page/csp.js
Rich-Harris: What would those checks look like? I would assume that if someone is capable of injecting a `crypto` global, they are also capable of passing whatever duck typing checks we have in place. Unless it's accident rather than malice, in which case `crypto.getRandomValues` would fail anyway

packages/kit/src/runtime/server/page/csp.js
Rich-Harris: From https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues:

> All elements in the array are overwritten with random numbers

So it's safe. I would expect the perf impact to be minimal but there's no real reason to keep allocating new memory and garbage collecting it

packages/kit/src/runtime/server/page/csp.js
Rich-Harris: we could, but every adapter that used esbuild would need to do it separately, including ones we don't maintain. it seemed simpler to solve it here. if esbuild's behaviour changed, we'd catch it in CI (that's how i caught this one)

packages/kit/src/runtime/server/page/csp.js
Rich-Harris: they were until we added our own sha256 implementation. removed

packages/kit/src/runtime/server/page/crypto.js
Rich-Harris: yep, done

packages/kit/src/runtime/server/page/csp.js
```diff
@@ -0,0 +1,197 @@
...
+
+if (typeof crypto !== 'undefined') {
+	const array = new Uint8Array(16);
+
+	generate_nonce = () => {
+		crypto.getRandomValues(array);
+		return base64(array);
+	};
+
+	generate_hash = sha256;
```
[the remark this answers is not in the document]
dominikg: fair enough :+1:

Rich-Harris:
I'm going to admit defeat on `strict-dynamic` for now, in the interests of getting this merged. We can add support in a follow-up PR

Issue #3473, opened 2022-01-20
adapter-static can't build initial project


nosovk:
### Describe the bug

Current version of adapter-static broken from [1.0.0-next.24 version](https://github.com/MailCheck-co/mailcheck.site/commit/abecae86f209ffc9abdae02d59611bb406235472)
After that update adapter-static failing to build Kit template.
Current Skeleton App not compatible with svelte kit, or svelte kit is not compatible with static adapter.

### Reproduction

Create new kit project (npm init svelte@next my-app)
install @sveltejs/adapter-static@next
Add it to svelte.config.js as adapter instead of adapter-auto.
Try to build app.

### Logs

```shell
Run npm run preview to preview your production build locally.

> Using @sveltejs/adapter-static
> body used already for: 
    at consumeBody (file:///C:/proj/staticAdapterBroken/node_modules/@sveltejs/kit/dist/install-fetch.js:
4912:9)
    at Response.text (file:///C:/proj/staticAdapterBroken/node_modules/@sveltejs/kit/dist/install-fetch.j
s:4877:24)
    at visit (file:///C:/proj/staticAdapterBroken/node_modules/@sveltejs/kit/dist/chunks/index5.js:483:39
)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

### System Info

```shell
System:
    OS: Windows 10 10.0.19043
    CPU: (8) x64 Intel(R) Core(TM) i5-8350U CPU @ 1.70GHz
    Memory: 13.38 GB / 31.88 GB
  Binaries:
    Node: 16.9.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.15 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 7.22.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (97.0.1072.62)
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.14
    @sveltejs/adapter-static: ^1.0.0-next.26 => 1.0.0-next.26
    @sveltejs/kit: next => 1.0.0-next.236
    svelte: ^3.44.0 => 3.46.2
```

### Severity

blocking all usage of SvelteKit

### Additional Information

[...]

Issue #591, opened 2021-03-22
Cloning of response leads to issues


benmccann:
This line:

https://github.com/sveltejs/kit/blob/8805c6daa6b11a46d2d373c3db890227b27e660d/packages/kit/src/runtime/server/page.js#L128

Causes issues in Node where it dies unexpectedly without any message on server. The Node high water mark limit is much smaller than browser. Can increase high water mark as workaround

There's multiple issues talking about this in `node-fetch` (e.g. [671](https://github.com/node-fetch/node-fetch/pull/671) and [151](https://github.com/node-fetch/node-fetch/issues/151)). Their recommendation is to stream and provide back pressure: https://github.com/node-fetch/node-fetch/issues/151#issuecomment-482777213

Rich-Harris:
If someone can create a repro it would help a lot

benmccann:
There was a comment on Discord that they're only reproducing this if the request goes through Nginx: https://discord.com/channels/457912077277855764/819723698415599626/823759556789272597

stalkerg:
@Rich-Harris yes, self-locking happens here https://github.com/sveltejs/kit/blob/master/packages/kit/src/runtime/server/page.js#L141 

If an HTTP packet contains more than two TCP packages (the first packet is usually the header) and one of these TCP packages for the body did not fit into the write buffer.
Basically, `await clone.text()` waiting for a response from the stream, but this stream waiting drying (reading) buffer by original `response` what can't be returned from the function because we waiting `await clone.text()`.

Basically, `const clone = response.clone();` can work only if all tcp package for HTTP body fit into write buffer OR JS interpretation/compiler/VM smart enough to detect self-locks what NodeJS not.

Rich-Harris:
Is there any way to interpret this other than '`fetch` is fundamentally broken'?

stalkerg:
If you talking about sveltekit fetch - then yes. 
PS one more link https://github.com/whatwg/streams/issues/506#issuecomment-242073695

benmccann:
The way to fix this problem is to resolve the cloned response in parallel with the original. For example:
```
const fetch = require('node-fetch');

const response = await fetch('https://example.com');
const r1 = await response.clone();

const results = await Promise.all([response.json(), r1.text()]);
```

The response in SvelteKit is used in two places:
1. it's serialized and put into the page for hydration
2. it's used in the server-side implementation of `load` as desired by the user

The problem here is that we're trying to do these basically in serial by basically intercepting the response.

[...]

PR #3521, opened 2022-01-24
[fix] reading from same response body twice during prerender (#3473)

Theo-Steiner:

This PR fixes the behavior described in #3473, where building the default template with adapter static fails as during the pre-rendering the same response body was read twice. As @bluwy pointed out in the original issue, this is related to fetch (and accordingly node-fetch) only allowing a single read of a response body and can be circumvented by using the clone() API that is provided for this purpose.
I also wrote a test that fails without this PR, and I think that would be generally useful to have.

### Tests

``test/test.js:953:2 › Load › handles large responses`` fails due to problems discussed in #591

### Changesets


packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -208,7 +208,8 @@ export async function load_node({
 					const proxy = new Proxy(response, {
 						get(response, key, _receiver) {
 							async function text() {
-								const body = await response.text();
+								const clonedResponse = response.clone();
+								const body = await clonedResponse.text();
```
bluwy, line 212:
```suggestion
								const cloned_response = response.clone();
								const body = await cloned_response.text();
```

These should be snake-case for internal variables.

bluwy (commented):
Looks like cloning it has an impact on large responses (probably expected) and caused the tests to fail 🤔

packages/kit/src/runtime/server/page/load_node.js
Theo-Steiner: Fixed, Thanks! 😊

Theo-Steiner:
> Looks like cloning it has an impact on large responses (probably expected) and caused the tests to fail 🤔

 it looks like we bumped into #591 here...

Rich-Harris:
Thank you.

I pushed a commit that changes the behaviour so that the prerenderer gets access to the same buffered body as the code that inlines responses into the HTML (i.e. without cloning the response), though as it was on its way to GitHub I realised that this would fail in the case where the data wasn't buffered during `load`.

I guess we'll need a hybrid approach in which we use the buffered body if it's available, otherwise buffer it at prerender time. Working on it.

Rich-Harris (approved):

Issue #3526, opened 2022-01-24
Syntax error in cjs code for netlify-adapter can break deploys


goynang:
There is an error in the CJS code generated for netlify-adapter where line 6 of `.netlify/handler.js` ends up as...

    var ./server/app.js = require('./server/app.js');

...which is a syntax error and leads to bad deploy unless you force esbuild to be used instead. The CJS version of `handler.js` (before code replacements take place) looks like...

    var APP = require('APP');

...whereas the ESM version looks like...

    import { App } from 'APP';

The code substitution is based on replacing `APP` with `./server/app.js` (lines 53-55 of `adapter-netlify/index.js`) so mistakenly replaces both uses within the require statement (the variable and the path).

It looks like rollup is altering the import from `App` to `APP` when it changes it to a require statement for CJS.

In addition it looks like the adapter is hardcoded to build CJS functions targeted for node 12...

    const output = esbuild.transformSync(input, { format: 'cjs', target: 'node12' }).code;

...rather than something newer or respecting netlify.toml

_Originally posted by @goynang in https://github.com/sveltejs/kit/issues/2687#issuecomment-1019579774_

PR #3528, opened 2022-01-24
Avoid clash for CJS code generation (App vs APP) for #3526

goynang:
Fixes https://github.com/sveltejs/kit/issues/3526

### Tests
- [x] (failing even without my change!)] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets

Fixes problem with CJS code generated for the netlify adapter to avoid syntax error.

### DIFF
--- a/packages/adapter-netlify/rollup.config.js
+++ b/packages/adapter-netlify/rollup.config.js
@@ -19,6 +19,6 @@ export default [
 			}
 		],
 		plugins: [nodeResolve(), commonjs(), json()],
-		external: ['APP', ...require('module').builtinModules]
+		external: ['App', ...require('module').builtinModules]
 	}
 ];


benmccann:
@goynang a couple suggestions for next time - if there's an open issue, please link to it. also put an "x" instead of a check in the boxes so that github formats it properly. I've edited your PR description to make both these changes

goynang:
@benmccann - thanks for that and duly noted.

Whilst I'm convinced the syntax error and Netlify deploy problem comes from the generation of the invalid CJS files... my fix doesn't work.

It looks like the rollup process is altering the name of the variable from `App` to `APP` which then triggers the problem where the string 'APP' is mistakenly substituted with the path to '.server/app.js' generating the syntax error.

I'm not familiar with rollup but I assume it does this as 'APP' is treated as an external lib and so it infers the name from it when it generates the CJS require statement (i.e. it guesses 'APP' rather than the 'App' the original import statement uses).

I'll dig a bit further but will likely delete this PR and let someone that knows rollup better than myself tackle this. I was hoping this might be my first contribution to Kit but I'll try again in future!

Rich-Harris:
closing in favour of #3546 — thanks!

PR #3540, opened 2022-01-25
[chore] remove TODO

benmccann:
It looks like we're now using the browser `fetch` types instead of the `node-fetch` types so there's no longer a type checking warning. I'm not exactly sure how `install-fetch` doesn't run into an issue with the types being different or maybe it does and that's ignored elsewhere, but at least in this portion of the code there no longer seems to be an issue


Rich-Harris (approved):

Rich-Harris:
`node-fetch` implements the fetch API, so the types match by design (there are some edge cases, e.g. it uses node streams instead of web streams to represent bodies, and `Headers` has a `raw` method, but for it's better to use standard types everywhere anyway)

PR #3541, opened 2022-01-26
[chore] remove `InternalHandle`

benmccann:
We were setting the type of the user-provided `handle` function as `InternalHandle`, which is not correct and was making it ward to work with

### DIFF
--- a/.changeset/tender-buses-talk.md
+++ b/.changeset/tender-buses-talk.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[chore] remove InternalHandle
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -122,7 +122,9 @@ export async function respond(request, options, state = {}) {
 				let decoded = decodeURI(event.url.pathname);
 
 				if (options.paths.base) {
-					if (!decoded.startsWith(options.paths.base)) return;
+					if (!decoded.startsWith(options.paths.base)) {
+						return new Response(undefined, { status: 404 });
+					}
 					decoded = decoded.slice(options.paths.base.length) || '/';
 				}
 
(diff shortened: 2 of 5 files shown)


Rich-Harris (approved):

Rich-Harris (changes requested):
actually, looking at the code, i think this _does_ exist for a reason. lemme come back to this

benmccann:
my motivation for this was to move AMP validation to a hook, but it depended on this. I pushed a branch for that here: https://github.com/sveltejs/kit/tree/amp-hook

I couldn't figure out how to get the types to work otherwise. I could probably do it by casting the types and/or using `@ts-expect-error`, but it felt rather ugly to try that

Rich-Harris:
so, here's the problem. if a lambda is rendering a page `foo` that depends on endpoint `bar`, and the code for `bar` isn't in the lambda, then instead of returning a 404 response, `respond` needs to signal that it needs to fall back to a real HTTP request. It only guarantees to generate a response if the request came from the user, rather than from `load`: https://github.com/sveltejs/kit/blob/1474aee74338ef80cce355ae851377fa5a903f5c/packages/kit/src/runtime/server/index.js#L176-L189

at the same time, this does mean that the `handle` signature is wrong. since there's no guarantee that `resolve` will return something, it'd be possible to write buggy code:

```js
  const response = resolve(event);
  response.headers.set('x-vegetable', 'potato'); // errors — response is undefined
  return response;
}
```

i _think_ that's the only place where we're using the lack of a response that way (though at the same time i'm pretty sure the distinction between `Handle` and `InternalHandle` predates the server-side code-splitting stuff, so maybe not?). Perhaps we could solve it some other way, e.g. a special header on the response, or a single `Response` object that both sides have a reference to and that could be used as a sentinel value.

benmccann:
Ok. I think I've fixed that by moving where the real HTTP request occurs

Rich-Harris (approved):

Rich-Harris:
oh, nice — nailed it. this is way better

PR #3546, opened 2022-01-26
give module id an invalid name to prevent invalid replacement

### LINKED ISSUE #3526: Syntax error in cjs code for netlify-adapter can break deploys

Rich-Harris:
fixes #3526. #3528 was close, but still had the bug — in order to fix it we need the module ID to be something that Rollup can't (in its desire to generate readable code) turn into an identical variable name. We can't use something like `{APP}` because that seems to cause issues with string replacement (possibly word boundary related? couldn't quite be bothered to figure it out), but if we prefix the ID with a number than Rollup is forced to add another character, thus hacking around the issue.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-netlify/index.js
+++ b/packages/adapter-netlify/index.js
@@ -51,7 +51,7 @@ export default function ({ split = false } = {}) {
 			const redirects = [];
 
 			const replace = {
-				APP: './server/app.js'
+				'0APP': './server/app.js'
 			};
 
 			if (esm) {
--- a/packages/adapter-netlify/rollup.config.js
+++ b/packages/adapter-netlify/rollup.config.js
@@ -19,6 +19,7 @@ export default [
 			}
 		],
 		plugins: [nodeResolve(), commonjs(), json()],
-		external: ['APP', ...require('module').builtinModules]
+		external: (id) => id === '0APP' || id.startsWith('node:'),
+		preserveEntrySignatures: true
 	}
 ];
--- a/packages/adapter-netlify/src/handler.d.ts
+++ b/packages/adapter-netlify/src/handler.d.ts
@@ -1,3 +1,3 @@
-declare module 'APP' {
+declare module '0APP' {
 	export { App } from '@sveltejs/kit';
 }
--- a/packages/adapter-netlify/src/handler.js
+++ b/packages/adapter-netlify/src/handler.js
@@ -1,5 +1,5 @@
 import './shims';
-import { App } from 'APP';
+import { App } from '0APP';
 
 /**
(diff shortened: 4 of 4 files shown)


packages/adapter-netlify/index.js
```diff
@@ -51,7 +51,7 @@ export default function ({ split = false } = {}) {
 			const redirects = [];
 
 			const replace = {
-				APP: './server/app.js'
+				'0APP': './server/app.js'
```
benmccann, line 54: we should probably add a comment about why we're prefixing with `0`

Rich-Harris, line 54:
```suggestion
				'0APP': './server/app.js' // digit prefix prevents CJS build from using this as a variable name
```

packages/adapter-netlify/index.js
```diff
@@ -51,7 +51,7 @@ export default function ({ split = false } = {}) {
 			const redirects = [];
 
 			const replace = {
-				APP: './server/app.js'
+				'0APP': './server/app.js' // digit prefix prevents CJS build from using this as a variable name
```
benmccann, line 54: i might add a bit more detail
```suggestion
				'0APP': './server/app.js' // digit prefix prevents CJS build from using this as a variable name to ensure replace doesn't replace occurrences of variables with this string
```

benmccann (approved):

packages/adapter-netlify/index.js
Rich-Harris, line 54:
```suggestion
				'0APP': './server/app.js' // digit prefix prevents CJS build from using this as a variable name, which would also get replaced
```

Issue #3330, opened 2022-01-13
Selectively Disable AMP Validation


john-michael-murphy:
### Describe the problem

Sveltekit’s AMP validation isn’t always correct. Take this example for, instance: 

```javascript // hooks.js

export async function handle({ request, resolve }) {
	rendered.body = addDynamicallyGeneratedAmpShell(rendered.body)
	return rendered;
}
```

In this scenario, the amp validator will return a bunch or errors (e.g. `The mandatory attribute '⚡' is missing in tag 'html’.`) even though the final rendered body is amp valid. 

### Describe the proposed solution

Perhaps instead of making `amp` a boolean, we could make it an object that accepts options. I’d be happy to open a PR to add this feature if that’d be helpful. 

### Importance

would make my life easier

PR #3554, opened 2022-01-26
[chore] refactor AMP validation

benmccann:
Make the AMP validation happen in a `handle` hook. This should fix https://github.com/sveltejs/kit/issues/3330 (it won't allow disabling it, but it will make it so that it's correct) and be a step towards https://github.com/sveltejs/kit/issues/2603. It wouldn't be hard to move this to an external package for users to call. However, the actual AMP HTML generation is pretty tied into SvelteKit and I'm not sure how we'd change that

### DIFF
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -12,6 +12,7 @@ import { SVELTE_KIT, SVELTE_KIT_ASSETS } from '../constants.js';
 import { get_mime_lookup, resolve_entry, runtime } from '../utils.js';
 import { coalesce_to_error } from '../../utils/error.js';
 import { load_template } from '../config/index.js';
+import { sequence } from '../../hooks.js';
 
 /**
  * @param {import('types/config').ValidatedConfig} config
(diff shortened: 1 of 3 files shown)


packages/kit/src/core/dev/plugin.js
```diff
@@ -165,10 +166,12 @@ export async function create_plugin(config, cwd) {
 							? await vite.ssrLoadModule(`/${config.kit.files.hooks}`)
 							: {};
 
+						const handle = user_hooks.handle || (({ event, resolve }) => resolve(event));
+
 						/** @type {import('types/internal').Hooks} */
 						const hooks = {
 							getSession: user_hooks.getSession || (() => ({})),
-							handle: user_hooks.handle || (({ event, resolve }) => resolve(event)),
+							handle: amp ? sequence(handle, amp) : handle,
```
Rich-Harris, line 174: i think these need to be reversed

```suggestion
							handle: amp ? sequence(amp, handle) : handle,
```

Rich-Harris (changes requested):
it might be worth adding a test for this somehow?

packages/kit/src/core/dev/plugin.js
benmccann: I believe they're applied left-to-right. Wouldn't we want the amp validator to run last? I.e. after the HTML is completely generated make sure it's valid

benmccann:
We do have amp tests already in `packages/kit/test/apps/amp` that test that the validator works correctly. This is just moving code rather than adding anything new, so I don't necessarily think we need new tests

packages/kit/src/core/dev/plugin.js
Rich-Harris: they're not applied left-to-right _or_ right-to-left. it just applies the first in the sequence, which is responsible for calling the second (which is responsible for calling the third, etc). so `amp` needs to call the user's `handle`, and inspect the result of calling that

benmccann: done

Rich-Harris (approved):

Issue #3436, opened 2022-01-20
No support for async vite config function


ispyinternet:
### Describe the bug

In `svelte.config.js` should it be possible to provide an aysnc function that returns the vite config?

### Reproduction

https://github.com/ispyinternet/sveltekit-async-vite-config

### System Info

```shell
System:
    OS: Linux 4.14 Amazon Linux 2
    CPU: (2) x64 Intel(R) Xeon(R) CPU E5-2665 0 @ 2.40GHz
    Memory: 434.50 MB / 16.73 GB
    Container: Yes
    Shell: 4.2.46 - /bin/bash
  Binaries:
    Node: 14.18.3 - /usr/bin/node
    npm: 7.5.3 - ~/.npm-global/bin/npm
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.13 
    @sveltejs/kit: next => 1.0.0-next.234 
    svelte: ^3.44.0 => 3.46.2
```

### Severity

annoyance

bluwy:
Similar #3077

PR #3565, opened 2022-01-27
Support an async function to provide vite config

ispyinternet:
fixes https://github.com/sveltejs/kit/issues/3436

### DIFF
--- a/.changeset/sour-shirts-camp.md
+++ b/.changeset/sour-shirts-camp.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Allow config.kit.vite to be an async function
--- a/packages/kit/src/core/build/build_client.js
+++ b/packages/kit/src/core/build/build_client.js
@@ -60,7 +60,7 @@ export async function build_client({
 	});
 
 	/** @type {[any, string[]]} */
-	const [merged_config, conflicts] = deep_merge(config.kit.vite(), {
+	const [merged_config, conflicts] = deep_merge(await config.kit.vite(), {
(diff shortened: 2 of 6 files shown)


ispyinternet:
At least lets me inject some async config into dev mode. Not sure about preview: https://github.com/sveltejs/kit/blob/master/packages/kit/src/core/preview/index.js#L72

benmccann:
You need to run `pnpm format`

dominikg:
It would be good to update the docs and config type to show that an async function can be used too. 

Please also add a changeset.

sidenote:
Got confused at first because your PR never handles the vite config as object case, but it looks like the validator turns an object into `()=>object`. I didn't expect that but makes sense to have it in a single place

Rich-Harris:
This seems like a harmless change so I'm 👍 , though I'm curious what the use case is? You can use top-level await in your `svelte.config.js`, so it seems like anything enabled by this change _should_ already be possible

ispyinternet:
Ah can I? I tried to export an async function that returned a svelte.config.is object but that didn't work.

Basically I have async config that I inject into the application code (typescript files not .svelte so I can't use svelte preprocess). I'm using rollup/plugin-replace via vite 🤷

Rich-Harris:
yep, you can do this:

```js
export default {
  kit: {
    vite: {
      plugins: [
        replace(await load_replacements());
      ]
    }
  }
};
```

given that, do we still need this change?

Rich-Harris (approved):

ispyinternet:
Sorry no reply @Rich-Harris tied up elsewhere!

PR #3576, opened 2022-01-27
Exclude routes

Rich-Harris:
For some reason the `gh pr checkout 2932` workflow failed — I can't push to the branch in #2932 — so I've created this branch to continue work on it. Thanks to @tony19 for making this happen. 

For simplicity, and consistency with the `kit.package.files`, `kit.package.exports` and `kit.serviceWorker.files` options, we should only allow a function, rather than an array of strings and regexes and functions. The question is whether to make it a `(filepath: string) => boolean` function, or keep the `basename` argument and either a) introduce inconsistency, or b) impose a breaking change on those other options.

The breaking change would be unfortunate because at the moment it's very easy to use with globbing libraries...

```js
import mm from 'micromatch';

export default {
  kit: {
    package: {
      files: mm.matcher('!**/build.*')
    }
  }
};
```

...so I'm inclined to go with `(filepath: string) => boolean`, since it's trivial to derive the basename.

I also think `kit.excludes` is probably a bit too mysterious. Perhaps `kit.routes` (or `kit.routing.include`, though we don't currently have anything else to go under `routing`) would be better, though at that point we'd have `kit.routes` and `kit.router`, which makes me want to put `kit.router` and `kit.hydrate` under `kit.browser` or something.

### Tests

### Changesets


documentation/docs/14-configuration.md
```diff
@@ -123,6 +126,35 @@ An object containing zero or more of the following `string` values:
...
+
+Each array item is one of the following:
+
+ * `String` - The string is matched against the file's basename, and the file is excluded if it matches.
+ * `RegExp` - The regular expression is tested against the file's basename, and the file is excluded if it matches.
+ * `({ filepath: string, basename: string }) => boolean` - The function receives the file's path and basename, and returns `true` to exclude the file.
+
+Double-underscore prefixes (e.g., `__foo.svelte`) are reserved for special files, and thus cannot be excluded.
+
+**Default:** `[/^[_.]/]` *(exclude files that begin with `_` or `.`)*
```
dreitzner, line 141: Does not match pattern in line 36

dreitzner (changes requested):

ignatiusmb (approved):

autr:
Whatever happened to this? None of these options seem available in V1 releases.

https://github.com/sveltejs/kit/issues/2930

PR #3580, opened 2022-01-28
[docs] add packaging guidelines

benmccann:
I've spent a lot of time fixing various packages that people have run into. At first I was doing this because most of the time it was a Vite issue. However, now it's usually a packaging issue. I'd love to have some docs to point people at because bad packaging is a commonly-made mistake

### DIFF
--- a/documentation/faq/70-packages.md
+++ b/documentation/faq/70-packages.md
@@ -2,10 +2,14 @@
 question: How do I fix the error I'm getting trying to include a package?
 ---
 
-SSR in Vite is not yet stable. Libraries work best with Vite when they distribute both CJS and ESM in their package and you may wish to work with library authors to make this happen.
+Vite's SSR support has become fairly stable since Vite 2.7. Most issues related to including a library are due to incorrect packaging.
 
-Svelte components must be written entirely in ESM. It is encouraged to make sure the dependencies of external Svelte components provide an ESM version. However, in order to handle CJS dependencies [`vite-plugin-svelte` will look for any CJS dependencies of external Svelte components](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#what-is-going-on-with-vite-and-pre-bundling-dependencies) and ask Vite to pre-bundle them by automatically adding them to Vite's `optimizeDeps.include` which will use `esbuild` to convert them to ESM.
+Libraries work best with Vite when they distribute an ESM version and you may wish to suggest this library authors. Here's a few things to keep in mind when checking if a library is packaged correctly:
+- `exports` takes precedence over the other fields like `main`, `module`, and `browser`. Adding an `exports` field may not be backwards-compatible as it prevents deep imports.
+- `main` should be defined if `exports` is not. It should be a CommonJS file unless `"type": "module"` is set, in which case it should be an ESM file. If a `module` field is defined, it should refer to an ESM file.
+- ESM files should end with `.mjs` unless `"type": "module"` is set in which any case CommonJS files should end with `.cjs`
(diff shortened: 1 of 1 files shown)


dummdidumm (approved):
These sound like good guidelines

documentation/faq/70-packages.md
```diff
@@ -2,10 +2,14 @@
 question: How do I fix the error I'm getting trying to include a package?
 ---
 
-SSR in Vite is not yet stable. Libraries work best with Vite when they distribute both CJS and ESM in their package and you may wish to work with library authors to make this happen.
+Vite's SSR support has become fairly stable since Vite 2.7. Most issues related to including a library are due to incorrect packaging.
 
-Svelte components must be written entirely in ESM. It is encouraged to make sure the dependencies of external Svelte components provide an ESM version. However, in order to handle CJS dependencies [`vite-plugin-svelte` will look for any CJS dependencies of external Svelte components](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#what-is-going-on-with-vite-and-pre-bundling-dependencies) and ask Vite to pre-bundle them by automatically adding them to Vite's `optimizeDeps.include` which will use `esbuild` to convert them to ESM.
+Libraries work best with Vite when they distribute an ESM version and you may wish to suggest this library authors. Here's a few things to keep in mind when checking if a library is packaged correctly:
```
bluwy, line 7:
```suggestion
Libraries work best with Vite when they distribute an ESM version and you may wish to suggest this to library authors. Here are a few things to keep in mind when checking if a library is packaged correctly:
```

documentation/faq/70-packages.md
```diff
@@ -2,10 +2,14 @@
...
+- `exports` takes precedence over the other fields like `main`, `module`, and `browser`. Adding an `exports` field may not be backwards-compatible as it prevents deep imports.
+- `main` should be defined if `exports` is not. It should be a CommonJS file unless `"type": "module"` is set, in which case it should be an ESM file. If a `module` field is defined, it should refer to an ESM file.
```
bluwy, line 9: If `"type": "module"` is set, should `main` always be an ESM file? I think it can be CJS too, depending on the library output. But either ways, both would prevent the opposite format from loading, which is probably intentional by Node.

The only way for dual format packages per [node](https://nodejs.org/api/packages.html#dual-commonjses-module-packages) seems to rely on `exports`

arxpoetica (approved):
Hooray, this is much needed documentation!

documentation/faq/70-packages.md
antony: I think older node doesn't know about exports, so you need main to point to the cjs file to create a hybrid package.

The even more fun part is that if you leave the main attribute populated and then try to use the package with Sapper (SSR), it won't work, even on modern node. So I found myself deleting the main field entirely.

documentation/faq/70-packages.md
```diff
@@ -2,10 +2,14 @@
...
+- ESM files should end with `.mjs` unless `"type": "module"` is set in which any case CommonJS files should end with `.cjs`
+- Svelte components should be distributed entirely as ESM and have a `svelte` field defining the entry point
```
antony, line 11: I thought we were planning to deprecated the `svelte` field. I remember rich talking about it.

benmccann: there was some discussion in the Kit repo (e.g. could it be a boolean rather than a path), but no decisions were made about changing anything yet, and so I think for now it's still the recommended course of action to have it

documentation/faq/70-packages.md
benmccann: Yeah, this was worded slightly more restrictive than actually necessary. I've updated it so that it should be accurate now

PR #3586, opened 2022-01-28
cleanup

antony:

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/templates/default/package.template.json
+++ b/packages/create-svelte/templates/default/package.template.json
@@ -1,6 +1,7 @@
 {
 	"name": "~TODO~",
 	"version": "0.0.1",
+	"private": true,
 	"scripts": {
 		"dev": "svelte-kit dev",
 		"build": "svelte-kit build",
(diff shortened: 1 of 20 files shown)


packages/create-svelte/templates/default/package.template.json
```diff
@@ -1,6 +1,7 @@
 {
 	"name": "~TODO~",
 	"version": "0.0.1",
+	"private": true,
```
dominikg, line 4: should we do this in a separate PR ? .template.json won't be found by changesets

packages/create-svelte/templates/skeleton/package.template.json
```diff
@@ -1,6 +1,7 @@
 {
 	"name": "~TODO~",
 	"version": "0.0.1",
+	"private": true,
```
dominikg, line 4: see above

packages/create-svelte/templates/default/package.template.json
antony: good point, I hadn't spotted this. Agreed.

dominikg (approved):

PR #3612, opened 2022-01-28
ci: update template repo from release action every time create-svelte…

dominikg:
… was released

### Tests

### Changesets


dominikg:
word of warning, this is totally untested but in theory it should work. :see_no_evil:

Rich-Harris:
recapping discord discussion: personally i think we should keep things as they are, since this is really just a slightly more indirect version of the postpublish script we already have. also, it took us over a week to get the damn thing working at all and i'm too scarred to risk changes

dominikg:
if it is working with `changeset release` running postpublish script we don't need this.

Issue #3428, opened 2022-01-20
Distributing multiple copies of same file


benmccann:
### Describe the bug

`node_modules/@sveltejs/kit/dist/chunks/index.js` and `node_modules/@sveltejs/kit/dist/assets/index.js` seem to have quite a lot of duplicated content in them. It makes debugging a pain because I don't know which line of code is actually being run and it makes installing the package take longer than necessary

### Reproduction

Run `npm init svelte @next` and look in `node_modules/@sveltejs/kit`

### System Info

```shell
any
```

### Severity

annoyance

JEbertPrime:
I might be making a mistake setting up, but the second path you provided (`node_modules/@sveltejs/kit/dist/assets/index.js`) seems to not be present in a normal install. Wondering if you mean `node_module/@sveltejs/kit/assets/server/index.js` which seems to have quit a bit of replicated code

benmccann:
The paths seem to have changed a bit though there's still some duplication present. E.g. if I do `grep -r "ignore W/ prefix"` I see that line in both of the following files:
```
./dist/chunks/index.js
./assets/server/index.js
```

JEbertPrime:
The duplicated code is a result of the `src/runtime/server/index.js` module not being properly externalized during the build process. When the rollup config is updated to externalize that import, there's no duplication. Tests are passing for this change. #3628

JEbertPrime:
Windows test `url-supplied anchor is ignored with onMount() scrolling on navigation to page ` fails. Unsure about why, but it appears(?) that this test fails regularly. 
Some thoughts about this solution:
- It muddies up the rollup config, which probably won't make maintenance easier
- My solution works but is kinda gross. Having a regex match in the config is probably less than ideal
- There is probably a reason why there aren't any imports between dist and assets folders. FWIW this import (from `src/core/dev/plugin.js`) seems to be the only one that references a module in the `runtime` folder (as evidenced by `grep -r '^import.*runtime'`) so this probably doesn't need to be a generalizable solution

_Originally posted by @JEbertPrime in https://github.com/sveltejs/kit/issues/3628#issuecomment-1025802097_

PR #3628, opened 2022-01-29
[fix] Minor change to rollup config to prevent duplication - issue #3428

JEbertPrime:
Externalized a dependency to reduce code duplication in `dist/chunks/index.js`

### Tests

### Changesets


JEbertPrime:
Windows test failed bc path matching didn't account for differences in paths between OSs. Otherwise tests pass, although some skip. At least locally (ubuntu 20.04) it seems a similar amount skip as on main branch

JEbertPrime:
Windows test `url-supplied anchor is ignored with onMount() scrolling on navigation to page ` fails. Unsure about why, but it appears(?) that this test fails regularly. 
Some thoughts about this solution:
- It muddies up the rollup config, which probably won't make maintenance easier
- My solution works but is kinda gross. Having a regex match in the config is probably less than ideal
- There is probably a reason why there aren't any imports between dist and assets folders. FWIW this import (from `src/core/dev/plugin.js`) seems to be the only one that references a module in the `runtime` folder (as evidenced by `grep -r '^import.*runtime'`) so this probably doesn't need to be a generalizable solution

Rich-Harris:
Thank you. I changed the implementation a bit to make it clearer that the built version of the dev plugin is importing from an external file — this should hopefully make it a bit easier to diagnose if we end up moving files around

benmccann:
What's the reason for externalizing this file vs removing the assets server bundle below? (the answer might be good to add as a code comment in the rollup config)

```
	{
		input: 'src/runtime/server/index.js',
		output: {
			format: 'esm',
			file: 'assets/server/index.js'
		},
		plugins: [
			resolve({
				extensions: ['.mjs', '.js', '.ts']
			})
		]
	},
```

benmccann:
I'm also slightly confused what `${runtime}` is and why we don't use it here, but do use it elsewhere. E.g. I see:
```
node_modules/@sveltejs/kit/dist/chunks/index3.js:import { set_prerendering } from '${runtime}/env.js';
```

Rich-Harris:
Good point, we can use the existing conditional import mechanism. Updated the PR.

> What's the reason for externalizing this file vs removing the assets server bundle below?

`assets` contains code that apps import, such as code for SSR. If we removed it, apps wouldn't work any more, you would only be able to `svelte-kit dev`

benmccann (approved):
awesome! thanks @JEbertPrime and @Rich-Harris for the help on this!

PR #3629, opened 2022-01-30
[fix] navigation `scrollTop` reset

### LINKED ISSUE #2733: Page scroll position not reset to top on navigation (regression)

nhe23:
Fixes #2733 by also scrolling the first child-element of #target that isn't scrolled to the top. 
Currently if something like this
```js
<div style="max-height: 100vh; overflow-y: auto;">
	<main>
		<slot />
	</main>
	<Footer />
</div>
```
is inside of a __layout component the page is not scrolled all the way to the top when using client-side navigation. One example where this happens is when a [DaisyUI Drawer Component](https://github.com/saadeghi/daisyui/issues/431#issuecomment-1023514827) is used. 

### Tests

### Changesets


Rich-Harris:
Thank you — I don't yet 100% understand what this is doing 😅 but will try and wrap my head round it using the provided test. In the meantime though it's probably a good idea to hold off until @nikosgpet is able to provide a repro for https://github.com/sveltejs/kit/issues/2733#issuecomment-1026080056 and we can verify that it's fixed

nhe23:
There were two use cases in which the page was not scrolled to the top using client side navigation. 
1. 'Overflowing' container inside of __layout.svelte (see PR description) -> fixed by scrolling first element inside of layout that is not scrolled up to the top (hence the overflowing container)
2.  'Overflowing' body (see [#2733 (comment)](https://github.com/sveltejs/kit/issues/2733#issuecomment-1028302825)) -> fixed by scrolling document.body if scrollTop is > 0

There still might be other use cases in which the scrolling does not work ... however we cannot fix what we don't know 😄

packages/kit/src/runtime/client/renderer.js
```diff
@@ -384,6 +384,16 @@ export class Renderer {
 					deep_linked.scrollIntoView();
 				} else {
 					scrollTo(0, 0);
+					if (document.body.scrollTop > 0) document.body.scrollTo(0, 0);
```
benmccann, line 387: can you add a comment explaining how this line is different than the preceding one?

packages/kit/src/runtime/client/renderer.js
```diff
@@ -384,6 +384,16 @@ export class Renderer {
...
+					// In some cases it is needed scroll first root element of target that
```
benmccann, line 388: I don't understand this comment. Can you expand it to give an example of such a case?

benmccann:
Ok, I ran the new tests without the fixes and understand a bit better what is happening now:
In the body case, the layout adds a scrollbar to body and when you navigate it replaces the page but leaves the body / layout scrolled
The other case is almost the same. The layout wraps the page in a div with a scrollbar and when you navigate that div isn't scrolled back up to the top

packages/kit/src/runtime/client/renderer.js
```diff
@@ -384,6 +384,16 @@ export class Renderer {
...
+					// is not scrolled to the top
+					for (const target_child of this.target.childNodes) {
+						const target_child_element = target_child.firstChild?.parentElement;
+						if (target_child_element && target_child_element.scrollTop > 0) {
+							target_child_element.scrollTo(0, 0);
+							break;
```
benmccann, line 394: I'm not sure it's always safe to `break` here. What if you had one overflowed div inside another?

packages/kit/src/runtime/client/renderer.js
benmccann, line 387: could we remove this line if we also remove the `break`?

packages/kit/src/runtime/client/renderer.js
```diff
@@ -384,6 +384,16 @@ export class Renderer {
 					deep_linked.scrollIntoView();
 				} else {
 					scrollTo(0, 0);
+					if (document.body.scrollTop > 0) document.body.scrollTo(0, 0);
+					// In some cases it is needed scroll first root element of target that
+					// is not scrolled to the top
```
benmccann, line 389:
```suggestion
					// If there is an element with the overflow CSS property it may have a scrollbar
					// We need to ensure the scroll is reset on these elements
```

packages/kit/test/apps/basics/test/test.js
```diff
@@ -234,6 +234,18 @@ test.describe('Scrolling', () => {
 		expect(await in_view('#input')).toBe(true);
 		expect(await page.$eval('#input', (el) => el === document.activeElement)).toBe(true);
 	});
+
+	test.only('scroll to top works with overflowing root container', async ({ page, clicknav }) => {
+		await page.goto('/scroll-top/overflow');
```
benmccann, line 239: I feel like `scroll-top` isn't super descriptive about what this is testing because we have other scroll tests that test scrolling to the top. How about something like `scroll-overflowed`?

packages/kit/src/runtime/client/renderer.js
```diff
@@ -384,6 +384,16 @@ export class Renderer {
...
+					for (const target_child of this.target.childNodes) {
+						const target_child_element = target_child.firstChild?.parentElement;
```
benmccann, line 391: why do you get the parent of the child? wouldn't that just give you back the original element?

benmccann:
Thanks for your patience on this one while we fixed the other scroll handling issues.

Handling the `body` case here seems fairly reasonable. The `div` case though seems trickier because the document could take any form and I'm not sure how we'd know which divs need to be scrolled. I somewhat think that the answer might be to simply say that is the user's responsibility. I'm curious what @Rich-Harris's opinion about this would be

packages/kit/src/runtime/client/renderer.js
nhe23: I think I tested this exact usecase and it still worked. Elements that are siblings to the first overflown element don't matter as they cannot overflow more than the first overflown element. Overflowing elements inside the first overflown element also don't matter as they would only scroll inside their parent.
However it could be argued that after a page navigation every scroll position in the DOM should be at 0 because that's the "normal" behavior. But this could be very inefficient if there are many elements in the DOM. That's why I thought breaking as early as possible would be the best solution.

packages/kit/src/runtime/client/renderer.js
nhe23: target_child is the node and not the element itself. As element nodes must have at least one child in order to be an element-node (the node could also be just text) I implicitly check this by selecting the parent of the child. Also the node object does not have the same properties that an element object has. I'm happy to change the implementation if there is a better way 😄

Rich-Harris: you can use [`this.target.children`](https://developer.mozilla.org/en-US/docs/Web/API/Element/children), which only includes elements

Rich-Harris:
Ok, I've finally had a chance to try and get my head round this PR.

Unfortunately I just don't think it can work. Because it restricts the scroll reset to direct children of the target element, and breaks when we first encounter a scrolled element, it fails when there are multiple scrolled elements or if there's an additional wrapper element anywhere. [See this demo for an example](https://stackblitz.com/edit/sveltejs-kit-template-default-2va9e7?file=src%2Froutes%2Findex.svelte&terminal=dev) — try scrolling both columns to the bottom, then clicking the button.

There's another tricky problem we haven't yet solved: restoring the scroll positions of non-root elements on `popstate`.

I'm not sure what a working solution _does_ look like; it might be worth seeing how/if other frameworks tackle this case.

PH4NTOMiki:
@Rich-Harris I know that Next.js and Remix don't handle anything other than body/html, and I just checked Chrome doesn't handle it as well, so I don't think that this is a issue framework needs solving.

Rich-Harris:
> It's probably a good idea to remove the 'scroll first overflow element of target'-thing and just scroll the body

Is that different to what Kit currently does? I'm not 100% sure I understand

> We could then add a FAQ that explains what behaviour prevents scrolling and what the user can do to solve this. Thoughts?

Right now I'm not sure there _is_ much that a user could do. It would be good to brainstorm some possible solutions

PH4NTOMiki:
I just checked and I was right, Next, Remix and Nuxt all don't handle scroll to 0 for non-root elements on navigation, 

But one way we can at least partially solve it is to do something similar to this PR but remove break, and use top-level-components(like Layout) to limit the scope of the scroll elements, because statistically the most elements are going to be in the lower-level Slot components than top-level ones. 

The second option is to write it in FAQ/docs because user knows what elements are scrollable and can account(set scrollTop of those elements) for those elements in `beforeNavigate` like for ex. I use for expanding navbar, I have a isNavbarOpen boolean and in `beforeNavigate` I set it to `false`.

mrkishi:
I also don't think it makes sense to tackle this automatically. If I were to intentionally add an overflowing element, there would be a reason behind it and it might be incompatible with some automatic scrolling.

However, it might be worth considering whether we could give `before/afterNavigate` access to custom user state from the history's session storage. That way users would at least be able to behave a little more consistently should they wish to.

Rich-Harris:
I think providing access to history state is the most promising solution, yeah. If we do a good job of designing it, we could address https://github.com/sveltejs/kit/issues/3236#issuecomment-1011436921 at the same time.

I'm thinking along these lines:

```svelte
<script>
  import { beforeNavigate, afterNavigate } from '$app/navigation';

  let sidebar;

  beforeNavigate(({ state }) => {
    state.sidebar_scroll = sidebar.scrollTop;
  });

  afterNavigate(({ type, state }) => {
    if (type === 'popstate') {
      sidebar.scrollTop = state.sidebar_scroll;
    }
  });
</script>

<aside bind:this={sidebar}>
  <!-- ... -->
</aside>

<style>
  aside {
    overflow-y: auto;
  }
</style>
```

PH4NTOMiki:
> > But one way we can at least partially solve it is to do something similar to this PR but remove break
> 
> I definitely understand the appeal, but something I've learned the hard way as an OSS maintainer is that it's better for something to remain broken than to _appear_ fixed

I understand and agree, but I did some testing(basic stuff just to see the stats) like this(it was conducted on a page with 3405 DOM nodes, so pretty large):
```js
const domNodes = Array.from(document.querySelectorAll('html, body, body *')); // length is 3405
```
and then run this many times:
```js
console.time();
domNodes.forEach(node => node.scrollTo(0, 0));
console.timeEnd();
```
and the times came at 12ms most of the times, at most 15ms
so even if we traverse the whole tree and do scrollTo we don't get much impact, and most web apps don't have that many nodes. so it should be much less time, also if I added a check for `node.scrollTop>0` it added 2-5ms to the results so if we decide to do this we should just skip checking it and do scrollTo for every node.

mrkishi:
It might be very performant, but it's still implementation specific. By the way, if we were to scroll them all to the top, I'd also expect us to save and restore their positions on history navigation—but how would we correctly identify the elements?

Of note, even Chromium browsers don't save and scroll all overflown elements on history navigation, and they are operating on a level where uniquely identifying elements isn't an issue..

With that in mind, I propose we:
- Provide the custom history state API on `before/afterNavigate`; this could be used not only for scrolling but other navigation page-specific state, so it's a win all-around imo.
- Introduce a `sveltekit:track` [1] attribute: this attribute would give elements supposed to have scrolling handled by SvelteKit a unique id, `sveltekit:track="sidebar"`. This would mean we could target only the children _intended_ to be scrolled, and also save their positions to history consistently.

[1] 🐐 `sveltekit:scroll`? I don't know what to call it; `track` isn't very meaningful, but `scroll` could cause confusion with anchors' `noscroll`.

PH4NTOMiki:
> Of note, even Chromium browsers don't save and scroll all overflown elements on history navigation, and they are operating on a level where uniquely identifying elements isn't an issue..

Yeah, I wrote that here https://github.com/sveltejs/kit/pull/3629#issuecomment-1057191975 
We just shouldn't even bother with that, Chrome and every other framework don't do it,
but scrolling to 0 should be something we should do as a framework if it is possible.

Issue #3503, opened 2022-01-22
Ability to include credentials on external server-side fetch() calls


johnnysprinkles:
### Describe the problem

***Edit***: Original title was "Make `locals` available to page load() function"

One common scenario in SSR rendered apps is that the initial render makes some async API calls, then later on when the user is interacting with the page, calls are made to that same API directly from the browser. For the server initiated API calls, it can simply forward on the end user's auth cookies. This is generally a good pattern as it makes your frontend service a dumb and unprivileged middleman.

Say we have a load function like this in one of our route components:

```javascript
export async function load({ fetch }) {
    let response = await fetch('https://api.example.com/mylatest', {
      credentials: 'include',
      headers: {
        cookie: 'jwt_token=???',
      },
    });
}
```

There doesn't seem to be a way to get at the end user's incoming request. This isn't specific to `fetch` or cookies, more generally we need the full detail of the request to service it properly. Maybe we look at the IP and do something different based on a coarse geolocation. Maybe we look at cookies, not just for forwarded authentication but even just for display preferences. Maybe we look at user-agent and fetch a smaller dataset if it's a phone than if it's a desktop. OK these examples are a bit contrived, it is 99.9% about forward auth.

So I think what SvelteKit wants you to do is use hooks to examine the low-level details of the request and turn it into higher level details that go in `locals`. So I can use `handle()` and put stuff in `event.locals`, but unless I'm missing something, there's no way to get that from load(), you can only get that from a data endpoint.

One workaround that is serviceable is to put whatever you need in session. But then it leaks out to the client (clearly visible with cmd-U show source), when it's actually only needed server-side.

So I'd propose simply plumbing locals through into the load() functions.

### Describe the proposed solution

See above.

### Alternatives considered

See above.
**Edit**: See below, maybe just update externalFetch?

### Importance

would make my life easier

johnnysprinkles:
Or, what might actually make more sense is to make "incoming_request" available to `externalFetch`.

johnnysprinkles:
A one-liner here would do it (verified locally):
https://github.com/sveltejs/kit/blob/ee906a3b2b103e4523eb1b6bb13f04d03eb35238/packages/kit/src/runtime/server/page/load_node.js#L204

Changed to:

    response = await options.hooks.externalFetch.call(null, external_request, event);

Also I think this might end up resolving several issues related to server-side data fetching for people who can't or don't want to have their API as a subdomain as their website.

Conduitry:
[...]

PR #3631, opened 2022-01-30
[fix] Server-side fetch request should have headers (#696)

aolose:
Server-side fetch in load function will include original request headers. （#696）

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -98,7 +98,7 @@ export async function load_node({
 				for (const [key, value] of event.request.headers) {
 					if (opts.headers.has(key)) continue;
 					if (key === 'cookie' || key === 'authorization') continue;
-					opts.headers.append(key, value);
+					opts.headers.set(key, value);
 				}
 
 				const resolved = resolve(event.url.pathname, requested.split('?')[0]);
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -92,11 +92,14 @@ export async function load_node({
 					};
 				}
 
+				opts.headers = new Headers(opts.headers);
+
 				// merge headers from request
-				opts.headers = new Headers({
-					...Object.fromEntries(event.request.headers),
-					...opts.headers
-				});
+				for (const [key, value] of event.request.headers.entries()) {
+					if (!opts.headers.has(key)) {
+						opts.headers.append(key, value);
+					}
+				}
 
 				const resolved = resolve(event.url.pathname, requested.split('?')[0]);
 
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -92,14 +92,11 @@ export async function load_node({
 					};
 				}
 
-				opts.headers = new Headers(opts.headers);
-
 				// merge headers from request
-				for (const [key, value] of event.request.headers.entries()) {
-					if (!opts.headers.has(key)) {
-						opts.headers.append(key, value);
-					}
-				}
+				opts.headers = new Headers({
+					...Object.fromEntries(event.request.headers),
+					...opts.headers
+				});
 
 				const resolved = resolve(event.url.pathname, requested.split('?')[0]);
 
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -97,9 +97,11 @@ export async function load_node({
 				// merge headers from request
 				for (const [key, value] of event.request.headers) {
 					if (opts.headers.has(key)) continue;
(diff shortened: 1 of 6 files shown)


benmccann:
It looks like there's another PR to fix this issue as well: https://github.com/sveltejs/kit/pull/2911. I haven't looked at either in any detail yet. Can you comment on the differences between the two approaches?

aolose:
> It looks like there's another PR to fix this issue as well: #2911. I haven't looked at either in any detail yet. Can you comment on the differences between the two approaches?

the param of `function load_node ` have changed after commit [`ee906a3b`](https://github.com/sveltejs/kit/commit/ee906a3b2b103e4523eb1b6bb13f04d03eb35238#diff-7266c49c71803f38dda70345d857f13a7a2494cc649b70af6b689627199f5b79)


```diff
* @param {{
- *   request: import('types/hooks').ServerRequest;
+ *   event: import('types/hooks').RequestEvent;
```

The code below is  unable to achieve goals.  [`228f0e660`](https://github.com/sveltejs/kit/pull/2911/commits/228f0e660a7f6713ff9eb9ac49244d70faa9b9df) 

```js
const headers = /** @type {import('types/helper').RequestHeaders} */ ({
    ...request.headers,
    ...opts.headers
});
```
Headers can't be merged. 

I try to modify it [`6f5308e7`](https://github.com/aolose/kit/commit/6f5308e79e0e18bec6dc7d9565696c607a6ef098):
```js
opts.headers = new Headers({
	...Object.fromEntries(event.request.headers),
	...opts.headers
});
```
It's working, but it can't pass tests.

johnnysprinkles:
This also relates to my #3503 where I'm grasping around for ways to get at the incoming request headers in order to fulfill a server-side rendered response (and one that makes data fetch calls).

I suggested either making "event" available in externalFetch, or for a narrower scoped change, making "fetch" have the cookies the way it would if you did a `credentials: "include"` fetch on the client. Haven't heard if either way is preferred, or neither.

Question though -- should this behavior be gated by the presence of "credentials: include"? At the very least it seems like it should respect "credentials: omit" the same as they do here right?

https://github.com/sveltejs/kit/blob/35461704dcc2e3859abb9a77e36163cc722c8708/packages/kit/src/runtime/server/page/load_node.js#L179

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -94,6 +94,13 @@ export async function load_node({
 
 				opts.headers = new Headers(opts.headers);
 
+				// merge headers from request
+				for (const [key, value] of event.request.headers.entries()) {
```
Rich-Harris, line 98: The `.entries()` is unnecessary, `headers` is already iterable

```suggestion
				for (const [key, value] of event.request.headers) {
```

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -94,6 +94,13 @@ export async function load_node({
 
 				opts.headers = new Headers(opts.headers);
 
+				// merge headers from request
+				for (const [key, value] of event.request.headers) {
+					if (opts.headers.has(key)) continue;
+					if (key === 'cookie' || key === 'authorization') continue;
```
benmccann, line 100: were we only going to skip these for the `omit` case? right now it looks like it's always skipping them

Conduitry: We're skipping copying them over directly right here. These headers should already have gotten handled earlier, if it was appropriate to copy them over.

Rich-Harris:
Realised there's at least a couple of headers that should get special casing — `if-none-match` should be removed (it's harmless, but can't possibly match), and `referer` should be set to the page URL. 

Adding some tests alongside this behaviour

bothness:
This fix appears to be causing problems when I'm using the `fetch` command in a route's `load` function within the module context.

I'm getting the following error **"Unexpected token < in JSON at position 2"**, which seems to mean that I'm getting an HTML error page in place of the JSON file that I'm trying to request (it worked perfectly in **1.0.0-next.252** but breaks in **1.0.0-next.252**).

I'm guessing this issue can be solved by explicitly adding certain headers to the request, eg. allowing CORS? It would be great if requests could be sent with suitable default headers to prevent this issue!

packages/kit/src/runtime/server/page/load_node.js
bothness: I seem to be getting issues from version 1.0.0-next.253 upwards whenever I make server-side fetch requests with an "authorization" header. I'm guessing it may relate to this code change?

See my comment here: https://github.com/sveltejs/kit/pull/3631#issuecomment-1065196387

squelix:
I'm getting an error too. I think we need more doc to say when to use fetch from global and fetch from loadInput.

benmccann:
@bothness comments on merged PRs will end up getting lost. If you have encountered a bug, please file a new issue

bothness:
Thanks for spotting this @benmccann. I'll file a fresh issue.

PR #3635, opened 2022-01-30
hopefully fixes #3552 Windows tests being flaky

### LINKED ISSUE #3552: fix the flaky tests
### Describe the bug

The tests that routinely fail on Windows — 'prefetches programmatically' and 'url-supplied anchor is ignored with onMount()' spring to mind — are a real nuisance. Getting these to pass reliably is probably the single biggest contribution someone could make relative to the development effort involved (though having said that, I've tried and failed)

### Reproduction

Open a random PR and look at the checks

### Logs

_No response_

### System Info

```shell
N/A
```

### Severity

annoyance

### Additional Information

_No response_

PH4NTOMiki:

Hopefully this PR will solve the issues with running the tests on Windows #3552 

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/utils.js
+++ b/packages/kit/test/utils.js
@@ -186,7 +186,7 @@ export const config = {
 		command: process.env.DEV ? 'npm run dev' : 'npm run build && npm run preview',
 		port: 3000
 	},
-	workers: 8,
+	workers: process.platform === 'win32' ? 4 : 8,
 	retries: process.env.CI ? 5 : 0,
 	projects: [
 		{


benmccann:
I don't think this is a good idea as it's just covering up rather than fixing the flakiness. It seems to me like there's some sort of race condition and this is impacting the speed at the tests run and impacting whether or not that race condition is hit. We should fix the root cause instead of trying to work around it

PH4NTOMiki:
@benmccann that's possible because it's running in the parallel, but in that case it would probably be same/similar for other platforms.

benmccann:
I don't think it'd necessarily be the same on the other platforms. Windows could be slower / faster than the other platforms, resourced differently on the CI, etc.

bluwy:
> I don't think this is a good idea as it's just covering up rather than fixing the flakiness. It seems to me like there's some sort of race condition and this is impacting the speed at the tests run and impacting whether or not that race condition is hit. We should fix the root cause instead of trying to work around it

I think for now we can still merge this so that CI would more consistently pass. We should definitely not close #3552 yet though since there are actual flaky tests happening for other environments too.

I think we can move to 4 workers for now for windows to not prolong the CI time.

PH4NTOMiki:
I guess I need to experiment some more, because even 2 now fails for some reason, but even MacOS is failing now for some reason, I'll push some more commits(empty) and see how it goes.

And I was looking at some of the test, it could be that some of them are failing because of this:
```
page.$eval(selector, pageFunction[, arg, options])
CAUTION
This method does not wait for the element to pass actionability checks and therefore can lead to the flaky tests. Use locator.evaluate(pageFunction[, arg, options]), other Locator helper methods or web-first assertions instead.
```
source: https://playwright.dev/docs/api/class-page#page-eval-on-selector
That's for the scrolling problem and some others, but I don't know enough of Playwright to be sure of anything.

PH4NTOMiki:
So 2 seems to be the most stable(other than 1 obviously but that's just too slow) and 2 doesn't seem to be that slower than 4 or even 8, so that's not an issue, and as I said I'll try to some more experimenting on my Windows machine and try to fix those flaky tests for good, but this could be a temporary fix, just to improve test consistency for now.

antony:
I'm also hesitant to merge this. Anything which hides flakiness will just mean that we are less likely to properly address it.

bluwy:
I don't think this PR is hiding the flakiness, but rather increasing the chances Windows CI passes right now. The tests are still flaky according to the playwright logs, so we have to address them soon too. For now I think this is nice to have as Windows CI typically has tighter resource constraints.

benmccann:
Tests that fail rarely are the most frustrating. If they fail frequently they are much easier to debug and much easier to tell if you've properly fixed them. I'd rather move it to 16 workers

PH4NTOMiki:
@Rich-Harris any input?
I also tried replacing those `page.$eval(selector, function)` functions with `page.locator(selector).evaluate(function)` locally, and I didn't see much change, but I'll try some more, also for those in_view function, I know that the code is using navigation-end event to wait for the page change, but maybe we could try to add `page.waitForTimeout(100)` or something like that after the event, just to be absolutely sure that the navigation did happen, that could have something to do with the racing conditions like @benmccann said.
I'm not really sure about this, but I'm just brainstorming.

Rich-Harris:
> @Rich-Harris any input?

agree with ben — we really need to try and identify the underlying race condition!

PH4NTOMiki:
Yeah, I agree, also I noticed some odd phenomenon, in most tests locally the issues arise in the build+js test, not dev+js, wonder what that could be

Rich-Harris:
closing in favour of #3741

PR #3637, opened 2022-01-30
docs: Update adapter-vercel README to recommend adapter-auto

leerob:
Rather than encouraging folks to install the Vercel adapter directly, we should recommend `adapter-auto` for maximum interoperability between platforms.


Rich-Harris:
Good catch, thanks. Though we _do_ need documentation specific to the Vercel adapter, since there are cases where you need to specify `external` modules, and `adapter-auto` doesn't take a config object (similarly, you might need to specify `split: true` for Netlify apps). I'll close this in favour of #3644

PR #3644, opened 2022-01-31
recommend using adapter-auto instead of cloudflare/netlify/vercel

Rich-Harris:
Supersedes #3637 — we need to be able to document the Vercel-specific options, for people who outgrow adapter-auto.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-netlify/README.md
+++ b/packages/adapter-netlify/README.md
@@ -2,6 +2,8 @@
 
 Adapter for Svelte apps that creates a Netlify app, using serverless functions for dynamically generating pages.
 
+If you're using [adapter-auto](../adapter-auto), you don't need to install this unless you need to specify Netlify-specific options, since it's already included.
+
 ## Installation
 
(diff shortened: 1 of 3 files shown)


leerob (approved):
Thanks, agreed this is a better solution 😄

packages/adapter-netlify/README.md
```diff
@@ -1,6 +1,8 @@
 # adapter-netlify
 
-Adapter for Svelte apps that creates a Netlify app, using serverless functions for dynamically generating pages.
+A SvelteKit adapter that creates a Netlify app.
+
+If you're using [adapter-auto](../adapter-auto), you don't need to install this unless you need to specify Netlify-specific options, since it's already included.
```
benmccann, line 5: i'm not an English major, but do we need another comma here? the reason I think that is that you could take out the "unless..." section and it would read "you don't need to install this since it's already included". (if we do change it then it'd need to be changed in the others as well)
```suggestion
If you're using [adapter-auto](../adapter-auto), you don't need to install this, unless you need to specify Netlify-specific options, since it's already included.
```

benmccann (commented):
seems fine to me

packages/adapter-netlify/README.md
Rich-Harris: I think it's fine as-is. You'd never have three commas in a sentence as short as this one unless you're Russ Hanneman
 
image

Issue #3636, opened 2022-01-30
clicking hash link recorded in history twice


kyontan:
### Describe the bug

When opening `#anchor-a` then click a link to `#anchor-b`, history **newly** records entries `#anchor-a` and `#anchor-b` though `#anchor-a` is previous location.

Therefore, hisotry looks like:

```
#anchor-a (initial view)
#anchor-b
#anchor-b (when clicked #anchor-b)
#anchor-c
#anchor-c (when clicked #anchor-c)
#anchor-d
#anchor-d (when clicked #anchor-d)
...
```

Disabling `router` by `export const router = false;` mitigate this issue, so that I think router has some issue around this.

### Reproduction

https://stackblitz.com/edit/sveltekit-ivmzya?devtoolsheight=33&file=src/routes/index.svelte

Easily understand by following steps:

1.  `Open in New Window` in Stackblitz
2.  Clicking anchor links some times
3.  Check a browser history, looks like:

<img width="319" alt="スクリーンショット 2022-01-31 3 59 07" src="https://user-images.githubusercontent.com/802339/151713646-85bb8908-f253-40de-9879-d6cb080fadeb.png">

(Also I notified `null` is oddly appealing on first clicking `#anchor-b`)

### System Info

```shell
# npx envinfo --system --binaries --browsers --npmPackages "{svelte,@sveltejs/*,vite}"
  System:
    OS: macOS 11.4
    CPU: (8) x64 Apple M1
    Memory: 24.88 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.2 - ~/.nodebrew/current/bin/node
    Yarn: 1.22.17 - ~/.nodebrew/current/bin/yarn
    npm: 8.1.2 - ~/.nodebrew/current/bin/npm
  Browsers:
    Firefox: 95.0
    Safari: 14.1.1
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.17
    @sveltejs/kit: next => 1.0.0-next.251
    svelte: ^3.46.3 => 3.46.3
```
```


### Severity

blocking an upgrade

### Additional Information

_No response_

kyontan:
I think replacing `<a href="#anchor">...</a>` with `<a on:click={() => goto("#anchor")}>...</a>` could workaround this.

Glench:
Maybe related to #3621?

bluwy:
[...]

Issue #3621, opened 2022-01-29
Browser "back" button doesn't scroll back from anchors in current page


Glench:
### Describe the bug

It's a little easier to describe this bug with the reproduction, but the gist is that when you click a link with an anchor (e.g. `<a href="#foo">`), then click the browser's back button, the page doesn't scroll back to where you were before you clicked the anchor. Instead it just stays at the same scroll position as the target, although it does properly change the url hash.

Testing the behavior on other sites (e.g. [any wikipedia page with a table of contents](https://en.wikipedia.org/wiki/Barack_Obama)) confirms that the browser does normally remember your scroll position when navigating between anchor elements on the same page.

### Reproduction

1. Go here as an example: https://sveltesaas.com/articles/sveltekit-github-pages-guide/
2. Scroll down and click the "Install SvelteKit" link in the Getting Started section. The page will properly scroll down to the correct section.
3. Click your browser's "back" button. The URL will correctly change but the page will not scroll back to where you were before, as it usually does when navigating between anchors on the same page.

### System Info

```shell
Firefox 96.0.2 on Mac 12.1
```

### Severity

annoyance

geoffrich:
I can repro this on the [Kit docs](https://kit.svelte.dev/docs). Interestingly, it only happens on the first hash link you click. So after clicking one hash link, the back button will not return you to the top of the page. But if you click two hash links in a row, the back button works -- it will first take you back to the first hash link, and then to the top of the page.

Glench:
> I can repro this on the [Kit docs](https://kit.svelte.dev/docs). Interestingly, it only happens on the first hash link you click. So after clicking one hash link, the back button will not return you to the top of the page. But if you click two hash links in a row, the back button works -- it will first take you back to the first hash link, and then to the top of the page.

Ohh interesting — you're right! Weird.

vicky1999:
Can I work on this issue?

benmccann:
I believe this is fixed now. @Glench can you upgrade to the latest version and test it out? Let me know if it's still present or if it has resolved for you

charbelnicolas:
I think this is still broken :(

Glench:
I just tried on the SvelteKit docs and the scrolling seems to work fine.

charbelnicolas:
> I just tried on the SvelteKit docs and the scrolling seems to work fine.

[...]

PR #3663, opened 2022-02-01
scroll on browser back button click

vicky1999:
solves - #3621 and maybe #3636 

### Tests

### Changesets

### Video Test: 

https://user-images.githubusercontent.com/24916780/152032395-44620051-5d47-4454-bf01-c33b2121278f.mov

### DIFF
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -186,9 +186,9 @@ export class Router {
 
 			// Check if new url only differs by hash
 			if (url.href.split('#')[0] === location.href.split('#')[0]) {
-				// Call `pushState` to add url to history so going back works.
+				// Call `replaceState` to add url to history so going back works.
 				// Also make a delay, otherwise the browser default behaviour would not kick in
-				setTimeout(() => history.pushState({}, '', url.href));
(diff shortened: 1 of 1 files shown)


benmccann:
Can you explain how you came to this solution and why you are making this change?

vicky1999:
When using `pushState`, the history is recorded twice while navigating through hashes ( #3636 ).  So, I tried `replaceState` to replace the history entry.

vicky1999:
@benmccann I have tested the approach.  and it is working.  please let me know if there's any other way to solve this problem.

bluwy:
I think there's a bigger issue with the behavour now, that somewhere in the code `pushState` is being called twice. With this PR, we're only making the first `pushState` a `replaceState` so we're pushing a new history entry once through the second `pushState`, and I dont think this is the right solution. We should find out where the second `pushState` happens and prevent that. Or maybe it's safe to remove the first `pushState`.

Maybe this was already discussed before and I missed it. But historically when we call the `setTimeout` `pushState` line, no other `pushState` is called later.

benmccann:
Yes, I agree this is not the right solution, so I'm going to go ahead and close this. If you can point out on the issues where the two calls happen that may help us in designing a better solution

Issue #3569, opened 2022-01-27
App-level types


Rich-Harris:
### Describe the problem

Various types in a Kit app are app-level, by which I mean they're the same wherever they're used — `event.locals`, [`event.platform`](#3429), `session`, and arguably `stuff` (which, even though it's scoped to `load`, can be used globally as `$page.stuff` and is therefore probably best thought of and used as a consistently-shaped object, where input `stuff` and output `stuff` in `load` are both `Partial<Stuff>`).

It's wasteful, therefore, that to take full advantage of type safety it's necessary to use generic arguments everywhere:

```html
<script context="module">
  /**
   * @type {import('@sveltejs/kit').Load<{
   *   pageParams: PageParams;
   *   session: { foo: number };
   *   stuff: Stuff;
   * }, {
   *   props: Props;
   *   stuff: Stuff
   * }>} */
  export function load({...}) {...}
</script>
```

```js
/**
 * @type {import('@sveltejs/kit').RequestHandler<
 *   Locals,
 *   Platform,
 *   Output
 * >}
 */
export function get({...}) {...}
```

```js
/**
 * @type {import('@sveltejs/kit').Handle<
 *   Locals,
 *   Platform
 * >}
 */
export async function handle(...) {...}
```

> Aside: why is it `pageParams` rather than `params`?

It's also a problem that `RequestHandler` and `Handle` take positional generic arguments — it meant that the addition of `Platform` was a breaking change, for example. And why doesn't `RequestHandler` accept `params`? If we fixed that, it would be another breaking change.

With app-level types (and assuming we _were_ to treat `stuff` as app-level), the examples above could be rewritten thusly:

```html
<script context="module">
  /**
   * @type {import('@sveltejs/kit').Load<{
   *   pageParams: PageParams;
   *   props: Props;
   * }>} */
  export function load({...}) {...}
</script>
```

```js
/** @type {import('@sveltejs/kit').RequestHandler<Output>} */
export function get({...}) {...}
```

```js
/** @type {import('@sveltejs/kit').Handle} */
export async function handle(...) {...}
```

### Stretch goal — no manual typing at all

A priori, it's silly that you have to type `params` — SvelteKit already has that information. It should make it available... [somehow](https://github.com/sveltejs/kit/issues/647).

In an ideal world, it wouldn't be necessary to type `props` either — that would be inferred from the props in the `<script>` block.

[...]

PR #3670, opened 2022-02-01
add a SvelteKit namespace for app-level types (#3569)

Rich-Harris:
Ref #3569.

This is a breaking change, as it affects how `Load` and `ErrorLoad` are typed if you're using generic arguments. If you're not, TypeScript will likely start yelling at you because `session` and `locals` etc are no longer typed as `any`.

By creating an app-level `namespace`, we're able to provide types for `event.locals`, `event.platform`, `page.session` and `page.stuff` throughout the app with no duplication. One caveat — it's no longer possible to specify input `stuff` and output `stuff` separately for a `load` function (unless you create your own type, I suppose) which might be a nuisance if you're ultra-pedantic. Personally I think the ergonomics are vastly better this way.

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/templates/default/src/app.d.ts
+++ b/packages/create-svelte/templates/default/src/app.d.ts
@@ -1,0 +1,13 @@
+/// <reference types="@sveltejs/kit" />
+
+declare namespace SvelteKit {
+	interface Locals {
+		userid: string;
+	}
+
+	interface Platform {}
+
+	interface Session {}
+
+	interface Stuff {}
+}
--- a/packages/create-svelte/templates/default/src/app.d.ts
+++ b/packages/create-svelte/templates/default/src/app.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="@sveltejs/kit" />
 
-declare namespace SvelteKit {
+declare namespace App {
 	interface Locals {
 		userid: string;
 	}
--- a/packages/create-svelte/templates/default/src/app.d.ts
+++ b/packages/create-svelte/templates/default/src/app.d.ts
@@ -1,5 +1,7 @@
 /// <reference types="@sveltejs/kit" />
 
+// See https://kit.svelte.dev/docs#typescript
+// for information about these interfaces
 declare namespace App {
 	interface Locals {
 		userid: string;
--- a/packages/create-svelte/templates/skeleton/src/app.d.ts
+++ b/packages/create-svelte/templates/skeleton/src/app.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="@sveltejs/kit" />
 
-declare namespace SvelteKit {
+declare namespace App {
 	interface Locals {}
 
 	interface Platform {}
--- a/packages/create-svelte/templates/skeleton/src/app.d.ts
+++ b/packages/create-svelte/templates/skeleton/src/app.d.ts
@@ -1,5 +1,7 @@
 /// <reference types="@sveltejs/kit" />
 
+// See https://kit.svelte.dev/docs#typescript
+// for information about these interfaces
 declare namespace App {
 	interface Locals {}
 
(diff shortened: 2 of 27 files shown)


Conduitry:
`7:12  error  An empty interface is equivalent to `{}`  @typescript-eslint/no-empty-interface`

How helpful.

packages/create-svelte/templates/skeleton/src/app.d.ts
```diff
@@ -0,0 +1,11 @@
+/// <reference types="@sveltejs/kit" />
+
+declare namespace App {
```
dummdidumm, line 3: We should add a link to the docs here so people know what this is about

packages/create-svelte/templates/default/src/app.d.ts
```diff
@@ -0,0 +1,13 @@
+/// <reference types="@sveltejs/kit" />
+
+declare namespace App {
```
dummdidumm, line 3: We should add a link to the docs here so people know what this is about

packages/kit/types/ambient-modules.d.ts
```diff
@@ -1,5 +1,12 @@
 /* eslint-disable import/no-duplicates */
 
+declare namespace App {
+	interface Locals {}
+	interface Platform {}
+	interface Session {}
+	interface Stuff {}
```
dummdidumm, line 7: Any possibility that someone would want this to be a primitive type which can't be expressed as an object-like interface?

dummdidumm (commented):
I really like this change, especially because it allows us to remove so many type-level code and the generics hack.

Couple of thoughts:
- is `app.d.ts` the best name? You can add anything that is global/ambient there, I slightly prefer the previous name, but I'm not passionate about it and can live with it either way
- Is `App` a good namespace name? My fear is that it's rather generic and the possibility of namespace clashes might be higher than with another name. What about `$App` since that somewhat corresponds to all the `$app/..` imports?
- Are all of these globally valid? For example `Stuff` does differ depending on your route. You might have `foo` inside `stuff` in one part of your app and `bar` in another part. This is not possible to express with this way - but I don't know how we could express this either. The only way I can think of is people wrapping the `stuff` store and reexporting it with a more specific type like `export const myStuff = stuff as Writable<SomeSubStuff>`

packages/kit/types/ambient-modules.d.ts
Rich-Harris: No, they're all required to be objects

packages/create-svelte/templates/skeleton/src/app.d.ts
Rich-Harris: good catch, done

packages/create-svelte/templates/default/src/app.d.ts
Rich-Harris: done

dummdidumm:
> Would those clashes ever be coming from dependencies? I haven't ever seen that happen but maybe there are packages that do that? If it's only likely to conflict with the user's own namespaces then I think that's probably fine, they would just need to rename stuff (and even then, only if their own namespace contained interfaces with conflicting names).

As soon as something make TS load a namespace it's part of the global types, and they can then clash. Doesn't matter where they come from. They also do need to have the same interface definitions inside for the clash to be noticeable though, which I guess would be very rare. I don't know, maybe this fear is unfounded.
 
> `$App` to me implies a closer connection to the `$app` modules than really exists. I initially used `SvelteKit` per [#3569 (comment)](https://github.com/sveltejs/kit/issues/3569#issue-1116462759), but it felt weird that those interfaces weren't coming _from_ SvelteKit.

👍 

> It _can_ differ, but since it's globally accessible as `$page.stuff` I think we want to encourage people to treat it as such. I'd expect the cases where it differs from route to route to be fairly rare and somewhat contrived.

Agree that the benefits outweigh the edge cases of pedantic typing wizards

ignatiusmb (approved):
This is refreshing to see. `adapter-cloudflare` docs needs updating as well, but other than that, this looks good!

PR #3674, opened 2022-02-02
Remove `target` option

Rich-Harris:
Ref #2221. This simplifies `svelte.config.js` and `src/app.html`, and removes the tight coupling between them, by getting rid of the `target` option. Instead, the init script hydrates the init script's `parentNode`. As a bonus, it makes it easier to embed multiple SvelteKit apps on a single page, for the rare cases where that's necessary.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/config/options.js
+++ b/packages/kit/src/core/config/options.js
@@ -252,7 +252,7 @@ const options = object(
 			),
 
 			// TODO remove this for 1.0
-			target: error((keypath) => `${keypath} is no longer required`),
+			target: error((keypath) => `${keypath} is no longer required, and should be removed`),
 
 			trailingSlash: list(['never', 'always', 'ignore']),
 
--- a/packages/kit/src/core/config/options.js
+++ b/packages/kit/src/core/config/options.js
@@ -251,7 +251,8 @@ const options = object(
 					`${keypath} has been removed — use the handle hook instead: https://kit.svelte.dev/docs#hooks-handle'`
 			),
 
-			target: string(null),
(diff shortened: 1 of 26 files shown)


packages/kit/src/core/config/options.js
```diff
@@ -251,7 +251,8 @@ const options = object(
 					`${keypath} has been removed — use the handle hook instead: https://kit.svelte.dev/docs#hooks-handle'`
 			),
 
-			target: string(null),
+			// TODO remove this for 1.0
+			target: error((keypath) => `${keypath} is no longer required`),
```
dummdidumm, line 255: "No longer required" sounds like optional, which doesn't sound like "remove it". This hopefully clarifies it.

```suggestion
			target: error((keypath) => `${keypath} is no longer required. Remove this option.`),
```

Mlocik97: `${keypath} is no longer supported, You should remove this option` sound better.

Rich-Harris: softened it a bit :)

```suggestion
			target: error((keypath) => `${keypath} is no longer required, and should be removed`),
```

james-camilleri:
I presume that any other elements that were previously in the app.html body now need to be moved to the primary layout.svelte? Some of my sites are using an external service (Snipcart) that needs some configuration passed via divs and an external script tag. I usually put them in the app.html body, but this update seems to be clearing the body tag indiscriminately, even if I wrap the %svelte% placeholder in a nested div. Although perhaps there's something I've implemented incorrectly from my end.

Issue #3532, opened 2022-01-24
Shadow endpoints (makes `load` boilerplate unnecessary in many cases)


Rich-Harris:
### Describe the problem

A significant number of pages have `load` functions that are essentially just boilerplate:

```svelte
<!-- src/routes/blog/[slug].svelte -->
<script context="module">
  /** @type {import('@sveltejs/kit').Load */
  export async function load({ params }) {
    const res = await fetch(`/blog/${params.slug}.json`);
    const { post } = await res.json();

    return {
      props: { post }
    };
  }
</script>

<script>
  export let post;
</script>
```

```js
// src/routes/blog/[slug].json.js
import * as db from '$lib/db';

export function get({ params }) {
  const post = await db.get('post', params.slug);

  if (!post) {
    return { status: 404 };
  }

  return {
    body: { post }
  }  
}
```

The `load` function even contains a bug — it doesn't handle the 404 case. 

While we definitely do need `load` for more complex use cases, it feels like we could drastically simplify this. This topic has come up before (e.g. #758), but we got hung up on the problems created by putting endpoint code inside a Svelte file (you need magic opinionated treeshaking, and scoping becomes nonsensical).

### Describe the proposed solution

I think we can solve the problem very straightforwardly: if a route maps to both a page and an endpoint, we regard the endpoint (hereafter 'shadow endpoint') as providing the data to the page. In other words, `/blog/[slug]` contains both the data (JSON) _and_ the view of the data (HTML), depending on the request's `Accept` header. (This isn't a novel idea; this is how HTTP has always worked.)

In the example above, no changes to the endpoint would be necessary other than renaming `[slug].json.js` to `[slug].js`. The page, meanwhile, could get rid of the entire `context="module"` script.

One obvious constraint: handlers in shadow endpoints need to return an object of props. Since we already handle serialization to JSON, this is already idiomatic usage, and can easily be enforced.

### POST requests

This becomes even more useful with `POST`. Currently, SvelteKit can't render a page as a result of a `POST` request, unless the handler redirects to a page. At that point, you've lost the ability to return data (for example, form validation errors) from the handler. With this change, validation errors could be included in the page props.

We do, however, run into an interesting problem here. Suppose you have a `/todos` page like the one from the demo app...

```js
import * as db from 'db';

export async function get({ locals }) {
  const todos = (await db.get('todo', locals.user)) || [];
  return {
    body: { todos }
  };
}

export async function post({ locals, request }) {
  const data = await request.formData();
[...]

PR #3679, opened 2022-02-02
Shadow endpoints

Rich-Harris:
This PR implements shadow endpoints — see #3532.

* [x] Load shadow data during SSR
* [x] Serialize shadow data
* [x] Load shadow data during hydration (from serialized data) and client-side navigation
* [x] Make data available at `${path}/__data.json`
* [x] Prevent prerendering of shadowed pages with non-`GET` methods
* [x] skip page response when `accept` does not include `text/html`
* [x] `invalidate('/foo')` should invalidate `/foo/__data.json` (at least conceptually)
* [x] Update docs

### Tests

### Changesets


Rich-Harris:
As mentioned in #3532, we need to accommodate prerendering. I think the neatest way to do this is to make shadow endpoint data available at `${path}/__data.json`. Since double-underscore-prefixed routes are already [verboten](https://github.com/sveltejs/kit/blob/21c01432fb9056abbef9c131706e324486c06662/packages/kit/src/core/create_manifest_data/index.js#L90-L92), this is safe to do, and provides a nice debugging experience (no need to fire up `curl` to set `accept` headers, just append `/__data.json` to the URL).

Maybe `GET ${path}` directly with an `accept` header that includes `application/json` but not `text/html` should redirect to this file? Not 100% sure what the HTTP etiquette is here.

documentation/docs/01-routing.md
```diff
@@ -114,28 +111,90 @@ The job of this function is to return a `{ status, headers, body }` object repre
...
 
 ```js
 export function post(event) {...}
+export function put(event) {...}
+export function patch(event) {...}
+export function del(event) {...} // `delete` is a reserved word
 ```
 
-Since `delete` is a reserved word in JavaScript, DELETE requests are handled with a `del` function.
+These functions can, like `get`, return a `body` that will be passed to the page as props. Whereas 4xx/5xx responses from `get` will result in an error page rendering, similar responses to non-GET requests do not, allowing you to do things like render form validation errors:
```
Conduitry, line 138: I'm confused about this part of this feature. Is this for AJAX forms? Old-fashioned no-JS forms? Both?

Rich-Harris: old-fashioned. if you hit the route with an `accept: application/json` header, it will behave like a standalone endpoint

Conduitry: I think most of my questions are answered by looking at the code sample below. I don't know whether there's something I would have preferred this paragraph before the code sample to say that would have made me less temporarily confused.

But, what happens if `GET`s or `POST`s return a 200 or non-200 response that's not JSON? Does that always render an error page?

Conduitry: Oh, whoops, GitHub didn't update to tell me about your comment that came in while I was writing my second one.

That might be worth mentioning in the docs here, along with the `accept: application/json` thing.

documentation/docs/03-loading.md
```diff
@@ -111,6 +109,10 @@ For a route filename example like `src/routes/a/[b]/[...c]` and a `url.pathname`
 }
 ```
 
+#### props
+
+If the page you're loading has an endpoint, the data returned from it is accessible inside the leaf component's `load` function as `props`.
```
Conduitry, line 114: Is it an empty object otherwise, or is it undefined?

packages/create-svelte/templates/default/package.json
```diff
@@ -5,8 +5,8 @@
 	"scripts": {
 		"dev": "svelte-kit dev",
 		"build": "svelte-kit build --verbose",
-		"preview": "svelte-kit preview",
-		"start": "svelte-kit start"
+		"package": "svelte-kit package",
+		"preview": "svelte-kit preview"
```
Conduitry, line 9: Wow, `start` has been a relic for a while.

I'm curious about the inclusion of `package` though. Is this part of the template that users get with `npm init svelte@next` or is it one of the test apps that get run in CI?

documentation/docs/03-loading.md
Rich-Harris: An empty object. will update

packages/create-svelte/templates/default/package.json
Rich-Harris: this was copy-pasta more than anything — this package.json isn't what ends up in a newly created project, it's just local stuff. package.template.json is where the action is

documentation/docs/01-routing.md
Rich-Harris: 6cb75dbd69f97942de78b57729dd7194465fb411

Conduitry: Sounds good!

documentation/docs/03-loading.md
Conduitry: :+1: Sounds good

packages/create-svelte/templates/default/svelte.config.js
```diff
@@ -1,3 +1,4 @@
+import path from 'path';
```
Conduitry, line 1:
```suggestion
```

This can also be removed now.

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -300,12 +335,148 @@ export async function load_node({
...
+
+		if (prerender && (mod.post || mod.put || mod.del || mod.patch)) {
+			throw new Error('Cannot prerender pages that have shadow endpoints with mutative methods');
+		}
+
+		const method = event.request.method.toLowerCase().replace('delete', 'del');
+		const handler = mod[method];
+
+		if (!handler) {
+			// TODO figure out what to do here. is 405 Method Not Allowed
```
Conduitry, line 382: "TODO figure out what to do here" always makes me nervous. What are the variables? What might we be falling through to?

packages/kit/src/runtime/server/index.js
```diff
@@ -3,6 +3,9 @@ import { render_page } from './page/index.js';
 import { render_response } from './page/render.js';
 import { respond_with_error } from './page/respond_with_error.js';
 import { coalesce_to_error } from '../../utils/error.js';
+import { decode_params } from './utils.js';
+
+const DATA_SUFFIX = '__data.json';
```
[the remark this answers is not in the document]
Rich-Harris: no, because if it was a root endpoint (`src/routes/index.js`), substracting `DATA_SUFFIX.length` would leave `''` rather than `'/'`, which would break. this way, you might be left with `/foo/bar/` or whatever, but that's fine — it will still match, regardless of whether the `trailingSlash` config causes redirects

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -300,12 +335,148 @@ export async function load_node({
...
+}
+
+/**
+ * @param {Headers | Partial<import('types/helper').ResponseHeaders>} headers
+ * @param {import('types/helper').JSONValue} body
+ */
+function validate_shadow_output(headers, body) {
+	if (headers instanceof Headers && headers.has('set-cookie')) {
+		throw new Error(
+			'Shadow endpoint request handler cannot use Headers interface with Set-Cookie headers'
```
Conduitry, line 475: Is this a security issue? a technical limitation? I don't see why this would be disallowed.

packages/kit/src/runtime/server/page/load_node.js
Rich-Harris: in a situation like this...

```
src/routes/specific.js
src/routes/specific.svelte
src/routes/[unspecific].js
src/routes/[unspecific].svelte
```

...where you `POST` to `/specific`, one of two things could happen if `src/routes/specific.js` doesn't export `post`: we could throw a 405, or we could fall through to `src/routes/[unspecific].js` to see if that's able to fulfil the request. But that means responding with a 404 rather than a 405 if it's unable to do so.

packages/kit/src/runtime/server/index.js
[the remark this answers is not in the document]
Conduitry: I guess I'm worried about an endpoint called `foo__data.json` (which would technically be allowed) and that causing an issue, but probably I shouldn't be worrying about that.

packages/kit/src/runtime/server/page/load_node.js
Rich-Harris: Because you can put multiple `set-cookie` headers _into_ `Headers`, but you can't get them out, because Chrome engineers are actively hostile to server-side use cases for `Headers` and have blocked spec changes that would make it possible

packages/kit/src/runtime/server/page/load_node.js
Conduitry: Hm. I think a 405 makes the most sense, then. I don't see a ton of value in letting shadow endpoints implicitly fall through to other endpoints. It would also make apps harder to reason about.

I assume that if `src/routes/specific.js` had a `post` that explicitly returned `{ fallthrough: true }` that that fallthrough would be handled the same way any other would? If so, I'm find with disallowing implicit fallthrough here.

packages/kit/src/runtime/server/index.js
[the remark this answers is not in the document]
Rich-Harris: oh, that's a good point. yeah. we should change it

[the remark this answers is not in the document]
Conduitry: I assume that would just be a matter of adding the `/` here and then doing `|| '/'` below?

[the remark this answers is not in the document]
Rich-Harris: yep, done

packages/kit/src/runtime/server/page/load_node.js
Rich-Harris: cool — removed the comment

Conduitry (approved):

deshartman:
Hi,

This change has broken the demo application build process for static adapter. It now shows a 500 error:

---------

Steps to reproduce:
`npm init svelte@next functions`

Use TypeScript? … Yes
Add ESLint for code linting? … Yes
Add Prettier for code formatting? … Yes

cd functions
`npm install`

-------

Change adaptor to static in svelte.config.js
`import adapter from '@sveltejs/adapter-static';`

Install the static adapter:
`npm install @sveltejs/adapter-static@next`

Build project
`npm run build`

npm run build

> functions@0.0.1 build
> svelte-kit build

vite v2.8.0 building for production...
✓ 34 modules transformed.
.svelte-kit/output/client/_app/assets/svelte-logo-87df40b8.svg                           1.85 KiB
.svelte-kit/output/client/_app/assets/fira-mono-greek-ext-400-normal-9e2fe623.woff2      7.33 KiB
.svelte-kit/output/client/_app/assets/fira-mono-cyrillic-ext-400-normal-3df7909e.woff2   15.40 KiB
.svelte-kit/output/client/_app/assets/fira-mono-latin-ext-400-normal-6bfabd30.woff2      11.10 KiB
.svelte-kit/output/client/_app/assets/fira-mono-cyrillic-400-normal-c7d433fd.woff2       8.89 KiB
.svelte-kit/output/client/_app/assets/fira-mono-greek-400-normal-a8be01ce.woff2          10.27 KiB
.svelte-kit/output/client/_app/assets/fira-mono-latin-400-normal-e43b3538.woff2          15.90 KiB
.svelte-kit/output/client/_app/assets/fira-mono-all-400-normal-1e3b098c.woff             75.55 KiB
.svelte-kit/output/client/_app/manifest.json                                             2.51 KiB
.svelte-kit/output/client/_app/pages/__layout.svelte-1fdcf5b9.js                         4.76 KiB / gzip: 1.86 KiB
.svelte-kit/output/client/_app/error.svelte-6f6d00da.js                                  1.56 KiB / gzip: 0.75 KiB
.svelte-kit/output/client/_app/pages/index.svelte-92dd2217.js                            4.08 KiB / gzip: 1.69 KiB
.svelte-kit/output/client/_app/pages/about.svelte-64b29582.js                            2.38 KiB / gzip: 1.09 KiB
.svelte-kit/output/client/_app/pages/todos/index.svelte-0cc5f931.js                      5.91 KiB / gzip: 2.40 KiB
.svelte-kit/output/client/_app/chunks/singletons-a6a7384f.js                             0.05 KiB / gzip: 0.07 KiB
.svelte-kit/output/client/_app/start-f44d78f5.js                                         23.36 KiB / gzip: 7.95 KiB
.svelte-kit/output/client/_app/assets/pages/todos/index.svelte-784042c1.css              3.70 KiB / gzip: 1.04 KiB
.svelte-kit/output/client/_app/assets/pages/index.svelte-1d45bd0c.css                    1.40 KiB / gzip: 0.50 KiB
.svelte-kit/output/client/_app/assets/pages/about.svelte-bf4528fa.css                    0.11 KiB / gzip: 0.10 KiB
.svelte-kit/output/client/_app/chunks/vendor-ee294e9e.js                                 14.16 KiB / gzip: 6.04 KiB
.svelte-kit/output/client/_app/assets/pages/__layout.svelte-15681802.css                 4.96 KiB / gzip: 1.55 KiB
vite v2.8.0 building SSR bundle for production...
✓ 36 modules transformed.
.svelte-kit/output/server/manifest.json                         2.15 KiB
.svelte-kit/output/server/app.js                                63.28 KiB
.svelte-kit/output/server/entries/endpoints/todos/index.ts.js   1.23 KiB
.svelte-kit/output/server/entries/pages/__layout.svelte.js      4.60 KiB
.svelte-kit/output/server/entries/pages/error.svelte.js         0.72 KiB
.svelte-kit/output/server/entries/pages/index.svelte.js         8.07 KiB
.svelte-kit/output/server/entries/pages/about.svelte.js         1.30 KiB
.svelte-kit/output/server/entries/pages/todos/index.svelte.js   5.84 KiB
.svelte-kit/output/server/chunks/index-ee9fdf1a.js              4.24 KiB

> 500 /todos
    at file:///functions/node_modules/@sveltejs/kit/dist/chunks/index5.js:408:11
    at visit (file:///functions/node_modules/@sveltejs/kit/dist/chunks/index5.js:585:4)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)


---------

What has changed:
- Missing [uid].json.ts
- Missing index.json.ts
- New file index.ts
-

PR #3716, opened 2022-02-03
[fix] Allow numeric as value of header

MacFJA:

### Tests

### Changesets

----

In the current version of SvelteKit, if an endpoint return an header with a numeric value, an error is raise:
> `TypeError: value.forEach is not a function`

The test on the value only assume that value can be either a `string` or an Array (the assumption is on the array).
https://github.com/sveltejs/kit/blob/89a97551f9e771e0e00064646a5196b943480ab1/packages/kit/src/utils/http.js#L10-L12

Per definition, a header value can only be a text (there is no typing in the HTTP protocol).
The type definition also define string and an array of string as only possibility
https://github.com/sveltejs/kit/blob/89a97551f9e771e0e00064646a5196b943480ab1/packages/kit/types/helper.d.ts#L7

But nothing prevent a number (integer or decimal) to be defined (if not typing check is done) and there are not, per say, wrong (for example `content-length` header is a numerical value)

----

Another approach can be to ignore all values that are not a `string` or an **array of `string`**
```js
if (typeof value === 'string') { // Allow string value
	headers.set(key, value);
} else if (Array.isArray(value)) { // Allow array
	value
		.filter((value) => typeof value === 'string') // Ignore every value that are not string
		.forEach((value) => {
			headers.append(key, value);
		});
} // values that are not a string or an array are ignored
```

----

The fact that user should set correctly headers can also be a way to handle this case.
Then feel free to close the PR :wink:

### DIFF
--- a/.changeset/sharp-mugs-melt.md
+++ b/.changeset/sharp-mugs-melt.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Handle numeric headers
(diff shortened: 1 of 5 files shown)


Rich-Harris (approved):

Rich-Harris:
thanks!

Issue #3696, opened 2022-02-03
Overhaul docs


Rich-Harris:
### Describe the problem

https://kit.svelte.dev/docs needs some love. The API was a nice idea but we decided it's ultimately more trouble than it's worth — previewing changes is much harder, and it prohibits us from experimenting with things like mdsvex (which is very useful in the [Svelte Cubed documentation](https://svelte-cubed.vercel.app/docs/getting-started), for example)

### Describe the proposed solution

The maintainers recently settled on the outline of a plan:

1. [x] #3704 
2. [x] Deploy the site from this repo, rather than `sveltejs/sites`
3. [x] Remove kit.svelte.dev from `sveltejs/sites`
4. [x] Prerender docs using markdown in this repo, rather than api.svelte.dev
5. [x] Remove Kit docs from api.svelte.dev
6. [x] Multiple pages and full text search
7. [x] #3324 so that the docs are typechecked

Other thoughts:

* We should move most of the type declarations to the `TypeScript` section — there's a lot of duplicative clutter that helps no-one
* It's important that we're able to accommodate internationalisation work in future, though we don't yet know exactly what form that will take in SvelteKit. I do like the idea of having translations be visible in the repo, where it's easy to (for example) see if a given piece of the primary documentation is more recent than its translations.

### Importance

would make my life easier

benmccann:
In case it helps with the next step, here's the old code for generating docs locally. It's possible there might need to be some changes copied over from the API if this docs generation code has changed at all:

https://github.com/sveltejs/sites/blob/a043d3fdf6c3fbfd8f5dc84979eb03a50657b7d4/sites/kit.svelte.dev/src/routes/docs/index.json.js
https://github.com/sveltejs/sites/blob/a043d3fdf6c3fbfd8f5dc84979eb03a50657b7d4/sites/kit.svelte.dev/src/utils/generate_docs.js

lukaszpolowczyk:
**Collapsing elements in the side panel**

While we're on the subject of shortening documentation page scrolling, why not add element collapsing in the side panel? Because the list in the side panel is pretty long too, and often whole sections I skip anyway.
(Even the collapsing can be maintained between sessions)

PR #3717, opened 2022-02-04
Full text search

Rich-Harris:
Step 6 of #3696.

Preview deployment: https://kit-svelte-dev-git-flexsearch-svelte.vercel.app/

WIP. Adds a search bar powered by https://github.com/nextapps-de/flexsearch. Needs lots of design work and debugging (and @geoffrich, if you get a chance, I'd be very grateful for any a11y pointers), but as a proof of concept it does the trick.

Once it's in better shape, we can move to multi-page docs (and move migrating/FAQ into the docs — they don't need to be separate links, I don't think).

One known issue: I haven't figured out how to make a 'clean' `marked` instance — once `marked.use(...)` gets called, the options are used everywhere, which means that the `plaintext` stuff in this PR affects the generation of the docs themselves (which should only affect local development, but is annoying nonetheless).

* [x] improve search (make the headings themselves searchable, not just the content that belongs to them)
* [x] improve design (group items?)
* [ ] ensure a11y is up to snuff
* [x] move to a multi-page setup
* [x] update all the internal docs links
* [x] fix links to FAQ
* [x] fix FAQ CSS
* [x] fix bug that happens when you bring up the search box with Cmd-K while the page is scrolled — if nav bar isn't visible, it flashes in awkwardly
* [x] fix markdown transformation after `content.json` has been generated
* [x] ensure it works well on mobile
* [x] see if we can get away with keeping all this stuff on the main thread, or if we need to faff about with workers
* [x] while waiting for a query, list recent searches (powered by localStorage) with delete buttons, or show a 'no recent searches' message
* [ ] (eventually) extract this stuff out so that it's usable by other sites, e.g. svelte.dev (which might also include tutorials and blog posts etc)
* [x] fade search bar in on hydration, since it's not present at SSR time
* [x] replace ⌘K with the windows equivalent when appropriate, hide on mobile
* [x] include faqs/migration
* [x] continue to show links in mobile nav
* [x] fix scrollbar bug (dragging the scrollbar kills the modal)
* [ ] contrast in search box https://github.com/sveltejs/kit/pull/3717#discussion_r800106735
* [ ] check things are in the right place in safari https://github.com/sveltejs/kit/pull/3717#discussion_r800107719
* [x] fix nav overlapping https://github.com/sveltejs/kit/pull/3717#issuecomment-1031776732

Tasks specific to multi-page redesign:

* [x] next/prev links
* [x] only load TOC in layout, not the content itself
* [ ] use shadow endpoint for content (there's a bug that needs fixing first — the data isn't refetched when the URL changes)
* [x] remove contents menu on mobile in favour of a) search and b) links in footer

### Tests

### Changesets


sites/kit.svelte.dev/src/lib/search/Search.svelte
```diff
@@ -0,0 +1,253 @@
...
+		type="search"
+		placeholder="Search (Cmd-K)"
+	/>
+</div>
+
+{#if searching}
+	<div class="modal-background" on:click={() => (searching = false)}>
+		<div class="search-box" on:click={(e) => e.stopPropagation()}>
+			<!-- svelte-ignore a11y-autofocus -->
+			<input
```
geoffrich, line 101: This is going to need some a11y improvements + screen reader testing to make sure it behaves properly. Comboboxes like this are one of the trickier patterns from an a11y perspective. I'll do some research and get back to you.

Potential resources/patterns I'll be diving into:
- [WAI-ARIA combobox pattern](https://www.w3.org/TR/wai-aria-practices-1.1/#combobox)
- [React ARIA - Creating an accessible autocomplete experience](https://react-spectrum.adobe.com/blog/building-a-combobox.html)
- [Lion combobox component](https://lion-web.netlify.app/components/inputs/combobox/overview/)
- Other framework docs search widgets

sites/kit.svelte.dev/src/lib/search/Search.svelte
```diff
@@ -0,0 +1,253 @@
...
+	.search-box input {
+		width: 100%;
+		padding: 0.5rem;
+		border-radius: var(--border-r);
+		margin: 0 0 1rem 0;
+	}
+
+	.results {
+		flex: 1;
+		overflow: scroll;
```
geoffrich, line 209: I think this should be overflow: auto. Otherwise scrollbars show up on Windows when there's nothing to scroll.

sites/kit.svelte.dev/src/lib/search/Search.svelte
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	.search-box input:focus-visible {
+		border: 2px solid var(--flash);
+		outline: none;
+	}
+
+	.results {
+		flex: 1;
+		overflow: scroll;
```
Rich-Harris, line 255:
```suggestion
		overflow: auto;
```

sites/kit.svelte.dev/src/lib/search/Search.svelte
```diff
@@ -0,0 +1,323 @@
...
+			e.preventDefault();
+			searching = !searching;
+		} else if (document.activeElement === button) {
+			console.log(e);
+		}
+	}}
+/>
+
+<div class="search-container">
+	<input
```
Rich-Harris, line 95:
```suggestion
	<input
		aria-label="search"
```

sites/kit.svelte.dev/src/lib/search/Search.svelte
```diff
@@ -0,0 +1,253 @@
...
+
+			if (d_top > 0) ul.scrollTop -= d_top;
+			if (d_bottom < 0) ul.scrollTop -= d_bottom;
+		}
+	}
+</script>
+
+<svelte:window
+	on:keydown={(e) => {
+		if (e.code === 'KeyK' && e.metaKey) {
```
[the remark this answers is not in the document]
Rich-Harris: I was basically trying to replication the behaviour of the Algolia search widget — what does Ctrl+K or Win+K do on https://v3.nuxtjs.org/?

sites/kit.svelte.dev/src/lib/search/Search.svelte
```diff
@@ -0,0 +1,253 @@
...
+			e.preventDefault();
+			searching = !searching;
+		}
+	}}
+/>
+
+<div class="search-container">
+	<input
```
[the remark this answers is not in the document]
Rich-Harris: done

sites/kit.svelte.dev/src/lib/search/Search.svelte
Rich-Harris: fixed

sites/kit.svelte.dev/src/lib/search/Search.svelte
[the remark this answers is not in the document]
geoffrich: Ctrl + K brings up the search widget for me on that page.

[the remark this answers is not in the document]
brittneypostma: Ctrl + K does the same for me in FireFox Dev and Chrome.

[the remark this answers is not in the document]
TheOnlyTails: > what does Ctrl+K or Win+K do on [v3.nuxtjs.org](https://v3.nuxtjs.org/)?

For me (Firefox Dev) <kbd>Ctrl</kbd> + <kbd>K</kbd> brings out the widget, while <kbd>Win</kbd> + <kbd>K</kbd> is a Windows shortcut.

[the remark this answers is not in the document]
brittneypostma: This is what I get with `Ctrl` + `K` on Firefox Dev 
image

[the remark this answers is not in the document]
TheOnlyTails: Ah, I meant on the Nuxt website. I get the same on the deploy preview.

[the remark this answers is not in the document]
Rich-Harris: hopefully this fixes windows? https://github.com/sveltejs/kit/pull/3717/commits/6637f801c0f9e33cbc90eeb558e8cb81b5820484

Rich-Harris:
Bugs:

* [ ] clicking a link to another page (e.g. `/faq` from `/docs`) results in a full page load, rather than a client-side navigation. not sure why
* [ ] clicking a hash link doesn't clear the modal
* [ ] clicking a link doesn't populate `$recent`
* [ ] on reflection, treating tab and shift-tab the same as arrowdown and arrowup doesn't really make sense. the modal should trap focus, but we shouldn't interfere with tab behaviour beyond that. this will allow us to add a close button, and delete buttons on the recent searches

sites/kit.svelte.dev/src/lib/search/Search.svelte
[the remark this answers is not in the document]
brittneypostma: > hopefully this fixes windows? [6637f80](https://github.com/sveltejs/kit/commit/6637f801c0f9e33cbc90eeb558e8cb81b5820484)

Is there a deploy link that is working? I'll test it out.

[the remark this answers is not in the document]
brittneypostma: Oh, apparently the one in the first comment works 😅 Windows is fixed!!

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
```diff
@@ -0,0 +1,416 @@
...
+			$searching = false;
+		}
+	}}
+/>
+
+{#if $searching && index}
+	<div
+		class="modal-background"
+		on:click={() => ($searching = false)}
+		on:wheel={(e) => e.preventDefault()}
```
geoffrich, line 110: Nit: when I hover over the modal and scroll, the underlying page still scrolls. This only happens if the modal is not showing a scrollbar (i.e. not enough results to scroll). Not sure if it's worth addressing.

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
```diff
@@ -0,0 +1,416 @@
...
+
+		const prefix = index > 20 ? `…${content.slice(index - 15, index)}` : content.slice(0, index);
+		const suffix = content.slice(
+			index + query.length,
+			index + query.length + (80 - (prefix.length + query.length))
+		);
+
+		return (
+			escape(prefix) +
+			`<mark>${escape(content.slice(index, index + query.length))}</mark>` +
```
geoffrich, line 76: Marking up content like this is having the screen readers I tested (NVDA and VoiceOver) break up the words, which could be confusing. For example, if I search "rou", the first result's heading is output as `<mark>Rou</mark>ting`, which is read by NVDA as "highlighted    Rou  out of highlighted  ting". VoiceOver doesn't announce "highlight", but does separate into separate words, i.e. "Rou ting".

I wonder if it would be a good idea to output the marked-up strings twice: once with the `<mark>` elements and once without. The marked version would be hidden from screen readers with aria-hidden, and the unmarked version would be hidden from sighted users [with CSS](https://piccalil.li/quick-tip/visually-hidden/). This way, screen readers would have the full string read out without being broken up.

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
```diff
@@ -0,0 +1,416 @@
...
+				}}
+				on:input={(e) => {
+					$query = e.target.value;
+					update();
+				}}
+				value={$query}
+				placeholder="Search"
+			/>
+
+			<button aria-label="Close" on:click={() => ($searching = false)}>
```
geoffrich, line 179: This is misaligned on iOS Safari:
close

sites/kit.svelte.dev/src/lib/search/Search.svelte
```diff
@@ -0,0 +1,112 @@
...
+	<span>
+		<svg viewBox="0 0 24 24">
+			<path
+				fill="currentColor"
+				d="M9.5,3A6.5,6.5 0 0,1 16,9.5C16,11.11 15.41,12.59 14.44,13.73L14.71,14H15.5L20.5,19L19,20.5L14,15.5V14.71L13.73,14.44C12.59,15.41 11.11,16 9.5,16A6.5,6.5 0 0,1 3,9.5A6.5,6.5 0 0,1 9.5,3M9.5,5C7,5 5,7 5,9.5C5,12 7,14 9.5,14C12,14 14,12 14,9.5C14,7 12,5 9.5,5Z"
+			/>
+		</svg>
+
+		<!-- TODO should be ctrl on non-mac -->
+		Search <kbd>{navigator.platform === 'MacIntel' ? '⌘' : 'Ctrl'}</kbd> <kbd>K</kbd>
```
geoffrich, line 28: Can we make this the label for the input instead of using aria-label?

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
```diff
@@ -0,0 +1,416 @@
...
+						<li>
+							<a on:click={() => navigate(result.href)} href={result.href}>
+								<small>{result.breadcrumbs.join('/')}</small>
+								<strong>{@html excerpt(result.title, $query)}</strong>
+								<span>{@html excerpt(result.content, $query)}</span>
+							</a>
+						</li>
+					{/each}
+				{:else}
+					<li class="info">{recent_searches.length ? 'Recent searches' : 'No recent searches'}</li>
```
geoffrich, line 196: This probably makes more sense as a heading than a list item.

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
```diff
@@ -0,0 +1,416 @@
...
+				if (e.key === 'ArrowDown') {
+					group.next(selector);
+				} else {
+					group.prev(selector);
+				}
+			}
+
+			if (e.code === 'Space' && e.target.nodeName !== 'INPUT') {
+				e.preventDefault();
+			}
```
geoffrich, line 134: What is this doing?

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
```diff
@@ -0,0 +1,416 @@
...
+		}}
+		use:trap
+	>
+		<div class="search-box">
+			<!-- svelte-ignore a11y-autofocus -->
+			<input
```
geoffrich, line 140: After further reflection, I don't think this is a combobox or autocomplete (as I suggested in another comment), since we aren't having the user select a value. It's just an input that dynamically filters a list of search results. With that in mind, I think we can apply some recommendations from ["Considering dynamic search results and content"](https://www.scottohara.me/blog/2022/02/05/dynamic-results.html) by Scott O'Hara (posted today!)

- Add [instructions](https://www.scottohara.me/blog/2022/02/05/dynamic-results.html#the-text-field) that say something like "Results will appear as you type", linked to the input with aria-describedby. These could be visually hidden so they're only shown to assistive tech users, if desired.
- Add assertive live region to [announce when no results are found](https://www.scottohara.me/blog/2022/02/05/dynamic-results.html#communicating-no-results)

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
```diff
@@ -0,0 +1,416 @@
...
+<svelte:window
+	on:keydown={(e) => {
+		if (e.code === 'KeyK' && (navigator.platform === 'MacIntel' ? e.metaKey : e.ctrlKey)) {
+			e.preventDefault();
+			$query = '';
+			$searching = !$searching;
+		}
+
+		if (e.code === 'Escape') {
+			$searching = false;
```
geoffrich, line 101: When closing the modal, we need to move focus somewhere so it's not lost (though only if the user didn't select an option, i.e. they're not navigating). The candidates would either be 1) the search input in the nav bar (it's where focus initially was, but could be confusing to non-sighted users, since doing so would read out "Search input" again) or 2) the skip link in the nav (basically resetting to the top of the page)

geoffrich (commented):
One issue I noticed when testing keyboard nav: when following a hash link, my focus ends up at the top of the page instead of the link destination, effectively making the search box useless for keyboard users. Repro steps:

1. Tab to search box
2. Enter search query (e.g. route)
3. Select search result with arrow keys and hit enter
4. After navigation completes, hit tab. The first item on the page is focused (the skip link). 

Expected behavior: The focus is at the heading that was linked to.

This only happens if you're not initially on the page. Once you're on the docs page, any search result link you follow that is also on the docs page lands keyboard focus in the correct position.

Not sure if this is an issue with the implementation or with SvelteKit.

sites/kit.svelte.dev/src/lib/search/Search.svelte
geoffrich: I changed my mind and don't think this is really a combo box. See my [new comment](https://github.com/sveltejs/kit/pull/3717#discussion_r800108915).

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
Rich-Harris: it's preventing the spacebar from scrolling the page if the input isn't focused (i often forget whether enter or spacebar activates a focused link, and when i press the spacebar it's jarring for the page to jump down)

Rich-Harris: added a note

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
```diff
@@ -0,0 +1,416 @@
...
+				{:else}
+					<li class="info">{recent_searches.length ? 'Recent searches' : 'No recent searches'}</li>
+					{#each recent_searches as search, i}
+						<!-- svelte-ignore a11y-mouse-events-have-key-events -->
+						<li class="recent">
+							<a on:click={() => navigate(search.href)} href={search.href}>
+								<small>{search.breadcrumbs.join('/')}</small>
+								<strong>{search.title}</strong>
+
+								<button
```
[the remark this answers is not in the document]
Rich-Harris: yep, fixed

sites/kit.svelte.dev/src/routes/__layout.svelte
```diff
@@ -13,12 +16,18 @@
 <SkipLink href="#main" />
 <Nav {page} logo="/images/svelte-kit-horizontal.svg">
 	<svelte:fragment slot="nav-center">
+		{#if browser}
+			<Search />
```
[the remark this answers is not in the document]
Rich-Harris: The `<li>` is weird, but unfortunately you're right, it's necessary with `<Nav>` as it exists right now. A lot of components in `site-kit` are going to need to be rethought, but doing so would break svelte.dev, so in the short term it's probably easiest to do the weird thing but leave a note

sites/kit.svelte.dev/src/routes/__layout.svelte
```diff
@@ -33,6 +42,10 @@
 	</svelte:fragment>
 </Nav>
 
+{#if browser}
+	<SearchBox />
```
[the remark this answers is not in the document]
Rich-Harris: ah, yep. fixed

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
Rich-Harris: ah, interesting. the only real reason to use `<mark>` is for semantic correctness, which I had naively assumed would be the more accessible choice. if it's actually harmful to accessibility to mark incomplete words, then maybe it should just be a `<span class="highlight">` instead?

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
Rich-Harris: Ugh safari. It was broken on desktop safari too, which i fixed by adding `position: relative` to `.search-box`. Hopefully that fixes iOS safari as well (will be able to check later when i can borrow an iphone)

sites/kit.svelte.dev/src/lib/search/Search.svelte
Rich-Harris: yep, done

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
Rich-Harris: Updated — I used an `<h2>` because I'm never sure what level is appropriate, but `<h1>` seems wrong. Lmk if you have any thoughts on that!

Conduitry:
The current preview deployment seems to be broken, in case you hadn't noticed. The endpoint https://kit-svelte-dev-git-flexsearch-svelte.vercel.app/docs/sections.json for example is returning a 500 with an error message `ENOENT: no such file or directory, scandir '../../documentation/docs'`.

Conduitry:
I'm seeing an error now for https://kit-svelte-dev-git-flexsearch-svelte.vercel.app/docs/__data.json which sounds like it could be related to #3760.

Rich-Harris:
gah

image

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
```diff
@@ -0,0 +1,416 @@
...
+
+			<ul class="results">
+				{#if $query}
+					{#each results as result, i}
+						<!-- svelte-ignore a11y-mouse-events-have-key-events -->
+						<li>
+							<a on:click={() => navigate(result.href)} href={result.href}>
+								<small>{result.breadcrumbs.join('/')}</small>
+								<strong>{@html excerpt(result.title, $query)}</strong>
+								<span>{@html excerpt(result.content, $query)}</span>
```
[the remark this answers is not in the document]
Rich-Harris: Fixed the first by using `--second` everywhere instead of `--flash`:

image

The second is a bit trickier — could use `--text` everywhere, but it would make the results look very noisy, which would make them much _harder_ to parse for people for whom the current contrast is sufficient. Not sure what the best solution is, will keep fiddling

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
geoffrich: Upon further reflection, I think this is fine to leave as-is:

1. `<mark>` is the more semantically correct option here
1. NVDA lets users [configure this behavior](https://www.nvaccess.org/files/nvda/documentation/userGuide.html#DocumentFormattingSettings), so users can turn off the "highlight" announcement if they want. 
1. Since I'm not certain that hearing incomplete words like this is an accessibility issue, we probably shouldn't over-optimize for that assumption. 

> To a developer or user who is used to the way JAWS or NVDA announces content, the behavior of VO might seem fragmented and annoying, but to someone who regularly uses VO, this pattern of speech output is usual, and may be the expected or even preferred pattern. It’s not necessarily accurate to call this a usability problem at all, and it’s not safe to assume that the typical behavior of an assistive technology is something that needs to be “fixed”, simply because it’s inconsistent with the behavior of similar technologies. In any case, this quirk only exists at all when navigating with the default side-swipe gesture pattern, it doesn’t occur when navigating using rotor lines navigation, or when VO reads continuously (but the problems caused by overriding semantics do still occur in all circumstances).
> &mdash;[TPGI](https://www.tpgi.com/using-the-text-role/)

Rich-Harris:
the deploy is working now, i fooked up the prerendering

sites/kit.svelte.dev/src/routes/docs/[slug].svelte
```diff
@@ -0,0 +1,60 @@
...
+	<meta name="twitter:description" content="{section.title} • SvelteKit documentation" />
+	<meta name="Description" content="{section.title} • SvelteKit documentation" />
+</svelte:head>
+
+<div class="content listify">
+	<h1>{section.title}</h1>
+
+	<a href="https://github.com/sveltejs/kit/edit/master/documentation/docs/{section.file}">
+		<Icon size={50} name="edit" /> Edit this page on GitHub
+	</a>
```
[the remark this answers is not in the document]
Comment (Rich-Harris) [reply sites/kit.svelte.dev/src/routes/docs/[slug].svelte:39]: fixed — was getting hidden by the subsequent header's padding/negative margin combo (which exists so that the header doesn't stick to the very top of the page behind the nav bar when you click a `#` link)

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
Rich-Harris: thanks — have implemented both of this

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
Rich-Harris: I tried focusing the skip link but it meant that it became visible...

image

...which I think we probably want to avoid. I just pushed a change that focuses the `<body>` instead, maybe that's ok?

Rich-Harris:
> Not sure if this is an issue with the implementation or with SvelteKit.

It's a bug in Kit itself — opened https://github.com/sveltejs/kit/issues/3770

ghostdevv:
Am wondering how integrating with this off-site will work. Currently the api is used by the [svelte discord bot](https://github.com/pngwn/svelte-bot) to add search functionality inside of the discord. Will the bot be able to use the kit endpoints or need to implement it's own scraping and parsing of the data whether that's from repo or site

Rich-Harris:
@ghostdevv at present the content that powers the search box is available at [/content.json](https://kit-svelte-dev-git-flexsearch-svelte.vercel.app/content.json). The table of doc contents is at [/docs.json](https://kit-svelte-dev-git-flexsearch-svelte.vercel.app/docs.json), and individual sections (with content) are at [/docs/introduction.json](https://kit-svelte-dev-git-flexsearch-svelte.vercel.app/docs/introduction.json) etc. The bot might need some rejiggering but I _think_ everything it needs should be available.

Having said that those URLs might change if this switches over to use shadow endpoints — /docs/introduction/__data.json etc.

ghostdevv:
@Rich-Harris thanks! We can work with that for sure

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
geoffrich: This is fixed now, though there's on more Safari issue I noticed -- seems like there's an overlapping icon on the search box. image

PR #3741, opened 2022-02-05
Address playwright race conditions and environment timeouts

### LINKED ISSUE #3552: fix the flaky tests

harvey-k:

### Tests

### Changesets

Attempting to resolve #3552

First, the obligatory "it works on my machine"...

In attempting to address this issue, I encountered the following underlying difficulties:
 1) Addressing "prefetches programmatically" test race by waiting for playwright network event
 2) Handling environment dependent network requests issued by "prefetches programmatically" test
 3) Addressing visibility issues by using playwright's page.isVisible method
 4) Random test timeouts on my local machine, and apparently also in other environments according to [discord](https://discord.com/channels/457912077277855764/750401468569354431/939343653434982451). These were addressed by bumping the window.setTimeout threshold from 2s to 5s AND the test harness from 10s to 15s in utils.js, both of which were required.

Further info on changes for "prefetches programmatically" test, the following debug output
```
console.log(`waiting for ${baseURL}/routing/prefetched.json`);
requests.forEach((req) => {
  console.log(`req was: ${req}`);
});
```
Resulted in
```
waiting for http://localhost:3000/routing/prefetched.json
req was: http://localhost:3000/src/routes/routing/prefetched/index.svelte
req was: http://localhost:3000/@fs/C:/kit/packages/kit/src/runtime/env.js
req was: http://localhost:3000/routing/prefetched.json

waiting for http://localhost:3000/routing/prefetched.json
req was: http://localhost:3000/src/routes/routing/prefetched/index.svelte
req was: http://localhost:3000/routing/prefetched.json

waiting for http://localhost:3000/routing/prefetched.json
req was: http://localhost:3000/_app/pages/routing/prefetched/index.svelte-b0244d07.js
req was: http://localhost:3000/routing/prefetched.json
```
Note depending upon the vite mode, sometimes a third "fs" path would be fetched.

Possible improvements:
  1) Unfortunately, the sets_paths and trailingSlash tests still fail on windows, seemingly reliably at least. Should I open a separate for this?
  2) Should the integration timeout values just be changed to the following, in order to account for additional failures mentioned in the [discord discussion](https://discord.com/channels/457912077277855764/750401468569354431/939343653434982451):
     timeout: process.env.CI ? 45000 : 15000,

All feedback welcome!

Last but not least thank you so much for your efforts on this project, SvelteKit and Svelte are wonderful!

### DIFF
--- a/packages/kit/test/apps/basics/test/test.js
+++ b/packages/kit/test/apps/basics/test/test.js
@@ -1627,9 +1627,21 @@ test.describe.parallel('Routing', () => {
 			let requests = [];
 			page.on('request', (r) => requests.push(r.url()));
 
-			await app.prefetch('/routing/prefetched');
-			expect(requests.length).toBe(2);
-			expect(requests[1]).toBe(`${baseURL}/routing/prefetched.json`);
+			// also wait for network processing to complete, see
+			// https://playwright.dev/docs/network#network-events
+			await Promise.all([
+				app.prefetch('/routing/prefetched'),
+				page.waitForResponse(`${baseURL}/routing/prefetched.json`)
+			]);
+
+			// svelte request made is environment dependent
+			if (process.env.DEV) {
+				expect(requests.filter((req) => req.endsWith('index.svelte')).length).toBe(1);
+			} else {
+				expect(requests.filter((req) => req.endsWith('.js')).length).toBe(1);
+			}
+
+			expect(requests.includes(`${baseURL}/routing/prefetched.json`)).toBe(true);
 
 			requests = [];
 			await app.goto('/routing/prefetched');
--- a/packages/kit/test/utils.js
+++ b/packages/kit/test/utils.js
@@ -60,7 +60,7 @@ export const test = base.extend({
 			if (javaScriptEnabled) {
 				await page.evaluate(() => {
 					window.navigated = new Promise((fulfil, reject) => {
-						const timeout = setTimeout(() => reject(new Error('Timed out')), 2000);
+						const timeout = setTimeout(() => reject(new Error('Timed out')), 5000);
 						addEventListener(
 							'sveltekit:navigation-end',
 							() => {
@@ -86,7 +86,7 @@ export const test = base.extend({
 			if (javaScriptEnabled) {
 				await page.evaluate(() => {
 					window.navigated = new Promise((fulfil, reject) => {
-						const timeout = setTimeout(() => reject(new Error('Timed out')), 2000);
+						const timeout = setTimeout(() => reject(new Error('Timed out')), 5000);
 						addEventListener(
 							'sveltekit:navigation-end',
 							() => {
@@ -111,20 +111,25 @@ export const test = base.extend({
 	in_view: async ({ page }, use) => {
 		/** @param {string} selector */
 		async function in_view(selector) {
-			// @ts-expect-error
-			return page.$eval(selector, async (element) => {
-				const { top, bottom } = element.getBoundingClientRect();
+			const isVisible = await page.isVisible(selector);
+			if (!isVisible) {
+				// @ts-expect-error
+				return page.$eval(selector, async (element) => {
+					const { top, bottom } = element.getBoundingClientRect();
 
-				if (top > window.innerHeight || bottom < 0) {
-					// slightly more useful feedback than true/false
-					return {
-						top,
-						bottom
-					};
-				}
+					if (top > window.innerHeight || bottom < 0) {
+						// slightly more useful feedback than true/false
+						return {
+							top,
+							bottom
+						};
+					}
 
-				return true;
-			});
+					return true;
+				});
+			}
+
+			return isVisible;
 		}
 
 		use(in_view);
@@ -181,7 +186,13 @@ export const test = base.extend({
 /** @type {import('@playwright/test').PlaywrightTestConfig} */
 export const config = {
 	// generous timeouts on CI, especially on windows
-	timeout: process.env.CI ? (process.platform === 'win32' ? 45000 : 30000) : 10000,
+	timeout: process.env.CI
+		? process.platform === 'win32'
+			? 45000
+			: 30000
+		: process.platform === 'win32'
+		? 15000
+		: 10000,
 	webServer: {
 		command: process.env.DEV ? 'npm run dev' : 'npm run build && npm run preview',
 		port: 3000


packages/kit/test/utils.js
```diff
@@ -181,7 +186,13 @@ export const test = base.extend({
 /** @type {import('@playwright/test').PlaywrightTestConfig} */
 export const config = {
 	// generous timeouts on CI, especially on windows
-	timeout: process.env.CI ? (process.platform === 'win32' ? 45000 : 30000) : 10000,
+	timeout: process.env.CI
```
benmccann, line 189: can you add some parentheses here or will prettier remove them? I'm having a hard time reading this

packages/kit/test/utils.js
```diff
@@ -111,20 +111,25 @@ export const test = base.extend({
...
-				const { top, bottom } = element.getBoundingClientRect();
-
-				if (top > window.innerHeight || bottom < 0) {
-					// slightly more useful feedback than true/false
-					return {
-						top,
-						bottom
-					};
-				}
+			const isVisible = await page.isVisible(selector);
```
benmccann, line 114: I might reverse the condition to indent less code:
```
if (visible) return true;
return page.$eval(...
```

packages/kit/test/apps/basics/test/test.js
```diff
@@ -1627,9 +1627,21 @@ test.describe.parallel('Routing', () => {
 			let requests = [];
 			page.on('request', (r) => requests.push(r.url()));
 
-			await app.prefetch('/routing/prefetched');
-			expect(requests.length).toBe(2);
-			expect(requests[1]).toBe(`${baseURL}/routing/prefetched.json`);
+			// also wait for network processing to complete, see
```
benmccann, line 1630: this is really great. thank you!!

benmccann:
The timeout thing is quite interesting. My first reaction was that it should never take more than 2 seconds to finish. Then I started thinking about how that could possibly be and I wonder if having more worker threads than CPUs means the worker processes slow down too much or block each other. It could explain how reducing the worker threads in https://github.com/sveltejs/kit/pull/3635 helps. I don't like that PR if it's just changing when we hit race conditions. But it might be more legitimate if we're talking about timeouts. If that's it, then raising from 2s to 5s is maybe not the cleanest solution. Maybe instead we should base the number of workers on `os.cpus().length`?

packages/kit/test/utils.js
benmccann: or maybe a better solution would be to make this always be something like `60000 / num_cpus`

benmccann: of maybe we don't need to get so fancy with this and can just set a single constant if we stop creating more workers than CPUs

PH4NTOMiki:
You're right @benmccann this PR is really awesome, and as I said in the #3635 PR that was just a little boost to get more tests to pass, it was definitely not a solution, this is a real fix. 
Awesome stuff @harvey-k

PH4NTOMiki:
> Maybe instead we should base the number of workers on `os.cpus().length`?

That command mostly returns threads(number of) and not CPU cores, just something to keep in mind, but could work.

packages/kit/test/utils.js
harvey-k: Yeah, prettier removed them...  Let me put in the simpler logic I had in the PR comment.

packages/kit/test/utils.js
harvey-k: Done!

harvey-k:
> The timeout thing is quite interesting. My first reaction was that it should never take more than 2 seconds to finish. Then I started thinking about how that could possibly be and I wonder if having more worker threads than CPUs means the worker processes slow down too much because they're fighting for CPU time. It could explain how reducing the worker threads in #3635 helps. I don't like that PR if it's just changing when we hit race conditions. But it might be more legitimate if we're talking about timeouts. If that's it, then raising from 2s to 5s is maybe not the cleanest solution. Maybe instead we should base the number of workers on `os.cpus().length`?

At face value, I agree. However, in the context of #3552, let me try and articulate the questions/concerns I see being raised:

1) Should the integration test suite favor overall test throughput (a higher `workers` value) or favor lower individual test latency?

2) How can we enforce some sort of performance baseline, and can we do so by favoring lower individual test latency in the first question?

Does that sound about right? Unfortunately, my proposed changes fail to address either of these questions. Should they be broken out into separate issues? Anyways, below is my two cents on them.

For the first item, my interpretation of the current code is that the setting of the playwright `workers` config option to a higher value is that the current decision is to attempt achieve higher test throughput. E.g, please run these fast as possible, I want to get back to coding.

With regards to the second question, IMHO, the integration test timeout values should be very gentle guardrails, focusing on if the situation under test executes successfully. Enforcing a performance baseline is a separate concern, and would be better addressed by a testing framework designed with this as a primary goal. I'm not sure playwright falls under this category, although my only experience with playwright is this PR.

Finally, on my current system os.cpus().length happens to equal the currently used value of 8. Just for discussion, the [default playwright parallelism level](https://playwright.dev/docs/api/class-testconfig#test-config-workers) is half the number of processors.

Further confusing all this is the high variance in host test execution speed.  From a recent run, [a bit over 10 minutes on windows](https://github.com/sveltejs/kit/runs/5080027286?check_suite_focus=true) to [4.5 minutes on linux](https://github.com/sveltejs/kit/runs/5080027304?check_suite_focus=true).

Apologies in advance if all of this should've been posted to the issue instead of the PR here, I'm a bit new to this.

packages/kit/test/utils.js
```diff
@@ -111,6 +111,9 @@ export const test = base.extend({
 	in_view: async ({ page }, use) => {
 		/** @param {string} selector */
 		async function in_view(selector) {
+			const visible = await page.isVisible(selector);
+			if (visible) return true;
+
```
bluwy, line 116: I remembered that `page.isVisible` checks if the element is anywhere on the page, regardless if it's in the viewport or not. For this function however, we want to check if the element is in the viewport or not. So perhaps we should remove these lines

PH4NTOMiki: I actually think you should have two test, first that isVisible, and if it fails return false, if not then check if the element is actually in the current viewport,
just my opinion

bluwy:
I'm personally still think #3635 is a good addition. More workers come with the expense of more memory used, which can likely cause longer timeouts, so we have to find the right balance point between workers and timeouts. That said this PR looks great so far (besides my comment above), thanks for digging into this!

packages/kit/test/apps/basics/test/test.js
```diff
@@ -1627,9 +1627,21 @@ test.describe.parallel('Routing', () => {
...
+			// https://playwright.dev/docs/network#network-events
+			await Promise.all([
+				app.prefetch('/routing/prefetched'),
+				page.waitForResponse(`${baseURL}/routing/prefetched.json`)
```
bluwy, line 1634: Micro nitpick: I think we should flip these two lines, according to playwright's convention.

benmccann:
> Finally, on my current system os.cpus().length happens to equal the currently used value of 8. Just for discussion, the [default playwright parallelism level](https://playwright.dev/docs/api/class-testconfig#test-config-workers) is half the number of processors.

I think this is a really sensible default. It leaves a few CPUs free so that it doesn't lock up your machine. Maybe we could do something like `process.env.ci ? num_cpus : num_cpus / 2`;

On my machine, `os.cpus().length` is also 8. I changed it to use the Playwright default and it actually ran faster going from 6m to 5m, so I don't think we'll be losing much performance :laughing:

packages/kit/test/utils.js
benmccann: I don't think we necessarily need both checks and would probably opt for just one to keep the code simpler

Another thought though - just below this we call `page.$eval`. The docs for that have:

>This method does not wait for the element to pass actionability checks and therefore can lead to the flaky tests. Use [locator.evaluate(pageFunction[, arg, options])](https://playwright.dev/docs/api/class-locator#locator-evaluate), other [Locator](https://playwright.dev/docs/api/class-locator) helper methods or web-first assertions instead.

Perhaps we should call `page.waitForFunction` instead?

harvey-k: Oof... sorry about that! I took another go at the logic with locator.

packages/kit/test/apps/basics/test/test.js
harvey-k: Done!

packages/kit/test/utils.js
```diff
@@ -111,6 +111,9 @@ export const test = base.extend({
 	in_view: async ({ page }, use) => {
 		/** @param {string} selector */
 		async function in_view(selector) {
+			const box = await page.locator(selector).boundingBox();
+			if (box !== null) return true;
+
```
bluwy, line 116: Cool, didn't know there's the `boundingBox` selector. That should make this function a lot simpler and safer. I still don't understand these lines here though. Why are we returning true when the box exists (anywhere on the page)?

I think we could probably simplify the function altogether as:

```js
const box = await page.locator(selector).boundingBox();
const viewport = await page.viewportSize();
return box && viewport && box.top <= viewport.height && box.bottom >= 0
```

I don't think we need to return an object of `top`/`bottom` like before. I don't see it used anywhere.

Rich-Harris: Yeah, the `top`/`bottom` stuff was useful at one point for testing the tests, but is no longer needed. Nothing depends on `width` or `height` so I think this can be even simpler — just `return !!box`. Will update the PR

Rich-Harris (approved):

Rich-Harris:
This is fantastic, thanks so much!

PR #3177, opened 2022-01-02
fix: skip navigation if only url hash change

bluwy:
Fixes #3105
Fixes #2667

If only the URL hash change when navigating via anchor tags, e.g. `<a href="#foo">`, skip router navigation and let the default browser behaviour to kick in.

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [x] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #3757, opened 2022-02-07
Fix client side router issue with handling hash

### LINKED ISSUE #3746: hash handling in the client-side router
### Describe the bug

When the page is visited with a hash segment(`/page1#bottom`) and for ex. there is navigation link to `/page1` (that same page just without the hash part) and the browser would do a full page reload instead of just removing the hash part (by calling `history.pushState`) and scroll to top.

### Reproduction

Visit https://kit.svelte.dev/docs#loading
Click Docs link in the top middle in navbar
The browser would do a full page reload

### Logs

_No response_

### System Info

```shell
latest sveltekit
```

### Severity

annoyance

### Additional Information

_No response_

PH4NTOMiki:

fixes #3746 issue

### Tests

### Changesets

### DIFF
--- a/.changeset/chatty-ducks-shop.md
+++ b/.changeset/chatty-ducks-shop.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+fixes bug with hash routing
--- a/.changeset/chatty-ducks-shop.md
+++ b/.changeset/chatty-ducks-shop.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-fixes bug with hash routing
+Prevent full reload when router navigates and only removes hash
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -184,7 +184,9 @@ export class Router {
 			// Ignore if <a> has a target
 			if (a instanceof SVGAElement ? a.target.baseVal : a.target) return;
 
-			// Check if new url only differs by hash
+			// Check if new url only differs by hash and use the browser default behavior in that case
+			// This will ensure the `hashchange` event is fired
+			// Removing the hash does a full page navigation in the browser, so make sure a hash is present
 			const [base, hash] = url.href.split('#');
 			if (hash && base === location.href.split('#')[0]) {
 				// Call `pushState` to add url to history so going back works.
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -185,10 +185,8 @@ export class Router {
 			if (a instanceof SVGAElement ? a.target.baseVal : a.target) return;
 
 			// Check if new url only differs by hash
-			if (
-				url.href.split('#').length > 1 &&
-				url.href.split('#')[0] === location.href.split('#')[0]
-			) {
+			const [base, hash] = url.href.split('#');
+			if (hash && base === location.href.split('#')[0]) {
 				// Call `pushState` to add url to history so going back works.
 				// Also make a delay, otherwise the browser default behaviour would not kick in
 				setTimeout(() => history.pushState({}, '', url.href));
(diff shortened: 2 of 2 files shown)


.changeset/chatty-ducks-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+fixes bug with hash routing
```
ignatiusmb, line 5:
```suggestion
Prevent full reload when router navigates and only removes hash
```

packages/kit/src/runtime/client/router.js
```diff
@@ -185,7 +185,10 @@ export class Router {
 			if (a instanceof SVGAElement ? a.target.baseVal : a.target) return;
 
 			// Check if new url only differs by hash
-			if (url.href.split('#')[0] === location.href.split('#')[0]) {
+			if (
+				url.href.split('#').length > 1 &&
+				url.href.split('#')[0] === location.href.split('#')[0]
+			) {
```
ignatiusmb, line 191: We should try to not repeat splitting more than once

```suggestion
			const [base, hash] = url.href.split('#');
			if (hash && base === location.href.split('#')[0]) {
```

ignatiusmb (commented):
LGTM, I've added a couple of suggestions

PH4NTOMiki:
Thank you @ignatiusmb for the review, committed

benmccann:
wtf. why do browsers behave this way?! I never would have expected removing a hash to result in a full page navigation. I guess we missed that when implementing https://github.com/sveltejs/kit/pull/3177

I would expect the `hashchange` event to be fired if you remove the `hash`, but it's not with this PR. But I suppose that's okay since I'm assuming browsers won't do it since they issue a full navigation - I haven't tested this though and browsers thus far have defied my expectations. What do you think @bluwy?

PH4NTOMiki:
yeah, I agree @benmccann but the funny thing is that from experience(a vanilla web app I developed a long time ago) I know that it doesn't, I actually use `history.pushState` specifically(and only) to remove the hash part, and I never had issues with it, so there must be some other player at scene. I'm not sure what tho, but as I wrote in #3746 I think this is a bigger problem than originally thought. I'll try some more fiddling with this. And I'll write anything I found in the #3746 issue.

benmccann (approved):

packages/kit/src/runtime/client/router.js
```diff
@@ -185,7 +185,8 @@ export class Router {
 			if (a instanceof SVGAElement ? a.target.baseVal : a.target) return;
 
 			// Check if new url only differs by hash
```
benmccann, line 187:
```suggestion
			// Check if new url only differs by hash and use the browser default behavior in that case
			// This will ensure the `hashchange` event is fired
			// Removing the hash does a full page navigation in the browser, so make sure a hash is present
```

benmccann:
I expanded the code comments since this section of code is a bit bug-prone and will go ahead and merge this

Thanks for the fix!!

PH4NTOMiki:
@bluwy @ignatiusmb Also i'm not sure about the `/page1#` specific use case, should that be treated by that conditional block or not(do we consider it to fall in that category or not), my first commit was just checking for the splitted array length to be longer than 1, so that meant that the condition was triggered even if the hash was just '#' but I'm not sure, but the rewrite @ignatiusmb suggested and I committed changed that, because the truthiness of empty string is false, so we're skiping that conditional block.
So which is correct to add that.

And @benmccann to clarify myself, on vanilla web app, if you do `history.pushState({}, '', location.href.split('#')[0])` it would do nothing, but if you do `location.hash = ''` it would trigger a full page reload.

PH4NTOMiki:
@benmccann I think we should have waited to clarify this (the '/page1#' edge-case) before commit, I was just writing previous comment when you merged it., I'll push a PR if needed to fix that if needed

benmccann:
I think `/page1#` should go into the `if` block based on the browser's behavior

Issue #3725, opened 2022-02-04
update hrefs to match with createIndexFiles set to false


iham:
### Describe the problem

i try to build a completely static site - as a result - from a sveltekit project.
idea is to not host the app, but its output - SSG so to speak
in general it works but adding `createIndexFiles: false` to prerender, creates hickups.

a website with two pages
/routes/index.svelte
/routes/about/index.svelte

in my __layout.svelte a nav with a tags which hrefs point to / and /about
```
<nav>
  <a href="/">HOME</a>
  <a href="/about">ABOUT</a>
</nav>
```

npm run build creates 
/build/index.html
/build/about.html

but the navigation does not respect the "changes"
```
<nav>
  <a href="/" class="svelte-1rhg2al">HOME</a>
  <a href="/about" class="svelte-1rhg2al"> ABOUT </a>
</nav>
```

i have searched the docs, examples and tried several ideas on my own, but sadly didn't find a solution.

### Describe the proposed solution

hrefs should be updated correctly when createIndexFiles is set to false and the routing changes from /about/index.html to /about.html

### Alternatives considered

if this works - the docs should be showing an example on that use case

### Importance

i cannot use SvelteKit without it

Rich-Harris:
If you're using `createIndexFiles: false`, the assumption is that you're using a static file host that responds to `/` with `index.html` and `/about` with `about.html`. If that's not the case, you shouldn't use that option.

Rich-Harris:
we no longer have `createIndexFiles` — everything is driven by `trailingSlash` — so I'll close this

PR #3761, opened 2022-02-07
[docs] additional clarification about createIndexFiles: false

benmccann:
closes https://github.com/sveltejs/kit/issues/3725


Rich-Harris:
https://github.com/sveltejs/kit/issues/3799

iham:
@Rich-Harris good point!
having the hint how to deal with endpoints `/a-page` converted to `a-page.html` when using `createIndexFiles` is set to `false` would be nice.

some mentioning that this can be a thing to setup in your webserver or some services already do it that way (like gh-pages)

my local nginx used for testing needs to be configured correctly to deal with that using `try_files $uri $uri.html $uri/index.html =404;` for example.

PR #3763, opened 2022-02-07
Fix client side router issue with handling hash

PH4NTOMiki:
Just a addition to #3757 for #3746


bluwy (approved):

bluwy:
Good catch. Gonna merge this without a changeset as this is a follow-up to #3757 and it's not released yet.

PR #3767, opened 2022-02-07
[docs] clarify page components

benmccann:
I just saw someone on Discord get confused by the term "page components" as they thought it mean the components within a page rather than the component for a page. Removing the word component seems like it's potentially clearer and doesn't really lose anything


bluwy (approved):

documentation/docs/03-loading.md
```diff
@@ -127,7 +127,7 @@ If the page you're loading has an endpoint, the data returned from it is accessi
 
 #### stuff
 
-`stuff` is passed from layout components to child layouts and page components and can be filled with anything else you need to make available. For the root `__layout.svelte` component, it is equal to `{}`, but if that component's `load` function returns an object with a `stuff` property, it will be available to subsequent `load` functions.
+`stuff` is passed from layouts to children layouts and pages and can be filled with anything else you need to make available. For the root `__layout.svelte` component, it is equal to `{}`, but if that component's `load` function returns an object with a `stuff` property, it will be available to subsequent `load` functions.
```
Conduitry, line 130: "Children" should remain "child" here I think. It's being used more as an adjective.

benmccann: yes, I read it as an adjective as well, but still thought it should be children. Maybe we should just change it to "descendant" to remove any confusion?

Rich-Harris (approved):

PR #3769, opened 2022-02-07
[docs] add a bit more help to the changelog for breaking changes

benmccann:
- there was a breaking change in 250 that we didn't call out
- we're using different formats to tag breaking changes - should we use something like Vite's semantic commit bot to enforce PR titles?
- add some more details to 234 which people seem to be getting really hung up on

### DIFF
--- a/packages/kit/CHANGELOG.md
+++ b/packages/kit/CHANGELOG.md
@@ -28,7 +28,7 @@
 
 * Add App namespace for app-level types ([#3670](https://github.com/sveltejs/kit/pull/3670))
 
-- Remove target option ([#3674](https://github.com/sveltejs/kit/pull/3674))
+- [breaking] remove target option ([#3674](https://github.com/sveltejs/kit/pull/3674))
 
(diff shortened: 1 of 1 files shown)


bluwy (commented):
I think we can mark the `Remove target option` change as breaking too. IMO SvelteKit should probably make a warning instead, but it throws an error which blocks the server startup.

benmccann:
I like the idea of changing it from an exception to a warning. It'd no longer be breaking in that case though, so maybe we shouldn't  mark it as such in the changelog

Conduitry:
SvelteKit already treats all unknown options under `kit` in `svelte.config.js` as errors, so I don't think we should treat `target` any differently in that regard. Marking it optional for the time being would just be delaying people's pain with little apparent purpose - especially if we've already released versions where including it is an error. (And leaving it forever optional, but a no-op, would be a very confusing API.)

benmccann:
ok, I've listed it as breaking in the changelog

bluwy (approved):

PR #3771, opened 2022-02-08
[docs] update docs for fallthrough routes

benmccann:
The docs needed to mention `{fallthrough: true}`


packages/kit/src/core/config/options.js
```diff
@@ -252,7 +252,7 @@ const options = object(
 			),
 
 			// TODO remove this for 1.0
-			target: error((keypath) => `${keypath} is no longer required, and should be removed`),
+			target: warn((keypath) => `${keypath} is no longer required, and should be removed`),
```
Conduitry, line 255: This seems unrelated to these changes (did you push to the wrong branch?) but, as I mentioned over in #3769, I don't think we should change this anyway.

Rich-Harris:
Now that we have multi-page docs, we have a bit more breathing room to actually explain this stuff. Updated the PR

documentation/docs/01-routing.md
```diff
@@ -277,17 +277,61 @@ A route can have multiple dynamic parameters, for example `src/routes/[category]
...
+src/routes/[b].svelte
+src/routes/[c].svelte
+src/routes/[...catchall].svelte
+src/routes/foo-[bar].svelte
+```
 
-Finally, if you have multiple routes that match a given path, SvelteKit will try each of them until one responds. For example if you have these routes...
+SvelteKit needs to know which route is being requested. To do so, it sorts them according to the following rules...
+
+- More specific routes are higher priority
```
benmccann, line 294: it's a bit unclear here whether a more specific page would beat a less specific standalone endpoint

Rich-Harris: it's immediately followed by an example that should eliminate any ambiguity though, no?

documentation/docs/01-routing.md
```diff
@@ -277,17 +277,61 @@ A route can have multiple dynamic parameters, for example `src/routes/[category]
...
+```
+
+```js
+// src/routes/[a].js
+export function get({ params }) {
+	if (params.a === 'foo-def') {
+		return { fallthrough: true };
+	}
+
+	// ...
```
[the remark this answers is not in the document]
Rich-Harris: i don't think there's much difference in clarity. if we move towards using shiki-twoslash for docs then we can't have syntactically invalid snippets, so i don't think we should make changes we know we'll have to undo later

documentation/docs/01-routing.md
benmccann: It's a lot harder to reverse engineer the rules from an example. Why not just state it up front?

PR #3780, opened 2022-02-08
Fix inconsistent rendering of shadow routes on client/server-side

### LINKED ISSUE #3727: Shadow endpoints navigation within same route
### Describe the bug

How can we navigate if route component and endpoint won't change?
Currently its not work as expected.
If we have say:
/example/[...rest].js
/example/[...rest].svelte

And try to use links on page like:
href="/example", href="/example/1", href="example/2" — navigation between them won't work (page not updates, only address bar changes)

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-x92deg?devtoolsheight=33&file=src/routes/example/[...rest].js

### Logs

_No response_

### System Info

```shell
kit next 260
```

### Severity

### Additional Information

_No response_

### LINKED ISSUE #3777: Shadow endpoints don't update with params, don't receive URL searchParams
### Describe the bug

When changing routes, the shadow endpoint is not called when expected and even when it is called it does not receive URL searchParams. Only an initial, full page-refresh when on that URL works as expected.

### Reproduction

Using out-the-box new Svelte-Kit project.

`__layout.svelte` (simply some links to click)
```
<ul>
    <li><a href="/">home</a></li>
    <li><a href="/a?page=1">a / 1</a></li>
    <li><a href="/a?page=2">a / 2</a></li>
    <li><a href="/b?page=1">b / 1</a></li>
    <li><a href="/b?page=2">b / 2</a></li>
</ul>

<slot></slot>
```

`index.svelte` (to allow changing to / from route)
```
Home
```

`[id].svelte` (route to display data from endpoint)
```
<script lang="ts">
  export let id
  export let page
</script>

<pre>id: {id} / page: {page}</pre>
```

`[id].ts` (endpoint just mirrors param / url.searchParams)
```

export const get: RequestHandler = async ({ params, url }) => {
  const { id } = params
  const page = url.searchParams.get('page')

  return {
    body: {
      id,
      page,
    }
  }
}
```

Clicking the links changes the id param and / or the page url searchParam. The id param is picked up on the initial entry to a route, but the searchParam is null. Neither changes on subsequent clicks. A full page reload makes both be recognized and reflected in the page. Going back to home and then to the route makes the id param be picked up for that initial change (and page be null), but is then frozen again if navigating to a different id page.

Adding back the original load function to `[id].svelte`, even calling the new shadow endpoint, makes things work:

```
  import type { Load } from '@sveltejs/kit'

  export const load: Load = async ({ fetch, props, params, url }) => {
    const { id } = params
    const page = url.searchParams.get('page')

    const resp = await fetch(`/${id}/__data.json?page=${page}`)
    const data = await resp.json()

    return {
      props: { ...data }
    }
  }
</script>
```

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.2
    CPU: (6) x64 Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
    Memory: 1.67 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.17.3 - /usr/local/bin/node
    npm: 7.20.2 - /usr/local/bin/npm
  Browsers:
    Chrome: 98.0.4758.80
    Chrome Canary: 100.0.4876.0
    Edge: 98.0.1108.43
    Firefox: 96.0.3
    Safari: 15.3
    Safari Technology Preview: 15.4
  npmPackages:
    @sveltejs/adapter-auto: ^1.0.0-next.18 => 1.0.0-next.18 
    svelte: ^3.44.0 => 3.46.4
```

### Severity

### Additional Information

_No response_

### LINKED ISSUE #3778: Routing endpoints missing searchParams
### Describe the bug

The new [routing endpoints](https://kit.svelte.dev/docs#routing-endpoints) are great, but I think I've come across a bug. If you use `goto` to set a querystring/searchParam, the routing endpoint can't read the searchParam when using client-side routing. But it can if you refresh the page (server-side routing).

### Reproduction

Use this basic [reproduction on stackblitz](https://stackblitz.com/edit/sveltejs-kit-template-default-viwedt?file=src/routes/index.svelte).

1. Click _Open in a New Window_ at the top-right of stackblitz, since the built-in browser doesn't update the URL visibly
2. Enter a value in the text box and submit the form (client-side routing)
3. You'll see the querystring printed is blank, but the browser's URL contains a query string
4. Refresh the page (server-side routing) and you'll see the querystring comes through
5. Press back, then forwards (client-side routing), and you'll see it's blank again

I've left a couple lines commented out to illustrate the conditions under which it works as expected:
- If you use a `load` function instead of a routing endpoint (uncomment the load function)
- Or if you use a standard `<form>` submission, instead of `goto` (uncomment the `<form>` tag and comment out the one beneath)

### Logs

_No response_

### System Info

```shell
System: (note: reproduced the issue on stackblitz)
    OS: macOS 12.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 90.80 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.3.1 - /opt/homebrew/bin/node
    npm: 8.3.0 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 98.0.4758.80
    Firefox: 96.0.1
    Safari: 15.2
  npmPackages:
    @sveltejs/adapter-auto: ^1.0.0-next.17 => 1.0.0-next.17 
    @sveltejs/kit: ^1.0.0-next.260 => 1.0.0-next.260 
    svelte: ^3.44.0 => 3.46.3
```

### Severity

### Additional Information

My _guess_, not being familiar with sveltekit's codebase, is that the issue could be with how `goto` sets HTML5 history - perhaps it's not including the query/searchParams there. But that's a rather wild guess.

PH4NTOMiki:

This PR fixes inconsistent URL handling for shadow routes on the client/server-side. and because of that it also 
fixes #3777
fixes #3778

### Tests

### Changesets


PH4NTOMiki:
@benmccann @bluwy please take a look, I'm pretty sure about the params part of the issue, I'll try to deal with the URLSearchParams tomorrow if they don't work with this.

PH4NTOMiki:
should also fix #3777 and #3778

packages/kit/src/runtime/client/renderer.js
```diff
@@ -615,6 +615,9 @@ export class Renderer {
 		if (props) {
 			// shadow endpoint props means we need to mark this URL as a dependency of itself
 			node.uses.dependencies.add(url.href);
+			for (const key in params) {
```
PH4NTOMiki, line 618: This 3 lines may be redundant because we are checking `has_shadow` instead of uses.dependencies, but maybe for being safe, in case we need this info for something else.

PH4NTOMiki: By 3 lines I mean 618 and the 2 below it

benmccann:
@PH4NTOMiki if you put "fixes https://github.com/sveltejs/kit/issues/3727 https://github.com/sveltejs/kit/issues/3777 https://github.com/sveltejs/kit/issues/3778" in the PR description it will only link one of the issues. you need to put the word fixes before each issue. I will edit the PR description to update

dominikg (approved):

packages/kit/src/runtime/client/renderer.js
benmccann: @Rich-Harris suggested we might not want to automatically add all params because while it makes sense in isolation, it may be too eager if we add a layout equivalent of shadow endpoints

PH4NTOMiki: @benmccann Yeah, I understand, but what do you suggest, how do we deal with that now, maybe add some check if it's the last branch, or something like that, 
but when/if you add layout shadows, how would you track the params that are used then?

Rich-Harris (changes requested):
Thank you! This needs tests :)

Please only fix one issue per PR — I know it's tempting to do several simultaneously, but it slows things down (if this was two PRs we could add a test for the `url.search` addition and merge that immediately), makes reviews harder, and muddies the changelog and git history.

On the invalidation issue, the most straightforward logic would just be whether or not the pathname or query has changed. We need to have a proper think about what invalidation logic would look like if we had layout-shadowing endpoints, but this PR probably isn't the right place for it.

PH4NTOMiki:
@Rich-Harris thanks Rich for reviewing, and thank you for the hard work on Svelte, I was planning to fix just #3727 but when i was reading the comments on that issue, many said the related/duplicate problem was with URLSearchParams and since they were kinda related I decided to fix them both in this PR.
So should I convert this PR just for the `url.search` and create a separate PR for invalidation logic?
or close this, and create 2 new?

Thankss

Rich-Harris:
Whichever you prefer — converting this PR would be fine (we always squash commits before merging)

PH4NTOMiki:
@Rich-Harris done

timwis:
Thanks for this fix, @PH4NTOMiki! Looking forward to trying it out.

PH4NTOMiki:
@Rich-Harris @benmccann can we merge this?

PH4NTOMiki:
@bluwy can we merge this, cause it's a pretty serious bug.

benmccann:
@PH4NTOMiki lint and the tests are failing. can you fix those issues?

Rich-Harris (approved):

PR #3785, opened 2022-02-09
Allow HTML shadow endpoint responses

pzuraq:

Putting this up to see if there's interest in adding this functionality (which restores something that shadow endpoints were previously capable of). If there is, I can add tests and fix the types.

Linked issue: https://github.com/sveltejs/kit/issues/3532#issuecomment-1033201702

### Tests

### Changesets


packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -58,9 +58,12 @@ export async function load_node({
 	 */
 	let set_cookie_headers = [];
 
-	/** @type {import('types/helper').Either<import('types/endpoint').Fallthrough, import('types/page').LoadOutput>} */
+	/** @type {import('types/helper').Eithber<import('types/endpoint').Fallthrough, import('types/page').LoadOutput>} */
```
benmccann, line 61:
```suggestion
	/** @type {import('types/helper').Either<import('types/endpoint').Fallthrough, import('types/page').LoadOutput>} */
```

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -58,9 +58,12 @@ export async function load_node({
...
 	let loaded;
 
+	/** @type Response | undefined */
```
benmccann, line 64:
```suggestion
	/** @type {Response | undefined} */
```

Rich-Harris:
Brought this up to date and added some tests.

I'm a little uneasy about it though. What happens when SvelteKit loads a non-JSON endpoint during client-side navigation? Does it error? Reload the page, causing the request to be repeated? It will also add some confusion and complexity when we add layout endpoints as well, since it wouldn't make sense to respect an HTML response from a `__layout.js` file.

My gut says this isn't something SvelteKit should support.

pzuraq:
So our use case for this was as follows.

- We have profile pages for our users (example: https://www.bitski.com/@pzuraq)
- Some users have custom profile pages which are implemented in whatever framework they choose, all that matters is that it produces HTML/css/js (examples: https://www.bitski.com/@mlsxsecretwalls, https://www.bitski.com/@rtfkt)
- Custom pages are hosted on our CDN. When a user makes a request to the profile page, we first check to see if the page exists in the CDN, and if it does we then proxy through the HTML from the CDN. When I first figured out SvelteKit could do this it blew my mind 😄
- When SvelteKit has already loaded and we are navigating _client-side_, we first make a HEAD request to see if the custom page exists. If it does, we then do a full page navigation and reload the page to get the new HTML.

Currently we do all of this in hooks since it stopped working in endpoints, and given this is an unusual use case I would be ok with just keeping the functionality in hooks. We are going to be exploring more customization options in the future though, and I think SvelteKit’s flexibility here is really a good thing as it enables exploration and trying new things out. I’m not sure we would have figured this out if the only option was to use hooks in the first place.

pzuraq:
👍 yeah that’s fair, just wanted to make sure this wasn’t an oversight. Sounds good!

PR #3786, opened 2022-02-09
Re-evaluate even when transitioning between pages of the same component

phi-jp:
It seems that the processing in the script tag is not re-evaluated because the same component in routes is reused even if the params are different.

```html:routes/posts/[id].svelte
<script>
  // not called after the second time
  console.log('Hello, svelte!');
</script>

<div>
  <a href='/posts/1'>post 1</a>
</div>
<div>
  <a href='/posts/2'>post 2</a>
</div>
```

I tried resetting the component once so that it would be evaluated again.

Please merge it if you like.

----

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -354,6 +354,11 @@ export class Renderer {
 
 		if (this.started) {
 			this.current = navigation_result.state;
+			
+			this.root.$set({
+				components: [],
+			});
+			await tick();
 
 			this.root.$set(navigation_result.props);
 			this.stores.navigating.set(null);
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -354,9 +354,9 @@ export class Renderer {
 
 		if (this.started) {
 			this.current = navigation_result.state;
-			
+
 			this.root.$set({
-				components: [],
+				components: []
 			});
 			await tick();
 


Rich-Harris:
This is intentional. If you want to respond to navigations between `/a/b` and `/a/c` where both point to `/a/[x]`, use `beforeNavigate` and `afterNavigate`: https://kit.svelte.dev/docs/modules#$app-navigation

phi-jp:
@Rich-Harris Thank you for your easy-to-understand advice.

The code in [id] .svlete worked fine with that.
If you want to re-evaluate the processing of a child component, do you mean to use `beforeNavigate / afterNavigate`?

Below is an example of reproduction.

```html:[id].svelte
<script context="module">
  export async function load({params}) {
    let post = {
      id: params.id,
      title: `Page ${params.id}`,
    };
    return {
      props: {
        post,
      }
    };
  }
</script>

<script>
  export let post;
  import Post from "./Post.svelte";
</script>

<Post post='{post}' />
```

```html:Post.svelte
<script>
  import {onMount} from 'svelte';
  export let post;
  export let title = '';

  onMount(() => {
    // not called after the second time
    title = `${post.id}. ${post.title}`;
  });
  
  let getPostTitle = () => {
    return `${post.id}. ${post.title}`;
  };
</script>

<!-- Transiting the same component produces different results -->
<div>{post.id}. {post.title}</div>
<div>{title}</div>
<div>{getPostTitle()}</div>
```

phi-jp:
@babichjacob @Rich-Harris thank you.
I solved it if I followed the advice.

I am currently developing multiple products in Japan and are considering SvelteKit.
It seems that this will proceed.

I'm really thankful to you.
We hope for the development of Svelte / SvelteKit.

PR #3790, opened 2022-02-09
Precompress wasm files

stephtr:
When enabling the `precompress` option, WebAssembly files currently aren't compressed.
In my case however, compression brings me down from 4 MB to 1.5 MB, which seems to be in line with [what can be observed in the wild](https://almanac.httparchive.org/en/2021/webassembly#can-we-improve-compression).

- [ ] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs<br />_just a small change_

### Tests
- [ ] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`<br />_Test failing is independent of this change._

### Changesets

### DIFF
--- a/packages/adapter-static/index.js
+++ b/packages/adapter-static/index.js
@@ -50,7 +50,7 @@ export default function ({ pages = 'build', assets = pages, fallback, precompres
  * @param {string} directory
  */
 async function compress(directory) {
-	const files = await glob('**/*.{html,js,json,css,svg,xml}', {
+	const files = await glob('**/*.{html,js,json,css,svg,xml,wasm}', {
 		cwd: directory,
 		dot: true,
 		absolute: true,
--- a/.changeset/purple-dragons-work.md
+++ b/.changeset/purple-dragons-work.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-static': patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/adapter-static/index.js
```diff
@@ -50,7 +50,7 @@ export default function ({ pages = 'build', assets = pages, fallback, precompres
  * @param {string} directory
  */
 async function compress(directory) {
-	const files = await glob('**/*.{html,js,json,css,svg,xml}', {
+	const files = await glob('**/*.{html,js,json,css,svg,xml,wasm}', {
```
benmccann, line 53: It looks like this line also exists in `adapter-node`. We'd probably want to keep them in sync
https://github.com/sveltejs/kit/blob/e364454f6f7f057c758aab197e6015b6b93f287d/packages/adapter-node/index.js#L81

Rich-Harris (approved):

Rich-Harris:
Thank you. Rather than worrying about keeping these in sync, this logic probably belongs in Kit itself — will open a fresh issue for that

stephtr:
Thank you too!

PR #3793, opened 2022-02-09
fixes shadow hydration escaping

### LINKED ISSUE #3773: [bug?] Shadow Endpoints fails to hydrate big data sent to their matching page because it is truncated. causing SyntaxError: Unexpected end of JSON input
### Describe the bug

i originally came to this issue when trying to use shadow endpoints for the first time:

image

EDIT: there are more discovery points in subsequent comments below.

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-ahljwi?file=src/routes/about.svelte

`npm run dev`

and go to the /about endpoint

### Logs

```shell
9:03:33 PM [vite] hmr update /src/routes/about.svelte
```

### System Info

this is my mac envinfo since stackblitz doesnt seem to be able to run envinfo

```shell
  System:
    OS: macOS 11.4
    CPU: (8) x64 Apple M1
    Memory: 51.47 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.4.1 - /usr/local/bin/node
    npm: 7.18.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 98.0.4758.80
    Safari: 14.1.1
  npmPackages:
    @sveltejs/adapter-netlify: ^1.0.0-next.44 => 1.0.0-next.44 
    @sveltejs/kit: ^1.0.0-next.260 => 1.0.0-next.260 
    svelte: ^3.44.3 => 3.46.4 
```

### Severity

blocking an upgrade

### Additional Information

_No response_

PH4NTOMiki:

fixes #3773
there are many better ways to solve this, if anyone has a suggestion.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/utils/escape.js
+++ b/packages/kit/src/utils/escape.js
@@ -1,6 +1,5 @@
 /** @type {Record<string, string>} */
-const escape_json_string_in_html_dict = {
-	'"': '\\"',
+const escape_json_in_html_dict = {
 	'<': '\\u003C',
 	'>': '\\u003E',
 	'/': '\\u002F',
@@ -15,7 +14,38 @@ const escape_json_string_in_html_dict = {
 	'\u2029': '\\u2029'
 };
 
-/** @param {string} str */
+/** @type {Record<string, string>} */
+const escape_json_string_in_html_dict = {
+	'"': '\\"',
+	...escape_json_in_html_dict
+};
+
+/**
+ * escape a json string to be embedded into a script data tag
+ *
+ * <script>
+ *   output here
+ * </script>
+ * @param {string} str
+ */
+export function escape_json_in_html(str) {
+	return escape(
+		str,
+		escape_json_in_html_dict,
+		(code) => `\\u${code.toString(16).toUpperCase()}`
+	);
+}
+
+/**
+ * escape a json string to be embedded into a larger json object thats going to be embedded in html
+ *
+ * <script>
+ *   {
+ *     "foo":"output here"
+ *   }
+ * </script>
+ * @param {string} str
+ */
 export function escape_json_string_in_html(str) {
 	return escape(
 		str,
--- a/packages/kit/src/utils/escape.js
+++ b/packages/kit/src/utils/escape.js
@@ -29,11 +29,7 @@ const escape_json_string_in_html_dict = {
  * @param {string} str
  */
 export function escape_json_in_html(str) {
-	return escape(
-		str,
-		escape_json_in_html_dict,
-		(code) => `\\u${code.toString(16).toUpperCase()}`
-	);
+	return escape(str, escape_json_in_html_dict, (code) => `\\u${code.toString(16).toUpperCase()}`);
 }
 
 /**
--- a/packages/kit/src/utils/escape.js
+++ b/packages/kit/src/utils/escape.js
@@ -21,7 +21,7 @@ const escape_json_string_in_html_dict = {
 };
 
 /**
- * Escape a JSON string to be embedded into a `<script>` tag
+ * Escape a stringified JSON object that's going to be embedded in a `<script>` tag
  * @param {string} str
  */
 export function escape_json_in_html(str) {
@@ -29,7 +29,7 @@ export function escape_json_in_html(str) {
 }
 
 /**
- * Escape a stringified JSON object that's going to be embedded in a `<script>` tag
+ * Escape a string JSON value to be embedded into a `<script>` tag
  * @param {string} str
  */
 export function escape_json_string_in_html(str) {
--- a/packages/kit/src/utils/escape.js
+++ b/packages/kit/src/utils/escape.js
@@ -29,13 +29,7 @@ export function escape_json_in_html(str) {
 }
 
 /**
- * escape a json string to be embedded into a larger json object thats going to be embedded in html
- *
- * <script>
- *   {
- *     "foo":"output here"
- *   }
- * </script>
+ * Escape a stringified JSON object that's going to be embedded in a `<script>` tag
  * @param {string} str
  */
 export function escape_json_string_in_html(str) {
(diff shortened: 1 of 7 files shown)


PH4NTOMiki:
I also tried using devalue package but it's the same, html becomes malformed. This way is vulnerable to XSS probably so not the best way, but just a suggestion.

PH4NTOMiki:
Actually I found out that Next.js is using the same(similar) strategy with regex like this PR:
https://github.com/vercel/next.js/blob/694407450638b037673c6d714bfe4126aeded740/packages/next/server/htmlescape.ts

I'll improve this a little but this isn't that bad.

dominikg:
we have a utility for that https://github.com/sveltejs/kit/blob/59badb515755a9294365e8084f9beefa10a90773/packages/kit/src/utils/escape.js#L19

dominikg:
and it would be great to have a test in https://github.com/sveltejs/kit/tree/master/packages/kit/test/apps/basics/src/routes/xss  so that we can catch regressions.

PH4NTOMiki:
> 

Thanks, I actually adopted the code from Next.js in the latest commit, but I'll implement that code now you suggested because DRY.

dominikg:
@PH4NTOMiki  i pushed a fix for the invalid double-qoute escape and added a test. hope you don't mind. this is kind of urgent :grimacing:

PH4NTOMiki:
> @PH4NTOMiki i pushed a fix for the invalid double-qoute escape and added a test. hope you don't mind. this is kind of urgent grimacing

Yeah, of course, thank you

packages/kit/test/apps/basics/test/test.js
```diff
@@ -1992,4 +1992,14 @@ test.describe.parallel('XSS', () => {
 		// @ts-expect-error - check global injected variable
 		expect(await page.evaluate(() => window.pwned)).toBeUndefined();
 	});
+
+	test('no xss via shadow endpoint', async ({ page }) => {
+		await page.goto('/xss/shadow')
```
benmccann, line 1997:
```suggestion
		await page.goto('/xss/shadow');
```

benmccann:
Looks like you'll need to run `pnpm format`

packages/kit/src/utils/escape.js
```diff
@@ -15,7 +14,34 @@ const escape_json_string_in_html_dict = {
...
+	'"': '\\"',
+	...escape_json_in_html_dict
+};
+
+/**
+ * escape a json string to be embedded into a script data tag
+ *
+ * <script>
+ *   output here
+ * </script>
```
benmccann, line 28:
```suggestion
 * Escape a JSON string to be embedded into a `<script>` tag
```

packages/kit/src/utils/escape.js
```diff
@@ -15,7 +14,34 @@ const escape_json_string_in_html_dict = {
...
+}
+
+/**
+ * escape a json string to be embedded into a larger json object thats going to be embedded in html
+ *
+ * <script>
+ *   {
+ *     "foo":"output here"
+ *   }
+ * </script>
```
benmccann, line 42:
```suggestion
 * Escape a stringified JSON object that's going to be embedded in a `<script>` tag
```

benmccann (commented):
I had a couple suggestions for the jsdocs, but otherwise lgtm

I would like to get another set of eyes from @Conduitry though since he was more involved in https://github.com/sveltejs/kit/pull/2534

dominikg:
these docs describe where the escaped result can be safely used, not how the return value looks like.

see https://github.com/sveltejs/kit/blob/647131e00149cb5299042eedeb880efcd855a2df/packages/kit/src/runtime/server/page/load_node.js#L260

it's actually a string prop on a larger json thats added to a script block elsewhere later. Not fond of this complexity but it is what it is.

I'd like to merge this first and we can improve the docs in a followup PR.

packages/kit/src/utils/escape.js
```diff
@@ -15,7 +14,24 @@ const escape_json_string_in_html_dict = {
...
+ * @param {string} str
+ */
+export function escape_json_in_html(str) {
+	return escape(str, escape_json_in_html_dict, (code) => `\\u${code.toString(16).toUpperCase()}`);
+}
+
+/**
+ * Escape a stringified JSON object that's going to be embedded in a `<script>` tag
+ * @param {string} str
+ */
```
benmccann, line 34: oh wait. I think my documentation suggestions were incorrect. this should be better
```suggestion
 * Escape a stringified JSON object that's going to be embedded in a `<script>` tag
 * @param {string} str
 */
export function escape_json_in_html(str) {
	return escape(str, escape_json_in_html_dict, (code) => `\\u${code.toString(16).toUpperCase()}`);
}

/**
 * Escape a string JSON value to be embedded into a `<script>` tag
 * @param {string} str
 */
```

benmccann:
Something in this PR seems to have broken https://kit.svelte.dev/docs

PH4NTOMiki:
@dominikg what do you think?
@benmccann could you roll over back to https://github.com/sveltejs/kit/pull/3793/commits/886ad462f5881694c2856d93f78b64d92029632b
I think that should work?

PH4NTOMiki:
Or should I try to PR that again?

PH4NTOMiki:
@benmccann also, is there any change that that Vite update could be causing anything? maybe together with this, or alone?

PH4NTOMiki:
I think it's the `"` fix that @dominikg pushed, https://github.com/sveltejs/kit/pull/3793/commits/8c8e4e68b166a0afd323eb52c7d87c37c93d59a0 this commit, I think is the cause

benmccann:
It works if I revert this line:
```
body += `<script type="application/json" data-type="svelte-props">${escape_json_in_html(s(shadow_props))}</script>`;
```

PH4NTOMiki:
Isn't that the current/latest line?

benmccann:
Yes, and it's broken

PR #3797, opened 2022-02-09
Fixes hydration issue caused by #3793

PH4NTOMiki:
### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.
Should fix issue cause by PR #3793

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [ ] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #3798, opened 2022-02-09
[fix] replace broken escaping with a working version

benmccann:
Cleaned up version of https://github.com/sveltejs/kit/pull/3797

### DIFF
--- a/packages/kit/src/utils/escape.js
+++ b/packages/kit/src/utils/escape.js
@@ -15,7 +15,7 @@ const escape_json_in_html_dict = {
 };
 
 /** @type {Record<string, string>} */
-const escape_json_string_in_html_dict = {
+const escape_json_value_in_html_dict = {
 	'"': '\\"',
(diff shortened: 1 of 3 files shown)


packages/kit/src/utils/escape.js
```diff
@@ -25,42 +25,40 @@ const escape_json_string_in_html_dict = {
...
+				/**
+				 * @type { Record<string, string> }
+				 */
+				 const escape_lookup = {
+					'&': '\\u0026',
+					'>': '\\u003e',
+					'<': '\\u003c',
+					'\u2028': '\\u2028',
+					'\u2029': '\\u2029'
+				};
```
Rich-Harris, line 40: can we hoist this like the other dictionaries, so we're not recreating this object unnecessarily?

benmccann: done

Rich-Harris (approved):

PR #3801, opened 2022-02-09
move createIndexFiles option to adapter-static

### LINKED ISSUE #3799: Remove `prerender.createIndexFiles` option in favour of adapter control
### Describe the problem

In #2632 we added the `prerender.createIndexFiles` option, and I'm pretty sure it was a mistake. On most platforms you shouldn't be touching it at all. The only time the user should be configuring it is when they're targeting a static file host (i.e. they're using `adapter-static`) and the index files are problematic for some reason (e.g. they cause the host to redirect `/foo` to `/foo/`, breaking relative URLs).

There may be situations where an adapter needs to set this option, with knowledge of how it will interact with the platform (e.g. Netlify's default is to map `about.html` to `/about` and `about/index.html` to `/about/` — I think we're probably getting this wrong at the moment).

Remove `prerender.createIndexFiles` in favour of making it part of the adapter API...

```js
const prerendered = await builder.prerender({
  all: false,
  dest,
  index: false
});
```

...and exposing it via the `adapter-static` options.

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

Rich-Harris:
Fixes #3799. 

One thing I've noticed about this option is that people (including myself) get confused talking about it because the default is `true`, and it feels unusual to have to explicitly set an option to `false`. Two possible solutions:

1. we come up with a better name for it that means the opposite, allowing a default of `false`
2. we change the default to `false`, and make `about/index.html` instead of `about.html` opt-in

Might be worth doing a quick survey of popular static file hosts and seeing if `about.html` will work as expected in the majority of cases.

Also, it might make sense to vary the default based on the value of `trailingSlash`.

* [ ] update docs

---

### Tests

### Changesets

### DIFF
--- a/.changeset/flat-stingrays-talk.md
+++ b/.changeset/flat-stingrays-talk.md
@@ -1,0 +1,6 @@
+---
+'@sveltejs/adapter-static': patch
+'@sveltejs/kit': patch
+---
+
+Move createIndexFiles option from kit to adapter-static
--- a/packages/adapter-static/index.d.ts
+++ b/packages/adapter-static/index.d.ts
@@ -5,6 +5,7 @@ interface AdapterOptions {
 	assets?: string;
 	fallback?: string;
 	precompress?: boolean;
+	createIndexFiles?: boolean;
 }
 
 declare function plugin(options?: AdapterOptions): Adapter;
--- a/packages/adapter-static/index.js
+++ b/packages/adapter-static/index.js
@@ -7,7 +7,13 @@ import zlib from 'zlib';
 const pipe = promisify(pipeline);
 
 /** @type {import('.')} */
-export default function ({ pages = 'build', assets = pages, fallback, precompress = false } = {}) {
+export default function ({
+	pages = 'build',
+	assets = pages,
+	fallback,
+	precompress = false,
+	createIndexFiles = true
+} = {}) {
 	return {
 		name: '@sveltejs/adapter-static',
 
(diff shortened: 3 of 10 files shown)


benmccann:
This seems like a good idea to me. It'd be nice to do it automatically for `adapter-netlify` (I guess based on the value of `trailingSlash`?) and some of the others

Rich-Harris:
Via https://www.zachleat.com/web/trailing-slash/, I found https://github.com/slorber/trailing-slash-guide. My takeaway is that we should indeed vary it based on `trailingSlash` (which effectively means defaulting to `false`).

I even wonder if it's worth exposing the option at all. I suspect there aren't really any cases where you'd need to control `createIndexFiles` independently of `trailingSlash`, and on the basis that APIs are easy to add and hard to remove, we should make the breaking change now and see if people complain.

Rich-Harris:
Is that true? If I create `file.html` and `folder/index.html` in the relevant directory and fire up Apache, this is what I get:

image

image

`/file` renders correctly, `/file/` 404s. `/folder` redirects to `/folder/` which renders correctly. Every webserver seems to have subtly different semantics, but AFAICT prettifying `foo.html` and `bar/index.html` to `/foo` and `/bar/` respectively seems pretty consistent

benmccann:
Huh. TIL. I didn't think you could leave off the `.html` extension in the `file` case. Consider me corrected then

Issue #3649, opened 2022-01-31
escape_json_string_in_html causes ~23 MB of allocations for 1 MB of JSON


tv42:
### Describe the bug

I was troubleshooting why my SvelteKit app SSR was consuming a lot of RAM. There's likely other reasons, but this jumped out of the profile:

![image](https://user-images.githubusercontent.com/419909/151869026-3d1e0cb9-786f-4c34-a652-06040a4c8bff.png)

`escape_json_string_in_html` (which is just a wrapper for `escape`, https://github.com/sveltejs/kit/blob/59badb515755a9294365e8084f9beefa10a90773/packages/kit/src/utils/escape.js#L53) allocates 23 MB for a 1MB input.

### Reproduction

From a Page's `load` function, make a fetch to a 1 MB JSON file, and return it as a prop.

The SSR HTML will end with something like

```
 <script type="application/json" data-type="svelte-data" data-url="/api/graphql" data-body="foo">...</script>
```

containing said HTML, but creating that HTML apparently takes about 23x the size of the JSON in allocations.

### System Info

```shell
System:
    OS: Linux 5.15 NixOS 21.11 (Porcupine) 21.11 (Porcupine)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 6.33 GB / 62.61 GB
    Container: Yes
    Shell: 5.1.8 - /nix/store/xdxv22yb7qnjfyf997r9fx804761z14c-bash-interactive-5.1-p8/bin/bash
  Binaries:
    Node: 16.13.0 - /nix/store/26lslfmvwyhspppbinznh111fyaxx62z-nodejs-16.13.0/bin/node
    npm: 8.1.0 - /nix/store/26lslfmvwyhspppbinznh111fyaxx62z-nodejs-16.13.0/bin/npm
  Browsers:
    Firefox: 96.0.2
  npmPackages:
    @sveltejs/adapter-cloudflare-workers: ^1.0.0-next.30 => 1.0.0-next.30
    @sveltejs/kit: ^1.0.0-next.239 => 1.0.0-next.239
    svelte: ^3.46.1 => 3.46.2
```
```


### Severity

annoyance

### Additional Information

_No response_

bluwy:
@tv42 there has been changes to the function (renamed `escape_json_in_html`). Can you confirm that it still takes up large amount of memory with the latest version?

benmccann:
[...]

PR #3804, opened 2022-02-09
refactor: simplify escape_json_in_html util

dominikg:
followup of https://github.com/sveltejs/kit/pull/3798

moved JSON.stringify into `escape_json_in_html` and reused it in place of escape_json_value_in_html

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -1,7 +1,7 @@
 import { normalize } from '../../load.js';
 import { respond } from '../index.js';
 import { s } from '../../../utils/misc.js';
-import {escape_json_in_html} from '../../../utils/escape.js';
+import { escape_json_in_html } from '../../../utils/escape.js';
 import { is_root_relative, resolve } from '../../../utils/url.js';
 import { create_prerendering_url_proxy } from './utils.js';
 import { is_pojo } from '../utils.js';
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -255,8 +255,8 @@ export async function load_node({
 							if (!opts.body || typeof opts.body === 'string') {
 								// the json constructed below is later added to the dom in a script tag
 								// make sure the used values are safe
-								const statusNumber = Number(response.status);
-								if (isNaN(statusNumber)) {
+								const status_number = Number(response.status);
+								if (isNaN(status_number)) {
 									throw new Error(
 										`response.status is not a number. value: "${
 											response.status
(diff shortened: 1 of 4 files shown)


packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -257,7 +257,7 @@ export async function load_node({
 								fetched.push({
 									url: requested,
 									body: /** @type {string} */ (opts.body),
-									json: `{"status":${response.status},"statusText":${s(response.statusText)},"headers":${s(headers)},"body":"${escape_json_value_in_html(body)}"}`
+									json: `{"status":${response.status},"statusText":${s(response.statusText)},"headers":${s(headers)},"body":${escape_json_in_html(body)}}`
```
benmccann, line 260: just to be safe I think we should double check that the `status` is a number. I have no idea whether that would already be validated on all platforms
```suggestion
									json: `{"status":${Number(response.status)},"statusText":${s(response.statusText)},"headers":${s(headers)},"body":${escape_json_in_html(body)}}`
```

benmccann:
lgtm. thanks for this!

packages/kit/src/runtime/server/page/load_node.js
dominikg: this would result in `"status": NaN` for non numberish which would then result in parse errors down the road.
Should we validate that response.status is an actual integer in the valid http status range here or is it done elsewhere already?

dominikg: a check could look like this

```js
const statusNumber = Number(response.status);
if (isNaN(statusNumber)) {
 throw new Error(`response.status is not a number. value: "${response.status}" type: ${typeof response.status}`);
}
```
and then use statusNumber in the json string.

benmccann: that sounds fine. whether it's worth adding the extra validation to throw an error or not is something I debated. I think either way would be fine. I just want to be sure we're not inadvertently injecting something from the user into the page without any validation or escaping with the `Number` alone would do

bluwy:
I wonder if this would help https://github.com/sveltejs/kit/issues/3649 too

packages/kit/src/runtime/server/page/load_node.js
dominikg: i think throwing here is better, the subsequent parse error would be less tracable. added

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -253,11 +253,21 @@ export async function load_node({
 							}
 
 							if (!opts.body || typeof opts.body === 'string') {
+								// the json constructed below is later added to the dom in a script tag
+								// make sure the used values are safe
+								const statusNumber = Number(response.status);
```
benmccann, line 258: this should be `status_number`

dominikg: updated

PR #3809, opened 2022-02-09
upgrade to TypeScript 4.5

benmccann:


packages/kit/src/packaging/index.js
```diff
@@ -54,14 +54,17 @@ export async function make_package(config, cwd = process.cwd()) {
...
-			if (fs.existsSync(dts_path)) {
-				fs.unlinkSync(dts_path);
-
-				const dir = path.dirname(dts_path);
-				if (fs.readdirSync(dir).length === 0) {
-					fs.rmdirSync(dir);
+			const base = svelte_ext ? file : file.slice(0, -ext.length);
+			const dts_files = [base + '.d.ts', base + '.d.mts', base + '.d.cts'];
+			for (const dts_file of dts_files) {
+				const dts_path = path.join(package_dir, dts_file);
```
ignatiusmb, line 60: We can probably simplify this to

```suggestion
			for (const e of ['.d.ts', '.d.mts', '.d.cts']) {
				const dts_path = path.join(package_dir, base + e);
```

package.json
```diff
@@ -39,7 +39,7 @@
 		"playwright-chromium": "^1.17.0",
 		"prettier": "^2.5.0",
 		"rollup": "^2.60.2",
-		"typescript": "~4.4.4"
+		"typescript": "^4.5.5"
```
bluwy, line 42: Should we keep this as `~4.5.5` given that TypeScript minor versions could be breaking too?

packages/create-svelte/templates/default/package.json
```diff
@@ -13,7 +13,7 @@
 		"@sveltejs/kit": "workspace:*",
 		"svelte": "^3.44.0",
 		"svelte-preprocess": "^4.9.8",
-		"typescript": "^4.4.3"
+		"typescript": "^4.5.5"
```
bluwy, line 16:
```suggestion
		"typescript": "~4.5.5"
```

Might be good to recommend it here too.

benmccann: done

package.json
benmccann: done

dummdidumm (approved):

PR #3814, opened 2022-02-10
Fixes client-side not navigating if same shadow route module used

### LINKED ISSUE #3727: Shadow endpoints navigation within same route

PH4NTOMiki:

Fixes #3727
Needs tests

### Tests

### Changesets

### DIFF
--- a/.changeset/tame-mayflies-yell.md
+++ b/.changeset/tame-mayflies-yell.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+fix issue with not navigating if the same shadow route module is used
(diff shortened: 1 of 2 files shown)


PH4NTOMiki:
@bluwy @benmccann can we merge this?
It's quite a bug.

Rich-Harris:
Thank you. I went with a slightly different solution in #3925 — eventually we'll probably want a more granular solution when we have shadowed layouts, which means applying the same kind of dependency detection on the server side (and probably communicating it to the client via custom headers), which means that populating `node.uses` is arguably preferable to adding special casing in the change detection block

Issue #3759, opened 2022-02-07
Shadow endpoint -- issue rendering with no get endpoint


florianlouvet:
### Describe the bug

When using the new shadow endpoints, it seems it is required to define a `get` endpoint so that sveltekit renders the page (even if no data needs to be loaded and passed down to the page).
Not providing any `get` endpoints returns a 405 error.

<img width="1054" alt="Screenshot 2022-02-07 at 16 03 25" src="https://user-images.githubusercontent.com/21657360/152813743-75fffc4c-c005-48c8-9631-4a28ec17fc29.png">

Providing an empty `get` function solves this problem.
```javascript
export async function get() {
    return {
        body: { }
    };
}
```
It would be interesting not to have to provide an empty `get` function.

### Reproduction

Minimal reproduction can be found here: https://github.com/florianlouvet/shadow-endpoint-svelte-bug

When navigating to the `/login` page the 405 is raised. In the `login.js` endpoint file there is the empty get endpoints necessary to render the page (solving this issue in the meantime).

### Logs

```shell
405

get method not allowed

Error: get method not allowed
    at load_shadow_data (file:///Users/florianlouvet/own/shadow-endpoint-svelte-bug/.svelte-kit/runtime/server/index.js:1841:12)
    at async load_node (file:///Users/florianlouvet/own/shadow-endpoint-svelte-bug/.svelte-kit/runtime/server/index.js:1523:5)
    at async respond$1 (file:///Users/florianlouvet/own/shadow-endpoint-svelte-bug/.svelte-kit/runtime/server/index.js:2120:15)
    at async render_page (file:///Users/florianlouvet/own/shadow-endpoint-svelte-bug/.svelte-kit/runtime/server/index.js:2321:19)
    at async resolve (file:///Users/florianlouvet/own/shadow-endpoint-svelte-bug/.svelte-kit/runtime/server/index.js:2564:11)
    at async respond (file:///Users/florianlouvet/own/shadow-endpoint-svelte-bug/.svelte-kit/runtime/server/index.js:2497:20)
    at async file:///Users/florianlouvet/own/shadow-endpoint-svelte-bug/node_modules/@sveltejs/kit/dist/chunks/index.js:244:24
```

### System Info

```shell
System:
    OS: macOS 12.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 163.64 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
[...]

PR #3816, opened 2022-02-10
add empty get handler if not given in shadow endpoint

alexanderschau:
This PR adds a `get` handler with an empty body to shadow endpoints that do not have one.

Solves #3759

### Tests

### Changesets


packages/kit/test/apps/basics/test/test.js
```diff
@@ -463,6 +463,12 @@ test.describe.parallel('Shadowed pages', () => {
 
 		expect(headers.head).toEqual(headers.get);
 	});
+
+	test.only('Works with missing get handler', async ({ page, clicknav }) => {
```
Conduitry, line 467:
```suggestion
	test('Works with missing get handler', async ({ page, clicknav }) => {
```

Is there some way to make .only fail in CI like we have in the core Svelte repo?

Rich-Harris: opened https://github.com/sveltejs/kit/pull/3922

Rich-Harris:
Thank you. It turned out this only solved half the problem — SSR was fine, but client-side navigation still failed. I moved the test to the main test suite (which covers both SSR and client-side navigation with the same tests) and added some logic for that case

PR #3819, opened 2022-02-10
Add back HN example

benmccann:
Should this be in the `sites` directory or an `examples` directory? If the latter, then should we rename `sites/kit.svelte.dev` to just site?

There's not much point to having this be the sole project in the sites repo. It's annoying to me not to be able to test against it anymore. I could setup pnpm `overrides`, but then I have to keep making sure not to accidentally check it in, etc. `kit-sandbox` is pretty useless for playing around with because it doesn't have any pages in it. I could create some, but it's just an extra headache I don't want to have to deal with compared to having a more fleshed out site to play around with.


benmccann:
Closing in favor of https://github.com/sveltejs/sites/pull/300

PR #3823, opened 2022-02-10
[fix] spelling of .DS_Store

malinowskip:
The spelling of `.DS_Store` in the default config is incorrect. As a result, `.DS_Store` files are not filtered out from the `files` array passed to the service worker.

### Tests

### Changesets

### DIFF
--- a/.changeset/empty-islands-lie.md
+++ b/.changeset/empty-islands-lie.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-fix casing of .DS_Store in the default config for `serviceWorker`
+fix casing of `.DS_Store` in the default config for `serviceWorker`


Conduitry:
Should we just use a case insensitive regex, or is this _always_ `.DS_Store`?

Rich-Harris (approved):

Rich-Harris:
thanks!

> Should we just use a case insensitive regex, or is this _always_ `.DS_Store`?

i've never seen it cased any other way (except in our docs, apparently)

Conduitry:
This isn't just changing the docs, it's also changing Kit, so this would need to include a changeset in order to be released.

malinowskip:
OK, I’ve added a changeset. Let me know in case any other adjustments are needed.

.changeset/empty-islands-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+fix casing of .DS_Store in the default config for `serviceWorker`
```
benmccann, line 5:
```suggestion
fix casing of `.DS_Store` in the default config for `serviceWorker`
```

benmccann (approved):

Issue #3331, opened 2022-01-14
Unable to crawl non-soft nav links


john-michael-murphy:
### Describe the problem

The NYT has run into a scenario where we need svelte kit to crawl links that have soft nav disabled. AFAIK the only way to disable soft nav is to set `rel=‘external’`, which disables crawling. 

https://github.com/sveltejs/kit/blob/2097f00f131aef42ffe4e8fd16da6e302ce22901/packages/kit/src/core/adapt/prerender/crawl.js#L171

### Describe the proposed solution

It would be great if there were a way to selectively bypass sveltekit's link interception, without disabling crawling. 

### Importance

would make my life easier

Rich-Harris:
External links are now crawled, so I'll close this: https://github.com/sveltejs/kit/pull/3826

PR #3826, opened 2022-02-10
prerender external links

Rich-Harris:
In #1679 we stopped crawling links with `rel="external"` during prerendering. `rel="external"` is often used to force a full page navigation (instead of a client-side navigation), and in those cases is orthogonal to the question of whether something should be prerendered (see e.g. #3331).

It seems like the simplest thing to do would simply be to follow those links, but @v1ack perhaps you can elaborate on the motivation for #1679?

The alternative would be to introduce a new construct for disabling client-side navigation on a link (i.e. a new `sveltekit:blah` attribute) but it would be better to avoid that unless we really need it.

### Tests

### Changesets


v1ack:
External needs when another page on the same host handled by another service. 

For example, my case is 
- main site on GitHub pages (v1ack.github.io) running on svelte kit 
- docs site for lib (https://v1ack.github.io/svelte-virtual-scroll-list/) running from another repository 

If I place link to lib site on main site without #1679 - crawler fails with 404.

Maybe the logic of crawler can be changed to something like this: 
- if we faced to `external` - try to crawl 
- if `external` page doesn't exist - make a warning (not error) and continue crawling

Rich-Harris:
Yeah, I think I'm inclined to say that this can be handled via the `prerender.onError` option:

```js
// svelte.config.js
export default {
  kit: {
    prerender: {
      onError: ({ status, path }) => {
        if (path.startsWith('/svelte-virtual-scroll-list')) return;
        throw new Error(`${status} ${path}`);
      }
    }
  }
};
```

v1ack:
Ok, it looks good 
But I think it better to be mentioned in docs, near `external`option, maybe as a link to FAQ

Rich-Harris:
Added some docs: https://github.com/sveltejs/kit/pull/3826/commits/6a7f51b20e2a4ff3c256e312cff9b9cfabc2625a

babichjacob:
Partially related discussion on Discord for those curious: https://discord.com/channels/457912077277855764/750401468569354431/943224194634764369

PR #3827, opened 2022-02-10
Fix Cloudflare Workers adapter target

pzuraq:
When attempting to upload a SvelteKit app which contains modern ES
features such as optional chaining or nullish coalescing, workers is
unable to parse the built output.

```
Module parse failed: Unexpected token (13066:47)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|       if (typeof window !== "undefined") {
|         const windowAny = window;
>         const dataLayer = windowAny.dataLayer ?? [];
|         windowAny.dataLayer = dataLayer;
|         gtag = function() {
Error: webpack returned an error. You may be able to resolve this issue by running npm install
```

I could not find any official documentation of what version of ES
workers support, but ES2020 is the year that nullish coalescing was
introduced, and setting it to ES2019 appears to fix the issue.

Unsure how to test this

### Tests

### Changesets

### DIFF
--- a/.changeset/wicked-terms-agree.md
+++ b/.changeset/wicked-terms-agree.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-cloudflare-workers': patch
+---
+
+Fix Cloudflare adapter targets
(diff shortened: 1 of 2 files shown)


benmccann (approved):

bluwy:
Why is webpack being used to bundle stuff in cloudflare workers? The [docs](https://github.com/sveltejs/kit/tree/master/packages/adapter-cloudflare-workers#basic-configuration) recommend to set `type = "javascript"` in the wrangler config, so webpack shouldn't be used in the first place.

PR #3829, opened 2022-02-10
Copy endpoint response Headers instead of setting directly

pzuraq:
Headers that are created as part of a Response object (e.g. directly
from `fetch` or via `Response.redirect`) are not mutable in some
environments, including the browser and Cloudflare Workers. If an
endpoint returns an instance of a Response currently, it will attempt
to mutate the headers, which throws an error. This only fails in
Cloudflare, so it's hard to catch before a deployment.

Thought about it a bit and I'm unsure how to test this since it's environment specific. We could try to hijack the `Response` global potentially.

### Tests

### Changesets

### DIFF
--- a/.changeset/fast-coats-attack.md
+++ b/.changeset/fast-coats-attack.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Copy endpoint response Headers instead of setting directly
+[fix] avoid mutating response `Headers`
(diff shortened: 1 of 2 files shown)


benmccann (approved):

.changeset/fast-coats-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Copy endpoint response Headers instead of setting directly
```
benmccann, line 5:
```suggestion
[fix] avoid mutating response `Headers`
```

Issue #2390, opened 2021-09-08
Prerender adds trailing slash even for json/webmanifest files


Konakona-chan:
### Describe the bug

Hi, for my project I need to generate webmanifest during build. With `trailingSlash: 'always'` build fails, because prerender.js extracts link from `<link rel="manifest" href="/manifest.webmanifest" />` and fails with:
```
Attempt to 404 /manifest.webmanifest/ (linked from /)
```

### Reproduction

svelte.config.js
```
import preprocess from 'svelte-preprocess';
import adapter from '@sveltejs/adapter-static';

/** @type {import('@sveltejs/kit').Config} */
const config = {
	preprocess: preprocess(),

	kit: {
		adapter: adapter(),
		target: '#svelte',
		trailingSlash: 'always'
	}
};

export default config;
```

index.svelte
```
<svelte:head>
    <link rel="manifest" href="/manifest.webmanifest" />
</svelte:head>
```

manifest.webmanifest.ts
```
import type { RequestHandler } from "@sveltejs/kit";


export const get: RequestHandler = () => {
    return { body: {/* code skipped */} };
};
```

`npm run build` fails with:
```
> Using @sveltejs/adapter-static
> 404 /manifest.webmanifest/ (linked from /)
```

### System Info

```shell
System:
    OS: Windows 10 10.0.17763
    CPU: (8) x64 Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
    Memory: 3.17 GB / 15.88 GB
  Binaries:
    Node: 14.17.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.4 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 7.19.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 93.0.4577.63
    Edge: Spartan (44.17763.1911.0)
    Internet Explorer: 11.0.17763.1852
  npmPackages:
    @sveltejs/adapter-static: ^1.0.0-next.18 => 1.0.0-next.18
    @sveltejs/kit: next => 1.0.0-next.164
    svelte: ^3.34.0 => 3.42.4
```

### Severity

serious, but I can work around it

### Additional Information

Renaming `manifest.webmanifest.ts` to `manifest.json.ts` does not help.

The only workaround I know is to use .kit.trailingSlash="ignore" option (I can not use "never" for compatibility reasons).

rmunn:
[...]

PR #3851, opened 2022-02-11
Skip trailingSlash for paths with extension

kdheepak:
Fixes https://github.com/sveltejs/kit/issues/2390

Currently, when using `trailingSlash: 'always'` in sveltekit's `svelte.config.js`, the `normalize` function adds a trailing slash to _all_ paths that don't contain a trailing slash. This causes issues when you want to use `trailingSlash: 'always'` _and_ also want to host some dynamic content on an endpoint, for example when using `rss.xml` as a endpoint.

This PR changes the behavior of `normalize` to not add a trailing slash when the filename in the path contains an extension.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/adapt/prerender/prerender.js
+++ b/packages/kit/src/core/adapt/prerender/prerender.js
@@ -106,8 +106,8 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 		// has_ext('rss.xml') === true
 		// has_ext('robots.txt') === true
 		// has_ext('manifest.webmanifest') === true
-		const fname = path.split('/').pop()
-		return fname.slice((Math.max(0, fname.lastIndexOf(".")) || Infinity) + 1).length > 0;
+		const fname = path.split('/').pop();
+		return fname.slice((Math.max(0, fname.lastIndexOf('.')) || Infinity) + 1).length > 0;
 	}
 
 	/**
--- a/packages/kit/src/core/adapt/prerender/prerender.js
+++ b/packages/kit/src/core/adapt/prerender/prerender.js
@@ -97,11 +97,25 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 	});
 
 	/**
+	 * Checks if path has extension
+	 * @param {string} path
+	 */
+	function has_ext(path) {
+		// has_ext('.htcaccess') === false
+		// has_ext('path/to/.htcaccess') === false
+		// has_ext('rss.xml') === true
+		// has_ext('robots.txt') === true
+		// has_ext('manifest.webmanifest') === true
+		const fname = path.split('/').pop()
+		return fname.slice((Math.max(0, fname.lastIndexOf(".")) || Infinity) + 1).length > 0;
+	}
+
+	/**
 	 * @param {string} path
 	 */
 	function normalize(path) {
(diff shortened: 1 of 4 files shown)


packages/kit/src/core/adapt/prerender/prerender.js
```diff
@@ -96,12 +96,26 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 		}
 	});
 
+	/**
+	 * Checks if path has extension
+	 * @param {string} path
+	 */
+	function has_ext(path) {
+		// has_ext('.htcaccess') === false
+		// has_ext('path/to/.htcaccess') === false
```
Conduitry, line 105:
```suggestion
		// has_ext('.htaccess') === false
		// has_ext('path/to/.htaccess') === false
```

packages/kit/src/core/adapt/prerender/prerender.js
```diff
@@ -96,12 +96,26 @@ export async function prerender({ cwd, out, log, config, build_data, fallback, a
 		}
 	});
 
+	/**
+	 * Checks if path has extension
+	 * @param {string} path
+	 */
+	function has_ext(path) {
+		// has_ext('.htaccess') === false
```
benmccann, line 104: I don't understand. Why do you want a trailing slash on `.htaccess`?

kdheepak: We wouldn't want to do that, but I added that in as an example of where a `.file` would have a trailing slash. If you had a route that was `htaccess` (or any name for that matter) it would still add a trailing slash if you used `trailingSlash: 'always'`. This behavior is the same as it was before this PR.

This PR only addresses when the filename in the path "has an extension", and `.htaccess` is not a file with an extension. Thoughts?

kdheepak: I think having a per page / per endpoint way to override the value of the `trailingSlash` option would be awesome, and may make this PR unnecessary. I tried looking into how to do that but quickly got overwhelmed. If someone can direct me, I'm happy to take that on.

benmccann: I kind of think dot files should be considered as having an extension by default if we're going to handle files with extensions

If we wanted to make `trailingSlash` more configurable I think it'd have to be a resolve option similar to `ssr`. I'll let @Rich-Harris weigh in on what the best API for this might be though

kdheepak:
The only way I know how to fix this lint:

Screen Shot 2022-02-11 at 12 20 11 PM

is by making an explicit check for undefined. Typescript has `arr.pop()` returning `undefined` when the `arr` array is empty, which I don't believe will ever happen in this case.

packages/kit/src/core/adapt/prerender/prerender.js
kdheepak: Alternatively, never adding a trailing slash for endpoints makes sense in my opinion, regardless of the `trailingSlash` configuration value. The behavior for pages can be left as is.

kdheepak:
Then I will close this PR :)

kdheepak:
@Rich-Harris thoughts on adding a `export let trailingSlash: 'always'` option on a per page or per endpoint level?

Issue #3848, opened 2022-02-11
Return redirects and file information from `prerender`


Rich-Harris:
### Describe the problem

As of right now, https://kit.svelte.dev/docs is erroring if you hit it directly. As far as I can tell this is because of a combination of two bugs:

* Vercel doesn't know that `/docs` should redirect straight to `/docs/introduction`, because we're not preserving redirects during prerendering. Because of that it's SSRing `/docs` 
* SSRing `/docs` fails in production, and I _think_ that's because the layout is fetching `docs.json` here, which is failing because that file can't be dynamically server-rendered (it needs to read files that don't exist in the production deployment; it's designed to be prerendered) https://github.com/sveltejs/kit/blob/d788cf27e0f6014bd0773c70e6200f2bfe6d96ef/sites/kit.svelte.dev/src/routes/docs/__layout.svelte#L5

### Describe the proposed solution

If `prerendered` contained redirect information we could add that directly to the generated `routes` config, saving us from attempting to server render `/docs`. (I'll open a separate issue for reading prerendered dependencies without attempting to server-render)

### Importance

would make my life easier

PR #3858, opened 2022-02-11
prerendering overhaul

Rich-Harris:
This began as a fix for #3848 but has turned into something more comprehensive, as I ran into some encoding bugs while working on it

### Tests

### Changesets

### DIFF
--- a/packages/adapter-cloudflare-workers/files/entry.d.ts
+++ b/packages/adapter-cloudflare-workers/files/entry.d.ts
@@ -8,3 +8,10 @@ declare module 'MANIFEST' {
 	export const manifest: SSRManifest;
 	export const prerendered: Set<string>;
 }
+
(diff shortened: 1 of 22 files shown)


Rich-Harris:
Looks like this does indeed fix the problem:

* ✅ https://kit-svelte-dev-git-gh-3848-svelte.vercel.app/docs
* ❌ https://kit.svelte.dev/docs

kdheepak:
I have two related comments:

1) I was trying to test this implementation of the normalize function in the old code but the normalize function is being passed in `/rss.xml/` which then returns `/rss.xml/`. I haven't tested this PR to see if that is an issue or not here, but I thought I'd make a note about it

2) Pages that are error codes (i.e. `404/`) are left as is. Does it make sense to always convert `404/index` to `404.html`? Similar question for other response codes.

PR #3868, opened 2022-02-12
Adds context to platform in SvelteKit endpoints

### LINKED ISSUE #2807: Expose cloudflare specific functionality (event.waitUntil) in hooks

mcfadyeni:

### Changes
- Adds the `context` variable, which Pages Functions provides, into `platform` in endpoints and hooks.
- This is important because `context` contains `waitUntil`, which quite a few error-reporting platforms and other packages make use of (such as Sentry via [Toucan.js](https://github.com/robertcepa/toucan-js)).
- Closes #2807.

### Tests

### Changesets

### DIFF
--- a/.changeset/sharp-moles-confess.md
+++ b/.changeset/sharp-moles-confess.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-cloudflare': patch
+---
+
+Allows access of the context variable on platform inside endpoints. Closes #2807.
--- a/.changeset/sharp-moles-confess.md
+++ b/.changeset/sharp-moles-confess.md
@@ -2,4 +2,4 @@
 '@sveltejs/adapter-cloudflare': patch
 ---
 
-Allows access of the context variable on platform inside endpoints. Closes #2807.
+Add `context` to `event.platform` object
--- a/packages/adapter-cloudflare/README.md
+++ b/packages/adapter-cloudflare/README.md
@@ -53,7 +53,7 @@ When configuring your project settings, you must use the following settings:
 
 ## Environment variables
 
-The [`env`](https://developers.cloudflare.com/workers/runtime-apis/fetch-event#parameters) object, containing KV namespaces etc, is passed to SvelteKit via the `platform` property, meaning you can access it in hooks and endpoints:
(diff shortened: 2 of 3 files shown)


Rich-Harris:
thanks!

half2me:
Is there a reason why `adapter-cloudflare-workers` missed out on this? Can we add this there too?

isaac-mcfadyen:
> Is there a reason why `adapter-cloudflare-workers` missed out on this? Can we add this there too?

From what it looks like here, `adapter-cloudflare-workers` isn't passed `platform` at all (so no `env` or `context`):
https://github.com/sveltejs/kit/blob/3d728d2f59062cf40d13896cf8709ebac8110c83/packages/adapter-cloudflare-workers/files/entry.js#L53

Issue #3707, opened 2022-02-03
Better source comments/documentation in packages


JEbertPrime:
### Describe the problem

Familiarizing myself with the internal package source code has been frustrating. Some of the more complex functions can be fairly opaque. The JSDoc type declarations are useful, but it seems like a missed opportunity to be using them exclusively as type declarations when we could add more descriptive information. I understand that investing too heavily into internal documentation pre-1.0 might cause some headaches if there are significant changes to the codebase, but I think the benefits probably outweigh that.

### Describe the proposed solution

- Add basic descriptions to JSDoc annotations. Short function and  parameter descriptions like the ones described on the [JSDoc website](https://jsdoc.app/about-getting-started.html) are probably all this needs.
- [nice to have] Configuration for JSDoc site generation to create some human-readable docs for development purposes.
- Add some instructions for updating the JSDoc descriptions to `CONTRIBUTING.md`

### Alternatives considered

Wait until 1.0 release, when more stability can be expected for the internal packages.

### Importance

nice to have

### Additional Information

I'm happy to work on this incrementally as I go through the codebase and learn how it all works.

benmccann:
> I'm happy to work on this incrementally as I go through the codebase and learn how it all works.

That would be very much loved by all. It's been a barrier to anyone trying to get involved in the Svelte codebase including some of the maintainers from other parts of the Svelte ecosystem. There's very few people with the knowledge to do add comprehensive docs, but I think if folks can just learn little areas of the codebase and document it once they have a good understanding like you suggested that's probably one of the most realistic ways to improve it.

JEbertPrime:
Cool! Do you have any preference for how progress should be made on this? My initial thought is to open up directory-specific PRs so changes can be reviewed/approved in small-ish quantities (e.g. 'Documentation update for `./packages/kit/src/runtime/app`').

benmccann:
Oh, I just realized we're in the SvelteKit repo and not Svelte core :laughing:  I just assumed Svelte core because internal docs are way more of a pain point there...

Doing updates for a particular directory would make sense. I think especially at the beginning for the first one or two I wouldn't do too much so that you can get feedback on the initial PR and make sure it's going in a direction everyone likes

[...]

PR #3870, opened 2022-02-12
Source documentation for `./packages/kit/src/core/create_app` - issue #3707

JEbertPrime:
### Documentation standards I made up
Bases on some quick checking of other libraries' documentation practices, I followed some rules that seemed pretty sensible.
- I wrote documentation for any function or object which had an existing JSDoc type declaration
- Descriptions of objects/functions were written with sentence case and ended with a period
- Parameter descriptions are separated from the parameter they describe with a dash, are not capitalized, and do not end in a period
- In cases where a parameter type is defined as an object, I re-wrote to match the [format described here](https://jsdoc.app/tags-param.html#parameters-with-properties). It should be noted that VS Code's typescript hover info doesn't include these descriptions, but this seemed like the best option since it's recommended by JSDoc

These are the only things I did that I feel should be standardized in any way (and really, this isn't super important). This seems like a good enough way to handle incrementally adding more info to source documentation - it took minimal effort to write out short descriptions while I was reading through these functions. I started with this folder since it only has one file, after this I'll be working through the other folders in `src/core`.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/create_app/index.js
+++ b/packages/kit/src/core/create_app/index.js
@@ -53,8 +53,10 @@ function trim(str) {
  * @param {string} base - the base path to prepend to imports
  */
 function generate_client_manifest(manifest_data, base) {
-	/**Maps relative path of component to the component's index in `manifest_date.components`.
-	 * @type {Record<string, number>} */
+	/**
+	 * Maps relative path of component to the component's index in `manifest_date.components`.
+	 * @type {Record<string, number>}
+	 */
 	const component_indexes = {};
 
 	/**
--- a/packages/kit/src/core/create_app/index.js
+++ b/packages/kit/src/core/create_app/index.js
@@ -4,7 +4,7 @@ import { s } from '../../utils/misc.js';
 import { mkdirp } from '../../utils/filesystem.js';
 
 /**
- * Maps file contents to file path
+ * Maps file contents to file path.
  *  @type {Map<string, string>} */
 const previous_contents = new Map();
 
@@ -48,15 +48,17 @@ function trim(str) {
 }
 
 /**
- * Generates code for the `manifest.js` file which imports svelte components and exports the app's `routes`, an an array of arrays containing a pattern to match the route and two arrays of components (`[__layout, page]` and `[error]`)
+ * Generates code for the `manifest.js` file which imports svelte components and exports the app's `routes`, an an array of arrays containing a pattern to match the route and two arrays of components (`[__layout, page]` and `[error]`).
  * @param {ManifestData} manifest_data - object containing paths to the apps assets, user defined components, routes, and the app's __layout and error components.
  * @param {string} base - the base path to prepend to imports 
  */
 function generate_client_manifest(manifest_data, base) {
-	/** @type {Record<string, number>} */
+	/**Maps relative path of component to the component's index in `manifest_date.components`.
+	 * @type {Record<string, number>} */
 	const component_indexes = {};
 
-	/** @param {string} c */
+	/**Wraps `path.relative`. 
+	 * @param {string} c - component path */
 	const get_path = (c) => path.relative(base, c);
 	
 	const components = `[
--- a/packages/kit/src/core/create_app/index.js
+++ b/packages/kit/src/core/create_app/index.js
@@ -48,8 +48,8 @@ function trim(str) {
 }
 
 /**
- * Generates code which imports svelte components and exports the app's `routes`, an an array of arrays containing a pattern to match the route and an array of components ([layout, component, error])
- * @param {ManifestData} manifest_data - object containing paths to the apps assets, user defined components, routes, and __layout and error components
+ * Generates code for the `manifest.js` file which imports svelte components and exports the app's `routes`, an an array of arrays containing a pattern to match the route and two arrays of components (`[__layout, page]` and `[error]`)
+ * @param {ManifestData} manifest_data - object containing paths to the apps assets, user defined components, routes, and the app's __layout and error components.
  * @param {string} base - the base path to prepend to imports 
  */
 function generate_client_manifest(manifest_data, base) {
@@ -58,7 +58,7 @@ function generate_client_manifest(manifest_data, base) {
 
 	/** @param {string} c */
 	const get_path = (c) => path.relative(base, c);
-
+	
 	const components = `[
 		${manifest_data.components
 			.map((component, i) => {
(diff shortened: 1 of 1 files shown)


packages/kit/src/core/create_app/index.js
```diff
@@ -35,21 +38,26 @@ export function create_app({ manifest_data, output, cwd = process.cwd() }) {
...
- * @param {string} base
+ * Generates code for the `manifest.js` file which imports svelte components and exports the app's `routes`, an an array of arrays containing a pattern to match the route and two arrays of components (`[__layout, page]` and `[error]`).
+ * @param {ManifestData} manifest_data - object containing paths to the apps assets, user defined components, routes, and the app's __layout and error components.
+ * @param {string} base - the base path to prepend to imports
  */
 function generate_client_manifest(manifest_data, base) {
-	/** @type {Record<string, number>} */
+	/**Maps relative path of component to the component's index in `manifest_date.components`.
+	 * @type {Record<string, number>} */
 	const component_indexes = {};
```
benmccann, line 57:
```suggestion
	/**
	 * Maps relative path of component to the component's index in `manifest_date.components`.
	 * @type {Record<string, number>}
	 */
	const component_indexes = {};
```

packages/kit/src/core/create_app/index.js
```diff
@@ -21,11 +25,11 @@ export function write_if_changed(file, code) {
 /** @typedef {import('types/internal').ManifestData} ManifestData */
 
 /**
- * @param {{
- *   manifest_data: ManifestData;
- *   output: string;
- *   cwd: string;
- * }} options
+ * Updates or creates two files - manifest.js and root.svelte - in the specified output directory.
+ * @param {Object} options
```
benmccann, line 29: I don't think this is correct. There aren't four parameters, but a single parameter. See https://stackoverflow.com/a/31573441 for suggestions

benmccann: Ah, sorry if the `options.` was already there. I think I missed that on my first time around

packages/kit/src/core/create_app/index.js
```diff
@@ -3,12 +3,16 @@ import path from 'path';
...
+/**
+ * Maps file contents to file path.
+ * @type {Map<string, string>}
+ */
 const previous_contents = new Map();
 
 /**
- * @param {string} file
- * @param {string} code
+ * Checks `previous_contents` map for `file` key and performs strict inequality comparison of `code` against the value, then writes `code` to `file` if there are changes.
```
benmccann, line 13: The description is currently a bit too long and isn't really much easier to read than the code :smile: How about?
```suggestion
 * Checks `previous_contents` map for `file` and writes `code` to `file` if there are changes.
```

packages/kit/src/core/create_app/index.js
```diff
@@ -21,11 +25,11 @@ export function write_if_changed(file, code) {
...
 
 /**
- * @param {{
- *   manifest_data: ManifestData;
- *   output: string;
- *   cwd: string;
- * }} options
+ * Updates or creates two files - manifest.js and root.svelte - in the specified output directory.
+ * @param {Object} options
+ * @param {ManifestData} options.manifest_data - details the apps assets, user defined components, routes, and __layout and error components
```
benmccann, line 30: I worry this description has the potential to get outdated if we change what is contained within the `ManifestData`. I'd rather ensure that the `ManifestData` type is well documented and make this description more minimal

packages/kit/src/core/create_app/index.js
```diff
@@ -62,7 +75,10 @@ function generate_client_manifest(manifest_data, base) {
 			.join(',\n\t\t\t\t')}
 	]`.replace(/^\t/gm, '');
 
-	/** @param {string[]} parts */
+	/**
+	 * Matches components by path and returns an array of those components accessed within the `c`array.
```
benmccann, line 79:
```suggestion
	 * Matches components by path and returns an array of those components accessed within the `c` array.
```

benmccann: `accessed`? I'm not sure what that's trying to communicate, but it doesn't sound right

packages/kit/src/core/create_app/index.js
```diff
@@ -35,21 +39,30 @@ export function create_app({ manifest_data, output, cwd = process.cwd() }) {
...
 function trim(str) {
 	return str.replace(/^\t\t/gm, '').trim();
 }
 
 /**
- * @param {ManifestData} manifest_data
- * @param {string} base
+ * Generates code for the `manifest.js` file which imports svelte components and exports the app's `routes`, an an array of arrays containing a pattern to match the route and two arrays of components (`[__layout, page]` and `[error]`).
+ * @param {ManifestData} manifest_data - object containing paths to the apps assets, user defined components, routes, and the app's __layout and error components.
+ * @param {string} base - the base path to prepend to imports
```
benmccann, line 53: it doesn't import anything. we could simplify these descriptions a lot
```suggestion
 * Generates code for the `manifest.js` file which exports the app's `routes` and the root layout/error components.
 * @param {ManifestData} manifest_data - a listing of paths to the app's components and assets.
 * @param {string} base - the base path to prepend to imports
 * @return {string}
```

packages/kit/src/core/create_app/index.js
Rich-Harris: I'd go further — I'd suggest that the `previous_contents` stuff is all implementation detail, irrelevant to understanding what the function's purpose is. All you really need to know is that it writes a file if the contents have changed. But then again you can figure that out from the name of the function!

packages/kit/src/core/create_app/index.js
JEbertPrime: My bad  - I glanced at the array and saw the import statements, didn't see they were wrapped in functions.

packages/kit/src/core/create_app/index.js
JEbertPrime: I was trying to communicate that the array returned contains references to the import functions listed in the `c` array within the generated code. I'm unsure of the best way to describe the form of the arrays it returns, since just saying they contain the components seems a little wrong

packages/kit/src/core/create_app/index.js
JEbertPrime: Agreed. Probably best to just refer to the type declaration for any objects which are typed.

JEbertPrime:
Some takeaways from the suggested changes so far:
- Need better linting for description format
- Descriptions are only needed for larger, more abstract functions (possibly this should just be limited to exported functions)
- Favor brevity over exhaustive descriptions
- In cases where a type is already defined, refer to that type's definition rather than re-describing
I'll commit these changes when I get a chance

Rich-Harris:
Closing this is as several of the places where descriptions were added in this PR already have descriptions in the current `master` (albeit in a different module, as the codebase has changed somewhat). Thanks

Issue #3105, opened 2021-12-24
Focus broken when navigating with hash


AnishDe12020:
### Describe the bug

In an effort to make tab navigation better on my site, I was implementing a skip to content button which would skip the navbar and directly go to the content. I have implemented it as it is normally implemented but I am running into a small problem. Normally, focussing on the skip to content button and then pressing enter gets you to the content, and pressing the tab key after that makes you cycle through the content. In my implementation, I am brought to the skip to content button twice before I can cycle through the site's content. 

### Reproduction

Add a skip to content link to a page (basically a link that is the first thing to come into focus and shows up when tabbed on landing on the website and else remains hidden). Set the `href` to an id the link points to (usually something like `main` or `content` that is after that page's navbar/header). Navigate to the site, press tab, press enter, again press tab and the skip to content link will again show up.

### Logs

```shell
None. Check additional information for screen recording, details on code, source code, and a hosted site where it can be reproduced.
```

### System Info

```shell
System:
    OS: Linux 5.15 Pop!_OS 21.10
    CPU: (8) x64 Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
    Memory: 3.19 GB / 15.52 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 16.9.1 - ~/.nvm/versions/node/v16.9.1/bin/node
    Yarn: 1.22.17 - /usr/bin/yarn
    npm: 7.21.1 - ~/.nvm/versions/node/v16.9.1/bin/npm
  Browsers:
    Brave Browser: nightly
    Chrome: 96.0.4664.110
    Firefox: 95.0.1
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.3 
    @sveltejs/kit: next => 1.0.0-next.201 
    svelte: ^3.44.0 => 3.44.2
```

### Severity

serious, but I can work around it

### Additional Information

Hosted Preview - https://portfolio-nlaow838l-anishde12020.vercel.app/
Source code (THIS SPECIFIC COMMIT) - https://github.com/AnishDe12020/portfolio/commit/d2406ffeb65e0f1da29d6d95e8a2c323051dd3b6

<details>
<summary>Details on the code</summary>
I have added a button to the layout of my website, that is in `__layout.svelte`. Here is the code:

```svelte
<script>
  import "../app.css";
  import WipBanner from "../components/wip-banner.svelte";
  import Header from "../components/header.svelte";
[...]

PR #3873, opened 2022-02-12
scroll restoration overhaul

PH4NTOMiki:

### Tests

### Changesets

Should fix double recording navigation issue(still checking on that) and removes the need for `scroll` event listener.
Related to #3835


Rich-Harris:
Afraid I'm seeing some wildly inconsistent behaviour here:

https://user-images.githubusercontent.com/1162160/153728097-9f09ecb7-686e-4b61-9dc0-2e5061163d5b.mov

packages/kit/src/runtime/client/router.js
```diff
@@ -72,6 +72,20 @@ export class Router {
...
+	save_scroll_state() {
+		const new_state = {
+			...(history.state || {}),
+			'sveltekit:scroll': scroll_state()
+		};
+
+		// need this in case it was set in previous state
+		if (new_state['sveltekit:scroll'].x === 0 && new_state['sveltekit:scroll'].y === 0) {
+			delete new_state['sveltekit:scroll'];
+		}
```
Rich-Harris, line 84: Not sure I understand the purpose of this?

PH4NTOMiki:
@Rich-Harris @benmccann we have a decision to make, we are currently decision-less about this thing:
so, the code that's currently being used is not firing the events/hooks, (`beforeNavigate` and that stuff) if the user is navigating to the hash on the same page, we're re-rendering the same page, and then telling the browser to deal with the navigation event itself(that's happening because we never call `event.preventDefault()` in that if check)
so the choices are:
a) remove that code (if) completely and things should work correctly, the `beforeNavigate` would fire, but I tried this and some tests fail
b) do this:

```javascript
if (hash !== undefined && base === location.href.split('#')[0]) {
    event.preventDefault();
    this.save_scroll_state();
    const info = this.parse(url);
    if (info) {
        await this.renderer.update(info, [], false);
        location.hash = hash; // this will automatically pushState and set the state to null and then trigger hashchange event
        this.save_scroll_state(); // by calling this method, it will also replaceState with object instead of null, we could also use history.replaceState({}, '', url.href);
    }
    return;
}
```

or even simple(but without re-rendering):

```javascript
if (hash !== undefined && base === location.href.split('#')[0]) {
    event.preventDefault();
    this.save_scroll_state();
    const info = this.parse(url);
    if (info) {
        location.hash = hash; // this will automatically pushState and set the state to null and then trigger hashchange event
        this.save_scroll_state(); // by calling this method, it will also replaceState with object instead of null, we could also use history.replaceState({}, '', url.href);
    }
    return;
}
```

packages/kit/src/runtime/client/router.js
PH4NTOMiki: if an x and y coordinates are 0, that's the default position so we don't need to store it.

PH4NTOMiki:
1@Rich-Harris also scrolling test are wrong, at least about hash handling, they go success but they should fail, also I'll implement the fix for scrolling now(option B) and remove the listener tomorrow, because it's more complicated than I originally though, and It's 10pm where I live (London time + 1 :) ) so I'm tired now. I'll implement it tomorrow.

packages/kit/src/runtime/client/router.js
```diff
@@ -152,7 +152,7 @@ export class Router {
 		addEventListener('sveltekit:trigger_prefetch', trigger_prefetch);
 
 		/** @param {MouseEvent} event */
-		addEventListener('click', (event) => {
+		addEventListener('click', async (event) => {
```
benmccann, line 155: why did you have to make this `async`?

benmccann:
Can you update the PR title and description to reflect what this change does? It's hard to understand the context for it right now since I don't think it removes the need for the scroll event listener and the description is unclear as to whether it fixes the duplicate history entry

packages/kit/src/runtime/client/router.js
```diff
@@ -190,12 +190,12 @@ export class Router {
...
 			const [base, hash] = url.href.split('#');
 			if (hash !== undefined && base === location.href.split('#')[0]) {
-				// Call `pushState` to add url to history so going back works.
-				// Also make a delay, otherwise the browser default behaviour would not kick in
-				setTimeout(() => history.pushState({}, '', url.href));
+				event.preventDefault();
 				const info = this.parse(url);
 				if (info) {
-					return this.renderer.update(info, [], false);
+					await this.renderer.update(info, [], false);
```
benmccann, line 196: do we actually need to call `renderer.update`? I think we just need to update the hash in the page store

benmccann:
I sent a PR with a failing test that's commented out: https://github.com/sveltejs/kit/pull/3880. After it's merged you can rebase this PR and uncomment it

packages/kit/src/runtime/client/router.js
```diff
@@ -190,12 +190,12 @@ export class Router {
 			// Removing the hash does a full page navigation in the browser, so make sure a hash is present
 			const [base, hash] = url.href.split('#');
 			if (hash !== undefined && base === location.href.split('#')[0]) {
-				// Call `pushState` to add url to history so going back works.
-				// Also make a delay, otherwise the browser default behaviour would not kick in
-				setTimeout(() => history.pushState({}, '', url.href));
+				event.preventDefault();
```
jasonlyu123, line 193: This would disable the focus management done by the browser. Making https://github.com/sveltejs/kit/issues/3105 a problem again.

jasonlyu123: Oh never mind. It's handled by hash setter and I tested with the wrong commit. But I notice the test for this is wrong. It actually won't fail.

PH4NTOMiki: Yeah, I'll deal with it. thanks

packages/kit/src/runtime/client/router.js
PH4NTOMiki: and if we decide to let renderer.update stay there, I think we should use `.then()` instead of `await` because it's the only need for `async` function. 
But as you said, I don't think it's actually needed, because we are re-rendering the same page, and that's not how browser normally behaves(when in not SPA environment) so we're not following the spec if we let it remain.

packages/kit/src/runtime/client/router.js
jasonlyu123: I mean this one https://github.com/sveltejs/kit/blob/2fefa5e9cf01510dd03c1a9cc0b62c9686621faa/packages/kit/test/apps/basics/test/test.js#L1732

There should be another focusable element between `<p id="p1">` and `<p id="p2">` here
https://github.com/sveltejs/kit/blob/4453c67a9526c3752d7eee7e53b7b1390c631e09/packages/kit/test/apps/basics/src/routes/routing/hashes/target.svelte#L7

I think it's somewhat unrelated to this PR. So you don't have to deal with it. I can create another PR for this later.

PH4NTOMiki:
It makes no sense that these test are failling, because I'm calling `store.set` maybe there is some racing.

packages/kit/src/runtime/client/router.js
PH4NTOMiki: @benmccann @Rich-Harris we can't do that because the stores don't react to that, we would have to trigger every store change manually, like I tried in https://github.com/sveltejs/kit/pull/3873/commits/6340fea3a8861de2ffc87ae168665456270d0ad0

PH4NTOMiki:
I'll now check why it failed, I have an idea

PH4NTOMiki:
@benmccann I think the test is wrong because you are assigning a number to const, and then comparing it twice to different values.

PH4NTOMiki:
@benmccann the tests are again being flaky after that latest in_view updates, but I just tested it locally and it passes.

PH4NTOMiki:
@Rich-Harris @benmccann I tried without that setTimeout hack and setting the stores manually, but it gets either too complicated, or unreliable, so this the best option that definitely works, and passes the tests.
So I think you should merge it, and if there is a better option we'll do another PR.

packages/kit/src/runtime/client/router.js
benmccann: I would stick with `await` / `async` as I think it's easier to understand and more consistent with the rest of the code

packages/kit/src/runtime/client/router.js
```diff
@@ -72,6 +73,30 @@ export class Router {
...
 
+	save_scroll_position() {
+		sessionStorage.setItem(
+			`sveltekit:scroll:${this.current_history_index}`,
+			JSON.stringify(scroll_state())
+		);
+	}
+
+	/**
+	 *
```
benmccann, line 84: no need for a blank line here
```suggestion
```

packages/kit/test/utils.js
```diff
@@ -111,9 +111,33 @@ export const test = base.extend({
 	in_view: async ({ page }, use) => {
 		/** @param {string} selector */
 		async function in_view(selector) {
-			const box = await page.locator(selector).boundingBox();
-			const view = await page.viewportSize();
-			return box && view && box.y < view.height && box.y + box.height > 0;
+			return await page.$eval(selector, async (/** @type {Element} */ element) => {
```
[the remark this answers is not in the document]
PH4NTOMiki: Yeah, I was just tetsing it obviously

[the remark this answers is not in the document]
PH4NTOMiki: I was angry the test were failing even thought I know it works so I added that for testing.

PH4NTOMiki:
> I'm having an extremely difficult time keeping up with the state of this PR. 46 commits is entirely too many. Can you mark this as draft until it's in its final state? Is it ready for review now or is there more that you're changing still?

It's done now, those were mostly test retries.

PH4NTOMiki:
> > I'm having an extremely difficult time keeping up with the state of this PR. 46 commits is entirely too many. Can you mark this as draft until it's in its final state? Is it ready for review now or is there more that you're changing still?
> 
> It's done now, those were mostly test retries.

Or I can close this PR and open a new with only the changes, it would be much clearer.

benmccann:
Ok. It's not a blocker to make all the tests pass. We know they're flaky and still commit PRs even if one test run fails. We should get them passing regularly, but I'd prefer to do one thing per PR and address the flakiness separate from fixing bugs

packages/kit/test/utils.js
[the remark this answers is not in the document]
benmccann: can we revert this change in this PR? i'm not sure whether it's needed at all as I haven't been keeping up with the discussion on the other PR, but if so it'd be better to send a new PR for it

packages/kit/src/runtime/client/router.js
```diff
@@ -72,6 +73,29 @@ export class Router {
...
+			`sveltekit:scroll:${this.current_history_index}`,
+			JSON.stringify(scroll_state())
+		);
+	}
+
+	/**
+	 * @param {number} index
+	 * @returns {{x: number, y: number} | null}
+	 */
+	get_scroll_position(index) {
```
benmccann, line 87: can we either make this method and `save_scroll_position` a standalone function next to `scroll_state` or prefix with `#` to make them private? I don't think they need to be exposed outside of this class

benmccann:
I did just notice one issue - though it's also present in `master`, so isn't a blocker. If you visit http://localhost:3000/docs/layouts#nested-layouts then press Ctrl+W followed by Ctrl+Shift+T it will reopen at the top of the page rather than the anchor

packages/kit/src/runtime/client/router.js
PH4NTOMiki: done

PH4NTOMiki:
> I did just notice one issue - though it's also present in `master`, so isn't a blocker. If you visit http://localhost:3000/docs/layouts#nested-layouts then press Ctrl+W followed by Ctrl+Shift+T it will reopen at the top of the page rather than the anchor

fixed that and ready to merge

Rich-Harris:
We can't merge this until there's a test that fails without this PR — are you able to add one?

packages/kit/src/runtime/client/router.js
```diff
@@ -55,12 +55,26 @@ export class Router {
...
+		this.current_history_index = Number(
+			sessionStorage.getItem('sveltekit:index') ?? history.state?.['sveltekit:index'] ?? 0
+		);
 
-		if (this.current_history_index === 0) {
+		if (!history.state || history.state['sveltekit:index'] == undefined) {
 			// create initial history entry, so we can return here
-			history.replaceState({ ...history.state, 'sveltekit:index': 0 }, '', location.href);
+			history.replaceState(
+				{ ...(history.state || {}), 'sveltekit:index': this.current_history_index },
```
Rich-Harris, line 69: it's fine to spread `undefined`

```suggestion
				{ ...history.state, 'sveltekit:index': this.current_history_index },
```

packages/kit/src/runtime/client/router.js
```diff
@@ -55,12 +55,26 @@ export class Router {
...
+
 		// keeping track of the history index in order to prevent popstate navigation events if needed
-		this.current_history_index = history.state?.['sveltekit:index'] ?? 0;
+		/** @type {number} */
+		this.current_history_index = Number(
+			sessionStorage.getItem('sveltekit:index') ?? history.state?.['sveltekit:index'] ?? 0
+		);
 
-		if (this.current_history_index === 0) {
+		if (!history.state || history.state['sveltekit:index'] == undefined) {
```
Rich-Harris, line 66:
```suggestion
		if (history.state?.['sveltekit:index'] === undefined) {
```

packages/kit/src/runtime/client/router.js
```diff
@@ -190,9 +205,17 @@ export class Router {
...
-				// Call `pushState` to add url to history so going back works.
-				// Also make a delay, otherwise the browser default behaviour would not kick in
-				setTimeout(() => history.pushState({}, '', url.href));
+				this.#save_scroll_position();
+
+				// Call `replaceState` in timeout because this event will add pushState automatically and we need to store our own history state
+				// Timeout is being used to delay replaceState till next tick
+				setTimeout(() => {
+					history.replaceState(
+						{ ...(history.state || {}), 'sveltekit:index': ++this.current_history_index },
```
Rich-Harris, line 214:
```suggestion
						{ ...history.state, 'sveltekit:index': ++this.current_history_index },
```

packages/kit/src/runtime/client/router.js
```diff
@@ -55,12 +55,26 @@ export class Router {
 		// make it possible to reset focus
 		document.body.setAttribute('tabindex', '-1');
 
+		history.scrollRestoration = 'manual';
```
benmccann, line 58: we're doing this twice now. here and line 139. do we need both?

packages/kit/src/runtime/client/router.js
```diff
@@ -190,9 +205,17 @@ export class Router {
...
-				// Call `pushState` to add url to history so going back works.
-				// Also make a delay, otherwise the browser default behaviour would not kick in
-				setTimeout(() => history.pushState({}, '', url.href));
+				this.#save_scroll_position();
+
+				// Call `replaceState` in timeout because this event will add pushState automatically and we need to store our own history state
+				// Timeout is being used to delay replaceState till next tick
+				setTimeout(() => {
+					history.replaceState(
+						{ ...history.state, 'sveltekit:index': ++this.current_history_index },
```
benmccann, line 214: why are we storing `sveltekit:index` in the history state here and in the session elsewhere?

packages/kit/src/runtime/client/router.js
```diff
@@ -55,12 +55,26 @@ export class Router {
...
-		if (this.current_history_index === 0) {
+		if (history.state?.['sveltekit:index'] === undefined) {
 			// create initial history entry, so we can return here
-			history.replaceState({ ...history.state, 'sveltekit:index': 0 }, '', location.href);
+			history.replaceState(
+				{ ...history.state, 'sveltekit:index': this.current_history_index },
+				'',
+				location.href
+			);
+		} else if (sessionStorage.getItem('sveltekit:left_page') === 'true') {
```
benmccann, line 73: maybe we should call this something like `sveltekit:exited` since left can have multiple meanings. e.g. left as in LTR or left as in exit

Rich-Harris:
When I enable the test included in this PR on `master`, the app behaves as expected. The test fails because playwright reports the URL as incorrect — probably a race condition — but other than that things are fine.

By contrast, when I check out this PR the behaviour is totally broken.

I'm sorry, but I really don't understand what this PR is doing!

PH4NTOMiki:
> When I enable the test included in this PR on `master`, the app behaves as expected. The test fails because playwright reports the URL as incorrect — probably a race condition — but other than that things are fine.
> 
> By contrast, when I check out this PR the behaviour is totally broken.
> 
> I'm sorry, but I really don't understand what this PR is doing!

No worries, but how do you mean, the behaviour is totally broken
https://kit-svelte-dev-z8e747sj9-svelte.vercel.app/ latest commit, looks good to me, even the Ctrl+W and Ctrl+Shift+T (reopen closed tab issue @benmccann mentioned that was present isn't there anymore)

Rich-Harris:
Ah. Turns out `master` behaves as expected on Firefox but not Chrome. This PR behaves as expected on Chrome but not Firefox...

PH4NTOMiki:
@Rich-Harris can you tell what's wrong on Firefox and I'm sure I can fix it

Rich-Harris:
https://user-images.githubusercontent.com/1162160/154121599-5cd3171d-cfc4-4929-bb62-f4630c747fd4.mov

PH4NTOMiki:
@Rich-Harris okay, I understand the issue, I just found that hashchange acts differently on FF, and that's the reason, but I think I know how to fix it.

packages/kit/src/runtime/client/router.js
benmccann: why not just store the scroll state in the history directly?

PH4NTOMiki:
@Rich-Harris fixed, please test now, it should work in Chrome/Edge and FF

Rich-Harris:
Thanks. It turns out I can get the test to pass with just the most recent change, so I've opened a separate PR with just that: https://github.com/sveltejs/kit/pull/3931

Is there a separate issue that this PR is also solving? Given how volatile scroll handling has turned out to be, I'd favour a cautious (and test-driven) approach to making changes

PH4NTOMiki:
> Thanks. It turns out I can get the test to pass with just the most recent change, so I've opened a separate PR with just that: #3931
> 
> Is there a separate issue that this PR is also solving? Given how volatile scroll handling has turned out to be, I'd favour a cautious (and test-driven) approach to making changes

Yeah, to be honest I thought of that as well, but this solution is much safer, than relaying on the scroll event, both Remix(https://github.com/remix-run/remix/blob/1fd70960e4d88740df5bf407a6ba2cd2b9549459/packages/remix-react/scroll-restoration.tsx) and Next.js(https://github.com/vercel/next.js/blob/81b537c04e694930a2fb9e6870ba9c5805269683/packages/next/shared/lib/router/router.ts) are using a solution similar to this. I based this solution on theirs. because scroll event is pretty bad to rely on, and on low powered devices(and Svelte is great for those) scroll event can cause throttling(lagging) of scrolling. I've seen it happen many many times. So consider this, but that PR is okay too, if that's what you want.

PH4NTOMiki:
In this PR we are fully using our own scroll handling, so in the future if we decide to add some scroll behavior that's more complicated, we can do it with PR, but not with that with scroll event listener.

PH4NTOMiki:
and there is also that Ctrl+shift+t bug that's not present in this PR, I know it's not that big of deal, but still, one of many upsides of this PR comparing to that one.

Rich-Harris:
For the sake of the git history I think what I'll do is merge #3931 first, so that this PR becomes purely about eliminating the scroll handler and the ctrl+shift+t bug (which we'd ideally have a test for — not yet sure if that's possible with playwright)

benmccann:
Can we add a comment explaining why the `hash_navigating` variable is needed and listening to `hashchange` is not sufficient?

Rich-Harris:
> Can we add a comment explaining why the `hash_navigating` variable is needed and listening to `hashchange` is not sufficient?

https://github.com/sveltejs/kit/pull/3937

Rich-Harris:
#3936 addresses the Cmd-Shift-T bug separately from everything else

Rich-Harris:
Ignore #3936, I've opened #3938 which supersedes it.

I think that PR covers everything here, but more simply — as far as I can see we don't need to track `sveltekit:index` in both `history.state` _and_ `sessionStorage`, we don't need to track `sveltekit:exited`, and we only need to update `sessionStorage` in `beforeunload`, unless I'm missing something?

PR #3876, opened 2022-02-12
[Fix]: Adding compatibility for the search shortcut to other keyboard layouts

Swarleys:
If you're using a keyboard layout different to QWERTY  like Dvorak or Colemak the shortcut of CMD+K will not work since the k is in another position, but if we change from event.code to event.key it will work. The other possibility
 is with event.keyCode that returns on all the layouts 75 but is not so clear what we are checking then.

- [ x] This message body should clearly illustrate what problems it solves.

### Tests

### Changesets

### DIFF
--- a/sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
+++ b/sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
@@ -106,7 +106,7 @@
 
 <svelte:window
 	on:keydown={(e) => {
-		if (e.code === 'KeyK' && (navigator.platform === 'MacIntel' ? e.metaKey : e.ctrlKey)) {
+		if (e.key === 'k' && (navigator.platform === 'MacIntel' ? e.metaKey : e.ctrlKey)) {
 			e.preventDefault();
 			$query = '';
 			$searching = !$searching;


sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
```diff
@@ -106,7 +106,7 @@
 
 <svelte:window
 	on:keydown={(e) => {
-		if (e.code === 'KeyK' && (navigator.platform === 'MacIntel' ? e.metaKey : e.ctrlKey)) {
+		if (e.key === 'k' && (navigator.platform === 'MacIntel' ? e.metaKey : e.ctrlKey)) {
```
Rich-Harris, line 109: edge case, but do we need to consider the situation where caps lock is on?

```suggestion
		if ((e.key === 'k' || e.key === 'K') && (navigator.platform === 'MacIntel' ? e.metaKey : e.ctrlKey)) {
```

Swarleys: There is no need, I just try it and keeps working.

Rich-Harris: oh, fun — i tried on https://keycode.info/ and it shows up as `K` with caps lock (hence my comment) but if you're also holding the cmd key down it's `k`. Keyboard events are endlessly weird!

Rich-Harris (approved):

Rich-Harris:
thanks!

PR #3877, opened 2022-02-12
ToC updates twice on mount

janosh:
```js
afterNavigate(() => {
  update();
  highlight();
});
```

already calls `update();` and `highlight();` on mount. No need to do it again in `onMount`.

Not sure I get it. Does the rest of the `onMount` function accomplish anything?

```js
onMount(() => {
  // wait for fonts to load...
  const timeouts = [setTimeout(onresize, 1000), setTimeout(onscroll, 5000)];
  return () => {
    window.removeEventListener('scroll', onscroll, true);
    window.removeEventListener('resize', onresize, true);
    timeouts.forEach((timeout) => clearTimeout(timeout));
  };
});
```

### DIFF
--- a/sites/kit.svelte.dev/src/lib/docs/Contents.svelte
+++ b/sites/kit.svelte.dev/src/lib/docs/Contents.svelte
@@ -20,9 +20,6 @@
 		// wait for fonts to load...
 		const timeouts = [setTimeout(onresize, 1000), setTimeout(onscroll, 5000)];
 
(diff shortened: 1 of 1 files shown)


Rich-Harris (approved):

Rich-Harris:
good catch, thanks — the rest of the `onMount` does serve a purpose, but is currently referencing non-existent functions. fixed in #3878

PR #3878, opened 2022-02-12
Fix contents onmount

Rich-Harris:
follow-up to #3877

### Tests

### Changesets

### DIFF
--- a/sites/kit.svelte.dev/src/lib/docs/Contents.svelte
+++ b/sites/kit.svelte.dev/src/lib/docs/Contents.svelte
@@ -16,16 +16,11 @@
 	/** @type {number[]} */
 	let positions = [];
 
-	onMount(() => {
-		// wait for fonts to load...
-		const timeout = setTimeout(() => {
-			update();
-			highlight();
-		}, 1000);
+	onMount(async () => {
+		await document.fonts.ready;
 
-		return () => {
-			clearTimeout(timeout);
-		};
+		update();
+		highlight();
 	});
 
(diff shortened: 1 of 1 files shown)


sites/kit.svelte.dev/src/lib/docs/Contents.svelte
```diff
@@ -18,16 +18,13 @@
...
 	onMount(() => {
 		// wait for fonts to load...
-		const timeouts = [setTimeout(onresize, 1000), setTimeout(onscroll, 5000)];
-
-		update();
-		highlight();
+		const timeout = setTimeout(() => {
+			update();
+			highlight();
+		}, 1000);
```
benmccann, line 24: does this mean there won't be any contents until 1 second after load? I don't really understand why we need these function calls both here an in `afterNavigate`

benmccann (commented):
looks like this PR needs a rebase

sites/kit.svelte.dev/src/lib/docs/Contents.svelte
Rich-Harris: the comment above explains. it's there so that positions are correctly calculated if fonts are slow to load

Rich-Harris: updated to use `document.fonts.ready`. Preloading is probably a step too far — would be finicky to get the URL to preload (unless Vite provided it to us), and means that other resources are deprioritised

benmccann (approved):

PR #3879, opened 2022-02-12
Ensure handleError hook is called for shadow endpoint errors

### LINKED ISSUE #3807: Shadow Endpoint errors don't hit handleError in SSR
### Describe the bug

When an error is throw in an endpoint `get()` method, it's not passed to `handleError` if called as a shadow endpoint via SSR. 

Calling the endpoint manually with `fetch()`, or fetching the shadow endpoint by means of a client-side navigation all trigger `handleError` as expected.

I believe this is the problem I'm trying to wrap my head around in [this discussion](https://github.com/sveltejs/kit/discussions/3768).

### Reproduction

I've made a simple reproduction here: https://github.com/georgecrawford/sveltekit-shadow-endpoint-handle-error (also available live in StackBlitz here: https://stackblitz.com/edit/node-wy166v?file=src%2Fhooks.js&terminal=dev).

There are two endpoints, both of which throw an `Error` inside `get()`. When the endpoint is called from a `load()` function, both SSR and browser routing will cause the error to be handled by `handleError`. However, when an endpoint is called as a shadow endpoint, the SSR version doesn't hit `handleError` (whereas the browser-routed shadow load does as expected).

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 11.6.2
    Memory: 202.62 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 16.4.0 - /usr/local/bin/node
    npm: 7.19.0 - /usr/local/bin/npm
    Watchman: 2021.09.06.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 98.0.4758.80
    Chrome Canary: 100.0.4878.0
    Firefox: 92.0.1
    Safari: 15.2
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.262
    svelte: ^3.44.0 => 3.46.4
```

### Severity

blocking an upgrade

### Additional Information

_No response_

georgecrawford:
Fixes #3807 

I'm honestly not sure this is the right place, or the right treatment for errors thrown in shadow endpoints, but it seems wrong that they weren't being passed to the `handleError` hook. I added a failing test, and tracked down a possible fix location. The test still fails as the assertion for `page.textContent('#message')` is getting the error message, but expecting an empty string. I don't know what's desired here (I just copied the `'error in endpoint'` test above), so the failure might indicate the fix is more complex than I thought. Anyway, it's a start!

### Tests

### Changesets

### DIFF
--- a/.changeset/selfish-walls-grow.md
+++ b/.changeset/selfish-walls-grow.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Ensure handleError hook is called for shadow endpoint errors
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -66,6 +66,7 @@ export async function load_node({
 		? await load_shadow_data(
 				/** @type {import('types/internal').SSRPage} */ (route),
 				event,
+				options,
 				!!state.prerender
 		  )
 		: {};
@@ -360,10 +361,11 @@ export async function load_node({
  *
  * @param {import('types/internal').SSRPage} route
  * @param {import('types/hooks').RequestEvent} event
+ * @param {import('types/internal').SSROptions} options
  * @param {boolean} prerender
  * @returns {Promise<import('types/endpoint').ShadowData>}
  */
-async function load_shadow_data(route, event, prerender) {
+async function load_shadow_data(route, event, options, prerender) {
 	if (!route.shadow) return {};
 
 	try {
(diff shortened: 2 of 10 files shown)


Rich-Harris:
Thanks. `handleError` is only called for _unexpected_ errors — returning `{ status: 500 }` shouldn't result in `handleError` running. This distinction wasn't covered by the existing tests, so I added some and added the `handle_error` call where it needs to go: 8411533b142f451e4d89e9ded051fe8f7649e589

Rich-Harris (approved):

georgecrawford:
Thanks @Rich-Harris. I was pretty sure I was close but not quite right on this one!

PR #3885, opened 2022-02-13
fix in_view function

bluwy:
@PH4NTOMiki's https://github.com/sveltejs/kit/pull/3873#issuecomment-1037486329 reminded me of https://github.com/sveltejs/kit/pull/3741#discussion_r802313189. The `in_view` function now isn't correct and reports `true` if the element is anywhere on the page, not stricly in the viewport as expected.

With this PR, `in_view` only returns true if the element is actually in the viewport.

## Notes

According to [playwright's docs](https://playwright.dev/docs/api/class-locator#locator-bounding-box):

> This method returns the bounding box of the element, or null if the element is not visible

Playwright's term for "is not visible" is [confusing](https://playwright.dev/docs/actionability#visible):

> Element is considered visible when it has non-empty bounding box and does not have visibility:hidden computed style. Note that elements of zero size or with display:none are not considered visible.

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/utils.js
+++ b/packages/kit/test/utils.js
@@ -112,7 +112,8 @@ export const test = base.extend({
 		/** @param {string} selector */
 		async function in_view(selector) {
 			const box = await page.locator(selector).boundingBox();
-			return !!box;
+			const view = await page.viewportSize();
+			return box && view && box.y < view.height && box.y + box.height > 0;
 		}
 
 		use(in_view);


packages/kit/test/utils.js
```diff
@@ -112,7 +112,8 @@ export const test = base.extend({
 		/** @param {string} selector */
 		async function in_view(selector) {
 			const box = await page.locator(selector).boundingBox();
-			return !!box;
+			const view = await page.viewportSize();
+			return box && view && box.y < view.height && box.y + box.height > 0;
```
benmccann, line 116: Is this part right `box.y + box.height > 0`? Or should it be `-` instead of `+`?

bluwy: Looks right to me. The `box.y` is the top value of the box, plus the height and comparing it with 0 ensures that at least a part of the bounding box is inside the viewport. For example, cases where `box.y` is negative (slightly outside of the viewport), but its height is still visible in the viewport.

benmccann: oh, I see. yeah, makes sense

PH4NTOMiki:
@bluwy @benmccann the tests using in_view are being flaky again.

bluwy:
Yeah, I don't think the tests were de-flaked in the first place.

PH4NTOMiki:
I'm trying this now https://stackoverflow.com/a/68848306

PH4NTOMiki:
@bluwy what do you think of that, it seems to pass consistently.

PH4NTOMiki:
> @bluwy what do you think of that, it seems to pass consistently.

Everywhere except Windows

PH4NTOMiki:
@bluwy @benmccann This works consistently even on Windows, it's the old deprecated way, but it works:
```javascript
			return await page.$eval(selector, async (/** @type {Element} */ element) => {
				// @ts-ignore
				let top = element.offsetTop;
				// @ts-ignore
				let left = element.offsetLeft;
				// @ts-ignore
				const width = element.offsetWidth;
				// @ts-ignore
				const height = element.offsetHeight;

				// @ts-ignore
				while (element.offsetParent) {
					// @ts-ignore
					element = element.offsetParent;
					// @ts-ignore
					top += element.offsetTop;
					// @ts-ignore
					left += element.offsetLeft;
				}

				return top < window.pageYOffset + window.innerHeight &&
					left < window.pageXOffset + window.innerWidth &&
					top + height > window.pageYOffset &&
					left + width > window.pageXOffset
					? true
					: { top, left, width, height };
			});
		}

		use(in_view);
	}
```

Maybe you or I can do a PR for it and merge it.

bluwy:
Thanks for digging into this @PH4NTOMiki. My first encounter with the scrolling shenanigans was also through that stackoverflow post, and the function had evolved over time. I'm not sure if using `ResizeObservers` would help though if it would still sometimes fail in windows.

Re the _old deprecated way_, do you know why that works more reliably? I'm not 100% sure if the function is safe too.

Also maybe related, perhaps we can setup the CI to emit the screenshots for failed tests. That might show us if the element is actually inside the view, or it's an error with `in_view` incorrectly detecting stuff.

benmccann:
It's already outputting screenshots and failure traces on the CI: https://github.com/sveltejs/kit/blob/0c6bc25f6f2ef3bad181ac116ed90143e4038ab6/packages/kit/test/utils.js#L194

I'm not sure how to retrieve them though. I guess we'd have to setup something to upload them somewhere?

benmccann:
There were a couple tests I couldn't get to fail on my machine. A user in Discord provided me with these traces awhile back to help us debug. Unfortunately, I haven't gotten to investigate yet. But maybe someone else will be able to :grin: 

[playwright-test-results.zip](https://github.com/sveltejs/kit/files/8062404/playwright-test-results.zip)

bluwy:
> I'm not sure how to retrieve them though. I guess we'd have to setup something to upload them somewhere?

I've done something like this before using `upload-artifact`. Will send a PR.

PH4NTOMiki:
@bluwy I think it works better because it doesn't use boundingClientRect(boundingBox in playwright is similar). It uses element static properties.

PR #3895, opened 2022-02-14
add: new gitignore files & organize

thesyedbasim:
Add more files to gitignore templates, and organize them.

### Tests
- [ x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets

### DIFF
--- a/packages/create-svelte/templates/default/.gitignore
+++ b/packages/create-svelte/templates/default/.gitignore
@@ -1,10 +1,29 @@
-.DS_Store
-node_modules
+# dependencies
+/node_modules
+
+# testing
+/coverage
+
+# production
 /build
-/.svelte-kit
-/package
+
+# misc
+.DS_Store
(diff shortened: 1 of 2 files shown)


Conduitry:
I'm -1 on adding all of these. If someone is using tool X they should be responsible for adding the relevant files for that tool. If we add a bunch of things, people will then request that we add more entries to support a tool that they say is very popular.

I'm against .DS_Store as well, actually - that should be in your global system ignore config - but I know that in reality most people aren't going to be careful about that.

Rich-Harris:
Thanks, but I agree with @Conduitry, it's not our place to add this stuff. I'd be on board with removing `.DS_Store`

thesyedbasim:
Thank you for the feedback.

I also agree with your reason @Conduitry 

Thank you.

PR #3900, opened 2022-02-14
use old-style padding

Rich-Harris:
kit.svelte.dev looks terrible on some iphones because Safari only recently began supporting logical properties, and it's not an evergreen browser. ~~I have a hunch that `padding-inline` and `padding-block` are to blame — this is a small PR to test out that theory. If it holds, we should probably remove all those declarations until such time as we need to support languages that need them.~~

### Tests

### Changesets

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -311,7 +311,7 @@ importers:
     specifiers:
       '@sveltejs/adapter-auto': workspace:*
       '@sveltejs/kit': workspace:*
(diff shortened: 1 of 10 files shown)


Rich-Harris:
Hunch confirmed. Annoyingly, this means reverting https://github.com/sveltejs/sites/pull/238. What a bloody awful piece of software.

**Safari: Creating Unnecessary Work For Developers Since 2003™**

Issue #3894, opened 2022-02-14
HEAD method not allowed


livehtml:
### Describe the bug

I've just run my SvelteKit app in production (adapter node) and periodically gets an error in logs:
405 HEAD Method not allowed

Other methods I can disable with nginx via limit_except, but what about HEAD? Is this method should be supported? How can I enable and correctly handle it?

### Reproduction

Build and run with adapter node

### System Info

```shell
System:
    OS: Linux 5.4 Ubuntu 20.04.3 LTS (Focal Fossa)
    CPU: (8) x64 DO-Premium-Intel
    Memory: 12.50 GB / 15.63 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 16.14.0 - /usr/bin/node
    npm: 8.3.1 - /usr/bin/npm
  npmPackages:
    @sveltejs/adapter-node: ^1.0.0-next.68 => 1.0.0-next.68 
    @sveltejs/kit: next => 1.0.0-next.267 
    svelte: ^3.46.4 => 3.46.4
```

### Severity

annoyance

antony:
As far as I'm aware, we don't prevent the `HEAD` method, and this may be a misconfiguration of your server.

Try `npm run preview` locally and see if you can reproduce the issue.

livehtml:
Hmm, may be, but how can fix it?

When I run ```node build``` and then ```curl -i --head http://127.0.0.1:3000```

I've got:
```
HTTP/1.1 405 Method Not Allowed
content-type: text/html
etag: "qkw4jc"
Vary: Accept-Encoding
Date: Mon, 14 Feb 2022 12:53:45 GMT
Connection: keep-alive
Keep-Alive: timeout=5
```

livehtml:
I've just tested on SvelteKit Demo app and found that if we request http://127.0.0.1:3000 — HEAD is allowed (200 OK), but if http://127.0.0.1:3000/todos — got the 405 Method Not Allowed

geoffrich:
Looks like we return a 405 for shadow endpoints if there isn't a handler:

https://github.com/sveltejs/kit/blob/0c6bc25f6f2ef3bad181ac116ed90143e4038ab6/packages/kit/src/runtime/server/page/load_node.js#L386-L394

dominikg:
[...]

PR #3903, opened 2022-02-14
[fix] default handlers for head

dominikg:
see #3894

add default handling for HEAD requests. follow "get" handling but do not return body in the end.
custom `head` handlers in endpoint files are respected and take precedence over get.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/endpoint.js
+++ b/packages/kit/src/runtime/server/endpoint.js
@@ -96,7 +96,7 @@ export async function render_endpoint(event, mod) {
 		}
 	}
 
-	return new Response(method !==  'head' ? normalized_body : '' , {
+	return new Response(method !== 'head' ? normalized_body : '', {
 		status,
 		headers
 	});
--- a/packages/kit/src/runtime/server/endpoint.js
+++ b/packages/kit/src/runtime/server/endpoint.js
@@ -1,6 +1,6 @@
 import { to_headers } from '../../utils/http.js';
 import { hash } from '../hash.js';
-import { is_pojo, normalized_request_method } from './utils.js';
+import { is_pojo, normalize_request_method } from './utils.js';
 
 /** @param {string} body */
 function error(body) {
@@ -40,7 +40,7 @@ export function is_text(content_type) {
  * @returns {Promise<Response | undefined>}
  */
 export async function render_endpoint(event, mod) {
-	const method = normalized_request_method(event);
+	const method = normalize_request_method(event);
 
 	/** @type {import('types/endpoint').RequestHandler} */
 	let handler = mod[method];
--- a/packages/kit/src/runtime/server/endpoint.js
+++ b/packages/kit/src/runtime/server/endpoint.js
@@ -97,7 +97,7 @@ export async function render_endpoint(event, mod) {
 		}
 	}
 
-	return new Response(method !== 'head' ? normalized_body : '', {
+	return new Response(method !== 'head' ? normalized_body : undefined, {
 		status,
 		headers
 	});
--- a/packages/kit/src/runtime/server/endpoint.js
+++ b/packages/kit/src/runtime/server/endpoint.js
@@ -40,9 +40,13 @@ export function is_text(content_type) {
  * @returns {Promise<Response | undefined>}
  */
 export async function render_endpoint(event, mod) {
+	const method = event.request.method.toLowerCase().replace('delete', 'del'); // 'delete' is a reserved word
 	/** @type {import('types/endpoint').RequestHandler} */
-	const handler = mod[event.request.method.toLowerCase().replace('delete', 'del')]; // 'delete' is a reserved word
+	let handler = mod[method];
 
+	if (!handler && method === 'head') {
+		handler = mod.get;
+	}
 	if (!handler) {
 		return;
 	}
@@ -92,7 +96,7 @@ export async function render_endpoint(event, mod) {
 		}
 	}
 
-	return new Response(normalized_body, {
+	return new Response(method !==  'head' ? normalized_body : '' , {
 		status,
 		headers
 	});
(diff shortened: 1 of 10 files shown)


packages/kit/src/runtime/server/endpoint.js
```diff
@@ -40,9 +40,13 @@ export function is_text(content_type) {
  * @returns {Promise<Response | undefined>}
  */
 export async function render_endpoint(event, mod) {
+	const method = event.request.method.toLowerCase().replace('delete', 'del'); // 'delete' is a reserved word
```
Conduitry, line 43: This code was already present, but I feel like `.replace()` isn't the best way to deal with this, for a few reasons. What if we just do something like:

```suggestion
	let method = event.request.method.toLowerCase();
	if (method === 'delete') { // 'delete' is a reserved word
		method = 'del';
	}
```

dominikg: moved it to a util as it was used in two places

packages/kit/src/runtime/server/endpoint.js
```diff
@@ -92,7 +96,7 @@ export async function render_endpoint(event, mod) {
 		}
 	}
 
-	return new Response(normalized_body, {
+	return new Response(method !== 'head' ? normalized_body : '', {
```
Rich-Harris, line 100: strictly speaking i think this should be `undefined`, though i'm not sure there's an observable distinction

```suggestion
	return new Response(method !== 'head' ? normalized_body : undefined, {
```

dominikg: done, had a test fail locally with a redirect loop on first run but that seems to have been a fluke, second run passed

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -424,8 +424,9 @@ async function load_shadow_data(route, event, prerender) {
 			data.body = body;
 		}
 
-		if (mod.get) {
-			const result = await mod.get.call(null, event);
+		const resultHandler = (method === 'head' && mod.head) || mod.get;
+		if (resultHandler) {
+			const result = await resultHandler.call(null, event);
```
Rich-Harris, line 429: `resultHandler` feels a bit clunky (it suggests the other `handler` is some _other_ kind of handler). I suggest we just call it `get`. Also, the `.call(null, ` is only necessary with dynamic dispatch

```suggestion
		const get = (method === 'head' && mod.head) || mod.get;
		if (get) {
			const result = await get(event);
```

Rich-Harris:
I just pushed a cosmetic change — the camelCased variables conflicted with the code style, and it was a bit of a wall-of-text, but it was tricky to do in a commit suggestion

packages/kit/test/apps/basics/test/test.js
```diff
@@ -459,6 +491,23 @@ test.describe.parallel('Endpoints', () => {
 		expect(response.headers()['set-cookie']).toBeDefined();
 	});
 
+	test('HEAD with matching headers but without body', async ({ request }) => {
+		const url = '/endpoint-output/body';
+		const headResponse = await request.head(url);
```
Rich-Harris, line 496: see above re camelCase — can we make this `responses.head` and `responses.get` etc? it's more consistent with the code style elsewhere and the extra newlines make it a bit easier to read IMHO

packages/kit/src/runtime/server/endpoint.js
```diff
@@ -1,6 +1,6 @@
 import { to_headers } from '../../utils/http.js';
 import { hash } from '../hash.js';
-import { is_pojo } from './utils.js';
+import { is_pojo, normalized_request_method } from './utils.js';
```
Rich-Harris, line 3: Most of our function names are verby — they describe what they do rather than what they return. I think this should be the same

```suggestion
import { is_pojo, normalize_request_method } from './utils.js';
```

packages/kit/src/runtime/server/endpoint.js
```diff
@@ -40,9 +40,14 @@ export function is_text(content_type) {
  * @returns {Promise<Response | undefined>}
  */
 export async function render_endpoint(event, mod) {
+	const method = normalized_request_method(event);
```
Rich-Harris, line 43:
```suggestion
	const method = normalize_request_method(event);
```

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -4,7 +4,7 @@ import { s } from '../../../utils/misc.js';
 import { escape_json_in_html } from '../../../utils/escape.js';
 import { is_root_relative, resolve } from '../../../utils/url.js';
 import { create_prerendering_url_proxy } from './utils.js';
-import { is_pojo, lowercase_keys } from '../utils.js';
+import { is_pojo, lowercase_keys, normalized_request_method } from '../utils.js';
```
Rich-Harris, line 7:
```suggestion
import { is_pojo, lowercase_keys, normalize_request_method } from '../utils.js';
```

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -383,8 +383,8 @@ async function load_shadow_data(route, event, prerender) {
 			throw new Error('Cannot prerender pages that have shadow endpoints with mutative methods');
 		}
 
-		const method = event.request.method.toLowerCase().replace('delete', 'del');
-		const handler = mod[method];
+		const method = normalized_request_method(event);
```
Rich-Harris, line 386:
```suggestion
		const method = normalize_request_method(event);
```

packages/kit/src/runtime/server/utils.js
```diff
@@ -49,3 +49,11 @@ export function is_pojo(body) {
 
 	return true;
 }
+/**
+ * @param {import('types/hooks').RequestEvent} event
+ * @returns string
+ */
+export function normalized_request_method(event) {
```
Rich-Harris, line 56:
```suggestion
export function normalize_request_method(event) {
```

Rich-Harris (approved):

PR #3907, opened 2022-02-14
handleResponse hook for logging/monitoring which route was matched

### LINKED ISSUE #3840: Implement post-response hook for logging and monitoring, including matched route and params
### Describe the problem

I've split this out from https://github.com/sveltejs/kit/issues/1194#issuecomment-1036108720, which is a slightly different issue and request.

I'm keen to implement good logging and monitoring for my SvelteKit app. For monitoring, in particular, it's extremely useful to know which route, and which params, produced the final response. I use Prometheus, and with my old Express.js app I'd typically label each finished request/response with details like `{method: 'GET', route: '/users/[id]/details', status: 200}`. This generic route parameter means that I can aggregate all similar requests to the same route and understand how long the responses took to generate on average (or stddev, etc). At present, since I don't know which route rendered the response in `handle()`, the best I can do is to use `url.pathname`, which means I can't aggregate across similar request to the same route.

Perhaps we could consider a `handleResponse` hook, with a signature something like this (and apologies, my TypeScript knowledge is almost nil, so this is pseudocode-typing!):

```
/* @returns <void> */
hooks.handleResponse = function({
    event: RequestEvent, 
    response: Response, 
    responseData: {
        duration: IntMilliseconds, 
        route: SomeKindOfRouteObject, 
        params: Record<string, string>,
    }
}) {
    ...
}
```

I could make use of details from the `route` object (ideally, I'd get the path of the `.svelte` or `.js` file which finally served the response, or something else akin to the Express route (e.g. `'/users/[id]/details'`). I'd also log the response duration, and I might make use of `params` in my logging output.

### Alternatives considered

At the moment, I don't believe it's possible to determine which route/params combo produced the response which `resolve()` from `hooks.handle()` returns, so I don't think I can get this data without a code change.

### Importance

### Additional Information

Something along these lines will be vital to people running SvelteKit apps in production, IMO.

georgecrawford:
This is a very speculative PR, with no tests or types yet. I'm interested what the view is from the maintainers about the use-case and the approach.

Fixes #3840 

From that issue:
> I'm keen to implement good logging and monitoring for my SvelteKit app. For monitoring, in particular, it's extremely useful to know which route, and which params, produced the final response. I use Prometheus, and with my old Express.js app I'd typically label each finished request/response with details like {method: 'GET', route: '/users/[id]/details', status: 200}. This generic route parameter means that I can aggregate all similar requests to the same route and understand how long the responses took to generate on average (or stddev, etc). At present, since I don't know which route rendered the response in handle(), the best I can do is to use url.pathname, which means I can't aggregate across similar request to the same route.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/build/build_server.js
+++ b/packages/kit/src/core/build/build_server.js
@@ -41,6 +41,7 @@ const get_hooks = hooks => ({
 	getSession: hooks.getSession || (() => ({})),
 	handle: hooks.handle || (({ event, resolve }) => resolve(event)),
 	handleError: hooks.handleError || (({ error }) => console.error(error.stack)),
+	handleResponse: hooks.handleResponse || (() => {}),
 	externalFetch: hooks.externalFetch || fetch
 });
 
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -107,6 +107,7 @@ export async function create_plugin(config, cwd) {
 								return {
 									type: 'page',
 									pattern: route.pattern,
+									key: route.key,
 									params: get_params(route.params),
 									shadow: route.shadow
 										? async () => {
@@ -122,6 +123,7 @@ export async function create_plugin(config, cwd) {
 							return {
 								type: 'endpoint',
 								pattern: route.pattern,
+								key: route.key,
 								params: get_params(route.params),
 								load: async () => {
 									const url = path.resolve(cwd, route.file);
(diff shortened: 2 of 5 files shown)


Rich-Harris:
It took a while, but we're finally in a position to address this issue. A new hook isn't ideal, and nor is exposing the internal `route` object, but by getting rid of fallthrough routes (#4330) we can determine the route before we even call `handle`, which means we can expose the internal `route.key` property on the event, which means you can do the logging inside `handle`.

Closing in favour of #4345

PR #3914, opened 2022-02-15
[feat] transformPage

### LINKED ISSUE #3511: `transformPage`
### Describe the problem

Until recently, the way to manipulate HTML was to do something like this in `handle`:

```js
  const rendered = await resolve(request);

  if (rendered.headers['content-type'] === 'text/html') {
    rendered.body = rendered.body.replace(/old/g, 'new');
  }

  return rendered;
}
```

With #3384, we're now dealing with `Response` objects meaning that you'd now do this:

```js
  let response = await resolve(event);

  if (response.headers.get('content-type') === 'text/html') {
    const body = await response.text();
    response = new Response(body.replace(/old/g, 'new'), {
      status: response.status,
      headers: response.headers
    });
  }

  return response;
}
```

For all the advantages brought by using `Request` and `Response`, this does not feel like an improvement. Verbosity aside, doing `await response.text()` and creating a new response feels like a lot of overhead.

We could do this:

```js
  return await resolve(event, {
    transformPage: ({ html }) => html.replace(/old/g, 'new')
  });
}
```

Then this line... https://github.com/sveltejs/kit/blob/0cca67063a4f38806eb4428b76548b8d948035ca/packages/kit/src/runtime/server/page/render.js#L218 ...would become something like 

```js
const html = options.transformPage(options.template({ head, body, assets }));
```

### Alternatives considered

The alternative is to do nothing. It depends on whether we consider manipulating the HTML response to be 'supported' — if it's a rarely-used hack, perhaps it's not necessary, but if we consider it to be a valid and common use case then the API probably pays for itself.

Minification is one use case; injecting the correct `<html lang="...">` attribute is another (albeit one that Kit will eventually have full support for anyway). I've seen it be used for other stuff — here's a snippet from the NYT covid tracker that allows multiple Kit apps to exist on the same page...

```js
  const rendered = await resolve(request);

  if (typeof rendered.body === 'string') {
    rendered.body = rendered.body
      .replace(/start\({([\s\S]+?)}\);/, (match, $1) => {
        const opts = $1.replace(/target: .+?,/, 'target,');
        return `const target = document.querySelector('#__covidtracker__');\n\t\t\ttarget.removeAttribute('id');\n\t\t\tstart({${opts}});`;
      })
      .replace(/__covidtracker__/g, `svelte-${hash(request.path)}`);
  }

  return rendered;
}
```

...and there are doubtless other uses we can't anticipate. So I vote for its inclusion, but others might feel differently.

### Importance

nice to have

### Additional Information

_No response_

pixelmund:
fixes #3511 

This PR implements the transformPage option on `resolve`. 

One question that came up was, are there or will there be more properties than `html` on transformPage?

### Tests

### Changesets

### DIFF
--- a/.changeset/forty-cycles-punch.md
+++ b/.changeset/forty-cycles-punch.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+feat: transformPage
--- a/packages/kit/src/core/build/build_server.js
+++ b/packages/kit/src/core/build/build_server.js
@@ -89,6 +89,7 @@ export class App {
 			root,
 			service_worker: ${has_service_worker ? "base + '/service-worker.js'" : 'null'},
 			router: ${s(config.kit.browser.router)},
+			transformPage: ({ html }) => html,
 			template,
 			template_contains_nonce: ${template.includes('%svelte.nonce%')},
 			trailing_slash: ${s(config.kit.trailingSlash)}
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -110,9 +110,9 @@ export async function create_plugin(config, cwd) {
 									params: get_params(route.params),
 									shadow: route.shadow
 										? async () => {
-											const url = path.resolve(cwd, /** @type {string} */(route.shadow));
-											return await vite.ssrLoadModule(url);
-										}
+												const url = path.resolve(cwd, /** @type {string} */ (route.shadow));
+												return await vite.ssrLoadModule(url);
+										  }
 										: null,
 									a: route.a.map((id) => manifest_data.components.indexOf(id)),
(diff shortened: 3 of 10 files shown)


benmccann (commented):
Thanks. Would you be able to add some docs?

I have somewhat mixed feelings about this because it doesn't seem like it should be necessary, but at the same time this is definitely a friendlier API and this is a common enough usecase that I suppose it's worth it. It looks like Rich would agree it's worth it since he filed the issue request :smile: It's too bad the Request / Response API isn't nicer, but we can't do much about that. So anyway, I guess I'm in favor of this change

pixelmund:
Yeah ofc, i'll try to add docs once i got some free time.

Rich-Harris (approved):

Rich-Harris:
thank you!

PR #3933, opened 2022-02-15
ci: tar test results before upload, filter retries, limit retention 

dominikg:
to avoid blowing the quota

### Tests

### Changesets

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,9 +60,14 @@ jobs:
           PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: ${{ steps.browser_cache.outputs.cache-hit == 'true' }}
       - run: pnpm build --filter ./packages --filter !./packages/create-svelte/templates
       - run: pnpm test
+      - name: archive test results
+        if: failure()
+        shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -60,9 +60,14 @@ jobs:
...
+        shell: bash
+        run: find . -type d -name test-results | tar -czf test-results.tar.gz --exclude='*-retry*' --files-from=-
       - name: Upload failed tests screenshots
         if: failure()
         uses: actions/upload-artifact@v2
         with:
+          retention-days: 3
           name: test-failure-${{ github.run_id }}-${{ matrix.os }}-${{ matrix.node-version }}
-          path: packages/kit/**/test-results
+          path: test-results.tar.gz
```
benmccann, line 73: I'd rather avoid this if possible. They're already compressed I believe

dominikg: We should upload a single file to avoid perf issues/errors, see https://github.com/actions/upload-artifact#too-many-uploads-resulting-in-429-responses

we could lose the .gz part though as downloads are currenly always zipped again on-the-fly.

But the uncompressed tar would increase our quota-use

benmccann (approved):

PR #3940, opened 2022-02-15
[docs] code clarifications around CSS rendering

benmccann:
changes that will help me navigate this code the next time I come back to it

Companion PR: https://github.com/sveltejs/svelte/pull/7271

### DIFF
--- a/packages/kit/src/core/build/build_client.js
+++ b/packages/kit/src/core/build/build_client.js
@@ -89,6 +89,11 @@ export async function build_client({
 		plugins: [
 			svelte({
 				extensions: config.extensions,
+				// In AMP mode, we know that there are no conditional component imports. In that case, we
+				// don't need to include CSS for components that are imported but unused, so we can just
+				// include rendered CSS.
+				// This would also apply if hydrate and router are both false, but we don't know if one
+				// has been enabled at the page level, so we don't do anything there.
 				emitCss: !config.kit.amp,
 				compilerOptions: {
 					hydratable: !!config.kit.browser.hydrate
--- a/packages/kit/src/core/generate_manifest/index.js
+++ b/packages/kit/src/core/generate_manifest/index.js
@@ -13,7 +13,8 @@ export function generate_manifest(
 	routes = build_data.manifest_data.routes,
 	format = 'esm'
 ) {
-	/** @type {Map<string, {index:number, path:string}>} */
+	/** @typedef {{ index: number, path: string }} LookupEntry */
+	/** @type {Map<string, LookupEntry>} */
 	const bundled_nodes = new Map();
 
 	// 0 and 1 are special, they correspond to the root layout and root error nodes
@@ -53,6 +54,9 @@ export function generate_manifest(
 		assets.push(build_data.service_worker);
 	}
 
+	/** @param {string} id */
+	const get_index = (id) => id && /** @type {LookupEntry} */ (bundled_nodes.get(id)).index;
+
 	// prettier-ignore
 	return `{
 		appDir: ${s(build_data.app_dir)},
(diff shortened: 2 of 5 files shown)


packages/kit/src/runtime/server/page/render.js
```diff
@@ -146,7 +146,7 @@ export async function render_response({
 
 	let { head, html: body } = rendered;
 
-	const inlined_style = Array.from(styles.values()).join('\n');
+	const rendered_styles = Array.from(styles.values()).join('\n');
```
Rich-Harris, line 149: doesn't changing this from `inlined_style` to `rendered_styles` exacerbate the confusion? this contains styles that were imported, _not_ necessarily rendered

benmccann: Maybe I'm still confused. What controls if a style will appear in `node.css` vs `node.styles`? How do the contents of these two variables differ with respect to changing the `css` compiler option?

Rich-Harris: In dev, everything goes into `node.styles`, because that's the only way to avoid FOUC: https://github.com/sveltejs/kit/blob/9bb08d732fcecfca0177e08f0df211cd0065d843/packages/kit/src/core/dev/plugin.js#L96-L102

In build, a stylesheet ends up in `node.styles` if it's below the threshold or we're in AMP mode: https://github.com/sveltejs/kit/blob/9bb08d732fcecfca0177e08f0df211cd0065d843/packages/kit/src/core/build/build_server.js#L240-L252

It will _also_ end up in `node.css`, meaning SvelteKit injects a `<link>`, but with the `disabled` attribute so that the Vite runtime doesn't load stylesheets that are already inlined:

https://github.com/sveltejs/kit/blob/9bb08d732fcecfca0177e08f0df211cd0065d843/packages/kit/src/runtime/server/page/render.js#L228-L230

Rich-Harris: I will freely concede that this is non-obvious and that there are opportunities to both name things less ambiguously and provide clearer comments in the code.

packages/kit/src/core/generate_manifest/index.js
```diff
@@ -71,8 +72,8 @@ export function generate_manifest(
 							params: ${get_params(route.params)},
 							path: ${route.path ? s(route.path) : null},
 							shadow: ${route.shadow ? importer(`${relative_path}/${build_data.server.vite_manifest[route.shadow].file}`) : null},
-							a: ${s(route.a.map(component => component && bundled_nodes.get(component).index))},
-							b: ${s(route.b.map(component => component && bundled_nodes.get(component).index))}
+							a: ${s(route.a.map(component => component && bundled_nodes.get(component)?.index))},
+							b: ${s(route.b.map(component => component && bundled_nodes.get(component)?.index))}
```
Rich-Harris, line 76: why do we need `?.` all of a sudden?

packages/kit/src/core/build/build_client.js
```diff
@@ -89,6 +89,11 @@ export async function build_client({
 		plugins: [
 			svelte({
 				extensions: config.extensions,
+				// In AMP mode, we know that there are no conditional component imports. In that case, we
+				// don't need to include CSS for components that are imported but unused, so we can just
+				// include rendered CSS.
+				// This would also apply if hydrate and router are both false, but we don't know if one
+				// has been enabled at the page level, so we don't do anything there.
```
Rich-Harris, line 96: would it be worth deduping these?

packages/kit/src/core/generate_manifest/index.js
benmccann: it's a result of typing `bundled_nodes`. as far as the typescript compiler knows getting a value from a map can always return `undefined`. it didn't type check it before when the type was `any`

packages/kit/src/core/build/build_client.js
benmccann: I might except I figure we're going to remove `config.kit.amp` which means a lot of this will get deleted, so I don't know if it's really worth it

packages/kit/src/core/generate_manifest/index.js
Rich-Harris: _We_ know that it's defined though, even if TypeScript doesn't. I'm not sure if there are any performance downsides to using `?.` instead of `.` but it feels like the correct thing to do is give TypeScript the information it needs rather than affecting runtime behaviour. One sec

Rich-Harris (approved):

PR #3941, opened 2022-02-15
Correctly identify readable node streams

### LINKED ISSUE #3939: ReadableStream not correctly detected
### Describe the bug

I use `createServer` from `@graphql-yoga/common` which provides the method `.handleRequest(...): Promise<Response>`. This response contains a `Readable` from `node:stream`.

Because of the check in
https://github.com/sveltejs/kit/blob/03a3215a12b99e82e84af19f472c7b27a13b9943/packages/kit/src/runtime/server/utils.js#L42-L44
the body is regarded as a pojo because `_writableState` is `undefined`. I think the check is to strict because we want a readable not writable.

The node sources contain this nice function
```js
function isReadableNodeStream(obj, strict = false) {
  return !!(
    obj &&
    typeof obj.pipe === 'function' &&
    typeof obj.on === 'function' &&
    (
      !strict ||
      (typeof obj.pause === 'function' && typeof obj.resume === 'function')
    ) &&
    (!obj._writableState || obj._readableState?.readable !== false) && // Duplex
    (!obj._writableState || obj._readableState) // Writable has .pipe.
  );
}
```
https://github.com/nodejs/node/blob/20347d51549bc7319ec3a44fe8b5fd25994df462/lib/internal/streams/utils.js#L14

After I copied the function above `is_pojo` and called it to test for being a readable stream it worked as expected. I will open a PR for this.

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-imy1rc?file=src%2Froutes%2Fgraphql.js&terminal=dev

Visit `/graphql` and execute the query
```graphql
query { ok }
```

### Logs

_No response_

### System Info

```shell
System:
    OS: Linux 5.10 NixOS 22.05 (Quokka) 22.05 (Quokka)
    CPU: (8) x64 Intel(R) Core(TM) i7-7700T CPU @ 2.90GHz
    Memory: 3.13 GB / 31.28 GB
    Container: Yes
    Shell: 5.1.12 - /run/current-system/sw/bin/bash
  Binaries:
    Node: 14.18.3 - /nix/store/8m7yyfkcd972irpz592ksvcc866a3af7-nodejs-14.18.3/bin/node
    npm: 6.14.15 - /nix/store/8m7yyfkcd972irpz592ksvcc866a3af7-nodejs-14.18.3/bin/npm
  Browsers:
    Chromium: 98.0.4758.80
    Firefox: 97.0
  npmPackages:
    @sveltejs/adapter-node: ^1.0.0-next.68 => 1.0.0-next.68 
    @sveltejs/kit: ^1.0.0-next.267 => 1.0.0-next.267 
    svelte: ^3.46.4 => 3.46.4
```

### Severity

blocking an upgrade

### Additional Information

_No response_

fehnomenal:
Fixes #3939

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/utils.js
+++ b/packages/kit/src/runtime/server/utils.js
@@ -32,20 +32,6 @@ export function decode_params(params) {
 	return params;
 }
 
-/** @param {any} obj **/
-function is_readable_node_stream(obj) {
-	// Copied from nodejs sources
-	return !!(
-		(
-			obj &&
-			typeof obj.pipe === 'function' &&
-			typeof obj.on === 'function' &&
-			(!obj._writableState || obj._readableState?.readable !== false) && // Duplex
-			(!obj._writableState || obj._readableState)
-		) // Writable has .pipe.
-	);
-}
-
 /** @param {any} body */
 export function is_pojo(body) {
 	if (typeof body !== 'object') return false;
@@ -53,7 +39,9 @@ export function is_pojo(body) {
 	if (body) {
 		if (body instanceof Uint8Array) return false;
 
-		if (is_readable_node_stream(body)) return false;
+		// body could be a node Readable, but we don't want to import
+		// node built-ins, so we use duck typing
+		if (body._readableState && typeof body.pipe === 'function') return false;
 
 		// similarly, it could be a web ReadableStream
 		if (typeof ReadableStream !== 'undefined' && body instanceof ReadableStream) return false;
--- a/packages/kit/src/runtime/server/utils.js
+++ b/packages/kit/src/runtime/server/utils.js
@@ -32,6 +32,20 @@ export function decode_params(params) {
 	return params;
 }
 
+/** @param {any} obj **/
+function is_readable_node_stream(obj) {
+	// Copied from nodejs sources
+	return !!(
+		(
+			obj &&
+			typeof obj.pipe === 'function' &&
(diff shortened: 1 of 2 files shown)


packages/kit/src/runtime/server/utils.js
```diff
@@ -32,16 +32,28 @@ export function decode_params(params) {
...
+	return !!(
+		(
+			obj &&
+			typeof obj.pipe === 'function' &&
+			typeof obj.on === 'function' &&
+			(!obj._writableState || obj._readableState?.readable !== false) && // Duplex
+			(!obj._writableState || obj._readableState)
+		) // Writable has .pipe.
+	);
+}
```
Rich-Harris, line 47: Thanks. We need to make this check as fast as possible (it's an edge case, we don't want to dwell on it) — we already know `obj` is truthy, so a sufficient check is probably that it has a `_readableState` property and (belt and braces, once we've established that) that `pipe` is a function:

```suggestion
	return obj._readableState && typeof obj.pipe === 'function';
}
```

Open question whether we need the indirection of a separate function call at that point

fehnomenal: Yeah you're right. I applied your suggestion and inlined the function.

Rich-Harris (approved):

Rich-Harris:
thanks!

Issue #3917, opened 2022-02-15
Clarify meaning of `shadow endpoint`.


baseballyama:
### Describe the problem

I got the below error but I took time to understand the meaning of "shadow endpoint"

> Body returned from shadow endpoint request handler must be a plain object

This is because the document doesn't appear "shadow endpoint".

### Describe the proposed solution

Write the meaning of `Sahdow Endpoint` to the Document.

https://kit.svelte.dev/docs/routing#endpoints

> If an endpoint has the same filename as a page (except for the extension), the page will get its props from the endpoint. So a page like src/routes/items/[id].svelte could get its props from src/routes/items/[id].js:
> (Such an endpoint that fires without a `load` function is called `Shadow Endpoint`.)

(I added 2nd line but please correct the sentence if necessary....)

### Alternatives considered

Remove `Sahdow Endpoint` from the error message and use another sentence.

### Importance

nice to have

### Additional Information

If other users don't actually have this issue, please close it.

Rich-Harris:
We should fix the error message rather than the documentation — 'shadow endpoints' are intended to be the default (i.e. just 'endpoints'), with 'standalone endpoints' being the exception

baseballyama:
Thank you for answering!
Right. I created PR.

I have 1 question.
The word "Shadow" is used in Code also.
Should we refactor it?
I mean should we remove "Shadow" and add "Standalone"?

netaisllc:
Consider more effective docs around this pattern.  Leaving the effectiveness of the pattern aside for the moment, I find the the docs are too quiet on the following points:
* what is the behavior of Kit when a route exists with both a `load` function and an endpoint of the same name?
* if both are executed in what order does that happen and what are the consequences?
* is the pattern of using both a really lame idea (which I assume it is)

baseballyama:
> what is the behavior of Kit when a route exists with both a load function and an endpoint of the same name?
> if both are executed in what order does that happen and what are the consequences?

Right. I think we should add some sentences to the document.

> is the pattern of using both a really lame idea (which I assume it is)

Personally, I think it is better to avoid using endpoints and standalone endpoints at the same time from the point of view of code readability. (So IMO kit doesn't need to support it.)

PR #3943, opened 2022-02-16
[fix] Remove 'Shadow' from error message

baseballyama:
fix: https://github.com/sveltejs/kit/issues/3917

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -382,7 +382,7 @@ async function load_shadow_data(route, event, options, prerender) {
 		const mod = await route.shadow();
 
 		if (prerender && (mod.post || mod.put || mod.del || mod.patch)) {
-			throw new Error('Cannot prerender pages that have shadow endpoints with mutative methods');
+			throw new Error('Cannot prerender pages that have endpoints with mutative methods');
 		}
 
 		const method = normalize_request_method(event);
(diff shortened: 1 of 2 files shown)


packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -488,15 +488,15 @@ function validate_shadow_output(result) {
...
+				'Endpoint request handler cannot use Headers interface with Set-Cookie headers'
 			);
 		}
 	} else {
 		headers = lowercase_keys(/** @type {Record<string, string>} */ (headers));
 	}
 
 	if (!is_pojo(body)) {
-		throw new Error('Body returned from shadow endpoint request handler must be a plain object');
+		throw new Error('Body returned from endpoint request handler must be a plain object');
```
Rich-Harris, line 499: So... this is a bit clunky but strictly speaking I think we need something like this, since standalone endpoint request handlers can return either a POJO or any valid `Request` body. Thoughts?

```suggestion
		throw new Error('Body returned from page's endpoint request handler must be a plain object');
```

Rich-Harris: Yeah, I guess that's a fair point, we _do_ already separate out 'standalone endpoints' and document the extra flexibility they have. So it's probably fine as-is. Thanks for the gut check!

Rich-Harris (approved):

PR #3944, opened 2022-02-16
fix: fix placeholder color search input

caesar-exploit:
Rich, I am trying to connect to the "repro" I have good ideas to improve but I still don't understand why I can't sync my commits. I hope to solve this soon or if someone can help me.

### Tests

### Changesets

### DIFF
--- a/sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
+++ b/sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
@@ -222,7 +222,8 @@
 </div>
 
 <style>
-	input {
+input {
+		--c: #444;
 		font-family: inherit;
 		font-size: 1.6rem;
 		width: 100%;
@@ -238,9 +239,14 @@
 	}
 
 	input:focus-visible {
+		--c: white;
 		background: var(--second);
-		color: white;
+		color: var(--c);
 		outline: none;
+	}
+
+	input::placeholder {
+		color: var(--c);
 	}
 
 	button[aria-label='Close'] {


benmccann:
What do you mean you can't connect to the repro? Do you mean https://kit-svelte-dev-n29rspmw6-svelte.vercel.app/?

What do you mean that you can't sync your commits? What are you trying to do and what error do you get?

What does this PR change? I don't notice anything that looks different visually

benmccann (approved):

sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
```diff
@@ -238,11 +239,16 @@
...
 	input:focus-visible {
+		--c: white;
 		background: var(--second);
-		color: white;
+		color: var(--c);
 		outline: none;
 	}
 
+	input::placeholder {
+		color: var(--c);
```
Rich-Harris, line 249: I think this should be faded, in the way that placeholders are by default on a white background — something like `rgba(255, 255, 255, 0.5)` (but only when `:focus-visible`, obviously)

Rich-Harris (changes requested):
Thank you! As commented inline, the placeholder shouldn't be the same color as the actual text — it ought to be faded. We could achieve this either with a new selector...

```css
input:focus-visible::placeholder {
  color: rgba(255, 255, 255, 0.5);
}
```

...or an additional custom property, which should be called something descriptive (`--c` is unnecessarily terse — `--color` and `--placeholder` would be better).

Rich-Harris:
Closing in favour of #3981

PR #3953, opened 2022-02-16
ci: remove '.' from archive tar hierachy and omit empty results

dominikg:
getting screenshots from the downloaded archive is a bit of a hassle. you get a tar.gz inside a .zip and then there's a lot of nested directories inside the tar.

This PR removes the root '.' directory inside the tar and also omits empty `test-results` directories to avoid confusion.

### Tests

### Changesets

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
       - name: archive test results
         if: failure()
         shell: bash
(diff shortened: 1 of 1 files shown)


dominikg:
the failing test on this PR gives us the opportunity to compare.

with these changes: https://github.com/sveltejs/kit/suites/5328621593/artifacts/166138163
without: https://github.com/sveltejs/kit/suites/5327204116/artifacts/166055201

PR #3960, opened 2022-02-16
docs(routing): Accessing slug values

gristow:
Add an example of how to access the value of a slug to the documentation. (Previously, documentation only showed how to create a route with a slug, but provided no example of how to use the slug value.)

  - Not a new feature, just documenting an existing one.
  - Not a code change, just documenting existing feature

### Tests

### Changesets
  - N/A, only documentation


Rich-Harris:
Thank you, though it's premature to talk about the page store at this point in the docs, and in most cases it's better to access it via `load` (or just getting data directly from an endpoint and not referencing `params` at all in your page code). Also, the language isn't quite right — these aren't slugs, they're parameters ('slug' is merely an example of a parameter name).

I've opened #3962 which links to the later documentation that explains this stuff in more detail.

gristow:
Got it, and thank *you*!!!

PR #3969, opened 2022-02-16
Fall back to full page reload if link href doesn't match anything

### LINKED ISSUE #3935: Rethinking `rel=external`
### Describe the problem

Right now, SvelteKit requires us to add a `rel="external"` attribute to links that should cause a regular browser navigation instead of being handled by the client-side router. This allows us to direct users to routes in the same domain that are handled outside SvelteKit.

There are a few issues I can see:

- `rel="external"` is supposed to denote a link points to an external site. While different routes on a single domain being served by SvelteKit and other services could mean exactly that, they are orthogonal concepts—they could arguably be considered a single site. The impact is probably minor but unknown: search engines could use this unfavorably.
- It's required not only for external services, but also for static assets and standalone endpoints, which is surprising.
- It optimizes for the exceptional case:
    - The router will render a 404 for any unmatched routes under the app's domain, with the benefit that it doesn't cause a full page navigation and doesn't hit the server.
    - However, 404s are the exception, not the norm: if you have a `/external` link and it doesn't match a route, the expectation is that it exists as a standalone endpoint, static asset or separate service. If it was in fact a link to SvelteKit-handled route that should 404, it'd arguably be the exception.

With that in mind, I propose that the client-side router should relegate links without a matching route to the browser for a full page navigation by default.

- In the usual case, that would be the desired effect: it'd be a static asset, a standalone endpoint or an external server.
- In the exceptional case (invalid or stale urls, etc), the experience would gracefully degrade: SvelteKit would end up ssr'ing the 404 page and the client-side app would restart normally.

On cases where the link _would_ match a SvelteKit route (eg. `/[...path].svelte` as a catch-all would mean any route would match), we'd still need an attribute to disable SvelteKit's client-side routing. We could keep `rel="external"` or come up with a SvelteKit-specific attribute like `sveltekit:noroute` (we have the `sveltekit:noscroll` prior art, but it does sound odd 🐐).

### Alternatives considered

This change would mean the exceptional case (a non-route link should be handled by SvelteKit's client-side router) _would require_ a round-trip. We could, instead, aim to minimize the amount of `sveltekit:noroute` links: instead of requiring it on static assets, standalone endpoints and external services, we could require it only on external services by either:

- Making the client-side router aware of standalone endpoints and static assets: not ideal as it'd increase the payload for the app, in some cases considerably (#2656 _got rid_ of endpoints from the payload).
- Automatically adding the attribute at compile-time if we see a matching endpoint or asset: viable improvement.

Still, I don't think it'd be worth the hassle to add the attribute to all externally handled routes just to skip a navigation on 404s, as I'd rather strive to eliminate 404s instead.

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
Currently, clicking a link like `<a href="/foo">` will fail if `/foo` doesn't 'belong' to the app — it will result in the error page being rendered with a synthetic 404. In cases where `/foo` is a real resource (handled by a separate app on the same domain, or a file in `static`, etc) we need to add `rel="external"` to trick the router into ignoring it in order to avoid that unwanted behaviour.

With this PR, if a link doesn't match a route in the manifest, we fall back to a full page navigation instead, allowing the browser to access the resource. In the case where the link doesn't go anywhere, Kit will still respond with a 404, it'll just be from the server instead of the client-side router.

Closes #3935.

### Tests

### Changesets

### DIFF
--- a/.changeset/four-news-turn.md
+++ b/.changeset/four-news-turn.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Fall back to full page reload if link href does not match route manifest
--- a/packages/kit/src/runtime/app/navigation.js
+++ b/packages/kit/src/runtime/app/navigation.js
@@ -45,8 +45,8 @@ async function invalidate_(resource) {
 /**
  * @type {import('$app/navigation').prefetch}
  */
-function prefetch_(href) {
-	return router.prefetch(new URL(href, get_base_uri(document)));
+async function prefetch_(href) {
+	await router.prefetch(new URL(href, get_base_uri(document)));
 }
 
 /**
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -149,7 +149,7 @@ export class Renderer {
 		/** @type {Map<string, import('./types').NavigationResult>} */
 		this.cache = new Map();
 
-		/** @type {{id: string | null, promise: Promise<import('./types').NavigationResult> | null}} */
+		/** @type {{id: string | null, promise: Promise<import('./types').NavigationResult | undefined> | null}} */
 		this.loading = {
 			id: null,
 			promise: null
@@ -316,6 +316,11 @@ export class Renderer {
 		const token = (this.token = {});
 		let navigation_result = await this._get_navigation_result(info, no_cache);
 
+		if (!navigation_result) {
+			location.href = info.url.href;
+			return;
+		}
+
 		// abort if user navigated during update
 		if (token !== this.token) return;
 
@@ -407,7 +412,7 @@ export class Renderer {
 
 	/**
 	 * @param {import('./types').NavigationInfo} info
-	 * @returns {Promise<import('./types').NavigationResult>}
+	 * @returns {Promise<import('./types').NavigationResult | undefined>}
 	 */
 	load(info) {
 		this.loading.promise = this._get_navigation_result(info, false);
@@ -459,7 +464,7 @@ export class Renderer {
 	/**
 	 * @param {import('./types').NavigationInfo} info
 	 * @param {boolean} no_cache
-	 * @returns {Promise<import('./types').NavigationResult>}
+	 * @returns {Promise<import('./types').NavigationResult | undefined>}
 	 */
 	async _get_navigation_result(info, no_cache) {
 		if (this.loading.id === info.id && this.loading.promise) {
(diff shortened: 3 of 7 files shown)


Conduitry:
Are you thinking we'd adjust any of the rules that the router already uses to know to perform a regular navigation to a link, or just add this one on top of them?

We should also be able to tidy the /chat link in the header with this. I don't remember what repo that lives in.

Rich-Harris:
We're not really adding any rules, we're just changing what happens when the router fails to match anything

Rich-Harris:
Hmm. This is failing because `adapter-static` has a test for SPAs: if you deploy your SPA somewhere and use a fallback `200.html` page to capture all requests... https://github.com/sveltejs/kit/blob/905457befb1e558848b0f032031019b62ad08bef/packages/adapter-static/test/utils.js#L45-L47 ...then previously SvelteKit would be responsible for rendering a 404 if nothing matched.

With this change that won't be the case. I _think_ that's okay, but it means users will see the platform 404 page rather than the app 404 page. Does that make it a breaking change? Does it change our plans at all?

Conduitry:
Wait, what's changing with the static adapter output? How does anything decide whether to render the host's 404 page instead of your app's catch-all page? That would only happen if it were configured to only catch certain routes and not all of them, right?

Rich-Harris:
I was slightly mistaken in my diagnosis. Nothing is changing with the output, but one of the tests navigates to `/nosuchpage` and expects to see a SvelteKit error page in response. Instead, SvelteKit attempts a full page navigation to `/nosuchpage`, which causes an infinite loop.

mrkishi:
That makes sense as browser delegation shouldn't happen on first render, only on subsequent navigations, yet `start({ spa: true })` will [trigger a `goto()`](https://github.com/sveltejs/kit/blob/08b18817ef07325ed19d0d5a74e72d41cdb1acd1/packages/kit/src/runtime/client/start.js#L50) during initialization.

I guess we could add an extra `forceClient` option to `goto()` that'd make its way into `renderer.handle_navigation`, but since its use would be extremely limited [1], keeping that flag private inside the router and flipping it after `init_listeners` could be an alternative.

[1] the one scenario where this would make sense is an application that wanted to serve a url from outside SvelteKit when hit directly, but still have it accessible as a SvelteKit route when navigating client-side.

Rich-Harris:
Thanks, that was what I landed on when I shut my laptop last night, so I was pleased to wake up to your branch. Feeling good about this change

PH4NTOMiki:
@Rich-Harris this kinda has a flaw, when using SPA mode, when 404 occurs this leads to infinite reloads. We have to fix this somehow, maybe set some sessionStorage value that counts number of redirects and if greater than 1 then stop. or something like that.

PH4NTOMiki:
I know there is code for `initialized` in this PR but still it breaks somehow

PH4NTOMiki:
`Not found` gets rendered for a small time, and then reload, and then again.

PR #3965, opened 2022-02-16
[fix] make demo app work without JS

geoffrich:
Fixes #3952

After the introduction of shadow endpoints, the demo app would throw a "Body returned from shadow endpoint request handler must be a plain object" error when adding a todo with JS disabled. This was because we were returning a Response, not a POJO.

After fixing that issue, I also had to modify the request passed to the api in the `get` handler. When rendering a page in response to a POST, the `get` handler is called with a POST request. Since we passed the original request to the `api` method, we were sometimes making a POST request when we needed a GET. By making a new request, we make sure the method is GET in all cases. I added an explanatory comment to the code since it's a little weird.

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

Issue #3952, opened 2022-02-16
Demo App: Todo doesn't work without Javascript


decherd:
### Describe the bug

The Demo App Todo page does not work without Javascript and returns a 500 status code along with this error: Body returned from shadow endpoint request handler must be a plain object.

### Reproduction

I don't think I need to link back to this repo. Again, it is just the demo app.

### Logs

```shell
Body returned from shadow endpoint request handler must be a plain object
Error: Body returned from shadow endpoint request handler must be a plain object
    at validate_shadow_output (file:///Users/drew/Projects/demo_app2/.svelte-kit/runtime/server/index.js:1999:9)
    at load_shadow_data (file:///Users/drew/Projects/demo_app2/.svelte-kit/runtime/server/index.js:1910:38)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async load_node (file:///Users/drew/Projects/demo_app2/.svelte-kit/runtime/server/index.js:1565:5)
    at async respond$1 (file:///Users/drew/Projects/demo_app2/.svelte-kit/runtime/server/index.js:2194:15)
    at async render_page (file:///Users/drew/Projects/demo_app2/.svelte-kit/runtime/server/index.js:2395:19)
    at async resolve (file:///Users/drew/Projects/demo_app2/.svelte-kit/runtime/server/index.js:2660:11)
    at async Object.handle (/Users/drew/Projects/demo_app2/src/hooks.ts:9:24)
    at async respond (file:///Users/drew/Projects/demo_app2/.svelte-kit/runtime/server/index.js:2567:20)
    at async file:///Users/drew/Projects/demo_app2/node_modules/@sveltejs/kit/dist/chunks/index.js:244:24
```

### System Info

```shell
System:
    OS: macOS 12.1
    CPU: (4) x64 Intel(R) Core(TM) i5-6360U CPU @ 2.00GHz
    Memory: 157.93 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.0 - /usr/local/bin/node
    Yarn: 1.10.1 - /usr/local/bin/yarn
    npm: 8.3.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 98.0.4758.80
    Firefox: 66.0.4
    Safari: 15.2
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.17
    @sveltejs/kit: next => 1.0.0-next.272
[...]

PR #3970, opened 2022-02-16
Alternative fix for #3952

Rich-Harris:
This is an alternative to #3965 — was easier to do as a separate PR than to make commit suggestions. Rather than each handler needing to `await response.json()`, we can just change the return value from the `api` helper to make it a valid shadow endpoint response.

I found one other issue with the current version of the app. If you patch or delete a todo with JS disabled, the URL will update to `/todos?_method=PATCH`. This is no good — reloading the page will cause an error, because method overrides can't be used with GET requests. To fix that, I changed the handlers so that they're redirecting back to `/todos` instead (which we can do, because we're not using the return values for anything — we're relying on invalidation instead).

### Tests

### Changesets

### DIFF
--- a/.changeset/calm-spiders-turn.md
+++ b/.changeset/calm-spiders-turn.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+make demo app work without JS
--- a/packages/create-svelte/templates/default/src/routes/todos/_api.ts
+++ b/packages/create-svelte/templates/default/src/routes/todos/_api.ts
@@ -1,5 +1,5 @@
 /*
-	This module is used by the /todos endpoint to 
+	This module is used by the /todos endpoint to
 	make calls to api.svelte.dev, which stores todos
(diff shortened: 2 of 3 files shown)


Rich-Harris:
Gah, I guess this doesn't have any error handling, since we're not just returning the response directly

geoffrich:
Thanks, this seems cleaner!

> Gah, I guess this doesn't have any error handling, since we're not just returning the response directly

I don't think that's a dealbreaker, since even if we return a non-successful response from the endpoint, we don't do anything with it on the page. We're not providing an error handler to the `enhance` action, and the page isn't exposing any error props to be populated by the endpoint response. It would be good to have more robust error handling in the demo app, but maybe we could look at that in a separate PR?

Rich-Harris:
Yeah, you're right, we need to improve error handling all round. Fine line between doing things 'right' and making the example too verbose to be easily understood by a newcomer

PR #3972, opened 2022-02-17
Include remote_addr of client in request header

TheRealThor:
Currently the IP address (remote_addr) of clients that send a request is not available in endpoints or handle because the header object of the raw request is simply copied. However the remote_addr lies within the socket object of the raw request. The IP address of the requesting client is an important env needed for various use cases, like security and logging tasks. Hence, it is useful to include this env into the request object available for endpoints. 

- [X ] This message body should clearly illustrate what problems it solves.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/node.js
+++ b/packages/kit/src/node.js
@@ -53,6 +53,7 @@ function get_raw_body(req) {
 /** @type {import('@sveltejs/kit/node').GetRequest} */
 export async function getRequest(base, req) {
 	let headers = /** @type {Record<string, string>} */ (req.headers);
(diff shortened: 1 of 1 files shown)


Conduitry:
This doesn't feel to me like the right way to handle this. It's not a real header. And, while `req.socket.remoteAddress` may be available in Node environments, it may not be in other environments that use this same utility function.

This feels like a job for https://kit.svelte.dev/docs/adapters#supported-environments-platform-specific-context for each relevant adapter.

Rich-Harris:
@Conduitry is right, stuff like this belongs in `event.platform` if it's needed

PR #3975, opened 2022-02-17
Simplify hash change click handler

PH4NTOMiki:

### Tests

### Changesets

Simplifies hash handling initiated by click. as @benmccann pointed out in https://github.com/sveltejs/kit/pull/3873#discussion_r805219310 we don't really need to call renderer.update when the only thing needed update is page store, so that's what we are doing here, this is close to the browser default behavior, and since 
@Rich-Harris merged pulls #3931 and #3938 we can now do that.

### DIFF
--- a/.changeset/sharp-olives-sparkle.md
+++ b/.changeset/sharp-olives-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Update page store without rerunning load when hash changes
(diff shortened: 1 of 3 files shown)


PH4NTOMiki:
Yeah, no problem, yeah that makes sense, I didn't want to change to much on my own. but that's definitely better.

Rich-Harris (approved):

PH4NTOMiki:
@Rich-Harris just one more thing, in normal click navigation, first the URL changes, and then store changes, in this hash case, it's the opposite, we can fix that by calling `update_page_store` inside that hashchange handler(inside if block in it)

Issue #3324, opened 2022-01-13
Use shiki-twoslash for docs


Rich-Harris:
### Describe the problem

These are the endpoint docs rendered on my screen:

<img width="1089" alt="image" src="https://user-images.githubusercontent.com/1162160/149367599-b4b64c3f-488a-4204-a916-6efe41f3038e.png">

There's a full page of TypeScript throat-clearing before we get to the bit that people actually want to see:

<img width="1072" alt="image" src="https://user-images.githubusercontent.com/1162160/149367796-c8978e92-86f4-4d13-8adb-321e0ab84d8c.png">

It's useful to have the types documented, but not like this. 

Another problem with our docs: the types can get out of date, [as seen here](https://github.com/sveltejs/kit/pull/3323). It would be nice if _somehow_ snippets in our docs were typechecked.

### Describe the proposed solution

We could potentially solve both problems with @orta's https://shikijs.github.io/twoslash. 

Questions to which I don't know the answer:

* Can it be made to work with Svelte components?
* Does it work with JSDoc syntax?
* Do we have any concerns about accessibility, if type information is hidden behind hovers?

Aside: doing this would presumably require cooperation between this repo, [sveltejs/sites](https://github.com/sveltejs/sites), and [sveltejs/action-deploy-docs](https://github.com/sveltejs/action-deploy-docs). Before leaping into it we might want to consider whether the current setup is to our liking, or whether we might want to move to multi-page docs with search and interactive components (likely powered by mdsvex).

### Importance

nice to have

geoffrich:
re: a11y - the type annotations shiki-twoslash shows on hover are not accessible, by keyboard or by screen reader. Right now it just hide/shows a pseudoelement on `:hover`. I'm honestly not sure if there's a good way to make it accessible, especially since tooltips are infamously tricky in that regard. 

Though I definitely see the value it brings, especially around keeping the docs from getting out of date. I don't think the type annotations not being accessible is a dealbreaker if we make sure the type info is available somewhere else (details/summary block? separate page? not sure). I think having the type info available outside of a hover is good _anyway_, since people might want to reference them without needing to keep hovering over a section of the docs.

jasonlyu123:
It seems like we could use our CSS to provide the hover, focus style. So that we can make the tooltip also show up with focus. The one on the website seems to be from the CSS in their site instead of the library. 

>Can it be made to work with Svelte components?

[...]

PR #3986, opened 2022-02-17
first stab at shiki-twoslash integration

Rich-Harris:
#3324. This is very WIP. If anyone can help me figure out how to do the hovers, I'd be grateful — the [docs](https://shikijs.github.io/twoslash/) have inline `<data-lsp>` elements which aren't being generated in this setup, and I'm not sure why (nor am I sure where the client-side JS is supposed to come from).

* [ ] Hovers
* [ ] Update all the docs
* [ ] Improve styles (match colour scheme, give empty lines a height)
* [ ] Would be cool if we could show the filename outside the code block, if one is specified
* [ ] Stretch goal: make it work with `.svelte` blocks (this could be a follow-up PR)
* [ ] We should have a reference of all the available types on https://kit.svelte.dev/docs/typescript (and maybe occurences to them in code blocks could link to the references?)

---

### Tests

### Changesets

### DIFF
--- a/documentation/docs/01-routing.md
+++ b/documentation/docs/01-routing.md
@@ -47,46 +47,21 @@ A file or directory can have multiple dynamic parts, like `[id]-[category].svelt
 
 Endpoints are modules written in `.js` (or `.ts`) files that export functions corresponding to HTTP methods. Their job is to allow pages to read and write data that is only available on the server (for example in a database, or on the filesystem).
 
-```ts
-// Type declarations for endpoints (declarations marked with
-// an `export` keyword can be imported from `@sveltejs/kit`)
-
-export interface RequestHandler<Output = Record<string, any>> {
-	(event: RequestEvent): MaybePromise<
-		Either<Output extends Response ? Response : EndpointOutput<Output>, Fallthrough>
-	>;
-}
-
-export interface RequestEvent {
-	request: Request;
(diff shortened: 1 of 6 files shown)


Rich-Harris:
cc'ing @orta in case there's a straightforward answer to the question around hovers, as I'm a bit flummoxed. TIA!

Am also curious about whether it's possible to create our own theme. At the moment I can't even seem to change the theme to one of the other presets, not sure why.

jasonlyu123:
Regarding changing the theme preset, it seems to be because `createShikiHighlighter` is caching the result. So changing the option still bring up the cached version. It seems to change after restarting the dev server. 

From the [docs of shiki](https://github.com/shikijs/shiki/blob/main/docs/themes.md) we can use the `css-variable` theme and set the css-variable like

```css
:root {
    --shiki-token-keyword: var(--code-keyword);
    --shiki-token-function: var(--code-function);
    --shiki-token-string-expression: var(--code-template-string);
    --shiki-color-background: var(--code-bg);
}
```
圖片

Or load the theme from a JSON file.

Rich-Harris:
Alright, it's starting to come together! Thanks all https://kit-svelte-dev-git-shiki-twoslash-svelte.vercel.app/docs/routing

benmccann (approved):

PR #3989, opened 2022-02-18
allow to type `Params` of `RequestHandler`

ivanhofer:

### Tests

### Changesets

Currently it is not possible to type `Params` of a `RequestHandler` handler function

```ts

type ReturnType = { id: string }

// current
export const get: RequestHandler<ReturnType> = ({ params }) => {
   return {
      body: {
         // `params.id` is of type `string | undefined`
         id: params.id
      },
   }
}

// with the changes of this PR
type ParamsType = { id: string }

export const get: RequestHandler<ReturnType, ParamsType> = ({ params }) => {
   return {
      body: {
         // `params.id` is of type `string`
         id: params.id
      },
   }
}

```


dummdidumm:
If we add this, I feel like we should make this a breaking change and switch the order of generics. Feels strange to me to have the input generic be the second generic, but maybe that's just me

ivanhofer:
I was also thinking about this, but didn't want to introduce a breaking change. I'll update it

.changeset/proud-cooks-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+allow to type `Params` of `RequestHandler`
```
benmccann, line 5:
```suggestion
[breaking] allow providing `Params` type argument for `RequestHandler`
```

benmccann:
this PR would need to be rebased

ivanhofer:
> this PR would need to be rebased

done

Rich-Harris:
agree that input should go first, though i'll note that if we manage to pull off #647 we'll have to revert this breaking change with another. i don't think that's a reason not to do it though

Rich-Harris (approved):

PR #3993, opened 2022-02-18
adapter-node: Make remoteAddress of requesting client available

TheRealThor:
This makes available to endpoints the IP address of client requesting the resource via platform object.

Improved solution to [#3972 ](https://github.com/sveltejs/kit/pull/3972)

### Tests

### Changesets

### DIFF
--- a/packages/adapter-node/src/handler.js
+++ b/packages/adapter-node/src/handler.js
@@ -45,7 +45,7 @@ const ssr = async (req, res) => {
 		return res.end(err.reason || 'Invalid request body');
 	}
 
-	setResponse(res, await app.render(request, { 'platform':  {'remoteAddress' : req.socket.remoteAddress}  }));
+	setResponse(res, await app.render(request, { platform:  {remoteAddress : req.socket.remoteAddress}  }));
 };
 
 /** @param {import('polka').Middleware[]} handlers */
--- a/packages/adapter-node/src/handler.js
+++ b/packages/adapter-node/src/handler.js
@@ -45,7 +45,7 @@ const ssr = async (req, res) => {
 		return res.end(err.reason || 'Invalid request body');
 	}
 
-	setResponse(res, await app.render(request));
+	setResponse(res, await app.render(request, { 'platform':  {'remoteAddress' : req.socket.remoteAddress}  }));
 };
 
 /** @param {import('polka').Middleware[]} handlers */


benmccann:
Is this necessary? I would think that you could get the client address from an existing HTTP header. E.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

It feels weird to me to expose `remoteAddress` specifically. If it is really necessary, I'd probably rather expose the whole request - though others may have different opinions about this

TheRealThor:
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

This header only is set when a proxy is used because in this case "remoteAddress" would only show the IP address of the proxy (who in fact send the request). 

I already [proposed to show the IP-address as a additional header](https://github.com/sveltejs/kit/pull/3972) information to the request, where Rich Harris suggested it should be a adapter/platform relevant information. 

Whether the whole request (raw request) or just the IP address ist exposed, in any case the IP address of the remote client is an important information for security purposes and "it feels weird" to me that to date nobody thought about that in the first place.

mrkishi:
Yeah, I wouldn't make the assumption apps are always behind a proxy, at least not with node and custom adapters support.

I don't know about exposing the raw request, but even if it made sense in the end, I wonder if it _also_ makes sense to lift common features so that they work across adapters. While not all adapters may have the remote address available, as pointed out in the other pr, most will, each exposing it in their own way (eg. `req.remoteAddress`, `CF-Connecting-IP` and `X-Nf-Client-Connection-Ip` headers). If they all provided it as a compatible `remoteAddress`, apps would be at least _somewhat_ portable. What do you think, Ben?

ps. it might be worth it to make it `getRemoteAddress()` in case some adapters have to do work to get it, but I expect at most they'd need to do a header lookup and I don't know that's measurably different than adding to the closure. But if it turns out there's a lot we could normalize under `platform`, maybe there's a better/leaner API?

Conduitry:
I'm a bit worried about lulling people into a false sense of security by having similar APIs in `platform` across different adapters. But that said, it does provide a lower-stakes way to test out things like this before potentially promoting them to the shared API.

I agree that we don't want to assume that certain headers are present from a proxy in front of the Node app. But also, `req.socket.remoteAddr` would be the proxy address if there were a proxy, and not the user's address, right? I'm not sure what the best API is here.

TheRealThor:
> But also, `req.socket.remoteAddress` would be the proxy address if there were a proxy, and not the user's address, right?

The regular case is that the client is directly connected to the server, especially to a node-server, as proposed here.  
However, If its an external/untrusted proxy , behind which the client sits, still _only_ `req.socket.remoteAddress` can be trusted as header data like "X-Forwarded-For" etc can be set by the client. Even if it's an internal/trusted proxy, `req.socket.remoteAddress` still might be needed  to see if the sender is indeed the trusted proxy. 

The special case, that someone uses an internal proxy, making remoteAddress always returning that proxy address should be addressed in a later commit and/or by the user of the proxy with a header solution.

Generally it would be logical to have the remoteAddress in the request object, making it optional so that it could be set by the relevant adapter.

Rich-Harris:
Doing `app.render(request, { platform: { req } })` is definitely the easiest option, though I do quite like @mrkishi's idea of making things like this standard across environments — `event.remoteAddress` or whatever. Are we worried that it wouldn't be accessible within some adapters? Or that once we start adding things like this we'll be faced with an onslaught of requests for random properties to be added to solve niche scenarios?

mrkishi:
A _best-effort_ `platform.remoteAddress` is growing on me. Cross-compatibility would not be guaranteed so you'd still need to go to the docs for migration, yet it'd be a lot more convenient to find out if your new adapter also supports it.

I wouldn't be worried about a false sense of API compatibility because, alternatively, users would have used some adapter-specific API and they would've encountered a similar error when migrating. If that is a prevalent worry, we could also go with a Proxy or accessor API so that useful errors are possible, at a (small?) performance cost, eg. 
```js
platform.get('remoteAddress') === new Error(`adapter-x has no support for "${property}". see https://adapter-docs/ for details.`)
```

benmccann:
I believe `remoteAddress` with Node's meaning of it is only available in `adapter-node`. In `adapter-vercel`, etc. you could use `X-Forwarded-For`, but it has a different meaning. I suppose that's okay, but it's a bit weird that it'd have different behaviors on different platforms.

mrkishi:
I suppose we'd need to add options to `adapter-node` telling it to trust certain proxies, as you'd have to do on a regular node server.

evdama:
> #3972
Having the requesing client IP address available is something we very much need in order to use it to query the ipregistry.org API for example in order to get information about this IP belongs to a botnet for example, or the latitude/longitute for mapping.

I've been using Sapper in the past with some express middleware (request-ip) which looks for a bunch of information inside the request header https://github.com/pbojinov/request-ip#how-it-works

The user ip is determined by the following order:

```
The user ip is determined by the following order:

X-Client-IP
X-Forwarded-For (Header may return multiple IP addresses in the format: "client IP, proxy 1 IP, proxy 2 IP", so we take the the first one.)
CF-Connecting-IP (Cloudflare)
Fastly-Client-Ip (Fastly CDN and Firebase hosting header when forwared to a cloud function)
True-Client-Ip (Akamai and Cloudflare)
X-Real-IP (Nginx proxy/FastCGI)
X-Cluster-Client-IP (Rackspace LB, Riverbed Stingray)
X-Forwarded, Forwarded-For and Forwarded (Variations of #2)
req.connection.remoteAddress
req.socket.remoteAddress
req.connection.socket.remoteAddress
req.info.remoteAddress
If an IP address cannot be found, it will return null.
```

evdama:
Maybe using 'req.socket.remoteAddress` is a good first step afer which, at least adapter-node, could implement something like this https://github.com/pbojinov/request-ip/blob/915d984b46d262e4418a889abc7601c38d47f049/src/index.js#L48 no?

benmccann:
> I suppose we'd need to add options to adapter-node telling it to trust certain proxies, as you'd have to do on a regular node server.

This is what we already do for `x-forwarded-proto` and `x-forwarded-host`. I think if we do it here we need to call the field something other than `remoteAddress` since it no longer behaves the same as Node's `remoteAddress`. I also wonder if there might be some case where I user would want both the `remoteAddress` and the `x-forwarded-host` and by having it behind a flag you can only access one or the other.

evdama:
The horrible war in 🇺🇦 and the impossed sanctions on 🇷🇺 create another use case for having the remote clients IP address available.

Using the remote IP of a client we are going to call the [ipregistry.org](https://ipregistry.co/) API whether this is a russian user or not. If so, he/she is redirected to a dedicated page, telling him/her that they can visist our website, but that payments are disabled.

evdama:
> ps. it might be worth it to make it `getRemoteAddress()` in case some adapters have to do work to get it, but I expect at most they'd need to do a header lookup

- Using something like `getRemoteAddress()` might indeed be a good way to abstract accross different platforms and/or ways users deploy their SvelteKit apps (proxy, no proxy... etc).
- Depending on the various cases `request-ip` [does a header lookup](https://github.com/pbojinov/request-ip#how-it-works).
- getting this feature somewhat harmonized accross platforms/adapters would then also make it easier to migrate (except for some cases were docs could pave the way)

@benmccann @mrkishi @TheRealThor @Rich-Harris @Conduitry what do you think?

Rich-Harris:
Ok, recapping some discussion between the maintainers:

* Client IP address is useful information to surface regardless of adapter, and we want the experience of using SvelteKit to be normalised across adapters to the greatest degree possible
* That means that we'd ideally have a property like `event.clientAddress`
* To populate that property, adapters would need to pass in a function to `server.render`:
    ```js
    response = await server.render(request, {
      getClientAddress: () => {...}
    });
    ```
* SvelteKit could make `event.clientAddress` a getter, so that we don't need to invoke the function unnecessarily
* If an adapter doesn't provide `getClientAddress`, we should throw an error like '`adapter-foo` does not make `clientAddress` available'
* On platforms like Netlify, Cloudflare Pages, Vercel etc we can trust that the IP address is correct. In cases like `adapter-node` we probably need the user to opt in, because we shouldn't trust the proxy on their behalf (just like how we treat `HOST_HEADER` and `PROTOCOL_HEADER` currently)

Rich-Harris:
going to close this in favour of #4289 — thanks for getting the ball rolling

Issue #3987, opened 2022-02-17
[docs] Inline code blocks are not rendered properly in the sidebar


malinowskip:
### Describe the problem

Inline code blocks are rendered as `<code>Some heading in the docs</code>` in the sidebar nav.
<img width="266" alt="Zrzut ekranu 2022-02-17 o 23 02 37" src="https://user-images.githubusercontent.com/47485431/154578376-eea1e125-ccb6-42ae-806e-b93c0f629b28.png">

### Describe the proposed solution

Remove backticks from [section headings](https://github.com/sveltejs/kit/blob/ec62766ac115ba02ed60b717ae9ba0ca9bafab6f/documentation/docs/15-seo.md?plain=1#L25).

### Alternatives considered

Using `{@html ...}` to render section headings in the sidebar causes problems (code block contents are also rendered as HTML, even if escaped), so making it work would require other rewrites.

### Importance

nice to have

### Additional Information

I also noticed that backticks are not rendered as code blocks in the new search utility.

<img width="541" alt="Zrzut ekranu 2022-02-17 o 23 23 21" src="https://user-images.githubusercontent.com/47485431/154581015-3d6aeb01-1582-4382-8a0b-2993b8c8c4ea.png">

bluwy:
Fixed in #4011

PR #3990, opened 2022-02-18
Removing the backticks to fix the problem of #3987

Swarleys:
Removing the backticks of title and meta so it will not be rendered as `<code>` in the sidebar nav, as opened on the issue https://github.com/sveltejs/kit/issues/3987

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [X] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [X] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [ ] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [ ] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #4000, opened 2022-02-18
[docs] render code blocks in sidebar nav

malinowskip:
Possible solution to https://github.com/sveltejs/kit/issues/3987

### Tests

### Changesets


bluwy:
Looks like this makes it stand out in the sidebar

image

Even if we can dim down the styles, I wonder if it would be better to avoid code blocks / backticks altogether so the font looks more consistent (#3990)

Rich-Harris:
thanks, this was fixed separately in #4011

Issue #3996, opened 2022-02-18
Generated TypeDoc


benmccann:
### Describe the problem

It's hard to keep the docs in sync with the types and our current method of displaying them isn't always the most readable

### Describe the proposed solution

I took a stab at playing around with this. There are two approaches I tried:
- ~~https://github.com/sveltejs/kit/tree/typedoc-markdown~~
- https://github.com/sveltejs/kit/tree/typedoc-json

At first I tried having it spit out `.md` files. That was a bit tricky because they don't quite match the format expected with what we've hand-written. E.g. it includes `#` and `##` whereas our parser expects headings to start only at `###`

Rich suggested a better approach might be to generate JSON instead and work from that. That turned out to be the more fruitful and customizable approach I think

![Screenshot from 2022-02-18 10-01-30](https://user-images.githubusercontent.com/322311/154741114-4cdb36da-0509-4594-9d02-3a199d72cf78.png)

### Alternatives considered

Copy `.d.ts` files into docs still

### Importance

nice to have

benmccann:
Closing in favor of https://github.com/sveltejs/kit/issues/4002

PR #4005, opened 2022-02-19
Typedoc

benmccann:
Closes https://github.com/sveltejs/kit/issues/3996

Preview: https://kit-svelte-dev-l413933f6-svelte.vercel.app/types

Things left to do:
- [ ] currently displaying only property names. need to display the types as well. it's a little tricky because the types can be complex. probably need a shared fragment or component to do this
- [ ] it would be nice to link to types that are from SvelteKit. maybe link to MDN as well. need to search through the json file to find which exist
- [ ] shouldn't check in the `.json` file. generate it at build time and `.gitignore` it
- [ ] there's a few things like type aliases, the location of the source file, etc. that i'm currently not displaying. we probably don't need to display a lot of these, but there might be a couple we'd want to add
- [ ] the docs don't link to this new page at all. perhaps it shouldn't even be a new page and should just be on the TypeScript page or something. I haven't thought about this yet
- [ ] the styles need to be cleaned up. I copied them from the faq page as a start, but there's probably a lot that can be removed or written without `global`
- [ ] typedoc is generating a good handful of errors when running it

Possible limitations:
- [ ] the `.json` output from `typedoc` doesn't include the comments. I'm not sure if there's a way to get them? https://typedoc.org/guides/doccomments/ would seem to suggest it's possible

### DIFF
--- a/documentation/types.json
+++ b/documentation/types.json
@@ -1,0 +1,5111 @@
+{
+	"id": 0,
+	"name": "SvelteKit - v1.0.0-next.278",
+	"kind": 1,
+	"kindString": "Project",
+	"flags": {},
+	"originalName": "SvelteKit",
+	"children": [
+		{
+			"id": 1,
+			"name": "App",
+			"kind": 128,
+			"kindString": "Class",
+			"flags": {},
+			"children": [
+				{
+					"id": 2,
+					"name": "constructor",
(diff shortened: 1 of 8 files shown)


benmccann:
Closing in favor of https://github.com/sveltejs/kit/pull/4011

Issue #4002, opened 2022-02-18
Document types


Rich-Harris:
### Describe the problem

The TypeScript section of the docs documents the `App` namespace, but none of the types that are exported from `@sveltejs/kit` or the `$app` modules. It would be good to have this information present in a reference (albeit human-readable — none of the incomprehensible garbage you get from stuff like typedoc by default).

In addition, we could definitely make stuff like this a bit nicer to look at:

<img width="952" alt="image" src="https://user-images.githubusercontent.com/1162160/154771804-d8265bf2-3637-4b2d-b92c-135fd07d5afd.png">

### Describe the proposed solution

This could take many forms, and will probably need a bit of experimentation to get right.

### Importance

nice to have

PR #4010, opened 2022-02-19
simplify types

Rich-Harris:
The motivation behind this is an idea I had about documenting types (#4002), but I think it's valuable in its own right so I'm opening this separately from any docs changes.

Right now we have types scattered amongst a bunch of nebulously-scoped `.d.ts` files. It would be much simpler all round if we just split the types into public (`index.d.ts`), private (`internal.d.ts`) and ambient (`ambient.d.ts`). I don't think the current setup serves any real purpose.

I haven't made any changes to the types themselves — just where they live — but I do think there's a worthwhile conversation to be had about where we draw the line between public and private. There's definitely some things in `internal.d.ts` that wouldn't be terribly out of place in `index.d.ts`. I don't think we need to solve that in this PR though.

### Tests

### Changesets


dummdidumm (approved):
Agree with this change. The files are not that huge that it gets out of hand and I also often wasn't sure what goes where

PR #4011, opened 2022-02-19
Document types

Rich-Harris:
Follow-up to #4010. The idea here is to make types linkable throughout the docs — rather than the unreadable gibberish you get from automatic doc generators, this literally just copies the source `.d.ts` into the docs, extracts leading JSDoc comments, and crosslinks everything.

So for example in the [endpoint docs](https://kit-svelte-dev-git-document-types-svelte.vercel.app/docs/routing#endpoints), the `RequestHandler` type becomes a link...

image

...that points [here](https://kit-svelte-dev-git-document-types-svelte.vercel.app/docs/types#sveltejs-kit-requesthandler):

image

* [x] Prevent type docs linking to themselves (e.g. `RequestHandler` -> `RequestHandler` in the example above)
* [x] Improve the styles a bit
* [ ] ~~Add more JSDoc comments explaining the various types~~ this can happen in a follow-up
* [ ] ~~Rethink `App` vs `App` — maybe we rename the class?~~ follow-up in #4030
* [x] Include types in search index
* [x] Either remove links from `lsp` attributes, or have the tooltip actually link to them
* [x] make comments work on mobile (i.e. not stretch out into the distance) somehow


pngwn:
Can we set a max length on the snippets? Some of them are very difficult to read on mobile.

Rich-Harris:
Do you mean running Prettier on the snippets with a shorter line length, or something?

pngwn:
Yeah, that would be good if it is possible.

Rich-Harris:
Made that change. The JSDoc comments will need to be updated (and maintained) manually though

Rich-Harris:
Would be good if this linked to `Adapter`. At the very least the HTML shouldn't be visible:

image

PR #4013, opened 2022-02-19
Update 60-env-vars.md

d3sandoval:
Vite doesn't use `process.` it's `import.meta`

Fixes `process not defined` errors when using environment variables

@bluwy edit: Explain that `import.meta.env.` is used for build time.

### Tests

### Changesets

### DIFF
--- a/documentation/faq/60-env-vars.md
+++ b/documentation/faq/60-env-vars.md
@@ -2,8 +2,8 @@
 title: How do I use environment variables?
 ---
 
-Vite uses [dotenv](https://github.com/motdotla/dotenv) to load environment variables from a file named `.env` or similar. Only environment variables prefixed with `VITE_` are exposed ([you can set `envPrefix` to change this](https://vitejs.dev/config/#envprefix)). Vite will use these and statically replace them at build-time.
+Vite uses [dotenv](https://github.com/motdotla/dotenv) to load environment variables from a file named `.env` or similar. Only environment variables prefixed with `VITE_` are exposed ([you can set `envPrefix` to change this](https://vitejs.dev/config/#envprefix)). You can access the variable using `import.meta.env.VITE_ENV_VAR`, and Vite will statically replace them at build-time.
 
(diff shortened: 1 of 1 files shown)


Rich-Harris:
`import.meta.env` is for _build time_, not run time — I think the existing docs are correct? (Notwithstanding the fact that not every environment exposes `process`, of course)

Conduitry:
This is code that, on the server side, isn't changed by Vite at all. It's just `process.env.whatever` which Vite leaves alone and Node then interprets as it normally would, at runtime.

Conduitry:
Whoops, I'm on the mobile app and I missed Rich's comment. Yes the `import.meta` thing is replaced at build time but only for env vars beginning with `VITE_`. If you want runtime stuff, Vite is not involved.

d3sandoval:
Interesting. The FAQ is a bit confusing then... If I'm using an environment variable in a .js file, `process` isn't defined:
```js
/* supabaseClient.js */
import { createClient } from '@supabase/supabase-js';

const supabaseUrl = process.env.VITE_SUPABASE_URL;
const supabaseAnonKey = process.env.VITE_SUPABASE_ANON_KEY;

export const supabase = createClient(supabaseUrl, supabaseAnonKey);
```

```js
/* src/routes/auth/index.js */
<script>
	import { supabase } from '$lib/supabaseClient';

	let loading = false;
	let email;

	console.log(supabase);
</script>
```

image

But if I change `process` to `import.meta`, it runs successfully. I'm assuming this means I can only access `process` from a `.svelte` file?

bluwy:
Yeah I think the above paragraph "Vite uses dotenv..." could have mention the `import.meta.env.*` syntax as that's for build-time replacements. The paragraph later explains for runtime replacements. 

> I'm assuming this means I can only access process from a .svelte file?

You should be using `import.meta.env` in your example code. `process.env` should only be used in server code as that's determined in runtime, as in the browser runtime there's no `process`.

d3sandoval:
Ohhh I see! That makes sense. Appreciate the help.

bluwy:
Updated the docs to cover the discussion we had. Ready for a quick review.

Rich-Harris (approved):

Rich-Harris:
much better than what we had before 👍

PR #4015, opened 2022-02-20
Fix html attribute escaping

### LINKED ISSUE #4014: Missing ampersand escaping in html attributes
### Describe the bug

`escape_html_attr` doesn't escape `&`, so values like `&quot;` get transformed by the browser.

### Reproduction

Use `fetch` in a page `load`:
```js
	const response = await fetch('/url-with-&quot;');
	return { status: 200, props: await response.json() };
}
```

The page will include:
```html
<script type="application/json" data-type="svelte-data" data-url="/url-with-&quot;">
```

Then:
```js
script.dataset.url === '/url-with-"'
```

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    Memory: 1.04 GB / 7.88 GB
  Binaries:
    Node: 17.4.0 - ~\scoop\apps\nodejs\current\node.EXE
  Browsers:
    Chrome: 98.0.4758.102
    Edge: Spartan (44.19041.1266.0), Chromium (98.0.1108.50)
  npmPackages:
    @sveltejs/eslint-config: github:sveltejs/eslint-config#v5.8.0 => 5.8.0
```

### Severity

annoyance

### Additional Information

_No response_

mrkishi:
Fixes #4014.

This just adds the missing `&` and removes superfluous `<>` from the html attribute escape dictionary.

`<>` are treated as raw characters inside a double-quoted attribute value. This means they can contain `</script>` and friends inside, but if we're already in rawtext mode (meaning, inside an open `<script>` or other greedy html tags), then we're already inside an xss vector and we'd have to escape every single character, as `kill(everyone)` would be a valid value.

Try it: `<tag data-value="<script>alert('pwned')</script>">` is properly handled by browsers.

### Tests

Sorry but tests timeout at some point on this weak machine.

### Changesets

### DIFF
--- a/packages/kit/src/utils/escape.js
+++ b/packages/kit/src/utils/escape.js
@@ -59,8 +59,7 @@ function escape(str, dict, unicode_encoder) {
 
 /** @type {Record<string, string>} */
 const escape_html_attr_dict = {
-	'<': '&lt;',
-	'>': '&gt;',
+	'&': '&amp;',
 	'"': '&quot;'
 };
 
--- a/packages/kit/src/utils/escape.js
+++ b/packages/kit/src/utils/escape.js
@@ -57,19 +57,25 @@ function escape(str, dict, unicode_encoder) {
 	return result;
 }
 
-/** @type {Record<string, string>} */
+/**
+ * When inside a double-quoted attribute value, only `&` and `"` hold special meaning.
+ * @see https://html.spec.whatwg.org/multipage/parsing.html#attribute-value-(double-quoted)-state
+ * @type {Record<string, string>}
+ */
 const escape_html_attr_dict = {
 	'&': '&amp;',
 	'"': '&quot;'
 };
 
 /**
- * use for escaping string values to be used html attributes on the page
- * e.g.
- * <script data-url="here">
+ * Formats a string to be used as an attribute's value in raw HTML.
+ *
+ * It escapes unpaired surrogates (which are allowed in js strings but invalid in HTML), escapes
+ * characters that are special in attributes, and surrounds the whole string in double-quotes.
  *
  * @param {string} str
- * @returns string escaped string
+ * @returns {string} Escaped string surrounded by double-quotes.
+ * @example const html = `<tag data-value=${escape_html_attr('value')}>...</tag>`;
  */
 export function escape_html_attr(str) {
 	return '"' + escape(str, escape_html_attr_dict, (code) => `&#${code};`) + '"';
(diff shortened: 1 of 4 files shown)


benmccann:
Why does `&` need to be encoded? OWASP suggests simply encoding `"` to `&#x22;` (should we use that rather than `&quot;`?)

https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#rule-2-attribute-encode-before-inserting-untrusted-data-into-html-common-attributes

mrkishi:
> Why does `&` need to be encoded?

It's not _exactly_ a security issue in this case. `&` won't break the surrounding quotes and enable a direct xss attack, however, it _will_ change the value we'll get back from the dom.

```js
const parent = document.body;
const text = 'hello &quot;world&quot;';
parent.insertAdjacentHTML('afterbegin', `<div data-string="${text}"></div>`);
if (parent.firstChild.dataset.string !== text) {
	throw new Error('oops! lost information');
}
```

> should we use that rather than `&quot;`?

I've no idea. I assume `quot` has a higher chance of being compressible, but newer compression algorithms like brotli are pre-primed with web-specific data, and both might be in there already!

packages/kit/src/utils/escape.spec.js
```diff
@@ -0,0 +1,24 @@
...
+const attr = suite('escape_html_attr');
+
+attr('escapes special attribute characters', () => {
+	assert.equal(
+		escape_html_attr('some "values" are &special here, <others> aren\'t.'),
+		'"some &quot;values&quot; are &amp;special here, <others> aren\'t."'
+	);
+});
+
+attr('escapes invalid surrogates', () => {
```
benmccann, line 14: it might be worth a comment explaining what this is testing

packages/kit/src/utils/escape.js
```diff
@@ -59,8 +59,7 @@ function escape(str, dict, unicode_encoder) {
 
 /** @type {Record<string, string>} */
 const escape_html_attr_dict = {
-	'<': '&lt;',
-	'>': '&gt;',
+	'&': '&amp;',
```
benmccann, line 62: it's maybe worth a comment that this is for pulling the attribute out of the dom and the other one is for security

benmccann (approved):

Conduitry (approved):
This makes sense to me. Thank you!

dominikg (approved):

PR #4017, opened 2022-02-20
Fix check_port() in dev command

jonahsnider:
Fixes a bug where `check_port()` was not called in the `dev` command.

### Tests

### Changesets

### DIFF
--- a/.changeset/late-ties-relax.md
+++ b/.changeset/late-ties-relax.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Fix check_port() in dev command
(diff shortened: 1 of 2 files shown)


bluwy:
I don't think we need this here. Currently port checking is done by the Vite dev server, and I think Vite prints a message if the port is occupied?

jonahsnider:
Vite will error if the port is unavailable but the error message doesn't include as much helpful information as `check_port()`. Additionally, it's a bit weird to have the `dev` command be the only one that has a different error message for the port. IMO having a consistent error output in this case should be preferred.

bluwy:
We had a similar PR before and it was closed (can't find, am on mobile), mainly because to correctly determine when to check_port, we have to resolve the Vite config to retrieve the `strictPort` option, and resolving it before starting the server would unnecessarily slow down the initial startup time and make the code a bit out of place. I think it's fine if the message is slightly different as long as it gets the message across though.

jonahsnider:
That makes sense, I'll close this PR then.

PR #4024, opened 2022-02-20
Improve `escape_html_attr` performance

### LINKED ISSUE #4016: Utf16 handling in html attributes and text
### Describe the bug

#4014 follow up.

The `escape_html_attr()` helper currently escapes invalid utf16 code units (ie. unpaired surrogates). ~~Its sibling `escape_json_in_html()` doesn't.~~

Do we actually need to escape them? They are technically invalid according to the HTML spec, but treated as a soft-error, and apparently browsers will just treat them as unknown characters without any extra issues that I could find.

I guess this could also be affected by whatever platform we're running on, since the utf16 to utf8 conversion usually happens there (from what I could gather?). Do we know of any issues from passing invalid utf16 to any platform?

If there are no known issues, we could drop the escaping from attributes.
~~If there are, it's unlikely that they wouldn't also affect text nodes (ie. `<script>`'s contents), in that case we'd also need that on `escape_json_in_html`.~~ This is actually already the case.

If it turns out they're needed, I have a regex-based version ready that improves performance quite a bit over the character-by-character loop.

### Reproduction

n/a

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    Memory: 1.04 GB / 7.88 GB
  Binaries:
    Node: 17.4.0 - ~\scoop\apps\nodejs\current\node.EXE
  Browsers:
    Chrome: 98.0.4758.102
    Edge: Spartan (44.19041.1266.0), Chromium (98.0.1108.50)
  npmPackages:
    @sveltejs/eslint-config: github:sveltejs/eslint-config#v5.8.0 => 5.8.0
```

### Severity

annoyance

### Additional Information

_No response_

mrkishi:
Closes #4016.
Depends on #4015.

Replaces the character-by-character loop with a regex-based `str.replace` call.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/utils/escape.js
+++ b/packages/kit/src/utils/escape.js
@@ -25,39 +25,6 @@ export function escape_json_in_html(val) {
 }
 
 /**
- * @param str {string} string to escape
- * @param dict {Record<string, string>} dictionary of character replacements
- * @param unicode_encoder {function(number): string} encoder to use for high unicode characters
- * @returns {string}
- */
-function escape(str, dict, unicode_encoder) {
-	let result = '';
-
-	for (let i = 0; i < str.length; i += 1) {
-		const char = str.charAt(i);
-		const code = char.charCodeAt(0);
-
-		if (char in dict) {
-			result += dict[char];
-		} else if (code >= 0xd800 && code <= 0xdfff) {
-			const next = str.charCodeAt(i + 1);
-
-			// If this is the beginning of a [high, low] surrogate pair,
-			// add the next two characters, otherwise escape
-			if (code <= 0xdbff && next >= 0xdc00 && next <= 0xdfff) {
-				result += char + str[++i];
-			} else {
-				result += unicode_encoder(code);
-			}
-		} else {
-			result += char;
-		}
-	}
-
-	return result;
-}
-
-/**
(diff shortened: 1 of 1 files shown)


packages/kit/src/utils/escape.js
```diff
@@ -24,54 +24,43 @@ export function escape_json_in_html(val) {
...
-
 /** @type {Record<string, string>} */
 const escape_html_attr_dict = {
-	'<': '&lt;',
-	'>': '&gt;',
+	'&': '&amp;',
 	'"': '&quot;'
 };
 
+const escape_html_attr_regex = new RegExp(
```
benmccann, line 33: can we add a comment about why this is needed?

PH4NTOMiki: Why delete < and >
they break HTML if not escaped?

PH4NTOMiki: and we fixed it with #3798 and #3804

mrkishi: @PH4NTOMiki That was about `escape_json_in_html`, this is about `escape_html_attr`.

@benmccann Sorry about the comments, the previous code didn't have them and I thought they weren't necessary. I'll add some.

Rich-Harris (approved):

Issue #2576, opened 2021-10-09
Add typescript version of svelte.config.js


parischap:
### Describe the problem

In the FAQ, it is said I can add a middleware in dev mode by creating a vite plugin.

But my middleware is a typescript  file and I cannot import it into my svelte.config.js.

I tried to rename svelte.config.js to svelte.config.ts but got an error "Cannot find module 'C:\Users\JEROME\Documents\MesDeveloppements\SitesInternet\tarteen\svelte.config.cjs'".

### Describe the proposed solution

Allow svelte config file to be written in typescript.

### Alternatives considered

Don't see any

### Importance

would make my life easier

Kapsonfire-DE:
Need to push this - its really an issue

martinjeromelouis:
Thanks for pushing.

Actually I found out this is even more complex than I thought. All that runs in the context of the svelte.config.js file is not "visible" from your code running in the sveltekit context. For instance, if you need to initialize some stuff for your middleware, e.g. a database for the express-session middleware, or a logger, or whatever, there is no way (or I didn't find any) you can pass these initialized variables to your hook.ts. Whatever module you import in svelte.config.js will be loaded and executed again if you import it again in hook.ts.

Furthermore, I also didn't find a way to retrieve in hooks.ts anything sent by a middleware. For instance, the cookie-session middleware will set a req.session object. But I don't see it on the request object I get from the handle function in hooks.ts. I also tried to create a middleware that sets a req.locals.dummy variable. I do get a request.locals in the handle function in hooks.ts but it's empty.

So if you pass middlewares through svelte.config.js, you have two separate codes executing : one in the pre VITE context and another one in the VITE context. And there seems to be no communication between the two.

Finally, the VITE_ENV variables are not available in svelte.config.js (because VITE is not started yet). So you have to load them again (for instance for your database keys,...).

For the moment, I have given up trying. I will try to make do without the usual express middleware.

I think all would be much easier if there were a file called main.js/ts loaded by sveltekit at server start. And the same client side.

Would really appreciate any help.

dummdidumm:
This does not sound related to wanting TypeScript support for the `svelte.config.js`. Furthermore, `svelte.config.js` is reserved for configuring what to run, initialization of middlewares/databases etc should go into `hooks.ts`, #1538 has more details.

martinjeromelouis:
I don't understand.

[...]

Issue #1919, opened 2021-07-16
svelte.config.ts – Svelte Config TS Support


ScarletFlash:
### Describe the problem

Overview of available config fields currently is only available from [docs](https://kit.svelte.dev/docs#configuration).

### Describe the proposed solution

It would be great if we had build-in Interface, describes all available fields for config file.
Moreover, it would be great if we had build-in support for `.ts`-file extension for config files.

### Alternatives considered

–

### Importance

would make my life easier

dummdidumm:
This interface already exists, and the JSDoc at the top of the file which comes with the starter tepmlates should enable the type hints: `/** @type {import('@sveltejs/kit').Config} */` . Adding `.ts` support for the config files is out of scope since it would require SvelteKit to also install TypeScript or some other form of TS-JS-transpiler for everyone, even if they don't use TS, and it adds additional maintenance overhead for a file that should be pretty simple anyway.

michael0liver:
I was also searching for this feature because it is [supported](https://vitejs.dev/config/#config-intellisense) with vite:
> Vite also directly supports TS config files. You can use `vite.config.ts` with the defineConfig helper as well.

Not essential but its nice to have, especially when you have local custom svelte preprocessors and vite plugins written in TS.

PR #4031, opened 2022-02-21
Support `svelte.config.ts`

tommywalkie:
This is a continuation of https://github.com/sveltejs/kit/issues/1919 and most likely https://github.com/sveltejs/kit/issues/2576, making use of Vite's `loadConfigFromFile` to handle TypeScript config file. For now this was only tested on a local SvelteKit project.

As mentionned in most threads, the remaining issue/quirk is making it work with other Svelte tools (like the VSCode plugin, which is expecting a `svelte.config.js` file).

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/config/index.js
+++ b/packages/kit/src/core/config/index.js
@@ -29,14 +29,29 @@ export function load_template(cwd, config) {
 
 export async function load_config({ cwd = process.cwd() } = {}) {
 	const config_file = path.join(cwd, 'svelte.config.js');
+	const ts_config_file = path.join(cwd, 'svelte.config.ts');
+	let config;
 
-	if (!fs.existsSync(config_file)) {
-		throw new Error(
-			'You need to create a svelte.config.js file. See https://kit.svelte.dev/docs/configuration'
-		);
+	if (fs.existsSync(config_file)) {
+		config = await import(url.pathToFileURL(config_file).href);
+	} else {
+		if (fs.existsSync(ts_config_file)) {
+			const { loadConfigFromFile } = await import('vite');
+			const resolved_config = await loadConfigFromFile(
+				{
+					command: 'build',
+					mode: process.env.NODE_ENV || 'production'
+				},
+				ts_config_file,
+				cwd
+			);
+			config = { default: resolved_config?.config };
+		} else {
+			throw new Error(
+				'You need to create a svelte.config.js file. See https://kit.svelte.dev/docs/configuration'
+			);
+		}
 	}
-
-	const config = await import(url.pathToFileURL(config_file).href);
 
 	const validated = validate_config(config.default);
 
--- a/packages/kit/src/core/config/index.js
+++ b/packages/kit/src/core/config/index.js
@@ -27,30 +27,31 @@ export function load_template(cwd, config) {
 	return fs.readFileSync(template, 'utf-8');
 }
 
-export async function load_config({ cwd = process.cwd() } = {}) {
+/**
+ * @param {{
+ *   cwd?: string;
+ *   mode?: string;
+ * }} opts
+ */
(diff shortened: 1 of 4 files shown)


dummdidumm:
I don't know how I feel about this. It feels wrong to support this half-way. There's not much value if you can use this in your build but your IDE/other tooling can't deal with it. This doesn't mean I'm not open to changing this on the tooling-side of things btw.

tommywalkie:
@dummdidumm Yea pretty much. It does run the config file fine, then you inevitably find out the language server highlights your codebase in red. If anything, this could help in the future.

packages/kit/src/core/config/index.js
```diff
@@ -29,15 +29,30 @@ export function load_template(cwd, config) {
...
+	let config;
 
-	if (!fs.existsSync(config_file)) {
-		throw new Error(
-			'You need to create a svelte.config.js file. See https://kit.svelte.dev/docs/configuration'
-		);
+	if (fs.existsSync(config_file)) {
+		config = await import(url.pathToFileURL(config_file).href);
+	} else {
+		if (fs.existsSync(ts_config_file)) {
```
benmccann, line 38: instead of having the `if` inside the `else` you could make it an `else if`

packages/kit/src/core/config/index.js
```diff
@@ -29,15 +29,30 @@ export function load_template(cwd, config) {
...
+			const { loadConfigFromFile } = await import('vite');
+			const resolved_config = await loadConfigFromFile(
+				{
+					command: 'build',
+					mode: process.env.NODE_ENV || 'production'
```
benmccann, line 43: the `|| 'production'` shouldn't be needed because it is done here: https://github.com/sveltejs/kit/blob/a8a3b14c4420a135088ef0f5e263fd24e88f0cd6/packages/kit/src/cli.js#L91

i'm worried that this might be putting it into `production` mode when you run `svelte-kit dev`

benmccann:
It seems somewhat reasonable to me. I'm not sure we should assume everyone uses VS Code. I'm a bit more split on whether we'd want to advertise it in the docs or not

packages/kit/src/core/config/index.js
tommywalkie: Maybe this command need a `|| 'production'` (?)

https://github.com/sveltejs/kit/blob/a8a3b14c4420a135088ef0f5e263fd24e88f0cd6/packages/kit/src/cli.js#L147-L153
 
Otherwise, I considered using `development` as the default value for `NODE_ENV`.

benmccann: yeah, it seems reasonable to add a `production` default to the `package` command

Rich-Harris:
I'm a strong 👎 on this. It adds moving parts to something that should be extremely simple and foolproof, and does so using an undocumented (and thus liable to change) API that's designed to load a _Vite_ config, not a _SvelteKit_ config (even if it happens to work today, is it guaranteed to work in future, despite whatever changes — to either Vite or Kit — we haven't anticipated?). As noted above, it will also cause problems with other tooling.

You can already type your config files (as shown throughout the docs). The use case presented in #2576 isn't something we should encourage in the first place (I didn't actually realise that we're promoting it in the FAQ; I should pay more attention to that page). The cost to benefit ratio is all wrong.

bluwy:
-1 for me too. This wouldn't work for `vite-plugin-svelte` and I think it will be vital when we need to configure it in a SvelteKit project. I think we should wait for https://github.com/sveltejs/rfcs/pull/59 first and slowly migrate the tooling towards it before making a new change.

Rich-Harris:
I'm going to close this as the consensus seems to be that it's not a change we should make. Perhaps we can address it at an ecosystem level eventually per the mentioned RFC. Thanks

PR #4034, opened 2022-02-21
rename `app.render` to `server.respond`

### LINKED ISSUE #4030: Rename `App`
### Describe the problem

Internally, we have two things called `App` — the class that is created by `svelte-kit build`, and instantiated by adapters...

```js
import { App } from 'APP';
import { manifest } from 'MANIFEST';

const app = new App(manifest);

// later
const response = await app.render(request);
```

...and the namespace for specifying app-level interfaces like `locals`:

```ts
declare namespace App {
  interface Locals {}
  interface Platform {}
  interface Session {}
  interface Stuff {}
}
```

This hasn't really been a problem, but now that we're adding docs for the types (#4011), it's going to become a source of confusion and broken links.

I think it makes more sense to rename the class than the namespace. The best suggestion so far is `Server`:

```js
// created in build_server.js
export class Server {
  constructor(manifest) {
    // ...
  }
}
```

```js
// created by an adapter
import { Server } from 'SERVER';
import { manifest } from 'MANIFEST';

// later
const response = await server.render(request);
```

At the risk of inviting bikeshedding, it might also make sense to change `server.render` to `server.respond` or `server.handle` or something, since 'render' sort of implies something side-effect free (which PUT/POST/DELETE requests aren't), and also implies that we're doing SSR when in some cases we're just rendering an empty shell page.

Aside: I've been meaning to overhaul `Router` and `Renderer` — having two separate classes seemed like a good idea at the time, but in practice it hasn't worked out, and it would make sense to unify them into a single class. There'd be a nice symmetry if we had a `class Server` and a `class Client`.

### Alternatives considered

We could rename the namespace instead, though it's less clear what we could rename it _to_, and it would be a breaking change for more people (all Kit users rather than just adapter authors).

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
Closes #4030 — `class App` becomes `class Server`

### Tests

### Changesets

### DIFF
--- a/.changeset/late-bobcats-brake.md
+++ b/.changeset/late-bobcats-brake.md
@@ -7,4 +7,4 @@
 '@sveltejs/kit': patch
 ---
 
-Rename app.render to server.respond
+[breaking] rename `app.render` to `server.respond`
--- a/.changeset/late-bobcats-brake.md
+++ b/.changeset/late-bobcats-brake.md
@@ -1,0 +1,10 @@
+---
+'@sveltejs/adapter-cloudflare': patch
+'@sveltejs/adapter-cloudflare-workers': patch
+'@sveltejs/adapter-netlify': patch
+'@sveltejs/adapter-node': patch
+'@sveltejs/adapter-vercel': patch
+'@sveltejs/kit': patch
+---
+
+Rename app.render to server.respond
--- a/packages/adapter-cloudflare-workers/files/entry.d.ts
+++ b/packages/adapter-cloudflare-workers/files/entry.d.ts
@@ -1,5 +1,5 @@
-declare module 'APP' {
-	export { App } from '@sveltejs/kit';
+declare module 'SERVER' {
+	export { Server } from '@sveltejs/kit';
 }
 
 declare module 'MANIFEST' {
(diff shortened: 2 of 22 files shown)


.changeset/late-bobcats-brake.md
```diff
@@ -0,0 +1,10 @@
+---
+'@sveltejs/adapter-cloudflare': patch
+'@sveltejs/adapter-cloudflare-workers': patch
+'@sveltejs/adapter-netlify': patch
+'@sveltejs/adapter-node': patch
+'@sveltejs/adapter-vercel': patch
+'@sveltejs/kit': patch
+---
+
+Rename app.render to server.respond
```
benmccann, line 10:
```suggestion
[breaking] rename `app.render` to `server.respond`
```

benmccann (approved):

benmccann:
CC'ing the community adapter authors: @jthegedus @pluvial @geoffrich @MacFJA @HalfdanJ @ryanbethel

As part of putting up our [new types documention](https://kit.svelte.dev/docs/types) we renamed the `App.render` to `Server.respond` to avoid conflicting with [the `App` namespace`](https://kit.svelte.dev/docs/types#the-app-namespace), so a small update will be needed to any adapter

PR #4043, opened 2022-02-22
[chore] refactor out Prefetcher

benmccann:
Mostly this just helps organize the code a bit better. Though theoretically we could also offer a build-time option to disable prefetching that would save these bytes from being imported and used


Rich-Harris:
I was actually planning to go in the opposite direction — combining `Router` and `Renderer` in a single `Client` class

benmccann:
I'm not against that idea since the `Router`/`Renderer` separation isn't very clean. I'd like to take a stab at seeing if we can fix that, but I'm not sure if we can and think you might be right that combining them would be the best choice. However, that seems like a separate question to me. The reason having them separate isn't working because they're too tightly coupled. But that doesn't really apply to prefetching which can pretty easily be separated

benmccann:
It didn't really use the router at all. It just used a couple util methods that were defined there. I moved those to `utils.js` to make it clearer

Rich-Harris:
`prefetcher` still has a property called `router` though! This PR adds 30 lines of code and I can't understand what we're getting in return

PR #4054, opened 2022-02-22
[fix] remove private methods because Safari sucks

benmccann:
private methods apparently don't work on Safari 14.1 according to a user report on Discord

### DIFF
--- a/.changeset/clean-islands-do.md
+++ b/.changeset/clean-islands-do.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] remove private methods to make Safari 14.1 work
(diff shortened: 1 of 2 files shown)


Rich-Harris (approved):

ksmoore17:
Thank you. Not the discord user, but I was getting the same or similar error from my out of date safari browser:
`SyntaxError: Unexpected private name #update_scroll_positions. Cannot parse class method with private name.`

Leaving this comment because this issue was hard to find.

Changing from `"@sveltejs/kit": "next"` to `"@sveltejs/kit": "^1.0.0-next.282"` in `package.json` and `npm install` fixed.

Issue #19, opened 2020-10-15
Template tests


Rich-Harris:
If the app template doesn't include tests, then one of the first questions people will ask is 'how do we add tests'? We should probably have *some* way of answering this question

benmccann:
Jest just merged support for async transforms, which means pre-processing will now be supported with the next release of Jest

dominikg:
I'm not entirely sure i understand what this issue is about but some thoughts crossed my mind that i'd like to share in hopes they will be helpful

1) playwright-core without bundled browser
to reduce download size for integration tests with playwright, you can use playwright-core and an existing chrome installation passed to it via CHROME_BIN env or by guessing the path https://github.com/svitejs/svite/blob/3a9e1a069450cd3fa009716192dc645802e5370e/scripts/jestGlobalSetup.js#L35

2) use a browser server
chromium.launchServer creates a service that can be reused across tests so they don't have to start/stop chrome all over again https://github.com/svitejs/svite/blob/3a9e1a069450cd3fa009716192dc645802e5370e/scripts/jestGlobalSetup.js#L59

3) compile / transform with the tools we already have available
Would it be possible to levarage a vite dev server instance for transforms via https://github.com/vitejs/vite/blob/26c46b92b2b5b21e21380459d1d1a76cb929a69a/packages/vite/src/node/server/index.ts#L192
Or maybe a special vite-plugin-svelte instance that works standalone?

Rich-Harris:
This could be useful: https://github.com/microsoft/playwright-test

kenkunz:
(Sorry… I just realized I was auto-commenting on this ticket from my forked-repo commits. Ignore the first two above – I overwrote my history thinking it wouldn't be viewed by anyone else.)

[...]

PR #4056, opened 2022-02-22
add option to create tests with Playwright

Rich-Harris:
Thought I'd tackle one of the earliest issues — #19. This adds a 'Add Playwright for testing?' prompt when you create a new project:

image

After that, `npm test` runs `test/test.js` with Playwright.

Aside: we had to create our own test fixtures for Kit's own tests, notably `clicknav`, which allows us to wait for client-side routing to happen following a click. Am reluctantly coming round to the view that my position in https://github.com/sveltejs/kit/pull/2917#issuecomment-979419299 and #3241 was misguided; in order to use things like `page.click` and `waitForNavigation` idiomatically, we need to wait until navigation has occurred before doing `pushState`. This isn't just the tail wagging the dog — premature `pushState` also causes a subtle bug where relative URLs are briefly incorrect.

### Tests

### Changesets


benmccann (approved):
just fyi, `lint` is failing

this seems like a good idea to me. the big questions I had about doing this were:
- the navigation stuff (which you already mentioned. though I wonder if that should be done before this)
- whether we'd also want to setup vitest or something similar for unit testing
- whether it makes any sense to test components outside of a web browser in some scenarios (I legitimately don't know the answer to this as I've done limited frontend testing)

packages/create-svelte/bin.js
```diff
@@ -92,6 +92,14 @@ async function main() {
 					initial: false,
 					active: 'Yes',
 					inactive: 'No'
+				},
+				{
+					type: 'toggle',
+					name: 'playwright',
+					message: 'Add Playwright for testing?',
```
benmccann, line 99: Maybe we should expand this message in case we offer Vitest or similar in the future alongside Playwright?
```suggestion
					message: 'Add Playwright for end-to-end browser testing?',
```

Rich-Harris:
> the navigation stuff (which you already mentioned. though I wonder if that should be done before this)

probably. we can open a separate PR and leave this open till then

> whether we'd also want to setup vitest or something similar for unit testing

will confess i have no idea what this looks like in practice. have people set this up in kit apps? any pitfalls? either way i tend to think unit tests are somewhat less important, so we can tackle that later

> whether it makes any sense to test components outside of a web browser in some scenarios

FWIW it's fairly straightforward to test SSR and endpoints with Playwright. Though I wonder if we should have JS and JS-less projects in the playwright.config.js by default, the same as we do in Kit's own tests

packages/create-svelte/bin.js
```diff
@@ -92,6 +92,14 @@ async function main() {
 					initial: false,
 					active: 'Yes',
 					inactive: 'No'
+				},
+				{
+					type: 'toggle',
+					name: 'playwright',
+					message: 'Add Playwright for end-to-end browser testing?',
```
Rich-Harris, line 99: actually on second thoughts i don't think we should call this end-to-end testing @benmccann. integration testing is more accurate, but probably easier to just do this

```suggestion
					message: 'Add Playwright for browser testing?',
```

benmccann: sure. happy to avoid the debate over what the correct term is :smile:

packages/create-svelte/shared/+playwright+default/tests/test.js
```diff
@@ -0,0 +1,6 @@
+import { expect, test } from '@playwright/test';
+
+test('about page has expected h1', async ({ page }) => {
```
dummdidumm, line 3: Does playwrigt support writing your code in TypeScript? If yes, should we add a variant for TS users?

bluwy: Yeah, playwright does [support typescript](https://playwright.dev/docs/test-typescript) ootb, down to the playwright config too.

Rich-Harris:
Let's do the TS stuff in a follow-up

randyridge:
Hi there, I just tried this out, with the build and preview command in playwright.config.js I found that my build runs longer than the default playwright test timeout, I ended up having to move the build portion to my package.json test script "pnpm build && playwrite test".

PR #4059, opened 2022-02-22
[feat] allow returning strings from shadow endpoints

benmccann:
I had some JSON serialized to disk that I was trying to return from a shadow endpoint and was surprised that I couldn't. The restriction was only documented under the "standalone endpoints" section, so I also didn't see it there for a few days until I happened across it later. Removing the unnecessary restriction should hopefully be more intuitive

### DIFF
--- a/.changeset/nasty-poems-fix.md
+++ b/.changeset/nasty-poems-fix.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[feat] allow returning strings from shadow endpoints
--- a/documentation/docs/01-routing.md
+++ b/documentation/docs/01-routing.md
@@ -265,7 +265,7 @@ export default config;
 
 ### Standalone endpoints
 
-Most commonly, endpoints exist to provide data to the page with which they're paired. They can, however, exist separately from pages. Standalone endpoints have slightly more flexibility over the returned `body` type — in addition to objects, they can return a string or a `Uint8Array`.
+Most commonly, endpoints exist to provide data to the page with which they're paired. They can, however, exist separately from pages. Standalone endpoints have slightly more flexibility over the returned `body` type — in addition to objects, they can return a `Uint8Array`.
 
 > Support for streaming request and response bodies is [coming soon](https://github.com/sveltejs/kit/issues/3419).
 
--- a/packages/kit/src/runtime/server/endpoint.js
+++ b/packages/kit/src/runtime/server/endpoint.js
@@ -1,17 +1,12 @@
 import { to_headers } from '../../utils/http.js';
 import { hash } from '../hash.js';
-import { is_pojo, normalize_request_method } from './utils.js';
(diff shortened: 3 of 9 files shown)


Rich-Harris:
This feels like a documentation issue to me — the `JSON.parse` should happen in the endpoint, I think

benmccann:
There doesn't necessarily have to be a `JSON.parse`. E.g. if you're sending a `string` via a client-side navigation, there's no reason to parse it to a JSON object and then back to a string again.

I also think if we can just make it work it's better than making people be aware of a restriction in the documentation that they might not have seen

Conduitry:
What's the proposed change in behavior in this PR? That a `body` that's a string should be assumed to already be a stringified representation of a JSON object? I'm not sure how I feel about that. I know there's precedent for that sort of thing (like in the replace Rollup plugin, a string replacement happens literally in the JS code but all other object types are `JSON.stringify`ed first before the replacement happens) but that still feels ugly to me.

benmccann:
The behavior change is that an API endpoint should be allowed to return a string. I'm honestly surprised this is controversial. All data must be serialized to a string, so why not allow it to be returned directly? It will only be assumed to be JSON when the `accept` header specifies JSON

Rich-Harris:
*Standalone* endpoints can return a string. That's always been possible. But it strikes me as very weird that POSTing to this page would yield the props `{a: 1, b: 2}`:

```js
export function get() {
  return {
    body: '{"a":1}'
  };
}

export function post() {
  return {
    body: '{"b":2}'
  };
}
```

In any case, if we were to make it behave the same as a standalone endpoint you'd need to specify the right content type...

```diff
export function get() {
  return {
+   headers: {
+     'content-type': 'application/json',
+   }
    body: '{"a":1}'
  };
}

export function post() {
  return {
+   headers: {
+     'content-type': 'application/json',
+   }
    body: '{"b":2}'
  };
}
```

...since otherwise it will just return that JSON as `text/plain`. (`accept` headers tell the server what responses are acceptable, they don't tell it 'return whatever you were planning to return, but add this content type header'.) But it would be absurd to do that instead of this:

```diff
export function get() {
  return {
-    body: '{"a":1}'
+    body: JSON.parse('{"a":1}')
  };
}

export function post() {
  return {
-    body: '{"b":2}'
+    body: JSON.parse('{"b":2}')
  };
}
```

Page endpoints deal in POJOs. I think it's much easier and more understandable all round if we stick to that.

benmccann:
I'm going to close this. It's not a priority for me

> But it strikes me as very weird that POSTing to this page would yield the props {a: 1, b: 2}:

That would strike me as something very weird as well and if it was happening with this PR it certainly was not intentional.

> if we were to make it behave the same as a standalone endpoint you'd need to specify the right content type...

Shouldn't that be happening automatically? If you want to hit a shadow endpoint from outside the application I'd expect it to have the correct headers

Issue #3974, opened 2022-02-17
Unable to use `index.svelte` with a shadow endpoint `index.ts`


kdheepak:
### Describe the bug

I get the following error on internal routing to the root url:

```
404

Failed to load data

_load@https://svelte-kit-test.kdheepak.com/_app/start-f0f2a705.js:1:22348
async*_get_navigation_result@https://svelte-kit-test.kdheepak.com/_app/start-f0f2a705.js:1:19338
update@https://svelte-kit-test.kdheepak.com/_app/start-f0f2a705.js:1:17196
handle_navigation@https://svelte-kit-test.kdheepak.com/_app/start-f0f2a705.js:1:17109
_navigate@https://svelte-kit-test.kdheepak.com/_app/start-f0f2a705.js:1:12898
init_listeners/<@https://svelte-kit-test.kdheepak.com/_app/start-f0f2a705.js:1:10324
EventListener.handleEvent*init_listeners@https://svelte-kit-test.kdheepak.com/_app/start-f0f2a705.js:1:9696
it@https://svelte-kit-test.kdheepak.com/_app/start-f0f2a705.js:1:24041
async*@https://svelte-kit-test.kdheepak.com/:28:8
```

And I get the following in the console:

```
https://svelte-kit-test.kdheepak.com/__data.json
```

<img width="1440" alt="Screen Shot 2022-02-17 at 1 20 44 AM" src="https://user-images.githubusercontent.com/1813121/154434446-ebca3894-0cc3-4d1f-bda3-e809ce6ed8fa.png">

### Reproduction

Reproducible using https://github.com/kdheepak/svelte-kit-test and adapter-static, on github pages.

1) Go to <https://svelte-kit-test.kdheepak.com/>
2) Click on link that takes you to <https://svelte-kit-test.kdheepak.com/link>
3) Click on link that takes you to <https://svelte-kit-test.kdheepak.com/>

### Logs

```shell
_load@https://svelte-kit-test.kdheepak.com/_app/start-f0f2a705.js:1:22348
async*_get_navigation_result@https://svelte-kit-test.kdheepak.com/_app/start-f0f2a705.js:1:19338
update@https://svelte-kit-test.kdheepak.com/_app/start-f0f2a705.js:1:17196
handle_navigation@https://svelte-kit-test.kdheepak.com/_app/start-f0f2a705.js:1:17109
_navigate@https://svelte-kit-test.kdheepak.com/_app/start-f0f2a705.js:1:12898
[...]

PR #4066, opened 2022-02-23
[fix] create missing root's `__data.json`

baseballyama:
fix: https://github.com/sveltejs/kit/issues/3974

The cause was that root's `__data.json` had not been created.
It seems to be a SPA-specific issue, so I created a test in adapter-static.

--- 

### Tests

### Changesets


packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -346,7 +346,7 @@ export async function load_node({
 
 	// generate __data.json files when prerendering
 	if (shadow.body && state.prerender) {
-		const pathname = `${event.url.pathname}/__data.json`;
+		const pathname = `${event.url.pathname.replace(/\/$/, '')}/__data.json`;
```
Conduitry, line 349: Are there other paths besides literally `'/'` that might end with a `/` that we have to worry about here? Can we check for the literal string `'/'` here which would (presumably) be slightly less computationally intensive?

Rich-Harris: yeah, this is also fixing the `trailingSlash: 'always'` case

Rich-Harris (approved):

packages/kit/src/runtime/server/page/load_node.js
Conduitry: Gotcha, okay, that makes me sufficiently disinterested in microoptimizations, then,

baseballyama:
Ah yes. Thank you for simplifying the test!

Rich-Harris:
Thank you for the fix!

Issue #4045, opened 2022-02-22
mailto: links open twice


ljani:
### Describe the bug

Clicking a [`mailto:test@example.com`](mailto:test@example.com) link on Firefox 98.0b7 seems to open Outlook twice. Other sites (such as the link in previous sentence) are fine.

Chromium seems to be fine.

### Reproduction

```sh
git clone https://github.com/ljani/sveltekit-mailto-issue
cd sveltekit-mailto-issue
npm install
npm run dev
```

### System Info

```shell
  System:
    OS: Windows 10 10.0.22000
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 19.85 GB / 31.72 GB
  Binaries:
    Node: 16.13.2 - C:\Program Files\nodejs\node.EXE
    npm: 8.1.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (98.0.1108.55)
    Internet Explorer: 11.0.22000.120
  npmPackages:
    @sveltejs/adapter-auto: ^1.0.0-next.24 => 1.0.0-next.24
    @sveltejs/kit: ^1.0.0-next.281 => 1.0.0-next.281
    svelte: ^3.46.4 => 3.46.4
```

### Severity

annoyance

### Additional Information

This used to work fine in an earlier version of sveltekit, I might have time to look it up later.

EDIT: Bumped dependencies to the latest versions.

ljani:
`@sveltejs/kit@1.0.0-next.225` works
`@sveltejs/kit@1.0.0-next.226` does not exist
`@sveltejs/kit@1.0.0-next.227` is broken

PR #4072, opened 2022-02-23
[fix] Ignore click event if href is mailto link

baseballyama:
fix: https://github.com/sveltejs/kit/issues/4045

I didn't know how to test this, so I didn't implement a test. Instead, I put a comment in the code.
If you have any ideas for the test, please let me know!

### Tests

### Changesets

### DIFF
--- a/.changeset/tiny-badgers-love.md
+++ b/.changeset/tiny-badgers-love.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Ignore click event if href is mailto link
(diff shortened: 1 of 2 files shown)


Conduitry:
Should we check for other custom URL schemes here as well? Is there a similar problem with, say, `tel:` links? Can we do this without hard coding specific schemes? It seems like any URL beginning with `foo:` should not be handled by the router and should be left to the browser.

baseballyama:
@Conduitry 

Ahh, this is really good point.

I think we can divert `URL` object.
Such `mailto:` or `tel:` doesn't have origin, so we can easily judge it by `url.origin`.
```js
if (!is_svg_a_element && url.origin === 'null') return;
```

What do you think about it?

Rich-Harris (approved):

Rich-Harris:
yeah, that looks like a good solution to me. thanks!

PR #4076, opened 2022-02-23
independent page type

### LINKED ISSUE #3730: Store types
### Describe the problem

Previously it was possible to import the types of the various SvelteKit stores from , such as `$page`. This was useful for creating helper functions which relied on stores, like so:

```ts
export function getReturnState(
  page: Page,
  returnState: Record<string, unknown> = {}
): Record<string, unknown> {
  return {
    returnTo: `${page.path}${page.query}`,
    returnState,
  };
}
```

Ideally a way to access an interface that matches the stores. There doesn't seem to be one on the latest SvelteKit (`next-260`).

### Alternatives considered

We could do `typeof $page` possibly, but this doesn't work for plain `.ts` file since we can't use `$` syntax there, and `typeof page` is a readable store which is not what we want.

### Importance

would make my life easier

### Additional Information

_No response_

ignatiusmb:
Closes #3730 

### Tests

### Changesets


Rich-Harris:
I guess we'll need to figure out a place to document this. Should it go in https://kit.svelte.dev/docs/modules#$app-stores or do we need to rig up a new section in https://kit.svelte.dev/docs/types?

Also, do we need the `export` keywords here? If an unexported `interface` is importable, should it just be `const page` etc rather than `export const page`?

packages/kit/types/ambient.d.ts
```diff
@@ -119,16 +119,18 @@ declare module '$app/stores' {
 		session: Writable<App.Session>;
 		updated: typeof updated;
 	};
-	/**
-	 * A readable store whose value contains page data.
-	 */
-	export const page: Readable<{
+
+	interface Page<Params extends Record<string, string> = Record<string, string>> {
```
benmccann, line 123: if this is going to close https://github.com/sveltejs/kit/issues/3730 then I think this interface would need to be exported?

Rich-Harris:
In #4099, we're moving `Navigation` (formerly 'Navigating') to the `@sveltejs/kit` types, and I think it makes sense to do the same thing here, since these are fairly universal concepts even if they're currently only used inside `$app/stores`. At the very least it makes documenting things somewhat easier

Rich-Harris (approved):

PR #4084, opened 2022-02-23
handle doc 404s

Rich-Harris:
Received a report that a link to `/docs/typescript` was 500ing rather than 404ing. This fixes it. An alternative/additional fix would be to set up redirects every time we make a change to the structure of the docs, but I think that would be a bad use of time at least until 1.0

### Tests

### Changesets


Rich-Harris:
Gah, it doesn't fix it. If you try to hit a non-prerendered page, it will fall back to the serverless function, which attempts to read a non-existent file from the filesystem. Might have to do a gross workaround

Conduitry:
There was an issue opened for this recently, which I was having trouble finding. Turns out because it was opened in the Svelte repo. I've moved it here: #4086.

Rich-Harris:
So... we could certainly work around this, but it wouldn't be that pretty, and it would just be masking the real issue which is that the endpoint shouldn't be deployed _anyway_.

I think the solution is to add `export const prerendered` to endpoints, and exclude prerendered endpoints from the generated route manifests.

benmccann:
Guess I should have read the comments before approving :grin: 

Are there any non-prerendered pages on the SvelteKit site? Maybe you could use `adapter-static` with the `fallback` option?

Or you might be able to do something like:
```
	if (process.env.NODE_ENV === 'production' && !$prerendering) {
		return { status: 404 };
	}
```

Rich-Harris:
adapter-static is an option, but only because we happen not to have any dynamic pages _yet_ — i'd rather fix the underlying issue. deploying a 'real' app also means we get custom error pages instead of the minimalist Vercel black-on-white status code

Rich-Harris:
have gone with the temporary workaround since prerendered endpoints might be a bit of a rabbit hole (e.g. i think we would want to do https://github.com/sveltejs/kit/issues/4112 first)

benmccann (approved):

Rich-Harris:
gah, still 500ing https://kit-svelte-dev-h9zzmhvx6-svelte.vercel.app/docs/foo

Rich-Harris:
It should be loading /docs.json from prerendered files, but isn't 🤔

sites/kit.svelte.dev/src/routes/docs/[slug].js
```diff
@@ -0,0 +1,28 @@
+import { mode, prerendering } from '$app/env';
+import { read } from '$lib/docs/server';
+
+/** @type {import('@sveltejs/kit').RequestHandler} */
+export async function get({ params }) {
+	// TODO https://github.com/sveltejs/kit/issues/4093
+	if (!prerendering && mode === 'production') {
```
Comment (Conduitry) [inline sites/kit.svelte.dev/src/routes/docs/[slug].js:7]: Is there a reason this is checking for `mode === 'production'` rather than `!dev`?

Comment (Rich-Harris) [reply sites/kit.svelte.dev/src/routes/docs/[slug].js]: yes, me forgetting that such a thing existed

Rich-Harris:
closing this in favour of #4093

PR #4096, opened 2022-02-23
[fix] Use headers.raw() on `split_headers`

### LINKED ISSUE #4087: Netlify Adapter splits set-cookie headers at commas, disrupting the expires parameter
### Describe the bug

Specifying the `expires` attribute when setting a cookie on an endpoint causes the Netlify Adapter to emit two `set-cookie` headers (neither with expiry) instead of one single header with the designated expiry. More specifically, a header like this:

```
set-cookie: flavor=chocolate%20chip; Expires=Wed, 23 Feb 2022 17:17:53 GMT
```

Gets converted into this:

```
set-cookie: flavor=chocolate%20chip; Expires=Wed
set-cookie: 23 Feb 2022 17:17:53 GMT
```

I suspect that the [Netlify handler code](https://github.com/sveltejs/kit/blob/master/packages/adapter-netlify/src/handler.js#L76) is too simple, as it splits the `set-cookie` header at every comma. Since the date format requires a comma per the [date specification](https://httpwg.org/specs/rfc7231.html#http.date), this means expiry gets split by the adapter.

The code under suspicion:

```
headers.forEach((value, key) => {
    if (key === 'set-cookie') {
        m[key] = value.split(', ');
    } else {
        h[key] = value;
    }
});
```

### Reproduction

I have [deployed a tiny app](https://flamboyant-jackson-c3e06d.netlify.app/) on Netlify showcasing the issue. It allows you to attempt to set and unset a `flavor` cookie, and if you observe the network request/response in the developer tools, you can see the cookie is not set correctly.

* See the reproduction code on Github: https://github.com/Auroratide/sveltekit-cookie-expiry

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 11.6.1
    CPU: (8) arm64 Apple M1
    Memory: 151.03 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.5.0 - /opt/homebrew/bin/node
    Yarn: 1.22.10 - /opt/homebrew/bin/yarn
    npm: 8.4.1 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 98.0.4758.102
    Firefox: 97.0.1
    Safari: 15.1
  npmPackages:
    @sveltejs/adapter-netlify: ^1.0.0-next.46 => 1.0.0-next.46 
    @sveltejs/kit: next => 1.0.0-next.283 
    svelte: ^3.44.0 => 3.46.4
```

### Severity

### Additional Information

_No response_

JeanJPNM:

### Tests

### Changesets

Fixes #4087 by using the `headers.raw()` method from node-fetch to get the array of cookie headers.

### DIFF
--- a/.changeset/three-elephants-breathe.md
+++ b/.changeset/three-elephants-breathe.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-netlify': patch
+---
+
+Fix incorrect set-cookie header handling on adapter-netlify
--- a/packages/adapter-netlify/src/handler.js
+++ b/packages/adapter-netlify/src/handler.js
@@ -73,7 +73,8 @@ function split_headers(headers) {
 
 	headers.forEach((value, key) => {
 		if (key === 'set-cookie') {
-			m[key] = value.split(', ');
+			// @ts-expect-error (headers.raw() is non-standard)
+			m[key] = headers.raw()[key];
 		} else {
 			h[key] = value;
 		}


Auroratide:
Interesting! Didn't know about the `raw()` method, as it's specific to node-fetch.

I started on a branch of my own by adding some [testing](https://github.com/Auroratide/kit/blob/fix-set-cookie-expiry/packages/adapter-netlify/test/headers.js), as there didn't seem to be any for the adapter. Are tests like these desired for a function of this scope?

Rich-Harris (approved):

Rich-Harris:
thanks! definitely not opposed to adding some tests — our convention for unit tests is to have e.g. a `headers.spec.js` file alongside `headers.js` rather than putting them all in a central `test.js` file.

PR #4099, opened 2022-02-23
document module types

Rich-Harris:
Per [#4076](https://github.com/sveltejs/kit/pull/4076#issuecomment-1048861921), we kinda need a way to document everything exported from the `$app/*` modules. This WIP gets us part of the way there, but it means [/docs/types](https://kit-svelte-dev-git-document-module-types-svelte.vercel.app/docs/types#$app-env) and [/docs/modules](https://kit-svelte-dev-git-document-module-types-svelte.vercel.app/docs/modules#$app-env) duplicate each other, so we need to figure out the right way to handle that.


Rich-Harris:
alright, i think this is ready

Rich-Harris:
merging so we can make progress on #4021 etc

PR #4101, opened 2022-02-24
[chore] cleanup `Router` and `Renderer`

benmccann:
I know there's been some talk of combining `Router` and `Renderer` into a single class, but I think the original idea of having separate `Router` and `Renderer` implementation came from a good place and we just never implemented it very well

The benefits of having a separate `Router` and `Renderer` include being able to hydrate without loading the `Router` code
 and more possibilities for improved `Router` testing.

This removes `renderer` as a field from `Router`. By removing cycles between the two classes, it's far easier to understand what's happening and what should happen where. One of the main things needed to accomplish this is to move prefetching to the `Renderer`. It makes more sense in `Renderer` than `Router` since it almost renders the page by calling `load`, etc. whereas the `Router` is meant to be just related to navigation

Things this PR  cleans up the API quite a bit. The set of public methods on the two classes is much saner now.

### DIFF
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -174,8 +174,7 @@ export class Renderer {
 			if (!ready || !this.router) return;
 			this.session_id += 1;
 
-			const info = this.router.parse(new URL(location.href));
-			if (info) this.update(info, [], undefined, true);
+			this.invalidate(location.href);
 		});
 		ready = true;
 	}
@@ -207,7 +206,7 @@ export class Renderer {
 		addEventListener('touchstart', trigger_prefetch);
 		addEventListener('mousemove', handle_mousemove);
 		addEventListener('sveltekit:trigger_prefetch', trigger_prefetch);
-		addEventListener('hashchange', () => this.update_page_store(new URL(window.location.href)));
+		addEventListener('hashchange', () => this._update_page_store(new URL(window.location.href)));
 	}
 
 	disable_scroll_handling() {
@@ -221,6 +220,7 @@ export class Renderer {
 	}
 
 	/**
+	 * TODO: we should ensure the routes version from the server matches the version here
 	 * @param {{
 	 *   status: number;
 	 *   error: Error;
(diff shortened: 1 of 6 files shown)


Rich-Harris:
Is this intended as an alternative to combining them into a single `Client` class? I was planning to address any outstanding PRs that affect this code (like https://github.com/sveltejs/kit/pull/3629, possibly others? not sure) before touching it, so that they don't get stale

packages/kit/src/runtime/client/renderer.js
```diff
@@ -174,12 +174,41 @@ export class Renderer {
...
+				event.target?.dispatchEvent(
+					new CustomEvent('sveltekit:trigger_prefetch', { bubbles: true })
+				);
+			}, 20);
+		};
+
+		addEventListener('touchstart', trigger_prefetch);
+		addEventListener('mousemove', handle_mousemove);
+		addEventListener('sveltekit:trigger_prefetch', trigger_prefetch);
+		addEventListener('hashchange', () => this._update_page_store(new URL(window.location.href)));
```
PH4NTOMiki, line 209: This need to be in that if (`hash_navigating`) that gets inited as false, and mutated to true in click event listener if the navigation only changes hash.

PH4NTOMiki: Actually I just checked and you have duplicate for every event listener, so this is probably WIP still, or?

benmccann: Ummm. I'm pretty sure there is not a duplicate for every event listener unless I'm missing something. There are two `hashchange` listeners though, which I hadn't noticed. Perhaps they should be combined so that the handlers execute in a consistent order

PH4NTOMiki: Yeah, I saw those 2 and assumed that there are more duplicates, sorry for that, but if that's the case then I guess we can merge them, if code should be after that other one for consistency.

benmccann:
Closing in favor of https://github.com/sveltejs/kit/pull/4161 after discussion with the other maintainers

PR #4104, opened 2022-02-24
semi public types

Rich-Harris:
Deploy preview: https://kit-svelte-dev-git-semi-public-types-svelte.vercel.app/docs/types#additional-types

---

I tried to do this in #4021 but the merge was too gnarly. Easier to start fresh.

The basic thinking here is that we have public types in `index.d.ts` and `ambient.d.ts`, and private-but-documented types in `private.d.ts`. Changes to these types should be made carefully to avoid breakage.

`internal.d.ts` is for things that aren't referenced by public types, and can be changed freely.

~~So far I haven't actually moved most of the types, I've just set up the logic for extracting/rendering/linking them.~~

### DIFF
--- a/packages/kit/types/internal.d.ts
+++ b/packages/kit/types/internal.d.ts
@@ -303,7 +303,6 @@ export interface SSRState {
 
 export type StrictBody = string | Uint8Array;
 
-// TODO should this be public?
 export type ValidatedConfig = RecursiveRequired<Config>;
 
 export * from './index';
--- a/packages/kit/types/internal.d.ts
+++ b/packages/kit/types/internal.d.ts
@@ -1,16 +1,12 @@
 import { OutputAsset, OutputChunk } from 'rollup';
 import {
-	SSRManifest,
-	ValidatedConfig,
 	RequestHandler,
 	Load,
 	ExternalFetch,
 	GetSession,
 	Handle,
 	HandleError,
-	RequestEvent,
-	RequestOptions,
-	Server
+	Config
 } from './index';
 import {
 	Either,
@@ -18,8 +14,13 @@ import {
 	HttpMethod,
 	JSONObject,
 	MaybePromise,
+	RequestEvent,
+	RequestOptions,
+	ResolveOptions,
 	ResponseHeaders,
 	RouteSegment,
+	Server,
+	SSRManifest,
 	TrailingSlash
 } from './private';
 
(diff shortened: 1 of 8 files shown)


Rich-Harris:
Ok I think this is ready for review. There's some further tweaks I'd like to make to the types (particularly `EndpointOutput`) but they probably warrant a separate PR.

The big change here, other than that private types are now documented, is that a bunch of formerly public types are now in `private.d.ts`. I've been quite liberal in moving things over, on the basis that we can easily move things from `private.d.ts` to `index.d.ts`, but we can't go the other way without breaking changes. The test for whether something goes in `index.d.ts` was 'is a user likely to need to import this type?'

packages/kit/types/internal.d.ts
```diff
@@ -365,8 +303,8 @@ export interface SSRState {
 
 export type StrictBody = string | Uint8Array;
 
-type ToJSON = { toJSON(...args: any[]): Exclude<JSONValue, ToJSON> };
-
-export type TrailingSlash = 'never' | 'always' | 'ignore';
+// TODO should this be public?
```
benmccann, line 306: maybe remove this TODO now that it's no longer public?

benmccann (approved):
this looks great!

frederikhors:
@Rich-Harris, after upgrade from `1.0.0-next.285` to `1.0.0-next.286` I have an error using this code:

```svelte
<script context="module" lang="ts">
	import type { LoadInput, LoadOutput } from "@sveltejs/kit";

	export function load({ url }: LoadInput): LoadOutput {
		//...
	}
</script>
```

Now if I copy the code from https://kit.svelte.dev/docs/loading:

```svelte
<script context="module" lang="ts">
  export function load({ url }) {
    //...
  }
</script>
```

the `url` is `any`.

Why?

Am I wrong using it?

tv42:
This change made using Typescript harder. Previously, I could say 

```
  export async function load({ fetch }: LoadInput): Promise<LoadOutput> {
```

Now it seems I have to use this form

```
  export const load: Load = ({ fetch }) => {
```

I know this is a style thing, and a lot of people seem to write arrow expressions everywhere, but I'm more inclined to agree with https://deno.land/manual/contributing/style_guide#top-level-functions-should-not-use-arrow-syntax

benmccann:
I'm not disagreeing about whether we should export `LoadInput` and `LoadOutput`, but I would think if you want to avoid arrow expressions you could write it as:

```
  export const load: Load = function({ fetch }) {
```

PR #4109, opened 2022-02-24
rename `EndpointOutput` to `RequestHandlerOutput` and tidy up

Rich-Harris:
This clarifies the terminology a bit — an _endpoint_ is a module corresponding to a route, a _request handler_ is a function exported from an endpoint. Therefore `RequestHandlerOutput` makes more sense than `EndpointOutput`.

It also neatens the type up. Before:

image

After:

image

Before:

image

After:

image

No changeset, since now that these types are private there's no breakage involved.

### Tests

### Changesets

### DIFF
--- a/packages/kit/types/index.d.ts
+++ b/packages/kit/types/index.d.ts
@@ -16,8 +16,8 @@ import {
 	MaybePromise,
 	PrerenderOnErrorValue,
 	RequestEvent,
-	RequestHandlerOutput,
 	ResolveOptions,
+	ResponseHeaders,
 	TrailingSlash
 } from './private';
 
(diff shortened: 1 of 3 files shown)


Conduitry:
Shouldn't this have a changeset (probably marked as a breaking change) so that the published types are updated, instead of just the site?

Rich-Harris:
It's not a breaking change, since `EndpointOutput` isn't a public type (as of #4104). As far as a consumer of `RequestHandler` is concerned, nothing has changed

Rich-Harris:
Just realised there's an errant `>` in the tooltip, not sure why

image

bjon:
@Rich-Harris  Its breaking for us. We're using EndpointOutput.

e.g.
```
export function internalError(
  msg = `Internal server error`,
): EndpointOutput<string> {
  return {
    status: 500,
    headers: {
      "Content-Type": contentType.textPlain,
    },
    body: msg,
  };
}
```

Conduitry:
It sounds like #4104 would have been the breaking change for you then, not this change.

bjon:
@Conduitry Yes you're right. Was on .278

bjon:
@Conduitry Yes you're right. Was on .278

Rich-Harris:
At any rate that's a convincing reason to move `RequestHandlerOutput` from `private.d.ts` to `index.d.ts`, at which point this PR will be an unbreaking change for you (sort of)

PR #4115, opened 2022-02-24
remove unnecessary units

icecream17:
This is extremely minor so I haven't filled in most of the checklist

Changes `0px 0px` to `0 0`

Using units after 0 causes a lint warning in vscode, `css(zeroUnits)`, and also it's consistent with the omission in `margin: 0 0 0.5rem 0`

There are some cases in `calc` where units are required, but this doesn't apply since the change is in [`drop-shadow`](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/drop-shadow())

### Tests

### Changesets

### DIFF
--- a/.changeset/late-bats-rule.md
+++ b/.changeset/late-bats-rule.md
@@ -1,0 +1,5 @@
+---
+"create-svelte": patch
+---
+
+remove unnecessary CSS units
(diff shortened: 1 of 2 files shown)


Rich-Harris (approved):

Rich-Harris:
thanks!

PR #4116, opened 2022-02-24
Try to fix SPA initial behavior for 404s

PH4NTOMiki:

### Tests

### Changesets

One of the ways to fix https://github.com/sveltejs/kit/pull/4049#issuecomment-1048624895
we can probably completely discard that `initialization` boolean in router if we decide to do this.

### DIFF
--- a/.changeset/blue-parents-invite.md
+++ b/.changeset/blue-parents-invite.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+fixes SPA mode error handling
--- a/.changeset/blue-parents-invite.md
+++ b/.changeset/blue-parents-invite.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-fixes SPA mode error handling
+Only fall back to full page reload if pathname has changed
--- a/packages/kit/src/runtime/client/router.js
+++ b/packages/kit/src/runtime/client/router.js
@@ -56,15 +56,18 @@ export class Router {
 	 *    base: string;
 	 *    routes: import('types').CSRRoute[];
 	 *    trailing_slash: import('types').TrailingSlash;
-	 *    renderer: import('./renderer').Renderer
+	 *    renderer: import('./renderer').Renderer;
+	 *    is_fallback: boolean;
 	 * }} opts
 	 */
-	constructor({ base, routes, trailing_slash, renderer }) {
+	constructor({ base, routes, trailing_slash, renderer, is_fallback }) {
 		this.base = base;
 		this.routes = routes;
 		this.trailing_slash = trailing_slash;
 		/** Keeps tracks of multiple navigations caused by redirects during rendering */
 		this.navigating = 0;
+
+		this.is_fallback = is_fallback;
 
 		/** @type {import('./renderer').Renderer} */
 		this.renderer = renderer;
@@ -283,10 +286,9 @@ export class Router {
 
 	/**
 	 * @param {URL} url
-	 * @param {boolean} [skip_browser_reload]
 	 * @returns {import('./types').NavigationInfo | undefined}
 	 */
-	parse(url, skip_browser_reload) {
+	parse(url) {
 		if (this.owns(url)) {
 			const path = decodeURI(url.pathname.slice(this.base.length) || '/');
 
@@ -294,8 +296,7 @@ export class Router {
 				id: url.pathname + url.search,
 				routes: this.routes.filter(([pattern]) => pattern.test(path)),
 				url,
-				path,
-				initial: skip_browser_reload ?? !this.initialized
+				path
 			};
 		}
 	}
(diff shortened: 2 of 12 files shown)


PH4NTOMiki:
@Rich-Harris @benmccann what do you think of this, am I missing something?

benmccann:
I have some other ideas about how this should be cleaned up. I'll try to share something shortly

PH4NTOMiki:
Okay, fully understand, just one of ideas, definitely not the best one.

packages/kit/src/runtime/client/router.js
```diff
@@ -294,7 +295,7 @@ export class Router {
 				routes: this.routes.filter(([pattern]) => pattern.test(path)),
 				url,
 				path,
-				initial: !this.initialized
+				initial: skip_browser_reload ?? !this.initialized
```
benmccann, line 298: I haven't gotten to sit down and understand this PR yet, but I'm concerned that this variable no longer represents the `initial` page load, but is still being called `initial`. I think we'd need a better way to represent this

I sent a PR before this one to cleanup the router APIs and moved this flag to `Renderer.has_hydrated`: https://github.com/sveltejs/kit/pull/4101

PH4NTOMiki: Yeah, that makes sense

benmccann:
It was unclear to me what this is supposed to be fixing, but it looks like it's referred to in the comments of another PR (https://github.com/sveltejs/kit/pull/4049#issuecomment-1048624895). A closed PR is a hard place to have a discussion about this. Can you open a new issue for it?

Rich-Harris:
Thanks. I've added a test that fails without this PR.

Been thinking a little more about the circumstances in which this issue arises. We basically only need the guard when a) we're hydrating a page from a fallback `200.html` (or equivalent) page, or b) we've invalidated a URL or the `session` and we haven't yet navigated away from the initial page. The current check is a little too broad. Will push up some changes shortly

packages/kit/src/runtime/client/router.js
```diff
@@ -449,6 +450,10 @@ export class Router {
 		}
 
 		if (details) {
+			if (this.is_fallback && info.url.pathname !== location.pathname) {
```
benmccann, line 453: what is the second half of this check doing? I don't understand how an app can be initialized as a fallback/SPA and then later not become one. Maybe this variable is representing something else in this case as should have a different name?

Rich-Harris: an SPA can have _some_ pages prerendered. the bug we're trying to avoid is infinite reloads when the initial hydration fails (or subsequent invalidations that aren't caused by navigation), rather than preventing full page reloads across the board

packages/kit/src/runtime/client/router.js
```diff
@@ -448,6 +450,10 @@ export class Router {
...
```
benmccann, line 453: we should probably add a comment to clarify at least
```suggestion
			// if navigating to a prerendered page
			if (this.is_fallback && info.url.pathname !== location.pathname) {
```

but another way to write this might be to move the URL check to `renderer`. overwriting `is_fallback` seems a bit funny to me because at this point it is still a fallback. it's not until after we've left and navigated to the new page that it's not a fallback

Rich-Harris: Not sure I understand what you mean about moving the URL check to `renderer`. The responsibility for this stuff (to the extent that the two things are separate — I'm going to open a PR today to merge them, because it really will simplify things) lies with the router

> it's not until after we've left and navigated to the new page that it's not a fallback

pushing state _is_ navigating to the new page

benmccann: I mean you could change `if (!navigation_result && this.router?.is_fallback) {` to `if (!navigation_result && this.router?.is_fallback && info.url.pathname !== location.pathname) {`

If I understand correctly, when this is triggered we will do a server-side nav. This page is still a fallback page in my mind until the server-side nav has happened

Rich-Harris: Actually you know what? If we were to compare `url.pathname` with `location.pathname` at that point, it would arguably be simpler to do away with the fallback/non-fallback distinction altogether and just always bail out of full page reloads if they're the same (i.e. it's not a navigation, it's a hydration/invalidation). commit inbound

PH4NTOMiki: Yeah, that's simpler and should work.

.changeset/blue-parents-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+fixes SPA mode error handling
```
Rich-Harris, line 5:
```suggestion
Only fall back to full page reload if pathname has changed
```

Rich-Harris (approved):

PR #4118, opened 2022-02-24
generated tsconfig.json

Rich-Harris:
Groundwork for #647. This adds a generated `.svelte-kit/tsconfig.json` and checks that the user's config (if such exists) `extends` it.

This allows us to make changes like #3064 that existing apps can take advantage of (rather than only newly created apps), including the change I have planned that will address #647.

It also allows us to base certain values on the user's configuration. For example, if someone were to specify a different `lib` folder...

```js
// svelte.config.js
export default {
  kit: {
    files: {
      lib: 'src/my-stuff'
    }
  }
};
```

...then we can automatically generate the correct alias:

```js
{
  "paths": {
    "$lib": ["../src/my-stuff"],
    "$lib/*": ["../src/my-stuff/*"]
  }
}
```

I'm not sure how much of the existing `compilerOptions` is necessary, but I figured I'd keep it there for now unless people have thoughts.

### Tests

### Changesets


packages/kit/src/core/tsconfig.js
```diff
@@ -0,0 +1,101 @@
...
+
+	const paths = {};
+
+	paths['$lib'] = [path.relative(SVELTE_KIT, config.kit.files.lib)];
+	paths['$lib/*'] = [path.relative(SVELTE_KIT, config.kit.files.lib) + '/*'];
+
+	if (user_file) {
+		// we have to eval the file, since it's not parseable as JSON (contains comments)
+		const user_tsconfig_json = fs.readFileSync(user_file, 'utf-8');
+		const user_tsconfig = (0, eval)(`(${user_tsconfig_json})`);
```
Conduitry, line 23: Eesh, this makes me extremely uncomfortable. I don't know what else I would suggest. Regexes to strip the comments (somehow) and then parse it?

Conduitry: A very quick search reveals https://stackoverflow.com/a/62945875 which suggests `str.replace(/\\"|"(?:\\"|[^"])*"|(\/\/.*|\/\*[\s\S]*?\*\/)/g, (m, g) => g ? "" : m)` which I have not tested at all.

Conduitry: I do realize that, yes, any code in your app's directory is stuff you're supposed to trust, and that this is probably an area that it's not worth trying to protect people from themselves, but still. This gives me the creeps.

Rich-Harris: yeah, i didn't think an `eval` would get past the Conduitry filter, even if it's made indirect with `(0, eval)`. What's the danger though? Presumably an attacker capable of modifying `tsconfig.json` could just as equally modify `src/hooks.js`, which has far more potential for mischief

Conduitry: Yeah, I suppose that, rationally, there's not any great harm in this, nor any great benefit in avoiding this one particular attack vector. I'll resolve this conversation, but I got my eye on you Mr. Harris.

packages/kit/src/core/tsconfig.js
```diff
@@ -0,0 +1,101 @@
...
+	fs.writeFileSync(
+		`${SVELTE_KIT}/tsconfig.json`,
+		JSON.stringify(
+			{
+				compilerOptions: {
+					moduleResolution: 'node',
+					module: 'es2020',
+					lib: ['es2020', 'DOM'],
+					target: 'es2020',
+					// svelte-preprocess cannot figure out whether you have a value or a type, so tell TypeScript
```
Conduitry, line 75: Are we worried about losing these comments in the generated tsconfig file that gets included? Does tooling generally provide a way to look at the file being extended, and do we care if that becomes less readable.

Or, another way of asking it: was the only reason for the user-visible comments so that people could know which config options not to mess with?

Also sort of relatedly: Do you foresee us potentially wanting to warn or error out if the user config does override certain essential values from the generated config?

packages/kit/src/core/tsconfig.js
```diff
@@ -0,0 +1,101 @@
...
+					sourceMap: true,
+					esModuleInterop: true,
+					skipLibCheck: true,
+					forceConsistentCasingInFileNames: true,
+					baseUrl: path.relative(SVELTE_KIT, '.'),
+					allowJs: true,
+					checkJs: true,
+					paths
+				},
+				include: ['../**/*.d.ts', '../**/*.js', '../**/*.ts', '../**/*.svelte'],
```
Conduitry, line 94: I recall there was a feature recently request to be able to customize the location/name of `.svelte-kit` directory, which would potentially mess up some of the stuff here, like these relative paths. This seems like the sort of thing that it would be easy to forget about if we do implement a feature where that can be customized. Do you think it's worth it to make this more flexible now?

packages/kit/src/core/tsconfig.js
Rich-Harris: I'm not particularly worried about the comments. Though I do think it's probably a good idea to warn if the user specifies incompatible values (as already happens with `compilerOptions.paths`)

packages/kit/src/core/tsconfig.js
Rich-Harris: yeah, in some places this PR is doing it right, but i got lazy with `include`. definitely needs fixing. though actually we can probably get away with this, since `baseUrl` is set:

```js
include: ['**/*.d.ts', '**/*.js', '**/*.ts', '**/*.svelte']
```

Similarly, `paths` should come out as e.g. `src/lib` rather than `../src/lib`, it turns out

.changeset/silver-horses-check.md
```diff
@@ -0,0 +1,6 @@
+---
+'create-svelte': patch
+'@sveltejs/kit': patch
+---
+
+Extend user tsconfig from generated .svelte-kit/tsconfig.json
```
benmccann, line 6: I haven't looked closely, but my assumption is that this is probably not a strictly breaking change, but we would want people to update their apps. Maybe we should mark it as `[breaking]` to make sure people update their apps?

Conduitry:
Is there anything similar we'd want to do with jsconfig.json? Does that support `extends` in the same way?

Rich-Harris:
jsconfig.json works exactly the same way, yeah (in fact i've been testing all this with a jsconfig)

.changeset/silver-horses-check.md
Rich-Harris: they'll see a warning like this:

image

i reckon that'll be more effective than relying on people to read the changelogs

Conduitry:
Oh whoops, I missed in the diff that you'd update the jsconfig in the starter templates as well to extend the same file. Never mind, nice!

packages/kit/src/core/tsconfig.js
```diff
@@ -0,0 +1,101 @@
...
+						(/** @type {string} */ relative) => path.resolve('.', relative) === config.kit.files.lib
+					) ||
+					!lib_.some(
+						(/** @type {string} */ relative) =>
+							path.resolve('.', relative) === config.kit.files.lib + '/*'
+					);
+
+				if (missing_lib_paths) {
+					console.warn(
+						`\u001B[1m\u001B[93mYour compilerOptions.paths in ${user_file} should include the following:\u001B[39m\u001B[22m`
```
benmccann, line 47: I see the same leading/trailing characters here and below. It could be worth refactoring them out into variables with a name or comment that explains what they are

packages/kit/src/core/tsconfig.js
```diff
@@ -0,0 +1,101 @@
...
+				}
+			}
+		} else {
+			let relative = path.relative('.', out);
+			if (relative.startsWith(SVELTE_KIT)) relative = './' + relative;
+
+			console.warn(
+				`\u001B[1m\u001B[93mYour ${user_file} should include the following:\u001B[39m\u001B[22m`
+			);
+			console.warn(`"extends": "${relative}"`);
```
benmccann, line 60: This is great! I wonder if people will not know exactly where to add it or realize they can remove everything else. Maybe we could tweak the wording slightly to make those points clearer?

```suggestion
			console.warn(
				`\u001B[1m\u001B[93mYour ${user_file} should extend SvelteKit's. New projects simply have:\u001B[39m\u001B[22m`
			);
			console.warn(`{ "extends": "${relative}" }`);
```

packages/kit/src/core/tsconfig.js
```diff
@@ -0,0 +1,101 @@
...
+/** @param {import('types').ValidatedConfig} config */
+export function generate_tsconfig(config) {
+	const out = path.resolve(SVELTE_KIT, 'tsconfig.json');
+
+	const user_file = exists('tsconfig.json') || exists('jsconfig.json');
+
+	const paths = {};
+
+	paths['$lib'] = [path.relative(SVELTE_KIT, config.kit.files.lib)];
+	paths['$lib/*'] = [path.relative(SVELTE_KIT, config.kit.files.lib) + '/*'];
```
benmccann, line 18:
```suggestion
	const paths = {
		'$lib': [path.relative(SVELTE_KIT, config.kit.files.lib)],
		'$lib/*': [path.relative(SVELTE_KIT, config.kit.files.lib) + '/*'];
	};
```

benmccann: I think this could be moved down to line 65 as it is not used until then

packages/kit/src/core/tsconfig.js
```diff
@@ -0,0 +1,101 @@
...
+
+	if (user_file) {
```
benmccann, line 20: I was trying to see it this block mutates anything. I don't think it does, but it's a little hard to see. Maybe the whole thing could be moved to a function like `validate_user_config` to make it clearer that it has no side effects? (assuming that's all it does from my quick look at it)

benmccann:
I left a few stylistic comments, but LGTM

packages/kit/src/core/tsconfig.js
Rich-Harris: we can actually just use `kleur`, not sure what i was thinking

janosh:
I'm getting warnings in lots of projects following this PR.

```
Your tsconfig.json should extend the configuration generated by SvelteKit:
{
}
```

Just adding the above lines to `tsconfig.json` doesn't seem like a good solution as it would break all CI scripts that run tests without calling `svelte-kit dev/build` first:

```
Error: failed to resolve "extends":"./.svelte-kit/tsconfig.json"
```

What's the recommended setup now?

Conduitry:
I'd say the recommendation is to wait for #4182 to be released.

applemonkey496:
After this PR, I am now unable to resolve imports to asset files. It seems that the generated tsconfig defaults to only including `*.js`, `*.ts`, and `*.svelte` files. Unfortunately this also means that importing such files as `assets/styles.css` result in a vite error when building the project.

I have resolved this issue by adding the following to my tsconfig:
```jsonc
{
    "include": ["src/**/*"],
    "exclude": [],
}
```

I do think that this makes more sense than only allowing `.js`, `.ts`, and `.svelte` files, because many users are going to want to make use of vite's [asset handling](https://vitejs.dev/guide/assets.html) features. 

I think it would be beneficial and avoid some confusion to implement the following behaviour as default:
- include *all* files inside the directory specified in `config.kit.files.assets`
- include all `.js`, `.ts`, and `.svelte` files inside `src/` (existing behaviour) and inside the `hooks`, `lib`, and `routes` folders specified in `config.kit.files` (in case the user specifies custom locations for these folders)

benmccann:
@applemonkey496 could you file a new issue about this so that we can track it? It will get lost as a comment on a closed PR

Issue #4052, opened 2022-02-22
Configure output directory (currently fixed to ".svelte-kit")


Marr11317:
### Describe the problem

In monorepos, builds are often centralized into a single top-level build directory. 

Is there anything that prevents that change from being innocent and simple?

### Describe the proposed solution

In order for svelte-kit to support that, the `SVELTE_KIT` constant in `constants.js` should be replaced by a configuration option. 

setting the `kit.vite.build.outDir` options prints an error: 

```
build_server: The value for kit.vite.build.outDir specified in svelte.config.js has been ignored. This option is controlled by SvelteKit.
```

should the vite option be used, or should there be svelte-kit specific option?

### Importance

nice to have

Marr11317:
I couldn't find any existing bug report/feature request for this, which surprises me. I hope I am not duplicating things here. Feel free to close if there's already a similar issue.

Rich-Harris:
The directory is used for more than Vite builds, so we can't use `vite.build.outDir` — Kit needs to control it. Having it be configurable seems reasonable, we would just need to bikeshed a name. `buildDir`? `outDir`? `tmpDir`? something else?

Marr11317:
> The directory is used for more than Vite builds, so we can't use `vite.build.outDir` — Kit needs to control it. 

Can't we use `vite.build.outDir` for more than vite builds?

Marr11317:
> we would just need to bikeshed a name. `buildDir`? `outDir`? `tmpDir`? something else?

I have a preference for `outDir`, just because it's clearer than `tmpDir`, is more than just a `buildDir`, and is in line with vite.

flcarre:
Figuring the same issue for adding sveltekit to https://github.com/jhipster, as a monorepo project generator, while we can't change the output dir, we can't merge the sveltekit pr

DominikPieper:
It would be a great help to create a better Nx integration

flcarre:
Finaly found a workarround using adapter-static

`  kit: {
    adapter: adapter({
      pages: 'target/classes/static/',
      assets: 'target/classes/static/',
      fallback: 'index.html',
    }),
    appDir: '_app',`

Conduitry:
@Rich-Harris You wanted to keep this open until the case of `outDir` being outside the CWD was sorted out, right?

Rich-Harris:
[...]

PR #4120, opened 2022-02-24
Generated types

Rich-Harris:
This code is a bit ratty, but I figured I'd push it up anyway. It generates `.d.ts` files for every page/endpoint file in `src/routes` (but tucked away in `.svelte-kit` via the magic of `rootDirs`). Easier to show than tell:

image

An eventual follow-up would be to automatically type `Props` (and the return value of a shadow endpoint, if no `load` function is given) based on a component's `<script>` tag.

### Tests

### Changesets


Rich-Harris:
This is ready for review but lint is failing because `pnpm check` apparently runs before the test apps have been built (which creates the `.svelte-kit/tsconfig.json` files that `pnpm check` depends on). Not sure why, since the job includes `pnpm build`: https://github.com/sveltejs/kit/blob/eb7492fbe9d53a28190d49811a77c5306ca5f062/.github/workflows/ci.yml#L13-L27 Any ideas?

documentation/docs/14-types.md
```diff
@@ -3,3 +3,70 @@ title: Types
...
 
 **TYPES**
+
+### Generated types
+
+The [`RequestHandler`](#sveltejs-kit-requesthandler) and [`Load`](#sveltejs-kit-load) types both accept a `Params` argument allowing you to type the `params` object:
+
+```diff
+/// file: src/routes/[foo]/[bar]/[baz].js
+-/** @type {import('@sveltejs').RequestHandler} */
```
Conduitry, line 13: This diff feels a little confusing. Is this a migration guide kind of thing leaking out to somewhere we should be more forward looking?

documentation/docs/01-routing.md
```diff
@@ -74,11 +78,13 @@ export async function get({ params }) {
 	return {
 		status: 404
 	};
-};
+}
 ```
 
 > All server-side code, including endpoints, has access to `fetch` in case you need to request data from external APIs. Don't worry about the `$lib` import, we'll get to that [later](/docs/modules#$lib).
 
+The type of the `get` function above comes from `./[id].d.ts`, which is a file generated by SvelteKit (in a hidden directory, using the [`rootDirs`](https://www.typescriptlang.org/tsconfig#rootDirs) option) that provides type safety when accessing `params`. See the section on [generated types](/docs/types#generated-types) for more detail.
```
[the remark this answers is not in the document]
Rich-Harris: We don't, currently. It might be a bit much to introduce it in this sentence alongside everything else. If we implemented #4052 that might help

documentation/docs/14-types.md
Rich-Harris: It's meant to be read as 'you _could_ type `params` like this' in order to set up 'but you _should_ do it like this', but will concede it might not be landing. Any thoughts on a better alternative?

documentation/docs/01-routing.md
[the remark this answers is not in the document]
Conduitry: That's fair. I think the current text here is good enough for the time being, then.

documentation/docs/14-types.md
Conduitry: Maybe just be more explicit about it? Mention the old types inline in the paragraph, and then just write the manually typed params in the code block, which would even let you use JS syntax highlighting and hover types, rather than syntax highlighted as a diff.

Conduitry: Oh, I see the link to the types in the diff code block already works in the deployed site, so that part isn't as much of a concern.

What is missing, though, is the path of this file, so the part about renaming the directory just below this is confusing.

documentation/docs/14-types.md
```diff
@@ -3,3 +3,70 @@ title: Types
...
+```svelte
+<script context="module">
+	/** @type {import('./[baz]').Load} */
+	export async function load({ params, fetch, session, stuff }) {
+		// ...
+	}
+</script>
+```
+
+> For this to work, your own `tsconfig.json` or `jsconfig.json` should extend from the generated `.svelte-kit/tsconfig.json`:
```
Conduitry, line 70: Oh, I guess we are mentioning .svelte-kit here. I dunno. It is officially part of the API now that we want people to extend the config in it. I'm not sure whether that changes my opinion about mentioning it up higher instead of calling it a hidden folder.

documentation/docs/14-types.md
Rich-Harris: Tweaked it a bit. The file path is there, though maybe it's liable to get missed?

image

Rich-Harris: Made it a bit more obvious 4471a8b2061ab1b27ef6b0f762cb30d68e5289f3

Conduitry: It wasn't displaying the filename `src/routes/[foo]/[bar]/[baz].js` at the top of the `diff` code block before

image

but I do see it there now. I think the new version is good.

packages/kit/src/core/tsconfig.js
```diff
@@ -22,6 +22,19 @@ export function generate_tsconfig(config) {
 	/** @param {string} file */
 	const config_relative = (file) => path.relative(SVELTE_KIT, file);
 
+	const dirs = new Set([
+		project_relative(path.dirname(config.kit.files.routes)),
+		project_relative(path.dirname(config.kit.files.lib))
```
dummdidumm, line 27: Not sure if this works for everyone. People could have other stuff besides lib/routes below their src and want it to have included in their tsconfig. This also might exclude `app.d.ts` as a result. What's the reason for changing it from the previous implementation?

Rich-Harris: Before, it was `../**/*.js` etc, which included stuff in things like `build` — it was causing our own CI to fail. This scopes it to `../src` by default (since that's the common parent of `src/routes` and `src/lib`), which works unless you start putting things outside `src` in which case you would need to modify your own `include` config (but hopefully at that point you know what you're doing)

dummdidumm: Ah I missed the `path.dirname`, I thought this would result in `src/lib/**/*.ts` etc, but it results in `src/**/*.ts` - all good then 👍

documentation/docs/14-types.md
dummdidumm: I would also add a note that that folder is not available immediately, so you'll need to run a command first (we also need to restart the language-server for it to resolve the extended tsconfig again, adding it to this list https://github.com/sveltejs/language-tools/blob/master/packages/svelte-vscode/src/extension.ts#L144).
What if we add a postinstall hook to the `package.json` of the create-svelte templates to immediately generate the types?

Conduitry: Do `postinstall` hooks run after you run `npm install` on the repo whose `package.json` they're in, or do they run when you've just installed a package whose `package.json` has one?

Rich-Harris: Just checked, they do. Will add a `svelte-kit init` command and a `"postinstall": "svelte-kit init"` run script

Rich-Harris: actually i'll do that in a separate PR, because there might be some bikeshedding involved

pheuter:
Would it be fair to understand these changes as striving towards the direction of something like [tRPC](https://trpc.io/) where you have the client being able to use types auto-generated from server endpoints?

ambiguous48:
> Would it be fair to understand these changes as striving towards the direction of something like [tRPC](https://trpc.io/) where you have the client being able to use types auto-generated from server endpoints?

@pheuter It seems like it, to me at least, given that #647 is included in the 1.0 milestone.

Although, that may just be wishful thinking on my part. I recently added tRPC to my sveltekit app, in what I can only describe as the cherry to the delicous cake that is sveltekit. The effortless type safety has made what was already a ridiculously fun dev experience even better. Needless to say, I'd be pretty excited if sveltekit were to move in that direction.

And I mention tRPC in particular becuase of how simple or lightweight their approach is, which I feel aligns with svelte's own. In contrast, my original approach of implementing "end-to-end" typesafety was through graphQL. And while it could lead to the same result, the bloat accompanying graphQL quickly gets out of hand where you start questioning whether or not you're actually improving the dx at all. It ends up sticking out like a sore thumb alongside the simplicity of your typical sveltekit app.

P.S. I'm not an expert on the matter by any means, do take what I say with a grain of salt.

BrettBedarf:
Is this intended to work with the same for .ts files in addition to .js with jsdoc comments?

I am able to import the generated `RequestHandler` type from its full project path but not relative.

image

Minimal example: https://github.com/BrettBedarf/generated-types-minimal

`/routes/ts/[slug].json.ts` vs `/routes/js/[slug].json.js`

I think the issue is typescript trying to resolve the same file name does not "merge" the import

If I manually change the generated .d.ts file to `[slug].json.gen.d.ts` and change the import to `import type { RequestHandler } from './[slug].json.gen';` I am able to use the type

ambiguous48:
@BrettBedarf I'm seeing a similar issue. Eslint is reporting that the file is importing itself. It seems that an endpoints with a `.ts` extension wil try to import itself rather than importing it's generated types.

In the meantime, try the following workaround. Basically just add a `.d` to the import.

```diff
- import type { RequestHandler } from './[slug].json';
+ import type { RequestHandler } from './[slug].json.d';
```

BrettBedarf:
@ambiguous48 good workaround thanks!

TravisSpomer:
> Is this intended to work the same for .ts files in addition to .js with jsdoc comments?

It doesn't seem so—seems like a really strange oversight. But you can hack it to work in TypeScript like this (at least for pages):

```ts
import type { Load } from "./filename"
export async function load({ params }: Parameters<Load>[0]): Promise<ReturnType<Load>>
{ /* ... */ }
```

If the generated types file exported `LoadInput` and `LoadOutput` then it would be a lot simpler.

PR #4122, opened 2022-02-25
Fix race condition in clicknav test utility func

### LINKED ISSUE #4121: `clicknav` in test utils has a race condition
### Describe the bug

The current implementation of the `clicknav` utility function in https://github.com/sveltejs/kit/blob/fd8bca9796e2a9dbdbe27ae147393e70fcc6ce77/packages/kit/test/utils.js#L76 has a race condition. [The Playwright docs](https://playwright.dev/docs/navigations#asynchronous-navigation) mention that it's important to put the `waitForNavigation` call *before* the `.click()` call that can trigger navigation, specifically to avoid a race condition where the navigation happens before the `waitForNavigation` call can set up the wait condition.

In other words, [this line](https://github.com/sveltejs/kit/blob/fd8bca9796e2a9dbdbe27ae147393e70fcc6ce77/packages/kit/test/utils.js#L76) in the current implementation of `clicknav`:

```js
await Promise.all([page.click(selector), page.waitForNavigation(options)]);
```

should be changed to:

```js
await Promise.all([page.waitForNavigation(options), page.click(selector)]);
```

### Reproduction

Can't reproduce reliably as it's a race condition (and one that I'd expect to trigger only rarely), but it's clearly mentioned in [the Playwright documentation](https://playwright.dev/docs/navigations#asynchronous-navigation) so we might as well do it correctly.

Note that it *is* being done correctly [here](https://github.com/sveltejs/kit/blob/93d22b784afb7ff0c69936e9843b4fb6a57adca0/packages/kit/test/apps/basics/test/test.js#L409) and [here](https://github.com/sveltejs/kit/blob/93d22b784afb7ff0c69936e9843b4fb6a57adca0/packages/kit/test/apps/basics/test/test.js#L415).

### Logs

_No response_

### System Info

```shell
N/A
```

### Severity

annoyance

### Additional Information

_No response_

rmunn:
[The Playwright docs](https://playwright.dev/docs/navigations#asynchronous-navigation) mention that when using `page.waitForNavigation()`, it's important to use `Promise.all` and await the navigation first, before triggering the event (such as a link click) that will begin navigation. Otherwise there could be a race condition where the navigation happens too fast, before the `waitForNavigation()` can get set up, resulting in `waitForNavigation()` timing out and the test failing.

It's very unlikely to happen in practice, but "very unlikely" isn't the same as "impossible", so we might as well do it the way the Playwright docs recommend, with waitForNavigation coming first.

**NOTE:** I ran the tests (for the entire repo) both before and after making this change. They passed both before and after, but there was one test that was flaky after making the change. It failed twice, then passed on the third try with no code changes. That test was the 'prevents navigation triggered by back button' test [here](https://github.com/sveltejs/kit/blob/06ceef5ef052861951eabfb525430fad43dafb89/packages/kit/test/apps/basics/test/test.js#L126). Since it doesn't use the `clicknav` fixture at all, and the only change this PR makes is to the `clicknav` fixture, I don't believe the failure is related.

No changeset included as I don't believe this is something that should be mentioned in the changelog.

Fixes #4121.

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/apps/basics/test/test.js
+++ b/packages/kit/test/apps/basics/test/test.js
@@ -1422,8 +1422,8 @@ test.describe.parallel('Page options', () => {
 		expect(await page.textContent('button')).toBe('clicks: 0');
 
 		if (javaScriptEnabled) {
-			await Promise.all([page.click('[href="/no-hydrate/other"]'), page.waitForNavigation()]);
-			await Promise.all([page.click('[href="/no-hydrate"]'), page.waitForNavigation()]);
+			await Promise.all([page.waitForNavigation(), page.click('[href="/no-hydrate/other"]')]);
+			await Promise.all([page.waitForNavigation(), page.click('[href="/no-hydrate"]')]);
 
 			await page.click('button');
 			expect(await page.textContent('button')).toBe('clicks: 1');
@@ -1457,7 +1457,7 @@ test.describe.parallel('Page options', () => {
 			await page.click('button');
 			expect(await page.textContent('button')).toBe('clicks: 1');
 
-			await Promise.all([page.click('[href="/no-router/b"]'), page.waitForNavigation()]);
+			await Promise.all([page.waitForNavigation(), page.click('[href="/no-router/b"]')]);
 			expect(await page.textContent('button')).toBe('clicks: 0');
 
 			await page.click('button');
@@ -1466,7 +1466,7 @@ test.describe.parallel('Page options', () => {
 			await clicknav('[href="/no-router/a"]');
 			expect(await page.textContent('button')).toBe('clicks: 1');
 
-			await Promise.all([page.click('[href="/no-router/b"]'), page.waitForNavigation()]);
+			await Promise.all([page.waitForNavigation(), page.click('[href="/no-router/b"]')]);
 			expect(await page.textContent('button')).toBe('clicks: 0');
 		}
 	});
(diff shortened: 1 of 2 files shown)


bluwy (approved):

ignatiusmb:
There's also some inconsistencies in `packages/kit/test/apps/basics/test/test.js`, can you adjust them as well while we're here?

rmunn:
> There's also some inconsistencies in `packages/kit/test/apps/basics/test/test.js`, can you adjust them as well while we're here?

@ignatiusmb Done, in commit https://github.com/sveltejs/kit/pull/4122/commits/62502effd711379d03b46582a1f781b61deb3dc7.

baseballyama (approved):
> NOTE: I ran the tests (for the entire repo) both before and after making this change. They passed both before and after, but t[here](https://github.com/sveltejs/kit/blob/06ceef5ef052861951eabfb525430fad43dafb89/packages/kit/test/apps/basics/test/test.js#L126) was one test that was flaky after making the change. It failed twice, then passed on the third try with no code changes. That test was the 'prevents navigation triggered by back button' test here. Since it doesn't use the clicknav fixture at all, and the only change this PR makes is to the clicknav fixture, I don't believe the failure is related.

Yeah the day before yesterday, I fixed tests flakiness.
However, there still seems to be some flakiness with `goto`  (There may be others) on Windows.
I will continue to investigate the cause of this.

ignatiusmb:
Thanks!

PR #4128, opened 2022-02-25
Improve JSON payload handling

mrkishi:
This fixes security issues related to script elements carrying JSON payloads, while refactoring the internal API so it's less likely to be used incorrectly.

## Changes to the internal API

Since `escape_json_in_html` worked in chunks of data, it was easy to mistake it for a safe method regardless of context. However, the transformed payload wouldn't be useable everywhere in a document. The only place it could safely be placed is inside a `<script>` tag.

Therefore, instead of escaping values ad-hoc and later interpolating strings to assemble the final script tag, payloads are now kept as objects until the time to render them to HTML. At that point, calling `render_json_payload_script` will render the whole script element to html, at once.

This change is intended to minimize the risk that escaped values are accidentally mishandled. As a matter of fact, we previously forgot to escape `statusText` and `headers`, current attack vectors.

It also uses `sveltekit:data-{property}` as attribute names for the payload's associated key-value data. While this is not final and is open to bikeshed, it avoids conflicts with user attributes as was previously the case.

## Changes to escaped characters

These changes are not strictly required, but since we're now ensuring the payload is in a script tag, it is possible to reduce the amount of escaped characters. Some information:

 - Escaping `<` and `>` is not strictly necessary. These have no meaning to HTML parsers inside script elements, with the exception of the literal sequence `</script`, which could be handled by escaping any of those characters. [1]

 - `\u2028` and `\u2029` are blips of history. They used to cause issues because they weren't valid characters inside JS strings. However, not only were they always valid in non-js script elements, they are also valid in JS string since the "Subsume JSON" proposal was standardized in ES2019. [2]

 - Care must be taken about HTML comments. Even though `<script>` is famously a raw text element that will parse `"</script>"` as a valid closing tag, this is not true inside HTML comments. A rogue `<!--` can deactivate parsing until the next `-->`, after which text will continue to be consumed by the script element. [3]

With this in mind, we can limit the escaped characters to `<`, but `\u2028` and `\u2029` are escaped for the sake of older browsers.

[1] https://html.spec.whatwg.org/multipage/syntax.html#cdata-rcdata-restrictions
[2] https://github.com/tc39/proposal-json-superset
[3] https://html.spec.whatwg.org/multipage/parsing.html#script-data-double-escaped-state

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -42,13 +42,7 @@ export async function load_node({
 
 	let uses_credentials = false;
 
-	/**
-	 * @type {Array<{
-	 *   url: string;
-	 *   body: string;
-	 *   json: import('types').JSONValue;
-	 * }>}
-	 */
+	/** @type {Array<import('./types').Fetched>} */
 	const fetched = [];
 
 	/**
@@ -248,8 +242,6 @@ export async function load_node({
 							}
 
 							if (!opts.body || typeof opts.body === 'string') {
-								// the json constructed below is later added to the dom in a script tag
-								// make sure the used values are safe
 								const status_number = Number(response.status);
 								if (isNaN(status_number)) {
 									throw new Error(
@@ -258,11 +250,11 @@ export async function load_node({
 										}" type: ${typeof response.status}`
 									);
 								}
-								// prettier-ignore
+
 								fetched.push({
 									url: requested,
-									body: /** @type {string} */ (opts.body),
-									json: {
+									body: opts.body,
+									response: {
 										status: status_number,
 										statusText: response.statusText,
 										headers,
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -1,7 +1,5 @@
 import { normalize } from '../../load.js';
 import { respond } from '../index.js';
-import { s } from '../../../utils/misc.js';
-import { escape_json_in_html } from '../../../utils/escape.js';
 import { is_root_relative, resolve } from '../../../utils/url.js';
 import { create_prerendering_url_proxy } from './utils.js';
 import { is_pojo, lowercase_keys, normalize_request_method } from '../utils.js';
@@ -48,7 +46,7 @@ export async function load_node({
 	 * @type {Array<{
 	 *   url: string;
 	 *   body: string;
-	 *   json: string;
+	 *   json: import('types').JSONValue;
 	 * }>}
 	 */
 	const fetched = [];
@@ -264,7 +262,12 @@ export async function load_node({
 								fetched.push({
 									url: requested,
 									body: /** @type {string} */ (opts.body),
-									json: `{"status":${status_number},"statusText":${s(response.statusText)},"headers":${s(headers)},"body":${escape_json_in_html(body)}}`
+									json: {
+										status: status_number,
+										statusText: response.statusText,
+										headers,
+										body
+									}
 								});
 							}
 
--- a/packages/kit/src/utils/escape.js
+++ b/packages/kit/src/utils/escape.js
@@ -1,27 +1,42 @@
-// dict from https://github.com/yahoo/serialize-javascript/blob/183c18a776e4635a379fdc620f81771f219832bb/index.js#L25
-/** @type {Record<string, string>} */
-const escape_json_in_html_dict = {
-	'<': '\\u003C',
-	'>': '\\u003E',
-	'/': '\\u002F',
-	'\u2028': '\\u2028',
-	'\u2029': '\\u2029'
-};
-
-const escape_json_in_html_regex = new RegExp(
-	`[${Object.keys(escape_json_in_html_dict).join('')}]`,
-	'g'
-);
-
 /**
- * Escape a JSONValue that's going to be embedded in a `<script>` tag
- * @param {import('types').JSONValue} val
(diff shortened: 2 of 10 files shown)


packages/kit/src/utils/escape.js
```diff
@@ -1,27 +1,42 @@
-// dict from https://github.com/yahoo/serialize-javascript/blob/183c18a776e4635a379fdc620f81771f219832bb/index.js#L25
-/** @type {Record<string, string>} */
-const escape_json_in_html_dict = {
-	'<': '\\u003C',
-	'>': '\\u003E',
-	'/': '\\u002F',
-	'\u2028': '\\u2028',
```
dominikg, line 7: if i understand it correctly \u2028 and \u2029 escapes existed not because they could break the html parser, but because they could break the clientside JSON.parse when reading the script tag content.

Is that still safe?

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -264,7 +262,12 @@ export async function load_node({
 								fetched.push({
 									url: requested,
 									body: /** @type {string} */ (opts.body),
-									json: `{"status":${status_number},"statusText":${s(response.statusText)},"headers":${s(headers)},"body":${escape_json_in_html(body)}}`
+									json: {
```
dominikg, line 265: love that this asymetric escape json string build is being refactored!

now it's no longer a json string, is there a better name for it? 

Unfortunately i think there was a reason behind the different treatment of headers and body. It may no longer be relevant due to the changes in escaping but in a previous version applying the same escaping rules to headers meant that there were cases where header values got mangled that shouldn't have been.

packages/kit/src/utils/escape.js
mrkishi: Browsers implemented subsume json around 2018. I forgot we still had to step through JS for the parsing, so it might be worth it to keep them, then, for the sake of older browsers.

packages/kit/src/runtime/server/page/load_node.js
mrkishi: This might be worth looking into, then, because I don't see how it'd have altered the output json—after `JSON.parse`, you'd get the same string back, even on the previous implementation.

Maybe the issue was with double-escaping at some point?

packages/kit/src/utils/escape.spec.js
```diff
@@ -21,4 +50,5 @@ attr('escapes invalid surrogates', () => {
 	assert.equal(escape_html_attr('\ud800\ud800\udc00\udc00'), '"&#55296;\ud800\udc00&#56320;"');
 });
 
+json.run();
```
benmccann, line 53: I saw @Rich-Harris comment on a recent PR that it didn't need to create a test suite. The test setup has changed a lot and I'm not sure under which circumstances a test suite is needed and when it is not needed. Flagging this for his attention and my education :smile:

packages/kit/src/utils/escape.js
```diff
@@ -1,27 +1,42 @@
...
- * @param {import('types').JSONValue} val
+ * Generates a raw HTML string containing a safe script element carrying JSON data and associated attributes.
+ *
+ * It escapes all the special characters needed to guarantee the element is unbroken, but care must
+ * be taken to ensure it is inserted in the document at an acceptable position for a script element,
+ * and that the resulting string isn't further modified.
+ *
+ * Attribute names must be type-checked so we don't need to escape them.
+ *
+ * **INTERPOLATING THIS STRING INTO AN UNRELATED SCRIPT ELEMENT IS NOT SAFE.**
```
benmccann, line 10: I'm not sure what this line means. This is outputting a `<script>` tag. Are you thinking someone might try to put this `<script>` tag inside another `<script>` tag? I'm not quite sure what that would mean or why anyone would try to do that

packages/kit/src/utils/escape.spec.js
mrkishi: I thought suites were just informational groups to help test navigation—without them, all tests go into a single anonymous suite. I'll remove them if that causes issues!

packages/kit/src/utils/escape.js
mrkishi: Yeah, it was supposed to empathize the raw html string is not properly vetted for every context. Maybe I should reword it.

It's not that people would intentionally try to do that, but marking a string as "safe for html" in your head is risky: each html context has different and incompatible escaping rules, so care must always be taken to properly place these strings in the document, even if something was "escaped."

packages/kit/src/runtime/server/page/load_node.js
mrkishi: I git-blamed this down to inception and those were indeed never escaped—I'm just unsure why it would've caused a problem. My search-fu failed me in trying to find a discussion on this.

I believe they are being handled correctly, for what it's worth, but let me know if you see the issue!

dominikg: this comment is the only ref i could find of it https://github.com/sveltejs/kit/pull/2534#issue-1013386805 . basically my initial approach at improving this was encoding the whole `json` object and that failed a testcase. If tests are working now and headers isn't used anywhere else it should be fine.

Are headers even needed on the clientside in the script body? it's not like we can magically turn them back into http headers somehow and what if there was an http-only cookie in there?

mrkishi: They are sent for the sake of being included in the `new Response()` we create that can be consumed by `load()` client-side.

But, yeah, tests pass now and I believe this is working.

Issue #4038, opened 2022-02-21
Clashing endpoint-powered pages break route sorting


mrkishi:
### Describe the bug

When multiple endpoint-powered pages have a clashing route path, both client- and server-side routers break, not following sorting rules.

Given:

```
/[id].js
/[id].svelte
/[slug].js
/[slug].svelte
```

A client-side navigation to any matching route renders `[id].svelte` with undefined props, without hitting the endpoint.

A full-page navigation _does_ hit the endpoints and falls through them, but you get back their JSON payload instead of the rendered page.

### Reproduction

[Minimal reproduction available here.](https://github.com/mrkishi/mrkishi-repro-sveltekit-clashing/commit/ca1c21dd2ad2a35b6d0a62db4044d5bda9fe3d9c)

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (4) x64 Intel(R) Core(TM) i5-3570 CPU @ 3.40GHz
    Memory: 1.44 GB / 7.88 GB
  Binaries:
    Node: 17.5.0 - ~\scoop\apps\nodejs\current\node.EXE
    npm: 8.4.0 - ~\scoop\apps\nodejs\current\bin\npm.CMD
  Browsers:
    Chrome: 98.0.4758.102
    Edge: Spartan (44.19041.1266.0), Chromium (98.0.1108.56)
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.17
    @sveltejs/kit: next => 1.0.0-next.279
    svelte: ^3.44.0 => 3.46.4
```

### Severity

serious, but I can work around it

isaacfranco:
I saw that the strategy to associate a page route with it's shadow endpoint requires that they are just after the other after route sorting.

The problem is that when we have multiple endpoint-powered pages on the same folder, all it's endpoints stay before all the pages, and the association strategy doesn't work.

On the later part of packages/kit/src/core/create_manifest_data/index.js there is a comparator function that to the rout sorting, and there is a code that put pages after endpoints:
```javascript
	if (a.is_page !== b.is_page) {
		return a.is_page ? 1 : -1;
	}
```

if we remove that, the bug is gone, but I'm feeling that it will not respect the priorities especified on the documentation anymore.

Someone with any thoughts?

isaacfranco:
I can try to keep the sort comparator intact, but change this:
```javascript
while (i--) {
		const route = routes[i];
		const prev = routes[i - 1];

[...]

PR #4139, opened 2022-02-26
fix  Clashing endpoint-powered pages break route sorting(#4038)

aolose:
issue #4038

fix:
1. `shadow` of shadowed page not change  after merge matching page/endpoint pairs   
2. fallthrough response  didn't working at client-side  （client-side don't know if the returned content is a fallthrough）
3. request from fallthrough shadowed page not match the right endpoint

update: [`0e123e6`](https://github.com/sveltejs/kit/pull/4139/commits/0e123e63623f089c7813b45c6a5c03c6437a640c)

There are multiple  fallthrough shadow page , like this:
`[a]  -- fallthrough -> [b] --fallthrough --> [c] `
Before you need to request three times from client-side,
now you just need send one request.

update: [`7397a20`](https://github.com/sveltejs/kit/pull/4139/commits/7397a202c46566d1dfbfac84df0b0d437ab853fd)
add test :
1. client-side  fallthrough from shadow page to another shadow page
2. client-side fallthrough from shadow page  to another shadow page then fallthrough  to no-shadow page

```
cd packages\kit\test\apps\basics
npx  playwright test -g 'Shadow fallthrough shadow'
```

### Tests

### Changesets


packages/kit/src/core/dev/plugin.js
```diff
@@ -105,7 +105,9 @@ export async function create_plugin(config, cwd) {
 						}),
 						routes: manifest_data.routes.map((route) => {
 							if (route.type === 'page') {
+								/**@type import('types').SSRPage**/
```
Rich-Harris, line 108: `@type` doesn't do anything with `return`. You would need to declare a new variable or make it inline (`return /** @type {...} */ (...)`) but I don't think there's much value in it

```suggestion
```

packages/kit/src/runtime/client/renderer.js
```diff
@@ -718,7 +718,7 @@ export class Renderer {
 			if (cached) return cached;
 		}
 
-		const [pattern, a, b, get_params, has_shadow] = route;
+		const [pattern, a, b, get_params, has_shadow, routeKey] = route;
```
Rich-Harris, line 721: removes the `has_shadow` flag, per comment above — would need corresponding change on line 772. note that we use snake_case, not camelCase

```suggestion
		const [pattern, a, b, get_params, shadow_key] = route;
```

packages/kit/src/runtime/client/renderer.js
```diff
@@ -775,12 +775,12 @@ export class Renderer {
 						const res = await fetch(
 							`${url.pathname}${url.pathname.endsWith('/') ? '' : '/'}__data.json${url.search}`,
 							{
+								// @ts-ignore
 								headers: {
-									'x-sveltekit-load': 'true'
+									'x-sveltekit-load': routeKey
```
Rich-Harris, line 780:
```suggestion
									'x-sveltekit-load': shadow_key
```

packages/kit/src/runtime/client/renderer.js
```diff
@@ -791,7 +791,7 @@ export class Renderer {
 									state: this.current
 								};
 							}
-
+							if (res.status === 204) return;
```
Rich-Harris, line 794: what is this for?

packages/kit/src/runtime/server/index.js
```diff
@@ -151,6 +151,38 @@ export async function respond(request, options, state = {}) {
...
+											});
+										}
+									}
+									return response;
+								}
+							}
+						}
+					}
+				}
+
```
Rich-Harris, line 185: Could this not be replaced with something along these lines inside the `for` loop below?

```js
if (key && route.key !== key) continue;
```

packages/kit/types/internal.d.ts
```diff
@@ -71,7 +71,14 @@ export type CSRComponent = any; // TODO
 
 export type CSRComponentLoader = () => Promise<CSRComponent>;
 
-export type CSRRoute = [RegExp, CSRComponentLoader[], CSRComponentLoader[], GetParams?, HasShadow?];
+export type CSRRoute = [
+	RegExp,
+	CSRComponentLoader[],
+	CSRComponentLoader[],
+	GetParams?,
+	HasShadow?,
```
Rich-Harris, line 79:
```suggestion

```

packages/kit/types/internal.d.ts
```diff
@@ -211,6 +218,7 @@ export interface SSREndpoint {
 	type: 'endpoint';
 	pattern: RegExp;
 	params: GetParams;
+
```
Rich-Harris, line 221: not sure what these extra newlines are for?
```suggestion
```

packages/kit/types/internal.d.ts
```diff
@@ -236,7 +244,9 @@ export interface SSROptions {
 	dev: boolean;
 	floc: boolean;
 	get_stack: (error: Error) => string | undefined;
+
```
Rich-Harris, line 247:
```suggestion
```

packages/kit/types/internal.d.ts
```diff
@@ -236,7 +244,9 @@ export interface SSROptions {
...
 	handle_error(error: Error & { frame?: string }, event: RequestEvent): void;
+
```
Rich-Harris, line 249:
```suggestion
```

packages/kit/types/internal.d.ts
```diff
@@ -247,10 +257,13 @@ export interface SSROptions {
 	};
 	prefix: string;
 	prerender: boolean;
+
```
Rich-Harris, line 260:
```suggestion
```

packages/kit/types/internal.d.ts
```diff
@@ -247,10 +257,13 @@ export interface SSROptions {
...
 	read(file: string): Buffer;
+
```
Rich-Harris, line 262:
```suggestion
```

packages/kit/types/internal.d.ts
```diff
@@ -247,10 +257,13 @@ export interface SSROptions {
...
 	root: SSRComponent['default'];
 	router: boolean;
 	service_worker?: string;
+
```
Rich-Harris, line 266:
```suggestion
```

packages/kit/types/internal.d.ts
```diff
@@ -262,11 +275,13 @@ export interface SSROptions {
 		assets: string;
 		nonce: string;
 	}): string;
+
```
Rich-Harris, line 278:
```suggestion
```

Rich-Harris (changes requested):
Thank you — I left comments inline. This PR will need some tests before we can merge it!

packages/kit/src/runtime/client/renderer.js
aolose: When a request endpoint  from shadow page and endpoints return a fallthrough,
then client-side need to  know  the response is a fallthrough. 
We need a flag to tell the client-side that. so I return a 204 status to client-site.

packages/kit/src/runtime/server/index.js
aolose: Yes, it's better.

Rich-Harris (changes requested):
Thank you, but this will still need tests before it can be properly reviewed or merged

Rich-Harris:
Can you undo the `pnpm-lock.yaml` changes please?

Rich-Harris:
Thank you. I spent some time looking into this PR — unfortunately `master` has diverged quite a bit, so it isn't a straightforward merge. As explained in #4203 I think we need to use a simpler (if less sophisticated) implementation, since this is something of an edge case, so I'll close this in favour of that PR

aolose:
> Thank you. I spent some time looking into this PR — unfortunately `master` has diverged quite a bit, so it isn't a straightforward merge. As explained in #4203 I think we need to use a simpler (if less sophisticated) implementation, since this is something of an edge case, so I'll close this in favour of that PR

I agree width you, less sophisticated  is better.

PR #4140, opened 2022-02-26
restore `tabindex` after navigation

### LINKED ISSUE #3501: SvelteKit should not set tabIndex=-1
### Describe the bug

SvelteKit adds tabIndex=-1 to the document body in order to be able to manage focus upon navigation: https://github.com/sveltejs/kit/issues/307

This is not a "safe" change, in that it may impact the behavior of components on the page. In particular, this change means that some actions that would normally not trigger a `focus` event (like mousedown on some text) will now fire a `focus` event on the document body. These extra `focus` events may confuse code that does its own focus management (for example, a focus trap). They were the source of two different user-reported bugs in my component library: https://github.com/rgossiaux/svelte-headlessui/issues/36 and https://github.com/rgossiaux/svelte-headlessui/issues/39 

Looking at the motivation for introducing this originally, I don't see why it can't be replaced with something better. Can SvelteKit clean up after itself after navigation by removing the `tabIndex`? Alternatively, can it do something that doesn't require setting this at all? For example, Headless UI has a function to focus on the first focusable element within a container: https://github.com/tailwindlabs/headlessui/blob/cae976a18b7955670cd28fb1067a49917bda2c09/packages/%40headlessui-react/src/utils/focus-management.ts#L105 This works by using a CSS selector to identify potentially-focusable elements, then attempting to focus them one by one until a success.

### Reproduction

Not sure this needs a special test repro since this happens on every SvelteKit app, but I can create one if you disagree. 

The bugs that this caused can be reproduced at this commit: https://github.com/rgossiaux/svelte-headlessui/tree/ac1f86ac156a94c86e81647c55d8bb56c035edc6

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 4.11 GB / 64.00 GB
    Shell: 3.3.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 16.8.0 - /var/folders/rn/2kdgb0jj25q2v656j3qc_ywr0000gn/T/fnm_multishells/23841_1640400023703/bin/node
    npm: 7.21.0 - /var/folders/rn/2kdgb0jj25q2v656j3qc_ywr0000gn/T/fnm_multishells/23841_1640400023703/bin/npm
  Browsers:
    Chrome: 97.0.4692.99
    Safari: 15.2
```

### Severity

### Additional Information

_No response_

mrkishi:
Fixes #3501.

As noted in the issue, adding `tabindex="-1"` to `<body>` causes a few focus management problems with body "stealing" input from other elements (eg. scrolling overflown children).

Since it's only needed to reset the page's focus during navigation, we can safely [?] remove it afterwards.

[?]: I'm not able to test this on Apple devices.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -369,6 +369,11 @@ export class Renderer {
 			const { scroll, keepfocus } = opts;
 
 			if (!keepfocus) {
+				// reset page selection and focus
+				// we try to mimick browsers' behaviour as closely as possible by targetting the
+				// viewport, but unfortunately it's not a perfect match—eg. shift-tabbing won't
+				// immediately cycle from the end of the page
+				// see https://html.spec.whatwg.org/multipage/interaction.html#get-the-focusable-area
 				const root = document.documentElement;
 				const tabindex = root.getAttribute('tabindex');
 
@@ -376,6 +381,7 @@ export class Renderer {
 				root.tabIndex = -1;
 				root.focus();
 
+				// restore `tabindex` as to prevent the document from stealing input from elements
 				if (tabindex !== null) {
 					root.setAttribute('tabindex', tabindex);
 				} else {
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -369,8 +369,18 @@ export class Renderer {
 			const { scroll, keepfocus } = opts;
 
 			if (!keepfocus) {
+				const root = document.documentElement;
+				const tabindex = root.getAttribute('tabindex');
+
 				getSelection()?.removeAllRanges();
-				document.body.focus();
+				root.tabIndex = -1;
+				root.focus();
+
+				if (tabindex !== null) {
+					root.setAttribute('tabindex', tabindex);
+				} else {
+					root.removeAttribute('tabindex');
+				}
 			}
 
 			// need to render the DOM before we can scroll to the rendered elements
--- a/packages/kit/src/runtime/client/renderer.js
+++ b/packages/kit/src/runtime/client/renderer.js
@@ -369,11 +369,11 @@ export class Renderer {
 			const { scroll, keepfocus } = opts;
 
 			if (!keepfocus) {
-				// reset page selection and focus
-				// we try to mimick browsers' behaviour as closely as possible by targetting the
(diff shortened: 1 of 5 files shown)


geoffrich:
I think this is a good change - I can do some testing on an iPhone to make sure things are working as expected there.

Is there a reason you chose to start focusing `html` instead of `body`? I'm not sure it matters much, but was wondering if there was a reason to change the element we reset focus to.

We'll also need to update this line in kit.svelte.dev, which assumed `body` was focusable:

https://github.com/sveltejs/kit/blob/09bf92f853a75c887738e0bccd64dc68a81b6bc1/sites/kit.svelte.dev/src/lib/search/SearchBox.svelte#L54-L63

packages/kit/src/runtime/client/renderer.js
```diff
@@ -369,8 +369,18 @@ export class Renderer {
 			const { scroll, keepfocus } = opts;
 
 			if (!keepfocus) {
+				const root = document.documentElement;
```
benmccann, line 372: can we add some comment here about what this is doing and why

mrkishi: pushed some comments

packages/kit/src/runtime/client/renderer.js
```diff
@@ -369,8 +369,24 @@ export class Renderer {
 			const { scroll, keepfocus } = opts;
 
 			if (!keepfocus) {
+				// reset page selection and focus
+				// we try to mimick browsers' behaviour as closely as possible by targetting the
+				// viewport, but unfortunately it's not a perfect match—eg. shift-tabbing won't
+				// immediately cycle from the end of the page
+				// see https://html.spec.whatwg.org/multipage/interaction.html#get-the-focusable-area
```
benmccann, line 376: Thanks! I noticed a typo in "targeting". A few capital letters might help in identifying the beginning of sentences
```suggestion
				// Reset page selection and focus
				// We try to mimick browsers' behaviour as closely as possible by targeting the
				// viewport, but unfortunately it's not a perfect match — e.g. shift-tabbing won't
				// immediately cycle from the end of the page
				// See https://html.spec.whatwg.org/multipage/interaction.html#get-the-focusable-area
```

benmccann (approved):

bluwy (approved):

PR #4148, opened 2022-02-27
[docs] clarify what "it" refers to

pevey:
Edits ambiguous pronoun to make it clear that 'load' could run on either the client or the server.  I have seen a couple of tutorials recently with content creators stating the putting things like client secrets in this function within <script context="module"> is okay because it runs on the server.  If my understanding is correct, this is not safe and could possibly be made a little more clear.

### Tests

### Changesets

### DIFF
--- a/documentation/docs/03-loading.md
+++ b/documentation/docs/03-loading.md
@@ -26,7 +26,7 @@ If the data for a page comes from its endpoint, you may not need a `load` functi
 
 > Note the `<script context="module">` — this is necessary because `load` runs before the component is rendered. Code that is per-component instance should go into a second `<script>` tag.
 
-`load` is similar to `getStaticProps` or `getServerSideProps` in Next.js, except that `load` in SvelteKit runs on both the server and the client. In the example above, if a user clicks on a link to this page the data will be fetched from `cms.example.com` without going via our server.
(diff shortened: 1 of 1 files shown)


documentation/docs/03-loading.md
```diff
@@ -26,7 +26,7 @@ If the data for a page comes from its endpoint, you may not need a `load` functi
 
 > Note the `<script context="module">` — this is necessary because `load` runs before the component is rendered. Code that is per-component instance should go into a second `<script>` tag.
 
-`load` is similar to `getStaticProps` or `getServerSideProps` in Next.js, except that it runs on both the server and the client. In the example above, if a user clicks on a link to this page the data will be fetched from `cms.example.com` without going via our server.
+`load` is similar to `getStaticProps` or `getServerSideProps` in Next.js, except that `load` in SvelteKit runs on both the server and the client. In the example above, if a user clicks on a link to this page the data will be fetched from `cms.example.com` without going via our server.
```
benmccann, line 29:
```suggestion
`load` is similar to `getStaticProps` or `getServerSideProps` in Next.js, except that `load` runs on both the server and the client. In the example above, if a user clicks on a link to this page the data will be fetched from `cms.example.com` without going via our server.
```

PR #4155, opened 2022-02-28
port banner from svelte.dev

benmccann:
copying over from https://github.com/sveltejs/sites/pull/308


Rich-Harris (approved):

pngwn (approved):
lgtm! we don't need all of the CSS here but it is easy to strip it out when the time comes.

PR #4160, opened 2022-03-01
differentiate between `input` and `output` params inside `Load`

ivanhofer:
- [] Ideally, include a test that fails without this PR but passes with it.

### Tests

### Changesets

I want to be able to type the Params of a `Load` function that get's it's props from an Endpoint.
Currently the `InputParams` are always typed as `Record<string, any>`

Here is a simplified example:

`/routes/[id].ts`
```ts

export type Product = {
   name: string,
   color: string,
}

type Params = { id: string}
type OutputType = Product

export const get: RequestHandler<Params, OutputType> = ({ params }) => {
   return {
      body: {
         name: `${params.id} product`,
         color: 'green',
      }
   }
}
```

> `/routes/[id].svelte`
```svelte
   import type { Load } from '@sveltejs/kit'
   import type { get } from './[id]'

   type Params = { id: string }
   type InputProps = NonNullable<Awaited<ReturnType<typeof get>>['body']> // would be `Product`
   type OutputProps = Params & InputProps

   export const load: Load<Params, InputProps, OutputProps> = async ({
      params,
      props, // <= `InputProps`
      session,
   }) => {

      // ... some logic

      return {
         props: { // <= `OutputProps`
            product: props,
            id: params.id,
         },
      }
   }
</script>
```

This would be a breaking change


benmccann:
you'll need to run `pnpm format` because `lint` is currently failing

packages/kit/types/index.d.ts
```diff
@@ -115,8 +115,14 @@ export interface HandleError {
...
-export interface Load<Params = Record<string, string>, Props = Record<string, any>> {
-	(input: LoadInput<Params>): MaybePromise<Either<Fallthrough, LoadOutput<Props>>>;
+export interface Load<
+	Params = Record<string, string>,
+	InputProps = Record<string, any>,
+	OutputProps = Record<string, any>
+> {
+	(input: LoadInput<Params, InputProps>): MaybePromise<
+		Either<Fallthrough, LoadOutput<OutputProps>>
+	>;
```
Rich-Harris, line 125: It would be good if we could omit `OutputProps` in the common case where they're just being passed through, I think?

```suggestion
export interface Load<
	Params extends Record<string, string> = Record<string, string>,
	InputProps extends Record<string, any> = Record<string, any>,
	OutputProps extends Record<string, any> = InputProps
> {
	(input: LoadInput<Params, InputProps>): MaybePromise<
		Either<Fallthrough, LoadOutput<OutputProps>>
	>;
```

.changeset/yellow-geese-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+allow providing `InputProps` type argument for `Load`
```
Rich-Harris, line 5:
```suggestion
allow `InputProps` and `OutputProps` to be typed separately in `Load`
```

packages/kit/types/index.d.ts
Rich-Harris: In fact, I think this would make it a non-breaking change?

ivanhofer: Would still be a breaking change.
Before this PR `props` (`InputProps`) could hold any `Record`, If someone would have set the `Props` before (`OutputProps`), this would now become the `InputProps` so this could throw some errors.
Here is an example

Before:
```ts
   import type { Load } from '@sveltejs/kit'

   type Props = { value: string } 

   export const load: Load<any, Props> = async ({ props }) => {

      console.log(props.something) // valid

      return { /* ... */ }
   }
</script>
```
After:
```ts
   import type { Load } from '@sveltejs/kit'

   type Props = { value: string } 

   export const load: Load<any, Props> = async ({ props }) => {

      console.log(props.something) // not valid anymore because only `value` is a known property

      return { /* ... */ }
   }
</script>
```

.changeset/yellow-geese-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+allow `InputProps` and `OutputProps` to be typed separately in `Load`
```
Rich-Harris, line 5:
```suggestion
[breaking] allow `InputProps` and `OutputProps` to be typed separately in `Load`
```

packages/kit/types/index.d.ts
Rich-Harris: Ah yep, gotcha. I've updated the changeset to reflect that. Breakingness aside, does this suggestion make sense?

ivanhofer: Yes, they make sense. I haven't spotted the missing `extends` in the generics because I was lazy and just copied the `Props` code from somewhere else ^^
Maybe it would make sense to take another look and update places where the `extends` is missing in the current code base? Should I take a look and include the fixes in this PR?

Also having the same `InputProps` and `OutputProps` could be a valid case 👍

Rich-Harris (approved):

packages/kit/types/index.d.ts
Rich-Harris: There's probably some other places that should use `extends`, but I'll merge this PR in the meantime anyway. Thank you!

PR #4161, opened 2022-03-01
Combine `Router` and `Renderer`

Rich-Harris:
This has been on my wishlist for a while — taking the `Router` and `Renderer` classes, which are somewhat tightly coupled and have a confusing relationship to each other, and converting them to a single `client` entity. It _does_ mean a big ass `client.js` file (though we could probably extract out some utilities), but overall I think it's much easier to understand the flow of things when it's all colocated.

I opted for a revealing module pattern in this PR rather than a new `class` for a couple of reasons:

* It minifies better — the client is more than 2kb lighter before gzip (less of a difference after gzip but every little helps)
* We can have private methods and state without using syntax that Safari doesn't like
* No `this` nonsense — `$app/navigation` can simply re-export the public methods without needing to wrap them

There's still some tidying up to do, but on the whole this feels like a good direction to me — less source code, less built code, less indirection.

### Tests

### Changesets


benmccann (commented):
This still feels like throwing the baby out with the bathwater to me. I totally agree the current classes are terribly structured, but I think there are better solutions than putting everything into a single file. At the end of the day, I might just have to agree to disagree, but I'd love if we could get other opinions on this approach vs https://github.com/sveltejs/kit/pull/4101

packages/kit/src/runtime/client/client.js
```diff
@@ -0,0 +1,1249 @@
...
+			params,
+			stuff,
+			branch,
+			status,
+			error
+		});
+	}
+
+	/** @param {URL} url */
+	function _parse(url) {
```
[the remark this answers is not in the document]
Rich-Harris: changed to `get_navigation_intent`

packages/kit/src/runtime/client/client.js
```diff
@@ -0,0 +1,1249 @@
...
+		_hydrate: async ({ status, error, nodes, params }) => {
+			const url = new URL(location.href);
+
+			/** @type {Array<import('./types').BranchNode | undefined>} */
+			const branch = [];
+
+			/** @type {Record<string, any>} */
+			let stuff = {};
+
+			/** @type {import('./types').NavigationResult | undefined} */
```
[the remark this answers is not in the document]
Rich-Harris: Agree that the terminology could be tightened up a bit, though I think there's value in thinking of a 'load' as something that relates to a single node (i.e. layout/page component) in the branch, since that's where you might find a `load` function

packages/kit/src/runtime/client/client.js
```diff
@@ -0,0 +1,1249 @@
...
+// state we're navigating from
+
+/** @typedef {{ x: number, y: number }} ScrollPosition */
+/** @type {Record<number, ScrollPosition>} */
+let scroll_positions = {};
+try {
+	scroll_positions = JSON.parse(sessionStorage[SCROLL_KEY]);
+} catch {
+	// do nothing
+}
```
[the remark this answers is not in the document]
Rich-Harris: I've always felt iffy about IIFEs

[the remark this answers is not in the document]
benmccann: Another option to make it a `const` might be:
```
const scroll_positions = {};
try {
	Object.assign(scroll_positions, JSON.parse(sessionStorage[SCROLL_KEY]));
} catch {
	// do nothing
}
```

Rich-Harris:
I would also welcome more opinions on this vs #4101. To me though, this approach is pure upside.

It seems to me that the desire to keep the router and the 'renderer' (weird name) separate is based on a thinly-sketched idea about modularity, whether that means being able to use different routers inside SvelteKit (the value proposition for which hasn't been clearly articulated AFAICT; if there are shortcomings with SvelteKit's router we should address them, but the point at which we start talking about supporting non-filesystem-based routers is the point at which we start building an entirely different project with entirely different assumptions), or using SvelteKit's router inside other frameworks (which puts a massive burden on us to turn implementation details into stable and documented APIs).

Ultimately it may come down to a matter of philosophy. If SvelteKit were a laptop, I think of it as a MacBook Pro rather than a commodity device running a Linux distribution — extensible and flexible in the ways that matter, but very much geared towards productivity, delight, craftsmanship and cohesion rather than modularity for its own sake.

mrkishi (commented):
This is much easier to follow already!

I added some suggestions. They're mostly nitpicks, but might as well make use of the brand new file diff.

packages/kit/src/runtime/client/client.js
```diff
@@ -0,0 +1,1231 @@
...
+								url,
+								params,
+								stuff: node_loaded.stuff
+							});
+
+							if (error_loaded?.loaded?.error) {
+								continue;
+							}
+
+							if (error_loaded?.loaded?.loaded.stuff) {
```
mrkishi, line 715: That was my bad. 🤦 
```suggestion
							if (error_loaded?.loaded?.stuff) {
```

packages/kit/src/runtime/client/client.js
```diff
@@ -0,0 +1,1231 @@
...
+
+				if (!a.href) return;
+
+				const is_svg_a_element = a instanceof SVGAElement;
+				const url = get_href(a);
+				const url_string = url.toString();
+				if (url_string === location.href) {
+					if (!location.hash) event.preventDefault();
+					return;
+				}
```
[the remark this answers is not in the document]
Rich-Harris: I'm not sure I understand — why 1071 rather than here? I do think we could simplify it — `url.toString()` is just a roundabout way of saying `url.href`, and there's certainly no reason to create a new variable

```suggestion
				if (url.href === location.href) {
					if (!location.hash) event.preventDefault();
					return;
				}
```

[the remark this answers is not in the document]
mrkishi: That'd be just for the sake of allowing the other checks to run. If there's a `rel=external` or `download` attribute but it ended up here, couldn't `preventDefault()` mess things up? Unless it'd have no effect—in which case, my bad!

packages/kit/src/runtime/client/client.js
```diff
@@ -0,0 +1,1231 @@
...
+			navigation_result = await load_root_error_page({
+				status: 404,
+				error: new Error(`Not found: ${intent.url.pathname}`),
+				url: intent.url
+			});
+		}
+
+		if (!navigation_result) {
+			location.href = intent.url.href;
+			return;
```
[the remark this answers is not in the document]
Rich-Harris: There's a few places where we're doing a full page navigation — created a new `traditional_navigation` helper (open to better name suggestions) that encapsulates this behaviour, and replaced all `location.href = ...` occurrences with it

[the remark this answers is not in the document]
mrkishi: That makes sense! `traditional_navigation` sounds alright, but I had to think to connect the dots. Consider `navigate_natively`, `perform_native/browser_navigation` or something that sounds like an action.

packages/kit/src/runtime/client/client.js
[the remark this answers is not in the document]
Rich-Harris: ah, I see, that makes sense. (and it turns out i already applied that suggestion locally and didn't think i'd committed it, so i'll resolve this)

packages/kit/src/runtime/client/client.js
```diff
@@ -0,0 +1,1233 @@
...
+		/** @type {Array<(opts: { from: URL, to: URL | null, cancel: () => void }) => void>} */
+		before_navigate: [],
+
+		/** @type {Array<(opts: { from: URL | null, to: URL }) => void>} */
+		after_navigate: []
+	};
+
+	/** @type {import('./types').NavigationState} */
+	let current = {
+		// @ts-ignore - we need the initial value to be null
```
benmccann, line 81: is this true? could we not just do `new URL(window.location.href)`?

packages/kit/src/runtime/client/client.js
```diff
@@ -0,0 +1,1233 @@
...
+	const stores = {
+		url: notifiable_store({}),
+		page: notifiable_store({}),
+		navigating: writable(/** @type {import('types').Navigation | null} */ (null)),
+		session: writable(session),
+		updated: create_updated_store()
+	};
+
+	/** @type {{id: string | null, promise: Promise<import('./types').NavigationResult | undefined> | null}} */
+	const loading = {
```
benmccann, line 66: a better name for this might be `load_cache`

a related note - I just saw someone on Discord raise the issue that if they move their mouse between two links they will be prefetched over and over again. possibly we could make this more than a single-membered cache in the future

packages/kit/src/runtime/client/client.js
```diff
@@ -0,0 +1,1233 @@
...
+ *   trailing_slash: import('types').TrailingSlash;
+ * }} opts
+ * @returns {import('./types').Client}
+ */
+export function create_client({ target, session, base, trailing_slash }) {
+	/** @type {Map<string, import('./types').NavigationResult>} */
+	const cache = new Map();
+
+	/** @type {Set<string>} */
+	const invalid = new Set();
```
benmccann, line 55: perhaps `invalidated` would be slightly clearer?

packages/kit/src/runtime/client/client.js
```diff
@@ -0,0 +1,1233 @@
...
+	let $session;
+
+	let ready = false;
+	stores.session.subscribe(async (value) => {
+		$session = value;
+
+		if (!ready) return;
+		session_id += 1;
+
+		const intent = get_navigation_intent(new URL(location.href));
```
benmccann, line 108: should we just call `invalidate` here?

packages/kit/src/runtime/client/client.js
```diff
@@ -0,0 +1,1233 @@
...
+	 * @param {import('./types').NavigationIntent} intent
+	 * @param {string[]} redirect_chain
+	 * @param {boolean} no_cache
+	 * @param {{hash?: string, scroll: { x: number, y: number } | null, keepfocus: boolean}} [opts]
+	 */
+	async function update(intent, redirect_chain, no_cache, opts) {
+		const current_token = (token = {});
+		let navigation_result = await get_navigation_result(intent, no_cache);
+
+		if (!navigation_result && intent.url.pathname === location.pathname) {
```
benmccann, line 196: something that might be worth a comment - I'm not understanding the second half of this check. if you want to navigate to the page we're already on, isn't that essentially just a no-op? why do we want to show a 404 page in that case?

packages/kit/src/runtime/client/client.js
```diff
@@ -0,0 +1,1233 @@
...
+			});
+		}
+
+		if (!navigation_result) {
+			await traditional_navigation(intent.url);
+			return; // unnecessary, but TypeScript prefers it this way
+		}
+
+		// abort if user navigated during update
+		if (token !== current_token) return;
```
benmccann, line 210: I noticed when working on https://github.com/sveltejs/kit/pull/4101 that this token stuff looks broken in `master`. I didn't attempt to fix it and it looks like it's not addressed here either. Assuming I haven't missed something, `current_token` is only ever set to `{}` and so this is a no-op check

packages/kit/src/runtime/client/client.js
Rich-Harris: Conceptually, upon hydration we're navigating from `null` to `location.href`, and this is reflected in places like https://github.com/sveltejs/kit/blob/7373dd5396d6998749ac1766e359cfa28b523888/packages/kit/src/runtime/client/client.js#L396 and https://github.com/sveltejs/kit/blob/7373dd5396d6998749ac1766e359cfa28b523888/packages/kit/src/runtime/client/client.js#L564-L568

packages/kit/src/runtime/client/client.js
mrkishi: `update()` can be called by a different navigation/redirect between the time it started and the time this check is made, meaning `token` will point to a different `{}` than this invocation's `current_token`, and in that case we abort (because this call to `update()` was superseded).

packages/kit/src/runtime/client/client.js
Rich-Harris: It doesn't have quite the right semantics — if `session` changes we want to re-run `load` functions that depend on `session`, not those that `fetch(location.href)`

packages/kit/src/runtime/client/client.js
Rich-Harris: renamed. re multiple prefetches, the current behaviour is by design, but if they want to open an issue then we can discuss further

packages/kit/src/runtime/client/client.js
Rich-Harris: added a comment

packages/kit/src/runtime/client/client.js
Rich-Harris: it's working as intended as far as I can see. consider this simplified version:

```js
let token;

async function log(message) {
  const current_token = (token = {});
  await Promise.resolve();
  if (token !== current_token) return;

  console.log(message);
}

// only b is logged; a is aborted
log('a');
log('b');
```

benmccann: Oh. I see. That's a bit of a subtle thing that relies on `===` checking that they're the same instance. I was just looking at it and seeing that they were both always `{}` and doing more of a checking that the values were equivalent in my head. Maybe we could either add a comment or rename to `current_token_instance` and `token_instance`?

Rich-Harris: I think it's clear enough, given the comment immediately above the check

packages/kit/src/runtime/client/client.js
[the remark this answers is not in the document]
Rich-Harris: I like `native` — have renamed it to `native_navigation`

PR #4182, opened 2022-03-02
add `svelte-kit sync`

### LINKED ISSUE #4165: svelte-kit sync
### Describe the problem

In #4120, we introduce _generated types_, which give you type safety when using `params` inside pages and endpoints. (In future, we want to extend that to props as well.)

One minor wrinkle is that the types aren't generated until you run `svelte-kit dev` or `svelte-kit build`, which means that if you were to (for example) typecheck in CI _before_ building the app, it would fail. 

It would be good to collate all the work that takes the filesystem as input (i.e. generating `manifest.js`, `Root.svelte` and now the route-level types) into a new function — let's call it `sync` — and expose it via the CLI as `svelte-kit sync`. (Open to alternative names but this one seems good enough to me.)

We could then add this to the default template as a postinstall script:

```js
// package.json
"scripts": {
  // ...
  "postinstall": "svelte-kit sync"
}
```

### Importance

would make my life easier

Rich-Harris:
Closes #4165. This adds a `sync` module responsible for writing stuff to `outDir` — `generated/manifest.js`, `generated/root.svelte`, `runtime`, `types` and `tsconfig.json`.

It also adds a `svelte-kit sync` command that runs when you first set up a new project, to ensure that types are set up and correct before you run typechecking.

I've been meaning to tidy up this corner of the codebase anyway, and this was a nice opportunity to do it.

### Tests

### Changesets


packages/create-svelte/templates/default/package.template.json
```diff
@@ -5,7 +5,8 @@
 		"dev": "svelte-kit dev",
 		"build": "svelte-kit build",
 		"package": "svelte-kit package",
-		"preview": "svelte-kit preview"
+		"preview": "svelte-kit preview",
+		"prepare": "svelte-kit sync"
```
benmccann, line 9: I would think that we should call them both `prepare` or call them both `sync`

benmccann:
Is this command cheap enough that we can run it whenever we want to run `pnpm check`? Or would it be helpful to also have a `svelte-kit tsconfig` command that just generates that file? (https://github.com/sveltejs/kit/pull/4190#discussion_r818634026)

packages/create-svelte/templates/default/package.template.json
Conduitry: Calling the script `prepare` is required so that it gets automatically run at the correct time by npm. I don't know that we also want that to inform what we call the CLI command.

benmccann: Ah, I thought `build` invoke `sync` via code: https://github.com/sveltejs/kit/pull/4182/files#diff-09b70580810ae66bf783d467b3ea4f54cd1ac4cb57532b38b788145e27cc0983R26

Does this mean it will be run twice?

mrkishi: `prepare` only runs (automatically) at certain points, like `install`: https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-scripts.

Rich-Harris:
For a large-ish app it takes about 100ms on my machine — it's basically free

PR #4185, opened 2022-03-02
ignore .css files when building route manifest - 

### LINKED ISSUE #3997: CSS file present in src/routes/ breaks production build
### Describe the bug

If a non-empty `.css` file is present in the `src/routes/` directory with the same name as a `.svelte` file in the same directory, we receive an error in build.

### Reproduction

To reproduce using the SvelteKit skeleton project:

1. `npm init svelte@next sveltekit-css-repro` (select skeleton, no to all options)
2. `cd sveltekit-css-repro/`
3. `npm install`
4. `npm run build` (succeeds)
5. `echo "h1 { font-size: 1.5em; }" > src/routes/index.css`
6. `npm run build` (fails, log below)
7. `mv src/routes/{index,index-styles}.css`
8. `npm run build` (succeeds)

### Logs

```shell
$ npm run build

> sveltekit-css-repro@0.0.1 build
> svelte-kit build

vite v2.8.4 building for production...
✓ 13 modules transformed.
.svelte-kit/output/client/_app/manifest.json                    1.15 KiB
.svelte-kit/output/client/_app/layout.svelte-ddbe22e6.js        0.53 KiB / gzip: 0.35 KiB
.svelte-kit/output/client/_app/error.svelte-60b5f229.js         1.56 KiB / gzip: 0.75 KiB
.svelte-kit/output/client/_app/pages/index.svelte-4fdc3974.js   0.80 KiB / gzip: 0.47 KiB
.svelte-kit/output/client/_app/start-0e28adf2.js                23.61 KiB / gzip: 8.09 KiB
.svelte-kit/output/client/_app/chunks/vendor-f0095a1c.js        7.21 KiB / gzip: 2.96 KiB
vite v2.8.4 building SSR bundle for production...
✓ 12 modules transformed.
.svelte-kit/output/server/manifest.json                    0.87 KiB
.svelte-kit/output/server/app.js                           64.10 KiB
.svelte-kit/output/server/entries/pages/layout.svelte.js   0.24 KiB
.svelte-kit/output/server/entries/pages/error.svelte.js    0.72 KiB
.svelte-kit/output/server/entries/pages/index.svelte.js    0.32 KiB
.svelte-kit/output/server/chunks/index-2dc61825.js         2.29 KiB
> Cannot read properties of undefined (reading 'file')
    at file:///home/yann/git/sveltekit-css-repro/node_modules/@sveltejs/kit/dist/chunks/index4.js:78:108
    at Array.map (<anonymous>)
    at generate_manifest (file:///home/yann/git/sveltekit-css-repro/node_modules/@sveltejs/kit/dist/chunks/index4.js:71:14)
    at build (file:///home/yann/git/sveltekit-css-repro/node_modules/@sveltejs/kit/dist/chunks/index3.js:637:46)
    at async file:///home/yann/git/sveltekit-css-repro/node_modules/@sveltejs/kit/dist/cli.js:1054:23
```

### System Info

```shell
System:
  OS: Linux 5.10 Debian GNU/Linux 10 (buster) 10 (buster)
  CPU: (8) x64 Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz
  Memory: 8.23 GB / 15.60 GB
  Container: Yes
  Shell: 5.0.3 - /bin/bash
Binaries:
  Node: 17.5.0 - ~/.nvm/versions/node/v17.5.0/bin/node
  Yarn: 1.22.17 - ~/.nvm/versions/node/v17.5.0/bin/yarn
  npm: 8.5.1 - ~/.nvm/versions/node/v17.5.0/bin/npm
npmPackages:
  @sveltejs/kit: next => 1.0.0-next.278
  svelte: ^3.44.0 => 3.46.4
```

### Severity

### Additional Information

Further discussion in Svelte Discord thread https://discord.com/channels/457912077277855764/944288129114861588. Thanks to nudge#9063 for the help diagnosing!

A workaround is fairly trivial, ensure `.css` files inside `src/routes/**/*` do not match the filename of a `.svelte` file in the same directory.

The error is unclear, though. In a medium to large sized SvelteKit project, this could present a challenge upgrading.

Rich-Harris:
fixes #3997. I think we can punt on the question of whether we want to treat all `.js`/`.ts` files similarly — I think the `.css` case is fairly unambiguous. If we decide to make it more expansive in future, `.css` files would be included in that, so this is forward-compatible

then again this could be a can of worms if people start asking about `.scss` and what-have-you...

### Tests

### Changesets

### DIFF
--- a/.changeset/sweet-houses-worry.md
+++ b/.changeset/sweet-houses-worry.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Ignore .css files when building route manifest
--- a/packages/kit/src/core/create_manifest_data/index.js
+++ b/packages/kit/src/core/create_manifest_data/index.js
@@ -76,6 +76,10 @@ export default function create_manifest_data({
 
 			const ext = config.extensions.find((ext) => basename.endsWith(ext)) || path.extname(basename);
 
+			// treat .css files as a special case (https://github.com/sveltejs/kit/issues/3997)
+			// TODO do we want to treat all non-.js/.ts files similarly?
+			if (ext === '.css') return;
+
 			const name = ext ? basename.slice(0, -ext.length) : basename;
 
 			// TODO remove this after a while
--- a/packages/kit/test/apps/basics/src/routes/index.css
+++ b/packages/kit/test/apps/basics/src/routes/index.css
@@ -1,0 +1,4 @@
+/* https://github.com/sveltejs/kit/issues/3997 */
+h1 {
+	color: blue;
+}
--- a/packages/kit/test/apps/basics/src/routes/index.svelte
+++ b/packages/kit/test/apps/basics/src/routes/index.svelte
@@ -11,6 +11,8 @@
 </script>
 
 <script>
+	import './index.css';
+
 	/** @type {number} */
 	export let answer;
 </script>


Conduitry:
I'm not sure where the line is between stuff that should happen in the default https://kit.svelte.dev/docs/configuration#routes function vs stuff that should be stripped out in a separate step.

benmccann:
What happens to file types besides `.js`/`.ts` and `.css`/`.scss`/etc.? Do we want to generate routes for `.json` files? I don't know what we do today, but I would expect that we don't and that they should just go into `static/`. It seems potentially safer to me to whitelist `.js` / `.ts` than to blacklist `.js`?

Rich-Harris:
If we include rather than exclude, then we need a way to make it extensible to handle edge cases like `.res` (ReScript) files and whatever other wacky shit people want to use

> Do we want to generate routes for `.json` files?

no, they will always belong in `static`

benmccann:
Should I ask if we also need to exclude image files? I have a feeling we'll need a config option either way whether it's to add stuff like `.res` or `.scss`. It seems less common to use other JS formats like `.res` than the combined use of all other extensions like `.scss`, so it might be the option that works out-of-the-box more of the time.

yanneves:
Just my two cents as another possible avenue, we could tidy up the error to say "hey, that .css file doesn't belong under routes/, consider static/ or somewhere outside routes/ if you want to import it"

The route manifest and this directory are doing a lot of heavy lifting, and it wasn't obvious from the outset that I was interfering with it - in retrospect, it makes sense why this broke, and I think noping this use case doesn't detract from the developer experience

I'd like to think that could also reduce collateral continuing to implement advanced features (looking at you, shadow endpoints) with this directory

Rich-Harris:
Opened https://github.com/sveltejs/kit/pull/4197 as a possible alternative

Rich-Harris:
Closing this in favour of #4197

PR #4190, opened 2022-03-03
Upgrade to TypeScript 4.6

benmccann:
I had to add `pnpm build` to run when running `pnpm check` for the test projects so that the generated `tsconfig.json` files would exist. Apparently TypeScript 4.5 didn't care if the files were missing. We can replace this with `pnpm sync` once it exists (https://github.com/sveltejs/kit/pull/4182)


dummdidumm (approved):

packages/kit/package.json
```diff
@@ -62,13 +62,13 @@
...
-		"check:integration:options-2": "cd test/apps/options-2 && pnpm check",
+		"check:integration:amp": "cd test/apps/amp && pnpm build && pnpm check",
+		"check:integration:basics": "cd test/apps/basics && pnpm build && pnpm check",
+		"check:integration:options": "cd test/apps/options && pnpm build && pnpm check",
+		"check:integration:options-2": "cd test/apps/options-2 && pnpm build && pnpm check",
 		"check:prerendering": "npm run check:prerendering:basics && npm run check:prerendering:options",
-		"check:prerendering:basics": "cd test/prerendering/basics && pnpm check",
-		"check:prerendering:options": "cd test/prerendering/options && pnpm check",
+		"check:prerendering:basics": "cd test/prerendering/basics && pnpm build && pnpm check",
+		"check:prerendering:options": "cd test/prerendering/options && pnpm build && pnpm check",
```
Rich-Harris, line 71: we don't want to build these projects every time we check them, it'll slow things right down. in CI we do `pnpm -r build` before running the checks. if we really want to eliminate the chance of checking prematurely in _all_ cases then we should use `svelte-kit sync` (#4182)

packages/kit/src/core/tsconfig.js
```diff
@@ -29,9 +29,9 @@ export function generate_tsconfig(config) {
 	/** @type {string[]} */
 	const include = [];
 	dirs.forEach((dir) => {
-		include.push(config_relative(`${dir}/**/*.js`));
-		include.push(config_relative(`${dir}/**/*.ts`));
-		include.push(config_relative(`${dir}/**/*.svelte`));
+		include.push(config_relative(`./${dir}/**/*.js`));
+		include.push(config_relative(`./${dir}/**/*.ts`));
+		include.push(config_relative(`./${dir}/**/*.svelte`));
```
Rich-Harris, line 34: why do we need this?

packages/kit/src/packaging/test/fixtures/assets/jsconfig.json
```diff
@@ -3,8 +3,8 @@
 		"checkJs": true,
 		"baseUrl": ".",
 		"paths": {
-			"$lib/*": ["src/lib/*"]
+			"$lib/*": ["./src/lib/*"]
 		}
 	},
-	"include": ["src/**/*.d.ts", "src/**/*.js", "src/**/*.svelte"]
+	"include": ["./src/**/*.d.ts", "./src/**/*.js", "./src/**/*.svelte"]
```
Rich-Harris, line 9: IIUC the `./` prefix is unnecessary if you specify `baseUrl`

dummdidumm: Oh you're right, yes. They are unnecessary then.

packages/kit/src/core/tsconfig.js
dummdidumm: Probably the same reasoning as for adding the `./` prefix in the other places

Rich-Harris:
```suggestion
		include.push(config_relative(`${dir}/**/*.js`));
		include.push(config_relative(`${dir}/**/*.ts`));
		include.push(config_relative(`${dir}/**/*.svelte`));
```

packages/kit/src/packaging/test/fixtures/assets/jsconfig.json
```diff
@@ -3,8 +3,8 @@
 		"checkJs": true,
 		"baseUrl": ".",
 		"paths": {
-			"$lib/*": ["src/lib/*"]
+			"$lib/*": ["./src/lib/*"]
```
Rich-Harris, line 6:
```suggestion
			"$lib/*": ["src/lib/*"]
```

packages/kit/src/packaging/test/fixtures/assets/jsconfig.json
Rich-Harris, line 9:
```suggestion
	"include": ["src/**/*.d.ts", "src/**/*.js", "src/**/*.svelte"]
```

packages/kit/test/apps/amp/tsconfig.json
```diff
@@ -2,8 +2,8 @@
 	"compilerOptions": {
 		"noEmit": true,
 		"paths": {
-			"$lib": ["src/lib"],
-			"$lib/*": ["src/lib/*"],
+			"$lib": ["./src/lib"],
+			"$lib/*": ["./src/lib/*"],
```
Rich-Harris, line 6:
```suggestion
			"$lib": ["src/lib"],
			"$lib/*": ["src/lib/*"],
```

packages/kit/package.json
benmccann: `pnpm -r build` only builds the code. It doesn't build these test projects. I thought it'd be more of a negative impact to build these during `pnpm build`

Yeah, my plan was to switch these to `sync` as soon as it's available

packages/kit/src/packaging/test/fixtures/assets/jsconfig.json
benmccann: Just to be clear, you want me to add `baseUrl` everywhere instead? TypeScript 4.6 will enforce one or the other

packages/kit/package.json
Rich-Harris: > It doesn't build these test projects

it does!

packages/kit/src/packaging/test/fixtures/assets/jsconfig.json
Rich-Harris: I'm saying that in cases where there's already a `baseUrl` (directly or via `extends`) there's no need to add `./` prefixes

packages/kit/src/core/tsconfig.js
benmccann: removed

packages/kit/package.json
benmccann: Ah, you're right. I was thinking of `pnpm build` which has a filter on it.

Anyway, will we just switch these to `sync` when it's available? I'm not sure it matters what happens in the meantime since it will be in this state for only a short time. I can remove `pnpm build` here if you want and then we will just add it back as `pnpm sync`?

benmccann: Ok. Can we just leave this as `build` then and have your PR update it to `sync`? If I try to remove it then I'm going to have to update the `.github/workflows` just for you to remove my changes in your PR, which I think would be more work for both of us

benmccann: updated to use `sync`

PR #4191, opened 2022-03-03
Should fix 4177 navigation to pages with iframe with relative urls are broken

### LINKED ISSUE #4177: navigation to pages with iframes with relative URLs are broken after #4070
### Describe the bug

* You have a page at "/foo/bar" which loads an iframe whose source is the relative URL "baz" (the absolute URL is thus "/foo/baz")
* You navigate from "/foo" to "/foo/bar"
* The iframe loads with the wrong relative URL--it should load with the relative URL of the page it's on (so "/foo/baz") but instead it loads relative to the original page ("/baz")
* Refresh and everything works; the error is only on navigation.

This is a regression. I bisected the kit version responsible for this and it's 1.0.0-next.285 which looks to correspond to this PR: #4070

### Reproduction

You can repro on this branch: https://github.com/rgossiaux/svelte-headlessui/pull/74

* Clone the repo
* `npm install`
* `git checkout upgrade-kit` 
* `npm run dev`
* Go to the `/docs` home page
* Click on Dialog on the sidebar

The iframe will 404. After a refresh it will load.

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 5.59 GB / 64.00 GB
    Shell: 3.3.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 16.8.0 - /var/folders/rn/2kdgb0jj25q2v656j3qc_ywr0000gn/T/fnm_multishells/2969_1644654672256/bin/node
    npm: 7.21.0 - /var/folders/rn/2kdgb0jj25q2v656j3qc_ywr0000gn/T/fnm_multishells/2969_1644654672256/bin/npm
  Browsers:
    Chrome: 98.0.4758.109
    Safari: 15.2
  npmPackages:
    @sveltejs/kit: ^1.0.0-next.285 => 1.0.0-next.285
    svelte: ^3.44.0 => 3.44.3
```

### Severity

blocking an upgrade

### Additional Information

_No response_

PH4NTOMiki:

### Tests

### Changesets

Fixes #4177
I just did what @Rich-Harris wrote in https://github.com/sveltejs/kit/issues/4177#issuecomment-1057250020 that should fix it.

### DIFF
--- a/.changeset/sour-needles-compete.md
+++ b/.changeset/sour-needles-compete.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Change URL just before running root.\$set
--- a/.changeset/sour-needles-compete.md
+++ b/.changeset/sour-needles-compete.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Change URL just before running root.\$set
+Update history immediately before updating DOM
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -187,7 +187,7 @@ export function create_client({ target, session, base, trailing_slash }) {
 	 * @param {import('./types').NavigationIntent} intent
 	 * @param {string[]} redirect_chain
 	 * @param {boolean} no_cache
-	 * @param {{hash?: string, scroll: { x: number, y: number } | null, keepfocus: boolean}} [opts]
+	 * @param {{hash?: string, scroll: { x: number, y: number } | null, keepfocus: boolean, details: { replaceState: boolean, state: any } | null}} [opts]
 	 */
 	async function update(intent, redirect_chain, no_cache, opts) {
 		const current_token = (token = {});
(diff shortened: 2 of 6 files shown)


Rich-Harris:
Thanks — can you add a test please?

PH4NTOMiki:
@Rich-Harris But actually I think this PR may make tests more flaky(and that is the reason we did #4070), is there any way to solve this without doing `pushState` before `root.$set`?
What does it even do, I don't really understand?

Rich-Harris:
Since the DOM update happens in the same turn of the event loop as `history.pushState`, I don't think this will increase test flakiness — AFAICT Playwright waits until the _next_ tick before resolving `waitForNavigation` promises. I added a test, and everything is still green (except windows, which is timing out for unrelated reasons), so I think we're good to go

.changeset/sour-needles-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Change URL just before running root.\$set
```
Rich-Harris, line 5:
```suggestion
Update history immediately before updating DOM
```

Rich-Harris (approved):

Issue #3792, opened 2022-02-09
Option to make prerender opt-out rather than opt-in


madeleineostoja:
### Describe the problem

For large headless / static sites with a bit of api glue code, marking every single route with `export const prerender = true` is both unergonomic and unintuitive, considering other major serverless frameworks (eg: Next) automatically prerender all routes when possible.

A common pattern is to have all routes prerendered, and then an `api/` route or similar for endpoints that would get translated to serverless functions you can request to.

### Describe the proposed solution

There should be an option to make prerendering opt-out rather than opt-in for serverless adapters. Ideally an adapter like Vercel should attempt to prerender all routes by default, while keeping endpoints as serverless functions. That would be a fairly major breaking change, so making this a configuration option (false by default) would be a non-breaking and huge erganomic upgrade. Though I still think the default behaviour is odd, given kit's emphasis on performance and DX.

### Alternatives considered

1. Mark every route with `export const prerender = true`, unintuitive and prone to user-error
2. Use `adapter-static` and manually deploy serverless functions for api routes / endpoints to provider of choice, largely negating the whole benefit of Sveltekit's deep serverless integration

### Importance

nice to have

Issue #2271, opened 2021-08-23
`files` in `$service-worker` should contain prerendered files


frederikhors:
Using `adapter-static` and the code:

```ts
import { files } from '$service-worker';
```

I expect to find also the `<NAME>.html` fallback file listed in `files`.

Am I wrong?

ignatiusmb:
Please provide a complete and minimal reproducible example of what happen and should've happen.

Mlocik97:
this is specific for adapter, so it's not good if it's provided by `$service-worker`, why You don't simply add manually that file to it? It's just one line of code.

benmccann:
For fallback we can add just the one. For prerendered files there might be too many to do this. We'd have to have the adapter generate the manifest after prerendering I guess?

benmccann:
SvelteKit's own service worker should contain prerendered files now. Unfortunately, it still doesn't happen early enough in the build process that plugins like `vite-plugin-pwa` could pick them up

Issue #923, opened 2021-04-07
Get access to the list of pages generated 


wighawag:
**Is your feature request related to a problem? Please describe.**
on my fully static website I usually have a service worker that cache the pages on first load

Currently svelte-kit gives us a list of build files but not any list of pages

**Describe the solution you'd like**
it would be great if there was a way to get the list of pages (dynamic ones would be given their non actualised params)

**Describe alternatives you've considered**
One workaround is to read from the file system the routes, but this duplicate what svelte kit is already doing and would be brittle to emulate

**How important is this feature to you?**
This is important as otherwise I have to do some workaround as mentioned above

devgar:
Routify uses a `layout` store which instantiates more immediate parent layout. his layout store contains a list of children `nodes` which represents routes.

This is how a navbar can be auto-generated by routify:
```html
<!-- src/pages/_layout.svelte --> <script> import { layout } from '@roxi/routify' </script> {#each $layout.children as node} <a href="{node.path}">{node.title}</a> {/each}
```

lucianoratamero:
hi everyone! 

@wighawag I'm having the same issues right now. I tried many ways to work around it, to no success =/
after digging around the source code, I found this [`build_service_worker`](https://github.com/sveltejs/kit/blob/353afa1bbc8799abc9247544422dd9b2f240f878/packages/kit/src/core/build/index.js#L465) function, which is the one that provides [these variables](https://kit.svelte.dev/docs#modules-$service-worker) to be used on the service worker. it writes a file at build time that makes them available, then goes forward to build the service worker file for production.

it didn't seem to be that much of a hassle to add the routes to the file as well, since they're already available in the `manifest` object that is injected into the function. as a proof of concept, I was able to get everything working. :]

my solution was to add the following code to [the line 483 of build.js](https://github.com/sveltejs/kit/blob/353afa1bbc8799abc9247544422dd9b2f240f878/packages/kit/src/core/build/index.js#L483) (985 of the `node_modules/@svelte/kit/dist/chunks/index4.js`, if you're using the npm installed package):

```
export const pageRoutes = [
	${manifest.routes
		.filter((route) => route.type === "page")
		.map(
			(page) =>
			`${s(`${page.path.endsWith("/") ? page.path : page.path + "/"}`)}`
			)
			.join(",\n\t\t\t\t")}
];
```

I'd be happy to discuss this further, and provide a PR, if everyone is fine with this proposal!

oneezy:
[...]

PR #4192, opened 2022-03-03
Prerender during build

### LINKED ISSUE #4112: Prerender during build rather than adapt
### Describe the problem

Currently, prerendering happens inside adapters. This creates various minor problems:

* `svelte-kit preview` doesn't use prerendered pages — loss of fidelity, and slow performance that's unreflective of production build
* adapters have to be careful to prerender before generating manifests
* prerendering doesn't happen if you haven't configured an adapter, or you're doing `svelte-kit build` locally with `adapter-auto` (unless you prefix it like `VERCEL=1 npm run build`)

I think prerendering should happen during build, _before_ the adapter runs. 

The fly in the ointment is `adapter-static`, which needs to specify that all pages should be prerendered (not just those explicitly marked as capable of being prerendered) unless a fallback is specified. A `config.kit.prerender.all` or `config.kit.prerender.default` option could stand in for that, and could also mean 'prerender everything without `export const prerender = false`. Together with a `builder.prerenderFallback` method, we could solve the `adapter-static` problem in a way that gives everyone a bit more flexibility. (Would probably need to make `adapter-static` aware of the `all`/`default` option via the builder so that it can instruct people to update their config accordingly.)

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

Rich-Harris:
Closes #4112
Closes https://github.com/sveltejs/kit/issues/3792
Closes https://github.com/sveltejs/kit/issues/2271
Closes https://github.com/sveltejs/kit/issues/923

* [x] Add new `config.kit.prerender.default` option (defaults to `false`)
* [x] Prerender during `build`, and pass `builder.prerendered` object to adapters
* [x] Implement `builder.writePrerendered(dest, { fallback })` method
* [x] Include `default` option in `builder.prerendered`, so e.g. `adapter-static` can warn if `default` was `false` when there's no `fallback`
* [x] Use prerendered pages in `svelte-kit preview`
* [x] Make `prerendered` available to service workers
* [x] Update all adapters
* [x] Update docs

### Tests

### Changesets


Rich-Harris:
The UI for linking issues is broken, because GitHub is a modern app built with 00's technology, but this will close the following issues:

* #4112
* #3792
* #2271
* #923

documentation/docs/13-configuration.md
```diff
@@ -224,6 +224,7 @@ See [Prerendering](/docs/page-options#prerender). An object containing zero or m
 
 - `concurrency` — how many pages can be prerendered simultaneously. JS is single-threaded, but in cases where prerendering performance is network-bound (for example loading content from a remote CMS) this can speed things up by processing other tasks while waiting on the network response
 - `crawl` — determines whether SvelteKit should find pages to prerender by following links from the seed page(s)
+- `default` — set to `true` to prerender every page without `export const prerender = false`
```
benmccann, line 227: I wonder if we should combine `default` and `enabled` into a single option. E.g. with values like `true | false | 'never'` or `'default-true' | 'default-false' | 'never'`

Rich-Harris: I think I prefer the booleans, personally

PR #4197, opened 2022-03-03
add config.kit.endpointExtensions option

Rich-Harris:
Alternative to #4185 — prevents files other than `.js` and `.ts` files from being treated as endpoints, unless you specify `endpointExtensions`

### Tests

### Changesets


Conduitry:
I do feel a bit better about this than the other option. Do we want to rename the existing `extensions` option to `pageExtensions` or something similar? Or move both under a single object?

Rich-Harris:
The wrinkle there is that `config.extensions` is used by other tooling, like mdsvex

Conduitry:
Oh. Hm. And it lives outside of `kit`. Never mind, I don't have a good suggestion about this then.

documentation/docs/13-configuration.md
```diff
@@ -135,6 +136,10 @@ When pages are prerendered, the CSP header is added via a `<meta http-equiv>` ta
 
 > When `mode` is `'auto'`, SvelteKit will use nonces for dynamically rendered pages and hashes for prerendered pages. Using nonces with prerendered pages is insecure and therefore forbidden.
 
+### endpointExtensions
+
+An array of valid file extensions for endpoint modules.
```
Conduitry, line 141: "Valid" seems like a strange word to be using here to me. It's not like Kit will throw an error for other ones, will it? Just ignore them. I don't have another word to suggest at the moment.

Rich-Harris:
```suggestion
An array of file extensions that SvelteKit will treat as endpoints. Files with extensions that match neither `config.extensions` nor `config.kit.endpointExtensions` will be ignored by the router.
```

benmccann (approved):

Issue #4186, opened 2022-03-02
SvelteKit fails to build if user provides default error page


benmccann:
### Describe the bug

A user reported that their project is failing when they run `npm run build`. I did some preliminary investigation and found that it's failing because it can't find the default error page in the manifest. If I delete `src/routes/error.svelte` from their project then the build passes. I guess Vite doesn't see the default error page being used or imported anywhere in this case

### Reproduction

https://github.com/deozza/appquarium-admin-sveltekit

I made a fork which runs the latest SvelteKit and get the same issue: https://github.com/benmccann/appquarium-admin-sveltekit

### Logs

`build-server` is generating the error:
```shell
> Cannot read property 'file' of undefined
```

### System Info

```shell
latest
```

### Severity

blocking all usage of SvelteKit

PR #4200, opened 2022-03-03
fix issue #4186

aolose:
issue #4186

fix bug:

You can see the `.svelte-kit/runtime/components/error.svelte`   add the `resolved` to `input['entries/pages']` 

And then the `rooters/error.svelte`    overrided the `resolved` input['entries/pages']`

The `input` will effect rollupOptions.input 
```js
rollupOptions: {
input,
...
```

Finally, the  `.svelte-kit/runtime/components/error.svelte` didn't out put and the  generate manifest.json   does not contain it.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/build/build_server.js
+++ b/packages/kit/src/core/build/build_server.js
@@ -154,14 +154,21 @@ export async function build_server(
 		}
 	});
 
+	const fallback = `${get_runtime_path(config)}/components`;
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Thank you. This isn't quite the right fix — opened #4284, so I'll close this

PR #4203, opened 2022-03-04
Ensure props are loaded from matching endpoint during client-side navigation

### LINKED ISSUE #4038: Clashing endpoint-powered pages break route sorting

Rich-Harris:
Fixes #4038.

Credit to @aolose for the first version of this (https://github.com/sveltejs/kit/pull/4139). I tried to work on that branch but things had diverged too much, and I was concerned that the implementation came with fair bit of intrinsic complexity. This implementation is slightly less sophisticated — if an endpoint falls through during client-side navigation, Kit won't find which endpoint _actually_ matches and allow the client to short-circuit the next attempt — but it's less code and easier to follow. Fallthroughs are something of an edge case so I'm okay with them not being fully optimised.

### Tests

### Changesets


packages/kit/types/internal.d.ts
```diff
@@ -179,6 +177,8 @@ export interface ShadowEndpointOutput<Output extends JSONObject = JSONObject> {
 	body?: Output;
 }
 
+type ShadowKey = string;
```
benmccann, line 180: could we add a comment describing the format of this key and why we need it?

PR #4206, opened 2022-03-04
Deduplicate CI testing

mrkishi:
Let's see if this helps. I'm not sure the port assignment will work.

### Tests

### Changesets


packages/kit/package.json
```diff
@@ -72,7 +73,8 @@
 		"format": "npm run check-format -- --write",
 		"check-format": "prettier --check . --config ../../.prettierrc --ignore-path .gitignore",
 		"prepublishOnly": "npm run build",
-		"test": "npm run test:unit && npm run test:typings && npm run test:packaging && npm run test:prerendering && npm run test:integration",
+		"test": "npm run test:unit && npm run test:typings && npm run test:packaging",
+		"test:all": "npm run test:unit && npm run test:typings && npm run test:packaging && npm run test:prerendering && npm run test:integration",
```
Rich-Harris, line 77:
```suggestion
		"test:all": "npm run test && npm run test:prerendering && npm run test:integration",
```

benmccann:
Personally, I'd remove most of the ~25 different test scripts from `packages/kit/package.json`. If I want to run an individual set of tests I'll do something like `cd test/apps/basics && pnpm test` which easier for me to remember

Rich-Harris:
`test:integration:basics` etc aren't there so you can type `npm run test:integration:basics` instead of `cd test/apps/basics && pnpm test`, it's there so that `test:all` can look like this...

```
"test:all": "npm run test && npm run test:prerendering && npm run test:integration"
```

...instead of like this:

```
"test:all": "npm run test && cd test/prerendering/basics && pnpm test && cd ../options && pnpm test && cd ../../apps/amp && pnpm test && cd ../basics && pnpm test && cd ../options && pnpm test && cd ../options-2 && pnpm test"
```

Open to a third way if it's possible to DRY that out without introducing a lot of indirection

mrkishi:
This should do!

packages/kit/test/utils.js
```diff
@@ -169,7 +169,8 @@ export const config = {
 	use: {
 		screenshot: 'only-on-failure',
 		trace: 'retain-on-failure'
-	}
+	},
+	workers: process.env.CI ? 2 : undefined
```
benmccann, line 173: we had removed this purposefully not long ago. a lot of the CIs have only 1 CPU and we were worried that having multiple workers was making them compete for CPU and causing timeouts

packages/kit/package.json
```diff
@@ -60,7 +60,8 @@
 		"build": "rollup -c && node scripts/cp.js src/runtime/components assets/components && npm run types",
 		"dev": "rollup -cw",
 		"lint": "eslint --ignore-path .gitignore --ignore-pattern \"src/packaging/test/**\" \"{src,test}/**/*.{ts,mjs,js,svelte}\" && npm run check-format",
-		"check": "tsc && npm run check:integration && npm run check:prerendering",
+		"check": "tsc",
+		"check:all": "tsc && npm run check:integration && npm run check:prerendering",
```
benmccann, line 64: would it make sense for this to also use the `-r` syntax in the same way you updated `test:all`?

packages/kit/test/utils.js
mrkishi: This is a test, but adding full parallelization to the tests (at the pnpm level) didn't cause contention to the point of increasing run time, in fact it reduced a few percent. This is also a "max workers" setting—shouldn't playwright still use 1 worker when there's only one cpu?

packages/kit/package.json
mrkishi: Oops, I thought I commit that.

packages/kit/test/utils.js
benmccann: I don't think adding more workers would increase overall runtime, but it may make individual tests take longer to run

But anyway it looks like I may have been wrong that some environments only support a single CPU? I think I'm seeing now that most support 2, so this would be fine

benmccann (approved):

PR #4213, opened 2022-03-04
replace timestamp in $service-worker with version

### LINKED ISSUE #4196: `timestamp` should be replaced with `version` in service workers
### Describe the problem

#3412 added a `version` export to `$app/env`, which defaults to the timestamp at the time of builld. The `$service-worker` module doesn't expose it — it instead exposes a different `timestamp`.

Add `version` to `$service-worker`, and replace `timestamp` with an object that will throw a noisy error if you try to stringify it (which will cover the common case where it's used as a cache key).

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

Rich-Harris:
fixes #4196. Service workers that currently use `timestamp` to create cache keys will see this:

image

### Tests

### Changesets

### DIFF
--- a/packages/kit/types/ambient.d.ts
+++ b/packages/kit/types/ambient.d.ts
@@ -214,7 +214,7 @@ declare module '$lib' {}
 
 /**
  * ```ts
- * import { build, files, timestamp } from '$service-worker';
+ * import { build, files, prerendered, version } from '$service-worker';
  * ```
  *
  * This module is only available to [service workers](/docs/service-workers).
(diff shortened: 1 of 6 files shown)


packages/kit/types/ambient.d.ts
```diff
@@ -233,9 +233,9 @@ declare module '$service-worker' {
 	 */
 	export const prerendered: string[];
 	/**
-	 * The result of calling `Date.now()` at build time. It's useful for generating unique cache names inside your service worker, so that a later deployment of your app can invalidate old caches.
+	 * See [`config.kit.version`](/docs/configuration#version). It's useful for generating unique cache names inside your service worker, so that a later deployment of your app can invalidate old caches.
```
benmccann, line 236: hmm. we may have broken scroll handling recently. When I go to https://kit.svelte.dev/docs/configuration#version it doesn't scroll to the correct location

Conduitry: It works for me on Firefox. But yet, it does not work for me on Chrome.

PR #4220, opened 2022-03-04
typecheck adapters

### LINKED ISSUE #625: Type checking in adapters

Rich-Harris:
Closes #625

### Tests

### Changesets


packages/adapter-cloudflare-workers/index.js
```diff
@@ -67,7 +67,7 @@ export default function () {
 	};
 }
 
-/** @param {import('@sveltejs/kit').Builder} builder */
+/** @param {any} builder */
```
Rich-Harris, line 70: this is a bit cheeky but i'm not sure if it's worth making `Builder` public

packages/adapter-netlify/index.js
```diff
@@ -153,7 +153,7 @@ function get_netlify_config() {
 
 /**
  * @param {NetlifyConfig} netlify_config
- * @param {import('@sveltejs/kit').Builder} builder
+ * @param {any} builder
```
Rich-Harris, line 156: see https://github.com/sveltejs/kit/pull/4220/files#r819806068

packages/adapter-cloudflare-workers/index.js
benmccann: The only thing we gain by not making it public is that we can rename it from `Builder` to something else, but we can't change any of the methods on it in any manner, so the API is already 95% locked. Personally, I would be in favor of making it public

Rich-Harris: true. updated

benmccann (approved):

Issue #4215, opened 2022-03-04
`$lib` in ambient modules overrides actual index file definitions


dummdidumm:
### Describe the bug

In #4099 module types were documented. As part of that `$lib` was added in the form of `declare module '$lib' {}` with a comment so it shows up in the docs. This has bad consequences for everyone using a `index.ts/js` file however to reexport stuff from lib: That definition overrides the actual contents of that index file.

### Reproduction

Folder structure:
```
src
  lib
    index.ts
  routes
    index.svelte
```
Contents of `index.ts`:
```ts
export const foo = 'bar';
```
Contents of `index.svelte`:
```svelte
<script lang="ts">
  import { foo } from '$lib'; // error: $lib has no exported member foo
</script>
```

### System Info

```shell
System:
    OS: Windows 10 10.0.19042
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 17.56 GB / 31.74 GB
  Binaries:
    Node: 16.13.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.17 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.1.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (98.0.1108.62)
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.30
    @sveltejs/kit: next => 1.0.0-next.291
    svelte: ^3.44.0 => 3.46.4
```

### Severity

serious, but I can work around it

### Additional Information

A workaround is too add `declare module '$lib' { export let MyComponent: any; ...}` to `app.d.ts` which will widen the type of that module to include the given exports.

Rich-Harris:
bleurgh, good catch. Is there a good way to make it show up in the right place in the docs without including it in `ambient.d.ts`? https://kit.svelte.dev/docs/modules#$lib

PR #4227, opened 2022-03-05
[fix] Support override  by index file definitions

baseballyama:
fix: https://github.com/sveltejs/kit/issues/4215

### Check List

- [x] `import { foo } from '$lib';` doesn't show type error
  - [x] I can jump type definition when I click `foo`
- [x] after change `lib` path at `svelte.config.js`, still `import { foo } from '$lib';` doesn't show type error
  - [x] I can jump type definition when I click `foo`
- [x] Docs show `$lib` even after the change

But it seems like a very simple solution, am I missing something?

### Tests

### Changesets

### DIFF
--- a/.changeset/odd-dots-hang.md
+++ b/.changeset/odd-dots-hang.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] Support override \$lib by index file definitions
--- a/.changeset/odd-dots-hang.md
+++ b/.changeset/odd-dots-hang.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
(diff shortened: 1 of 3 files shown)


.changeset/odd-dots-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] Support override \$lib by index file definitions
```
Rich-Harris, line 5:
```suggestion
Remove declared `$lib` module
```

Rich-Harris (approved):

Rich-Harris:
Thank you — I realised we can actually put whatever we like on that page by changing the `extract-types.js` module, so I opted to do that instead of declaring an unused namespace

Issue #4188, opened 2022-03-03
`version.json` doesn't consider `paths.assets`


Rich-Harris:
### Describe the bug

Mentioned in https://github.com/sveltejs/kit/issues/3666#issuecomment-1042102935:

> it seems that the path to `version.json` is not considering the `assets` path setting. Let's say `paths.assets` is `https://example.com/assets`, it would then still attempt to load `https://example.com/_app/version.json` (instead of `https://example.com/assets/_app/version.json`)

### Reproduction

Will put together a repro later, if the fix isn't obvious

### System Info

```shell
System:
    OS: macOS 12.0.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 408.91 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.1 - ~/.nvm/versions/node/v16.13.1/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v16.13.1/bin/yarn
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.1/bin/npm
  Browsers:
    Chrome: 98.0.4758.109
    Firefox: 97.0.1
    Safari: 15.1
  npmPackages:
    @sveltejs/eslint-config: github:sveltejs/eslint-config#v5.8.0 => 5.8.0
```

### Severity

annoyance

georgecrawford:
@Rich-Harris It could be a misconfiguration on my side, but is it possible that we're now mistakenly _caching_ the `version.json` response? I'm seeing this:

```
$ curl -i https://domain.com/_app/version.json
HTTP/1.1 200 OK
Date: Thu, 05 May 2022 22:06:20 GMT
Content-Type: application/json
Content-Length: 54
<...snip>
Cache-Control: public,max-age=31536000,immutable

{"version":"33a6aec40c50fa9f56eb5be2f4b36f30da58fc8f"}
```

yanni4night:
I am afraid this will cause a CORS error if paths.assets has a different domain

PR #4234, opened 2022-03-05
[fix] consider `paths.assets` for `version.json`

baseballyama:
fix: https://github.com/sveltejs/kit/issues/4188

### Checklist
I checked behavior with adapter-node.
- [x] It can fetch `version.json` if `kit.paths.assets` has custom value
- [x] It can fetch `version.json` if `kit.paths.assets` doesn't set value

I couldn't find how to write tests.
So could you please teach me if there are some sample existing tests?

### Tests

### Changesets

### DIFF
--- a/.changeset/silent-crews-change.md
+++ b/.changeset/silent-crews-change.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+consider `paths.assets` for `version.json`
--- a/.changeset/silent-crews-change.md
+++ b/.changeset/silent-crews-change.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-consider `paths.assets` for `version.json`
+fetch `version.json` relative to `paths.assets`, if set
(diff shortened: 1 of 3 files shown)


Rich-Harris:
Thanks — this _would_ fix the issue (agree that it's quite hard to add a test for this, I suppose we could set `paths.assets` to `http://localhost:${TEST_PORT}/assets` and check it works, but that feels like a task for another day) but it's not 100% future-proof, because one day we'd like to introduce dynamic basepaths (#595) so that Kit apps work on IPFS/the internet archive/situations where the basepath is set by an environment variable/etc.

I think we can support that by just using `paths.assets` instead of `paths.base` (if `assets` is unset, it defaults to `base`) — 5cb887020877461b6302f8b6a083e9b5422ef27e

baseballyama:
`VITE_SVELTEKIT_APP_VERSION_FILE` is super specific name, so I assigned complete path.
But I agree with you that building complete path at runtime is more flexible if we consider dynamic basepath.👍

.changeset/silent-crews-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+consider `paths.assets` for `version.json`
```
Rich-Harris, line 5:
```suggestion
fetch `version.json` relative to `paths.assets`, if set
```

Rich-Harris (approved):

PR #4239, opened 2022-03-05
[fix] Fix build

baseballyama:
Follow up: https://github.com/sveltejs/kit/pull/4235

The latest release was failed due to some build configuration.
So I tried to fix it.

### Tests

### Changesets

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -380,7 +380,7 @@ importers:
   sites/kit.svelte.dev:
     specifiers:
       '@sveltejs/adapter-auto': workspace:*
-      '@sveltejs/adapter-static': workspace:^1.0.0-next.28
+      '@sveltejs/adapter-static': workspace:^1.0.0-next.29
       '@sveltejs/kit': workspace:*
       '@sveltejs/site-kit': ^2.0.3
       '@types/node': ^16.11.11
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -384,6 +384,7 @@ importers:
       '@sveltejs/kit': workspace:*
       '@sveltejs/site-kit': ^2.0.3
       '@types/node': ^16.11.11
+      devalue: ^2.0.1
       flexsearch: ^0.7.21
       marked: ^4.0.5
       prism-svelte: ^0.5.0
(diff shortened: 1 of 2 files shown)


pnpm-lock.yaml
```diff
@@ -380,10 +380,11 @@ importers:
   sites/kit.svelte.dev:
     specifiers:
       '@sveltejs/adapter-auto': workspace:*
-      '@sveltejs/adapter-static': workspace:^1.0.0-next.28
+      '@sveltejs/adapter-static': workspace:^1.0.0-next.29
```
Conduitry, line 383: The static adapter should presumably be completely removed from the kt.svelte.dev site.

baseballyama: Maybe it's just my local, but after removing` @sveltejs/adapter-static`, `pnpm build` is failed.

PH4NTOMiki:
Can you try to merge it @Conduitry

benmccann:
Is every project going to need `devalue`? I don't think that would be a good situation

I'm guessing the need for it might be related to this: https://github.com/sveltejs/kit/pull/3942#issuecomment-1041622382

Conduitry:
@benmccann `devalue` is only needed, I believe, because of the changes in https://github.com/sveltejs/kit/pull/4221/commits/ab4ef9cf5130bb30169e60444331352365f68ba6 - I wouldn't say no to reverting those as well, because it's clearly not the proper solution. (I'm not even sure whether Rich intended that to be part of the final PR.)

Rich-Harris:
It would be nice to get rid of `devalue` if we can, so this problem doesn't pop up. The rationale behind ab4ef9cf5130bb30169e60444331352365f68ba6 was to make it easier to debug Kit while working on kit.svelte.dev, and I do think there's some value in preserving that — definitely wouldn't be upset if we decided against it but for now I reckon I'll merge this so that things are at least working again

Rich-Harris (approved):

PH4NTOMiki:
@Rich-Harris thanks for merging that

PH4NTOMiki:
@Rich-Harris I think you need to rerun that action that failed previously.
not sure thought

baseballyama:
Maybe should we swap (rebase) latest 2 commits for releasing?

bitdom8:
Latest build doesn't run 

    pnpm run build

ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and 'C:\Users\Administrator\Documents\client\package.json' contains "type": "module". To treat 
it as a CommonJS script, rename it to use the '.cjs' file extension.

Issue #3014, opened 2021-12-10
Link in html files are wrong for routes when using adapter-static


myhrmans:
### Describe the bug

When building using adapter-static and going to the output folder to check the index.html files you can see that the links
`		<link rel="modulepreload" href="/_app/start-0680aa2f.js">

		<link rel="modulepreload" href="/_app/chunks/vendor-3d58c1d9.js">

		<link rel="modulepreload" href="/_app/chunks/preload-helper-ec9aa979.js">

		<link rel="modulepreload" href="/_app/chunks/singletons-ff603286.js">

		<link rel="stylesheet" href="/_app/assets/start-61d1577b.css">`

or in this case; the entire file is identical. 
The links are relative but when creating routes they should be relative to where the index files are stored. 

the viewed file in the below image is` /_app/home/prototype/index.html` but the links in it refer to `/_app...` but should be `../../../_app`
![image](https://user-images.githubusercontent.com/14261388/145533945-2571fe4b-2a2c-483d-b7db-22e28f53cee5.png)

Opening one of these index files will result in a "FILE_NOT_FOUND"

### Reproduction

https://github.com/myhrmans/sveltekit-link-generation-bug

### System Info

```shell
System:
    OS: Linux 5.11 Ubuntu 20.04.3 LTS (Focal Fossa)
    CPU: (8) x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
    Memory: 440.27 MB / 31.28 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: 1.22.10 - /usr/bin/yarn
    npm: 8.1.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Chrome: 96.0.4664.93
    Chromium: 96.0.4664.45
    Firefox: 95.0
```

### Severity

blocking all usage of SvelteKit

locriacyber:
You were using `file://` transport, which is platform-dependent. You are not supposed to visit a website from disk. Run a HTTP server and visit the website that way. `python -m http.server` is a simple HTTP server.

myhrmans:
Wrong button...

Rich-Harris:
Closing, as SvelteKit uses relative links in all cases except an explicit `paths.asset` or fallback prerendering, so this should work even with `file://` URLs

PR #4250, opened 2022-03-06
use relative asset paths

Rich-Harris:
This was just a quick experiment inspired by #3014 to see if it's possible to create prerendered SvelteKit apps that don't even need a webserver. 

It basically works — if you have routes like `src/routes/index.html.svelte` instead of `index.svelte`, you can double-click on the resulting `build/index.html` file, and you have an (almost) fully working site, because the generated CSS and JS are loaded from `file://` URLs. Might be useful in certain contexts, and is a step towards #595. 

One thing it doesn't account for is fallback routes (i.e. if you visit `/foo/bar/baz` in an SPA and it's served by the fallback page, the relative URLs will be wrong). The biggest issue though is that Vite uses absolute URLs — if you import an asset from JS or CSS, it will be imported from an absolute URL rather than a relative one. I think that could be changed (along with https://github.com/vitejs/vite/issues/2009) for the benefit of all, as it would make every Vite app more portable. Unless and until that happens, I'm going to mark this as draft and not spend any more time on it.

### Tests

### Changesets


locriacyber:
I was able to track it down to `.svelte-kit/output/client/_app/chunks/preload-helper-e4860ae8.js`. Content of the file:
```js
const a = "modulepreload",
  o = {},
  u = "/_app/",
  h = function (s, n) {
    return !n || n.length === 0
      ? s()
      : Promise.all(
          n.map((e) => {
            if (((e = `${u}${e}`), e in o)) return
            o[e] = !0
            const t = e.endsWith(".css"),
              i = t ? '[rel="stylesheet"]' : ""
            if (document.querySelector(`link[href="${e}"]${i}`)) return
            const r = document.createElement("link")
            if (
              ((r.rel = t ? "stylesheet" : a),
              t || ((r.as = "script"), (r.crossOrigin = "")),
              (r.href = e),
              document.head.appendChild(r),
              t)
            )
              return new Promise((l, c) => {
                r.addEventListener("load", l),
                  r.addEventListener("error", () =>
                    c(new Error(`Unable to preload CSS for ${e}`))
                  )
              })
          })
        ).then(() => s())
  }
export { h as _ }
```

Note that "/_app/" was here.

If there is a way to change the way `import(xxxxx)` is compiled, this could be solved.

This was mentioned in https://github.com/vitejs/vite/issues/2024.

Rich-Harris:
Ok I think this is ready. The one caveat is that we need to use the user-specified `base` for the SSR build (rather than `./`) because otherwise imported assets will have the wrong URL, which results in things like images 404ing for a split second before hydration kicks in. 

Nothing really changes as a result of this PR (other than apps being more resilient to things like finding themselves on the internet archive, etc), but it hopefully paves the way for #595 etc.

Rich-Harris:
> Ok I think this is ready

It's not ready. A weird thing I noticed is that since this change, workers are built to the wrong place. They're _already_ built to the wrong place, but now they're built to a wronger place. If you visit https://kit.svelte.dev/ and inspect the network tab, you'll see this file:

https://kit.svelte.dev/_app/assets/search.0db017ff.js

That's the web worker that powers the search function. It should be in `_app/immutable` but for some reason it's not, it's just in `_app`, which means it won't get properly cached.

As of this PR, it's not even inside `_app` — it's built to `/assets/search.0db017ff.js`. I've tried setting `worker.rollupOptions.output` but it doesn't seem to be having any effect. Still investigating.

Rich-Harris:
I believe I've found two bugs in Vite:

1. `worker.rollupOptions.output` is not respected if it's set inside a plugin's `config` hook

2. When building with `base: './'`, workers are instantiated with `new Worker('../../path/to/worker.js')`, rather than with a root-relative path beginning with `base`. But that's wrong — the path will be relative to the _document_, not the module instantiating a worker. I think it should be something like this:
    ```js
    new Worker(new URL('../../path/to/worker.js', import.meta.url).href)
    ```

Will see if I can cobble together a repro later, but cc'ing @bluwy for visibility in the meantime

patak-dev:
> 2. When building with `base: './'`, workers are instantiated with `new Worker('../../path/to/worker.js')`, rather than with a root-relative path beginning with `base`. But that's wrong — the path will be relative to the _document_, not the module instantiating the worker. I think it should be something like this:
>    ```js
>    new Worker(new URL('../../path/to/worker.js', import.meta.url).href)
>    ```

This should be fixed in the next patch by:
- https://github.com/vitejs/vite/pull/9204

> 1. `worker.rollupOptions.output` is not respected if it's set inside a plugin's `config` hook

For this one, I'll wait for the repro.

Rich-Harris:
thank you, done! https://github.com/vitejs/vite/issues/9205

benmccann:
@Rich-Harris both fixes were released as part of Vite 3.0.3

benmccann (approved):
Is there anything special we need to do for https://kit.svelte.dev/docs/modules#$app-paths-assets?

Assuming not, this looks good to me

Rich-Harris:
I don't think so. This PR still uses an absolute base in SSR mode, as I couldn't figure out an elegant way to use `renderBuiltUrl` to create relative asset paths at SSR time; will leave that problem for another day. Just waiting to see if the windows failure is flakiness or not

Amerlander:
Is there an option to keep absolute paths when using static adapter?
I'm not able to use vite imagetools with that: https://github.com/JonasKruckenberg/imagetools/issues/367

Issue #4264, opened 2022-03-07
Remove network calls from demo app


benmccann:
### Describe the bug

I have seen so many people complaining that SvelteKit is slow. No one realizes that a locally created demo app would be making network calls.

E.g. https://discord.com/channels/457912077277855764/939868205869072444/950313663850508328

This would also solve our issues around eventual consistency: https://github.com/sveltejs/kit/issues/1564

### Reproduction

I don't know if these people have particularly slow network connections or the app or host is sometimes being quite slow

### System Info

```shell
N/A
```

### Severity

annoyance

Conduitry:
Even without the speed concerns, I'd be in favor of a tiny ad-hoc in-memory data store thing for the todo app rather than calling out to the Cloudflare KV API.

am283721:
I'm wondering if it's an issue with the API being served from somewhere in North America and certain users on other continents experiencing unusually slow network performance?

I also think some sort of in-memory data store would provide a better/more consistent user experience as long as it's still able to demonstrate the same functionality that is shown with the current setup. Specifically things like utilizing fetch, handling redirects, method overrides, etc... Having a demo that interacts with an external API is very useful as most people will need to do the same, and showing some best practices helps them get started on the right foot.

An update here could also be an opportunity to add functionality demonstrating shadow endpoints (unless that would create too much confusion for new users).

baseballyama:
IMO...

By default, the Demo app will use an in-memory object.
And as default, commenting out files that connect to the API,
If users want to try to use API also, then users can use the commented-out files.
Then users don't need to feel bad experience.
And even users can try persistence and dynamic SSR if necessary.

am283721:
I was thinking using something like [json-server](https://github.com/typicode/json-server) could provide a good middle-ground, where a full rest api could be utilized but the backend is running in the same environment as the frontend.

To me, persisting data between a local dev environment and a deployed environment isn't that important, considering in most real-world scenarios you wouldn't want that either.

The main downside I can see of using something like `json-server` is introducing a 3rd party dependency to the demo app that might require future maintenance. Secondarily, it adds a little more code and complexity to the demo app.

aolose:
I prefer the app to send these requests, it is closer to reality. We need a more friendly interactive experience.
For example, you could add a status display to each new todo item instead of waiting for the request to complete before adding it to the list.

Rich-Harris:
[...]

PR #4266, opened 2022-03-07
[chore] use in memory object instead of Cloudflare KV API for demo app

baseballyama:
fix: https://github.com/sveltejs/kit/issues/4264

I am accessing the Demo app from Tokyo but the response is not good.
I considered using indexedDB or Session Storage but did not adopt them because it is important that this process be performed on the back end.
And the process should not be complex to avoid making noise, so I finally opted for a simple in-memory object.
Any opinions for a better demo app?

### Tests

### Changesets


gtm-nayan:
@baseballyama I thought using a Map would be more semantically correct, and abstracting the db operations would help with readability, so I've made a PR over at your branch https://github.com/baseballyama/kit/pull/6

gtm-nayan:
The abstractions should make it easier to add some persistence as well if needed.

baseballyama:
I don't think there is consensus on this PR yet, but I have modified it so that we can try both in-memory objects and Cloudflare KV API.
This allows us to use in-memory objects by default, but also to demonstrate using the API.
I'm not sure if we should inform users of this by some means, but I don't think it is necessary at this time.

This is a just proposal, so please close it if it is not in line with the core team's decision.

Issue #1712, opened 2021-06-19
Cloudflare Worker Adapter - Durable Objects


DrewRidley:
**Is your feature request related to a problem? Please describe.**
Cloudflare recently introduced the ability to maintain websocket connections with workers, in addition to a stateful storage solution called 'Durable Objects'. For workers, these objects are accessible through a ``env`` argument passed in during the ``fetch`` invocation. More information about durable objects and their utilization can be found here.
https://developers.cloudflare.com/workers/learning/using-durable-objects

**Describe the solution you'd like**
In my opinion, the cloudflare worker adapter should expose a way to mutate and interact with this ``env`` object passed in by cloudflare during a function invocation.

**Describe alternatives you've considered**
I have not found any alternate ways of using durable objects without modifying the adapter.

**How important is this feature to you?**
This feature is fairly important to me personally, since it would dramatically simplify my stack if my state coordination can be done in the same worker.

DrewRidley:
I can look into this feature in the coming days to see how feasible it would be to introduce it as a PR, however, if the issue requires more significant changes to the adapter I might want to wait and consult some others to learn more about the impact this feature request might have.

DrewRidley:
I have done some more research and found a few areas the adapter can be improved. Namely, the documentation needs a significant upgrade. The documentation is confusing for new users and does not explain what the ``toml`` file should look like. Secondly, the adapter does not currently use ``kv-asset-handler`` to its full potential and does not set any browser cache headers on the static files served. At the very least, the adapter needs to include some way for users to be able to specify their cache preferences for static files. 

Since I have never contributed to ``sveltekit`` before, this raises a few questions. Namely, im wondering whether an RFC would be necessary for such changes. Any help or clarification would be much appreciated!

DrewRidley:
As a short update;

I managed to get the module syntax fully compiled. I am just waiting for cloudflare to merge my PR of ``kv-asset-handler`` which introduces support for the module syntax there. The adapter also has some new configuration options so my PR will include some changes to the documentation.

Most notably however, my PR will introduce a breaking change. This will likely need to be discussed further with a larger group of the community. The reason I would like to introduce the breaking change is that eventually cloudflare plans to deprecate the old syntax, and force usage of their new modular syntax. For this reason, I think it makes sense for the modular syntax to be the default for all workers created using this adapter.

[...]

PR #4276, opened 2022-03-08
refactor: use Module Worker pattern instead of Service Worker, add error handling

dominikg:
as discussed on discord.

- switch to ModuleWorker pattern
- add error handling for static kv requests
- pass platform to server.respond

### Tests

### Changesets

### DIFF
--- a/packages/adapter-cloudflare-workers/files/entry.js
+++ b/packages/adapter-cloudflare-workers/files/entry.js
@@ -1,61 +1,101 @@
 import { Server } from 'SERVER';
 import { manifest, prerendered } from 'MANIFEST';
 import { getAssetFromKV } from '@cloudflare/kv-asset-handler';
+import static_asset_manifest_json from '__STATIC_CONTENT_MANIFEST';
+const static_asset_manifest = JSON.parse(static_asset_manifest_json);
 
 const server = new Server(manifest);
 
 const prefix = `/${manifest.appDir}/`;
 
-addEventListener('fetch', (/** @type {FetchEvent} */ event) => {
-	event.respondWith(handle(event));
-});
+export default {
+	/**
+	 * @param {Request} req
+	 * @param {any} env
+	 * @param {any} context
+	 */
+	async fetch(req, env, context) {
+		const url = new URL(req.url);
+
+		// static assets
+		if (url.pathname.startsWith(prefix)) {
+			/** @type {Response} */
+			const res = await get_asset_from_kv(req, env, context);
+			if (is_error(res.status)) {
+				return res;
+			}
+			return new Response(res.body, {
+				headers: {
+					// include original cache headers, minus cache-control which
+					// is overridden, and etag which is no longer useful
+					'cache-control': 'public, immutable, max-age=31536000',
+					'content-type': res.headers.get('content-type'),
+					'x-robots-tag': 'noindex'
+				}
+			});
+		}
+
+		// prerendered pages and index.html files
+		const pathname = url.pathname.replace(/\/$/, '');
+		let file = pathname.substring(1);
+
+		try {
+			file = decodeURIComponent(file);
+		} catch (err) {
+			// ignore
+		}
+
+		if (
+			manifest.assets.has(file) ||
+			manifest.assets.has(file + '/index.html') ||
+			prerendered.has(pathname || '/')
+		) {
+			return get_asset_from_kv(req, env, context);
+		}
+
+		// dynamically-generated pages
+		try {
+			return await server.respond(req, {
+				platform: { env, context },
+				getClientAddress() {
+					return req.headers.get('cf-connecting-ip');
+				}
+			});
+		} catch (e) {
+			return new Response('Error rendering route: ' + (e.message || e.toString()), { status: 500 });
+		}
+	}
+};
 
 /**
- * @param {FetchEvent} event
- * @returns {Promise<Response>}
+ * @param {Request} req
+ * @param {any} env
+ * @param {any} context
  */
-async function handle(event) {
-	const { request } = event;
-
-	const url = new URL(request.url);
-
-	// generated assets
-	if (url.pathname.startsWith(prefix)) {
-		const res = await getAssetFromKV(event);
-		return new Response(res.body, {
-			headers: {
-				'cache-control': 'public, immutable, max-age=31536000',
-				'content-type': res.headers.get('content-type')
-			}
-		});
-	}
-
-	// prerendered pages and index.html files
-	const pathname = url.pathname.replace(/\/$/, '');
-	let file = pathname.substring(1);
-
+async function get_asset_from_kv(req, env, context) {
 	try {
-		file = decodeURIComponent(file);
-	} catch (err) {
-		// ignore
-	}
-
-	if (
-		manifest.assets.has(file) ||
-		manifest.assets.has(file + '/index.html') ||
-		prerendered.has(pathname || '/')
-	) {
-		return await getAssetFromKV(event);
-	}
-
-	// dynamically-generated pages
-	try {
-		return await server.respond(request, {
-			getClientAddress() {
-				return request.headers.get('cf-connecting-ip');
+		return await getAssetFromKV(
+			{
+				request: req,
+				waitUntil(promise) {
+					return context.waitUntil(promise);
+				}
+			},
+			{
+				ASSET_NAMESPACE: env.__STATIC_CONTENT,
+				ASSET_MANIFEST: static_asset_manifest
 			}
-		});
+		);
 	} catch (e) {
-		return new Response('Error rendering route:' + (e.message || e.toString()), { status: 500 });
+		const status = is_error(e.status) ? e.status : 500;
+		return new Response(e.message || e.toString(), { status });
 	}
+}
+
+/**
+ * @param {number} status
+ * @returns {boolean}
+ */
+function is_error(status) {
+	return status > 399;
 }
--- a/packages/adapter-cloudflare-workers/index.js
+++ b/packages/adapter-cloudflare-workers/index.js
@@ -6,7 +6,7 @@ import toml from '@iarna/toml';
 import { fileURLToPath } from 'url';
 
 /** @type {import('.')} */
-export default function () {
+export default function (options = {}) {
 	return {
 		name: '@sveltejs/adapter-cloudflare-workers',
 
@@ -53,12 +53,13 @@ export default function () {
 			);
 
 			await esbuild.build({
-				entryPoints: [`${tmp}/entry.js`],
-				outfile: `${entrypoint}/${main_path}`,
 				target: 'es2020',
 				platform: 'browser',
+				...options,
+				entryPoints: [`${tmp}/entry.js`],
+				outfile: `${entrypoint}/${main_path}`,
 				bundle: true,
-				external: ['__STATIC_CONTENT_MANIFEST'],
+				external: ['__STATIC_CONTENT_MANIFEST', ...(options?.external || [])],
 				format: 'esm'
 			});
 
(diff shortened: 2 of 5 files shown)


dominikg:
note: `pnpm check` returned an unrelated error for adapter-netlify when i ran it locally,  `pnpm test` failed a lot with create page timeouts.  lets see what ci shows here.

packages/adapter-cloudflare-workers/files/entry.js
```diff
@@ -1,57 +1,96 @@
...
+			{
+				request: req,
+				waitUntil(promise) {
+					return context.waitUntil(promise);
+				}
+			},
+			{
+				ASSET_NAMESPACE: env.__STATIC_CONTENT,
+				ASSET_MANIFEST: static_asset_manifest
+			}
```
Rich-Harris, line 82: mentioned this in discord, but for posterity: would be good to understand why (if) this is necessary

dominikg: taken directly from the docs of kv-asset-handler: https://github.com/cloudflare/kv-asset-handler#es-modules

maybe someone from CF or with more experience using it can chime in why it is done that way?

lukeed: > The binding name to the KV Namespace populated with key/value entries of files for the Worker to serve. By default, Workers Sites uses a [binding to a Workers KV Namespace](https://developers.cloudflare.com/workers/reference/storage/api/#namespaces) named __STATIC_CONTENT.

Awkward, but is what it is. Both are required

Rich-Harris: I don't understand how the existing adapter works if this is true? Is it a different between the service worker and module formats? (Also that link is broken btw :)

hmnd: @Rich-Harris Service workers get KV namespaces and env variables bound to global, whereas for module workers, they are provided in an env parameter passed to the exported [`fetch` function](https://github.com/sveltejs/kit/blob/4db18af228ac4960adbfce04ef46fda1cf219f2a/packages/adapter-cloudflare-workers/files/entry.js#L17). The current implementation for service workers doesn't pass any options to `getAssetFromKV`, so it [defaults to using the globally bound `__STATIC_CONTENT`](https://github.com/cloudflare/kv-asset-handler/blob/fd27ccfeb541312f5e6c6be74932efc32ae4ce1f/src/index.ts#L24-L25).
Sorry to butt in, just looking forward to this getting merged ;)

hmnd:
BTW this should also fix #2844

hmnd (changes requested):
Really excited to finally get env var support for this adapter.
Some more changes are needed to `index.js` for this to work. See [my fork here](https://github.com/hmnd/svelte-kit/blob/refactor/adapter-cloudflare-workers/packages/adapter-cloudflare-workers/index.js). I've added support for using the value of `build.upload.main` in there too.

packages/adapter-cloudflare-workers/files/entry.js
PH4NTOMiki: > (Also that link is broken btw :)

https://developers.cloudflare.com/workers/runtime-apis/kv/#referencing-kv-from-durable-objects-and-workers-using-modules-syntax I think Luke meant this, but in any case this explains it.

Rich-Harris: Thanks. Those docs suggest we should do this...

```js
const valueFromKV = await this.env.NAMESPACE.get('someKey')
return new Response(valueFromKV)
```

...rather than importing `getAssetFromKV` and passing in `ASSET_NAMESPACE` and `ASSET_MANIFEST`. (Though I don't quite understand how that sets the correct content type etc, and it's not clear from the docs how `someKey` relates to `request.url`) Honestly I'm a bit flummoxed at this point!

PH4NTOMiki: That specific example is for Durable Objects, so not 100% applicable but consider this code example https://blog.cloudflare.com/workers-javascript-modules/#:~:text=body%3A%20error.stack%0A%20%20%7D)%0A%7D-,//%20filename%3A%20./src/worker.js,-import%20%7B%20logError%20%7D%20from

lukeed: You can do it directly without the `kv-asset-handler` package, but then you have to do the lookups & the content-type setting yourself. (This is what I do, I don't use this pkg). The `0.2.0` of this pkg had an API change to allow for/accommodate Module Worker syntax. Before then, it just assumed & grabbed the namespace binding (aka global variable) that Workers Sites uses. In a Module Worker world, there are no globals and you have to be explicit, hence the 2nd object of extra stuff.

The namespace is to actually do the lookups. 
The manifest is to know what paths exist (or don't) in the namespace. 

The pkg [deals with mime types](https://github.com/cloudflare/kv-asset-handler/blob/master/src/index.ts#L160) as is & does some other footwork for cache management, which we don't really need here imo.

dominikg:
> Really excited to finally get env var support for this adapter. Some more changes are needed to `index.js` for this to work. See [my fork here](https://github.com/hmnd/svelte-kit/blob/refactor/adapter-cloudflare-workers/packages/adapter-cloudflare-workers/index.js). I've added support for using the value of `build.upload.main` in there too.

thank you for this! Unfortunately it didn't show me a button to accept your changes. so i copied them over.

Readme and changeset probably need a few more words too.

packages/adapter-cloudflare-workers/index.js
```diff
@@ -86,6 +94,24 @@ function validate_config(builder) {
 			);
 		}
 
+		// @ts-ignore
+		const main_file = wrangler_config.build?.upload?.main;
```
dominikg, line 98: is this needed or could/should we hardcode/use the filename for main instead? 

The checks below feel a bit brittle, what if there is no glob but an excact matching rule instead?

dominikg:
On a more general note/question. How should breaking changes like this be handled once 1.0 is reached?
In kit we did add logging for some required changes, should the adapter read existing wrangler.toml and warn if it is not type module?

packages/adapter-cloudflare-workers/files/entry.js
dominikg: > This is what I do, I don't use this pkg

is that in another importable package or can you share it here? Or should we stick to kvAssetHandler regardless to reduce maintenance here. Its 312 lines currently and thats counting the comments. How bad can it be? (I feel i might regret this question later)

dominikg:
this could also help adding DO support: see https://github.com/sveltejs/kit/issues/1712#issuecomment-1022449210

packages/adapter-cloudflare-workers/index.js
hmnd: May be a bug with wrangler, but main seems to be treated according to the rules, regardless of the specified `format`. My build kept failing to upload until I changed the name to .mjs or added a rule to treat *.js as ESM. `build.upload.main` is actually required, so we can probably get rid of all the `?.` there.

packages/adapter-cloudflare-workers/files/entry.js
lukeed: https://github.com/lukeed/worktop/blob/next/packages/worktop/src/cfw.kv.assets.ts

```ts
import { serve } from 'worktop/cfw.kv.assets';

// ...

let res = await serve(env.__STATIC_CONTENT, request);
return res || new Response('Missing', { status: 404 });
```

Doesn't need the mapping, but you can do some routing with it outside of the `serve` call & then pass down the pre-determined pathname as the 2nd parameter instead of `request`; eg `serve(env.__STATIC_CONTENT, '/foobar.webm')`;

Rich-Harris: Note that we have access to all relevant MIME types in `manifest`, along with a list of the static assets themselves. Is there a way we can use this rather than importing a list of types, and `__STATIC_CONTENT_MANIFEST` for that matter?

Right now the MIME types are on the private `manifest._.mime` object, but we could make it public on `manifest.mime` if it's useful here.

lukeed: Yes, didn't know that existed. If that could be passed down to the handler then, that's all you'd need. No external library, just a KV read and probably grab the `Range` logic from my last link.

No idea what `__STATIC_CONTENT_MANIFEST` is – thought this was SK-specific too.

Rich-Harris: I've opened https://github.com/sveltejs/kit/pull/4302 — if we merge that, then this PR can use `manifest.mime` to construct its own responses.

Presumably we'd want to set ETags as well? Not sure what that looks like in practice

dominikg: Handling cache headers / etags is needed. The current approach of far-future caching hashed assets is good, this PR had just added some error handling to it.

If we were to replace kvAssetHandler with our own logic, do we need to handle edge caching too? What about prerender results? At some point we're ending up replementing a lot of things that are mostly generic. 

Would it be more useful to do this in a reusable/external way, eg in worktop or via sending PRs to kvAssetHandler. Or even having a separate "cloudflare-adapter-utils" library that can be shared between the two cf adapters.

I'm no longer confident we can easily merge the two because of the fundamental difference that one needs a local build step and the other one builds on cf.

Rich-Harris: We should serve prerender results, yeah. As for ETags, we probably want to generate content hashes at build time, which has implications for the shape of `SSRManifest`

benmccann:
@dominikg it looks like this PR will need a rebase

dominikg:
> @dominikg it looks like this PR will need a rebase

done, thx.

isaac-mcfadyen:
> BTW this should also fix #2844

That should already be closed... unless I'm misunderstanding what you mean by environment the current `adapter-cloudflare` passes the environment inside of `platform` to endpoints (I closed another issue that mentioned this but didn't catch yours when I made the PR).

packages/adapter-cloudflare-workers/files/entry.js
pzuraq: `__STATIC_CONTENT_MANIFEST` is also necessary because Wrangler adds an additional fingerprint to every asset it uploads, and it contains the mapping from the requested asset to the fingerprinted one.

Rich-Harris: What kind of fingerprint?

pzuraq: Here's an example with a test app I deployed:

Screen Shot 2022-04-11 at 10 26 19 AM

I'm unsure what adds these extra fingerprints, or why they need to exist, but without the `__STATIC_CONTENT_MANIFEST` the files will not be found.

This is definitely not ideal, but I think it's acceptable given as soon as Pages is out of beta, everyone is going to drop Workers and migrate over. If for some reason that doesn't happen, I'd be happy to dig in further and try to optimize this more, but until that's certain I think it's best to use the standard Cloudflare methodology for the time being.

packages/adapter-cloudflare-workers/index.js
```diff
@@ -51,13 +54,18 @@ export default function () {
...
 			await esbuild.build({
 				entryPoints: [`${tmp}/entry.js`],
-				outfile: `${entrypoint}/index.js`,
-				bundle: true,
+				outfile: `${entrypoint}/${main_path}`,
 				target: 'es2020',
-				platform: 'browser'
+				platform: 'browser',
+				bundle: true,
+				external: ['__STATIC_CONTENT_MANIFEST'],
```
pzuraq, line 61: One thing I'd like to see from #4576 is the ability to pass in `esbuild` options, which is something `adapter-cloudflare` supports. We occasionally need to mark external dependencies for legacy libraries that reference Node utils, or provide polyfills.

dominikg: added and rebased.  cf adapter allows overriding `platform` and `target` so it is allowed here too or should it be fixed for workers?

Rich-Harris (approved):

rauleite:
I had to add this line too, in `[build.upload]`:

`dir = "./.cloudflare/worker"`

Otherwise it will try to use the default "./build" directory.

Issue #3852, opened 2022-02-11
Routing in app built by adapter-static is broken in next.266


jamesb93:
### Describe the bug

Since upgrading to the latest version of SvelteKit a site which is compiled with the static adapter no longer works when served via Nginx.

For example routing to https://mywebsite.com works but routing to https://mywebsite.com/hello results in 404. If I were to navigate to the second link internally via an `<a>` the 404 would not appear. I can resolve the issue by pinning the SvelteKit version to something like 250 but being on the HEAD manifests the error.

### Reproduction

1. Build any site with static adapter and SvelteKit version 266
2. Host output with a web server such as nginx
3. Route to anything except index.html directly or navigate there and refresh. For example https//mysite.com/anotherpage

### System Info

```shell
System:
    OS: macOS 12.2
    CPU: (8) arm64 Apple M1
    Memory: 553.13 MB / 16.00 GB
    Shell: 3.3.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 17.4.0 - /opt/homebrew/bin/node
    Yarn: 1.22.10 - /opt/homebrew/bin/yarn
    npm: 8.3.1 - /opt/homebrew/bin/npm
  Browsers:
    Brave Browser: 96.1.32.115
    Chrome: 98.0.4758.80
    Firefox: 97.0
    Safari: 15.3
  npmPackages:
    @sveltejs/adapter-static: next => 1.0.0-next.28 
    @sveltejs/kit: next => 1.0.0-next.266 
    svelte: ^3.46.4 => 3.46.4
```

### Severity

serious, but I can work around it

fzembow:
@jamesb93 looks like they changed the config so that `trailingSlash` must be set to `"always"` in the kit config in order to generate nested index.htmls like was the default before.

https://kit.svelte.dev/docs/configuration#trailingslash

@Rich-Harris not sure about changing the default (per [your PR](https://github.com/sveltejs/kit/pull/3801)) ... at least on a website hosted using the default settings for S3 hosting, `something.html` is _not_ available at `/something`

jamesb93:
Thanks @fzembow, this explains why I'm seeing the issue I can see.

Oliver-Piorun:
@fzembow Hi! Setting `trailingSlash` to `"always"` does not work for me. It breaks fetching of assets (for example).

Without it:
Browser tries to fetch `/logo.png`

WIth it:
[...]

PR #4279, opened 2022-03-09
[docs] highlight need to set trailingSlash with adapter-static

benmccann:
Closes https://github.com/sveltejs/kit/issues/3852


Conduitry:
Do we want to add a note about what `'always'` does, saying how it writes the file to `/a/index.html` instead of `/a.html`? This might be something just for the adapter readme and not the main docs.

PR #4289, opened 2022-03-10
expose `event.clientAddress`

Rich-Harris:
Supersedes #3993. This adds a new property to `event`, `clientAddress`, which adapters are expected to populate with the client IP address.

Will be very grateful for any contributions to this PR!

* [X] Populate `clientAddress` in dev and preview
* [x] Update adapters
    * [X] Cloudflare/Cloudflare Workers
    * [X] Netlify
    * [x] Node (possibly using [request-ip](https://github.com/pbojinov/request-ip)?) Needs to be behind an option of some sort, since we don't want to trust proxies on behalf of users
    * [X] Vercel
    * [x] Figure out which may have multiple values appended together (e.g. `x-forwarded-for` may) and whether we should split them
* [x] Update docs (happens automatically by updating types)
* [ ] ~~Warn on startup if `getClientAddress` is not provided~~ changed my mind
* [x] Throw on accessing `event.clientAddress` if `getClientAddress` is not provided

### Tests

### Changesets


packages/kit/src/core/dev/plugin.js
```diff
@@ -259,62 +259,72 @@ export async function create_plugin(config, cwd) {
...
+								trailing_slash: config.kit.trailingSlash
 							},
-							template_contains_nonce: template.includes('%svelte.nonce%'),
-							trailing_slash: config.kit.trailingSlash
-						});
+							{
+								getClientAddress: () => {
+									const { remoteAddress } = req.socket;
+									if (remoteAddress) return remoteAddress;
+									throw new Error('Could not determine clientAddress');
```
Conduitry, line 324: When are we imagining this might happen?

packages/kit/src/core/preview/index.js
```diff
@@ -137,7 +137,16 @@ export async function preview({ port, host, config, https: use_https = false })
...
 			}
 
-			setResponse(res, await server.respond(request));
+			setResponse(
+				res,
+				await server.respond(request, {
+					getClientAddress: () => {
+						const { remoteAddress } = req.socket;
+						if (remoteAddress) return remoteAddress;
+						throw new Error('Could not determine clientAddress');
```
Conduitry, line 146: When are we imagining this might happen?

packages/kit/src/core/dev/plugin.js
benmccann: From https://nodejs.org/api/net.html#socketremoteaddress:

> Value may be `undefined` if the socket is destroyed (for example, if the client disconnected).

Rich-Harris:
I just checked, and on Vercel `x-forwarded-for` only contains the client IP address so we're in the clear there. For Node we will want to do something like this

```js
const [clientAddress] = headers.get('x-forwarded-for').split(',');
```

packages/adapter-node/index.d.ts
```diff
@@ -19,6 +19,7 @@ interface AdapterOptions {
 			host?: string;
 		};
 	};
+	trustProxy?: boolean;
```
benmccann, line 22: this should probably be added to the README

packages/adapter-netlify/src/handler.js
```diff
@@ -10,7 +10,12 @@ export function init(manifest) {
 	const server = new Server(manifest);
 
 	return async (event, context) => {
-		const rendered = await server.respond(to_request(event), { platform: { context } });
+		const rendered = await server.respond(to_request(event), {
+			platform: { context },
+			getClientAddress() {
+				return event.headers['x-nf-client-connection-ip'];
```
benmccann, line 16: @ascorbic I wanted to confirm with you if we're doing this correctly because it seems to be undocumented, but is what is suggested in this support post: https://answers.netlify.com/t/is-the-client-ip-header-going-to-be-supported-long-term/11203

benmccann, line 16: @ascorbic I wanted to confirm with you if we're doing this correctly because it seems to be undocumented, but is what is suggested in this support post: https://answers.netlify.com/t/is-the-client-ip-header-going-to-be-supported-long-term/11203

packages/adapter-node/index.d.ts
Rich-Harris: done

Rich-Harris:
The last outstanding issue (I think) is whether the `trustProxy` option should be called something else (for the rare cases where the server is running without a proxy), and whether we should fall back to `req.socket.remoteAddress` when it's `false`. I'm inclined to say 'no' to the latter case, since people won't then see the error message, and will likely get the wrong IP address.

Conduitry:
What's the purpose of `trustProxy` for the Node adapter if it just toggles between two behaviors: `event.clientAddress` always throws an error, and `event.clientAddress` always uses the same series of fallbacks to determine the client IP? If it were an option, I'd expect it to let me specify how the client IP is determined in some way - what headers are looked at, etc.

Rich-Harris:
@Conduitry i have no strong opinions here whatsoever — what sort of API would make sense to you?

Conduitry:
@Rich-Harris I don't know. I was intending that more as an argument for just getting rid of `trustProxy` always using the same order of fallbacks as `trustProxy: true` does. We could document in the Node adapter's readme what this order is. I'd expect this to be a somewhat niche feature. If we eventually need to allow this to be customized, we can take inspiration from whatever issues are opened and what people are asking for.

Conduitry:
I suppose what strikes me as odd about the current proposed API is that it feels like getting people to sign a waiver or check a terms-and-conditions box before they can use a certain feature at all. It seems akin to another proposed change on Svelte core (which I also don't like) of renaming `{@html}` to `{@dangerousInnerHtml}` for the benefit of people who don't read and the expense of the irritation of everyone else.

Conduitry:
I'll let others speak to the details of this, but I'm now retracting my above suggestion, because from some discussions it sounds like there is no reasonable default for which header(s) to use. If when someone is opting in to this behavior they also need to opt in to using specific headers (as opposed to just opting in to the feature as a whole), I no longer have the same qualms about requiring them to opt in.

packages/adapter-netlify/src/handler.js
ascorbic: @benmccann Yes, that's correct. We recently cleaned up the internal headers sent to functions so that all remaining ones are now officially supported. https://answers.netlify.com/t/upcoming-change-stripping-exposed-netlify-headers-from-function-and-proxy-requests/52665

evdama:
The way I see it is that `request-ip` for has a particular order of its fallbacks... the [same order as is now about to be merged into sveltekit](https://github.com/sveltejs/kit/pull/4289/commits/7aca944ad2608fe40014f9ea5a2fbdc573ab2580#diff-a1607538c0a408f7696fff4d8e6efdc5752eee5b8db7222e36ad3e726a5c14a1) (thanks a ton btw 🥳 ).

My experience of using Sapper with nodejs/express and `request-ip` inside a firebase cloud function is that it worked correctly as it traversed through its fallbacks i.e. I got the correct client IP address and not some intermediary IP from some of Google's proxies or other infrastructure components.

Given that, I'd say why not try without having a `trustProxy` check in place, and see how it goes... it's always possible to add more config when and where needed later, based on user reports, just like mentioned by Conduitry.

TheRealThor:
I don't understand the
`throw new Error('You must enable the adapter-node trustProxy option to read clientAddress');`

If I don't enable the trustProxy options, the clientAddress should still be read from the socket that is connected to the server otherwise this feature would be limited to use case of a node-server behind a http proxy, which I don't assume is the majority of node servers out there

Rich-Harris:
Alright, `adapter-node` continues to be a PITA. 

I've been reading https://adam-p.ca/blog/2022/03/x-forwarded-for/ and the inescapable conclusion is that request-ip is horribly insecure, and we mustn't emulate its behaviour. Instead, users of this adapter must specify which header to get `event.clientAddress` from (e.g. `True-Client-IP` or `CF-Connecting-IP`). 

In the case where the header is `X-Forwarded-For`, we need to specify which address to use. The best API I could come up with is specifying an `xForwardedForIndex` option, where positive numbers count from the left and negative numbers count from the right.

Rich-Harris:
Also, in 1fa68232c30808dd9a6e33fb64f37ee4ce27d94c `socket.remoteAddress` is returned from `getClientAddress` if no header is specified or the header isn't present, so the proxy-less case works OOTB.

evdama:
- I find that a good solution as node can be deployed in various ways and to many hosting platforms/providers, so there's no way to know what header to use here, thus the user has to provide this information via config 👍
- For other adapters such as vercel, netlify etc. it's straight forward as it's known beforehand which header carries the IP address of the remote client connecting.

TheRealThor:
Thank you @Rich-Harris, I think it perfect now.

AlessioGr:
Would be awesome if someone could make an example as to how/where to use it. I'd like to have the IP address available in my svelte component - where exactly do I start? Where can I find the `event.clientAddress` field?

ollydixon:
Hey guys, this returns an invalid address on our production servers. Using node express (same server groups) returns a valid address so something is broken with ```clientAddress```

@AlessioGr yeah they are bad at giving examples for people learning.

```
export async function post(data) {
    data.clientAddress
}
```

ollydixon:
https://github.com/sveltejs/kit/issues/5348

CAPTAIN320:
> Would be awesome if someone could make an example as to how/where to use it. I'd like to have the IP address available in my svelte component - where exactly do I start? Where can I find the `event.clientAddress` field?

I made an example of how to get a user's ip address and subsequently the user's location using sveltekit.

Repo - https://github.com/CAPTAIN320/sveltekit-user-ip-location-example

CAPTAIN320:
> Would be awesome if someone could make an example as to how/where to use it. I'd like to have the IP address available in my svelte component - where exactly do I start? Where can I find the `event.clientAddress` field?

https://github.com/CAPTAIN320/sveltekit-user-ip-location-example

a4vg:
@CAPTAIN320 I see you're [exporting a variable on hooks.server.ts](https://github.com/CAPTAIN320/sveltekit-user-ip-location-example/blob/f98db3d045693835f1204260dd4777255a3febe0/src/hooks.server.ts#L3). Since it runs on the server side, isn't the variable shared by all clients? The correct approach should be getting the ip in the `load` function of `+page.server.ts` and returning it so it's available on `+page.svelte` in the `data` variable.

CAPTAIN320:
> @CAPTAIN320 I see you're [exporting a variable on hooks.server.ts](https://github.com/CAPTAIN320/sveltekit-user-ip-location-example/blob/f98db3d045693835f1204260dd4777255a3febe0/src/hooks.server.ts#L3). Since it runs on the server side, isn't the variable shared by all clients? The correct approach should be getting the ip in the `load` function of `+page.server.ts` and returning it so it's available on `+page.svelte` in the `data` variable.

@a4vg Getting the variable from the load function and having your app is deployed to a Paas such as Vercel, will result in the location of the vercel server being returned. Which was why I was getting US location for all my users.
Getting it from hooks.server.ts however, will prevent this behaviour and return the client ip instead.

From then on you can import it to any *.server.ts file.

a4vg:
@CAPTAIN320 Maybe the solution works, but you should write the IP to `event.locals` ([docs](https://kit.svelte.dev/docs/hooks#server-hooks-handle)) and then access it on the `load` function instead of exporting a variable from `hooks.server`. The reason is because `hooks.server` runs on the server side and the variables that you put there will be shared by all users (correct me if I'm wrong, I'm just starting to use sveltekit), so it won't work as expected when multiple users try to access the application, one user could end up getting the IP from another user.

------

Now, for anyone else who can't still get the public IP address...

I was using Heroku FWIW, but I had the same issue where I couldn't get the real IP. `getClientAddress` was reporting a local IP.

I'm using the node-adapter and after reading this note on [the docs](https://github.com/sveltejs/kit/blob/master/packages/adapter-node/src/handler.js#L98):
> If you need to read the left-most address instead (and don't care about spoofing) — for example, to offer a geolocation service, where it's more important for the IP address to be real than trusted, you can do so by **inspecting the x-forwarded-for header within your app**.

I added
```ts
// +page.server.ts
export const load: PageServerLoad = ({ request }) => {
    const ip = request.headers.get('x-forwarded-for') || '';
```

Another option is setting the env var `ADDRESS_HEADER="X-Forwarded-For"` and `getClientAddress` will start reporting the correct IP.

The node-adapter code shows exactly how `getClientAddress` works https://github.com/sveltejs/kit/blob/96993cf69a38e44ddaf31ee5fa2436f889772fbc/packages/adapter-node/src/handler.js#L131-L136

Carlos-err406:
setting the `ADDRESS_HEADER="X-Forwarded-For"` 
and using `getClientAddress()` 
worked just fine for me

fxmt2009:
I'm a beginner. How do you do that? Where do you set the env var ADDRESS_HEADER? 

@Carlos-err406 
setting the ADDRESS_HEADER="X-Forwarded-For"
and using getClientAddress()
worked just fine for me

Carlos-err406:
i did the test on a remote server
(if is worth anything im using adapter node)
i set up the app to log the output of the getClientAddress function so i could check on it and indeed accessing from my PC to the app logged the samy IP i get from whatsmyip.com
tried from my phone as well (so a different IP was logged) 

as for where to set the enviroment variable it was on my root .env as a regular enviroment variable, no further implications were required
(i ran the app with `node -r dotenv/config build` as the documentation says, if it helps)

fxmt2009:
That was detailed and specific. Impressive and helpful. Thank you.

leoj3n:
Here is a more complete example for anyone that it helps...



`+page.server.js`:

```js
/** @type {import('./$types').PageServerLoad} */
export const load = async (event) => {
    let requestIp;
    
    try {
      requestIp = event.getClientAddress(); // IP from Client Request
      console.log('IP Address from Client Request: ', requestIp);
    } catch (error) {
      console.log('Error reading IP');
    }

    return { clientAddress: requestIp }
}
```

`+page.svelte`:

```svelte
<script>

	/** @type {import('./$types').PageData} */
	export let data;

	onMount(async () => {
		await import('@mux/mux-player');
	});
</script>

<mux-player
  playback-id="..."
  metadata-viewer-user-id="{data.clientAddress}"
></mux-player>
```

This worked deployed to Vercel with `@sveltejs/adapter-vercel`.

Didn't seem to need to add any environment config like `ADDRESS_HEADER="X-Forwarded-For"`.

travisdmathis:
anyone that's trying to figure out how to do this in 2024 from an API endpoint in Svelte 5, it's like this

```
export async function POST({ getClientAddress, request, url }) {
 let clientIp = getClientAddress();
 ```
 
 Save yourself some time.

Issue #3030, opened 2021-12-12
Environment variables on `process.env` sometimes not available in hooks.js in local dev


josefaidt:
### Describe the bug

I'm looking to use the `handle` hook to add authentication and authorization middleware to my API routes, however the variables needed to verify user JWT's are _sometimes_ not available when the hook runs, but _only_ in local development. In my experience these variables were available just once after upgrading dependencies to use Vite 2.7.1, however on subsequent attempts/restarts of the dev server these variables were no longer available and thus all API attempts would be unauthorized. I've attempted to remove the `.svelte-kit` directory, node_modules, lockfile, and reinstall to no avail.

It is important to note this is only experienced during local development with `svelte-kit dev`, and may actually be a Vite-related issue (given the Svelte-Kit scope I decided to file here first to see what y'all's thoughts are).

This can be mitigated by adding a small Vite plugin to load secrets with `dotenv`

```diff
// svelte.config.js
// ...
/** @type {import('@sveltejs/kit').Config} */
const config = {
  kit: {
    adapter: adapter(),

    // hydrate the <div id="svelte"> element in src/app.html
    target: '#svelte',

+   vite: {
+     plugins: [
+       (function LoadSecrets() {
+         return {
+           name: 'load-secrets',
+           configureServer: async () => {
+             ;(await import('dotenv')).config()
+           }
+         }
+       })()
+     ]
+   }
  }
}
```

And for reference, the `hooks.js` file:

```js
export async function handle({ request, resolve }) {
  let user = await authenticate(request) // verifies JWT, requires 3 server-side env vars

  request.locals.user = user
  request.locals.isAuthenticated = !!user

  if (request.path.startsWith('/api')) {
    if (!user) {
      return {
        status: 401,
        body: JSON.stringify({
          error: {
            message: 'Unauthorized'
          }
        })
      }
    }

    if (request.path.startsWith('/api/admin') && !isAdmin(user)) {
      return {
        status: 403,
        body: JSON.stringify({
          error: {
            message: 'Invalid Admin'
          }
        })
      }
    }
  }

  const response = await resolve(request)
  return response
}
```

### Reproduction

1. Create a Svelte-Kit app with `npm init svelte@next my-app`
2. Create a `.env` file in the project root, add an environment variable
	```
	MY_VAR="hello, world!"
	```
3. Create `src/hooks.js`:
	```js
    export async function handle({ request, resolve }) {
      console.log(process.env['MY_VAR']) // sometimes undefined
[...]

Issue #3004, opened 2021-12-08
Env variables are not accessed by vite in PROD 


narender2031:
### Describe the bug

I am using the docker image for the production build. I have set my environment variables in docker and they are accessible by using `env`. But when I build the app these are not accessed by Vite. 

Please help me here!

### Reproduction

Create Svelte kit app by started. 
- create the docker image
- Create the env variables in the docker system. 
- use the VITE_ as a prefix. 

### System Info

```shell
Undefined
```

### Severity

blocking all usage of SvelteKit

Conduitry:
`VITE_` variables are bundled during build, not looked at at runtime in prod mode. If your Dockerfile is setting them at app runtime with `ENV`, this is expected. If that's not what's happening, we need a proper reproduction.

narender2031:
> `VITE_` variables are bundled during the build, not looked at runtime in prod mode. If your Dockerfile is setting them at app runtime with `ENV`, this is expected. If that's not what's happening, we need proper reproduction.

@Conduitry Yes, Dockerfile is setting them at app runtime with `ENV`. 
Do you know any workaround for this.

narender2031:
@Conduitry 

.env
```
VITE_SOME_ENV=hello
```

variable.ts
```
console.log(import.meta.env) // {VITE_SVELTEKIT_AMP: '', BASE_URL: '/_app/', MODE: 'production', DEV: false, PROD: true}
export const variables = {
	SOME_ENV: <string>import.meta.env.VITE_SOME_ENV,
};
```
NOTE: import.meta.env don't have `VITE_SOME_ENV` after the build

api.ts

```
import { variable } from './variable.ts'

const data = variable.SOME_ENV // const data = "undefined "
```
after the build `const data = "undefined "`  it replaced by `undefined`.


My docker image already has all the env variable inside the system `env`. When I build the file my env data is replaced by undefined.

rmunn:
What happens if you change `import.meta.env.SOME_ENV` to `import.meta.env['SOME_ENV']`? I remember a while ago Vite was doing text replacement specifically looking for the text `import.meta.env.` (with a trailing period), but I haven't looked at it in months and don't know if that's still current Vite behavior. But if `import.meta.env['SOME_ENV']` works for you then that could be an easy workaround.

benmccann:
@rmunn there should be some improvement on that front with Vite 2.7: https://github.com/vitejs/vite/pull/5404

[...]

PR #4293, opened 2022-03-11
feat: Runtime SVELTE_PUBLIC_* env variables

bfanger:
Using environment variables in SvelteKit is possible via [vite's `VITE_*` variables](https://vitejs.dev/guide/env-and-mode.html#production-replacement).

The downside is that these are **compile-time** environment variables, the values are injected into the javascript files during the **dev** and **build** steps.

This PR adds **runtime** environment variables, this allow you to change the _variable_ in the _environment_ without needing to recompile.

### Use-cases:
- Setting an environment variable and restarting the server works. This allows changing a [Config Var](https://devcenter.heroku.com/articles/config-vars) on Heroku to activate instantly. (No longer needs a image rebuild & deploy)
- Using an Azure Devops Build pipeline to [create an Artifact](https://docs.microsoft.com/en-us/azure/devops/pipelines/publish-pipeline-artifact?view=azure-devops&tabs=yaml) and deploy that build to differrent environments using release pipelines.

## Usage example:

```
import { env } from "$app/env";

console.log(env.SVELTE_PUBLIC_API_ENDPOINT);
```

## How it works
At server startup a `<script type="svelte/env">` tag is generated containing values from `process.env` that start with `SVELTE_PUBLIC_`.
When importing the  `$app/env` the `env` export is populated with values from `process.env` on the server and on the client it uses the values from the 'svelte/env' tag.

## Possible future improvements:

- Add support for .env files. (Workaround `node -r dotenv/config node_modules/.bin/svelte-kit dev`)
- Allow overriding the `SVELTE_PUBLIC_` prefix/filter in the svelte.config
- Allow adapters to provide enviroment variables. (`process.env` is a NodeJS api)
- The `SVELTE_PUBLIC_` prefix naming inspiration came from [NEXT_PUBLIC_](https://nextjs.org/docs/basic-features/environment-variables), but these are also compile-time variables (via webpack)

### Tests

### Changesets

Related #3030
[Alternative implementation using hooks.ts](https://github.com/bfanger/svelte-project-template/blob/main/src/hooks.ts)

### DIFF
--- a/.changeset/new-apples-confess.md
+++ b/.changeset/new-apples-confess.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Add support for runtime environment variables (SVELTE*PUBLIC*\*)
--- a/packages/kit/src/runtime/app/env.js
+++ b/packages/kit/src/runtime/app/env.js
@@ -14,5 +14,22 @@ export const mode = import.meta.env.MODE;
  * @type {import('$app/env').amp}
  */
 export const amp = !!import.meta.env.VITE_SVELTEKIT_AMP;
+/**
+ * @type {import('$app/env').env}
+ */
+export const env = extractEnv();
 
 export { prerendering } from '../env.js';
+
+function extractEnv() {
+	if (typeof process === 'object' && typeof process.env === 'object') {
+		return process.env;
+	}
+	if (typeof document !== 'undefined') {
+		const el = document.querySelector('script[type="svelte/env"]');
+		if (el) {
+			return new Function('return ' + el?.textContent)();
+		}
+	}
+	return {};
+}
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -7,6 +7,8 @@ import { s } from '../../../utils/misc.js';
 import { create_prerendering_url_proxy } from './utils.js';
 import { Csp, csp_ready } from './csp.js';
 
+const envScript = renderEnvScript();
+
 // TODO rename this function/module
 
 const updated = {
@@ -137,6 +139,7 @@ export async function render_response({
 
 		try {
 			rendered = options.root.render(props);
+			rendered.head += envScript;
 		} finally {
 			unsubscribe();
 		}
@@ -354,4 +357,20 @@ function serialize_error(error) {
 		serialized = '{}';
 	}
 	return serialized;
+}
+
+function renderEnvScript() {
+	/** @type {Record<string, string | undefined>} */
+	const env = {};
+	if (typeof process === 'object' && typeof process.env === 'object') {
+		Object.entries(process.env).forEach(([name, value]) => {
+			if (name.startsWith('SVELTE_PUBLIC_')) {
+				env[name] = value;
+			}
+		});
+	}
+	if (Object.keys(env).length === 0) {
+		return '';
+	}
+	return `\n    <script type="svelte/env">${devalue(env)}</script>`;
 }
--- a/packages/kit/types/ambient.d.ts
+++ b/packages/kit/types/ambient.d.ts
@@ -67,6 +67,11 @@ declare module '$app/env' {
 	 * By default, `svelte-kit dev` runs with `mode=development` and `svelte-kit build` runs with `mode=production`.
 	 */
 	export const mode: string;
+	/**
+	 * Runtime environment variables.
+	 * Only variables starting with "SVELTE_PUBLIC_" are available on the client, the server has access to all variables.
+	 */
+	export const env: Record<string, string | undefined>;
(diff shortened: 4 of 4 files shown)


Conduitry:
Does `process.env.*` even work in non-Node environments?

I'm -1 on doing something like this. If people want to expose runtime variables to the client (of whichever kind are supported in their deployment environment), they can just do so manually in `getSession`.

dominikg:
This scenario and how it can be achieved is mentioned in the faq:  https://kit.svelte.dev/faq#env-vars
and semi-recently here https://github.com/sveltejs/kit/issues/3004#issuecomment-1026471269

first class support for it in kit - and if that is something we want to add  at all - should have been discussed in a feature-request first.
Also the way this PR adds a script tag for the client is not safe, crafted env values could cause script injection, see https://github.com/sveltejs/kit/pull/4128

bfanger:
Agree with the `process.env` argument (on the [svelte-adapter-deno](https://github.com/pluvial/svelte-adapter-deno) should use `Deno.env.toObject()` for example), any pointers on how to implement this with multiple runtimes in mind are welcome.

If a hacker can modify the env variables to generate executable javascript you are already lost, but thanks for the pointer to the `render_json_payload_script` utility.

I didn't consider getSession as I made assumptions it was like something PHP's `$_SESSION` and should be a place for data that depends on the user session. Diving deeper in how getSession is implemented and behaves in kit it's has the exact behavior needed for sharing the environment variables. 

- Works in adapter-static 
- Doen't result in additional fetch requests to sync the session when navigating.

`getSession()` is like a `exposeTheseVariablesToClientOnce(serverRequest)`

A slight advantage of this PR over getSession is that you can use `import { env } from "$app/env"` inside an endpoint.
(But as dotenv is not part of kit, its identical to process.env )

Feel free to treat my PR's as a feature-requests. it just has an example implementation as a bonus.

Rich-Harris:
Thanks for the PR — we definitely _do_ need to come up with a consistent/abstracted approach to dealing with environment variables, since `process.env` isn't available in all platforms. Quite how we do that is very much up for debate — we need ticks in all four quadrants:

|                        | build time               | run time                                  |
|------------------------|--------------------------|-------------------------------------------|
| public (client/server) | ✅ import.meta.env.VITE_* | ❌                                         |
| private (server-only)  | ❌                        | ⚠️ process.env.* (doesn't work everywhere) |

And all this would ideally

* work with `.env` and `.env.*` files
* enable dead code elimination etc at build time
* allow autocompletion/auto-imports
* feel consistent across the four quadrants, but with clearly understandable distinctions between them (e.g. `$app/env/public` vs `$app/env/private` or whatever)

Aside: I've always found `import.meta.env` to be a deeply weird construct — it feels like we're just using the syntax that happens to be lying around, rather than designing something. 

I'm not wholly convinced that `getSession` is the answer, since you're then restricted to using these values inside components. You can certainly imagine a module like this:

```js
import { env } from '$app/env';
import { create_api } from '$lib/db';

export const api = create_api(env.PUBLIC_BASE_URL);
```

That said, it's not wholly clear to me what kinds of things would sit in that top right quadrant. Isn't that sort of thing generally known at build time?

I guess I should move this comment into a new issue.

bfanger:
I quickly ran into the `Function called outside component initialization` when trying to use the `getSession` approach.

I could use dependency injection and pass the $session.ENV as a parameter to the functions, but in my opinion that creates unnecessary boilerplate code.
Injecting the values into a global store/variable is also not a good idea, because that creates subtle bugs where the order of navigation matters.

So i'll keep using the workaround with the [handle Hook](https://kit.svelte.dev/docs/hooks#handle), which i'll share here:

**src/app.html**
Change:

```
    %svelte.head%
```

Into

```
    %svelte.head%
    <script type="svelte/env"></script>
```

**src/hooks.js**

```js
import dotenv from "dotenv";

dotenv.config();

const envScript = `<script type="svelte/env">${JSON.stringify(
  Object.fromEntries(
    Object.entries(process.env).filter(([key]) =>
      key.startsWith("SVELTE_PUBLIC_")
    )
  )
)}</script>`;

  const body = await response.text();
  return new Response(
    body.replace('<script type="svelte/env"></script>', envScript),
    response
  );
}
```

**src/lib/env.js**

```js
let env = {};
if (typeof process === "object" && typeof process.env === "object") {
  env = process.env;
} else if (typeof document !== "undefined") {
  const el = document.querySelector('script[type="svelte/env"]');
  if (el) {
    env = JSON.parse(el.textContent);
  }
}
export default env;
```

shiftlabs1:
@bfanger I just came across your proposed solution here while looking for the best way to deal with this issue. I found though that the first time my app runs, it fails as the variables are only loaded after the actual path resolves. these values become available only after a reload is done.

PR #4295, opened 2022-03-11
[fix] check for app.d.ts rather than global.d.ts

benmccann:
Looks like we forgot to update this line when `global.d.ts` was renamed to `app.d.ts`


dominikg (approved):

Rich-Harris:
ah i wondered why this file wasn't getting included. d'oh

PR #4302, opened 2022-03-11
make manifest.mimeTypes part of the public API

Rich-Harris:
This moves `manifest._.mime` to `manifest.mimeTypes`, so that adapters can use it. This will enable #4276 to serve static assets without pulling in a mime library

### Tests

### Changesets


benmccann (approved):
lgtm. Though I wonder if we should call it `mimeType`?

Rich-Harris:
Changed it to `mimeTypes`

PR #4304, opened 2022-03-12
[docs] adapter-static: document prerender.default and info about disabling ssr in dev

Theo-Steiner:
#4192 introduced a configuration setting for prerendering by default, which explicitly has to be set to true for adapter-static to work in non-SPA mode. Currently adapter-static only tells the user to do so when actually building, but having it in the docs is good for obvious reasons.
I also lumped in a small mention about disabling ssr in the handle() hook, to make developing an SPA with svelte-kit aligned with how it will actually works when built. I hope it's okay that I made a single PR for two changes, just let me know if changes are required!

### Tests

### Changesets

### DIFF
--- a/packages/adapter-static/README.md
+++ b/packages/adapter-static/README.md
@@ -12,18 +12,18 @@ import adapter from '@sveltejs/adapter-static';
 
 export default {
 	kit: {
+		// Enable prerendering by default unless when in SPA mode
+		prerender.default: true,
 		adapter: adapter({
 			// default options are shown
 			pages: 'build',
 			assets: 'build',
 			fallback: null,
 			precompress: false
-		})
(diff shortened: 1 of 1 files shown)


packages/adapter-static/README.md
```diff
@@ -12,6 +12,8 @@ import adapter from '@sveltejs/adapter-static';
 
 export default {
 	kit: {
+		// Enable prerendering by default unless when in SPA mode
```
benmccann, line 15:
```suggestion
		// Enable prerendering by default. Remove if using SPA mode
```

Rich-Harris (approved):

Rich-Harris:
thanks!

PR #4305, opened 2022-03-12
allow `InputProps` and `OutputProps` to be typed in generated `Load`

ivanhofer:

### Tests

### Changesets

Same fix as here: https://github.com/sveltejs/kit/pull/4160
But this time for the generated `Load` type.

I also have restricted all the `Params` and `Props` types with the `extends` keyword.
Maybe it makes sense to create a "base type" and use it everywhere:

```diff
+type BaseParams = Record<string, string>
+type BaseProps = Record<string, any>

export interface Load<
-	Params extends Record<string, string> = Record<string, string>,
-	InputProps extends Record<string, any> = Record<string, any>,
-	OutputProps extends Record<string, any> = InputProps
+	Params extends BaseParams = BaseParams,
+	InputProps extends BaseProps = BaseProps,
+	OutputProps extends BaseProps = InputProps
> {
	(input: LoadInput<Params, InputProps>): MaybePromise<
		Either<Fallthrough, LoadOutput<OutputProps>>
	>;
}


ivanhofer:
I also have noticed the same issue with the generated `RequestHandler` type and included the fix in this PR.

I'm not sure what to do with the `Body` type from the "private" types. Can I move it to `index.d.ts`?

Rich-Harris:
Thanks! Yeah, it makes sense to expose `Body`. I've updated this PR so that it's exposed as `ResponseBody`

Rich-Harris (approved):

Issue #4303, opened 2022-03-12
Cannot get props in errors returned by shadowed route


ItalyPaleAle:
### Describe the bug

When a shadowed (get) route returns an error, the `load` method does not receive anything from the body of the response of the shadowed route. See repro for example.

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-euapgy

Navigate (manually) to the `/err` page and you'll notice that `message` is `undefined`.

### System Info

```shell
System:
    OS: macOS 12.1
    CPU: (8) arm64 Apple M1
    Memory: 99.84 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.6.2 - ~/Library/Caches/fnm_multishells/67621_1646621511738/bin/node
    npm: 7.20.3 - ~/Library/Caches/fnm_multishells/67621_1646621511738/bin/npm
  Browsers:
    Chrome: 99.0.4844.51
    Edge: 97.0.1072.55
    Firefox: 99.0
    Firefox Developer Edition: 98.0
    Safari: 15.2
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.295 
    svelte: ^3.46.4 => 3.46.4
```

### Severity

serious, but I can work around it

Rich-Harris:
This is a duplicate of #3715 — the solution is being worked on in #4032

PR #4313, opened 2022-03-12
[fix] Cannot get props in errors returned by shadowed route #4303

aolose:
Allow display custom error message from shadow.

e.g
```
export function get() {
	return {
		body: {
			error: 'A custom error message'
		},
		status: 451
	};
}
```

fix: #4303

test:
```
cd packages\kit\test\apps\basics
npx  playwright test -g 'Display custom error message on client side shadow page'
npx  playwright test -g 'Display custom error message on server side shadow page'
```

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -650,8 +650,8 @@ export function create_client({ target, session, base, trailing_slash }) {
 							status = res.status;
 							let error_message = 'Failed to load data';
 							try {
-								const { message } = await res.json();
-								if ('string' === typeof message) error_message = message;
+								const { error } = await res.json();
+								if ('string' === typeof error) error_message = error;
 							} catch (e) {
 								// ignore
 							}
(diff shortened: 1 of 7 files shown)


packages/kit/src/runtime/client/client.js
```diff
@@ -648,7 +648,14 @@ export function create_client({ target, session, base, trailing_slash }) {
 							props = await res.json();
 						} else {
 							status = res.status;
-							error = new Error('Failed to load data');
+							let error_message = 'Failed to load data';
+							try {
+								const { message } = await res.json();
```
ItalyPaleAle, line 653: Should this be `message` or `error`?

ItalyPaleAle:
Thank you for submitting a PR this quickly after I opened an issue :)

packages/kit/src/runtime/client/client.js
aolose: I think `error` is more explicit.

Rich-Harris:
Thank you — this is being worked on in #4032

Issue #4265, opened 2022-03-07
Why is `preredering` from `$app/env` always `false`, even during `pnpm build`?


ar4hc:
### Describe the bug

i use hooks.js to init a redis connection.
And in the build process hooks gets called and tries to connect to redis.
Which is in a docker container and far away and not availabe during build, 
so it breaks and crashes the build.

So i tried `prerendering`, and return something "empty" but this fails too as it is always reported as `false`...

My code looks like this:

```js
import { prerendering } from '$app/env';

console.info('Prerendering:', prerendering);

const client = createClient(
  // [...]
);

export const handle = async ({ event, resolve }) => {
		if (!client.isOpen) await client.connect();
  // [...]
}
```

and in handle() i'd try to wrap some of that code in `if (prerendering)...`, and avoid the db connection.

i'm not sure i understand how `prerendering` and hooks should work here, but i think it should work that way...

### Reproduction

run pnpm build in shell, `docker build`, or `docker compose build`

### Logs

```shell
$ pnpm build

> classifier-gui@0.0.1 build /Users/ar/src/lincare/classifier/classifier-gui
> svelte-kit build

vite v2.8.6 building for production...
✓ 54 modules transformed.
.svelte-kit/output/client/_app/manifest.json                               2.11 KiB
.svelte-kit/output/client/_app/pages/__layout.svelte-687f8954.js           2.00 KiB / gzip: 1.06 KiB
.svelte-kit/output/client/_app/pages/__error.svelte-6723fd33.js            3.69 KiB / gzip: 1.20 KiB
[...]
vite v2.8.6 building SSR bundle for production...
✓ 62 modules transformed.
.svelte-kit/output/server/manifest.json                                    3.27 KiB
.svelte-kit/output/server/index.js                                         67.47 KiB
.svelte-kit/output/server/entries/endpoints/index.js                       2.17 KiB
.svelte-kit/output/server/entries/endpoints/faxlist.js                     1.17 KiB
.svelte-kit/output/server/entries/endpoints/api/index.js                   0.07 KiB
[...]

Prerendering: false <--- ouput from the console.log above...
```

### System Info

```shell
System:
    OS: macOS 11.6.4
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 1.43 GB / 32.00 GB
    Shell: 5.8.1 - /usr/local/bin/zsh
  Binaries:
    Node: 17.6.0 - /usr/local/bin/node
[...]

PR #4322, opened 2022-03-14
[fix] prerendering always false in global code of hooks.js

Theo-Steiner:

This fixes the behavior documented in #4265, where ``prerendering`` imported from ``'$app/env'`` is false outside the actual hooks in hooks.ts even when called during prerender.
This occurs because the hooks module is imported by the same module that exports the override function that allows to set prerendering in the first place. Therefore, when the global code of the hooks module is executed as it is imported, prerendering can never be any other value than it's default value.

This PR addresses this behavior by extracting the setting of prerendering into its own module (notify_prerendering.js), which can be imported without triggering the global code of the hooks module. As the notify_prerendering.js module will never be used anywhere other than during prerendering, it is not included into the ``writeServer()`` output of the builder.

### Tests

### Changesets

### Alternatives considered
Instead of making these changes, this behavior could also be documented under $app/prerendering as people can circumvent the problem with a setup function as described in #4265.


Rich-Harris:
Thank you. There's actually a simpler fix, which is more in line with plans for future changes (namely the addition of scoped `__middleware.js` files, which would likely take the place of `hooks.js`) — we import the hooks lazily rather than eagerly. I've updated the branch

Rich-Harris (approved):

Issue #4291, opened 2022-03-11
Replace fallthrough routes with param validators


Rich-Harris:
### Describe the problem

SvelteKit has an unusual feature called [fallthrough routes](https://kit.svelte.dev/docs/routing#advanced-routing-fallthrough-routes) which allows ambiguity between routes to be resolved by the routes themselves.

It's a powerful feature, but a niche one, and it comes at a significant cost. It adds non-trivial complexity to the SvelteKit codebase, makes things conceptually less clear, leads to undefined behaviour (should `event.locals` persist between different routes that are candidates for the same request?), has undesirable API implications (`load` has to return something, even if it doesn't return anything), and terrible performance implications in the case where a route _does_ fall through.

We could get 90% of the value of fallthrough routes with a fraction of the cost if we had some way of validating route parameters _before_ attempting to render or navigate.

### Describe the proposed solution

Credit for this idea goes to @mrkishi; any flaws were introduced by me. 

Suppose you're building a calendar app, and for some reason you're forced to have routes like this:

```
src/routes/[date]
src/routes/[eventid]
```

When someone visits `/2022-03-12`, either route could match. Rather than adding fallthrough logic to `src/routes/[date]`, we could add a validator that SvelteKit uses to check the format:

```js
// src/params/date.js
export default param => /^\d{4}-\d{2}-\d{2}$/.test(param);
```

Seeing that `2022-03-12` passes the test, we can select the `src/routes/[date]` route immediately.

If we know that every `eventid` is a UUID, we could test that too. Since UUIDs are a generic concept, we need a more generic name than `eventid`:

```
src/routes/[eventid=uuid]
```

Now, we can add a UUID validator...

```js
// src/params/uuid.js
export default param => /[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[89aAbB][a-f0-9]{3}-[a-f0-9]{12}/.test(param);
```

...and if someone visits `/neither-a-date-nor-a-uuid` we can 404 without doing anything else. Not only that, but the routes can have complete confidence that their parameters are valid.

(We could also do enum checks for things like `src/routes/[year]/[month]/[day]`...

```js
// src/params/month.js
const months = new Set('jan feb mar apr may jun jul aug sep oct nov dec'.split(' '));
export default param => months.has(param);
```

...though I don't have any immediate thoughts on how that would internationalize.)

### Validating vs parsing

[...]

PR #4328, opened 2022-03-15
render page with correct status code when non-get endpoint returns validation error

Rich-Harris:
Currently, if a page endpoint returns an error status code, it gets discarded. Just a failing test for now. I have a hunch that fixing it will be easier once we've killed fallthrough (#4291)

### Tests

### Changesets


packages/kit/src/runtime/load.js
```diff
@@ -3,6 +3,10 @@
  * @returns {import('types').NormalizedLoadOutput}
  */
 export function normalize(loaded) {
+	if (!loaded) {
+		return {};
+	}
+
```
Conduitry, line 9: Does this mean that returning a falsy value from a `load` function is now equivalent to returning `{}`? Is this desirable? If there were other corresponding changes to public types for this, I missed them.

Conduitry: Oh, I see elsewhere you also remove a comment `// TODO do we still want to enforce this now that there's no fallthrough?`, so, yes, it sounds like this is a conscious decision made at the same time as fixing the main bug. The types should still be updated.

Rich-Harris: updated the types and the docs

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -346,7 +344,7 @@ export async function load_node({
 				return proxy;
 			},
 			stuff: { ...stuff },
-			status: is_error ? status ?? null : null,
+			status: is_error ? status ?? null : shadow.status ?? null,
```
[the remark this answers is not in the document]
Rich-Harris: how do you mean — like this?

```suggestion
			status: (is_error ? status : shadow.status) ?? null,
```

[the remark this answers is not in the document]
Conduitry: That's what I had meant, yeah. I think it's a bit easier on the eyes, but I don't feel strongly about it.

I am still curious about needing `?? null` at all though, and what would behave differently if `status` were `undefined` rather than `null`.

PR #4330, opened 2022-03-15
Remove fallthrough

Rich-Harris:
#4291, part 1 of 2. This removes fallthrough routes but doesn't add parameter validators (that's part 2).

I have a hunch that we can simplify `client.js` further, but that can happen in a follow-up

### Tests

### Changesets


UltraCakeBakery:
I am relying quite a lot on fallthrough routes in their current state. I don't want to get locked out of bug fixes while waiting for part two..

Is this a [WIP] kind of Pull Request, or will this actually get merged without any alternative to fallthrough routes?

Rich-Harris:
It won't get merged until part two is ready — working on that now

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -325,8 +323,15 @@ export async function load_node({
 		loaded = await module.load.call(null, load_input);
 
 		if (!loaded) {
+			// TODO do we still want to enforce this now that there's no fallthrough?
 			throw new Error(`load function must return a value${options.dev ? ` (${node.entry})` : ''}`);
 		}
+
+		// TODO remove for 1.9
```
benmccann, line 330: this typo exists a few places below as well
```suggestion
		// TODO remove for 1.0
```

documentation/docs/01-routing.md
```diff
@@ -323,11 +323,10 @@ A route can have multiple dynamic parameters, for example `src/routes/[category]
 It's possible for multiple routes to match a given path. For example each of these routes would match `/foo-abc`:
 
 ```bash
+src/routes/[...catchall].svelte
```
benmccann, line 326: should you be allowed to have `[...catchall].svelte` and `[b].svelte`? It seems like `[b].svelte` will never be called, so maybe this should throw an error during build

Conduitry: They would both be allowed once we implement part 2 of this with route validators, right? (And that is going to happen immediately after this PR.)

Rich-Harris: `[...catchall]` is a rest parameter. regardless of part 2, it will match `/things/b/will/not/match`. We _could_ throw an error for `[a].js` alongside `[b].svelte`, but I think we can do that separately to avoid bloating this PR

packages/kit/src/runtime/server/index.js
```diff
@@ -257,6 +243,8 @@ export async function respond(request, options, state) {
 
 						return response;
 					}
+
+					break;
```
dominikg, line 247: this break at the very end confuses me a bit, is there a way to restructure the loop to avoid it? (or explain it to me like i'm 5)

documentation/docs/01-routing.md
benmccann: I assumed this was the section where we showed the priority order of routes (GitHub collapsed the context around this). It looks like `[..catchall].svelte` is still last in the priority ordering. I wonder if we should show it last here as well for consistency?

PR #4332, opened 2022-03-15
[breaking] rename `xForwardedForIndex` to `XFF_DEPTH`

benmccann:
Having to specify a negative number is a bit of an unusual API, so flip it and make it positive

This is a breaking change for anyone already using `xForwardedForIndex`

### DIFF
--- a/packages/adapter-node/README.md
+++ b/packages/adapter-node/README.md
@@ -94,23 +94,17 @@ MY_ORIGINURL=https://my.site \
 node build
 ```
 
-### xForwardedForIndex
+### xForwardedForNumProxies
 
-If the `ADDRESS_HEADER` is `X-Forwarded-For`, the header value will contain a comma-separated list of IP addresses. For example, if there are three proxies between your server and the client, proxy 3 will forward the addresses of the client and the first two proxies:
+If the `ADDRESS_HEADER` is `X-Forwarded-For`, the header value will contain a comma-separated list of IP addresses. `xForwardedForNumProxies` should specify how many trusted proxies sit between your server and the client. E.g. if there are three trusted proxies, proxy 3 will forward the addresses of the client and the first two proxies:
 
 ```
 <client address>, <proxy 1 address>, <proxy 2 address>
 ```
 
-To get the client address we could use `xForwardedFor: 0` or `xForwardedFor: -3`, which counts back from the number of addresses.
+To get the client address we could use `xForwardedFor: 3`.
 
-**X-Forwarded-For is [trivial to spoof](https://adam-p.ca/blog/2022/03/x-forwarded-for/), howevever**:
-
-```
-<spoofed address>, <client address>, <proxy 1 address>, <proxy 2 address>
-```
-
-For that reason you should always use a negative number (depending on the number of proxies) if you need to trust `event.clientAddress`. In the above example, `0` would yield the spoofed address while `-3` would continue to work.
+If there may be a variable number of proxies, you would have to read the `X-Forwarded-For` header yourself and read the IP address from the left, but be very careful that you do not use the result for any applications with possible security implications.
 
 ## Custom server
 
--- a/packages/adapter-node/README.md
+++ b/packages/adapter-node/README.md
@@ -104,7 +104,7 @@ If the `ADDRESS_HEADER` is `X-Forwarded-For`, the header value will contain a co
 
 To get the client address we could use `XFF_DEPTH=3`.
 
-This is [the most secure way](https://adam-p.ca/blog/2022/03/x-forwarded-for/) to read from the `X-Forwarded-For` header. It assumes a constant number of proxies between the client and server. If you are not using this value for a security-sensitive application, you may find it more reliable in some instances to avoid this assumption and read the first entry from the `X-Forwarded-For` header yourself (e.g. some users may be connecting through a corporate proxy while others are not breaking the assumption of a constant number of proxies).
+> If you need to read the left-most address instead (and don't care about spoofing) — for example, to offer a geolocation service, where it's more important for the IP address to be _real_ than _trusted_, you can do so by inspecting the `x-forwarded-for` header within your app.
 
 ## Custom server
 
--- a/packages/adapter-node/README.md
+++ b/packages/adapter-node/README.md
@@ -94,17 +94,17 @@ MY_ORIGINURL=https://my.site \
 node build
 ```
 
-### xForwardedForNumProxies
+### `XFF_DEPTH`
 
-If the `ADDRESS_HEADER` is `X-Forwarded-For`, the header value will contain a comma-separated list of IP addresses. `xForwardedForNumProxies` should specify how many trusted proxies sit between your server and the client. E.g. if there are three trusted proxies, proxy 3 will forward the addresses of the client and the first two proxies:
+If the `ADDRESS_HEADER` is `X-Forwarded-For`, the header value will contain a comma-separated list of IP addresses. The `XFF_DEPTH` environment variable should specify how many trusted proxies sit between your server and the client. E.g. if there are three trusted proxies, proxy 3 will forward the addresses of the client and the first two proxies:
 
 ```
 <client address>, <proxy 1 address>, <proxy 2 address>
 ```
 
-To get the client address we could use `xForwardedFor: 3`.
+To get the client address we could use `XFF_DEPTH=3`.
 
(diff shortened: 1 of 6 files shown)


packages/adapter-node/README.md
```diff
@@ -94,23 +94,17 @@ MY_ORIGINURL=https://my.site \
...
 
-If the `ADDRESS_HEADER` is `X-Forwarded-For`, the header value will contain a comma-separated list of IP addresses. For example, if there are three proxies between your server and the client, proxy 3 will forward the addresses of the client and the first two proxies:
+If the `ADDRESS_HEADER` is `X-Forwarded-For`, the header value will contain a comma-separated list of IP addresses. `xForwardedForNumProxies` should specify how many trusted proxies sit between your server and the client. E.g. if there are three trusted proxies, proxy 3 will forward the addresses of the client and the first two proxies:
 
 ```
 <client address>, <proxy 1 address>, <proxy 2 address>
 ```
 
-To get the client address we could use `xForwardedFor: 0` or `xForwardedFor: -3`, which counts back from the number of addresses.
+To get the client address we could use `xForwardedFor: 3`.
```
Rich-Harris, line 105:
```suggestion
To get the client address we could use `xForwardedForDepth: 3`.
```

```suggestion
To get the client address we could use `xForwardedForNumProxies: 3`.
```

packages/adapter-node/README.md
```diff
@@ -94,23 +94,17 @@ MY_ORIGINURL=https://my.site \
...
 
-**X-Forwarded-For is [trivial to spoof](https://adam-p.ca/blog/2022/03/x-forwarded-for/), howevever**:
-
-```
-<spoofed address>, <client address>, <proxy 1 address>, <proxy 2 address>
-```
-
-For that reason you should always use a negative number (depending on the number of proxies) if you need to trust `event.clientAddress`. In the above example, `0` would yield the spoofed address while `-3` would continue to work.
+If there may be a variable number of proxies, you would have to read the `X-Forwarded-For` header yourself and read the IP address from the left, but be very careful that you do not use the result for any applications with possible security implications since `X-Forwarded-For` is [trivial to spoof](https://adam-p.ca/blog/2022/03/x-forwarded-for/).
```
Rich-Harris, line 107: Is this a realistic scenario? Surely there's going to be a fixed number of proxies unless you changed your network architecture?

I think it might be worth preserving the previous illustration, because most people aren't going to know the implications of 'read from the left' or 'read from the right' (at least, _I_ wouldn't have known before this week)

packages/adapter-node/README.md
```diff
@@ -94,23 +94,17 @@ MY_ORIGINURL=https://my.site \
 node build
 ```
 
-### xForwardedForIndex
+### xForwardedForNumProxies
```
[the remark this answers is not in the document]
benmccann: I also wondered about something like `numProxies` or `numTrustedProxies`. What do you think about dropping the `xForwardedFor` prefix?

mrkishi:
I just realized this should also be an environment variable, since it's not app- but deployment-specific. So `ADDRESS_HEADER` and...

`X_FORWARDED_FOR_DEPTH`?
`X_FORWARDED_FOR_TRUST`?
`X_FORWARDED_FOR_TRUSTED_PROXIES`?

packages/adapter-node/README.md
benmccann: I was thinking that there could be proxies that are not controlled by your own server infrastructure. E.g. some users may have the traffic flow through a corporate proxy at the user's workplace while others wouldn't, so there could always be variable number of proxies. This could probably be worded more clearly

benmccann:
yeah, making it an environment variable probably makes sense

packages/adapter-node/README.md
Rich-Harris: Surely the relevant number isn't the proxies between the server and the client, but the (trusted) proxies between the server and the public internet?

benmccann: Why would that be? Any proxy could be added to the `X-Forwarded-For`

Sometimes you could have:
```
<client address>, <trusted proxy address>
```

And sometimes you could have:
```
<client address>, <trusted proxy address>, <untrusted proxy address>
```

In this case it doesn't work to go a set number from the right. And actually now I'm remembering @mrkishi suggesting that we would need to not just use an index, but check the addresses in the list and I'm wondering if we should just remove the ability to lookup by index as it actually does not seem very safe to me.

mrkishi: Using an index is fine, you just need to be careful do it from the right (as we're already doing). If you have an untrusted proxy _after_ a trusted one, you can't trust the header at all, so in that case you shouldn't set the header in order to get `remoteAddress` instead.

A list of trusted proxy "addresses" is a valid option, but since it was considered too troublesome, we can just let it up to the user since they already have access to the header. Even with a list of trusted proxies, you should _still_ read from the right: keep discarding trusted addresses from the right and use the first untrusted one as the client address.

packages/adapter-node/README.md
```diff
@@ -94,23 +94,17 @@ MY_ORIGINURL=https://my.site \
 node build
 ```
 
-### xForwardedForIndex
+### `XFF_DEPTH`
 
-If the `ADDRESS_HEADER` is `X-Forwarded-For`, the header value will contain a comma-separated list of IP addresses. For example, if there are three proxies between your server and the client, proxy 3 will forward the addresses of the client and the first two proxies:
+If the `ADDRESS_HEADER` is `X-Forwarded-For`, the header value will contain a comma-separated list of IP addresses. The `XFF_DEPTH` environment variable should specify how many trusted proxies sit between your server and the client. E.g. if there are three trusted proxies, proxy 3 will forward the addresses of the client and the first two proxies:
```
Rich-Harris, line 99: What if we said this, so that we don't need to worry about what proxies sit in front of the client? Or is that too confusing given the possibility of apps running on intranets etc?

```suggestion
If the `ADDRESS_HEADER` is `X-Forwarded-For`, the header value will contain a comma-separated list of IP addresses. The `XFF_DEPTH` environment variable should specify how many trusted proxies sit between your server and the public internet. E.g. if there are three trusted proxies, proxy 3 will forward the addresses of the client and the first two proxies:
```

packages/adapter-node/README.md
```diff
@@ -94,23 +94,17 @@ MY_ORIGINURL=https://my.site \
...
 
 ```
 <client address>, <proxy 1 address>, <proxy 2 address>
 ```
 
-To get the client address we could use `xForwardedFor: 0` or `xForwardedFor: -3`, which counts back from the number of addresses.
+To get the client address we could use `XFF_DEPTH=3`.
```
Rich-Harris, line 105: Some guides will tell you to read the left-most address, but this leaves you vulnerable to spoofing:

```
<spoofed address>, <client address>, <proxy 1 address>, <proxy 2 address>
```

Instead, we read from the _right_, accounting for the number of trusted proxies. In this case, we would use `XFF_DEPTH=3`.

packages/adapter-node/src/handler.js
```diff
@@ -7,17 +7,32 @@ import { getRequest, setResponse } from '@sveltejs/kit/node';
...
 
+const xff_depth = get_xff_depth();
 const address_header = ADDRESS_HEADER && (process.env[ADDRESS_HEADER] || '').toLowerCase();
 const protocol_header = PROTOCOL_HEADER && process.env[PROTOCOL_HEADER];
 const host_header = (HOST_HEADER && process.env[HOST_HEADER]) || 'host';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
+function get_xff_depth() {
+	const value = process.env['XFF_DEPTH'];
```
Rich-Harris, line 23: should it be possible to change the name of this header, like all the others? if not, what makes it different?

packages/adapter-node/src/handler.js
```diff
@@ -7,17 +7,32 @@ import { getRequest, setResponse } from '@sveltejs/kit/node';
...
+	let xff_depth;
+	try {
+		xff_depth = value ? parseInt(value) : 1;
```
Rich-Harris, line 26: this shouldn't be in a try-catch. `parseInt` will never throw, it'll just return `NaN`

Rich-Harris: opened https://github.com/sveltejs/kit/pull/4357 as it was easier than piecemeal suggestions for a replacement for `get_xff_depth`

packages/adapter-node/README.md
Rich-Harris: yeah, much better. i like the first one

benmccann: I'm not sure I understand the difference between these and what I said, but @mrkishi's wording is fine with me if you'd like to change it to that

Rich-Harris (approved):

PR #4333, opened 2022-03-15
fix: change `remove` to `unlink` in vite.watcher

ydcjeff:
Chokidar does not have `remove` event, only `unlink` for watching file removal.
https://github.com/paulmillr/chokidar#api

### Tests

### Changesets

### DIFF
--- a/.changeset/dull-emus-tap.md
+++ b/.changeset/dull-emus-tap.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/kit": patch
+---
+
+fix: correctly detect removal of route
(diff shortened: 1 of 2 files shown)


Conduitry:
Should this actually be a "fix" and not a "chore"? Would SvelteKit currently fail to update the manifest when a file is removed?

ydcjeff:
@Conduitry Yes, I think this should be a fix. Currently, SvelteKit does not update the manifest when route file is removed.

ydcjeff:
Should this PR need to run `changeset`?

Conduitry:
Yes, a changeset is necessary so there's an entry in the changelog and so that an automatic PR is opened to release `@sveltejs/kit` with those changes.

benmccann (approved):

PR #4334, opened 2022-03-15
Param validators

Rich-Harris:
WIP. Adds parameter validators (#4291). Part 2 of 2, follows #4330. 

* [ ] Server-side param validation
* [ ] Documentation
* [ ] General tidy up

### Tests

### Changesets


packages/kit/src/core/dev/plugin.js
```diff
@@ -123,14 +127,27 @@ export async function create_plugin(config, cwd) {
...
-						})
+						}),
+						validators: async () => {
+							/** @type {Record<string, import('types').ParamValidator>} */
+							const validators = {};
+
+							for (const key in manifest_data.validators) {
+								const url = path.resolve(cwd, manifest_data.validators[key]);
+								const module = await vite.ssrLoadModule(url);
+								validators[key] = module.validate;
```
dominikg, line 146: throw an error if validate doesn't exist on the loaded module?

Rich-Harris: good catch — fixed

packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -286,12 +289,32 @@ export default function create_manifest_data({
...
+	const params_base = path.relative(cwd, config.kit.files.params);
+
+	/** @type {Record<string, string>} */
+	const validators = {};
+	if (fs.existsSync(config.kit.files.params)) {
+		for (const file of fs.readdirSync(config.kit.files.params)) {
+			const ext = path.extname(file);
+			const type = file.slice(0, -ext.length);
+
+			if (/^[a-zA-Z_$][a-zA-Z0-9_$]*$/.test(type)) {
```
dominikg, line 301: if validator names can appear in filenames i'd forbid `$`

documentation/docs/01-routing.md
```diff
@@ -318,6 +318,27 @@ A route can have multiple dynamic parameters, for example `src/routes/[category]
...
 
+#### Validation
+
+A route like `src/routes/archive/[page]` would match `/archive/3`, but it would also match `/archive/potato`. We don't want that. You can ensure that route parameters are well-formed by adding _validators_ to your [`params`](/docs/configuration#files) directory...
+
+```js
+/// file: src/params/integer.js
+/** @type {import('@sveltejs/kit').ParamValidator} */
+export function validate(param) {
+	return /^\d+$/.test(param);
```
dummdidumm, line 329: Reading this through the lens of someone who didn't follow the discussion, the given documentation might be a little too short. This additional comment might help.

```suggestion
	// A validator is a function that returns a `boolean` indicating if a given route parameter is valid.
	return /^\d+$/.test(param);
```

packages/kit/src/runtime/client/client.js
```diff
@@ -11,14 +11,22 @@ import {
...
+import { components, dictionary, validators } from '__GENERATED__/client-manifest.js';
 
 const SCROLL_KEY = 'sveltekit:scroll';
 const INDEX_KEY = 'sveltekit:index';
 
+const routes = parse(components, dictionary, validators);
+
+// we import the root layout/error components eagerly, so that
+// connectivity errors after initialisation don't nuke the app
+const default_layout = components[0]();
```
dominikg, line 27: is this part of the validators effort or something different? 
what happens when one of them errors?

Rich-Harris: it's just moving this existing logic https://github.com/sveltejs/kit/blob/fac284475decb1531ee9c2413f58020256f719d2/packages/kit/src/core/sync/write_manifest.js#L69-L71

dummdidumm (commented):
I still don't feel like having a dedicated folder who contains app-wide parameter validation is the right API choice, due to colocation and having this munged in one place - but this doesn't mean I'm completely against this.
The documentation feels a little on the short side, too, maybe a few more sentences would help.

packages/kit/src/core/generate_manifest/index.js
```diff
@@ -41,10 +42,13 @@ export function generate_manifest({ build_data, relative_path, routes, format =
 	});
 
 	/** @type {(path: string) => string} */
-	const importer =
+	const load =
```
benmccann, line 45: I personally liked `importer` a bit more to avoid conflating with the page `load`

documentation/docs/01-routing.md
```diff
@@ -318,6 +318,27 @@ A route can have multiple dynamic parameters, for example `src/routes/[category]
 
 > `src/routes/a/[...rest]/z.svelte` will match `/a/z` as well as `/a/b/z` and `/a/b/c/z` and so on. Make sure you check that the value of the rest parameter is valid.
 
+#### Validation
+
+A route like `src/routes/archive/[page]` would match `/archive/3`, but it would also match `/archive/potato`. We don't want that. You can ensure that route parameters are well-formed by adding _validators_ to your [`params`](/docs/configuration#files) directory...
```
Rich-Harris, line 323: Perhaps this, rather than adding a comment to the code sample (which feels a bit duplicative with the type annotation):

```suggestion
A route like `src/routes/archive/[page]` would match `/archive/3`, but it would also match `/archive/potato`. We don't want that. You can ensure that route parameters are well-formed by adding a _validator_ — which takes the parameter string (`"3"` or `"potato"`) and returns `true` if it is valid — to your [`params`](/docs/configuration#files) directory...
```

packages/kit/src/runtime/server/index.js
```diff
@@ -114,6 +115,8 @@ export async function respond(request, options, state) {
 		transformPage: default_transform
 	};
 
+	// TODO match route before calling handle?
```
benmccann, line 118: probably don't need to add this TODO so that we don't need to remember to remove it in https://github.com/sveltejs/kit/pull/4344

packages/kit/src/core/sync/create_manifest_data/index.js
dominikg: working with them is just harder and they should be urlencoded too (which again could cause trouble with vite path to module resolution). sticking to characters that don't have special meaning in the contexts they are used avoids these issues.

documentation/docs/01-routing.md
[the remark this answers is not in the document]
Rich-Harris: I updated `configuration.md`

dummdidumm: I thought about something like this, too, at first, but didn't find a concise way to write it out - I like the way you phrased it 👍

packages/kit/src/runtime/client/ambient.d.ts
```diff
@@ -1,6 +1,7 @@
-declare module '__GENERATED__/manifest.js' {
-	import { CSRComponent, CSRRoute } from 'types';
+declare module '__GENERATED__/client-manifest.js' {
+	import { CSRComponentLoader } from 'types';
 
-	export const fallback: [CSRComponent, CSRComponent];
-	export const routes: CSRRoute[];
+	export const components: CSRComponentLoader[];
+	export const dictionary: Record<string, [number[], number[], 1]>;
```
benmccann, line 5: it could be worth a JSDoc explaining what dictionary is. it wasn't really clear to me from the name alone. I see that it's value is set to `routes`. I wonder if just calling it `routes` might be a more informative name?

benmccann (commented):
lgtm

packages/kit/src/utils/routing.js
```diff
@@ -0,0 +1,65 @@
+/** @param {string} key */
+export function parse_route_key(key) {
```
benmccann, line 2: do we want to call this `parse_route_id` to align with https://github.com/sveltejs/kit/pull/4345?

packages/kit/src/core/sync/create_manifest_data/index.js
Rich-Harris: fair enough, have tightened it up. figured we really need to do it for parameter names as well on the same basis, which is technically a breaking change — added a changeset

packages/kit/src/runtime/server/index.js
Rich-Harris: #4344 already removes it, it'll probably be an easier merge if we leave it in place

packages/kit/src/utils/routing.js
Rich-Harris: we're going to merge #4345 into this PR, so no

packages/kit/src/runtime/client/ambient.d.ts
Rich-Harris: added some docs

PR #4345, opened 2022-03-16
expose `event.routeId` and `page.routeId`

### LINKED ISSUE #3840: Implement post-response hook for logging and monitoring, including matched route and params

Rich-Harris:
closes #3840 (and #3907) by exposing the internal `route.key` as `event.routeKey`. This is a string based on the filename of the page/endpoint:

| filename                                            | `event.routeKey`                             |
|-----------------------------------------------------|---------------------------------|
| src/routes/index.svelte                             | `""`                            |
| src/routes/about.svelte                             | `"about"`                       |
| src/routes/blog/index.js                            | `"blog"`                        |
| src/routes/blog/index.svelte                        | `"blog"`                        |
| src/routes/blog/[slug].svelte                       | `"blog/[slug]"`                 |
| src/routes/blog/archive/[page=integer]/index.svelte | `"blog/archive/[page=integer]"` |
| src/routes/[...catchall]                            | `"[...catchall]"`               |

### Tests

### Changesets

### DIFF
--- a/.changeset/mean-crews-unite.md
+++ b/.changeset/mean-crews-unite.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Expose event.routeKey
+Expose `event.routeId` and `page.routeId`
--- a/.changeset/mean-crews-unite.md
+++ b/.changeset/mean-crews-unite.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Expose event.routeKey
--- a/packages/adapter-node/src/handler.js
+++ b/packages/adapter-node/src/handler.js
@@ -44,7 +44,8 @@ const ssr = async (req, res) => {
 		request = await getRequest(origin || get_origin(req.headers), req);
 	} catch (err) {
 		res.statusCode = err.status || 400;
-		return res.end(err.reason || 'Invalid request body');
+		res.end(err.reason || 'Invalid request body');
+		return;
 	}
 
 	if (address_header && !(address_header in req.headers)) {
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -127,6 +127,7 @@ export async function create_plugin(config, cwd) {
 
 							return {
 								type: 'endpoint',
+								key: route.key,
 								pattern,
 								names,
 								types,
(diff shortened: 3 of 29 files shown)


Conduitry:
Do we want to expose this in `$page` as well? Or are you thinking that, if someone really wanted this on the client, they could expose it in `$session` manually?

Rich-Harris:
We definitely could put it in page (one of the changes in this series of PRs makes all page keys available to the client as a side effect), I was just being as conservative as possible. Do you have a use for it?

Rich-Harris:
In that case is `page.routeKey` the right property, or should it just be `event.key` and `page.key`?

.changeset/mean-crews-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Expose event.routeKey
```
Rich-Harris, line 5:
```suggestion
Expose `event.routeId` and `page.routeId`
```

Rich-Harris:
merging this into #4344 to try and minimise #4354-style git fuckups

PR #4351, opened 2022-03-16
render fallback page regardless of trailingSlash

### LINKED ISSUE #4340: The generated `fallback` HTML file is empty when `trailingSlash` is `"always"`
### Describe the bug

When using the static adapter and specifying a `fallback` page to be generated, it (`build/200.html` in the example) is 0 bytes when `kit.trailingSlash` is `"always"`. Specifying a different value of `trailingSlash` or omitting it to use the default makes the file have its expected HTML again. Regular pages in the app (`/` (the index) in the example) have built HTML as expected in either case.

### Reproduction

https://github.com/babichjacob/repro-trailingslash-fallback

This is a minimal skeleton template with the adapter switched from auto to static and other changes in `svelte.config.js` as mentioned.

### Logs

```shell
> # there is no error
> svelte-kit build

vite v2.8.6 building for production...
✓ 12 modules transformed.
.svelte-kit/output/client/_app/manifest.json                    1.15 KiB
.svelte-kit/output/client/_app/layout.svelte-7fa5b089.js        0.53 KiB / gzip: 0.35 KiB
.svelte-kit/output/client/_app/error.svelte-e35f4419.js         1.56 KiB / gzip: 0.75 KiB
.svelte-kit/output/client/_app/pages/index.svelte-296a81ee.js   0.80 KiB / gzip: 0.47 KiB
.svelte-kit/output/client/_app/chunks/vendor-f45b3020.js        7.21 KiB / gzip: 2.96 KiB
.svelte-kit/output/client/_app/start-221a0adb.js                21.20 KiB / gzip: 7.88 KiB
vite v2.8.6 building SSR bundle for production...
✓ 11 modules transformed.
Generated an empty chunk: "hooks"
.svelte-kit/output/server/manifest.json                        1.09 KiB
.svelte-kit/output/server/index.js                             64.87 KiB
.svelte-kit/output/server/entries/fallbacks/layout.svelte.js   0.24 KiB
.svelte-kit/output/server/entries/fallbacks/error.svelte.js    0.72 KiB
.svelte-kit/output/server/entries/pages/index.svelte.js        0.32 KiB
.svelte-kit/output/server/chunks/index-eddbeae8.js             2.29 KiB
.svelte-kit/output/server/chunks/hooks-1c45ba0b.js             0.00 KiB

  Wrote site to "build"
  ✔ done
```

### System Info

```shell
System:
    OS: macOS 11.6
    CPU: (12) x64 AMD Ryzen 5 2600 Six-Core Processor            
    Memory: 1009.41 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.2 - ~/.nvm/versions/node/v16.13.2/bin/node
    npm: 8.4.0 - ~/.nvm/versions/node/v16.13.2/bin/npm
  Browsers:
    Chrome: 99.0.4844.51
    Firefox: 98.0
    Safari: 15.1
  npmPackages:
    @sveltejs/adapter-static: ^1.0.0-next.29 => 1.0.0-next.29 
    @sveltejs/kit: next => 1.0.0-next.298 
    svelte: ^3.46.4 => 3.46.4
```

### Severity

annoyance

### Additional Information

[I found that](https://github.com/sveltejs/kit/blob/7f958dd49cfbbc4712001b1b21ce1474b44da69e/packages/kit/src/core/adapt/builder.js#L143-L146) `build/200.html` comes from the `fallback.html` file in `.svelte-kit/output/prerendered` being copied as-is to the `build` folder under the given name. I haven't done any experimentation, but maybe [`respond`](https://github.com/sveltejs/kit/blob/268ee6b921ee9cab9f5995b6c5144b3d903ed02a/packages/kit/src/runtime/server/index.js#L128-L144) or [`render_response`](https://github.com/sveltejs/kit/blob/268ee6b921ee9cab9f5995b6c5144b3d903ed02a/packages/kit/src/runtime/server/page/render.js#L32) might not have been programmed to accommodate the trailing slash configuration in this specific combination of options?

Rich-Harris:
Trailing slash normalisation was causing this to be a redirect. Fixes #4340

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -17,7 +17,7 @@ export async function respond(request, options, state) {
 
 	const normalized = normalize_path(url.pathname, options.trailing_slash);
 
-	if (normalized !== url.pathname) {
+	if (normalized !== url.pathname && !state.prerender?.fallback) {
 		return new Response(undefined, {
 			status: 301,
 			headers: {
--- a/.changeset/few-carrots-sell.md
+++ b/.changeset/few-carrots-sell.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Populate fallback page when trailingSlash is "always"
--- a/packages/kit/test/prerendering/options/svelte.config.js
+++ b/packages/kit/test/prerendering/options/svelte.config.js
@@ -4,7 +4,9 @@ import adapter from '../../../../adapter-static/index.js';
 /** @type {import('@sveltejs/kit').Config} */
 const config = {
 	kit: {
-		adapter: adapter(),
+		adapter: adapter({
+			fallback: '200.html'
+		}),
 
 		csp: {
 			directives: {
--- a/packages/kit/test/prerendering/options/test/test.js
+++ b/packages/kit/test/prerendering/options/test/test.js
@@ -31,4 +31,10 @@ test('does not copy `public` into `_app`', () => {
 	assert.ok(!fs.existsSync(`${build}/_app/robots.txt`));
 });
 
+// https://github.com/sveltejs/kit/issues/4340
+test('populates fallback 200.html file', () => {
+	const content = read('200.html');
+	assert.ok(content !== '');
+});
+
 test.run();


packages/kit/src/runtime/server/index.js
```diff
@@ -17,7 +17,7 @@ export async function respond(request, options, state) {
 
 	const normalized = normalize_path(url.pathname, options.trailing_slash);
 
-	if (normalized !== url.pathname) {
+	if (normalized !== url.pathname && !state.prerender?.fallback) {
```
Conduitry, line 20: I forget where we're at with using optional chaining on the server. I recall there being issues in some hosting providers because of the JS features they supported. Are those now addressed and/or would this be getting transpiled (as either part of Kit's build, or during the adapt phase)?

Rich-Harris: i think we're in the clear — we only support 14 and above: https://node.green/#ES2020-features-optional-chaining-operator------optional-property-access

i think we're using optional chaining in a ton of places already

PR #4357, opened 2022-03-16
XFF_DEPTH_ENV

Rich-Harris:
suggested changes to #4332 (cc @benmccann)

### DIFF
--- a/packages/adapter-node/index.js
+++ b/packages/adapter-node/index.js
@@ -18,22 +18,18 @@ export default function ({
 		host: host_env = 'HOST',
 		port: port_env = 'PORT',
 		origin: origin_env = 'ORIGIN',
+		xffDepth: xff_depth_env = 'XFF_DEPTH',
 		headers: {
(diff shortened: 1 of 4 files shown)


packages/adapter-node/index.js
```diff
@@ -18,6 +18,7 @@ export default function ({
 		host: host_env = 'HOST',
 		port: port_env = 'PORT',
 		origin: origin_env = 'ORIGIN',
+		xffDepth: xff_depth_env = 'XFF_DEPTH',
```
benmccann, line 21: No objections to this because I know it's consistent with the others, but it really seems unnecessary

benmccann (approved):

packages/adapter-node/index.js
Rich-Harris: i guess the question becomes 'why is it necessary for the others but not this?' and if it _isn't_ necessary for the others, then why do we need to make the env vars configurable at all?

PR #4360, opened 2022-03-16
make env vars a bit easier to understand/follow in adapter-node

Rich-Harris:
I find the environment variable options in `adapter-node` confusing in the extreme — they're options that set variables that allow you to set variables (that tell the app how to read variables, in the case of headers). The least we can do is avoid giving them inconsistent aliases on top of that.

### Tests

### Changesets


packages/adapter-node/index.js
```diff
@@ -10,22 +10,30 @@ const pipe = promisify(pipeline);
...
+	if (opts.env) {
+		throw new Error(
+			`options.env has been removed in favour of options.environment, which has different properties. Consult the adapter-node README: https://github.com/sveltejs/kit/tree/master/packages/adapter-node`
+		);
+	}
+
+	const {
+		out = 'build',
+		precompress,
+		environment: {
```
benmccann, line 25: would something like `envAliases` make sense as a name?

Rich-Harris: i feel like to be properly accurate it would need to be `envVarAliases` which seems like a step too far

packages/adapter-node/tsconfig.json
```diff
@@ -12,5 +12,5 @@
 			"@sveltejs/kit": ["../kit/types/index"]
 		}
 	},
-	"include": ["index.js", "src/**/*.js", "ambient.d.ts"]
+	"include": ["index.js", "src/**/*.js", "ambient.d.ts", "src/types.d.ts"]
```
mrkishi, line 15:
```suggestion
	"include": ["index.js", "src/**/*.js", "ambient.d.ts"]
```

Conduitry (approved):

Issue #4362, opened 2022-03-16
`event.url` is not set when `--https` flag is used


DrMagPie:
### Describe the bug

when using `svelte-kit dev` command with `--https` flag `event.url` is not properly configured. please see following output.
```
URL {
  href: 'https://undefined/',
  origin: 'https://undefined',
  protocol: 'https:',
  username: '',
  password: '',
  host: 'undefined',
  hostname: 'undefined',
  port: '',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

if `--https` flag is not used output looks like this
```
URL {
  href: 'http://localhost:3000/',
  origin: 'http://localhost:3000',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'localhost:3000',
  hostname: 'localhost',
  port: '3000',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

### Reproduction

to see the issue use following `hooks.ts` file
```ts
import type { Handle } from "@sveltejs/kit";
export const handle: Handle = async ({ event, resolve }) => {
  console.log(event.url)
  const response = await resolve(event);
  return response;
};
```
then just start dev server and visit any page
`yarn dev --https`

### Logs

```shell
❯ yarn dev 
yarn run v1.22.17
warning package.json: No license field
$ svelte-kit dev

  SvelteKit v1.0.0-next.299

  local:   http://localhost:3000
  network: not exposed

  Use --host to expose server to other devices on this network


URL {
  href: 'http://localhost:3000/',
  origin: 'http://localhost:3000',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'localhost:3000',
  hostname: 'localhost',
  port: '3000',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}

❯ yarn dev --https
yarn run v1.22.17
warning package.json: No license field
$ svelte-kit dev --https

  SvelteKit v1.0.0-next.299

  local:   https://localhost:3000
  network: not exposed

  Use --host to expose server to other devices on this network


URL {
  href: 'https://undefined/',
  origin: 'https://undefined',
  protocol: 'https:',
  username: '',
  password: '',
  host: 'undefined',
  hostname: 'undefined',
  port: '',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

### System Info

```shell
[...]

PR #4364, opened 2022-03-16
[fix] populate event.url when using http2

benmccann:
fixes https://github.com/sveltejs/kit/issues/4362

### DIFF
--- a/.changeset/five-turtles-rhyme.md
+++ b/.changeset/five-turtles-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] populate event.url when using http2
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -197,7 +197,9 @@ export async function create_plugin(config, cwd) {
 					try {
 						if (!req.url || !req.method) throw new Error('Incomplete request');
 
-						const base = `${vite.config.server.https ? 'https' : 'http'}://${req.headers.host}`;
+						const base = `${vite.config.server.https ? 'https' : 'http'}://${
+							req.headers[':authority'] || req.headers.host
+						}`;
 
 						const decoded = decodeURI(new URL(base + req.url).pathname);
 


packages/kit/src/core/dev/plugin.js
```diff
@@ -197,7 +197,9 @@ export async function create_plugin(config, cwd) {
 					try {
 						if (!req.url || !req.method) throw new Error('Incomplete request');
 
-						const base = `${vite.config.server.https ? 'https' : 'http'}://${req.headers.host}`;
+						const base = `${vite.config.server.https ? 'https' : 'http'}://${
+							req.headers[':authority'] || req.headers.host
```
Conduitry, line 201: This is what I had suggested in #4362, but I don't know whether it's the most correct option. Perhaps something like `req.httpVersionMajor === 2 ? req.headers[':authority'] : req.headers.host` would be slightly better?

Conduitry:
```suggestion
							req.httpVersionMajor === 2 ? req.headers[':authority'] : req.headers.host
```

(to make it easy in case that's what we want to do)

bjon: This will make http3 act as http1

Conduitry: What should we be using in HTTP/3? (For the record, this code is just run during local dev, not in the production apps, and Vite doesn't currently use HTTP/3 for anything.)

benmccann: Can you explain what you mean when you say it will make it act as http1? Is there some way that http3 is different than http2 here?

Conduitry: It would act like HTTP/1 if we are checking for `req.httpVersionMajor === 2` and we instead get `req.httpVersionMajor === 3`. That said, I have no idea whether there's any native HTTP/3 support in Node, or how we could find ourselves with a `req` object here with `req.httpVersionMajor === 3`.

Unless and until Vite starts sending us HTTP/3 connections here (somehow), I don't think that's something we need to worry about.

benmccann: ah, I didn't understand that @bjon's comment was referring to @Conduitry's suggestion rather than my own code. It makes more sense now :smile: 

anyway, I believe the code as it currently is in this PR might be more correct than the suggestion: https://github.com/nodejs/node/commit/f5c0e282ccf98e17f295c11850649ad19a6fff51

Conduitry: :+1: Sounds good to me. I'll mark this conversation resolved.

.changeset/five-turtles-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] populate event.url when using http2
```
Conduitry, line 5:
```suggestion
[fix] correctly populate `event.url.host` in dev mode when using `--https`
```

Conduitry (approved):

Issue #4341, opened 2022-03-15
TypeScript installation works with `tests/test.ts`, not `tests/test.js`


jtlapp:
### Describe the bug

I answered 'yes' to installing with TypeScript, but `yarn test` and `npm test` would bomb with this error:

`Error [ERR_REQUIRE_ESM] [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/joe/repos/cavesite/tests/test.js`

I solved the problem by renaming `tests/test.js` to `tests/test.ts`.

The main problem with this is that it sent me for a spin around the Internet trying to figure out how to get things working.

(Aside: I'm still trying to figure out how to run `src/**/*.test.ts` unit tests, via either playwright or jest.)

### Reproduction

1. Install SvelteKit with TypeScript.
2. Run `yarn test` or `npm test`.

### Logs

```shell
Error [ERR_REQUIRE_ESM] [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/joe/repos/cavesite/tests/test.js
require() of ES modules is not supported.
require() of /Users/joe/repos/cavesite/tests/test.js from /Users/joe/repos/cavesite/node_modules/@playwright/test/lib/loader.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename test.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/joe/repos/cavesite/package.json.


    at new NodeError (internal/errors.js:322:7)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1102:13)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:93:18)
    at Loader._requireOrImport (/Users/joe/repos/cavesite/node_modules/@playwright/test/lib/loader.js:244:14)
    at Loader.loadTestFile (/Users/joe/repos/cavesite/node_modules/@playwright/test/lib/loader.js:142:18)
    at Runner._runFiles (/Users/joe/repos/cavesite/node_modules/@playwright/test/lib/runner.js:295:44)
    at Runner._run (/Users/joe/repos/cavesite/node_modules/@playwright/test/lib/runner.js:231:12)
[...]

Issue #4194, opened 2022-03-03
Use `.ts` extension when creating Playwright config for TypeScript projects


benmccann:
### Describe the bug

I've seen a few reports of this. It appears to have been a known issue when the feature was added: https://github.com/sveltejs/kit/pull/4056#discussion_r813166060

I saw that someone fixed it by prefixing with some environment variable. Unforunately, I can't find what that was at the moment

### Reproduction

Create TypeScript + Playwright project from wizard

### System Info

```shell
latest
```

### Severity

serious, but I can work around it

vhscom:
```json
"test": "PW_EXPERIMENTAL_TS_ESM=1 playwright test"
```

example implementation: https://codeberg.org/vhs/svelte-headlessui-starter

vhscom:
I believe you're looking for:

```json
"test": "PW_EXPERIMENTAL_TS_ESM=1 playwright test"
```

example implementation: https://github.com/vhscom/svelte-headlessui-starter

vhscom:
Last update I saw for ESM in TS for Playwrite dated Jan 11: https://github.com/microsoft/playwright/issues/7121#issuecomment-1009140849. I'd expect the ENV var to simply become optional later, but since this flag was recently introduced I haven't dug in deeper yet.

frederikhors:
I opened this some time ago and I'm waiting for the next version. After this TS code is usable. https://github.com/microsoft/playwright/issues/12307

EdricChan03:
> I believe you're looking for:
> 
> ```json
> "test": "PW_EXPERIMENTAL_TS_ESM=1 playwright test"
> ```
> 
> example implementation: https://github.com/vhscom/svelte-headlessui-starter

Testing this on the latest versions appears to result in the following issue:

```
Cannot find package 'url.js' imported from /path/to/node_modules/@sveltejs/kit/dist/chunks/index2.js
```
([Full build logs for context](https://github.com/EdricChan03/playwright-ts-repro/runs/5529113293?check_suite_focus=true#step:5:20))

(And a [reproduction of the issue](https://github.com/EdricChan03/playwright-ts-repro/tree/repro/cant-find-package-urljs))

vhscom:
@EdricChan03 the `url.js` appears to have started with https://github.com/sveltejs/kit/blob/master/packages/kit/CHANGELOG.md#100-next292

vhscom:
[...]

PR #4365, opened 2022-03-16
[fix] use .ts extension for tests when using TypeScript

benmccann:
fixes https://github.com/sveltejs/kit/issues/4194
fixes https://github.com/sveltejs/kit/issues/4341


Rich-Harris:
Version that I think is slightly more future-proof: #4368

PR #4368, opened 2022-03-17
Add `.ts` files in `create-svelte/shared` to both `+typescript` and `-typescript` (as `.js` files)

Rich-Harris:
Alternative to #4365. With this PR, instead of needing to maintain parallel `test/test.ts` and `test/test.js` files (or whatever other stuff gets added in future), we can just have a `.ts` file in a shared directory that doesn't specify typescript, and it will get added to both `+typescript` and `-typescript` variants, appropriately converted in the latter case.

fixes https://github.com/sveltejs/kit/issues/4341. closes #4365

### Tests

### Changesets


benmccann (approved):
I don't know if there's any sort of test we should have for this?

Rich-Harris:
re tests — maybe, though I kinda feel like that applies to the package as a whole. one day

Issue #4294, opened 2022-03-11
Unexpected prerendering trailing slashes routes with "." behavior.


Unthrottled:
### Describe the bug

I am pre-rendering a site that has a multiple routes defined with different parameters for say: `routes/message/[message].svelte`.

When I specify routes to pre-render, with trailing slashes, such as `/message/its.ya.boi`, when I run the build, the resulting rendered file built is `message/its.ya.boiindex.html`. However, I was expecting `message/its.ya.boi/index.html` to be output.

### Reproduction

[Demo repository](https://github.com/Unthrottled/svelte-kit-pre-render-problem)

Here is the configuration I have for the basic set up.

svelte.config.js

```javascript
import adapter from '@sveltejs/adapter-static';
import preprocess from 'svelte-preprocess';

/** @type {import('@sveltejs/kit').Config} */
const config = {
	// Consult https://github.com/sveltejs/svelte-preprocess
	// for more information about preprocessors
	preprocess: preprocess(),

	kit: {
		adapter: adapter(),
		trailingSlash: 'always',
		prerender :{
			default: true,
			crawl: true,
			enabled: true,
			onError: 'continue',
			entries: ['/', '/message/hey.gurl', '/message/how.you', '/message/doin'],
		},
	}
};

export default config;
```

When I run `npm run build` in the `build/message` directory I get these following results:

![image](https://user-images.githubusercontent.com/15972415/157903731-b66d3a37-033d-44d4-b4ce-f08b7384fe9a.png)

Was expecting to have a `hey.gurl/index.html` and a `how.you/index.html`, instead of the combined name file.

### Logs

```shell
I don't know if you wanted the logs of the build command, but here it is:


C:\Users\birdm.DESKTOP-FO92PV5\workspace\svelte-kit-pre-render-problem>npm run build

> svelte-kit-pre-render-problem@0.0.1 build C:\Users\birdm.DESKTOP-FO92PV5\workspace\svelte-kit-pre-render-problem
> svelte-kit build

vite v2.8.6 building for production...
✓ 16 modules transformed.
.svelte-kit/output/client/_app/manifest.json                                1.44 KiB
.svelte-kit/output/client/_app/layout.svelte-77538e00.js                    0.53 KiB / gzip: 0.35 KiB
.svelte-kit/output/client/_app/error.svelte-3b6a161c.js                     1.56 KiB / gzip: 0.75 KiB
[...]

Issue #4275, opened 2022-03-08
Named layouts — solves pathless routes and granular layout resets


Rich-Harris:
### Describe the problem

There's a whole slew of issues surrounding layout resets, such as these:

* https://github.com/sveltejs/kit/issues/4103
* https://github.com/sveltejs/kit/issues/4133
* https://github.com/sveltejs/kit/issues/2647
* https://github.com/sveltejs/kit/issues/1685

There might well be others. The essence of them is that `__layout.reset.svelte` is too blunt an instrument for the problem it purports to solve — your `/` page might be unique (but you can't have a root-level reset), or your nested route might want to reset an intermediate layout but _not_ the root layout, or you might just want to reset a layout for a specific component without the overhead of creating a neighbouring reset component.

A [separate issue](#3832) asks for 'pathless' routes, so that your marketing pages and your app itself can both live at the root level, even though they have totally different layouts. But it's actually the same problem, in disguise.

### Describe the proposed solution

Credit for this belongs mostly to @mrkishi, not me. The idea is to have _named layouts_, and give components a way to declare which layouts they belong to.

We name a layout like so:

```
routes/
├ __layout.{js,svelte}
└ __layout-special.{js,svelte}
```

Any page can now mount inside the `special` layout rather than the default layout by specifying so in its own filename:

```
routes/
├ __layout.{js,svelte}
├ __layout-special.{js,svelte}
├ normal-page.{js,svelte}
├ abnormal-page#special.{js,svelte}
```

This applies at arbitrary depths: in this case, `/deeply/nested/path` will ignore the `__layout`, `deeply/__layout` and `deeply/nested/__layout` files.

```
routes/
├ __layout.{js,svelte}
├ __layout-special.{js,svelte}
├ deeply
│ ├ nested
│ │ ├ path#special.{js,svelte}
│ │ └ __layout.{js,svelte}
│ └ __layout.{js,svelte}
```

If a folder specifies a layout name, its contents will inherit it — `/deeply/nested/path/contains/more/stuff` would use `__layout-special` as well as any `__layout` files contained inside `path#special`:

```
routes/
├ __layout.{js,svelte}
├ __layout-special.{js,svelte}
├ deeply
│ ├ nested
│ │ ├ path#special
│ │ │ ├ contains
│ │ │ │ ├ more
│ │ │ │ │ └ stuff
```

[...]

PR #4370, opened 2022-03-17
Simplify `create_manifest_data`

Rich-Harris:
Started work on #4275, but found `create_manifest_data` quite difficult to work with because of some unnecessary complexity. This puts us on better footing by eliminating that complexity. 

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/utils/routing.js
+++ b/packages/kit/src/utils/routing.js
@@ -82,8 +82,6 @@ export function exec(match, names, types, matchers) {
 	/** @type {Record<string, string>} */
 	const params = {};
 
-	console.log({ match, names, types, matchers });
-
 	for (let i = 0; i < names.length; i += 1) {
 		const name = names[i];
 		const type = types[i];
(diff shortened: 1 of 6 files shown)


packages/kit/src/utils/routing.js
```diff
@@ -1,36 +1,72 @@
...
+
+/** @param {string} id */
+export function parse_route_id(id) {
 	/** @type {string[]} */
 	const names = [];
 
 	/** @type {string[]} */
 	const types = [];
 
+	// `/foo` should get an optional trailing slash, `/foo.json` should not
```
benmccann, line 11: I didn't really look at the changes here. any chance this might fix https://github.com/sveltejs/kit/issues/4294 ?

Rich-Harris: I don't _think_ so. Will need to take a closer look at that issue

Rich-Harris: It does not. Fixing that issue will involve some fundamental changes to how `trailingSlash` is handled

Issue #4383, opened 2022-03-18
`beforeNavigate` fires when the page is reloaded


aradalvand:
### Describe the bug

Previously, I was using the `sveltekit:navigation-start` and `sveltekit:navigation-end` events that were fired on `window` to show/hide a navigation loader on the page.

Now, it seems like the aforementioned events [were removed](https://github.com/sveltejs/kit/blob/master/packages/kit/CHANGELOG.md#100-next285) in favor of the `beforeNavigate`/`afterNavigate` lifecycle functions.

However, I expected the `beforeNavigate` lifecycle event to get fired only when client-side routing is triggered, just like how the good old `sveltekit:navigation-start` used to work, but I noticed that it gets fired even when you refresh the page, or manually modify the URL in the browser address bar and hit enter to leave the page.
https://github.com/sveltejs/kit/blob/e28bc390e033e38585fd1640da65c22f973afd2c/packages/kit/src/runtime/client/router.js#L110-L119
I don't understand the reason behind this, if someone wants to listen for `beforeunload` they could do so by simply registering a listener for that event; correct me if I'm wrong but there's no need for SvelteKit to additionally fire `beforeNavigate`, as that can be completely undesirable in certain cases, such as in the case of a client-side navigation loader, and it's also not consistent with the behavior of the old `sveltekit:navigation-start` event.
My primary suggestion is that the behavior should change to match that of the `sveltekit:navigation-start` event, namely that it should only be fired when client-side navigation occurs. This is the most expected behavior.
But at the very least, I think the callback function should be passed a boolean argument that indicates whether the navigation was triggered by the client-side router or not:
```js
beforeNavigate(({ client }) => {
    if (client) {
      // Blah blah blah...
    }
});
```

It's also worth noting that the `$navigating` store works as expected in this regard, namely that it gets populated only in the case of client-side routing. So, the current behavior of `beforeNavigate` is also inconsistent with the `$navigating` store.

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (4) x64 Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
    Memory: 885.61 MB / 7.89 GB
  Binaries:
    Node: 14.15.0 - C:\Program Files\nodejs\node.EXE     
    Yarn: 1.22.10 - ~\AppData\Roaming\npm\yarn.CMD       
    npm: 6.14.8 - C:\Program Files\nodejs\npm.CMD        
  Browsers:
[...]

PR #4384, opened 2022-03-18
Clarify what happens when `beforeNavigate` is called during `beforeunload`

Rich-Harris:
#4383


dummdidumm (approved):

Rich-Harris:
actually on second thoughts this is wrong. The current API is badly designed:

* `to` should be populated in either case
* it should fire on all navigations away from the page, regardless of whether that's because it's an external URL or because the router is disabled

PR #4388, opened 2022-03-18
Named layouts

### LINKED ISSUE #1685: Allow layout reset for index page
We can currently have routes described by either `<name>.svelte` or `<name>/index.svelte`, whose path when served will be `/name` unless configured otherwise. We can also use layout resets in the `<name>/__layout.reset.svelte` to override the root __layout.svelte. However, we don't seem to be able to apply this reset pattern to the index page, `/` (for example, we can't do `index/__layout.reset.svelte). My use case is that all the non-index pages use a standard banner *except* the home page, where I'd like a different style of banner, so while for any other page the existing functionality works perfectly, we can't apply it to the index.

Be able to reset the layout for the index page. For example, perhaps allow for a pattern like `index/__layout.reset.svelte` and `index/index.svelte`, which would be served from `/` but with the layout overridden.

Use a different name, like `home/index.svelte` with a layout reset, and then in the config make it be served from `/`, but I don't believe this is possible either...

Medium

### LINKED ISSUE #2647: Reset layout level for X levels in nested layouts
### Describe the problem

Having following structure
```
routes
routes/__layout.svelte
routes/admin/__layout.svelte
routes/admin/spec/__layout.reset.svelte
routes/admin/spec/test.svelte
```
In test.svelte i dont want the `__layout` from /routes/admin... so I reset.. but is still want to inherit `routes/__layout.svelte`
i already imported it and it shows up many things of the original 
`__layout.svelte` but some parts are missing because the load function isnt used this way

Adding the option to reset layout only by X levels with new file format
    
`__layout.reset.svelte` -> reset full inheritance of nested layouts
`__layout.reset.1.svelte` -> reset layout of only 1 level
`__layout.reset.2.svelte` -> reset layout of only 2 level

and so on

### Alternatives considered

```
    import { load as loadOrigin } from "../../../__layout.svelte";
    export async function load(arg) {
        let originLoadRet = await loadOrigin(arg);

        return {
            props:  {
                originLoadRet
            }
        }
    }

</script>
<script lang="ts">
    export let originLoadRet : any;
    import LayoutOrigin from "../../../__layout.svelte";
</script>


<LayoutOrigin {...originLoadRet.props}>
    <slot></slot>
</LayoutOrigin>
```

### Importance

would make my life easier

### Additional Information

If the reset level is higher than the totally nested layouts a warning is given and it acts as full reset

### LINKED ISSUE #4103: Allow layout reset without __layout.reset.svelte
### Describe the problem

Let's say we need `/login` to not inherit root layout from `__layout.svelte`. To do this one needs to create `login` folder with `__layout.reset.svelte` and `index.svelte`:
```
login/
  __layout.reset.svelte
  index.svelte
```
This feels like a lot of noise for a one page layout reset. It gets more frustrating when there are many such pages.

I propose to allow `.reset.svelte` to reset the layout without the need of `__layout.reset.svelte`. The `/login` path described above would be implemented simply like this:
```
login.reset.svelte
```

### Alternatives considered

One can put together all pages needing a reset in folder with layout reset:
```
user
  __layout.reset.svelte
  login.svelte
  signup.svelte
```
This means that the paths will be `/user/login` and `/user/signup`, not the desired `/login` and `/signup`

### Importance

would make my life easier

### Additional Information

_No response_

### LINKED ISSUE #4133: Resetting layouts only up to a certain level
### Describe the problem

Currently `__layout.reset.svelte` files undo all the layouts that have been applied from the root level. This behaviour is undesirable when we only want to reset the immediate parent layout and leave the remaining ones. 

Another thing that makes it undesirable, importing global stylesheets in the root layout is a common pattern from what I've seen and undoing the root layout means having to do those imports all over again in the reset file.

My first thought was something in the filename of the resets like `__layout.reset.1.svelte`.

Seeing #4103 and #2694 makes me wonder what if we ditch the __layout.reset.svelte file altogether and instead allow page level components (incl `__layout.svelte` files) to do `export const reset` like with `prerender` and `router` except that `reset` would be number instead of a boolean.

Say we have a file structure like this.
```files
routes/
├─ index.svelte
├─ __layout.svelte
├─ nested/
│  ├─ __layout.svelte
│  ├─ page1.svelte
│  ├─ deeper-nested/
│  │  ├─ page2.svelte
```

#### Example 1
If page2 exports `reset = 1`, then it'd only reset `nested/__layout.svelte`
If it exports `reset = 2`, then it'd reset both `nested/__layout.svelte` and `routes/__layout.svelte`

#### Example 2
If `nested/__layout.svelte` exports `reset = 1`, then it'd reset its parent layout, `routes/__layout.svelte`
And if it didn't export `reset` at all, then it'd just stack like it currently does.

#### Example 3?
If the current behavior of resetting all the way to the top is desired in some cases, then exporting `reset = 0` could be treated as a special value to make that easier (this can be scrapped if it's thought to be ambiguous/confusing)

### Alternatives considered

N/A

### Importance

would make my life easier

### Additional Information

_No response_

### LINKED ISSUE #4275: Named layouts — solves pathless routes and granular layout resets

Rich-Harris:
* Closes #4275 
* Closes #4103 
* Closes #4133
* Closes #2647
* Closes #1685

Differences from #4275: 

* it uses a `@` character instead of `#`, because URL encoding
* no parent layouts on directories, only on pages and layouts (same expressive power)
* layouts can inherit from siblings (including sibling default, if explicit with `__layout-name@default.svelte`)
* no `page@.svelte` or `page@~.svelte`, they are now unnecessary

[Docs here](https://kit-svelte-dev-git-named-layouts-svelte.vercel.app/docs/layouts#named-layouts). I'm not completely happy with them but I think the interactive version will be much clearer. 

### Tests

### Changesets


packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -14,18 +14,28 @@ import { parse_route_id } from '../../../utils/routing.js';
  *   rest: boolean;
  *   type: string | null;
  * }} Part
+ *
  * @typedef {{
```
mrkishi, line 18: Jsdoc comments appear in all types when merged.
```suggestion
 * }} Part
 */
 
 /**
 * @typedef {{
```

packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -40,172 +50,185 @@ export default function create_manifest_data({
...
+		error: posixify(path.relative(cwd, `${fallback}/error.svelte`)),
+		layouts: {
+			default: { file: posixify(path.relative(cwd, `${fallback}/layout.svelte`)), name: 'default' }
+		}
+	});
 
-			if (name.startsWith('__') && !specials.has(name)) {
-				throw new Error(`Files and directories prefixed with __ are reserved (saw ${file})`);
-			}
+	const routes_base = path.relative(cwd, config.kit.files.routes);
```
mrkishi, line 70: Unit tests pass with this on Windows:
```suggestion
	const routes_base = posixify(path.relative(cwd, config.kit.files.routes));
```

documentation/docs/02-layouts.md
```diff
@@ -62,11 +62,115 @@ We can create a layout that only applies to pages below `/settings` (while inher
...
+
+Named layouts can be created at any depth, and will apply to any components in the same subtree. For example, `__layout-foo` will apply to `one` and `two`, but not `three` or `four`:
+
+```
+src/routes/
+├ a/
+│ ├ b/
+│ │ ├ c/
+│ │ │ ├ __layout-foo.svelte
+│ │ │ ├ one@foo.svelte
```
benmccann, line 95: what if someone has a route like `folding@home.svelte` where they don't want the `@` to represent a named layout? should we add a mention here of some way to escape the `@` symbol or use route matchers?

packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -14,18 +14,34 @@ import { parse_route_id } from '../../../utils/routing.js';
  *   rest: boolean;
  *   type: string | null;
  * }} Part
+ */
+
+/**
+ * @typedef {{
```
benmccann, line 20: can we add a comment as to what this is supposed to represent? e.g. "A route along with the page and endpoint corresponding to that route"

documentation/docs/02-layouts.md
Rich-Harris: i've opened https://github.com/sveltejs/kit/issues/4403

packages/kit/src/core/sync/create_manifest_data/index.js
Rich-Harris: added something. tbh these types probably won't last long, i'd like to collapse all this a bit. but that'll probably happen as part of layout endpoints

documentation/docs/02-layouts.md
```diff
@@ -113,64 +113,60 @@ src/routes/
...
 │ │ ├ c/
-│ │ │ ├ __layout@foo.svelte
-│ │ │ └ page.svelte
+│ │ │ └ index@home.svelte
 │ │ └ __layout.svelte
-│ ├ __layout.svelte
-│ └ __layout-foo.svelte
+│ └ __layout.svelte
+│ __layout.svelte
+└ __layout-home@.svelte
```
benmccann, line 166: Could we just call it `__layout-home.svelte` instead?

documentation/docs/02-layouts.md
```diff
@@ -113,64 +113,60 @@ src/routes/
...
+__layout.svelte
+  __layout-a@.svelte
+    __layout-b@a.svelte
+      __layout-c@b.svelte
+        nested@c.svelte
+```
+
+#### Resetting layouts
+
+Sometimes you need a page to use the 'home' layout, regardless of its URL. You can do this by creating a top-level named layout that inherits from the default layout by using `@` without a name:
```
benmccann, line 155:
```suggestion
You can create a named layout that does not inherit from others by using `@` without a name. E.g. you could do this to have a page use the 'home' layout, regardless of its URL:
```

documentation/docs/02-layouts.md
mrkishi: So, reading your recent comments it seems we are still unclear about the semantics of inheritance and the `@` symbol.

What's unanimous so far is that `__layout.svelte` is a default layout, `__layout-foo.svelte` is a layout named `foo`, and `__layout[-foo]@bar.svelte` is a layout that extends the nearest layout named `bar`. 

The confusing parts are:

What do we mean by `__layout-foo@.svelte`? There are two options:

1) A layout `foo` that inherits from the nearest unnamed layout, aka. the default layout;
2) A layout `foo` that inherits from _no_ layout, aka. a layout reset.

Likewise, what do we mean by `__layout-foo.svelte`?

1) A layout that extends the nearest `foo`.
2) A layout that extends the nearest `foo` or, if that doesn't exist, the nearest default layout.

We must pick the same answer for both questions, as they are opposite sides of the same issue. As it stands, the first option is implemented.

We could switch to the second option, but I'm not sure it'd be more intuitive. For instance, with option 1, `__layout-foo.svelte` is equivalent to `__layout-bar@foo.svelte`. With option 2, you'd have too look up a parent layout `foo` differently depending on usage: `@foo` means `foo`, while `__layout-foo` means `foo` _or_ default layout.

I'd prefer option 1 and just work on rephrasing the docs. 

That said... I know @Rich-Harris dislikes reserved names, and we tried not to go there, but at this point I'm wondering if we should just _not_ support `@.svelte` at all and make `default` a reserved name for the default layout. It'd be clearer than picking a meaning for the empty shorthand.

benmccann: > What do we mean by `__layout-foo@.svelte`?

If I read the docs correctly, then it's currently implemented as option 1. Layout resets are to be done using named layouts. I mostly agree with this, but question why `@` is needed at all in this case instead of just calling it `__layout-foo.svelte`

> Likewise, what do we mean by __layout-foo.svelte?

I'd go with option 1 here. And I'd probably have it throw an exception if `foo` does not exist.

mrkishi: > If I read the docs correctly, then it's currently implemented as option 1.

Yes, that's correct.

> I mostly agree with this, but question why `@` is needed at all in this case instead of just calling it `__layout-foo.svelte`

Because if we keep option 1, `__layout-foo.svelte` would be a reset instead of inheriting from the default layout, and in this case the intention would be to have `__layout-home@.svelte` inherit from `__layout.svelte` so we could reset to "the root layout" instead of an empty one.

mrkishi: I still believe the first option is easier to understand and, if we did consider reserving the `default` layout name, it'd be even easier. The whole ruleset would be:

- `__layout[-layout][@parent].svelte` defines a layout named `name = layout ?? 'default'` that extends the nearest layout named `parent ?? name`.
- `some-page[@parent].svelte` defines a page that extends the nearest layout named `parent ?? 'default'`.
- If `@parent` is explicitly specified it must be known, ie. `...@unknown.svelte` is an error.
- "Nearest layout named foo" means the first layout named `'foo'` found in the current folder or its parents.

In this case, `__layout-home@.svelte` wouldn't be a thing and we'd have to be explicit with `__layout-home@default.svelte`.

benmccann: Ok. I think that makes sense and if we just clarify that in the docs it should probably be sufficient

benmccann: I like being explicit with `default`

documentation/docs/02-layouts.md
Rich-Harris: marking this resolved as the API has changed since this suggestion

documentation/docs/02-layouts.md
Rich-Harris: closing as API was updated

packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -229,151 +272,169 @@ export default function create_manifest_data({
...
+	const parts = file.split('/');
+	const filename = /** @type {string} */ (parts.pop());
+	const extension = /** @type {string} */ (extensions.find((ext) => file.endsWith(ext)));
+	const base = filename.slice(0, -extension.length);
+
+	let layout_id = base.includes('@') ? base.split('@')[1] : DEFAULT;
+
+	// walk up the tree, find which __layout and __error components
+	// apply to this page
+	while (true) {
```
benmccann, line 302: looks like you need to update this line to satisfy eslint
```suggestion
	while (parts.length > 0) {
```

packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -229,151 +272,169 @@ export default function create_manifest_data({
...
+		const node = tree.get(parts.join('/'));
+		const layout = node?.layouts[layout_id];
+
+		errors.unshift(node?.error);
+		layouts.unshift(layout?.file);
+
+		if (layout?.name.includes('@')) {
+			layout_id = layout.name.split('@')[1];
+		} else {
+			if (parts.length === 0) break;
```
benmccann, line 312: you could change this line together with the code change above
```suggestion
		} else if (parts.length > 0)  {
```

packages/kit/src/core/sync/create_manifest_data/index.js
Conduitry: Or, alternatively, something like this:

```suggestion
	for (;;) {
```

(or add an `eslint-ignore` comment).

documentation/docs/02-layouts.md
```diff
@@ -62,11 +62,75 @@ We can create a layout that only applies to pages below `/settings` (while inher
...
+├ a/
+│ ├ b/
+│ │ ├ c/
+│ │ │ ├ __layout.svelte                     # skipped
+│ │ │ └ page-with-root-layout@home.svelte
+│ │ └ __layout.svelte                       # skipped
+│ └ __layout.svelte                         # skipped
+├ __layout.svelte                           # inherited
+└ __layout-home@default.svelte              # inherited
+```
```
mrkishi, line 134: Not great englishing on my part, but this is probably a good place to take resets home (and make them searchable):

````suggestion
#### Layout resets

Taking into account scoping and inheritance rules, we can reset layouts to any depth we wish by carefully creating named layouts at the right position in the tree:

```
src/routes/
├ __layout.svelte
├ __layout-blank.svelte
└ a/
  ├ __layout@a.svelte
  ├ __layout-a.svelte
  └ b/
    ├ __layout.svelte
    └ c/
      ├ __layout.svelte
      ├ page-with-c's-layout.svelte
      ├ page-with-a's-layout@a.svelte
      └ page-with-empty-layout@blank.svelte
```

````

File: packages/kit/src/core/sync/create_manifest_data/index.js (hunk 7 above)
Comment (Rich-Harris) [reply packages/kit/src/core/sync/create_manifest_data/index.js:303]: `parts.length > 0` isn't correct, because it will fail for files in the root directory. i added an eslint-ignore comment

File: packages/kit/src/core/sync/create_manifest_data/index.js (hunk 8 above)
Comment (Rich-Harris) [reply packages/kit/src/core/sync/create_manifest_data/index.js:313]: unfortunately the algorithm would then be incorrect

File: documentation/docs/02-layouts.md [hunk 10]
```diff
@@ -62,11 +62,75 @@ We can create a layout that only applies to pages below `/settings` (while inher
...
```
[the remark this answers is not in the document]
Rich-Harris: I think it's important to include the description of what this enables (aka why you'd use it), including the explanation that it's cool and normal to have a layout that just contains a `<slot/>` for this purpose

packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -40,172 +61,194 @@ export default function create_manifest_data({
...
+		.forEach((unit) => {
+			// TODO when we introduce layout endpoints and scoped middlewares, we
+			// will probably want to have a single unified route type here
+			// (created in the list_files(...).forEach(...) callback)
+			if (unit.page) {
+				routes.push({
+					type: 'page',
+					id: unit.id,
+					pattern: unit.pattern,
+					path: unit.id.includes('[') ? '' : `/${unit.id.replace(/@(?:~|[a-zA-Z0-9_-]*)/g, '')}`,
```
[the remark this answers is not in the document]
Rich-Harris: if the names weren't valid, they shouldn't have got this far anyway

documentation/docs/02-layouts.md
Rich-Harris: I've been using this ropey-as-hell REPL app to generate the file tree layouts: https://svelte.dev/repl/5f66998f2c924a9eadb0954896682616?version=3.46.6. To make it consistent with the others it'd look more like this:


````suggestion
#### Layout resets

Taking into account scoping and inheritance rules, we can reset layouts to any depth we wish by carefully creating named layouts at the right position in the tree:

```
src/routes/
├ a/
│ ├ b/
│ │ ├ c/
│ │ │ ├ __layout.svelte
│ │ │ ├ page-with-c's-layout.svelte
│ │ │ ├ page-with-a's-layout@a.svelte
│ │ │ └ page-with-empty-layout@blank.svelte
│ │ └ __layout.svelte
│ ├ __layout@a.svelte
│ └ __layout-a.svelte
├ __layout.svelte
└ __layout-blank.svelte
```

````

Though I wonder if the `__layout-a`/`__layout@a` pair is a bit cryptic as things stand. Maybe something like this would be easier to follow? Though it uses a slightly different sorting to the other blocks, so they'd probably need to change. 

```
src/routes/
├ __layout.svelte        # (1)
├ __layout-root.svelte   # (2)
├ foo/
│ ├ __layout-foo.svelte  # (3)
│ ├ __layout@foo.svelte  # (4)
│ ├ bar/
│ │ ├ __layout.svelte    # (5)
│ │ ├ one.svelte         # 1 -> 3 -> 4 -> 5 -> one.svelte
│ │ ├ two@foo.svelte     # 1 -> 3 -> 4 -> two@foo.svelte
│ │ └ three@root.svelte  # 2 -> three@root.svelte
│ └ four.svelte          # 1 -> 3 -> 4 -> four.svelte
```

````

File: documentation/docs/02-layouts.md (hunk 10 above)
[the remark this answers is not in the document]
Comment (benmccann) [reply documentation/docs/02-layouts.md:133]: Ok. I added that back to the code suggestion. The notion of "skipping" default layouts was just not quite how I thought of it as I think of it more as a lookup of the name. Not that either is wrong, but it's just a bit easier for me to conceptualize that way

File: documentation/docs/02-layouts.md (hunk 9 above)
Comment (mrkishi) [reply documentation/docs/02-layouts.md:134]: Yeah, and that's a use-case I would expect most people would never touch, so it was conflicting adding it.

Would inverting it to `__layout-foo@default.svelte` be easier or hard to grasp, I wonder? Or perhaps moving it a layer down like `/a/b/__layout-extends-a.svelte` (used `a` to be referenced at `c` since `/foo/bar` wouldn't be deep enough to get the point across)?

File: documentation/docs/02-layouts.md (hunk 10 above)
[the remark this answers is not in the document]
Comment (Rich-Harris) [reply documentation/docs/02-layouts.md:133]: FYI you can include code blocks in suggestions by incrementing the number of backticks, like so:

`````
````suggestion
### Some markdown

```js
// some code
```
````
`````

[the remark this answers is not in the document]
Comment (Rich-Harris) [reply documentation/docs/02-layouts.md:133]: (I edited your comment already)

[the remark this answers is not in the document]
Comment (Rich-Harris) [reply documentation/docs/02-layouts.md:133]: tweaks:

````suggestion
Layouts that don't specify a parent layout will inherit from default (i.e. unnamed) layouts in parent directories. If you'd like to inherit from a default layout in the _current_ directory you can specify `@default`. For example, a layout called `__layout-home@default.svelte` that contained a lone `<slot/>` in the root directory would effectively enable any page, anywhere in the app, to inherit directly from the root layout — ignoring any intermediate layouts — by specifying `@home`:

```
src/routes/
├ a/
│ ├ b/
│ │ ├ c/
│ │ │ ├ __layout.svelte
│ │ │ └ page-with-root-layout@home.svelte
│ │ └ __layout.svelte
│ └ __layout.svelte
├ __layout.svelte                           # selected by `@default`
└ __layout-home@default.svelte              # selected by `@home`
```
````

File: documentation/docs/02-layouts.md (hunk 9 above)
Comment (Rich-Harris) [reply documentation/docs/02-layouts.md:134]: I started writing "I think inverting it probably makes a ton of sense, since `@default` has already been discussed"...

````suggestion
#### Layout resets

Taking into account scoping and inheritance rules, we can reset layouts to any depth we wish by carefully creating named layouts at the right position in the tree:

```
src/routes/
├ __layout.svelte                # (1)
├ __layout-root.svelte           # (2)
├ foo/
│ ├ __layout.svelte              # (3)
│ ├ __layout-foo@default.svelte  # (4)
│ ├ bar/
│ │ ├ __layout.svelte            # (5)
│ │ ├ one.svelte                 # 1 -> 3 -> 5 -> one.svelte
│ │ ├ two@foo.svelte             # 1 -> 3 -> 4 -> two@foo.svelte
│ │ └ three@root.svelte          # 2 -> three@root.svelte
│ └ four.svelte                  # 1 -> 3 -> four.svelte
```

````

...but then we miss the fact that `four.svelte` inherits from a layout that inherits from the `foo` layout. Gah...

Comment (mrkishi) [reply documentation/docs/02-layouts.md:134]: I'm not sure I follow—the example is supposed to be about resetting to layouts at arbitrary depths, and `__layout-foo@default` would indeed be an empty `<slot/>` anchor into `/foo/__layout`. Or at least that was my original intention with the "using `a`'s layout" example.

Comment (mrkishi) [reply documentation/docs/02-layouts.md:134]: Although, yeah, `/foo/bar` isn't deep enough since `/foo/bar/__layout-alt.svelte` would have the same meaning.

It'd make more sense having `/foo/__layout-foo@default.svelte` and `/foo/bar/baz/page@foo.svelte` to hit the idea that `page` shares the same layout as `/foo/four.svelte`.

Comment (Rich-Harris) [reply documentation/docs/02-layouts.md:134]: just gonna go with the most recent suggestion for now

Comment (Rich-Harris) [reply documentation/docs/02-layouts.md:134]: oh hang on, missed your comments because github's UI sucks balls

**Comment by ondrejjcizek** [comment]
I still don't get it. When I have nested folder and there I created layout-foo.svelte
Now I after update 307 I can use  in the same folder => index@foo.svelte
But I have there layout from foo and default layout also. How does it work then?

**Comment by smithliam** [comment]
How can I simulate the same `__layout.reset.svelte` functionality with this new feature that mostly feels like a downgrade?😐

**Comment by ObieMunoz** [comment]
> I still don't get it. When I have nested folder and there I created layout-foo.svelte Now I after update 307 I can use in the same folder => [index@foo.svelte](mailto:index@foo.svelte) But I have there layout from foo and default layout also. How does it work for index.svelte in nested folder then? When I want to use as a reset layout as before?

I'm currently having this exact same issue.  Default layout still appears and wraps around the named layout. Is there a solution you have found?

**Comment by ObieMunoz** [comment]
> > ]: I placed the named layout in /routes instead of having it in the nested folder and it works as intended
> 
> Interesting, that doesn't seem to be the case for me. I'm trying to implement an auth guard behavior and been racking my brain on it a bit.
> 
> I'm trying to use the default layout as an auth guard to redirect unauthenticated users for all pages, and then using a named `__layout@public.svelte` to opt-out of authentication for pages like login/sign-up. I have console logs inside the __layout files and it seems on my login/sign-up pages, both the named layout and default layout are both firing.
> 
> My file structure is like this:
> 
> ```
> /routes
>   __layout.svelte
>   __layout-public.svelte
>   /login
>      index@public.svelte
> ```

That's pretty much what my new structure looks like.  I had originally tried doing:

```
/routes
  /page1
      __layout-test.svelte
      [slug]@test.svelte
  __layout.svelte
```
but it was wrapping the default __layout.svelte (contains header and footer) around the "test" layout.

I switched to this structure:
```
/routes
  /page1
      [slug]@test.svelte
  __layout.svelte
  __layout-test.svelte
```
and it works the way I thought it should, only rendering the test layout.

danfoust:
@ObieMunoz Interesting, that's what I had, and it wasn't working.  The named layout was in the same directory as the main layout.  I scrapped my project and starting over from scratch I can't reproduce, and it seems to be working as expected.

ondrejjcizek:
> How can I simulate the same `__layout.reset.svelte` functionality with this new feature that mostly feels like a downgrade?😐
It seems to be better way how to improve old classic `__layout.reset.svelte` 😁

PR #4390, opened 2022-03-19
Create Svelte: Default App CSS fix

aMediocreDad:
This is a simple change to the CSS in the skeleton app. 

To test the issue, access any of the kit-demo links (e.g. https://cloudflare.demo.svelte.dev/) on an iOS device and attempt to tap the counter button multiple times. It will result in the page zooming in due to the native "double-tap-to-zoom" interaction on iOS. 

Adding `touch-action: manipulation` to the button's css will disable the "double-tap-to-zoom"-interaction when tapping the buttons. 

That is all.

- [ ] ~Ideally, include a test that fails without this PR but passes with it.~ (Not aware that this is possible to test)

### Tests

### Changesets

### DIFF
--- a/.changeset/lemon-keys-run.md
+++ b/.changeset/lemon-keys-run.md
@@ -1,5 +1,5 @@
 ---
-'default-template': patch
+'create-svelte': patch
 ---
 
 Add touch-interaction property to button selector in Counter.svelte
--- a/.changeset/lemon-keys-run.md
+++ b/.changeset/lemon-keys-run.md
@@ -2,4 +2,4 @@
 'create-svelte': patch
 ---
 
-Add touch-interaction property to button selector in Counter.svelte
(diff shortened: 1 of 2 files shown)


.changeset/lemon-keys-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'default-template': patch
```
Rich-Harris, line 2:
```suggestion
'create-svelte': patch
```

.changeset/lemon-keys-run.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add touch-interaction property to button selector in Counter.svelte
```
Rich-Harris, line 5:
```suggestion
Fix iOS double-tap zoom on counter buttons
```

Rich-Harris (approved):
TIL! Thank you

Issue #4314, opened 2022-03-12
`stuff` not populated when using `goto` to reload page on an `__error` page


geoffrich:
### Describe the bug

This is a bit of an edge case, but if you:

1. Throw an error on a given route and show an error page
2. Provide a "reload" button on that error page that runs `goto($page.url.toString())`
3. The route passes `stuff` to its parent layout

On hitting the reload button, the page's `stuff` will not be returned, though any props will be. Navigating away from and back to the page will populate stuff appropriately.

### Reproduction

[Stackblitz](https://stackblitz.com/edit/sveltejs-kit-template-default-urpphf?file=src/routes/__layout.svelte)

`/page2` is configured to return the following from `load`.
```js
{
	stuff: {
		page2: true
	},
	props: {
		number: 2
	}
}
```

Repro steps:

1. Click on "Page 2" to navigate to the page with the error. The route is configured to only throw an error on first load. The error page will be shown.
2. Click the "Reload" button on the error page, which calls `goto($page.url.toString())`. The page will successfully render with the props passed from `load`, but `page2: true` will not be in stuff.
3. Click on "Home" to navigate home. Then click on "Page 2" again. This time `page2: true` _will_ be in stuff.

Expected behavior: `page2: true` is present in `stuff` even after reloading an error page with `goto`.

### Logs

```shell
n/a
```

### System Info

```shell
Can't run npx envinfo on StackBlitz.

@sveltejs/kit: 1.0.0-next.295
```

### Severity

annoyance

### Additional Information

I ran into this behavior on one of my SvelteKit sites. I'm rendering an `__error` page with a reload button. The page uses `stuff` to pass a title to the parent `__layout`, but the title is undefined after a reload from the error page.

PR #4392, opened 2022-03-19
[fix] Apply updated `stuff` (and other `page` properties)  when `get_navigation_result_from_branch` called

baseballyama:
fix: https://github.com/sveltejs/kit/issues/4314

### Tests

### Changesets


Rich-Harris:
Thank you. I made a change so that it only updates the page store when something has actually changed, rather than on every update

Rich-Harris (approved):

Issue #3862, opened 2022-02-12
[adapter-cloudflare-workers] Support writing rendered page to Cloudflare cache to achieve ISR


ivanvanderbyl:
### Describe the problem

The initial build on Cloudflare is quick, however, we'd like to cache it for a certain number of seconds on the Cloudflare cache or KV store to achieve something akin to Incremental Static Rendering in Next, where it will only re-render every N seconds.

This would also dramatically improve TTFB.

### Describe the proposed solution

Add an option (or use information from load()) to indicate to the Cloudflare Worker to write the rendered result to Cloudflare cache with a TTL after rendering. Check if rendered result exists in cache before rendering.

### Importance

would make my life easier

### Additional Information

I'd be happy to implement it, if this aligns with the roadmap etc.

Rich-Harris:
Do you mean https://kit.svelte.dev/docs/loading#output-maxage? Or something else?

ivanvanderbyl:
Hey @Rich-Harris — my understanding is that `maxage` controls the data fetching, but the page will still be rendered on each request, is that right? What I want to do is cache the rendered output on the KV store so that it get synced to every region, then have it expire after `maxage`. Effectively only rendering public pages once, incrementally, and reducing the TTFB as much as possible. 

I did an experiment with a custom worker and pushing the output to KV store and it increased our pagespeed score from 87 to 100.

Rich-Harris:
`maxage` does two things:

1. it causes a `cache-control` header to be added to the response
2. during client-side navigation, it causes data to be cached in memory (so `load` doesn't re-run until sufficient time has elapsed)

The `cache-control` header will be `public` (i.e. CDN-cacheable) unless you use `session` or make a credentialled `fetch` (i.e. a `fetch` without `credentials: 'omit'`) during SSR.

I'm not sure whether the cached page is synced to other regions or not, would be curious to know. If it isn't, then I can see the value of this — though it can't happen willy-nilly, because in many cases the SSR'd page _will_ use `session` or a credentialled `fetch`. Ideally it wouldn't be Cloudflare-specific, but would instead be a generic API that adapters could use or ignore (as I don't think my bosses would forgive me if we supported ISR on Cloudflare before Vercel! Relatedly, the API would ideally support [on-demand ISR](https://nextjs.org/blog/next-12-1#on-demand-incremental-static-regeneration-beta) as well as on-request).

isaac-mcfadyen:
[...]

PR #4412, opened 2022-03-21
Add cloudflare cache

johnwalshuk:

### Tests

### Changesets

Resolves https://github.com/sveltejs/kit/issues/3862. 

Add cache to dynamically generated pages at the edge by using the cloudflare cache api. Pages that have a cache policy (maxage) will be stored.

### DIFF
--- a/packages/adapter-cloudflare/files/worker.js
+++ b/packages/adapter-cloudflare/files/worker.js
@@ -50,8 +50,7 @@ export default {
 
 		// dynamically-generated pages
 		try {
-			// `default` cache property exists in the cloudflare workers environment only.
-			// @ts-expect-error
+			// @ts-expect-error - `default` exists only in the cloudflare workers environment
 			const cache = caches.default;
 			let response = await cache.match(req);
 
--- a/packages/adapter-cloudflare/files/worker.js
+++ b/packages/adapter-cloudflare/files/worker.js
@@ -65,7 +65,7 @@ export default {
 				// Use context so you can return the response without blocking on
 				// writing to cache
 				try {
-					context(cache.put(req, response.clone()));
+					context.waitUntil(cache.put(req, response.clone()));
 				} catch {
 					// noop
 				}
--- a/packages/adapter-cloudflare/files/worker.js
+++ b/packages/adapter-cloudflare/files/worker.js
@@ -50,12 +50,25 @@ export default {
 
 		// dynamically-generated pages
 		try {
-			return await server.respond(req, {
-				platform: { env, context },
-				getClientAddress() {
-					return req.headers.get('cf-connecting-ip');
-				}
-			});
+			const cache = caches.default;
+			let response = await cache.match(req);
+
+			if (!response) {
+				response = await server.respond(req, {
+					platform: { env, context },
+					getClientAddress() {
+						return req.headers.get('cf-connecting-ip');
+					}
+				});
+
+				// Use context so you can return the response without blocking on
+				// writing to cache
+				try {
+					context(cache.put(req, response.clone()));
+				} catch {}
+			}
+
+			return response;
 		} catch (e) {
 			return new Response('Error rendering route: ' + (e.message || e.toString()), { status: 500 });
 		}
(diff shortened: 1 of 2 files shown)


PH4NTOMiki:
@johnwalshuk You aren't checking for the max-age(time nor even is it present) as far as i can see

johnwalshuk:
> @johnwalshuk You aren't checking for the max-age(time nor even is it present) as far as i can see

Cloudflare inspects the response headers automatically. See https://developers.cloudflare.com/workers/runtime-apis/cache/#headers

PH4NTOMiki:
Didn't know that, but still, relying on it isn't really advisable.

lukeed:
@PH4NTOMiki why?

The only things to be wary of are [invalid parameters](https://developers.cloudflare.com/workers/runtime-apis/cache/#invalid-parameters) and [errors](https://developers.cloudflare.com/workers/runtime-apis/cache/#errors). But there's a try/catch in place already so both can be ignored in this case. 

Personally I [manually check](https://github.com/lukeed/worktop/blob/next/packages/worktop/src/cache.ts#L22-L30) before any calls to `cache.put` just to be sure, but what's here (after suggestion) should be okay for general use. If anything needs to be added, it's appending `'private=Set-Cookie'` when there's a cookie present.

PH4NTOMiki:
@lukeed okay, if you say that, then I stand corrected. but as you said personally, checking it is nice, and it's clearer what's going on in the code.

lukeed (approved):

packages/adapter-cloudflare/files/worker.js
```diff
@@ -50,12 +50,28 @@ export default {
 
 		// dynamically-generated pages
 		try {
-			return await server.respond(req, {
-				platform: { env, context },
-				getClientAddress() {
-					return req.headers.get('cf-connecting-ip');
+			// @ts-ignore
```
benmccann, line 53: Is there some package like https://www.npmjs.com/package/@cloudflare/workers-types that we can use so that we don't have to `@ts-ignore` this?

If there's no solution then it should be turned into `@ts-expect-error` along with a comment explaining why it cannot be addressed

johnwalshuk: Hey @benmccann, I tried to add the package above but it started to throw ~30 errors in the JS files so for now I've used `@ts-expect-error` with a comment as to why.

packages/adapter-cloudflare/files/worker.js
```diff
@@ -50,12 +50,35 @@ export default {
 
 		// dynamically-generated pages
 		try {
-			return await server.respond(req, {
-				platform: { env, context },
-				getClientAddress() {
-					return req.headers.get('cf-connecting-ip');
+			// `default` cache property exists in the cloudflare workers environment only.
+			// @ts-expect-error
```
benmccann, line 54:
```suggestion
			// @ts-expect-error - `default` exists only in the cloudflare workers environment
```

benmccann (approved):

jahir9991:
how can I disable this feature?

mastermakrela:
@jahir9991  > how can I disable this feature?

I'm not sure if it's the right solution (or if we had the same problem 😅).
But after adding the following headers, the Responses aren't cached any more:

```
headers: {
  "Cache-Control": "no-store",
  "max-age": "no-cache",
}
```

lukeed:
The cache-control header is the only relevant header here

jahir9991:
how can configure it for disable cache  in global layer

lukeed:
You can use a global [`handle`](https://kit.svelte.dev/docs/hooks#handle) hook to ensure Cache-Control headers are doing what you want:

```js
// src/hooks.js

/** @type {import('@sveltejs/kit').[Handle](https://kit.svelte.dev/docs/types#sveltejs-kit-handle)} */
  event.request = new Request(event.request);
  event.request.headers.set('cache-control', 'no-cache');
 
  let response = await resolve(event);

  response = new Response(response.body, response);
  response.headers.set('cache-control', 'no-cache');

  return response;
}
```
> untested, but should work based on docs

Issue #2515, opened 2021-09-28
redirect to another domain


wbster:
### Describe the bug

the link leads to someone else's site
[`http://localhost:3000//https://google.com/`](http://localhost:3000//www.google.com/)
can be used by hackers

### Reproduction

found on all versions of sveltekit. Redirect to another site, can be used by hackers

### System Info

```shell
all systems
```

### Severity

blocking all usage of SvelteKit

Conduitry:
How do we reproduce this? If I go to, for example, https://kit.svelte.dev//https://google.com/ I get a 404, as expected. Does this only happen in dev mode?

bluwy:
Reproducible in dev mode. Tried setting `config.kit.server.fs.strict: true` but with the same behaviour. This is likely an issue in Vite's dev server, and should be reported there if not duplicate.

Conduitry:
Added the "vite" label - I also removed the "1.0" milestone, since this appears to be a dev-only issue.

wbster:
Should I tell them about this problem or have they already been informed?

rmunn:
Can easily reproduce with a new Svelte-Kit skeleton app, but I've tried multiple ways to reproduce this with a `create-vite` app and haven't been able to reproduce it with only Vite involved. So I can't rule out a Svelte-Kit issue just yet.

BTW, two findings from my repro attempts might be useful.

1. The final `/` and initial `//` are required to trigger this bug. http://localhost:3000/https://www.google.com/ did not redirect, but http://localhost:3000//https://www.google.com/ did. Also, http://localhost:3000//https://www.google.com did not redirect.
2. I added a middleware before the Svelte-Kit middleware that would just log the request URL and call `next()`. In the examples above that did not redirect, that middleware logged the URL and then logged the various other URLs that Svelte-Kit loaded (e.g. `/.svelte-kit/dev/runtime/internal/start.js`). But on the attempt that triggered the redirect, my logging middleware was never hit; the http://localhost:3000//https://www.google.com/ URL never showed up in the logs.

bluwy:
I was just about to refer to open an issue in Vite as I'm able to reproduce this in a normal Vite app. SvelteKit uses the Vite dev server under-the-hood so I think the core fix is in Vite. I've also check Vite's issue tracker and there doesn't seem to be an existing one.

rmunn:
[...]

PR #4414, opened 2022-03-21
[fix] don't redirect to external URLs

benmccann:
fixes https://github.com/sveltejs/kit/issues/2515

### DIFF
--- a/.changeset/curvy-foxes-change.md
+++ b/.changeset/curvy-foxes-change.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-[fix] don't redirect to external URLs
+[fix] don't redirect to external URLs when normalizing paths
--- a/.changeset/curvy-foxes-change.md
+++ b/.changeset/curvy-foxes-change.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] don't redirect to external URLs
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -18,15 +18,14 @@ export async function respond(request, options, state) {
 
 	const normalized = normalize_path(url.pathname, options.trailing_slash);
 
-	if (
-		normalized !== url.pathname &&
-		!state.prerender?.fallback &&
-		!/^(https?:)?\/\//.test(normalized)
-	) {
+	if (normalized !== url.pathname && !state.prerender?.fallback) {
 		return new Response(undefined, {
 			status: 301,
 			headers: {
-				location: normalized + (url.search === '?' ? '' : url.search)
+				location:
+					// ensure paths starting with '//' are not treated as protocol-relative
+					(normalized.startsWith('//') ? url.origin + normalized : normalized) +
+					(url.search === '?' ? '' : url.search)
 			}
 		});
 	}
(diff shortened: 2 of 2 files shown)


mrkishi:
I believe it'd make sense to put the full url into the `location` header instead. Even if unusual, those filtered paths are valid.

Conduitry:
Apologies for the comment spam. I'm now re-reading @mrkishi's comment in light of what this PR is supposed to actually be fixing, and I'm a bit confused.

This does feel like an incorrect fix to me, but I'm not sure what a correct fix would be. When we're redirecting to the normalized version of a particular URL, we should probably make sure we're always specifying a URL that's host-relative. `normalize_path` is always going to be getting passed just the path part, and so that's what `normalized` is going to contain as well. In what circumstances is this generating a redirect that the browser is interpreting as an absolute URL redirect, and how can we stop precisely those situations, without having a hardcoded `https?://` check?

mrkishi:
I share your concerns when absolute URLs are involved, @Conduitry, but wouldn't other things break if we didn't have the correct origin/host settings?

As to why these paths cause issues: we normalize the pathname and send it back as a root-relative URL in the location header. That's fine for most pathnames, but if they start with a double-slash you end up with a scheme-relative URL instead: `//path` resolves to `scheme://path` instead of `scheme://host//path`.

Conduitry:
Okay, yet more comment spam:

I think the `(https?:)?` part of the check is unnecessary. This is _always_ going to be checking something that comes from `normalize_path`, which is _always_ going to be something that begins with `/`.

I also think that preventing the redirect from happening here is unnecessarily prohibitive. I don't think we want to refuse to handle paths that happen to begin with two slashes (as weird as that might be). We should instead make sure we return a `Location` header that the browser won't misinterpret as a protocol-relative URL. The only way to do that might simply be to prepend the original origin, which brings us back to @mrkishi's suggestion. I don't know whether it would be better to avoid including the origin whenever possible, or to consistently include the origin, forcing everyone with weird setups to potentially worry about that for the sake of this weird case.

.changeset/curvy-foxes-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] don't redirect to external URLs
```
Conduitry, line 5:
```suggestion
[fix] don't redirect to external URLs when normalizing paths
```

Conduitry (approved):
I've gone back and forth a few times on whether it would be better to always use the origin or only use it when the path begins with `//`. I think this limited change will, overall, be better.

It's less likely to break already-working apps. Apps that legitimately use paths beginning with `//` are probably pretty uncommon. And, if the origin setting is incorrect, getting redirected there when normalizing a `//` path seems pretty unlikely to cause confusing behavior for a real user - they would likely just be sent to a hostname that doesn't resolve.

saostad:
This is the workaround I put in my project:
```json
"@sveltejs/adapter-auto": "^2.0.0",
"@sveltejs/kit": "^1.15.2",
"svelte": "^3.58.0",
"svelte-check": "^3.2.0",
"svelte-preprocess": "^5.0.3",
"vite": "^4.2.1"
```

```js
function validateUrl(url: string) {
		if (url.startsWith('http://') || url.startsWith('https://')) {
			return url;
		} else {
			return `http://${url}`;
		}
	}
```

PR #4424, opened 2022-03-22
Fix preprocess etc

locriacyber:
Ship svelte-preprocess by default, make produced project work with latest svelte version + minor fixes (check commits)

svelte-preprocess is useful to bundle by default, so the framework user can use SASS without further config.
 
---

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/package.json
+++ b/packages/create-svelte/package.json
@@ -19,6 +19,7 @@
 		"@types/gitignore-parser": "^0.0.0",
 		"@types/prettier": "^2.4.2",
 		"@types/prompts": "^2.0.14",
+		"eslint": "^8.11.0",
 		"gitignore-parser": "^0.0.2",
 		"prettier": "^2.5.0",
 		"prettier-plugin-svelte": "^2.5.0",
@@ -28,6 +29,7 @@
 		"tiny-glob": "^0.2.9"
 	},
 	"scripts": {
+		"start": "node bin.js",
 		"build": "node scripts/build-templates",
 		"check": "tsc",
 		"lint": "eslint --ignore-path .gitignore --ignore-path ../../.gitignore \"./*.js\" && npm run check-format",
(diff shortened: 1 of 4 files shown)


Conduitry:
What is this changing?

locriacyber:
See first message.

packages/create-svelte/package.json
```diff
@@ -19,6 +19,7 @@
 		"@types/gitignore-parser": "^0.0.0",
 		"@types/prettier": "^2.4.2",
 		"@types/prompts": "^2.0.14",
+		"eslint": "^8.11.0",
```
Conduitry, line 22: Why is this being added?

packages/create-svelte/package.json
```diff
@@ -28,6 +29,7 @@
 		"tiny-glob": "^0.2.9"
 	},
 	"scripts": {
+		"start": "node bin.js",
```
Conduitry, line 32: Why is this being added?

packages/create-svelte/package.json
benmccann, line 22: why?

packages/create-svelte/package.json
benmccann, line 32: why?

packages/create-svelte/templates/skeleton/svelte.config.js
```diff
@@ -1,9 +1,14 @@
 import adapter from '@sveltejs/adapter-auto';
+import preprocess from 'svelte-preprocess';
 
 // This config is ignored and replaced with one of the configs in the shared folder when a project is created.
 
 /** @type {import('@sveltejs/kit').Config} */
 const config = {
+	// Consult https://github.com/sveltejs/svelte-preprocess
+	// for more information about preprocessors
+	preprocess: preprocess(),
```
benmccann, line 10: why?

benmccann:
> svelte-preprocess is useful to bundle by default, so the framework user can use SASS without further config.

Most people are not using Sass, so I don't think we'd want to add this by default

packages/create-svelte/package.json
locriacyber: This package needs eslint to run `npm run lint`

packages/create-svelte/package.json
locriacyber: A simple way to test the package

PR #4425, opened 2022-03-22
fixes scroll restoration after native navigation to another sveltekit route return

### LINKED ISSUE #4386: history scroll restoration usage with native navigation
### Describe the problem

I first noticed this when I was browsing around a website built with Next.js, and since #3873 and #3938 we're using similar technique for restoring scroll position as Next, so this bug applies too, but in much smaller scale since we listen to all anchor clicks, and not just Link component like Next does.

The issue is when the page does native navigation to another link that is being handled by SvelteKit, and that second navigation overwrites scroll position inside session storage, so when user returns to the first page, second scroll position is being read.

This is pretty edge case for SvelteKit because of listening to all anchor elements, but still it could happen.

I can push PR for this.

### Alternatives considered

Leave things the way they currently are. Since this is edge case that could be okay.

### Importance

nice to have

### Additional Information

_No response_

PH4NTOMiki:

### Tests

### Changesets

Closes #4386
Probably needs test.

### DIFF
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -10,7 +10,8 @@ import {
 	get_href,
 	initial_fetch,
 	notifiable_store,
-	scroll_state
+	scroll_state,
+	get_random_string
 } from './utils.js';
 import { parse } from './parse.js';
 
@@ -34,7 +35,7 @@ const default_error = components[1]();
 // state we're navigating from
 
 /** @typedef {{ x: number, y: number }} ScrollPosition */
-/** @type {Record<String, ScrollPosition>} */
+/** @type {Record<string, ScrollPosition>} */
 let scroll_positions = {};
 try {
 	scroll_positions = JSON.parse(sessionStorage[SCROLL_KEY]);
@@ -42,7 +43,7 @@ try {
 	// do nothing
 }
 
-/** @param {String} index */
+/** @param {string} index */
 function update_scroll_positions(index) {
 	scroll_positions[index] = scroll_state();
 }
@@ -124,12 +125,11 @@ export function create_client({ target, session, base, trailing_slash }) {
 	let router_enabled = true;
 
 	// keeping track of the history index in order to prevent popstate navigation events if needed
-	/** @type {Number} */
+	/** @type {number} */
 	let current_history_index = history.state?.[INDEX_KEY] ?? 0;
 	// keeping track of the position index in order to restore scroll position
-	/** @type {String} */
-	let current_position_index =
-		history.state?.[RESTORATION_KEY] ?? Math.random().toString(32).slice(2);
+	/** @type {string} */
+	let current_position_index = history.state?.[RESTORATION_KEY] ?? get_random_string();
 
 	if (current_history_index === 0) {
 		// create initial history entry, so we can return here
@@ -269,7 +269,9 @@ export function create_client({ target, session, base, trailing_slash }) {
 			const { details } = opts;
 			const change = details.replaceState ? 0 : 1;
 			details.state[INDEX_KEY] = current_history_index += change;
-			details.state[RESTORATION_KEY] = current_position_index = Math.random().toString(32).slice(2);
+			if (!details.replaceState) {
+				details.state[RESTORATION_KEY] = current_position_index = get_random_string();
+			}
 			history[details.replaceState ? 'replaceState' : 'pushState'](details.state, '', url);
 		}
 
@@ -1150,7 +1152,7 @@ export function create_client({ target, session, base, trailing_slash }) {
 				// we need to update history, otherwise we have to leave it alone
 				if (hash_navigating) {
 					hash_navigating = false;
-					current_position_index = Math.random().toString(32).slice(2);
+					current_position_index = get_random_string();
 					history.replaceState(
 						{
 							...history.state,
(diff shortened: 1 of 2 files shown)


PatrickG:
I was about to look into this issue as well.
I think it would be far easier to change `scroll_positions` into an array and set the `.length` property to `current_history_index + 1` when clicking a link or using `goto`.

PH4NTOMiki:
@Rich-Harris what do you think about this?

PH4NTOMiki:
@PatrickG I don't think that what you suggested would solve #4386 because we need to make collisions impossible, and that won't do it, this will because it uses random keys.

PH4NTOMiki:
And in case I wasn't really clear in the issue, the issue goes like this: 

1. go to some Kit app (the `history.state` would be like this: `{sveltekit:index: 0}`) and scroll to the end/bottom of the page
2. you click on some internal link in that app and for some reason(maybe `router=false`, maybe `app version changed` etc.) navigation is handled by the `browser native` navigation
3. after that the `history.state` would again be `{sveltekit:index: 0}` and that would cause overwrite of position state.
4. go back using browser back button, you would't be returned to the bottom of the page where scroll position should be because you left it there in step 1. because it was overwritten by the second navigation.

PatrickG:
> And in case I wasn't really clear in the issue, the issue goes like this:
> 
> 1. go to some Kit app (the `history.state` would be like this: `{sveltekit:index: 0}`) and scroll to the end/bottom of the page
> 2. you click on some internal link in that app and for some reason(maybe `router=false`, maybe `app version changed` etc.) navigation is handled by the `browser native` navigation
> 3. after that the `history.state` would again be `{sveltekit:index: 0}` and that would cause overwrite of position state.
> 4. go back using browser back button, you would't be returned to the bottom of the page where scroll position should be because you left it there in step 1. because it was overwritten by the second navigation.

Alright, then i misunderstood the issue.
I thought the issue is:
- start page (index 0): click link -> second page (index 1)
- scroll down on second page (index 1)
- go back to start page (index 0)
- klick another link -> thrid page (index 1) -> page scrolls down unintentionally

benmccann:
I agree we should match the browser behavior and not share scroll state across tabs. FYI, Rich is on vacation this week

packages/kit/src/runtime/client/client.js
```diff
@@ -33,15 +34,15 @@ const default_error = components[1]();
 // state we're navigating from
 
 /** @typedef {{ x: number, y: number }} ScrollPosition */
-/** @type {Record<number, ScrollPosition>} */
+/** @type {Record<String, ScrollPosition>} */
```
benmccann, line 37: We normally use lowercase `string` and `number`

https://stackoverflow.com/a/42654496

packages/kit/src/runtime/client/client.js
```diff
@@ -259,6 +269,7 @@ export function create_client({ target, session, base, trailing_slash }) {
 			const { details } = opts;
 			const change = details.replaceState ? 0 : 1;
 			details.state[INDEX_KEY] = current_history_index += change;
+			details.state[RESTORATION_KEY] = current_position_index = Math.random().toString(32).slice(2);
```
PH4NTOMiki, line 272: I'm not sure whether I should put this in `if(!details.replaceState)` contitional

packages/kit/src/runtime/client/client.js
```diff
@@ -19,6 +19,7 @@ import { components, dictionary, matchers } from '__GENERATED__/client-manifest.
 
 const SCROLL_KEY = 'sveltekit:scroll';
 const INDEX_KEY = 'sveltekit:index';
+const RESTORATION_KEY = 'sveltekit:restoration';
```
PH4NTOMiki, line 22: And I'm not sure about this name.

packages/kit/src/runtime/client/client.js
PH4NTOMiki: I did put in inside conditional

packages/kit/src/runtime/client/client.js
PH4NTOMiki: Still not sure, not changing for now

PH4NTOMiki:
> Your solution does fix that as well.

I thought that was working already, do you have a example where it wasn't?

PH4NTOMiki:
@benmccann is it okay to merge this, it works locally, I tested it and it works, and it's really difficult to add test for this because it's pretty edge-case-y.

PH4NTOMiki:
@Rich-Harris can you check this out?

PH4NTOMiki:
@Rich-Harris can you merge this, or just say why it's not merged.

Rich-Harris:
Can you add a test?

PH4NTOMiki:
@Rich-Harris It's pretty hard thing to test for, because the bug happens in some specific conditions, that are hard to create in testing environment. But if you check out the code, you can see that there is not any way that this code could break anything, and I checked out the deployment https://kit-svelte-dev-zu82n3ukp-svelte.vercel.app/ and it works in 3 of my browsers(Edge, Firefox, Chrome)

PH4NTOMiki:
@Rich-Harris but I can hope we can merge it anyway(without the test)

Rich-Harris:
I haven't managed to write a test that fails on `master`, but I _was_ able to simply the implementation somewhat. There's no need to maintain `SCROLL_INDEX_KEY` separately — all we need to do is ensure that cross-document navigations within the same app (or between two SvelteKit apps) don't corrupt the scroll position history, which is easy to do by offsetting the history index. Closing this in favour of #4640 — thanks

PR #4438, opened 2022-03-23
allows for __tests__ & __test__ directories

### LINKED ISSUE #4435: __tests__ directories break the build
### Describe the problem

https://github.com/sveltejs/kit/blob/master/packages/kit/src/core/sync/create_manifest_data/index.js#L85

Keeps me from placing `__tests__` directories alongside my source files:

```
src/index.svelte
src/__tests_/index.svelte
```

I have no idea why this is blocked, so I don't know the ramifications of what I'm asking for. The most brute-force method would be to just hard-code the allowance of `__tests__` directories.

Edit [line 28](https://github.com/sveltejs/kit/blob/master/packages/kit/src/core/sync/create_manifest_data/index.js#L28) to be:

`const specials = new Set(['__layout', '__layout.reset', '__error', '__tests__']);`

### Alternatives considered

Structuring my code like this:

```
src/index.svelte
src/index.test.svelte
```

or:

```
src/index.svelte
src/_tests_/index.svelte
```

The former gets very cluttered, and the latter just doesn't match the conventions used by my team on other projects so will be another tiny little speed-bump in adoption.

### Importance

would make my life easier

### Additional Information

I'm happy to submit a PR, but didn't want to bother if it was just going to be closed.

cormacmccarthy:
fixes #4435

### Tests

### Changesets

### DIFF
--- a/.changeset/modern-cameras-fetch.md
+++ b/.changeset/modern-cameras-fetch.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+allows for **tests** directories
--- a/.changeset/modern-cameras-fetch.md
+++ b/.changeset/modern-cameras-fetch.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-allows for **tests** directories
+allow files and directories named `__tests__` in the routes directory
--- a/.changeset/modern-cameras-fetch.md
+++ b/.changeset/modern-cameras-fetch.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-allow files and directories named `__tests__` in the routes directory
+allow files and directories named `__tests__` and `__test__` in the routes directory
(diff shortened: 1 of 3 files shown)


.changeset/modern-cameras-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+allows for **tests** directories
```
mrkishi, line 5:
```suggestion
allow files and directories named `__tests__` in the routes directory
```

benmccann (approved):

rmunn:
I'd suggest also allowing `__test__` in addition to `__tests__`, because I've seen both `test` and `tests` being used as directory names in projects I've been involved with. So while I haven't personally seen either `__test__` or `__tests__` myself, I wouldn't be surprised if both are in use, with more or less equal usage (I'd guess). So if we're special-casing one, we should special-case both and do it in the same PR so we don't have to revisit the question later on when someone asks for `__test__`.

dominikg (approved):

PR #4453, opened 2022-03-25
chore(cloudflare): optimize Cache usage; strict TS source

lukeed:

Rely on the Cache API sooner within `adapter-cloudflare`. Makes use of [`worktop`](https://github.com/lukeed/worktop) for all Cache operations, which safely writes a `Response` into the Cache only when permitted.

Also converted the `_worker.ts` file to TypeScript for strict type checking, which is doable since we build with `esbuild` anyway.

### Tests

### Changesets


lukeed:
CI is passing. Pages app seems unrelated

Rich-Harris:
The `.svelte-kit/cloudflare` output directory _does_ exist locally if you set `CF_PAGES=1` (which is how `adapter-auto` detects that it needs to delegate to `adapter-cloudflare`). The build failure is being caused by this...

```
13:46:21.752 | packages/create-svelte/templates/default build: > Using @sveltejs/adapter-cloudflare
-- | --
13:46:21.753 | packages/create-svelte/templates/default build:   Detected environment: Cloudflare Pages. Using @sveltejs/adapter-cloudflare
13:46:21.780 | packages/create-svelte/templates/default build: ✘ [ERROR] Could not resolve "worktop/cfw.cache"
13:46:21.780 | packages/create-svelte/templates/default build:     .svelte-kit/cloudflare-tmp/_worker.ts:3:23:
13:46:21.781 | packages/create-svelte/templates/default build:       3 │ import * as Cache from 'worktop/cfw.cache';
13:46:21.781 | packages/create-svelte/templates/default build:         ╵                        ~~~~~~~~~~~~~~~~~~~
13:46:21.781 | packages/create-svelte/templates/default build:   You can mark the path "worktop/cfw.cache" as external to exclude it from the bundle, which will remove this error.
```

...which makes me think it _is_ related to this PR

lukeed:
Ah. Fails because the default template is built from within the `create-svelte` package (`packages/create-svelte/templates/default/.svelte-kit/cloudflare-tmp`) and the `worktop` package isn't resolvable from that location since root & `create-svelte` don't have it installed. This would be true for any runtime dependency, which none of the other adapters currently have.

Rich-Harris:
Of course, that makes sense. In other adapters, we solve this problem by bundling earlier. We could do that here: https://github.com/sveltejs/kit/pull/4473

Rich-Harris (approved):
This LGTM but I'll confess I don't really understand why we need this to be a `.ts` file. I get that it doesn't make a huge difference seeing as we're bundling with esbuild, but I think there's probably some value in using `.js` consistently across the entire repo, rather than introducing an exception. Opened https://github.com/sveltejs/kit/pull/4474

lukeed:
I don't really care about `*.ts` or not – esbuild is already there so it's free either way. I just want the strict types loaded to further prevent type & usage errors we saw in the originating Module Worker transform PR

.changeset/empty-falcons-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/adapter-cloudflare': patch
+---
+
+Convert entry to TypeScript; check for Cache match sooner; use `worktop` for types & Cache operations
```
Rich-Harris, line 5:
```suggestion
Check for Cache match sooner; use `worktop` for types & Cache operations
```

Rich-Harris (approved):

Issue #4327, opened 2022-03-14
Generated types invalid for routes with a rest parameter


TravisSpomer:
### Describe the bug

The automatically generated types file for routes with a rest parameter are incorrect.

If you create a file called `[...foo].svelte` and then inspect the `[...foo].d.ts` file that gets generated, you'll see that the code isn't valid:

```ts
// this file is auto-generated
import type { Load as GenericLoad } from '@sveltejs/kit';
export type Load<Props = Record<string, any>> = GenericLoad<{ ...foo: string }, Props>;
```

If the `...` in front of `foo` in `GenericLoad` is removed, the generated code would be valid and correct.

### Reproduction

Just create a file called `[...foo].svelte` in any project. It doesn't need to have any contents.

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (8) x64 Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz   
    Memory: 21.02 GB / 31.93 GB
  Binaries:
    Node: 16.14.0 - C:\Program Files\nodejs\node.EXE        
    Yarn: 1.22.5 - C:\Program Files (x86)\Yarn\bin\yarn.CMD 
    npm: 8.5.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (99.0.1150.39)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @sveltejs/adapter-static: ^1.0.0-next.29 => 1.0.0-next.29
    @sveltejs/kit: ^1.0.0-next.296 => 1.0.0-next.296
    svelte: ^3.46.4 => 3.46.4
```

### Severity

annoyance+ (it's a compilation error that might break builds in some environments)

PR #4462, opened 2022-03-29
Generated types invalid for routes with matched parameter

istarkov:
# Generated types invalid for routes with matched parameter

Ref #4327 #4361

## Bug description

The automatically generated types file for routes with matched parameters are invalid.
i.e. route like `/[lang=is_lang]/[name]` will generate following

```ts
export type RequestHandler<Output extends ResponseBody = ResponseBody> = GenericRequestHandler<{ lang=is_lang: string; name: string }, Output>;
```

After this PR

```ts
export type RequestHandler<Output extends ResponseBody = ResponseBody> = GenericRequestHandler<{ lang: string; name: string }, Output>;
```

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/sync/write_types.js
+++ b/packages/kit/src/core/sync/write_types.js
@@ -29,7 +29,7 @@ export function write_types(config, manifest_data) {
 		/** @type {string[]} */
 		const params = [];
 
-		const pattern = /\[(?:\.{3})?([^\]]+)\]/g;
+		const pattern = /\[(?:\.{3})?([^\]=]+)[^\]]*\]/g;
 		let match;
 
 		while ((match = pattern.exec(key))) {


benmccann:
Cool. Thanks for the fix. Perhaps it would be good to expose `extract_params` and add a unit test for it

istarkov:
Based on this comment https://github.com/sveltejs/kit/pull/4361#issue-1171548905 
> We could get the canonical list of params from the manifest data for routes

Seems like implementation will be changed seriously some day, as extracting params using different code doesn't look good. 

This is just a quick fix allowing to use modern features. 

PS: For me types generated for components like __layout should be changed too, i.e params for `__layout` should be `Record<string, string>` instead of `{}`

Rich-Harris:
Thank you. Since `write_types` was written, a `parse_route_id` function was added which we can use instead — closing this in favour of #4472

PR #4473, opened 2022-03-31
bundle worktop

Rich-Harris:
Bundles worktop so that apps don't need to install it — see https://github.com/sveltejs/kit/pull/4453#issuecomment-1084763864


lukeed (approved):
This feels wrong, but it gets the job done i guess

I'll update my PR with the extra changes necessary for the build paths

packages/adapter-cloudflare/tsconfig.json
```diff
@@ -11,5 +11,5 @@
 			"@sveltejs/kit": ["../kit/types/index"]
 		}
 	},
-	"include": ["**/*.js", "ambient.d.ts", "files/worker.ts"]
+	"include": ["**/*.js", "ambient.d.ts", "src/worker.ts"]
```
lukeed, line 14:
```suggestion
	"include": ["index.js", "ambient.d.ts", "src/worker.ts"]
```

PR #4474, opened 2022-03-31
Convert adapter-cloudflare back to JS

Rich-Harris:
i think this gives us the same level of type safety

### DIFF
--- a/packages/adapter-cloudflare/src/worker.js
+++ b/packages/adapter-cloudflare/src/worker.js
@@ -2,18 +2,12 @@ import { Server } from 'SERVER';
 import { manifest, prerendered } from 'MANIFEST';
 import * as Cache from 'worktop/cfw.cache';
 
-import type { Module, Bindings } from 'worktop/cfw';
-import type { Durable } from 'worktop/cfw.durable';
-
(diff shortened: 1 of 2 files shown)


lukeed (approved):
Personally would prefer a `@typedef` to break out some of the long-running type info, but as long as the type check is happy then I'm fine

packages/adapter-cloudflare/src/worker.js
```diff
@@ -2,18 +2,12 @@ import { Server } from 'SERVER';
...
-interface Environment extends Bindings {
-	ASSETS: Durable.Object;
-}
-
 const server = new Server(manifest);
 
 const prefix = `/${manifest.appDir}/`;
 
-const worker: Module.Worker<Environment> = {
+/** @type {import('worktop/cfw').Module.Worker<{ ASSETS: import('worktop/cfw.durable').Durable.Object }>} */
```
lukeed, line 9: Optional, free to disregard

```suggestion
/**
 * @typedef Environment
 * @property {import('worktop/cfw.durable').Durable.Object} ASSETS
 */
 
/** @type {import('worktop/cfw').Module.Worker<Environment>} */
```

PR #4475, opened 2022-03-31
apply linking logic to inline code spans

Rich-Harris:
This is a possible solution to #4445 — it makes references to types in inline `<code>` spans link to their definitions, and [updates the `RequestHandler` type accordingly](https://kit-svelte-dev-git-link-inline-codespans-svelte.vercel.app/docs/types#sveltejs-kit-requesthandler):

image

This seems more scalable/less invasive than the alternative, to me

### DIFF
--- a/packages/kit/types/index.d.ts
+++ b/packages/kit/types/index.d.ts
@@ -215,13 +215,9 @@ export interface ParamMatcher {
 }
 
 /**
- * A function exported from an endpoint that corresponds to an
- * HTTP verb (`get`, `put`, `patch`, etc) and handles requests with
- * that method. Note that since 'delete' is a reserved word in
- * JavaScript, delete handles are called `del` instead.
+ * A `(event: RequestEvent) => RequestHandlerOutput` function exported from an endpoint that corresponds to an HTTP verb (`get`, `put`, `patch`, etc) and handles requests with that method. Note that since 'delete' is a reserved word in JavaScript, delete handles are called `del` instead.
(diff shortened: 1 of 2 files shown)


packages/kit/types/index.d.ts
```diff
@@ -215,13 +215,9 @@ export interface ParamMatcher {
 }
 
 /**
- * A function exported from an endpoint that corresponds to an
- * HTTP verb (`get`, `put`, `patch`, etc) and handles requests with
- * that method. Note that since 'delete' is a reserved word in
- * JavaScript, delete handles are called `del` instead.
+ * A `(event: RequestEvent) => RequestHandlerOutput` function exported from an endpoint that corresponds to an HTTP verb (`get`, `put`, `patch`, etc) and handles requests with that method. Note that since 'delete' is a reserved word in JavaScript, delete handles are called `del` instead.
```
benmccann, line 218: should we make a similar change to the description of `Load` to make it easier to read sans generics as well?

Rich-Harris: sure, we can, i just figured we should get the plumbing in first and then make the actual changes separately (apart from this one for demo purposes) rather than worrying about all the content changes straight away

benmccann: I think it's the only other one we might consider to changing from a quick skim though I did earlier, so figured we might as well just get it out of the way. Anyway, this change looks good to me

Issue #4478, opened 2022-04-01
Playwrite with typescript not working


JackDra:
### Describe the bug

Basic sveltekit app with typescript and palywrite enabled is not working for me.

### Reproduction

Setting up a basic sveltekit app:
```bash
npm init svelte@next my-app
cd my-app
npm install
npm run dev -- --open
```

With all options enabled, I.e. typescript and playwrite enabled causes the tests not to work.

### Logs

```shell
$ npm run test

> my-app@0.0.1 test
> playwright test


Running 0 test using 0 worker


Error: tests\test.ts: Cannot import a typescript file from an esmodule.



=================
 no tests found.
=================

```

### System Info

```shell
I am on windows machine, I am not sure if this issues happens on other platforms but the error does not seem platform dependent


$ npx envinfo --system --binaries --browsers --npmPackages "{svelte,@sveltejs/*,vite}"
Need to install the following packages:
  envinfo
Ok to proceed? (y) y

  System:
    OS: Windows 10 10.0.19041
    CPU: (12) x64 Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz  
    Memory: 17.81 GB / 31.71 GB
  Binaries:
    Node: 16.14.2 - C:\Program Files\nodejs\node.EXE        
    npm: 8.5.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (99.0.1150.55)
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.34 
    @sveltejs/kit: next => 1.0.0-next.304 
    svelte: ^3.46.0 => 3.46.6 
```

### Severity

serious, but I can work around it

### Additional Information

To work around this, and probably the way to solve the issue has 2 steps:

playwrite has an alpha release that fixes this issue, see: https://github.com/microsoft/playwright/issues/11755#issuecomment-1082544994

Steps for me were:
```bash
npm up @playwright/test@next
npm exec playwright install
```

And then there is also an issue where the default playwrite file `playwright.config.js` is made in javascript, and needs to be build as a typescript file if typescript is enabled ` playwright.config.ts`. So my workaround is just to change the extension of the filename to `.ts`

DoisKoh:
I'm having the exact same issue. Switching to 1.20.0 of Playwright also circumvents this.

JackDra:
[...]

PR #4479, opened 2022-04-01
FIX: Separate playwrite templates for js vs ts

JackDra:
contributes to #4478

As noted on #4478. We still need to wait for playwright to release the current `@playwright/test@next` version.

### Summary
Playwright file when building with typescript requires the `playwright.config` file to have extension `.ts`.
This PR adds replaces the single `+playwright` folder with 2 folders `+playwright+typescript` and `+playwright-typescript`, where the `+typescript` file has its `playwright.config` file with extension `.ts`.

### Tests

I have manually tested by doing:
```bash
cd packages/create-svelte
node bin.js
```
Then following the prompts to make a app.
- Selecting playwright with no typescript produces `playwright.config.js` in the created app directory
- Selecting playwright with typescript produces `playwright.config.ts` in the created app directory

I also ran the actual test suite in each of the created folders:
```bash
npm install
npm run test
```
For the typescript + playwright, I had to do the extra step get the next version of playwright:
```bash
npm install @playwright/test@next
npm install
npm run test
```

There are no tests in the `packages/create-svelte` currently. 
I'm happy to look into adding tests to ensure all the required files are created depending on the users input?...

### Changesets


Rich-Harris:
Thanks. If we're going to do this, we may as well just automatically install the `next` version for TypeScript users. I'll open an issue reminding us to remove that once 1.21 is released

illright:
I don't think this is resolved. When I create a new SvelteKit project (`pnpm create svelte@next`, Skeleton + TS + No ESLint + No Prettier + Playwright), it still generates a `playwright.config.js` instead of `.ts`, which doesn't work. Also I had to up the Playwright version manually.

Are these changes not released yet, perhaps?

JackDra:
> Are these changes not released yet, perhaps?

Yeah, the latest release of "create-svelte" was:
https://github.com/sveltejs/kit/releases/tag/create-svelte%402.0.0-next.127
which was 20 days ago, this PR was merged 9 days ago

PR #4495, opened 2022-04-02
[Fix] Disable type checking for non-typescript projects by default

### LINKED ISSUE #4427: [create-svelte] Non TS demo app has type errors for the enhance action
### Describe the bug

Without Typescript declarations marking them as optional, the type of the parameters to the action is inferred as `{ pending: any; error: any; result: any; }` with all three fields required causing errors everywhere `enhance` is used in todos/index.svelte.

### Reproduction

Default create-svelte template without Typescript

### Logs

_No response_

### System Info

```shell
.
```

### Severity

annoyance

### Additional Information

https://discord.com/channels/457912077277855764/956030069229359154

### LINKED ISSUE #4428: extends... in jsconfig causes warnings in editors
### Describe the bug

VSCode gives type warnings in vanilla js

Removing the following from `jsconfig.json` fixes this: 

```
``` 

This shouldn't happen if you are not using typescript.

### Reproduction

Create a brand new SvelteKit project, skeleton, no typescript.
Add following to `index.svelte` ```html
<script>
    import { session } from '$app/stores'
</script>

{$session.lang}
```

VSCode will now mark `lang` with a warning/error saying `Property 'lang' lang does not exist on type 'Session'`


### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (12) x64 Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz
    Memory: 13.28 GB / 31.71 GB
  Binaries:
    Node: 16.14.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.3.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (99.0.1150.46)
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.33
    @sveltejs/kit: next => 1.0.0-next.302
    svelte: ^3.44.0 => 3.46.4
```

### Severity

annoyance

### Additional Information

_No response_

Theo-Steiner:

Ever since #4118 .js files are type checked for newly created projects despite ``? Use TypeScript? › No`` being selected during the create-svelte wizard.
This has been described in #4428 and also causes unwanted error messages as #4427 mentions.

I think type checking for non-typescript projects really should be disabled by default as quite a few newcomers on the discord ([today](https://discord.com/channels/457912077277855764/959745024881492019), [yesterday](https://discord.com/channels/457912077277855764/959361729047978035)) are confused by type errors popping up, even though they didn't select typescript in the CLI. 
This behavior seems to surprise even experienced developers: just yesterday Vercel's Lee Robinson stumbled over this in his stream on SvelteKit.
The red squiggly lines on a newly scaffolded out default app also are a bad look IMO.

Maybe we could put a section into the docs for how to re-enable this behavior though? Since it definitely can be helpful when desired.

Closes #4428, closes #4427.

### Tests

### Changesets


dummdidumm (commented):
Agree. AFAIK this was changed by accident when the generated tsconfig was introduced, before that the starter template didn't have type checking for JS.
I'm thinking about whether or not we should turn this around and instead have the generated tsconfig contain less (no checkJs for example) and add that to the tsconfig instead when generating a TS project.

Rich-Harris:
I almost wonder if we need to have two separate prompts when setting up a project, where if you don't opt in to TypeScript-the-language you can still opt in to typechecking: 

```
? Use TypeScript? › Yes
```

```
? Use TypeScript? › No
? Add typechecking for JavaScript? › No / Yes
```

Personally I would choose No and Yes, but I don't know if I'm in a sufficiently small minority that that would be annoying for most non-TypeScript users.

gtm-nayan:
> Personally I would choose No and Yes
Users who choose to do that with the todos demo would still get the type errors I mentioned in #4427. I've been suggesting people to add a `form.d.ts` with the action's types when they complain about this over on Discord. I tried to give it a shot to conditionally include the file but I'm failing to understand the +- thing.

> that would be annoying for most non-TypeScript users
That shouldn't be a problem with [dynamic prompts]. (https://github.com/terkelg/prompts#dynamic-prompts), could also consider making it a single selection between, `Typescript`, `Vanilla JS`, and `JS with some type checking`.

Theo-Steiner:
> Personally I would choose No and Yes, but I don't know if I'm in a sufficiently small minority that that would be annoying for most non-TypeScript users.

I would think that JSdoc users are enough of a minority to justify **not** introducing an additional prompt for enabling JSdoc. Especially given that the overhead for manually turning it on is literally just flipping one boolean.

Also the additional configuration necessary mentioned by @gtm-nayan could mean adding an additional maintenance burden to create-svelte.

mrkishi:
`checkJs` isn't aimed exclusively at JSDoc users, though. Personally I unconditionally enable `checkJs` on all projects I browse as I consider the type-checking analysis useful—even if at times I come across misplaced errors.

gtm-nayan's suggestion of a single selection between TS, JS and JS with type-checking has my vote.

Theo-Steiner:
I tried my hand at modifying create-svelte to allow for the three way selection favored by gtm-nayan and mrkishi.
Would love to hear your feedback!
And I also got stuck at the same problem as gtm-nayan, not knowing how to type the enhance action in the checkJs case.

gtm-nayan:
Something like [this](https://github.com/sveltejs/kit/commit/05867e0cc8435d9566215b695879e01ec54fb72d) maybe?

Still stuck on the `ts` -> `dts` part tho

Rich-Harris:
Another idea is to add JSDoc annotations to the source code, and only strip them out if the user chooses `typescript`. I'd argue this is better than having `.d.ts` files. I opened a new PR for that: #4621

Rich-Harris:
Going to close this in favour of #4621 — thanks

PR #4503, opened 2022-04-04
[Docs] Search - only include type/migration docs when explicitly searched for

### LINKED ISSUE #3865: Search results from Migrating from Sapper litter the search engine.
### Describe the problem

A search in the docs turns up results with Migrating from Sapper.
What are the chances that you're looking for this? Practically none.
This is annoying, results from Migrating from Sapper litter the search engine.

Move all results with Migrating from Sapper to the bottom of the search engine.
Or maybe even separate, specially marked.

### Alternatives considered

Remove from search results and separate Migrating from Sapper from the rest of the documentation.

### Importance

would make my life easier

### Additional Information

_No response_

Theo-Steiner:
Closes #3865

When searching the SvelteKit docs, currently both the type docs and the migration docs tend to get into the way. 
This PR makes it so, that the search is filtered to only include routes containing "type" or "migrat" (stem of migrate/migration/migrating) if the query itself also contains those terms. 

I initially wanted to address this by leveraging [the boost option provided by flexsearch](https://github.com/nextapps-de/flexsearch#index-options) to give type/migration related search results a lower priority, but this option cannot yet be enabled in combination with {tokenizer: "forward"}.

PS: I also removed an unused variable from the codebase in the process.

### Tests

### Changesets

### DIFF
--- a/sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
+++ b/sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
@@ -9,7 +9,6 @@
 	let modal;
 
 	let results = [];
-	let backspace_pressed;
 
 	let index;
 	let lookup;
--- a/sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
+++ b/sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
@@ -64,7 +64,13 @@
 	}
 
 	function update() {
-		results = (index ? index.search($query) : []).map((href) => lookup.get(href));
+		const filter = /type|migrat/;
+		let searchResults = index ? index.search($query.toLowerCase()) : [];
+		// only include type and migration docs if user is explicitly searching for them
+		if (!filter.test($query)) {
+			searchResults = searchResults.filter((href) => !filter.test(href));
+		}
+		results = searchResults.map((href) => lookup.get(href));
 	}
 
 	function escape(text) {


sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
```diff
@@ -65,7 +64,13 @@
...
 
 	function update() {
-		results = (index ? index.search($query) : []).map((href) => lookup.get(href));
+		const filter = /type|migrat/;
+		let searchResults = index ? index.search($query) : [];
+		// only include type and migration docs if user is explicitly searching for them
+		if (!filter.test($query.toLowerCase())) {
+			searchResults = searchResults.filter((href) => !filter.test(href));
+		}
+		results = searchResults.map((href) => lookup.get(href));
```
gtm-nayan, line 73:
```suggestion
		const requestedTypesOrMigrating = filter.test($query.toLowerCase())
		results = searchResults
			.map((href) => lookup.get(href))
			.filter(result => requestedTypesOrMigrating || !filter.test(result.breadcrumbs[0].toLowerCase()));
```
I might've messed up the boolean algebra (blame github's editor), but this should have fewer false negatives if I didn't

Theo-Steiner: Thanks! That definitely is a better approach 🥰 
But there's actually a bug in this: breadcrumbs might be of length 0 because it is only a sliced version of the original breadcrumbs.
I've also mashed everything into one big reduce to not have to iterate over the array twice

Rich-Harris:
Thank you but I'm not at all sure this is desirable. This is all that shows up if you search for `migrat`:

image

There'd be no way to search for the replacement for `preload`, for example.

And if I start typing `requestha`, there's a pretty good chance I'm looking for https://kit.svelte.dev/docs/types#sveltejs-kit-requesthandler — it would be deeply weird (and hard to discover) if I instead needed to search for `type requesthandler` or `requesthandler type`.

I could maybe be persuaded that #3865 is a big enough issue that we should implement some sort of prioritisation, so that migration docs go at the bottom if you search for something like `stores`...

image

...but this isn't the solution, I'm afraid.

PR #4509, opened 2022-04-04
Upgrade to Playwright 1.20.2

benmccann:


Rich-Harris (approved):

Conduitry:
This perhaps should have had a changeset, if it's updating one of the create-svelte templates.

Rich-Harris:
It won't actually change anything, since 1.20.2 is covered by ^1.20.0 (and the dev dependency in `create-svelte` itself is only for our benefit, to remove red squigglies)

PR #4510, opened 2022-04-04
dev server hints if `paths.base` prefix is missing; `--open` uses `paths.base`

### LINKED ISSUE #4418: Redirect to base by default
### Describe the problem

When `kit.paths.base` is defined, by default the app opens to a "Not found" root route. The developer has to manually change the route to get to the desired location.

It would be helpful if dev server redirected to base route by default.

### Alternatives considered

To solve the issue I created a vite plugin - [vite-plugin-redirect](https://github.com/ebeloded/vite-plugin-redirect), which works like this:
```js
import redirect from 'vite-plugin-redirect';

const base = '/admin';

const config = {
  kit: {
    paths: {
      base
    },
    vite: {
      plugins: [redirect({ '/': base })]
    }
  }
};
```

Unless there is a reason to not redirect to base, I think this functionality should be built-in.

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
closes #4418 — if `path.base` is `/blah` and the developer visits `/foo` in dev, they are redirected to `/blah/foo`

### Tests

### Changesets

### DIFF
--- a/.changeset/lovely-hairs-applaud.md
+++ b/.changeset/lovely-hairs-applaud.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Add missing paths.base prefix in dev, if necessary
--- a/.changeset/lovely-hairs-applaud.md
+++ b/.changeset/lovely-hairs-applaud.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Add missing paths.base prefix in dev, if necessary
+Hint if `paths.base` is missing in dev
--- a/.changeset/new-rules-vanish.md
+++ b/.changeset/new-rules-vanish.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/kit": patch
+---
+
+Respect `paths.base` when using `--open`
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -203,7 +203,14 @@ export async function create_plugin(config, cwd) {
 
 						if (req.url === '/favicon.ico') return not_found(res);
 
-						if (!decoded.startsWith(config.kit.paths.base)) return not_found(res);
+						if (!decoded.startsWith(config.kit.paths.base)) {
+							res.writeHead(307, {
+								location: config.kit.paths.base + req.url
+							});
+							res.end();
+
+							return;
+						}
 
 						/** @type {Partial<import('types').Hooks>} */
 						const user_hooks = resolve_entry(config.kit.files.hooks)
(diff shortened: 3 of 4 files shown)


Conduitry:
Is this a good idea? It feels like it's asking for trouble, although I'm admittedly not clear on what that trouble would look like exactly.

In https://github.com/sveltejs/kit/issues/4418#issuecomment-1075481738 I suggest another way of handling this.

mrkishi:
Yeah, it does sound like arbitrary `/foo` redirects could mask broken links and fetches that wouldn't work in production.

Rich-Harris:
Masking broken links is a good reason not to do it. Updated the PR

Conduitry:
:+1: The changeset message and the PR title should be updated as well.

.changeset/lovely-hairs-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Add missing paths.base prefix in dev, if necessary
```
Rich-Harris, line 5:
```suggestion
Hint if `paths.base` is missing in dev
```

.changeset/new-rules-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sveltejs/kit": patch
```
Conduitry, line 2: Is it going to matter if these are double quotes instead of single quotes?

Rich-Harris: this was created via the link in the comment — https://github.com/sveltejs/kit/pull/4510#issuecomment-1087947182 — so i certainly hope not

Conduitry: Huh, weird that the CLI does something different than the bot. Anyway, if this wasn't the result of you manually writing the file, I'm not worried.

Conduitry (approved):

packages/kit/src/core/dev/plugin.js
```diff
@@ -203,7 +204,13 @@ export async function create_plugin(config, cwd) {
 
 						if (req.url === '/favicon.ico') return not_found(res);
 
-						if (!decoded.startsWith(config.kit.paths.base)) return not_found(res);
+						if (!decoded.startsWith(config.kit.paths.base)) {
+							const suggestion = normalize_path(
+								config.kit.paths.base + req.url,
+								config.kit.trailingSlash
+							);
+							return not_found(res, `Not found (did you mean ${suggestion}?)`);
```
mrkishi, line 212: Perhaps?
```suggestion
							return not_found(res, `Not found (did you mean <a href="${suggestion}">${suggestion}</a>?)`);
```

Conduitry: This error message gets returned in a plain text response (with no `content-type`). I don't think it's worth it to make it an HTML response and then worry about escaping the suggestion appropriately.

Rich-Harris: yeah, that's the journey i went on too

PR #4516, opened 2022-04-04
opt into `strict: true` in `tsconfig.json` by default

### LINKED ISSUE #4395: Make strict mode the default when choosing Typescript
### Describe the problem

From a recent conversation I learned that nowadays almost every Typescript starter template of other frameworks (example: the vite templates) and the typescript cli itself default to turning on strict mode. This issue is intended to discuss turning strict mode on for create-svelte, too, and after that I'm happy to open PRs for the vite/rollup template doing the same.

Arguments in favor of strict mode:
- way easier to have it turned on from the start than turning it on later and converting an existing code base
- the TS ecosystem as a whole recommends turning it on by default
- the strict null checks are very nice for catching nullpointer errors early
- if it's off by default, people might be surprised about nullpointers not getting caught because they thought TS should catch these things, too

Arguments against it:
- Potentially steeper learning curve for those new to it

Turn on strict mode in the starter templates 

### Alternatives considered

Only turn on a subset of it, for example only the strict null checks option which is arguably the most important one

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
closes #4395. This is a breaking change

### Tests

### Changesets

### DIFF
--- a/.changeset/pretty-oranges-impress.md
+++ b/.changeset/pretty-oranges-impress.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] Opt in to TypeScript strict: true by default
--- a/packages/kit/src/core/sync/write_tsconfig.js
+++ b/packages/kit/src/core/sync/write_tsconfig.js
@@ -63,7 +63,8 @@ export function write_tsconfig(config) {
 						$lib: [project_relative(config.kit.files.lib)],
 						'$lib/*': [project_relative(config.kit.files.lib + '/*')]
 					},
-					rootDirs: [config_relative('.'), './types']
+					rootDirs: [config_relative('.'), './types'],
+					strict: true
 				},
 				include,
 				exclude: [config_relative('node_modules/**'), './**']


dummdidumm:
This is closely related to #4495, we should at least release them together to give non-TS users a way to opt out.

The alternative would be to not add `strict` to the generated tsconfig, and remove `checkJs` as well. This means we would need to adjust #4495 to add `strict` accordingly.

Rich-Harris:
Thinking further on this, it feels like `strict` probably belongs in the user-editable `tsconfig`/`jsconfig`, rather than the one we extend from. There's a few other compiler options I'd put in the same category

dummdidumm:
Agree. I'd move the following to the user config:
skipLibCheck, esModuleInterop (honestly not even sure if it works for Svelte files since TS can't see the other files during transpiration)

The following I'm not sure about:
resolveJsonModule (if user doesn't have a `declare module "*.json"` somewhere they would get errors)

Rich-Harris:
I don't think Kit or the default apps rely on `resolveJsonModule` though? Feels like that belongs to user config too

Rich-Harris:
Closing this in favour of #4633, which documents the generated tsconfig and moves non-essential stuff into user config, where it can be changed freely

PR #4517, opened 2022-04-05
[chore] clean up playwright-chromium root dependency

Conduitry:
There was an unneeded (and outdated) version of `playwright-chromium` in the root `package.json` of this repo. This removes it.

### Tests

### Changesets


Conduitry:
A bunch of tests are failing. What do they do that I wouldn't be testing with a `pnpm test` from the root of this repo?

mrkishi:
It seems like that's due to `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD`. Tests probably assume the browser is in `PATH`, in this case through `node_modules`?

Conduitry:
Oh, I see. Does it make sense then to have this _only_ at the root level (and not in `adapter-static`) - and to also bump it to the same version as is being used for `@playwright/test` elsewhere?

mrkishi:
My mistake—`PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD` is `1` throughout the workflow _except_ when `pnpm install`ing, and I see  `playwright` did download the browsers. I don't know why it couldn't launch them, then.

@pngwn since you got caching working in the first place, do you have an opinion if it's worth migrating to `playwright install --with-deps chromium` these days?

Rich-Harris:
what's the status of this?

Conduitry:
The (probably) most pressing part was to stop having two different versions of playwright-chromium in the project, which was since addressed in another PR.

I've merged master back in and resolved the conflicts. CI is running now. If that fails, I'd be in favor of just closing this without spending more time investigating why that's happening when the tests work locally.

Rich-Harris (approved):

PR #4520, opened 2022-04-05
reword `adapter-node`s header security warning

### LINKED ISSUE #3391: [docs] improve explanation around trusting forwarding headers
### Describe the bug

https://kit.svelte.dev/docs#configuration-headers says:

> You should only do this if you trust the reverse proxy, which is why it isn't the default.

Is that an accurate description of the issue? If so, maybe we could link to a resource on it. I think though the issue isn't typically that the reverse proxy would be malicious, but that the client could be. And I think the condition under which "you should only do this if" would be a different condition since you can never trust the client.

Perhaps it should say something like:

> Clients may send malicious values for headers, so you should only do this if allowing the client to set an arbitrary value does not pose a security risk to your application

### Reproduction

https://kit.svelte.dev/docs#configuration-headers

### Logs

_No response_

### System Info

```shell
n/a
```

### Severity

annoyance

### Additional Information

_No response_

mrkishi:
Closes #3391. Clarifies the security implication of setting `adapter-node`'s `_HEADER` options.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-node/README.md
+++ b/packages/adapter-node/README.md
@@ -41,7 +41,7 @@ With this, a request for the `/stuff` pathname will correctly resolve to `https:
 PROTOCOL_HEADER=x-forwarded-proto HOST_HEADER=x-forwarded-host node build
 ```
 
-> [`x-forwarded-proto`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) and [`x-forwarded-host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) are de facto standard headers that forward the original protocol and host if you're using a reverse proxy (think load balancers and CDNs). You should only set these variables if you trust the reverse proxy.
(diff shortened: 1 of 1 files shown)


packages/adapter-node/README.md
```diff
@@ -41,7 +41,7 @@ With this, a request for the `/stuff` pathname will correctly resolve to `https:
 PROTOCOL_HEADER=x-forwarded-proto HOST_HEADER=x-forwarded-host node build
 ```
 
-> [`x-forwarded-proto`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) and [`x-forwarded-host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) are de facto standard headers that forward the original protocol and host if you're using a reverse proxy (think load balancers and CDNs). You should only set these variables if you trust the reverse proxy.
+> [`x-forwarded-proto`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) and [`x-forwarded-host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) are de facto standard headers that forward the original protocol and host if you're using a reverse proxy (think load balancers and CDNs). You should only set these variables if your server is behind a trusted reverse proxy, otherwise it'd be possible for clients to spoof these headers.
```
Conduitry, line 44:
```suggestion
> [`x-forwarded-proto`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) and [`x-forwarded-host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) are de facto standard headers that forward the original protocol and host if you're using a reverse proxy (think load balancers and CDNs). You should only set these variables if your server is behind a trusted reverse proxy; otherwise, it'd be possible for clients to spoof these headers.
```

I think?

Rich-Harris (approved):

PR #4529, opened 2022-04-05
document how to render nested error pages for 404s

Rich-Harris:
people struggle with this, we've had a bunch of confusing issues raised around it


dreitzner (approved):
That is a great explanation and clarifies the issue for me

benmccann (approved):

PR #4532, opened 2022-04-05
[fix] type check exception handling on form action

avarun42:

### Tests

### Changesets

`form.ts` in the default demo app currently uses an `any` type for catching an error thrown during the request. Switch to an `unknown` type and explicitly check for the correct type before passing to the `error` function which expects an argument of type `Error`.


Rich-Harris (approved):

Rich-Harris:
thanks!

Issue #1634, opened 2021-06-03
Assets path wrong when importing from nested routes


rohanharikr:
**Describe the bug**

My project structure is as follows:
```
routes
│   login.svelte
└───wizard
│   │   email.svelte
```

When adding an img tag that uses an image from the `static` folder
`<img src="logo.svg"/>`

This works correctly for the `login` route but not for `/wizard/email` route.

A fix I found for this was to prefix `/static/` to all images used,
for example: `<img src="/static/logo.svg"/>`

Just wondering why its not working with just `logo.svg`

Note: I am in dev enviroment

goncalolobo3:
Where is your static folder? Depending on where it is, and what config you have in your project, you might be able to change the file path. 
In the meantime, use "/static/" when using images.

Rich-Harris:
You shouldn't use `/static/` as a prefix, you should just just `/`. It's a bug that the `/static/` prefix works (to do with how Vite handles requests — fixed in #4535).

This is expected behaviour — relative URLs are relative to the current page. You need to use absolute or root-relative URLs, either like `/logo.svg` or `${assets}/logo.svg` (using https://kit.svelte.dev/docs/modules#$app-paths-assets)

PR #4535, opened 2022-04-05
remove Vite static asset middleware

Rich-Harris:
Noticed in #1634 (and I can't believe this hasn't come up sooner) that Vite will serve assets in the `static` folder, with that prefix — so `/favicon.png` can also be requested as `/static/favicon.png`. 

Fixed by removing `viteServeStaticMiddleware` in the plugin (which, aside: this feels like a _massive_ code smell and an indication that the design of the Vite dev server isn't optimal for the SSR framework use case).

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -378,7 +378,8 @@ function remove_html_middlewares(server) {
 	const html_middlewares = [
 		'viteIndexHtmlMiddleware',
 		'vite404Middleware',
-		'viteSpaFallbackMiddleware'
+		'viteSpaFallbackMiddleware',
+		'viteServeStaticMiddleware'
 	];
 	for (let i = server.stack.length - 1; i > 0; i--) {
 		// @ts-expect-error using internals until https://github.com/vitejs/vite/pull/4640 is merged
--- a/packages/kit/test/apps/basics/test/test.js
+++ b/packages/kit/test/apps/basics/test/test.js
@@ -2232,6 +2232,11 @@ test.describe.parallel('Static files', () => {
 		response = await request.get('/favicon.ico');
 		expect(response.status()).toBe(200);
 	});
+
+	test('does not use Vite middleware', async ({ request }) => {
+		const response = await request.get('/static/static.json');
+		expect(response.status()).toBe(404);
+	});
 });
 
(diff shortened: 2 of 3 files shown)


packages/kit/test/apps/basics/test/test.js
```diff
@@ -2232,6 +2232,11 @@ test.describe.parallel('Static files', () => {
 		response = await request.get('/favicon.ico');
 		expect(response.status()).toBe(200);
 	});
+
+	test('does not use Vite middleware', async ({ request }) => {
```
PH4NTOMiki, line 2236:
```suggestion
	test('does not use Vite static middleware', async ({ request }) => {
```

packages/kit/src/core/dev/plugin.js
```diff
@@ -378,7 +378,8 @@ function remove_html_middlewares(server) {
 	const html_middlewares = [
 		'viteIndexHtmlMiddleware',
 		'vite404Middleware',
-		'viteSpaFallbackMiddleware'
+		'viteSpaFallbackMiddleware',
+		'viteServeStaticMiddleware'
```
benmccann, line 382:
```suggestionx
		// https://github.com/vitejs/vite/issues/7363
		'viteServeStaticMiddleware'
```

Conduitry:
I'm not completely clear on what this new approach is doing, but the changeset and the test description should probably be updated.

packages/kit/src/core/dev/plugin.js
Rich-Harris: resolving as outdated

Rich-Harris:
updated those descriptions

bluwy (approved):

benmccann:
Could you explain why this was changed? The last iteration made sense to me whereas this iteration does not

Rich-Harris:
The static middleware is needed to serve things like image files in `src`

istarkov:
Seems like this breaks [assets imported as urls](https://vitejs.dev/guide/assets.html#importing-asset-as-url) used in development mode, i.e before
```
import myimage from '../../static/myimage.png?url';
...
<img src={myimage} />
```
worked in dev mode under `/static/myimage.png` path, now its not. (no issues with prod)

PR #4536, opened 2022-04-06
[feat] custom load dependencies

### LINKED ISSUE #1684: (re)trigger the load function

icalvin102:
Closes #1684 

### Problem
The `invalidate` function only re-runs `load` if fetch is used. This is not possible in cases where an `APIClient` does not allow for customization of the fetching process or when using non REST APIs like GraphQL.

### Solution
Add a `dependencies` array to the `LoadOutput` (suggested by @Rich-Harris in https://github.com/sveltejs/kit/pull/1709).
The dependencies will be registered the in the same way `fetch` does it. This allows not only to `invalidate` real resources but entire pages or groups of pages.

**Example Usage**

```svelte
<script context="module">
import { api } from '$lib/myapi';
export async function load({params}){
    const item = await api.getItem(params.id);
    return { 
        props: { item },
        dependencies: [`/item-${params.id}`, 'custom:item-page']
    }
}
</script>

<script>
import Item from '$lib/Item.svelte';
export let item;
</script>

<Item {item} />
<button on:click={()=> invalidate('custom:item-page') }>Reload Page</button>
<button on:click={()=> invalidate(`/item-${item.id}`) }>Reload Item</button>
```

### Tests

### Changesets


dreitzner:
@icalvin102 just a quick question.... 
Is this problem not solved by [page endpoints](https://kit.svelte.dev/docs/routing#endpoints-page-endpoints)?
To the best of my understanding this should generate a load function on build.
By moving the `$lib/myapi` integration to the endpoint there should be no trouble on calling the load again ;)

icalvin102:
@dreitzner page endpoints will only work if your API is written in sveltekit AFAIK.
Page endpoints wont work in SPA mode and/or when using third party APIs that rely on browser state like cookies that are not accessible from the sveltekit server.

documentation/docs/04-loading.md
```diff
@@ -134,3 +134,11 @@ If the `load` function returns a `props` object, the props will be passed to the
...
 
 The combined `stuff` is available to components using the [page store](/docs/modules#$app-stores) as `$page.stuff`, providing a mechanism for pages to pass data 'upward' to layouts.
+
+#### dependencies
+
+If the `load` function depends on custom resources a `dependencies` array can returned. A registered dependency can later be used as argument of `invalidate` to rerun the `load` function.
+
+> Note that the `fetch` function alreay registers the requested resources as dependencies. Manual registration is only necessary if it's not possible to use SvelteKits `fetch` function. 
+
+The items should be [properly encoded](https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding) URIs. Both absolute and relative URIs are acceptable.
```
Rich-Harris, line 144:
```suggestion
An array of strings representing URLs the page depends on, which can subsequently be used with [`invalidate`](/docs/modules#$app-navigation-invalidate) to cause `load` to rerun. You only need to add them to `dependencies` if you're using a custom API client; URLs loaded with the provided `fetch` function are added automatically.

URLs can be absolute or relative to the page being loaded, and must be [encoded](https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding).
```

Rich-Harris (approved):

Rich-Harris:
thank you!

PR #4538, opened 2022-04-06
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/packages/kit/.eslintrc.json
+++ b/packages/kit/.eslintrc.json
@@ -13,6 +13,10 @@
 		"afterNavigate": true
 	},
 	"rules": {
+		"no-restricted-properties": [
+			"error",
+			{ "property": "substr", "message": "Use .slice instead of .substr." }
+		],
 		"@typescript-eslint/no-empty-interface": "off"
 	}
 }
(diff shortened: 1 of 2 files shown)


benmccann:
Should we use `substring` rather than `slice`? I always have trouble remembering what the `slice` API is since it's so generically named whereas `substring` exists in Java as well, so I end up using it a lot more often

CommanderRoot:
`slice` and `substring` are very similar but [differ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#differences_between_substring_and_slice) in some regards.
`slice` is in my opinion a bit stricter in how it handles the input variables and it's also generally faster (probably overall irrelevant)

Conduitry:
🤷 I personally always use `slice` because it's the same API on strings and on arrays, so I'd be in favor of this PR as-is.

Rich-Harris (approved):

Rich-Harris:
thanks 👍

PR #4522, opened 2022-04-05
[docs] Search - deprioritize type/migration and include category

Theo-Steiner:
This is attempt number two to improve searching the docs. 
Sorry for keeping y'all from working on more important PRs 🙇‍♂️

After a very clumsy solution in #4503, I think I now found a way to deprioritize type/migration related search results that as a side effect also genuinely makes the search better as well!

To recap the problem: when searching the documentation, type and migration related search results clutter the view by appearing at the very top, basically doubling the search results the user has to scan to find what they're looking for.

I embarrassingly misunderstood the flexsearch docs and thought the ``boost()`` function that allows influencing the ranking  was not available for ``tokenizer: "forward"`` but this limitation is only about how ``boost`` works internally.

For the ranking manipulation to work, I included the "category" (the very first breadcrumb) into the content for each entry and returned a boost of 0.5 (lower than 1 deprioritizes) for entries with relevant categories.
This works like a charm and has the nice side effect that when searching for let's say "faq" now all pages inside the faq section appear as search results.

Closes #3865
### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [X] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [X] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [X] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [ ] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #4541, opened 2022-04-06
Deprioritise migration guide

Rich-Harris:
Supersedes #4522. Adds a declarative method for ranking doc sections, since `boost` is unreliable. Only applies to the migration guide, _not_ the types, since it's quite likely people are searching for those.


benmccann:
FYI @Theo-Steiner - thanks for kicking off this investigation!

benmccann (approved):

PR #4542, opened 2022-04-06
Demand better repros

Rich-Harris:
People keep trying to get away with shit repros. If we update the text to explain what constitutes a good repro, that will probably still happen, but we can at least point to this as we close the issue


Conduitry (approved):
Absolutely in favor of this.

PR #4545, opened 2022-04-06
skip client-side nav for <a sveltekit:reload>

### LINKED ISSUE #546: sveltekit:refresh
At present you can force an `<a>` tag to trigger a full page reload on click by adding either `target="_self"` or `rel="external"`.

Neither is particularly explicit. They cause a reload because &lt;handwave&gt;that's how SPA routers work&lt;/handwave&gt;, but in the name of clearly communicating intent and having a useful thing to add to the docs, it might be good to add a `sveltekit:refresh` attribute alongside `sveltekit:prefetch` and `sveltekit:noscroll`. 

It's duplicative, but maybe that's not the worst thing in the world?

Rich-Harris:
closes #546, though there's still room for bikeshedding.

One issue here is that the `sveltekit:reload` attribute gets squigglied: 

image

I actually have no idea where `SvelteKitAnchorProps` (which determines which attributes are valid) is defined?

### Tests

### Changesets

### DIFF
--- a/.changeset/mighty-cows-notice.md
+++ b/.changeset/mighty-cows-notice.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Skip client-side navigation for links with sveltekit:reload
--- a/.changeset/orange-wolves-fly.md
+++ b/.changeset/orange-wolves-fly.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] Skip prerendering for rel="external" links
--- a/documentation/docs/07-a-options.md
+++ b/documentation/docs/07-a-options.md
@@ -20,6 +20,22 @@ Note that prefetching will not work if the [`router`](/docs/page-options#router)
 
 You can also programmatically invoke `prefetch` from `$app/navigation`.
 
+### sveltekit:reload
+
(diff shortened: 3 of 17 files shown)


Rich-Harris:
in fact, the type error mentioned above is blocking CI. cc @dummdidumm, who likely knows where and how we'd fix this

dummdidumm:
That needs a change in language tools and new release of `svelte-check` - will do that tomorrow

benmccann:
`reload` sounds to me like we're staying on the same page similar to @Conduitry's objection to `refresh` in the issue https://github.com/sveltejs/kit/issues/546#issuecomment-799850700. I wonder if there's a name that better indicates it's a new page. Perhaps `sveltekit:fullnav` or `sveltekit:servernav` or something?

Rich-Harris:
In the context of a link that points to a specific location, I think `reload` is clear enough. Better to use vernacular that's in widespread use than to come up with something that sounds clunky, I think

dummdidumm:
New version of `svelte-check` which supports the `sveltekit:reload` attribute released.

mrkishi:
I just realized we could now ignore `rel="external"` links when prerendering since those will be dedicated for truly external links. `sveltekit:reload` links should still be crawled.

benmccann (approved):
> In the context of a link that points to a specific location, I think reload is clear enough. Better to use vernacular that's in widespread use than to come up with something that sounds clunky, I think

Okay. I think I'm convinced.

PR #4553, opened 2022-04-07
[chore] bump `eslint` from version 7 to 8

stvnwn:
When a user creates a SvelteKit app with `eslint` support using `create-svelte`, it ships the last `eslint` release under major version 7. With this pull request, `create-svelte` will now ship the latest `eslint` release (for now).

I'm worried that `eslint-plugin-svelte3` doesn't support `eslint@^8.0.0`, from [this discussion](https://github.com/sveltejs/eslint-plugin-svelte3/pull/155#issuecomment-1003056483).

`eslint-plugin-svelte3`'s peerDependencies [lists `eslint@>=6.0.0`](https://github.com/sveltejs/eslint-plugin-svelte3/blob/55fc0bdcb76748edec2b5613113574ba45b1a419/package.json#L28) since June 22, 2019, around when `eslint@6.0.1` was released.

Related: #3261

---

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/shared/+eslint/package.json
+++ b/packages/create-svelte/shared/+eslint/package.json
@@ -1,6 +1,6 @@
 {
 	"devDependencies": {
-		"eslint": "^7.32.0",
+		"eslint": "^8.12.0",
 		"eslint-plugin-svelte3": "^3.2.1"
 	}
 }
--- a/packages/create-svelte/shared/+eslint/package.json
+++ b/packages/create-svelte/shared/+eslint/package.json
@@ -1,6 +1,6 @@
 {
 	"devDependencies": {
 		"eslint": "^8.12.0",
-		"eslint-plugin-svelte3": "^3.2.1"
+		"eslint-plugin-svelte3": "^4.0.0"
 	}
 }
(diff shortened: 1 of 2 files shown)


dummdidumm:
I think we should first release a new major for `eslint-plugin-svelte3` which officially supports ESLint 8, drops NodeJS 10 support etc and then continue here. @Conduitry would you be ok with that?

Conduitry:
That sounds fine to me. Is there any benefit to dropping Node 12 while we're at it, because it's EOL at the end of the month anyway?

benmccann:
The eslint plugin tests currently use an internal eslint class. eslint 8 adds an `exports` map that stops us from using it. I've sent https://github.com/eslint/eslint/pull/15776 to unblock it.

If that doesn't work we can see if we can hack around that by doing something like `require.resolve('eslint/package.json')` and then constructing the path of the module we want to require since we can probably `require` an absolute or relative path, regardless of export maps.

packages/create-svelte/shared/+eslint/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"devDependencies": {
-		"eslint": "^7.32.0",
+		"eslint": "^8.12.0",
 		"eslint-plugin-svelte3": "^3.2.1"
```
benmccann, line 4:
```suggestion
		"eslint-plugin-svelte3": "^4.0.0"
```

Issue #686, opened 2021-03-25
Remove 'Update latest tag on published packages' step


benmccann:
https://www.npmjs.com/package/@sveltejs/kit shows an old version of the README and people continually run into issues installing the `latest` instead of `next` package

Rich-Harris:
is that just a matter of removing `--tag=next` from https://github.com/sveltejs/kit/blob/d1925cfa7a469841c78302b49b3a6d11fc6f4954/package.json#L12 ? we can still stay in `pre` mode as far as `changesets` is concerned, and publish packages with a `-next.n` version suffix?

if so then 👍 this probably makes sense, though presumably we'd need to do some deprecated `@next` releases so that people are pointed towards `latest`? otherwise this would be equally confusing for the people currently using `@next` packages.

benmccann:
@pngwn do you know if we can just remove `--tag=next` or if more is needed?

pngwn:
We need to call changeset publish to get other tags and releases in the repo rather than recursively publishing manually.

Rich-Harris:
but `changeset publish` would publish `next` tags because we're in prerelease mode, no? Will `changeset publish` respect the `"private": true` package.json fields?

pngwn:
> But changeset publish would publish next tags because we're in prerelease mode, no?

You're right, i'm not sure how we add additional tags for specific releases, I'll look into that.

> Will changeset publish respect the "private": true package.json fields?

It will, yes.

benmccann:
`pnpm publish` might not let us add two tags, but we should be able to add a second tag after the package is published

dominikg:
we are using `changeset publish` and it is run by `@changeset/action'` here:  https://github.com/sveltejs/kit/blob/32a7ccfbd1ab5e1928da27149fe67d88d3308f7f/.github/workflows/release.yml#L32
 
This action [emits outputs](https://github.com/changesets/action#outputs) that could be used to generate
`npm dist-tag add package@version latest` calls in a conditional step  

```
- name: Tag published packages as latest too
        if: steps.changesets.outputs.published == 'true'
        # somehow take that json output of steps.changesets.outputs.publishedPackages and turn it into dist-tag commands
        run: # execute commands
```
problems:
1) github actions expressions syntax is not a joy to work with  
2) running `dist-tag` shortly after publishing a version for the first time may fail if the registry hasn't propagated the new version.

Rich-Harris:
[...]

PR #4564, opened 2022-04-07
ci: add step to release script that updates the latest tag 

dominikg:
see https://github.com/sveltejs/kit/issues/686

What this step does.
- put publishedPackages into an env var
- echo and pipe that into jq that maps it into `package@version` strings
- loop over these and call `echo npm dist-tag add package@version latest` for each published package

**The called command is `echo`ed for now to make sure it is correct.** 

After we verified it works we can remove the echo.

### Tests

### Changesets

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,6 +40,15 @@ jobs:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           UPDATE_TEMPLATE_SSH_KEY: ${{ secrets.UPDATE_TEMPLATE_SSH_KEY }}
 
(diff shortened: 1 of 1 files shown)


dominikg:
this step would essentially turn into a noop (set latest to same version it was set to by the publish step before) once we exit pre-release mode but it should be removed then.

dominikg:
`jq` is included in the ubuntu environment https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md

Rich-Harris (approved):

Rich-Harris:
nice! excited to see if this works

PR #4552, opened 2022-04-07
docs(rest): adding url search params example

MateusAndrade:
### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [x] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

## Changes

Hi, thanks for the amazing job on this project. I have been using SvelteKit for a while, but wasn't able to find a proper example on how to access params sent using URL params: `/foo?a=1&b=2`.

This PR adds an example regarding the use of URL search params using [endpoints](https://kit.svelte.dev/docs/routing#endpoints).

PR #4565, opened 2022-04-07
docs(endpoint): adding example regarding parsing url params

MateusAndrade:

### Tests

### Changesets

## Changes

Hi, thanks for the amazing job on this project. I have been using SvelteKit for a while, but wasn't able to find a proper example on how to access params sent using URL params: `/foo?a=1&b=2`.

This PR adds an example regarding the use of URL search params using [endpoints](https://kit.svelte.dev/docs/routing#endpoints).

Note: I sent this fix at #4552 , but it looks like the pull bot ended up removing my changes. 😓

### DIFF
--- a/documentation/docs/01-routing.md
+++ b/documentation/docs/01-routing.md
@@ -241,6 +241,25 @@ export async function post({ request }) {
 }
 ```
 
+#### URL parsing
+
+The `url` object is an instance of the standard [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) class at the `searchParams` key. As such, you can access the params sent using query params.
+
+```js
+// @filename: index.js
+// /foo?limit=10&offset=20
+/** @type {import('@sveltejs/kit').RequestHandler} */
+export async function get({ url }) {
+	const limit = url.searchParams.get("limit");
+	const offset = url.searchParams.get("offset"); 
+
+	const response = await expensiveSearch({ limit, offset });
+	const items = await response.json();
+
+	return { items, status: 201 };
+}
+```
+
 #### Setting cookies
 
 Endpoints can set cookies by returning a `headers` object with `set-cookie`. To set multiple cookies simultaneously, return an array:


Conduitry:
We already mention [here](https://kit.svelte.dev/docs/loading#input-url) that `url.searchParams` is an instance of `URLSearchParams`. _Maybe_ we could also link to the MDN page here from the routing section, but I don't think we want to be in the business of telling people how to use web APIs that are well-documented elsewhere.

MateusAndrade:
> We already mention [here](https://kit.svelte.dev/docs/loading#input-url) that `url.searchParams` is an instance of `URLSearchParams`. _Maybe_ we could also link to the MDN page here from the routing section, but I don't think we want to be in the business of telling people how to use web APIs that are well-documented elsewhere.

Nice point! What about link/explain that the object received at an endpoint inherits this property? I found it confusing to understand how to access this parameters without rendering using the `load`

documentation/docs/01-routing.md
```diff
@@ -241,6 +241,25 @@ export async function post({ request }) {
 }
 ```
 
+#### URL parsing
+
+The `url` object is an instance of the standard [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) class at the `searchParams` key. As such, you can access the params sent using query params.
```
Rich-Harris, line 246: i think we can simplify this wording

```suggestion
The `url` object is a standard [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object. As such, you can access the query via `url.searchParams`:
```

documentation/docs/01-routing.md
```diff
@@ -241,6 +241,25 @@ export async function post({ request }) {
...
+// /foo?limit=10&offset=20
+/** @type {import('@sveltejs/kit').RequestHandler} */
+export async function get({ url }) {
+	const limit = url.searchParams.get("limit");
+	const offset = url.searchParams.get("offset"); 
+
+	const response = await expensiveSearch({ limit, offset });
+	const items = await response.json();
+
+	return { items, status: 201 };
```
Rich-Harris, line 259: this is incorrect (and part of the reason CI is failing) — you need to return a `body` key

```suggestion
	return {
	  status: 201,
	  body: { items }
	};
```

documentation/docs/01-routing.md
```diff
@@ -241,6 +241,25 @@ export async function post({ request }) {
 }
 ```
 
+#### URL parsing
+
+The `url` object is an instance of the standard [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) class at the `searchParams` key. As such, you can access the params sent using query params.
+
+```js
+// @filename: index.js
```
Rich-Harris, line 249: typechecking will fail without this

```suggestion
// @filename: ambient.d.ts
declare global {
  const search: (opts: { limit: string | null, offset: string | null }) => Response;
}

export {};

// @filename: index.js
// ---cut---
```

documentation/docs/01-routing.md
```diff
@@ -241,6 +241,25 @@ export async function post({ request }) {
...
+// /foo?limit=10&offset=20
+/** @type {import('@sveltejs/kit').RequestHandler} */
+export async function get({ url }) {
+	const limit = url.searchParams.get("limit");
+	const offset = url.searchParams.get("offset"); 
+
+	const response = await expensiveSearch({ limit, offset });
```
Rich-Harris, line 256: i think we can make this simpler, the `expensive` draws attention to the wrong thing

```suggestion
	const response = await search({ limit, offset });
```

documentation/docs/01-routing.md
```diff
@@ -241,6 +241,25 @@ export async function post({ request }) {
...
+
+The `url` object is an instance of the standard [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) class at the `searchParams` key. As such, you can access the params sent using query params.
+
+```js
+// @filename: index.js
+// /foo?limit=10&offset=20
+/** @type {import('@sveltejs/kit').RequestHandler} */
+export async function get({ url }) {
+	const limit = url.searchParams.get("limit");
+	const offset = url.searchParams.get("offset");
```
Rich-Harris, line 254:
```suggestion
	const limit = url.searchParams.get('limit');
	const offset = url.searchParams.get('offset'); 
```

Rich-Harris (changes requested):
I'm inclined to agree with @Conduitry that we need to avoid bloating the docs. Perhaps we should instead have a small section on web standards, near the top of the docs, that explains 'you're gonna encounter `fetch` and `URL` and `Request` and `Response` and other things that you might not be overly familiar with, but don't worry, that's because we're using web standards, and time spent learning them (over here on MDN) will be well-rewarded'. Could have a couple of minimal examples there.

In the meantime, I made some suggestions in case we had a change of heart and _did_ want to go down this route

documentation/docs/01-routing.md
MateusAndrade: Nice!!! I made some confusion because I'm using this with Svelte 😅

documentation/docs/01-routing.md
MateusAndrade: Nice! I forgot about that.

MateusAndrade:
> In the meantime, I made some suggestions in case we had a change of heart and _did_ want to go down this route

No problem, I ended up applying these suggestions and changing the status code at the example to 200 since it's a GET.  😄

documentation/docs/01-routing.md
```diff
@@ -241,6 +241,36 @@ export async function post({ request }) {
 }
 ```
 
+#### URL parsing
+
+The `url` object is a standard [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) object. As such, you can access the query via `url.searchParams`:
```
Rich-Harris, line 246: just realised we have 'the `url` object ... URL object' which sounds clunky

```suggestion
The `url` object is an instance of the standard [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) class. As such, you can access the query via `url.searchParams`:
```

Rich-Harris:
https://github.com/sveltejs/kit/pull/4618

Rich-Harris:
merged #4618, so I'll close this — thanks

PR #4573, opened 2022-04-08
ci: release action - add NPM_TOKEN to npmrc 

dominikg:
and remove echo from dist-tag call

### Tests

### Changesets

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,10 +44,14 @@ jobs:
         if: steps.changesets.outputs.published == 'true'
         # wait a bit and then run "npm dist-tag add package@version latest" for all published packages
         env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           published_packages: ${{steps.changesets.outputs.publishedPackages}}
         run: |
           sleep 20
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -44,10 +44,14 @@ jobs:
         if: steps.changesets.outputs.published == 'true'
         # wait a bit and then run "npm dist-tag add package@version latest" for all published packages
         env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           published_packages: ${{steps.changesets.outputs.publishedPackages}}
         run: |
           sleep 20
-          echo $published_packages | jq -r 'map((.name + "@" + .version)) | .[]' | while read pkgver; do echo npm dist-tag add $pkgver latest || continue;done
+          cat << EOF > "$HOME/.npmrc"
```
dominikg, line 51: this was taken from https://github.com/sveltejs/vite-plugin-svelte/blob/a5d36899e2a987830d7ab049add2f8cb01a06797/.github/workflows/release.yml#L37

home/.npmrc was chosen to avoid overriding a project specific key

do we need to clean that up?

Rich-Harris: nah, if it works it works

Rich-Harris (approved):

PR #4588, opened 2022-04-11
apply set-cookie headers from page dependencies

### LINKED ISSUE #1198: Make server version of fetch() in load() copy over 'set-cookie' to the page response

Rich-Harris:
fixes #1198. `set-cookie` headers from endpoints were being handled in a fairly naive way — if the response was read inside `load`, the `set-cookie` headers were merged with existing ones, but otherwise they were ignored.

With this PR, `set-cookie` headers are automatically parsed (not just if the response is read) using [set-cookie-parser](https://www.npmjs.com/package/set-cookie-parser) (which doesn't mind if multiple headers have been concatenated), and kept track of as `load` runs. This means that if two endpoints are `fetch`ed in `load`, cookies from the first response will be applied to the second request, and cookies from both responses are added to the page response.

Very possible that I've overlooked some nuance around domains or somesuch, so would be grateful for a sanity check.

### Tests

### Changesets


packages/adapter-static/package.json
```diff
@@ -29,9 +29,11 @@
 	},
 	"devDependencies": {
 		"@sveltejs/kit": "workspace:*",
+		"cookie": "^0.4.1",
```
benmccann, line 32: https://github.com/sveltejs/kit/pull/4592
```suggestion
		"cookie": "^0.5.0",
```

PR #4599, opened 2022-04-12
Add platform to root relative requests

pzuraq:
Currently `platform` is not passed forward to root relative requests. If request hooks or endpoints rely on the values in `platform`, this will cause them to not work and throw errors. This PR fixes this by adding the platform.

There is no way to test this as far as I can tell currently because there is no way to provide a platform for local development.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -195,6 +195,7 @@ export async function load_node({
 						{
 							getClientAddress: state.getClientAddress,
 							initiator: route,
+							platform: state.platform,
 							prerender: state.prerender
 						}
 					);


benmccann (approved):

.changeset/honest-wasps-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+add `platform` to requests triggered by `fetch` in `load`
```
Conduitry, line 5: Do we want to specify that this is just fixing this for requests that happen during SSR? (_Does_ it just fix it for requests that happen during SSR?) Not worth mentioning?

pzuraq: It does only do this in SSR I believe, since it's about nested fetch requests in the same SSR event. Standard fetch requests already work.

Rich-Harris, line 5:
```suggestion
add `platform` to requests triggered by `fetch` in `load` during SSR
```

Rich-Harris:
thanks!

Issue #4488, opened 2022-04-01
Remove `@playwright/test@next`


Rich-Harris:
### Describe the problem

Ref #4479. In order to run Playwright in TypeScript projects with a `playwright.config.ts` file, we need to use `@playwright/test@next`, hence this file: https://github.com/sveltejs/kit/blob/master/packages/create-svelte/shared/%2Bplaywright%2Btypescript/package.json

Ideally we'd be using a stable version instead.

### Describe the proposed solution

Remove the file once 1.21 is released

### Importance

nice to have

JackDra:
Ref to keep track of playwright 1.21 release: https://github.com/microsoft/playwright/releases

PR #4601, opened 2022-04-12
[chore] upgrade to Playwright 1.21.0

benmccann:
Closes https://github.com/sveltejs/kit/issues/4488


Conduitry (approved):
I'm still not sure what's up with the #4517 stuff, but this change looks good to me. We should upgrade the root-level instance of the package as well until we can figure out why we can't remove it.

PR #4443, opened 2022-03-24
Create fallback HTML file when prerendering is disabled

dimfeld:
Fixes #4441. With `prerendering.enabled = false` the fallback HTML file was not created.

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [X] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [X] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

Wasn't quite sure of the best way to test this. Looks like it would require creating a new sample app in the `prerendering` test directory with this configuration and a single test that the fallback file is written. Should I do that?

### Tests
- [X] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [X] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #4604, opened 2022-04-13
Revert change to render fallback when prerendering is disabled

### LINKED ISSUE #4492: SvelteKit v1.0.0-next.304 is suddenly executing code during build, prerender problem?
### Describe the bug

I am building a full stack SvelteKit adapter-node application with database connection. It is build in a container to be run in a container. Todays `npm update` killed my build process because `npm run build` started to _execute_ my source code.

So far this would build, but with `SvelteKit v1.0.0-next.304` (and its dependencies) build started to fail:
```
// hooks.ts
export const handle = async ({ event, resolve }) => {
	// simulate crash because DB can not be connected 
        // without credentials and access during build process
	console.log('Going to simulate crash')
	process.exit();

	return response;
};
```
My last _good_ `npm update` was 3 days ago.

Note: The build process always worked, but I had to use `npm ci --production --ignore-scripts` later with the built code.

This may have to do with the [pre-rendering](https://kit.svelte.dev/docs/configuration#prerender), but I have not touched those settings.

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-9crzvj?file=src%2Fhooks.js
```
const config = {
	kit: {
		prerender: {
			enabled: false
		}
	}
}
```

Run `npm run build` and it will execute the code and exit the process

### Logs

_No response_

### System Info

```shell
WORKS

  System:
    OS: macOS 12.2.1
    CPU: (8) arm64 Apple M1
    Memory: 105.05 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.19.0 - /opt/homebrew/opt/node@14/bin/node
    npm: 6.14.16 - /opt/homebrew/opt/node@14/bin/npm
  Browsers:
    Chrome: 100.0.4896.60
    Firefox: 98.0.2
    Safari: 15.3
  npmPackages:
    @sveltejs/adapter-auto: ^1.0.0-next.33 => 1.0.0-next.33
    @sveltejs/adapter-node: ^1.0.0-next.73 => 1.0.0-next.73
    @sveltejs/kit: ^1.0.0-next.303 => 1.0.0-next.303
    svelte: ^3.46.0 => 3.46.4

WORKS NOT ANYMORE

  npmPackages:
    @sveltejs/adapter-auto: ^1.0.0-next.34 => 1.0.0-next.34
    @sveltejs/adapter-node: ^1.0.0-next.73 => 1.0.0-next.73
    @sveltejs/kit: ^1.0.0-next.304 => 1.0.0-next.304
    svelte: ^3.46.6 => 3.46.6
```

### Severity

blocking an upgrade

### Additional Information

_No response_

dimfeld:
Fixes #4492. 

Loading `hooks.js` and/or executing the `handle` hook during the build process causes problems for some people, and so rendering the fallback page event when prerendering is disabled presents those people with build failures that can not be worked around. This PR reverts  #4443, which enabled that behavior after the prerender process was moved into the build step.

For those like me who need a fallback page but are saddled with code that causes the build process to fail because Node.js throws a tantrum when you even try to `import` a page component, a configuration workaround is presented in the README for `adapter-static`. (An example is dependencies that are ESM compatible, but not in the exact one way that Node requires them to be.)

### Tests

### Changesets

### DIFF
--- a/packages/adapter-static/README.md
+++ b/packages/adapter-static/README.md
@@ -71,6 +71,8 @@ export default {
 
 When operating in SPA mode, you can omit `config.kit.prerender.default` (or set it to `false`, its default value), and only pages that have the [`prerender`](https://kit.svelte.dev/docs/page-options#prerender) option set will be prerendered at build time.
 
+If Node.js throws errors when loading your pages during the build process and this is hard for you to work around, setting `config.kit.prerender.entries` to `[]` will prevent the build process from attempting to load any page or layout components. Do not set `config.kit.prerender.enabled` to `false` because that will also prevent the fallback page from being generated.
+
 > ⚠️ During development, SvelteKit will still attempt to server-side render your routes. This means accessing things that are only available in the browser (such as the `window` object) will result in errors, even though this would be valid in the output app. To align the behavior of SvelteKit's dev mode with your SPA, you can [call `resolve()` with a parameter of `{ssr: false}` inside the `handle()` hook](https://kit.svelte.dev/docs/hooks#handle).
 
 ## GitHub Pages
--- a/packages/adapter-static/README.md
+++ b/packages/adapter-static/README.md
@@ -71,7 +71,7 @@ export default {
 
 When operating in SPA mode, you can omit `config.kit.prerender.default` (or set it to `false`, its default value), and only pages that have the [`prerender`](https://kit.svelte.dev/docs/page-options#prerender) option set will be prerendered at build time.
 
-If Node.js throws errors when loading your pages during the build process and this is hard for you to work around, setting `config.kit.prerender.entries` to `[]` will prevent the build process from attempting to load any page or layout components. Do not set `config.kit.prerender.enabled` to `false` because that will also prevent the fallback page from being generated.
+SvelteKit will still crawl your app's entry points looking for prerenderable pages. If `svelte-kit build` fails because of pages that can't be loaded outside the browser, you can set `config.kit.prerender.entries` to `[]` to prevent this from happening. (Setting `config.kit.prerender.enabled` also has this effect, but would prevent the fallback page from being generated.)
 
 > ⚠️ During development, SvelteKit will still attempt to server-side render your routes. This means accessing things that are only available in the browser (such as the `window` object) will result in errors, even though this would be valid in the output app. To align the behavior of SvelteKit's dev mode with your SPA, you can [call `resolve()` with a parameter of `{ssr: false}` inside the `handle()` hook](https://kit.svelte.dev/docs/hooks#handle).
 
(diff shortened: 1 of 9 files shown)


packages/adapter-static/README.md
```diff
@@ -71,6 +71,8 @@ export default {
 
 When operating in SPA mode, you can omit `config.kit.prerender.default` (or set it to `false`, its default value), and only pages that have the [`prerender`](https://kit.svelte.dev/docs/page-options#prerender) option set will be prerendered at build time.
 
+If Node.js throws errors when loading your pages during the build process and this is hard for you to work around, setting `config.kit.prerender.entries` to `[]` will prevent the build process from attempting to load any page or layout components. Do not set `config.kit.prerender.enabled` to `false` because that will also prevent the fallback page from being generated.
```
dummdidumm, line 74: Interesting, I just converted a project to SvelteKit yesterday, using the SPA mode, and I did not have this problem. Setting `prerender.enabled` to `false` worked for me, the final build folder contained the specified `index.html`

Rich-Harris: As of this PR, the fallback `index.html` wouldn't be generated if `prerender.enabled` is set to `false`

Rich-Harris, line 74: I think we should try to explain _why_ this is happening:

```suggestion
SvelteKit will still crawl your app's entry points looking for prerenderable pages. If `svelte-kit build` fails because of pages that can't be loaded outside the browser, you can set `config.kit.prerender.entries` to `[]` to prevent this from happening. (Setting `config.kit.prerender.enabled` also has this effect, but would prevent the fallback page from being generated.)
```

Rich-Harris (approved):

packages/adapter-static/README.md
dimfeld: Thanks for the editing! Yes this reads much better.

dimfeld: @dummdidumm Here's the timeline of events :)

- Before version 292, the fallback page would be created even if prerendering was disabled.
- 292 moved the prerender into the build process and as a side effect caused the fallback page to not be generated when `prerender.enabled: false`.
- I submitted a PR to change this which was merged into 304, but it caused problems for people who were doing things like instantiating database clients in the global scope from `hooks.js`, hence this PR.
- While investigating solutions, I discovered that setting `prerender.entries: []` would work for cases where Node.js will complain if it even tries to `import` the page/layout components.
- And so this PR, which reverts back to the 292 behavior while documenting what to do it that causes issues.

dummdidumm: ah yeah my bad, obviously this worked for me because the change wasnt reverted at that time, thanks for the clarification 😄

PR #4611, opened 2022-04-13
install turborepo

Rich-Harris:
first crack at getting [turborepo](https://turborepo.org) set up. if i've done this correctly (big if), this will make things like docs PRs much faster to get through CI, since we can replay remote caches for unchanged packages rather than re-running the entire CI workflow every time. similarly, tweaking a changeset shouldn't cause another 10 minute wait to get a bank of green lights.

to take advantage of this, we need to do e.g. `pnpm turbo run lint` instead of `pnpm lint`. i haven't perfectly preserved all the `--filter` options we had in place previously, since in theory if everything is cached we can be more relaxed about that sort of thing


Rich-Harris:
once this is merged, we should update the build commands on vercel (on [both](https://vercel.com/svelte/kit/5FyWUo4wf1AFnquBfpC1GNbJRQQG) [projects](https://vercel.com/svelte/kit-svelte-dev/86jEoAEavP7RoZCaUqb3toXBc9Lm)) to use `pnpm turbo run build` rather than `pnpm build -r`

image

PR #4617, opened 2022-04-14
fix prettier config

Rich-Harris:
don't exactly know how this broke, but this seems to fix the lint failures that suddenly started happening

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -7,11 +7,12 @@
 		{
 			"files": ["*.svelte"],
 			"options": {
-				"svelteBracketNewLine": true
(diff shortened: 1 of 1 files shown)


.prettierrc
```diff
@@ -7,11 +7,12 @@
 		{
 			"files": ["*.svelte"],
 			"options": {
-				"svelteBracketNewLine": true
+				"bracketSameLine": false
```
dummdidumm, line 10: Strange, both should work. The former is deprecated in favor of the latter, so this is a good change anyway.

dominikg (approved):

PR #4623, opened 2022-04-14
fix some turbo stuff

Rich-Harris:
we can speed up deploys of kit.svelte.dev quite a bit by only installing and building the bits we need, but in order to do that we need to explicitly mark the generated `types.js` file as an output of the `build` script in the `kit` package


Rich-Harris:
going to merge this immediately since it's blocking other builds

PR #4633, opened 2022-04-15
document generated tsconfig.json, move compilerOptions into user config

Rich-Harris:
https://kit-svelte-dev-git-document-tsconfig-svelte.vercel.app/docs/types#generated-types-default-tsconfig-json

We generate a bunch of config options, but don't explain what they are and which ones can be overridden. This fixes that


documentation/docs/15-types.md
```diff
@@ -74,3 +74,66 @@ export async function get({ params }) {
...
+
+Others are required for SvelteKit to work properly, and should also be left untouched unless you know what you're doing:
+
+```json
+/// file: .svelte-kit/tsconfig.json
+{
+	"compilerOptions": {
+		// you must use `import type` to
+		// import types, rather than `import`,
+		// for svelte-preprocess to work
```
dummdidumm, line 106: The last line isn't 100% correct, it's more a hint like "if all of the imports are used as types, throw an error". Writing this, I'm wondering how this plays out in conjunction with "preserveValueImports".. maybe we can remove this option. Need to test this out, don't know when I will get to that.

Btw since TS 4.5 you can mix value and type imports by doing `import { type Foo, bar } ..`.

Rich-Harris: I updated it — it's slightly tricky because we need to communicate how it affects how people should write code, more than what the option actually does, but it also needs to make sense in context

PR #4636, opened 2022-04-15
[feat] invalidate with predicate

### LINKED ISSUE #4540: Add more flexibility to invalidate
### Describe the problem

I find the current implementation of `invalidate` to be quite limiting in practice even tough I really like the base concept.
The fact the `href` argument of `invalidate` has to mach the `url` of a fetched resource exactly makes it hard to deal with more complex URLs that are generated and/or contain query parameters.

**Example**

I fetched `/collections`, `/collections?category=foo` and `/collections?category=bar` in different `load` functions.
Now I want to `invalidate` all `load` functions that fetched the `/collections` endpoint with and without query params. At the moment this is only possible if you keep track of all fetched URLs and invalidate them individually.

Extend invalidate to take a `predicate` function as argument (similar to `Array.prototype.some`).
The `predicate` is run for each item of [`dependencies`](https://github.com/sveltejs/kit/blob/83f60af73dd6b7a86167281bca7f85122c83818e/packages/kit/src/runtime/client/client.js#L474) and will cause a rerun of `load` if `true` is returned.

**Signature:**
```ts
function invalidate(href: string): Promise<void>;
function invalidate(predicate: (dep: string) => boolean): Promise<void>;
```

**Example usage:**
```ts
<script context="module">
export async function load({fetch, url}) {
  const response = await fetch(`/collection${url.search}`);
  const data = await response.json();
  
  return { props: { data }};
}
</script>
<script>
const isCollectionURL = (dep) => dep.includes('/collections');
</script>

<button on:click={() => invalidate(isCollectionURL)}>Reload collection</button>
```

This would make the `invalidate` function far more useful IMO and would not break how it's currently used.

### Alternatives considered

Keep track of the fetched resources in `stuff` and filter the list of resources your self before passing it to invalidate.
This is quite cumbersome though and would not work well with APIClients that generate the URLs under the hood.

### Importance

### Additional Information

This feature is fairly simple to implement and I'm willing to implement it my self. I just wanted to get some feedback before making a PR

icalvin102:
Closes #4540

### Problem

The current implementation of `invalidate` is quite limiting because the provided argument has to match the URL of the fetched resource exactly. It's not possible to invalidate resources that match a certain pattern. (Take a look at #4540 for more detailed problem description) 

### Solution

Extend invalidate to take a `predicate` function as argument (similar to `Array.prototype.some`).
The `predicate` is run for each item of [`dependencies`](https://github.com/sveltejs/kit/blob/83f60af73dd6b7a86167281bca7f85122c83818e/packages/kit/src/runtime/client/client.js#L474) and will cause a rerun of `load` if `true` is returned.

**Signature:**
```ts
function invalidate(href: string): Promise<void>;
function invalidate(predicate: (dep: string) => boolean): Promise<void>;
```

**Example usage:**
```svelte
<script context="module">
export async function load({fetch, url}) {
  const response = await fetch(`/collections${url.search}`);
  const data = await response.json();
  
  return { props: { data }};
}
</script>
<script>
  const isCollectionURL = (dep) => dep.includes('/collections');
</script>

<button on:click={() => invalidate(isCollectionURL)}>Reload collection</button>
```

This makes the `invalidate` function far more useful IMO and would not break how it's currently used.

### Tests

### Changesets

### DIFF
--- a/.changeset/hungry-weeks-exercise.md
+++ b/.changeset/hungry-weeks-exercise.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+`invalidate` with `predicate` function
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -59,8 +59,8 @@ export function create_client({ target, session, base, trailing_slash }) {
 	/** @type {Map<string, import('./types').NavigationResult>} */
 	const cache = new Map();
 
-	/** @type {Set<string | ((href: string) => boolean)>} */
-	const invalidated = new Set();
+	/** @type {Array<((href: string) => boolean)>} */
+	const invalidated = [];
 
 	const stores = {
 		url: notifiable_store({}),
@@ -228,7 +228,7 @@ export function create_client({ target, session, base, trailing_slash }) {
 		// abort if user navigated during update
 		if (token !== current_token) return;
 
-		invalidated.clear();
+		invalidated.length = 0;
 
 		if (navigation_result.redirect) {
 			if (redirect_chain.length > 10 || redirect_chain.includes(url.pathname)) {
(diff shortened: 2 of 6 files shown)


icalvin102:
I have included a test. However the test fails because playwright is unable to serialize the `predicate` function and throws a `Unexpeced value` error in https://github.com/sveltejs/kit/blob/f445802cba871393898cadc2ffcc950b45443179/packages/kit/test/utils.js#L27

This is a problem with playwright and not with the actual feature. I don't have much experience with playright and am not sure how to solve this. It would be nice if someone could take a look at this or guide me in the right direction.

Rich-Harris (approved):

Rich-Harris:
thank you!

PR #4637, opened 2022-04-15
remove .npmrc and .prettierrc from new projects

Rich-Harris:
this might be controversial but i figured the easiest way to discuss it would be to throw up a PR.

when you create a new project, you immediately end up with a whole bunch of files:

image

the more stuff we have there, the more intimidating it is to figure out what's doing what, and which files deserve attention. i feel like we could remove two of them — `.npmrc` and `.prettierrc`. there's still a lot of stuff, but it's not quite as overwhelming:

image
 
`.npmrc` is presumably there because SvelteKit requires Node 14.13 or higher. But Node 12 is about to fall out of LTS, and I'd expect vanishingly few people to still be using anything below 14.13 (which was released in September 2020). Moreover, _it doesn't work_ — all it does is prevent you from _installing_ SvelteKit with an older version of Node; it only circumstantially prevents you from _running_ it. I think it's an inelegant, flawed solution, and we're better off just documenting the version requirement. (If you do try to run `svelte-kit` or an app built with `adapter-node` with Node 12, it will immediately fail with a syntax error due to the use of optional chaining.)

`.prettierrc` is redundant; the config can be expressed equivalently in `package.json`. Unfortunately we can't do the same for `.eslintrc.cjs` because one of the variants has `() => require('typescript')`.

### Tests

### Changesets

### DIFF
--- a/.changeset/quick-owls-lie.md
+++ b/.changeset/quick-owls-lie.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Reduce number of files in new project
--- a/packages/create-svelte/shared/+prettier/.prettierrc
+++ b/packages/create-svelte/shared/+prettier/.prettierrc
@@ -1,6 +1,0 @@
-{
-	"useTabs": true,
-	"singleQuote": true,
-	"trailingComma": "none",
-	"printWidth": 100
-}
--- a/packages/create-svelte/shared/+prettier/package.json
+++ b/packages/create-svelte/shared/+prettier/package.json
@@ -2,5 +2,11 @@
 	"devDependencies": {
 		"prettier": "^2.5.1",
 		"prettier-plugin-svelte": "^2.5.0"
+	},
+	"prettier": {
+		"useTabs": true,
+		"singleQuote": true,
+		"trailingComma": "none",
+		"printWidth": 100
 	}
 }
--- a/packages/create-svelte/templates/default/.npmrc
+++ b/packages/create-svelte/templates/default/.npmrc
@@ -1,1 +1,0 @@
-engine-strict=true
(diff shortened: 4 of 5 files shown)


dominikg:
Not so sure about prettier. This is a choice in `create-svelte` and not seeing a config file might confuse users aswell.

If we decide to put it in package.json we should filter it out in `svelte-kit package` alongside scripts.

Personally i'd rather keep it separate, there are ways to hide common config files from view in vscode if users are turned off by too much clutter. https://github.com/antfu/vscode-file-nesting-config

dummdidumm:
In favor of removing .npmrc, against removing .prettierrc. people opting in to prettier know what that file does because it's by far the most used config file for it, so for me it would be more confusing to not have it.

benbender:
I'm using [.editorconfig](https://editorconfig.org/) instead of .prettierrc as it has broader support (incl. prettier) and helps with consistency.

.npmrc could be replaced with an explicit [engines](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines)-entry in package.json.

Maybe these could be alternatives?

Rich-Harris:
`engines` applies to packages (and we already have one set in the Kit package.json) — it's not a replacement for `engine-strict`, it's the thing that `engine-strict` applies to

benbender:
> `engines` applies to packages (and we already have one set in the Kit package.json) — it's not a replacement for `engine-strict`, it's the thing that `engine-strict` applies to

it depends. F.e. [vercel](https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version) reads and uses the field to setup the deployment.

benmccann:
There was no end of people running into trouble using old versions of Node before we added the `.npmrc`. I'm not sure it's worth the added noise and support burden to remove it. It's rarely seen anyway since it's hidden

PR #4640, opened 2022-04-15
fix cross-document scroll bug (#4386)

### LINKED ISSUE #4386: history scroll restoration usage with native navigation

Rich-Harris:
fixes #4386, supersedes #4425. this is a WIP though because for some reason the test succeeds even _without_ this PR — somehow it's measuring before JS kicks in, I think. Need to investigate

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -127,11 +127,19 @@ export function create_client({ target, session, base, trailing_slash }) {
 	let router_enabled = true;
 
 	// keeping track of the history index in order to prevent popstate navigation events if needed
-	let current_history_index = history.state?.[INDEX_KEY] ?? 0;
+	let current_history_index = history.state?.[INDEX_KEY];
 
-	if (current_history_index === 0) {
+	if (!current_history_index) {
+		// we use Date.now() as an offset so that cross-document navigations
+		// within the app don't result in data loss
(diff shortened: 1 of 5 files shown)


PH4NTOMiki:
@Rich-Harris that's why said it's hard to test it, because it doesn't happen every time, sometimes the browser returns the previous page from ITS cache, and not with Svelte Client Router , and I don't know how it decides when to do that, and that's actually a bug of itself because sometimes it could return to old app version.

Rich-Harris:
Managed to coax the test into failing somewhat reliably without this fix. It's not 100% (scroll behaviour seems somewhat non-deterministic in browsers) but good enough that I'm happy merging this

Issue #4642, opened 2022-04-16
Error building project - "Cannot find package 'cookie'"


callumgare:
### Describe the bug

Following the instructions on https://github.com/sveltejs/kit/blob/master/CONTRIBUTING.md exactly I get the following error:

```
kit.svelte.dev:build: > Cannot find package 'cookie' imported from /private/tmp/kit/sites/kit.svelte.dev/.svelte-kit/output/server/index.js
kit.svelte.dev:build: Did you mean to import cookie@0.5.0/node_modules/cookie/index.js?
kit.svelte.dev:build: Did you mean to import cookie@0.5.0/node_modules/cookie/index.js?
kit.svelte.dev:build:     at new NodeError (node:internal/errors:372:5)
kit.svelte.dev:build:     at packageResolve (node:internal/modules/esm/resolve:908:9)
kit.svelte.dev:build:     at moduleResolve (node:internal/modules/esm/resolve:957:20)
kit.svelte.dev:build:     at defaultResolve (node:internal/modules/esm/resolve:1173:11)
kit.svelte.dev:build:     at ESMLoader.resolve (node:internal/modules/esm/loader:604:30)
kit.svelte.dev:build:     at ESMLoader.getModuleJob (node:internal/modules/esm/loader:318:18)
kit.svelte.dev:build:     at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:80:40)
kit.svelte.dev:build:     at link (node:internal/modules/esm/module_job:78:36)
kit.svelte.dev:build:  ELIFECYCLE  Command failed with exit code 1.
kit.svelte.dev:build: Error: command finished with error: command (sites/kit.svelte.dev) pnpm run build exited (1)
command (sites/kit.svelte.dev) pnpm run build exited (1)
```

### Reproduction

I have literally just done the steps in https://github.com/sveltejs/kit/blob/master/CONTRIBUTING.md (deleting everything and trying again several times):

```bash
npm i -g pnpm
git clone git@github.com:sveltejs/kit.git
cd kit
pnpm install
pnpm build
```

### System Info

```shell
System:
    OS: macOS 12.3
    CPU: (10) arm64 Apple M1 Max
    Memory: 109.58 MB / 32.00 GB
    Shell: 3.4.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 17.9.0 - /opt/homebrew/bin/node
    npm: 8.5.5 - /opt/homebrew/bin/npm
  Browsers:
[...]

PR #4643, opened 2022-04-16
Fix url bug in the hooks documentation examples

callumgare:
In a hook `event.request.url` will be a complete URL I believe (https://developer.mozilla.org/en-US/docs/Web/API/Request/url) meaning `event.request.url.startsWith('/custom')` will never be true. This commit fixes that.

<details>
<summary>Checklist boilerplate</summary>

  **I haven't bothered with this since it's such a small and hopefully non-controversial change.**
  **N/A but I have tried out the old example and run into this issue and changing it fixes it**

### Tests
  **I couldn't get the project to build due to an error I was running into (https://github.com/sveltejs/kit/issues/4642) so I couldn't run these but since it's purely a docs change the code base shouldn't be impacted.**

### Changesets
  **N/A**
</details>

### DIFF
--- a/documentation/docs/05-hooks.md
+++ b/documentation/docs/05-hooks.md
@@ -14,7 +14,7 @@ This function runs every time the SvelteKit server receives a [request](/docs/we
 /// file: src/hooks.js
 /** @type {import('@sveltejs/kit').Handle} */
 export async function handle({ event, resolve }) {
-	if (event.request.url.startsWith('/custom')) {
+	if (event.url.pathname.startsWith('/custom')) {
 		return new Response('custom response');
 	}
 


Mlocik97:
Doesn't we need to jsut add `event.url.pathname.startWith()` instead?  Because url in event object is already Instance of URL. So constructing URL from URL actualy does nothing.

callumgare:
Opps! Quite right you are. I didn't realise there was both an `event.request.url` and a `event.url`, one with a string and one with a URL object. I've updated the commit now.

Mlocik97 (approved):

Conduitry (approved):

Issue #4093, opened 2022-02-23
Mark endpoints as prerenderable


Rich-Harris:
### Describe the problem

If an endpoint loads data from the filesystem...

```js
import fs from 'fs';
import marked from 'marked';

export function get({ params }) {
  const markdown = fs.readFileSync(`content/${params.slug}.md`, 'utf-8');

  return {
    body: marked(markdown)
  };
}
```

...then in most environments it will won't work in production — it's only suitable for prerendering. SvelteKit doesn't know that, and will include the endpoint in the built app. This is bad for two reasons:

* it means deploying unnecessary code
* a mistyped URL will 500 instead of 404

We're experiencing this right now with https://kit.svelte.dev/docs/nope.

### Describe the proposed solution

If we could mark the endpoint with `export const prerender = true`, just like we do with pages, we could avoid including the endpoint in the deployed app.

The semantics are slightly different to `export const prerender = true` in pages, where it means 'you can prerender this page' as opposed to 'you can ONLY prerender this endpoint', but I think that's probably fine.

### Importance

would make my life easier

Conduitry:
Would this mark all defined HTTP methods for that endpoint as needing to be prerendered? Do we need more granularity than that?

Rich-Harris:
On most platforms, it goes to the filesystem first — if there's a `foo.json` then it will either respond with that when you `POST /foo.json` or it will 405. So I think we have to treat the existence of `export const prerender = true` alongside `export function post` etc as an error

vwkd:
> 'you can ONLY prerender this endpoint', but I think that's probably fine.

I think that’s probably confusing. Why? If you have two things which by default both are true, then saying one is true doesn’t automatically imply the other is false.

Maybe the option could disable the opposite. Maybe it could be `dynamicRendering = false` or something g like that.

Theo-Steiner:
I think this would be a great feature, especially for apps where a part of the data from let's say a CMS changes only occasionally.
Right now, you can return maxage from load functions to not fetch on every subsequent navigation, but according to the docs this only works in non __layout.svelte routes and therefore isn't really an option for all use cases. (Especially if you would want to eliminate boilerplate by fetching the data in the layout and pass it down using stuff)
[...]

PR #4653, opened 2022-04-18
allow endpoints to be treated as prerender entry points

### LINKED ISSUE #2541: Adapter static does not export an index.html for root redirect
### Describe the bug

When using adapter static with a redirect server root, no root index.html is exported.

This results in a 404 error when fetching the root index.html, instead of redirecting to the wanted location.

src/
 - routes/
   - index.ts --> this one redirects to en/index.svelte
   - en/
     - index.svelte

# What happens:

No root index.html is created when exporting with adapter static.

That is normal, since server routes are simply ignored, but :

# What i'd like

 1) Emit a warning about using adapter static with a repo that has server routes.
 2) // MAYBE?? Provide a better way to use redirect routes

# Workaround

replace index.ts with index.svelte and put the script content in `<script context="module>`

### Reproduction

coming soon

### Logs

_No response_

### System Info

```shell
Github Pages
```

### Severity

annoyance

### Additional Information

_No response_

Rich-Harris:
fixes #2541. An endpoint like `src/routes/index.js` or `src/routes/sitemap.xml.js` or whatever should be prerenderable.

Though now that I've implemented it and I'm writing this comment, it occurs to me that this should probably only apply to endpoints that are explicitly marked as prerenderable, unless `config.kit.prerender.default === true`. Will leave this open pending #4093.

### Tests

### Changesets

### DIFF
--- a/.changeset/small-countries-doubt.md
+++ b/.changeset/small-countries-doubt.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Allow endpoints to be treated as prerender entry points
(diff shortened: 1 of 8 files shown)


Rich-Harris:
obsolete since #6392

PR #4657, opened 2022-04-19
feat: add support for Netlify Edge Functions

ascorbic:
This PR adds support for the beta release of [Netlify Edge Functions](https://www.netlify.com/blog/announcing-serverless-compute-with-edge-functions) to the Netlify adapter. It adds an option `edge` to the adapter, which if set to true, compiles an edge function that runs on the Deno runtime as well as an edge functions manifest, instead of the Node-based Netlify Function. 

### Tests

### Changesets


packages/adapter-netlify/index.js
```diff
@@ -26,111 +46,29 @@ export default function ({ split = false } = {}) {
...
-			const redirects = [];
-
-			const replace = {
-				'0SERVER': './server/index.js' // digit prefix prevents CJS build from using this as a variable name, which would also get replaced
-			};
-
-			if (esm) {
-				builder.copy(`${files}/esm`, '.netlify', { replace });
+			if (edge) {
+				await generate_edge_functions({ builder });
```
Rich-Harris, line 62: It looks like `generate_edge_functions` doesn't support function splitting — is it worth making this clear if someone tries to use both together?

```suggestion
			if (edge) {
				if (split) {
					throw new Error('Cannot use `split: true` alongside `edge: true`');
				}
				
				await generate_edge_functions({ builder });
```

Rich-Harris (commented):
Thanks, and congrats on the launch! I made a small tweak to the README (the GitHub UI didn't let me use the suggestion tool, since the diff had both red and green, but our convention is to document all the available options in the code sample), and added an inline comment about handling `edge: true` alongside `split: true`

packages/adapter-netlify/index.js
ascorbic: Ah, good point!

packages/adapter-netlify/README.md
```diff
@@ -17,9 +17,15 @@ import adapter from '@sveltejs/adapter-netlify';
...
+		// default options are shown
 		adapter: adapter({
+			// if true, will create a Netlify Edge Function rather
+			// than using standard Node-based functions
+			edge: false,
+
 			// if true, will split your app into multiple functions
-			// instead of creating a single one for the entire app
+			// instead of creating a single one for the entire app.
+			// if `edge` is true, this option will be disregarded
```
ascorbic, line 28:
```suggestion
			// if `edge` is true, this option cannot be used
```
@Rich-Harris I think the wording needs to be stornger, as we're throwing an error now if they combine them

Rich-Harris (approved):

Rich-Harris:
thank you!

Masstronaut:
@ascorbic thrilled to see you implemented this!

PR #4658, opened 2022-04-19
add `--watch` flag to `svelte-kit package`

### LINKED ISSUE #2333: Watch mode for package command
### Describe the problem

Some people use SvelteKit to create a package but use/test this package in a different environment than the `routes` folder in the same project. For them to get updates, they have to run the package command over and over again or write a custom script for that, which might not work for windows users. Details: https://github.com/sveltejs/component-template/issues/29#issuecomment-908589456

Add a `--watch` option to the package command which rebuilds the package without deleting its top level folder

### Alternatives considered

Don't add it, but at least don't delete the top level folder (only empty its content) so that the changes are properly picked up by watchers.

### Importance

nice to have

### Additional Information

_No response_

Rich-Harris:
Closes #2333. Undertook a bit of refactoring while I was at it, both to make incremental changes possible (side note: would be nice if `emit_dts` could be made to work incrementally in future) and also just as a general tidy-up.

Also, it seems it was failing to build packages with `jsconfig.json` instead of `tsconfig.json`. Fixed that in this PR.

### Tests

### Changesets


packages/kit/src/packaging/index.js
```diff
@@ -1,134 +1,38 @@
...
-			out_contents = await transpile_ts(filename, source.toString('utf-8'));
-			out_contents = resolve_$lib_alias(out_file, out_contents, config);
-		} else {
-			out_file = file;
-			out_contents = source;
-		}
-
-		write(path.join(package_dir, out_file), out_contents);
+	rimraf(dir);
+	mkdirp(dir); // TODO https://github.com/sveltejs/kit/issues/2333
```
ignatiusmb, line 25:
```suggestion
	mkdirp(dir);
```

packages/kit/src/packaging/test/fixtures/files-exclude/svelte.config.js
```diff
@@ -4,6 +4,7 @@ const config = {
 		package: {
 			files(filepath) {
 				const ext = filepath.slice(filepath.lastIndexOf('.') + 1);
+				console.error({ filepath, ext });
```
ignatiusmb, line 7:
```suggestion
```

ignatiusmb (approved):
A bit of cleanup and we should be good

PR #4662, opened 2022-04-19
Add link to base path suggestion

plmrry:
When visiting `http://localhost:3000/` with a base path configured, it would be nice to render the suggested path as a link.

Also note that [vite does something similar](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/middlewares/base.ts#L43-L46)

### Tests

### Changesets


Conduitry (changes requested):
As written, this isn't going to work. See the conversation in https://github.com/sveltejs/kit/pull/4510#discussion_r842198962

plmrry:
Oops, sorry. I tried to search for related issues/PRs but clearly not hard enough! Will close.

Issue #4659, opened 2022-04-19
Trouble catching `fetch` errors in module context `load`


stephenlrandall:
### Describe the bug

Adding module context `load` functions with `fetch` attempts to non-existent resources seems to introduce errors that can't be caught. Similar `fetch` calls in the non-module context script tags can have their errors caught in the expected way:

```
res = await fetch('/resource/no/exist').catch(err => ...);
```

Additionally, `fetch` errors in `__layout.svelte` specifically will cause an infinite loop as the `load` function is run over and over again.

### Reproduction

In the template SvelteKit to-do application, add something to the effect of:

```
<script context="module">
	export async function load({ fetch }) {
		console.log("Attempting to fetch...");

		const res = await fetch("/-favicon.png").catch(_ => console.log("Caught an error!"))

		if (res.ok)
			return { status: 200 };
		else
			return {
				status: res.status,
				props: { error: res.statusText }
			};
	}
</script>
```

If this is in `__layout.svelte`, the "Attempting to fetch..." line will be printed indefinitely as the load hangs. If this is in a specific route (like the About page), navigation will not hang but the error will also not get caught.

If this `fetch` is not in module context, the error is caught and the "Caught an error!" line is printed. If the `fetch` is changed to an existing resource ("-favicon" -> "favicon", for example), everything works fine as well.

### System Info

```shell
System:
    OS: macOS 12.3.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 780.53 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.9.0 - /usr/local/bin/node
    npm: 8.5.5 - /usr/local/bin/npm
  Browsers:
    Safari: 15.4
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.34 
    @sveltejs/kit: next => 1.0.0-next.314 
    svelte: ^3.46.0 => 3.47.0
```

### Severity

serious, but I can work around it

### Additional Information

I can work around this issue by moving the `fetch` calls into the main script to correctly catch any errors, but this is obviously non-ideal as content will flash/jump as it's loaded in.

Really hoping this is me missing something obvious, and apologies if that's the case! I tried asking about this in the Svelte Discord a couple times but got no response.

plmrry:
Isn't this expected behavior? `fetch` does not throw on 404s.

stephenlrandall:
[...]

Issue #3068, opened 2021-12-17
Thrown exceptions from __layout.svelte load result in plain text error responses


Conduitry:
### Describe the bug

Throwing an exception (or returning a rejecting promise) from a `load` function should be equivalent to returning `{ status: 500, error }` (apart from the former also resulting in a call to the `handleError` hook). However, currently, it results in a plain text error message response from the server, which should be returned only when the `__error.svelte` component fails to render.

### Reproduction

Starting from the demo app, add this to `__layout.svelte`:

```svelte
<script context="module">
	export function load() {
		throw new Error("FOO");
	}
</script>
```

Navigate to any page.

### System Info

```shell
System:
    OS: Linux 5.10 Ubuntu 20.04.3 LTS (Focal Fossa)
    CPU: (8) x64 Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz
    Memory: 6.36 GB / 12.32 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Yarn: 1.22.15
    npm: 8.1.4
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.4
    @sveltejs/kit: next => 1.0.0-next.202
    svelte: ^3.44.0 => 3.44.3
```

### Severity

serious, but I can work around it

### Additional Information

This is inconvenient because I'd like to use the `handleError` hook as the sole place for performing certain kinds of error logging, but I'd also prefer to not return a plain text error response to users. This means that currently my `__layout.svelte`'s `load` function needs to catch the error and reimplement the `handleError` handling in a different way (since it has access to `load` arguments instead of to the `request` object).

Conduitry:
The issue seems to be in https://github.com/sveltejs/kit/blob/master/packages/kit/src/runtime/server/page/respond_with_error.js where the first call to `load_node` for the layout component is returning a rejecting promise. I could make it not do that and (when `is_leaf` and `is_error` are both `false` - i.e., when this is a layout component) instead catch the error thrown by `module.load.call` and turn it into a `{ status, error }` object but something about that feels a bit unclean. It might be better to make this the responsibility of `respond_with_error`, but that also feels unclean, and also trickier to do, and requires more duplication of `load_node` logic.

Conduitry:
Implementation aside, I'm now having a bit of a question about what the desired behavior even is.

[...]

PR #4665, opened 2022-04-19
handle non-200 status codes from root `__layout.svelte`

Rich-Harris:
#4659. Just a failing test for now. Looks like the 'bail out and render a generic 500 if an error occurs while rendering an error page' thing only applies to errors that are thrown; `return { status: 404 }` doesn't do the trick.

### Tests

### Changesets


Conduitry:
I haven't fully been following what's happening in that issue - but from your description in this PR, it sounds at least somewhat adjacent to #3068 - although the symptoms are different

PR #4675, opened 2022-04-20
hardcode team name

Rich-Harris:
Secrets are redacted from GitHub Actions logs. Since we have a `TURBO_TEAM` environment variable, to enable remote caching, we have a secret whose value is `svelte`. That results in some... interesting redactions:

image

Since it's not really a secret that we're the `svelte` team, this replaces the secret with a hardcoded value.

image


pngwn (approved):

Conduitry:
Looks good to me, but I don't know how this could be failing linting now ...

https://github.com/sveltejs/kit/runs/6102827202

Rich-Harris:
The lint error is something unrelated that snuck in. Will be fixed by https://github.com/sveltejs/kit/pull/4674

Conduitry (approved):

PR #4686, opened 2022-04-21
remove default `<meta name="description">`

Rich-Harris:
The project templates include a default `<meta name="description">`, which is well-intentioned (you need one to get a good lighthouse score, etc) but ultimately counterproductive since it takes precedence over ones in `<svelte:head>`

### Tests

### Changesets

### DIFF
--- a/.changeset/stupid-owls-provide.md
+++ b/.changeset/stupid-owls-provide.md
@@ -2,4 +2,4 @@
 'create-svelte': patch
 ---
 
-Remove default <meta name="description">
+Remove default <meta name="description"> and add separate descriptions to each page
--- a/.changeset/stupid-owls-provide.md
+++ b/.changeset/stupid-owls-provide.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Remove default <meta name="description">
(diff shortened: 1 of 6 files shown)


Conduitry:
Do we want to add one to the root `<svelte:head>` in our demo apps, or at least the non-skeleton app?

Rich-Harris:
Yeah, good point — we already have a `<title>` in the demo app, so it makes sense for `<meta name="description">` to live there too

Rich-Harris:
Am in two minds about that. It's definitely _simpler_, but reusing descriptions for multiple pages is discouraged by SEO-heads. Short of faffing about with `stuff` (which feels out of scope for this), I think adding them to every page is probably the better choice

.changeset/stupid-owls-provide.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Remove default <meta name="description">
```
Conduitry, line 5:
```suggestion
Remove default <meta name="description"> and add separate descriptions to each page
```

Or something similar.

Conduitry (approved):

PR #4687, opened 2022-04-21
feat(ci): validate github actions with actionlint

jbergstroem:
Actionlint makes sure github actions are staying in line (as well as inline shell scripts through shellcheck). Fire off actionlint if there are changes to github actions in CI. Also, minor changes (quoting variables, adding `read -r`) to the release aciton to make actionlint happy.

### Tests

### Changesets

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -1,0 +1,21 @@
+name: Actionlint
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/*.yml'
+
+jobs:
+  actionlint:
+    name: Verify actions
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Actionlint
+        env:
+          ACTIONLINT_VERSION: 1.6.12
+        run: |
+          wget -q -O- "https://github.com/rhysd/actionlint/releases/download/v${ACTIONLINT_VERSION}/actionlint_${ACTIONLINT_VERSION}_linux_amd64.tar.gz" | tar -x -z -C . actionlint && \
+          mv actionlint /usr/local/bin
+      - name: Run Actionlint
+        run: |
+          actionlint -format '{{range $err := .}}::error file={{$err.Filepath}},line={{$err.Line}},col={{$err.Column}}::{{$err.Message}}{{end}}' -ignore 'SC2016:' .github/workflows/*.yml
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,10 +53,10 @@ jobs:
           published_packages: ${{steps.changesets.outputs.publishedPackages}}
         run: |
           sleep 20
-          cat << EOF > "$HOME/.npmrc"
-            //registry.npmjs.org/:_authToken=$NPM_TOKEN
+          cat << EOF > "${HOME}/.npmrc"
+            "//registry.npmjs.org/:_authToken=$NPM_TOKEN"
           EOF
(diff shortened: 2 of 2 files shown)


.github/workflows/actionlint.yml
```diff
@@ -0,0 +1,21 @@
+name: CI
```
dominikg, line 1: there is a workflow named CI already: https://github.com/sveltejs/kit/blob/master/.github/workflows/ci.yml#L1

.github/workflows/actionlint.yml
```diff
@@ -0,0 +1,21 @@
...
+  actionlint:
+    name: Lint Actions
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Actionlint
+        env:
+          ACTIONLINT_VERSION: 1.6.12
+        run: |
+          wget -q -O- "https://github.com/rhysd/actionlint/releases/download/v${ACTIONLINT_VERSION}/actionlint_${ACTIONLINT_VERSION}_linux_amd64.tar.gz" | tar -x -z -C . actionlint && \
```
dominikg, line 17: why does this download and execute a binary?

.github/workflows/actionlint.yml
jbergstroem: Yes, but then this job would fire on each PR; not only those that change relevant files which is why I took this approach instead.

.github/workflows/actionlint.yml
jbergstroem: Because thats the only way to execute the linter

.github/workflows/actionlint.yml
dominikg: This job would need a different name to tell them apart

.github/workflows/actionlint.yml
dominikg: Are there other ways to use this that do not involve executing arbitrary downloaded binaries in a CI env? The possible risk outweighs the benefits imho

jbergstroem: Thats mostly how CI jobs seem to work as far as I can tell. The action downloads / installs dependencies required for the work to be done, then executes it. Is it for instance very different to installing `prettier`?

.github/workflows/actionlint.yml
jbergstroem: Ah, my bad. Apologies

.github/workflows/actionlint.yml
jbergstroem: There are other ways of achieving the same, but they are all downloading the binary: https://github.com/rhysd/actionlint/blob/main/docs/usage.md#tools-integration

dominikg: yes, prettier is loaded from npm which does not allow republishing packages with the same version and any funky prettier release would be found out almost immediately due to its popularity.
The github release can be overwritten so it may work fine today but not tomorrow.

Funnily enough prettier itself seems to use  node-actionlint https://github.com/prettier/prettier/blob/66585d9f250f11569456421e66a2407397e98f69/package.json#L155

jbergstroem: Do you want me to switch to that?

dominikg: thats even less used than actionlint itself so i'd be still hesitant to add it to a CI with access to secrets. Running it in a properly secured local env and PRing any findings seems like a safer way to use this.

jbergstroem: I'll close the PR then.

jbergstroem:
Closing due to security concerns of using actionlint. If you see this PR and want to contribute – file a new PR with the improvements that actionlint suggests instead!

Issue #4549, opened 2022-04-06
Replace `maxage` in `load` with more detailed `cache` option


Rich-Harris:
### Describe the problem

Pages can return a `maxage` property from `load` which a) sets a `Cache-Control` header on the server response, and b) causes the returned value to be cached in memory so that navigating back to the page won't result in a repeated `load`.

This mostly works, but has limitations:

* determining whether to add `public` is based on a heuristic (no `fetch` without `credentials: 'omit'`, and no `session` access) that [doesn't cover every edge case](https://github.com/sveltejs/kit/issues/793#issuecomment-811305016)
* there's nowhere to express `revalidate`
* some people want to be able to express `s-maxage` rather than `maxage`, though I'll confess I don't fully understand why

### Describe the proposed solution

Replace `maxage` with a `cache` object:

```svelte
<script context="module">
  /** @type {import('@sveltejs/kit').Load} */
  export async function load() {
    return {
      props: {...},
      cache: {
        maxage: 3600,
        private: false // if unset, uses existing heuristic
      }
    };
  }
</script>
```

In future, we could add `revalidate` and `smaxage` if those prove necessary.

### Importance

nice to have

tcc-sejohnson:
Hey @Rich-Harris -- would you like me to tackle this one? Happy to grab it if no one else is currently working on it.

Rich-Harris:
That would be great @tcc-sejohnson, thank you!

tcc-sejohnson:
@Rich-Harris 

These changes will cause merge conflicts with the existing pull request I have open (#4515) -- the typedef for `LoadInput` has completely moved and changed to include the `ErrorLoadInput` properties. 

If you're bullish on that change, I can base the current changes off of the tip of my feature branch to make the eventual merge easier. If you're bearish on it or think it needs a significant rework, I can base the changes off of current `master` and we can just deal with the merge conflicts when we get to them.

tcc-sejohnson:
Disregard previous -- I see the PR has been merged. I'll work on this over the weekend and report back with any questions.

tcc-sejohnson:
@Rich-Harris 

Completion candidate: #4690 

I exhaustively updated the tests to cover every edge case I could think of.

tcc-sejohnson:
@Rich-Harris 

Glad to see it all worked out (other than the snafu with the doc site -- oops). Just wanted to let you know this issue is still open. I can't close it :)

andreivreja:
[...]

PR #4690, opened 2022-04-22
feat: Upgrade page caching

tcc-sejohnson:

### Tests

### Changesets

This completes #4549. A couple of notes for consideration before merging:

The `cache.private` option will override (perhaps confusingly?) behavior that's currently in place -- for example, `cache.private === false` will cause the `cache-control` header to be `public`, even if a credentialed fetch or the `session` object is accessed (whether in `init` or in `load`). Because of this, I was _extra specific_ in the tests, asserting that the override worked in every case. Even if this is a bit confusing at first, I can't think that going half way (having it not override the defaults) in some cases would be any clearer.

### DIFF
--- a/.changeset/gold-bikes-chew.md
+++ b/.changeset/gold-bikes-chew.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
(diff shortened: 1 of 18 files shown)


tcc-sejohnson:
Note: Cannot replicate test failure locally. ☹️

Rich-Harris (approved):

Rich-Harris:
thank you, this is a great PR!

Issue #4685, opened 2022-04-21
kit start mounting wrong node


john-michael-murphy:
### Describe the bug

For a number of reasons related to platform integration, nytimes wraps sveltekit html output in a template that will occasionally hoist script tags into the `body` tag:

src/hooks.js
```javascript
/** @type {import('@sveltejs/kit').Handle} */
export async function handle({ event, resolve }) {
	return resolve(event, {
		transformPage: ({ html }) => {
                          // wraps the output in the nytimes renderer, which sometimes 
                          // hoists script tags contained in html out of the generated component html and into the <body> tag.
			return render_page_in_template(html);
		},
	});
}
```

Since [kit now assumes that the mount script is always an immediate child of the root mount node](https://github.com/sveltejs/kit/blob/28139749c4bf056d1e04f55e7f955da33770750d/packages/kit/src/runtime/server/page/render.js#L161), this means that kit will mount the `body` node instead of the parentNode defined in app.html template. 

In practice this means that if we wanted to embed a component generated by kit into the NYTs homepage (which hoists scripts), we're at risk of that component blowing out the entire DOM. 

I really like the ergonomics introduced in https://github.com/sveltejs/kit/pull/3674, but I wonder if instead setting `data-hydrate` on `<script  data-hydrate="foo" />` we set it on a tag that's not vulnerable to hoisting `<div  data-hydrate="foo" />`

Thanks so much for all the amazing work y'all have done so far. And thanks in advance for your wisdom on this. 

### Reproduction

In src/hooks.js write a small script that hoists kit script tags out of their default parent node. 

### System Info

```shell
System:
    OS: macOS 12.3.1
    CPU: (10) x64 Apple M1 Max
    Memory: 2.44 GB / 64.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.18.3 - ~/bin/nvm/versions/node/v14.18.3/bin/node
    Yarn: 1.22.17 - /opt/homebrew/bin/yarn
    npm: 6.14.15 - ~/bin/nvm/versions/node/v14.18.3/bin/npm
  Browsers:
    Chrome: 100.0.4896.127
    Firefox: 99.0
    Safari: 15.4
  npmPackages:
    @sveltejs/adapter-static: 1.0.0-next.29 => 1.0.0-next.29 
    @sveltejs/kit: 1.0.0-next.305 => 1.0.0-next.305 
[...]

PR #4698, opened 2022-04-22
make hydration resilient to hoisted `<script>` elements

Rich-Harris:
This is a possible solution to #4685, though in the process of working on it I realised it's probably easier to solve it in userland — will add details on the issue

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/apps/basics/test/test.js
+++ b/packages/kit/test/apps/basics/test/test.js
@@ -2459,3 +2459,12 @@ test.describe.parallel('XSS', () => {
 		);
 	});
 });
+
+test.describe.parallel('Hydration', () => {
+	test('Hydration survives scripts being hoisted', async ({ page, javaScriptEnabled }) => {
+		await page.goto('/hoisted-script');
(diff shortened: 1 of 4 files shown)


packages/kit/test/apps/basics/test/test.js
```diff
@@ -2459,3 +2459,12 @@ test.describe.parallel('XSS', () => {
 		);
 	});
 });
+
+test.describe.parallel('Hydration', () => {
```
john-michael-murphy, line 2463: 🎉

benmccann:
It seems like solving this in userland is probably the better approach. If I understand correctly, NYT modifies all `<script>` tags and that breaks SvelteKit. The solution here is moving some stuff from a `<script>` tag to a `<span>` tag so NYT's code won't clobber it. But then what do we do if someone wants to write code that modifies all `<span>` tags? I think it would be better for the NYT code to blacklist the SvelteKit `<script>` tag and write a test to ensure that if SvelteKit's code ever changes it that they catch it

PR #4707, opened 2022-04-22
[chore] add Node 18 to CI

benmccann:
Can't drop Node 14 support yet because AWS Lambda still doesn't support Node 16


Rich-Harris (approved):

Rich-Harris:
Lambda is the new IE

Mlocik97:
well Node 14 has still support until `2023-04-30` so there is whole 1 year for them to move to newer version.

Issue #3557, opened 2022-01-26
CSP support for AMP pages


Rich-Harris:
### Describe the problem

We added CSP support in #3499, but it didn't apply to AMP pages.

### Describe the proposed solution

I think we just have to add [this monstrosity](https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/secure-pages/) but I didn't look too closely yet

### Importance

nice to have

Rich-Harris:
This... isn't actually possible. If you follow the instructions in https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/secure-pages/, your page won't validate:

<img width="659" alt="image" src="https://user-images.githubusercontent.com/1162160/169090859-6187dd07-8a98-4fbb-bb17-69668297a306.png">

AMP is a trainwreck.

PR #4710, opened 2022-04-23
remove built-in AMP support

### LINKED ISSUE #1308: AMP rendered as a separate page with canonical link
I want to be able to generate a separate page for every single one of my pages in my project. The reason is, I want to avoid interactive components when the customer views in AMP and show the fully-fledged page when in normal mode.

Eg:
AMP page: https://sandeep.dev/amp/choosetochallenge-removing-the-subconscious-bias
Normal page: https://sandeep.dev/choosetochallenge-removing-the-subconscious-bias

You can see that the amp page is served from an amp folder

Additional property for the amp option in config which can mention the custom folder to which it will be compiled to and a global variable to define {#if} blocks to avoid certain components during amp compilation. 

Svelte kit should be able to automatically include canonical like back to the main site from amp page and also mention the `amphtml` link in the main page.

https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/discovery/

Our application will need amp pages or it is a dealbreaker.

### LINKED ISSUE #2603: Move amp support to a plugin

### LINKED ISSUE #4029: AMP hybrid mode just like next.js
### Describe the problem

We have a big website that is built using Svelte, everything is fine and works perfect so thank you svelte <3

But, now, we have a business need where we need some pages to be "Google AMP".

We've tried `amp:true` so far but it forces the entire website to become an AMP. 

Is there a way you are adding this feature soon? Otherwise we are thinking about using the endpoints and serving custom AMP but could we get some general guidelines or recommendations if possible?

We would love to have a hybrid mode to activate AMP only on some routes. 

### Alternatives considered

Using endpoints to serve our own AMP code but meh. 

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
* fixes #2603
* fixes #1308
* fixes #4029

This removes the `amp: true` config option in favour of a more appropriate solution (we despise AMP and want to hasten its demise, but still need to provide _some_ way for people to deal with it).

It adds a new `@sveltejs/kit/amp` module which exports a `transform` function, which:

* [x] removes all `<script>` tags
* [x] ensures there are no `<link rel="stylesheet">` elements
* [x] adds an `amp-custom` attribute to `<style>`
* [x] removes `<meta http-equiv>` elements
* [x] adds the AMP boilerplate
* [ ] ~~adds `<amp-install-serviceworker>` if appropriate~~
* [x] CSP #3557

Some of these are TODO and I haven't quite figured out how to solve them. Also TODO:

* [x] docs

Using it is relatively straightforward — set `inlineStyleThreshold` to `Infinity` (external stylesheets are disallowed), and add the following to `src/hooks.js`:

```js
import * as amp from '@sveltejs/kit/amp';

  return resolve(event, {
    transformPage: ({ html }) => amp.transform(html);
  });
}
```

One thing we lose is the ability to exclude CSS from components that are imported but not rendered for a given page. Given AMP's wholly arbitrary CSS size limit, that can be a problem. This can be solved (at least for prerendered pages, where it doesn't matter how expensive this sort of processing is) with something like https://github.com/purifycss/purifycss:

```js
import purify from 'purify-css';
import * as amp from '@sveltejs/kit/amp';

  return resolve(event, {
    transformPage: ({ html }) => {
      html = amp.transform(html);

      // remove unused CSS
      let css = '';
      const markup = html.replace(
        /<style([^>]*?)>([^]+?)<\/style>/,
        (match, attributes, contents) => {
          css = contents;
          return `<style${attributes}></style>`;
        }
      );

      css = purify(markup, css);
      return markup.replace('</style>', `${css}</style>`);
    }
  });
}
```

It's not quite as turnkey as the current solution, but it's fine.

Another thing we lose is validation. This, too, can easily be solved in userland by using `amphtml-validator` directly.

### Tests

### Changesets


Rich-Harris:
cc @john-michael-murphy as I know this'll probably affect you lot

john-michael-murphy:
Migration looks doable. Thank you for the heads up. AMP is such a thorn!

Rich-Harris:
There's no good way to do automatic service worker installation (the transformer can't read the config, or `paths.base` etc), so i'm going to leave that to users, who can easily inject `<amp-install-serviceworker>` themselves

documentation/docs/16-seo.md
```diff
@@ -82,8 +82,40 @@ export async function get() {
...
+		// since <link rel="stylesheet"> isn't
+		// allowed, inline all styles
+		inlineStyleThreshold: Infinity
+	}
+};
+
+export default config;
+```
+
+...and transforming the HTML using `transformPage` along with `amp.transform` imported from `@sveltejs/kit/amp`:
```
Conduitry, line 108:
```suggestion
...and transforming the HTML using `transformPage` along with `transform` imported from `@sveltejs/kit/amp`:
```

I'd just refer to this as `transform`, since that's the name of the export.

Rich-Harris:
i moved `@sveltejs/kit/amp` to `@sveltejs/kit-amp` — @benmccann made a good point about potentially bundling the validator without penalising people who install `@sveltejs/kit`

.npmrc
```diff
@@ -1,2 +1,2 @@
 link-workspace-packages = true
-engine-strict = true
+engine-strict = true
```
benmccann, line 2:
```suggestion
engine-strict = true

```

benmccann (commented):
having kit in the name `kit-amp` seems a bit redundant. maybe just `amp`?

Rich-Harris:
> having kit in the name `kit-amp` seems a bit redundant. maybe just `amp`?

originally i was thinking that we might end up having kit-specific stuff in there, but i'm inclined to do this just because my muscles are so used to `cd pac[tab]/k[tab]/t[tab]/a[tab]/b[tab` taking me to `packages/kit/test/apps/basics` and it doesn't any more, and it will make me resent AMP even more than i currently do

PR #4712, opened 2022-04-23
Throwable responses from endpoints

coryvirok:
The idea here is that endpoints can not only return a response, but also throw one.

E.g.

```js
  // return a response from an endpoint
  export async function get() {
    return { status: 200, body: {} }
  }

  // throw a response from an endpoint
  export async function get() {
    throw { status: 401 }
  }
```

This is useful for building request validation libraries. E.g. to build an auth guard
for an endpoint without this change, you would need something like:

```js
  export async function get() {
    if (!isAuthenticated()) {
      return { status: 401 }
    }
  }
```

And with this change, you can do:

```js
  export async function get() {
    // Throws a 401 if the request is not authenticated
    verifyAuthenticated()
  }
```

This is a common pattern in other web frameworks and makes it easier to structure
your code.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/endpoint.js
+++ b/packages/kit/src/runtime/server/endpoint.js
@@ -75,7 +75,22 @@ export async function render_endpoint(event, mod) {
 			  });
 	}
 
-	const response = await handler(event);
+	let response;
+	try {
+		response = await handler(event);
+	} catch (err) {
+		const parsed_status = parseInt(err.status);
+		if (isNaN(parsed_status)) {
+			throw err;
+		}
+		response = {
+			status: parsed_status,
+			headers: err.headers,
+			body: err.body,
+			error: err.error
+		};
+	}
+
 	const preface = `Invalid response from route ${event.url.pathname}`;
 
 	if (typeof response !== 'object') {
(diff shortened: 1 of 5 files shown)


Rich-Harris:
Thanks for the PR. We discussed this a bit in the maintainer's discord and decided against merging this, on the grounds that `throw` isn't a control flow mechanism — the endpoint with `isAuthenticated()` is much more understandable than the one with `verifyAuthenticated()`. It also makes things more complicated — https://kit.svelte.dev/docs/hooks#handleerror would need to change:

> If an error is thrown during rendering, this function will be called with the error and the event that caused it, **unless the error has a numeric `status` property, in which case it will be treated as an expected error rather than an unexpected one**. This allows you to send data to an error tracking service, or to customise the formatting before printing the error to the console.

The modest convenience doesn't outweigh the potential for confusion.

A better solution might be to add a helper function like this:

```js
// src/lib/authed.js
export function authed(fn) {
  return event => {
    if (!event.session.user) {
      return { status: 401 };
    }

    return fn ? fn(event) : {};
  }
}
```

```js
// src/routes/whatever.js
import { authed } from '$lib/authed';

export const get = authed(); // guard without doing any actual work in the endpoint

export const get = authed(async event => {
  // in here, we know that `event.session.user` is populated
  return {...};
});
```

coryvirok:
Alright. Thanks for the consideration and response. I'll go with the pattern you described or possibly use middleware/endpoint layouts when they're available.

coryvirok:
For my own reference, the pattern I suggested here was essentially implemented via the error helper - https://kit.svelte.dev/docs/errors#expected-errors

Issue #1329, opened 2021-05-04
Path handling during build with adapter-static


mvolfik:
**Describe the bug**
Magic happens. Build fails with weird 404 when `paths: { base: "/" }` and there's `<a href="/">`, and creates a directory `buildfavicon.ico` instead of nonexistent file `build/favicon.ico`.

<details>
<summary><b>Logs</b></summary>

```
yarn run v1.22.10
warning package.json: No license field
$ svelte-kit build
vite v2.2.4 building for production...
✓ 14 modules transformed.
.svelte/output/client/_app/manifest.json                    1.18kb
.svelte/output/client/_app/pages/index.svelte-31924b7e.js   0.41kb / brotli: 0.26kb
.svelte/output/client/_app/layout.svelte-f07a7e7f.js        0.49kb / brotli: 0.29kb
.svelte/output/client/_app/assets/start-a8cd1609.css        0.16kb / brotli: 0.10kb
.svelte/output/client/_app/error.svelte-6645eba8.js         1.19kb / brotli: 0.57kb
.svelte/output/client/_app/chunks/vendor-43cd886a.js        5.05kb / brotli: 1.98kb
.svelte/output/client/_app/start-fec3b019.js                16.00kb / brotli: 5.12kb
vite v2.2.4 building SSR bundle for production...
✓ 9 modules transformed.
.svelte/output/server/app.js   8.86kb

Run npm run preview to preview your production build locally.

> Using @sveltejs/adapter-static
> 404  (linked from /)
Error: 404  (linked from /)
    at error (file:///tmp/test/node_modules/@sveltejs/kit/dist/chunks/index5.js:92:11)
    at visit (file:///tmp/test/node_modules/@sveltejs/kit/dist/chunks/index5.js:159:5)
    at async visit (file:///tmp/test/node_modules/@sveltejs/kit/dist/chunks/index5.js:219:6)
    at async prerender (file:///tmp/test/node_modules/@sveltejs/kit/dist/chunks/index5.js:228:5)
    at async Object.prerender (file:///tmp/test/node_modules/@sveltejs/kit/dist/chunks/index5.js:295:5)
    at async adapt (/tmp/test/node_modules/@sveltejs/adapter-static/index.cjs:10:4)
    at async adapt (file:///tmp/test/node_modules/@sveltejs/kit/dist/chunks/index5.js:321:2)
    at async file:///tmp/test/node_modules/@sveltejs/kit/dist/cli.js:590:5
[...]

PR #4713, opened 2022-04-23
clarify that base path can be the empty string

Rich-Harris:
Apparently some people get confused by this (e.g. #1329)

### Tests

### Changesets

### DIFF
--- a/.changeset/old-cars-switch.md
+++ b/.changeset/old-cars-switch.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Tweak error message for non-validating base paths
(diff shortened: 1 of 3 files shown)


benmccann (approved):

Conduitry:
This broke tests for me in https://github.com/sveltejs/kit/tree/master/packages/kit/src/core/config/index.spec.js (which I would expect because it changed an error message, and they're now looking for the wrong one), but CI is still passing, which is bugging me.

mrkishi:
Apparently the tests [haven't been running](https://github.com/sveltejs/kit/actions/runs/2213292708) since https://github.com/sveltejs/kit/pull/4707, the green/neutral checks are Netlify/Cloudflare/Vercel checks.

I think `include` should be an array instead, #4727 seems to be running the tests.

Conduitry:
Ah, thank you!

PR #4726, opened 2022-04-24
explain how to make credentialled fetches in server-side code

### LINKED ISSUE #3380: Credentialed `fetch` for hooks
### Describe the problem

I'm a bit confused about how authentication is done in SvelteKit as I couldn't find sufficient information in the docs, so I think I need some guidance.

I have an ASP.NET Core back-end application, which acts as my API and will be served on a subdomain in production (e.g. `api.my-domain.com`), and a SvelteKit front-end application which would be the main site (e.g. `my-domain.com`).

Authentication is already implemented in the back-end project, once a user logs in, a cookie is generated that contains a token (sort of like JWT). Any subsequent requests that carry that cookie will be authenticated. Pretty straightforward stuff.

Now, in the docs, in the section about the `load` function's `fetch`, it's mentioned that:
> Cookies will only be passed through if the target host is the same as the SvelteKit application or a more specific subdomain of it. [(Here)](https://kit.svelte.dev/docs#:~:text=additional%20network%20request.-,Cookies%20will%20only%20be%20passed%20through%20if%20the%20target%20host%20is%20the%20same%20as%20the%20SvelteKit%20application%20or%20a%20more%20specific%20subdomain%20of%20it.,-session)

And this works beautifully. When a load function sends a fetch request to the same target host, the client's credentials are properly passed through. As a result, when, for example, in a layout's `load` function I send a fetch request to my back-end API to retrieve, say, the current user's authentication state and information, it works as expected during SSR, that is, the auth cookie is passed through and the right information is retrieved. The fetch request that takes place on the server, in a sense, **emulates** a fetch request that's sent directly from the user's browser (in terms of credentials).

But I also noticed that SvelteKit has something called the [`getSession` hook](https://kit.svelte.dev/docs#hooks-getsession) which is simply a function that's run on every request and can be used to retrieve information about the current request, including the current user's authentication status, and that information is then put in a `session` store that is going to be available to use in pages and components.
In the `getSession` hook, however, I tried to send a fetch request to my back-end API to retrieve some information about the current user, but the problem is I need the client's credentials to be passed through, just like what happens in `load` functions (provided that you used the special `fetch` that's received as an argument, of course). But there's no such special `fetch` in hooks, and by using the global, plain `fetch` there, it's as though the server is sending the request and therefore the client's credentials aren't passed through.

Couldn't hook functions also receive a special `fetch` that passes the client's credentials through and effectively emulates the client? Just like how the load function's `fetch` works. This would enable scenarios like this.

If there is already a straightforward workaround for this, or I'm missing something obvious, let me know.

Thank you.

Rich-Harris:
Closes #3380. Rather than introducing a new version of `fetch`, I think it makes sense to just document the fact that the way to make credentialled fetches in server-side code is to pass cookie/auth headers.

### DIFF
--- a/documentation/docs/04-loading.md
+++ b/documentation/docs/04-loading.md
@@ -79,7 +79,7 @@ If the page you're loading has an endpoint, the data returned from it is accessi
 
 #### fetch
 
-`fetch` is equivalent to the native `fetch` web API, with a few additional features:
+`fetch` is equivalent to the [native `fetch` web API](https://developer.mozilla.org/en-US/docs/Web/API/fetch), with a few additional features:
 
 - it can be used to make credentialed requests on the server, as it inherits the `cookie` and `authorization` headers for the page request
 - it can make relative requests on the server (ordinarily, `fetch` requires a URL with an origin when used in a server context)
--- a/documentation/docs/04-loading.md
+++ b/documentation/docs/04-loading.md
@@ -79,9 +79,13 @@ If the page you're loading has an endpoint, the data returned from it is accessi
 
 #### fetch
 
-`fetch` is equivalent to the native `fetch` web API, and can make credentialed requests. It can be used across both client and server contexts.
+`fetch` is equivalent to the native `fetch` web API, with a few additional features:
 
-> When `fetch` runs on the server, the resulting response will be serialized and inlined into the rendered HTML. This allows the subsequent client-side `load` to access identical data immediately without an additional network request.
+- it can be used to make credentialed requests on the server, as it inherits the `cookie` and `authorization` headers for the page request
+- it can make relative requests on the server (ordinarily, `fetch` requires a URL with an origin when used in a server context)
+- requests for endpoints go direct to the handler function during server-side rendering, without the overhead of an HTTP call
+- during server-side rendering, the response will be captured and inlined into the rendered HTML
+- during hydration, the response will be read from the HTML, guaranteeing consistency and preventing an additional network request
 
 > Cookies will only be passed through if the target host is the same as the SvelteKit application or a more specific subdomain of it.
 
--- a/documentation/docs/01-web-standards.md
+++ b/documentation/docs/01-web-standards.md
@@ -12,7 +12,7 @@ In particular, you'll get comfortable with the following:
 
 SvelteKit uses [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/fetch) for getting data from the network. It's available in [hooks](/docs/hooks) and [endpoints](/docs/routing#endpoints) as well as in the browser.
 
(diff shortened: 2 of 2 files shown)


georgecrawford:
Am I right that there are other differences between the globally-available fetch and the version passed to `load()` as well? For example:

```js
<script context="module">
    const oldFetch = fetch;

    export async function load({fetch}) {
        await fetch('/test');
        await oldFetch('/test');
    }
</script>
```

The first of these works fine with SSR, whilst the second throws:
```
TypeError [ERR_INVALID_URL]: Invalid URL
    at new NodeError (node:internal/errors:371:5)
    at onParseError (node:internal/url:552:9)
    at parse (<anonymous>)
    at new URL (node:internal/url:628:5)
    at new Request (file:///app/server/node_modules/@sveltejs/kit/dist/install-fetch.js:5907:16)
    at file:///app/server/node_modules/@sveltejs/kit/dist/install-fetch.js:6189:19
    at new Promise (<anonymous>)
    at fetch (file:///app/server/node_modules/@sveltejs/kit/dist/install-fetch.js:6187:9)
    at load (/src/routes/__layout.svelte:38:8)
    at load_node (file:///app/server/.svelte-kit/runtime/server/index.js:1864:30) {
  input: '/test',
  code: 'ERR_INVALID_URL'
}
```

Could this, and any other differences, be added to the docs in this MR please? I just don't know what they are, or why they exist! 

Thank you

Rich-Harris:
done!

benmccann:
I wonder how to deal with the fact that we have this section about `fetch` as well as the one in https://kit.svelte.dev/docs/loading#input-fetch. The latter says "fetch is equivalent to the native fetch web API", but then here we're listing some ways that it's not so we might need to update it with a caveat that points to this section?

georgecrawford:
I’d absolutely agree - this was the source of some of my confusion, as it’s where I first encountered documentation for fetch as a param to load(). 

________________



George

itssumitrai:
While not related to this, but there is a difference I found between the LoadInput `fetch` and the global `fetch`. I am not able to abort the fetch request with `AbortController` for the LoadInput `fetch`.

Rich-Harris:
@itssumitrai 

> While not related to this

if you have a separate issue, please raise a separate issue :)

documentation/docs/04-loading.md
```diff
@@ -79,9 +79,13 @@ If the page you're loading has an endpoint, the data returned from it is accessi
 
 #### fetch
 
-`fetch` is equivalent to the native `fetch` web API, and can make credentialed requests. It can be used across both client and server contexts.
+`fetch` is equivalent to the native `fetch` web API, with a few additional features:
```
mrkishi, line 82:
```suggestion
`fetch` is equivalent to the [native `fetch` web API](https://developer.mozilla.org/en-US/docs/Web/API/fetch), with a few additional features:
```

mrkishi (approved):

PR #4727, opened 2022-04-25
fix ci tests

mrkishi:

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/config/index.spec.js
+++ b/packages/kit/src/core/config/index.spec.js
@@ -240,7 +240,7 @@ test('fails if paths.base is not root-relative', () => {
 				}
 			}
 		});
-	}, /^config\.kit\.paths\.base option must be a root-relative path that starts but doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs\/configuration#paths$/);
+	}, /^config\.kit\.paths\.base option must either be the empty string, or a root-relative path that starts but doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs\/configuration#paths$/);
 });
 
 test("fails if paths.base ends with '/'", () => {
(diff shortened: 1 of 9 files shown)


packages/kit/src/core/preview/index.js
```diff
@@ -161,7 +161,7 @@ export async function preview({ port, host, config, https: use_https = false })
 	});
 
 	return new Promise((fulfil) => {
-		http_server.listen(port, host || '0.0.0.0', () => {
+		http_server.listen(port, host, () => {
```
Conduitry, line 164: Just to confirm: When the user passes `--host` with no argument (which indicates they want to bind on all interfaces), `host` will be an empty string, and passing `''` to `.listen` will have that effect?

mrkishi: Passing `''` would be equivalent to `undefined` and results in `0.0.0.0` or `::`, depending on the network support (and I just tested to confirm it). The difference in this case is that we will bind all ipv6 and ipv4 interfaces, whereas before we only did ipv4.

Conduitry: :+1: Sounds good. I think that warrants a changeset, and then this is ready to go!

packages/kit/src/core/config/index.spec.js
```diff
@@ -240,7 +240,7 @@ test('fails if paths.base is not root-relative', () => {
 				}
 			}
 		});
-	}, /^config\.kit\.paths\.base option must be a root-relative path that starts but doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs\/configuration#paths$/);
+	}, /^config\.kit\.paths\.base option must either be the empty string, or a root-relative path that starts but doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs\/configuration#paths$/);
```
benmccann, line 243:
```suggestion
	}, /^config\.kit\.paths\.base option must either be the empty string or a root-relative path that starts but doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs\/configuration#paths$/);
```

packages/kit/src/core/config/index.spec.js
```diff
@@ -252,7 +252,7 @@ test("fails if paths.base ends with '/'", () => {
...
```
benmccann, line 255:
```suggestion
	}, /^config\.kit\.paths\.base option must either be the empty string or a root-relative path that starts but doesn't end with '\/'. See https:\/\/kit\.svelte\.dev\/docs\/configuration#paths$/);
```

Conduitry (approved):

PR #4728, opened 2022-04-25
[chore] simplify env GH action

Conduitry:
`.github/actions/env` has been bugging me for a while. We had built code that needed to be checked in - and which apparently hadn't been kept up to date in a while, because whenever I ran `pnpm build` locally, it changed this file.

We'll see what happens when CI runs, but this attempts to replace that whole file with a re-creation of just the functionality we need, which lets us drop the build process for the file and remove the whole `@actions/core` dependency.

### Tests

### Changesets


Conduitry:
After #4713 is resolved, I'll merge master back in here and we'll see what happens.

mrkishi (approved):

pngwn (approved):

PR #4731, opened 2022-04-25
[adapter-vercel] Make all root routes be called `index` instead of an empty string

Schniz:
The current implementation of `edge: true, split: true` will fail to deploy on Vercel because the Edge Function name is empty. This PR fixes that by renaming them to `index` instead.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-vercel/index.js
+++ b/packages/adapter-vercel/index.js
@@ -340,7 +340,7 @@ async function v3(builder, external, edge, split) {
 						.slice(1, -2) // remove leading / and trailing $/
 						.replace(/\\\//g, '/')}(?:/__data.json)?$`; // TODO adding /__data.json is a temporary workaround — those endpoints should be treated as distinct routes
 
-					await generate_function(route.id, src, entry.generateManifest);
+					await generate_function(createValidFunctionName(route.id), src, entry.generateManifest);
 				}
 			};
 		});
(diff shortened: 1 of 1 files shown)


packages/adapter-vercel/index.js
```diff
@@ -340,7 +340,7 @@ async function v3(builder, external, edge, split) {
 						.slice(1, -2) // remove leading / and trailing $/
 						.replace(/\\\//g, '/')}(?:/__data.json)?$`; // TODO adding /__data.json is a temporary workaround — those endpoints should be treated as distinct routes
 
-					await generate_function(route.id, src, entry.generateManifest);
+					await generate_function(createValidFunctionName(route.id), src, entry.generateManifest);
```
Rich-Harris, line 343:
```suggestion
					await generate_function(route.id || 'index', src, entry.generateManifest);
```

packages/adapter-vercel/index.js
```diff
@@ -385,3 +385,20 @@ function write(file, data) {
...
+ * createValidFunctionName("nested") // "nested"
+ * createValidFunctionName("nested/") // "nested/index"
+ * ```
+ *
+ * @param {string} routeId
+ * @returns {string} route id that is a valid function name
+ */
+function createValidFunctionName(routeId) {
+	return routeId.replace(/(\/|^)$/, '$1index');
+}
```
Rich-Harris, line 404:
```suggestion
```

Rich-Harris (commented):
thanks! it's actually way simpler — a route ID never ends with `/` (i.e. there's no possibility of separate `nested` and `nested/` routes), so the only case we need to consider is the empty string

.changeset/shiny-needles-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/adapter-vercel': patch
+---
+
+Fix `edge:true, split: true` build error for root routes when deploying to Vercel
```
Conduitry, line 5:
```suggestion
Fix `edge: true, split: true` build error for root routes when deploying to Vercel
```

Rich-Harris (approved):

PR #4739, opened 2022-04-25
update rollup+plugins

Rich-Harris:
since the eager-but-dumb renovate bot is going to pester us about `@rollup/plugin-commonjs` otherwise, this PR updates Rollup itself alongside its plugins. 

I also took the opportunity to tidy up the `adapter-begin` package, since it no longer needs Rollup (or indeed anything) given that it just points to a different package.

Supersedes #4715.


Rich-Harris:
Updating Rollup to 2.70 somehow causes the docs to fail typechecking. I don't have the energy for this bullshit

PR #4742, opened 2022-04-25
default to target: node14.8

### LINKED ISSUE #1538: Hook for initialization?

Rich-Harris:
Fixes #1538, where I recommended that people do async setup work in `src/hooks.js` (or any other module, really) using top-level await. This works wonderfully in development, but it turns out `svelte-kit build` fails because Vite runs everything through `esbuild` (not sure why, if I'm perfectly honest, but that's a separate conversation) and we currently use `target: 'es2020'` which doesn't allow top-level await:

> chunks/hooks-4fb5bb73.js:206:0: ERROR: Top-level await is not available in the configured target environment ("es2020")

This changes it to `node16`. I tried `node14` at first, but that still didn't work even though top-level await definitely _is_ supported there. I think this is probably fine, and at any rate, you can override it in `config.kit.vite.build.target`.

### Tests

### Changesets

### DIFF
--- a/.changeset/hungry-coats-sort.md
+++ b/.changeset/hungry-coats-sort.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Default to target: node16, so that top-level await is supported in user code
--- a/.changeset/hungry-coats-sort.md
+++ b/.changeset/hungry-coats-sort.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Default to target: node16, so that top-level await is supported in user code
+Default to target: node14.8, so that top-level await is supported in user code
--- a/packages/kit/src/core/build/build_server.js
+++ b/packages/kit/src/core/build/build_server.js
@@ -190,7 +190,7 @@ export async function build_server(
 
 	const default_config = {
 		build: {
-			target: 'es2020'
+			target: 'node16'
 		},
 		ssr: {
 			// when developing against the Kit src code, we want to ensure that
--- a/packages/kit/src/core/build/build_server.js
+++ b/packages/kit/src/core/build/build_server.js
@@ -190,7 +190,7 @@ export async function build_server(
 
 	const default_config = {
 		build: {
-			target: 'node16'
+			target: 'node14.8'
 		},
 		ssr: {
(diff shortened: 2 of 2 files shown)


bluwy:
> Vite runs everything through `esbuild`

It's using `esbuild` in the `renderChunk()` phase for minification. ([source](https://github.com/vitejs/vite/blob/3edd1af56e980aef56641a5a51cf2932bb580d41/packages/vite/src/node/plugins/esbuild.ts#L228-L258))

> I tried `node14` at first, but that still didn't work even though top-level await definitely _is_ supported there

esbuild keeps a[ compat table for top level await](https://github.com/evanw/esbuild/blob/e8768569a042be09be0b01445fbe399ed043793b/internal/compat/js_table.go#L494-L501). It only exists in 14.8, so specifying `node14.8` should work too.

Rich-Harris:
> specifying `node14.8` should work too

nice, it does

bluwy (approved):
Looks like the `package.json` have `"node": ">=14.13"` so we should be safe with the change too.

.changeset/hungry-coats-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Default to target: node16, so that top-level await is supported in user code
```
Conduitry, line 5:
```suggestion
Default to target: node14.8, so that top-level await is supported in user code
```

Mlocik97: some mismatch but I think we don't want to rewrite version in this part, instead set target in config to node16.

PR #4749, opened 2022-04-26
use spaces in README.md files

Rich-Harris:
In general we use tabs everywhere, because we are sensible. GitHub's CSS, however, is _not_ sensible:

image

This PR adjusts the Prettier config so that we can link to individual package READMEs as documentation without feeling embarrassed. (I thought `.editorconfig` was supposed to fix this, which would be better since code copied from the READMEs would contain tabs, but clearly it doesn't.)

image

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -11,6 +11,13 @@
 			}
 		},
 		{
+			"files": ["packages/*/README.md"],
+			"options": {
+				"useTabs": false,
+				"tabWidth": 2
+			}
+		},
+		{
 			"files": [
 				"**/CHANGELOG.md",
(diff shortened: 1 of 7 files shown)


geoffrich:
FWIW, you can customize your tab size in GitHub settings: https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-user-account-settings/managing-your-tab-size-rendering-preference

Rich-Harris:
> FWIW, you can customize your tab size in GitHub settings

yep. this is less about making it look nice for _me_, more about making it look nice for everyone we send to those readmes. i expect very few people bother updating those settings

PR #4751, opened 2022-04-26
add "did you mean" 404 response when requesting unprefixed path in `svelte-kit preview`

### LINKED ISSUE #2838: Make dev & preview be friendlier for base!='/'

Rich-Harris:
closes #2838 by showing the same messages in `svelte-kit preview` as in `dev` when visiting a pathname that doesn't begin with `config.kit.paths.base`

### Tests

### Changesets


benmccann (approved):
note that https://github.com/sveltejs/kit/pull/4761 is changing the logic in dev mode so if that PR is merged first then this one will need to be updated

PR #4757, opened 2022-04-27
[chore] disallow test.only tests during linting

Conduitry:
We can't (yet) tell uvu to disallow `test.only` tests (https://github.com/lukeed/uvu/issues/174), so I'd like to prevent this during linting instead.

The first commit of this PR purposefully does not fix the existing `test.only` so that we can make sure it's caught by CI. Once that happens, I'll push another commit to fix it.

### Tests

### Changesets

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,5 +6,11 @@
 	},
 	"settings": {
 		"import/ignore": "/template/"
+	},
+	"rules": {
+		"no-restricted-properties": [
+			"error",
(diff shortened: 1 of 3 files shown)


Conduitry:
Yup, caught by CI: https://github.com/sveltejs/kit/runs/6186484274

I'm pushing another commit to fix the linting error.

I'm not sure why there are _two_ ESLint config files - one for `packages/kit` and one for everything else - nor why the Kit one purposefully doesn't inherit from the top-level one - but we can worry about that another time, and for now I've just put the new rule in both.

PR #4760, opened 2022-04-27
pass config to vite-plugin-svelte instead of reloading it

### LINKED ISSUE #4909: Gracefully recover from malformed svelte.config.js during development
### Describe the problem

At present if I make changes to `svelte.config.js`, the during development, the DEV server tries to load it and crashes because I've set my editor to autosave and have not yet completed typing. This is frustrating, especially when working on a Vite plugin or configuring other build scripts that require frequent changes.

If the config load succeeded before, reuse the old config and a present a warning about the malformed `svelte.config.js` in the terminal. I assume it's as simple as a adding a try-catch somewhere, though I may be wrong.

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
At present, `svelte.config.js` is loaded multiple times — once by SvelteKit, then again by vite-plugin-svelte (twice, during `svelte-kit build`). Aside from being marginally less efficient than reusing the config, this can result in surprising behaviour if you're doing any kind of setup work in the config file.

This uses the new `configFile: false` option and passes config directly to the plugin instead.

It also fixes #4909.

### Tests

### Changesets


packages/kit/src/core/build/build_server.js
```diff
@@ -232,9 +232,12 @@ export async function build_server(
...
 			svelte({
 				extensions: config.extensions,
+				preprocess: config.preprocess,
 				compilerOptions: {
+					...config.compilerOptions,
 					hydratable: !!config.kit.browser.hydrate
-				}
+				},
+				configFile: false
 			})
```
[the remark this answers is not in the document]
dominikg: yeah, sveltekit config validation fortunately allows unknown keys and copies them along, the above should work. 

Even when we move the vite plugin options to their own key, the config file is the only place for users to customize vite-plugin-svelte in kit, unless we kind of duplicate the options inside the kit namespace or seal it off and only pass through a subset

packages/kit/src/core/dev/index.js
```diff
@@ -60,15 +60,18 @@ export async function dev({ cwd, port, host, https, config }) {
 		plugins: [
 			svelte({
 				extensions: config.extensions,
+				preprocess: config.preprocess,
```
dominikg, line 63:
```suggestion
				...config,
```

packages/kit/src/core/build/build_server.js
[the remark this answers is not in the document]
dominikg: damnit, v-p-s is going to complain about the inline `kit` key. need to add that to the allowed list.

Rich-Harris:
this should be good to merge. it now gracefully restarts the entire dev server on config changes

packages/kit/package.json
```diff
@@ -10,7 +10,7 @@
 	"homepage": "https://kit.svelte.dev",
 	"type": "module",
 	"dependencies": {
-		"@sveltejs/vite-plugin-svelte": "^1.0.0-next.32",
+		"@sveltejs/vite-plugin-svelte": "1.0.0-next.44",
```
dominikg, line 13: is the fixed version intentional?

Rich-Harris: nope!

```suggestion
		"@sveltejs/vite-plugin-svelte": "^1.0.0-next.44",
```

bluwy (approved):

PR #4773, opened 2022-04-29
fix: `sveltekit:prefetch` loads wrong page with anchor links (#4768)

### LINKED ISSUE #4768: `sveltekit:prefetch` potentially loads a wrong page when an anchor link is involved
### Describe the bug

When using a link with `sveltekit:prefetch` that contains an anchor-link e.g. `'/blog#modal'` the page gets requested on hover, which is the intended behaviour.
But if you then decide to not click on that link and visit the same page without the anchor-link `/blog` you will see the contents of the link you hovered before.

In this implementation the `onMount` hook accesses the `$page` store which holds a wrong value.

I would assume when using `prefetch` sveltekit only loads the necessary `load` functions and components.
The `$page` store should get updated when you actually click on the link.

### Reproduction

Here is the link to a Repo showing the issue: https://stackblitz.com/edit/sveltejs-kit-template-default-uubxoo

1. load the homepage
2. hover over one of the last two links
3. click the first link

You should see that the page was loaded with some content that belongs to the other links

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.22000
    CPU: (12) x64 Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz
    Memory: 16.90 GB / 31.73 GB
  Binaries:
    Node: 16.13.2 - ~\AppData\Local\Volta\tools\image\node\16.13.2\node.EXE
    Yarn: 1.22.17 - ~\AppData\Local\Volta\tools\image\yarn\1.22.17\bin\yarn.CMD
    npm: 8.1.2 - ~\AppData\Local\Volta\tools\image\node\16.13.2\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (100.0.1185.50), ChromiumDev (99.0.1141.0)
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.37 
    @sveltejs/kit: next => 1.0.0-next.320 
    svelte: ^3.46.0 => 3.47.0
```

### Severity

### Additional Information

_No response_

tcc-sejohnson:
Fixes #4768.

EDIT:

The fix was a bit more complicated than expected, but not very invasive. Essentially, if a route is preloaded, then the same route with a different hash is preloaded, we want to update the cached route's hash (so that when the navigation actually happens, the `page` store receives the correct route including the hash) but leave the rest of the preload cache untouched. The problem is that the lion's share of the cache is a promise, and we obviously don't want to `await` that promise to change the URL's `hash` attribute. Instead, we wrap the promise in another promise, which awaits the internal promise, changes the url's hash, and returns the load cache (wrapped in a promise, obviously).

The only downside I could see to this approach is that it may trigger some kind of stack overflow if you were to wrap the promise too many times (I imagine it's creating a new closure for every `prefetch`), but I added a test to prefetch 1000 times and it runs just fine.

I also added tests to ensure all of the assumptions are correct -- that identical hash routes are `prefetched` only once and that identical routes with different hashes are `prefetched` only once.

~~Added a test that fails if `/routing/prefetched/hash-route#please-dont-show-me` is prerendered, then `/routing/prefetched/hash-route` is visited and the page shows the hash route's content instead of the "naked" route's.~~

~~The problem was super hard to isolate, and it actually turned out there was more than one bug involved here, but the solution was only a one-line change. When `client.js` called `get_navigation_intent` while prefetching, it set the intent's `id` to `url.pathname + url.search`, meaning if the hash route changed, subsequent calls to `prerender` wouldn't fully run, as they'd see that `intent.id` hadn't changed. This resulted in two bugs:~~

~~1. When trying to prefetch different hash routes, only the first prefetch would ever be run -- i.e. if `/prefetch#one` was prerendered, then `/prefetch#two` was subsequently hovered, then `/prefetch` was hovered after that, only `/prefetch#one` would ever be prefetched.~~
~~2. As a result of 1, prefetching `/prefetch#one` meant that the internal `page.url` would be `/prefetch#one`. When the user actually clicked `/prefetch` or `/prefetch#two`, which triggers assignment of the internal `page` object to the `page` writable store, the `page` assigned would have the incorrectly-cached `/prefetch#one` url.~~

~~Changing the cache `id` to `url.pathname + url.hash + url.search` fixes the issues and does not break any other tests (at least, locally... we'll see how CI does). Maybe someone can think of some edge case this doesn't cover, but I can't.~~

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/apps/basics/src/routes/routing/prefetched/hash-route.svelte
+++ b/packages/kit/test/apps/basics/src/routes/routing/prefetched/hash-route.svelte
@@ -1,0 +1,27 @@
+<script>
+	import { onMount } from 'svelte';
+	import { page } from '$app/stores';
+
+	const modalContents = [
+		{
+			id: 'please-dont-show-me',
+			title: 'Oopsie'
+		}
+	];
+
+	let modal = undefined;
+
+	const checkIfModalShouldBeShown = () => {
+		console.log('mount');
+		const modalToShow = $page.url.hash.substring(1);
+		modal = modalContents.find(({ id }) => id === modalToShow);
+	};
+
+	onMount(checkIfModalShouldBeShown);
+</script>
+
+<svelte:window on:popstate={checkIfModalShouldBeShown} />
+
+<h1>
+	{modal?.title ?? ''}
+</h1>
--- a/packages/kit/test/apps/basics/src/routes/routing/prefetched/hash-route.svelte
+++ b/packages/kit/test/apps/basics/src/routes/routing/prefetched/hash-route.svelte
@@ -1,9 +1,9 @@
 <script context="module">
-	let calls = 0;
+	let loadCalls = 0;
 
 	export const load = () => {
-		calls += 1;
-		return { props: { calls } };
+		loadCalls += 1;
+		return { props: { calls: loadCalls } };
 	};
 </script>
 
--- a/packages/kit/test/apps/basics/src/routes/routing/prefetched/hash-route.svelte
+++ b/packages/kit/test/apps/basics/src/routes/routing/prefetched/hash-route.svelte
@@ -25,7 +25,6 @@
 	let modal = undefined;
 
 	const checkIfModalShouldBeShown = () => {
-		console.log('mount');
 		const modalToShow = $page.url.hash.substring(1);
 		modal = modalContents[modalToShow];
 	};
--- a/packages/kit/test/apps/basics/src/routes/routing/prefetched/hash-route.svelte
+++ b/packages/kit/test/apps/basics/src/routes/routing/prefetched/hash-route.svelte
@@ -1,20 +1,33 @@
+<script context="module">
+	let calls = 0;
+
+	export const load = () => {
+		calls += 1;
+		return { props: { calls } };
+	};
+</script>
+
 <script>
 	import { onMount } from 'svelte';
 	import { page } from '$app/stores';
 
-	const modalContents = [
-		{
-			id: 'please-dont-show-me',
+	export let calls;
+
+	const modalContents = {
+		'please-dont-show-me': {
 			title: 'Oopsie'
+		},
+		'please-dont-show-me-jr': {
+			title: 'Oopsie Jr.'
 		}
-	];
+	};
 
 	let modal = undefined;
 
 	const checkIfModalShouldBeShown = () => {
 		console.log('mount');
 		const modalToShow = $page.url.hash.substring(1);
-		modal = modalContents.find(({ id }) => id === modalToShow);
+		modal = modalContents[modalToShow];
 	};
 
 	onMount(checkIfModalShouldBeShown);
@@ -25,3 +38,7 @@
 <h1>
 	{modal?.title ?? ''}
 </h1>
+
+<p>
+	Loaded {calls} times.
+</p>
--- a/.changeset/old-comics-happen.md
+++ b/.changeset/old-comics-happen.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+fix: sveltekit:prefetch correctly distinguishes between hash routes
--- a/.changeset/old-comics-happen.md
+++ b/.changeset/old-comics-happen.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-fix: sveltekit:prefetch correctly distinguishes between hash routes
+fix: page store correct after navigation when an identical route with a different hash had been prefetched
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -686,6 +686,7 @@
     "slow-dots-know",
     "slow-fans-own",
     "slow-suns-reply",
+    "small-crabs-sparkle",
     "small-ligers-wave",
     "small-pigs-doubt",
     "smart-apes-drum",
--- a/packages/kit/CHANGELOG.md
+++ b/packages/kit/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @sveltejs/kit
 
+## 1.0.0-next.324
+
+### Patch Changes
+
+- Throw errors on encountering named layout in directory ([#4756](https://github.com/sveltejs/kit/pull/4756))
+
 ## 1.0.0-next.323
 
 ### Patch Changes
--- a/packages/kit/package.json
+++ b/packages/kit/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@sveltejs/kit",
-	"version": "1.0.0-next.323",
+	"version": "1.0.0-next.324",
 	"repository": {
 		"type": "git",
 		"url": "https://github.com/sveltejs/kit",
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -72,9 +72,10 @@ export function create_client({ target, session, base, trailing_slash }) {
 		updated: create_updated_store()
 	};
 
-	/** @type {{id: string | null, promise: Promise<import('./types').NavigationResult | undefined> | null}} */
+	/** @type {import('./types').NavigationResultCache} */
 	const load_cache = {
 		id: null,
+		url: null,
 		promise: null
 	};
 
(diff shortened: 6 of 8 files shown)


packages/kit/test/apps/basics/src/routes/routing/prefetched/hash-route.svelte
```diff
@@ -0,0 +1,27 @@
...
+		{
+			id: 'please-dont-show-me',
+			title: 'Oopsie'
+		}
+	];
+
+	let modal = undefined;
+
+	const checkIfModalShouldBeShown = () => {
+		console.log('mount');
```
ivanhofer, line 15: my bad. I forgot to remove this debug line before I pushed the example. I would remove it

ivanhofer (approved):
Thanks @tcc-sejohnson for your fast response on this issue!

I'm not entirely sure what happens (and what should happen) if a `load` function gets involved here. Hash is usually a concept the server doesn't know. Would this code now call the `load`function on each hover of all links?

packages/kit/test/apps/basics/src/routes/routing/prefetched/hash-route.svelte
tcc-sejohnson: Hmm, I'll update the test to check (and just for assertion in the future), but I don't believe it'll be an issue. `load` can potentially run on both the client and the server, but the only implementation I changed here is the client one. I would expect the client `load` to rerun.

ivanhofer: Right, prefetch only happens on the client 👍

tcc-sejohnson:
> Thanks @tcc-sejohnson for your fast response on this issue!
> 
> I'm not entirely sure what happens (and what should happen) if a `load` function gets involved here. Hash is usually a concept the server doesn't know. Would this code now call the `load`function on each hover of all links? 

I don't believe it'll be an issue. `load` can potentially run on both the client and the server, but the only implementation I changed here is the client one. I would expect the client `load` to rerun, right?

ivanhofer:
> > Thanks @tcc-sejohnson for your fast response on this issue!
> > I'm not entirely sure what happens (and what should happen) if a `load` function gets involved here. Hash is usually a concept the server doesn't know. Would this code now call the `load`function on each hover of all links?
> 
> I don't believe it'll be an issue. `load` can potentially run on both the client and the server, but the only implementation I changed here is the client one. I would expect the client `load` to rerun, right?

Right, prefetch happens only on the client.
And yes, I would expect to rerun it if the funtion reads from the hash. I think SvelteKit has some smart checks that only re-runs the load function if something changes. Not sure if the hash is also part of that diffing process.

tcc-sejohnson:
Moved tests around prefetching into their own group, as the parallel execution of the prefetching tests and some of the routing tests caused the routing tests to be flaky. 

Added two tests to check that multiple calls to `prerender` for the same hash route do not result in duplicate runs of `load`, and to check that multiple calls to `prerender` for different hash routes on the same base route do rerun `load`.

benmccann:
This doesn't make sense to me. `/page`, `/page#hash1`, and `/page#hash2` are all the same page. The page contents should not differ between those URLs and nothing should happen if you navigate between those URLs

tcc-sejohnson:
@benmccann 

I agree that the page contents should not change based on the URL hash -- the tests here simply use that as the method to verify that the correct link has been prefetched and that the prefetch behavior (as far as when to re-prefetch a page) is correct.

As I outlined in the original section, there are two bugs:

1. No re-prefetching of routes that are different by hash only
2. Incorrect value sent to page store on navigation to a "naked" route after prefetch of an identical hash route.

I'm pretty positive 2 is _actually_ a bug -- the `page` store should never tell me I'm at `/prefetched#some-hash` when the link I clicked is actually `/prefetched#another-hash` or `/prefetched`. 

1 is more of a judgment call. Do we re-prefetch a page on a hash route change, or make sure that we only fix 2 and assert that prefetching `/prefetched#some-hash` is the same as prefetching `/prefetched` and refuse to prefetch again?

benmccann:
1 is the expected behavior and not a bug.

> Do we re-prefetch a page on a hash route change, or make sure that we only fix 2 and assert that prefetching /prefetched#some-hash is the same as prefetching /prefetched and refuse to prefetch again?

No, we shouldn't re-prefetch. Prefetching `/prefetched#some-hash` and `/prefetched` should be the same so we shouldn't need to prefetch again.

tcc-sejohnson:
@benmccann 

My initial thought was to agree with you, but rereading the `prefetch` documentation seeds some doubt. One of the guarantees `prefetch` makes is that it runs the `load` function of the page when prefetching. `load` receives the URL, and it is very possible users want to do something in load with the URL hash (obviously only applicable on the client, but that's what the `browser` import is for).

If I hover `prefetched#some-hash`, my route is prefetched and `load` is called with `prefetched#some-hash`. If I then click on `prefetched`, load would not rerun with the correct URL. While it may be the "same page" from a semantic standpoint, I think it's very valid to want to do something in JS with the hash, no? It feels very wrong to tell users "Your load function _might_ be called with the correct URL, or it might not -- depends on whether you use hashes and prefetching."

benmccann:
> it is very possible users want to do something in load with the URL hash

They should not be. Maybe we should strip out the hash to make this impossible. `load` is meant to run once for the page to load the page contents. The contents should not differ by hash and `load` should not be run again if the hash changes

tcc-sejohnson:
Yep, definitely defer to you there -- with those semantics, I fully agree. 

For this PR, I'll update to ensure the correct `page` (with the correct hash) is passed to the `page` store when a link is clicked after prefetching an identical link with a hash.

Would you like me to create an additional issue around stripping the hash out of the URL passed to `load` on the client?

ivanhofer:
Yes, the `hash` should probably be removed from the `load` function, since it could lead to unexpected issues when running on a server and inside the browser. This definitely should be mentioned somewhere in the documentation.

tcc-sejohnson:
Turns out it was a lot tougher to do things without re-prefetching, as the cache is a promise that we obviously don't want to have to wait for in order to update the URL.

Solution I came up with was to:

- Add the URL to the cache object
- When validating the cache, if the cache ID and the new `NavigationIntent` ID are the same, check to see if only the URL hash changed. If it did, create a new promise closure around the old cache promise. Await the old promise, update the URL hash to the new URL hash, and return it. Return the closure promise. 

The only downside I could see with this is that it's creating a new closure every time a the cache is checked and the hash is different, meaning if you prefetch the same route with different hashes a bunch of times, you're creating this big annoying closure stack. In practice, I didn't see any issues -- I added a test that prefetches two different hash routes a thousand times and it ran very quickly and didn't cause any weird issues.

Without a significant rework, I can't think of any other way we could fix this, but I'm very open to ideas. 😺

tcc-sejohnson:
Given the decision in #4788, I expect this is just about ready for primetime -- though someone smarter than me might want to evaluate the solution for any footguns I'm not seeing.

Rich-Harris:
Given that we decided to just strip `url.hash` out inside `load`, I'm wondering if we need `validate_cache`? If the bug is that `$page.url` contains the hash that was present when the first link was prefetched, what if we just reset the URL?

This fixes the original repro, and the tests in this PR:

```diff
		if (started) {
			current = navigation_result.state;

+			if (navigation_result.props.page) {
+				navigation_result.props.page.url = url;
+			}
+
			root.$set(navigation_result.props);
		} else {
			initialize(navigation_result);
		}
```

Opened #5039 as an alternative (branched off of this PR)

tcc-sejohnson:
@Rich-Harris 

Darn it, I knew there was a simpler solution. I had tried that at some point (it took me _forever_ to figure out that `root.$set` was where the page store was getting set). Back when we had `url.hash` in `load`, we had to use the re-`prefetch`ing strategy to ensure we re-prefetched the route on hash changes, so this wouldn't have worked, but you're totally right -- this is the right way to go now that we don't care about the hash from a `prefetch`ing perspective. You can go ahead and close this PR unless there's something else we need to do on it.

We can probably also remove the "prefetch a route 1000 times" test I added -- I had it there just to verify that the promise-wrapping wouldn't cause issues, but there's no possibility of that now.

Rich-Harris:
Roger that, closing in favour of #5039. thanks!

Issue #4783, opened 2022-04-30
[adapter-cloudflare-workers] add Link headers to preloaded assets


maraisr:
### Describe the problem

Cloudflare supports Early Hints, which is as far as my understanding goes — will async fetch any assets as the response is served, and allows the browser to fetch assets before parsing the document.

Would it be possible at all to flush these headers.

👉 https://developers.cloudflare.com/cache/about/early-hints
eg:

```
Link: </_app/start-b98cfa77.js>; rel=preload
```

### Describe the proposed solution

I haven't delved too deeply, but feel the assets are available for the page (its injected in the html), so surely the header can also be flushed.

### Alternatives considered

Do nothing, the html `modulepreload` seems alright.

### Importance

nice to have

Issue #4158, opened 2022-02-28
Different link preloads for static rendering 


PH4NTOMiki:
### Describe the bug

Currently `adapter-static` renders same HTML output(.html files) for all `spa: true` / `fallback` mode routes.

What should happen is different link preloads for every HTML page(file) so loading happens quicker.

Kinda related: #4157

### Reproduction

.

### System Info

```shell
latest SvelteKit
```

### Severity

serious, but I can work around it

Conduitry:
I don't think this makes sense. How do you have multiple SPA fallbacks with static files?

PH4NTOMiki:
I don't, there is just one(`index.html`), but other routes get outputed if there is no dynamic parts/slugs in them.

Rich-Harris:
What link preloads are we talking about?

PH4NTOMiki:
<link rel=modulepreload>

Rich-Harris:
Right, but which modules are we preloading? We can't know which modules a given page depends on without server-rendering them, but the whole reason to use SPA+fallback is that you _can't_ server-render those pages.

PH4NTOMiki:
Yeah, but I'm not specifically talking about fallback page(`200.html` for ex.) but other routes that are not server rendered(`{ssr:false}`) but create their own html files(`books.html` in my case is one of them) needs resources `/_app/pages/__layout.svelte-xxxxxxxx.js` (every page including fallback should include Layout preload) and `/_app/pages/books/index.svelte-xxxxxxxx.js` and we know those things in advance so we should include them when prerendering because it delays rendering(slightly) to have to parse `start.js` to find what resources it needs to render the current page. 

This is related to issue #4157 I also created where I'm requesting to prerender even the parameterized routes when `{ssr:false}` and `adapter-static` is being used so I can manually create rewrite file for my specific hosting service for routes, essentially my request in that issue is if there is a route for `/users/[user_id].svelte` and `ssr` is `false` to prerender/output it as `/users/_slug_id_.html` or something similar, and in that case we do link preloads for them too, and that would make fallback file(at least for my case, and I believe for some others too) unnecessary.

Does that makes sense?

PH4NTOMiki:
@Rich-Harris I could maybe try to do PR for this, but I don't know much about Svelte/Vite compilation

teemingc:
I _think_ this is resolved by https://github.com/sveltejs/kit/pull/15718

PR #4785, opened 2022-04-30
feat: supports link header

maraisr:

### Tests

### Changesets

---

This is my attempt at solving #4783 albeit not specifically for Cloudflare, but think all servers could benefit from it.

Probably this PR won't be solution that'll land as only SSR'd non-preloaded pages will have this header. But really all static html files could use it.

So, my thinking lies more about exposing the assets consumed by a page to an adaptor.

Adaptors know how to deal with their targets, as some use a `_headers` file, others allow it as well as `new Response('', {headers})`, and other are entirely different. So maybe instead of this functionality being first-class, defer it to an adaptor.

This way Cloudflare's adaptor can build a `_headers` file to contain the headers for all static (pre-rendered) html files, as well as include the `Link` header for ssr responses.

Will mark PR as draft, but feel free to open and merge if you see fit.

### DIFF
--- a/documentation/docs/14-configuration.md
+++ b/documentation/docs/14-configuration.md
@@ -29,7 +29,6 @@ const config = {
 				// ...
 			}
 		},
-		link_header: false,
 		endpointExtensions: ['.js', '.ts'],
 		files: {
 			assets: 'static',
@@ -138,12 +137,6 @@ export default config;
 When pages are prerendered, the CSP header is added via a `<meta http-equiv>` tag (note that in this case, `frame-ancestors`, `report-uri` and `sandbox` directives will be ignored).
 
 > When `mode` is `'auto'`, SvelteKit will use nonces for dynamically rendered pages and hashes for prerendered pages. Using nonces with prerendered pages is insecure and therefore forbidden.
-
-# link_header
-
-When `true` the server will inject a `Link: </my-files.js>; rel=preload` header defined by [MDN Link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link). This can be used to let the browser know about modulepreloads before the html is parsed.
-
-> Cloudflare with [Early Hits](https://developers.cloudflare.com/cache/about/early-hints) enabled can then further optimize this by fetching these assets as it flies through the edge.
 
 ### endpointExtensions
 
--- a/documentation/docs/14-configuration.md
+++ b/documentation/docs/14-configuration.md
@@ -29,6 +29,7 @@ const config = {
 				// ...
 			}
 		},
+		link_header: false,
 		endpointExtensions: ['.js', '.ts'],
 		files: {
 			assets: 'static',
@@ -137,6 +138,12 @@ export default config;
 When pages are prerendered, the CSP header is added via a `<meta http-equiv>` tag (note that in this case, `frame-ancestors`, `report-uri` and `sandbox` directives will be ignored).
 
 > When `mode` is `'auto'`, SvelteKit will use nonces for dynamically rendered pages and hashes for prerendered pages. Using nonces with prerendered pages is insecure and therefore forbidden.
+
+# link_header
+
+When `true` the server will inject a `Link: </my-files.js>; rel=preload` header defined by [MDN Link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link). This can be used to let the browser know about modulepreloads before the html is parsed.
+
+> Cloudflare with [Early Hits](https://developers.cloudflare.com/cache/about/early-hints) enabled can then further optimize this by fetching these assets as it flies through the edge.
 
 ### endpointExtensions
 
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -286,6 +286,7 @@ export async function create_plugin(config, cwd) {
 							{
 								amp: config.kit.amp,
 								csp: config.kit.csp,
+								link_header: false,
 								dev: true,
 								floc: config.kit.floc,
 								get_stack: (error) => {
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -286,7 +286,6 @@ export async function create_plugin(config, cwd) {
 							{
 								amp: config.kit.amp,
 								csp: config.kit.csp,
-								link_header: false,
 								dev: true,
 								floc: config.kit.floc,
 								get_stack: (error) => {
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -233,7 +233,9 @@ export async function render_response({
 					// don't load stylesheets that are already inlined
 					// include them in disabled state so that Vite can detect them and doesn't try to add them
 					attributes.push('disabled', 'media="(max-width: 0)"');
-				} else if (options.link_header) link_header_preload.add(path);
+				} else {
+					link_header_preload.add(`<${path}>; rel="preload"`);
+				}
 
 				return `\n\t<link ${attributes.join(' ')}>`;
 			})
@@ -243,7 +245,7 @@ export async function render_response({
 			for (const dep of modulepreloads) {
 				const path = options.prefix + dep;
 				head += `\n\t<link rel="modulepreload" href="${path}">`;
-				if (options.link_header) link_header_preload.add(path);
+				link_header_preload.add(`<${path}>; rel="modulepreload"`);
 			}
 
 			const attributes = ['type="module"', `data-hydrate="${target}"`];
@@ -310,12 +312,7 @@ export async function render_response({
 	});
 
 	if (link_header_preload.size) {
-		headers.set(
-			'link',
-			Array.from(link_header_preload)
-				.map((href) => `<${href}>; rel=preload`)
-				.join(',')
-		);
+		headers.set('link', Array.from(link_header_preload).join(','));
 	}
 
 	if (cache) {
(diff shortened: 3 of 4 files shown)


documentation/docs/14-configuration.md
```diff
@@ -138,6 +139,12 @@ When pages are prerendered, the CSP header is added via a `<meta http-equiv>` ta
 
 > When `mode` is `'auto'`, SvelteKit will use nonces for dynamically rendered pages and hashes for prerendered pages. Using nonces with prerendered pages is insecure and therefore forbidden.
 
+# link_header
```
benmccann, line 142: this should be `linkHeader` since it's exposed externally. we use camelCase for external APIs and snake_case for internal APIs

benmccann:
I wonder if we should not include the `modulepreload` in the HTML if the header is present? I'm curious what browsers this feature is supported in. MDN doesn't have any data for this and I thought I remembered the feature or a similar one being removed from browsers

documentation/docs/14-configuration.md
```diff
@@ -29,6 +29,7 @@ const config = {
 				// ...
 			}
 		},
+		link_header: false,
```
Rich-Harris, line 32:
```suggestion
		linkHeader: false,
```

documentation/docs/14-configuration.md
Rich-Harris, line 142:
```suggestion
# linkHeader
```

documentation/docs/14-configuration.md
```diff
@@ -138,6 +139,12 @@ When pages are prerendered, the CSP header is added via a `<meta http-equiv>` ta
 
 > When `mode` is `'auto'`, SvelteKit will use nonces for dynamically rendered pages and hashes for prerendered pages. Using nonces with prerendered pages is insecure and therefore forbidden.
 
+# link_header
+
+When `true` the server will inject a `Link: </my-files.js>; rel=preload` header defined by [MDN Link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link). This can be used to let the browser know about modulepreloads before the html is parsed.
+
+> Cloudflare with [Early Hits](https://developers.cloudflare.com/cache/about/early-hints) enabled can then further optimize this by fetching these assets as it flies through the edge.
```
Rich-Harris, line 146: I don't think it's worth mentioning early hints here — since we return the response body along with the headers, I'd be very surprised if early hints offers anything over and above a `Link` header with an ordinary response

```suggestion
When `true` the server will inject a [`Link` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link). This can be used to let the browser eagerly fetch essential resources before the HTML is parsed.
```

documentation/docs/14-configuration.md
Rich-Harris: Also, this section should be moved — everything else is in alphabetical order

documentation/docs/14-configuration.md
Rich-Harris: this should be moved so that it's in alphabetical order

packages/kit/src/runtime/server/page/render.js
```diff
@@ -304,6 +309,15 @@ export async function render_response({
...
 
+	if (link_header_preload.size) {
+		headers.set(
+			'link',
+			Array.from(link_header_preload)
+				.map((href) => `<${href}>; rel=preload`)
+				.join(',')
+		);
+	}
+
```
Rich-Harris, line 320: I'm pretty sure this is incorrect — modules should have `rel=modulepreload`, not `rel=preload`

packages/kit/src/core/dev/plugin.js
```diff
@@ -286,6 +286,7 @@ export async function create_plugin(config, cwd) {
 							{
 								amp: config.kit.amp,
 								csp: config.kit.csp,
+								link_header: false,
```
Rich-Harris, line 289: this should be moved to its alphabetical place

packages/kit/types/internal.d.ts
```diff
@@ -242,6 +242,7 @@ export type SSRNodeLoader = () => Promise<SSRNode>;
 export interface SSROptions {
 	amp: boolean;
 	csp: ValidatedConfig['kit']['csp'];
+	link_header: boolean;
```
Rich-Harris, line 245: should be moved to its alphabetical place

Rich-Harris (changes requested):
Thanks — is there any reason we wouldn't just always include this header? Would be simpler if we didn't have to introduce another config option here

packages/kit/src/runtime/server/page/render.js
maraisr: Ya you're right 😅 
- https://datatracker.ietf.org/doc/html/rfc8288#section-2.1.1.1
- https://www.iana.org/assignments/link-relations

fixed. <3

maraisr:
Thanks for that one @Rich-Harris  I have updated the PR — and you're right, I cannot see why this header SHOULD not have been included by default.

Have added that in.

As per other points mentioned around what do we do for pre-rended files, do we need something similar to this by in lastmile? Aka post process once that file is either read from disk, from KV or so forth?

benmccann:
This PR doesn't actually implement early hints as discussed in the linked issue (e.g. nowhere are we returning a 103 status code). What this PR does do is implement headers for server push is which is being removed from browsers. I don't think this PR would necessarily hurt, but I think the benefits are likely to be minimal. The benefits might be meaningful for a page with high SSR render time if early hints were indeed implemented

From https://medium.com/reloading/preload-prefetch-and-priorities-in-chrome-776165961bbf:

> Preloading via headers _may_ offer a very negligible improvement, but require the document to be committed before processing happens.

> You can provide preload links in either form, but there is one important difference you should understand: as allowed by the spec, many servers initiate an HTTP/2 Server Push when they encounter a preload link in HTTP header form. The performance implications of H/2 Server Push are different from those of preloading (see below), so you should make sure you don’t unintentionally trigger pushes.

From https://www.ctrl.blog/entry/http2-push-chromium-deprecation.html:
> Chromium developers have [announced](https://groups.google.com/a/chromium.org/g/blink-dev/c/K3rYLvmQUBY/m/vOWBKZGoAQAJ?pli=1) that they plan to remove support for HTTP/2 server push from the market-leading browser engine.
> The Chrome team has measured the adoption of HTTP/2 push, and found that only 0,05 % of all HTTP/2 sessions seen by Chrome users use server push.
> Servers aren’t aware of the state of the client’s cache. They can’t know what resources the client has stored from previous visits, and risk pushing assets needlessly. This can waste people’s precious data quotas, slow-down page-loading performance, and cost you more money. 

From https://brianli.com/2020/12/chrome-to-drop-support-for-http2-server-push/:
> Firefox and Safari both support HTTP/2 push and we are not aware of any plans for removal.

maraisr:
Sorry to have caused you trouble.

mrkishi:
@maraisr Thank you for investigating this. I believe this PR could still move forward (although possibly with changes).

I think there's some miscommunication at play here.

Hints can be added both to `Link:` headers and `<link>` tags, and they should indeed provide similar performance. _However_, headers are more convenient when doing further optimization since they arrive early and don't require an html parser... And that's why they were once used for HTTP/2's Server Push (which while deprecated is still supported by some servers and clients) and they're similarly being used for Early Hints by reverse proxies.

Now for the kicker: `Link:` headers on responses with status codes 200, 301 and 302 are _cached and reused_ by Cloudflare to implement Early Hints. Even if we don't explicitly implement 103 status codes, make Svelte 4's `render()` streaming and turn SvelteKit into a streaming-first framework, users can still benefit from link headers when paired with the right reverse proxy with an extra configuration.

Since we already have `<link>` tags, adding the equivalent header shouldn't cause issues—although because of Server Push it could change the performance characteristics of some proxy/client combos, so perhaps it'd be safer to add a `nopush` hint.

Apart from the `nopush` hint, it's possible we'd need `as=style` for css, although I'm not sure it makes a difference these days.

But would adding this to SSR now become a chore later if we decide to move the responsibility to adapters (as noted in OP)?

Rich-Harris:
I think we should reopen this PR until we've discussed this further, but I'm unable to do so - @maraisr can you hit the reopen button please? Thanks 🙏

Rich-Harris:
Some further relevant discussion about flushing JS/CSS-related tags (basically everything in `src/app.html` up until `%svelte.head%`, plus those tags) as soon as we match a route, rather than after the route has been rendered: https://github.com/sveltejs/kit/discussions/4831#discussioncomment-2695990.

Also related: https://github.com/sveltejs/kit/issues/4158#issuecomment-1108794924

mrkishi:
Since early headers would require solving the bigger streaming questions (early status codes) it sounds like it could continue to be discussed in #4831, and for now we just focus on improvements assuming late headers. In that case, this seems mostly there as a stop-gap, except for the adapter-specific questions. Or should we focus on the full streaming solution?

Regarding the adapters, do you have opinions on where to take the adapter api in order to expose this sort of per-page information? For instance, it'd be nice if they were able to setup their platform's necessary plumbing to add the headers to prerendered pages.

mrkishi:
> it sounds like the impact is quite limited when the page has been cached

Yeah, that's expected. When a page is cached it makes sense not to use early hints since they wouldn't be early as they wouldn't be waiting for origin. But since this is regarding SSRed pages, most won't be cached. And iirc Cloudflare also doesn't cache html pages by default even with `cache-control` headers, you have to add a Page Rule for that.

> I don't think this would really make sense to live in adapters since it's a runtime rather than build-time concern?

I wasn't clear there, but this is a concern when prerendering pages. Granted, the upside is limited for the same reason as Cloudflare won't use early hints when cached, but this is something that'd impact _other_ headers anyway, so it's worth considering how it'd be handled anyway. We currently have no way to specify headers on prerendered pages outside of meta tags.

PR #4786, opened 2022-04-30
Allow user to define custom layoutPrefix for layout page

heinthanth:
This patch allow user to define custom layoutPrefix to be use as layout page. There're several users who don't like `__layout.svelte` convention. So, now, they can define `layoutPrefix: '__tmpl'` in `svelte.config.js` and can use `__tmpl.svelte` as their layout page.

### Tests

### Changesets


benmccann:
I'm not sure I love this idea. It makes projects using the feature far less approachable for people who are already familiar with SvelteKit. What's the reason for suggesting this?

heinthanth:
Nothing special reason behind this pull request. 
Being able to customize is a must-have feature of a framework.
But some people ( including me ) don't like the default naming convention.

So, I thought of two options:
1. the first one is to create a patch file and apply it every time `npm install`.
2. the second one is to create a pull request so that everyone can take advantage of it.

I even hacked the kit to exclude `.svelte` from `[name]-[hash].js` client side output to make svelte detection harder.

Conduitry:
> Being able to customize is a must-have feature of a framework.

Being able to rename arbitrary features is definitely not a must-have feature of a framework.

Mlocik97:
sry, I'm against merging this. As it was already said, it would just add chaos for new developers that would try to contribute to existing projects that use it's own name convention.

Conduitry:
The whole reason for using two-underscore prefixes for `__layout` and `__error` and preventing users from having other files with this prefix is so that there's no chance of future new `__` files SvelteKit might want to add colliding with other files users might have in their apps. Without this, SvelteKit adding any `__foo.js` with a special meaning would be a breaking change if people were previously allowed to have `__foo.js` with no special meaning.

This PR brings us right back into that, by letting people introduce collisions, which we don't want.

There have been discussions in other issues about renaming `__layout`, which may well still happen. We're going to be adding middleware for endpoints sometime before 1.0, and that may end up being something like `__middleware.js`. `__layout` might end up just being for things that truly are related to the visual layout. This is all still up in the air.

There's going to be an enforced convention, and it's not settled on yet. We'll endeavor to make these good names, but having them be configurable is not on the table. If you really hate the name for whatever reason and you don't care about the other people who might be reading your code, you're welcome to fork SvelteKit.

Issue #4764, opened 2022-04-28
Package now compiles Typescript files to CJS instead of ESM


jacob-8:
### Describe the bug

Upgrading from @sveltejs/kit@1.0.0-next.314 to @sveltejs/kit@1.0.0-next.315 will change the output of svelte-kit package as follows:
 
![image](https://user-images.githubusercontent.com/7559478/165654223-72f2ebdc-6b3c-45e1-98ed-3d48f29d7a4f.png)

![image](https://user-images.githubusercontent.com/7559478/165654318-67ad98d3-21e9-410a-b835-b45fdd67629a.png)

My package is now broken when trying to install in another app (`ReferenceError: exports is not defined` in reference to an action defined in a .js file) if I upgrade SvelteKit and I'm forced to stay .314. 

### Reproduction

- Clone https://github.com/jacob-8/svelte-pieces/
- `pnpm i`
- `pnpm package` - look at any compiled .js filed inside /package
- update to @sveltejs/kit@1.0.0-next.315 or higher
- `pnpm i`
- `pnpm package` and inspect different results in compiled .js files in /package

### System Info

```shell
System:
    OS: Windows 10 10.0.22000
    CPU: (8) x64 Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz
    Memory: 1.14 GB / 11.77 GB
  Binaries:
    Node: 16.13.2 - C:\Program Files\nodejs\node.EXE
    npm: 8.1.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (100.0.1185.50)
    Internet Explorer: 11.0.22000.120
  npmPackages:
    @sveltejs/adapter-auto: 1.0.0-next.40 => 1.0.0-next.40
    @sveltejs/kit: 1.0.0-next.314 => 1.0.0-next.315
    @sveltejs/vite-plugin-svelte: ^1.0.0-next.41 => 1.0.0-next.41
    svelte: ^3.47.0 => 3.47.0
```

### Severity

blocking an upgrade

tcc-sejohnson:
Weird... I just checked the changelog for 314 => 315, and [this is the only line I could possibly see that might cause this issue](https://github.com/sveltejs/kit/blob/c8f71e3191015d74a5ac446e064b5d21aa408500/packages/kit/rollup.config.js#L57), but it shouldn't be the problem. I'll try to look further into this when I get home to at least triage.

[...]

Issue #4460, opened 2022-03-28
Kit Packaged libs don't work outside of dev


ghostdevv:
### Describe the bug

One of my packages [svelte-copy](https://github.com/ghostdevv/svelte-copy), which uses svelte kit package, fails to build when using netlify. I build a [repro here](https://github.com/ghostdevv/svelte-copy-require-repro). I get the error of `Must use import to load ES Module: /var/task/node_modules/svelte-copy/index.js`

I tracked this down to `.netlify/server/entries/pages/index.svelte.js` which has the line `var import_svelte_copy = require("svelte-copy")`

### Reproduction

https://github.com/ghostdevv/svelte-copy-require-repro

### Logs

```shell
500
Must use import to load ES Module: /var/task/node_modules/svelte-copy/index.js
require() of ES modules is not supported.
require() of /var/task/node_modules/svelte-copy/index.js from /var/task/.netlify/server/entries/pages/index.svelte.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /var/task/node_modules/svelte-copy/package.json.
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /var/task/node_modules/svelte-copy/index.js
require() of ES modules is not supported.
require() of /var/task/node_modules/svelte-copy/index.js from /var/task/.netlify/server/entries/pages/index.svelte.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /var/task/node_modules/svelte-copy/package.json.
```

### System Info

```shell
Binaries:
    Node: 16.9.1 - ~/.nvm/versions/node/v16.9.1/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v16.9.1/bin/yarn
    npm: 7.24.2 - ~/.nvm/versions/node/v16.9.1/bin/npm
  Browsers:
    Brave Browser: 99.1.36.116
    Chromium: 99.0.4844.74
    Firefox: 98.0.1
```

### Severity

blocking all usage of SvelteKit

ghostdevv:
[...]

PR #4791, opened 2022-05-01
update generated tsconfig

ghostdevv:

### Tests

### Changesets

It's probably best to discuss the side effects of this change, it fixes my issue of https://github.com/sveltejs/kit/issues/4460 and also https://github.com/sveltejs/kit/issues/4764 - However this was changed in https://github.com/sveltejs/kit/pull/4633 and maybe this change is what should stay and alternative solution to the issues can be done such as what I proposed here https://github.com/sveltejs/kit/issues/4460#issuecomment-1111603855

### DIFF
--- a/.changeset/hip-ears-reflect.md
+++ b/.changeset/hip-ears-reflect.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+add lib, module, and target to generated tsconfig
--- a/.changeset/hip-ears-reflect.md
+++ b/.changeset/hip-ears-reflect.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-add lib, module, and target to generated tsconfig
+[breaking] add lib, module, and target to generated tsconfig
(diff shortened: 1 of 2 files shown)


Conduitry:
Maybe we should be using `esnext` as the target for all of these things? We don't want to output CJS, but we also want to make people opt in to whatever specific degree of JS transpiling they need for their situation.

ghostdevv:
Good idea, have switched that

.changeset/hip-ears-reflect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+add lib, module, and target to generated tsconfig
```
Rich-Harris, line 5: technically a breaking change, not that it should actually break anything

```suggestion
[breaking] add lib, module, and target to generated tsconfig
```

Rich-Harris (approved):

Rich-Harris:
Thanks, this seems like the right fix to me

PR #4792, opened 2022-05-02
feat: allows platform in externalFetch

maraisr:

### Tests

### Changesets

---

Within Cloudflare workers and [service bindings](https://blog.cloudflare.com/introducing-worker-services/) the externalFetch function _could_ make sense to support a `env.MY_API.fetch` to perform that request.

As such; lets allow `state.platform` to be sent into `externalFetch` thus allowing something like;

```ts

// within Cloudflare workers _or_ pages

/** @type {import('@sveltejs/kit').ExternalFetch} */
export async function externalFetch(request, {env}) {
	if (request.url.startsWith('https://api.example.com')) {
		return env.API.fetch(request.clone());
	}

	return fetch(request);
}
```

Maybe it makes sense to support a similar suite of options that the current _handle_ hook does. 

---

Either this, or do you see the `load` module export handle this internally? aka

```svelte
<script context="module" lang="ts">
	export const prerender = false;

	export async function load({ params, fetch, platform }) {
		const req = await (platform?.env.API.fetch ?? fetch)(`https://api.example.com/orgs/${params.slug}`);

		if (!req.ok) return { status: 404 };

		return {
			status: 200,
			props: {
				org: await req.json().then((d) => d.data),
			},
		};
	}
</script>
```

> granted would mean load can be conditionally be given platform options. browser/ssr, which imo is more cumbersome. So would be more inclined to the first (and this PR) approach.

### DIFF
--- a/packages/kit/src/core/build/build_server.js
+++ b/packages/kit/src/core/build/build_server.js
@@ -93,7 +93,7 @@ export class Server {
 				getSession: module.getSession || (() => ({})),
 				handle: module.handle || (({ event, resolve }) => resolve(event)),
 				handleError: module.handleError || (({ error }) => console.error(error.stack)),
-				externalFetch: module.externalFetch || fetch
+				externalFetch: module.externalFetch || ((r) => fetch(r))
 			};
 		}
 
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -247,7 +247,7 @@ export async function load_node({
 					}
 
 					const external_request = new Request(requested, /** @type {RequestInit} */ (opts));
-					response = await options.hooks.externalFetch.call(null, external_request);
+					response = await options.hooks.externalFetch.call(null, external_request, state.platform);
 				}
 
 				const set_cookie = response.headers.get('set-cookie');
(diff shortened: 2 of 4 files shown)


packages/kit/src/core/build/build_server.js
```diff
@@ -93,7 +93,7 @@ export class Server {
 				getSession: module.getSession || (() => ({})),
 				handle: module.handle || (({ event, resolve }) => resolve(event)),
 				handleError: module.handleError || (({ error }) => console.error(error.stack)),
-				externalFetch: module.externalFetch || fetch
+				externalFetch: module.externalFetch || ((r) => fetch(r))
```
maraisr, line 96: we need to do this coz `fetch(a, b)` will be wrong when called during execution. So lets pick only first arg.

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -247,7 +247,7 @@ export async function load_node({
 					}
 
 					const external_request = new Request(requested, /** @type {RequestInit} */ (opts));
-					response = await options.hooks.externalFetch.call(null, external_request);
+					response = await options.hooks.externalFetch.call(null, external_request, state.platform);
```
maraisr, line 250: maybe entire `event` could suffice?

PR #4810, opened 2022-05-03
[feat] support wrangler2 in adapter-cloudflare

maraisr:

### Tests

### Changesets

---

Wrangler 2 has a new config file, and looks somewhat different.

This aims to allow the validator to still pass when building for Cloudflare workers.

Another solution is to potentially allow bypassing validation with a flag until the wrangler 2 toml scheme is finalised.

cc @threepointone

### DIFF
--- a/packages/adapter-cloudflare-workers/index.js
+++ b/packages/adapter-cloudflare-workers/index.js
@@ -90,7 +90,7 @@ function validate_config(builder) {
 			throw err;
 		}
 
-		// rudamentary way to detect a wrangler2 config file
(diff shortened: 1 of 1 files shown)


packages/adapter-cloudflare-workers/index.js
```diff
@@ -90,27 +90,32 @@ function validate_config(builder) {
 			throw err;
 		}
 
+		// rudamentary way to detect a wrangler2 config file
```
benmccann, line 93:
```suggestion
		// rudimentary way to detect a wrangler2 config file
```

PR #4811, opened 2022-05-04
Fail prerendering if page attempts to access `session` in `load` - fixes #3722

### LINKED ISSUE #3722: `getSession()` does not get called on page load. should warn user
### Describe the bug

Unfortunately, I haven't been able to get our sveltekit app to work in production past kt@next.206 and adapter-node@next.56.

Since the ssr was always somewhat flunky with our - aside from the login page- completely private app we previously disabled it via the config and on all update attempts in the handle hook.

Now, the issue is always the same, everything runs fine locally, until it gets deployed, where you can't get past the login screen anymore, since the session never gets populated. I recently found out that the issue seems to lie somewhere within adapter-node, since running `dev` and `preview` works flawlessly, but running the build output directly via `node _build` presents the same issue our staging environment showed locally.

Any help/pointers would be appreciated, ready to replace our auth code if necessary/it's just us working against sveltekit.

If you have any further questions, feel free to ask !

### Reproduction

https://github.com/luhagel/sveltekit-node-session-repro

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 11.6
    CPU: (12) x64 AMD Ryzen 5 3600X 6-Core Processor             
    Memory: 9.02 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.4.0 - /usr/local/bin/node
    npm: 8.3.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 98.0.4758.80
    Firefox: 96.0.2
    Safari: 15.0
  npmPackages:
    @sveltejs/adapter-node: ^1.0.0-next.67 => 1.0.0-next.67 
    svelte: ^3.44.0 => 3.46.4
```

### Severity

blocking an upgrade

### Additional Information

_No response_

tcc-sejohnson:
It doesn't make sense to prerender a page that tries to access `session`, as it can never be populated by the server. Similarly to how we throw if we try to prerender a page that has a mutative shadow endpoint, we now throw if a page tries to access `session` during prerendering. This exception can be skipped by setting `continueOnError` to `true`, but it will still warn users by logging the error to the console. Fixes #3722.

As is, this will fail on server-side requests while in dev as well (so that developers don't have to wait until they try to build for prod to get slapped in the face with it).

### Tests

### Changesets


tcc-sejohnson:
Hmm. Strange. I linted/checked locally...

I'll run again and push if necessary.

Rich-Harris (approved):

Rich-Harris:
excellent, thank you!

PR #4812, opened 2022-05-04
feat: Fail prerendering for pages with mutative endpoints - fixes #3410 and #4252

### LINKED ISSUE #3410: Fail prerendering if a page is shadowed by an endpoint
### Describe the bug

If you create a `bob/index` with a POST and `bob/index.svelte` --> posting will not work when deployed to Vercel.

Vercel throws error 405.

### Reproduction

Deploy this [repo](https://github.com/21RISK/sveltekit) to Vercel, and visit `/bob` and hit the submit button - and then the error will appear.

As long as it exists 😅, you can also try it here: https://sveltekit-p9hxntt45-jeddah.vercel.app/bob

### Logs

_No response_

### System Info

```shell
Latest version `next` of kit and vercel adapter.
```

### Severity

### Additional Information

I can simply work around by not enabling `prerender=true` on the sveltekit component.

Screen Shot 2022-01-19 at 12 53 36

Let me know if you need more information 😎

### LINKED ISSUE #4252: Prerendered page alongside non-GET request handler should be an error
### Describe the bug

Related to #3294. I don't think there's a situation in which it would make sense to have a non-GET request handler alongside a prerendered page.

In the same vein as https://github.com/sveltejs/kit/issues/4093#issuecomment-1049117727, SvelteKit should throw an error if a prerenderable page (either `export const prerender = true`, or `config.kit.prerender.default = true` without a `export const prerenderable = false`) has an endpoint with non-GET methods.

### Reproduction

Add a non-GET method to any page endpoint and mark the page as prerenderable: https://stackblitz.com/edit/sveltejs-kit-template-default-8la3qs?file=src%2Froutes%2Fabout.js&terminal=dev

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.0.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 169.98 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.1 - ~/.nvm/versions/node/v16.13.1/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v16.13.1/bin/yarn
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.1/bin/npm
  Browsers:
    Chrome: 98.0.4758.109
    Firefox: 97.0.1
    Safari: 15.1
  npmPackages:
    svelte: ^3.43.0 => 3.44.2
```

### Severity

annoyance

### Additional Information

_No response_

tcc-sejohnson:
There are a few issues up around the confusing things that happen if a page is prerendered with mutative methods. Thankfully, this already errors at buildtime. However, it can be a bit of a footgun during development, as there's no warning what you're doing isn't valid until you try to build.

Added tests to make sure that, if a page should be prerendered (both a build-time and at runtime), it fails to do so with a descriptive message.

Fixes #3410 and fixes #4252.

### Tests

### Changesets


tcc-sejohnson:
Only downside of this that I can think of is that it'll only fail in dev when a server-side navigation occurs... but I don't think it's even possible to check for prerendering in the client. (A quick search of `/packages/kit/src/runtime/client` for "prerender" yields... 0 results.)

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -58,7 +58,7 @@ export async function load_node({
 				/** @type {import('types').SSRPage} */ (route),
 				event,
 				options,
-				!!state.prerender
+				node.module.prerender ?? state?.prerender?.default ?? false
```
benmccann, line 61: It took me awhile to read this line. It might help to refactor it out into a variable since the variable name would make things clearer
```
const prerender_enabled = node.module.prerender ?? state?.prerender?.default ?? false;
```

Do we still need the `state.prerender` reference? I'd expect it to be something like `prerender_enabled && state.prerender`

tcc-sejohnson: Agreed, a variable makes that easier to understand. Added.

> Do we still need the state.prerender reference? I'd expect it to be something like prerender_enabled && state.prerender

I'm not sure _exactly_ what you mean there. We have to check for `node.module.prerender`, as that corresponds to the module-scoped `export const prerender`. If that's not defined, we have to check for the default behavior (if `kit.prerender.default === true`, we should prerender this page _unless_ the page opts out, which we already checked for).

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -52,13 +52,15 @@ export async function load_node({
 	/** @type {import('types').LoadOutput} */
 	let loaded;
 
+	const should_prerender = node.module.prerender ?? state?.prerender?.default ?? false;
```
benmccann, line 55: "should" is very vague. it doesn't really clarify under what conditions prerendering happens. I think `prerender_enabled` would make it clearer that we're checking whether it's enabled for the page

I also realized this could be written as `!!(node.module.prerender ?? state?.prerender?.default)` which might be easier to read by removing the double nullish coalescing?

tcc-sejohnson: I'm just copying the naming from [here](https://github.com/sveltejs/kit/blob/39eaf0905f28c55ba7ee3fe9ba67ec487290a6b2/packages/kit/src/runtime/server/page/respond.js#L75) for consistency:

```ts
const should_prerender = leaf.prerender ?? state.prerender.default;
```

If you want to go with `prerender_enabled`, I'm happy to make that change! Just want to confirm before we name it inconsistently.

As for the double-bang, it's funny -- I work a _lot_ in C# and Go. In the process of opening the 3-4 PRs I currently have going, I actually had to Google what the heck the double-bang was doing, as the coalescing chain is what I'm used to! 😆 If it's clearer to JS developers to use the double-bang, I'll change it along with the variable name.

packages/kit/src/runtime/server/page/load_node.js
benmccann: I could be wrong about that. I haven't really looked at any of the shadow endpoint code yet. I'll let Rich take a look at this PR and the remaining questions as he's more familiar

tcc-sejohnson: Yeah, full disclosure, I _definitely_ don't understand everything in the codebase, so I could definitely be wrong here!

packages/kit/src/runtime/server/page/load_node.js
tcc-sejohnson: @benmccann - I just had one of those "eyes pop open in the middle of the night" moments about this one. Doing something completely unrelated, hadn't thought about this in days, and there it was -- like some ugly toad jumping out of the grass on an evening walk. The reason it feels weird to name the variable `prerender_enabled` is because that's _not_ what the variable means. There's a setting in config, `prerender.enabled`, which controls whether Kit even _tries_ to prerender your app. I believe that flag is available in this context as `state.prerender.enabled`. This check is for whether or not this node _should_ prerender, which is determined by the combination of the node's exported `prerender` attribute and the config's prerendering default.

benmccann: I know it's not referring to whether it's enabled at the application level, but it does refer to whether it's enabled at the page level. Anyway, I'll let Rich weigh in with his thoughts

Rich-Harris: i think `should_prerender` is fine — i agree that `prerender_enabled` sounds too much like it's just `config.kit.prerender.enabled`

Rich-Harris:
Thank you! There was one small change I needed to make — whereas in the page `respond.js` module we're checking whether we're _actually_ prerendering at the moment, in the context of this PR we need to check whether we _will_ prerender the page. In the first case we need to check state, but in the second (in e.g. dev mode) we need to check configuration.

Though thinking about it, `state.prerender.default` is now redundant... one sec...

Rich-Harris:
I renamed `state.prerender` to `state.prerendering` to make the distinction clearer

Rich-Harris:
Ah whoops, looks like I broke something. Will look into it in a bit

tcc-sejohnson:
Bah, I looked into it for a bit, but I'm just not sure what's causing the issue. To save you some time, what's failing is a test in `adapter-static` that's making sure that when `config.kit.prerender.default = false`, pages that do not set `export const prerender = true` are not prerendered. For some reason, your last commit has made the prerendering process prerender `index.html` even though `index.svelte` does not export the `prerendering` attribute. I might have more time to look into it later.

sarajw:
In case this is useful to you, my experiences:

I tried to update my blog that uses adapter-static the day before yesterday and it didn't build right - only one of my *.svelte pages (incidentally the simplest and most recently edited) was built into a *.html file.

I could open the site via that one rendered html page and - I'm not smart enough yet to understand how it all works - I guess the rest was populating itself via JS and it sort of worked. Apart from, the blog index page which complained about the JSON it was referencing having a rogue `<` in it at 0.

Anyway eventually I paid attention to some warnings I was getting in the terminal, saying I had to add 
    `prerender: { default: true }`
to svelte.config.js, and
    `"extends": "./.svelte-kit/tsconfig.json"`
to jsconfig.json.

Then after that it built fine again. Are these lines I will have to change back once any fixes happen?

Thanks!

tcc-sejohnson:
Hey @sarajw! This is an unrelated issue and is actually intended behavior (unless there's something I'm not understanding about your issue). The terminal warnings you were seeing were trying to tell you this. For `adapter-static`, you must specify either `kit.prerender.default = true` (to prerender your site into static HTML) or `adapter.fallback = <some_page>` (to run in SPA mode). You can read more about both of these options in the adapter documentation -- and feel free to ask more questions in the SvelteKit help channel on Discord!

sarajw:
> Hey @sarajw! This is an unrelated issue and is actually intended behavior (unless there's something I'm not understanding about your issue). The terminal warnings you were seeing were trying to tell you this. For `adapter-static`, you must specify either `kit.prerender.default = true` (to prerender your site into static HTML) or `adapter.fallback = <some_page>` (to run in SPA mode). You can read more about both of these options in the adapter documentation -- and feel free to ask more questions in the SvelteKit help channel on Discord!

Absolutely fine if that's the case, it was just new and surprising. I had already posted in the Discord SvelteKit channel and not got much help or information.

tcc-sejohnson:
> Absolutely fine if that's the case, it was just new and surprising. I had already posted in the Discord SvelteKit channel and not got much help or information.

And to be clear, I'm not saying you're _not_ experiencing a bug -- just that I don't see any buggy behavior described in what you said. If you still think you are, please do create a repro and file an issue. 🙂 I personally think it's kind of weird that `adapter-static` doesn't "just work" without changing the mentioned settings, so it may be worth creating an issue to discuss ergonomics? 🤷

Rich-Harris (approved):

Rich-Harris:
thank you!

PR #4824, opened 2022-05-04
Generate tsconfig when running svelte-kit package

ghostdevv:

### Tests

### Changesets

Currently sveltekit package does not update the generated tsconfig when packaging, this fixes that

### DIFF
--- a/packages/kit/src/packaging/index.js
+++ b/packages/kit/src/packaging/index.js
@@ -90,9 +90,6 @@ export async function build(config, cwd = process.cwd()) {
  * @param {import('types').ValidatedConfig} config
  */
 export async function watch(config, cwd = process.cwd()) {
-	// Make sure generated tsconfig is up-to-date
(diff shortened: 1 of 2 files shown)


packages/kit/src/packaging/index.js
```diff
@@ -86,6 +90,9 @@ export async function build(config, cwd = process.cwd()) {
  * @param {import('types').ValidatedConfig} config
  */
 export async function watch(config, cwd = process.cwd()) {
+	// Make sure generated tsconfig is up-to-date
+	write_tsconfig(config);
+
```
benmccann, line 95: you were also adding this to the `build` command which is called on the next line, so we probably don't need to do it twice?
```suggestion
```

ghostdevv: Yep!

Rich-Harris (approved):

PR #4843, opened 2022-05-07
chore: make sure pnpm7 is used for this repo

dominikg:
using pnpm6 and 7 in parallel would cause problems with lockfile format changes and incompatibilities in package script syntax.

packageManager field allows people to use corepack if they like and the field is evaluated by some tools like `@antfu/ni`. renovate bot also needs a hint, otherwise it is using pnpm7 and it's PRs may fail.

Once we decide to update to 7 we just bump the versions in package.json and everyone using the repo gets notified on next use (or corepack handles it automatically).

### Tests

### Changesets


.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: pnpm/action-setup@v2.2.1
         with:
-          version: 6.23.2
+          version: 6.32.11
```
bluwy, line 20: We can also omit this as `pnpm/action-setup` reads `packageManager` too.

dominikg: https://github.com/pnpm/action-setup/blob/57b9359b4c69c79ca51422d02ca739a7957f4aed/src/install-pnpm/run.ts#L25-L27 

this is not great. basically curlbashing a hardcoded version of pnpm to then install another version? I don't get it.

what about this?
read packageManager from package.json and send that to  `npm i -g`
```bash
PNPM_VER=$(jq -r '.packageManager | if .[0:5] == "pnpm@" then .[5:] else "packageManager in package.json does not start with pnpm@\n" | halt_error(1)  end' package.json)
npm i -g pnpm@$PNPM_VER
```

that would exit with 1 if packageManager isn't set or doesn't start with pnpm@

benmccann (approved):

.github/workflows/ci.yml
benmccann: I removed the version for now and will see if we can get the issue fixed in `action-setup`: https://github.com/pnpm/action-setup/issues/38

packages/kit/test/utils.js
```diff
@@ -162,7 +162,8 @@ export const config = {
 	],
 	use: {
 		screenshot: 'only-on-failure',
-		trace: 'retain-on-failure'
+		trace: 'retain-on-failure',
+		channel: 'chrome'
```
benmccann, line 166: should we add a comment here explaining why we added this?

package.json
```diff
@@ -40,5 +40,9 @@
 		"turbo": "^1.2.6",
 		"typescript": "~4.6.2"
 	},
+	"packageManager": "pnpm@7.1.0",
+	"engines": {
+		"pnpm": "^7.1.0"
```
benmccann, line 45:
```suggestion
		"pnpm": "^7.0.0"
```

benmccann, line 45:
```suggestion
		"pnpm": "^7.0.0"
```

PR #4846, opened 2022-05-07
prevent scroll when resetting page focus (#4065)

timothycohen:
Fixes #4065. While resetting focus, Safari was scrolling over the top margin.

As it’s a Safari specific issue, the change in test from fail to pass is only visible when running Playwright with WebKit. 

This is my first PR, so let me know if I missed something in the workflow, thanks!

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [x] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #4847, opened 2022-05-07
WebKit tests

timothycohen:
While writing a test for a Safari only bug (PR #4846), the test suite exposed 13 false positives originating from browser specific behavior in the tests. This PR slightly tweaks the integration tests to be more browser agnostic and allows them to be run using WebKit. It doesn’t include changes to the Playwright config, so the WebKit tests would still be completely opt-in.

### "includes correct page request headers"
Test targeted headers that Safari doesn't support

### 1. "resets focus" <br/> 2. "focus works if page load has hash" <br/> 3. "focus works when navigating to a hash on the same page"
"Press tab to highlight each item on a webpage" accessibility setting is turned off in Playwright. Need to use Alt+Tab for WebKit instead.

### "client-side error from load() is malformed"
The test matched error.stack instead of error.message and WebKit doesn't show the message as the first line of the stack trace

### "prevents redirect loops"
Test checked for chrome specific error page.

### "prevents navigation triggered by back button"
The browsers block attempts to prevent navigation on a frame that's never had a user gesture.

### "should load stuff after reloading by goto" 
Module flag wasn't reset between playwright runs

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/apps/basics/src/routes/load/fetch-headers.svelte
+++ b/packages/kit/test/apps/basics/src/routes/load/fetch-headers.svelte
@@ -16,6 +16,7 @@
 	export let headers;
 
 	const json = JSON.stringify({
+		connection: headers['connection'],
 		'sec-fetch-dest': headers['sec-fetch-dest'],
 		'sec-fetch-mode': headers['sec-fetch-mode'],
 		'if-none-match': headers['if-none-match'],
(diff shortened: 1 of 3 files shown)


Rich-Harris (approved):

Rich-Harris:
thank you!

PR #4851, opened 2022-05-08
feat: `create-svelte` support for scoped packages

### LINKED ISSUE #4754: `create-svelte`'s `to_valid_package_name` modifies valid package names
### Describe the bug

According to [NPM's docs for `package.json`](https://docs.npmjs.com/creating-a-package-json-file);

> The "name" field contains your package's name, and must be lowercase and one word, and may contain hyphens and underscores.

Which if taken in a vacuum, the `to_valid_package_name` function correctly enforces.
What it doesn't correctly enforce is the use of scopes, eg `@sveltejs/kit`. Instead, it replaces the `@` and `/` with hyphens.

### Reproduction

The easiest method is to locally install `create-svelte` and pass the following when creating the package;

```js
import { create } from "create-svelte";

create("./emptyDir", { name: "@scope/package",
 /* these two don't matter, so long as they're valid */ template: "skeleton", types: "typescript" });
```

Observe the `package.json` `"name"` field doesn't read `"@scope/package"`, bet instead reads `"-scope-package"`

I traced where that `name` field goes, and the only place it's modified is [this line](https://github.com/sveltejs/kit/blob/master/packages/create-svelte/index.js#L139)

[Repo with minimum reproduction](https://github.com/willster277/sveltejs-kit-issue-4754)

### Logs

```shell
N/A
```

### System Info

```shell
System:
    OS: Linux 5.16 Pop!_OS 21.10
    CPU: (8) x64 Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz
    Memory: 9.84 GB / 15.36 GB
    Container: Yes
    Shell: 5.1.8 - /bin/bash
  Binaries:
    Node: 17.6.0 - ~/.nvm/versions/node/v17.6.0/bin/node
    npm: 8.5.1 - ~/.nvm/versions/node/v17.6.0/bin/npm
  Browsers:
    Chrome: 100.0.4896.127
    Chromium: 100.0.4896.127
    Firefox: 99.0
```

### Severity

annoyance

### Additional Information

This is working for me, though it's possible these simple lookbehinds miss a rare use case;
```diff
/** @param {string} name */
function to_valid_package_name(name) {
  return name
    .trim()
    .toLowerCase()
    .replace(/\s+/g, '-')
    .replace(/^[._]/, '')
-    .replace(/[^a-z0-9~.-]+/g, '-');
+    .replace( /[^a-z0-9~.-@\/]+/g, '-' )
+    .replace( /^(?<=.)@/g, '-' )
+    .replace( /^(?<!@.+)\//g, '-' );
}
```

tcc-sejohnson:
Fixes #4754.

Currently `create-svelte` does not support creating projects with scoped names, i.e. `@my-fancy-scope/my-fancy-package`. This is a crack at fixing that in a way that's as un-frustrating to developers as possible.

The current approach is to try to turn the base path of the provided project location into a package name. This has some weird edge cases, and is especially gross with `@scope/package` semantics, as that's obviously two folders (or a folder and a file with no extension) from a filesystem perspective.

For ease of understanding, here's the current behavior vs. the new behavior:

- Current
    - Accept the project path as the first positional argument
    - If that's not provided, ask for a path to put the project in
    - Do our best to yoink the package name from the provided path (eg. `./@myscope/mypackage` becomes `mypackage`, and `./.my folder name with spaces` becomes `my-folder-name-with-spaces`).
- New
    - Continue to accept the project path as the first positional argument
    - Prompt user for package name. Use a validator to make sure the provided name is valid (accepts both scoped and normal package names).
        - This does NOT kick the user out of the interface for providing a bad package name -- it just provides a descriptive error message and prevents them from continuing until they provide a valid one.
    - If the user provided the CLI folder path, continue as normal using that folder.
    - If not:
        - Prompt the user `Where should we create your project? (leave blank to use default: ${provided_package_name})`
        - If blank, use the provided package name as the folder. Else, use the provided folder.Examples:
            - `@my-scope/my-package => ./@my-scope/my-package`
            - `my-package => my-package`
            - All package characters are allowed filesystem characters, so there shouldn't be any problems with this approach.

All in all, this adds one prompt and makes the project naming much more user-controllable. At worst, it requires typing out one more answer than current -- at best, it's still the same number of inputs.

One other fringe benefit of this is that the `to_valid_package_name` (and the scoped package edge-cases that come with it) function is no longer needed at all -- since the package name is validated at the CLI, we know it's a valid package name already.

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -21,16 +21,41 @@ async function main() {
 	console.log(gray(`\ncreate-svelte version ${version}`));
 	console.log(disclaimer);
 
-	let cwd = process.argv[2] || '.';
-
-	if (cwd === '.') {
-		const opts = await prompts([
+	const package_name = await prompts(
+		[
 			{
 				type: 'text',
-				name: 'dir',
-				message: 'Where should we create your project?\n  (leave blank to use current directory)'
+				name: 'packagename',
+				message: 'What would you like to name your project?',
+				validate: (value) => {
+					const regex = new RegExp(/^(@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/);
+					const match = regex.test(value);
+					return !match
+						? 'Value is not a valid NPM package name. Package names must be lowercase and one word, and may contain hyphens and underscores, or be scoped and of the form @scope/package.'
+						: true;
+				}
 			}
-		]);
+		],
+		{
+			onCancel: () => process.exit(1)
+		}
+	);
+
+	let cwd = process.argv[2] || `./${package_name.packagename}`;
+
+	if (cwd !== process.argv[2]) {
+		const opts = await prompts(
+			[
+				{
+					type: 'text',
+					name: 'dir',
+					message: `Where should we create your project?\n  (leave blank to use default: ${cwd})`
+				}
+			],
+			{
+				onCancel: () => process.exit(1)
+			}
+		);
 
 		if (opts.dir) {
 			cwd = opts.dir;
@@ -113,7 +138,7 @@ async function main() {
 		)
 	);
 
-	options.name = path.basename(path.resolve(cwd));
+	options.name = package_name.packagename;
 
 	await create(cwd, options);
 
--- a/packages/create-svelte/index.js
+++ b/packages/create-svelte/index.js
@@ -72,7 +72,7 @@ function write_common_files(cwd, options, name) {
 
 	pkg.dependencies = sort_keys(pkg.dependencies);
 	pkg.devDependencies = sort_keys(pkg.devDependencies);
-	pkg.name = to_valid_package_name(name);
+	pkg.name = name;
 
 	fs.writeFileSync(pkg_file, JSON.stringify(pkg, null, '  '));
 }
@@ -133,14 +133,4 @@ function sort_keys(obj) {
 		});
 
 	return sorted;
-}
-
-/** @param {string} name */
-function to_valid_package_name(name) {
-	return name
-		.trim()
-		.toLowerCase()
-		.replace(/\s+/g, '-')
-		.replace(/^[._]/, '')
-		.replace(/[^a-z0-9~.-]+/g, '-');
 }
--- a/.changeset/cuddly-stingrays-boil.md
+++ b/.changeset/cuddly-stingrays-boil.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Now asks for project name instead of inferring it from filepath


packages/create-svelte/bin.js
```diff
@@ -21,16 +21,41 @@ async function main() {
...
 				type: 'text',
-				name: 'dir',
-				message: 'Where should we create your project?\n  (leave blank to use current directory)'
+				name: 'packagename',
+				message: 'What would you like to name your project?',
+				validate: (value) => {
+					const regex = new RegExp(/^(@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/);
+					const match = regex.test(value);
+					return !match
+						? 'Value is not a valid NPM package name. Package names must be lowercase and one word, and may contain hyphens and underscores, or be scoped and of the form @scope/package.'
```
Mlocik97, line 34: isn't `match` already boolean value? Also I don't know if You want to put error message here. (When you are not throwing it there)...

`return match ?? 'Value is not...'`

packages/create-svelte/index.js
```diff
@@ -72,7 +72,7 @@ function write_common_files(cwd, options, name) {
 
 	pkg.dependencies = sort_keys(pkg.dependencies);
 	pkg.devDependencies = sort_keys(pkg.devDependencies);
-	pkg.name = to_valid_package_name(name);
+	pkg.name = name;
```
Mlocik97, line 75: I think the way you implemented this is worse than actual implementation...

now, if you initialize project in existing folder like `some@folder#`, it will throw error instead converting it to correct package name `some-folder`

dominikg:
Some thoughts:

If you use create-svelte to initialize an application, asking for a package name isn't really relevant and could be confusing. In fact if the user doesn't want to publish a package, it would be best if `"private": true` was set to prevent mishaps

If you use it to initialize a library, esp. within an org, it is very likely that the org directory already exists on disk and you run `pnpm create svelte` inside that, so creating a deep path might backfire (or users prefer it without the @ prefix).

If we end up asking for a package name and validating it, this one is the official utility to do so: https://github.com/npm/validate-npm-package-name

packages/create-svelte/bin.js
```diff
@@ -21,16 +21,41 @@ async function main() {
...
+	const package_name = await prompts(
+		[
 			{
 				type: 'text',
-				name: 'dir',
-				message: 'Where should we create your project?\n  (leave blank to use current directory)'
+				name: 'packagename',
+				message: 'What would you like to name your project?',
+				validate: (value) => {
+					const regex = new RegExp(/^(@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/);
```
dominikg, line 31: ~  is no longer allowed for new packages, see https://github.com/npm/validate-npm-package-name

packages/create-svelte/bin.js
tcc-sejohnson: Ah, you're sort of right. The error message would never be returned with `??` (nullish coalescing) -- but I should be using `matched || "Blah blah blah"`. I'll fix it later.

packages/create-svelte/index.js
tcc-sejohnson: Incorrect, it is impossible to provide an invalid package name as currently implemented -- the folder name is no longer used as the package name; the package name (which is supplied by the user and validated before acceptance) is used as the default folder name. Instead of _converting_ the folder name to a valid (and sometimes really weird) package name, we're now just starting with a guaranteed-to-be-valid package name.

Mlocik97: you actually just repeated my words about how it's from another input... I just say this is not really good solution and it should be derived from folder name. That's how `npm init -y` behave itself.

tcc-sejohnson: > now, if you initialize project in existing folder like some@folder#, it will throw error instead converting it to correct package name some-folder even when name of folder is valid.

I was responding to this, which is just plain incorrect. 

As for the rest of your comment, not wanting to add another input is a perfectly valid concern, but not one best discussed here, on a hidden comment, on an unrelated line of code. I'm going to resolve this thread, as that same concern is going to come up in my top-level reply to the last comment on this PR -- feel free to provide your feedback there after I've made it. If you have a better idea on how to support scoped package names (which can't really be derived from a provided path without some _very_ surprising edge cases), I would love to hear them!

packages/create-svelte/bin.js
tcc-sejohnson: Interesting! That's probably a second bug fix for us -- the current function we're using to convert folder names to package names is still allowing them. I'll likely change this implementation to use the library you linked in your top level comment anyway -- I didn't realize that was an official NPM library when I saw it!

tcc-sejohnson:
> If you use create-svelte to initialize an application, asking for a package name isn't really relevant and could be confusing. 

I'd agree... to an extent?

IMO, if you're creating a Node application, the package name is _always_ relevant, but others may disagree. (More below.)

> In fact if the user doesn't want to publish a package, it would be best if "private": true was set to prevent mishaps

I'm not really sure what mishaps we'd be preventing (oops, I accidentally went all the way through the multistep process of releasing a package when it was supposed to be private? 😂). Maybe I'm just not thinking of something obvious. 

> If you use it to initialize a library, esp. within an org, it is very likely that the org directory already exists on disk and you run pnpm create svelte inside that, so creating a deep path might backfire (or users prefer it without the @ prefix).

Yeah, I supposed what the default is depends on what people are normally doing. I normally run `npm create svelte` from my `repos` directory and then provide my package name as the response to "where should we create it" -- which doesn't work for scoped packages. Even so, `npm init svelte .` is two characters longer and will initialize the project in the current directory without prompting for a location. 🙂

> If we end up asking for a package name and validating it, this one is the official utility to do so: https://github.com/npm/validate-npm-package-name

Thank you! I'm going to incorporate this into this PR.

Just to add to the discussion and source your feedback (and @Mlocik97's as he seems pretty invested in this one):

I cannot think of a better way to support scoped packages. No matter what, it seems like there's no way to get around asking for a package name. Deriving a scoped package from the file system seems error-prone and confusing. You could add a second prompt after asking for the location ("Would you like the current package name?: ${package_name_derived_from_folder}"), but that still involves a second prompt that, unlike the current implementation, can't be skipped by providing an argument to the CLI.

The only other palatable option I could think of is providing an override package name to the CLI as the second positional argument (`npm init svelte . @my-scope/my-package`), but that seems... hidden. 

If we don't want to support scoped package names at all, we can just close this PR and the associated issue. 🙂

packages/create-svelte/index.js
Mlocik97: I look on your code, and I don't see how it could derive package name, because I don't see any derivation logic as is used now.

Mlocik97:
> Yeah, I supposed what the default is depends on what people are normally doing. I normally run npm create svelte from my repos directory and then provide my package name as the response to "where should we create it" -- which doesn't work for scoped packages. Even so, npm init svelte . is two characters longer and will initialize the project in the current directory without prompting for a location. :slightly_smiling_face:

I really don't know what is bad about this... how `npm init svelte` vs `npm init svelte .` works is actually correct for me... I wouldn't change this.

> I cannot think of a better way to support scoped packages. No matter what, it seems like there's no way to get around asking for a package name. Deriving a scoped package from the file system seems error-prone and confusing.

How, `npm init -y` does it. Yeah they don't validate it, but they still derive it. I expect same for every `npm init <something>`. It definitelly should be derived and user shouldn't be asked for package name, as dominikg already said. It could be even more confusing behavior. Let make behavior as close to `npm init -y` as possible. Consistency is important...

I already wrote how it should work (in original issue), and I don't think it's bug-prone or confusing, actually opposit.

Mlocik97:
What behavior I expect is:

1 `npm init svelte` inside `my-app` folder.

will generate app inside that folder and set pkg.name to `my-app`.

2. `npm init svelte` inside `som@eap#` folder

will generate app inside that folder and set pkg.name to `som-eap`

3. `npm init svelte @comp/someapp`

will generate app inside `someapp` folder and set pkg.name to `@comp/someapp`

4. `npm init svelte @com#s/s@app/thing`

will generate app inside `s@app/thing` folder and set pkg.name to `@com-s/s-app-thing`

5. `npm init svelte some/app`

will generate app inside `some/app` and set pkg.name to `app`

Mlocik97:
actually I got another idea.

What about having it all as it is now... aka exactly how `to_valid_package_name()` works, but do something like:

```js
const scope = process.argv[3];

pkg.name = scope ? `@${to_valid_package_name(scope)/to_valid_package_name(name);}` : to_valid_package_name(name);
```

tcc-sejohnson:
@Mlocik97

I had a similar idea that I like _slightly_ better.

Right now, the CLI looks like:

```bash
npm init svelte {path}
```

We just update it to look like (similar but not identical to your suggestion):

```bash
npm init svelte {path} {packageNameOverride}
```

For calls that specify path but not a package name override, we'll parse the package name from the path as we currently do (with an update that checks to see if the second folder from the leaf starts with an @ - if it does, it'll parse it as a scoped package).

For calls that specify a path _and_ a package name, it'll use both verbatim (obviously validating the package name argument).

Give me some time today to implement. I think it's the cleanest solution -- the CLI will still work _exactly_ as it currently does, but with the option to specify the package name separate from the path on the CLI.

dominikg:
Please do not put work into this before it was discussed and decided **if** `create-svelte` should support it and how exactly.

more alternatives
1) just leave it as is and you'll just update package.json manually after create-svelte is done. It's  a one-time one-line edit for you keeps create-svelte small and doesn't ask unrelated questions to new users who might not even know about npm package scopes
2) create a custom starter template for your scope and clone that instead of using create-svelte
3) build a different cli that asks advanced questions

Mlocik97:
dominkg

if it would be introducing another param, I don't think it's problem. without that param whole create-svelte would works exactly as it does now. So I don't see any breaking change or confusion... we just introduce another param for scope or pkg.name overwrite. I don't see how this would be problematic.

tcc-sejohnson

I don't think dominikg said this PR is wrong, more than he said they need time to discuss it with team to decide how it will work.

tcc-sejohnson:
@Mlocik97

I understand! I still have the work saved. I would rather not clog up the PR list (it's already pretty long). Whenever the maintainers join the discussion on the linked issue, if they decide to support a change, I'll open a new PR. They seem pretty overwhelmed (not in a bad way!) with the number of issues/PRs open in the 1.0 milestone right now, so I didn't want to have something requiring additional discussion open. If/when they tag more "help wanted" issues (or when I find a way to get in on conversations involving non-help-wanted issues), I'll jump on them.

PR #4876, opened 2022-05-11
fix: more reliable typing of JSONObject

GauBen:

### Tests

### Changesets

---

Before:
image

It ends with:

```
 Type 'Stuff' is not assignable to type 'JSONObject'.
    Index signature for type 'string' is missing in type 'Stuff'
```

After:
image

I have no idea what's the difference, but `Record` works.

I don't plan to investigate any further.

---

```ts
type Record<K extends keyof any, T> = {
    [P in K]: T;
};
```

### DIFF
--- a/.changeset/gorgeous-ads-kneel.md
+++ b/.changeset/gorgeous-ads-kneel.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/kit": patch
+---
+
+fix: more reliable typing of JSONObject
--- a/packages/kit/types/private.d.ts
+++ b/packages/kit/types/private.d.ts
@@ -136,9 +136,7 @@ export interface CspDirectives {
 
 export type HttpMethod = 'get' | 'head' | 'post' | 'put' | 'delete' | 'patch';
 
-export interface JSONObject {
-	[key: string]: JSONValue;
-}
+export type JSONObject = Record<string, JSONValue>;
 
 export type JSONValue =
 	| string


Rich-Harris:
This is failing CI because this...

```ts
export type JSONObject = Record<string, JSONValue>;
```

...introduces a circular dependency between `JSONObject` and `JSONValue`. I don't know why that doesn't happen with interfaces or how to fix it.

ignatiusmb:
Although it used to be about type or interfaces, [that was now fixed](https://github.com/microsoft/TypeScript/issues/14174). The difference here lies in how we define them, one with the a generics and the other directly as an object literal. A type alias can't be referenced by itself at the top level, and the trick with generics is that they will unpack/expand anything that is passed upon them, which means it won't count as 'literals' even though we're abstracting that object literal in another type — what the `Record` implementation is.

Which also means it won't satisfy this rule (from the [archived spec](https://github.com/microsoft/TypeScript/blob/c300fea3250abd7f75920d95a58d9e742ac730ee/doc/spec-ARCHIVED.md#3.10))
> Note that object type literals, function type literals, and constructor type literals do not depend on types referenced within them and are therefore permitted to circularly reference themselves through type aliases.

Rich-Harris:
So is there any way to make [this](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgFIGUDyA5TAjAKwgTGQG8AoASAG0BrCATwC5kBnMKUAcwF1WMOAGpwANgFcIAbgoBfChTCMADikHYRElAF5qAH3ace+5CHEBbPNBN4A9rdEQ4IE2dGiT4kABMIMUBDeJgAqtuom6pqSNLwRWLiExGAyFKCQsIgooerk1GBh8QAUAHSlcFDcbKzOjDEAlKwAogAeCBK+ADyRYpIANMjZ8QB8MvIUAPTjyAC0s3Pzcwpp0PBIyADqABZwkABu0LnIR8gw9qwAjKMKMF4kwLYgyIhIymBsAPoEbA-vtokkhT+BAE8XwRBIdUOxwQD2+jmKols3EBhDqVwoMJAHGQAHdtntoKwtjsIPsoMhtOQTmdkOdkLIUs8IK8Pl8fkCkoU8SSyWiKEyWZ9viBfv8wIUyNTbBd6WigA) work?

ignatiusmb:
I suppose, [yeah](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgFIGUDyA5TAjAKwgTGQG8AoASAG0BrCATwC5kBnMKUAcwF1WMOAGpwANgFcIAbgoBfChTCMADikHYRElAF5qAH3ace+5CHEBbPNBN4A9rdEQ4IE2dGiT4kABMIMUBDeJgAqtuom6pqSNLwRWLiExGAyFKCQsIgooerk1GBh8QAUAHSlcFDcbKzOjDEAlKwAogAeCBK+ADyRYpIANMjZ8QB8MvIUAPTjyAC0s3PzcwpKqsgA6gAWcJAAbtDI2rnIR8gw9qwAjKMKMF4kwLYgyIhIymBsAPoEbA-vtokkhT+BAE8XwRBIdUOxwQD2+jmKols3EBhDqVwoMJAHGQAHdNjtoKwNlsILsoPtyCczshzshZClnhBXh8vj8gUlCniSWS0RRGczPt8QL9-mBCmQqbYLnS0UA)?

Rich-Harris:
closing in favour of #4897

Issue #4864, opened 2022-05-10
Document that `invalidate` on current URL reloads the page


tv42:
### Describe the problem

The docs for `invalidate` at https://kit.svelte.dev/docs/modules#$app-navigation-invalidate say

> Causes any `load` functions belonging to the currently active page to re-run if they `fetch` the resource in question.

The way it is written, it sounds like "if `load` calls `fetch` with URL X, then `invalidate(X)` reloads the page".

However, observed behavior is that you can also invalidate the *page itself*.

https://stackoverflow.com/questions/71743867/refresh-current-page-in-sveltekit

I think this should be explicitly documented, both for discoverability and to say that that is the intended behavior, not a random side effect that might go away.

### Describe the proposed solution

> Causes any `load` functions belonging to the currently active page to re-run if they `fetch` the resource in question, or if the current page URL is invalidated.

### Alternatives considered

Do only what the docs promised, but then something else needs to provide the reload functionality -- not a good idea, and breaks existing code.

### Importance

would make my life easier

PR #4891, opened 2022-05-12
Fix double-mount bug

### LINKED ISSUE #2130: __layout.svelte may run twice after browser reload in dev

### LINKED ISSUE #4799: Double component mount on $session change after component change and page refresh in the dev mode
### Describe the bug

If I start the svelte kit in the dev mode (`npm run dev`) then change a `*.svelte` file and then refresh the http://localhost:3000 page in the browser, then `onMount` will be called twice (on the first load and on the `$session` change).

### Reproduction

1. Create the svelte kit app:

**_src/routes/index.svelte:_**
```
<script>
	import {session} from '$app/stores'
	import {onMount} from 'svelte'

	onMount(() => {
		console.log('onMount')
	})
</script>

<div>{$session.counter}</div>

<button type='button' on:click={() => {
	session.update(session => {
		return {
			counter: session.counter + 1
		}
	})
}}>Click me</button>
```

**_app.html:_**
```
<!DOCTYPE html>
<html lang="en">
	<head>
		<meta charset="utf-8" />
		%svelte.head%
	</head>
	<body>
		<div>%svelte.body%</div>
	</body>
</html>

```

**_hooks.js:_**
```
export function getSession(event) {
	return {
		counter: 0
	}
}
```

**_package.json:_**
```
{
  "name": "my-app",
  "version": "0.0.1",
  "scripts": {
    "dev": "svelte-kit dev",
    "build": "svelte-kit build",
    "package": "svelte-kit package",
    "preview": "svelte-kit preview",
    "prepare": "svelte-kit sync"
  },
  "dependencies": {
    "@sveltejs/adapter-auto": "next",
    "@sveltejs/kit": "next",
    "svelte": "^3.48.0"
  },
  "type": "module"
}
```

**_svelte.config.js:_**
```
import adapter from '@sveltejs/adapter-auto'

export default {
	kit: {
		adapter: adapter()
	}
}
```

2. Execute `npm install`
3. Execute `npm run dev`
4. Open the http://localhost:3000/ page in the browser and open the browser console
5. Click to the "Click me" button. As you can see, `onMount` doesn't triggered on click (this is the correct behavior).
6. Now change the `index.svelte` file. For example, replace `console.log('onMount')` to `console.log('onMount2')`
7. Refresh the page in the browser (F5, Ctrl+R, etc.)
8. Click to the "Click me" button

# Expected result
`onMount` is triggered only on the page load and doesn't triggered on click

# Actual result
`onMount` is triggered on the page load and on the click (i.e. on the `session.update()`)

### Logs

_No response_

### System Info

```shell
@sveltejs/kit: 1.0.0-next.324
@sveltejs/vite-plugin-svelte: 1.0.0-next.42
svelte: 3.48.0
vite: 2.9.7
browser: Chromium 101.0.4951.41
OS: Linux Ubuntu 20.10
nodejs: v17.0.1
```

### Severity

annoyance

### Additional Information

_No response_

Rich-Harris:
Stumbled upon a nasty bug while investigating #4864 — if you edit `src/routes/a/[b].svelte`, then any navigation from `/a/1` to `/a/2` will cause `[b].svelte` to be reloaded (with the HMR cache-busting timestamp), then unmounted and remounted.

I think the reason is that we have an `import('/src/routes/[b].svelte')` in the HTML _and_ in the routes manifest; the latter import is cachebusted but the one in the HTML isn't.

This only occurs during dev.

* fixes #4799
* fixes #2130 

### Tests

### Changesets


Rich-Harris:
The reason this is failing CI is that changing how the initial page components are loaded has surfaced an unrelated bug: `vitePreload` fails in `svelte-kit preview` if `paths.assets` is specified, because in preview we replace it with `/_svelte_kit_assets` in order to serve assets locally. Previously the bug was masked (at least in the tests) because the components were already loaded via `import(...)` and `<link>`, meaning the router didn't try to load them via `vitePreload`.

This will hopefully be fixed by https://github.com/vitejs/vite/pull/7644

PR #4912, opened 2022-05-13
New page option for "key"

### LINKED ISSUE #3506: Svelte is reusing component with its state on different pages using the same route
### Describe the bug

Svelte routing reuses the component with its local state.
I have a route (/counter/[id]) That is using the sveltekit demo counter component.
Navigating from one counter/[id] page to another keep the state of the counter component
The only way to make it work as I want is to use the `target="_self"` attribute on the link which is forcing a full page reload (not ideal)

Sorry if this is expected behavior, it looks like a bug to me.

### Reproduction

Demo: https://routing-issue-theta.vercel.app/
Repo: https://github.com/plrenaudin/svelte-routing-issue

### Logs

_No response_

### System Info

```shell
System:
    OS: Linux 5.14 Solus 4.3
    Memory: 1.73 GB / 15.46 GB
    Container: Yes
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.1 - ~/.nvm/versions/node/v14.16.1/bin/node
    npm: 7.24.2 - ~/.nvm/versions/node/v14.16.1/bin/npm
  Browsers:
    Brave Browser: unknown
    Firefox: 96.0
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.14 
    @sveltejs/kit: next => 1.0.0-next.239 
    svelte: ^3.44.0 => 3.46.2
```

### Severity

### Additional Information

_No response_

johnnysprinkles:
Fixes #3506

Highly doubt this would be accepted but, maybe still useful in case anyone wants to patch it in.

I wanted to consider each logical URL as distinct as far as routing is concerned. Unlike query strings which could be considered the same location with different params, "path params" aren't exactly a real thing at least as far as old school CGI is concerned. `/foo/1` and `/foo/2` are just two different locations.

If you want to tear down all the state and re-instantiate your route component when the path params change, and not have to mess around with `afterNavigation`, this gives you a way.

### Tests

### Changesets

### DIFF
--- a/.changeset/small-mirrors-warn.md
+++ b/.changeset/small-mirrors-warn.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': minor
+---
+
+New page option for "key"
--- a/documentation/docs/11-page-options.md
+++ b/documentation/docs/11-page-options.md
@@ -22,6 +22,26 @@ In certain circumstances you might need to disable [client-side routing](/docs/a
 
 Note that this will disable client-side routing for any navigation from this page, regardless of whether the router is already active.
 
+### key
+
+By default, changing route parameters isn't considered a route change in SvelteKit. It's considered the same route with
+params and loaded data that update reactively via props.
+
+If instead you want to treat each logical URL as a separate route, with its own component instance and lifecycle, you can do that
+by exporting a key function. Internally, SvelteKit wraps the top-level components in a `{#key}` block so they can be destroyed
+and recreated based on arbitrary conditions, even when navigating to another route of the same pattern.
+
+The function accepts an object with url (URL instance) and params (a plain JS object), and returns a key used for
+equality comparison in a `{#key}` block.
+
+```html
+<script context="module">
+	export function key({ url, params }) {
+		return url.pathname;
+	}
+</script>
+```
+
 ### hydrate
 
 Ordinarily, SvelteKit [hydrates](/docs/appendix#hydration) your server-rendered HTML into an interactive page. Some pages don't require JavaScript at all — many blog posts and 'about' pages fall into this category. In these cases you can skip hydration when the app boots up with the app-wide [`browser.hydrate` config option](/docs/configuration#browser) or the page-level `hydrate` export:
--- a/packages/kit/src/core/sync/write_root.js
+++ b/packages/kit/src/core/sync/write_root.js
@@ -21,16 +21,20 @@ export function write_root(manifest_data, output) {
 
 	let l = max_depth;
 
-	let pyramid = `<svelte:component this={components[${l}]} {...(props_${l} || {})}/>`;
+	let pyramid = `{#key key_${l}}<svelte:component this={components[${l}]} {...(props_${l} || {})}/>{/key}`;
 
 	while (l--) {
 		pyramid = `
 			{#if components[${l + 1}]}
-				<svelte:component this={components[${l}]} {...(props_${l} || {})}>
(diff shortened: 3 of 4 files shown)


Rich-Harris:
As you predicted, I'm a little conflicted about this. Wanting to re-run the contents of `<script>` on state changes (and navigation is really just a type of state change) feels like imposing a React mindset (where `const` means the opposite of `const`) on Svelte.

If you had a situation like this...

```svelte
{#if number > 10}
  <BigNumber {number}/>
{:else}
  <LittleNumber {number}/>
{/if}
```

...you wouldn't expect `<BigNumber/>` to be destroyed and recreated when `number` goes from 11 to 12 — instead, in Svelte, we make things reactive to changing props. So why would we expect a page to be destroyed and recreated when _its_ props change?

Perhaps we can come up with a convincing reason why navigation is a sufficiently special kind of state change that it _does_ make sense to destroy and recreate components. But that reasoning would apply to layouts as well as leaves — if the argument is that arriving at `/id/4` should behave the same way whether you came from `/id/3` or `/foo`, then we need to destroy and remount `id/__layout.svelte` as well as `id/[x].svelte`. In most cases, we probably don't want that, since it will destroy sidebar scroll state etc. I mention this because I think that if the behaviour is to have first class support in the framework it's important for us to articulate a coherent argument for why it's is desirable.

Regardless, I don't think the `key` approach — ingenious as it is — will fly, unfortunately. If navigating from `/id/3` to `/id/4` should cause a destroy-and-remount, then so should navigating from `/id/4` to `/id/4`, and there's no realistic case in which the `key` will differ in that case. A better approach might be to return `preserve: false` from `load`. (I'm not endorsing this approach, because I'm sceptical of this idea overall, but if we _were_ to add a feature like this that's probably what I'd suggest!)

johnnysprinkles:
Yes but you might want to destroy and recreate `<BigNumber/>` for some reason, who knows why, otherwise why does the `#key` block even exist? For anything at a lower level than the route component you always have that option, this would just extend that option all the way up.

From a user perspective, it seems like resetting the state would actually be the common case. Say you're on Redfin browsing real estate and click "show more" on the description. If you navigate to another property, _of course_ you'd expect the description to start out collapsed. The "morph in place" T-1000 feel, where remnants of the previous page kind of hang around and maybe get reset and replaced by new data, or maybe they don't, is perhaps part of the reason SPAs get a bad rap.

Navigation is special! Well at least it is if you're building something with named locations, i.e. more of a website than an application. The instructions might say "In your web browser, go to /id/3" but they wouldn't say "In your web browser, rearrange to the shape of /id/3". On the web we go places, or least create a convincing illusion of going places.

Assuming you did want a clean slate when going between IDs, you could make some `reset()` function that you call from `afterNavigate` but that's error prone since you're likely to miss something as you add more state. A better option is to put all the real functionality in a sub-component and have the route component be a thin wrapper with a key block. This is ok but not that elegant, reminds me of the old Redux container components that you called `connect()` on. I am doing that currently in production though.

Your point about layout is interesting... I still haven't dug into how the nested layouts work, totally on my TODO list to learn that. Since multiple routes share a layout, yeah I'm not even sure what the right thing to do would be.

Good point about links to the page you're currently on, didn't even think of that. And it is important, since being on a homepage and clicking the logo to refresh the homepage is such a common pattern, that should translate to the SPA world.

Anyway, I like `preserve: false` just fine, maybe I'll explore that direction and possibly send a different PR for that.

Issue #4367, opened 2022-03-17
Error: Dynamic require of "crypto" is not supported


CoolOppo:
### Describe the bug

I'm getting this error on my Cloudflare Pages site but can't reproduce locally. I am feeding in 'esnext' to the adapter because I have to for anything to compile. Might be part of the problem, idk, but if that's the cause then I have another question: why am I not supposed to use ECMA from the last 2 years? ;)

from my svelte config:
```ts
adapter: adapter({
			target: "esnext",
			minify: false,
			logLevel: "error",
		})
```

I believe it's caused by this line:
https://github.com/sveltejs/kit/blob/827d1623d64a9886032c8bfc8e2315420d1599d0/packages/kit/src/runtime/server/page/csp.js#L24

### Reproduction

I am not trying to be a dick, I am very busy and don't have the time right now to reproduce. I've explained all steps necessary to reproduce; it's an extremely simple setup. Just want to report this bug. Sorry if that isn't allowed; I know you're all busy too.

### System Info

```shell
Cloudflare Workers node version 17.5.0
```

### Severity

blocking all usage of SvelteKit

CoolOppo:
Oh I think it may be caused by this line, just added a service worker so that may be it!

https://github.com/sveltejs/kit/blob/27a9888161b86bcc8418f87f93ed54959c631079/packages/kit/src/runtime/server/page/render.js#L276

benmccann:
Looks like this might get fixed when AWS gets off their asses to implement Node 16 support, so that we can drop Node 14

https://github.com/sveltejs/kit/blob/827d1623d64a9886032c8bfc8e2315420d1599d0/packages/kit/src/runtime/server/page/csp.js#L23

CoolOppo:
Just fixed it...using [this ESBuild plugin](https://github.com/sinnaj-r/esbuild-plugin-node-polyfills) in the meantime.

Instructions:

*svelte.config.js*
```js
import ESBuildNodePolyfillsPlugin from 'esbuild-plugin-node-polyfills'

// ...

adapter: adapter({
			target: "esnext",
			minify: true,
			logLevel: "error",
			plugins: [ESBuildNodePolyfillsPlugin]
		}),
```

CoolOppo:
This is a very ratchet way of doing this of course...iirc the changelog for adapter-cloudflare actually said that support for ESbuild options was removed, lol. But when you put in the options, they get in there. :p

Rich-Harris:
[...]

PR #4915, opened 2022-05-14
prevent esbuild trying to bundle node:crypto

Rich-Harris:
There's no good place to add a test for this, but I think it's fine because we can get rid of this whole mess as soon as we drop support for Node 14.

The issue was reported in https://github.com/sveltejs/kit/issues/4367#issuecomment-1122474272, and basically looks like this: if `config.kit.vite.build.minify === 'esbuild'`, the `'crypto'` string here...

https://github.com/sveltejs/kit/blob/7609ad0759c895600198eaf4345d735c7c6d1166/packages/kit/src/runtime/server/page/csp.js#L23-L25

...gets inlined into the `import(name)` expression, which means that a _subsequent_ esbuild run tries to import the `node:crypto` module, it will fail unless `platform: 'node'` is specified.

Making the expression more complex seems to prevent esbuild from being cute.

### Tests

### Changesets

### DIFF
--- a/.changeset/small-pumpkins-grab.md
+++ b/.changeset/small-pumpkins-grab.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Prevent esbuild trying to bundle node:crypto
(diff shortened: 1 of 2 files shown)


benmccann (approved):

Rich-Harris:
This will be superseded by a PR I aim to submit later today that moves the crypto stuff into a node polyfill module

xpluscal:
Has this ever been resolved?

PR #4920, opened 2022-05-14
[fix] deduplicate injected css during dev

### LINKED ISSUE #3497: Tailwind CSS is loaded twice in dev mode
### Describe the bug

It's actually the same bug as reported here: https://github.com/sveltejs/kit/issues/2192

I'm just gonna re-open a new once since I'm getting the same issue on recent ` SvelteKit v1.0.0-next.239`.

I've tried several css loading techniques, they all result in the exact same issue.

When running `svelte-kit dev` and taking a look at the source code, you have tailwindcss twice in there.

it's not happening on the client side, since I guess deduplication or something is happening, but its still confusing and annoying.

Take a look at this repo to reproduce: https://github.com/mrtnbroder/sveltekit-tailwindcss-twice-bug

in `__layout.svelte`, where `global.css` is simply the following:

```
@import "tailwindcss/base";
@import "tailwindcss/components";
@import "tailwindcss/utilities";
```

I've tested these `__layout.svelte`'s

```html
<script lang="ts">
  import "../styles/global.css"
</script>
```
OR
```
<style global lang="postcss">
  @import "../styles/global.css"
<style>
```
OR
```
<style global lang="postcss">
  @import "tailwindcss/base";
  @import "tailwindcss/components";
  @import "tailwindcss/utilities";
<style>
```
OR
```
<style global lang="postcss">
  @tailwind base;
  @tailwind components;
  @tailwind utilities";
<style>
```

### Reproduction

https://github.com/mrtnbroder/sveltekit-tailwindcss-twice-bug

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.1
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 2.13 GB / 16.00 GB
  Binaries:
    Node: 16.10.0 - ~/.nodenv/versions/16.10.0/bin/node
    npm: 7.24.0 - ~/.nodenv/versions/16.10.0/bin/npm
    Watchman: 2022.01.03.00 - /usr/local/bin/watchman
  Browsers:
    Brave Browser: 97.1.34.80
    Chrome: 97.0.4692.99
    Firefox: 95.0.2
    Safari: 15.2
  npmPackages:
    @sveltejs/kit: ^1.0.0-next.239 => 1.0.0-next.239
    svelte: ^3.46.2 => 3.46.2
```

### Severity

annoyance

### Additional Information

_No response_

mrkishi:
Fixes #3497

During dev, kit inlines css files in order to prevent FOUC. The current implementation injects all imported modules indiscriminately, but this conflicts with vite's own `@import` inlining.

Given a component:
```svelte
<script>
	import '$lib/entry.css';
</script>
```

And the respective css files:
```css
/* entry.css */
@import './base.css';
html { color: green; }
/* entry.css eof */

/* base.css */
html { color: red; }
/* base.css eof */
```

By walking the `ModuleGraph` we end up inlining both `entry.css`'s and `base.css`'s _transformed_ content. Since vite inlines `base` into `entry`, we get the following injected css:
```css
/* entry.css */
/* base.css */
html { color: red; }
/* base.css eof */
html { color: green; }
/* entry.css eof */

/* base.css */
html { color: red; }
/* base.css eof */
```

The first block is `entry` and the second is `base`. If hydration is turned off, you'll get the unexpected rules applied and a confusing inspector experience. If it's turned on, you'll get a flash of _incorrectly styled_ content.

This PR walks through `ssrTransformResult` dependencies instead, such that `base` is ignored since it's already inlined into `entry`. I'm not sure how to test for this, unfortunately.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -73,7 +73,7 @@ export async function create_plugin(config, cwd) {
 								if (!node) throw new Error(`Could not find node for ${url}`);
 
 								const deps = new Set();
-								find_deps(node, deps);
+								await find_deps(vite, node, deps);
 
 								/** @type {Record<string, string>} */
 								const styles = {};
@@ -387,14 +387,36 @@ function remove_html_middlewares(server) {
 }
 
 /**
+ * @param {import('vite').ViteDevServer} vite
  * @param {import('vite').ModuleNode} node
  * @param {Set<import('vite').ModuleNode>} deps
  */
-function find_deps(node, deps) {
-	for (const dep of node.importedModules) {
-		if (!deps.has(dep)) {
-			deps.add(dep);
-			find_deps(dep, deps);
+function find_deps(vite, node, deps) {
+	// since `ssrTransformResult.deps` contains URLs instead of `ModuleNode`s, this process is asynchronous.
+	// instead of using `await`, we resolve all branches in parallel.
+	/** @type {Promise<void>[]} */
+	const branches = [];
+
+	/** @param {import('vite').ModuleNode} node */
+	function add(node) {
+		if (!deps.has(node)) {
+			deps.add(node);
+			branches.push(find_deps(vite, node, deps));
 		}
 	}
+
+	/** @param {string} url */
+	async function add_by_url(url) {
+		branches.push(vite.moduleGraph.getModuleByUrl(url).then((node) => node && add(node)));
+	}
+
+	if (node.ssrTransformResult) {
+		if (node.ssrTransformResult.deps) {
+			node.ssrTransformResult.deps.forEach(add_by_url);
+		}
+	} else {
+		node.importedModules.forEach(add);
+	}
+
+	return Promise.all(branches).then(() => {});
 }
--- a/.changeset/yellow-apricots-stare.md
+++ b/.changeset/yellow-apricots-stare.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] deduplicate injected css during dev
--- a/packages/kit/test/apps/basics/src/routes/css/_base.css
+++ b/packages/kit/test/apps/basics/src/routes/css/_base.css
@@ -1,0 +1,3 @@
+.overridden {
+	color: red;
+}
--- a/packages/kit/test/apps/basics/src/routes/css/_styles.css
+++ b/packages/kit/test/apps/basics/src/routes/css/_styles.css
@@ -1,3 +1,9 @@
+@import '_base.css';
+
 .styled {
 	color: red;
+}
+
+.overridden {
+	color: green;
 }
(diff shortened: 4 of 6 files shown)


dominikg:
idea for testing: 
- setup a small e2e app with easy to grep styles 
- use `fetch(test_page_url).then(res=>res.text())`  to get ssr result
- check head in both ssr result and playwright page.get...
- add a check with el.getComputedStyle for good measure

mrkishi:
I didn't mean the technical steps of testing, but rather that I was unsure of a _useful_ test we could get out of this. 😛 

Checking the actual ssr output is brittle and tests for an implementation detail that doesn't add much, imo.

I think having a css rule override and checking the right rule is applied is about as good as we could get from this, since it could catch other manifestations of this bug—although it could just as easily happen in different orders and render such a test ineffective anyway.

Perhaps some clever use of css variables can make the test at least order-resistant, but I'll have to check it out later.

packages/kit/src/core/dev/plugin.js
```diff
@@ -387,14 +387,36 @@ function remove_html_middlewares(server) {
...
+
+	if (node.ssrTransformResult) {
+		if (node.ssrTransformResult.deps) {
+			node.ssrTransformResult.deps.forEach(add_by_url);
+		}
+	} else {
+		node.importedModules.forEach(add);
+	}
+
+	return Promise.all(branches).then(() => {});
```
Rich-Harris, line 421: This looks suspicious to me — it will only await the promises that are added to `branches` _before_ the `Promise.all`. Don't we need to await the ones added later?

Demo here — even though we push to the array before the first promise resolves, the result is `[1]` rather than `[1, 2]`:

```js
function resolve(value) {
  return Promise.resolve().then(() => {
    console.log(`resolving ${value}`);
    return value;
  });
}

array = [resolve(1)];

Promise.all(array).then((values) => {
  console.log('result', values);
});

array.push(resolve(2));

// resolving 1
// resolving 2
// result [1]
```

Given that we're mostly CPU-bound here, I wonder if we shouldn't just do everything serially?

mrkishi: All pushes to `branches` happen synchronously—each `find_deps` call has their own `branches`.

I thought this would be IO bound, because otherwise `getModuleByUrl` wouldn't need to be async, but I'll gladly make this serial instead if you think it isn't worth it.

mrkishi:
I couldn't think of any way to create a test for duplicate or arbitrarily ordered css rules without analyzing the generated output, so instead I just added another variation on the css tests that check the right rules are applied in the end. That should cover _enough_, especially considering this is such an edge case.

packages/kit/src/core/dev/plugin.js
Rich-Harris: d'oh! of course

Rich-Harris (approved):

mrkishi:
@vhscom only duplication. But what FOUC? All styles are inlined, I don't see FOUC normally.

Issue #4916, opened 2022-05-14
Drop Node 14 support


Rich-Harris:
### Describe the problem

Now that even AWS lambda have finally got their act together and shipped Node 16 support (just in time for Node 18 to be the [current LTS version](https://nodejs.org/en/about/releases/)), I'm not sure there's much of a reason to continue supporting Node 14?

Dropping it would simplify the test matrix slightly, and would allow us to get rid of this grotesque crypto stuff:

https://github.com/sveltejs/kit/blob/7609ad0759c895600198eaf4345d735c7c6d1166/packages/kit/src/runtime/server/page/csp.js#L23-L33

If Lambda supporting 16 isn't a good enough reason, then presumably we have to continue supporting 14 until it's out of LTS on April 30 next year. That doesn't seem ideal.

### Describe the proposed solution

Only support Node 16 and above

### Importance

would make my life easier

Mlocik97:
I would say it's little early... but not that much it would be big problem...

For example my systems will update to Node 16 together with system update (to Ubuntu 22.04.1 LTS at 4th august).

ghostdevv:
I don't use Node 14 myself but it would be good to keep in mind there might be some teams still using it, we still got just under a year of Node 14 being supported by Node:

![image](https://user-images.githubusercontent.com/47755378/168422690-22b0812b-53a1-4db9-8ee8-7fc49ed6dcd2.png)

Rich-Harris:
We debated it among the maintainers and support for dropping 14 was universal, since if Lambda supporting 16 isn't a solid enough reason then there'd be no point at which we can drop it prior to April 2023.

All the relevant deployment platforms support 16, and it's very easy for people to switch locally with tools like `nvm`, so the only reason we can see for someone being _unable_ to use 16 is that they have some dependency that only works with 14 (which candidly I'm unsympathetic to — 16 has been out for over a year and has been the active LTS version for well over 6 months)

PR #4922, opened 2022-05-14
[breaking] drop Node 14 support

benmccann:
closes https://github.com/sveltejs/kit/issues/4916

### DIFF
--- a/packages/adapter-vercel/index.js
+++ b/packages/adapter-vercel/index.js
@@ -396,7 +396,7 @@ function get_node_version() {
 	const major = parseInt(full.split('.')[0]); // '16.5.0' --> 16
 
 	if (major < 16) {
-		throw new Error(`SvelteKit only support Node.js version 16 or greater (currently using v${full}). Consult the documentation: https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version`)
+		throw new Error(`SvelteKit only supports Node.js version 16 or greater (currently using v${full}). Consult the documentation: https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version`)
 	}
 
 	return { major, full };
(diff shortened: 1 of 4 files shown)


packages/adapter-vercel/index.js
```diff
@@ -395,8 +395,8 @@ function get_node_version() {
 	const full = process.version.slice(1); // 'v16.5.0' --> '16.5.0'
 	const major = parseInt(full.split('.')[0]); // '16.5.0' --> 16
 
-	if (major < 14) {
-		throw new Error(`SvelteKit only support Node.js version 14 or greater (currently using v${full}). Consult the documentation: https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version`)
+	if (major < 16) {
+		throw new Error(`SvelteKit only support Node.js version 16 or greater (currently using v${full}). Consult the documentation: https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version`)
```
mrkishi, line 399:
```suggestion
		throw new Error(`SvelteKit only supports Node.js version 16 or greater (currently using v${full}). Consult the documentation: https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version`)
```

Rich-Harris:
AFK or I'd do this myself right now - we should get rid of the node 14 crypto stuff as well

Rich-Harris:
Actually, scratch that. We should change install-fetch to be install-polyfills or something and assign `require('node: crypto').webcrypto` to `globalThis.crypto`, and put any other things we need in there (like `AbortController` and `ReadableStream` etc). Will do that in a follow-up

Rich-Harris (approved):

Cinci7:
Your not done yet should of read the whole lesson plan when you had the chance.

PR #4923, opened 2022-05-14
[fix] get packaging test working on Windows

mrkishi:
Packaging's "watches for changes" test hasn't worked on Windows. Upon investigation it was because of two reasons:

- We need to wait for `chokidar`'s `ready` event before writing the files;
- Typescript uses `crlf` on the generated `.d.ts` files by default, causing a mismatch with our expected output.

### Tests

### Changesets


dominikg (approved):

Rich-Harris:
Niiiiice

Issue #3419, opened 2022-01-19
Handling streaming request/response bodies


Rich-Harris:
### Describe the problem

> Prior discussions: #70, #1563

As of #3384, hooks and endpoints receive a standard `Request` object, which among other things makes it possible to POST files:

```js
// handles submission from a <form enctype="multipart/form-data">
export async function post({ request }) {
  const body = await request.formData();
  const file = await body.get('myfile');
  // ...
}
```

The file is buffered in memory, however, which means there's a practical limit on how large it can be.

Similarly, there's no good way to read a stream of binary data, or respond with a stream.

### Describe the proposed solution

We need to solve this at a high level and at a low level. At the high level, it would be good to have utility functions for dealing with multipart form data specifically — something like this, perhaps:

```js
import { multipart } from '$app/somewhere';

export async function post({ request }) {
  for await (const part of multipart(request)) {
    if (part.filename) {
      const uploader = create_s3_uploader(BUCKET, part.filename);
      for (const chunk of part) {
        uploader.write(chunk);
      }
    }
  }

  return { status: 201 };
}
```

For octet streams it might look like this:

```js
import { stream } from '$app/somewhere';

export async function post({ request }) {
  for await (const chunk of stream(request)) {
    // ...
  }

  return { status: 201 };
}
```

At the low level, it should be possible to create your own stream reader and return your own `ReadableStream` bodies:

```js
export async function post({ request }) {
  const reader = request.body.getReader();

  let chunk;
  while (chunk = await reader.read()) {
    if (chunk.done) break;
    await do_something_with(chunk.value);
  }

  return {
    body: new ReadableStream({
      start(controller) {
        // ...
      },
      cancel() {
        // ...
      }
    })
  };
}
```

Unfortunately there's a wrinkle: `ReadableStream` isn't available in older versions of Node, and the `Request` and `Response` objects (as polyfilled by `node-fetch`) use node streams instead of `ReadableStream`. It may therefore be necessary to polyfill `ReadableStream` during development (and in `adapter-node`, and lambdas) and convert between that and node streams. (On the `request` object, this would probably mean creating a Proxy that replaces `body` with a `ReadableStream` with the same contents as the node stream. Hopefully this is actually possible.)

### Importance

would make my life easier

Rich-Harris:
[...]

PR #4934, opened 2022-05-15
Node polyfills

Rich-Harris:
This replaces `@sveltejs/kit/install-fetch` with `@sveltejs/kit/node/polyfills`, and adds `crypto` to the list of web APIs that are guaranteed to be present.

* [ ] ~~add `AbortController` (#4846)~~
* [ ] add web streams (#3419)
* [ ] add other interfaces listed on [Minimum Common Web Platform API](https://common-min-api.proposal.wintercg.org/)? if we wanted to get really sophisticated we'd detect which APIs were used, so that adapters could only bundle those
* [ ] swap out `node-fetch` for `undici`, which is more standards compliant (though I think we're blocked on https://github.com/nodejs/undici/issues/1262)
* [x] only add polyfills if they're not already there (since Node 18 provides a lot of stuff anyway. in future maybe we should have `@sveltejs/kit/node/polyfills/16` and `@sveltejs/kit/node/polyfills/18`)
* [x] document which APIs are guaranteed to be present

### Tests

### Changesets


ghostdevv:
Will this mean that libraries that rely on node dependencies in a non-node environment such as CloudFlare Pages have a chance of working? i.e. stripe

Rich-Harris:
No, it's the opposite — it's bringing the Node environment up to date with other more web-like ones, where `fetch` and `crypto` etc are already universally supported. If people need to use Node-specific dependencies in non-Node environments, they will need to use things like https://www.npmjs.com/package/rollup-plugin-node-builtins (though some of these have significant caveats).

Ideally, those libraries should be updated to use web APIs, instead of assuming they're running in Node.

ghostdevv:
Ah that makes sense, will try the link you gave but haven't had luck with similar packages and kit in the past :pray:

Rich-Harris:
Gah. We can't use native `fetch` in Node 18 because it's secretly just undici, and we can't use undici because of https://github.com/nodejs/undici/issues/1262. That's why those tests are failing.

The response returned from undici's `fetch` omits `set-cookie` headers:

```js
import { createServer } from 'http';
// import fetch from 'node-fetch';

const PORT = 3000;

createServer((req, res) => {
  res.writeHead(200, {
    'set-cookie': 'foo=bar'
  });
  res.end('ok');
}).listen(PORT, async () => {
  const res = await fetch(`http://localhost:${PORT}`);
  console.log(res.headers.get('set-cookie'));
});
```

This logs `null`; if you uncomment the `import fetch from 'node-fetch'` it logs `foo=bar`.

I'm _pretty_ sure this is specific to undici (hopefully we can open a pull request), but we should probably check that you can read `set-cookie` in Deno and Cloudflare Workers (and the runtimes that sit on top of them — Netlify and Vercel edge functions).

benmccann:
Just FYI, Vite is likely switching to `node-native-fetch`: https://github.com/vitejs/vite/pull/8157

Rich-Harris:
Added `crypto` to the [web standards](https://kit.svelte.dev/docs/web-standards) section of the docs, so marking this ready for review. The other stuff (upgrading to `undici`, etc) can wait until another time, no point waiting on that for everything else.

packages/kit/src/node/index.js
```diff
@@ -56,6 +57,7 @@ export async function getRequest(base, req) {
 	if (req.httpVersionMajor === 2) {
 		// we need to strip out the HTTP/2 pseudo-headers because node-fetch's
 		// Request implementation doesn't like them
+		// TODO is this still true with undici?
```
benmccann, line 60: maybe it'd be clearer to say Node 18 for people who aren't familiar with Undici?
```suggestion
		// TODO is this still true with Node 18
```

PR #4939, opened 2022-05-16
explicit recursive body validations

### LINKED ISSUE #4926: Cannot use nested interfaces to type endpoint output
### Describe the bug

This is a follow-up on #1997, which was recently fixed in #4897. 

The new `BodyValidator` type seems to work well, but it doesn't seem to work with interfaces that have properties which are themselves interfaces. 

### Reproduction

If you prefer a repository: check out [this file](https://github.com/flut1/sveltekit-endpoint-output-nested-interfaces-repro/blob/main/src/routes/index.ts) in [flut1/sveltekit-endpoint-output-nested-interfaces-repro](https://github.com/flut1/sveltekit-endpoint-output-nested-interfaces-repro)

```TypeScript
import type { RequestHandler } from "@sveltejs/kit";

interface GetOutput {
    item: { message: string }
}

// No TS error
export const get: RequestHandler<any, GetOutput> = async () => {
    return {
        status: 200,
        body: {
            item: { message: 'hello' },
        },
    };
};

interface Item {
    message: string;
}

interface PostOutput {
    item: Item;
}

// TS error: Type '{ message: string; }' is not assignable to type 'never'.ts(2322)
export const post: RequestHandler<any, PostOutput> = async () => {
    return {
        status: 200,
        body: {
            item: { message: 'hello' },
        },
    };
};
```

### Logs

Output of svelte-check:

```shell
sveltekit-endpoint-output-nested-interfaces-repro/src/routes/index.ts:26:14
Error: Type '() => Promise<{ status: number; body: { item: { message: string; }; }; }>' is not assignable to type 'RequestHandler<any, PostOutput>'.
  Type 'Promise<{ status: number; body: { item: { message: string; }; }; }>' is not assignable to type 'MaybePromise<RequestHandlerOutput<PostOutput>>'.
    Type 'Promise<{ status: number; body: { item: { message: string; }; }; }>' is not assignable to type 'Promise<RequestHandlerOutput<PostOutput>>'.
      Type '{ status: number; body: { item: { message: string; }; }; }' is not assignable to type 'RequestHandlerOutput<PostOutput>'.
        The types of 'body.item' are incompatible between these types.
          Type '{ message: string; }' is not assignable to type 'never'. 
export const post: RequestHandler<any, PostOutput> = async () => {
```

### System Info

```shell
System:
  OS: macOS 12.3.1
  CPU: (8) arm64 Apple M1
  Memory: 111.95 MB / 8.00 GB
  Shell: 5.8 - /bin/zsh
Binaries:
  Node: 16.15.0 - ~/.nvs/default/bin/node
  npm: 8.5.5 - ~/.nvs/default/bin/npm
Browsers:
  Firefox: 100.0
  Safari: 15.4
npmPackages:
  @sveltejs/adapter-auto: next => 1.0.0-next.42 
  @sveltejs/kit: next => 1.0.0-next.330 
  svelte: ^3.44.0 => 3.48.0 
```

### Severity

annoyance

### Additional Information

_No response_

ignatiusmb:
Taking the inverse of the current approach, rather than naively `extend`-ing `JSONValue`, we will explicitly recurse when hitting an object. Other than that, we'll check whether it's an invalid type that cannot be converted to JSON string rather than checking whether it's a valid JSON type, [that should be much simpler](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#exceptions).

Fixes #4926

### Tests

### Changesets

### DIFF
--- a/.changeset/five-cows-happen.md
+++ b/.changeset/five-cows-happen.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/kit": patch
+---
+
+fix `BodyValidator` for nested interfaces
--- a/packages/kit/test/typings/endpoint.test.ts
+++ b/packages/kit/test/typings/endpoint.test.ts
@@ -92,6 +92,18 @@ export const generic_case: RequestHandler<Record<string, string>, ExamplePost> =
 	};
 };
 
+interface NestedChild {
+	message: string;
+}
+interface ParentWrapper {
+	fields: NestedChild;
+}
+export const nested_interfaces: RequestHandler<Record<string, string>, ParentWrapper> = () => {
+	return {
+		body: {} as ParentWrapper
+	};
+};
+
 // --- invalid cases ---
 
 // @ts-expect-error - body must be JSON serializable
--- a/packages/kit/test/typings/endpoint.test.ts
+++ b/packages/kit/test/typings/endpoint.test.ts
@@ -106,10 +106,46 @@ export const nested_interfaces: RequestHandler<Record<string, string>, ParentWra
 
 // --- invalid cases ---
 
-// @ts-expect-error - body must be JSON serializable
-export const error_unserializable_literal: RequestHandler = () => {
+// @ts-expect-error - bigint cannot be converted to JSON string
+export const error_unserializable_literal_bigint: RequestHandler = () => {
+	return {
+		body: BigInt('')
+	};
+};
+// @ts-expect-error - bigint cannot be converted to JSON string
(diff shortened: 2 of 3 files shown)


dummdidumm (approved):
At first I thought it might be better to check the positive case (extends string|number|etc) but after reading the linked docs I agree that the negative check is better and more robust

Rich-Harris:
this goes straight over my head but i like it!

Issue #1575, opened 2021-05-28
Why is redirect from `__layout.svelte` ignored, when an unkown route is hit?


alexbjorlig:
I have a small reproducible example [here](https://github.com/alexbjorlig/sveltekit-redirect-question).

In the above repo, there is `__layout.svelte` file, that will redirect like this: 

```ts
return {
  status: 302,
  redirect: '/login'
}
```

Works great when a route is matched, but if there is an error the redirect is not respected.
The load function does actually run, but the browser is not redirected to login.

Is this a bug, not allowed - or do I misunderstand something here?

benmccann:
Can you explain your use case and why you'd want to redirect instead of showing the error page?

alexbjorlig:
I expect the following: 

1. User attempts to access application, an unknown route is hit
2. __layouat.svelte is then processed, and determines in this case it's best to redirect to `/login`


BUT the above is not what happens, **despite** the layout's load function actually being called(!)

benmccann:
Right. But what's your use case? I.e. what application are you building, what is the user doing in your application, what are you trying to accomplish?

alexbjorlig:
My usecase here is to redirect user to login, when trying to access a route that does not exists. Combined with layout resets, that would be a powerfull way to show some users 404 pages, and redirect others to login.

benmccann:
If you have a redirect on the error page as requested in https://github.com/sveltejs/kit/issues/1574 why do you need one on the layout to do that?

alexbjorlig:
Ok so It's a bit hard to explain without a conversation for me. But let's be concrete. Why is a redirect from __layout.svelte ignored? The load function runs, and returns a redirect. The redirect is then, silenty, ignored. That's is the purpose of this issue to put clarification to. As a developer, I expect a redirect to happen when I return a redirect 😅

benmccann:
> Why is a redirect from __layout.svelte ignored?

It's not ignored in the general case. But if you hit an error then it might not have a chance to execute. That actually seems like the more intuitive behavior in my mind.

Kapsonfire-DE:
But the __error.svelte is using the __layout.svelte, isnt it?
so the redirect should execute. this is not the case - I had this issue before. 
Why I needed it: I didnt want to give the user a 404 Error page for my mobile game - Instead I wanted to recover to a proper state (home route). I ended up, to make a redirect on the __error.svelte itsself - this worked.

[...]

PR #4945, opened 2022-05-16
document that status, error, redirect, cache are ignored when rendering error pages

Rich-Harris:
See #1575. We might want to add some warnings when these properties are encountered when rendering error pages, but the first step is to document the expected behaviour

### DIFF
--- a/documentation/docs/04-loading.md
+++ b/documentation/docs/04-loading.md
@@ -103,7 +103,7 @@ If the page you're loading has an endpoint, the data returned from it is accessi
 
 ### Output
 
-If you return a Promise from `load`, SvelteKit will delay rendering until the promise resolves. The return value has several properties, all optional:
+If you return a Promise from `load`, SvelteKit will delay rendering until the promise resolves. The return value has several properties listed below, all of which are optional.
 
 > `status`, `error`, `redirect` and `cache` are ignored when rendering error pages.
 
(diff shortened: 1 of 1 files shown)


documentation/docs/04-loading.md
```diff
@@ -144,6 +144,8 @@ This will be merged with any existing `stuff` and passed to the `load` functions
 
 The combined `stuff` is available to components using the [page store](/docs/modules#$app-stores) as `$page.stuff`, providing a mechanism for pages to pass data 'upward' to layouts.
 
+> `status`, `error`, `redirect` and `cache` are ignored when rendering error pages.
```
benmccann, line 147: putting this note in `stuff` seems like a surprising place for it. maybe under `status` would work?

Rich-Harris: oh, this was supposed to go at the end of `Output`, but i managed to put it before `dependencies`. i think it probably belongs at the end or at the top, rather than between items

benmccann: end of `Output` sounds good too

documentation/docs/04-loading.md
```diff
@@ -105,6 +105,8 @@ If the page you're loading has an endpoint, the data returned from it is accessi
 
 If you return a Promise from `load`, SvelteKit will delay rendering until the promise resolves. The return value has several properties, all optional:
```
benmccann, line 106: Should we no longer have this end with `:` since the next thing is the warning box?
```suggestion
If you return a Promise from `load`, SvelteKit will delay rendering until the promise resolves. The return value has several properties listed below, all of which are optional.
```

benmccann (approved):
lgtm. just one possible suggestion

PR #4950, opened 2022-05-16
delay automatic service worker registration until load event

### LINKED ISSUE #3342: Wait until the load event to register a service worker
### Describe the problem

The current best practice, [as per Google Developers](https://developers.google.com/web/fundamentals/primers/service-workers/registration), seems to be to register a service worker when the `load` event fires on the `window`:

```js
if ('serviceWorker' in navigator) {
  window.addEventListener('load', function() {
    navigator.serviceWorker.register('/service-worker.js');
  });
}
```

Currently, if a service worker is present in the project, SvelteKit generates the following registration code, which seems to be suboptimal:

```js
if ('serviceWorker' in navigator) {
  navigator.serviceWorker.register('/service-worker.js');
}
```

I can testify to this. My application was exported statically and served with `sirv` from localhost. If I delay the registration of the service worker, I get +5 points of performance in Lighthouse, which come from reducing Time to Interactive almost in half and clearing away the 40ms of Total Blocking Time.

In this article, there's a section called **Reasons to register early** that warns that this supposedly isn't a one-size-fits-all solution, however, I couldn't make sense of the scenario that they provide. Thus, I'd appreciate if someone took a closer look to that section.

I propose to add the delaying code to the service worker registration to follow the current best practices.

### Alternatives considered

We could leave it as it is until further rigorous testing is conducted, however, I trust that this best practice has been verified to be true.

### Importance

nice to have

### Additional Information

_No response_

Rich-Harris:
closes #3342. This is a more sensible default, and since we can disable automatic service worker registration we don't need to provide any more granular control.

### Tests

### Changesets

### DIFF
--- a/.changeset/swift-horses-march.md
+++ b/.changeset/swift-horses-march.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Delay automatic service worker registration until load event
--- a/.changeset/swift-horses-march.md
+++ b/.changeset/swift-horses-march.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Delay automatic service worker registration until load event
+[breaking] delay automatic service worker registration until load event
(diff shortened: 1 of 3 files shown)


.changeset/swift-horses-march.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Delay automatic service worker registration until load event
```
benmccann, line 5:
```suggestion
[breaking] delay automatic service worker registration until load event
```

benmccann (approved):

PR #4956, opened 2022-05-16
clarify rest routes

### LINKED ISSUE #4946: Clarify that rest routing parameter can be non-existent
### Describe the problem

So, I had read the advanced [advanced routing](https://kit.svelte.dev/docs/routing#advanced-routing) section of the docs and knew about [...rest] routing. But I believe that if I had the following:

`src/routes/[...rest]/page1.svelte`

Then the following routes would route to page1.svelte:

`localhost:3000/a/b/c/page1.svelte`
`localhost:3000/a/b/page1.svelte`

Which is correct. What I didn't realize is that `/page1.svelte` would also route to page1.svelte. Because with regular dynamic parameters you must have something there to be the parameter. For rest routes, the documents say:

>src/routes/a/[...rest]/z.svelte will match /a/z as well as /a/b/z and /a/b/c/z and so on. Make sure you check that the value of the rest parameter is valid.

And that does explain it correctly but it kind of flew over my head at /a/z match meant there was no ...rest parameter.

Specify in the Rest Parameters description that it's different than regular dynamic routes in that it can have multiple dynamic parameters OR no parameters. Just clarify that there doesn't need to be a parameter.

### Alternatives considered

Keep it the way it is.

### Importance

nice to have

### Additional Information

_No response_

Rich-Harris:
closes #4946

### DIFF
--- a/documentation/docs/02-routing.md
+++ b/documentation/docs/02-routing.md
@@ -302,7 +302,9 @@ Files and directories with a leading `_` or `.` (other than [`.well-known`](http
 
 #### Rest parameters
 
-A route can have multiple dynamic parameters, for example `src/routes/[category]/[item].svelte` or even `src/routes/[category]-[item].svelte`. (Parameters are 'non-greedy'; in an ambiguous case like `/x-y-z`, `category` would be `x` and `item` would be `y-z`.) If the number of route segments is unknown, you can use rest syntax — for example you might implement GitHub's file viewer like so...
+A route can have multiple dynamic parameters, for example `src/routes/[category]/[item].svelte` or even `src/routes/[category]-[item].svelte`. (Parameters are 'non-greedy'; in an ambiguous case like `/x-y-z`, `category` would be `x` and `item` would be `y-z`.)
+
+If the number of route segments is unknown, you can use rest syntax — for example you might implement GitHub's file viewer like so...
 
 ```bash
 /[org]/[repo]/tree/[branch]/[...file]
@@ -320,7 +322,7 @@ A route can have multiple dynamic parameters, for example `src/routes/[category]
 }
 ```
 
-> `src/routes/a/[...rest]/z.svelte` will match `/a/z` as well as `/a/b/z` and `/a/b/c/z` and so on. Make sure you check that the value of the rest parameter is valid.
+> `src/routes/a/[...rest]/z.svelte` will match `/a/z` (i.e. there's no parameter at all) as well as `/a/b/z` and `/a/b/c/z` and so on. Make sure you check that the value of the rest parameter is valid, for example using a [matcher](#advanced-routing-matching).
 
 #### Matching
 


documentation/docs/02-routing.md
```diff
@@ -302,7 +302,9 @@ Files and directories with a leading `_` or `.` (other than [`.well-known`](http
 
 #### Rest parameters
 
-A route can have multiple dynamic parameters, for example `src/routes/[category]/[item].svelte` or even `src/routes/[category]-[item].svelte`. (Parameters are 'non-greedy'; in an ambiguous case like `/x-y-z`, `category` would be `x` and `item` would be `y-z`.) If the number of route segments is unknown, you can use rest syntax — for example you might implement GitHub's file viewer like so...
+A route can have multiple dynamic parameters, for example `src/routes/[category]/[item].svelte` or even `src/routes/[category]-[item].svelte`. (Parameters are 'non-greedy'; in an ambiguous case like `/x-y-z`, `category` would be `x` and `item` would be `y-z`.)
+
+But if the number of route segments is unknown, you can use rest syntax — for example you might implement GitHub's file viewer like so...
```
benmccann, line 307: it sounds a bit strange to start a paragraph with "But" to me as it seems like a new paragraph should be a new thought. I wonder if we even need the preceding paragraph at all?
```suggestion
If the number of route segments is unknown, you can use rest syntax. For example, you might implement GitHub's file viewer like so...
```

documentation/docs/02-routing.md
```diff
@@ -320,7 +322,7 @@ A route can have multiple dynamic parameters, for example `src/routes/[category]
 }
 ```
 
-> `src/routes/a/[...rest]/z.svelte` will match `/a/z` as well as `/a/b/z` and `/a/b/c/z` and so on. Make sure you check that the value of the rest parameter is valid.
+> `src/routes/a/[...rest]/z.svelte` will match `/a/z` (i.e. there's no parameter at all) as well as `/a/b/z` and `/a/b/c/z` and so on. Make sure you check that the value of the rest parameter is valid, for example using a [matcher](#advanced-routing-matching).
```
benmccann, line 325:
```suggestion
> `src/routes/a/[...rest]/z.svelte` will match `/a/z` (i.e. there's no parameter at all) as well as `/a/b/z` and `/a/b/c/z` and so on. Make sure you check that the value of the rest parameter is valid — for example, using a [matcher](#advanced-routing-matching).
```

documentation/docs/02-routing.md
Rich-Harris: turns out this is actually duplicative with text above. fixed: c84e11ff2877f2bcd4d43fb1ebdf24047ed358fa

benmccann (approved):
lgtm. I just had one nitpick about a comma. not sure if you saw that suggestion

documentation/docs/02-routing.md
Rich-Harris: I think we can get away with the less punctuation-heavy version, since it's a short clause and this is an aside

PR #4958, opened 2022-05-16
warn if load uses `window.fetch`

### LINKED ISSUE #3731: Make it clearer when client-side load is re-fetching data because the server-side load didn't use its wrapped fetch function
### Describe the problem

When using an external SDK in a `load` function without telling it what `fetch` to use (or if it doesn't have the ability to be told what `fetch` to use), the network response is not serialized into the server-rendered DOM, and on the client-side, the browser's native `fetch` doesn't know to first look for these serialized responses before making a network call. This results in degraded performance, because the same request is then performed again in the browser.

The solution here is for people to call the SDK on the server in an endpoint (preferably a shadow endpoint for this page, which will make this very easy), or to make sure they only call it client-side in `onMount`, say. However, there's no currently obvious way for people to know that this undesirable thing is even happening.

In dev mode, we could override the browser's `window.fetch` with our own version that we can configure to complain when it is called (and then call the underlying native `fetch`). We could enable this during the `load` for the initial hydration of the page.

For bonus points, the `load`'s `fetch` could call this patched `window.fetch` function in such a way as to tell it that it _shouldn't_ complain during this request. This would reduce false positives where there is (for whatever reason) a genuine reason to request additional data in the client's `load` that wasn't requested in the server's `load`, as long as it is done via the appropriate `fetch` function.

### Alternatives considered

We could just document this, but that doesn't seem very discoverable. If someone's doing something that's probably wrong, we would like to tell them about it, rather than assuming that they've read the documentation and are keeping this in the back of their mind all the time.

### Importance

nice to have

### Additional Information

This proposal arose from a conversation in the maintainers' chat about #3729.

Rich-Harris:
closes #3731

### Tests

### Changesets


packages/kit/src/runtime/client/client.js
```diff
@@ -48,6 +48,19 @@ function update_scroll_positions(index) {
 	scroll_positions[index] = scroll_state();
 }
 
+const fetch = window.fetch;
```
Conduitry, line 51: It won't matter a ton if these aren't treeshaken fully in prod mode, but would it be better to just declare them as undefined `let`s here (which definitely would get treeshaken) and then assign them inside the `if (import.meta.env.DEV)` below?

Rich-Harris: the later calls to `fetch`...

https://github.com/sveltejs/kit/blob/7556a10afa3333791eaeeeb1d4321d339b382cd6/packages/kit/src/runtime/client/client.js#L547

https://github.com/sveltejs/kit/blob/7556a10afa3333791eaeeeb1d4321d339b382cd6/packages/kit/src/runtime/client/client.js#L645-L647

...would attempt to call an undefined function if the `let` wasn't initialised. This felt like the least worst option but i'm certainly open to alternatives

Conduitry: `let original_fetch;` maybe? I dunno. I think that would work without further changes to the rest of the existing code?

I think it's worth stumbling over ourselves a little bit in service of having dev mode code completely treeshaken out in prod mode.

Rich-Harris: if we do `original_fetch`, then those later uses of `fetch` need to use `original_fetch` instead, which means we need that to exist in prod too, no?

Conduitry: I haven't looked closely at the code, but I has assumed that we'd want to be using the global `fetch`/`window.fetch` everywhere, so that any calls would be run through our version that warned users about calling it. But it sounds like that's not the case? Do we want all later calls to _not_ trigger the warning? If so, yeah, I agree that what you have now is the least-bad version I can think of (and apologies for the noise!).

Rich-Harris: Yeah, we want the cases in the snippets embedded above to call the original `fetch` and not the intercepted version:

```js
  // this calls our fetch, which
  //   1. tracks the dependency
  //   2. loads the data from the HTML, if it was serialized and we're hydrating
  //   3. defers to original fetch (should not warn, because the user did everything right)
  const res = await fetch('https://api.example.com/whatever.json');
  // ...
}
```

```js
  // this calls the intercepted window.fetch, which should print a warning
  const res = await fetch('https://api.example.com/whatever.json');
  // ...
}
```

Conduitry (approved):

mrkishi:
This will probably result in a few false positives, since any unrelated `fetch` made during load will _also_ print the warning. Maybe the message should account for that?

But since this is a dev warning, what if we checked for `load_node` in `new Error().stack` instead? Don't all major browsers support `async` stack traces these days? We would miss `return Promise.resolve().then(() => fetch()).then(() => { props: ... })`, but I think this would be a lot less common than the false positives we'd get with `loading += 1`.

I wouldn't be confident messing with non-standard stringly-typed stack traces on production, but during dev that could be harmless.

Rich-Harris:
oh, interesting. i figured false positives would be rare enough that it wouldn't really matter, but using the stack trace is a great idea

mrkishi:
It could've been cool, but not if Safari has a say in it, of course.

Chrome:
```
Error
    at load (http://localhost:1313/src/routes/foo.svelte:110:12)
    at async load_node (http://localhost:1313/@fs/F:/projects/svelte/kit/packages/kit/src/runtime/client/client.js?t=1652762312775:562:19)
    at async load_route (http://localhost:1313/@fs/F:/projects/svelte/kit/packages/kit/src/runtime/client/client.js?t=1652762312775:674:14)
    at async update (http://localhost:1313/@fs/F:/projects/svelte/kit/packages/kit/src/runtime/client/client.js?t=1652762312775:213:38)
    at async navigate (http://localhost:1313/@fs/F:/projects/svelte/kit/packages/kit/src/runtime/client/client.js?t=1652762312775:899:21)
```

Firefox:
```
load@http://localhost:1313/src/routes/foo.svelte:110:12
async*load_node@http://localhost:1313/@fs/F:/projects/svelte/kit/packages/kit/src/runtime/client/client.js?t=1652762312775:562:37
load_route@http://localhost:1313/@fs/F:/projects/svelte/kit/packages/kit/src/runtime/client/client.js?t=1652762312775:674:20
async*update@http://localhost:1313/@fs/F:/projects/svelte/kit/packages/kit/src/runtime/client/client.js?t=1652762312775:213:44
navigate@http://localhost:1313/@fs/F:/projects/svelte/kit/packages/kit/src/runtime/client/client.js?t=1652762312775:899:27
create_client/_start_router/<@http://localhost:1313/@fs/F:/projects/svelte/kit/packages/kit/src/runtime/client/client.js?t=1652762312775:1115:13
EventListener.handleEvent*_start_router@http://localhost:1313/@fs/F:/projects/svelte/kit/packages/kit/src/runtime/client/client.js?t=1652762312775:1060:20
start@http://localhost:1313/@fs/projects/svelte/kit/packages/kit/src/runtime/client/start.js:42:10
async*@http://localhost:1313/bar:23:8
```

Safari's superior clean aesthetic, no noise in these machines:
```
@http://localhost:3000/src/routes/foo.svelte:110:21
asyncFunctionResume@[native code]
@[native code]
promiseReactionJobWithoutPromise@[native code]
promiseReactionJob@[native code]
```

Rich-Harris:
Maybe not getting the warning in Safari is a small price to pay to eliminate false positives in other browsers?

Rich-Harris:
Or I guess we could UA-sniff Safari and only have the false positives there

packages/kit/src/runtime/client/client.js
```diff
@@ -559,7 +586,14 @@ export function create_client({ target, session, base, trailing_slash }) {
...
 
-			const loaded = await module.load.call(null, load_input);
+			let loaded;
+
+			try {
+				if (import.meta.env.DEV) loading += 1;
+				loaded = await module.load.call(null, load_input);
+			} finally {
+				if (import.meta.env.DEV) loading -= 1;
+			}
```
Conduitry, line 596: Will Rollup/Terser/whoever optimize out this try/finally block in production? Should we un-DRY this a bit and have two separate dev and prod mode branches? Am I going overboard in trying to reduce bytes? Probably?

Rich-Harris: pushed a change. reducing bytes is always worthwhile

PR #4967, opened 2022-05-17
revert to using cjs for lambda

Rich-Harris:
fixes #4933. I don't truly understand why this was failing, but reverting to `cjs` rather than `esm` seems to fix it. If it turns out that other apps _need_ `esm` then we can expose it as a config option

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #4969, opened 2022-05-17
Use `@vercel/nft` to create lambdas

### LINKED ISSUE #4933: @sveltejs/adapter-vercel 1.0.0-next.51 breaks production project
### Describe the bug

The "start" method goes from...
```js
start({
	target: document.querySelector('[data-hydrate="16zlsws"]').parentNode,
	paths: {"base":"","assets":""},
	session: {},
	route: true,
	spa: false,
	trailing_slash: "never",
	hydrate: {
		status: 200,
		error: null,
		nodes: [
			import("/_app/pages/__layout.svelte-f2210a94.js"),
				import("/_app/pages/index.svelte-61687a47.js")
		],
		params: {},
		routeId: ""
	}
});
```
...to...
```js
start({
	target: document.querySelector('[data-hydrate="2eq20f"]').parentNode,
	paths: {"base":"","assets":""},
	session: {token:void 0},
	route: true,
	spa: false,
	trailing_slash: "never",
	hydrate: {
		status: 500,
		error: {name:"Error",message:"Dynamic require of \"tty\" is not supported",stack:"Error: Dynamic require of \"tty\" is not supported"},
		nodes: [
			import("/_app/undefined"),
				import("/_app/pages/__error.svelte-c1882640.js")
		],
		params: {},
		routeId: ""
	}
});
```

when updating the vercel adapter from `next.50` to `next.51`.

This change triggers the following error message when trying to open up the project in production.
```
Uncaught (in promise) TypeError: Failed to fetch dynamically imported module: https://my-website.vercel.app/_app/undefined
```

### Reproduction

Here's a MVP that will break in production with `adapter.51`, but will work with `adapter.50`: https://github.com/MathiasWP/vercel-adapter-bug-reproduction

The project has 2 branches, `main` with adapter `next.51`, and a branch called `adapter.50` with the `next.50` version.

Here is a deployed version with the `next.50` adapter (working): https://vercel-adapter-bug-reproduction-mps8glry7-mathiaswp.vercel.app/
And here is the exact same application with the `next.51` adapter (not working): https://vercel-adapter-bug-reproduction.vercel.app/

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 11.4
    CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
    Memory: 23.68 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.0 - /usr/local/bin/node
    npm: 8.1.0 - /usr/local/bin/npm
  Browsers:
    Brave Browser: 101.1.38.111
    Chrome: 101.0.4951.64
    Edge: 101.0.1210.47
    Firefox: 100.0
    Safari: 14.1.1
  npmPackages:
    @sveltejs/adapter-vercel: ^1.0.0-next.51 => 1.0.0-next.51 
    @sveltejs/kit: 1.0.0-next.330 => 1.0.0-next.330 
    svelte: ^3.46.0 => 3.48.0
```

### Severity

blocking an upgrade

### Additional Information

_No response_

Rich-Harris:
An alternative to #4967 — fixes #4933.

This is something I've been meaning to look into for a while. By using [`@vercel/nft`](https://github.com/vercel/nft), we can get a list of all the files a function depends on, and copy them into the function directory rather than relying on a bundler like `esbuild` to bundle everything into a single file.

Doing so has a couple of advantages:

* we don't encounter issues like https://github.com/evanw/esbuild/issues/1921
* apps should be able to use native (i.e. non-JavaScript) dependencies like `sharp`

I think continuing to use `esbuild` for Edge Functions is the right strategy, at least for the time being (since these can't use native dependencies anyway).

There's some trickery involved in getting this to work — in order for relative imports to continue working, we need to find the common ancestor directory of all dependencies (which could even be outside the workspace, if using linked dependencies) and copy the structure into the lambda, and to support `pnpm` we need special handling for symlinks. It's working, though it feels like quite a lot of grunt work, so I'm not 100% sure I'm doing it right.

Todo:

* [ ] use the same process for the v1 API

### Tests

### Changesets


Rich-Harris:
Getting it to work with the v1 API is proving finicky, but since v3 is coming out of beta soon I think we should just revert to CJS output for v1 and count down the days until we can remove it altogether.

People who want to use top-level await will need to opt into v3 by setting the `ENABLE_VC_BUILD=1` environment variable until it leaves beta.

Future optimisation: would be good to deduplicate shared dependencies between functions. Not going to worry about that right now though.

.changeset/few-scissors-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/adapter-vercel': patch
+---
+
+Revert to cjs mode when building for lambda
```
Conduitry, line 5: This should be updated.

Rich-Harris: good catch, updated this changeset and added another one

Rich-Harris:
Builds are completing successfully with this PR, but the build process on Vercel is then hanging with messages like these:

> Attempting to remove file /vercel/path0/node_modules/.cache/turbo/c5e740db3d8bebe3/sites/kit.svelte.dev/.vercel/output/functions/render.func/node_modules/.pnpm/shiki@0.9.11/node_modules/onigasm; a subdirectory is required
>
> Attempting to remove file /vercel/path0/node_modules/.cache/turbo/c5e740db3d8bebe3/sites/kit.svelte.dev/.vercel/output/functions/render.func/node_modules/.pnpm/lru-cache@5.1.1/node_modules/yallist; a subdirectory is required

Rich-Harris:
This should be fixed in the next version of `turbo`

PR #4972, opened 2022-05-17
add test to prevent accidental breakage of data-hydrate mechanism

### LINKED ISSUE #4685: kit start mounting wrong node

Rich-Harris:
supersedes #4698, closes #4685. This is a compromise measure — it keeps the existing behaviour, but adds a test to make sure we don't accidentally change the hydration mechanism in a way that breaks apps manipulating the markup in certain ways.

### Tests

### Changesets

### DIFF
--- a/.changeset/eight-rivers-battle.md
+++ b/.changeset/eight-rivers-battle.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] change data-hydrate to data-sveltekit-hydrate
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -158,7 +158,7 @@ export async function render_response({
 	const init_app = `
 		import { start } from ${s(options.prefix + options.manifest._.entry.file)};
 		start({
-			target: document.querySelector('[data-hydrate="${target}"]').parentNode,
+			target: document.querySelector('[data-sveltekit-hydrate="${target}"]').parentNode,
 			paths: ${s(options.paths)},
 			session: ${try_serialize($session, (error) => {
 				throw new Error(`Failed to serialize session data: ${error.message}`);
@@ -243,7 +243,7 @@ export async function render_response({
 				.map((dep) => `\n\t<link rel="modulepreload" href="${options.prefix + dep}">`)
 				.join('');
 
-			const attributes = ['type="module"', `data-hydrate="${target}"`];
+			const attributes = ['type="module"', `data-sveltekit-hydrate="${target}"`];
 
 			csp.add_script(init_app);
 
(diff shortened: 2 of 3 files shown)


mrkishi:
Should we make it `data-sveltekit-hydrate` (or `sveltekit:hydrate` 🐐 if our non-standard adventures make it to 1.0) as well? It'd be good regardless, but especially so if we're officially making it user-facing.

Rich-Harris:
cc @john-michael-murphy — we're planning to change `data-hydrate` to `data-sveltekit-hydrate`

mrkishi (approved):

john-michael-murphy:
@Rich-Harris sorry for delay! Been in OOO.  Thanks for the heads up.

PR #4979, opened 2022-05-18
Add dir=ltr to generated html-tags by default

benbender:
The web is expanding it's native handling towards logical properties for layouts to not be bound to left-to-right-languages anymore.

To follow this trend in svelte, it would be nice to have `dir=ltr` on all `html`-tags created by the framework per default. `dir=ltr` is the logical default, given we already have `lang=en` set. With this little addition, css/js can correctly target the writing-mode right from the project setup. Additionally, tools like `postcss-logical` are depending on this attribute to be set.

Would make my life a little bit easier as I don't have to do it manually on project-setup. Thanks for this awesome framework!

### Tests

### Changesets

### DIFF
--- a/.changeset/blue-trees-worry.md
+++ b/.changeset/blue-trees-worry.md
@@ -1,0 +1,7 @@
+---
+'@sveltejs/adapter-static': patch
+'default-template': patch
+'create-svelte': patch
+---
+
+Add dir=ltr as a default to all generated html-tags in templates
--- a/packages/adapter-static/test/apps/prerendered/src/app.html
+++ b/packages/adapter-static/test/apps/prerendered/src/app.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" dir="ltr">
 	<head>
 		<meta charset="utf-8" />
 		<meta name="viewport" content="width=device-width, initial-scale=1" />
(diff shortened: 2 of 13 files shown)


Rich-Harris:
Thank you. My feeling is that we probably _shouldn't_ do this, because it reinforces that `ltr` is the 'default'.

Fully aware that that's a hypocritical stance when we also have `lang="en"`! The difference is that that's there because HTML validators complain if no `lang` is provided, and we haven't yet implemented any form of built-in internationalisation. As soon as we do, the template will remove the `lang="en"` in favour of something less narcissistic.

benbender:
It's a bit funny, because I would argue from the direct opposite viewpoint ;) Imho the problem is the *implicit* default where the explicit markup pinpoints towards the possibility to change the writing-mode. Imho the thing that's hardly missing is education/awareness about possibilities and consequences. On top, as noted, the missing attribute may have a negative impact on tooling/styling which already supports/handles different writing modes...

But either way, I expected this one be a no-brainer and obviously I was wrong. If this change is something the maintainers would consider anyways, the right move may be to have this as an issue instead of a pr?

Rich-Harris:
An issue (or [discussion](https://github.com/sveltejs/kit/discussions) is always a good first move, since more people are likely to see it and weigh in, though I understand why you began with a PR for a small change like this.

I'll close this, on the basis that we should err on the side of not including stuff in `app.html` that isn't necessary. If people need the attribute for styling, it's very easy to add, and in the vast majority of cases I don't think it's needed. But that's certainly not set in stone, if people in an issue or discussion think it's necessary. Thanks!

PR #4983, opened 2022-05-18
breaking: Remove `hash` from `url` passed to `load`

### LINKED ISSUE #4788: `hash` is present in the `URL` passed to `load`
### Describe the bug

In working on #4773, @benmccann stated a couple of behaviors:

(For the duration of this post, `fragment === hash`.)

1. `prefetch` should not rerun for identical routes with different hashes (eg. `/prefetch` and `/prefetch#some-hash`).
2. People should _not_ be using the `URL` `hash` property in `load`.

I would tend to agree that either both or neither of these statements should be true; if 1 then 2 OR if not 1 then not 2. The promise of `prefetch` is that `load` will be run -- if the `hash` is meant to be part of the `URL` supplied to `load`, then `load` should rerun when the hash changes. If the `hash` should _not_ be part of the `URL` supplied to `load`, `load` should not rerun when the `hash` changes.

If 1 is correct, [this branch](https://github.com/tcc-sejohnson/kit/tree/sejohnson-remove-hash-from-load-url) will take care of 2. (It was very easy to change and test.)

However, I'm having trouble with the justification of 1 and 2. Looking at the [URL spec](https://url.spec.whatwg.org/#concept-url-fragment), I can't find anywhere that says "the content of the document should not change" based on the fragment. [The HTML spec](https://html.spec.whatwg.org/multipage/browsing-the-web.html#scroll-to-fragid) also doesn't seem to require that the content of the page remain static regardless of the fragment. (Please note, I'm not an expert in the nitty gritty of the URL or HTML specs, so please correct me if I'm wrong and I'll edit this part to include a correction.) The URL spec defines the fragment as "[a] [string](https://infra.spec.whatwg.org/#ascii-string) that can be used for further processing on the resource the [URL](https://url.spec.whatwg.org/#concept-url)’s other components identify." The phrase "can be used for further processing on the resource" in the context of a web document would say to me "...can be used for further processing by JavaScript...", which would suggest we would want to enable users to have access to it in `load`.

The very existence of things like `ReactRouter`'s [`HashRouter`](https://v5.reactrouter.com/web/api/HashRouter) seem to go against the idea that the document's content should not change based on the hash.

I'm opening this issue as a place to discuss further. If we commit to both 1 and 2, this will be fixed by opening a pull request against the previously-linked branch. If we commit to not 1 and not 2, this will be fixed by leaving 2 alone (exposing the hash to `load` is the current behavior), and a small modification to #4477 to fix 1. I'm most concerned with SvelteKit continuing to align with expected spec behavior (and the expectations of its users)!

### Reproduction

N/A.

### Logs

```shell
N/A.
```

### System Info

```shell
N/A.
```

### Severity

annoyance

### Additional Information

_No response_

tcc-sejohnson:
Fixes #4788. More specifically, this sets the `hash` of the `url` passed to `load` to the empty string, unifying the client- and server-side behavior. See the linked issue for discussion around why this is important.

### Tests

### Changesets


tcc-sejohnson:
> Nice, thank you! Thinking aloud: I wonder if we should return a `Proxy` from `get url` that errors if you try to access `url.hash` (in bother server and client contexts, I suppose) so that people aren't confused about why it's empty?

Wow, I don't know how I've gone this long without seeing a JavaScript `Proxy` object. That's pretty awesome. Give me five...

.changeset/shaggy-bugs-move.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+accessing url.hash from load now results in a helpful error
```
Rich-Harris, line 5:
```suggestion
[breaking] accessing url.hash from load now results in a helpful error
```

Rich-Harris (approved):

PR #4995, opened 2022-05-19
[chore] upgrade to Vite 2.9.9

benmccann:
use a newer version for the site and testing

### DIFF
--- a/sites/kit.svelte.dev/package.json
+++ b/sites/kit.svelte.dev/package.json
@@ -20,7 +20,6 @@
 		"shiki-twoslash": "^3.0.2",
 		"svelte": "^3.43.0",
 		"typescript": "~4.6.2",
-		"vite": "^2.9.0",
 		"vite-imagetools": "^4.0.3"
 	},
 	"type": "module"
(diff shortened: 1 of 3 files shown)


sites/kit.svelte.dev/package.json
```diff
@@ -20,7 +20,6 @@
 		"shiki-twoslash": "^3.0.2",
 		"svelte": "^3.43.0",
 		"typescript": "~4.6.2",
-		"vite": "^2.9.0",
```
Rich-Harris, line 23: this needs to be here otherwise the docs won't build (because shiki-twoslash needs to be able to see the types)

benmccann: ok. added it back

Rich-Harris (approved):

Rich-Harris:
we should probably start a vite 3 branch as well, make sure there are no surprises

benmccann:
Yeah, I was going to start a Vite 3 branch after the next alpha. There are some changes I want to leverage and wanted to test them out to make sure they'll work for us

PR #4999, opened 2022-05-20
ignore packaging test fixtures

Rich-Harris:
no idea why this didn't show up sooner, but `git clean -xdf && pnpm i && pnpm build` creates a bunch of untracked files that should be ignored


Rich-Harris:
actually that's not true, it was only happening on #4710 because i borked up the rimraf implementation. carry on

PR #5005, opened 2022-05-20
Upgrade to Vite 3

benmccann:
Announcement blog post: https://vitejs.dev/blog/announcing-vite3.html

Some changes noticeable for SvelteKit:
- relative base support (https://github.com/sveltejs/kit/issues/595)
- default port changed to 5173 in development mode
- `fixStacktrace` now defaults to `false`
- We can now set `appType: 'custom'` so that HTML middlewares are not added and we no longer have to remove many of them

v3 also added a new hook `configurePreviewServer` that has been backported to 2.9 and already leveraged in SvelteKit `master` which allowed us to switch to the `vite` CLI


richarddavenport:
@benmccann Does this allow debugging using IDE's now? For reference: https://github.com/vitejs/vite/pull/3928

benmccann:
I doubt it. That PR was closed without being merged

asuraphel:
@benmccann looks like @Rich-Harris is waiting for reply on [his question](https://github.com/sveltejs/kit/pull/5005#issuecomment-1134887926)

benmccann:
I've updated the description to be clearer

Rich-Harris:
The tests that are failing are the new `create-svelte` tests, which `npm install` the `next` version of SvelteKit instead of `workspace:*` (because they're instantiating a new project in order to `check` it). Because of that, there's a dependency mismatch that should resolve itself as soon as this PR is merged.

Not sure if we should come up with some tricksy workaround in the meantime, but I have a hunch that it won't be the last time we experience this issue in some form

Rich-Harris:
> i think this might work

[RON HOWARD VOICE] it did not work

PR #5012, opened 2022-05-21
Root devdeps

Rich-Harris:
Moves shared `devDependencies` into the workspace root

### Tests

### Changesets


Rich-Harris:
an alternative is to put _all_ devDeps in the workspace root (except e.g. the ones shiki-twoslash needs to have in the site dir for building the docs)

benmccann (approved):

package.json
```diff
@@ -30,6 +30,7 @@
 		"@rollup/plugin-replace": "^4.0.0",
 		"@sveltejs/eslint-config": "github:sveltejs/eslint-config#v5.8.0",
 		"@svitejs/changesets-changelog-github-compact": "^0.1.1",
+		"@types/node": "~16.11.36",
```
benmccann, line 33: Why tilde instead of caret?

Conduitry: I'm not sure how `@types/node` versions correspond to Node versions. If we're saying we support Node 16+, we should make sure we only have Node 16.0 types available, and not those from any later versions of Node 16. I don't know how we enforce this. Would that be `~16.0.0`?

If that's the purpose of the tilde, and if that's how the `@types/node` versions work, we should make sure we don't bring in types for versions of Node we don't support (so that the typechecking can yell at us immediately if we use unsupported features), so that's why no caret, but also ~16.11 isn't what we want to be depending on either.

mrkishi: Since we support Node v16, we don't want types for v17+ APIs readily available. They shouldn't show up in IDE features and should be marked as an error by default.

If we want to use some new feature, we can guard it with runtime checks and explicitly `@ts-ignore` it.

mrkishi: Ah, I didn't see your comment, @Conduitry.

> ~16.11 isn't what we want to be depending on either

Why not?

mrkishi: Errr, wait. I completely confused things in my head. I intended to use `~` for Typescript, not Node...

Conduitry: If `@types/node@16.11.x` is types for Node 16.11.x (is it?), then we don't want to use those types, as they would include features added after Node 16.0, and we're claiming SvelteKit works with any version of Node 16+ (which would include Node 16.0). If that's how versioning on `@types/node` indeed works, it would be nice to be warned by the typechecker that we are using features of Node that we shouldn't.

mrkishi: Since they're manually maintained, there's a [best-effort process](https://github.com/definitelytyped/definitelytyped#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library) to match the major and minor versions with Node's. I'm not sure they backport fixes to previous minor versions when a gap is found late (eg. 16.1 backported to 16.0), though.

We could make it `~16.0`, but maybe we need some CI tests to make sure it doesn't get changed accidentally. Apparently [Svelte's `=8`](https://github.com/sveltejs/svelte/commit/e90feb11f946948d79fa970f67d07a992838e9ea) didn't last 3 months, eh, [Conduitry](https://github.com/sveltejs/svelte/commit/71be8a6683d6c76c26230ac9d7e3ff27a1f2094a)? 😆 

I joke, but all package managers encourage us to update even if pinned to exact versions when updating post-`package.json`, so a CI check isn't that ridiculous.

PR #5016, opened 2022-05-21
replace `%svelte.body%` etc with `%sveltekit.body%`

Rich-Harris:
Never got round to creating an issue for this, but I think it's a no-brainer — we should have been more explicit from the start that `%namespace.body%` is a SvelteKit thing, not a Svelte thing

### Tests

### Changesets

### DIFF
--- a/.changeset/olive-lies-train.md
+++ b/.changeset/olive-lies-train.md
@@ -3,4 +3,4 @@
 '@sveltejs/kit': patch
 ---
 
-[breaking] Replace %svelte.body% etc with %sveltekit.body%
+[breaking] Replace `%svelte.body%` with `%sveltekit.body%`, etc.
--- a/.changeset/olive-lies-train.md
+++ b/.changeset/olive-lies-train.md
@@ -1,0 +1,6 @@
+---
+'create-svelte': patch
+'@sveltejs/kit': patch
+---
+
+[breaking] Replace %svelte.body% etc with %sveltekit.body%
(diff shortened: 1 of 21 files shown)


.changeset/olive-lies-train.md
```diff
@@ -0,0 +1,6 @@
+---
+'create-svelte': patch
+'@sveltejs/kit': patch
+---
+
+[breaking] Replace %svelte.body% etc with %sveltekit.body%
```
benmccann, line 6: The etc makes more sense to me at the end
```suggestion
[breaking] Replace `%svelte.body%` with `%sveltekit.body%`, etc.
```

benmccann (approved):
I saw this the other day and was also thinking it's inconsistent with what we've done elsewhere

Mlocik97:
I find it unnecesary long... can't we just have `%kit.body%` ? shorter to write if someone will write their own custom template or modify it.

tcc-sejohnson:
@Rich-Harris

While we're at it, the root context for `$app/stores` is called `__svelte__` ([see here](https://github.com/sveltejs/kit/blob/master/packages/kit/src/runtime/app/stores.js)). Seems like renaming it to `__sveltekit__` would fit well with this PR?

Rich-Harris:
We can make the `__svelte__` change separately since it's internal rather than external — it would probably be better to use a private object as the key rather than a magic string, but I don't know off the top of my head if that'd be a rabbit hole

PR #5018, opened 2022-05-21
Remove FLoC handling

furudean:
[FLoC was canned by WICG in favor of the Topics API](https://github.com/WICG/floc#replaced-by-the-topics-api). I couldn't find any information online if it was removed from chrome yet (the only browser where it was ever implemented) but it definitely looks like its on its way out.

### Tests

### Changesets


mrkishi:
FLoC never made it past origin trials and it's indeed not active in any browsers anymore.

Having said that, origin trials for its successor have already begun and [the FLoC opt-out header is still accepted in the current Topics API draft](https://github.com/patcg-individual-drafts/topics#:~:text=Note%3A%20The%20old%20Permissions%2DPolicy%3A%20interest%2Dcohort%3D()%20from%20FLoC%20will%20also%20forbid%20topic%20calculation.).

We should probably add a Topics API opt-out option when removing FLoC's, but since it's still in flux and the deprecated option is still valid, I wouldn't oppose postponing any action until the spec moves forward.

furudean:
I see, maybe we should change this to `Permissions-Policy: browsing-topics=()` instead to stay up to date? Or we can wait and see if topics also get scrapped... lol.

Though at this point we are definitely asking a different question--are topics as bad as FLoC, and should Svelte Kit do something about it by default?

mrkishi:
> are topics as bad as FLoC

Getting your hands wet might not be as bad as drowning in water, but I wouldn't want it 24/7 either. I expect most on the team to share the uneasiness towards the new iteration as well.

If you _want_ to update the implementation and docs to get it in sync with Topics API instead of FLoC, I believe you'll find no objections. But I don't speak for everyone so you might want to wait for other opinions.

mrkishi:
> We should probably add a Topics API opt-out option when removing FLoC's

I believe this would still be valuable even if privacy experts ultimately decide the Topics API isn't that harmful—it's just that if it turns out it _is_, then it should be enabled by default.

Rich-Harris:
If we don't want to disable it by default then the opt-out would just be this in `hooks.js`, no?

```js
  response.headers.set('Permissions-Policy', 'browsing-topics=()');
  return response;
}
```

I'd argue we only want config for it if it's something we want to enable by default, which probably _does_ depend on how harmful it is.

.changeset/friendly-eggs-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+remove floc
```
Rich-Harris, line 5:
```suggestion
remove FLoC protection, now that we vanquished Google
```

Rich-Harris:
thank you!

PR #5030, opened 2022-05-23
fix: use named imports from vite to support upcoming vite 3.0 esm

dominikg:
vite 3.0 is going to be esm first :tada: https://github.com/vitejs/vite/pull/8178

No default export means we have to change a few `import vite from 'vite'` to prevent errors like this one:
https://github.com/vitejs/vite-ecosystem-ci/runs/6549407184?check_suite_focus=true#step:7:392

### Tests

### Changesets

### DIFF
--- a/.changeset/short-guests-sort.md
+++ b/.changeset/short-guests-sort.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+use named imports of vite to support upcoming vite 3.0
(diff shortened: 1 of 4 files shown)


bluwy (approved):

Rich-Harris:
Nice, this is great news. Is there any reason to prefer named imports over a namespace import? I tend to think `vite.build(...)` is clearer than `build(...)` — should we just change `import vite` to `import * as vite` instead of changing all the invocations?

dominikg:
I feel it's a bit more explicit to use the named exports, but the readability is indeed reduced. :thinking:   Indifferent about it if there are no other technical benefits in avoiding * imports

Rich-Harris:
Yeah, there's no technical benefits, it's really just a stylistic preference

PR #5042, opened 2022-05-23
use `crypto.randomUUID()`

### LINKED ISSUE #4929: Drop external uuid dependency in favor of native functionality
### Describe the problem

The default demo [currently uses](https://github.com/sveltejs/kit/blob/d723e876be53db66f0549def417d94a3c0014653/packages/create-svelte/templates/default/src/hooks.ts#L1) a dev dependency for generating random v4 UUIDs inside the hooks file which has been available natively in Node and the Web Crypto API for some time now. In Node `crypto.randomUUID()` it [has been available](https://nodejs.org/docs/latest-v14.x/api/crypto.html#crypto_crypto_randomuuid_options) since v14.17.0. The Web Cryptography API, which appears to support web workers as well, uses the [same method name](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID).

Dropping this dependency now that [support for Node 14 has ended](https://github.com/sveltejs/kit/issues/4916) seems like a good time to cut the baggage.

### Alternatives considered

Leave the dependency in place. It's [slower than native methods](https://dev.to/galkin/crypto-randomuuid-vs-uuid-v4-47i5) but probably not noticeable in the demo.

### Importance

nice to have

### Additional Information

_No response_

Rich-Harris:
Now that we only support Node 16 and above, we can drop the dependency on `@lukeed/uuid` in favour of Using The Platform.

Closes #4929

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/templates/default/src/hooks.ts
+++ b/packages/create-svelte/templates/default/src/hooks.ts
@@ -1,11 +1,10 @@
-import { v4 as uuid } from '@lukeed/uuid';
 import type { Handle } from '@sveltejs/kit';
 import * as cookie from 'cookie';
 
 /** @type {import('@sveltejs/kit').Handle} */
 export const handle: Handle = async ({ event, resolve }) => {
 	const cookies = cookie.parse(event.request.headers.get('cookie') || '');
-	event.locals.userid = cookies['userid'] || uuid();
+	event.locals.userid = cookies['userid'] || crypto.randomUUID();
 
 	const response = await resolve(event);
 
--- a/.changeset/fast-islands-lick.md
+++ b/.changeset/fast-islands-lick.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Use crypto.randomUUID() instead of @lukeed/uuid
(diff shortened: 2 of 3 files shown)


benmccann (approved):

packages/create-svelte/templates/default/src/hooks.ts
```diff
@@ -1,11 +1,10 @@
-import { v4 as uuid } from '@lukeed/uuid';
 import type { Handle } from '@sveltejs/kit';
 import * as cookie from 'cookie';
 
 /** @type {import('@sveltejs/kit').Handle} */
 export const handle: Handle = async ({ event, resolve }) => {
 	const cookies = cookie.parse(event.request.headers.get('cookie') || '');
-	event.locals.userid = cookies['userid'] || uuid();
+	event.locals.userid = cookies['userid'] || crypto.randomUUID();
```
Conduitry, line 7: Are we doing something elsewhere that lets us rely on `crypto` as a global? I just tried that in another unrelated project, and it didn't work.

Conduitry: Ah, I guess this is the new `polyfills.js`. Never mind!

PR #5043, opened 2022-05-23
[chore] return config from server start methods

benmccann:
We shouldn't read the Kit config in the CLI, but rather in the Vite plugin so that it will work with the `vite` CLI as well. This doesn't move the config reading all the way to the plugin yet, but does remove it from the CLI which is a step in that direction

### DIFF
--- a/packages/kit/src/cli.js
+++ b/packages/kit/src/cli.js
@@ -61,25 +61,23 @@ prog
 		/** @type {() => Promise<void>} */
 		let close;
 
-		/** @param {import('types').ValidatedConfig} config */
-		async function start(config) {
+		async function start() {
 			const { dev } = await import('./core/dev/index.js');
 
-			const { address_info, server_config, close } = await dev({
+			const { address_info, config, close } = await dev({
 				port,
 				host,
-				https,
-				config
+				https
 			});
 
 			welcome({
 				port: address_info.port,
 				host: address_info.address,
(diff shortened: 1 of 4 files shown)


packages/kit/src/cli.js
```diff
@@ -61,25 +61,23 @@ prog
...
 				port: address_info.port,
 				host: address_info.address,
-				https: !!(https || server_config.https),
-				open: first && (open || !!server_config.open),
-				base: config.kit.paths.base,
-				loose: server_config.fs.strict === false,
-				allow: server_config.fs.allow
+				https: !!(https || config.server.https),
+				open: first && (open || !!config.server.open),
+				base: config.base,
```
Rich-Harris, line 78: this will break when we switch to always using `./` with Vite 3

benmccann: Why? Because of the leading `.` or something else?

Rich-Harris: Because it will open to `localhost:${port}` instead of `localhost:${port}/${config.kit.base}`

benmccann: Fixed.

I'm not that familiar with the base path stuff. I was assuming `config.kit.paths.base` would just be passed in to Vite as `config.base`, but it looks we're hardcoding the Vite `config.base` to `'/'`. I switched it to return the SvelteKit config from the server start methods so we can get the correct base path

PR #5045, opened 2022-05-24
add types to pkg.exports

### LINKED ISSUE #5028: Typescript cannot find types when using `NodeNext` or `Node16` modules
### Describe the bug

Typescript 4.7+ includes new the new `"Node16"`/`"NodeNext"` options for configuration options `"module"` and `"moduleResolution"`. These adhere strictly to Node's `package.json` `"exports"` fields rules -- when that field is set the other root fields (like `"types"`) are ignored.

The `package.json` for `@sveltejs/adapter-static`, and presumably other SvelteKit packages, does not list types in its `"exports"` field, causing Typescript to be unable to find types when using `NodeNext` module strategies.

It's an easy fix, and Typescript 4.7 is still in beta so the impact of this will slowly grow as adoption of latest Typescript versions increases.

Here's the diff generated via `patch-package` for `@sveltejs/adapter-static`:

```diff
diff --git a/node_modules/@sveltejs/adapter-static/package.json b/node_modules/@sveltejs/adapter-static/package.json
index c3f39a0..6c0dac8 100644
--- a/node_modules/@sveltejs/adapter-static/package.json
+++ b/node_modules/@sveltejs/adapter-static/package.json
@@ -12,6 +12,7 @@
   "main": "index.js",
   "exports": {
     ".": {
+      "types": "./index.d.ts",
       "import": "./index.js"
     },
     "./package.json": "./package.json"
```

### Reproduction

With Typescript 4.7-beta and "module" set to "ESNext" or "NodeNext", we get an error finding the types:

https://www.typescriptlang.org/play?moduleResolution=99&target=99&jsx=0&allowSyntheticDefaultImports=true&ts=4.7.0-beta#code/JYWwDg9gTgLgBAQwCYLDAplAyjBNgDGcAZlBCHAOQACAzgG7oA2GAVrQPTKoZQC0tXPgKUA3EA

If we change the "module" to "CommonJs", the types are found successfully (ignore the newly-revealed error, it's unrelated to this issue):

https://www.typescriptlang.org/play?target=99&ts=4.7.0-beta&jsx=0&module=1&allowSyntheticDefaultImports=true#code/JYWwDg9gTgLgBAQwCYLDAplAyjBNgDGcAZlBCHAOQACAzgG7oA2GAVrQPTKoZQC0tXPgKUA3EA

### Logs

_No response_

### System Info

```shell
System:
  OS: Windows 10 10.0.22000
  CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700KF
  Memory: 43.34 GB / 63.82 GB
Binaries:
  Node: 16.14.2 - ~\AppData\Local\Volta\tools\image\node\16.14.2\node.EXE
  npm: 8.10.0 - ~\AppData\Local\Volta\tools\image\npm\8.10.0\bin\npm.CMD
Browsers:
  Edge: Spartan (44.22000.), Chromium (101.0.121)
npmPackages:
  @sveltejs/adapter-static: ^1.0.0-next.31 => 1.0.0-next.31 
  @sveltejs/kit: ^1.0.0-next.338 => 1.0.0-next.338
```

### Severity

### Additional Information

_No response_

Rich-Harris:
fixes #5028 (I think?). No types for `@sveltejs/kit/*` but I don't think we need them because of `types/ambient.d.ts`

### Tests

### Changesets

### DIFF
--- a/packages/adapter-auto/package.json
+++ b/packages/adapter-auto/package.json
@@ -11,8 +11,8 @@
 	"type": "module",
 	"exports": {
 		".": {
-			"import": "./index.js",
-			"types": "./index.d.ts"
+			"types": "./index.d.ts",
+			"import": "./index.js"
 		},
 		"./package.json": "./package.json"
 	},
--- a/packages/adapter-auto/package.json
+++ b/packages/adapter-auto/package.json
@@ -11,7 +11,8 @@
 	"type": "module",
 	"exports": {
 		".": {
-			"import": "./index.js"
+			"import": "./index.js",
+			"types": "./index.d.ts"
 		},
 		"./package.json": "./package.json"
 	},
--- a/.changeset/stupid-maps-hide.md
+++ b/.changeset/stupid-maps-hide.md
@@ -1,0 +1,12 @@
+---
+'@sveltejs/adapter-auto': patch
+'@sveltejs/adapter-cloudflare': patch
+'@sveltejs/adapter-cloudflare-workers': patch
+'@sveltejs/adapter-netlify': patch
+'@sveltejs/adapter-node': patch
+'@sveltejs/adapter-static': patch
+'@sveltejs/adapter-vercel': patch
+'@sveltejs/amp': patch
+---
+
+Add types to pkg.exports
--- a/packages/adapter-cloudflare-workers/package.json
+++ b/packages/adapter-cloudflare-workers/package.json
@@ -11,8 +11,8 @@
 	"type": "module",
 	"exports": {
 		".": {
-			"import": "./index.js",
-			"types": "./index.d.ts"
+			"types": "./index.d.ts",
+			"import": "./index.js"
 		},
 		"./package.json": "./package.json"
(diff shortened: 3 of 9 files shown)


mrkishi (approved):
I won't pretend to know why it's needed, but it seems to work here both in VSCode and `tsc`.

packages/adapter-auto/package.json
```diff
@@ -11,7 +11,8 @@
 	"type": "module",
 	"exports": {
 		".": {
-			"import": "./index.js"
+			"import": "./index.js",
+			"types": "./index.d.ts"
```
bluwy, line 15: Think we should flip `import` and `types` so `types` comes first so it's more compliant to the resolve algorithm. The Svelte repo also did the same.

dummdidumm: Yes, types should be first

Rich-Harris: fixed

PR #5047, opened 2022-05-24
ensure correct case is used for static assets

### LINKED ISSUE #4931: Asset extension matching not consistent between DEV and PREVIEW
### Describe the bug

Preview build of sveltekit doesn't match asset paths the same way as dev build.

### Reproduction

Put let's say image.PNG inside of static then in index.svelte `` and let's run npm run dev, then run build and preview and the image won't be displayed anymore.

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19041
    CPU: (6) x64 Intel(R) Core(TM) i5-8600K CPU @ 3.60GHz
    Memory: 6.21 GB / 15.94 GB
  Binaries:
    Node: 16.15.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.8.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1023.0), Chromium (97.0.1072.76)
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.328
    svelte: ^3.44.0 => 3.48.0
```

### Severity

annoyance

### Additional Information

_No response_

Rich-Harris:
closes #4931

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -195,9 +195,7 @@ export async function create_plugin(config) {
 							const file = config.kit.files.assets + pathname;
 
 							if (fs.existsSync(file) && !fs.statSync(file).isDirectory()) {
-								const has_correct_case = fs
-									.readdirSync(path.dirname(file))
-									.includes(path.basename(file));
+								const has_correct_case = fs.realpathSync.native(file) === path.resolve(file);
 
 								if (has_correct_case) {
 									req.url = encodeURI(pathname); // don't need query/hash
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -195,9 +195,15 @@ export async function create_plugin(config) {
 							const file = config.kit.files.assets + pathname;
 
 							if (fs.existsSync(file) && !fs.statSync(file).isDirectory()) {
-								req.url = encodeURI(pathname); // don't need query/hash
-								asset_server(req, res);
-								return;
+								const has_correct_case = fs
+									.readdirSync(path.dirname(file))
+									.includes(path.basename(file));
+
+								if (has_correct_case) {
+									req.url = encodeURI(pathname); // don't need query/hash
+									asset_server(req, res);
+									return;
+								}
 							}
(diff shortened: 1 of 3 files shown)


packages/kit/src/core/dev/plugin.js
```diff
@@ -195,9 +195,15 @@ export async function create_plugin(config) {
 							const file = config.kit.files.assets + pathname;
 
 							if (fs.existsSync(file) && !fs.statSync(file).isDirectory()) {
-								req.url = encodeURI(pathname); // don't need query/hash
-								asset_server(req, res);
-								return;
+								const has_correct_case = fs
+									.readdirSync(path.dirname(file))
+									.includes(path.basename(file));
```
mrkishi, line 200: How about using `realpath.native` instead?

```suggestion
								const has_correct_case = fs.realpathSync.native(file) === path.resolve(file);
```

packages/kit/test/apps/basics/test/test.js
```diff
@@ -2570,6 +2570,11 @@ test.describe.parallel('Static files', () => {
 		const response = await request.get(path);
 		expect(response.status()).toBe(200);
 	});
+
+	test('Filenames are case-sensitive', async ({ request }) => {
+		let response = await request.get('/static.JSON');
```
mrkishi, line 2575:
```suggestion
		const response = await request.get('/static.JSON');
```

mrkishi (approved):

packages/kit/src/core/dev/plugin.js
anarchodin: I started getting unexpected 404s after updating SvelteKit today. As it turns out, if you have a symlinked subdirectory in `static`, realpath and resolve _don't_ give you the same result.

Monkey-patching the check to use the initial version stopped the 404s, since the directory _does_ indeed contain a file by that name. You might want to double-check this fix.

(As a detail, this was happening on Windows, but it seems like it should also bite POSIX systems.)

(Also, considering this is merged and resolved, I should probably ensure @mrkishi and @Rich-Harris get pinged about this. 😃)

PR #5049, opened 2022-05-24
use free variable for native fetch

Rich-Harris:
i have ABSOLUTELY NO IDEA how this passed CI in the first place, but apparently #5041 broke a bunch of stuff somehow? my browser suddenly started complaining that you can't call `fetch` if it's a property of something that isn't `window`. seems like a weird constraint but who am i to judge.

no changeset because #5041 wasn't released yet, so no need

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -616,7 +616,7 @@ export function create_client({ target, session, base, trailing_slash }) {
 					increment();
 					loaded = await module.load.call(null, load_input);
 				} finally {
-					decrement();
+					unlock_fetch();
 				}
 			} else {
 				loaded = await module.load.call(null, load_input);
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -11,7 +11,7 @@ import {
 	notifiable_store,
 	scroll_state
 } from './utils.js';
-import { increment, decrement, initial_fetch, native_fetch } from './fetcher.js';
+import { increment as lock_fetch, decrement as unlock_fetch, initial_fetch, native_fetch } from './fetcher.js';
 import { parse } from './parse.js';
 
 import Root from '__GENERATED__/root.svelte';
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -11,7 +11,7 @@ import {
 	notifiable_store,
 	scroll_state
 } from './utils.js';
-import { increment as lock_fetch, decrement as unlock_fetch, initial_fetch, native_fetch } from './fetcher.js';
+import { lock_fetch, unlock_fetch, initial_fetch, native_fetch } from './fetcher.js';
 import { parse } from './parse.js';
 
 import Root from '__GENERATED__/root.svelte';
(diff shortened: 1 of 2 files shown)


mrkishi:
Huh. Are the failures on Safari/Firefox? We were fooled by green checks pretty hard.

packages/kit/src/runtime/client/client.js
```diff
@@ -11,7 +11,7 @@ import {
 	notifiable_store,
 	scroll_state
 } from './utils.js';
-import * as fetcher from './fetcher.js';
+import { increment, decrement, initial_fetch, native_fetch } from './fetcher.js';
```
mrkishi, line 14: Maybe renaming `increment` and `decrement` to something else would make it clearer, at least locally if not in `fetcher.js`?

```suggestion
import { increment as lock_fetch, decrement as unlock_fetch, initial_fetch, native_fetch } from './fetcher.js';
```

Rich-Harris:
No, just running `pnpm test` locally. Truly weird

Conduitry:
Somewhat recently, there was a change to have Playwright use whatever version of Chrome it could find on your system, rather than using separate downloaded binaries. I wonder whether that could be related to the difference in behavior in the tests?

I'd be in favor of reverting that change anyway. Requiring the user have Chrome installed on their system makes local dev less portable, and not using a specific version makes tests less reproducible.

packages/kit/src/runtime/client/client.js
```diff
@@ -613,10 +613,10 @@ export function create_client({ target, session, base, trailing_slash }) {
 
 			if (import.meta.env.DEV) {
 				try {
-					fetcher.increment();
+					increment();
```
mrkishi, line 616: Following previous comment:
```suggestion
					lock_fetch();
```

packages/kit/src/runtime/client/client.js
```diff
@@ -613,10 +613,10 @@ export function create_client({ target, session, base, trailing_slash }) {
...
 					loaded = await module.load.call(null, load_input);
 				} finally {
-					fetcher.decrement();
+					decrement();
```
mrkishi, line 619: Continuing:
```suggestion
					unlock_fetch();
```

mrkishi (approved):

sartoshi-foot-dao:
My latest local builds failing for perhaps the same reason:

```
> Using @sveltejs/adapter-vercel
✘ [ERROR] Could not resolve "@sveltejs/kit/install-fetch"

    .svelte-kit/vercel-tmp/serverless.js:1:29:
      1 │ import { installFetch } from '@sveltejs/kit/install-fetch';
        ╵                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  The path "./install-fetch" is not exported by package "@sveltejs/kit":

    node_modules/@sveltejs/kit/package.json:47:13:
      47 │   "exports": {
         ╵              ^

  You can mark the path "@sveltejs/kit/install-fetch" as external to exclude it from the bundle,
  which will remove this error.

> Build failed with 1 error:
.svelte-kit/vercel-tmp/serverless.js:1:29: ERROR: Could not resolve "@sveltejs/kit/install-fetch"
```

Hope this helps.

PR #5057, opened 2022-05-24
disable local playwright tracing by default

mrkishi:
This PR disables playwright tracing on local e2e tests unless `process.env.TRACE` is enabled. Tracing has a performance impact, so it makes sense to only trace tests when we're interested in them.

An alternative would be to enable retries locally and use `trace: 'on-first-retry'` instead. This would be valuable if the majority of contributors use traces as a first debugging step, otherwise requiring manual activation (ie. this pr) might be best.

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/utils.js
+++ b/packages/kit/test/utils.js
@@ -162,7 +162,7 @@ export const config = {
 	],
 	use: {
 		screenshot: 'only-on-failure',
-		trace: 'retain-on-failure',
+		trace: process.env.CI || process.env.TRACE ? 'retain-on-failure' : 'off',
 		// use stable chrome from host OS instead of downloading one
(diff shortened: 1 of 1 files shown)


dominikg:
If trace on retry is used, the result archiving needs to be updated to include retried results. They are currently filtered to keep the archive smaller
https://github.com/sveltejs/kit/blob/4592b76d87c25e5cbaaec1d7f1ad08dfd322fcf3/.github/workflows/ci.yml#L48

mrkishi:
This was only about local tests. Do you believe we should move CI from `retain-on-failure` to `on-first-retry`? I'm not sure it's a good deal, having tracing always enabled on CI might even surface timing-sensitive errors we miss locally due to the different execution.

PR #5060, opened 2022-05-24
update turbo

Rich-Harris:
Need to update `turbo` for #4969 to pass CI, but it's causing `pnpm check` to fail so I figured I'd work on it in a separate branch.

I have no idea why it's failing suddenly; it's very strange. I assumed it was because it caused `typescript` to upgrade for some reason, but bumping TypeScript separately doesn't seem to have the same effect.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: pnpm
       - run: pnpm install --frozen-lockfile
-      - run: pnpm turbo run test
+      - run: pnpm test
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Ok, it's happening because the new `turbo` checks test apps as well as `packages/*` — previously it followed `pnpm` by ignoring everything inside `test`

Rich-Harris:
Turns out `turbo run test` no longer works, because it now runs all of `packages/kit/test/*/*` simultaneously, and there's a battle for port 3000. I think we need to remove those directories from the pnpm workspace, and run the tests manually

PR #5063, opened 2022-05-24
fix: Generated types no longer permanently delete themselves

### LINKED ISSUE #5059: Generated types for all files are deleted and never repopulated when sync is run
### Describe the bug

I believe the fix for #4973 introduced this bug because it uses `rimraf` and then `write_if_changed` to repopulate the cleared `types` directory. This means that in some cases when `svelte-kit sync` runs, the entire `types` directory is removed but no files get re-added because they have not changed.

Stopping the dev process and re-running seems to correctly restore the types (probably because the `previous_contents` `Map` referenced by `write_if_changed` is now cleared).

### Reproduction

Have a file with generated types in `.svelte-kit/types`, run `svelte-kit sync`, see `.svelte-kit/types` disappear and not come back.

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.3.1
    CPU: (32) x64 Intel(R) Xeon(R) W-3245 CPU @ 3.20GHz
    Memory: 35.00 GB / 96.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 18.1.0 - ~/.asdf/installs/nodejs/18.1.0/bin/node
    npm: 8.8.0 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 101.0.4951.64
    Firefox: 90.0.2
    Safari: 15.4
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.43
    @sveltejs/kit: next => 1.0.0-next.338
    svelte: ^3.44.0 => 3.48.0
```

### Severity

annoyance

### Additional Information

_No response_

tcc-sejohnson:
Fixes #5059.

`svelte-kit sync` always completely deletes the `types` directory when it writes types. Then, it re-writes types by looking for changed files. If no files have been changed, the `types` directory is gone for good (or at least until a file is changed). See the repro in the linked issue.

This fixes the issue by always rewriting all of the types (since we're deleting everything, we need to rewrite everything). I'm not really sure how to test this change, though this does fix the repro.

If there's some larger issue here that needs a bigger fix, feel free to just close this and move on!

### Tests

### Changesets

### DIFF
--- a/.changeset/lemon-planets-sit.md
+++ b/.changeset/lemon-planets-sit.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] svelte-kit sync no longer permanently deletes the types directory
--- a/packages/kit/src/core/sync/utils.js
+++ b/packages/kit/src/core/sync/utils.js
@@ -17,6 +17,16 @@ export function write_if_changed(file, code) {
 	}
 }
 
+/**
+ * @param {string} file
+ * @param {string} code
+ */
+export function write(file, code) {
+	previous_contents.set(file, code);
+	mkdirp(path.dirname(file));
+	fs.writeFileSync(file, code);
+}
+
 /** @param {string} str */
 export function trim(str) {
 	const indentation = /** @type {RegExpExecArray} */ (/\n?(\s*)/.exec(str))[1];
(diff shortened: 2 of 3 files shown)


Rich-Harris (approved):

Rich-Harris:
thank you!

Issue #5066, opened 2022-05-24
Start from version `1.0.0-next.337` to current `1.0.0-next.343` , use `[...path].svelte` will overwrite assets import urls


TrungRueta:
### Describe the bug

start from version `1.0.0-next.337` to current `1.0.0-next.343`. when define a route file to catch error 404 (file named `[...path].svelte`) , it will catch asset urls imported in source too. This issue not happend until ver `1.0.0-next.336` . 

[Change log 337](https://github.com/sveltejs/kit/blob/master/packages/kit/CHANGELOG.md#100-next337)

i guess that the change in v*.337 regards to 
`Only run viteServeStaticMiddleware after server has attempted response (https://github.com/sveltejs/kit/pull/4974)` will be cause. But dont know how to help fix.

### Reproduction

- create new fresh svelte-kit project use `npm init svelte <folder project name>`
- add image into `/src/lib/assets/a.jpg`
- from file `/src/routes/index.svelte` import image to `<script>` block, then create `<img>` tag in template.
- in `/src/routes` create new file name `[...path].svelte`, define some html like a 404 catcher page.
- run `npm run dev`. see image url in chrome devtool will redirect to `[...path].svelte` page .
- rename file `[...path].svelte` to `[path].svelte`. run dev serer again, see img show correctly.
- try add file `[...path].svelte again, then run build `npm run build` and then `npm run preview`, the built version work ok so issue only happend with dev mode.

### System Info

```shell
System:
    OS: macOS 11.6
    CPU: (8) arm64 Apple M1
    Memory: 81.88 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.11.1 - ~/.nvm/versions/node/v16.11.1/bin/node
    Yarn: 1.22.17 - /opt/homebrew/bin/yarn
    npm: 8.3.0 - ~/.nvm/versions/node/v16.11.1/bin/npm
  Browsers:
    Chrome: 101.0.4951.64
    Safari: 15.3
  npmPackages:
    svelte: ^3.44.0 => 3.48.0
```

### Severity

blocking an upgrade

Rich-Harris:
closed via #5070

PR #4974, opened 2022-05-17
Only run `viteServeStaticMiddleware` after SvelteKit server has a chance to respond

Rich-Harris:
This closes #4353 by removing `viteServeStaticMiddleware` from the stack, and then (since we need it to serve things like `src/routes/foo/bar.svg`) running it after SvelteKit's server has a chance to respond to a request.

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [x] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #5070, opened 2022-05-25
Prefer (allowed) static assets to routes

Rich-Harris:
#5066 happens because of #4974, which we need because otherwise files in the project root will shadow stuff in `src/routes`. As of that PR, we attempt to render requests with `server.respond(...)`, and only defer to `viteServeStaticMiddleware` if that results in a 404.

In the case where you have rest routes, that's no good, because you can get a 200 response even when you're requesting a static asset.

I'm having a hard time figuring out what the correct fix for this is, so I'm just going to leave this failing test here for now.

### Tests

### Changesets


benmccann:
I think the correct fix is probably https://github.com/vitejs/vite/issues/7363

Rich-Harris:
In lieu of a fix in Vite itself, I think this is a reasonable solution — we mimic Vite's internal logic to determine if an extant file is allowed, and if so we defer to `serve_static_middleware`

packages/kit/src/vite/dev/index.js
```diff
@@ -189,7 +190,7 @@ export async function dev(vite, svelte_config) {
 
 		remove_html_middlewares(vite.middlewares);
 
-		vite.middlewares.use(async (req, res) => {
+		vite.middlewares.use(async (req, res, next) => {
```
Rich-Harris, line 193: didn't end up using this after all

```suggestion
		vite.middlewares.use(async (req, res) => {
```

7antra (approved):

packages/kit/src/vite/index.js
```diff
@@ -75,6 +75,9 @@ function kit() {
 	/** @type {import('vite').ConfigEnv} */
 	let vite_config_env;
 
+	/** @type {string[]} */
+	let allowed;
```
benmccann, line 79: do we need to store `allowed` separately when we already store `vite_config`? (although arguably we could store the `vite_config` in `configResolved`)

benmccann: right, but the only thing we use `vite_config` for is to get the `logLevel`. it's actually probably better for us to use the resolved config there, so if we switch `vite_config` to store the resolved config then we'd no longer need a separate `allowed` variable

Rich-Harris: there's also `protocol`. the reason i didn't just save the whole thing is that there isn't a convenient type for the resolved config but yeah, we can save those three things inside `configResolved`

Rich-Harris: oh! TIL. for some reason i didn't find that before — fixed

packages/kit/src/vite/dev/index.js
```diff
@@ -212,6 +213,23 @@ export async function dev(vite, svelte_config) {
...
+
+				// temporary — debugging windows via CI
+				if (file.includes('logo.svg')) {
+					console.error({ file, allowed });
+				}
+
+				if (
+					fs.existsSync(file) &&
+					!fs.statSync(file).isDirectory() &&
+					allowed.some((dir) => file.startsWith(dir))
```
benmccann, line 226: if we're going to do this we also need to check if `server.fs.strict` is enabled

Rich-Harris: done

packages/kit/test/apps/basics/test/test.js
```diff
@@ -2572,12 +2572,17 @@ test.describe('Static files', () => {
...
 		await page.goto('/assets');
 		const path = await page.textContent('h1');
 
-		const response = await request.get(path);
-		expect(response.status()).toBe(200);
+		const r1 = await request.get(path);
+		expect(r1.status()).toBe(200);
+		expect(await r1.text()).toContain('http://www.w3.org/2000/svg');
+
+		const r2 = await request.get('/package.json');
```
benmccann, line 2583: this seems surprising to me. I wouldn't expect you to be able to get the `package.json`, `node_modules`, etc. It will also make it available to everyone on your network. I'm not sure that's a huge risk, but I'm not really sure why we'd expose it either

Rich-Harris: It's _not_ getting package.json. The test is there to check that it will serve a route _called_ `/package.json` despite there being a package.json in the project directory outside the `allow` list. Check two lines further down.

node_modules on the other hand are on the allow list so that you can use node_modules!

benmccann, line 2583: it'd be nice to add a comment for package.json since it's easy to miss what it's testing
```suggestion
		// check that we can fetch a route which overlaps with the name of a file
		const r2 = await request.get('/package.json');
```

that's not exactly what I was suggesting with `node_modules`, but you can ignore because I didn't correctly understand what this PR was doing

7antra (approved):

PR #5073, opened 2022-05-25
[chore] provide Vite config via plugin

benmccann:
Kit is now simply a Vite plugin in dev mode

We could move it to a new location and export it as `@sveltejs/kit/vite`, but I didn't want to move files around in the same PR where I'm refactoring things to make it more manageable to understand what's changing. I'll do that after this PR is merged in a follow up


packages/kit/src/core/dev/plugin.js
```diff
@@ -1,373 +1,431 @@
...
+import { deep_merge } from '../../utils/object.js';
 
 // Vite doesn't expose this so we just copy the list for now
 // https://github.com/vitejs/vite/blob/3edd1af56e980aef56641a5a51cf2932bb580d41/packages/vite/src/node/plugins/css.ts#L96
 const style_pattern = /\.(css|less|sass|scss|styl|stylus|pcss|postcss)$/;
 
 const cwd = process.cwd();
 
+/** @type {import('types').ValidatedConfig} */
+export const svelte_config = await load_config();
```
Rich-Harris, line 26: this won't be reloaded when it changes. I don't think you can export `plugins`

benmccann: ah, so it's broken on `master` too then I guess?

benmccann: fixed

packages/kit/src/cli.js
```diff
@@ -62,35 +63,57 @@ prog
...
-			const { dev } = await import('./core/dev/index.js');
+			const svelte_config = await load_config();
+			const { svelte, sveltekit } = await import('./core/dev/plugin.js');
+			const vite_config = await svelte_config.kit.vite();
+
+			/** @type {import('vite').UserConfig} */
+			const config = {
+				plugins: [...(vite_config.plugins || []), svelte(svelte_config), sveltekit(svelte_config)]
+			};
+			config.server = config.server || {};
```
Rich-Harris, line 74:
```suggestion
			const config = {
				plugins: [...(vite_config.plugins || []), svelte(svelte_config), sveltekit(svelte_config)],
				server: {}
			};
```

benmccann: It's a more natural way to write it, but upsets TypeScript. TypeScript see the type annotation which says `server` can be optional and that takes precedence. If we want to do it the suggested way then I also need to add ` & {server: import('vite').ServerOptions}` to the type, which seemed even uglier to me. I just noticed the way I have it written currently can be simplified though

benmccann:
I'm going to merge this since it seems like we're pretty much agreed on the details at this point that way I can get started on the next PR. Feel free to leave comments still on either this PR or the next one and I can tweak any remaining items still

PR #5081, opened 2022-05-26
Expose `caches` to Cloudflare adapter through Platform

venashial:
This PR exposes [Cloudflare's Worker Cache API](https://developers.cloudflare.com/workers/runtime-apis/cache/) as `caches` in the `platform` object. This allows hooks to do elaborate caching, for example, [I use the Cache API](https://github.com/modrinth/knossos/blob/1b34ce1fb906f400fac72e753244890aba87009e/src/adapter/files/worker.js#L63-L85) to cache certain pages based on request cookies.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-cloudflare/README.md
+++ b/packages/adapter-cloudflare/README.md
@@ -53,7 +53,7 @@ When configuring your project settings, you must use the following settings:
 
 ## Environment variables
 
-The [`env`](https://developers.cloudflare.com/workers/runtime-apis/fetch-event#parameters) object, containing KV namespaces etc, is passed to SvelteKit via the `platform` property along with `context`, meaning you can access it in hooks and endpoints:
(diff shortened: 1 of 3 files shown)


packages/adapter-cloudflare/README.md
```diff
@@ -66,7 +66,8 @@ declare namespace App {
 +		};
 +		context: {
 +			waitUntil(promise: Promise<any>): void;
-+		}
++		};
++		caches: CacheStorage & { default: Cache }
```
benmccann, line 70: I wonder if we should show these in alphabetical order?

benmccann (approved):
lgtm, but I'm not the most familiar with Cloudflare, so will leave this for others to review as well

packages/adapter-cloudflare/README.md
venashial: I put `caches` at the bottom because it is the least likely to be used.

Rich-Harris (approved):

Rich-Harris:
thank you!

PR #5085, opened 2022-05-26
Check extensions for param matcher files

h8gi:
I get error messages (`"Matcher names must match /^[a-zA-Z_][a-zA-Z0-9_]*$/ — "${file}" is invalid"`) when my emacs creates backup files (like `#[id].js#`) in /params directory.
I think only `.js` and `.ts` files should be included in `matchers`.

### Tests

### Changesets


packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -267,6 +267,7 @@ export default function create_manifest_data({
 	if (fs.existsSync(config.kit.files.params)) {
 		for (const file of fs.readdirSync(config.kit.files.params)) {
 			const ext = path.extname(file);
+			if (!['.js', '.ts'].includes(ext)) break;
```
PH4NTOMiki, line 270: maybe it should be `continue` here? maybe i'm wrong

h8gi: Oh, I'm wrong!
It should be `continue`.

h8gi:
That sounds good!
For now, I rewrote `['.js', '.ts']` to `endpointExtensions`.

benmccann:
> For now, I rewrote ['.js', '.ts'] to endpointExtensions.

That's a bit confusing because you wouldn't expect `endpointExtensions` to affect matchers. I think you'd have to rename the option as well

benmccann:
@h8gi it looks like this PR needs a rebase

Rich-Harris (approved):

Rich-Harris (approved):

PR #5089, opened 2022-05-26
allow symlinked assets

### LINKED ISSUE #4709: Prerenderer don't follow symbolic link in assets folder
### Describe the bug

If prerendered page reference assets through a symbolic link, then it will not be accessible on `svelte-kit build`.

### Reproduction

https://github.com/locriacyber/debug-svelte-symbolic

Note:
- `static/data` is symbolic link to `data`
- The website works fine in `yarn dev` 
- But it breaks with `yarn build`

### Logs

```shell
> yarn build
yarn run v1.22.18
warning package.json: No license field
$ svelte-kit build
vite v2.9.5 building for production...
✓ 13 modules transformed.
.svelte-kit/output/client/_app/manifest.json                    1.14 KiB
.svelte-kit/output/client/_app/start-32c0b608.js                23.10 KiB / gzip: 8.67 KiB
.svelte-kit/output/client/_app/layout.svelte-23d0f029.js        0.53 KiB / gzip: 0.35 KiB
.svelte-kit/output/client/_app/error.svelte-bb9909e1.js         1.56 KiB / gzip: 0.74 KiB
.svelte-kit/output/client/_app/pages/index.svelte-8781c221.js   0.48 KiB / gzip: 0.33 KiB
.svelte-kit/output/client/_app/chunks/index-dd45c2ae.js         6.83 KiB / gzip: 2.79 KiB
vite v2.9.5 building SSR bundle for production...
✓ 11 modules transformed.
Generated an empty chunk: "hooks"
.svelte-kit/output/server/manifest.json                        1.09 KiB
.svelte-kit/output/server/index.js                             75.42 KiB
.svelte-kit/output/server/entries/fallbacks/layout.svelte.js   0.24 KiB
.svelte-kit/output/server/entries/fallbacks/error.svelte.js    0.72 KiB
.svelte-kit/output/server/entries/pages/index.svelte.js        0.28 KiB
.svelte-kit/output/server/chunks/index-2dc61825.js             2.29 KiB
.svelte-kit/output/server/chunks/hooks-1c45ba0b.js             0.00 KiB
> 404 /data/home.html (linked from /)
```

### Severity

annoyance, workaround complicated

### Workaround

Use the following file as `hooks.ts`

```ts
import type { Handle } from "@sveltejs/kit"
import * as fs from "fs/promises"
import * as path from "path"

// Workaround for https://github.com/sveltejs/kit/issues/4709
export const handle: Handle = async function({ event, resolve: res }) {
  if (event.url.pathname.startsWith("/data/")) {
    const relpath = event.url.pathname.slice("/data/".length)
    return new Response(await fs.readFile(path.join(await find_project_root(), "../data", relpath)))
  }

  const response = await res(event)
  return response
}

async function find_project_root(): Promise<string> {
    let patha = path.resolve(".")
    while (!(await fs.readdir(patha)).indexOf("package.json")) {
        const pathb = path.join(patha, "..")
        if (pathb == patha) throw Error("Cannot find project root")
        patha = pathb
    }
    return patha
}
```

gaarf:
This might fix #4709 

my monorepo setup uses symlinked static assets, and it worked great until https://github.com/sveltejs/kit/pull/5047 !
I believe that check for the casing was too restrictive, so I went ahead and made this PR addressing the issue. There's a test included in the `basics` app. The issue really only applies to dev, since the build copies symlink contents transparently.

Thanks in advance for the review 🙏 

### Tests

### Changesets


gaarf:
@anarchodin @mrkishi @Rich-Harris please have a look 🪟 🔗 ⬆️

Rich-Harris (approved):

Rich-Harris:
thank you! the correct fix looks a bit different — we can't just check the end of the path, because that won't work if the symlink source and target have different names. instead we're just checking that `file` exists in `fs.readdirSync(path.dirname(file))`, which is only true if the case is correct (and then doing that recursively)

PR #5094, opened 2022-05-27
Expose Vite plugin as @sveltejs/kit/vite

benmccann:


benmccann:
@nickbreaton @JessicaSachs FYI. I hope this might help you. I'd love some feedback if you're able to try it out at some point

JessicaSachs:
!!! NICE. I'm looking forward to trying this out post-conferences... So.... mid-June? Unless @bluewinds wants to give it a go first.

packages/kit/src/vite/plugin.js
```diff
@@ -0,0 +1,368 @@
...
+import { deep_merge } from '../utils/object.js';
+import { find_deps, get_default_config } from '../core/build/utils.js';
+
+const cwd = process.cwd();
+
+/**
+ * @param {import('types').ValidatedConfig} svelte_config
+ * @return {import('vite').Plugin}
+ */
+export const sveltekit = function (svelte_config) {
```
Rich-Harris, line 24: this needs to take a `Config['kit']`, not a `ValidatedConfig` (or rather, `Config` becomes what's currently `Config['kit']`, and mutatis mutandis for `ValidatedConfig`). we'll have to do the validation inside the plugin

```suggestion
/**
 * @param {import('types').Config['kit']} svelte_config
 * @return {import('vite').Plugin}
 */
export const sveltekit = function (svelte_config = {}) {
```

Rich-Harris:
notes to self:

* [ ] "The value for kit.vite.blah specified in svelte.config.js has been ignored" is incorrect, if the value was specified in vite.config.js
* [ ] Need to warn that `config.kit.vite` is ignored in the case where `vite.config.js` exists

Rich-Harris (approved):

PR #5099, opened 2022-05-27
auto-configure adapter-static based on environment

### LINKED ISSUE #4818: Make `adapter-static` zero-config by detecting platform
### Describe the problem

We've encountered Vercel users using `adapter-static` instead of `adapter-vercel` (or `adapter-auto`) to deploy their fully-prerendered sites, the result being that they have to mess around with config (changing `public` to `build`, and fiddling with `cleanUrls`). It's likely that some users of other zero-config platforms (currently Netlify and Cloudflare Pages) are in a similar situation.

Using `adapter-static` is basically unnecessary given that the same pages will be prerendered whatever adapter you use (the only difference being that `adapter-static` lets you specify a fallback page, which is unnecessary in these cases). It would be helpful therefore if we steered people back towards `adapter-auto`, making configuration unnecessary.

In `adapter-static`, we could use the same platform detection logic in `adapter-auto`...

https://github.com/sveltejs/kit/blob/b2c37078ae109f63c6646e1e1e5752fb54582aa5/packages/adapter-auto/adapters.js#L1-L17

...to detect that someone should just use `adapter-auto` or the adapter specific to their platform.

### Alternatives considered

Make `adapter-static` somehow work in a zero-config way on the platforms in question.

### Importance

nice to have

### Additional Information

Couple of caveats: functions will still be created for endpoints and for pages with dynamic parameters. Those functions will never be invoked (unless someone requests something that wasn't prerendered), but there may be costs associated just with creating them. We can mitigate this by declaring endpoints `prerenderOnly` or something, but not eliminate it.

On Netlify (when using Edge Functions) and Cloudflare Pages, the main handler will still be invoked (since it sits in front of requests for static assets), and there _is_ a cost to that.

As I write this, I'm starting to wonder what universally zero-config `adapter-static` would look like.

Rich-Harris:
fixes #4818. The simpler version of this would just be to set `pages = 'public', assets = 'public'` in the Vercel case, but that wouldn't result in the right behaviour when `trailingSlash` is `always`, so this PR uses the build output API instead.

Todo

* [x] get it to actually work properly on Vercel (including respecting `trailingSlash` when using the v1 API)
* [ ] defaults for other platforms
* [x] docs

### Tests

### Changesets


Rich-Harris:
Switched to using `routes` instead of `overrides` in the Vercel config, because `overrides` don't quite work how I'd hope:

* `/index.html` incorrectly serves the index page
* `/foo` and `/foo/` are interchangeable, and there's no way to enforce the use of one over the other (apart from brittle regex routing, which can't handle cases where a page pathname has a `.` character in its last segment

There is a downside, which is that there's a limit on the number of routes you can have (1,000, IIRC). Since we need two routes for each prerendered page (one for trailing slash redirects, one for the actual page) that means a site can only have 499 prerendered pages.

Rich-Harris:
marking this ready as it'll unblock some folks. would love to add support for more platforms in future, though ideally without adding TOML/YAML parsers to read configs

styfle:
> Since we need two routes for each prerendered page (one for trailing slash redirects, one for the actual page) 

@Rich-Harris That doesn't sound right. You only need routes for pages that don't exist. So creating the file for the page should be sufficient. And adding a trailing slash shouldn't been needed for every page, you can do it with these routes:

https://github.com/vercel/vercel/blob/b8b87b96da49fff38435d1e6ae123a721c8d7a4b/packages/routing-utils/src/superstatic.ts#L147

PR #5108, opened 2022-05-29
robustify error handling

Rich-Harris:
Currently, if you introduce an error into a component that causes rendering to fail...

```svelte
<p>{this_variable_does_not_exist}</p>
```

...fixing the component won't cause the page to recover, because Svelte itself is now in a broken state. (We plan to make Svelte more resilient to this sort of thing, but it's a ways off.) Adding `hot.optimistic: false` to the `vite-plugin-svelte` options causes the page to reload on HMR updates after a rendering error is encountered.

That solves half the problem. If the page reloads and it's _still_ broken, the server will render an error page. Since the error page generally won't use the component where the error was, subsequent HMR updates will be ignored, so fixing the component won't cause the error page to go away.

I think a reasonable way to fix that is to _always_ reload the page when an HMR update occurs, _if you're on an error page_.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -512,6 +512,9 @@ export const svelte = function (svelte_config) {
 			...svelte_config.compilerOptions,
 			hydratable: !!svelte_config.kit.browser.hydrate
 		},
+		hot: {
+			optimistic: false
+		},
 		configFile: false
 	});
 };
--- a/packages/kit/src/core/dev/plugin.js
+++ b/packages/kit/src/core/dev/plugin.js
@@ -512,9 +512,6 @@ export const svelte = function (svelte_config) {
 			...svelte_config.compilerOptions,
 			hydratable: !!svelte_config.kit.browser.hydrate
 		},
-		hot: {
-			optimistic: false
-		},
 		configFile: false
 	});
 };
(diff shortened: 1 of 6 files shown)


packages/kit/src/core/dev/plugin.js
```diff
@@ -512,6 +512,9 @@ export const svelte = function (svelte_config) {
 			...svelte_config.compilerOptions,
 			hydratable: !!svelte_config.kit.browser.hydrate
 		},
+		hot: {
+			optimistic: false
```
dominikg, line 516: rixo mentioned changing the default value in svelte-hmr so this isn't needed after, unless you want to prevent users overriding it

dominikg (approved):

packages/kit/src/core/dev/plugin.js
```diff
@@ -512,6 +512,9 @@ export const svelte = function (svelte_config) {
...
+		},
```
Rich-Harris, line 517: we can remove this once https://github.com/sveltejs/vite-plugin-svelte/pull/349 is released

```suggestion
		hot: {
			optimistic: false
		},
```

packages/kit/src/core/dev/plugin.js
Rich-Harris: yeah, that seems better. will apply https://github.com/sveltejs/kit/pull/5108/files#r884334210 later

Issue #5110, opened 2022-05-29
Adapters throw typescript errors on import when `moduleResolution` is set to "NodeNext" or "ESNext"


adam-coster:
### Describe the bug

The `index.d.ts` file of each adapter has its export like so:

https://github.com/sveltejs/kit/blob/ec7b804c18c7a7bb50573c16006bb699d68157d9/packages/adapter-auto/index.d.ts#L3-L4

But when Typescript (4.7+) is set to have `moduleResolution` of `NodeNext` or `ESNext` the import statement throws a compiler error. This can be fixed by changing the export to the following:

```ts
export default function plugin(): Adapter;
```

### Reproduction

Importing with `NodeNext` module resolution (should see the error):

https://www.typescriptlang.org/play?target=99&moduleResolution=99&ts=4.7.0-beta#code/JYWwDg9gTgLgBAQwCYLDAplAyjBNgDGcAZlBCHAOQACAzgG7oA2GAVrQPTKoZQC0CAK4wIlANxA

Same deal with `ESNext`:

https://www.typescriptlang.org/play?target=99&ts=4.7.0-beta#code/JYWwDg9gTgLgBAQwCYLDAplAyjBNgDGcAZlBCHAOQACAzgG7oA2GAVrQPTKoZQC0CAK4wIlANxA

### System Info

```shell
System:
    OS: Linux 5.10 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700KF
    Memory: 21.52 GB / 31.26 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 16.13.1 - ~/.volta/tools/image/node/16.13.1/bin/node
    npm: 8.9.0 - ~/.volta/tools/image/npm/8.9.0/bin/npm
  npmPackages:
    @sveltejs/adapter-auto: ^1.0.0-next.48 => 1.0.0-next.48 
    @sveltejs/kit: ^1.0.0-next.345 => 1.0.0-next.345 
    typescript: 4.8.0-dev.20220529
```

### Severity

serious, but I can work around it

### Additional Information

As far as I can guess, there aren't backwards compatibility problems if the export is changed to:

```ts
export default function plugin(): Adapter;
```

Since the packages are already set up as ESM.

adam-coster:
I cloned the SvelteKit repo to see if I could just make a PR real quick, but it turns out it's not that simple.

[...]

PR #5111, opened 2022-05-29
Update adapter entrypoint typings to be NodeNext/ESNext-compatible.

adam-coster:
As discussed in #5110, the entry `index.d.ts` files of the adapters all use a node-resolution-compatible export syntax like this:

```ts
 declare function plugin(): Adapter; 
 export = plugin; 
```

However, all of these packages use ESM (instead of CommonJS) so they can instead use:

```ts
export default function plugin(): Adapter;
```

Making this change resolves a Typescript 4.7+ error where `import adapter from '@sveltejs/adapter-auto';` throws a compiler error when the `tsconfig.json` is configured with `moduleResolution` set to `NodeNext` or `ESNext`.

To make this change, all of the `index.js` files that import these types have to have their JSDoc import statements changed to explicitly grab the `default` export. E.g. `/** @type {import('.')} **/` ➡️ `/** @type {import('.').default} **/`.

Since these packages are already modules these typings are more technically correct, though it's possible that this change will result in new Typescript errors for some compiler configurations. With Typescript 4.8 (dev), I can confirm that this change results in no errors with either of the following `tsconfig` configurations:

```jsonc
// Node resolution and ES2020 targets
"compilerOptions":{
    "target": "ES2020",
    "lib": ["ES2020"],
    "module": "ES2020",
    "moduleResolution": "Node",
}
```

```jsonc
// NodeNext resolution and ESNext targtes
"compilerOptions":{
    "target": "ESNext",
    "lib": ["ESNext"],
    "module": "ESNext",
    "moduleResolution": "NodeNext",
}
```

- [x ] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs

### Tests

> ⚠️ My test environment didn't have Chrome, so those tests did not pass.

### Changesets

### DIFF
--- a/.changeset/lemon-timers-deliver.md
+++ b/.changeset/lemon-timers-deliver.md
@@ -1,0 +1,11 @@
+---
+"@sveltejs/adapter-auto": patch
+"@sveltejs/adapter-cloudflare-workers": patch
+"@sveltejs/adapter-cloudflare": patch
+"@sveltejs/adapter-netlify": patch
+"@sveltejs/adapter-node": patch
+"@sveltejs/adapter-static": patch
+"@sveltejs/adapter-vercel": patch
+---
+
+Update adapter entrypoint typings to be NodeNext/ESNext-compatible.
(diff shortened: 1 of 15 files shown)


dummdidumm (approved):

Rich-Harris:
thank you!

Issue #5113, opened 2022-05-30
Swap out `node-fetch` for `undici`


rmunn:
### Describe the problem

https://github.com/sveltejs/kit/pull/4934 had a checklist step "swap out `node-fetch` for `undici`, which is more standards compliant (though I think we're blocked on https://github.com/nodejs/undici/issues/1262)". https://github.com/nodejs/undici/issues/1262 has just been closed via https://github.com/nodejs/undici/pull/1469, so it may be time to look again at swapping out `node-fetch`, at least once https://github.com/nodejs/undici/pull/1469 makes its way into an actual Node release.

### Describe the proposed solution

Swap out `node-fetch` for `undici` as described in https://github.com/sveltejs/kit/pull/4934#issue-1236378590

### Importance

would make my life easier

### Additional Information

This should allow us to make progress on https://github.com/sveltejs/kit/issues/3419 as well (see https://github.com/sveltejs/kit/pull/4934#issuecomment-1127679552).

OttoHatt:
I believe this'll be blocked by sveltejs/kit#5109 (cloudflare wrangler also uses Undici for requests).

rmunn:
> I believe this'll be blocked by #5109 (cloudflare wrangler also uses Undici for requests).

#5109 is marked as fixed by #5116, which has been [released in `next.346`](https://github.com/sveltejs/kit/blob/master/packages/kit/CHANGELOG.md#100-next346). I can't confirm that the fix works on Cloudflare as I don't have any Cloudflare sites set up at the moment; @OttoHatt, can you confirm that it works?

OttoHatt:
yes! on the newest version, #5109 is fixed.

PR #5117, opened 2022-05-30
Undici

Rich-Harris:
#5113. This replaces `node-fetch` with `undici`, which is more standards-compliant and is built in to Node 18. This will unlock using web streams for request and response bodies, on platforms that support them. 

TODO:

* [x] get remaining tests passing
* [x] replace `git` dependency with `undici@latest` when https://github.com/nodejs/undici/pull/1469 is released
* [x] only apply polyfills where necessary
* [x] document addition of web streams in https://kit.svelte.dev/docs/web-standards

### Tests

### Changesets


Rich-Harris:
The `handles large response` test is failing for reasons I have been unable to figure out

Rich-Harris:
oh and a bunch of packaging tests have started failing for absolutely no good reason whatsoever. jfc software is awful

Rich-Harris:
ok i think the unrelated packaging tests are failing because this PR installs `undici` from git, which means it's pulling in dev dependencies that mess up some stuff. programming is super fun

pixelmund:
Seems like undici had a new release, so we can remove the git dependency now.

half-metal:
Is this PR on the backburner for now, or waiting for something else from undici? Currently using another solution for streaming req/resp, but hoping to only use Sveltekit.

benmccann:
I fixed the merge conflict against `master`. The `Load › handles large responses` test is still failing though

benmccann:
I don't really understand why `handles large responses` is starting a second server. I guess it's to calculate `times_responded`? But could we simplify by having a single server with a `large-response.json` endpoint that increments a global `times_responded`?

Rich-Harris:
> I don't really understand why `handles large responses` is starting a second server

Because before undici it wasn't possible to generate a chunked response from a SvelteKit app. That's no longer the case, so I've updated the test — instead of generating the response in a second server, it does it in an endpoint

benmccann (approved):

PR #5021, opened 2022-05-21
fix: lint pipeline automatically formats files, misses errors in CI

tcc-sejohnson:
Every time I try to make a PR that changes `packages/kit` it tries to reformat the `paths-base/svelte.config.js` file. I figured this was some weird mis-cached formatting issue, so I ran `turbo lint --force` to bypass caching and see if there were any other mis-formatted things. Looks like there was one in `adapter-vercel`.

Note that running `turbo check --force` also fails, but I couldn't figure out what was causing it and I didn't have time to dig into it any further.

If this is some weird configuration issue on my end, feel free to just close the PR and forget about it.

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [ ] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #5120, opened 2022-05-30
Update dependencies

Rich-Harris:
The failing packaging tests in #5117 will hit us eventually even if they're only showing up there now because of a git dependency, so I figured it's as good a time as any to do some spring cleaning.

One thing I'm realising is that our new practice of installing shared dependencies in the workspace makes it much harder to determine which packages are affected by a given upgrade. I'm inclined to suggest reversing #5012. We could just have a script that checks that different `package.json` files don't reference different versions of the same dependency.


Rich-Harris:
Superseded by #5021. Will close this if that PR is accepted

Issue #5137, opened 2022-06-01
Lazily imported components suffer from FOUC


Rich-Harris:
### Describe the bug

A useful pattern is to dynamically import a component inside a `load` function:

```svelte
<script context="module">
  export async function load({ params }) {
    return {
      props: {
        Thing: params.thing = 'this'
          ? (await import('$lib/ThisThing.svelte')).default
          : (await import('$lib/ThatThing.svelte')).default
      }
    };
  }
</script>

<script>
	export let Thing;
</script>

<Thing/>
```

The problem is that results in FOUC, because SvelteKit only injects CSS for static dependencies when server-side rendering. The styles don't appear until the component is imported.

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-qstyga?file=src%2Froutes%2Findex.svelte,src%2Flib%2FLazy.svelte&terminal=dev

### System Info

```shell
latest
```

### Severity

serious, but I can work around it

PlopTheReal:
Not sure if that's related but as of 1.0.0-next.370, all the CSS of all my dynamic components gets loaded from the first page load.

PlopTheReal:
Sorry I haven't precised that I'm not loading the components with that pattern (lazy load when the component has mounted).
I've tried such pattern but I'm getting the following errors: 
```
in dev:
/web/svelttest/node_modules/vite/dist/node/chunks/dep-80fe9c6b.js:50310
            const err = new Error(`Cannot find module '${id}' imported from '${importer}'`);

in preview: 
Error: <Test> is not a valid SSR component. You may need to review your build config to ensure that dependencies are compiled, rather than imported as pre-compiled modules
```

Rich-Harris:
This was fixed by #5138.

@PlopTheReal it's unclear what your last comment has to do with this thread — if you're experiencing a bug, please open a new issue with a repro

PR #5138, opened 2022-06-01
Include styles for dynamically imported components

Rich-Harris:
#5137. This is a partial fix — it solves the SSR problem in prod builds, and solves it in dev when the page is reloaded (because `node.ssrTransformResult` has been computed by that point), but not for the initial page load. Not yet sure what the solution looks like. Maybe we can wait until `load` has completed before computing the styles. (The most obvious alternative is to eagerly load the dynamically imported modules, but that would be a bad idea in the case where they're dynamically imported precisely because they can't be loaded on the server.)

### Tests

### Changesets


dominikg:
maybe related: injecting a link for component styles is pending this: https://github.com/sveltejs/vite-plugin-svelte/issues/341

If we want to avoid the js, is there a way to compile only the css for a component?

PatrickG:
Can we merge this as is so that styles of dynamic imported components load at least in production builds without javascript?

Rich-Harris:
> It might be nice to see if we can use Vite's `importedCss` ([vitejs/vite#6629](https://github.com/vitejs/vite/pull/6629))

Not sure I understand what that is and what benefit it brings?

> If we want to avoid the js, is there a way to compile only the css for a component?

Not without changes to Svelte itself

> Can we merge this as is so that styles of dynamic imported components load at least in production builds without javascript?

We generally try and avoid that — it ends up being more work as you need to create a new issue and then (eventually) a new PR. Better to just fix it in one go, even if it means a slight delay

PR #5141, opened 2022-06-01
[chore] introduce `KitConfig` type

benmccann:
Creates a `KitConfig` and `ValidatedKitConfig` (last file in the PR) to start to decouple the Svelte and SvelteKit configuration. We will need to figure out what to do about `extensions` which stops us from being able to more thoroughly convert to `ValidatedKitConfig`

### DIFF
--- a/.changeset/moody-lobsters-cough.md
+++ b/.changeset/moody-lobsters-cough.md
@@ -1,5 +1,5 @@
 ---
-"@sveltejs/kit": patch
+'@sveltejs/kit': patch
 ---
 
-[chore] introduce `KitConfig` type
+[chore] introduce `UserConfig` type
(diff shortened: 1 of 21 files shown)


Rich-Harris:
Should we change `KitConfig` to `UserConfig`? It aligns with Vite, and this...

```js
/** @type {import('@sveltejs/kit').UserConfig} */
```

...feels a bit more natural than this, to me:

```js
/** @type {import('@sveltejs/kit').KitConfig} */
```

Presumably `ValidatedKitConfig` will be renamed to `ValidatedConfig` if we get rid of the current `ValidatedConfig`, but that's a task for future us

Rich-Harris:
I took the liberty of making that change; we can revert c0bcc3bf414fdcded188ebceb78b31a9b1d508f7 if you disagree with it

Rich-Harris:
fair enough. reverted

Rich-Harris (approved):

PR #5149, opened 2022-06-03
Fix page-endpoints with trailing slash always

### LINKED ISSUE #5148: Page Endpoint redirects with trailing slash always
### Describe the bug

When `trailingSlash` is set to `always`, the `/.../__data.json` request gets redirected to `/.../`.

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-xp9vnn

Click the `Click me!` link.

### Logs

_No response_

### System Info

```shell
Not relevant
```

### Severity

### Additional Information

To fix it, we need to subtract 1 from `DATA_SUFFIX.length` if `options.trailing_slash === 'always'` here:
https://github.com/sveltejs/kit/blob/a72123ad45fbfbc050441a227a204f9198818d33/packages/kit/src/runtime/server/index.js#L62

PatrickG:
Fixes #5148

### Tests

### Changesets


Rich-Harris:
Thank you! Interestingly, this PR exposed an entirely separate bug — it's possible to access `__data.json` for a standalone endpoint, when it should only apply to page endpoints. Updated the test so that it fails without this PR, and fixed that other bug for good measure

Rich-Harris (approved):

PR #5154, opened 2022-06-03
[docs] improve documentation for new developers

benmccann:
Adds code comments to help new devs around the codebase. I didn't add comments for methods where I felt the name would be self-explanatory, but tried to comment a lot of the more important methods whose functioning might be more complicated.

CC @gtm-nayan - feel free to let me know if there are others you'd like me to add here

### DIFF
--- a/packages/kit/src/core/build/utils.js
+++ b/packages/kit/src/core/build/utils.js
@@ -27,7 +27,7 @@ export async function create_build(config) {
 }
 
 /**
- * Adds the transitive JS and CSS to the js and css inputs.
+ * Adds transitive JS and CSS dependencies to the js and css inputs.
  * @param {string} file
  * @param {import('vite').Manifest} manifest
  * @param {Set<string>} css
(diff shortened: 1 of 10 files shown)


packages/kit/src/core/build/utils.js
```diff
@@ -24,6 +27,7 @@ export async function create_build(config) {
 }
 
 /**
+ * Adds the transitive JS and CSS to the js and css inputs.
```
Rich-Harris, line 30:
```suggestion
 * Adds transitive JS and CSS dependencies to the js and css inputs.
```

packages/kit/src/core/config/index.js
```diff
@@ -5,6 +5,8 @@ import { logger } from '../utils.js';
 import options from './options.js';
 
 /**
+ * Loads the template (app.html by default) and validates that it has the
```
Rich-Harris, line 8:
```suggestion
 * Loads the template (src/app.html by default) and validates that it has the
```

Rich-Harris (approved):

PR #5155, opened 2022-06-03
[breaking] remove `kit.browser.hydrate` config in favor of `compilerOptions.hydratable`

benmccann:
This is a duplicate option. Having two ways of setting it makes it really difficult to create the svelte plugin without loading the config first which ends up making plugin creation `async`. This would remove the need for `async` in https://github.com/sveltejs/kit/pull/5094

### DIFF
--- a/documentation/docs/16-seo.md
+++ b/documentation/docs/16-seo.md
@@ -88,11 +88,13 @@ An unfortunate reality of modern web development is that it is sometimes necessa
 /// file: svelte.config.js
 /** @type {import('@sveltejs/kit').Config} */
 const config = {
+	compilerOptions: {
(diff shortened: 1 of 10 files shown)


documentation/docs/16-seo.md
```diff
@@ -88,11 +88,13 @@ An unfortunate reality of modern web development is that it is sometimes necessa
 /// file: svelte.config.js
 /** @type {import('@sveltejs/kit').Config} */
 const config = {
+	compilerOptions: {
+		hydratable: false,
+	}
```
Rich-Harris, line 93:
```suggestion
	},
```

Rich-Harris (approved):

Issue #2393, opened 2021-09-08
`SyntaxError: Cannot use import statement outside a module` after  yarn3


erg:
### Describe the bug

There's an import error on startup after ``npm init svelte@next`` and ``yarn run dev`` with yarn3.

This looks like the same error message as #2161 but the issue seems different. I hope this is the correct repository for reporting this issue.

### Reproduction

Commands without output (except the last):
```
(base) ➜  npm init svelte@next my-app                                 
(base) ➜  cd my-app 
(base) ➜  my-app yarn set version berry    
(base) ➜  my-app yarn
/home/erg/my-app/.yarn/__virtual__/@sveltejs-kit-virtual-e79fbd740b/0/cache/@sveltejs-kit-npm-1.0.0-next.165-3f2325a2a6-6fd796f1e6.zip/node_modules/@sveltejs/kit/svelte-kit.js:2
import './dist/cli.js';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at Object.compileFunction (node:vm:352:18)
    at wrapSafe (node:internal/modules/cjs/loader:1031:15)
    at Module._compile (node:internal/modules/cjs/loader:1065:27)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.external_module_.Module._load (/home/erg/my-app/.pnp.cjs:10817:14)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:79:12)
    at node:internal/main/run_main_module:17:47
```

### Logs

```shell
Initial command:

# Answer defaults to the npm init questions (no/no/no/no)
(base) ➜  ~ npm init svelte@next my-app                                 
create-svelte version 2.0.0-next.78

Welcome to SvelteKit!

This is beta software; expect bugs and missing features.

If you encounter a problem, open an issue on https://github.com/sveltejs/kit/issues if none exists already.

✔ Which Svelte app template? › SvelteKit demo app
✔ Use TypeScript? … No / Yes
✔ Add ESLint for code linting? … No / Yes
✔ Add Prettier for code formatting? … No / Yes
✔ Copied project files

Want to add other parts to your code base? Visit https://github.com/svelte-add/svelte-adders, a community project of commands to add particular functionality to Svelte projects


Next steps:
  1: cd my-app
  2: npm install (or pnpm install, etc)
[...]

Issue #1557, opened 2021-05-25
Initialize SvelteKit project with Yarn


AleCandido:
**Describe the bug**
I was not able to initialize with `yarn`, i.e. the command `yarn create svelte@next foo` is failing.

**To Reproduce**
- `yarn`: 2.4.1
- `node`: v16.1.0

Then run `yarn create svelte@next foo`.

**Expected behavior**
Initialize the same way of `npm init svelte@next foo`

**Stacktraces**
<details><summary>yarn output</summary>
<pre><code>
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed in 1s 638ms
➤ YN0000: ┌ Fetch step
➤ YN0000: └ Completed
➤ YN0000: ┌ Link step
➤ YN0000: └ Completed
➤ YN0000: Done in 1s 699ms

(node:66799) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
/home/alessandro/.yarn/berry/cache/create-svelte-npm-2.0.0-next.71-3c988c4861-7.zip/node_modules/create-svelte/bin.js:2
import fs from 'fs';

^^^^^^
SyntaxError: Cannot use import statement outside a module
    at Object.compileFunction (node:vm:355:18)
    at wrapSafe (node:internal/modules/cjs/loader:1039:15)
    at Module._compile (node:internal/modules/cjs/loader:1073:27)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1138:10)
    at Module.load (node:internal/modules/cjs/loader:989:32)
    at Function.external_module_.Module._load (/tmp/xfs-14e6a52a/dlx-66788/.pnp.js:4968:14)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:76:12)
    at node:internal/main/run_main_module:17:47
</code></pre>
</details>

**Severity**
I have found a workaround and I can still use `npm`, so **low**.

**Additional context**
I found a workaround: searching on stackoverflow I found a [related question](https://stackoverflow.com/questions/58384179/syntaxerror-cannot-use-import-statement-outside-a-module) and somewhere else (I don't remember where, most likely in [yarn documentation](https://yarnpkg.com/getting-started/migration#step-by-step)) I get the second bit of the equation and the two things together made the command working:

1. add immediately a `package.json`
```json
{
   "type": "module"
}
```
[...]

Issue #993, opened 2021-04-13
Incompatible with Yarn 2


araujogui:
**Describe the bug**
Yarn 2.4.1 fails to run Svelte Kit's scripts.

**Logs**
```
➤ YN0000: ┌ Resolution step
➤ YN0032: │ fsevents@npm:2.3.2: Implicit dependencies on node-gyp are discouraged
➤ YN0032: │ fsevents@npm:2.3.2: Implicit dependencies on node-gyp are discouraged
➤ YN0002: │ @sveltejs/vite-plugin-svelte@npm:1.0.0-next.6 [6e7b7] doesn't provide rollup (p9cae2), requested by @rollup/pluginutils
➤ YN0000: │ Some peer dependencies are incorrectly met; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code
➤ YN0000: └ Completed in 11s 534ms
➤ YN0000: ┌ Fetch step
➤ YN0013: │ vite@npm:2.1.5 can't be found in the cache and will be fetched from the remote registry
➤ YN0013: │ which@npm:2.0.2 can't be found in the cache and will be fetched from the remote registry
➤ YN0013: │ wide-align@npm:1.1.3 can't be found in the cache and will be fetched from the remote registry
➤ YN0013: │ wrappy@npm:1.0.2 can't be found in the cache and will be fetched from the remote registry
➤ YN0013: │ yallist@npm:4.0.0 can't be found in the cache and will be fetched from the remote registry
➤ YN0000: └ Completed in 2s 749ms
➤ YN0000: ┌ Link step
➤ YN0007: │ svelte-preprocess@npm:4.7.0 [2d2f7] must be built because it never did before or the last one failed
➤ YN0000: └ Completed
➤ YN0000: Done with warnings in 14s 455ms
```

```
(node:5987) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
/home/araujo/Área de trabalho/eurora.store/.yarn/$$virtual/@sveltejs-kit-virtual-6e7b74d398/0/cache/@sveltejs-kit-npm-1.0.0-next.74-7ec168993f-79c42c4576.zip/node_modules/@sveltejs/kit/svelte-kit.js:2
import './dist/cli.js';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at Object.compileFunction (node:vm:355:18)
[...]

PR #5169, opened 2022-06-06
Added info for using yarn berry with yarn 3.2.1

Hecatron:
Hi,
I had a quick look around at seeing if I could use yarn berry with enableGlobalCache set to true.
I think yarn berry is generally discouraged although I've managed to I think get it working ok with a small change to the svelte.config.js

I suspect there's been some changes to yarn version 3 which now allows it to work (or at least allows the demo app to work)
This bug doesn't seem to exist anymore https://github.com/sveltejs/kit/issues/2393

Also spotted these related to berry
  * https://github.com/sveltejs/kit/issues/993
  * https://github.com/sveltejs/kit/issues/1557

Currently I'm using Windows and yarn version 3.2.1

### DIFF
--- a/documentation/faq/80-integrations.md
+++ b/documentation/faq/80-integrations.md
@@ -129,8 +129,8 @@ yarn install
 
 **Yarn 3 global cache**
 
-One of the more interesting features of Yarn Berry is the ability to have a single global cache for packages, instead of having multiple copies for each project on the disk. However, setting **enableGlobalCache** to true causes building to fail, so it is recommended to add the following to the `.yarnrc.yml` file:
+One of the more interesting features of Yarn Berry is the ability to have a single global cache for packages, instead of having multiple copies for each project on the disk. However, setting `enableGlobalCache` to true causes building to fail, so it is recommended to add the following to the `.yarnrc.yml` file:
 ```
 nodeLinker: node-modules
 ```
-This will cause packages to be downloaded into a local node_modules directory but avoids the above problem and is you're best bet for using version 3 of yarn at this point in time.
+This will cause packages to be downloaded into a local node_modules directory but avoids the above problem and is your best bet for using version 3 of Yarn at this point in time.


benmccann (commented):
I don't know how much we want to encourage Yarn usage, but I guess if it basically works out-of-the-box then it doesn't hurt to give a suggestion. I'm curious what the other maintainers think though

documentation/faq/80-integrations.md
```diff
@@ -106,3 +106,29 @@ onMount(() => {
 ### Does it work with Yarn 2?
 
 Sort of. The Plug'n'Play feature, aka 'pnp', is broken (it deviates from the Node module resolution algorithm, and [doesn't yet work with native JavaScript modules](https://github.com/yarnpkg/berry/issues/638) which SvelteKit — along with an [increasing number of packages](https://blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77) — uses). You can use `nodeLinker: 'node-modules'` in your [`.yarnrc.yml`](https://yarnpkg.com/configuration/yarnrc#nodeLinker) file to disable pnp, but it's probably easier to just use npm or [pnpm](https://pnpm.io/), which is similarly fast and efficient but without the compatibility headaches.
+
+### Yarn 3 Berry Example
+
+Currently ESM Support within the latest yarn (version 3) is considered experimental.  
+The below seems to work although you're results may vary.
```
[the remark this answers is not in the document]
grbd: Changes now made, I do recognise that there may be unforeseen problems I've not picked up on and that its unsupported etc. But I just noticed it worked as is and thought it might be useful to mention the config change.

Personally I've used pnpm before but then moved to yarn berry with enableGlobalCache since it seems like a cleaner option moving forward without all the filesystem links.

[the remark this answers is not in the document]
grbd: Ok I just did a bit more testing on this
It seems to work the first time you run "yarn dev", but then fails on attempts afterwards with a "p.replace is not a function" error. Until you delete the .svelte-kit directory

documentation/faq/80-integrations.md
```diff
@@ -103,6 +103,47 @@ onMount(() => {
...
+```json
+	kit: {
+		vite: {server: {fs: {allow: ['.yarn']}}},
+                ...
+	}
+```
+
+However please note that this seems to be a bit flakey, it runs the first time you run `yarn dev` but then fails with a `p.replace is not a function` error on the 2nd or follow up attempts to run the site in dev, until you delete the `.svelte-kit` directory
+
+Until the above is fixed I'd instead recommend adding the following to the `.yarnrc.yml` file
```
benmccann, line 145: I don't think we should document something that doesn't work very well
```suggestion

One of the more interesting features of Yarn Berry is the ability to have a single global cache for packages, instead of having multiple copies for each project on the disk. However, setting **enableGlobalCache** to true causes building to fail, so it is recommended to add the following to the `.yarnrc.yml` file:
```

grbd: Now patched

documentation/faq/80-integrations.md
```diff
@@ -103,6 +103,41 @@ onMount(() => {
...
 ```
 
-### Does it work with Yarn 2?
+### How do I use with Yarn?
+
+#### Does it work with Yarn 2?
 
 Sort of. The Plug'n'Play feature, aka 'pnp', is broken (it deviates from the Node module resolution algorithm, and [doesn't yet work with native JavaScript modules](https://github.com/yarnpkg/berry/issues/638) which SvelteKit — along with an [increasing number of packages](https://blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77) — uses). You can use `nodeLinker: 'node-modules'` in your [`.yarnrc.yml`](https://yarnpkg.com/configuration/yarnrc#nodeLinker) file to disable pnp, but it's probably easier to just use npm or [pnpm](https://pnpm.io/), which is similarly fast and efficient but without the compatibility headaches.
+
+#### Yarn 3 Berry Example
```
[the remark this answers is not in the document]
grbd: Now patched

documentation/faq/80-integrations.md
```diff
@@ -103,6 +103,34 @@ onMount(() => {
...
+cd myapp
+```
+
+And enable Yarn Berry:
+```sh
+yarn set version berry
+yarn install
+```
+
+#### Yarn 3 Global Cache
```
benmccann, line 130: This doesn't seem like a separate FAQ, so I wouldn't use a heading. You can make the text bold though:
```suggestion
**Yarn 3 global cache**
```

documentation/faq/80-integrations.md
```diff
@@ -103,6 +103,34 @@ onMount(() => {
...
+yarn install
+```
+
+**Yarn 3 global cache**
+
+One of the more interesting features of Yarn Berry is the ability to have a single global cache for packages, instead of having multiple copies for each project on the disk. However, setting **enableGlobalCache** to true causes building to fail, so it is recommended to add the following to the `.yarnrc.yml` file:
+```
+nodeLinker: node-modules
+```
+This will cause packages to be downloaded into a local node_modules directory but avoids the above problem and is you're best bet for using version 3 of yarn at this point in time.
```
benmccann, line 136: "you're" is short for "you are"
```suggestion
This will cause packages to be downloaded into a local node_modules directory but avoids the above problem and is your best bet for using version 3 of Yarn at this point in time.
```

benmccann (approved):
lgtm, but I will leave for the other maintainers to weigh in

Rich-Harris (approved):

Rich-Harris:
I think this strikes the right balance between providing people with the right information and steering people towards tools with fewer... idiosyncracies — thanks 👍

PR #5170, opened 2022-06-06
[chore] upgrade to Vite 2.9.10

benmccann:
This adds `configurePreviewServer`, which is necessary to use `vite.config.js` / `@sveltejs/kit/experiemental/vite` in preview mode


packages/kit/src/cli.js
```diff
@@ -200,15 +200,34 @@ prog
...
 
-			const { preview } = await import('./core/preview/index.js');
+			const { sveltekit_plugin } = await import('./core/preview/index.js');
+			const svelte_config = await load_config();
+			const vite_config = await svelte_config.kit.vite();
+
+			/** @type {import('vite').UserConfig} */
+			const config = {
+				...vite_config,
+				plugins: [...(vite_config.plugins || []), sveltekit_plugin]
```
PH4NTOMiki, line 212:
```suggestion
				plugins: [...(vite_config.plugins || []), ...sveltekit_plugin]
```

benmccann: `sveltekit_plugin` is just a single plugin, so it wouldn't make sense to spread it. This is different than dev because in dev we need `vite-plugin-svelte` which adds a second plugin, but in preview the `.svelte` files have already been compiled, so there's no need for `vite-plugin-svelte`

PR #5178, opened 2022-06-07
Revert "[breaking] remove `kit.browser.hydrate` config in favor of `compilerOptions.hydratable`"

Rich-Harris:
we messed this one up — it turns out it doesn't enable the thing we thought it would enable, so we're making a change elsewhere instead. in the meantime, there's no sense in removing the `browser.hydrate` option. sorry for the churn, for anyone who upgraded to next.349 already


dummdidumm (commented):
The changeset that was deleted needs to be undeleted I think, else all good

PR #5180, opened 2022-06-07
fix: add redirects for routes containing '__data.json' suffix

### LINKED ISSUE #4872: In adapter-netlify split:true config is breaking SSR and other things.
### Describe the bug

When `split: true` is enabled in the `svelte.config.js` for `adapter-netlify`, SSR routes when navigated to with client side navigation gets a 404. The endpoint route will work on refresh, but the todo page goes to a Netlify 404 page does not exist skipping the `__error.svelte` route.

### Reproduction

Repo of demo app with `adapter-netlify`: https://github.com/brittneypostma/test-ssr-split

Preview deploy of site with `split: false` and `__error.svelte` working: https://627c04fe0cebc30008f3f36b--test-ssr-split.netlify.app/endpoint

Preview deploy of site with `split: true` and `__error.svelte` not catching on /todos refresh: https://627c046d20bc620009be8eef--test-ssr-split.netlify.app/endpoint

### Logs

```shell
`GET /endpoint/__data.json 404`
```

### System Info

```shell
System:
    OS: macOS 12.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 127.41 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.15.0 - ~/.nvm/versions/node/v16.15.0/bin/node
    npm: 8.5.5 - ~/.nvm/versions/node/v16.15.0/bin/npm
  Browsers:
    Chrome: 101.0.4951.54
    Firefox Developer Edition: 101.0
    Safari: 15.4
  npmPackages:
    @sveltejs/adapter-netlify: 1.0.0-next.56 => 1.0.0-next.56 
    @sveltejs/kit: next => 1.0.0-next.326 
    svelte: ^3.46.0 => 3.48.0 
```

### Severity

### Additional Information

Only workaround is to turn off split, which isn't ideal.

ericapisani:
Fixes #4872

We also needed to `await` the `builder.createEntries` call as it was not completing before the `redirects` array was being written to the `build/redirects` file.

Demo of site before fix: https://627c046d20bc620009be8eef--test-ssr-split.netlify.app/endpoint
Demo of site after fix: https://ep-svelte.netlify.app/

## Way to reproduce/confirm fix
* Visit one of the above links
* Visit either 'endpoints' or 'todos' page

**Before:** 
* Would see a 404 on both pages
* Refreshing on 'endpoints' page would render the page correctly, but not on `todos`
* Checking the network tab when navigating to `todos` through client side navigation showed that the route for `todos/__data.json` (and the related one for `endpoints`) was returning a 404

**After:** 
* Correct page will render, both in client side navigation and when refreshing on the page itself

### Tests

### Changesets

### DIFF
--- a/.changeset/tasty-zebras-guess.md
+++ b/.changeset/tasty-zebras-guess.md
@@ -2,4 +2,4 @@
 '@sveltejs/adapter-netlify': patch
 ---
 
-fix: add redirects for routes containing '\_\_data.json' suffix
+fix: add redirects for routes containing `__data.json` suffix


nickytonline (approved):
Thanks for the fix @erica. Looks great!

Rich-Harris:
excellent, thanks so much!

PR #5183, opened 2022-06-08
fix: URL instance methods work in `load`

### LINKED ISSUE #5083: Can't `location.assign(url)` in `load()` function: throws `Illegal invocation`
### Describe the bug

I have a fallthrough route, which I want to _always be server-rendered_. The use-case is that we're migrating from Angular to SvelteKit, and any pages which SvelteKit doesn't recognise should:
- if invoked client-side, reload the URL to force a SSR
- return a page which boots AngularJS, with no hydration or SvelteKit router
- this ensures that only AngularJS is running in the browser, as SvelteKit will not send any of its own JS with the SSR response

The fix for [this issue](https://github.com/sveltejs/kit/issues/4788) (PR [here](https://github.com/sveltejs/kit/pull/4983)) introduced a `Proxy` wrapper for the `url` input parameter to the `load()` function. This now causes my browser-side interceptor code to throw a `TypeError: Illegal invocation`.

### Reproduction

Click the link to the fallthrough page in this repro: https://stackblitz.com/edit/sveltejs-kit-template-default-cn2bpx?file=src%2Froutes%2F[...fallthrough].svelte&terminal=dev, and you'll see:

```
500
Illegal invocation
TypeError: Illegal invocation
    at load (https://sveltejs-kit-template-default-cn2bpx--3000.local.webcontainer.io/src/routes/[...fallthrough].svelte?t=1653555167052:81:12)
    at load_node (https://sveltejs-kit-template-default-cn2bpx--3000.local.webcontainer.io/@fs/home/projects/sveltejs-kit-template-default-cn2bpx/.svelte-kit/runtime/client/start.js:1029:33)
    at load_route (https://sveltejs-kit-template-default-cn2bpx--3000.local.webcontainer.io/@fs/home/projects/sveltejs-kit-template-default-cn2bpx/.svelte-kit/runtime/client/start.js:1147:20)
    at async update (https://sveltejs-kit-template-default-cn2bpx--3000.local.webcontainer.io/@fs/home/projects/sveltejs-kit-template-default-cn2bpx/.svelte-kit/runtime/client/start.js:626:38)
    at async navigate (https://sveltejs-kit-template-default-cn2bpx--3000.local.webcontainer.io/@fs/home/projects/sveltejs-kit-template-default-cn2bpx/.svelte-kit/runtime/client/start.js:1372:3)
```

Change the `location.assign(url)` to `location.assign(url.href)` and it works fine.

### Logs

_No response_

### System Info

```shell
`@sveltejs/kit@1.0.0-next.345`
```

### Severity

### Additional Information

This may purely be a documentation/education thing. The [docs say](https://kit.svelte.dev/docs/loading#input-url) **`url` is an instance of [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL)**, but that's not entirely true, as the implicit casting to string with `location.assign(url)` is broken by the Proxy wrapper. Either we should document that limitation, or, if possible, improve the Proxy with a `toString()` method or whatever is required.

tcc-sejohnson:
Fixes #5083.

This is the "surgical approach" -- just replace the `hash` property with a custom getter. The other option is (if we're really committed to using a Proxy):

```typescript
new Proxy(url, {
    get: (target, property, receiver) => {
        if (property === 'hash') {
            throw new Error(
                'url.hash is inaccessible from load. Consider accessing hash from the page store within the script tag of your component.'
            );
        }

        if (!(property in target)) {
          return undefined;
        }
        const value = target[property];
        return typeof value === "function"
          ? (...args) => value.apply(target, args)
          : value;
    }
});
```

This scares me a little bit, because it's a lot harder to anticipate what other side effects it might exhibit. 

### Tests

### Changesets

### DIFF
--- a/.changeset/popular-pets-obey.md
+++ b/.changeset/popular-pets-obey.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] URL instance methods now work in `load`
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -2,7 +2,7 @@ import { onMount, tick } from 'svelte';
 import { writable } from 'svelte/store';
 import { coalesce_to_error } from '../../utils/error.js';
 import { normalize } from '../load.js';
-import { normalize_path } from '../../utils/url.js';
+import { LoadURL, normalize_path } from '../../utils/url.js';
 import {
 	create_updated_store,
 	find_anchor,
@@ -527,6 +527,7 @@ export function create_client({ target, session, base, trailing_slash }) {
 		}
 
 		const session = $session;
+		const load_url = new LoadURL(url);
 
 		if (module.load) {
 			/** @type {import('types').LoadEvent} */
@@ -536,18 +537,6 @@ export function create_client({ target, session, base, trailing_slash }) {
 				props: props || {},
 				get url() {
 					node.uses.url = true;
-					const load_url = new URL(url);
-
-					// We could use a Proxy here, but it causes weird effects.
-					// See: https://github.com/sveltejs/kit/issues/5083
-					Object.defineProperty(load_url, 'hash', {
-						get: () => {
-							throw new Error(
-								'url.hash is inaccessible from load. Consider accessing hash from the page store within the script tag of your component.'
-							);
-						}
-					});
-
 					return load_url;
 				},
 				get session() {
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -536,18 +536,19 @@ export function create_client({ target, session, base, trailing_slash }) {
 				props: props || {},
 				get url() {
 					node.uses.url = true;
+					const load_url = new URL(url);
 
-					return new Proxy(url, {
-						get: (target, property) => {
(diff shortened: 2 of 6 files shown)


Rich-Harris:
Thank you — this is definitely the right approach, rather than making the proxy more sophisticated — confirmed at [excruciating length](https://twitter.com/Rich_Harris/status/1539375179394056192) via Twitter.

I made a couple of small tweaks:

* using a `LoadURL` subclass instead of monkey patching — feels less naughty, somehow
* if someone wasn't using destructuring, we'd want `event.url === event.url` to be true, so create the `LoadURL` instance eagerly and return it from the getter
* use it for SSR as well as client-side navigation, so that the error is consistent

It occurred to me that we probably want to do something similar for the prerendering proxy (the one that blocks use of `url.search[Params]`), but I left that out for now just to make life easy

tcc-sejohnson:
Ugh, thanks for that Twitter thread. After reading a few dozen subthreads, I think I'm _starting_ to understand why the proxy approach doesn't work: It's because JavaScript is stupid. 🤣 

If we want to implement similar behavior for prerendering, I'm happy to tackle that. I can do it as part of this PR or as a separate one after this one is merged.

Rich-Harris (approved):

Issue #4902, opened 2022-05-13
afterNavigate not firing in dev mode after a code change


jp-simons:
### Describe the bug

In dev mode, after I make a code change, the route component re-mounts itself but it never re-calls `afterNavigate`. This leaves things in an inconsistent state. Not a huge annoyance but I do have to do ctrl+R in development.

### Reproduction

Repro is this branch of this repo:
https://github.com/johnnysprinkles/sveltekit_after_navigate_bug/tree/after_nv

It's the Demo app from create-svelte, but extended to have a parameterized route. The final commit in the history just adds some `console.log` statements to see what's going on in the lifecycle events.

To repro just clone that and run it with "npm run dev", and create some TODO items, then navigate to one of them (using the newly added "i" icon, or directly via URL e.g. http://localhost:3000/todos/4110e0d9-56a9-4106-9d83-1fada5087660

Now if you make an arbitrary change to routes/todos/[uid].svelte you'll see how onMount runs but afterNavigate doesn't. Looks like:

![Screen Shot 2022-05-13 at 10 00 41 AM](https://user-images.githubusercontent.com/32530010/168332719-0dd99f49-97c8-406d-8541-123d81a407fe.png)

### System Info

```shell
System:
    OS: macOS 11.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4750HQ CPU @ 2.00GHz
    Memory: 510.79 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.4.1 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 8.9.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 101.0.4951.64
    Firefox: 100.0
    Safari: 14.1.2
    Safari Technology Preview: 15.4
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.33
    @sveltejs/kit: next => 1.0.0-next.302
    svelte: ^3.46.0 => 3.46.4
```

### Severity

annoyance

Rich-Harris:
This one is tricky — since the component accepts the hot update, the main `client.js` has no opportunity to detect that that's happened and re-run `afterNavigate` functions that were added during the update. (The Vite client [does emit events](https://vitejs.dev/guide/api-hmr.html#hot-on-event-cb), but not the ones we need AFAICT.)

[...]

PR #5195, opened 2022-06-11
Exposing event to load(), and removing the automatic header forwarding

### LINKED ISSUE #696: Server-side fetch request should have headers

### LINKED ISSUE #3503: Ability to include credentials on external server-side fetch() calls

johnnysprinkles:
Fixes #3503 Fixes #696
Reverts #3631 Reverts #835

Okay this is the last one of my SvelteKit wishlist items. Not sure what you'll think but it certainly makes sense to me.

This does two things, each in its own commit.

1. Exposes `event` to the load() function
2. Reverts and deletes all the logic related to passing headers and cookies to the upstream service calls.

#3631 was clearly a mistake. If we have a network call graph that looks something like this:

Screen Shot 2022-05-30 at 9 38 42 PM

It doesn't make any sense for the Node service to pretend to be a browser. Some of the reasons:

* You're lying about the user agent. If you passed one at all it should be, perhaps, `node/sveltekit` or something
* You're passing headers unidirectionally, sending them out but but not sending the response headers back to the browser, because of course that wouldn't make any sense. The calls are many to one so what should you do, merge them?
* The type is different (HTML vs most likely JSON), so the `accept` header could be wrong
* Some of the headers are nonsensical for service-to-service calls, like `referer` which is a browser concern

There just seems to be some confusion and it seems like people are considering calls A-C in the graph to be the Z call proxied onward. When in fact those calls are unrelated to call Z, except that call Z is the reason calls A-C are being made.

But then even with all the complexity, we still can't pass along the most important information: the auth cookies. To get the full details of the incoming request you have to use an endpoint (or a hook). Endpoints should be a feature you can use if you want but shouldn't be mandatory.

In a more ops-driven, "close to the metal" kind of enterprise setup, our frontend services live among a constellation of other services, both in front as reverse proxies, and behind as upstream services. In this world things like SSL termination and endpoints are simply not wanted or needed. In this world if we want to avoid the moving target of CORS and serve our data on the same domain as our HTML, we simply put a reverse proxy in front and split traffic.

What this change does is restores control. If you want to call your backend services and spoof like you're a browser, you can do so! Copy all your headers over. If you want to forward a backend service's "set-cookie" directive to the browser, go ahead. But if you do nothing the calls to the backends get no headers, just as if you made a curl call, which seems like a pretty reasonable thing to do by default.

Here's some examples:

```javascript
export default async function load({ event, params, fetch }) {
  let response = await fetch('https://a.backend.com/data/' + event.id, {
    headers: {
      // Use end user credentials
      // You could parse the cookie string and slice out a specific auth cookie, or
      // or just pass the whole darn thing.
      cookie: event?.request.headers.get('cookie'), // only relevant server-side
      credentials: 'include', // only relevant client-side
    },
  });
  ...
}
```

If your backend sets a cookie and you want to set the same cookie in the browser, there's no way to do that explicitly, and I stripped out the automatic logic for that because, hey, making service calls shouldn't have side effects. You get the data, you do what you want with the data, end of story. So I extended some interfaces with a `set_cookies` string array.

```javascript
export default async function load({ event, params, fetch }) {
  let response = await fetch('https://a.backend.com/data/' + event.id, {
    headers: {
      cookie: event?.request.headers.get('cookie'), // only relevant server-side
      credentials: 'include', // only relevant client-side
    },
  });
  let data = await response.json();
  let set_cookie = response.headers.get('set-cookie');
  return {
    props: {
      widgets: data.widgets
    },
    set_cookies: set_cookie && [set_cookie],
}
```

**More rationale:** Even in the demo TODO app, you need access to the user's cookies to make the backend call. I know that backend is slated for removal in #4264 but as long as it remains it does prove a point. In my repro case for #4902 I have code like this:

https://github.com/johnnysprinkles/sveltekit_after_navigate_bug/commit/53dd01ad7f4f07cd1e5cfc31fb2782048ff258d1#diff-87b424ffe45157388b79be6ab8b5f706a521128cf102a15a3b58f480f4411a85R27

where I have to put the userid in session because there's no way to read it from the request in `load()`. If you don't want to put it in session and you don't want to use an endpoint, you're currently out of luck.

Anyway, I hope this might at least provoke some discussion. I know this goes against a SvelteKit value of abstracting away the differences between client and server, but that abstraction is kind of a fiction anyway.

Also, I have no idea how much this would break. Shadow endpoints? Various adapters?

### Tests

### Changesets

### DIFF
--- a/.changeset/khaki-ways-applaud.md
+++ b/.changeset/khaki-ways-applaud.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': minor
+---
+
+Exposing the event to server-side load() and removing the automatic header forwarding on th modded fetch()
--- a/documentation/docs/04-loading.md
+++ b/documentation/docs/04-loading.md
@@ -50,7 +50,7 @@ It is recommended that you not store per-request state in global variables, but 
 
 ### Input
 
-The `load` function receives an object containing eight fields — `url`, `params`, `props`, `fetch`, `session`, `stuff`, `status`, and `error`. The `load` function is reactive, and will re-run when its parameters change, but only if they are used in the function. Specifically, if `url`, `session` or `stuff` are used in the function, they will be re-run whenever their value changes, and likewise for the individual properties of `params`.
+The `load` function receives an object containing nine fields — `url`, `params`, `props`, `fetch`, `session`, `stuff`, `status`, `error`, and `event`. The `load` function is reactive, and will re-run when its parameters change, but only if they are used in the function. Specifically, if `url`, `session` or `stuff` are used in the function, they will be re-run whenever their value changes, and likewise for the individual properties of `params`.
 
 > Note that destructuring parameters in the function declaration is enough to count as using them.
 
@@ -104,6 +104,10 @@ If the page you're loading has an endpoint, the data returned from it is accessi
 #### error
 
 `error` is the error that was thrown (or returned from a previous `load`) when rendering an error page, or `null` otherwise.
+
+#### event
+
+`event` represents the incoming request. This will be null on the client.
 
 ### Output
 
--- a/documentation/docs/04-loading.md
+++ b/documentation/docs/04-loading.md
@@ -159,3 +159,7 @@ The combined `stuff` is available to components using the [page store](/docs/mod
 An array of strings representing URLs the page depends on, which can subsequently be used with [`invalidate`](/docs/modules#$app-navigation-invalidate) to cause `load` to rerun. You only need to add them to `dependencies` if you're using a custom API client; URLs loaded with the provided `fetch` function are added automatically.
 
 URLs can be absolute or relative to the page being loaded, and must be [encoded](https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding).
+
+#### set_cookies
+
+An array of strings, each in standard cookie form with key/value and semicolon delimited attributes, e.g. "cookie1=value;maxage=3600;path=/".
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -597,7 +597,8 @@ export function create_client({ target, session, base, trailing_slash }) {
 					return started ? native_fetch(normalized, init) : initial_fetch(requested, init);
 				},
 				status: status ?? null,
-				error: error ?? null
+				error: error ?? null,
+				event: null
 			};
 
 			if (import.meta.env.DEV) {
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -6,7 +6,6 @@ import { is_root_relative, resolve } from '../../../utils/url.js';
 import { create_prerendering_url_proxy } from './utils.js';
 import { is_pojo, lowercase_keys, normalize_request_method } from '../utils.js';
 import { coalesce_to_error } from '../../../utils/error.js';
-import { domain_matches, path_matches } from './cookie.js';
 
 /**
  * Calls the user's `load` function.
@@ -44,8 +43,6 @@ export async function load_node({
 
 	/** @type {Array<import('./types').Fetched>} */
 	const fetched = [];
-
-	const cookies = cookie.parse(event.request.headers.get('cookie') || '');
 
 	/** @type {import('set-cookie-parser').Cookie[]} */
 	const new_cookies = [];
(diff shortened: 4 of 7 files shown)


Conduitry:
As I mentioned in https://github.com/sveltejs/kit/issues/3503#issuecomment-1029094194, exposing `event` to `load` is bad because you completely lose the security of HTTP-only cookies.

johnnysprinkles:
I guess I never quite understood that point... this is a change inside the running NodeJS code, not sure how that could be a security issue. I looked back at your comment and see that the server-side fetch currently tries to simulate what the browser's `fetch` would do, but why impose those kind of elective limitations on your own code in the serverside?

johnnysprinkles:
Speaking of HTTP only cookie though, I was too aggressive in pruning and should have left in the code that strips out `set-cookie` and `etag` from the `fetched` headers. Because that information might not want to be exposed to client logic. We could look at the cookie value itself and look for `HttpOnly` but for now I just reverted that file back to how it was, so no `set-cookie` value every appears on the client via that "serialized fetch result" data passing mechanism. (See update.)

johnnysprinkles:
Maybe I'm thinking about this wrong... if you want a "pure" frontend, that is, a frontend which never speaks data, only listens to data and speaks HTML (and HTML-consumed code), there's no need for endpoints. But shadow endpoints aren't endpoints at all? They're actually just an `if (!browser) {}` code block pulled out into a separate file? If that's the code, maybe all my network topology diagram stuff really applies in there, in the shadow endpoint.

johnnysprinkles:
More thoughts about this... if right now we're implementing the browser's XHR security model in very broad strokes, it's interesting to think about what would be involved in implementing it exactly correctly:

* Respect the various "credentials" options, not just "omit"
* Understand and respect all the various `Access-Control-*` headers, and there's a lot of them
* Know if the request should be preflighted or not
* Change all the rules depending on if it's HTTP or HTTPS
* Keep this all up to date as things change (isn't Chrome always threatening to remove third party cookies altogether?)
* Maybe even use browser sniffing to make the behavior match the user's UA
* Ideally we'd even match older browser versions, keeping a historical record of how CORS works in all browsers across all versions

This doesn't seem like a business we want to be in.

I did a quick search and we have 79 open SvelteKit issues containing "fetch". Just on the first page I see a couple that this PR might resolve. Would be pretty nice to be able to close all those out.

PR #5197, opened 2022-06-11
feat: adapter-vercel should be able to generate sourcemap

ACHP:

### Tests

### Changesets

## Why
This pull-request adds a new option to enable sourcemap for the generated build. This is really important for error monitoring integration. (eg; Sentry in my case).
For now there is no easy way to debug a production app deployed with sveltkit on vercel.

### DIFF
--- a/packages/adapter-vercel/README.md
+++ b/packages/adapter-vercel/README.md
@@ -30,7 +30,11 @@ export default {
 
       // if true, will split your app into multiple functions
       // instead of creating a single one for the entire app
-      split: false
+      split: false,
+
+      // same as esbuild sourcemap option
+			// https://esbuild.github.io/api/#sourcemap
+      sourcemap: false
     })
   }
 };
--- a/packages/adapter-vercel/index.d.ts
+++ b/packages/adapter-vercel/index.d.ts
@@ -1,9 +1,11 @@
 import { Adapter } from '@sveltejs/kit';
+import type { BuildOptions } from 'esbuild';
 
 type Options = {
 	edge?: boolean;
 	external?: string[];
 	split?: boolean;
+	sourcemap?: BuildOptions['sourcemap'];
 };
 
 export default function plugin(options?: Options): Adapter;
--- a/packages/adapter-vercel/index.js
+++ b/packages/adapter-vercel/index.js
@@ -84,19 +84,19 @@ const redirects = {
 const files = fileURLToPath(new URL('./files', import.meta.url).href);
 
 /** @type {import('.').default} **/
-export default function ({ external = [], edge, split } = {}) {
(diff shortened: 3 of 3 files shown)


.changeset/mighty-melons-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/adapter-vercel': patch
+---
+
+Add new option to generate sourcemap for the `functions/node/render/index.js` generated file
```
benmccann, line 5:
```suggestion
Add new option to create a sourcemap for the generated functions
```

benmccann:
We probably should always generate sourcemaps, but I wonder what the default should be. It's not a simple boolean but an enum. To generate them as separate files probably is a good default, but I could see people wanting something different

Rich-Harris:
> I could see people wanting something different

I can't. I think we should just always do linked sourcemaps (i.e. `true`) — if someone comes up with a compelling rationale for exposing an option, we could easily add it in a non-breaking way in future, but until then I think we should abide by the principle that configuration is a last resort

benmccann:
Making it `linked` until someone needs something else sounds like a fine solution to me

ACHP:
Thank you for your reviews, I updated this PR according to your comments. However the CI seems broken on Windows but I don't think it's related to this pull-request

benmccann:
I reran the Windows CI and it passed. I wouldn't worry if it's only one test job that fails. We have some flaky tests

benmccann:
My one other question is whether we also need to make the same change to the other adapters that use esbuild. I'm guessing we probably do, but haven't checked

ACHP:
> My one other question is whether we also need to make the same change to the other adapters that use esbuild. I'm guessing we probably do, but haven't checked

Yes, there is no reason not to do it on other adapters I guess. I did it on the vercel adapter because this is what I'm working on at the moment and it's easy for me to test it, so I'm was confident enought to do it. Do you think other adapters should be updated in this PR as well or should we merge it as is and create other pull-request for different adapters ?

benmccann:
I'd rather we just do them all in this PR if you don't mind updating it. Only a few of the adapters use esbuild, so hopefully there won't be too much to update.

benmccann:
You don't have to provide the source maps to anyone outside your company, so I don't see that as a reason to make it configurable

Rich-Harris:
@8462852 to be clear, we're only talking about sourcemaps for your serverless/edge functions. The sourcemapped stack traces will not be visible to anyone who doesn't have access to https://vercel.com/YOUR_ORG/YOUR_PROJECT/YOUR_DEPLOYMENT_ID/functions?name=YOUR_FUNCTION

Rich-Harris (approved):

Rich-Harris:
thank you!

PR #5198, opened 2022-06-12
[docs] be more specific about the definition of page endpoints

gtm-nayan:

### Tests

### Changesets

### DIFF
--- a/documentation/docs/02-routing.md
+++ b/documentation/docs/02-routing.md
@@ -100,7 +100,7 @@ The job of a [request handler](/docs/types#sveltejs-kit-requesthandler) is to re
 
 #### Page endpoints
 
-If an endpoint has the same filename as a page (except for the extension), the page gets its props from the endpoint — via `fetch` during client-side navigation, or via direct function call during SSR.
+If an endpoint has the same filename as a page (including named layouts and matchers but not the extension), the page gets its props from the endpoint — via `fetch` during client-side navigation, or via direct function call during SSR.
 
 A page like `src/routes/items/[id].svelte` could get its props from the `body` in the endpoint above:
 
--- a/documentation/docs/02-routing.md
+++ b/documentation/docs/02-routing.md
@@ -100,7 +100,7 @@ The job of a [request handler](/docs/types#sveltejs-kit-requesthandler) is to re
 
 #### Page endpoints
 
-If an endpoint has the same filename as a page (including named layouts and matchers but not the extension), the page gets its props from the endpoint — via `fetch` during client-side navigation, or via direct function call during SSR.
+If an endpoint has the same filename (including named layouts and matchers but not the extension) as a page, the page gets its props from the endpoint — via `fetch` during client-side navigation, or via direct function call during SSR.
 
 A page like `src/routes/items/[id].svelte` could get its props from the `body` in the endpoint above:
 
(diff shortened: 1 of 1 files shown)


kristjanmar:
I think this is harder to understand than the current sentence and I'm not sure what it means. But I agree that the docs on page endpoints can be improved, especially with more info about how they are different from the load function.

gtm-nayan:
@kristjanmar how about now?

kristjanmar:
Much better.

"The filename for a page endpoint must also include any [named layouts](/docs/layouts#named-layouts) or [matchers](/docs/routing#advanced-routing-matching) in its page's name."

Maybe something like this would be optimal just to make it more clear that this is an IF:

"If the page name has [named layouts](/docs/layouts#named-layouts) or [matchers](/docs/routing#advanced-routing-matching), then the filename for the page endpoint also needs to include them."

documentation/docs/02-routing.md
```diff
@@ -100,7 +100,7 @@ The job of a [request handler](/docs/types#sveltejs-kit-requesthandler) is to re
 
 #### Page endpoints
 
-If an endpoint has the same filename as a page (except for the extension), the page gets its props from the endpoint — via `fetch` during client-side navigation, or via direct function call during SSR.
+If an endpoint has the same filename (except the extension) as a page, the page gets its props from the endpoint — via `fetch` during client-side navigation, or via direct function call during SSR. If a page has [named layouts](/docs/layouts#named-layouts) or [matchers](/docs/routing#advanced-routing-matching) in its filename then its page endpoint's filename must also include them.
```
benmccann, line 103:
```suggestion
If an endpoint has the same filename (except the extension) as a page, the page gets its props from the endpoint — via `fetch` during client-side navigation, or via direct function call during SSR. If a page uses syntax for [named layouts](/docs/layouts#named-layouts) or [matchers](/docs/routing#advanced-routing-matching) in its filename then the corresponding page endpoint's filename must also include them.
```

documentation/docs/02-routing.md
```diff
@@ -100,7 +100,7 @@ The job of a [request handler](/docs/types#sveltejs-kit-requesthandler) is to re
 
 #### Page endpoints
 
-If an endpoint has the same filename as a page (except for the extension), the page gets its props from the endpoint — via `fetch` during client-side navigation, or via direct function call during SSR.
+If an endpoint has the same filename (except the extension) as a page, the page gets its props from the endpoint — via `fetch` during client-side navigation, or via direct function call during SSR. If a page uses syntax for [named layouts](/docs/layouts#named-layouts) or [matchers](/docs/routing#advanced-routing-matching) in its filename then the corresponding page endpoint's filename must also include them.
```
benmccann, line 103: I found the first sentence easier to read before
```suggestion
If an endpoint has the same filename as a page (except for the extension), the page gets its props from the endpoint — via `fetch` during client-side navigation, or via direct function call during SSR. If a page uses syntax for [named layouts](/docs/layouts#named-layouts) or [matchers](/docs/routing#advanced-routing-matching) in its filename then the corresponding page endpoint's filename must also include them.
```

Rich-Harris (approved):

Rich-Harris:
much better 👍

PR #5204, opened 2022-06-14
fix FAQ h1 spacing

nimatrengo:
## This PR is not important as it is fixing a UX issue on the FAQ page. 
Texts were not selectable as the H1 had a negative margin (Not a good practice)

### Tests

### Changesets

### DIFF
--- a/sites/kit.svelte.dev/src/routes/faq/index.svelte
+++ b/sites/kit.svelte.dev/src/routes/faq/index.svelte
@@ -79,8 +79,8 @@
 	}
 
 	h2 {
-		margin: -4rem 0 1rem 0;
-		padding-top: 10rem;
+		margin: 0rem 0 1rem 0;
+		padding-top: 6rem;
 		padding-bottom: 0.2rem;
 		color: var(--text);
 		/* max-width: 24em; */


nimatrengo:
> This was added in #3717 but I don't know why

Not sure too, but it blocks users to select boxes. negative margin will cause impediments to users at any time.

gtm-nayan:
Last time something like this was there it was a workaround for the headings being covered by the navbar when hash links were used. It was fixed a different way for the docs but the faq page might've been left out.

nimatrengo:
> 

So from your perspective, is it a fix now? rather was it a workaround?

gtm-nayan:
Looks like the workaround isn't needed anymore, nothing seems to be breaking with your change. Thanks for the fix.

---
There is another style issue with the FAQ page's code blocks. One liner so probably could include it in this PR and just rename it.

Could you add `import '@sveltejs/site-kit/code.css';` to the FAQ page?

nimatrengo:
As I am not familiar with "Kit" and Vercel workflow, not sure why tests have been failed. Vercel - kit-svelte-dev git action is failed too. ⁉️

nimatrengo:
> Don't worry about the actions, although you have put the import in the wrong file, it should be in routes/faq/index.svelte instead of routes/index.svelte

Good catch, fixed. ✌️

gtm-nayan (approved):

Rich-Harris (approved):

Rich-Harris:
thanks!

PR #5221, opened 2022-06-18
create-svelte: Add descriptions to select options

### LINKED ISSUE #5217: Add description to options when initializing new project
### Describe the problem

I don't have a lot of knowledge of javascript.

The "get started" tells you to `npm init svelte my-app`

You get this question:
image

I have no idea what I want.
I choose the `skeleton project` because ??

Another question appears:
image

I think this might be a trick question..

Typescript is Javascript with type-checking right?🤔

I choose `Typescript`.

The rest of the questions is easier because it tells you what is is.
image

I would like to have two things:

1. Some information with these options. What do I get when I choose the option? Some information with a link to more information would be nice.
2. A recommended option. You (the svelte people) probably know what the best option is right? I would expect for the type checking this would be possible.

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

gtm-nayan:
closes #5217

### Tests

### Changesets

Here's how the descriptions look
image

### DIFF
--- a/.changeset/thick-swans-join.md
+++ b/.changeset/thick-swans-join.md
@@ -2,4 +2,4 @@
 'create-svelte': patch
 ---
 
-Add descriptions to select options
+Add descriptions to templates, and make TypeScript options more self-explanatory
--- a/.changeset/thick-swans-join.md
+++ b/.changeset/thick-swans-join.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Add descriptions to select options
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -73,7 +73,7 @@ async function main() {
 				{
 					type: 'select',
 					name: 'types',
-					message: 'Add type checking?',
+					message: 'Add type checking with TypeScript?',
 					initial: false,
 					choices: [
 						{
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -77,7 +77,7 @@ async function main() {
 					initial: false,
 					choices: [
 						{
-							title: 'With JavaScript comments (JSDoc)',
+							title: 'Yes, using JavaScript with JSDoc comments',
 							value: 'checkjs'
 						},
 						{
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -61,10 +61,11 @@ async function main() {
 					message: 'Which Svelte app template?',
 					choices: fs.readdirSync(dist('templates')).map((dir) => {
 						const meta_file = dist(`templates/${dir}/meta.json`);
-						const meta = JSON.parse(fs.readFileSync(meta_file, 'utf8'));
+						const { title, description } = JSON.parse(fs.readFileSync(meta_file, 'utf8'));
 
 						return {
-							title: meta.description,
+							title,
+							description,
 							value: dir
 						};
 					})
@@ -75,8 +76,16 @@ async function main() {
 					message: 'Add type checking?',
 					initial: false,
 					choices: [
-						{ title: 'Type-checked JavaScript', value: 'checkjs' },
-						{ title: 'TypeScript', value: 'typescript' },
+						{
+							title: 'JavaScript with checkJs',
+							value: 'checkjs',
+							description: 'using JavaScript comments for types.'
+						},
+						{
+							title: 'TypeScript',
+							value: 'typescript',
+							description: 'using TypeScript annotations for types.'
+						},
 						{ title: 'None', value: null }
 					]
 				},
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -84,7 +84,7 @@ async function main() {
 							title: 'With TypeScript annotations',
 							value: 'typescript'
 						},
-						{ title: 'None', value: null }
+						{ title: 'No', value: null }
 					]
 				},
 				{
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -77,14 +77,12 @@ async function main() {
 					initial: false,
 					choices: [
 						{
-							title: 'JavaScript with checkJs',
+							title: 'With JavaScript comments (JSDoc)',
 							value: 'checkjs',
-							description: 'using JavaScript comments for types.'
 						},
 						{
-							title: 'TypeScript',
+							title: 'With TypeScript annotations',
 							value: 'typescript',
-							description: 'using TypeScript annotations for types.'
 						},
 						{ title: 'None', value: null }
 					]
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -78,11 +78,11 @@ async function main() {
 					choices: [
 						{
 							title: 'With JavaScript comments (JSDoc)',
-							value: 'checkjs',
+							value: 'checkjs'
 						},
 						{
 							title: 'With TypeScript annotations',
-							value: 'typescript',
+							value: 'typescript'
 						},
 						{ title: 'None', value: null }
 					]
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -81,7 +81,7 @@ async function main() {
 							value: 'checkjs'
 						},
 						{
-							title: 'With TypeScript annotations',
+							title: 'Yes, using TypeScript syntax',
 							value: 'typescript'
 						},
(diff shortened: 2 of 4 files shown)


packages/create-svelte/bin.js
```diff
@@ -75,8 +76,16 @@ async function main() {
 					message: 'Add type checking?',
 					initial: false,
 					choices: [
-						{ title: 'Type-checked JavaScript', value: 'checkjs' },
-						{ title: 'TypeScript', value: 'typescript' },
+						{
+							title: 'JavaScript with checkJs',
```
Mlocik97, line 80: I prefer `Type-checked JavaScript` as title.

packages/create-svelte/bin.js
```diff
@@ -75,8 +76,16 @@ async function main() {
...
+							value: 'checkjs',
+							description: 'using JavaScript comments for types.'
+						},
+						{
+							title: 'TypeScript',
+							value: 'typescript',
+							description: 'using TypeScript annotations for types.'
+						},
 						{ title: 'None', value: null }
```
Mlocik97, line 89: maybe add description here too? For `None` description can be `Standard JS without any type checking` or something like that.

packages/create-svelte/bin.js
gtm-nayan: Tbh, I'm not all that happy with either, as the author of the issue pointed out "Isn't typescript supposed to be type checked javascript", and I agree with that. And 'Javascript with checkJs' probably just confuses people who don't know what checkJs is even more.

Lemme try something else.

packages/create-svelte/bin.js
gtm-nayan: None is fairly self-descriptive, don't think we need to add verbiage to it

packages/create-svelte/bin.js
Mlocik97: yes, `checkJS` is really confusing for beginners... actually I think TypeScript is TypeScript, another form of type checked JavaScript, but type checked JavaScript is also JS with JSDocs... so I think name `Type-Checked JavaScript` is fine for me.

packages/create-svelte/bin.js
schelv: I agree with @Mlocik97 here. 
It may seem a bit redundant, but it makes it easier to compare the options, and it confirms what the user is thinking.

packages/create-svelte/bin.js
schelv: I'm fine with the titles that are confusing. As long as the description is deconfusing.  

The description for Typescript could be:
 `A superset of JavaScript that adds optional static typing to the language. (for more info see https://www.typescriptlang.org)`

For the `Type-Checked JavaScript` option I'm still not sure what it means haha.
Is this the checkJS you are talking about? https://www.typescriptlang.org/tsconfig#checkJs

I've also found [this page] (https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html) which also describes other forms of js type checking besides the JSDoc. 
Am I correct in concluding that the JSDoc-based-type-checks are only a subset of the type checking done?
Maybe the description can refer to that page.

But having read all this information, the best of these option for a new project seems to be Typescript, right?
What is the situations where you want to have the js with typechecking?

gtm-nayan: > Is this the checkJS you are talking about? [typescriptlang.org/tsconfig#checkJs](https://www.typescriptlang.org/tsconfig#checkJs)

Yes it was, although now I've omitted the descriptions so the preview above is outdated.

> The description for Typescript could be:
 A superset of JavaScript that adds optional static typing to the language. (for more info see https://www.typescriptlang.org/)

That might be a bit too lengthy. If someone doesn't know then it's a quick search away, it's a different language so that'll provide a lot more information than we ever could in a CLI prompt.

> What is the situations where you want to have the js with typechecking?
If that situtation almost never happens wouldn't it be better to have typescript as the recommended option?

TypeScript is more syntax to learn and not the friendliest at times. It's also an additional build step so many devs prefer to stick with Javascript as long as JSDoc provides the same type-checking benefits.

> For the Type-Checked JavaScript option I'm still not sure what it means haha.

Quite a few people have gotten confused by it over on discord, and I've been meaning to try and make it clearer. Thank you for bringing it up.

I'm gonna resolve this so that people don't get confused by the outdated code snippet but if you have any further suggestions you can put them in the main thread. 🙂

packages/create-svelte/bin.js
```diff
@@ -75,8 +76,14 @@ async function main() {
 					message: 'Add type checking?',
 					initial: false,
 					choices: [
-						{ title: 'Type-checked JavaScript', value: 'checkjs' },
-						{ title: 'TypeScript', value: 'typescript' },
+						{
+							title: 'With JavaScript comments (JSDoc)',
```
benmccann, line 80: might need a bit more explanation of what this is. but hopefully this isn't too long
```suggestion
							title: 'Runs TypeScript's type checking against JavaScript comments (JSDoc)',
```

gtm-nayan: Guess it could be put into the descriptions, how's this? 

image
image

benmccann: I agree it looks nicer on the same line

gtm-nayan: That's coming from how the prompts library handles descriptions, if they're too long to fit on one line they start from the next, so it has to be put in the title

image

Mlocik97:
```
Type-Checked JavaScript - check for types in JS with JSDoc.
TypeScript - check for types with TS.
None - Don't check for types.

Comment (Mlocik97) [reply packages/create-svelte/bin.js:80]: Or how it was in https://github.com/sveltejs/kit/pull/5221#discussion_r902139470 but shortcut JS and TS, and maybe replace last part with `and run TSC to check types.`

Comment (gtm-nayan) [reply packages/create-svelte/bin.js:80]: Turns out we can cheat a bit with kleur to make parts of the title look like the description

image

Ended up with this, which I think looks nice and readable.

File: packages/create-svelte/bin.js [hunk 4]
```diff
@@ -75,8 +76,14 @@ async function main() {
 					message: 'Add type checking?',
```
Rich-Harris, line 76:
```suggestion
					message: 'Use TypeScript?',
```

packages/create-svelte/bin.js
Rich-Harris, line 80:
```suggestion
							title: 'Yes, using JSDoc comments',
```

packages/create-svelte/bin.js
```diff
@@ -75,8 +76,14 @@ async function main() {
...
 					initial: false,
 					choices: [
-						{ title: 'Type-checked JavaScript', value: 'checkjs' },
-						{ title: 'TypeScript', value: 'typescript' },
+						{
+							title: 'With JavaScript comments (JSDoc)',
+							value: 'checkjs'
+						},
+						{
+							title: 'With TypeScript annotations',
```
Rich-Harris, line 84:
```suggestion
							title: 'Yes, using TypeScript syntax',
```

packages/create-svelte/bin.js
babichjacob: I personally liked the question being asking if someone wants type checking, because the two answers are different ways to fulfill that. To me, it's a little weird to say that you want TypeScript using TypeScript syntax

gtm-nayan:
"Use TypeScript?" by itself would likely be interpreted as the TypeScript language itself.

```
Add TypeScript's type checking?
> Yes, using JavaScript with JSDoc comments
  Yes, using TypeScript syntax
  No
```

packages/create-svelte/bin.js
Rich-Harris:
```suggestion
							title: 'Yes, using JavaScript with JSDoc comments',
```

Rich-Harris:
I tweaked @gtm-nayan's suggestion in https://github.com/sveltejs/kit/pull/5221#issuecomment-1162495411 very slightly and committed it, so that the diff is once again legible. I think we're in a good place so I'll approve the PR, but will leave it open in case we want to bikeshed further

.changeset/thick-swans-join.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Add descriptions to select options
```
Rich-Harris, line 5:
```suggestion
Add descriptions to templates, and make TypeScript options more self-explanatory
```

Rich-Harris (approved):

dummdidumm (approved):

PR #5258, opened 2022-06-24
Generate sourcemaps

benmccann:
Vercel was done here: https://github.com/sveltejs/kit/pull/5197

We should do the others as well

### DIFF
--- a/.changeset/slimy-carpets-smash.md
+++ b/.changeset/slimy-carpets-smash.md
@@ -4,4 +4,4 @@
 '@sveltejs/adapter-netlify': patch
 ---
 
-Generate sourcemaps
+Generate sourcemaps for server-side functions when bundling with esbuild
--- a/.changeset/slimy-carpets-smash.md
+++ b/.changeset/slimy-carpets-smash.md
@@ -1,0 +1,7 @@
+---
+'@sveltejs/adapter-cloudflare': patch
+'@sveltejs/adapter-cloudflare-workers': patch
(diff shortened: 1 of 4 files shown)


Conduitry:
Can we make this configurable? (It should probably be configurable for Vercel as well.) Not everyone wants to expose their entire human-readable frontend codebase to the world, comments and all.

benmccann:
I believe this is only being done for the server-side, but am not super familiar with the Cloudflare adapters, so feel free to correct me

Rich-Harris:
correct, this is purely about making it possible for server-side logs to have correct stack traces. i've no idea if netlify/cloudflare etc can actually make use of them, but if so then this seems like a good change. will try and check later

benmccann:
Even if they don't use them yet, I'm not sure it'd be a bad idea to add them as it reduces the number of barriers to getting it working

Rich-Harris (approved):

INTERNAL-ERROR:
@Rich-Harris @benmccann 

Hey, I wrote issue [#5337](https://github.com/sveltejs/kit/issues/5337). It was resolved by downgrading to version 65 of netlify's svelte adapter, so something changed and broke between those versions. Just wanted to let you know, and please give me feedback on my conduct as this is my first open source issue.
Thanks!

benmccann:
You issue is unrelated to this change. The real cause is something else: https://github.com/sveltejs/kit/issues/5337#issuecomment-1176545668

PR #5263, opened 2022-06-24
Use ESM eslint config files

benmccann:
I'm not sure why this uses `.cjs` files. Maybe a holdover from an old version of eslint

### DIFF
--- a/.changeset/lovely-foxes-visit.md
+++ b/.changeset/lovely-foxes-visit.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Use ESM eslint config files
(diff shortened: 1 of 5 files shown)


avarun42:
does this actually work? making this change in my own SvelteKit project results in the error:
```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/varun/code/application/.eslintrc.js from /Users/varun/code/application/node_modules/@eslint/eslintrc/dist/eslintrc.cjs not supported.
Instead change the require of .eslintrc.js in /Users/varun/code/application/node_modules/@eslint/eslintrc/dist/eslintrc.cjs to a dynamic import() which is available in all CommonJS modules.
```

The [ESLint docs](https://eslint.org/docs/latest/user-guide/configuring/configuration-files#configuration-file-formats) also state:
> JavaScript (ESM) - use .eslintrc.cjs when running ESLint in JavaScript packages that specify "type":"module" in their package.json. Note that ESLint does not support ESM configuration at this time.

Issue #5269, opened 2022-06-25
Include typescript-svelte-plugin when using create-svelte to generate a new, type-checked project


theschmocker:
### Describe the problem

TypeScript support is great out-of-the box with VSCode, but requires some extra steps to get working properly with other editors' LSP clients (neovim's built-in LSP + [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig), in my use-case). I had to go looking in [sveltejs/language-tools](https://github.com/sveltejs/language-tools) to figure out that I needed to install typescript-svelte-plugin and enable it in my `tsconfig.json`.

### Describe the proposed solution

- typescript-svelte-plugin installed as a dev dependency by default in TypeScript and type-checked JS templates
- enabled the plugin in the generated `.svelte-kit/tsconfig.json`

### Alternatives considered

- Note the steps required to enable full LSP support for SvelteKit projects in the documentation
- Create an adder for https://github.com/svelte-add/svelte-add

### Importance

would make my life easier

### Additional Information

Pull request: #5270

theschmocker:
https://github.com/neovim/nvim-lspconfig/pull/1972

PR #5270, opened 2022-06-25
includes typescript-svelte-plugin in templates and generated tsconfig

theschmocker:

### Tests

### Changesets

Includes `typescript-svelte-plugin` in type-checked (TS, checked JS) kit templates and enables it the generated `.svelte-kit/tsconfig.json`

Related issue: #5269

### DIFF
--- a/.changeset/three-rivers-approve.md
+++ b/.changeset/three-rivers-approve.md
@@ -1,0 +1,6 @@
+---
+'create-svelte': patch
+'@sveltejs/kit': patch
+---
+
+includes typescript-svelte-plugin for out-of-the-box, non-vscode lsp client support
--- a/packages/create-svelte/shared/+checkjs/package.json
+++ b/packages/create-svelte/shared/+checkjs/package.json
@@ -5,6 +5,7 @@
 	},
 	"devDependencies": {
 		"typescript": "^4.7.2",
-		"svelte-check": "^2.7.1"
+		"svelte-check": "^2.7.1",
+		"typescript-svelte-plugin": "^0.3.3"
 	}
 }
--- a/packages/create-svelte/shared/+typescript/package.json
+++ b/packages/create-svelte/shared/+typescript/package.json
@@ -7,6 +7,7 @@
 		"typescript": "^4.7.2",
 		"tslib": "^2.3.1",
 		"svelte-check": "^2.7.1",
-		"svelte-preprocess": "^4.10.6"
+		"svelte-preprocess": "^4.10.6",
+		"typescript-svelte-plugin": "^0.3.3"
 	}
 }
--- a/packages/kit/src/core/sync/write_tsconfig.js
+++ b/packages/kit/src/core/sync/write_tsconfig.js
@@ -75,7 +75,10 @@ export function write_tsconfig(config, cwd = process.cwd()) {
 					lib: ['esnext', 'DOM'],
 					moduleResolution: 'node',
 					module: 'esnext',
-					target: 'esnext'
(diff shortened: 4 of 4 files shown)


benmccann:
I'll leave this for a bit to give the other maintainers a chance to see it, but I don't feel that we should be adding neovim support for all users

theschmocker:
> I'll leave this for a bit to give the other maintainers a chance to see it, but I don't feel that we should be adding neovim support for all users

To clarify, neovim's LSP client is my specific use case, but this should be editor-agnostic. typescript-plugin-svelte (from sveltejs/language-tools) hooks into TypeScript itself to provide intellisense for Svelte files and seems like a core part of the TypeScript+Svelte tooling. The official VSCode plugin [includes it implicitly](https://github.com/sveltejs/language-tools/tree/master/packages/typescript-plugin#usage). This would make it more explicit so that other tooling that leverages TypeScript (like language servers and their clients; not just neovim) can use it without extra work by the tool author or project modification by SvelteKit users

benmccann:
As a VS Code user, this seems bizarre to me. If it's part of the VS Code plugin I don't believe there's any reason I need to separately install it in my project. We should try to simplify things for SvelteKit users rather than editor plugin devs.

Maybe @dummdidumm would know more about what you're trying to accomplish than I do though

dummdidumm:
I too feel like this belongs to the setup docs of the IDE extension you are using, and `svelte-add` seems like a nice place to put the additional code that needs to be added for other IDEs than VS Code. Maybe we could also note it in a message when you chose the type check option, but not sure.

theschmocker:
It was just unclear to me as a user that although TypeScript language server and the Svelte language server work perfectly on their own, getting them to work together to provide a full framework experience requires that either a) there's a specific plugin for my editor which augments the existing LSP support or b) I add extra configuration to each project. LSP is great _not only_ because it makes life easier for plugin authors, but also because it makes life easier for users like myself. I don't need to worry about whether there's an actively-maintained, editor-specific plugin for the language/framework I'm using; I can get similar tools in any editor that supports LSP (emacs with lsp-mode comes to mind as another example).

I won't die on this hill, though. VSCode is great and popular, so it's completely understandable that effort is focused there. typescript-plugin-svelte looks to be included behind the scenes as an official part of the VSCode plugin, so my assumption was that adding typescript-svelte-plugin to SvelteKit and its templates would be an easy win for DX _outside_ of VSCode. If that's not the case and/or there are reasons not to do this that I have not considered, then I apologize for the unprompted PR and the time taken--I can look elsewhere to improve documentation instead

dummdidumm:
Nothing to apologize for!
`typescript-plugin-svelte` isn't a LSP, it's a plugin which typescript will load on startup in IDE contexts (not when using tsc). We need this because we have to tell TS how to deal with Svelte files in the context of JS/TS files, which the Svelte language service has no control over.
Adding it to the tsconfig means that people are very likely to not update it often, which means they miss out on bug fixes we might do later on. Optimizing for VS Code users who get these updates without worrying about it feels better to me since I assume this is the majority of devs. I agree that documentation about this should be better. Probably the best place for this is where the Svelte LSP integration is explained/implemented for the specific IDE. Once we overhaul the Svelte docs I also want to add a tooling section where this is mentioned.
Therefore closing this, thank you!

theschmocker:
> Adding it to the tsconfig means that people are very likely to not update it often

Ah, that makes sense. Thank you for elaborating. I've opened a PR to add a note to the Svelte docs over in https://github.com/neovim/nvim-lspconfig/pull/1972

PR #5280, opened 2022-06-26
[docs] enhance docs on adding types to app.d.ts

dummdidumm:
People still get confused by this too often, I hope this helps resolving their issues around imports quicker.

### Tests

### Changesets

### DIFF
--- a/packages/kit/types/ambient.d.ts
+++ b/packages/kit/types/ambient.d.ts
@@ -17,13 +17,29 @@
  *
  * By populating these interfaces, you will gain type safety when using `event.locals`, `event.platform`, `session` and `stuff`.
  *
- * Note that since it's an ambient declaration file, you can't use `import` statements — instead, use the `import(...)` function:
+ * Note that since it's an ambient declaration file, you have to be careful when using `import` statements. Once you add an `import`
+ * at the top level, the declaration file is no longer considered ambient and you loose access to these typings in other files.
+ * To avoid this, either use the `import(...)` function:
  *
  * ```ts
  * interface Locals {
  * 	user: import('$lib/types').User;
  * }
  * ```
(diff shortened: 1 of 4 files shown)


packages/kit/types/ambient.d.ts
```diff
@@ -17,13 +17,29 @@
  *
  * By populating these interfaces, you will gain type safety when using `event.locals`, `event.platform`, `session` and `stuff`.
  *
- * Note that since it's an ambient declaration file, you can't use `import` statements — instead, use the `import(...)` function:
+ * Note that since it's an ambient declaration file, you have to be careful when using `import` statements. Once you add an `import`
+ * at the top level, the declaration file is no longer considered ambient and you loose access to these typings in other files.
```
gtm-nayan, line 21:
```suggestion
 * at the top level, the declaration file is no longer considered ambient and you lose access to these typings in other files.
```

benmccann (approved):

chientrm (approved):

packages/kit/types/ambient.d.ts
```diff
@@ -17,13 +17,29 @@
...
+ * declare global {
+ * 	namespace App {
+ * 		interface Locals {
+ * 			user: User;
+ * 		}
+ * 		// ...
+ * 	}
+ * }
+ * ```
+ *
```
avarun42, line 42: Are there drawbacks to wrapping the namespace with `declare global`? If not, wouldn't it make sense to just do this by default in `create-svelte`? And if there are, then it probably shouldn't be recommended without a caveat.

dummdidumm: Problem is that it's a type error if there are _no_ imports, so unfortunately we can't add it by default

PR #5290, opened 2022-06-28
Revert "Undici"

Rich-Harris:
Gah — we might need to revert #5117. `undici` doesn't implement `multipart/form-data` parsing, unlike `node-fetch`. Since that's the default encoding, this means that forms are basically broken as a result of the switch. We just didn't have any tests covering it.

Related issue: https://github.com/nodejs/undici/issues/974


Rich-Harris:
closing in favour of #5292

PR #5291, opened 2022-06-28
stream request bodies

### LINKED ISSUE #3419: Handling streaming request/response bodies

Rich-Harris:
Closes #3419. Instead of buffering request bodies in Node, we can now turn them into `ReadableStream` objects.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/node/index.js
+++ b/packages/kit/src/node/index.js
@@ -2,51 +2,41 @@ import * as set_cookie_parser from 'set-cookie-parser';
 
 /** @param {import('http').IncomingMessage} req */
 function get_raw_body(req) {
-	return new Promise((fulfil, reject) => {
-		const h = req.headers;
+	const h = req.headers;
 
-		if (!h['content-type']) {
-			return fulfil(null);
-		}
+	if (!h['content-type']) {
+		return null;
+	}
 
-		req.on('error', reject);
+	const length = Number(h['content-length']);
 
-		const length = Number(h['content-length']);
+	// https://github.com/jshttp/type-is/blob/c1f4388c71c8a01f79934e68f630ca4a15fffcd6/index.js#L81-L95
+	if (isNaN(length) && h['transfer-encoding'] == null) {
+		return null;
+	}
 
-		// https://github.com/jshttp/type-is/blob/c1f4388c71c8a01f79934e68f630ca4a15fffcd6/index.js#L81-L95
-		if (isNaN(length) && h['transfer-encoding'] == null) {
-			return fulfil(null);
-		}
+	return new ReadableStream({
+		start(controller) {
+			req.on('error', (error) => {
+				controller.error(error);
+			});
 
-		let data = new Uint8Array(length || 0);
+			let size = 0;
 
-		if (length > 0) {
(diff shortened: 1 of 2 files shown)


packages/kit/src/node/index.js
```diff
@@ -2,51 +2,41 @@ import * as set_cookie_parser from 'set-cookie-parser';
...
-		}
+	if (!h['content-type']) {
+		return null;
+	}
 
-		req.on('error', reject);
+	const length = Number(h['content-length']);
 
-		const length = Number(h['content-length']);
+	// https://github.com/jshttp/type-is/blob/c1f4388c71c8a01f79934e68f630ca4a15fffcd6/index.js#L81-L95
```
benmccann, line 13: we could expand the comment to make it clearer what this does without needing to follow the link
```suggestion
	// check if no request body
	// https://github.com/jshttp/type-is/blob/c1f4388c71c8a01f79934e68f630ca4a15fffcd6/index.js#L81-L95
```

benmccann (approved):

PR #5292, opened 2022-06-28
monkey-patch formData to enable multipart parsing

Rich-Harris:
This is an alternative to reverting #5117 (#5290). It's ugly, but we can use the multipart form data parsing in `node-fetch` by monkey-patching the `undici` `Request` object's `formData` method to construct a `node-fetch` `Request` with the raw data and call `formData` on _that_. 

Hideous, but it works. If https://github.com/nodejs/undici/issues/974 is resolved, we can remove this hack.

Marking as draft for now as it should really have a test.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/node/index.js
+++ b/packages/kit/src/node/index.js
@@ -1,5 +1,6 @@
 import * as set_cookie_parser from 'set-cookie-parser';
 import { Request as NodeFetchRequest } from 'node-fetch';
+import { Readable } from 'stream';
 
 /** @param {import('http').IncomingMessage} req */
 function get_raw_body(req) {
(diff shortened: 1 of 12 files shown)


benmccann:
Hmm. Looks like this might only work on Node 18 for some reason. Maybe the polyfill isn't working on Node 16 as expected?

packages/kit/src/node/index.js
```diff
@@ -64,11 +65,21 @@ export async function getRequest(base, req) {
...
 		method: req.method,
 		headers,
 		body: await get_raw_body(req) // TODO stream rather than buffer
 	});
+
+	request.formData = async () => {
+		return new NodeFetchRequest(request.url, {
+			method: request.method,
+			headers: request.headers,
+			body: Buffer.from(await request.arrayBuffer())
```
jimmywarting, line 78: you should rather do something like `body: stream.Readable.from(request.body)` since undici uses web streams and node-fetch kind of expect a node stream

you also have https://nodejs.org/dist/latest-v18.x/docs/api/stream.html#streamreadablefromwebreadablestream-options

Rich-Harris:
Very weird that the `allows return value to be a Response` test fails in CI. It works locally using the same Node version

Rich-Harris:
Ok, so it was failing in CI because for some reason I don't fully understand some polyfills were being applied and some weren't, which led to an `instanceof` check failing. I think the best solution is to always apply all polyfills consistently

benmccann (approved):

Issue #5238, opened 2022-06-21
Type info tooltips look wacky when they're empty


Rich-Harris:
### Describe the bug

Not sure exactly what's going on here, but this looks daft:

<img width="716" alt="image" src="https://user-images.githubusercontent.com/1162160/174894929-609b22f9-dba2-470b-bd02-b24505a986ec.png">

### Reproduction

Visit https://kit.svelte.dev/faq#read-package-json

### System Info

```shell
N/A
```

### Severity

annoyance

PR #5304, opened 2022-06-28
fix: Type info tooltips look wacky when they're empty

PrashantSakre:

### Tests

### Changesets

Reference: #5238

* Margin is removed since not required.
* Attached image for reference.

Screenshot 2022-06-28 at 22 45 49


Rich-Harris:
Thank you, but the problem isn't that the empty tooltip is off-kilter, it's the fact that it gets rendered at all when there's no content.

The margin _is_ necessary btw — it's the difference between these:

image

image

Issue #4802, opened 2022-05-03
Failure in page endpoint results in incorrect message when navigated to client-side, but not when hard reloaded


tcc-sejohnson:
### Describe the bug

The error passed to the `load` function of `__error.svelte` is a generic `500: Failed to load data` when navigating client-side to a page with a page endpoint. However, refreshing the page or visiting it directly (causing it to be fetched directly from the server) results in the correct error. 

### Reproduction

1. Go here: https://github.com/tcc-sejohnson/kit-exception-bug-rero (yes, there's a typo...)
2. Read the README
3. See the error

### Logs

```shell
You asked for it.

Oh noooooo
Oh noooooo
Oh noooooo
Oh noooooo
Oh noooooo
Oh noooooo
```

### System Info

```shell
System:
    OS: Linux 5.16 Pop!_OS 22.04 LTS
    CPU: (12) x64 AMD Ryzen 5 5600X 6-Core Processor
    Memory: 26.11 GB / 31.32 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 18.0.0 - ~/.nvm/versions/node/v18.0.0/bin/node
    Yarn: 1.22.17 - ~/.yarn/bin/yarn
    npm: 8.6.0 - ~/.nvm/versions/node/v18.0.0/bin/npm
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.40 
    @sveltejs/kit: next => 1.0.0-next.324 
    svelte: ^3.44.0 => 3.48.0
```

### Severity

annoyance

### Additional Information

Seems like the error code is always correct at least... :)

cdcarson:
As of a month or two ago this issue was being tracked in this PR https://github.com/sveltejs/kit/pull/4032.

I described what I thought was a [decent fix](https://github.com/sveltejs/kit/pull/4032#issuecomment-1086906307), but I haven't heard back.

tcc-sejohnson:
Interesting! That PR appears to deal with some of the confusing bits brought up by #4801 as well. Error handling is one of the strangely-furry parts of Kit right now, unfortunately. 🙂

georgecrawford:
My PR is stale because I asked for help and advice for the maintainers. I didn’t want to pester but I’m stuck on determining the right way forward.

tcc-sejohnson:
George, have you tried asking for help in the Discord #contributing channel? That can be a good place to ask for help without feeling too "pester"-y. :)

[...]

PR #5314, opened 2022-06-29
Error handling fix

cdcarson:
This fixes https://github.com/sveltejs/kit/issues/4802, which was previously attempted in https://github.com/sveltejs/kit/pull/4032. 

Currently, when an error is thrown from an endpoint, the server ignores whether the request was the result of a `fetch` call from the client side. There are two cases where this matters: (1) a get request from the client router and (2) a `fetch` request from userland client code. 
- In all cases, the server returns HTML from `__error.svelte` rather than the error as JSON.
- In the case of client-side navigation, `__error.svelte` ends up getting hydrated with a default error rather than the thrown error. 
- In the case of a userland fetch, the HTML response is useless.

Two changes, as light-handed as possible:

1. In [packages/kit/src/runtime/server/index.js](https://github.com/sveltejs/kit/compare/master...cdcarson:error-handling-fix?expand=1#diff-70bbf390c5a53b49a9f23557426bbd1bee9188a247ce1ef34e5f4f217036c7f6) we check for the two cases above, and if applicable, return the error as JSON.  **NB:** Rather than worrying about serializing potentially complex or sensitive errors, it just returns  an `Error`-like object `{message: string}`. IMO, this will take care of most userland use cases (where you just want to show the user what went wrong.) It doesn't preclude folks from throwing more complex error structures as JSON.

2. Additionally, a minor change was necessary in [packages/kit/src/runtime/client/client.js](https://github.com/sveltejs/kit/compare/master...cdcarson:error-handling-fix?expand=1#diff-6bc786c5d82299b5eed36bb073cd7e3c8347af2951019da2b5dff1d5e3bad611).

There are three addition tests that cover off on the scenarios mentioned above.  All the existing relevant tests seem to be passing, but I'm not sure how much coverage there was to begin with. 
 
### Tests

### Changesets


Rich-Harris:
Alright, I got completely carried away with this PR, but I think it's done now. Thank you @cdcarson for getting the ball rolling, and thank you @georgecrawford for #4032 — apologies, I totally dropped the ball on that PR due to other things getting in the way.

In addition to turning _thrown_ errors into JSON responses, this PR now does the same thing for explicitly returned errors. If the `body` of a request handler from a page endpoint is an `Error`, the error page will be shown (rather than validation errors being mixed in with `get` properties), as in this example adapted from #4032:

```js
// post handler can return validation errors
export function post({ request, locals }) {
  if (!locals.user) {
    // fatal error — results in error page
    return {
      status: 401,
      body: new Error('show yourself, coward')
    };
  }

  if (!locals.user.mojo < 50) {
    // fatal error — results in error page
    return {
      status: 403,
      body: new Error('insufficient mojo')
    };
  }

  const [errors, item] = await create_item(await request.formData());

  if (errors) {
    // validation errors — mixed with GET props and rendered
    return {
      status: 400,
      body: { errors }
    };
  }

  // great success
  return {
    body: { item }
  };
}
```

Previously, the only way to trigger the error page in the first two cases (the 401 and 403) would have been to throw the error, which has two problems:

* the `status` will always be 500
* thrown errors go through `handleError` (i.e. they show up in logs etc) which probably isn't what you want

Turning errors into JSON involves being slightly opinionated. The `name`, `message` and `stack` properties are preserved (`stack` is neutered in production), unknown-but-enumerable properties (things like `code` or `frame`) are preserved, and if there's a `cause` property the process recurses. I didn't copy over the cyclical structure handling from #4032 as I think it's probably overkill — we don't do that for normal JSON responses, and I think it's a niche enough situation not to bother with — developers know how to debug stack overflows.

Since all this requires documentation, and since it was a bit awkward to fit into the existing docs, I overhauled the relevant section.

packages/kit/src/utils/http.js
```diff
@@ -19,3 +19,55 @@ export function to_headers(object) {
 
 	return headers;
 }
+
+/**
+ * @param {string} accept
```
benmccann, line 24: might not hurt to add descriptions for these two. I guess it's the accept header and types that we're okay using or something?

packages/kit/src/runtime/server/utils.js
```diff
@@ -55,3 +55,40 @@ export function normalize_request_method(event) {
...
+export function serialize_error(error, get_stack) {
+	return JSON.stringify(clone_error(error, get_stack));
+}
+
+/**
+ * @param {Error} error
+ * @param {(error: Error) => string | undefined} get_stack
+ */
+function clone_error(error, get_stack) {
+	const {
```
benmccann, line 72: I wonder if this will drop the error `frame` that is used to show where in the source code the error has occurred when there's a compilation error in dev mode

packages/kit/src/runtime/server/utils.js
```diff
@@ -55,3 +55,40 @@ export function normalize_request_method(event) {
...
+		name,
+		message,
+		// this should constitute 'using' a var, since it affects `custom`
+		// eslint-disable-next-line
+		stack,
+		// @ts-expect-error i guess typescript doesn't know about error.cause yet
+		cause,
+		...custom
```
benmccann, line 80: what is `error.custom`?

Rich-Harris: it's [rest syntax](https://v8.dev/features/object-rest-spread)

packages/kit/src/runtime/server/utils.js
Rich-Harris: not if it's enumerable

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -481,6 +493,17 @@ async function load_shadow_data(route, event, options, prerender) {
 			add_cookies(/** @type {string[]} */ (data.cookies), headers);
 			data.status = status;
 
+			if (body instanceof Error) {
```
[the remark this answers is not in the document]
Rich-Harris: feels like unnecessary indirection to me — you add function call overhead, a new utility somewhere in the codebase, an import declaration per consuming module... you shave off a few characters at the callsite but make the codebase larger and arguably less easy to read

packages/kit/src/runtime/server/utils.js
```diff
@@ -55,3 +55,40 @@ export function normalize_request_method(event) {
 	const method = event.request.method.toLowerCase();
 	return method === 'delete' ? 'del' : method; // 'delete' is a reserved word
 }
+
+/**
+ * @param {Error} error
+ * @param {(error: Error) => string | undefined} get_stack
+ */
+export function serialize_error(error, get_stack) {
+	return JSON.stringify(clone_error(error, get_stack));
```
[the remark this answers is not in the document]
Rich-Harris: added

packages/kit/src/utils/http.js
Rich-Harris: added

packages/kit/src/runtime/server/page/load_node.js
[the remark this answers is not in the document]
Rich-Harris: this is one of those cases where you basically _can't_ DRY it out, because of the control flow (i.e. mutating an existing object and conditionally returning it). Or rather you can, but the resulting function is larger than the code you saved by deduplicating. i reckon it's probably not worth it

benmccann (approved):

georgecrawford:
Great stuff @Rich-Harris - thanks so much for getting to this, and no worries for the delay. I was going for longest-open-PR 🥇 !

Issue #4938, opened 2022-05-16
Need CI to check template projects


benmccann:
### Describe the bug

I created a new project and `pnpm check` is currently failing with nine errors

### Reproduction

I chose "SvelteKit demo app" and then

1. Typescript
2. ESLint: Yes
3. Prettier: Yes
4. Playwright: Yes

If it makes a difference, I develop with pnpm.   (With pnpm v7, `pnpm init svelte` appears broken though.)

```
cd my-app
pnpm install
pnpm build
pnpm check
```

`pnpm build` is necessary to create `.svelte-kit/tsconfig.json`, which is extended by `tsconfig.json`.

### System Info

```shell
latest
```

### Severity

serious, but I can work around it

delhanty:
Note to future self when searching - also to get email updates on this PR:

The text in this PR was written by me in comment [#4805#issuecomment-1126938274](https://github.com/sveltejs/kit/issues/4805#issuecomment-1126938274), at Ben's request so "I" is actually me rather than Ben.  Also, there is another comment that I wrote that I would like to link now but, irritatingly, I can't because it's been marked as "off-topic".

jhubbardsf:
I just tried to reproduce this bug locally and couldn't. Tried with both `pnpm create` (pnpm v7.1.0) and `npm init` (npm v8.5.1). Selected demo, typescript, eslint, prettier, and playwright. Used pnpm to install, build, and check. 0 errors or warnings.

benmccann:
The errors have been fixed for new projects since this issue was filed. You can create a project from a few days ago and see the failure that was occurring with `npm init svelte@2.0.0-next.135 myapp`

I'm leaving this open since the issue is really about running these checks on the CI to prevent bad releases in the future

jhubbardsf:
Thanks for the clarification!

LucasSegersFabro:
heyy @benmccann 
could you check [this](https://github.com/sveltejs/kit/compare/master...LucasSegersFabro:chore/create-svelte-tests?expand=1) out?  

An approach to testing `create-svelte` on CI.  
we `npm install`, `npm build` and `npm check` a newly created project

upsides:
* should work on current CI (maybe I'd need some pointers to test it out)

downsides: 
* could be slow, as we install/build/check everytime
* assumes CI has npm installed

If it's all good I can work more on it (test different create configs, etc)

Thanks :)

benmccann:
[...]

PR #5322, opened 2022-06-30
test: Add tests for create-svelte templates

andyliuhaowen:
Fixes https://github.com/sveltejs/kit/issues/4938.

### Tests

~~Out of the four test cases that this PR is introducing, one is failing. The default template with JSDoc fails svelte-check.
But fixing that is IMO best suited for another separate PR.~~

Maybe not. Fixes for the errors newly detected by CI are included as a second commit.

### Changesets

### Design Decisions

1. Have `create-svelte/scripts/build-templates.js` export its main function as `build_template`.
**Alternatives considered**: forking a node process, but it feels silly to fork a node given the tests are running within `node` in the first place. That also poses the issue of weird `node` executable names.

2. Have the tests work under a created `.test-tmp` directory. **Alternatives considered**: Other names are considered but I can't imagine how those tests could function without an extra temp dir.


benmccann:
This feels heavyweight compared to what I was thinking. Can we just check the files without creating a project? Or is there something that makes project creation necessary?

andyliuhaowen:
1. The template files need to undergo transformations to be valid files for ts or checkjs. It's valid ts most likely because JSDoc are just strings, but we obviously couldn't use checkjs before performing those transformations. So invoking `build_templates` with a temp directory is necessary.
2. `svelte-check` loads the `svelte.config.js`, which imports `adapter-auto`. Then without `npm i` the config could not be loaded and `svelte-check` does not work.

andyliuhaowen:
Any idea how this process could be improved/made lightweight?

PR #5324, opened 2022-06-30
docs: Update broken links in CONTRIBUTING.md

mattcroat:
I think the previous entry points for the development server and production build moved to [Vite](https://github.com/sveltejs/kit/tree/master/packages/kit/src/vite)?

Let me know if I'm wrong! 😄

### Tests

### Changesets


Rich-Harris (approved):

Rich-Harris:
Yep — there's now a single Vite plugin that handles dev, build _and_ preview, whereas previously those were separate things handled programmatically. `build_server` is just one aspect of building (it's called after the client build has been created), so it makes more sense to just point everything at the plugin entry point, which I just did. Thanks!

PR #5330, opened 2022-06-30
[fix] don't log warning if root is configured

benmccann:
`root` is no longer set since https://github.com/sveltejs/kit/pull/5303. This logs an unnecessary warning when using Storybook

### DIFF
--- a/.changeset/quick-chicken-hide.md
+++ b/.changeset/quick-chicken-hide.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] don't log warning if root is configured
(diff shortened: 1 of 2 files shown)


Rich-Harris:
Can you elaborate? As far as I can tell, setting `root` just makes everything break. What are the situations in which it makes sense to be configured?

benmccann:
Storybook sets root to the expected directory. SvelteKit doesn't check what the value of the setting is and just prints an error if it's configured at all eventhough there's no actual issue. And SvelteKit doesn't actually set the value so the error message is also incorrect in saying that SvelteKit is taking control of it.

Rich-Harris:
What is 'the expected directory'? I'm just very unclear on how it could work. It feels like SvelteKit probably _should_ set the value to the project root and/or print an error if it's set to something else

Rich-Harris:
Ah. So I think what we should do is error if `root` is explicitly set to something _other_ than the cwd

benmccann:
Ok. Updated the PR to do that

Rich-Harris (approved):

Issue #5312, opened 2022-06-29
`svelte-kit preview --https ...` ignores custom cert in `config.kit.vite.[preview|server].https`


alipi:
### Describe the bug

The custom cert config in `config.kit.vite.[preview|server].https` is now ignored by `svelte-kit preview --https ...`.

This patch works, but I don't know if it's the correct fix:
```diff
$ cat @sveltejs+kit+1.0.0-next.355.patch
diff --git a/node_modules/@sveltejs/kit/dist/cli.js b/node_modules/@sveltejs/kit/dist/cli.js
index 8b3547b..b868a2e 100644
--- a/node_modules/@sveltejs/kit/dist/cli.js
+++ b/node_modules/@sveltejs/kit/dist/cli.js
@@ -181,7 +181,7 @@ prog
                        // optional config from command-line flags
                        // these should take precedence, but not print conflict warnings
                        if (host) vite_config.preview.host = host;
-                       if (https) vite_config.preview.https = https;
+                       if (https) vite_config.preview.https = vite_config.preview.https || https;
                        if (port) vite_config.preview.port = port;
 
                        const preview_server = await vite.preview(vite_config);

```

### Reproduction

svelte.config.js:
```
config.kit.vite.preview = {
   https: {
      key: readFileSync(resolve(__dirname, "my-dev.key"), "utf8"),
      cert: readFileSync(resolve(__dirname, "my-dev.crt"), "utf8")
   }
};
```
run:
`$ npx svelte-kit preview --https --host --port 5000`

Some internally created cert is returned instead of the one specified in the config.
```
$ openssl s_client -showcerts -servername localhost -connect localhost:5000 <<< "Q" | openssl x509 -text | grep -iA2 "Validity"
depth=0 CN = example.org, C = US, ST = Virginia, L = Blacksburg, O = Test, OU = Test
verify error:num=18:self-signed certificate
verify return:1
depth=0 CN = example.org, C = US, ST = Virginia, L = Blacksburg, O = Test, OU = Test
verify return:1
DONE
        Validity
            Not Before: Jun 29 10:05:25 2022 GMT
            Not After : Jul 29 10:05:25 2022 GMT
```

### System Info

```shell
$ npx envinfo --system --binaries --browsers --npmPackages "{svelte,@sveltejs/*,vite}"

  System:
    OS: Linux 5.10 Generic 36 (Generic) 36 (Generic)
[...]

Issue #5184, opened 2022-06-08
Bring `@sveltejs/kit/experimental/vite` out of experimental


benmccann:
### Describe the problem

Right now SvelteKit asks users to pass Vite configuration via the `kit.vite` option in `svelte.config.js`. This was originally nice because it means less config files and gives SvelteKit a bit more control. However, it's also non-standard. Tools like Vitest and Storybook look for a `vite.config.js` to read configuration from and it would be nice to support those tools out-of-the-box

### Describe the proposed solution

https://github.com/sveltejs/kit/pull/5094 creates a SvelteKit plugin that can be imported in a `vite.config.js`. If a `vite.config.js` is created, it is expected that Vite config is no longer specified in `svelte.config.js`, but in `vite.config.js`.

This new feature can be used like:

```
// vite.config.js
import { sveltekit } from '@sveltejs/kit/experimental/vite';

/** @type {import('vite').UserConfig} */
const config = {
  plugins: [sveltekit()]
};

export default config;
```

**Cleanup items**

If we decide that we like this approach (and I think we probably will) then it would make sense to make some changes:

* Make `vite.config.js` mandatory
* Add it to project templates
* Get rid of `config.kit.vite`
* Move `@sveltejs/kit/experimental/vite` to `@sveltejs/kit/vite` or `@sveltejs/vite-plugin-sveltekit`
* Update contributing guide with new code locations
* Add vitest option to `create-svelte`
* More controversially, we might want to get rid of `svelte-kit dev/build/preview` in favour of using `vite dev/build/preview`. This would mean that SvelteKit itself would no longer have a dependency on Vite, other than in the plugin; apps themselves would need to have a devDependency on `vite`
  * Consequently, figure out how to handle versions. It's likely that SvelteKit will expect a particular version of Vite, which might mean it needs to become a peerDependency if it's no longer a dependency (though if it's possible to avoid depending on a specific version, that would be ideal)

### Importance

nice to have

### Additional Information

CC people who might be interested in this: @nickbreaton @JessicaSachs @IanVS @brittneypostma 

Please feel free to share with others. I haven't used Storybook, Vitest, Histoire, etc. before so would love any help in testing! And would love tips on things we might want to document for the community to setup these tools. Whether there's something that needs to be improved or everything works, I'd love to hear about it either way!

dummdidumm:
Does this mean that `svelte.config.js` is no longer used/required then? Asking because of the implications for language tools

IanVS:
[...]

PR #5332, opened 2022-06-30
[breaking] graduate `@sveltejs/kit/vite` from experimental and require `vite.config.js`

### LINKED ISSUE #5308: Shared config object is mutated
### Describe the bug

SvelteKit's `vite/index.js` stores the config in `vite_user_config` and passes it into multiple builds. Vite mutates the config causing unexpected values for latter builds. This doesn't seem to be an issue for most users, but does cause issues with `vite-plugin-pwa`

### Reproduction

Update `svelte.config.js` for the site to the one below. Note that you need to comment out SvelteKit's `process.exit` for the Vite plugin to be invoked
```
import * as path from 'path';
import adapter from '@sveltejs/adapter-auto';
import { imagetools } from 'vite-imagetools';

let viteConfig;

const config = {
	kit: {
		adapter: adapter(),

		prerender: {
			default: true,
			entries: ['*', '/content.json']
		},

		vite: {
			plugins: [imagetools(), {
				name: "vite-plugin-pwa",
				enforce: "post",
				apply: "build",
				async configResolved(config) {
					viteConfig = config;
				},
				async closeBundle() {
					console.log(`IS SSR? ${viteConfig.build.ssr}`)
				}
			}],

			resolve: {
				alias: {
					$img: path.resolve('src/images')
				}
			},

			server: {
				fs: {
					strict: false
				}
			}
		}
	}
};

export default config;
```

### Suggested Solution

The config above isn't one that can be cloned. Instead we need to do something like leave `vite-plugin-svelte-kit` in place, but make `writeBundle` a no-op unless it's the client build (or equivalently split the config stuff out into a separate plugin that's used for the server build as well)

### System Info

```shell
`master`
```

### Severity

Prevents https://github.com/userquin/pwa-sveltekit-vite from working

### Additional Information

_No response_

benmccann:
Closes https://github.com/sveltejs/kit/issues/5184
Closes #5308
Closes https://github.com/sveltejs/kit/issues/5335
Closes https://github.com/sveltejs/kit/issues/5312


packages/kit/src/cli.js
```diff
@@ -298,10 +295,7 @@ export async function get_vite_config(svelte_config) {
...
 
-	const { sveltekit } = await import('./vite/index.js');
-
-	// TODO: stop reading Vite config from SvelteKit config or move to CLI
-	const vite_config = await svelte_config.kit.vite();
-	vite_config.plugins = [...(vite_config.plugins || []), ...sveltekit()];
-	return vite_config;
+	throw new Error(
+		'vite.config.js is missing. See an example here: https://github.com/sveltejs/kit/blob/master/packages/create-svelte/templates/skeleton/vite.config.js'
+	);
```
Rich-Harris, line 300: I think we should print a sample config, like we do for tsconfig.json and wrangler.toml etc

```suggestion
	console.error(colors.bold().red('> vite.config.js is required'));
	console.error(
		`
		Sample vite.config.js:

		import { sveltekit } from '@sveltejs/kit/vite';

		/** @type {import('vite').UserConfig} */
		const config = {
			plugins: [sveltekit()]
		};

		export default config;

		`.replace(/^\t{2}/gm, '')
	);

	process.exit(1);
```

packages/kit/src/cli.js
```diff
@@ -153,8 +153,7 @@ prog
 			process.env.NODE_ENV = process.env.NODE_ENV || 'production';
 			process.env.VERBOSE = verbose;
 
-			const svelte_config = await load_config();
-			const vite_config = await get_vite_config(svelte_config);
+			const vite_config = await get_vite_config();
 			await vite.build(vite_config); // TODO when we get rid of config.kit.vite, this can just be vite.build()
```
Rich-Harris, line 157: see the TODO — can we get rid of `get_vite_config` (or rather, replace it with a `check_vite_config_exists` or something) and let Vite worry about its own config loading?

Rich-Harris: (would check but i'm in the back of a car, tethering to a flaky hotspot — not very conducive to work)

Rich-Harris:
It occurs to me that if we make good on our plan to replace `svelte-kit dev/build/preview` with `vite dev/build/preview` we'll be inflicting two breaking changes instead of getting it all out of the way in one go. Is it worth trying to do both here? Feels like we get a bigger benefit of switching to vite.config.js immediately that way.

It would make the PR a bit bigger though — we would need to consider

* [ ] replacing `--verbose` with `--logLevel`
* [ ] removing Vite as a dependency or SvelteKit (it would need to become both a devDependency, so that we can access Vite's types in the codebase, and a peerDependency)
* [ ] adding Vite as a devDependency of the template
* [ ] updating the `"scripts"` of the template package.json files
* [ ] replacing `svelte-kit dev/build/preview` with stubs that tell people to install Vite as a devDependency and update their package.json files, and do `vite dev/build/preview`

documentation/docs/13-cli.md
```diff
@@ -2,43 +2,7 @@
...
-
-After you've built your app with `svelte-kit build`, you can start the production version (irrespective of any adapter that has been applied) locally with `svelte-kit preview`. This is intended for testing the production build locally, **not for serving your app**, for which you should always use an [adapter](/docs/adapters).
-
-Like `svelte-kit dev`, it accepts the following options:
-
-- `-p`/`--port`
-- `-o`/`--open`
-- `--host`
-- `--https`
+SvelteKit is built with Vite and so you will mostly use its CLI. However, SvelteKit includes a command line interface for packaging libraries.
```
Rich-Harris, line 5: note to self: without `dev`/`build`/`preview`, this paragraph transitions to the next quite strangely

documentation/docs/13-cli.md
```diff
@@ -47,3 +11,7 @@ Like `svelte-kit dev`, it accepts the following options:
 For package authors, see [packaging](/docs/packaging). `svelte-kit package` accepts the following options:
 
 - `-w`/`--watch` — watch files in `src/lib` for changes and rebuild the package
+
+### svelte-kit sync
+
+`svelte-kit sync` creates the generated files for your project such as types and a `tsconfig.json`.
```
Rich-Harris, line 17: we should note when this runs, and that people aren't expected to run it manually

Rich-Harris:
I guess we need to somehow run `sync` before build to avoid this warning. investigating

image

benmccann:
Any idea what might be creating that warning? I'm not sure where Vite might be trying to read that file...

We do `sync` in the build process now, but I guess it must be something that happens before this: https://github.com/sveltejs/kit/blob/ad0890c94c98effdfada00ff6a468e935106c69f/packages/kit/src/vite/index.js#L132

Rich-Harris:
Not sure. Also trying to figure out why the builds keep hanging (they time out after 45 minutes), or why `adapter-vercel` is complaining about a missing `source-map-support`

benmccann:
Looks like that warning message is from esbuild. It must be from dependency pre-bundling. Unfortunately, it looks like that is being invoked before any of the plugin hooks. Though that's surprising to me because I would think it would need to read the config since that can affect what to optimize

Rich-Harris:
It can read the config without invoking any plugins. We could workaround it with a prebuild script but that seems hackish. I wonder if there's any appetite in Vite for a pre-optimize hook...

Rich-Harris:
I can reproduce the hanging build locally with `vc build` (in the project root — I had to temporarily delete the other vercel project linked to the repo, though apparently it's possible to change the linked project with `vc link`). Haven't managed to diagnose yet

Rich-Harris:
Alright, the problem was `turbo` — there's something about this branch it doesn't like. I don't intend to spend any time figuring out why it hangs on this branch but not on master, so I've just updated the build command on Vercel:

```diff
-cd ../../ && npx turbo run build --force --filter=kit --filter=kit.svelte.dev && cd sites/kit.svelte.dev
+cd ../../packages/kit && pnpm build && cd ../../sites/kit.svelte.dev && pnpm build
```

jeffkelleran:
Leaving a comment for others that are upgrading older projects (mine was a couple of months old). I followed the prescribed steps: 1) Install vite as a dev dependency, 2) Update my scripts, and 3) Create vite.config.js. My svelte.config.js file contained config information for vite (config.kit.vite). That entry kept the dev server from starting.  Running the 'dev' script will give you an error: "config.kit.vite has been removed - use vite.config.js instead". I removed that entry and the dev server started up fine. Hope that helps somebody!

dolie:
Thank you @jeffkelleran !!!
I copy/paste `vite` property content into vite.config.js and it works fine !

PR #5333, opened 2022-06-30
try to get some tests passing

Rich-Harris:
I'm tearing my hair out over some of these. Running `pnpm test` locally consistently fails — some of the tests are flaky, others (like `handles external api`) fail reliably, and I'm struggling to understand why. It feels like a fairly recent phenomenon.

The `handles external api` case is particularly weird. The server that serves `/server-fetch-request-modified.json` is created twice (once for `javaScriptEnabled: true` and once for `false`), and listens on the same port in both cases. Adding logging shows that the first server is closed before the second server is started, so it's not weird that the original port is still available, but what _is_ weird is that the first server handles the request intended for the second server _even though it's already been closed_. Because of that, the request isn't intercepted in the right place.

I'm sure someone with a better understanding of all this could fix it in no time.

Sometimes tests fail with `ERR_NETWORK_CHANGED`; I have no idea what's causing those.

Flaky tests that fail with more specific errors (may add to this list as I find more):

* navigation is cancelled upon subsequent navigation
* links to unmatched routes result in a full page navigation, not a 404
* rest parameters do not swallow characters
* $page.url.hash is correctly set on navigation


packages/kit/test/apps/basics/test/test.js
```diff
@@ -1224,7 +1224,7 @@ test.describe.parallel('Headers', () => {
 	});
 });
 
-test.describe.parallel('Load', () => {
+test.describe('Load', () => {
```
benmccann, line 1227: What if we only make this one serial and leave the rest parallel? Does that fix the issue? It seems like there's at least some amount of strangeness coming from this second server, so it might be worth a shot

benmccann: Nevermind. That's not it. If I comment out the tests that start new servers I still get hangs

benmccann:
I've been testing various commits. It could just be lucky / unlucky with being able to reproduce it, but I'm getting hanging behavior with https://github.com/sveltejs/kit/commit/a6781ae7797b743e20de2b0e3a61e6d15bce8140, but not the commit before

PR #5345, opened 2022-07-02
Bundle SvelteKit with esbuild

Rich-Harris:
This began as an attempt to solve a mystery encountered in #5332, which turned out to be fixable by upgrading `@rollup/plugin-commonjs`. But I think it's worth keeping on its own merits — the build is faster, and the output is smaller (after minifying both outputs) for reasons I don't really understand. And since `esbuild` already lives in the dependency graph, it's 'free'.

### Tests

### Changesets


Rich-Harris:
Oof, maybe it's not quite that simple after all...

Rich-Harris:
Ok, life is too short for this nonsense. If anyone wants to try and figure out how to make this work be my guest

PR #5354, opened 2022-07-05
[create-svelte] [fix] todos failing to delete

### LINKED ISSUE #5350: create demo app behaviour
### Describe the bug

When creating a svelte-kit app  `npm create svelte my-app` with the option `SvelteKit demo app`, the `todos` app is not working properly in `localhost:3000/todos`. The todo/done checkbox does not keep its state. If you change a task to `done` it goes back to `not done` state.

### Reproduction

1. `npm create svelte my-app` with the option `SvelteKit demo app`.
2. `go to my-app folder and `yarn`
3. `npm run dev --- ---open`
4. go to localhost:3000/todos
5. create some todos and then check one of them.  It changes to `done` and then after some seconds, it goes back to `not done`. See: https://share.getcloudapp.com/jkuXv7v9

### Logs

_No response_

### System Info

```shell
Mac M1 with MacOS 12.4
Chrome Version 103.0.5060.53 (Official Build) (arm64)
```

### Severity

annoyance

### Additional Information

The matter with this possible bug is that it affects new users of svelte-kit.

gtm-nayan:
closes #5350

Didn't see anything in api.svelte.dev repo and worktop hasn't changed so....

image

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/templates/default/src/routes/todos/_api.ts
+++ b/packages/create-svelte/templates/default/src/routes/todos/_api.ts
@@ -18,7 +18,7 @@ const base = 'https://api.svelte.dev';
  */
 export function api(method: string, resource: string, data?: Record<string, unknown>) {
 	return fetch(`${base}/${resource}`, {
-		method,
+		method: method.toUpperCase(),
 		headers: {
 			'content-type': 'application/json'
 		},
--- a/.changeset/wild-terms-pump.md
+++ b/.changeset/wild-terms-pump.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+fix todos fail to delete in demo app


packages/create-svelte/templates/default/src/routes/todos/_api.ts
```diff
@@ -18,7 +18,7 @@ const base = 'https://api.svelte.dev';
  */
 export function api(method: string, resource: string, data?: Record<string, unknown>) {
 	return fetch(`${base}/${resource}`, {
-		method,
+		method: method.toUpperCase(),
```
Conduitry, line 21: ... This makes me uncomfortable. Does undici require uppercase method names while node-fetch didn't? Needing this feels like a workaround, and I'd prefer trying to get to the bottom of what changed rather than fixing it here in the template project. We shouldn't expect everyone using SvelteKit to always be calling their APIs with upper-case `method`s.

gtm-nayan: Hmm, I think it was just patch that was 400-ing, so I didn't consider the switch from node-fetch to undici as a possible cause, but not completely off the table I guess. I'll do a quick sanity check.

Rich-Harris (approved):

Issue #2040, opened 2021-07-30
Package output is missing .js extensions in imports when using Typescript


mattjennings:
### Describe the problem

This is a Typescript problem (see [this](https://github.com/microsoft/TypeScript/issues/16577), and [this](https://github.com/microsoft/TypeScript/issues/40878)), but package is affected by this because it produces ESM javascript.

Say your `src/lib` folder has two files in it: `index.ts` and `store.ts`. Your `index.ts` looks like this:

```typescript
export { myStore } from './store'
```

When you compile with `svelte-kit package`, the result is this:

```js
export { myStore } from './store'
```

Because there's no .js file extension, this is will likely fail in ESM. SvelteKit projects seem to handle this fine (I'm not sure why), but projects using [webpack](https://github.com/sveltejs/template-webpack) for example will fail with the following error:

```
Module not found: Error: Can't resolve './store' in 'node_modules/your-lib'
Did you mean 'store.js'?
BREAKING CHANGE: The request './store' failed to resolve only because it was resolved as fully specified
(probably because the origin is a '*.mjs' file or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

### Describe the proposed solution

Frustratingly, it seems Typescript's stance on this is [to add .js to your file extensions in your .ts files](https://github.com/microsoft/TypeScript/issues/16577#issuecomment-754941937). This works when compiling through `package`, but will not work if you want to consume `src/lib` yourself within the kit project (i.e under `/routes`, if you wanted to set up docs or something).

The only way around this that I can think of is to add .js extensions in a post-compilation script. Similar to how [tsc-esm](https://github.com/digital-loukoum/tsc-esm#readme) does it. It's dumb that kit has to care about this, but I think it's going to be a common gotcha for many library authors otherwise.

### Alternatives considered

- Compiling to commonjs but... no thanks

### Importance

would make my life easier

### Additional Information

In that `tsc-esm` package, it uses a utility file from [grubber](https://github.com/digital-loukoum/grubber/blob/main/source/utilities/patchJsImports.ts) to add the .js extensions. I wrote up a script and ran it against my package output:

```js
import patchJsImports from '@digitak/grubber/library/utilities/patchJsImports.js'

patchJsImports('./package')
```

and it worked on everything _except_ my index.js file, but I haven't looked much further into that.

mattjennings:
[...]

PR #5360, opened 2022-07-05
[fix] Adjust type imports to satisfy TS NodeNext moduleResolution

dummdidumm:
Discovered through comment in #2040
This is necessary so people can use this new resolution mode and still get proper typings for things like $app/stores
This isn't all that's needed though:
- Vite 3 is needed so we get their exports map which makes vite/client resolve correctly
- Svelte bump needed to properly expose CompileOptions. It's imported from a path which isn't accessible when adhering to export maps

### Tests

### Changesets

### DIFF
--- a/packages/kit/types/index.d.ts
+++ b/packages/kit/types/index.d.ts
@@ -1,7 +1,7 @@
 /// <reference types="svelte" />
 /// <reference types="vite/client" />
 
-import './ambient';
+import './ambient.js';
 
 import { CompileOptions } from 'svelte/types/compiler/interfaces';
 import {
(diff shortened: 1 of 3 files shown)


packages/kit/types/index.d.ts
```diff
@@ -1,7 +1,7 @@
 /// <reference types="svelte" />
 /// <reference types="vite/client" />
 
-import './ambient';
+import './ambient.js';
```
dummdidumm, line 4: This looks stupid but it's what TS wants when using the new resolution algorithm. Luckily this also works for the old one.

dummdidumm:
There's value in merging as-is already, it makes most of the types usable again with the new mode

PR #5386, opened 2022-07-06
adapter types

Rich-Harris:
In #5081, we ask people to manually update their `src/app.d.ts` to include the relevant platform types for their adapter. It's confusing and brittle (people who already have `adapter-cloudflare` installed won't see `env.caches` magically appear, even though it's now there).

In an ideal world we'd somehow magically inject these types, but failing that, asking people to add a [triple-slash directive](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-) is probably reasonable.

Marking as draft because we should do this for all adapters.

### Tests

### Changesets


Rich-Harris:
Netlify and Vercel are actually a little tricky, since you might use `edge: true` or `edge: false` and the `platform` object would differ. Not immediately sure of the best solution to that. No reason for it to hold this up though so I'm marking the PR ready

dummdidumm:
Idea: have two ambient type definitions which are in separate files and importable through `@sveltejs/adapter-vercel/edge` and `@sveltejs/adapter-vercel/noedge` or sth like that (this means the ambient file import needs to be removed from the other files)

Rich-Harris:
The challenge will be that one day we want to allow some endpoints (or even handlers) to run on the edge and others to run in a lambda, depending on their data requirements. Not sure how to reconcile that with a global `App.Platform` interface...

dummdidumm:
Oh yeah that sounds impossible to solve correctly with one global type definition

benmccann (approved):

Rich-Harris:
Perhaps it's an argument for normalizing/abstracting `platform` in those environments. Anyway we can cross that bridge later

PR #5392, opened 2022-07-06
use Vite defaults for port and strictPort

Rich-Harris:
These made sense when we were using `svelte-kit dev` etc, but I'm not sure they do any more

### Tests

### Changesets

### DIFF
--- a/.changeset/dull-ducks-report.md
+++ b/.changeset/dull-ducks-report.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] Use Vite defaults for port and strictPort
--- a/packages/kit/src/vite/index.js
+++ b/packages/kit/src/vite/index.js
@@ -156,10 +156,6 @@ function kit() {
 						input: `${get_runtime_path(svelte_config.kit)}/client/start.js`
 					}
 				},
-				preview: {
-					port: config.preview?.port ?? 3000,
-					strictPort: config.preview?.strictPort ?? true
-				},
 				resolve: {
 					alias: get_aliases(svelte_config.kit)
 				},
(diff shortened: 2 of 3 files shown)


benmccann:
`strictPort` probably makes sense to remove

The Vite default port is insane though. 4173 or something like that and I think it's different between dev and preview making it even more confusing

benmccann:
See https://github.com/vitejs/vite/pull/6330 and https://github.com/vitejs/vite/pull/8148 for the port defaults

Conduitry:
I'm torn. Yeah, those defaults are insane. Maybe as a compromise we could drop the default overrides from Kit itself, but re-add the same 3000 defaults to the starter templates?

dominikg:
They chose unique ports to be associated with vite. 
Different ports for dev and preview can make sense to recognize which mode it is in the url bar and also to run them in parallel.

It should be possible to provide different ports/overrides from the config hook.

bluwy:
Yeah 5173 and 4173 both represent `VITE` so there's a correlation, but one of the reasons why the port number doesn't matter is that users would just ctrl-click the url from the terminal to open up, or the browser already autocompletes it, so you won't always be typing that out.

I think if SvelteKit is truly a Vite plugin now, it make sense to use Vite's default config for it.

benmccann:
I've literally never ctrl+clicked the URL in the terminal and din't know you could until now. Even knowing that you can, I don't think I'd do it as it'll open as a new tab in whatever window I last used, which isn't necessarily where I want to open it. I agree putting the port in the template might be better than putting it in the plugin at least though

Rich-Harris:
I don't know why we'd want to override the default in the template. Is it a weird default? Maybe. Will people get used to it? Yes. Is it hard to change? No. Just feels like an unnecessary distinction from every other Vite app

Issue #5380, opened 2022-07-06
Slightly normalize requests before passing them to Undici


Conduitry:
### Describe the problem

The global `fetch` that now uses Undici under the hood is now a lot more finicky in a few areas than it previously was when using node-fetch. A partial list, which can be expanded when we run into more things:

- `fetch(..., { method: 'patch' })` does not work. Undici needs `PATCH`, although other verbs work in lowercase. This is, I understand, in compliance with some spec, but it's annoying
- `fetch(event.request)` in the `handle` hook to proxy a given incoming request also does not work because of the presence of the `connection` header in the request. This needs to be manually removed (`event.request.headers.delete('connection')`) before calling `fetch`.

### Describe the proposed solution

Lobby Undici to be less anal about the spec.

When that inevitably fails, normalize the values passed to the global `fetch`, preferably without mutating user input. This will admittedly get a bit hairy when trying to support `fetch(url, opts)` and `fetch(opts)` and `fetch(Request)`.

### Alternatives considered

Documentation, I guess.

### Importance

would make my life easier

Rich-Harris:
Given that Undici is identical to Node 18's built-in `fetch`, and is also used in other environments, I think we'd ultimately be doing our users a disservice if we did this:

```js
globalThis.fetch = (info, init) => undici.fetch(normalise_info(info), normalise_init(init));
```

_Maybe_ there's some merit in throwing an error if `method` is lowercase, etc, but ultimately I think people need to learn to do things the spec-compliant way, however annoying that is. (FWIW, `fetch(url, { method: 'patch' })` fails with a 400 error in the browser, so `node-fetch` really _is_ being too permissive.)

I do agree with deleting the `connection` header though

Rich-Harris:
Oh, I guess I thought this was happening because of a `fetch` inside a `load`. What's your use case for `fetch(event.request)`? Maybe that's something that should be on the user too

Conduitry:
It is happening because of a `fetch` in `load`, but this is getting mapped back to `handle` during SSR.

I have APIs on the same domain as is serving the frontend that are getting served by something outside of the SvelteKit app. There's something sitting between the browser and the application that ensures that appropriate URLs are mapped to the frontend and others are mapped to the APIs. This needs to work during SSR as well.

[...]

PR #5393, opened 2022-07-06
[fix] omit Connection header from requests made by load fetch

Conduitry:
This should resolve at least part of #5380. It might be that this ends up being the only thing we want to do as part of that issue. This PR is similar to #3690 which omitted the `Host` header.

### Tests

### Changesets

### DIFF
--- a/.changeset/smart-peas-compare.md
+++ b/.changeset/smart-peas-compare.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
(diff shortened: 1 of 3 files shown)


Conduitry:
So, this time _I_ lied about having run the tests locally. It looks like some tests need to be updated. Before I do so, I would like to confirm with @Rich-Harris that this is what you were suggesting happen in your comments on #5380.

Rich-Harris:
It is

Rich-Harris (approved):

Issue #5370, opened 2022-07-05
Invalid request body


KeithAndreRose:
### Describe the bug

I upgraded svelte-kit to its latest version today. I configured my project to use the new vite.config.js setup, however now  "Invalid request body" is the only response I get in the browser for all routes (except proxy rules). It's occurring while using the default starter/skeleton project as well.
<img width="1440" alt="Screen Shot 2022-07-05 at 6 23 32 PM" src="https://user-images.githubusercontent.com/37423459/177427295-cc9271a2-dd0e-4413-bde2-a4aba02387be.png">
<img width="1440" alt="Screen Shot 2022-07-05 at 6 23 51 PM" src="https://user-images.githubusercontent.com/37423459/177427296-7b67712d-fe49-4d2a-af05-7b2227143cbf.png">

There are no errors reported in the terminal.
Configured server proxy rules do resolve.

### Reproduction

npm create svelte my-app
cd my-app
npm install
npm run dev -- --open

### Logs

```shell
➜  my-app pnpm dev -d             

> my-app@0.0.1 dev /Users/keithrose/Workspace/svelte-test/my-app
> vite dev "-d"

(node:56433) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
  vite:vite-plugin-svelte svelte inspector path: /Users/keithrose/Workspace/svelte-test/my-app/node_modules/@sveltejs/vite-plugin-svelte/src/ui/inspector/ +0ms
  vite:config native esm config loaded in 144.40ms URL {
  href: 'file:///Users/keithrose/Workspace/svelte-test/my-app/vite.config.js',
  origin: 'null',
  protocol: 'file:',
  username: '',
  password: '',
  host: '',
  hostname: '',
  port: '',
  pathname: '/Users/keithrose/Workspace/svelte-test/my-app/vite.config.js',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
} +0ms
  vite:vite-plugin-svelte findSvelteDependencies: searching svelte dependencies in /Users/keithrose/Workspace/svelte-test/my-app +38ms
  vite:vite-plugin-svelte adding bare svelte packages to optimizeDeps.include: svelte/animate, svelte/easing, svelte/internal, svelte/motion, svelte/store, svelte/transition, svelte  +1ms
  vite:vite-plugin-svelte automatically excluding found svelte dependencies:  +0ms
[...]

PR #5395, opened 2022-07-06
[breaking] require Node 16.9

benmccann:
closes https://github.com/sveltejs/kit/issues/5370

not exactly sure why Node 16.9 is now required, but it fixes it, so :shrug:

### DIFF
--- a/.changeset/angry-crabs-walk.md
+++ b/.changeset/angry-crabs-walk.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] require Node 16.9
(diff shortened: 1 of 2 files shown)


Rich-Harris:
I was hoping to figure out why it's failing with 16.7 before resorting to this (though bumping the required version isn't the end of the world, since it's unlikely someone would be forced to be on 16.7)

Rich-Harris:
It turns out that undici requires 16.9 because it uses things like `util.toUSVString` (introduced in 16.8) and `Object.hasOwn` (introduced in 16.9), even though its `engines` field incorrectly says otherwise (https://github.com/nodejs/undici/pull/1526#issuecomment-1178207405).

I think that's unlikely to change, so I'll merge this

PR #5400, opened 2022-07-07
[chore] improve test setup so that Vite ecosystem CI runs

Rich-Harris:
This is the DRYer alternative to #5398 I was suggesting @benmccann

### DIFF
--- a/packages/adapter-static/test/utils.js
+++ b/packages/adapter-static/test/utils.js
@@ -108,5 +108,5 @@ function rimraf(path) {
 }
 
 export const plugin = process.env.CI
-	? (await import('../dist/vite.js')).sveltekit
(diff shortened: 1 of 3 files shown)


packages/adapter-static/test/utils.js
```diff
@@ -110,3 +106,7 @@ function create_server(port, handler) {
 function rimraf(path) {
 	(fs.rmSync || fs.rmdirSync)(path, { recursive: true, force: true });
 }
+
+export const plugin = process.env.CI
+	? (await import('../dist/vite.js')).sveltekit
+	: (await import('../src/vite/index.js')).sveltekit;
```
bluwy, line 112: Looks like the path here is incorrect in CI and causing it to fail

PR #5410, opened 2022-07-07
don't create changelog for test-basics

benmccann:
https://github.com/sveltejs/kit/pull/5394 looks screwed up. we probably need to do this before cutting a release


Rich-Harris:
ah whoops, good catch

PR #5416, opened 2022-07-08
[fix] allow Vite plugins to output mutable assets

benmccann:
@userquin is trying to get `vite-plugin-pwa` working with SvelteKit. That plugin adds a mutable manifest to the bundle. Since the bundle is written to the `outDir` we need the `outDir` to be set to `client` instead of `client/immutable`

### DIFF
--- a/packages/kit/src/vite/build/utils.js
+++ b/packages/kit/src/vite/build/utils.js
@@ -95,9 +95,9 @@ export const get_default_config = function ({ config, input, ssr, outDir }) {
 				input,
 				output: {
 					format: 'esm',
-					entryFileNames: ssr ? '[name].js' : '[name]-[hash].js',
-					chunkFileNames: 'chunks/[name]-[hash].js',
-					assetFileNames: 'assets/[name]-[hash][extname]'
+					entryFileNames: ssr ? '[name].js' : 'immutable/[name]-[hash].js',
+					chunkFileNames: 'immutable/chunks/[name]-[hash].js',
+					assetFileNames: 'immutable/assets/[name]-[hash][extname]'
 				},
 				preserveEntrySignatures: 'strict'
 			},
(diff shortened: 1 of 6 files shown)


benmccann:
Looks like `adapter-static` is ending up with `immutable` in the path twice somehow:
```
cd packages/adapter-static/test/apps/spa
pnpm start
404 - /_app/immutable/immutable/start-d357e0b9.js
```

packages/kit/src/vite/index.js
```diff
@@ -93,27 +93,27 @@ function kit() {
...
-			start: `${get_runtime_path(svelte_config.kit)}/client/start.js`
+			'immutable/start': `${get_runtime_path(svelte_config.kit)}/client/start.js`
 		};
 
-		// This step is optional — Vite/Rollup will create the necessary chunks
-		// for everything regardless — but it means that entry chunks reflect
-		// their location in the source code, which is helpful for debugging
+		// Put unchanging assets in immutable directory. We don't set that in the outDir so that other
+		// plugins can add mutable assets to the bundle. Also have the entry chunks reflect their
+		// location in the source code, which is helpful for debugging
```
Rich-Harris, line 101:
```suggestion
		// This step is optional — Vite/Rollup will create the necessary chunks other
		// for everything regardless — but it means that entry chunks reflect
		// their location in the source code, which is helpful for debugging
```

packages/kit/src/vite/build/utils.js
```diff
@@ -95,9 +95,9 @@ export const get_default_config = function ({ config, input, ssr, outDir }) {
 				input,
 				output: {
 					format: 'esm',
-					entryFileNames: ssr ? '[name].js' : '[name]-[hash].js',
-					chunkFileNames: 'chunks/[name]-[hash].js',
-					assetFileNames: 'assets/[name]-[hash][extname]'
+					entryFileNames: ssr ? '[name].js' : 'immutable/[name]-[hash].js',
```
benmccann, line 98: will this also put newly added chunks in `immutable`? If so, it wouldn't fix the issue this is attempting to address. I wonder if specifying the `input` name as I was doing before was safer

Rich-Harris: yes, because of the next line

benmccann:
this looks good to me. thanks for the help getting it working, Rich!

Rich-Harris (approved):

PR #5417, opened 2022-07-08
small fix in init_service_worker

NikolayMakhonin:
Replaced arrow function with anonymous function in the init_service_worker script for support old browsers

This is necessary for the application to work in WebView on Android >= 5.1 (Chromium >= 39).
Arrow functions are only supported in Android WebView >= 45, [see MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)

### Tests

### Changesets

### DIFF
--- a/.changeset/ninety-dodos-exercise.md
+++ b/.changeset/ninety-dodos-exercise.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/kit": patch
+---
+
+Use anonymous function in service worker init script to support legacy browsers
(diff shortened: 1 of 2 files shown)


benmccann (approved):

Rich-Harris:
Thanks — since a more comprehensive fix would take time, I think it makes sense to merge this to unblock you, but I added a comment to make a regression less likely — details in https://github.com/sveltejs/kit/pull/5417/files#r916857155

PR #4829, opened 2022-05-05
[Docs] Project Structure

Mlocik97:
# fixes https://github.com/sveltejs/kit/issues/4748 

[rendered](https://github.com/Mlocik97/kit/blob/master/documentation/docs/01-project-structure.md)

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [ ] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [ ] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpx changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

Issue #4748, opened 2022-04-26
Orientation


Rich-Harris:
### Describe the problem

We don't have anywhere in the docs that describes the structure of a SvelteKit app. For example, `app.html` isn't explained anywhere, and `%svelte.head%` and `%svelte.body%` (which should really be `%sveltekit.head%` and `%sveltekit.body%`) aren't documented.

### Describe the proposed solution

An 'Orientation' section near the top of the docs that gives a brief overview of the contents of a project freshly created with `npm init svelte`.

Mlocik97:
"Orientation" as name of section sounds weird/ugly for me, it's like talking about sex orientation... maybe "Structure" for describing structure of whole project folder, and resp. "HTML template" for `app.html` file docs?

jchanes04:
Seconding the choice of "Structure" for the name of this section. I'd recommend including information the common usage of a global.css file inside of `static` + `app.html` `head`. Some of the other parts of the docs could be moved here as well, such as explaining `routes` and `lib`.

AleCandido:
Actually, maybe a more general "Templates" or "Templates structure" sections might be appropriate: not only the basic SvelteKit structure is worth to be documented, but also all the other amenities provided by `create-svelte`, including:
1. the basic difference of the two templates (even a single sentence is fine)
2. references to the documentation for other tools (Prettier, TypeScript, ...) and possible relevant cross-projects ([`prettier-plugin-svelte`](https://github.com/sveltejs/prettier-plugin-svelte), [TypeScript integration](https://svelte.dev/blog/svelte-and-typescript), ...)

Point 2. has somehow interactions with ["Integrations"](https://kit.svelte.dev/docs/additional-resources#integrations), that might be expanded with the references above. Still, I'd mention explicitly everything that is referred and produced by the user facing `create-svelte` bootstrapping package.
At the moment, the easiest way to get an idea about is to have a look at the code (that is clean and templates are well separated, but not the best).

carusog:
I like @AleCandido idea of an overview section. Probably before anything else, optimally between "Introduction" and "Web Standards" to explore the "Project Structure" that has been generated by the command listed in "[Introduction > Getting Started](https://kit.svelte.dev/docs/introduction#getting-started)".

That "Project Structure" would be a great walk through that I was missing the first time I've read docs. 
Anecdotally, to better understanding Svelte Kit, I've created myself locally both JS and TS versions of the same starters code just to compare and to learn differences between their folder structures.

[...]

PR #5430, opened 2022-07-08
[docs] project structure

Rich-Harris:
This is a shorter/simpler version of #4829

Closes https://github.com/sveltejs/kit/pull/4829
Closes https://github.com/sveltejs/kit/issues/4748


Rich-Harris:
It's a little tricky, yeah. We can iterate on it over time. The main reason I included tsconfig.json is that we need to explain the whole `extends` thing. I wouldn't be opposed to adding `src/app.d.ts` in a follow-up

PR #5437, opened 2022-07-08
test: add automatic wait for vite client connected

dominikg:
to page navigation functions

This isn't the nicest way to do it and doesn't recognize cases where goBack leads to a load.
But it can stabilize tests that rely on the vite client and previously may have started work before that was ready.

logic taken from vite-plugin-svelte. cc @bluwy 

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/utils.js
+++ b/packages/kit/test/utils.js
@@ -97,14 +97,23 @@ export const test = base.extend({
 		// automatically wait for kit started event after navigation functions if js is enabled
 		const page_navigation_functions = ['goto', 'goBack', 'reload'];
 		page_navigation_functions.forEach((fn) => {
+			// @ts-ignore
 			const page_fn = page[fn];
 			if (!page_fn) {
 				throw new Error(`function does not exist on page: ${fn}`);
 			}
+			// @ts-ignore
 			page[fn] = async function (...args) {
 				const res = await page_fn.call(page, ...args);
 				if (javaScriptEnabled) {
-					await page.waitForSelector('body.started', { timeout: 5000 });
+					/**
+					 * @type {Promise<any>[]}
+					 */
+					const page_ready_conditions = [page.waitForSelector('body.started', { timeout: 5000 })];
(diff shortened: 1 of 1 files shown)


dominikg:
unfortunately it looks like that it now hangs the firefox tests. investigating tomorrow.

packages/kit/test/utils.js
```diff
@@ -97,14 +97,24 @@ export const test = base.extend({
 		// automatically wait for kit started event after navigation functions if js is enabled
 		const page_navigation_functions = ['goto', 'goBack', 'reload'];
 		page_navigation_functions.forEach((fn) => {
+			// @ts-ignore
```
benmccann, line 100: why is this needed now when it wasn't before?

dominikg: not sure but not `pnpm lint` fails without it.

dominikg:
firefox may be hanging due to https://github.com/microsoft/playwright/issues/15550

benmccann:
> firefox may be hanging due to https://github.com/microsoft/playwright/issues/15550

I've updated to Playwright 1.25.0, so I think that issue should be fixed now

Rich-Harris:
What's the status of this — is it still needed? Ready for review?

dominikg:
oh, didn't realize the firefox issue might be fixed. Not sure if it is still needed, i think separating the write tests into their own app made this mostly obsolete. lets see what happens after bringing this up 2 date.

dominikg:
now linux+chrome has fails with waiting for that console message. at this point it's safe to say that this does not improve stability

benmccann:
I suspect the reason this was not working is https://github.com/vitejs/vite/pull/7733

dominikg:
> I suspect the reason this was not working is [vitejs/vite#7733](https://github.com/vitejs/vite/pull/7733)

i saw that but vite plugin svelte still uses this check, looks like playwrights console api gives you debug too

PR #5442, opened 2022-07-09
use pull method

### LINKED ISSUE #5412: Memory Overflow when sending big files
### ⚠ Disclaimer
**First of all** I'm not a native english speaker so please forgive my sometimes bad english! ❤

**Second of all** I'm not entirely sure if this issue belongs to ``sveltejs/kit`` or more to ``sveltejs/kit/adapter-node`` but as they both are on the same repository I'm posting this issue here anyway.

**Third of all** I know there may be people saying this is more of a feature request than a bug. But in my opinion it's a bug as it's blocking me from using sveltekit in this project. 

### 🐞 The bug
Sveltekit processes any ``multipart/form-data`` that's hitting an POST endpoint completly in RAM before reaching out to the endpoint. This let's me and anyone who wants to enable file uploading run into 2 related but different problems.
1) It's impossible to check if a file is acceptable (Filesize, Filetype, etc.) from the endpoint before the entire file is already in memory. Even by utilizing a hook before hitting the real endpoint the entire form-data gets parsed and loaded into memory before the hook is called.
2) When transmitting large files (> 1GiB, 5GiB, 10GiB, etc.) the RAM will overflow leading to an **out of memory** error. This normaly can be avoided by streaming the files while parsing the form-data which leads to way lower ram utilization. Anyway because the entire form-data is directly completly read into memory by hitting the post handler enabled endpoint this can't be done in Sveltekit. As above a hook will not solve the problem as the entire form-data gets parsed and loaded into memory before the hook is called.

### 🌟A solution
I don't know if this is acceptable or not (maybe not because it will most likely introduce a breaking change) but a very simple solution for this problem would be to let the user decide if the body should be parsed for them. This in my opinion is the "cleanest way" of fixing this. As we already have the ``request`` ``.formData()``, ``.text()``, ``.json()``, etc. functions in the Web APIs Standard.

*Little side fact to this: first I thought my code was doing this behaviour because I used a ``.formData()`` which for me was the place to do such a thing but currently it's doing it regardless of the existence of it.*

### Skeleton Project
A skeleton project which generates a ~1GiB big file inside the browser and sends it to the node server can be found on: https://github.com/UnlimitedBytes/sveltekit-issue-overflow-formdata

### Can't recommend
Anyway **I do not** recommend using this to test the issue **a much better way** (which won't kill your browser ram too) would be to generate a random file on your os and upload it via a post form!

### Better way
To generate a random file with ~1GiB on different operating systems use:
```sh
# Create 1 GiB file on windows
fsutil file createnew bigfile.bin 1073741824

# Create 1 GiB file on ubuntu/debian
head -c 1073741824 </dev/urandom >bigfile.bin

# Create 1 GiB file on macos
mkfile -n 1g bigfile.bin
```

A simple form to upload a file should look like so:
```html
<form action="/upload" method="POST" enctype="multipart/form-data">
  <input type="file" name="file" /><br />
  <button type="submit">Send File</button>
</form>
```

**IMPORTANT ROUTE CODE**
In order to trigger the issue a POST endpoint handler is needed! So create an ``upload.js`` and/or an ``upload.svelte`` file with the following ``upload.js`` content:
```js
// upload.js
const delay = (ms) => new Promise((resolve) => setTimeout(resolve, ms));

export async function post() {
	// Give enough time to check memory
	await delay(10000);
	return {};
}
```

### Logs

```shell
![windows task manager showing node process on around one GiB RAM usage](https://i.imgur.com/Jmts2Wz.png)
```

### System Info

```shell
System:
  OS: Windows 10 10.0.19044
  CPU: (8) x64 AMD Ryzen™ 7 5800X3D @ 4.50GHz
  Memory: 6.05 GB / 32.00 GB
Binaries:
  Node: 18.4.0 - C:\Program Files\nodejs\node.EXE
  Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
  npm: 8.13.1 - ~\AppData\Roaming\npm\npm.CMD
Browsers:
  Edge: Spartan (44.19041.1266.0), Chromium (103.0.1264.44)
npmPackages:
  @sveltejs/adapter-node: next => 1.0.0-next.78
  @sveltejs/kit: next => 1.0.0-next.360
  svelte: ^3.44.0 => 3.49.0
  vite: ^2.9.13 => 2.9.13
```

### Severity

### Additional Information

_No response_

Rich-Harris:
Closes #5412.

I'm not 100% sure this is correct — in particular, I'm not sure what's supposed to happen if someone tries to upload a huge file and you early-exit with a 413; the response doesn't seem to get acknowledged while the request is still pending (which never stops being true, if you don't buffer it or create a reader).

But it does fix the vulnerability — data is only buffered if you ask for it to be buffered.

No test because I'm not really sure how you'd go about testing such a thing.

### Tests

### Changesets

### DIFF
--- a/.changeset/green-masks-confess.md
+++ b/.changeset/green-masks-confess.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Don't automatically buffer request bodies
--- a/packages/kit/src/node/index.js
+++ b/packages/kit/src/node/index.js
@@ -37,14 +37,15 @@ function get_raw_body(req) {
 		pull(controller) {
 			return new Promise((fulfil) => {
 				req.once('data', (chunk) => {
-					if (cancelled) return;
+					if (!cancelled) {
+						size += chunk.length;
+						if (size > length) {
+							controller.error(new Error('content-length exceeded'));
+						}
 
-					size += chunk.length;
-					if (size > length) {
-						controller.error(new Error('content-length exceeded'));
+						controller.enqueue(chunk);
 					}
 
-					controller.enqueue(chunk);
 					fulfil();
 				});
 			});
--- a/packages/kit/src/node/index.js
+++ b/packages/kit/src/node/index.js
@@ -35,7 +35,7 @@ function get_raw_body(req) {
 		},
 
 		pull(controller) {
-			return new Promise((fulfil, reject) => {
+			return new Promise((fulfil) => {
 				req.once('data', (chunk) => {
 					if (cancelled) return;
 
--- a/packages/kit/src/node/index.js
+++ b/packages/kit/src/node/index.js
@@ -28,7 +28,9 @@ function get_raw_body(req) {
 			});
 
 			req.on('end', () => {
-				controller.close();
+				if (!cancelled) {
+					controller.close();
+				}
 			});
 		},
 
--- a/packages/kit/src/node/index.js
+++ b/packages/kit/src/node/index.js
@@ -18,27 +18,38 @@ function get_raw_body(req) {
 		return null;
 	}
 
+	let size = 0;
+	let cancelled = false;
+
 	return new ReadableStream({
 		start(controller) {
 			req.on('error', (error) => {
 				controller.error(error);
 			});
 
-			let size = 0;
-
-			req.on('data', (chunk) => {
-				size += chunk.length;
-
-				if (size > length) {
-					controller.error(new Error('content-length exceeded'));
-				}
-
-				controller.enqueue(chunk);
-			});
-
 			req.on('end', () => {
 				controller.close();
 			});
+		},
+
+		pull(controller) {
+			return new Promise((fulfil, reject) => {
+				req.once('data', (chunk) => {
+					if (cancelled) return;
+
+					size += chunk.length;
+					if (size > length) {
+						controller.error(new Error('content-length exceeded'));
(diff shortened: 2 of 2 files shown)


benmccann (approved):

AlessioGr:
> data is only buffered if you ask for it to be buffered.

How can I do that? How do I ask for it to be buffered?

How can I customize or disable this size limit for some POST listeners?

irishburlybear:
Ok, Here is what I came up with. My setup is talking to a Directus Server via localhost. 

 const reader = await request.body.getReader() //<- This seems new....
    let buff //<-- create my bucket for data

//The lines below I pulled from the "use stream throttling" #5587
    for (;;){
        const {done, value} = await reader.read()
        if(done) break
        buff += value //For big files, keep piling on the data
        await new Promise((r) => setTimeout(r, 10)) //I think this is used for the throttle not sure if we/I need it..
    }

    console.log(buff.toString('utf-8')) //So I can see what the hell is going on...
    

    const req = await fetch(`http://localhost:8055/files`,{
        method:'POST',
        headers: {
            'Content-Type': request.headers.get('Content-Type') //I had to grab the headers from the original request because I lost them with the forwarding. Before the update, I didn't need to do this unless I got caught in a fluke..
        },

        body: buff //Assemble the data and send to Directus.
    })

PR #5446, opened 2022-07-09
[chore] upgrade turbo

benmccann:
I want to be on the latest version of turbo to see if it fixes the bug I encountered and if not so that I can file a bug pointing at this repo using the latest version


Conduitry:
I don't have a lot to say about the Turbo config, but it looks like in the lockfile that we're on the same version of Turbo as we were before. The semver range has changed, but the lockfile has the same version in it as before, so there hasn't actually been an upgrade.

Conduitry:
Oh, it looks like the upgrade just happened in #5332. So until _then_, the `inputs` didn't do anything, maybe,

dominikg:
omitting inputs would result in tests being rerun even for docs changes. Iirc that was one of the main reasons to introduce turbo. Avoiding redundant executions.

benmccann:
> omitting inputs would result in tests being rerun even for docs changes

Would it? Each package under `packages` would still be cached I believe, so I'm not sure there's anything that would really be rerun if something in `documentation` changes

Rich-Harris:
Without `input`, changing something in e.g. `README.md` would cause a package to be rebuilt unnecessarily. It's not redundant.

PR #5458, opened 2022-07-10
fix(5457): normalize path on prefetch

### LINKED ISSUE #5457: Prefetch and `trailingSlash: 'always'`
### Describe the bug

Not really a bug but an unexpected behavior.

How to reproduce:

1. Use `trailingSlash: 'always'`

```js
// svelte.config.js
const config = 
  kit: {
    trailingSlash: 'always',
  },
}
```

2. Use `sveltekit:prefetch` on a link without the trailing slash

```html
<a href="/page" sveltekit:prefetch>Link</a>
```

This will make the prefetch ineffective because two requests will be issued by the client.

1. When the link is hovered, `load_cache.id` is `/page` and `load_route` is called with `id` = `/page`
2. When the link is clicked, `load_cache.id` is `/page` and `load_route` is called with `id` = `/page/` (with the trailing slash, thus bypassing the cache)

### Reproduction

See above. I can create a repo if need be.

### Logs

_No response_

### System Info

```shell
System:
    CPU: (16) x64 AMD Ryzen 9 5900HX with Radeon Graphics
    Memory: 7.19 GB / 11.69 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 18.4.0 - ~/.volta/tools/image/node/18.4.0/bin/node
    Yarn: 3.2.1 - ~/.volta/tools/image/yarn/1.22.19/bin/yarn
    npm: 8.12.1 - ~/.volta/tools/image/node/18.4.0/bin/npm
  npmPackages:
    svelte: ^3.49.0 => 3.49.0
```

### Severity

annoyance

### Additional Information

_No response_

GauBen:
Fixes #5457

### DIFF
--- a/.changeset/new-geckos-vanish.md
+++ b/.changeset/new-geckos-vanish.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/kit": patch
+---
+
+Normalize paths on prefetch (fixes #5457)
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -898,14 +898,9 @@ export function create_client({ target, session, base, trailing_slash }) {
 			const params = route.exec(path);
 
 			if (params) {
+				const id = normalize_path(url.pathname, trailing_slash) + url.search;
 				/** @type {import('./types').NavigationIntent} */
-				const intent = {
-					id: url.pathname + url.search,
-					route,
-					params,
-					url
-				};
-
+				const intent = { id, route, params, url };
 				return intent;
 			}
 		}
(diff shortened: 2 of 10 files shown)


GauBen:
I have no idea what's happening but this is fun to watch.

As a sidenote, I was thinking about normalizing *before* matching:

```js
const path = decodeURI(url.pathname.slice(base.length) || '/');
const normalized_path = normalize_path(url.pathname, trailing_slash);

for (const route of routes) {
	  const params = route.exec(normalized_path);
	  
	  if (params) {
		  const id = normalized_path + url.search;
		  /** @type {import('./types').NavigationIntent} */
		  const intent = { id, route, params, url };
		  return intent;
	  }
}
```

It may solve current and future issues, or may not, I have no idea.

Rich-Harris (approved):

Rich-Harris:
Thanks!

> I was thinking about normalizing _before_ matching:

I don't think there's any benefit to doing so — it's the same amount of work in either case

PR #5464, opened 2022-07-11
[docs] fix vs code syntax highlighting

benmccann:


Conduitry:
What exactly about this is it that VS Code doesn't like? Is there a simple regex we could run on the entire codebase to make sure there aren't other instances of this?

benmccann:
You can't being to declare a multi-line type on the same line that you start the comment with `/**`. I don't know if there's a simple regex to run, but I ran `grep -r "** @type" .` which printed a ton and I manually scanned through it and I didn't see any others. It's possible I missed one, but I don't think there's many others if any.

Conduitry:
After some poking around, I came up with

```
\/\*\* (.(?!\*\/))*\n
```

which has a few false positives, but it did reveal another shiki-twoslash block in 16-types.md. I don't know whether that one's a problem.

benmccann:
> it did reveal another shiki-twoslash block in 16-types.md

Changing that one doesn't really affect how it's rendered in VS Code or, more importantly, on the site so I'd lean towards leaving it as is

Rich-Harris (approved):

PR #5473, opened 2022-07-12
feat: get client IP from Context object rather than the request header

ericapisani:
Originally the Edge function integrations were written under the assumption that we'd need to get the client IP from the request header. 

However, the client IP is available in the Netlify [`Context` object](https://docs.netlify.com/netlify-labs/experimental-features/edge-functions/api/#netlify-specific-context-object), so updating the adapter to reference that object rather than the request header.

### Tests

### Changesets


Rich-Harris (approved):

Rich-Harris:
thank you!

PR #5480, opened 2022-07-12
[chore] ensure secondary test server is closed

benmccann:
I noticed that one test had `close` in a `finally` and the others did not. It seems like a good idea, so I added it to the others


Rich-Harris:
I almost wonder if this should be part of the `start_server` API, i.e. the test body goes in the second argument, so that it's impossible to forget to do this. would be kinda ugly though so i'm just gonna merge this for now

Rich-Harris (approved):

PR #5486, opened 2022-07-12
[chore] speed up tests

benmccann:
The tests have stopped hanging for me since https://github.com/sveltejs/kit/pull/5482. There is still flakiness. I'd like to debug that, but running the tests takes too long right now, so I'm hoping to re-parallelize them. Has https://github.com/sveltejs/kit/pull/5482 fixed the increased flakiness with parallelization that others were seeing? If not, perhaps we could find a compromise option like parallelize unless on the CI?


Rich-Harris (approved):

Rich-Harris:
let's give it a whirl

Issue #5088, opened 2022-05-26
adapter-static bug in handling binary body in standalone endpoints


trenta3:
### Describe the bug

According to the [official documentation](https://kit.svelte.dev/docs/routing#endpoints-standalone-endpoints), Standalone endpoints can return `Uint8Array` types.
The given binary stream is treated correctly by adapterAuto but not by adapterStatic.

With an endpoint as simple as
```javascript
// src/routes/binary.dat.js
export async function get() {
    return { body: new Uint8Array([0xB0, 0xD8]) }
}
```
the returned file hexdump is correctly `b0 d8` when adapterAuto is used, and incorrectly `ef bf bd ef bf bd` when adapterStatic is used.

The sequence `ef bf bd` is unicode replacement character, which makes me suspect that the returned bytearray is incorrectly treated by adapterStatic as an UTF8 string or casted to a string.

### Reproduction

Repository with source code: https://github.com/trenta3/bug-sk-adapter-static
Diff with respect to the standard initialization template: https://github.com/trenta3/bug-sk-adapter-static/commit/96504ae7f83e46060e8e208e1a2a47c0c0c91ac7

### Logs

```shell
# When running in development mode
$ npm run dev
# In another terminal
$ curl http://localhost:3000/binary.dat 2>/dev/null | hexdump -C
00000000  b0 d8                                             |..|
00000002

# When running in building mode
$ npm run build
$ hexdump -C build/binary.bat
00000000  ef bf bd ef bf bd                                 |......|
00000006
```

### System Info

```shell
System:
    OS: Linux 5.17 Fedora Linux 35 (Workstation Edition)
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 7.81 GB / 15.47 GB
    Container: Yes
    Shell: 5.1.8 - /bin/bash
  Binaries:
    Node: 16.14.0 - /usr/bin/node
    npm: 8.5.5 - ~/bin/npm
  Browsers:
    Chrome: 101.0.4951.64
    Firefox: 100.0
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.48 
    @sveltejs/adapter-static: ^1.0.0-next.32 => 1.0.0-next.32 
    @sveltejs/kit: next => 1.0.0-next.345 
    svelte: ^3.44.0 => 3.48.0 
```

### Severity

serious, but I can work around it

martinszar:
[...]

PR #5497, opened 2022-07-13
[fix] handle binary data when prerendering

gbkwiatt:
fixes https://github.com/sveltejs/kit/issues/5088

body was always converted using `response.text()` which was always decoded with UTF-8 which then is breaking buffers

**I understand that this only solved problem with images, this is just to illustrate the issue, and maybe someone can come up with better solution**

### Tests

### Changesets


packages/kit/src/core/prerender/prerender.js
```diff
@@ -136,7 +136,12 @@ export async function prerender({ config, entries, files, log }) {
 			}
 		});
 
-		const text = await response.text();
+		let text = null;
+		if (response?.headers.get('Content-Type') && (response.headers.get('Content-Type')).startsWith('image/')) {
```
benmccann, line 140: you should use the existing `is_text` method: https://github.com/sveltejs/kit/blob/d430a33643b33f6caa0b0eb69ed558175dbcb343/packages/kit/src/runtime/server/endpoint.js#L32

Conduitry: What would happen if this just used the `arrayBuffer` version all of the time? Is there a reason to pass a string to `save()` sometimes?

benmccann: yeah, that's probably an even better idea assuming there's nothing that it prevent it from working

gbkwiatt: That all sounds like a good idea. Should I leave it with you and close the PR ? it seems like further down, lint doesn't like body to be not a string either.

benmccann (approved):

packages/kit/src/core/prerender/prerender.js
```diff
@@ -136,9 +136,9 @@ export async function prerender({ config, entries, files, log }) {
 			}
 		});
 
-		const text = await response.text();
+		const body = Buffer.from(await response.arrayBuffer());
```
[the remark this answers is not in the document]
Rich-Harris: actually, i guess not, since `Buffer.from(...).toString()` does something useful unlike `new Uint8Array(...).toString()`. But the type signature for `save` expects a `string | Uint8Array`, and we're creating a `Uint8Array` on line 149.

I feel like we need some more certainty here — we should probably have some tests as well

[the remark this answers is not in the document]
gbkwiatt: I think always using buffer is not a solution either, I've tried that and if there is no buffer, it stops whole thing from crawling any further without any error. So there is check needed whether it can be converted to buffer.

gbkwiatt:
> thank you!
> 
> as mentioned inline, I have some concerns about using `Buffer.from` here. i also think we need some tests before we can merge this

From myself, I can only say that your fix works fine on the project I am working on. It was a mix of 
```terminal
null
image/svg+xml
text/html
application/xml
text/plain
image/jpeg
text/html
```

gbkwiatt:
@Rich-Harris I wonder if you can assist me with writing the test. I am not really familiar with repo structure, I've started writing route for the image [kit](https://github.com/sveltejs/kit)/[packages](https://github.com/sveltejs/kit/tree/master/packages)/[kit](https://github.com/sveltejs/kit/tree/master/packages/kit)/[test](https://github.com/sveltejs/kit/tree/master/packages/kit/test)/[prerendering](https://github.com/sveltejs/kit/tree/master/packages/kit/test/prerendering)/[basics](https://github.com/sveltejs/kit/tree/master/packages/kit/test/prerendering/basics)/src/
to be prerendered and saved. But not sure what am I missing as it's not being saved in build folder even with crawl enabled. 
The idea was to test Buffers with output file and input file, although I am not sure that's what you want to test in here since you said you have doubts with `Buffer.from`.

Rich-Harris:
Do you have an `` somewhere in the test app?

gbkwiatt:
> Do you have an `` somewhere in the test app?

Yes that was it, I've had wrong path to image. I've pushed tests, not sure if it's sufficient but it's testing the problem I was facing.

packages/kit/test/prerendering/basics/test/test.js
```diff
@@ -8,6 +8,9 @@ const build = fileURLToPath(new URL('../build', import.meta.url));
 /** @param {string} file */
 const read = (file) => fs.readFileSync(`${build}/${file}`, 'utf-8');
 
+/** @param {string} file */
+const readNoUtf = (file) => fs.readFileSync(`${build}/${file}`);
```
Rich-Harris, line 12:
```suggestion
const read = (file, encoding = 'utf-8') => fs.readFileSync(`${build}/${file}`, encoding);
```

packages/kit/src/core/prerender/prerender.js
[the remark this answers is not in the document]
Rich-Harris: TIL `new Buffer(0) instanceof Uint8Array`, which is why TypeScript didn't complain about this

Rich-Harris (approved):

PR #5503, opened 2022-07-13
[chore] test using windows-latest

benmccann:
@dominikg reported that this repo doesn't build on WIndows 10. The `windows-latest` tag will use Windows Server 2022 instead of Windows Server 2019, so I was curious if that would pick up the build failure. It doesn't but maybe is still worth using `windows-latest` to align with the fact that we use `ubuntu-latest`. Or should we pin it to 2022 so that if things break when they update the latest we won't be surprised?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [16]
-        os: [ubuntu-latest , windows-2019]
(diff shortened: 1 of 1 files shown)


Rich-Harris (approved):

dominikg:
when github ci windows-latest switched from 2019 to 2022, the 2022 runners were not really stable. That was months ago so the situation has likely improved.

I'd hope that our CI passes on either though :+1:

PR #5506, opened 2022-07-13
remove `compression` middleware from adapter-node output server

Rich-Harris:
This is an alternative to #5491. Adding a new option that adds or disables the middleware feels like a bad precedent, and we've often wondered if it really belongs here (rather than at the load balancer level, for example).

Since we clearly document how to use the exported handler with a custom server, I think it's safe to err on the side of simplicity and just get rid of it

### Tests

### Changesets

### DIFF
--- a/.changeset/slimy-ways-applaud.md
+++ b/.changeset/slimy-ways-applaud.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-node': patch
+---
+
+[breaking] remove compression middleware from adapter-node output server
--- a/packages/adapter-node/package.json
+++ b/packages/adapter-node/package.json
@@ -38,10 +38,8 @@
 	"devDependencies": {
 		"@rollup/plugin-json": "^4.1.0",
 		"@sveltejs/kit": "workspace:*",
-		"@types/compression": "^1.7.2",
 		"@types/node": "^16.11.36",
 		"c8": "^7.11.3",
-		"compression": "^1.7.4",
 		"node-fetch": "^3.2.4",
 		"polka": "^1.0.0-next.22",
 		"rimraf": "^3.0.2",
--- a/packages/adapter-node/src/index.js
+++ b/packages/adapter-node/src/index.js
@@ -1,18 +1,12 @@
 import { handler } from './handler.js';
 import { env } from './env.js';
-import compression from 'compression';
 import polka from 'polka';
 
 export const path = env('SOCKET_PATH', false);
 export const host = env('HOST', '0.0.0.0');
 export const port = env('PORT', !path && '3000');
 
-const server = polka().use(
-	// https://github.com/lukeed/polka/issues/173
-	// @ts-ignore - nothing we can do about so just ignore it
-	compression({ threshold: 0 }),
-	handler
-);
+const server = polka().use(handler);
 
 server.listen({ path, host, port }, () => {
 	console.log(`Listening on ${path ? path : host + ':' + port}`);
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -132,10 +132,8 @@ importers:
     specifiers:
       '@rollup/plugin-json': ^4.1.0
       '@sveltejs/kit': workspace:*
-      '@types/compression': ^1.7.2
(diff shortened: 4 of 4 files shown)


benmccann (changes requested):
I much prefer the original PR https://github.com/sveltejs/kit/pull/5491. If we want to avoid the option, maybe we could merge this PR and then revert it after `compression` is fixed?

Rich-Harris:
Adding options for this sort of thing is just (sorry) bad API design. It's out of place, and having it be a boolean value assumes that we know best what options to set for `threshold`, `level` etc. It was a mistake to ever add the middleware

benmccann:
Sure, but it doesn't have to be a boolean. We can pass through the compression options by making it a build time option instead of runtime option

Rich-Harris:
We decided to go ahead with this for now, just so we have a working version. Whether and how we reinstate it is still up for debate

coryvirok:
Sad panda on this one. Anyone have a viable alternative to `compression` to use? I've tried creating a Vite plugin using `compression` but have run into the same hanging issue that this PR addresses.

 Also, does anyone know why this only started happening with the Vite 3 release?

benmccann:
It's not related to Vite 3, but because SvelteKit started streaming recently and compression's streaming implementation is broken. We're waiting for the fix: https://github.com/expressjs/compression/pull/183

itssumitrai:
Is there some alternative which could be applied in mean time to prevent the node hanging issue ? Most of us who use Sveltekit on production environment are currently completely blocked with the streaming & compression change. We have to keep Kit pinned at `1.0.0-next.378`

benmccann:
Compression has been removed, so if you update to the latest it will work. You can do compression in your load balancer or a hook if you need. We can consider adding compression back after it's been fixed

coryvirok:
If someone can figure out a way to implement that hook, I'd be interested. It's pretty crazy how many of the frameworks use the `compression` module and just how long it's been broken for. 

I host on Google Cloud Run which doesn't have an option to add compression so my users are sort of stuck with large payloads till there's a workaround.

fernandolguevara:
@coryvirok an alternative could be to wrap the sveltekit handler (generated by the adapter) with a custom server implementation and in this server u need to enable the comoression middleware.

lkj4:
Since I've upgraded to `next.405` I get following on `pnpm run build`. Is this related to this thread? Can't find the cause and don't know where to look (fwiw, I have a `vite.config.js` and `pnpm run dev` is working flawlessly).

```
failed to load config from /home/[...]/vite.config.js
error during build:
TypeError: Class extends value undefined is not a constructor or null
    at file:///home/[...]/node_modules/.pnpm/@sveltejs+kit@1.0.0-next.405_svelte@3.48.0+vite@3.0.5/node_modules/@sveltejs/kit/dist/node/polyfills.js:17900:25
    at ModuleJob.run (node:internal/modules/esm/module_job:183:25)
    at async Loader.import (node:internal/modules/esm/loader:178:24)
    at async loadConfigFromBundledFile (file:///home/[...]/node_modules/.pnpm/vite@3.0.5/node_modules/vite/dist/node/chunks/dep-5cb728cb.js:62986:21)
    at async loadConfigFromFile (file:///home/[...]/node_modules/.pnpm/vite@3.0.5/node_modules/vite/dist/node/chunks/dep-5cb728cb.js:62872:28)
    at async resolveConfig (file:///home/[...]/node_modules/.pnpm/vite@3.0.5/node_modules/vite/dist/node/chunks/dep-5cb728cb.js:62486:28)
    at async doBuild (file:///home/[...]/node_modules/.pnpm/vite@3.0.5/node_modules/vite/dist/node/chunks/dep-5cb728cb.js:43414:20)
    at async build (file:///home/[...]/node_modules/.pnpm/vite@3.0.5/node_modules/vite/dist/node/chunks/dep-5cb728cb.js:43403:16)
    at async CAC.<anonymous> (file:///home/[...]/node_modules/.pnpm/vite@3.0.5/node_modules/vite/dist/node/cli.js:747:9)
 ELIFECYCLE  Command failed with exit code 1.
```

PR #5507, opened 2022-07-13
[docs] include default value for trailingSlash

benmccann:


Rich-Harris:
yeah, maybe. might be tricky to do in a way that isn't messy/duplicative. either way i think we should close this to avoid doing things inconsistently

Issue #5367, opened 2022-07-05
Uppercase handler names (`export function GET() {...}` etc)


Rich-Harris:
### Describe the problem

Details in [this discussion](https://github.com/sveltejs/kit/discussions/5359). Consensus seems to be that this is a positive change.

### Describe the proposed solution

Replace this...

```js
// src/routes/some-endpoint.js
export function get({ params }) {...}
export function del({ params, request }) {...}
```

...with this:

```js
// src/routes/some-endpoint.js
export function GET({ params }) {...}
export function DELETE({ params, request }) {...}
```

### Alternatives considered

Not doing this.

### Importance

nice to have

UltraCakeBakery:
How about you make it case-insensitive?

UltraCakeBakery:
I removed my previous comment because I thought I misread something, but I now think I didn't. How about making the names of handler functions not case sensitive, and allowing users to export any function from an endpoint? We could make exceptions for functions that start with a `_` or have a reserved keyword.

Rich-Harris:
No, that's a bad idea. APIs should be as explicit and non-magical as possible — loosey-goosey things like accepting `get` or `GET` or `Get` or `gEt` create confusing documentation and pointless debates over best practice

UltraCakeBakery:
> No, that's a bad idea. APIs should be as explicit and non-magical as possible — loosey-goosey things like accepting get or GET or Get or gEt create confusing documentation and pointless debates over best practice

If you are talking about the APIs that users are building with SvelteKit, that should be left up to them and not SvelteKit. If you are talking about SvelteKit's api, I guess you're right. I'm not a fan of arbitrary limitations though and many others care about not having multiple naming conventions in their codebase. 

Maybe it is time to consider taking the same approach other file based routers have taken where you create `endpoint.method.js` files that `export default () => {}`. That way you leave the writing of the method to the file name. SvelteKit maybe should throw a warning when you are using a unsafe http method instead. That way you don't have to document anything else besides "http methods that are [not considered SAFE](https://www.iana.org/assignments/http-methods/http-methods.xhtml) will throw a warning, but still work for backwards compatibility and flexibility". The warning can be something along the lines of `X is not a safe http method. { method.toUpperCase() === method ? '' : 'Methods are case sensitive'}. Did you make a typo?`. Maybe not the most thought out idea but I like this direction more because it will also push people to put shared logic between endpoints in their `$lib` folder.

[...]

PR #5513, opened 2022-07-13
feat: uppercase endpoint methods

tcc-sejohnson:
Proof-of-concept implementation of #5367.

```
get => GET
post => POST
put => PUT
patch => PATCH
del => DELETE
```

The server will throw an error if an endpoint module exports one of the old names. This behavior should be removed for 1.0 (and is marked with a comment like other pre-1.0 throws are).

### Tests

### Changesets


gtm-nayan:
The tests were more regular than I thought they would be, nice.

I found this while I was taking a quick peek the other day:
https://github.com/sveltejs/kit/blob/606b9d6db9102745b5ae7d3f2bee6fa86571dafc/packages/kit/src/vite/build/build_server.js#L258-L295

https://github.com/sveltejs/kit/blob/606b9d6db9102745b5ae7d3f2bee6fa86571dafc/packages/kit/src/core/adapt/builder.js#L50

It's just passed to adapters so couldn't exactly tell what it's doing using `Go to references` and doesn't look like any of the adapters in the repo are relying on it but will likely need changing as well.

tcc-sejohnson:
> The tests were more regular than I thought they would be, nice.

Regex find and replace, buddy. 😉 

Good catch -- I updated those two files along with the type definition (so we can be reasonably certain there are no lowercase variants remaining).

packages/kit/src/runtime/server/endpoint.js
```diff
@@ -45,22 +45,29 @@ export function is_text(content_type) {
 export async function render_endpoint(event, mod, options) {
 	const method = normalize_request_method(event);
```
Rich-Harris, line 46: i'm 95% certain we can rely on `event.request.method` being uppercase already, since that's what `Request` mandates

```suggestion
	const { method } = event.request;
```

Rich-Harris: (suggestions I can't make because they're outside the diff and GitHub is a nuisance: removing the `normalize_request_method` import at the top of the file, and removing the `event.request.` on line 78)

packages/kit/src/runtime/server/utils.js
```diff
@@ -52,8 +52,7 @@ export function is_pojo(body) {
 
 /** @param {import('types').RequestEvent} event */
 export function normalize_request_method(event) {
-	const method = event.request.method.toLowerCase();
-	return method === 'delete' ? 'del' : method; // 'delete' is a reserved word
+	return event.request.method.toUpperCase();
 }
```
Rich-Harris, line 57: We can get rid of this whole function

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -409,13 +409,13 @@ async function load_shadow_data(route, event, options, prerender) {
 	try {
 		const mod = await route.shadow();
 
-		if (prerender && (mod.post || mod.put || mod.del || mod.patch)) {
+		if (prerender && (mod.POST || mod.PUT || mod.DELETE || mod.PATCH)) {
 			throw new Error('Cannot prerender pages that have endpoints with mutative methods');
 		}
 
 		const method = normalize_request_method(event);
```
Rich-Harris, line 416:
```suggestion
		const { method } = event.request;
```

packages/kit/src/vite/build/build_server.js
```diff
@@ -285,9 +281,7 @@ function get_methods(cwd, output, manifest_data) {
 		const file = route.type === 'endpoint' ? route.file : route.shadow;
 
 		if (file && lookup[file]) {
-			methods[file] = lookup[file]
-				.map((x) => /** @type {import('types').HttpMethod} */ (method_names[x]))
-				.filter(Boolean);
+			methods[file] = lookup[file].filter(isHttpMethod);
```
Rich-Harris, line 284:
```suggestion
			methods[file] = lookup[file].filter(is_http_method);
```

packages/kit/src/vite/build/build_server.js
```diff
@@ -4,7 +4,13 @@ import { mkdirp, posixify } from '../../utils/filesystem.js';
 import { get_vite_config, merge_vite_configs, resolve_entry } from '../utils.js';
 import { load_template } from '../../core/config/index.js';
 import { get_runtime_directory } from '../../core/utils.js';
-import { create_build, find_deps, get_default_config, remove_svelte_kit } from './utils.js';
+import {
+	create_build,
+	find_deps,
+	get_default_config,
+	remove_svelte_kit,
+	isHttpMethod
```
Rich-Harris, line 12:
```suggestion
	is_http_method
```

packages/kit/src/vite/build/utils.js
```diff
@@ -150,3 +150,14 @@ export function remove_svelte_kit(config) {
...
+// If we'd written this in TypeScript, it could be easy...
+/**
+ * @param {string} maybeHttpMethod
+ * @returns {maybeHttpMethod is import('types').HttpMethod}
+ */
+export function isHttpMethod(maybeHttpMethod) {
+	const method_names = ['GET', 'HEAD', 'PUT', 'POST', 'DELETE', 'PATCH'];
+
+	return method_names.includes(maybeHttpMethod);
+}
```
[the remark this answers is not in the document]
Rich-Harris: the return value is just a boolean, we don't need anything fancy. most of the time we let TypeScript infer return values, unless it's a longer function or the return value is particularly complex/inscrutable

Rich-Harris (commented):
this looks great to me, i just made a couple of small suggestions

packages/kit/src/vite/build/utils.js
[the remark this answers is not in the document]
tcc-sejohnson: I'm not home right now, but when I was making this commit, I actually just started with `.filter(export => method_names.includes(export))`. However, that ends up returning the same type as the original array (`Array<string>`). You've then got two options: Use a JSDoc cast to just force-cast the result to the correct type, or use a type predicate as the filter argument. The proposed solution (using a boolean return type) actually causes an error for `pnpm check` on my end.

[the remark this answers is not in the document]
Rich-Harris: ooooohhhhhhh i see! i'm a simple man and that flew right over my head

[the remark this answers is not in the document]
tcc-sejohnson: Yeah, confirmed -- changing return type to inferred means `lookup[file].filter(is_http_method)` is inferred as an `Array<string>`, which is not assignable to `methods[file]`, whereas it's an `Array<HttpMethod>` if you specify the return type of `is_http_method` is `str is HttpMethod` rather than just `bool`.

packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -409,13 +409,13 @@ async function load_shadow_data(route, event, options, prerender) {
 	try {
 		const mod = await route.shadow();
```
gtm-nayan, line 411:
```suggestion

	// TODO: Remove for 1.0
	['get', 'post', 'put', 'patch', 'del'].forEach((m) => {
		if (mod[m] !== undefined) {
			const replacement = m === 'del' ? 'DELETE' : m.toUpperCase();
			throw Error(`Endpoint method "${m}" has changed to "${replacement}"`);
		}
	});

```

The error should be here as well?

gtm-nayan:
For my own curiosity, @Rich-Harris what's the output of the get_methods function used for? Couldn't find anything in the repo. Also might be nice to add a doc comment.

packages/kit/src/runtime/server/page/load_node.js
Rich-Harris: ah yep, i'll rustle up a utility that can be used in both places

Rich-Harris:
> For my own curiosity, @Rich-Harris what's the output of the get_methods function used for?

This is... a good question. The only place it appears to be referenced is here: https://github.com/sveltejs/kit/blob/5eb76528b92c5f43c68e03eb5e7f73a37c586f3d/packages/kit/src/core/adapt/builder.js#L40-L51

I vaguely remember there being some reason that it was useful for adapters to have this information to hand. Perhaps if you're using [Architect](https://arc.codes/docs/en/reference/project-manifest/http), where you need to define routes like this:

```
@http
get /
get /pages
get /pages/:dateID
get /contact
post /contact
get /widgets/*
```

Rich-Harris:
alright, let's go... hopefully the positive emoji reactions on the discussion and the issue are representative of the community as a whole 🤞

PR #5516, opened 2022-07-14
[chore] avoid unnecessary test setup and fix type errors

benmccann:
I realized we had a lot of tests where we setup the test, visited zero or one pages, and then conditionally exited without doing any assertions. We can use a conditional `skip` to avoid doing the test setup or visiting any pages. There's almost 50 of these tests, so it could shave 20-30s off the run time.

I also fixed some type errors while I was at it.


dominikg (approved):

Rich-Harris (changes requested):
there's actually a bunch more tests that could be put in a `server.test.js` file — doing that now, hold off on merging please

Rich-Harris:
one downside of this new structure is that we have `test.describe('Foo', ...)` repeated in different test files. Feels a bit weird to have related tests disconnected like that. i'm trying to remember why we moved away from having individual `_test.js` files nestled among the routes; it might be nice if e.g. the `Error` tests were grouped like so:

```
src/routes/errors/
  _test-with-js.js
  _test-without-js.js
  _test.js
  index.svelte
  ...
```

Definitely not something we should worry about for this PR though

benmccann:
Oh, nice! I only looked at ones that were already sort of disabled. I didn't realize there was a whole host of ones that had been overlooked. This will be really nice on the speed front!

benmccann:
> i'm trying to remember why we moved away from having individual _test.js files nestled among the routes

I don't remember that we ever had it this way, but think I prefer the current way regardless. A lot of test groups use multiple routes to test different aspect of a functionality so I think putting the tests in the routes folders would break it up even more

Rich-Harris:
> A lot of test groups use multiple routes to test different aspect of a functionality

They definitely do, but if we were to reorganise them we'd touch up the tests in the process. E.g. the `Endpoints` group of tests includes a test that visits the `/delete-route` route, which has a folder all to itself and is otherwise untouched. With better organisation we wouldn't have random directories like that, we'd ensure that routes were logically grouped

PR #5517, opened 2022-07-14
[docs] add JSDocs for Vite build process

benmccann:
Also refactored out a couple of methods to make `writeBundle` a little shorter and moved a few log statements to better align with code boundaries

### DIFF
--- a/packages/kit/src/vite/index.js
+++ b/packages/kit/src/vite/index.js
@@ -391,14 +391,13 @@ function collect_output(bundle) {
 	const chunks = [];
 	/** @type {import('rollup').OutputAsset[]} */
 	const assets = [];
-	for (const key of Object.keys(bundle)) {
+	for (const value of Object.values(bundle)) {
 		// collect asset and output chunks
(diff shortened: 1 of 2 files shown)


packages/kit/src/vite/index.js
```diff
@@ -341,22 +361,47 @@ function kit() {
...
+	/** @type {import('rollup').OutputAsset[]} */
+	const assets = [];
+	for (const key of Object.keys(bundle)) {
+		// collect asset and output chunks
+		if (bundle[key].type === 'asset') {
+			assets.push(/** @type {import('rollup').OutputAsset} */ (bundle[key]));
+		} else {
+			chunks.push(/** @type {import('rollup').OutputChunk} */ (bundle[key]));
+		}
+	}
```
gtm-nayan, line 401: The key isn't used directly, might as well refactor it to Object.values while we're here.

ignatiusmb: Yep, we can also leverage TS control flow for discriminated unions and remove the type assertion.

Issue #5306, opened 2022-06-28
Prerender in subprocess 


benmccann:
### Describe the problem

SvelteKit `exit`s:

https://github.com/sveltejs/kit/blob/cfd38d40e035a6d9935d365acc2d079d9f15796a/packages/kit/src/vite/index.js#L312

This means you can't have another plugin after the SvelteKit plugin

### Describe the proposed solution

Add a new `close` hook

Have `prerender` call `server.close()`

### Importance

Prevents https://github.com/userquin/pwa-sveltekit-vite from working

Rich-Harris:
Adding a new hook just for this purpose would be unfortunate, I think — especially since (as far as the user is concerned) the connection between prerendering and `close` is unclear. We'd also be relying on people to implement `close` correctly, otherwise their builds would never complete (and people would assume it was because of a SvelteKit bug).

Instead, if we want to avoid exiting the main process, we should probably just do the prerendering in a child process that we can safely kill once prerendering is complete.

furudean:
A proper exit can't be guaranteed in all environments, especially serverless ones.

Conduitry:
> A proper exit can't be guaranteed in all environments, especially serverless ones.

This issue is just referring to things that would happen during _prerendering_ right, and not at runtime? I would certainly hope that we can rely on all the Node stuff we want while the app is being built, regardless of which platform it is being deployed to.

furudean:
Ack, my bad.

userquin:
how about just adding a global `end` build hook, just call it with some context (Vite  and/or SvelteKit) before `process.exit`  is called when an adapter is configured (should also be called when the adapter is not configured)? This way you don't need to change anything, just adding the global end hook call before the exit and also when the adapter is not configured.

IIRC I sent a PR for it a few months ago when trying to include pwa plugin in SvelteKit using the rollup interplugin communication api.

matiasfha:
Maybe completely wrong my questions here but, that this means that is not possible to run a plugin after the `closeBundle` hook. 
I'm looking for a way to run some transformation on the pre-renreder output files.

benmccann:
@matiasfha the prerendered output files get generated in `writeBundle`, so I would expect you to be able to use `writeBundle` if your plugin is after the SvelteKit plugin. You may have to wait for https://github.com/sveltejs/kit/pull/5439 to be released first though to ensure that your changes would get copied by the adapters

[...]

PR #5520, opened 2022-07-14
fix: avoid concealing build errors

### LINKED ISSUE #5413: Improper import in route does not throw but builds nothing
### Describe the bug

When a page (a .svelte-file in the src/routes directory) has a broken import: 
I run `npm run build`.
No errors are thrown, no warnings are given and the site doesnt build. 

Expected behaviour would be one of: 

* Build fails with error
* Build succeeds with warning
* Build succeeds without warning, but throws in runtime

### Reproduction

https://github.com/iverks/sveltekit_import_bug

`npm install`
`npm run build` - nothing gets built, but no errors

Then you can comment out the import in `index.svelte`, and it builds just fine

### Logs

```shell
PS C:\Users\ivism\Progging\bug> npm run build
npm WARN config global `--global`, `--local` are deprecated. Use `--location=global` instead.

> bug@0.0.1 build
> vite build

vite v2.9.13 building for production...
✓ 13 modules transformed.
```

### System Info

```shell
Testen on Windows 10 and windows 11.

From win10:
  System:
    OS: Windows 10 10.0.19042
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 4.37 GB / 15.71 GB
  Binaries:
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.5.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.423.0), Chromium (103.0.1264.44)
    Internet Explorer: 11.0.19041.1
```

### Severity

### Additional Information

About severity: I guess it could be considered "annoying", but its terrible DX and was almost impossible to debug.
Tested with adapter-node and adapter-static.
This is using the latest package.json, vite.config.js, svelte.config.js setup  supplied by `npm create svelte app-name`

gtm-nayan:
closes #5413 

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/vite/index.js
+++ b/packages/kit/src/vite/index.js
@@ -114,6 +114,11 @@ function kit() {
 	 */
 	let build_error;
 
+	/**
+	 * @type {Error | undefined}
+	 */
+	let write_bundle_error;
+
 	function vite_client_config() {
 		/** @type {Record<string, string>} */
 		const input = {
@@ -252,6 +257,9 @@ function kit() {
 		 * Clears the output directories.
 		 */
 		buildStart() {
+			// Reset errors in watch mode
+			build_error = write_bundle_error = undefined;
+
 			if (is_build) {
 				rimraf(paths.build_dir);
 				mkdirp(paths.build_dir);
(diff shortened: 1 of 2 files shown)


packages/kit/src/vite/index.js
```diff
@@ -99,6 +99,11 @@ function kit() {
 	 */
 	let paths;
 
+	/**
+	 * @type {Error | undefined}
+	 */
+	let build_error;
```
Rich-Harris, line 105:
```suggestion
	/** @type {Error | undefined} */
	let build_error;
```

Rich-Harris (approved):

bluwy (approved):

Rich-Harris:
This feels like quite a bit of complexity. IIRC the reason we're not doing the adapt step inside `writeBundle` is that calling `process.exit(0)` in that hook risks interfering with other plugins. What if we just followed through on #5306 — could we avoid all this?

benmccann:
I had wondered as well and https://github.com/sveltejs/kit/issues/5306 would certainly be nice to fix given the number of issues it has caused, but I think the answer is that we still need this because what would it mean to run the adapter on a failed build? I talked with patak that we need additional hooks at the end of the Vite build, so I think that may be part of the answer long-term as well

Rich-Harris:
If adapting was part of `writeBundle` then it _wouldn't_ run on a failed build. we wouldn't get that far into the function

benmccann:
Yes, but it would break Vite plugins that output anything during `writeBundle` as the adapter could no longer utilize their output

Rich-Harris:
Ah, I see. What does that mean in practice though? Are there any adapters that depend on the output of a Vite plugin? Even if so, it feels like all that would need to happen is that the SvelteKit plugin goes after any plugins that generate output the adapter depends on.

benmccann:
> Ah, I see. What does that mean in practice though? Are there any adapters that depend on the output of a Vite plugin?

It's not really specific to any adapter. E.g. `vite-plugin-pwa` generates files during `writeBundle`, but they weren't making them from `.svelte-kit` to `build` or wherever until we moved the adapt phase to happen later

> Even if so, it feels like all that would need to happen is that the SvelteKit plugin goes after any plugins that generate output the adapter depends on.

Unfortunately not because all `writeBundle` hooks run in parallel

packages/kit/src/vite/index.js
```diff
@@ -109,6 +109,12 @@ function kit() {
 	 */
 	let paths;
 
+	/** @type {Error | undefined} */
+	let build_error;
+
+	/** @type {Error | undefined} */
+	let write_bundle_error;
```
Rich-Harris, line 116:
```suggestion
	let errored = false;
```

packages/kit/src/vite/index.js
```diff
@@ -247,6 +253,9 @@ function kit() {
 		 * Clears the output directories.
 		 */
 		buildStart() {
+			// Reset errors in watch mode
+			build_error = write_bundle_error = undefined;
```
Rich-Harris, line 257:
```suggestion
			errored = false;
```

packages/kit/src/vite/index.js
```diff
@@ -256,6 +265,10 @@ function kit() {
 			}
 		},
 
+		buildEnd(err) {
+			build_error = err;
```
Rich-Harris, line 269:
```suggestion
			errored = !!err;
```

packages/kit/src/vite/index.js
```diff
@@ -266,80 +279,86 @@ function kit() {
...
-				if (svelte_config.kit.paths.assets) {
-					throw new Error('Cannot use service worker alongside config.kit.paths.assets');
+					await build_service_worker(options, prerendered, client.vite_manifest);
 				}
-
-				log.info('Building service worker');
-
-				await build_service_worker(options, prerendered, client.vite_manifest);
+			} catch (error) {
+				throw (write_bundle_error = /** @type {Error} */ (error));
```
Rich-Harris, line 361:
```suggestion
				errored = true;
				throw (error);
```

packages/kit/src/vite/index.js
```diff
@@ -354,6 +373,11 @@ function kit() {
 			if (!is_build) {
 				return; // vite calls closeBundle when dev-server restarts, ignore that
 			}
+
+			if (build_error || write_bundle_error) {
```
Rich-Harris, line 377:
```suggestion
			if (errored) {
```

Rich-Harris (commented):
given that we're not using the actual values of `build_error` or `write_bundle_error` anywhere, can we just track it as a boolean? suggestions inline

Rich-Harris:
in fact, could we make it even simpler than my suggestion above? rather than tracking failure, we could track success. presumably `writeBundle` isn't called if the build fails, so we could get rid of the `buildEnd` hook, and just set `completed_build = true` at the end of `writeBundle`. opening a PR against this branch, one sec

Rich-Harris:
https://github.com/sveltejs/kit/pull/5536

Rich-Harris:
closing since #5536 was merged — thanks all

Issue #4815, opened 2022-05-04
Returning non-200 status from `load` in `__layout.svelte` causes a `Uncaught (in Promise) Error` 


poppa:
### Describe the bug

Returning a non-`200` status from `load` in `__layout.svelte` in a Svelte-Kit app results in an _uncaught promise error_.

It seems to me said function is called recursively when returning a non-`200` status. If you track the number of calls and return `200` after the first call to `load` everything seems to be fine.  

### Reproduction

https://github.com/poppa/svelte-kit-uncaught-in-promise-sample

### Logs

```shell
__layout.svelte? [sm]:26 Uncaught (in promise) Error: Authentication Required
    at load (__layout.svelte? [sm]:26:14)
    at load_node (start.js:946:37)
    at Object._hydrate (start.js:1575:25)
    at async start (start.js:1676:3)
```

### System Info

```shell
System:
    OS: macOS 11.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 53.85 MB / 32.00 GB
    Shell: 3.3.1 - /usr/local/bin/fish
  Binaries:
    Node: 16.13.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 8.1.0 - /usr/local/bin/npm
  Browsers:
    Brave Browser: 101.1.38.111
    Chrome: 101.0.4951.54
    Firefox: 98.0.1
    Safari: 15.0
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.40 
    @sveltejs/kit: next => 1.0.0-next.324 
    svelte: ^3.44.0 => 3.48.0
```

### Severity

serious, but I can work around it

### Additional Information

You guys are doing a great job, and Svelte and Svelte-Kit is just awesome.

poppa:
Thanks for a quick reply 👍 

Personally I'm not a fan at all of the error page being rendered inside the main layout. In my world `__error.svelte` should be its own totally separate layout file, but that's just my opinion. 

I personally just didn't expect `__error.svelte` to issue a call to `load` in `__layout.svelte` (which seems to be the case) since it was said function that issued the render of `__error.svelte` in the first place.

Anyhow, keep up the good work 🥇

Rich-Harris:
[...]

PR #5532, opened 2022-07-14
Expose `status` and `error` to layouts, if rendering a generic error page

Rich-Harris:
This is a braindump follow-up to #4665, which addresses #4815. As of #4665, the behaviour seen in #4815 will be different — it will render the plaintext error page (because an error status is returned from `__layout` during the rendering of the `__error` page), which isn't what we want. 

One possible solution to that is to expose `status` and `error` to `__layout` components, so that if a layout does something like this...

```js
if (!$session.user) {
  return { status: 401 }
}
```

...it's possible to skip that when `load` is called a second time during rendering of the generic error page:

```diff
-if (!session.user) {
+if (!session.user && status === null) {
  return { status: 401 };
}
```

This does feel rather messy, and does feel like it hints at a deeper design flaw. It also adds some complexity, since ordinarily layouts _don't_ see these values (we only 'unwind' layout/leaf nodes until we hit an `__error` that we can render; we don't re-run `__layout` and so there's no opportunity to present `status`/`error`), there's a distinction between generic and non-generic error pages, which has to be somehow replicated in the client (at present it's easy, because if we initialise the client with an error, only the specified `__error` page gets to see it).

An alternative, as suggested in https://github.com/sveltejs/kit/issues/4815#issuecomment-1118289012, is for `__error` pages to be completely separate from the layout tree. I don't _love_ that, since aside from being a significant breaking change it would mean a lot of code duplication between `__error` and `__layout`.

Needs some further thought. In the meantime, I'm just going to leave this here.

### Tests

### Changesets


rmunn:
> An alternative, as suggested in [#4815 (comment)](https://github.com/sveltejs/kit/issues/4815#issuecomment-1118289012), is for `__error` pages to be completely separate from the layout tree. I don't _love_ that, since aside from being a significant breaking change it would mean a lot of code duplication between `__error` and `__layout`.

What about having error pages be separate from the layout tree _unless they opt in_? I.e. if you want your error page to display using the top-level `__layout.svelte` layout, you'd name it `__error@default.svelte`, but an `__error.svelte` page without an explicitly-inherited layout will render as a top-level component all by itself, with no layout components.

This would allow the fallback code to only load the top-level `__error.svelte` and not the top-level `__layout.svelte`, which might help mitigate issues like #2130 and #2154. (See https://github.com/sveltejs/kit/issues/2130#issuecomment-896512668 discussion for more about this.)

Haven't fully thought through the implications of that yet, but it seems like a possible solution here.

Rich-Harris:
Going to close this as it is obsolete in light of https://github.com/sveltejs/kit/discussions/5748.

Allowing error pages more control over layout is probably a smart idea, and one we should revisit after the redesign work is complete

PR #5533, opened 2022-07-14
move some dependencies

Rich-Harris:
While debugging a site deployment failure in #5529, I realised we can probably speed up deployments by only installing dependencies for `@sveltejs/kit` and `kit.svelte.dev`. That failed, because we still have some dependencies that are in the root package.json instead of in `packages`

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,6 @@
 	"homepage": "https://github.com/sveltejs/kit#readme",
 	"devDependencies": {
 		"@changesets/cli": "^2.22.0",
-		"@rollup/plugin-commonjs": "^22.0.1",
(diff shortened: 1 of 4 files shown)


benmccann:
Is there a rule we should follow as to what goes where? E.g. I think `rollup` is in both the root config and the sub-project configs, which seems duplicative

Rich-Harris:
dependencies should be local, with vanishingly few exceptions. though i figured out why the deploy was failing in #5529 (fixed in #5534) so i'm not deeply concerned about finishing this PR

Rich-Harris:
going to close this, i don't have the energy

PR #5541, opened 2022-07-15
[fix] ensure completed_build is reset

benmccann:
I missed this while reviewing https://github.com/sveltejs/kit/pull/5536 the first time around

### DIFF
--- a/packages/kit/src/vite/index.js
+++ b/packages/kit/src/vite/index.js
@@ -176,6 +176,7 @@ function kit() {
 			vite_config_env = config_env;
 			svelte_config = await load_config();
 			is_build = config_env.command === 'build';
+			completed_build = false;
 
 			paths = {
 				build_dir: `${svelte_config.kit.outDir}/build`,
--- a/.changeset/lucky-guests-act.md
+++ b/.changeset/lucky-guests-act.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] ensure completed_build is reset


packages/kit/src/vite/index.js
```diff
@@ -176,6 +176,7 @@ function kit() {
 			vite_config_env = config_env;
 			svelte_config = await load_config();
 			is_build = config_env.command === 'build';
+			completed_build = false;
```
gtm-nayan, line 179: Dominik mentioned `build --watch`, I think it should be in buildStart for that to work.

gtm-nayan: https://stackblitz.com/edit/vitejs-vite-azvzxr?file=vite.config.ts,src%2FApp.svelte&terminal=dev

dominikg (approved):

packages/kit/src/vite/index.js
benmccann: I don't understand. When the server restarts in that stackblitz, I see `config` called. @dominikg apparently mentioned something and I see @bluwy gave a thumbs up as well. What am I missing?

To me, `config` is a better place to put it because it happens sooner and we're already setting / resetting other variables there.

gtm-nayan: > When the server restarts in that stackblitz, I see config called.

The server doesn't restart in `build --watch`, config is only called when the plugin is loaded initially and we're already setting it to false where it's declared.

Let's say a build succeeds, the user then makes a change to some page which causes a build error, but config isn't called again so the flag isn't reset and it tries to go ahead with adapting in that case even though it shouldn't.

benmccann: Oh, I see. The stackblitz isn't using `build --watch` :laughing:

gtm-nayan: Oop, yeah mb, couldn't get it working according to their docs for startCommand, just realized I coulda just changed the dev script to `build --watch`.

PR #5556, opened 2022-07-15
document transition CSP caveat

### LINKED ISSUE #5215: sveltekit fails to automatically generate CSP hashes for inline styles
### Describe the bug

When using transitions, it seems svelte adds inline styles to the html but then sveltekit fails to automatically add a hash for it causing errors in the browser.

This is what my svelte.config.js has under CSP:

And this is the piece of code that adds inline styles:

Only way to get it working is to add the hash myself to the CSP but that's lame.

Am I doing something wrong? Or is it broken?

### Reproduction

Add a transition to your app and then set CSP for style-src to ['self']

### Logs

```shell
index-9c40d7a5.js:1 Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='), or a nonce ('nonce-...') is required to enable inline execution.
```

### System Info

```shell
System:
    OS: Linux 5.18 Arch Linux
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
    Memory: 28.28 GB / 31.31 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 18.3.0 - /usr/bin/node
    npm: 8.5.5 - /usr/bin/npm
  Browsers:
    Chromium: 102.0.5005.115
```

### Severity

annoyance

### Additional Information

The worst thing of all is if I add the hash myself then it doesn't work under dev mode.... So I have to remove and then add back for production...

Rich-Harris:
There's basically nothing we can do to allow Svelte-controlled CSS transitions on a page with a `style-src` CSP directive that doesn't include `unsafe-inline`. Best we can do is document it.

Closes #5215

### Tests

### Changesets

### DIFF
--- a/documentation/docs/15-configuration.md
+++ b/documentation/docs/15-configuration.md
@@ -154,7 +154,7 @@ When pages are prerendered, the CSP header is added via a `<meta http-equiv>` ta
 
 > When `mode` is `'auto'`, SvelteKit will use nonces for dynamically rendered pages and hashes for prerendered pages. Using nonces with prerendered pages is insecure and therefore forbidden.
 
-Note that most [Svelte transitions](https://svelte.dev/tutorial/transition) work by creating an inline `<style>` element. If you use these in your app, you must either leave the `style-src` directive unspecified or add `unsafe-inline`.
+> Note that most [Svelte transitions](https://svelte.dev/tutorial/transition) work by creating an inline `<style>` element. If you use these in your app, you must either leave the `style-src` directive unspecified or add `unsafe-inline`.
 
 ### moduleExtensions
 
--- a/documentation/docs/15-configuration.md
+++ b/documentation/docs/15-configuration.md
@@ -154,6 +154,8 @@ When pages are prerendered, the CSP header is added via a `<meta http-equiv>` ta
 
 > When `mode` is `'auto'`, SvelteKit will use nonces for dynamically rendered pages and hashes for prerendered pages. Using nonces with prerendered pages is insecure and therefore forbidden.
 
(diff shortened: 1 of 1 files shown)


documentation/docs/15-configuration.md
```diff
@@ -154,6 +154,8 @@ When pages are prerendered, the CSP header is added via a `<meta http-equiv>` ta
 
 > When `mode` is `'auto'`, SvelteKit will use nonces for dynamically rendered pages and hashes for prerendered pages. Using nonces with prerendered pages is insecure and therefore forbidden.
 
+Note that most [Svelte transitions](https://svelte.dev/tutorial/transition) work by creating an inline `<style>` element. If you use these in your app, you must either leave the `style-src` directive unspecified or add `unsafe-inline`.
```
benmccann, line 157: should this be prefixed with `>` to make it more noticeable?

benmccann: If we only get to pick one, then I'd choose this new one :smile: Two call out boxes isn't bad though. I also think we've got one or two that are a big block with multiple paragraphs like:
```
>When...
>
>Note...
```

Rich-Harris: yeah, multiple paragraphs is fine, though usually i think it's multiple paragraphs about a single concept, whereas these are two separate thoughts. will keep the two consecutive blockquotes 

```suggestion
> Note that most [Svelte transitions](https://svelte.dev/tutorial/transition) work by creating an inline `<style>` element. If you use these in your app, you must either leave the `style-src` directive unspecified or add `unsafe-inline`.
```

PR #5562, opened 2022-07-16
feat: `adapter-static` throws buildtime error when `!config.kit.prerender.default && !adapter.fallback`

### LINKED ISSUE #5185: People are confused by `adapter-static`
### Describe the problem

I've lost count of the number of times people have asked "why is my `adapter-static` build output missing `index.html`" (and other similar questions) in Discord. It's always fixed by either adding `config.kit.adapter.fallback = 'foo.html'` or `config.kit.prerender.default = true`. To be clear, this _is_ in the documentation, and not having one of these settings _does_ produce a yellow (?) warning during build, but I've found that people tend to lose the yellow warning in the output from `build`.

When trying to run `svelte-kit dev` or `svelte-kit build`, throw a full-on, red, crash-the-process error stating:

```
It looks like you're using adapter-static, but `config.kit.prerender.default` is `false` and `config.kit.adapter.fallback` is missing. Either set `config.kit.prerender.default` to true (to statically generate your entire site) or set `config.kit.adapter.fallback` to your desired fallback page (to operate in SPA mode). For more information, check out <link to FAQ in adapter-static docs about this issue>
```

Maybe I'm missing some context here, but I don't think there's any case where `adapter-static` should be run without one of those options set, and the fact that it "works" in dev confuses a lot of people.

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

Happy to work on this one if it's a feature we'd actually like to add.

tcc-sejohnson:
Closes #5185.

Quite possibly the smallest git diff I've PRed recently.

There are no additional tests as I'm not sure what the best way to test this is. It'd require a new app and a test that could try to run `npm run build` and make sure it fails with the necessary error.

### Tests

### Changesets


packages/kit/test/prerendering/disabled/svelte.config.js
```diff
@@ -3,7 +3,7 @@ import adapter from '../../../../adapter-static/index.js';
 /** @type {import('@sveltejs/kit').Config} */
 const config = {
 	kit: {
-		adapter: adapter(),
+		adapter: adapter({ fallback: '200.html' }),
```
benmccann, line 6: should we rename the app from `disabled` to `fallback`? This test app looks a bit weird as it doesn't check any of the output or anything

packages/adapter-static/index.js
```diff
@@ -14,8 +14,8 @@ export default function (options) {
 
 		async adapt(builder) {
 			if (!options?.fallback && !builder.config.kit.prerender.default) {
-				builder.log.warn(
-					'You should set `config.kit.prerender.default` to `true` if no fallback is specified'
+				throw Error(
+					"It looks like you're using adapter-static, but `config.kit.prerender.default` is `false` and `fallback` has not been provided as a configuration argument to adapter-static. Either set `config.kit.prerender.default` to true (to statically generate your entire site) or set `config.kit.adapter.fallback` to your desired fallback page (to operate in SPA mode). For more information, check out https://github.com/sveltejs/kit/tree/master/packages/adapter-static#spa-mode"
```
Rich-Harris, line 18:
```suggestion
				throw Error(
					'adapter-static requires `config.kit.prerender.default` to be `true` unless you set the `fallback: true` option to create a single-page app. See https://github.com/sveltejs/kit/tree/master/packages/adapter-static#spa-mode for more information'
```

Rich-Harris (approved):

packages/kit/test/prerendering/disabled/svelte.config.js
Rich-Harris: sure, done

PR #5563, opened 2022-07-16
fix: Defer warnings until after vite warning message

### LINKED ISSUE #5540: 'The following Vite config options will be overridden by SvelteKit' message is hidden
### Describe the bug

If you set a value in `vite.config.js` that SvelteKit needs to control, like `base`, a message is printed:

image

I had to add `clearScreen: false` to the Vite config to take that screenshot, because normally the message is immediately hidden by Vite's welcome message. Not sure if there's a way to either preserve the message or print it immediately after the welcome message?

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-pmx9s3?file=vite.config.js&terminal=dev

### Logs

_No response_

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Yarn: 1.22.10 - /bin/yarn
    npm: 7.17.0 - /bin/npm
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.60 
    @sveltejs/kit: next => 1.0.0-next.373 
    svelte: ^3.46.0 => 3.49.0 
    vite: ^3.0.0 => 3.0.0
```

### Severity

annoyance

### Additional Information

_No response_

tcc-sejohnson:
Closes #5540.

Shamelessly stole some code from @bluwy and @dominikg. In order to defer, we need to add a listener to the server, so I moved the warning into the `configureServer` function. Right now, it'll only work in dev, but I can probably work out a solution to add it to preview as well if necessary.

It's written so that if we ever need to do other stuff after the server starts listening we can pretty easily add it.

I wasn't really sure how to test this, unfortunately. Not sure if it's worth trying?

### Tests

### Changesets


packages/kit/src/vite/index.js
```diff
@@ -439,14 +442,36 @@ function collect_output(bundle) {
...
+		server.httpServer?.on('listening', () => {
+			setTimeout(() => {
+				actions.forEach((action) => action());
+			}, 0);
+		});
+		// @ts-ignore
+		// eslint-disable-next-line prefer-rest-params
+		return _listen.apply(this, arguments);
+	};
+	return server;
```
gtm-nayan, line 474:
```suggestion
		return _listen.apply(this, arguments);
	};
```

packages/kit/src/vite/index.js
```diff
@@ -388,7 +389,9 @@ function kit() {
 		 * @see https://vitejs.dev/guide/api-plugin.html#configureserver
 		 */
 		async configureServer(vite) {
-			return await dev(vite, vite_config, svelte_config);
+			const configWarning = create_overridden_config_action(original_config, vite_config);
+			const withWarning = add_post_listening_actions(vite, configWarning);
+			return await dev(withWarning, vite_config, svelte_config);
```
gtm-nayan, line 394:
```suggestion
			add_post_listening_actions(vite, configWarning);
			return await dev(vite, vite_config, svelte_config);
```

packages/kit/src/vite/index.js
```diff
@@ -388,6 +389,8 @@ function kit() {
 		 * @see https://vitejs.dev/guide/api-plugin.html#configureserver
 		 */
 		async configureServer(vite) {
+			const configWarning = create_overridden_config_action(original_config, vite_config);
+			add_post_listening_actions(vite, configWarning);
```
bluwy, line 393:
```suggestion
			const config_warning = create_overridden_config_action(original_config, vite_config);
			add_post_listening_actions(vite, config_warning);
```

I think this needs a snake case.

packages/kit/src/vite/index.js
```diff
@@ -439,14 +442,35 @@ function collect_output(bundle) {
...
+	const _listen = server.listen;
+	server.listen = function () {
+		server.httpServer?.on('listening', () => {
+			setTimeout(() => {
+				actions.forEach((action) => action());
+			}, 100);
+		});
+		// @ts-ignore
+		// eslint-disable-next-line prefer-rest-params
+		return _listen.apply(this, arguments);
```
bluwy, line 472: If we don't need to do anything special on restart, I think we could skip re-assigning server.listen, and attach the `listening` listener only 🤔  That way we can set this up for the `preview` too by using `configurePreviewServer`.

Re 100ms, I saw your discord message, and I'm not sure why it's nor working reliably. 100ms is probably fine for now though.

packages/kit/src/vite/index.js
tcc-sejohnson: I am physically unable to write JavaScript variables in snake case 😂

benmccann:
I don't think there would be any warnings printed in `preview`. I believe the options all get applied during `build` except for `server.preview`. As long as `build` and `dev` work then we should be okay.

packages/kit/src/vite/index.js
```diff
@@ -192,8 +196,6 @@ function kit() {
 
 				const new_config = vite_client_config();
 
-				warn_overridden_config(config, new_config);
```
benmccann, line 195: we should continue to warn in build mode. I'd don't think it'd happen without this

packages/kit/src/vite/index.js
```diff
@@ -76,6 +76,9 @@ export function sveltekit() {
  * @return {import('vite').Plugin}
  */
 function kit() {
+	/** @type {import('vite').UserConfig} */
+	let original_config;
```
benmccann, line 80: i'm actually surprised this works because i've run into Vite mutating the config

`original_vite_config` might be a clearer name to differentiate from the svelte config

bluwy: you're right, vite does mutate the config, or at least if other plugins mutate the config instead of returning a config object 🤔 I think it's rare enough that maybe it's fine for simplicity now. otherwise we'd have to deep clone it, or pre compute the warnings logs beforehand.

dominikg: vite config hook docs explicitly mention that plugins are allowed to mutate the config that is passed in
To get the unmodified one our best shot is reading the config again, the next best thing is a 'pre' plugin that clones it in it's own config hook. (hoping no other plugin has run before it)

dominikg: otoh even if it wasn't a user modification, the config that kit receives is the effective one, so it may still be useful to let the user know that something has a different idea. If they are using a plugin that needs a config that sveltekit does not allow the two are incompatible.

gtm-nayan:
As the timeout isn't completely reliable, I wonder if it'll be used for anything other than logging, so maybe we can [patch](https://github.com/sveltejs/kit/compare/master...gtm-nayan:sveltekit:deferred-logs-basic) print urls instead?

packages/kit/src/vite/index.js
benmccann: My concern was more that we read it, then it's mutated, and we no longer know what the original config is because our version has been mutated

We can't clone the entire config because it has functions and things in it that are hard to clone, but I think the fields we need are all simple fields, so we could first filter out the fields we need and then clone those

tcc-sejohnson:
I think @gtm-nayan is correct. This sort of patching gives me a headache, but it seems to work consistently and it's better than the other solution.

bluwy (approved):

packages/kit/src/vite/index.js
```diff
@@ -388,6 +386,12 @@ function kit() {
 		 * @see https://vitejs.dev/guide/api-plugin.html#configureserver
 		 */
 		async configureServer(vite) {
+			const print_urls = vite.printUrls;
```
dummdidumm, line 389: Should we add an explanatory comment here, why we do this?

```suggestion
			// This method is called by Vite after clearing the screen.
			// This patch ensures we can log any important messages afterwards for the user to see.
			const print_urls = vite.printUrls;
```

dummdidumm (approved):

Rich-Harris:
I tweaked the implementation a bit — functions that mutate external state as a side-effect make me very nervous, doubly so if it _sometimes_ has that side-effect but sometimes has a very different side-effect (logging) depending on an optional argument. It's a bit [Zalgo-y](https://blog.izs.me/2013/08/designing-apis-for-asynchrony/), and it's harder for things like TypeScript to make sense of. Even better, by just returning the formatted warning (for the caller to do with as they will) we end up with a bit less code.

Monkey-patching `printUrls` is a beautiful hack. Shit like this makes me proud to be a JavaScript developer

Issue #5575, opened 2022-07-17
CF Pages PR (#5453) incorrect


isaac-mcfadyen:
### Describe the bug

In the recent PR (https://github.com/sveltejs/kit/pull/5453) it was stated that, in order to make `_headers` and `_redirects` work you need to add them to the `/static` directory.

However, **these files do not work with `_worker.js`**.

I confirmed this with a Cloudflare Pages employee, who said that it's up to the framework or other platform bundling the `_worker.js` to handle these, as they are normally injected during the Functions build step which is overriden with `_worker.js`.

Since this information is incorrect it should be removed.

### Reproduction

N/A

### Logs

```shell
N/A
```

### System Info

```shell
N/A
```

### Severity

annoyance

PR #5453, opened 2022-07-10
doc: add notice about custom configuration for CF-pages

nosovk:
### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [X] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [X] This message body should clearly illustrate what problems it solves.
- [X] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [X] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [ ] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #5576, opened 2022-07-17
Updated adapter-cloudflare README

isaac-mcfadyen:
This PR fixes 2 things in the `adapter-cloudflare` README.md

**Number One:**
- Per https://github.com/sveltejs/kit/issues/4916, SvelteKit has dropped Node 14 support.
- This PR updates the `adapter-cloudflare` docs to remove mentions about SvelteKit supporting Node 14.

**Number Two:**
- `_headers` and `_redirects` config files that Cloudflare Pages uses only apply to static assets, not server-side rendered pages.
- Info about these files was added to the README.md, this PR modifies this to add that important detail.
- Closes https://github.com/sveltejs/kit/issues/5575, reverts https://github.com/sveltejs/kit/pull/5453

### Tests

### Changesets

### DIFF
--- a/packages/adapter-cloudflare/README.md
+++ b/packages/adapter-cloudflare/README.md
@@ -47,9 +47,9 @@ When configuring your project settings, you must use the following settings:
 - **Build command** – `npm run build` or `svelte-kit build`
 - **Build output directory** – `.svelte-kit/cloudflare`
 - **Environment variables**
-  - `NODE_VERSION`: `16` or `14`
+  - `NODE_VERSION`: `16`
 
-> **Important:** You need to add a `NODE_VERSION` environment variable to both the "production" and "preview" environments. You can add this during project setup or later in the Pages project settings. SvelteKit requires Node `14.13` or later, so you should use `14` or `16` as the `NODE_VERSION` value.
+> **Important:** You need to add a `NODE_VERSION` environment variable to both the "production" and "preview" environments. You can add this during project setup or later in the Pages project settings. SvelteKit requires Node `16` or later, so you should use `16` as the `NODE_VERSION` value.
 
 ## Environment variables
 
--- a/packages/adapter-cloudflare/README.md
+++ b/packages/adapter-cloudflare/README.md
@@ -83,7 +83,7 @@ declare namespace App {
 
 Functions contained in the `/functions` directory at the project's root will _not_ be included in the deployment, which is compiled to a [single `_worker.js` file](https://developers.cloudflare.com/pages/platform/functions/#advanced-mode). Functions should be implemented as [endpoints](https://kit.svelte.dev/docs/routing#endpoints) in your SvelteKit app.
 
-If you want to use `_headers` or `_redirects` custom [config files](https://developers.cloudflare.com/pages/platform/headers) to modify cloudflare behaviour you should put those configs in the `/static` folder of your SvelteKit project.
+The [`_headers` and `_redirects`](config files) files specific to Cloudflare Pages are not currently supported using SvelteKit. In order to return custom headers or redirect a user, return `headers` or a redirect using [SvelteKit Endpoints](https://kit.svelte.dev/docs/routing#endpoints).
 
 ## Changelog
 
--- a/packages/adapter-cloudflare/README.md
+++ b/packages/adapter-cloudflare/README.md
@@ -84,7 +84,8 @@ declare namespace App {
 Functions contained in the `/functions` directory at the project's root will _not_ be included in the deployment, which is compiled to a [single `_worker.js` file](https://developers.cloudflare.com/pages/platform/functions/#advanced-mode). Functions should be implemented as [endpoints](https://kit.svelte.dev/docs/routing#endpoints) in your SvelteKit app.
 
 The [`_headers` and `_redirects`](config files) files specific to Cloudflare Pages can be used for static asset responses (like images) by putting them into the `/static` folder.
-However, they are not currently supported for server-side rendered pages using SvelteKit. In order to return custom headers or redirect a user on SSRed pages, return `headers` or a redirect using [SvelteKit Endpoints](https://kit.svelte.dev/docs/routing#endpoints).
+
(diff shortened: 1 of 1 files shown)


packages/adapter-cloudflare/README.md
```diff
@@ -47,9 +47,9 @@ When configuring your project settings, you must use the following settings:
 - **Build command** – `npm run build` or `svelte-kit build`
 - **Build output directory** – `.svelte-kit/cloudflare`
 - **Environment variables**
-  - `NODE_VERSION`: `16` or `14`
+  - `NODE_VERSION`: `16`
 
-> **Important:** You need to add a `NODE_VERSION` environment variable to both the "production" and "preview" environments. You can add this during project setup or later in the Pages project settings. SvelteKit requires Node `14.13` or later, so you should use `14` or `16` as the `NODE_VERSION` value.
+> **Important:** You need to add a `NODE_VERSION` environment variable to both the "production" and "preview" environments. You can add this during project setup or later in the Pages project settings. SvelteKit requires Node `16` or later, so you should use `16` as the `NODE_VERSION` value.
```
PH4NTOMiki, line 52:
```suggestion
> **Important:** You need to add a `NODE_VERSION` environment variable to both the "production" and "preview" environments. You can add this during project setup or later in the Pages project settings. SvelteKit requires Node `16.9` or later, so you should use `16` as the `NODE_VERSION` value.
```
per https://github.com/sveltejs/kit/pull/5395

packages/adapter-cloudflare/README.md
```diff
@@ -83,7 +83,7 @@ declare namespace App {
 
 Functions contained in the `/functions` directory at the project's root will _not_ be included in the deployment, which is compiled to a [single `_worker.js` file](https://developers.cloudflare.com/pages/platform/functions/#advanced-mode). Functions should be implemented as [endpoints](https://kit.svelte.dev/docs/routing#endpoints) in your SvelteKit app.
 
-If you want to use `_headers` or `_redirects` custom [config files](https://developers.cloudflare.com/pages/platform/headers) to modify cloudflare behaviour you should put those configs in the `/static` folder of your SvelteKit project.
+The [`_headers` and `_redirects`](config files) files specific to Cloudflare Pages are not currently supported using SvelteKit. In order to return custom headers or redirect a user, return `headers` or a redirect using [SvelteKit Endpoints](https://kit.svelte.dev/docs/routing#endpoints).
```
nosovk, line 86: This is wrong. There is 2 different options. 
1. Cloudflare adapter for sites
2. Cloudflare adapter for workers

If you use cloudflare adapter for sites (new one) it actually supports headers and redirects, but they could be applied only for static content.

For example [here](https://securityheaders.com/?q=https%3A%2F%2Fmain.businessforukraine.pages.dev%2F&followRedirects=on) we added `x-robots-tag: noindex`

```_headers
/*
  X-Frame-Options: DENY
  X-Content-Type-Options: nosniff
  X-XSS-Protection: 1; mode=block
  Referrer-Policy: no-referrer-when-downgrade
  Permissions-Policy: document-domain=()

https://:businessforukraine.pages.dev/*
  X-Robots-Tag: noindex
  ```

all headers were added.

Comment (isaac-mcfadyen) [reply packages/adapter-cloudflare/README.md:86]: Are you prerendering that page?

`env.ASSETS.fetch` will apply `_headers` and `_redirects` but `env.ASSETS.fetch` isn't used in SvelteKit SSR for the actual page itself (only images and static assets and the like). See:
https://github.com/sveltejs/kit/blob/2a9c19252912d61bd840441fe80dc13366af41f7/packages/adapter-cloudflare/src/worker.js#L20-L21

I agree we could leave this in but should be clarified that this is for static asset behavior only.

Comment (nosovk) [reply packages/adapter-cloudflare/README.md:86]: Yep, we prerender most of content to get better performance. If you want to use shadowed endpoints or hooks for working with cookies this wouldn't work. But this is explicitly explained at cloudflare documentation for `_headers`

> Custom headers defined in the _headers file are not currently applied to responses from [Functions](https://developers.cloudflare.com/pages/platform/functions/), even if the function route matches the URL pattern. If your Pages application uses Functions, you must migrate any behaviors from the _headers file to the Response object in the appropriate /functions route. When altering headers for multiple routes, you may be interested in [adding middleware](https://developers.cloudflare.com/pages/platform/functions/#adding-middleware) for shared behavior.

from https://developers.cloudflare.com/pages/platform/headers/

This is the main benefit of new cloudflare adapter compared to old one. Previous one counted any request as worker invocation. New adapter allows to decrease workers consumption significantly.

Comment (isaac-mcfadyen) [reply packages/adapter-cloudflare/README.md:86]: I'm afraid you're mistake on that: during beta, static assets are still counted as Functions calls.

But yeah, I saw that warning. We should specifically state this in the SvelteKit docs rather than linking to an external page however. I'll modify my PR.

File: packages/adapter-cloudflare/README.md [hunk 3]
```diff
@@ -83,7 +83,8 @@ declare namespace App {
 
 Functions contained in the `/functions` directory at the project's root will _not_ be included in the deployment, which is compiled to a [single `_worker.js` file](https://developers.cloudflare.com/pages/platform/functions/#advanced-mode). Functions should be implemented as [endpoints](https://kit.svelte.dev/docs/routing#endpoints) in your SvelteKit app.
 
-If you want to use `_headers` or `_redirects` custom [config files](https://developers.cloudflare.com/pages/platform/headers) to modify cloudflare behaviour you should put those configs in the `/static` folder of your SvelteKit project.
+The [`_headers` and `_redirects`](config files) files specific to Cloudflare Pages can be used for static asset responses (like images) by putting them into the `/static` folder.
+However, they are not currently supported for server-side rendered pages using SvelteKit. In order to return custom headers or redirect a user on SSRed pages, return `headers` or a redirect using [SvelteKit Endpoints](https://kit.svelte.dev/docs/routing#endpoints).
```
Rich-Harris, line 87:
```suggestion

However, they will have no effect on responses dynamically rendered by SvelteKit, which should return custom headers or redirect responses from [endpoints](https://kit.svelte.dev/docs/routing#endpoints) or with the [`handle`](https://kit.svelte.dev/docs/hooks#handle) hook.
```

Rich-Harris (approved):

Rich-Harris:
thanks!

PR #5586, opened 2022-07-18
fix: use actual vite version in version check 

dominikg:
to allow custom resolutions, local links etc

the previous implementation did a read of local devDependencies. That does not work when pnpm overrides are used, or the local devDependency is set to a link of a local version etc.

Vite exports it's version so just use that instead.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/vite/index.js
+++ b/packages/kit/src/vite/index.js
@@ -2,7 +2,7 @@ import fs from 'fs';
 import path from 'path';
 import colors from 'kleur';
 import { svelte } from '@sveltejs/vite-plugin-svelte';
-import { searchForWorkspaceRoot, version as vite_version } from 'vite';
+import * as vite from 'vite';
 import { mkdirp, posixify, rimraf } from '../utils/filesystem.js';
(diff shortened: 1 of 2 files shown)


Rich-Harris (approved):

packages/kit/src/vite/index.js
```diff
@@ -2,7 +2,7 @@ import fs from 'fs';
 import path from 'path';
 import colors from 'kleur';
 import { svelte } from '@sveltejs/vite-plugin-svelte';
-import { searchForWorkspaceRoot } from 'vite';
+import { searchForWorkspaceRoot, version as vite_version } from 'vite';
```
bluwy, line 5: I don't think this works in Vite 2 though, `version` is only exported in Vite 3. Maybe we should check `vite/package.json` instead.

Rich-Harris: oh, whoops. presumably that means mucking about with `import-meta-resolve` which is a bit of a pain. what if we `import * as vite from 'vite'` and then check for the existence of the `version` export? if it's there, check it, if not, assume it's version 2

PR #5588, opened 2022-07-18
bump prettier format print width

ignatiusmb:
Makes it just a tad bit nicer to read, some examples (making it in one column so that the images are still visible)

|Before/After|
|---|
|1. `Load`|
|image|
|image|
|2. `RequestHandler` & `RequestHandlerOutput`|
|image|
|image|
|3. etc.|

### Tests

### Changesets

### DIFF
--- a/packages/kit/scripts/extract-types.js
+++ b/packages/kit/scripts/extract-types.js
@@ -51,7 +51,7 @@ function get_types(code, statements) {
 
 			const snippet = prettier.format(code.slice(start, statement.end).trim(), {
 				parser: 'typescript',
-				printWidth: 60,
+				printWidth: 80,
 				useTabs: true,
 				singleQuote: true,
 				trailingComma: 'none'


dummdidumm:
Interesting that it was set to 60 (default is 80 which I still think is small) - for small mobile phones?

ignatiusmb:
Yep, it was mentioned that [some were a bit difficult to read on mobile](https://github.com/sveltejs/kit/pull/4011#issuecomment-1046058352), I believe it was previously unformatted which makes it horizontally long to scroll on small mobile, but 60 is too small and makes it hard to read on desktop instead

Rich-Harris:
i'd guess most of the time our docs are being read on desktop, so this makes sense to be. the galaxy brain move would be to reformat dynamically based on width, but that would result in hydration shift unless we figured out a media-query-based solution, which would be hellishly complex

Rich-Harris (approved):

PR #5590, opened 2022-07-18
Separate write tests

Rich-Harris:
we think this will reduce test flakiness


packages/kit/test/apps/writes/test/test.js
```diff
@@ -0,0 +1,50 @@
...
+		const mounted = javaScriptEnabled ? 1 : 0;
+
+		// we write to the file, to trigger HMR invalidation
+		fs.writeFileSync(file, contents.replace(/PLACEHOLDER:\d+/, `PLACEHOLDER:${Date.now()}`));
+		await page.goto('/double-mount');
+		expect(await page.textContent('h1')).toBe(`mounted: ${mounted}`);
+		await page.click('button');
+		await page.waitForTimeout(100);
+		expect(await page.textContent('h1')).toBe(`mounted: ${mounted}`);
+		fs.writeFileSync(file, contents.replace(/PLACEHOLDER:\d+/, 'PLACEHOLDER:0'));
```
dominikg, line 48: In finally block and write back initial content?

packages/kit/test/apps/writes/package.json
```diff
@@ -0,0 +1,25 @@
+{
+	"name": "test-basics",
```
dominikg, line 2: Use a different name to avoid confusing it with basics app

benmccann:
do we want to turn HMR back on in the other tests now that these ones have been separated out? pro is that it'd be more like what users see. con is that it makes it easier to inadvertently break the tests in a hard-to-diagnose manner

packages/kit/test/apps/writes/src/routes/new-route/index.svelte
```diff
@@ -0,0 +1,14 @@
+<script>
+	import { page } from '$app/stores';
+</script>
+
+<h1>Great success!</h1>
+
+<a href="/routing/a">a</a>
```
dominikg, line 7: Some of the routes do not exist, are they still needed?

Rich-Harris:
> do we want to turn HMR back on in the other tests now

tempted to leave it off for now, on the basis that if we add a test that _depends_ on HMR we'll be reminded to put it in the `writes` test app

benmccann:
Perhaps we should also leave it off in the `writes` app then. It's not safe for two HMR tests to be in the same app and the existing tests don't need HMR on

dominikg:
> Perhaps we should also leave it off in the `writes` app then. It's not safe for two HMR tests to be in the same app and the existing tests don't need HMR on

the writes test is in serial mode so it should be safe to have it on there. testing writes with hmr=false feels like a big risk to miss possible errors happening for the user that has hmr=true by default.

For read-only tests the risk is smaller but still there, so if it's stable i'd reenable hmr everywhere.

Rich-Harris:
i'll try re-enabling it everywhere then we can see if we get any flakiness. hopefully not

Rich-Harris:
Saw a couple of flakes, but only one was a `dev` test. I think we can conclude that it's safe to keep HMR enabled

PR #5596, opened 2022-07-18
remove eslint

Rich-Harris:
i won't die on this hill if others disagree, but i don't think eslint provides any real value over and above what we get from typescript — at least, not enough to justify the time spent appeasing it. the one thing i'm not sure we can add to the typescript config is an equivalent of no-unused-vars — there's noUnusedLocals but that's not quite the same thing


benmccann (approved):
i'd be willing to give it a try for awhile and see if there's anything we miss. if there is we could perhaps add back a minimal eslint with just a couple rules

Issue #5115, opened 2022-05-30
Support multiple static asset dirs through by allowing string arrays for `config.kit.files.assets`


janosh:
### Describe the problem

I'd like to get static assets from multiple locations. In this example the SvelteKit site is inside the `site/` directory as part of a bigger project and I'd like to use static assets from `assets/` and `site/static/`.

```
.
├── assets
│   ├── ...
└── site
    ├── src
    ├── static
    └── ...
```

### Describe the proposed solution

`config.kit.files.assets` could support arrays:

```js
// svelte.config.jg

export default {
  kit: {
    adapter: adapter(),

    files: {
      assets:[ `../assets`, `./static` ],
    },
  },
}
```

This currently raises

> config.kit.files.assets should be a string, if specified

### Alternatives considered

Create an alias of `assets/` in `site/static/`. It works but feels hacky.

### Importance

nice to have

jochemvogel:
Would love to see this. I'm trying to exclude the `static/img` folder from the { [files](https://kit.svelte.dev/docs/modules#$service-worker-files) } in my service worker, because of the size.

I now do it with a workaround, but feels not very natural. Is there another solution for this problem?

Workaround: `const cachedFiles = files.filter(file => !file.startsWith('/img');`

alexanderl19:
Would also love to see this. It looks like vite only supports one public folder, so this might be a vite limitation?

fvilers:
Would need that too !

WillsterJohnson:
this would be very useful in a multisite monorepo, where shared assets are stored in `{root}/assets` and site specific assets are stored in `{root}/apps/appName/assets`.

Currently I'm creating a symlink such that `{root}/apps/appName/assets/global -> ../../../assets`

PR #5601, opened 2022-07-18
set Vite's `publicDir` and `base` options

benmccann:
Benefits:
1. A lot less code in SvelteKit
1. `publicDir` and `base` are now defined for Vite plugins (should help @userquin with his work supporting SvelteKit in `vite-plugin-pwa`)
1. The Vite CLI now correctly includes the base URL, so you can Ctrl+click it in the terminal to be taken to the correct location

I would suggest closing https://github.com/sveltejs/kit/issues/5115 with a suggestion that the issue be filed in the Vite repo


userquin:
@benmccann it is the webmanifest file, not the sw: pwa plugin will emit `manifest.webmanifest` asset in generateBundle but kit will copy it to _app folder.

Rich-Harris:
With this approach, the static assets get copied into `.svelte-kit/output/client`. Presumably we'd need to remove `writeStatic` from the adapter API if we did this

userquin:
> With this approach, the static assets get copied into `.svelte-kit/output/client`. Presumably we'd need to remove `writeStatic` from the adapter API if we did this

Why not just replace static with public folder? I don't need the assets to be written in the output folder.

I only need the publicDir to list the assets matching a globPattern

packages/kit/test/apps/options-2/test/test.js
```diff
@@ -7,7 +7,7 @@ test.describe.configure({ mode: 'parallel' });
 
 test.describe('paths.base', () => {
 	test('serves /basepath', async ({ page }) => {
-		await page.goto('/basepath');
+		await page.goto('/basepath/');
```
Rich-Harris, line 10: what's the reason for this change?

Conduitry:
This will also need changesets for all the adapters, correct?

Rich-Harris:
It would, yeah

packages/kit/test/apps/options-2/test/test.js
benmccann: Vite only serves the index page with a trailing slash

benmccann: Related issue: https://github.com/vitejs/vite/issues/8770

packages/kit/src/core/prerender/prerender.js
```diff
@@ -109,7 +109,7 @@ export async function prerender({ config, entries, files, log }) {
 		if (seen.has(decoded)) return;
 		seen.add(decoded);
 
-		const file = decoded.slice(config.paths.base.length + 1);
+		const file = decoded;
```
Rich-Harris, line 112: why did you remove this? no wonder the adapter-static tests are failing 😜

```suggestion
		const file = decoded.slice(config.paths.base.length + 1);
```

packages/kit/src/core/prerender/prerender.js
```diff
@@ -259,10 +254,10 @@ export async function prerender({ config, entries, files, log }) {
 		for (const entry of config.prerender.entries) {
 			if (entry === '*') {
 				for (const entry of entries) {
-					enqueue(null, config.paths.base + entry); // TODO can we pre-normalize these?
```
Rich-Harris, line 262: I don't understand what's going on with the removal of `config.paths.base`. Surely it needs to be there?

packages/kit/test/apps/options-2/test/test.js
Rich-Harris: okay, but the path is incorrect

packages/kit/src/core/prerender/prerender.js
Rich-Harris: also, not sure why you got rid of the comment. it's still a TODO

packages/kit/src/core/prerender/prerender.js
benmccann: Vite's `baseMiddleware` strips out the `base` path from the URL
https://github.com/vitejs/vite/blob/dfaeb2b7dc39822bf00fc97802b758b17129c37b/packages/vite/src/node/server/index.ts#L513
https://github.com/vitejs/vite/blob/dfaeb2b7dc39822bf00fc97802b758b17129c37b/packages/vite/src/node/server/middlewares/base.ts#L17

packages/kit/src/core/prerender/prerender.js
benmccann: I didn't understand the TODO. I thought it was referring to prepending the base

The base would not be needed as Vite's `baseMiddleware` strips out the `base` path from the URL as mentioned above https://github.com/sveltejs/kit/pull/5601#discussion_r924911591

packages/kit/test/apps/basics/test/server.test.js
```diff
@@ -473,11 +473,6 @@ test.describe('Static files', () => {
 		expect(await r2.json()).toEqual({ works: true });
 	});
 
-	test('Filenames are case-sensitive', async ({ request }) => {
-		const response = await request.get('/static.JSON');
-		expect(response.status()).toBe(404);
-	});
```
Rich-Harris, line 479: why was this test deleted?

benmccann: it looks like it doesn't need to be actually. I just restored it

Rich-Harris: Right, but many servers do, so it's essential that we do too — if we can't get Vite to be strict about casing it's a showstopper for this PR

benmccann:
Closing this for now as it doesn't seem feasible until maybe Vite 4 given current limitations in Vite

PR #5616, opened 2022-07-19
Handle the reader error

repsac-by:
The reader can throw an exception if there is no interception, then the application will crashing.

### Tests

### Changesets

### DIFF
--- a/.changeset/lucky-plums-listen.md
+++ b/.changeset/lucky-plums-listen.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/kit": patch
+---
+
+Handle errors in Node streams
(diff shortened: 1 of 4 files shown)


repsac-by:
> Thank you! Is this the sort of thing it's possible to add a test for, or is it more unpredictable than that?

This is possible if it is possible to expect the server to survive and not fail with an error. We cannot expect exactly a response from the server, since it will be interrupted in any case.

repsac-by:
I think to make two requests, if the second one was successful, then the server is alive. I will try to implement.

Rich-Harris:
thank you!

PR #5620, opened 2022-07-19
[fix] change Vite's output directory from _app to client

benmccann:
This fixes an issue that @userquin ran into trying to make `vite-plugin-pwa` work with SvelteKit. He was using Vite/Rollup's `emit` to output a service worker manifest and expecting it to be served from the root directory, but it was ending up under `_app`. `_app` is arguable only for stuff generated by SvelteKit though. I could imagine lots of other things like a `robots.txt` or something you might want to generate with a Vite plugin that need to go in a certain location and are impossible to put there today.

I also think the code ends up being a little nicer this was as we don't end up having to specify the app dir in quite as many places

I'm splitting this out from https://github.com/sveltejs/kit/pull/5601 as that PR was becoming too unwieldy

### DIFF
--- a/.changeset/strong-cats-wonder.md
+++ b/.changeset/strong-cats-wonder.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-[fix] change Vite's output directory from \_app to client
+[fix] change Vite's output directory from `_app` to client
--- a/.changeset/strong-cats-wonder.md
+++ b/.changeset/strong-cats-wonder.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] change Vite's output directory from \_app to client
--- a/packages/kit/src/core/generate_manifest/index.js
+++ b/packages/kit/src/core/generate_manifest/index.js
@@ -64,7 +64,6 @@ export function generate_manifest({ build_data, relative_path, routes, format = 
 
 	// prettier-ignore
 	return `{
-		appDir: ${s(build_data.app_dir)},
 		assets: new Set(${s(assets)}),
 		mimeTypes: ${s(get_mime_lookup(build_data.manifest_data))},
 		_: {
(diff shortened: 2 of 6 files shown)


.changeset/strong-cats-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] change Vite's output directory from \_app to client
```
Rich-Harris, line 5:
```suggestion
[fix] change Vite's output directory from `_app` to client
```

packages/kit/src/core/generate_manifest/index.js
```diff
@@ -64,7 +64,6 @@ export function generate_manifest({ build_data, relative_path, routes, format =
 
 	// prettier-ignore
 	return `{
-		appDir: ${s(build_data.app_dir)},
 		assets: new Set(${s(assets)}),
```
Rich-Harris, line 67:
```suggestion
		appDir: ${s(build_data.app_dir)},
		assets: new Set(${s(assets)}),
```

Rich-Harris (approved):

packages/kit/src/core/generate_manifest/index.js
```diff
@@ -64,7 +64,6 @@ export function generate_manifest({ build_data, relative_path, routes, format =
 
 	// prettier-ignore
 	return `{
-		appDir: ${s(build_data.app_dir)},
```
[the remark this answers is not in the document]
Rich-Harris: (i fixed this)

Rich-Harris:
This introduces a very minor bug — workers are no longer written to `_app/assets/worker-xyz123.js`, but are instead written to `assets/worker-xyz123.js`. Neither behaviour is correct (ideally it would be `_app/immutable/worker-xyz123.js`; that will be possible with the next release of Vite thanks to https://github.com/vitejs/vite/pull/9212), but putting workers outside `_app` feels _slightly_ naughtier.

Not enough to prevent merging this PR, but worth noting nonetheless

PR #5625, opened 2022-07-19
fix `pnpm format`

benmccann:
https://github.com/sveltejs/kit/pull/5596 removed a little too much


Rich-Harris:
It didn't remove those commands, it just moved them to `lint` (I didn't think it made sense to have `"lint"` scripts that were literally just calling a different run script) — they're now duplicated. Can we do #5628 instead?

Issue #5545, opened 2022-07-15
Data of type 'new Map()' is not getting fetched in browser, but works fine in server side.


yan5:
### Describe the bug

If we return data of type new Map() from corresponding page endpoints, the component getting rendered in browser is unable to get the data, while server side is able to get it correctly.

### Reproduction

index.svelte
---------------
```
<script>
	export let initialData;
	let map = initialData.map;
	console.log('in index.svelte map', map);
</script>
```

index.js
----------
```
export async function get({ params }) {
    let map = new Map();
    map.set("key", "value");
    let initialData = { "map": map }
    return {
        body: {
            initialData
        }
    };
};
```
Below is what we see, as you can see on server side it shows map values {'key' : 'value'}, but on the browser its empty.
Moreoever, the data type itself is changed, on server its still Map, but on browser its changed to dictionary ({}).

![Screenshot from 2022-07-15 19-22-36](https://user-images.githubusercontent.com/20642737/179213927-c0ce776e-b627-40fb-bb1a-22ae8cbc9c74.png)

But if we change the type to diction in index.js as follows 

```
export async function get({ params }) {
    let map = {};
    map["key"] = "value";
    let initialData = { "map": map }

    return {
        body: {
            initialData
        }
    }
}
```

then both server and browser gets the data correctly as shown below.

![Screenshot from 2022-07-15 19-26-37](https://user-images.githubusercontent.com/20642737/179214490-80ed3ffb-4835-4a4a-9822-cb43078b87ba.png)

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 22.04 LTS 22.04 LTS (Jammy Jellyfish)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 3.03 GB / 15.42 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 16.15.0 - ~/.nvm/versions/node/v16.15.0/bin/node
    npm: 8.13.2 - ~/.nvm/versions/node/v16.15.0/bin/npm
  Browsers:
    Brave Browser: 103.1.41.96
    Chrome: 103.0.5060.114
    Firefox: 102.0.1
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.55 
    @sveltejs/kit: next => 1.0.0-next.367 
    svelte: ^3.46.0 => 3.49.0 
[...]

Issue #5358, opened 2022-07-05
Dynamic Component SSR Error


kyjus25:
### Describe the bug

My goal is to dynamically render a component and have it also be server-side rendered in order to replace a system written in PHP that can perform the same task.

Originally I have followed this [REPL](https://svelte.dev/repl/16b375da9b39417dae837b5006799cb4?version=3.25.0), which is close, but does not SSR (moving the code to an async function instead of onMount does not work either)

However, I found that by using a page endpoint and feeding that to the `<svelte:component />` I was then able to dynamically  display a component that is SSR, **but it displays an error in the process**.

### Reproduction

I have prepared a repository that demonstrates the issue:
https://github.com/kyjus25/dynamic-component-ssr-error

It has the following structure:
```
-- lib
    -- MyComponent.svelte
-- routes
    -- index.svelte
    -- index.js
```

`index.js` will serve up a shadow endpoint that imports `MyComponent.svelte` for `index.svelte` to display as a `<svelte:component />`

The component does display properly, and the component is also SSR when viewing the source of the document, but the following error is displayed in the console.

### Logs

```shell
proxy.js:15 [HMR][Svelte] Unrecoverable HMR error in <Index>: next update will trigger a full reload
logError @ proxy.js:15
Proxy<Index> @ proxy.js:380
create_else_block @ root.svelte? [sm]:43
create_default_slot @ root.svelte? [sm]:38
create_slot @ index.mjs:69
create_fragment @ layout.svelte:21
init @ index.mjs:1891
Layout @ layout.svelte:97
createProxiedComponent @ svelte-hooks.js:341
ProxyComponent @ proxy.js:242
Proxy<Layout> @ proxy.js:349
create_if_block_2 @ root.svelte? [sm]:37
create_fragment @ root.svelte? [sm]:36
init @ index.mjs:1891
Root @ root.svelte? [sm]:16
createProxiedComponent @ svelte-hooks.js:341
ProxyComponent @ proxy.js:242
Proxy<Root> @ proxy.js:349
initialize @ start.js:786
_hydrate @ start.js:1741

index.svelte? [sm]:5 Uncaught (in promise) TypeError: switch_value is not a constructor
    at create_fragment (index.svelte? [sm]:5:34)
    at init (index.mjs:1891:37)
    at new Routes (index.svelte? [sm]:2:25)
    at createProxiedComponent (svelte-hooks.js:341:9)
    at new ProxyComponent (proxy.js:242:7)
[...]

Issue #4944, opened 2022-05-16
Enforce that endpoint responses can be serialized


paperdave:
### Describe the bug

When using page endpoints, the JSON output of the request isn't serialized before passing it to the component/load(), causing the SSR version of these pages to be different than the client rendered version. To extend this, you can pass any value through the endpoint, including functions and classes (these ones wont pass type checking), and it will be passed as props to the ssr component.

### Reproduction

Repo: https://github.com/davecaruso/bug-sveltekit-tojson-endpoints

- [Direct link to endpoint](https://github.com/davecaruso/bug-sveltekit-tojson-endpoints/blob/main/src/routes/index.js)
- [Direct link to page component](https://github.com/davecaruso/bug-sveltekit-tojson-endpoints/blob/main/src/routes/index.svelte)

### Logs

```shell
[no relevant logs]
```

### System Info

```shell
System:
    OS: Windows 10 10.0.22000
    CPU: (32) x64 AMD Ryzen 9 5950X 16-Core Processor
    Memory: 41.17 GB / 63.90 GB
  Binaries:
    Node: 17.8.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.17 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.5.5 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (101.0.1210.47)
    Internet Explorer: 11.0.22000.120
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.42 
    @sveltejs/kit: next => 1.0.0-next.330 
    svelte: ^3.44.0 => 3.48.0
```

### Severity

annoyance

### Additional Information

I believe the part of the code that has to be changed is here
https://github.com/sveltejs/kit/blob/master/packages/kit/src/runtime/server/page/load_node.js#L520
as `body` is just passed through. it should either be passed through `JSON.stringify` or some helper function to flatten all objects with `.toJSON`

Rich-Harris:
Hmm. Since no serialization is happening, I wonder if we should just enforce (albeit only at a type level) that the `body` returned from a page endpoint handler needs to be a POJO, rather than always traversing the object speculatively.

Is it important that you're able to rely on `toJSON` happening, or is it more of a shortcut?

davecaruso:
[...]

PR #5635, opened 2022-07-20
feat: warn if page endpoint body is not serializable

gtm-nayan:
#4944 #5358 #5545

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/load_node.js
+++ b/packages/kit/src/runtime/server/page/load_node.js
@@ -3,9 +3,10 @@ import * as set_cookie_parser from 'set-cookie-parser';
 import { normalize } from '../../load.js';
 import { respond } from '../index.js';
 import { LoadURL, PrerenderingURL, is_root_relative, resolve } from '../../../utils/url.js';
-import { check_method_names, is_pojo, lowercase_keys } from '../utils.js';
+import { check_method_names, lowercase_keys } from '../utils.js';
 import { coalesce_to_error } from '../../../utils/error.js';
 import { domain_matches, path_matches } from './cookie.js';
+import { is_pojo, warn_if_not_serdeable } from './utils.js';
 
 /**
  * Calls the user's `load` function.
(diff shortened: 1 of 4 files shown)


packages/kit/src/runtime/server/page/load_node.js
```diff
@@ -3,9 +3,10 @@ import * as set_cookie_parser from 'set-cookie-parser';
 import { normalize } from '../../load.js';
 import { respond } from '../index.js';
 import { LoadURL, PrerenderingURL, is_root_relative, resolve } from '../../../utils/url.js';
-import { check_method_names, is_pojo, lowercase_keys } from '../utils.js';
+import { check_method_names, lowercase_keys } from '../utils.js';
 import { coalesce_to_error } from '../../../utils/error.js';
 import { domain_matches, path_matches } from './cookie.js';
+import { is_pojo, warn_if_not_serdeable } from './utils.js';
```
dummdidumm, line 9: TIL that serdeable is an actual word 😄

packages/kit/src/runtime/server/page/utils.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * @param {any} v
+ */
+export function is_pojo(v) {
```
dummdidumm, line 4: It irks me a little that there's another function with the same name which does something different, and both have no short description so you have to jump into the code and usage locations to know what's this about. 
Also: What's the `TODO` about?

packages/kit/src/runtime/server/page/utils.js
```diff
@@ -0,0 +1,39 @@
...
+	// TODO: javascript-isms
+	return Object.getPrototypeOf(v) === Object.prototype;
+}
+
+/** @param {any} v */
+export function is_serializable_primitive(v) {
+	// TODO: javascript-isms
```
dummdidumm, line 11: What's the `TODO` about?

packages/kit/src/runtime/server/page/load_node.js
gtm-nayan: Serde is common in the rust circles. Guess it hasn't carried over much outside, will change if it's not clear.

gtm-nayan:
Ughh, __SVELTEKIT_DEV__ is throwing a reference error. https://github.com/sveltejs/kit/commit/baaf46e612228137e301eea78d530d5b92b44ff4 sounds like it's the same thing however here it might mean massive amounts of param drilling to do the same.

Rich-Harris:
It occurred to me that since we always need to serialize the data, we can do the checks at serialization time with a custom `replacer` argument to `JSON.stringify`, with very little overhead. I implemented that in a new PR since this one is a little out of date since #5778, so I'll close this — thanks

PR #5640, opened 2022-07-20
[fix] Ensure private cache when something is returned from getSession hook

### LINKED ISSUE #4268: getSession and maxage leaking session data
### Describe the problem

Using `maxage` and `getSession` together can lead to a `cache-control: public` header on a response containing user session data. If the `getSession` hook is intended exclusively for user session data, this combination could easily be misused and is a potential security issue, thus probably shouldn't be possible.

Where `maxage` is used, and `getSession` returns anything other than `void`, `cache-control` should be set to `private`.

### Alternatives considered

`cache-control` could be forced to `private` when any `getSession` hook is implemented. Is there a case for conditionally returning session data only for some routes though? This would allow implementations where some pages could be cached publicly while still using `getSession` for session data on others.

### Importance

nice to have

### Additional Information

_No response_

dummdidumm:
Fixes #4268

### Tests

### Changesets

### DIFF
--- a/.changeset/silly-otters-knock.md
+++ b/.changeset/silly-otters-knock.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Ensure private cache when something is returned from getSession hook
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -115,13 +115,7 @@ export async function render_response({
 			stores: {
 				page: writable(null),
 				navigating: writable(null),
-				/** @type {import('svelte/store').Writable<App.Session>} */
-				session: {
-					...session,
-					subscribe: (fn) => {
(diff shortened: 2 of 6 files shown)


Rich-Harris:
I'm trying to figure out if we could go in the other direction — rather than sacrificing cacheability for pages that don't make use of `session`, could we prevent it from being inlined into the HTML if it's not used? 

I guess someone could do this...

```html
<script>

  onMount(() => {
    return session.subscribe($session => {
      // `$session` is empty, because we thought we
      // didn't need to serialize it 
    });
  });
</script>
```

...which makes it somewhat trickier. 

Yeah, I think I've convinced myself this PR is the best solution

Rich-Harris (approved):

dummdidumm:
The `onMount` case, plus they could use `$session` on a different page which isn't accessed during SSR

PR #5641, opened 2022-07-20
fix prerendered path normalisation

### LINKED ISSUE #5225: Serviceworkers `prerendered` paths incorrect with `trailingSlash: 'always'`
### Describe the bug

When having set `trailingSlash` to `'always'`, the paths for page-endpoints (`__data.json`) and endpoints like `sitemap.xml` in the `prerendered` variable of the `$service-worker` module are normalized to have a trailing slash.

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-xkpctn
See console output

### Logs

_No response_

### System Info

```shell
Not relevant
```

### Severity

### Additional Information

I'm not sure, if it would be enough to revert this change: https://github.com/sveltejs/kit/commit/1ea4b06f158a8f52012e4a4cb5152a80b52f943b#diff-693bcce9e16dcf8cee3dc704a2e1c2f2024b1d540e31c3bb050a5112a61ef2d1

Rich-Harris:
fixes #5225

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/prerender/prerender.js
+++ b/packages/kit/src/core/prerender/prerender.js
@@ -149,11 +149,7 @@ export async function prerender({ config, entries, files, log }) {
 	 * @param {boolean} is_html
 	 */
 	function output_filename(path, is_html) {
-		const file = path.slice(config.paths.base.length + 1);
-
-		if (file === '') {
-			return 'index.html';
-		}
+		const file = path.slice(config.paths.base.length + 1) || 'index.html';
 
 		if (is_html && !file.endsWith('.html')) {
 			return file + (file.endsWith('/') ? 'index.html' : '.html');
@@ -287,9 +283,7 @@ export async function prerender({ config, entries, files, log }) {
 							location: resolved
 						});
 
-						prerendered.paths.push(
-							normalize_path(decoded, file.endsWith('/index.html') ? 'always' : 'never')
-						);
+						prerendered.paths.push(decoded);
 					}
 				}
 			} else {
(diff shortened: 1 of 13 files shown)


packages/kit/src/core/prerender/prerender.js
```diff
@@ -285,7 +285,9 @@ export async function prerender({ config, entries, files, log }) {
 							location: resolved
 						});
 
-						prerendered.paths.push(normalize_path(decoded, 'never'));
+						prerendered.paths.push(
+							normalize_path(decoded, file.endsWith('/index.html') ? 'always' : 'never')
```
dummdidumm, line 289: Why do we need to handle `/index.html` differently? Do we need to handle `/index.html` differently in all cases, so it should be moved into `normalize_path`?

packages/kit/src/vite/build/build_service_worker.js
```diff
@@ -55,9 +55,7 @@ export async function build_service_worker(
 			];
 
 			export const prerendered = [
-				${prerendered.paths
-					.map((path) => s(normalize_path(path, config.kit.trailingSlash)))
-					.join(',\n\t\t\t\t')}
+				${prerendered.paths.map((path) => s(path)).join(',\n\t\t\t\t')}
```
dummdidumm, line 58: Just to clarify: We don't normalize here because we already did in `prerender.js`?

dummdidumm (approved):
Left some questions mainly to clarify some things for myself, looks good otherwise

packages/kit/src/core/prerender/prerender.js
Rich-Harris: I started writing an explanation of this but it just made me realise we can simplify the implementation... one sec

packages/kit/src/vite/build/build_service_worker.js
Rich-Harris: exactly, yep. previously this was normalising pages _and_ endpoints; we only want pages to be normalised, hence moving the logic from here to there

PR #5645, opened 2022-07-21
only normalise internal URLs

### LINKED ISSUE #5087: trailingSlash: "always" changes external URLs
### Describe the bug

When using configuration: `trailingSlash: "always"` then external URLs will also get a trailing slash.

There is a workaround if you set some kind of taget to the anchor element. For example `target="_self"` but it's still a bug that shoud be fixed.

### Reproduction

You can see repro: https://sveltejs-kit-template-default-tq3lbk--3000.local.webcontainer.io/

"Click me" should link to the URL: https://example.com/foo.php?bar=baz&more=params
But it actually navigates to: https://example.com/foo.php/?bar=baz&more=params

Notice the extra slash before the question mark.

You can see source of this example: https://stackblitz.com/edit/sveltejs-kit-template-default-tq3lbk?file=src/lib/header/Header.svelte

### Logs

_No response_

### System Info

```shell
not applicable
```

### Severity

annoyance

### Additional Information

_No response_

Rich-Harris:
fixes #5087, by only normalising URLs we're navigating to once we've established that it's a page within the app

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/apps/basics/test/client.test.js
+++ b/packages/kit/test/apps/basics/test/client.test.js
@@ -553,12 +553,14 @@ test.describe('Routing', () => {
 			res.end('ok');
 		});
 
-		await page.goto(`/routing/slashes?port=${port}`);
-		await page.click(`a[href="http://localhost:${port}/with-slash/"]`);
+		try {
+			await page.goto(`/routing/slashes?port=${port}`);
+			await page.click(`a[href="http://localhost:${port}/with-slash/"]`);
 
-		expect(urls).toEqual(['/with-slash/']);
-
-		await close();
+			expect(urls).toEqual(['/with-slash/']);
+		} finally {
+			await close();
+		}
 	});
 });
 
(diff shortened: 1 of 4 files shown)


packages/kit/test/apps/basics/test/client.test.js
```diff
@@ -544,6 +544,22 @@ test.describe('Routing', () => {
...
+			urls.push(req.url);
+			res.end('ok');
+		});
+
+		await page.goto(`/routing/slashes?port=${port}`);
+		await page.click(`a[href="http://localhost:${port}/with-slash/"]`);
+
+		expect(urls).toEqual(['/with-slash/']);
+
+		await close();
```
benmccann, line 561: we should probably put this in a `finally` block

Rich-Harris:
can't reproduce these failures locally. hmm

Rich-Harris:
oh hang on yes i can. they're in `adapter-static` tests, not `kit` 🤦

benmccann (approved):
lgtm. I have a nagging feeling we should be able to test this in a lighter way, but don't have any immediate suggestions. I think Routify and React Router do some unit tests by mocking the browser, which might work here longer-term

Rich-Harris:
Maybe we could use network mocking https://playwright.dev/docs/test-configuration#network-mocking. I haven't used it so don't know if it's suitable, but possibly worth a try

PR #5648, opened 2022-07-21
set Vite's `publicDir` option

benmccann:
Trying https://github.com/sveltejs/kit/commit/14c18e3cdf023b587e1946e1b012f33cadb8d82a again, but in a way that will pass the CI this time

This sets the `publicDir` option, but still uses our own `sirv` middleware to serve static assets. Having `publicDir` set though means that assets end up where expected for Vite plugins like `vite-plugin-pwa` and that they can find where we keep our assets

### DIFF
--- a/packages/kit/src/vite/preview/index.js
+++ b/packages/kit/src/vite/preview/index.js
@@ -47,13 +47,13 @@ export async function preview(vite, config, protocol) {
 		// files in `static`
 		vite.middlewares.use(scoped(assets, mutable(config.kit.files.assets)));
 
-		// immutable generated client assets
+		// generated client assets
 		vite.middlewares.use(
 			scoped(
 				assets,
 				sirv(join(config.kit.outDir, 'output/client'), {
 					setHeaders: (res, pathname) => {
-						// only apply to build directory, not e.g. version.json
+						// only apply to immutable directory, not e.g. version.json
 						if (pathname.startsWith(`/${config.kit.appDir}/immutable`)) {
 							res.setHeader('cache-control', 'public,max-age=31536000,immutable');
 						}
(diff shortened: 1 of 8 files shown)


packages/kit/src/vite/preview/index.js
```diff
@@ -47,13 +47,13 @@ export async function preview(vite, config, protocol) {
 		// files in `static`
 		vite.middlewares.use(scoped(assets, mutable(config.kit.files.assets)));
```
Rich-Harris, line 49: we can get rid of this — it's covered by the next middleware now that we're using `publicDir`

```suggestion
```

packages/kit/src/vite/preview/index.js
```diff
@@ -47,13 +47,13 @@ export async function preview(vite, config, protocol) {
...
 
-		// immutable generated client assets
+		// generated client assets
```
Rich-Harris, line 50:
```suggestion
		// generated client assets and the contents of `static`
```

packages/kit/src/vite/preview/index.js
Rich-Harris: after getting rid of this, there's only one occurrence of `mutable` (for prerendered dependencies), so perhaps we should inline the `sirv` call. though we might then need to ensure that `prerendered/dependencies` always exists, as I think sirv will barf otherwise

benmccann: I think it's nicely structured as a separate method. I took a look at inlining it locally and it doesn't feel quite as clean, so if it's all the same I'd probably just leave it as is

Rich-Harris (approved):

PR #5651, opened 2022-07-21
fix: use postinstall hook to run sync instead of prepare 

### LINKED ISSUE #4585: Cannot run `npm ci --prod` with adapter-node
### Describe the bug

adapter-node says that `npm ci --prod` should be used to run the build.
However, this command fails

### Reproduction

`npm ci --prod`

### Logs

```shell
> guidest@0.0.1 prepare
> svelte-kit sync

sh: svelte-kit: not found
npm ERR! code 127
npm ERR! path /app
npm ERR! command failed
npm ERR! command sh -c svelte-kit sync

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2022-04-10T11_03_37_215Z-debug.log
The command '/bin/sh -c npm ci --prod' returned a non-zero code: 127
```

### System Info

```
Docker image node:16-alpine
```

### Severity

### Additional Information

Workaround: `sed -i '/prepare/d' package.json && npm ci --prod`

dominikg:
fixes #4585  and possibly a few others where people had issues with missing aliases

caveat:
this currently does not invoke sync for workspace projects depending on `"@sveltejs/kit":"workspace:*"` when running `pnpm install` in kit monorepo. 

We could think about keeping prepare in those, but i'm uneasy doing it do avoid it masking issues with postinstall.

### Tests

### Changesets


Rich-Harris:
Looks like we'll need to do some more setup work for our own CI.

If we move this to postinstall, do we still need to keep `svelte-kit sync`? Feels like at that point we could just have an internal script that we don't expose via a CLI, since we only really added it for the sake of `prepare`. The consequence of _that_ is that we don't need a `svelte-kit` CLI at all — we could move `svelte-kit package` into a separate package, which would make various things quite a bit easier.

dominikg:
That's a great idea, not going through the cli script for postinstall could make this easier.

Unfortunately I think we need to keep the sync command in a cli for cases where postinstall isnt run.

Rich-Harris:
Do we though? The only reason we included `sync` in the first place was so that people weren't faced with red squigglies in a new project — they can `npm i && npm run check` and the project will typecheck appropriately, which is handy for CI in some cases. But in any situation where you can run `svelte-kit sync` manually, you can also run sync via `vite dev` or `vite build`. I can't imagine there's a situation where you absolutely need to `sync`, but can't `build`, after you've already successfully installed the project

dominikg:
1) there is `npm install --ignore-scripts` so  postinstall is not guaranteed to be executed.
2) `dev` and `build` are a lot heavier than `sync` alone.
3) i think @dummdidumm and @jasonlyu123  recently discussed that the language server might want to call sync
4) our own repo CI currently fails because postinstall for workspace deps isn't called consistently in a pnpm workspace, which means we'd have to build all project for the lint task.

packages/kit/svelte-kit.js
```diff
@@ -1,2 +1,37 @@
...
+	if (!is_postinstall) {
+		console.warn(
+			`Your project at ${cwd} does not have a svelte.config.js  — skipping svelte-kit sync`
+		);
+	}
+} else {
+	const cliPath = is_workspace
+		? fileURLToPath(new URL('./src/cli.js', import.meta.url))
+		: '@sveltejs/kit/dist/cli.js';
+	await import(cliPath);
```
Rich-Harris, line 36: this should work, no?

```suggestion
	await import(
		fileURLToPath(new URL(`./${is_workspace ? 'src' : 'dist'}/cli.js`, import.meta.url))
	);
```

Rich-Harris: (otherwise it complains about `./dist/cli.js` not being in `pkg.exports`)

Rich-Harris:
I'll be honest, I don't understand what any of the new stuff in `packages/kit/svelte-kit.js` is for. Our CI workflow already ensures that SvelteKit is built, can't we get rid of all the `is_workspace` stuff?

Rich-Harris:
merged #5760 so will close this

Issue #5247, opened 2022-06-22
__layout props update on every route change


CaptainCodeman:
### Describe the bug

Not sure if this is a bug or should just be a discussion, but it was a surprise. When loading some data _once_ in a parent __layout I expected that data to then remain static. But the props appear to be re-assigned on every navigation, even for querystring parameter changes. This plays havoc with any `use:actions` which think things have been updated when they haven't (which may contain image gallery layout code, canvas setup etc... that I only want to re-run when the data actually changes)

### Reproduction

Simplified example to show the behavior:
https://github.com/CaptainCodeman/sveltekit-layout-props-example

Run the project and navigate around. Note the prop changes and action update console logs, even though the `load` function has only executed once:

<img width="484" alt="Screen Shot 2022-06-22 at 12 38 08 PM" src="https://user-images.githubusercontent.com/304910/175111993-c175289f-cdb8-4146-ad68-cc045420c05a.png">

I wouldn't expect the `data` prop to keep changing or for the `use:action` to keep being updated, but they are in both the __layout and an imported component.

### System Info

```shell
System:
    OS: macOS 12.4
    CPU: (6) x64 Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
    Memory: 335.20 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.14.0 - /usr/local/bin/node
    npm: 8.3.1 - /usr/local/bin/npm
  Browsers:
    Brave Browser: 102.1.39.111
    Chrome: 102.0.5005.115
    Chrome Canary: 105.0.5134.0
    Edge: 102.0.1245.41
    Firefox: 101.0.1
    Safari: 15.5
    Safari Technology Preview: 16.0
  npmPackages:
    @sveltejs/adapter-static: next => 1.0.0-next.34 
    @sveltejs/kit: next => 1.0.0-next.352 
    svelte: ^3.44.0 => 3.48.0
```

### Severity

annoyance

CaptainCodeman:
NOTE: the `data` is the same object reference each time, so I'm actually kind of surprised that it updates as it does but I guess it's doing it on the assignment operation and ignoring any reference check (which I thought Svelte did to avoid unnecessary updates).

I can get it to give the expected behavior by checking for the reference change myself. i.e. return dataraw from the load function and check for it differing from a local variable:

``` js
export let dataraw

let data

$: if (data !== dataraw) data = dataraw
```

[...]

Issue #2301, opened 2021-08-26
Rethink or clarify when `load` runs


benmccann:
### Describe the bug

The docs say:

>A component that defines a page or a layout can export a load function that runs before the component is created.

But I don't think that is when `load` runs? I think `load` is reactive to its input parameters and runs everytime they change

This is confusing to people. E.g. they don't expect it to run when `$session` changes https://github.com/sveltejs/kit/issues/2252. 

Also, sometimes `load` in your layout will run whenever the path changes and sometimes it will load only when the page first loads, which is super hard to discover. This also leads to weird workarounds like accessing `$page.path` to cause `load` to run eventhough you don't need to do anything with it (https://github.com/sveltejs/kit/issues/1165#issuecomment-829615802)

This all feels a little too magical to me. It's not very clear that it will happen (vs if I did something like `$: load = function(page) {}`. Maybe this is just how Svelte works, but it's not obvious to me. The Svelte docs don't have any stores as function parameters, and I never did something like `$page.subscribe(load)`.

We also have `page` accessible from two places: the `load` parameters and a store. Is there a reason we have two ways of doing the same thing? Does `load` only re-run when we access `page` via one of those locations or via both?

We've also gotten a request to not have `load` run on every URL change (https://github.com/sveltejs/kit/issues/2673)

At the very least we need to update the docs for `load`, but I wonder if we can't make things more explicit.

Part of the difficulty here is that components aren't recreated for each page view and so sometimes you get a new page component created and sometimes you don't. Given that, we sort of have to rerun `load` when `$page` changes. This is somewhat related to the transition bugs that we have. The reason we don't create new components for each page view is so that when viewing multiple pages rendered by the same svelte file you can have transitions - though I'm not sure that we couldn't accomplish that other ways

This also feels slightly related to the before / after hooks for routing (https://github.com/sveltejs/kit/issues/552) because some people try to put code that will run on every page navigation on `load` in `__layout.svelte`, but you might also be able to accomplish some of those use cases with routing hooks.

### Reproduction

n/a

### System Info

```shell
n/a
```

### Severity

serious, but I can work around it

rmunn:
[...]

Issue #2252, opened 2021-08-20
What's the point of rerunning the `load` function on `session` change?


csangonzo:
### Describe the problem

It's more like a genuine question about a new feature in sveltekit. I just noticed this in the docs:
```
The load function is reactive, and will re-run when its parameters change, but only if they are used in the function.
```
After login when I set `$session.user = response.user;` like in the realworld demo (https://github.com/sveltejs/realworld/blob/master/src/routes/login/index.svelte) I see that the login page's `load` function runs again.
I use access and refresh tokens in my app and I implemented a silent refresh method, something like the one mentioned here: https://hasura.io/blog/best-practices-of-using-jwt-with-graphql/ .
When I do a silent refresh I reset the session, but this causes a rerun of the load function now.
The problem with this is I have a lot of components that use data returned by the load function (like an editor that gets it's content first from the load function, or a form that has it's model from the load function) and if it runs again the data changes on the page, so if I had unsaved changes in my editor for example and my silent refresh runs, it loads the data again from the server so my changes will be lost.

### Describe the proposed solution

If someone could explain the benefits we get from this feature I'd be really glad, because right now this is the first thing that doesn't make any sense about sveltekit for me.

### Importance

i cannot use SvelteKit without it

### Additional Information

I follow the changes of sveltekit from day to day, last year I decided to ditch angular in favour of svelte and migrated almost every app to it that we had and I'm having problems with the authentication part because of this change. I'm open to alternative solutions as well.

dummdidumm:
This is beneficial because you are then able to refresh your page based on changed parameters. I can see that this behavior is sometimes not beneficial, not sure for the best way to handle this. 

Random things I just came up with:
- make the `load` function return a parameter like (`noFurtherUpdates: true`) to signal SvelteKit to not refresh again
- another option similar to `srr`, like `rerunLoad`, which can be configured globally and be overwritten on a page level (`export const rerunLoad = true`

Random thoughts on how to work around it right now:
- don't update the session store / keep the token somewhere else
- don't reference `session` in the pages, if possible (load won't rerun then)
- restructure how page props overwrite unsafed stuff

csangonzo:
Thanks for the response!
The real problem comes when I validate every page with session's user property, so I check `$session.user`, if it exists the user can stay on the page, if not then redirect.
I also use my access token in the load function (on the server and on the client as well) and it's in the session's user property. Since load can run on the client and on the server side, it was my only solution to store the token in.
[...]

PR #5654, opened 2022-07-21
[fix] Prevent rerender when route state did not change

### LINKED ISSUE #3732: page props are re-set whenever the `session` store is updated, and the `load` function is re-run the first time
### Describe the bug

In the docs, in the section about the `load` function, it's stated that:
> ...if url, session or stuff are used in the function, they will be re-run whenever their value changes... [(here)](https://kit.svelte.dev/docs#:~:text=if%20url%2C%20session%20or%20stuff%20are%20used%20in%20the%20function%2C%20they%20will%20be%20re%2Drun%20whenever%20their%20value%20changes)

But this doesn't seem to be the case. I have the following page in my SvelteKit app:
`index.svelte`:

```svelte
    import type { Load } from '@sveltejs/kit';

    export const load: Load = () => {
        console.warn('index load function');
        return {};
    }
</script>

<script lang="ts">

    function changeSession() {
        $session.someField = Math.random();
    }
</script>

<pre>{JSON.stringify($session, null, 4)}</pre>

<button on:click={changeSession}>
    Change $session
</button>
```

If you click on the "Change $session" button, you'll see the message "index load function" appearing in the console, which obviously means the load function is re-run, even though `session` isn't used in the `load` function. This is inconsistent with the behavior described in the docs.

The strange thing, however, is that from the second time onwards, when the value of the '$session' store changes, the `load` function is no longer re-run. This is confusing.

**The more severe bug**, however, which seems to be related to this one, is that every time the `$session` store changes, the page's props are apparently re-set, which means reactivity is re-triggered throughout the entire page, potentially causing massive problems.

Consider the following:
```svelte
    import type { Load } from '@sveltejs/kit';

    export const load: Load = () => {
        console.warn('index load function');
        return {
            props: {
                userBasicInfo: {
                    id: 2,
                }
            }
        };
    }
</script>

<script lang="ts">
    import { browser } from '$app/env';

    export let userBasicInfo;

    function changeSession() {
        $session.someField = Math.random();
    }

    $: console.log('userBasicInfo changed:', userBasicInfo);
</script>

<pre>{JSON.stringify($session, null, 4)}</pre>

<button on:click={changeSession}>
    Change Ssession
</button>

<hr />

{#if browser}
    {#await fetch(`https://reqres.in/api/users/${userBasicInfo.id}`).then(r => r.json())}
        Loading...
    {:then result}
        <pre>{JSON.stringify(result, null, 4)}</pre>
    {/await}
{/if}
```

In this setup, whenever you click on the "Change $session" button — which is to say whenever you update the `$session` store — the `fetch` request in the markup, which uses the `userBaicInfo` prop, is re-triggered, even though it obviously shouldn't. And you can see the `userBasicInfo changed: ...` message in the console which proves that it's because of the `userBasicInfo` prop getting re-set.

### Reproduction

You can create an empty SvelteKit skeleton project and then copy-paste the two code snippets above into the `index.svelte` file, and then `npm run dev` => open your browser and see the behavior for yourself.

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19042
    Memory: 1.09 GB / 7.89 GB
  Binaries:
    Node: 14.15.0 - C:\Program Files\nodejs\node.EXE     
    npm: 6.14.8 - C:\Program Files\nodejs\npm.CMD        
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (97.0.1072.76)
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    svelte: ^3.44.0 => 3.46.4
```

### Severity

### Additional Information

_No response_

dummdidumm:
Fixes #3732 (and maybe some other issue? Could have sworn I saw something similar about props running in another issue)

No test yet, want to confirm first that this is the right fix

### Tests

### Changesets

### DIFF
--- a/.changeset/red-rivers-approve.md
+++ b/.changeset/red-rivers-approve.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Prevent rerender when route state did not change
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -650,6 +650,7 @@ export function create_client({ target, session, base, trailing_slash }) {
 
 		/** @type {Array<import('./types').BranchNode | undefined>} */
 		let branch = [];
+		let is_previous_branch = true;
 
 		/** @type {Record<string, any>} */
 		let stuff = root_stuff;
@@ -686,6 +687,7 @@ export function create_client({ target, session, base, trailing_slash }) {
 					(stuff_changed && previous.uses.stuff);
 
 				if (changed_since_last_render) {
+					is_previous_branch = false;
 					/** @type {Record<string, any>} */
 					let props = {};
 
@@ -824,6 +826,12 @@ export function create_client({ target, session, base, trailing_slash }) {
 
 				branch.push(node);
 			}
+		}
+
+		if (is_previous_branch) {
+			// Nothing changed, don't rerender
+			// We can end up in here when $session is updated but not used in load
+			return;
 		}
 
 		return await get_navigation_result_from_branch({
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -117,6 +117,9 @@ export function create_client({ target, session, base, trailing_slash }) {
 		if (!ready) return;
 		session_id += 1;
 
+		const current_load_uses_session = current.branch.some((node) => node?.uses.session);
+		if (!current_load_uses_session) return;
+
 		update(new URL(location.href), [], true);
 	});
 	ready = true;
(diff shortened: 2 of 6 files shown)


Rich-Harris:
Ah, interesting. We definitely don't want to re-run `load` in that case, yeah. That said, there are cases where you _do_ want to trigger a navigation even though nothing changed — think of e.g. you're on `/` and you click the home icon in the top left, which should cause data to be re-fetched and the scroll to reset.

We need to separate those cases. Related: #2301

dummdidumm:
True, there's a difference between an implicit update through a session update an an explicit user interaction.
Also related:  #2252

dummdidumm:
I adjusted the code so that only the implicit `$session` update no longer initiates a rerender and added a test. I also fixed another test which I believe tested the wrong expectation (page store updating on session change, which it shouldn't). Maybe this can also be fixed differently while also fixing #5247 by somehow not telling Svelte to update specific components, will look into that.

Rich-Harris (approved):

PR #5655, opened 2022-07-21
(hopefully) fix a flaky test

Rich-Harris:
this is one of our flakier tests — i think i figured out why

### DIFF
--- a/packages/kit/test/apps/basics/test/client.test.js
+++ b/packages/kit/test/apps/basics/test/client.test.js
@@ -415,6 +415,9 @@ test.describe('Page options', () => {
 		await page.click('button');
 		expect(await page.textContent('button')).toBe('clicks: 1');
 
+		// wait until router has started before attempting backwards client-side navigation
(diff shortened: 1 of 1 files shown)


Rich-Harris:
alright, zero flakes on this test. still a couple of others but this was the worst I think

packages/kit/test/apps/basics/test/client.test.js
```diff
@@ -412,9 +412,15 @@ test.describe('Page options', () => {
...
 		expect(await page.textContent('button')).toBe('clicks: 0');
 
+		// wait until hydration before interacting with button
+		await page.waitForSelector('body.started');
+
 		await page.click('button');
 		expect(await page.textContent('button')).toBe('clicks: 1');
 
+		// wait until hydration before attempting backwards client-side navigation
+		await page.waitForSelector('body.started');
```
benmccann, line 422: Why would we need to do this again? Didn't it already happen with the line you added above?

PR #5657, opened 2022-07-21
replace transformPage with transformPageChunk

### LINKED ISSUE #4910: replace `transformPage` with `transformPageChunk` or similar
### Describe the problem

SvelteKit should have the ability to stream HTML. That could take several forms:

1. We immediately flush the contents of `src/app.html` up to `%svelte.head%`, plus the `<link>` elements that preload .js and .css files for the page (which we could determine as soon as the route is matched with some fairly minor changes), then flush the rest all in one go once the page is done rendering. This improves TTFB and lets the browser get a head start on fetching resources
2. We immediately flush the top (as in 1), then render a placeholder element for any `<svelte:head>` content discovered during rendering, then flush up to `%svelte.body%`, then flush the rest once the page renders
3. We do 2, but instead of rendering the whole body synchronously, we render layouts as soon as they're ready, meaning that in many cases we could flush a `<nav>` bar etc before the main content is ready

There's also a version where Svelte itself supports streaming rendering, and we pause on `{#await ...}` blocks, but that's a far more involved conversation that we don't need to have today.

In fact, we don't need to decide _any_ of this stuff today; we just need to identify the things that block us from implementing these ideas. 3 is impossible as long as we have `$page.stuff` (more on that in a separate issue), but all three versions are impossible as long as we have the `transformPage` API since it expects the entire page to be buffered.

Instead of `transformPage`, we could have something like `transformPageChunk` or `transformMarkup` which operates on a chunk at a time. In the common case, it would behave exactly the same:

```js
  return resolve(event, {
    transformMarkup: ({ html }) => html.replace(/cloud/g, 'butt')
  });
}
```

We can guarantee that the chunks are sensible (the entirety of a layout up to the `<slot/>` or following it, for example) so that it's not necessary to handle the case where one chunk contains 'clo' and the next contains 'ud'. That said, in some rare cases it might be necessary to buffer content, which means we would need to include a flag for the final chunk:

```js
  let buffer = '';
  return resolve(event, {
    transformMarkup: ({ html, last }) => {
      if (last) return minify(buffer + html);
      else buffer += html;
    }
  });
}
```

(Realistically, you'd probably just disable streaming in those situations, but `last` costs nothing.)

Until streaming is implemented (which probably won't happen straight away as it involves lots of decisions — is it driven by config? by a `resolve` option? do we do version 1, 2, or 3?) this function would behave exactly the same way as `transformPage` currently does.

### Alternatives considered

Keep the current API, but opt out of streaming when `transformPage` is provided. I think that would be a mistake; there are lots of valid uses for transforming markup that are no less valid in a streaming context.

### Importance

would make my life easier

### Additional Information

Aside: streaming implies everything gets a 200 response except early 404s. Apparently the way to make this not matter is to inject a `<meta name="robots" content="noindex">` into the page — doesn't matter where — if you encounter an error during rendering.

Another aside: out-of-order streaming (as suggested in 2, where we inject a placeholder element for `<svelte:head>` content, then render snippets of code that inject the HTML into the head once it's ready) relies on JavaScript. It should therefore a) be behind an option, and b) be disabled for search engine bots, since reliance on JS is bad for SEO despite some alarmingly common misconceptions. (We would probably need to make the list of bot UAs configurable.)

Rich-Harris:
Closes #4910.

This replaces the `transformPage` API with `transformPageChunk`, so that we can respond in a streaming fashion in future. We're not yet taking advantage of that ability, but updating the API is the first step.

Previously, `transformPage` took the entire HTML document:

```js
  return resolve(event, {
    transformPage: ({ html }) => html.replace('old', 'new')
  });
}
```

Now, it takes a _chunk_ of HTML plus a `done` boolean. Chunks are not guaranteed to be well-formed HTML (they might include the opening tag of an element but not its closing tag, for example) but nor are they arbitrary — they will always be split at sensible boundaries (like `%sveltekit.[property]%` or before the contents of a layout or page component). Because of that, many transforms will work identically:

```diff
export function handle({ event, resolve }) {
  return resolve(event, {
-    transformPage: ({ html }) => html.replace('old', 'new')
+    transformPageChunk: ({ html }) => html.replace('old', 'new')
  });
}
```

In some cases, it's necessary to operate on the entire document in one go — for example, to validate and transform an AMP document. In these situations it's easy to buffer the HTML and return the transformed document at the end:

```js
  let buffer = '';
  return resolve(event, {
    transformPageChunk: ({ html, done }) => {
      buffer += html;
      if (done) return amp.transform(html);
    }
  });
}
```

Right now, `done` will always be true (i.e. there'll only be one chunk, which is the entire document) but that will change in the near future.

### Tests

### Changesets

### DIFF
--- a/documentation/docs/06-hooks.md
+++ b/documentation/docs/06-hooks.md
@@ -67,7 +67,7 @@ You can add call multiple `handle` functions with [the `sequence` helper functio
 `resolve` also supports a second, optional parameter that gives you more control over how the response will be rendered. That parameter is an object that can have the following fields:
 
 - `ssr: boolean` (default `true`) — if `false`, renders an empty 'shell' page instead of server-side rendering
-- `transformPageChunk(opts: { html: string, done: boolean }): string` — applies custom transforms to HTML. If `done` is true, it's the final chunk
+- `transformPageChunk(opts: { html: string, done: boolean }): MaybePromise<string | undefined>` — applies custom transforms to HTML. If `done` is true, it's the final chunk. Chunks are not guaranteed to be well-formed HTML (they could include an element's opening tag but not its closing tag, for example) but they will always be split at sensible boundaries such as `%sveltekit.head%` or layout/page components.
 
 ```js
 /// file: src/hooks.js
--- a/documentation/docs/06-hooks.md
+++ b/documentation/docs/06-hooks.md
@@ -67,7 +67,7 @@ You can add call multiple `handle` functions with [the `sequence` helper functio
 `resolve` also supports a second, optional parameter that gives you more control over how the response will be rendered. That parameter is an object that can have the following fields:
 
 - `ssr: boolean` (default `true`) — if `false`, renders an empty 'shell' page instead of server-side rendering
-- `transformPage(opts: { html: string }): string` — applies custom transforms to HTML
+- `transformPageChunk(opts: { html: string, done: boolean }): string` — applies custom transforms to HTML. If `done` is true, it's the final chunk
 
 ```js
 /// file: src/hooks.js
@@ -75,7 +75,7 @@ You can add call multiple `handle` functions with [the `sequence` helper functio
 export async function handle({ event, resolve }) {
 	const response = await resolve(event, {
 		ssr: !event.url.pathname.startsWith('/admin'),
-		transformPage: ({ html }) => html.replace('old', 'new')
+		transformPageChunk: ({ html }) => html.replace('old', 'new')
 	});
 
 	return response;
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -171,7 +171,7 @@ export async function respond(request, options, state) {
 	/** @type {import('types').RequiredResolveOptions} */
 	let resolve_opts = {
 		ssr: true,
-		transformPage: default_transform
+		transformPageChunk: default_transform
 	};
 
 	// TODO match route before calling handle?
@@ -181,9 +181,17 @@ export async function respond(request, options, state) {
 			event,
 			resolve: async (event, opts) => {
 				if (opts) {
+					// TODO remove for 1.0
+					// @ts-expect-error
+					if (opts.transformPage) {
+						throw new Error(
+							'transformPage has been replaced by transformPageChunk — see https://github.com/sveltejs/kit/issues/4910 for more information'
+						);
+					}
+
 					resolve_opts = {
 						ssr: opts.ssr !== false,
-						transformPage: opts.transformPage || default_transform
+						transformPageChunk: opts.transformPageChunk || default_transform
 					};
 				}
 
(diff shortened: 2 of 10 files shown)


packages/kit/src/runtime/server/index.js
```diff
@@ -181,9 +181,17 @@ export async function respond(request, options, state) {
 			event,
 			resolve: async (event, opts) => {
 				if (opts) {
+					// TODO remove for 1.0
+					// @ts-expect-error
+					if (opts.transformPage) {
+						throw new Error(
+							'transformPage has been replaced by transformPageChunk — see https://github.com/sveltejs/kit/issues/4910 for more information'
```
Rich-Harris, line 188: point to this PR, rather than the issue

```suggestion
							'transformPage has been replaced by transformPageChunk — see https://github.com/sveltejs/kit/pull/5657 for more information'
```

benmccann:
It looks like all the examples are buffering the whole page before dealing with it, so I wonder if we can't have a `transformPage` still just with the clearly documented caveat that if you use it, you lose streaming and buffer the whole page? I haven't looked into the implementation details to know if that's possible or not

documentation/docs/06-hooks.md
```diff
@@ -67,15 +67,15 @@ You can add call multiple `handle` functions with [the `sequence` helper functio
 `resolve` also supports a second, optional parameter that gives you more control over how the response will be rendered. That parameter is an object that can have the following fields:
 
 - `ssr: boolean` (default `true`) — if `false`, renders an empty 'shell' page instead of server-side rendering
-- `transformPage(opts: { html: string }): string` — applies custom transforms to HTML
+- `transformPageChunk(opts: { html: string, done: boolean }): string` — applies custom transforms to HTML. If `done` is true, it's the final chunk
```
benmccann, line 70: I don't know if there's anything we should say here, but I'm not that familiar with how the chunking happens. Does it match a single write to the stream? Or is it split up so that each chunk is less than the TCP window size or something? E.g. I'm wondering if there's any risk that the `.replace('old', 'new')` example doesn't work because "ol" is one chunk and "d" in the next chunk

Rich-Harris: I addressed that above:

> Chunks are not guaranteed to be well-formed HTML (they might include the opening tag of an element but not its closing tag, for example) but nor are they arbitrary — they will always be split at sensible boundaries (like `%sveltekit.[property]%` or before the contents of a layout or page component)

I figured there wasn't much point going into detail in the docs until we actually start streaming responses

benmccann: I guess I should read PR descriptions :smile: Still, it might be worth including here so that people write their code correctly from the beginning and so that we don't break their applications when we do start streaming

Rich-Harris: added some detail

dummdidumm (approved):

PR #5659, opened 2022-07-21
[chore] simplify internal prerender API

benmccann:
Remove `entries` and `files` from `prerender` function since they're difficult to construct entities. This will make https://github.com/sveltejs/kit/issues/5306 much easier to implement

### DIFF
--- a/.changeset/soft-humans-cheer.md
+++ b/.changeset/soft-humans-cheer.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[chore] simplify internal prerender API
--- a/.changeset/soft-humans-cheer.md
+++ b/.changeset/soft-humans-cheer.md
@@ -1,5 +1,0 @@
----
-'@sveltejs/kit': patch
----
-
-[chore] simplify internal prerender API
--- a/packages/kit/src/core/prerender/prerender.js
+++ b/packages/kit/src/core/prerender/prerender.js
@@ -1,5 +1,5 @@
-import { readFileSync, writeFileSync } from 'fs';
-import { dirname, join } from 'path';
+import { readdirSync, readFileSync, statSync, writeFileSync } from 'fs';
+import { dirname, join, relative } from 'path';
 import { pathToFileURL, URL } from 'url';
 import { mkdirp } from '../../utils/filesystem.js';
 import { installPolyfills } from '../../node/polyfills.js';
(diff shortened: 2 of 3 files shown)


packages/kit/src/core/prerender/prerender.js
```diff
@@ -1,5 +1,5 @@
-import { readFileSync, writeFileSync } from 'fs';
-import { dirname, join } from 'path';
+import { readdirSync, readFileSync, statSync, writeFileSync } from 'fs';
+import { dirname, join, relative } from 'path';
 import { pathToFileURL, URL } from 'url';
 import { mkdirp } from '../../utils/filesystem.js';
```
Rich-Harris, line 4:
```suggestion
import { mkdirp, walk } from '../../utils/filesystem.js';
```

packages/kit/src/core/prerender/prerender.js
```diff
@@ -160,6 +160,7 @@ export async function prerender({ config, entries, files, log }) {
 		return file;
 	}
 
+	const files = get_files(client_out_dir);
```
Rich-Harris, line 163:
```suggestion
	const files = new Set(walk(client_out_dir));
```

packages/kit/src/core/prerender/prerender.js
```diff
@@ -351,3 +358,21 @@ export async function prerender({ config, entries, files, log }) {
...
+	for (const file of files) {
+		const name = join(curr_dir, file);
+		if (statSync(name).isDirectory()) {
+			get_files(dir, name).forEach((f) => result.add(f));
+		} else {
+			result.add(relative(dir, name));
+		}
+	}
+	return result;
+}
```
Rich-Harris, line 378:
```suggestion
```

Rich-Harris (commented):
there's an existing `walk` utility we can use

.changeset/soft-humans-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[chore] simplify internal prerender API
```
Conduitry, line 5: It doesn't sound like this should actually need a changeset, right? It's just an internal refactoring.

benmccann: I was debating. I figured if anything broke on a version with this update it'd be easier to track it down with the changeset, but users should not need to care about it otherwise

packages/kit/src/core/prerender/prerender.js
```diff
@@ -160,7 +160,7 @@ export async function prerender({ config, client_out_dir, manifest_path, log })
 		return file;
 	}
 
-	const files = new Set(walk(client_out_dir));
+	const files = new Set(walk(client_out_dir).map((f) => posixify(f)));
```
Rich-Harris, line 163:
```suggestion
	const files = new Set(walk(client_out_dir).map(posixify));
```

.changeset/soft-humans-cheer.md
Rich-Harris: We don't normally add changesets for internal-only changes — are you saying we should start requiring a changeset for every PR?

benmccann: I wasn't sure if we had a policy around it. I'll drop this file

Rich-Harris (approved):

PR #5663, opened 2022-07-21
feat: Consistent handling of environment variables

### LINKED ISSUE #4296: Consistent handling of environment variables
### Describe the problem

SvelteKit doesn't do anything special with environment variables; it delegates to Vite. This is good for the sorts of environment variables Vite deals in — `VITE_` prefixed values that are replaced at build time — but less good for other kinds.

Environment variables can be **static** (known at build time) or **dynamic** (not known until run time), and **private** (only accessible to code that runs on the server) or **public** (accessible to code that runs on the server or the client):

|                        | static               | dynamic                                  |
|------------------------|--------------------------|-------------------------------------------|
| public | ✅ import.meta.env.VITE_* | ❌ (1)                                         |
| private  | ⚠️ can be done with plugins (2)                     | ⚠️ process.env.* (3) |

1. To be fair, I'm not sure what sorts of things would go in this quadrant, though I would be eager to hear examples
2. You can add something like [plugin-replace](https://github.com/rollup/plugins/tree/master/packages/replace#readme) to selectively replace things at build time, but this is ad hoc, undiscoverable, and involves duplication between `.env` files and config
3. `process.env` is only available in Node-based environments (e.g. `adapter-node`, `adapter-netlify` and `adapter-vercel` — though once the Vercel adapter starts building Edge Functions that will no longer be the case)

It would be great if we had a consistent and idiomatic way of dealing with all of these.

I don't actually have any proposals to make, I just didn't want https://github.com/sveltejs/kit/pull/4293#issuecomment-1065300764 to get lost. Whatever solution we land on should

* work with `.env`/`.env.*` files
* enable dead code elimination etc at build time
* allow autocompletion/auto-imports (e.g. we could generate typings from your `.env` file)
* feel consistent across the four quadrants, but with clearly understandable distinctions between them (e.g. `$app/env/public` vs `$app/env/private` or whatever)

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

tcc-sejohnson:
Will close #4296 when finished.

This is a rough draft of the static environment variables solution. It adds a new config option: `config.kit.env.publicPrefix`, which defaults to `PUBLIC_`. Running `svelte-kit sync` will now dump all `publicPrefix`-prefixed environment variables to `$app/env`, and all variables _not_ prefixed to `$app/env/private`. It also creates types for these files (for autocompletion).

Additionally, it adds a neat new feature that prevents importing `$app/env/private` into any client-side code by analyzing the import module graph during both `dev` and `build`. Initially, I had concerns around catching dynamic imports, but it seems this solution does a great job at that.

There are some rough edges -- I _especially_ hate the amount of duplication between the Rollup and Vite algorithms in `vite/utils.js`. I'm sure there's a smarter way to structure that, but I've been screwing around with import graphs for too long right now to see it.

Update: Added support for the runtime variables part of the proposal.

Runtime variables are now exposed via `import { env } from '$app/env/runtime'`. `env` is typed as `App.RuntimeEnv`, which is just a convenience for developers (they can define what they expect the runtime env to be in order to provide intellisense when importing the module). This module is populated by calling `set_env`, which is itself called from `server.init`. Adapters are responsible for calling `server.init` when it is appropriate. For `adapter-node` (which I have updated as part of this PR for testing), this would be on startup.

### Tests

### Changesets


packages/kit/src/core/sync/write_env.js
```diff
@@ -0,0 +1,117 @@
...
+ */
+function const_declaration_template(pub, env) {
+	return `${autogen_comment}
+
+${Object.entries(env)
+	.map(
+		([k, v]) => `/**
+ * @type {import('$app/env${pub ? '' : '/private'}').${k}}
+ */
+export const ${k} = "${escape(v)}";
```
Rich-Harris, line 21: can this just use `JSON.stringify`?

```suggestion
export const ${k} = ${JSON.stringify(v)};
```

or even

```suggestion
export const ${k} = ${s(v)};
```

if we import it from https://github.com/sveltejs/kit/blob/master/packages/kit/src/utils/misc.js

packages/kit/src/core/sync/write_env.js
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * @param {boolean} pub
+ * @param {Record<string, string>} env
+ * @returns {string}
+ */
+function type_declaration_template(pub, env) {
+	return `declare module '$app/env${pub ? '' : '/private'}' {
+  ${Object.entries(env)
+		.map(([k]) => `export const ${k}: string;`)
```
Rich-Harris, line 36:
```suggestion
  ${Object.keys(env)
		.map((k) => `export const ${k}: string;`)
```

tcc-sejohnson: lol thanks

Originally I needed the pairs and my monkey brain apparently has 0 neuroplasticity 😂

packages/kit/src/core/sync/write_env.js
tcc-sejohnson: You know, I don't think I even knew you could pass anything but an object/array to `JSON.stringify`. I think that should work fine since it always wraps strings in double quotes.

tcc-sejohnson:
Added runtime variable support.

packages/kit/src/core/sync/write_env.js
```diff
@@ -0,0 +1,183 @@
...
+
+function runtime_env_template() {
+	return `/** @type {App.RuntimeEnv} */
+export let env = {};
+
+/** @type {(environment: Record<string, string>) => void} */
+export function set_env(environment) {
+  env = environment;
+}`;
+}
```
Rich-Harris, line 49: this could probably just be a file in https://github.com/sveltejs/kit/tree/master/packages/kit/src/runtime/app, since its content never changes

packages/kit/src/core/sync/write_env.js
```diff
@@ -0,0 +1,183 @@
...
+ * @param {Record<string, string>} env
+ * @returns {string}
+ */
+function const_declaration_template(pub, env) {
+	return `${autogen_comment}
+
+${Object.entries(env)
+	.map(
+		([k, v]) => `/**
+ * @type {import('$app/env${pub ? '' : '/private'}').${k}}
```
Rich-Harris, line 19: genuine question — i'm sure there's a good reason — but what's the reason to do this over just `string`?

```suggestion
 * @type {string}
```

packages/kit/src/core/sync/write_env.js
```diff
@@ -0,0 +1,183 @@
...
+ * in process.env to $app/env
+ * @param {import('types').ValidatedKitConfig} config
+ * @param {string | undefined} mode
+ * The Vite mode.
+ */
+function write_public_env(config, mode = undefined) {
+	const pub_out = path.join(config.outDir, 'runtime/app/env.js');
+
+	// public is a little difficult since we append to an
+	// already-existing file
```
Rich-Harris, line 76: what if we changed it to `$app/env/public`? it's arguably cleaner anyway

packages/kit/src/core/sync/write_env.js
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * Writes the type definitions for the environment variable files
+ * to types/ambient.d.ts
+ * @param {import('types').ValidatedKitConfig} config
+ * @param {Record<string, string>} pub
+ * @param {Record<string, string>} priv
+ */
+function write_typedef(config, pub, priv) {
+	const type_declaration_out = path.join(config.outDir, 'types/ambient.d.ts');
```
Rich-Harris, line 128: mostly unrelated to this PR, but this gets me thinking: maybe we should reference the Kit types in this ambient file, rather than here: https://github.com/sveltejs/kit/blob/ab65de0c065df5e274082805bc379186eb6c15d3/packages/create-svelte/templates/default/src/app.d.ts#L1

Tidier, and one less thing to break in userland. Or maybe even that's too complicated and we can reference those types directly in the generated config? not sure

packages/kit/src/core/sync/write_env.js
```diff
@@ -0,0 +1,183 @@
...
+ * @param {import('types').ValidatedKitConfig} config
+ * @param {string | undefined} mode
+ * The Vite mode.
+ */
+function write_public_env(config, mode = undefined) {
+	const pub_out = path.join(config.outDir, 'runtime/app/env.js');
+
+	// public is a little difficult since we append to an
+	// already-existing file
+	const pub = resolve_public_env(config.env.publicPrefix, mode);
```
Rich-Harris, line 77:
```suggestion
	const pub = loadEnv(mode, process.cwd(), config.env.publicPrefix);
```

packages/kit/src/core/sync/write_env.js
```diff
@@ -0,0 +1,183 @@
...
+	// but I'm also smart enough not to force
+	// someone else to read this with reduce later
+	Object.entries(env)
+		.filter(([k]) => (pub ? k.startsWith(public_prefix) : !k.startsWith(public_prefix)))
+		.forEach(([k, v]) => {
+			resolved[k] = v ?? '';
+		});
+
+	return resolved;
+}
```
Rich-Harris, line 183: i'm 99% sure we can just replace all this with `loadEnv` above

```suggestion
```

packages/kit/src/vite/build/build_server.js
```diff
@@ -88,6 +90,14 @@ export class Server {
 		};
 	}
 
+  init(options = { env: {} }) {
+    if (!initialized) {
+      set_env(options.env);
+      initialized = true;
+    }
+    return Promise.resolve();
```
Rich-Harris, line 98: what's this for?

```suggestion
```

packages/kit/src/vite/dev/index.js
```diff
@@ -235,6 +240,16 @@ export async function dev(vite, vite_config, svelte_config) {
...
 
+				/** @type {(env: Record<string, string>) => void} */
+				const set_env = (await import(`${runtime}/app/env/runtime.js`)).set_env;
+
+				set_env(
+					// This just makes sure no values are undefined
+					Object.entries(process.env).reduce((prev, [k, v]) => {
+						return { ...prev, [k]: v ?? '' };
+					}, {})
+				);
```
Rich-Harris, line 251: Is that a thing that can happen? I've never seen a value in `process.env` that wasn't a `string`

```suggestion
				set_env(process.env);
```

packages/kit/src/vite/preview/index.js
```diff
@@ -42,6 +42,12 @@ export async function preview(vite, config, protocol) {
 	});
 
 	const server = new Server(manifest);
+	server.init({
+		// This just makes sure no values are undefined
+		env: Object.entries(process.env).reduce((prev, [k, v]) => {
+			return { ...prev, [k]: v ?? '' };
+		}, {})
```
Rich-Harris, line 49: same question as above

```suggestion
		env: process.env
```

packages/kit/src/vite/dev/index.js
Rich-Harris: Looking at [`@types/node`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e296423148c71102ca1f05c6d632453c9e6e5189/types/node/process.d.ts#L112-L117)...

```ts
interface ProcessEnv extends Dict<string> {
    /**
     * Can be used to change the default timezone at runtime
     */
    TZ?: string;
}
```

...it seems that values [_can_ be undefined](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e296423148c71102ca1f05c6d632453c9e6e5189/types/node/globals.d.ts#L287-L289) as far as TypeScript is concerned:

```ts
interface Dict<T> {
    [key: string]: T | undefined;
}
```

But I've never actually seen that be the case, which makes me think maybe the types are wrong

packages/kit/src/cli.js
```diff
@@ -38,7 +38,12 @@ prog
 prog
 	.command('sync')
 	.describe('Synchronise generated files')
-	.action(async () => {
+	.option(
+		'--mode',
+		'Specify a mode for loading environment variables (default: "development")',
+		undefined
```
Rich-Harris, line 44:
```suggestion
		'--mode',
		'Specify a mode for loading environment variables',
		'development'
```

Rich-Harris:
Testing the static stuff. This works beautifully, with one caveat — importing `$app/env/private` in a public-facing module will cause the private variable to appear on the page via HMR, since that bypasses our graph traversal mechanism.

I thought we could account for that case by adding the following to the generated `$app/env/private`:

```js
if (!import.meta.env.SSR) {
  throw new Error('bikeshed this message please');
}
```

and it _does_ throw, but in a rather timid way:

image

No error overlay appears, and I'm not sure why. Possibly a Vite bug?

Rich-Harris:
Also, the illegal import error isn't preventing private variables from being SSR'd:

image

packages/kit/src/vite/dev/index.js
tcc-sejohnson: Oh you know what? This is type system funkiness. They _can_ be `undefined`, but only for keys which are also `undefined`. i.e. `process.env.FOO` might be undefined if `FOO` isn't an environment variable. So given that we're looping over keys which are guaranteed to exist, we can probably just hard cast it to a `Record<string, string>`.

I'll fix it if you don't. I think there's something similar in `sync` as well.

packages/kit/src/core/sync/write_env.js
tcc-sejohnson: Generally speaking I prefer to be more rather than less explicit (especially in generated code), but tbh it probably could just be string. I just prefer the hard coupling of type definitions to where they're defined. 🤷

packages/kit/src/core/sync/write_env.js
tcc-sejohnson: I suggested this at one point and Ben wanted to stick with `$app/env` since it's already a thing, but given that we have `/private` and `/runtime` now, I think I prefer `/public`

packages/kit/src/core/sync/write_env.js
tcc-sejohnson: I don't think I totally understand what you're suggesting. Can you flesh it out a bit more?

packages/kit/src/cli.js
tcc-sejohnson: We can do this, but "development" as the default is also defined at the `sync` level, so it's technically unnecessary, right?

packages/kit/src/core/sync/write_env.js
Rich-Harris: I guess what I'm asking is what effect does this actually have? `Right-click -> Go To Definition` takes me to `.svelte-kit/types/ambient.d.ts` in either case

packages/kit/src/core/sync/write_env.js
Rich-Harris: yeah, it's way simpler. i made that change

packages/kit/src/core/sync/write_env.js
Rich-Harris: Currently, in order to get the ambient module types defined in https://github.com/sveltejs/kit/blob/master/packages/kit/types/ambient.d.ts, you need to add this line to a module that TypeScript can 'see':

```ts
```

The expectation is that you include it in `app.d.ts`, and that's where it is in the templates. But someone might want to organise their types differently, or whatever, and that reference could easily get lost (or they could create a project without types then want to add them later, etc, and this is one more thing they'd need to worry about).

If we referenced the types from the generated tsconfig.json somehow, then they'd no longer need to worry about that (as long as they `extend` the config correctly).

Rich-Harris: opened an issue for this, we can tackle after this PR https://github.com/sveltejs/kit/issues/5677

Rich-Harris:
I just realised that we're creating `$app/env/public` and `$app/env/private` inside a folder called `runtime`, even though the whole deal is that they're build-time replacements. It's not a huge deal as it isn't user-facing, but it might be worth tidying that up later just for our own sanity

Rich-Harris:
I think this is looking in pretty good shape. Remaining tasks:

* [ ] tests
* [ ] update adapters
* [ ] changesets
* [ ] docs
* [ ] last-minute bikeshedding (e.g. I think `App.RuntimeEnv` could just be `App.Env`)

tcc-sejohnson:
@Rich-Harris 

Will work on those sometime this evening. For testing, what are you thinking? Given that this is both a dev-time error and a build-time error, I'm not sure what project it should live in or what strategy we want to use.

Rich-Harris:
Checking the illegal imports stuff is going to be tricky in the context of the integration tests — we might have to settle for (probably somewhat convoluted) unit tests

Rich-Harris:
More bike-shedding — I'm having second thoughts about calling it `$app/env/platform`. It doesn't feel like a good fit for the intended use case: https://github.com/sveltejs/kit/issues/4296#issuecomment-1186481682

tcc-sejohnson:
I don't see how it's _not_ a good fit for the intended use case, but I do see how it's a good fit for basically every non-node environment. The vars are provided by the platform. And even in the case of node, the vars are still "provided" by the platform, even if the platform is just forwarding the host machine's environment variables through `process.env`.

Rich-Harris:
I guess what I mean is the variables are specific to the _deployment_. If it was just 'platform' then you'd expect it to be common across apps, but it will differ from app to app, and in the case above (where they're using Node, and are rigorous about twelve factor) you might have a single build of the app with multiple deployments each with their own environment.

But `$app/env/deployment` also feels weird.

One option I quite like:

```diff
-import { browser, dev, prerendering } from '$app/env';
+import { browser, dev, env, prerendering } from '$app/env';
```

Rich-Harris:
Oh wait. That might be confusing, since you could import it into client-side code and it would always just be `{}`, yet it'd be typed as `App.Env`. Guess we can't do that

tcc-sejohnson:
Ehhh... that seems like splitting hairs to me. 🤷 Regardless of the deployment, the platform still provides the variables, and if we're really following 12-factor, the app shouldn't know about the concept of multiple environments at all -- so technically even the name "deployment" would be a leaky abstraction akin to accidentally calling Santa Claus "your mom" when talking to your child. 😉

I'm still all for platform and I don't love deployment, but definitely still listening.

Rich-Harris:
Added some docs: https://kit-svelte-9smins22h-svelte.vercel.app/docs/modules#$app-env-platform

Rich-Harris:
Updated the adapters except the `edge` variants of `adapter-vercel` and `adapter-netlify`, because I'm unclear on whether they make dynamic env vars available.

Rich-Harris:
Added a test. I _haven't_ added a test for illegal imports — given how finicky it is, I'd be happy to leave it untested for now, but if others feel differently then... have at it I guess 😆 

I think this is basically ready. So happy to have this, it's been a long time coming

bfanger:
I'm getting 
`failed to load module for ssr: /.svelte-kit/runtime/app/env/runtime` 
when using
 `import { env } from "$app/env/runtime";`

Rich-Harris:
@bfanger it's `$app/env/platform` not `$app/env/runtime`

Rich-Harris:
Not strictly related to the issue, but #5682 makes me realise that we're not exposing environment variables to service workers. That seems bad.

We _don't_ want to expose `$app` modules to service workers in general, so I wonder if rather than using that namespace, we should create a new one that is available in the app and in the service worker:

```
import { env } from '$env/platform';
import { STUFF } from '$env/private';
import { PUBLIC_STUFF } from '$env/public';
```

This would also address https://github.com/sveltejs/kit/pull/5663#issuecomment-1192843138. Of course, only the `public` vars would be accessible to the service worker.

As a consequence of this, `$app/env` starts to feel a little strange — perhaps we need to rename that.

bfanger:
I tried that too, but it gave the opposite error in node/vite.

A client-side navigation to /posts/1  worked but the env is an empty object (i expected the PUBLIC_* to be there, or a configuration to whitelist certain variables)  refreshing the page resulted in:

```
Cannot import $app/env/platform.js into client-side code:
- src/routes/posts/[id].svelte
  - src/lib/api.ts
    - $app/env/platform.js
 at prevent_illegal_imports (file:///.../kit/packages/kit/dist/vite.js:2606:9)
```

Rich-Harris:
Yeah you can't import `platform` into client-side code. It's a bug that it lets you navigate client-side, we should try to fix that. You will need to send the relevant vars from the server

bfanger:
For the runtime variables we could an `env` export.
```
import { env } from '$env/private';
import { env } from '$env/public';
```
`env` is an object containing the _runtime_ environment settings,  for  "$env/private" this would be an alias to `process.env` and for "$env/public" this could contain all variables starting with the `envPrefix` or via a new whitelist config `string[]` .
[It is possible](https://github.com/tcc-sejohnson/kit/pull/1) to make environment imports available in both server and the client.

Not sure about the naming, maybe  
`import { runtime } from '$env/public';`  or  `import { dynamic } from '$env/public';`   is clearer.

Rich-Harris:
Interesting... that _could_ work, though we'd need to inline the env vars in the HTML in all cases, penalizing people who don't use it. Maybe we could do some bundle analysis to see if it's needed, though it would be easier if it was a separate module than a named `env` export (which would also prohibit having an env var with that name)

madeleineostoja:
I like the proposal of moving this to `$env` rather than `$app/env`. To me it feels like `$app` is the place for exposing sveltekit internals (navigation, generated stores) whereas `$env` feels like a separate utility layer over user-set environment variables. That distinction gets muddier if there's any sveltekit-generated env variables put in `$env` later.

But I also understand the desire not to bloat the api with more standalone modules so 🤷🏻‍♀️

tcc-sejohnson:
I posted the following in the maintainers' chat. Moving it here so others can see.

I like the idea of moving to $env. How about:

```
$env/private
$env/public
$env/private/platform
$env/public/platform
```

I'd still like a hook here. Maybe getPlatformEnv? It'd have this signature:

```ts
(publicPrefix: string, env: Record<string, string>) => { public: App.PublicEnv, private: App.PrivateEnv }

// Defaults to:
(publicPrefix, env) => { 
  const entries = Object.entries(env);
  const public = entries.filter(([k]) => k.startsWith(publicPrefix)).reduce((acc, [k, v]) => { return { ...acc, [k]: v } }, {});
  const private = entries.filter(([k]) => !k.startsWith(publicPrefix)).reduce((acc, [k, v]) => { return { ...acc, [k]: v } }, {});
  return {
    public,
    private
  }
} 
```

This would allow for the best of both worlds where users "just get" public platform vars, and if there are too many and they want to pare them down or change them for some reason, they can. It also guarantees (if implemented) that they're not accidentally exposing private variables because the return is typed.

(I know those reduces are terrible, but I'm on mobile and don't want to write out for loops)

We may want to split that into two separate hooks? It seems like it might be useful to have `getPublicEnv` run per-request and receive an instance of the request URL. I dunno. I'm just spitballing on how to avoid having ALL variables automatically inlined. 🤷

PH4NTOMiki:
@tcc-sejohnson instead of using `reduce` to convert `entries` back to an object you can use `fromEntries` for ex. for public:
`const public = Object.fromEntries(entries.filter(([k]) => k.startsWith(publicPrefix)));`
It looks much cleaner, and its more obvious that its converting entries back to an object.

Rich-Harris:
I'd prefer we didn't add a hook, for two reasons. Firstly, it adds complexity - every new API is something that has to be tested and maintained and documented, and pushes us towards the place where people look at SvelteKit and go 'idk man, seems complicated'. Unless a feature has very clear demonstrable value and can't (unlike this) effectively be achieved on userland, it's something we should generally avoid.

Secondly, there's a timing issue. There are plenty of situations where you might want to access `env` inside the main body of hooks.js, but that's not possible if it has to go _through_ a hook first. That would be solvable with another hooks module dedicated to this problem, but that's yet more complexity.

Rich-Harris:
Also: per-request would be a footgun. It would be tempting to return only the variables that are useful for the page being requested, but in fact you would need to return all variables needed for the entire session including subsequent client-side navigations. This would be easy to get wrong and very difficult to catch before the bug hit production.

bfanger:
Agreed, a whitelisting approach (especially for the public dynamic env variables) is preferred, I think we might even be able to simplify the api:

Usage:

```js
import { VAR_NAME } from "$env";
```

By default this will work based on environment variables that are available at compile time.
The server bundle will contain the the value and on the client the import will work but the value will be undefined.

To change the behavior of the VAR_NAME you'll have two independent configuration options:

In svelte.config.js you can:

```js
env: {
    public: ['VAR_NAME'],
    dynamic: ['VAR_NAME'],
}
```

When the variable name is in the **public** array the value will also be available in the client bundle.

When the variable name is in the **dynamic** array the value won't be set at build time, but at runtime.

When the variable name is in **both**, then the value will be injected into the requests.

tcc-sejohnson:
@PH4NTOMiki 

Thanks! I don't know how I've never seen that before...

@Rich-Harris 

Yeah, the more I think about it the less I like it, for those reasons and more.

@bfanger 

Unfortunately some parts of that just won't work period, and I don't love the idea of an explicit whitelist when that's already effectively what a public prefix is.

1. Public and private _have_ to be separate modules in order to enable import analysis during dev and at buildtime.
2. We can't mix runtime and buildtime vars like that. If we can declare the buildtime vars as a `const VAR = 'some string';`, the compiler can statically replace references to it. For runtime vars, we have to use `let` with a setter. 

I do like some of these ideas, though. Here's my proposal:

```
$env/private
$env/public
```

I propose we make the behavior of the runtime/buildtime modules practically identical to the developer. I know, I know -- but hear me out:

Runtime and buildtime environment variables are practically identical during dev. You know which ones you need, and you know when you need them (runtime vs. buildtime -- which are basically the same thing during dev). The declaration for the two is very similar: one is declared with `const`, the other is declared with `let` (+ a setter).

Rather than just a `publicPrefix`:

```
env: {
  buildtimePublicPrefix: "BT_PUB",
  buildtimePrivatePrefix: "BT_PRIV",
  runtimePublicPrefix: "RT_PUB",
  runtimePrivatePrefix: "RT_PRIV",
}
```

Using this strategy, we actually could export all vars from just one public and one private module. If like @bfanger you just use runtime vars, you could set those prefixes to just "PUB"  and "PRIV" to shorten your var names.

For the runtime vars, we just use the same "dump the dev-time values" approach:

```ts
// $env/private
export let RT_PRIV_FOO = '';

export const __set_rt_priv_foo(val: string) => RT_PRIV_FOO = val;
```

We can then use code generation in `build_server` to make sure that the setters are all appropriately called in `init`.

This has a few advantages:
1. It's tree-shakeable! The runtime vars will still be dynamic, but the unused ones will go away.
2. Only two modules to import from
3. A dead-simple development experience (stick your vars -- all of them -- in a .env file and you're good to go)
4. Eliminates any chance of naming clashes (pretty much any other solution requires, one way or another, reserving env var names to make room for SK-specific exports from the module)
5. Autocomplete for runtime environment vars!

This is the first idea I've had that feels like something I would really want to use without reservation.

bfanger:
We _can_ mix runtime and buildtime vars because with a whitelist, the variable names are known at the moment the javascript is generated.
Can't we alias $env to server/env.js for server code and to client/env.js ?

I don't like prefixes because if I want to use let say, [GITHUB_RUN_ID](https://docs.github.com/en/actions/learn-github-actions/environment-variables) I'l have to re-export that env before the build and it adds clutter on all imports.

madeleineostoja:
Just want to chime in as a humble user to say that I actually love the prefix approach, it makes things very clear straight from variable definition, and it's something users will already be familiar with coming from the current vite (+ process.env) setup. I already have my vite prefix set to `PUBLIC_` (which I imagine is a reasonably common situation), so this would be a very simple migration

tcc-sejohnson:
> We can mix runtime and buildtime vars because with a whitelist, the variable names are known at the moment the javascript is generated.

Yeah, I guess I just really hate the idea of a manual whitelist. I really don't want to have to specify the names of my environment variables in yet another place when I can specify _just a few prefixes_ for potentially dozens of values and not worry about having to add to the list every time I add one. You're trying to avoid having to add another export to your build process... by adding another export to your config file.

I'm strongly against any sort of private/public combining for two reasons: It feels too magical, and it's super unclear where a variable is coming from.

For example:

```ts
import { FOO } from '$env';
```

Is that public? Private? Runtime or buildtime? Is it going to blow up my build or not? Intellisense doesn't know and I sure don't either. It's also harder to explain the semantics: "Sometimes importing a variable into client-side code from `$env` will blow up your build, sometimes it won't. There are rules for how that works but you're just going to have to memorize them, and when you see `FOO` in your code you're going to have to mosey on over to `svelte.config.js` to see if it's whitelisted or not."

I still stand by my prior proposal, but I'm totally still listening to the arguments from the Kit brain trust, including you. 🙂

packages/kit/src/core/sync/write_env.js
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * @param {string} id
+ * @param {Record<string, string>} env
+ * @returns {string}
+ */
+function create_module(id, env) {
+	const declarations = Object.entries(env)
+		.map(
+			([k, v]) => `/** @type {import('${id}'}').${k}} */\nexport const ${k} = ${JSON.stringify(v)};`
```
jasonlyu123, line 48: Found this would fail in some cases as the key might not be a valid javascript identifier. Especially on windows where there might be a  `ProgramFiles(x86)` variable

tcc-sejohnson:
Check out this branch: https://github.com/tcc-sejohnson/kit/tree/4296/sejohnson-environment-variables-2.0

It:
- Switched out `$app/env/*` for `$env/*` (haven't touched `$app/env` itself, but it should probably be renamed if we're going to `$env`)
- Implements most of the last proposal I made.

There are a couple of neat tricks here:
- Intellisense for both runtime and buildtime vars
- No naming conflicts
- Only two modules
- Simple (but flexible) API. Only adds 4 config options. Want all runtime environment variables to be private? Set `config.kit.env.runtimePrivatePrefix` to `''`.
- Import analysis still works great
- Both runtime and buildtime environment vars are tree-shakeable, and because runtime public vars have to be prefixed, we only have to send exactly those vars over the wire

To-do: Everything works great in SSR, but I haven't figured out how to inline the public runtime environment vars into the HTML sent from the server to the client.

I didn't do anything to this branch because I didn't want to break anything, though we should be able to PR that branch into this one if it ends up being the way we want to go. I've been playing around with it for the past half hour or so, and the ergonomics are _great_ so far (aside from, y'know, not having the runtime vars client-side part implemented yet).

bfanger:
@tcc-sejohnson How do you feel about removing the prefix from the generated variable names?

dummdidumm:
@bfanger I had the same thought, but came to the conclusion that it's better to keep the prefix part of the variable name, because
- doing a simple "I select this whole variable name in my env file and search for it" will yield results compared to "oh yeah I need to omit the prefix"
- it's more apparent from just looking at the name that it's private or public, else you'd have to go to the import at the top of the file
- possibly name clashes

tcc-sejohnson:
Yeah, @dummdidumm covered it pretty well. You can set one of the prefixes to the empty string, though, which would dump all env vars to whatever type.

Rich-Harris:
I really can't say I'm a fan of the multiple prefixes approach. Just on an aesthetic level, it's kinda brutal. But it also adds friction — one nice thing about the current approach in this PR is that you can easily do things like this...

```js
import { NODE_VERSION } from '$app/env/private';
import { env } from '$app/env/platform';

export function GET() {
  return {
    body: {
      builtWith: NODE_VERSION,
      runningOn: env.NODE_VERSION
    }
  };
}
```

...or use whatever other environment variables are useful but that _you don't own_, without needing to faff around aliasing them. It's hard to see what ergonomic advantages you get from having four different prefixes rather than four separate modules.

---

Combining stuff a la https://github.com/sveltejs/kit/pull/5663#issuecomment-1193357403 is a non-starter I'm afraid. The issue isn't whether transforms are running in SSR mode or not, it's whether a given module is _public_, which doesn't neatly align with SSR/non-SSR. `src/routes/index.svelte` is transformed in both SSR and non-SSR modes, but it shouldn't be able to access private env vars in either case.

---

Agree with @dummdidumm about removing prefixes. It's a layer of magic that will ultimately cause more pain and confusion than convenience.

---

I honestly think we're pretty close. We need to rename the modules — I suggest `$env/static/private`, `$env/static/public`, `$env/dynamic/private` and `$env/dynamic/public` — and implement the last of those, but then I think we've basically solved it?

tcc-sejohnson:
@Rich-Harris 

Fair enough. What do we want to do with `$app/env`? Anything? Seems weird to leave it.

Rich-Harris:
I have a bad feeling that it's going to take us a while to come up with an alternative to `$app/env` that isn't even worse. My first thought was `$app/context`, but then I remembered that we used to have three separate things called `context` in SvelteKit until we removed one and renamed the other to `stuff`.

We could probably punt on that for now, lest it kill this PR's momentum

Rich-Harris:
Question: should the contents of whichever `.env` file Vite has seen fit to load be available to `$env/dynamic/[type]` in `vite dev` and `vite preview`? Or just the contents of `process.env`?

I'm guessing @bfanger would say that vars from `.env` _should_ be available, even if it effectively means that the static/dynamic vars are identical during `dev` and `preview`

Rich-Harris:
(Or, third option — they're available to `vite dev` but _not_ in `vite preview`, so that the latter is a more faithful simulation of the production environment, where you're responsible for populating the env vars yourself)

tcc-sejohnson:
@Rich-Harris 

I'd say make them available in both `dev` and `preview`. Having to manually specify them as exports prior to running `preview` is just a pointless PITA. Just use `loadEnv(vite_config.mode, process.cwd(), '')` to set the environment in both cases.

Even better, factor out the logic used to filter/assign public and private env variables out of `write_env` and into one of the `utils` files and just share it between all three locations.

.changeset/beige-gorillas-tell.md
```diff
@@ -0,0 +1,9 @@
+---
+'@sveltejs/adapter-cloudflare': patch
+'@sveltejs/adapter-cloudflare-workers': patch
+'@sveltejs/adapter-netlify': patch
+'@sveltejs/adapter-node': patch
+'@sveltejs/adapter-vercel': patch
+---
+
+Initialise `$env/dynamic/private`
```
Rich-Harris, line 9:
```suggestion
Initialise `env`
```

packages/kit/src/core/sync/utils.js
```diff
@@ -31,3 +31,16 @@ export function trim(str) {
...
...DD69\uDD6D-\uDD72\uDD7B-\uDD82\uDD85-\uDD8B\uDDAA-\uDDAD\uDE42-\uDE44]|\uD835[\uDC00-\uDC54\uDC56-\uDC9C\uDC9E\uDC9F\uDCA2\uDCA5\uDCA6\uDCA9-\uDCAC\uDCAE-\uDCB9\uDCBB\uDCBD-\uDCC3\uDCC5-\uDD05\uDD07-\uDD0A\uDD0D-\uDD14\uDD16-\uDD1C\uDD1E-\uDD39\uDD3B-\uDD3E\uDD40-\uDD44\uDD46\uDD4A-\uDD50\uDD52-\uDEA5\uDEA8-\uDEC0\uDEC2-\uDEDA\uDEDC-\uDEFA\uDEFC-\uDF14\uDF16-\uDF34\uDF36-\uDF4E\uDF50-\uDF6E\uDF70-\uDF88\uDF8A-\uDFA8\uDFAA-\uDFC2\uDFC4-\uDFCB\uDFCE-\uDFFF]|\uD836[\uDE00-\uDE36\uDE3B-\uDE6C\uDE75\uDE84\uDE9B-\uDE9F\uDEA1-\uDEAF]|\uD838[\uDC00-\uDC06\uDC08-\uDC18\uDC1B-\uDC21\uDC23\uDC24\uDC26-\uDC2A]|\uD83A[\uDC00-\uDCC4\uDCD0-\uDCD6\uDD00-\uDD4A\uDD50-\uDD59]|\uD83B[\uDE00-\uDE03\uDE05-\uDE1F\uDE21\uDE22\uDE24\uDE27\uDE29-\uDE32\uDE34-\uDE37\uDE39\uDE3B\uDE42\uDE47\uDE49\uDE4B\uDE4D-\uDE4F\uDE51\uDE52\uDE54\uDE57\uDE59\uDE5B\uDE5D\uDE5F\uDE61\uDE62\uDE64\uDE67-\uDE6A\uDE6C-\uDE72\uDE74-\uDE77\uDE79-\uDE7C\uDE7E\uDE80-\uDE89\uDE8B-\uDE9B\uDEA1-\uDEA3\uDEA5-\uDEA9\uDEAB-\uDEBB]|\uD869[\uDC00-\uDED6\uDF00-\uDFFF]|\uD86D[\uDC00-\uDF34\uDF40-\uDFFF]|\uD86E[\uDC00-\uDC1D\uDC20-\uDFFF]|\uD873[\uDC00-\uDEA1\uDEB0-\uDFFF]|\uD87A[\uDC00-\uDFE0]|\uD87E[\uDC00-\uDE1D]|\uDB40[\uDD00-\uDDEF])*$/;
```
Rich-Harris, line 44: FWIW i'd be fine with us doing `/^[a-z_$][a-z0-9_$]*$/i` (and also checking for reserved words). if someone is using other characters in their env vars they're asking for pain

packages/kit/src/runtime/env-private.js
```diff
@@ -0,0 +1,9 @@
+// TODO need separate public/private types
```
Rich-Harris, line 1:
```suggestion
```

outdated

packages/kit/src/runtime/env-public.js
```diff
@@ -0,0 +1,9 @@
+// TODO need separate public/private types
```
Rich-Harris, line 1:
```suggestion
```

outdated

packages/kit/src/runtime/env-private.js
```diff
@@ -0,0 +1,9 @@
...
+/** @type {App.PrivateEnv} */
+export let env = {};
+
+// TODO put this in a separate module, so it's not user-accessible
```
Rich-Harris, line 5: outdated

```suggestion
```

packages/kit/src/runtime/env-public.js
```diff
@@ -0,0 +1,9 @@
...
+/** @type {App.PublicEnv} */
+export let env = {};
+
+// TODO put this in a separate module, so it's not user-accessible
```
Rich-Harris, line 5: outdated

```suggestion
```

packages/kit/src/core/sync/sync.js
```diff
@@ -24,12 +33,18 @@ export function update(config) {
...
 	return { manifest_data };
 }
 
-/** @param {import('types').ValidatedConfig} config */
-export function all(config) {
-	init(config);
-	return update(config);
+/**
+ * Run sync.init and sync.update in series, returning the result from
+ * sync.update.
```
benmccann, line 43: I imagine this could fit on one line
```suggestion
 * Run sync.init and sync.update in series, returning the result from sync.update.
```

packages/kit/src/vite/utils.js
```diff
@@ -141,3 +149,44 @@ export function resolve_entry(entry) {
 
 	return null;
 }
+
+/**
+ *
+ * @param {string} str
+ * @param {number} times
+ * @returns
```
benmccann, line 157:
```suggestion
 * @param {string} str
 * @param {number} times
```

packages/kit/src/vite/utils.js
```diff
@@ -141,3 +149,44 @@ export function resolve_entry(entry) {
...
+		return path.relative(process.cwd(), file);
+	});
+
+	const pyramid = stack.map((file, i) => `${repeat(' ', i * 2)}- ${file}`).join('\n');
+
+	return `Cannot import ${stack.at(-1)} into client-side code:\n${pyramid}`;
+}
+
+/**
+ * Load envinronment variables from process.env and .env files
```
benmccann, line 181:
```suggestion
 * Load environment variables from process.env and .env files
```

packages/kit/test/apps/basics/.env
```diff
@@ -0,0 +1,5 @@
+PRIVATE_STATIC="accessible to server-side code/replaced at build time"
+PRIVATE_DYNAMIC="accessible to server-side code/evaluated at run time"
+
+PUBLIC_STATIC="accessible anywhere/replaced at build time"
+PUBLIC_DYNAMIC="accessible anywhere/evaluated at run time"
```
benmccann, line 5:
```suggestion
PUBLIC_DYNAMIC="accessible anywhere/evaluated at run time"

```

packages/kit/test/apps/basics/.gitignore
```diff
@@ -1 +1,2 @@
-/test/errors.json
\ No newline at end of file
+/test/errors.json
+!/.env
```
benmccann, line 2:
```suggestion
!/.env

```

documentation/faq/60-env-vars.md
```diff
@@ -2,8 +2,6 @@
 title: How do I use environment variables?
 ---
 
-Vite uses [dotenv](https://github.com/motdotla/dotenv) to load environment variables from a file named `.env` or similar. Only environment variables prefixed with `VITE_` are exposed ([you can set `envPrefix` to change this](https://vitejs.dev/config/shared-options.html#envprefix)). You can access the variable using `import.meta.env.VITE_ENV_VAR`, and Vite will statically replace them at build-time.
+Environment variables are [loaded by Vite](https://vitejs.dev/guide/env-and-mode.html#env-files) from `.env` files and `process.env` during dev and build. These are available to your app via the [`$env/static/private`](/docs/modules#$env-static-private) and [`$env/static/public`](/docs/modules#$env-static-public) modules. Public variables — i.e. those that are safe to expose to users — must use a [prefix](/docs/configuration#kit-env-publicprefix) that defaults to `PUBLIC_`. Private variables cannot be imported into client-side code.
```
benmccann, line 5: I added this FAQ quite awhile ago when using Vite's environment variables was truly confusing. I think we should probably have some documentation of our own environment variable support in `documentation/docs/07-modules.md` and can remove this FAQ in favor of that

packages/kit/src/core/sync/sync.js
```diff
@@ -6,15 +6,24 @@ import { write_matchers } from './write_matchers.js';
...
+import { write_env } from './write_env.js';
 
-/** @param {import('types').ValidatedConfig} config */
-export function init(config) {
+/**
+ * Initialize SvelteKit's generated files.
+ * @param {import('types').ValidatedConfig} config
+ * @param {string | undefined} mode
+ */
+export function init(config, mode = undefined) {
```
benmccann, line 16:
```suggestion
 * @param {string} [mode]
 */
export function init(config, mode) {
```

documentation/faq/60-env-vars.md
Rich-Harris: we have https://kit-svelte-kxu7loqzq-svelte.vercel.app/docs/modules#$env-dynamic-private. did you mean something beyond that?

benmccann: No, that's good. I just didn't see `documentation/docs/07-modules.md` updated, but I guess that content gets inserted in when the docs are built

I'd say we just go ahead and remove this FAQ now then. No point in splitting documentation between multiple places

documentation/docs/15-configuration.md
```diff
@@ -157,9 +160,11 @@ When pages are prerendered, the CSP header is added via a `<meta http-equiv>` ta
 
 > Note that most [Svelte transitions](https://svelte.dev/tutorial/transition) work by creating an inline `<style>` element. If you use these in your app, you must either leave the `style-src` directive unspecified or add `unsafe-inline`.
 
-### moduleExtensions
+### env
 
-An array of file extensions that SvelteKit will treat as modules. Files with extensions that match neither `config.extensions` nor `config.kit.moduleExtensions` will be ignored by the router.
+Environment variable configuration:
+
+- `publicPrefix` — a prefix that signals that an environment variable is safe to expose to client-side code. See [`$env/static/public`](/docs/modules#$env-static-public) and [`$env/dynamic/public`](/docs/modules#$env-dynamic-public)
```
benmccann, line 167:
```suggestion
- `publicPrefix` — a prefix that signals that an environment variable is safe to expose to client-side code. See [`$env/static/public`](/docs/modules#$env-static-public) and [`$env/dynamic/public`](/docs/modules#$env-dynamic-public). Note that Vite's [`envPrefix`](https://vitejs.dev/config/shared-options.html#envprefix) must be set separately if you are using Vite's environment variable handling - though use of that feature should generally be unnecessary.
```

documentation/faq/60-env-vars.md
Rich-Harris: cool, done — this means that searching for 'environment variables' will no longer favour the FAQ over the docs

tcc-sejohnson:
I kindly ask that we not merge this PR until I at least _attempt_ to add unit tests for the import analysis (probably tonight) -- it's complicated enough that it'd be nice to verify the behavior.

packages/kit/src/core/sync/write_env.js
```diff
@@ -0,0 +1,76 @@
...
+		.map(([k, v]) => {
+			const comment = `/** @type {import('${id}'}').${k}} */`;
+
+			if (reserved.has(k)) {
+				let i = 1;
+				while (k + i in env) i++;
+				const name = k + 1;
+
+				return `${comment}\nconst ${name} = ${JSON.stringify(v)}; export { ${name} as ${k} }`;
+			}
```
tcc-sejohnson, line 55: I'd still like a resolution on this (see discussion starting [here](https://discordapp.com/channels/457912077277855764/997637728608985169/1001258747508830308)).

I'd really prefer to log a warning and ignore reserved keywords.

tcc-sejohnson: I mean, we're _already_ silently filtering out invalid JS identifiers. Seems like we should just silently filter out reserved JS keywords as well and make sure it's notated in the documentation. I'd be fine with that. Though the logging is probably preferable?

tcc-sejohnson:
Two things:

1. I added tests for the recursive import analysis. They rely on some... _interesting_ characteristics of JavaScript, but they test what we want. The biggest reason I really wanted to have them is for when/if there ends up being a bug: Just way easier to write a failing test when there's already a framework for them. This required factoring them back out into the `vite/utils.js` file, but it didn't really meaningfully increase complexity. It also made them more testable (no reliance on variables from an outer scope except through function arguments).
2. As part of that, I added a small enhancement so that the error pyramid will tag dynamically imported modules. I could see that being _really_ confusing if you didn't understand exactly how they work. Example:

```
Cannot import $env/static/private.js into client-side code:
- src/routes/index.svelte 
  - src/lib/test4.ts (imported by parent dynamically)
    - src/lib/test3.ts 
      - src/lib/test2.ts 
        - src/lib/test1.ts 
          - $env/static/private.js 
```

packages/kit/types/ambient.d.ts
```diff
@@ -85,6 +99,36 @@ declare module '$app/env' {
...
+ */
+declare module '$env/dynamic/private' {
+	export let env: App.PrivateEnv;
+}
+
+/**
+ * Similar to [`$env/dynamic/private`](https://kit.svelte.dev/docs/modules#$env-dynamic-private), but only includes
+ * variables that begin with [`config.kit.env.publicPrefix`](https://kit.svelte.dev/docs/configuration#kit-env-publicprefix)
+ * (which defaults to `PUBLIC_`), and can therefore safely be exposed to client-side code
+ *
```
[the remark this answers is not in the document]
Rich-Harris:
```suggestion
 *
 * Note that public dynamic environment variables must all be sent from the server to the client, causing larger network requests — when possible, use `$env/static/public` instead.
 *
```

packages/kit/src/core/sync/write_env.js
Rich-Harris: Added some warnings for this. They'll be hidden by the `vite dev` welcome message, but rather than making our workaround more all-encompassing (i started down this road, it gets a little messy) i opened an issue on vite: https://github.com/vitejs/vite/issues/9378

packages/kit/scripts/extract-types.js
```diff
@@ -117,9 +131,9 @@ modules.push({
 
 			modules.push({
 				name,
-				comment,
+				comment: comment.replace(/https:\/\/kit\.svelte\.dev/g, ''),
```
benmccann, line 134: this could use a comment. I don't understand why `kit.svelte.dev` was prepended to all the links and then is being removed here

packages/kit/rollup.config.js
```diff
@@ -23,6 +23,10 @@ export default [
 			'app/stores': 'src/runtime/app/stores.js',
 			'app/paths': 'src/runtime/app/paths.js',
 			'app/env': 'src/runtime/app/env.js',
+			'env/dynamic/private': 'src/runtime/env/dynamic/private.js',
+			'env/dynamic/public': 'src/runtime/env/dynamic/public.js',
+			'env-private': 'src/runtime/env-private.js',
```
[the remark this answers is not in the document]
Rich-Harris: felt unnecessary since we don't have entry points for `static`. in any case i'm thinking of these as temporary — this whole folder structure is due for a reshuffle (it's weird that we have `env` alongside `env-[type]`, for example)

packages/kit/types/ambient.d.ts
```diff
@@ -85,6 +99,38 @@ declare module '$app/env' {
...
+ * Note that public dynamic environment variables must all be sent from the server to the client, causing larger network requests — when possible, use `$env/static/public` instead.
+ *
+ * ```ts
+ * import { env } from '$env/dynamic/private';
```
benmccann, line 126:
```suggestion
 * import { env } from '$env/dynamic/public';
```

packages/kit/src/core/sync/sync.js
```diff
@@ -6,14 +6,24 @@ import { write_matchers } from './write_matchers.js';
...
+/**
+ * Initialize SvelteKit's generated files.
+ * @param {import('types').ValidatedConfig} config
+ * @param {string} mode
+ */
+export function init(config, mode) {
 	copy_assets(path.join(config.kit.outDir, 'runtime'));
+
 	write_tsconfig(config.kit);
+	write_env(config.kit, mode);
```
benmccann, line 20: why would this only be done at `init`? if someone changes an environment variable, don't we want to pick it up?

Rich-Harris: The only way they can change an env var that we can observe is if they edit an `.env` file, in which case Vite restarts (and `init` happens again)

packages/kit/scripts/extract-types.js
```diff
@@ -19,59 +20,72 @@ function get_types(code, statements) {
...
-			let start = statement.pos;
-			let comment = '';
-
-			// @ts-ignore i think typescript is bad at typescript
-			if (statement.jsDoc) {
-				// @ts-ignore
-				comment = statement.jsDoc[0].comment;
-				// @ts-ignore
-				start = statement.jsDoc[0].end;
+	if (statements) {
```
benmccann, line 23: personal preference, but I'd write this as shown below. right now, it's a bit hard to see where the `if` block ends and what comes after it since it doesn't fit on one screen. it'd also save having to have the code inside indented as much
```
if (!statements) {
  return { types, exports };
}
```

packages/kit/src/vite/index.js
```diff
@@ -14,7 +14,12 @@ import { generate_manifest } from '../core/generate_manifest/index.js';
 import { get_runtime_directory, logger } from '../core/utils.js';
 import { find_deps, get_default_config as get_default_build_config } from './build/utils.js';
 import { preview } from './preview/index.js';
-import { get_aliases, resolve_entry } from './utils.js';
+import {
+	get_aliases,
+	resolve_entry,
+	format_illegal_import_chain,
```
benmccann, line 20: this doesn't look used. is there some way we can get typescript to report on unused imports or is this something we're missing from removing eslint? I think there's some stuff like this that's exported from `vite/utils.js` that might not need to be

tcc-sejohnson: That function can likely be internal now since both functions that utilize it are in the same file.

packages/kit/scripts/extract-types.js
Rich-Harris: my personal preference goes the other way 😆 i know you love an early return, i much prefer positive conditions over negative ones — way fewer mental contortions — and not having duplicated identical `return` statements

packages/kit/src/vite/index.js
Rich-Harris: yep, this is an eslint casualty. fixed

benmccann: PR to stop it from happening again: https://github.com/sveltejs/kit/pull/5716

packages/kit/src/vite/utils.js
```diff
@@ -141,3 +149,135 @@ export function resolve_entry(entry) {
...
+		private: Object.fromEntries(entries.filter(([k]) => !k.startsWith(prefix)))
+	};
+}
+
+/**
+ * @param {(id: string) => import('rollup').ModuleInfo | null} node_getter
+ * @param {import('rollup').ModuleInfo} node
+ * @param {Set<string>} illegal_imports
+ * @param {string} out_dir
+ * The directory specified by config.kit.outDir.
```
benmccann, line 205: this can be one line
```suggestion
 * @param {string} out_dir The directory specified by config.kit.outDir
```

packages/kit/src/vite/utils.js
```diff
@@ -141,3 +149,135 @@ export function resolve_entry(entry) {
...
+ * @param {number} times
+ */
+function repeat(str, times) {
+	return new Array(times + 1).join(str);
+}
+/**
+ * Create a formatted error for an illegal import.
+ * @param {Array<{name: string, dynamic: boolean}>} stack
+ * @param {string} out_dir
+ * The directory specified by config.kit.outDir.
```
benmccann, line 164:
```suggestion
 * @param {string} out_dir The directory specified by config.kit.outDir
```

packages/kit/src/vite/utils.js
```diff
@@ -141,3 +149,135 @@ export function resolve_entry(entry) {
...
 }
+
+/**
+ * @param {string} str
+ * @param {number} times
+ */
+function repeat(str, times) {
+	return new Array(times + 1).join(str);
+}
+/**
```
benmccann, line 160:
```suggestion

/**
```

packages/kit/src/vite/utils.js
```diff
@@ -141,3 +149,135 @@ export function resolve_entry(entry) {
...
+		if (chain) return [{ name: node.id, dynamic }, ...chain];
+	}
+};
+
+/**
+ * Throw an error if a private module is imported from a client-side node.
+ * @param {import('vite').ModuleNode} node
+ * @param {Set<string>} illegal_imports
+ * @param {string} out_dir
+ * The directory specified by config.kit.outDir.
```
benmccann, line 254:
```suggestion
 * @param {string} out_dir The directory specified by config.kit.outDir
```

packages/kit/src/core/sync/write_env.js
```diff
@@ -0,0 +1,85 @@
...
+import { write_if_changed, reserved, valid_identifier } from './utils.js';
+
+const autogen_comment = '// this file is generated — do not edit it\n';
+
+/**
+ * Writes the existing environment variables in process.env to
+ * $env/static/private and $env/static/public
+ * @param {import('types').ValidatedKitConfig} config
+ * @param {string} mode
+ * The Vite mode.
```
benmccann, line 13:
```suggestion
 * @param {string} mode The Vite mode
```

Rich-Harris (approved):

campbell-hay:
I had to revert to version 70 (previous) because this version 71 was making my netlify function crash saying `server.init` did not exist or is undefined. Not sure how it is related but reverting back to 70 fixed the issue. 

For ref, I am using the netlify adapter without any option, so all defaults.

benmccann:
@campbell-hay it sounds like you upgraded `adapter-netlify`, but not SvelteKit. You need to update both of them together

campbell-hay:
@benmccann Oh good to know, my kit was 392 when it was not working with netlify adapter 71, now my kit 396. That means I could bump up the adapter to version 71 again then ?

benmccann:
Yes

JordanShurmer:
Looks like the **runtime** variables are missing from the documentation.

Is that how we would load optional variables though? loading from the dynamic/static gives an error during the build if the variable doesn't exist.

tcc-sejohnson:
https://kit.svelte.dev/docs/modules#$env-dynamic-private

Dynamic is runtime. You're likely importing it incorrectly -- dynamic only exports one object, `env`, containing the runtime environment vars.

orlov-vo:
How to make conditional reading of environment values between private/public scopes based on `browser` variable?

I tried do it like this, but I've got `Cannot import $env/dynamic/private.js into client-side code` error

```js
import { PUBLIC_API_BASE_URL } from '$env/static/public';

const getMyValue = () => {
    if (browser) return PUBLIC_MY_VALUE;
    return import('$env/dynamic/private').then(({ env }) => env.MY_VALUE);
};
```

DevOfManyThings:
@orlov-vo It defeats the point of private scoped variables if you can import them on the client, would you not be better off doing whatever logic you need to do in an endpoint with the private variable then just using the public one on the client?

tcc-sejohnson:
@orlov-vo 

You cannot. We completely disallow imports of `$env/*/private` in `.svelte` files. You would leak those variables to the client. If you want to reference private variables, you must do so from a file that is _guaranteed_ to only run on the server, such as an endpoint.

RamiAwar:
I'm a beginner here, and I can't really understand why my env vars aren't working.
- Updated to latest version of sveltejs/kit that includes these changes
- Calling `import { API_URL } from '$env/dynamic/private';` from inside lib/api.js which I assume is server side
- Set the environment variable in the shell before launching npm run dev 

Still get an undefined env var.

RamiAwar:
@tcc-sejohnson How do you validate your env vars? I saw you mention that in your initial design proposal, I'm interested in doing that as well.

tcc-sejohnson:
@RamiAwar 

That was something we decided to leave up to userland, though we solved some of it by accident. For the static modules, because you're importing the variables by name, if they're not defined at buildtime, your build will error. For dynamic vars, you can build an "env var validator" module, import it into `hooks.ts`, and call whatever `validate` method you want.

RamiAwar:
@tcc-sejohnson Just noticed this when deploying with the new changes. That's awesome! 

I feel like this should still be documented though imo, would be really useful to know that for static vars, build will error out if they're undefined.

sasselin:
This feature work only in preview? Not with npm run dev?

PR #5678, opened 2022-07-22
[fix] prerender in subprocess

benmccann:
fixes https://github.com/sveltejs/kit/issues/5306

the first half of this was done in https://github.com/sveltejs/kit/pull/5659. this finishes the job


packages/kit/src/vite/index.js
```diff
@@ -320,12 +319,26 @@ function kit() {
...
-				manifest_path,
-				log
-			});
+			const results_path = `${svelte_config.kit.outDir}/generated/prerendered.json`;
+
+			// do prerendering in a subprocess so any dangling stuff gets killed upon completion
+			spawnSync(
+				'node',
+				[
+					'./node_modules/@sveltejs/kit/dist/prerender.js',
```
dominikg, line 328: does this always work? Should either resolve the location or put it as bin in kit package

packages/kit/src/vite/index.js
```diff
@@ -320,12 +319,26 @@ function kit() {
...
+					results_path,
+					'--manifest_path',
+					manifest_path,
+					'--verbose',
+					'' + verbose
+				],
+				{ stdio: 'inherit' }
+			);
+
+			prerendered = JSON.parse(fs.readFileSync(results_path, 'utf8'));
```
dominikg, line 341: Do we need safeguards to ensure the file content is from the current run and there were no problems?

benmccann: ah, yeah, it was no longer failing the build if prerendering failed. fixed that

packages/kit/src/vite/index.js
benmccann: It should work for npm and pnpm unless you've done something weird like install sveltekit globally or only installed it in the monorepo root. One benefit of the code as it is now is that it doesn't require us to expose it externally. We could add some validation to make sure that it does something sensible if the file is not there at least

Is there an easy way to resolve the package location? It sounds nice, but also like a big can of worms if I remember my experience with that from Vite. I think we need an extra package to do that and a lot of them are broken. I can't remember if we finally found a good one that works and is easy to use?

dominikg: > Is there an easy way to resolve the package location?

`createRequire(import.meta.url).resolve('@sveltejs/kit/package.json')` and some path manipulation to append dist/prerender.js if you don't want to expose it

dominikg: Or with just the local file path like https://github.com/sveltejs/vite-plugin-svelte/blob/308c879a4881c61f625d0dc072d65d85352c49a1/packages/vite-plugin-svelte/src/ui/inspector/plugin.ts#L16

benmccann: done. thanks! that was easy :smile:

dominikg (approved):

packages/kit/src/core/prerender/prerender.js
```diff
@@ -1,18 +1,34 @@
...
 import { escape_html_attr } from '../../utils/escape.js';
+import { logger } from '../utils.js';
+import { load_config } from '../config/index.js';
 
 /**
  * @typedef {import('types').PrerenderErrorHandler} PrerenderErrorHandler
  * @typedef {import('types').Logger} Logger
  */
 
+const prog = sade('prerender');
```
Rich-Harris, line 19: do we need to use `sade` here? if it's only being used internally we could literally just do

```js
const [, , client_out_dir, results_path, manifest_path, verbose] = process.argv;
```

packages/kit/src/vite/index.js
```diff
@@ -320,12 +320,32 @@ function kit() {
...
+			const results_path = `${svelte_config.kit.outDir}/generated/prerendered.json`;
+
+			// do prerendering in a subprocess so any dangling stuff gets killed upon completion
+			const kit_dir = createRequire(import.meta.url)
+				.resolve('@sveltejs/kit/package.json')
+				.slice(0, -'package.json'.length);
+			const { status } = spawnSync(
+				'node',
+				[
+					kit_dir + 'dist/prerender.js',
```
Rich-Harris, line 332: would love it if we didn't lose the ability to test the `src` code — having to build `packages/kit` to test out simple changes is a pain

packages/kit/src/core/prerender/prerender.js
benmccann: done

packages/kit/src/vite/index.js
dominikg: https://github.com/sveltejs/kit/pull/5678#discussion_r927979693 it could work if you get the current file path (__dirname) and relative from that (checking for dist vs src in case the relative path is different)

benmccann: done

dominikg (approved):

benmccann:
https://github.com/sveltejs/kit/pull/5686 should be merged before this PR

.changeset/witty-avocados-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] remove Prerendered.assets
```
Rich-Harris, line 5: this doesn't reflect what's happening in this PR — was this a separate piece of work we were hoping to include here?

packages/kit/src/vite/index.js
```diff
@@ -318,14 +318,43 @@ function kit() {
 				})};\n`
 			);
 
+			// TODO where is this used?
 			process.env.SVELTEKIT_SERVER_BUILD_COMPLETED = 'true';
```
Rich-Harris, line 322: unrelated to this PR, but pretty sure we can get rid of this?

.changeset/witty-avocados-admire.md
benmccann: No. That should be updated. It accidentally got included from another PR I was considering sending

benmccann: No objections assuming it's not used anywhere. It was an internal-only thing, so we shouldn't worry about anyone else relying on it

packages/kit/src/vite/index.js
benmccann: No objections assuming it's not used anywhere. It was an internal-only thing, so we shouldn't worry about anyone else relying on it

.changeset/witty-avocados-admire.md
Rich-Harris:
```suggestion
prerender in a subprocess
```

Rich-Harris (approved):

PR #5683, opened 2022-07-23
Revert "set Vite's `publicDir` option"

Rich-Harris:
Reverts sveltejs/kit#5648.

I don't know this for certain (the stack trace is truncated, and I don't have time to investigate more deeply), but it's pretty clear that Vite is trying to process requests that originate from `<script>` and `<link>` elements, whereas previously we were bypassing its static file handling.

I'm just going to revert this now because it's bad breaking change that seems to have affected a lot of people — we can take another run at it later, when we can figure out how to make those requests bypass Vite.

### DIFF
--- a/.changeset/modern-jars-return.md
+++ b/.changeset/modern-jars-return.md
@@ -1,6 +1,0 @@
----
-'@sveltejs/adapter-node': patch
-'@sveltejs/kit': patch
----
-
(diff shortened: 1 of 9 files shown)


Rich-Harris:
ah shit, removing an old changeset doesn't cause a new changesets release...

Issue #5672, opened 2022-07-22
sourcing `js` file from static folder breaks on `1.0.0-next.390`


bhvngt:
### Describe the bug

After upgrading to the latest sveltekit version (1.0.0-next.390), I am not able to source `js` file from the static folder. Infact, it breaks for even css files. I believe this is due to #5648. Not sure, if there is any configuration that needs to be changed to make it work.

I am sourcing it by adding following code inside `app.html` head tag.

```
<script src="%sveltekit.assets%/sample.js"></script>
```

### Reproduction

[Reproduction](https://stackblitz.com/edit/sveltejs-kit-template-default-5jzsnx?file=static%2Fsample.js,package.json,src%2Fapp.html&terminal=dev)

### Logs

```shell
Failed to load url /sample.js (resolved id: /sample.js). This file is in /public and will be copied as-is during build without going through the plugin transforms, and therefore should not be imported from source code. It can only be referenced via HTML tags.
    at loadAndTransform (file:///home/projects/sveltejs-kit-template-default-5jzsnx/node_modules/vite/dist/node/chunks/dep-1513d487.js:39810:19
```

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (4) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    Yarn: 1.22.10 - /bin/yarn
    npm: 7.17.0 - /bin/npm
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.63 
    @sveltejs/kit: 1.0.0-next.390 => 1.0.0-next.390 
    svelte: ^3.46.0 => 3.49.0 
    vite: ^3.0.0 => 3.0.2
```

### Severity

blocking an upgrade

wickning1:
I'm getting this same error message with the global.css file after upgrading sveltekit and creating my vite.config.js for the first time. I'm not referencing it anywhere except the app.html.

homerjam:
I'm also experiencing this...
```html
<!-- app.html -->
...
<link rel="stylesheet" href="/fonts/fonts.css" />
...
```

[...]

PR #5686, opened 2022-07-23
Set Vite's publicDir and correctly serve public assets earlier in pipeline

benmccann:
This is a second stab at https://github.com/sveltejs/kit/pull/5648 which was reverted for causing https://github.com/sveltejs/kit/issues/5672

The issue was that we were serving static files from within the Kit middleware, so the transform middleware was being executed on them. This moves static serving into its own middleware so that they're served without the transform middleware being called on them

### DIFF
--- a/packages/kit/src/vite/dev/index.js
+++ b/packages/kit/src/vite/dev/index.js
@@ -181,8 +181,7 @@ export async function dev(vite, vite_config, svelte_config) {
 	});
 
 	const serve_static_middleware = vite.middlewares.stack.find(
-		(middleware) =>
-			/** @type {function} */ (middleware.handle).name === 'viteServeStaticMiddleware'
+		(middleware) => /** @type {function} */ (middleware.handle).name === 'viteServeStaticMiddleware'
 	);
 
 	remove_static_middlewares(vite.middlewares);
@@ -213,7 +212,6 @@ export async function dev(vite, vite_config, svelte_config) {
 	});
 
 	return () => {
-
 		vite.middlewares.use(async (req, res) => {
 			try {
 				if (!req.url || !req.method) throw new Error('Incomplete request');
--- a/packages/kit/src/vite/dev/index.js
+++ b/packages/kit/src/vite/dev/index.js
@@ -213,7 +213,8 @@ export async function dev(vite, vite_config, svelte_config) {
 
 	return () => {
 		const serve_static_middleware = vite.middlewares.stack.find(
-			(middleware) => /** @type {function} */ (middleware.handle).name === 'viteServeStaticMiddleware'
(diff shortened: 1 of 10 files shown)


bluwy (approved):

Rich-Harris:
Can we add a test to this to guard against a recurrence of #5672?

benmccann:
Yes, thanks for the reminder. I had intended to do that. Done!

packages/kit/src/vite/dev/index.js
```diff
@@ -180,13 +180,44 @@ export async function dev(vite, vite_config, svelte_config) {
 		extensions: []
 	});
 
+	vite.middlewares.use(async (req, res, next) => {
+		try {
+			if (!req.url || !req.method) throw new Error('Incomplete request');
```
Rich-Harris, line 185: is this actually a thing that can happen? if i remove this line and get rid of the `try-catch`, TypeScript doesn't complain

benmccann: No, I don't think we need to worry about it. I had copied it from the main middleware. I've removed it in both places. The `try`/`catch` on the otherhand can't be removed

Rich-Harris: Are you sure? I don't see any code there that's in danger of throwing

benmccann: If you try to remove it there's a test that fails with the error below showing where it can throw. It must bring down the whole server because all subsequent tests then fail.

```
[WebServer] file:///home/bmccann/src/kit/packages/kit/src/vite/dev/index.js:188
			const decoded = decodeURI(new URL(base + req.url).pathname);
			                ^

URIError: URI malformed
    at decodeURI (<anonymous>)
    at packages/kit/src/vite/dev/index.js:188:20
    at call (node_modules/.pnpm/vite@3.0.0/node_modules/vite/dist/node/chunks/dep-561c5231.js:45966:7)
    at next (node_modules/.pnpm/vite@3.0.0/node_modules/vite/dist/node/chunks/dep-561c5231.js:45910:5)
    at Function.handle (node_modules/.pnpm/vite@3.0.0/node_modules/vite/dist/node/chunks/dep-561c5231.js:45913:3)
    at Server.app (node_modules/.pnpm/vite@3.0.0/node_modules/vite/dist/node/chunks/dep-561c5231.js:45778:37)
    at Server.emit (node:events:527:28)
    at parserOnIncoming (node:_http_server:956:12)
    at HTTPParser.parserOnHeadersComplete (node:_http_common:128:17)
```

Rich-Harris: ah, gotcha. fair enough

PR #5694, opened 2022-07-24
feat: provide Vite hooks api

### LINKED ISSUE #5709: Vite Hooks Api
### Describe the problem

The SvelteKit's prerender and adapter process can have a race condition with another configured Vite plugins on `writeBundle`  and/or `closeBundle` hooks.

Some Vite plugins will require to run after `prerender` and before `adapter` processes: need access to collect prerendered pages and the result should be copied to the adapter output build folder (for example sitemap generation and pwa service worker).

Vite will run in parallel `writeBundle` and `closeBundle` plugins hooks and so we end up with the race condition, let me explain:
1) kit plugin + XXX plugin with async `closeBundle` and `enforce: 'pre'`: the XXX plugin will collect the prerendered/pages folder entries and generates a file in the kit output folder
2) kit plugin + YYY plugin with async `writeBundle` and `enforce: 'post'`: the YYY plugin will collect the prerendered/pages folder entries and generates a file in the kit output folder

In case 1), the XXX plugin will collect the pages properly, since it will run in the `closeBundle` hook, but since it will run in parallel, the adapter (that runs in the kit `closeBundle`) can finish before the XXX plugin finish run its `closeBundle` hook, and so the result it missing in the adapter output build folder.

In case 2), the YYY plugin will not collect the pages properly, since it will run in the `writeBundle` hook in parallel, so it can finish before the `prerender` process finish.

A good example is the `vite-plugin-pwa`: this plugin will collect all pages in the prerendered/pages to add them to the service worker's precache manifest (to allow work in offline), once collected it needs to build the service worker.
If we configure the `vite-plugin-pwa` build plugin with previous case 1), then the service worker is not copied to the adapter build output folder. If we configure it with previous case 2), then, the service worker is copied to the adapter build output folder, but the service worker's precache manifest will have not the pages.

Add build hooks to SvelteKit, it will control the internal process: allow configure Vite plugins via Rollup intercommunication plugin api and look for kit hooks to notify internal build process state.

### Alternatives considered

Add a new `preCloseBundle` hook in Vite: https://github.com/vitejs/vite/pull/9326

### Importance

### Additional Information

_No response_

userquin:
This PR will expose some hooks to avoid race condition on Vite, since all plugins will run in parallel and some will need the prerender process to finish: pwa plugin and sitemap for example.

There is a thread in Vite Discord server: https://discord.com/channels/804011606160703521/1000700061094776832

The latest changes @benmccann is making to the alignment of Vite and SvelteKit will not work with some plugins, even though the `process.exit` in the `closeBundle` has been removed. The main problem is that the async and parallel types of the plugins involved cause a race condition to exist on either `writeBundle` or `closeBundle`.

I also send a PR to include a new `preCloseBundle` hook: https://github.com/vitejs/vite/pull/9326

closes #5709

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/vite/index.js
+++ b/packages/kit/src/vite/index.js
@@ -15,7 +15,7 @@ import { get_runtime_directory, logger } from '../core/utils.js';
 import { find_deps, get_default_config as get_default_build_config } from './build/utils.js';
 import { preview } from './preview/index.js';
 import { get_aliases, resolve_entry } from './utils.js';
-import { create_vite_hooks_api } from './api.js';
+import { call_vite_config_api, call_vite_prerendered_api } from './api.js';
 
 const cwd = process.cwd();
 
@@ -58,10 +58,11 @@ const enforced_config = {
 };
 
 /**
+ * @param {import('types').ViteKitOptions} options
  * @return {import('vite').Plugin[]}
  */
-export function sveltekit() {
-	return [...svelte(), kit()];
+export function sveltekit(options) {
+	return [...svelte(), kit(options ?? {})];
 }
 
 /**
@@ -74,9 +75,10 @@ export function sveltekit() {
  * - https://rollupjs.org/guide/en/#build-hooks
  * - https://rollupjs.org/guide/en/#output-generation-hooks
  *
+ * @param {import('types').ViteKitOptions} options
  * @return {import('vite').Plugin}
  */
-function kit() {
+function kit(options) {
 	/** @type {import('types').ValidatedConfig} */
 	let svelte_config;
 
@@ -115,7 +117,7 @@ function kit() {
 
 	let completed_build = false;
 
-	const { call_config_hooks, call_prerendered_hooks, api } = create_vite_hooks_api();
+	// const { call_config_hooks, call_prerendered_hooks, api } = create_vite_hooks_api();
 
 	function vite_client_build_config() {
 		/** @type {Record<string, string>} */
@@ -197,7 +199,7 @@ function kit() {
 				const warning = warn_overridden_config(config, new_config);
 				if (warning) console.error(warning + '\n');
 
-				await call_config_hooks(svelte_config);
+				await call_vite_config_api(config, options, svelte_config);
 
 				return new_config;
 			}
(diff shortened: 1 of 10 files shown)


packages/kit/src/vite/index.js
```diff
@@ -194,6 +197,8 @@ function kit() {
 				const warning = warn_overridden_config(config, new_config);
 				if (warning) console.error(warning + '\n');
 
+				await call_config_hooks(svelte_config);
```
userquin, line 200: maybe it would be better to move this call to `configResolved` hook

userquin: I mean, there will be a race condition, another plugins can register in the kit hooks in their `config` hook, and so will be called on `configResolved` kit plugin hook

userquin:
I have also excluded IntelliJ/WebStorm stuff in `.gitignore` file.

userquin:
Maybe we can reverse the control, just registering the plugin names in the kit config and find them in the call, and so kit will control it, vite plugins will not need to do anything: kit logic will just check for api.hook funtión in each configured plugin.

userquin:
I just added in prerendering test a log with a dummy plugin: https://imgur.com/AekOLxE

userquin:
you should change the netlify adapter build script, it is failing in Windows  (it is using `cp` to copy a js file)

Rich-Harris:
Thank you but can you please open an issue before creating very large PRs like this? It adds a lot of code and complexity and it's completely unclear to me what it does or what problem it solves.

userquin:
> Thank you but can you please open an issue before creating very large PRs like this? It adds a lot of code and complexity and it's completely unclear to me what it does or what problem it solves.

I'll try to link a new issue later

userquin:
> Thank you but can you please open an issue before creating very large PRs like this? It adds a lot of code and complexity and it's completely unclear to me what it does or what problem it solves.

Created issue: since SvelteKit has removed the `process.exit` in the `closeBundle` hook, we can also remove the `onKitAdapter` hook, it can be done with a plugin with `closeBundle` hook configured after kit plugin.

benmccann:
I think https://github.com/vitejs/vite/discussions/9442 is a better solution to this problem, so let's see what happens there first

Rich-Harris:
Came here to post the same thing as @benmccann. This is a problem that should be solved at the Vite level. Will close this PR, thanks

PR #5701, opened 2022-07-25
📝 docs: Added documentations about Server Sent Events (SSE)

bastiaanv:
Adds documentation about Server Sent Events. See #887

### Tests

### Changesets

### DIFF
--- a/documentation/docs/03-routing.md
+++ b/documentation/docs/03-routing.md
@@ -419,3 +419,73 @@ assert.equal(
 ```
 
 To express a `%` character, use `%25`, otherwise the result will be malformed.
+
+### Server Sent Events (SSE)
+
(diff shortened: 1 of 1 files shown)


dummdidumm:
This is a good recipe, though I'm not sure if this is the right place for this. The other docs are related to explaining SvelteKit concepts, this is more of a how-to. We don't have a good section for things like this yet (learn.svelte.dev could become that).

Rich-Harris:
Agree, this is useful stuff but unfortunately it doesn't belong in the reference docs, it belongs in something like https://learn.svelte.dev/recipes (spoiler: that link 404s), so I have to close it — thank you and sorry!

There's also a significant wrinkle here: it assumes a single origin server. If you deploy your app to one of the many serverless platforms, different clients will connect to different instances of the server, which makes the whole thing fall apart. To build an app like the one in this PR we'd really need something along the lines of https://deno.com/deploy/docs/runtime-broadcast-channel; whether it's possible to expose that idea in a cross-platform way is very much TBD.

Issue #4423, opened 2022-03-22
Add Vitest to `create-svelte`


jtlapp:
### Describe the problem

I have spent many hours trying to figure out the best way to do unit testing with playwright. Sometimes I make progress, sometimes I get stuck. Right now I've got basic tests working but I'm stuck on mocking dependent modules.

At one point I tried to install jest but could not get it working with SvelteKit's configuration files. I'm thinking that the problem was supporting TypeScript. (That was two weeks ago, so I don't remember the details.)

Apparently, [playwright is not really designed for unit testing](https://github.com/microsoft/playwright/issues/8076), which probably explains Google's inability to pull up anything helpful.

So I'm here feeling displeased with SvelteKit. Something about the default installation is incompatible with jest or with jest + TypeScript. If some solutions are incompatible, what does SvelteKit recommend?

### Describe the proposed solution

I'd like to see something in the docs on doing unit testing -- running purely in node.js -- and maybe even an example unit test. I need a solution that supports mocking dependencies.

I'm not sure whether it's SvelteKit's support for unit testing that's incomplete or SvelteKit's support for TypeScript, but it looks like one of these to me.

### Importance

i cannot use SvelteKit without it

dummdidumm:
This may help: https://github.com/rossyman/svelte-add-jest

jtlapp:
Wow! That did it! I'm surprised I hadn't come across this before. Thank you so much!

There is one oddity: it completely bumped the playwright `test` script:

![Screen Shot 2022-03-22 at 13 43 16](https://user-images.githubusercontent.com/138133/159553411-a06c16f9-2021-45a8-87e6-343e4240c1fa.png)

But I am still able to run playwright tests from the command line, so I'll be able to add playwright back in.

jtlapp:
Except now I'm off trying to figure out why vite is pulling in `.spec.ts` files, despite `tsconfig.json` containing `"exclude": ["src/**/*.spec.ts"]`:

```
1:59:06 PM [vite] Error when evaluating SSR module /Users/joe/repos/cavesite/src/routes/index.spec.ts:
ReferenceError: describe is not defined
```

jtlapp:
But I'm having import problems. It's not pulling in postgres. These are unit tests I had working under playwright, converted to jest.

![capture1](https://user-images.githubusercontent.com/138133/159572135-f9d0c289-8baf-4a11-bf5d-2309d674a9ad.png)

[...]

PR #5708, opened 2022-07-25
Added Option to Automatically Add Vitest to New Projects using svelte-create

bertybot:
## Issue
Getting started with unit testing in Sveltekit can be confusing, which was brought up in this issue here https://github.com/sveltejs/kit/issues/4423

## Resolution
- Added the ability to automatically add Vitest with example tests to new projects using the svelte-create package. This should help people new to unit tests in Sveltekit get started with reasonable defaults by scaffolding the configuration for them, and giving them example tests to start out with.

## Testing
- Manually tested this with every config to make sure it worked.

## Possible problems
- I have not tested this Vitest config when emulating Sveltekit specific behavior such as `$app/env` and `$app/stores` I might need input from @benmccann  on this one...


packages/create-svelte/shared/+vitest+default/src/lib/Counter.test.ts
```diff
@@ -0,0 +1,31 @@
+import { render, screen, fireEvent } from '@testing-library/svelte';
```
benmccann, line 1: This perhaps deserves some discussion with the other maintainers, but my belief is we probably would not want to add a dependency on `@testing-library/svelte`. I've seen a few different solutions used for component testing and it's not obvious to me that there's a clear leader. For now, I would just have Vitest test a simple function and not test any `.svelte` file until there's a clearer consensus around how'd we like to recommend testing individual components.

I started a discussion about this here: https://github.com/sveltejs/kit/discussions/5285

bertybot: Okay, I'll do that. I will keep a separate branch with  `@testing-library/svelte` just, in case we change our mind later.

bertybot: Updated

packages/create-svelte/shared/+vitest-typescript/vitest.config.js
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from 'vite';
+import viteConfig from './vite.config';
+
+export default defineConfig({
```
benmccann, line 4: why do we need a separate `vitest.config.js` vs just using the existing `vite.config.js`?

benmccann (commented):
What is it that requires `+vitest+playwright+typescript` setup? Is there something in the Playwright config that needs to change?

packages/create-svelte/shared/+vitest-typescript/vitest.config.js
bertybot: > What is it that requires `+vitest+playwright+typescript` setup? Is there something in the Playwright config that needs to change?

When playwright is installed in need to edit the playwright config to only look for test in the test folder otherwise it would capture the vitest  tests too.

bertybot: > why do we need a separate `vitest.config.js` vs just using the existing `vite.config.js`?

Admittedly pretty wonky I think I originally did this to avoid touching the vite.config, since I felt like setting up the vitest config was safer.  Ill fix that

bertybot: fixed to add testing options to vite.config.js instead of adding vitest.config file

dominikg:
Do we want to add c8 + coverage script by default?
I think vitest asks you to install c8 if you run --coverage without it. Adding it by default means every user gets it, even if they are not interested in it.

dominikg:
I'd skip it until we have an idea/documentation to actually build a full coverage test suite, which may need an additional util for mocking kit things as required. 

Eg. hooks, page-endpoints and maybe even some parts of .svelte files are very difficult to test right now. 

So a coverage report by c8 would be very inaccurate, either by giving a false sense of security when showing 100% even though it's not seeing everything, or by showing 50% and the user has no way of getting a higher coverage because tools are missing to implement tests.

packages/create-svelte/shared/+vitest/vite.config.js
```diff
@@ -0,0 +1,12 @@
+import { sveltekit } from '@sveltejs/kit/vite';
+
+/** @type {import('vite').UserConfig} */
+const config = {
+	plugins: [sveltekit()],
+	test: {
+		include: ['src/**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}'],
```
dominikg, line 7: should we reduce the pattern to less combinations? projects are always esm so we could just use one extension (.js for js projects, .ts for typescript).

If and do we need both spec and test postfixes?

This is not a request, more the start of a mini-discussion

packages/create-svelte/shared/+vitest/vite.config.js
```diff
@@ -0,0 +1,12 @@
...
+		globals: true
```
dominikg, line 8: Another discussion ;) globals: true would mean we'd have to also provide additional env typing for the test files, just leave it off and resort to imports? It's cleaner anyways

bertybot: Yea, just gonna admit that a lot of my decisions, have been based off what I've seen other people doing, and what my team has been doing. 

I took a lot of inspiration from @davipon on his add vitest library
https://github.com/davipon/svelte-add-vitest/blob/main/templates/config/vite.config.js

I have yet to see someone not use globals I think because having to add them for every single test starts to get annoying, adding the imports just becomes repetitive boilerplate after a while. Definitely a preference though, the script already adds the vitest types to the tsconfig or Jsconfig based on what you choose.

packages/create-svelte/shared/+vitest/vite.config.js
bertybot: I think I just copied this off the [default for Vitest include](https://vitest.dev/config/#configuration) and added the src so that it targeted the Sveltekit source folder. Do agree with your sentiment though, I just thought it would be better to more align with the defaults and allow people to override them to what they need, but I wouldn't be apposed to just having it check 'src/**/*.test.{js,ts}'

Rich-Harris: agree with that latter suggestion — updated

packages/create-svelte/shared/+vitest+playwright+typescript/playwright.config.ts
```diff
@@ -0,0 +1,11 @@
+import type { PlaywrightTestConfig } from '@playwright/test';
+
+const config: PlaywrightTestConfig = {
+	webServer: {
+		command: 'npm run build && npm run preview',
+		port: 4173
+	},
+	testMatch: 'tests/**/*.{js,ts}'
```
Rich-Harris, line 8: is there any reason not to just add this to the default `playwright.config.ts`? seems like a reasonable glob to use, and that way we don't need to maintain multiple variants

packages/create-svelte/shared/+vitest+playwright-typescript/playwright.config.js
```diff
@@ -0,0 +1,10 @@
+/** @type {import('@playwright/test').PlaywrightTestConfig} */
+const config = {
+	webServer: {
+		command: 'npm run build && npm run preview',
+		port: 4173
+	},
+	testMatch: 'tests/**/*.js'
```
Rich-Harris, line 7: see https://github.com/sveltejs/kit/pull/5708/files#r947378121

Rich-Harris (commented):
Thank you, this is very cool. I made one change to how `create-svelte` works — a `.ts` file in the `shared` directory that isn't explicitly inside a `+typescript` directory is used as the source for the TypeScript _and_ JSDoc _and_ vanilla versions (i.e. the `.ts` file includes types in both TypeScript and JSDoc, and we strip one or both depending on the variant).

This way, we can have a single `sum.ts` instead of maintaining two or three versions:

```ts
/**
 * Add two numbers
 * @param {number} a
 * @param {number} b
 */
export function sum(a: number, b: number): number {
	return a + b;
}
```

That said, I don't know that `sum` is the right thing to have here — ideally we'd add realistic tests of functions that are actually used in the demo app. (The current demo app is due an overhaul for a variety of reasons, so maybe we should think about what sort of thing would make for a good unit test.)

It would be great if we had an example of a component unit test as well. I know we don't necessarily want to add testing-library, but does anyone have any suggestions for good component testing patterns that don't rely on a library?

packages/create-svelte/shared/+vitest+playwright+typescript/playwright.config.ts
bertybot: This is a good suggestion, I just thought it was too opinionated to add since I thought some people might like to keep their test files with their source file. I agree though that they should all probably be in tests as default, So I am good adding this to the default.

dominikg:
there are rudimentary examples with and without testing-library-svelte in https://github.com/vitest-dev/vitest/tree/main/examples/svelte/test

I think adding testing-library for component tests is a better suggestion than everyone coming up with their own set of helpers.

benmccann:
I don't have any objection to `testing-library-svelte`. But what I am worried about is adding a component test when it's not at all clear to me whether components are better tested with Vitest or Playwright. See the discussion in https://github.com/sveltejs/kit/discussions/5285 for more details

manuganji:
I'm using the newly recommended `tests` directory for keeping my tests and I have a few components inside. But I can't import them into my tests. I went looking through this template for guidance on how to write a component test. 

> Error: [vite-node] Failed to load /tests/components/TwoSubmitButtons.svelte

Should I make a separate issue for this or do you think it's solvable with the right `vite.config::test` settings? If you know which setting, could you please point me to it? This is the testing code in question:

```
import Form from '$lib/components/Form.svelte';
import TwoSubmitButtonsSvelte from './components/TwoSubmitButtons.svelte';
import { findByTestId, fireEvent, render, within } from '@testing-library/svelte';
import type { SpyInstance } from 'vitest';

describe('Test', () => {
  it('should render children buttons', () => {
    const props = { schema: {} };
    const { container } = render(TwoSubmitButtonsSvelte);
    expect(container.querySelectorAll('button[type=submit]')).toHaveLength(2);
  });
});
```

packages/create-svelte/shared/+vitest/vite.config.js
```diff
@@ -0,0 +1,12 @@
+import { sveltekit } from '@sveltejs/kit/vite';
+
+/** @type {import('vite').UserConfig} */
+const config = {
+	plugins: [sveltekit()],
+	test: {
+		include: ['src/**/*.{test}.{js,ts}'],
```
manuganji, line 7: I think this should be `include: ['tests/**/*.test.[t,j]s'],` from my experience in my project.

Rich-Harris:
@manuganji please open a new issue for things like this — questions like that don't belong on a PR discussion. thanks

packages/create-svelte/shared/+playwright+typescript/playwright.config.ts
```diff
@@ -4,7 +4,8 @@ const config: PlaywrightTestConfig = {
 	webServer: {
 		command: 'npm run build && npm run preview',
 		port: 4173
-	}
+	},
+	testMatch: 'tests/**/*.{js,ts}'
```
benmccann, line 8: I think adding this here as Rich suggested is a decent idea, but I might tweak it a bit. By default playwright looks for `test` or `spec` in the name and it might not be a bad idea to keep their default setting, but just add the `tests/` directory prefix so that we don't try to run non-test files
```suggestion
	testMatch: 'tests/**/.*(test|spec)\.(js|ts|mjs)'
```

ZetiMente: But the Playwright runner is going to conflict with the VitTest runner. You can't keep both Playwright default and Vitest default settings if the goal is to have Vitest runner only run its test and Playwright run only its test. Or are we having a preference for one default runner? 

I choose to have each runner with it's own files bc the Playwright runner has extra functionality but it required separating the Playwright files to have .e2e  to avoid changing Vitest default settings. `testMatch: 'tests/e2e/**.e2e.js'` for playwright.

The other option would be to make test/unit folder and test/e2e folder and change both the testMatch for both Playwright and ViTest to read only their folders. Could also add an integration/component folder but it doesn't seem if a decision on whether to use testing-library with svelte or Playwright has been made there.

packages/create-svelte/scripts/build-templates.js
```diff
@@ -206,22 +206,41 @@ async function generate_shared() {
 		}
 
 		if (name.endsWith('.ts') && !include.includes('typescript')) {
+			// file includes types in TypeScript and JSDoc —
+			// create .js file, with and without JSDoc
```
benmccann, line 210:
```suggestion
			// contents includes types in TypeScript and JSDoc
			// we will transform it to a version appropriate for each type checking option
```

packages/create-svelte/shared/+playwright+typescript/playwright.config.ts
benmccann: Rich and I were thinking that you'd differentiate by directory. So the Playwright tests would be in `tests/` and the Vitest tests would be in `src/`. I don't know if that's the most common way of doing it or not, but that's what we've been doing in SvelteKit's own tests and it's be working pretty well there

packages/create-svelte/shared/+vitest/src/lib/sum.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Add two numbers
+ * @param {number} a
+ * @param {number} b
+ */
+export function sum(a: number, b: number): number {
```
benmccann, line 6: it's weird that this isn't used anywhere except in the tests

One idea might be to convert the `<a href="https://kit.svelte.dev">kit.svelte.dev</a>` used in `+page.svelte` to a function like `{@html linkify('kit.svelte.dev')}`:
```
export function linkify(domain: string) {
  return `<a href="https://${domain}">${domain}</a>`;
}
```

benmccann:
I pushed some updates, so this should hopefully be close to being able to be merged. The one remaining item is @dominikg's comments about global types: https://github.com/sveltejs/kit/pull/5708#pullrequestreview-1051421077. Assuming we want to use globals, perhaps someone more familiar with TypeScript could add the types

dominikg:
I don't recommend using vite.config.ts . The way vite creates a temporary vite.config.hash.js isn't great imho.

dominikg:
If we suggest colocating test files, there should be documentation about how to handle + prefixed files.
See https://github.com/sveltejs/kit/issues/7121

dummdidumm (approved):

benmccann (approved):

dominikg:
unit test files are colocated in src, do we need to update tsconfig.json or show how that works at least so that users can have different ts settings for unit vs app code? or should the move it out of src in that case?

I'd really love to avoid going down the multiple tsconfig files road but i'm not sure you can have both playwright and vitest .spec.ts files with different gobal needs from a single tsconfig

benmccann:
@dominikg why would you need different TypeScript settings for unit vs app code?

dominikg:
> @dominikg why would you need different TypeScript settings for unit vs app code?

eg node env vs not

benmccann:
Isn't that just defined once per config? If you want multiple, I think you have to have multiple configs regardless of what directories things live in? I don't have a ton of experience writing tests in TypeScript projects, so I'm not understanding what alternative is being proposed and if there's some best practice around this that you think this PR might be violating?

dominikg:
it depends, for basic unit tests sharing the same tsconfig works. but I've seen multiple projects splitting it into a separate tsconfig.test.json with include and exclude. due to the way kit generates the root tsconfig that could become difficult.

we can start with this setup but advanced usecases will require additional setup which isn't clear and there is no docs about it.

dominikg:
is https://github.com/vitest-dev/vitest/issues/2298 a blocker here? not supporting these makes some unit tests harder/ impossible

Rich-Harris:
thanks all! very happy to get this in

PR #5730, opened 2022-07-27
[breaking] move packaging functionality into its own package

dummdidumm:
# How to migrate

1. `npm install -D @sveltejs/package` (or pnpm, or yarn)
2. If you have configured some package options, move them to the top level of `svelte.config.js`:
```diff
const config = {
	kit: {
- 		package: {..}
	},
+	package: {..}
};
```
3. Replace `svelte-kit package` with `svelte-package` in your `package.json` script

# PR description

Draft PR to move the package functionality out of SvelteKit. This makes the following decisions (all up for discussion):
- New package is called `@sveltejs/package` 
- Package has its own CLI, `svelte-package`, `svelte-kit package` will be removed
- Minimal SvelteKit interop; `kit.files.lib` is used to set `source` if `source` is unset
- We don't call `svelte-kit sync` in this package because you can use it standalone, but the template will call it before calling the package command

TODOS:
- SvelteKit and this new package share some utility code around filesystem stuff and options parsing. How do we best share this, if at all? Should we create a private package in the monorepo and enhance the rollup config to inline it, so no visible changes to the outside? Sounds like the best solution to me
- How to deploy this as a new package? Changeset masterminds please help. Is it enough to create a changeset and the rest is taken care of?
- Adjust SvelteKit code: remove packaging, remove the config (error and tell to switch to this new package)

### Tests

### Changesets


dummdidumm:
Tests for `create-svelte` are failing because the package isn't available yet I think

Rich-Harris:
Did as much of the grunt work of updating this post-#5748 as possible. Since `kit` is no longer bundled, the `@internal/shared` package isn't really an option unless we were to actually publish it to npm. I think it's preferable to just copy the few utility functions `kit` and `package` have in common, unless we find ourselves using them in more places.

I think we should get rid of the `exports` and `files` options in favour of having people include the `"exports"` field in their package.json files, pointing at the `src` code, so that the generated `pkg.exports` has the same shape but points to the package directory. This would allow people to use packages in a monorepo without needing to constantly build them. The `exports` would be treated as entry points, and we'd follow all the imports from those entry points to figure out which files need to be included in the package, making `files` unnecessary.

I don't have a strong opinion about whether that should happen in this PR or a follow-up.

dummdidumm:
I think we should merge this as-is and do the other stuff in follow-up PRs.

Rich-Harris (approved):

PR #5739, opened 2022-07-28
fix: `$env/*/private` throws for client-side imports on Windows and Tailwind doesn't freak out

### LINKED ISSUE #5723: $env/static/private can be imported in client modules
### Describe the bug

If I understood correctly, mechanisms should be in place that prevent $env/static/private to be imported in client modules (like routes). However, this does not seem to be the case. I can import from $env/static/private and the variables get transpiled into the client side javascript modules.

My excuses if this feature is not yet supposed to work.

### Reproduction

A simple repo showing the issue:
https://github.com/aloker/sveltekit-private-env

https://github.com/aloker/sveltekit-private-env/blob/main/src/routes/index.svelte

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19043
    Memory: 8.81 GB / 31.81 GB
  Binaries:
    Node: 16.15.1 - c:\tools\node\node.EXE
    Yarn: 1.22.19 - c:\tools\node\yarn.CMD
    npm: 8.13.2 - c:\tools\node\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (103.0.1264.71)
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.396 
    svelte: ^3.44.0 => 3.49.0 
    vite: ^3.0.0 => 3.0.3
```

### Severity

### Additional Information

_No response_

### LINKED ISSUE #5747: new env handling seems to break tailwindcss
### Describe the bug

while `@tailwind base` is in my app.css i get this error:

```
Error: Cannot import $env/dynamic/private.js into client-side code:
- src/routes/__layout.svelte 
  - src/routes/app.css 
    - $env/dynamic/private.js 
```

basicly the same error as https://github.com/sveltejs/kit/issues/5720

### Reproduction

https://github.com/benjaminlewandowski/sveltekit-tailwindcss-bugreport

### Logs

_No response_

### System Info

```shell
System:
    OS: Linux 5.18 Fedora Linux 36 (Workstation Edition)
    CPU: (16) x64 Intel(R) Core(TM) i7-10875H CPU @ 2.30GHz
    Memory: 12.38 GB / 31.04 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 18.6.0 - /run/user/1000/fnm_multishells/288221_1659039694408/bin/node
    Yarn: 1.22.19 - ~/.npm-global/bin/yarn
    npm: 8.13.2 - /run/user/1000/fnm_multishells/288221_1659039694408/bin/npm
  Browsers:
    Brave Browser: 103.1.41.100
    Chrome: 103.0.5060.134
    Firefox: 102.0
  npmPackages:
    @sveltejs/adapter-auto: 1.0.0-next.64 => 1.0.0-next.64 
    @sveltejs/kit: 1.0.0-next.396 => 1.0.0-next.396 
    svelte: 3.49.0 => 3.49.0 
    vite: 3.0.3 => 3.0.3
```

### Severity

blocking an upgrade

### Additional Information

demo repo is a minimal setup with that error

tcc-sejohnson:
Closes #5723 and closes #5747. 

The paths for both Vite and Rollup's module graphs needed to be normalized, along with the illegal import paths.

Because Tailwind needs to analyze basically every file in your project, Vite thinks it imports all of them, meaning it'll almost always import `$env/*/private`, causing our helpful module analysis to be, well, unhelpful. In dev, we now ignore non-`js|ts|svelte` files. Rollup doesn't have this problem (because it doesn't have to worry about HMR), so we can be stricter and just analyze all imports. I think this strikes a happy medium. "Provide armor against footguns in dev, and provide even better armor when we try to go to prod."

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/vite/utils.js
+++ b/packages/kit/src/vite/utils.js
@@ -250,42 +250,48 @@ const find_illegal_rollup_imports = (
 };
 
 /**
- * Throw an error if a private module is imported from a client-side node.
- * @param {import('vite').ModuleNode} node
- * @param {Set<string>} illegal_imports Illegal module IDs -- be sure to call vite.normalizePath!
- * @param {string} out_dir The directory specified by config.kit.outDir
- */
-export function prevent_illegal_vite_imports(node, illegal_imports, out_dir) {
-	const chain = find_illegal_vite_imports(node, illegal_imports);
-	if (chain) throw new Error(format_illegal_import_chain(chain, out_dir));
-}
-
-/*
  * Vite does some weird things with import trees in dev
  * for example, a Tailwind app.css will appear to import
  * every file in the project. This isn't a problem for
  * Rollup during build.
+ * @param {Iterable<string>} config_module_types
  */
-const module_types = new Set([
-	'.ts',
-	'.js',
-	'.svelte',
-	'.mts',
-	'.mjs',
-	'.cts',
-	'.cjs',
-	'.svelte.md',
-	'.svx',
-	'.md'
-]);
+const get_module_types = (config_module_types) => {
+	return new Set([
+		'.ts',
+		'.js',
+		'.svelte',
+		'.mts',
+		'.mjs',
+		'.cts',
+		'.cjs',
+		'.svelte.md',
+		'.svx',
+		'.md',
+		...config_module_types
+	]);
+};
+
+/**
+ * Throw an error if a private module is imported from a client-side node.
+ * @param {import('vite').ModuleNode} node
+ * @param {Set<string>} illegal_imports Illegal module IDs -- be sure to call vite.normalizePath!
+ * @param {Iterable<string>} module_types File extensions to analyze in addition to the defaults: `.ts`, `.js`, etc.
+ * @param {string} out_dir The directory specified by config.kit.outDir
+ */
+export function prevent_illegal_vite_imports(node, illegal_imports, module_types, out_dir) {
+	const chain = find_illegal_vite_imports(node, illegal_imports, get_module_types(module_types));
+	if (chain) throw new Error(format_illegal_import_chain(chain, out_dir));
+}
 
 /**
  * @param {import('vite').ModuleNode} node
  * @param {Set<string>} illegal_imports Illegal module IDs -- be sure to call vite.normalizePath!
+ * @param {Set<string>} module_types File extensions to analyze: `.ts`, `.js`, etc.
  * @param {Set<string>} seen
  * @returns {Array<import('types').ImportNode> | null}
  */
-function find_illegal_vite_imports(node, illegal_imports, seen = new Set()) {
+function find_illegal_vite_imports(node, illegal_imports, module_types, seen = new Set()) {
 	if (!node.id) return null; // TODO when does this happen?
 	const name = normalizePath(node.id);
 
@@ -297,7 +303,7 @@ function find_illegal_vite_imports(node, illegal_imports, seen = new Set()) {
 	}
 
 	for (const child of node.importedModules) {
-		const chain = child && find_illegal_vite_imports(child, illegal_imports, seen);
+		const chain = child && find_illegal_vite_imports(child, illegal_imports, module_types, seen);
 		if (chain) return [{ name, dynamic: false }, ...chain];
 	}
 
--- a/packages/kit/src/vite/utils.js
+++ b/packages/kit/src/vite/utils.js
@@ -1,6 +1,6 @@
 import fs from 'fs';
 import path from 'path';
-import { loadConfigFromFile, loadEnv } from 'vite';
+import { loadConfigFromFile, loadEnv, normalizePath } from 'vite';
 import { get_runtime_directory } from '../core/utils.js';
 
 /**
@@ -200,7 +200,7 @@ export function get_env(mode, prefix) {
 /**
  * @param {(id: string) => import('rollup').ModuleInfo | null} node_getter
  * @param {import('rollup').ModuleInfo} node
- * @param {Set<string>} illegal_imports Illegal module IDs -- be sure to call path.normalize!
+ * @param {Set<string>} illegal_imports Illegal module IDs -- be sure to call vite.normalizePath!
  * @param {string} out_dir The directory specified by config.kit.outDir
  */
 export function prevent_illegal_rollup_imports(node_getter, node, illegal_imports, out_dir) {
@@ -212,7 +212,7 @@ export function prevent_illegal_rollup_imports(node_getter, node, illegal_import
  * @param {(id: string) => import('rollup').ModuleInfo | null} node_getter
  * @param {import('rollup').ModuleInfo} node
  * @param {boolean} dynamic
- * @param {Set<string>} illegal_imports Illegal module IDs -- be sure to call path.normalize!
+ * @param {Set<string>} illegal_imports Illegal module IDs -- be sure to call vite.normalizePath!
  * @param {Set<string>} seen
  * @returns {Array<import('types').ImportNode> | undefined}
  */
@@ -223,7 +223,7 @@ const find_illegal_rollup_imports = (
 	illegal_imports,
 	seen = new Set()
 ) => {
-	const nodeId = path.normalize(node.id);
+	const nodeId = normalizePath(node.id);
 	if (seen.has(nodeId)) return;
 	seen.add(nodeId);
 
@@ -249,7 +249,7 @@ const find_illegal_rollup_imports = (
 /**
  * Throw an error if a private module is imported from a client-side node.
  * @param {import('vite').ModuleNode} node
- * @param {Set<string>} illegal_imports Illegal module IDs -- be sure to call path.normalize!
+ * @param {Set<string>} illegal_imports Illegal module IDs -- be sure to call vite.normalizePath!
  * @param {string} out_dir The directory specified by config.kit.outDir
  */
 export function prevent_illegal_vite_imports(node, illegal_imports, out_dir) {
@@ -259,13 +259,13 @@ export function prevent_illegal_vite_imports(node, illegal_imports, out_dir) {
 
 /**
  * @param {import('vite').ModuleNode} node
- * @param {Set<string>} illegal_imports Illegal module IDs -- be sure to call path.normalize!
+ * @param {Set<string>} illegal_imports Illegal module IDs -- be sure to call vite.normalizePath!
  * @param {Set<string>} seen
  * @returns {Array<import('types').ImportNode> | null}
  */
 function find_illegal_vite_imports(node, illegal_imports, seen = new Set()) {
 	if (!node.id) return null; // TODO when does this happen?
-	const nodeId = path.normalize(node.id);
+	const nodeId = normalizePath(node.id);
 	if (seen.has(nodeId)) return null;
 	seen.add(nodeId);
 
--- a/packages/kit/src/vite/utils.js
+++ b/packages/kit/src/vite/utils.js
@@ -223,29 +223,29 @@ const find_illegal_rollup_imports = (
 	illegal_imports,
 	seen = new Set()
 ) => {
-	const nodeId = normalizePath(node.id);
-	if (seen.has(nodeId)) return null;
-	seen.add(nodeId);
+	const name = normalizePath(node.id);
+	if (seen.has(name)) return null;
+	seen.add(name);
 
-	if (illegal_imports.has(nodeId)) {
-		return [{ name: nodeId, dynamic }];
+	if (illegal_imports.has(name)) {
+		return [{ name, dynamic }];
 	}
 
 	for (const id of node.importedIds) {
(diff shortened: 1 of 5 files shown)


dominikg:
is it possible to use vite's normalizePath here? https://vitejs.dev/guide/api-plugin.html#path-normalization

tcc-sejohnson:
I was hoping someone more knowledgeable would come along and make my day.

tcc-sejohnson:
> I'm surprised none of the browser tests failed on Windows. Maybe we need one that uses `$env/*/private`?

The problem is that these are buildtime/dev-time errors, meaning we'd need a test project dedicated to attempting to build and ensuring the build fails. There are a number of things that'd be useful for -- I'm happy to put an implementation together, but probably not for this PR?

benmccann:
I'm not sure you'd need anything too special or a separate test app. If you fail the build on one of the existing test apps it should turn the CI red

tcc-sejohnson:
> I'm not sure you'd need anything too special or a separate test app. If you fail the build on one of the existing test apps it should turn the CI red

Yes, but we're trying to verify that the build _does_ fail, and the build needs to _not_ fail in order for the rest of the tests to be able to run. Hence, the "metaproject" that tries to build purposefully-failing projects and verifies that they actually do fail.

benmccann:
Oh, right. Yeah, that'd be kind of a pain to test

tcc-sejohnson:
Rich, there's another env var bug I want to work on tonight, but it might step on this one's toes. Easier to merge this one first, if you think it looks good.

tcc-sejohnson:
Heads up @Rich-Harris -- I went ahead and fixed the other bug in this PR as well. It was a pretty simple change.

packages/kit/src/vite/utils.js
```diff
@@ -257,6 +260,14 @@ export function prevent_illegal_vite_imports(node, illegal_imports, out_dir) {
 	if (chain) throw new Error(format_illegal_import_chain(chain, out_dir));
 }
 
+/*
+ * Vite does some weird things with import trees in dev
+ * for example, a Tailwind app.css will appear to import
+ * every file in the project. This isn't a problem for
+ * Rollup during build.
+ */
+const module_types = new Set(['.js', '.ts', '.svelte']);
```
dominikg, line 269: do we need .mts,.mjs as well? or could it be inverted to stop looking when it encounters a `.css` ? Or is there another boolean on the module graph to see if it is a css module?

This is risky. Right now there is no great support for custom extensions, but it might change and at that point .anything could be a svelte file.

benmccann: I had suggested inverting seems risky. What else might this happen with? Postcss, less, etc.? I don't think we should ever support giving a different extension to a `.svelte` file. That just seems weird

dominikg: mdsvex with `.svx` or `.md` would like a word

dominikg: at the end of the day this is all tailwinds fault btw. it should not declare a dependency on a private env file in the first place

dominikg: looks like you can add configuration to tailwind to avoid this: https://github.com/sveltejs/kit/issues/5747#issuecomment-1199121656 

imho that's better than adding this restriction that might miss something.

Conduitry: Is there a reason to be telling Tailwind JIT to look at JS files anyway? Is that a recommendation on their website?

I agree that updating `tailwind.config.cjs` in a project seems better than adding this escape hatch here in Kit.

benmccann: > Is there a reason to be telling Tailwind JIT to look at JS files anyway? Is that a recommendation on their website?

That's the default config you get when setting up a new Tailwind project. The reason for it is here: https://github.com/sveltejs/kit/issues/5747#issuecomment-1199546971

Conduitry: Hmm, yeah, that's unfortunate. I'm not sure what I'd suggest here. I'm not saying I _never_ want to make a change within SvelteKit to accommodate this, but I do think we should revert that change from this PR and have it just be about the fairly uncontroversial Windows fix so we can get that out the door.

tcc-sejohnson: I get the reticence here, but it seems better to go this route (with the extra file extensions, probably?) and fix a bug that _actually prevents using SvelteKit_ than to sit around and try to find some other upstream fix. We're still being fully strict in `build`, so at worst, users would be able to develop their app and then have to go back and adjust some imports when they try to build.

tcc-sejohnson: Like, this issue is preventing _me_ from upgrading to use _the feature I implemented_, and there's no other way I can think of to fix it -- Tailwind really _is_ depending on those files during dev. This bug represents a complete blocker to using Tailwind in a SK project at all.

benmccann: Yeah, this seems like a pretty good compromise to me. Check `.js`/`.ts`/`.svelte` files during dev. Less common stuff like `.mdx` can get checked during build

tcc-sejohnson: Okay, quick poll on which filetypes to consider. I think we should at least add `mts|mjs`. I don't really know why we wouldn't go ahead and add the three common mdsvex types: `svelte.md|svx|md`. That should cover virtually all bases. Anything else is such an edge case that we'll catch it in build and not have to feel bad about it.

Conduitry: Hm. Yeah, I guess so. If we don't run into this sort of issue during `build` (i.e., we can continue to deny all imports of forbidden modules across the board), then during `dev` it's probably better to blacklist JS/TS/Svelte files than it is to whitelist CSS/Less/SASS/etc. files.

tcc-sejohnson: Yeah, this is _strictly_ a dev-time limitation. I think it has something to do with HMR.

If we can get a quick consensus on the filetypes I'll make an update and it'll be good to go. I think the complete list should be:

```
ts
js
svelte
mts
mjs
cts
cjs
svelte.md
svx
md
```

That's a relatively short but (I think?) all-encompassing list that should avoid weirdness about things that aren't actually imports and provide as close to a buildtime experience as possible.

dummdidumm: I approve that list

benmccann: sounds fine to me

tcc-sejohnson:
Okay, once we see the green christmas tree list we should be GTM @Rich-Harris

packages/kit/src/vite/utils.js
```diff
@@ -223,59 +223,84 @@ const find_illegal_rollup_imports = (
 	illegal_imports,
 	seen = new Set()
 ) => {
-	if (seen.has(node.id)) return;
-	seen.add(node.id);
+	const nodeId = normalizePath(node.id);
+	if (seen.has(nodeId)) return null;
+	seen.add(nodeId);
```
Rich-Harris, line 228:
```suggestion
	const name = normalizePath(node.id);
	if (seen.has(name)) return null;
	seen.add(name);
```

packages/kit/src/vite/utils.js
```diff
@@ -223,59 +223,84 @@ const find_illegal_rollup_imports = (
...
 
-	if (illegal_imports.has(node.id)) {
-		return [{ name: node.id, dynamic }];
+	if (illegal_imports.has(nodeId)) {
+		return [{ name: nodeId, dynamic }];
```
Rich-Harris, line 231:
```suggestion
	if (illegal_imports.has(name)) {
		return [{ name, dynamic }];
```

packages/kit/src/vite/utils.js
```diff
@@ -223,59 +223,84 @@ const find_illegal_rollup_imports = (
...
 	}
 
 	for (const id of node.importedIds) {
 		const child = node_getter(id);
 		const chain =
 			child && find_illegal_rollup_imports(node_getter, child, false, illegal_imports, seen);
-		if (chain) return [{ name: node.id, dynamic }, ...chain];
+		if (chain) return [{ name: nodeId, dynamic }, ...chain];
```
Rich-Harris, line 238:
```suggestion
		if (chain) return [{ name, dynamic }, ...chain];
```

packages/kit/src/vite/utils.js
```diff
@@ -223,59 +223,84 @@ const find_illegal_rollup_imports = (
...
 	}
 
 	for (const id of node.dynamicallyImportedIds) {
 		const child = node_getter(id);
 		const chain =
 			child && find_illegal_rollup_imports(node_getter, child, true, illegal_imports, seen);
-		if (chain) return [{ name: node.id, dynamic }, ...chain];
+		if (chain) return [{ name: nodeId, dynamic }, ...chain];
```
Rich-Harris, line 245:
```suggestion
		if (chain) return [{ name, dynamic }, ...chain];
```

packages/kit/src/vite/utils.js
```diff
@@ -223,59 +223,84 @@ const find_illegal_rollup_imports = (
...
 	}
+
+	seen.delete(nodeId);
```
Rich-Harris, line 248:
```suggestion
	seen.delete(name);
```

packages/kit/src/vite/utils.js
```diff
@@ -223,59 +223,84 @@ const find_illegal_rollup_imports = (
...
+	'.js',
+	'.svelte',
+	'.mts',
+	'.mjs',
+	'.cts',
+	'.cjs',
+	'.svelte.md',
+	'.svx',
+	'.md'
+]);
```
Rich-Harris, line 280: this hardcoded list makes me nervous. At the very least shouldn't it include `config.extensions` and `config.kit.moduleExtensions`?

packages/kit/src/vite/utils.js
```diff
@@ -223,59 +223,84 @@ const find_illegal_rollup_imports = (
...
 /**
  * @param {import('vite').ModuleNode} node
- * @param {Set<string>} illegal_imports
+ * @param {Set<string>} illegal_imports Illegal module IDs -- be sure to call vite.normalizePath!
  * @param {Set<string>} seen
  * @returns {Array<import('types').ImportNode> | null}
  */
 function find_illegal_vite_imports(node, illegal_imports, seen = new Set()) {
 	if (!node.id) return null; // TODO when does this happen?
+	const nodeId = normalizePath(node.id);
```
Rich-Harris, line 290:
```suggestion
	const name = normalizePath(node.id);
```

packages/kit/src/vite/utils.js
```diff
@@ -223,59 +223,84 @@ const find_illegal_rollup_imports = (
...
 
-	if (seen.has(node.id)) return null;
-	seen.add(node.id);
+	if (seen.has(nodeId) || !module_types.has(path.extname(nodeId))) return null;
+	seen.add(nodeId);
```
Rich-Harris, line 293:
```suggestion
	if (seen.has(name) || !module_types.has(path.extname(name))) return null;
	seen.add(name);
```

packages/kit/src/vite/utils.js
```diff
@@ -223,59 +223,84 @@ const find_illegal_rollup_imports = (
...
 
-	if (node.id && illegal_imports.has(node.id)) {
-		return [{ name: node.id, dynamic: false }];
+	if (nodeId && illegal_imports.has(nodeId)) {
+		return [{ name: nodeId, dynamic: false }];
```
Rich-Harris, line 296:
```suggestion
	if (name && illegal_imports.has(name)) {
		return [{ name, dynamic: false }];
```

packages/kit/src/vite/utils.js
```diff
@@ -223,59 +223,84 @@ const find_illegal_rollup_imports = (
...
 	}
 
 	for (const child of node.importedModules) {
 		const chain = child && find_illegal_vite_imports(child, illegal_imports, seen);
-		if (chain) return [{ name: node.id, dynamic: false }, ...chain];
+		if (chain) return [{ name: nodeId, dynamic: false }, ...chain];
```
Rich-Harris, line 301:
```suggestion
		if (chain) return [{ name, dynamic: false }, ...chain];
```

packages/kit/src/vite/utils.js
```diff
@@ -223,59 +223,84 @@ const find_illegal_rollup_imports = (
...
 	}
 
+	seen.delete(nodeId);
```
Rich-Harris, line 304:
```suggestion
	seen.delete(name);
```

Rich-Harris (changes requested):
we need a snake_case lint rule!

Rich-Harris (approved):

Rich-Harris:
actually i rescind my approval until i'm confident that the firefox tests are just flaking

PR #5760, opened 2022-07-29
Postinstall sync

Rich-Harris:
satisfying my curiosity about https://github.com/sveltejs/kit/pull/5651#issuecomment-1200008865


Rich-Harris:
ahhh. i see

Rich-Harris:
Ok, I think this is slightly simpler than #5651 — I prefer the idea of adding as little code as possible to deal with the case of our own CI stuff — but I'll wait for a 👍 before merging

packages/kit/svelte-kit.js
```diff
@@ -1,2 +1,11 @@
 #!/usr/bin/env node
-import './dist/cli.js';
+import fs from 'fs';
+import { fileURLToPath } from 'url';
+
+// in our own CI, and when deploying directly from this monorepo,
+// the `dist` directory will not exist yet
+if (fs.existsSync(fileURLToPath(new URL('./dist', import.meta.url)))) {
+	import('./dist/cli.js');
```
dominikg, line 8: Hmm, no import for the src cli? Could risk running sync from an outdated build

Rich-Harris: No, the only time we use `src/cli.js`, we import that file explicitly. This is just to cover the case where `postinstall` is running _in this repo_ either in GitHub Actions or when we deploy the app to Cloudflare Pages or Vercel (since `postinstall` wasn't running in the test apps _anyway_, I added `svelte-kit sync` to their `check` scripts, and there's no other time we would need to run `postinstall` in the repo itself. I tried using `--ignore-scripts` instead, but then the site failed to deploy because of a dependency on `sharp` which needs `postinstall`)

Rich-Harris: here's where we import `src/cli.js` explicitly: https://github.com/sveltejs/kit/blob/d7386d908ff8b48fd979b8e03470a37f9cd04985/packages/kit/test/cli.js#L5-L6

dominikg: So the only time we have to remember building it first is if we develop sync related hanges locally and want to run `node ../../../../../kit/src/svelte-kit.js sync` (or `pnpm svelte-kit sync`) in a test app to see what changes in the sync output we have to build kit first, otherwise it'll run the last built version (which may be from another branch even).

postinstall not running for workspace dependencies is a bit odd imho, i've asked about it in pnpm discord but didn't hear back.

packages/kit/test/apps/basics/package.json
```diff
@@ -6,8 +6,7 @@
 		"dev": "vite dev",
 		"build": "vite build",
 		"preview": "vite preview",
-		"prepare": "node ../../cli.js sync",
-		"check": "tsc && svelte-check",
+		"check": "svelte-kit sync && tsc && svelte-check",
```
dominikg, line 9:
```suggestion
		"check": "node ../../cli.js sync && tsc && svelte-check",
```
without this is always goes to dist for me.  
And do we want to keep the prepare script too to work around the fact that pnpm doesn't call postinstall for workspace dependencies in a monorepo?

packages/kit/svelte-kit.js
dominikg: I'm fine with this small limitation to keep the code cleaner. The is_workspace and is_self code in the other version was cumbersome.

packages/kit/src/cli.js
```diff
@@ -40,13 +41,25 @@ prog
 	.describe('Synchronise generated files')
 	.option('--mode', 'Specify a mode for loading environment variables', 'development')
 	.action(async ({ mode }) => {
-		if (!fs.existsSync('svelte.config.js')) {
-			console.warn('Missing svelte.config.js — skipping');
+		const event = process.env.npm_lifecycle_event;
+
+		// TODO remove for 1.0
+		if (event === 'prepare') {
```
dominikg, line 47: This is also a lot cleaner than reading the package.json and testing the prepare script ourselves.
However it is going to cause a false positive if users have added their own prepare scripts i think.

E.g. husky the popular git hooks automation uses it in their installations: https://typicode.github.io/husky/#/?id=install

so a test if it actually is our own prepare script is still needed i fear. (adding it here to only read the file if we are in prepare is a better way still)

packages/kit/test/apps/basics/package.json
Rich-Harris: Ah yeah, changed it to `../../cli.js` so we can test out changes to the `sync` script without rebuilding.

IIRC the `prepare` script was running before stuff was built, which meant it didn't work (though that would have been fixed by the `../../cli.js` I guess) but in any case `"prepare": "svelte-kit sync"` is now a no-op, so there's no reason to keep it

dominikg: I was thinking of prepare with the ../../cli.js call too, so that when you pull changes and do `pnpm i` it syncs to avoid red squiggles.

dominikg (approved):

PR #5768, opened 2022-07-31
fix sync scripts

Rich-Harris:
In eef885be16b06852c481c5eac333f48941045961 (#5760) I broke the `"check"` scripts — this fixes them


dominikg:
oops, sorry i didn't recognize it in the og PR. Should we add `pnpm -r check` to the ci?

PR #5773, opened 2022-07-31
Migration script

Rich-Harris:
This adds a new package, `svelte-migrate`, which will automate the automate-able bits of #5748:

```
npx svelte-migrate routes
```

It doesn't attempt to boil the ocean — it just moves files to their new homes and annotates the code with manual TODOs, which will be documented in a GitHub discussion.

(For clarity, we're doing the migration script first so we can dogfood it on our own apps and test suites, but we haven't implemented the changes yet, so don't get ahead of yourselves and run this on your own apps just yet!)

Bunch of stuff that still needs to happen:

* [x] fail gracefully when dealing with code that can't be parsed by TypeScript
* [x] add the documentation
* [x] inject errors into `+page.server.js` and `+server.js` telling people to update handler return values
* [x] inject errors into `+page.js` and `+layout.js` telling people to update `load` return values
* [x] inject error telling authors to replace multiple props with a single `data` prop
* [x] deal with the fact that `+error` can't have load, somehow
* [x] adjust imports in component code
* [x] replace `./__types/whatever` with `./$types`


Rich-Harris:
This is a bit loosey-goosey (I didn't add tests, on the basis that this is somewhat throwaway, but I could be persuaded of their value) but probably far enough along that we can start trying it out on codebases to see how far we get.

Rich-Harris:
Just noticed a bug — after using this to migrate `sites/kit.svelte.dev`, we're left with this:

image

It should be `src/routes/docs.json/+server.js`, not `src/routes/docs/index.json/+server.js`. One for the morning.

packages/migrate/migrations/routes/index.js
```diff
@@ -59,7 +59,7 @@ export async function migrate() {
 		config.kit?.routes ??
 		((filepath) => !/(?:(?:^_|\/_)|(?:^\.|\/\.)(?!well-known))/.test(filepath));
 
-	const files = glob(`${routes}/**`, { filesOnly: true });
+	const files = glob(`${routes}/**`, { filesOnly: true }).map((file) => file.replace(/\\/g, '/'));
```
Rich-Harris, line 62: oh huh — i thought it would always return paths with `/` regardless of platform. i may have confused input with output https://github.com/terkelg/tiny-glob#windows

Rich-Harris:
alright, time to give this a whirl

PR #5777, opened 2022-08-01
[fix] don't warn if custom `compilerOptions.paths` are used

metonym:
This is a QOL change that only emits a warning for `compilerOptions.paths` if it is missing or not set.

**Current DX**

A real world scenario is using SvelteKit for developing Svelte components in which the `tsconfig.json` [has a custom `compilerOptions.paths` set](https://github.com/metonym/svelte-bar-chart-race/blob/8faa606822e82d4e111bf9e551b398348355cd78/tsconfig.json#L5).

In this case, the warning is a false positive.

The following message is emitted when running `yarn build`.

```sh
Your compilerOptions.paths in tsconfig.json should include the following:
{
  "$lib":["src"],
  "$lib/*":["src/*"]
}
```

**Proposed DX**

SvelteKit only warns if `compilerOptions.paths` is missing or is empty. It will not warn if a custom `compilerOptions.paths` is set.

---

### Tests

### Changesets

### DIFF
--- a/.changeset/bright-scissors-drum.md
+++ b/.changeset/bright-scissors-drum.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] don't warn if custom compilerOptions.paths are used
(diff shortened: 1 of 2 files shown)


Rich-Harris:
Thanks, though the current behaviour is intentional — if a custom `compilerOptions.paths` is set but doesn't include the `$lib` aliases, it will result in frustrating errors for anyone using that directory. Is it causing unexpected problems?

metonym:
Makes sense, thanks for the response. It hasn't blocked usage – I suppose the workaround is to just include the expected `$lib` paths in the `tsconfig.json` in addition to any custom paths.

PR #5778, opened 2022-08-01
Implement #5748

Rich-Harris:
Edited: this now incorporates #5782 and #5810.

This PR implements #5748.

* [x] use the app's installed TypeScript rather than including it in SvelteKit (i.e. don't generate types otherwise)
* [x] create types for `parent`
* [x] make `write_types` work incrementally
* [x] update the docs
  * [x] document `POST`, `PATCH`, `PUT` and `DELETE` in `+page.server.js`
  * [x] get search working for `+page` etc
  * [x] 'The following types can be imported from `@sveltejs/kit`' is wrong in the modules section
* [x] get all the tests passing
* [x] ensure all the migration guides are correct
* [x] investigate why client-side navigation fails on the [preview deployment](https://kit-svelte-dev-git-migrate-tests-svelte.vercel.app/)
* [x] rename `GET` to `load` in `+page|layout.server.js` (`POST` etc can stay, pending resolution of https://github.com/sveltejs/kit/discussions/5875)
* [ ] ~~remove `session`, per https://github.com/sveltejs/kit/discussions/5883~~
* [x] implement `invalidate()` (i.e. re-run all `load` functions, if no argument is specified)
* [x] update types inside migrated files (`+page.server.js` should have `PageServerLoad` etc)
* [x] error if project has not been migrated
* [x] mother of all changesets (including for `svelte-migrate` and `create-svelte` etc)


documentation/docs/03-routing.md
```diff
@@ -45,206 +57,170 @@ A file called either `src/routes/about.svelte` or `src/routes/about/index.svelte
...
-	// redirect to the newly created item
-	return {
-		status: 303,
-		headers: {
-			location: `/items/${item.id}`
-		}
-	};
+Unexpected failures are easy to deal with, as it's the same as `GET` — `throw error(status)`.
+
+Validation errors cause the page to be re-rendered with the errors, so they can be used for UI that guides the user towards valid inputs:
```
cdcarson, line 177: In the case of an invalid post request, how does one show the invalid value in the input? E.g., I'd want to do this...

```
<p>Error: Must include the word "potato".</p>
<input name="description" value="tomato" />
```
According to my reading of this section and the [migration guide discussion](https://github.com/sveltejs/kit/discussions/5774#:~:text=POST%2C%20PUT%20and%20PATCH%20handlers%20should%20not%20return%20data.%20They%20can%20return%20an%20errors%20object%20of%20validation%20errors), the POST handler is only allowed to return a map of validation errors, not the invalid data itself. The data is populated from GET, not passed back from what the POST handler originally read. So, it appears the user could be shown the error, but not the invalid value that they submitted.

pnpm-lock.yaml
```diff
@@ -326,6 +325,7 @@ importers:
       svelte: 3.48.0
       svelte-preprocess: 4.10.7_lvfi2wesz6u4l5rfbnetbucfmm
       svelte2tsx: 0.5.11_lvfi2wesz6u4l5rfbnetbucfmm
+      typescript: 4.7.4
```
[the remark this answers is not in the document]
dummdidumm: I did, yes, I just forgot to also update the lock file - this commit moves typescript to devdependencies (we need it during development of kit, but you don't need it when using it).

packages/kit/src/vite/dev/index.js
```diff
@@ -250,9 +290,12 @@ export async function dev(vite, vite_config, svelte_config, illegal_imports) {
...
 
-				const runtime_base = process.env.BUNDLED
-					? `/${posixify(path.relative(cwd, `${svelte_config.kit.outDir}/runtime`))}`
-					: `/@fs${runtime}`;
+				// For some reason using runtime_prefix here is buggy, since Vite will later load the modules
+				// again with a slightly different url (with the drive letter) on windows
+				const runtime_base = `/@fs${
+					// Windows/Linux separation - Windows starts with a drive letter, we need a / in front there
+					runtime_directory.startsWith('/') ? '' : '/'
+				}${runtime_directory}`;
```
dummdidumm, line 298: This was needed for some reason - makes me uneasy and I don't know why it's needed (only?) in this place. Do Vite residents @bluwy and @dominikg know more?

dominikg: there is something similar in v-p-s codebase for the virtual css modules.. i don't remember what exactly in vite made in neccessary. https://github.com/sveltejs/vite-plugin-svelte/blob/ed2a22d0e411b71d39311a5026b3c72762e90e4b/packages/vite-plugin-svelte/src/utils/id.ts#L65

documentation/docs/03-routing.md
Rich-Harris: The plan is basically to take a different tack altogether: https://github.com/sveltejs/kit/discussions/5875

Until then you can have an `errors` prop on your page, and it will be filled in with the validation errors:

```svelte
<script>
  export let data;
  export let errors;
</script>
```

Rich-Harris: oh shit, wait, that doesn't answer your question at all. err...

this is nasty, but you could put the values on `event.locals` and read them in the `GET`

packages/kit/src/vite/dev/index.js
Rich-Harris: i'm gonna go with 'if it works it works' 😆

documentation/docs/05-loading.md
```diff
@@ -1,157 +0,0 @@
...
-
-> Code called inside `load` blocks:
->
-> - should use the SvelteKit-provided [`fetch`](/docs/loading#input-fetch) wrapper rather than using the native `fetch`
-> - should not reference `window`, `document`, or any browser-specific objects
-> - should not directly reference any API keys or secrets, which will be exposed to the client, but instead call an endpoint that uses any required secrets
-
-It is recommended that you not store per-request state in global variables, but instead use them only for cross-cutting concerns such as caching and holding database connections.
-
-> Mutating any shared state on the server will affect all clients, not just the current one.
```
dummdidumm, line 49: This is an important gotcha that some people are not aware of (as seen in related discussions). We should have this somewhere in the docs.

documentation/docs/05-load.md
```diff
@@ -0,0 +1,338 @@
...
+}
+```
+
+If an _unexpected_ error is thrown, SvelteKit will invoke [`handleError`](/docs/hooks#handleerror) and treat it as a 500 Internal Server Error.
+
+> In development, stack traces for unexpected errors are visible as `$page.error.stack`. In production, stack traces are hidden.
+
+### Redirects
+
+Redirecting users works similarly:
```
dummdidumm, line 314: This is more of a personal preference, but I prefer to have the sections more self-contained. If I jump right into this section I'm lost what this "similarly" refers to.

```suggestion
To redirect users, use the `redirect` helper from `@sveltejs/kit` to specify the location to which they should be redirected alongside a `3xx` status code.
```

documentation/docs/05-loading.md
gyurielf: > This is an important gotcha that some people are not aware of (as seen in related discussions). We should have this somewhere in the docs.

Agree with you. E.g:  I didn't know that, when I switched from SPA to SSR.

documentation/docs/05-load.md
martinbeentjes: With the beautiful Cmd+K search UI from the docs, I think it is important to keep every section self contained. If I look for 'redirects' I would like to only need to read that section, and not read back to what _similarly_ means. Your suggestion clearly says what I should do when I want to redirect an user. 

TL;DR: great suggestion 👍🏻

documentation/docs/05-load.md
```diff
@@ -180,96 +227,79 @@ export type PageLoad = import('@sveltejs/kit').Load<{}>;
...
+	}
 }
 ```
 
-The `redirect` string should be a [properly encoded](https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding) URI. Both absolute and relative URIs are acceptable. You can learn more about redirects and redirect codes [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections).
+### Invalidation
+
+SvelteKit tracks the dependencies of each `load` function to avoid re-running it unnecessarily during navigation. For example, a `load` function in a root `+layout.js` doesn't need to re-run when you navigate from one page to another unless it references `url` or a member of `params` that changed since the last navigation.
+
+Using [`invalidate(url)`](/docs/modules#$app-navigation-invalidate), you can re-run any `load` functions that depend on the invalidated resource (either implicitly, via [`fetch`](#fetch)), or explicitly via [`depends`](#depends). You can also invalidate _all_ `load` functions by calling `invalidate()` without an argument.
```
dummdidumm, line 338: One reading this again: Do you mean to rerun the `load` functions even if they didn't fetch anything/didn't declare any `depends`?

documentation/docs/03-routing.md
```diff
@@ -2,420 +2,295 @@
...
-...resulting in this ordering, meaning that `/foo-abc` will invoke `src/routes/foo-abc.svelte`, and `/foo-def` will invoke `src/routes/foo-[c].svelte` rather than less specific routes:
+Throughout the examples above, we've been importing types from a `$types.d.ts` file. This is a file SvelteKit creates for you in a hidden directory if you're using TypeScript (or JavaScript with JSDoc type annotations) to give you type safety when working with your root files.
 
-```bash
-src/routes/foo-abc.svelte
-src/routes/foo-[c].svelte
-src/routes/[a].js
-src/routes/[b].svelte
-src/routes/[...catchall].svelte
+For example, annotating `export let data` with `PageLoad` or `LayoutLoad` tells TypeScript that the type of `data` is whatever was returned from `load`:
```
msonnberger, line 272:
```suggestion
For example, annotating `export let data` with `PageData` tells TypeScript that the type of `data` is whatever was returned from `load`:
```

documentation/docs/05-loading.md
isaac-mcfadyen: I agree with keeping this: it comes up a lot of times with Serverless platforms like Cloudflare Workers where other requests are handled during async tasks, so storing request-specific things in global state can be disastrous.

documentation/docs/05-load.md
```diff
@@ -0,0 +1,338 @@
...
+export function load(event) {
+	return {
+		some: 'data'
+	};
+}
+```
+
+### Input properties
+
+The argument to a `load` function is a `LoadEvent` or (for server-only `load` functions) a `ServerLoadEvent`, which inherits `clientAddress`, `locals`, `platform` and `request` from `RequestEvent`. All events have the following properties:
```
isaac-mcfadyen, line 26:
```suggestion
The argument to a `load` function is a `LoadEvent` or (for server-only `load` functions a `ServerLoadEvent`), which inherits `clientAddress`, `locals`, `platform` and `request` from `RequestEvent`. All events have the following properties:
```

documentation/docs/03-routing.md
Rich-Harris:
```suggestion
For example, annotating `export let data` with `PageData` (or `LayoutData`, for a `+layout.svelte` file) tells TypeScript that the type of `data` is whatever was returned from `load`:
```

documentation/docs/05-load.md
Rich-Harris: not a typo! The way to tell if parens are in the right place is to remove the clause entirely and see if it makes sense:

> The argument to a `load` function is a `LoadEvent` or a `ServerLoadEvent`, which inherits `clientAddress`, `locals`, `platform` and `request` from `RequestEvent`.

vs

> The argument to a `load` function is a `LoadEvent` or, which inherits `clientAddress`, `locals`, `platform` and `request` from `RequestEvent`.

An alternative would be this, which is technically redundant but might make it clearer that `clientAddress` etc only apply to `ServerLoadEvent`:

```suggestion
The argument to a `load` function is a `LoadEvent` (or, for server-only `load` functions, a `ServerLoadEvent` which inherits `clientAddress`, `locals`, `platform` and `request` from `RequestEvent`). All events have the following properties:
```

documentation/docs/05-loading.md
Rich-Harris: yeah, this was a casualty of restructuring this section of the docs to accommodate new API — should be reinstated somewhere

Rich-Harris: added a note in c1e20860090554dcf947206be0b5bc1735932d99, will mark this resolved

documentation/docs/05-load.md
```diff
@@ -0,0 +1,334 @@
...
+
+### Invalidation
+
+SvelteKit tracks the dependencies of each `load` function to avoid re-running it unnecessarily during navigation. For example, a `load` function in a root `+layout.js` doesn't need to re-run when you navigate from one page to another unless it references `url` or a member of `params` that changed since the last navigation.
+
+Using [`invalidate(url)`](/docs/modules#$app-navigation-invalidate), you can re-run any `load` functions that depend on the invalidated resource (either implicitly, via [`fetch`](#fetch)), or explicitly via [`depends`](#depends). You can also invalidate _all_ `load` functions by calling `invalidate()` without an argument.
+
+### Shared state
+
+In many server environments, a single instance of your app will serve multiple users. For that reason, per-request state must not be stored in shared variables outside your `load` functions, but should instead be stored in `event.locals`.
```
dummdidumm, line 334: Since this not only relates to requests, but also the general user state, I suggest to enhance the section a little:

```suggestion
In many server environments, a single instance of your app will serve multiple users. For that reason, per-request state must not be stored in shared variables outside your `load` functions, but should instead be stored in `event.locals`. Similarly, per-user state must not be stored in global variables, instead make use of `$page.data` (which contains the combined data of all `load` functions) or use Svelte's [context feature](https://svelte.dev/docs#run-time-svelte-setcontext) to create scoped state.
```

PR #5782, opened 2022-08-01
it begins

Rich-Harris:
This PR will track the implementation of #5748. It _may_ become multiple PRs, we'll see. 

So far, it represents a tiny but important milestone — the first unit test of `create_manifest_data` passing with the new filesystem structure.

### Tests

### Changesets


packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -81,120 +39,160 @@ export default function create_manifest_data({
...
 
-			const id = file
-				.slice(0, -extension.length)
-				.replace(/(?:^|\/)index((?:@[a-zA-Z0-9_-]+)?(?:\.[a-z]+)?)?$/, '$1');
-			const project_relative = `${routes_base}/${file}`;
+			const project_relative = `${routes_base}/${filepath}`;
+			const segments = filepath.split('/');
+			const file = /** @type {string} */ (segments.pop());
+
+			if (file[0] !== '+') return; // not a route file
```
arxpoetica, line 66: There's gold in them hills.

johnnysprinkles: Not sure if this was discussed in that locked thread, but any way we could make those pluses optional for the people that really hate them? I'm not putting random junk in my route folders, so this would be a heck of a lot nicer:

    /routes
       /routes/foo
         /routes/foo/page.svelte
         /routes/foo/layout.svelte

csjh: > Not sure if this was discussed in that locked thread, but any way we could make those pluses optional for the people that really hate them? I'm not putting random junk in my route folders, so this would be a heck of a lot nicer:
> 
> ```
> /routes
>    /routes/foo
>      /routes/foo/page.svelte
>      /routes/foo/layout.svelte
> ```

Feels like that would defeat the purpose a bit :p

johnnysprinkles: Was discussed I see now, deleting comment

johnnysprinkles: I am struggling to see why any prefix is needed though. Wouldn't `analyze`ing every file in the directory and skipping irrelevant ones be pretty much the same as skipping all non-plus-prefixed ones?

Rich-Harris: This has been discussed ad nauseam on #5748. See in particular the 'Prefixes' section of https://github.com/sveltejs/kit/discussions/5748#discussioncomment-3282732

packages/kit/src/core/sync/write_types.js
```diff
@@ -42,24 +48,36 @@ export function write_types(config, manifest_data) {
...
-					write(`${outdir}/proxy${path.basename(route.page.module)}`, proxy);
-					imports.push(`import { load } from './proxy+page.js';`);
+				if (proxy && proxy.exports.includes('load')) {
+					const basename = path.basename(route.page.module);
+					write(`${outdir}/proxy${basename}`, proxy.code);
+					imports.push(`import { load } from './proxy${basename}';`);
 					exports.push(`export type Data = Awaited<ReturnType<typeof load>>;`);
 				} else {
-					// TODO bail out write file with `any` types
+					exports.push(`export type Data = any;`);
```
dummdidumm, line 57: should this be unknown or never instead?

packages/kit/src/core/sync/write_types.js
```diff
@@ -42,24 +48,36 @@ export function write_types(config, manifest_data) {
...
+				if (proxy && proxy.exports.includes('GET')) { // TODO handle validation errors from POST/PUT/PATCH
+					const basename = path.basename(route.page.server);
+					write(`${outdir}/proxy${basename}`, proxy.code);
+					imports.push(`import { GET } from './proxy${basename}';`);
+					declarations.push(`type ServerData = Awaited<ReturnType<typeof GET>>;`);
+				} else {
+					declarations.push(`type ServerData = any;`);
+				}
+			} else {
+				declarations.push(`type ServerData = null;`);
```
dummdidumm, line 74: `never`?

packages/kit/src/core/sync/write_types.js
```diff
@@ -42,24 +48,36 @@ export function write_types(config, manifest_data) {
...
+				const content = fs.readFileSync(route.page.server, 'utf8');
+				const proxy = tweak_types(content, server_names);
+
+				if (proxy && proxy.exports.includes('GET')) { // TODO handle validation errors from POST/PUT/PATCH
+					const basename = path.basename(route.page.server);
+					write(`${outdir}/proxy${basename}`, proxy.code);
+					imports.push(`import { GET } from './proxy${basename}';`);
+					declarations.push(`type ServerData = Awaited<ReturnType<typeof GET>>;`);
+				} else {
+					declarations.push(`type ServerData = any;`);
```
dummdidumm, line 71: should this be unknown or never instead?

packages/kit/src/core/sync/write_types.js
```diff
@@ -87,6 +105,35 @@ function tweak_types(content, names) {
...
+					const exported = element.name;
+					if (names.has(element.name.text)) {
+						const local = element.propertyName || element.name;
+						exports.set(exported.text, local.text);
+					}
+				});
+			}
+
+			// 93 because there's no ts.isExportKeyword function for some reason
+			if (node.modifiers?.some(modifier => modifier.kind === 93)) {
```
dummdidumm, line 126: You can use `ts.SyntaxKind.ExportKeyword` instead of `93`

dummdidumm (commented):
Left some small comment on the type writing code. It's a very clever solution, my biggest concern is that it contains stale data as long as people didn't safe the file - I don't see a way around this though. One way to make this a non-problem: If the function isn't modified (doesnt have the type annotations we replace) then don't create a proxy.
Also: Does this mean that now everybody has to have TypeScript installed in their code base? Those who vote "I don't want any types" don't have TS installed, so this should be optional. If the `sync` code is dynamically imported by the CLI this should work already.

packages/kit/src/core/sync/write_types.js
Rich-Harris: my thinking was that in the case where there's no `+page.server.js`, or it contains no `GET`, `data` would be `null` inside `load`:

```js
// +page.js
export function load({ data }) {
  if (data) {
    console.log('got data from server', data);
  } else {
    console.log('no data coming from the server, we\'re on our own');
  }

  // ...
}
```

I assumed `type ServerData = null` was the way to do that, but is `never` more appropriate in this case?

packages/kit/src/core/sync/write_types.js
Rich-Harris: yeah, should probably be `unknown`, but only if the process failed. Going to make it `null` if the process succeeded but there's no `load`

packages/kit/src/core/sync/write_types.js
Rich-Harris: changed to `unknown`

packages/kit/src/runtime/client/client.js
```diff
@@ -461,80 +451,52 @@ export function create_client({ target, session, base, trailing_slash }) {
...
+		const uses = {
+			params: new Set(),
+			url: false,
+			session: false,
+			dependencies: new Set()
 		};
 
 		/** @param dep {string} */
-		function add_dependency(dep) {
+		function depends(dep) {
```
dummdidumm, line 477: In the proposal it says that depends takes one or more string, so you can use it like `depends(a, b, ..)`, so I think this needs to be updated accordingly.

packages/kit/src/runtime/server/page/index.js
```diff
@@ -18,25 +30,317 @@ export async function render_page(event, route, options, state, resolve_opts) {
...
+
+		/** @type {Array<Promise<Record<string, any> | null>>} */
+		const load_promises = nodes.map((node, i) => {
+			if (error) throw error;
+			return Promise.resolve().then(async () => {
+				try {
+					const server_data = await server_promises[i];
+
+					if (node?.module?.load) {
+						const data = await node?.module?.load?.call(null, {
```
dummdidumm, line 136: I think this can use the newly added `load_data` function (that function needs to be adjusted then though)

packages/kit/src/core/sync/create_manifest_data/index.js
coryvirok: For the folks following along like myself...

The best rationale I've heard from anyone about the need for prefixes is that the maintainers want a future proof way to identify files of special significance. Without a prefix, they risk breaking ppl's codebases in the future if they introduce a new special/reserved file name.

Rich-Harris: That's one reason, yes — being able to add new features without forcing an otherwise unwarranted semver major is an important design consideration (the "integers are free!!!!" crowd disagrees with that statement, because they don't understand human psychology).

But there are also very good DX reasons:

* you see a `+` file, you immediately know that it's a route file (and if you're unfamiliar with that concept, searching for `+page` etc in the docs will immediately clue you in, without you needing to know any jargon)
* your files are lexically sorted in a very predictable way — child routes, route files, then colocated files. For example if you have a bunch of colocated files, `index.svelte` gets buried...

  ```
  src/routes
  ├ Foo.svelte
  ├ constants.js
  ├ index.svelte
  └ utils.js
  ```

  ...whereas `+page.svelte` sits apart:

  ```
  src/routes
  ├ +page.svelte
  ├ Foo.svelte
  ├ constants.js
  └ utils.js
  ```

* you can find all your route files in your editor very quickly — if you're using VSCode, `Cmd-P` then just type `+`, or start typing the parts of a route followed by `+`. If you start typing `pa` or `in` then you'll get a bunch of false positives, forcing you to type a lot more in order to filter the list. On a large codebase, details like that start to matter.

packages/kit/src/runtime/server/page/index.js
Rich-Harris: `load_data` is a vestige of the existing [`load_node`](https://github.com/sveltejs/kit/blob/bde443dad85868defb9bdfdc8f0a2347a2451582/packages/kit/src/runtime/server/page/load_node.js#L27), and should probably be got rid of as it doesn't know how to inherit data from `parent`

packages/kit/src/runtime/client/client.js
Rich-Harris: good catch, fixed

packages/kit/src/runtime/server/page/index.js
Rich-Harris: (though that does involve some slightly un-DRY code i suppose, for the `respond_with_error` function)

Rich-Harris:
Going to merge this into #5778, as the separate branches have served their purpose

PR #5792, opened 2022-08-02
Added cli for adding new route 'svelte-kit add route --name svelte' cli in svelte-kit

theetherGit:
I have added `svelte-kit  add route --name svelte` for future changes as @Rich-Harris mentioned in [discussion 5748](https://github.com/sveltejs/kit/discussions/5748). We going to need a cli to create routes so we don't have to add three files again and again. This supports typescript file generation and dynamic routes too.

###Example:
- Default command: svelte-kit  add route -n routeName (This is will generate `.js` files in route. -n or --name are required for route name)
- With typescript: svelte-kit  add route -t -n 'blog/[slug]' (This will generate `.ts` files in route. If you need route as dynamic you need to pass name in string).
---
### PR Requirements:
- [x] It's not a discussed issue or feature when i saw [discussion 5748](https://github.com/sveltejs/kit/discussions/5748), I wanted to have a cli for creating routes without any issues.
- [x] This might not solve a bigger problem but it's going to ease our issue related to adding three files over and over in every route so why not use cli for that.

### Tests

### Changesets
- Added a new cli command in svelte-kit cli.
- This can be released after [discussion 5748](https://github.com/sveltejs/kit/discussions/5748) is completed.

### DIFF
--- a/.changeset/cuddly-ads-kick.md
+++ b/.changeset/cuddly-ads-kick.md
@@ -1,0 +1,5 @@
+---
+'svelte-migrate': patch
+---
+
+handle more import cases
--- a/.changeset/famous-birds-boil.md
+++ b/.changeset/famous-birds-boil.md
@@ -1,0 +1,5 @@
+---
+'svelte-migrate': patch
+---
+
+Correctly rename files with spaces when migrating
--- a/.changeset/hot-beds-tickle.md
+++ b/.changeset/hot-beds-tickle.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] Import analysis doesn't get stuck in an infinite loop when encountering cyclical imports
--- a/.changeset/loud-parrots-sell.md
+++ b/.changeset/loud-parrots-sell.md
@@ -1,0 +1,5 @@
+---
+'svelte-migrate': patch
+---
+
+Add a README'
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -577,6 +577,7 @@
     "nice-ways-search",
     "nine-actors-whisper",
     "nine-beans-rhyme",
+    "nine-bees-agree",
     "nine-dots-occur",
     "nine-kangaroos-laugh",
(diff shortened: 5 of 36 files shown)


chvanam:
Would it be useful if the created route files contained some boilerplate functionnalities ? 
For example a GET handler in `+page.server.js` showing how to get prop data in `+page.svelte` with some comment about other handlers or an example load function in `+page.js` 🤔

The future documentation itself will undoubtly make page creations straight forward, just opening the discussion here.

theetherGit:
@chvanam it is possible. I wanted this to be approved first so in future i can add more things to cli and boilerplate code too.

j4w8n:
For what it's worth, I deployed an npm package that does this.

Has some boilerplate code for the files @chvanam
Supports typescript
Has a config file, so you can set defaults like creating .ts files instead of .js, whether you want the boilerplate code (I called it 'codekit'), and if the default route is a page or server.

You can't create named layouts with it yet, but there's an option to say that a page route should use a named layout.

https://www.npmjs.com/package/mkrt

theetherGit:
@j4w8n wow man that's a perfect module. I was already working on those additions and i got more ideas from your work.

Rich-Harris:
Hey — thank you for the PR, but this is really something that needs an issue first. It's probably not something we're going to add, but if it's a very popular feature request then we would need to figure out the answers to a ton of questions (what about layouts? what about `+server.js` files? do we automatically add all three page files even though most routes will only need two or one? what methods do we create handlers for? is typescript auto-detected? does it work relative to `config.kit.files.routes` or the current working directory? etc etc etc) before attempting an implementation. 

Realistically, answering all those questions would leave us with a process that wasn't really any easier than just creating the files manually.

theetherGit:
Thank You for your reply, I already had some of the questions before starting it, like we always not going to need all 3 files and relative to `config.kit.files.routes` or the current working directory. I was thinking about them and have some notes but that is for some other time. And for other queries i already have layout or code on my local. Again thank you so much.

PR #5804, opened 2022-08-03
[chore] upgrade to Undici 5.8.1

benmccann:
Had to update a test because Undici got a bit stricter about the spec in https://github.com/nodejs/undici/pull/1550

This test file will be moved in https://github.com/sveltejs/kit/pull/5778, but other than that should not cause any merge conflicts

### DIFF
--- a/.changeset/violet-jars-visit.md
+++ b/.changeset/violet-jars-visit.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[chore] upgrade to Undici 5.8.1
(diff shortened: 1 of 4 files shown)


.changeset/violet-jars-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[chore] upgrade to Undici 5.7.0
```
gyurielf, line 5: In the title: [chore] upgrade to Undici 5.8.1, but it's just 5.7.0 :)

benmccann: i'm aware. i'm trying to figure out why the tests are failing

Rich-Harris (approved):

PR #5810, opened 2022-08-04
Docs for the new API

dummdidumm:
Companion to #5782 

### Tests

### Changesets

### DIFF
--- a/documentation/docs/03-routing.md
+++ b/documentation/docs/03-routing.md
@@ -6,20 +6,32 @@ At the heart of SvelteKit is a _filesystem-based router_. This means that the st
 
 > You can change this to a different directory by editing the [project config](/docs/configuration).
 
-There are two types of route — **pages** and **endpoints**.
+Inside the routes folder, files with a `+` prefix have special meaning to SvelteKit.
 
-Pages typically generate HTML to display to the user (as well as any CSS and JavaScript needed for the page). By default, pages are rendered on both the client and server, though this behaviour is configurable.
+> The `+` prefix was chosen for several reasons:
+>
+> - It's distinctive and easy to spot — as a newcomer it immediately cues into that _something_ is going on, after that, you immediately _know_ what's going on.
+> - It nicely groups all special files. You are encouraged to colocate other files related only to that part of your app in the same folder without having to think about any prefixes for them.
+> - It's easy to search for, both in the docs and your editor's file search.
 
-Endpoints run only on the server (or when you build your site, if [prerendering](/docs/page-options#prerender)). This means it's the place to do things like access databases or APIs that require private credentials or return data that lives on a machine in your production network. Pages can request data from endpoints. Endpoints return JSON by default, though may also return data in other formats.
+The first type of files are those related to routes, or pages. A route is defined by a directory in your routes folder with a `+page.svelte` file. By default, pages are rendered on both the client and server, though this behaviour is configurable. A corresponding `+page.server.js` file is able to provide data to the page and receive data from it in the case of form POSTs. It only runs on the server. Additionally, a `+page.js` file is able to provide data to the page - if a `+page.server.js` is present, `+page.js` runs afterwards, receives the data from `+page.server.js` and can transform it - and can contain meta information relevant to the router.
+
+Related to pages are [layouts](/docs/layouts), named `+layout.svelte`. They contain common UI (like a header) or logic (like authentication) and wrap every page below it. Similar to pages a `+layout.server.js` and/or `+layout.js` file can be added. They function the same as their corresponding page files, except they can't receive data from layouts - they are read-only.
+
+In case of unexpected exceptions, you can fall back to a more gracious page using [`+error.svelte`](/docs/layouts#error-pages). If something goes wrong, the next `+error.svelte` file up the tree is used to display your fallback UI.
+
+Lastly, if you want to create endpoints, you create a `+server.js` file.
+
+Each of these special files is discussed in more detail in their respective sections.
 
 ### Pages
 
-Pages are Svelte components written in `.svelte` files (or any file with an extension listed in [`config.extensions`](/docs/configuration)). By default, when a user first visits the application, they will be served a server-rendered version of the page in question, plus some JavaScript that 'hydrates' the page and initialises a client-side router. From that point forward, navigating to other pages is handled entirely on the client for a fast, app-like feel where the common portions in the layout do not need to be rerendered.
+Pages are Svelte components written in `+page.svelte` files (or any `+page` file with an extension listed in [`config.extensions`](/docs/configuration)). By default, when a user first visits the application, they will be served a server-rendered version of the page in question, plus some JavaScript that 'hydrates' the page and initialises a client-side router. From that point forward, navigating to other pages is handled entirely on the client for a fast, app-like feel where the common portions in the layout do not need to be rerendered.
 
-The filename determines the route. For example, `src/routes/index.svelte` is the root of your site:
+The foldername determines the route. For example, `src/routes/+page.svelte` is the root of your site:
 
 ```html
-/// file: src/routes/index.svelte
+/// file: src/routes/+page.svelte
 <svelte:head>
 	<title>Welcome</title>
 </svelte:head>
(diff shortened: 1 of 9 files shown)


documentation/docs/03-routing.md
```diff
@@ -6,20 +6,32 @@ At the heart of SvelteKit is a _filesystem-based router_. This means that the st
 
 > You can change this to a different directory by editing the [project config](/docs/configuration).
 
-There are two types of route — **pages** and **endpoints**.
+Inside the routes folder, files with a `+` prefix have special meaning to SvelteKit.
 
-Pages typically generate HTML to display to the user (as well as any CSS and JavaScript needed for the page). By default, pages are rendered on both the client and server, though this behaviour is configurable.
+> The `+` prefix was chosen for several reasons:
+>
```
Mlocik97, line 12: is this supposed to be here?

documentation/docs/03-routing.md
```diff
@@ -45,206 +57,170 @@ A file called either `src/routes/about.svelte` or `src/routes/about/index.svelte
...
+	}
+
+	return {
+		answer: 42
+	};
+}
 ```
 
-These functions can, like `GET`, return a `body` that will be passed to the page as props. Whereas 4xx/5xx responses from `GET` will result in an error page rendering, similar responses to non-GET requests do not, allowing you to do things like render form validation errors:
+The first argument to `error` is an [HTTP status code](https://httpstatusdogs.com) (you'll most likely reach for `4xx` — client error, or `5xx` — server error), the second is an explanatory message. You can access both using the [page store](/docs/modules#$app-stores-page).
```
Mlocik97, line 150: Can we please not? I think link to <https://developer.mozilla.org/en-US/docs/Web/HTTP/Status> is better. I agree it's good to have some jokes and funny stuff in docs, but not for the sake of quality. It's needed to people know about what code to use.

Sry I'm funsucker.

documentation/docs/03-routing.md
```diff
@@ -315,6 +287,66 @@ export default config;
...
+| src/routes/data/+server.js            | /data      |
+
+Because the foldername describes the endpoint URL, you can't have a `+server.js` file next to a `+page.svelte` file — when hitting the URL, SvelteKit wouldn't know if you want to render the page or return data from the endpoint.
+
+#### GET, POST, PUT, PATCH, DELETE
+
+Endpoints can handle any HTTP method — not just `GET` — by exporting the corresponding function:
+
+```js
+// @noErrors
```
Mlocik97, line 308: this is weird in example.

documentation/docs/04-layouts.md
```diff
@@ -76,132 +76,120 @@ Some parts of your app might need something other than the default layout. For t
...
-├ __layout.svelte
--└ __layout-root.svelte
-+└ __layout-root@default.svelte
+│ │ ├ +layout@root.svelte
+│ │ ├ one/+page.svelte
+│ │ ├ two/+page.svelte
+│ │ └ three/+page.svelte
+│ └ +layout.svelte
+├ +layout.svelte
+-└ +layout-root.svelte
```
Mlocik97, line 130: will this render correctly? Without `-` symbol moving intendation of `└`? In VSCode preview it shows wrong.

documentation/docs/05-loading.md
```diff
@@ -93,65 +92,126 @@ If the page you're loading has an endpoint, the data returned from it is accessi
...
+		items: await getCartItems(user.cart);
+	};
+}
+```
 
-#### error
+#### depends
 
-`error` is the error that was thrown (or returned from a previous `load`) when rendering an error page, or `null` otherwise.
+`depends` is a helper method which allows to declare a dependency on a specific resource. The input is an array of strings representing URLs the page depends on, which can subsequently be used with [`invalidate`](/docs/modules#$app-navigation-invalidate) to cause `load` to rerun. You only need to invoke `depends` with them if you're using a custom API client; URLs loaded with the provided `fetch` function are added automatically.
```
Mlocik97, line 113: Text say "The input is an array of strings..." but in example it's not array of strings but just string. Confusing.

documentation/docs/05-loading.md
```diff
@@ -93,65 +92,126 @@ If the page you're loading has an endpoint, the data returned from it is accessi
...
+URLs can be absolute or relative to the page being loaded, and must be [encoded](https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding).
 
-If you return a Promise from `load`, SvelteKit will delay rendering until the promise resolves. The return value has several properties listed below, all of which are optional (as is the return value itself).
+```js
+/// file: src/routes/blog/+page.js
+/** @type {import('./$types').Load} */
+export async function load({ depends }) {
+	const url = `https://cms.example.com/articles.json`;
+	const response = await myOwnFetch(url);
+	depends(url);
```
Mlocik97, line 123: `depends([url])` ?

documentation/docs/03-routing.md
dummdidumm: This is from the old docs, I didn't realize it links to this site 😄 since it's there and noone else has complained I'm inclined to keep it for now

documentation/docs/03-routing.md
dummdidumm: This instructs wiki two slash, it's not visible in the output. You'll find some odd stuff like that in other places as well

documentation/docs/05-loading.md
dummdidumm: Good catch

Mlocik97:
Also, don't forget about FAQ section.

documentation/docs/03-routing.md
Mlocik97: oh, ok, didn't know.

documentation/docs/80-migrating.md
```diff
@@ -74,7 +74,14 @@ A common pattern in Sapper apps is to put your internal library in a directory i
...
+Routes now are made up of the folder name exclusively to remove ambiguity, the folder names leading up to a `+page.svelte` correspond to the route. See [the routing docs](/docs/routing) for an overview. The following shows a old/new comparison:
+
+| Old                       | New                       |
+| ------------------------- | ------------------------- |
+| routes/about/index.svelte | routes/about/+page.svelte |
+| routes/about.svelte       | routes/about/+page.svelte |
+
+Your custom error page component should be renamed from `_error.svelte` to `+error.svelte`. Any `_layout.svelte` files should likewise be renamed `+layout.svelte`. The double underscore prefix is reserved for SvelteKit; your own [private modules](/docs/routing#private-modules) are still denoted with a single `_` prefix (configurable via [`routes`](/docs/configuration#routes) config).
```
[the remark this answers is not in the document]
dummdidumm: This is the Sapper -> SvelteKit migration docs, in Sapper it was only a single underscore.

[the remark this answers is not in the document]
chvanam: Fair enough!

documentation/docs/03-routing.md
```diff
@@ -315,6 +287,66 @@ export default config;
...
 
+### Standalone endpoints
+
+Most commonly, endpoints exist to provide data to the page with which they're paired. They can, however, exist separately from pages. Standalone endpoints have more flexibility — they return [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) objects. This means you have more possibilities for your response `body` type besides JSON, for example [`Uint8Array`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array) or a [`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
+
+Standalone endpoints reside in a `+server.js` (or `.ts`) file, the folder describes the URL where the endpoint is available - just like for pages:
+
+| filename                              | endpoint   |
+| ------------------------------------- | ---------- |
+| src/routes/data/index.json/+server.js | /data.json |
```
rmunn, line 298: Is that route path correct? I think it would become `src/routes/data.json/+server.js` in the new world, not `data/index.json/+server.js`.

documentation/docs/04-layouts.md
Rich-Harris: we'll find out when this is merged into #5778 and the docs are able to build

documentation/docs/03-routing.md
Rich-Harris:
```suggestion
| src/routes/data.json/+server.js | /data.json |
```

PR #5825, opened 2022-08-05
feat: Increase security by virtualizing `$env/static/*`

tcc-sejohnson:
Previously, the contents of `$env/static/*` were written to the disk in `.svelte-kit/runtime/env/static/*.js`. We'd like to prevent dumping the environment to the disk, so this is a test implementation of using virtual modules.

A couple of questions for those wiser than I:

1. How do we make this work with `build`? It seems to work in `dev`, but throws a big red error when running `build`.
2. Are there any cases where this might come back to bite us?

Big thanks to @dominikg for the first "test" implementation of this.

---

# HEADS UP!

We're about to embark on a [significant redesign](https://github.com/sveltejs/kit/discussions/5748) that will touch many parts of the codebase. Until that work is finished, PRs are very likely to result in merge conflicts, and will likely be ignored until the redesign work is complete (by which time they will likely be stale). Please consider delaying your PR until then.

Thank you for understanding!

---

### Tests

### Changesets


tcc-sejohnson:
Note for those smarter than I am: I tried pulling the env logic out into its own plugin just for fun. (Code below.) It didn't work, unfortunately, with the same behavior: Fine in `dev`, broken on `build`.

```ts
/**
 * @return {import('vite').Plugin[]}
 */
export function sveltekit() {
	return [env(), ...svelte(), kit()];
}

/**
 * Returns the SvelteKit environment plugin. Replaces `$env/static/*` imports
 * with the actual environment variables.
 *
 */
function env() {
	/** @type {import('types').ValidatedConfig} */
	let svelte_config;

	/** @type {import('vite').ConfigEnv} */
	let vite_config_env;

	return {
		name: 'vite-plugin-svelte-kit',

		async config(_, config_env) {
			vite_config_env = config_env;
			svelte_config = await load_config();
		},

		async resolveId(id) {
			switch (id) {
				case '$env/static/private':
					return '\0$env/static/private';
				case '$env/static/public':
					return '\0$env/static/public';
			}
		},

		async load(id) {
			switch (id) {
				case '\0$env/static/private':
					return create_env_module(
						'$env/static/private',
						get_env(vite_config_env.mode, svelte_config.kit.env.publicPrefix).private
					);
				case '\0$env/static/public':
					return create_env_module(
						'$env/static/private',
						get_env(vite_config_env.mode, svelte_config.kit.env.publicPrefix).public
					);
			}
		}
	};
}
```

tcc-sejohnson:
Extracted the env module replacement out into its own Vite plugin, adding it to the array returned by `sveltekit`. Tadaa!

packages/kit/src/vite/index.js
```diff
@@ -61,7 +62,54 @@ const enforced_config = {
...
  */
 export function sveltekit() {
-	return [...svelte(), kit()];
+	return [...svelte(), env(), kit()];
+}
+
+/**
+ * Returns the SvelteKit environment plugin. Replaces `$env/static/*` imports
+ * with the actual environment variables.
+ *
```
benmccann, line 71:
```suggestion
```

packages/kit/src/vite/index.js
```diff
@@ -61,7 +62,54 @@ const enforced_config = {
...
+ */
+function env() {
+	/** @type {import('types').ValidatedConfig} */
+	let svelte_config;
+
+	/** @type {import('vite').ConfigEnv} */
+	let vite_config_env;
+
+	return {
+		name: 'vite-plugin-env',
```
benmccann, line 82: I think `vite-plugin-svelte` uses a convention of `:subplugin`, which seems nice.
```suggestion
		name: 'vite-plugin-svelte-kit:env',
```

packages/kit/src/vite/index.js
```diff
@@ -61,7 +62,53 @@ const enforced_config = {
...
+		async load(id) {
+			switch (id) {
+				case '\0$env/static/private':
+					return create_env_module(
+						'$env/static/private',
+						get_env(vite_config_env.mode, svelte_config.kit.env.publicPrefix).private
+					);
+				case '\0$env/static/public':
+					return create_env_module(
+						'$env/static/private',
```
PatrickG, line 106:
```suggestion
						'$env/static/public',
```

I guess this should be `$env/static/public`.

tcc-sejohnson: Thanks! You da bomb

tcc-sejohnson:
@Rich-Harris 

I think this should be good to go. The problem was in `utils.js`. Frankly, I have no idea what's going on with the previous code:

```ts
	if (!process.env.BUNDLED) {
		alias.push(
			{
				find: '$env/static/public',
				replacement: path.posix.join(config.outDir, 'runtime/env/static/public.js')
			},
			{
				find: '$env/static/private',
				replacement: path.posix.join(config.outDir, 'runtime/env/static/private.js')
			}
		);
	}

	alias.push({
		find: '$env',
		replacement: `${get_runtime_directory(config)}/env`
	});

	return alias;
```

But the following makes sense, I suppose, now that we're not aliasing `$env` itself?

```ts
	const base = `${get_runtime_directory(config)}/env`;

	alias.push(
		{
			find: '$env/dynamic/public',
			replacement: `${base}/dynamic/public.js`
		},
		{
			find: '$env/dynamic/private',
			replacement: `${base}/dynamic/private.js`
		}
	);

	return alias;
```

The root of the problem is that I just don't know what the heck `process.env.BUNDLED` denotes and why we needed to change the behavior of the specific aliases based on it...

packages/kit/src/vite/index.js
```diff
@@ -61,7 +62,53 @@ const enforced_config = {
...
+	/** @type {import('types').ValidatedConfig} */
+	let svelte_config;
+
+	/** @type {import('vite').ConfigEnv} */
+	let vite_config_env;
+
+	/** @typedef {{public: {[k: string]: string;};private: {[k: string]: string;};}} ResolvedEnv */
+
+	/** @type {ResolvedEnv} */
+	let resolved_environment;
```
dominikg, line 83: are these needed? Currently only used locally in config hook

packages/kit/src/vite/index.js
```diff
@@ -61,7 +62,53 @@ const enforced_config = {
...
+ * Returns the SvelteKit environment plugin. Replaces `$env/static/*` imports
+ * with the actual environment variables.
+ * @return {import('vite').Plugin}
+ */
+function env() {
+	/** @type {import('types').ValidatedConfig} */
+	let svelte_config;
+
+	/** @type {import('vite').ConfigEnv} */
+	let vite_config_env;
```
dominikg, line 78: are these needed, only used in config hook atm

packages/kit/src/vite/index.js
```diff
@@ -61,7 +62,53 @@ const enforced_config = {
...
+
+	/** @type {ResolvedEnv} */
+	let resolved_environment;
+
+	return {
+		name: 'vite-plugin-svelte-kit:env',
+
+		async config(_, config_env) {
+			vite_config_env = config_env;
+			svelte_config = await load_config();
```
dominikg, line 90: there were previous efforts to avoid reading svelte config multiple times, as it may lead to unexpected side effects when it contains TLA initialization code.

To avoid this here we'd either have to steal the ref from kit plugin somehow or move the resolve + load hooks there

cc @benmccann

benmccann: I thought we could load multiple times as long as we use the same loading helper (just have to make sure the loading helper uses the mtime trick)

tcc-sejohnson: I'll let you guys duke this one out as I have no idea one way or the other. I like this as its own plugin (it's just way less confusing and way more obvious what it's doing), but I can move it back into the main plugin if there's a need.

Rich-Harris: I moved it into the main plugin — if we do get round to creating a helper that allows us to load the config multiple times without penalty then we should, but in the meantime i think it's better that we load it as few times as possible

Rich-Harris (approved):

Rich-Harris (approved):

PR #5831, opened 2022-08-05
[fix] handle lone returns in migration tool, turn on strict mode

dummdidumm:
Adding strict mode uncovered that we previously failed on lone `return;` statements. This PR fixes that - but it brought up an interesting question: Should lone `return;` statements still be allowed in the new `load` function? Right now it's implemented as "no, you need to adjust this". If we allow it, we should adjust the types for the handler functions in `+page.server.js` to allow it as well. If we don't allow it, we should adjust the `load` function return type to reflect hat. 

### Tests

### Changesets


packages/migrate/migrations/routes/migrate_server/index.js
```diff
@@ -30,102 +30,104 @@ export function migrate_server(content) {
...
-									return (
-										ts.isPropertyAssignment(prop) &&
-										ts.isStringLiteral(prop.name) &&
-										/set-cookie/i.test(prop.name.text)
-									);
-								});
-
-								if (!set_cookie_value || is_string_like(set_cookie_value)) {
-									safe_headers = true;
+					if (expr) {
```
Rich-Harris, line 36: what if we changed the implementation of `rewrite_returns` instead? I don't think there's any reason we'd need to do anything with a `return;`, we can just leave it be:

```diff
-if (ts.isReturnStatement(node)) {
+if (ts.isReturnStatement(node) && node.expression) {
```
```

File: packages/migrate/migrations/routes/migrate_page_js/index.js [hunk 2]
```diff
@@ -24,21 +24,21 @@ export function migrate_page(content) {
...
 		return content;
 	}
 
-	const name = file.exports.map.get('load');
-
 	for (const statement of file.ast.statements) {
-		const fn = get_function_node(statement, name);
-		if (fn) {
+		const fn = name ? get_function_node(statement, name) : undefined;
+		if (fn?.body) {
```
Rich-Harris, line 36: not sure i understand this — when would a function node not have a body? do we need to tighten up the implementation of `get_function_node`?

dummdidumm: I don't think we can do much here, that's typescript giving it an optional type. I don't think this can happen except when you have a syntax error, and unfortunately we can't just use `!` 😄

packages/migrate/migrations/routes/migrate_server/index.js
dummdidumm: Depends on how we answer the questions in the PR description

Rich-Harris: At present the type of `Load` is this: https://github.com/sveltejs/kit/blob/ac5b00bd477b35f1c90cb31418b41a8ef0502589/packages/kit/types/index.d.ts#L180-L186

i.e. a `void` return is fine, and so assuming we're okay with that then `return;` is fine

packages/migrate/migrations/routes/migrate_page_js/index.js
Rich-Harris: ah, of course — no requirement for the AST to be well formed. error-tolerant parsers are new to me!

packages/migrate/migrations/routes/migrate_server/index.js
dummdidumm: Right, but currently the `RequestHandler` type does not allow a `void` return - which makes me think that we should probably align these. So `Load` is ok with a `void`, so `GET` should, too. In the case of standalone endpoints we always need to return a response now. If we assume that people have their returns correctly already, we can just skip the `return;` case - if you agree with that, I'll adjust the code accordingly.

Rich-Harris: If standalone endpoint returns void today, it'll be an error, so I think it's fine for `RequestHandler` to require a `Response` return value — it'll just continue to error if it doesn't

dummdidumm: True - ok I'll adjust this to leave lone returns as-is then

Rich-Harris (approved):

Issue #3176, opened 2022-01-02
Provide ability to bundle dev dependencies for Node adapter


Conduitry:
### Describe the problem

Currently, the Node adapter will not bundle any external server side dependencies. This means that you need to make all of the dependencies you might need at runtime be prod dependencies. This differs from its (recent) previous behavior, and from Sapper's behavior, and from (I would argue) expected behavior.

Until now, it's been standard operating procedure to bundle as many of the server-side dependencies as possible, to reduce the deployed app size, parse time, and startup time. If there's some good reason to stop doing this, it should certainly at least be documented.

This issue may also well apply to other adapters (other than the static one), but I haven't tried to validate that.

### Describe the proposed solution

I'm not entirely sure. It would be nice to be able to provide sensible default Vite configuration so that we have Vite do it, rather than do this in a separate esbuild step afterwards (either automatically or manually).

I don't know whether some different behavior from Vite would make this easier on our end, or whether this is something that Vite ought to change.

Below, I describe what I'm currently doing in userland to deal with this, but I'm not positive that this would have an equivalent in-framework analogue. (Do we have convenient access to the user's project's `package.json`?)

### Alternatives considered

Currently, I have a `vite.ssr.noExternal` setting of `process.env.NODE_ENV === "development" ? undefined : Object.keys(pkg.devDependencies)` (where `pkg` is the parsed JSON from `package.json`) to force Vite to bundle all of my dev dependencies.

The check for development vs prod mode is there because I was having an issue using `Object.keys(pkg.devDependencies)` in dev mode that I haven't fully tracked down or tried to reproduce in a minimal repo, but which seem to stem from having a CJS-only dependency that Vite doesn't try to convert to ESM during dev mode.

### Importance

would make my life easier

Karlinator:
So there used to be an `esbuild` step in `adapter-node`, which would bundle everything. It was removed in #2931.

@Rich-Harris was this an unintended side-effect of that? I can't see it mentioned in the PR or discussion.

rpf3:
I ran into a similar issue to @Conduitry since upgrading and taking inspiration from this thread I added my package dependencies manually to the `vite.ssr.noExternal` config and got things working again.  This definitely feels like a breaking change in how the Node adapter previously worked though.

A path forward would be a welcome discussion.

SomaticIT:
Hello,

[...]

PR #5832, opened 2022-08-05
Fix hard dependance on ES modules

JoelEllis:
Previously, __dirname was redefined to a value derived from import.meta.url. This meant when the adapter was transpiled to commonJS, __dirname would be undefined, crashing the server. Now it only redefined if it does not have a truthy value.

---

  - (I'm not sure adding esbuild for just one test is a great idea)

### Tests

### Changesets

### DIFF
--- a/packages/adapter-node/src/handler.js
+++ b/packages/adapter-node/src/handler.js
@@ -19,7 +19,7 @@ const address_header = env('ADDRESS_HEADER', '').toLowerCase();
 const protocol_header = env('PROTOCOL_HEADER', '').toLowerCase();
 const host_header = env('HOST_HEADER', 'host').toLowerCase();
 
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const __dirname = __dirname || path.dirname(fileURLToPath(import.meta.url));
 
 /**
  * @param {string} path


packages/adapter-node/src/handler.js
```diff
@@ -19,7 +19,7 @@ const address_header = env('ADDRESS_HEADER', '').toLowerCase();
 const protocol_header = env('PROTOCOL_HEADER', '').toLowerCase();
 const host_header = env('HOST_HEADER', 'host').toLowerCase();
 
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const __dirname = __dirname || path.dirname(fileURLToPath(import.meta.url));
```
benmccann, line 22: this is failing `lint`. will the following work?
```suggestion
const __dirname = globalThis.__dirname || path.dirname(fileURLToPath(import.meta.url));
```

benmccann (commented):
I'm curious if there's a reason you can't use ES modules. Is there something else in the ecosystem that needs to be addressed?

Rich-Harris:
Thanks, though as noted above this definitely isn't the right fix as it's invalid JS.

Can you open an issue with a reproduction please?

JoelEllis:
Sorry about leaving this for so long - I had a bit of a busy couple of weeks. 

So - the story for this. I've been building a SvelteKit app that needs to run on a server with a bit of a 'pet' set up that runs a bunch of other webapps on it. I wanted to run it in a way that I could upload a single JS file to the server for easier updates. Unfortunately, the /client/ and /static/ directories can't be all bundled into one file. I compromised a bit to wanting a small, quick upload to a single folder. 
Directly copying the build folder did not work, for two reasons:
 - NodeJS does not allow using ES modules by default without `"type": "module"` in a package.json
 - The dependencies are not bundled
   - node_modules are huge, and include all dev dependences, with no way to separate them out without deleting and reinstalling production only. 
 
 I discovered the second one only *after* automating copying a package json into the build :(.
Of course, the obvious solution - bundle, transpile and treeshake it all!I used ESBuild, and it worked great first try, except for the one line of code here - it couldn't deal with `import.meta.url`, and when running the code with node, it errored out.

After much fiddling and trial and error, I came up with this:
@sveltejs__adapter-node@1.0.0-next.85.patch
```patch
diff --git a/files/handler-393b2283.js b/files/handler-393b2283.js
index c5588bfa8e4eb54fed732724e95aa6dab677a944..c6a50d6de94641f8e7f71eddaf704b48017d190c 100644
--- a/files/handler-393b2283.js
+++ b/files/handler-393b2283.js
@@ -1052,7 +1052,14 @@ const address_header = env('ADDRESS_HEADER', '').toLowerCase();
 const protocol_header = env('PROTOCOL_HEADER', '').toLowerCase();
 const host_header = env('HOST_HEADER', 'host').toLowerCase();
 
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
+let __dirname;
+try {
+  __dirname = path.dirname(fileURLToPath(import.meta.url))
+} catch (e) {
+  try {
+    __dirname = path.dirname(resolve(".", process.argv[1] ))
+  } catch (e) { }
+}
 
 /**
  * @param {string} path
  ```
  
With the caveat that node must be run with the filename to be executed rather than the folder, defaulting to index,js or the settings in package.json (`"main"` I think?). There's probably a way to account for this, or a better way of doing it in general, but this worked for me. 

Again, apologies for leaving this so long, and for my writing skills. I also did this all via the web interface, so sorry for not running all the scripts as well.

benmccann:
Your latest proposed patch is even less likely to be merged, so I'm going to go ahead and close this.

If you're just trying to bundle dependencies, you can do something like this: https://github.com/sveltejs/kit/issues/3176#issuecomment-1176372356

Otherwise, you should use the custom server option: https://github.com/sveltejs/kit/tree/master/packages/adapter-node#custom-server

PR #5848, opened 2022-08-08
[package] remove tooling specific settings in `publishConfig` from generated package.json

### LINKED ISSUE #5830: Package command copies `publishConfig.directory` into the generated package.json
### Describe the problem

Referencing `svelte-kit package` generated packages through the [workspace protocol](https://pnpm.io/workspaces) of pnpm doesn't work out of the box. Most monorepo tools (pnpm included) will normally just symlink the root/source directory when you use a workspace version (like `workspace:*`). Since `svelte-kit package` outputs a new directory structure with its own `package.json` file, that doesn't work out of the box. You can however use a new feature ([see this issue](https://github.com/pnpm/pnpm/issues/3901)) in pnpm that allows you to specify the directory of the package (containing a new package.json) through the `publishConfig.directory`. This works great, and solves almost all of the problems. However, the package command copies the `publishConfig.directory` setting over to the new generated package.json file. This can cause a lot of issues, since that directory does not exist.

I suggest just removing the `publishConfig.directory` key if it exist from the generated `package.json` file, just like you already do for the script tag. I believe something like `delete pkg.publishConfig?.directory` close to [where you delete the script key](https://github.com/sveltejs/kit/blob/6abbdd7b927e5384dd1c2c8325180de083002caf/packages/kit/src/packaging/utils.js#L106) would suffice. However, it might be cases where this is not the desired behavior.

### Alternatives considered

It does not seem to cause immediate problems for pnpm at least, since it only looks at the `publishConfig.directory` key once for the root/source directory, and not recursively. So it might be fine to keep it as is. But to me having the `publishConfig.directory` inside the generated package.json doesn't make sense, and I can imagine several cases where it would be troublesome.

### Importance

nice to have

### Additional Information

_No response_

boyeborg:
I know you're currently doing a big rewrite. This is not an important fix, no rush in merging it.

Closes #5830 

**TL;DR**: Removes `publishConfig.directory` from the generated package.json when using the package command. This setting is usually added to tell monorepo tools or publishing tools where you're package is located, i.e. the output directory of the package command. Copying that over to the generated package.json file does not make contextual sense, since the referenced directory don't exist, and breaks some monorepo tools.

I work in several monorepos where we have a wokspace based setup (usually using pnpm). In those monorepos we have several svelte packages created by the SvelteKit package command. Usually when you run this setup, you can use the [workspace protocol](https://pnpm.io/workspaces) to locally link one of those packages to an application living inside your monorepo (e.g. linking the charting package to the admin dashboard app). However, since SvelteKit's package command generates a new folder structure with its own package.json file, this does not work out of the box. PNPM has however [recently added a feature](https://github.com/pnpm/pnpm/issues/3901) (or fixed an old one, depending on how you look at it) that fixes this problem. It allows you to add a new location for your outputted package within the original package.json file using the `publishConfig.directory` key. By setting that to the package command's output directory, pnpm will link everything correctly. However, the package command copies the `publishConfig.directory` setting to the generated package.json file as well. This might cause some issues, since the referenced directory does not exist. It only exist in context of the original package.json. If you for example try to publish the package, you might get errors when the publishing tool does not find the referenced directory.

This PR tries to fix that potential issue by explicitly removing the `publishConfig.directory` key from the generated package.json file if it exists. However, this might not always yield the expected result. There might exist cases where you want to use that setting in the generated package that I haven't thought of.

### Tests

### Changesets


dummdidumm:
Thank you! PR looks good, two things:
- Is there anything else in `publishConfig` that could make problems, so should the whole object be removed?
- We should mention that in the README that this field is removed so if someone needs this field they can at least know this from consulting the docs after being confused

boyeborg:
@dummdidumm Thanks!
I don't we should remove the whole object, since you can set things like custom npm registry and access there, that makes sense to copy over to the generated package. The [official docs from NPM](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#publishconfig) only states that you can set whatever config options that are available for the `npm publish` command. So I guess that's mostly `--registry`, `--access` and `--tag`. All of which seems reasonable to keep.
The [pnpm docs](https://pnpm.io/package_json#publishconfig) has a couple of extra uses for the `publishConfig` object (including the `directory` part), most of which fields you can set that pnpm will override the published package.json with when publishing. I guess that if you set any of those, and also use `svelte-kit package` to package the code you know what you're doing, and do it for a reason, so I don't think those should be removed.
But, since we're removing the `publishConfig.directory`, I think it also makes sense to remove the `publishConfig.linkDirectory`, since that's linked to the directory setting, and don't make sense without it. I'll add that to the list of keys. I'll also add the changes to the docs.

Rich-Harris (approved):

Rich-Harris:
thanks!

futantan:
hi @boyeborg I noticed that you are using `svelte-kit package` in monorepo, I'd link to ask you for help: How can we publish it to the npm because `svelte-kit package` generate a new folder contains a `package.json` file

```
.
├── README.md
├── package <-- here it what we what to publish
├── src
├── package.json
├── svelte.config.js
└── tsconfig.json
```

We want to use this in monorepo, so in the top level `package.json` of this npm package,  I added the following:
```
  "main": "package/index.js",
  "svelte": "package/index.js",
```
the question is how can I use `changeset publish` of the monorepo to publish the generated `package` folder?

boyeborg:
@futantan I'm not too familiar with how changeset work, but the way we do it is to add this to to root `package.json`:

```
{
  "publishConfig": {
    "directory": "package",
    "linkDirectory": true
  },
}
```

By having the `publishConfig.directory` set to point to the `svelte-kit package` output directory, all tooling we use understand that that's the home of the distributable package. If the changeset CLI doesn't use `publishConfig.directory`, I would encourage you to open an issue asking about it in the changeset repo.

PR #5853, opened 2022-08-08
Unbundling

Rich-Harris:
This is part of #5778, but it's a shift in strategy that's separate from #5748 so I figured I'd make this PR as a quick detour to explain it.

Currently, `@sveltejs/kit` is bundled. Imports like `@sveltejs/kit/vite` and `@sveltejs/kit/node` are bundled in `dist`, while the modules behind the `$app` alias are bundled in `assets`, and copied over to `.svelte-kit/runtime`. Bundling was necessary in the very earliest days, when we were using `.ts` files, and _in theory_ it has a benefit today in the form of slightly quicker installations, but in practice it adds considerable complexity.

It means that in order to test the `src` code locally without rebuilding on every change, we need to have different code paths (switching on `process.env.BUNDLED`) that are rather cryptic. It reduces the quality of stack traces. It adds a build step that isn't really necessary. 

In #5778, it makes things very complicated indeed. In order for the new `error` and `redirect` helpers to work, we need to be able to detect, from a `catch` clause, that they were used. The best (and most TypeScript-friendly) way to do this is an `instanceof` check that uses the internal `HttpError` and `Redirect` classes used by those helpers. In the interests of not over-committing API surface area, these live in a private module. But this means that `error` and `redirect` imported from the public `@sveltejs/kit` module are referring to different _copies_ of those classes than the code that does the `instanceof` checks, and there's really no way to fix that without introducing a bunch more complexity.

Unless, that is, we just get rid of the bundling altogether. Since SvelteKit has few dependencies, most of which have no or few dependencies of their own, I think the slight hit to installation time is more than justified by the substantial reduction in complexity.


Rich-Harris:
(An aside: the ability to do this without losing type safety is one of the main reasons I'm a vocal proponent of using TypeScript via JSDoc, rather than via `.ts` files.)

packages/kit/package.json
```diff
@@ -12,37 +12,33 @@
...
+		"kleur": "^4.1.4",
+		"magic-string": "^0.26.2",
+		"mime": "^3.0.0",
+		"node-fetch": "^3.2.4",
 		"sade": "^1.8.1",
-		"tiny-glob": "^0.2.9"
+		"set-cookie-parser": "^2.4.8",
+		"sirv": "^2.0.2",
+		"tiny-glob": "^0.2.9",
+		"typescript": "^4.7.4",
```
dominikg, line 25: oof

dominikg: sorry, should have been more wordy :see_no_evil: 
is it possible to not depend on typescript here? That's not small and even if users mostly are going to have it around during dev anyways, outright depending on it would lead to `npm i --production` pulling it too, which i think is not good / too heavy

Rich-Harris: this is the plan, yeah — it's only necessary for generated types, and if you don't have typescript in your app then there's no point in generating those types. so ideally we'll depend on the app's copy of typescript. i was just too lazy to do that

Issue #307, opened 2021-01-06
manage focus, and use aria-live to announce navigation


Rich-Harris:
https://github.com/sveltejs/sapper/issues/1083, basically.

There are two separate issues to consider in order to make client-side routing accessible:

## 1. Managing focus

In a server-rendered app, any time you navigate to a page, the `<body>` is focused. In Sapper and SvelteKit apps, we blur the current `activeElement` upon navigation, but this doesn't actually _reset_ the focus, which is what we need — it just removes it temporarily. As soon as you press the tab key (or shift-tab), focus moves to the element after (or before) whichever element was focused prior to the navigation, assuming it still exists in the DOM. This is not desirable.

Went down a bit of a rabbit hole trying to understand current recommendations:

* [Marcy Sutton](https://www.gatsbyjs.com/blog/2019-07-11-user-testing-accessible-client-routing/) advises focusing a 'skip back to navigation' link (i.e. a skip nav link in reverse)
* [Ryan Florence](https://twitter.com/ryanflorence/status/1130488429785915393) thinks the app developer should have control over where focus lands, but that it should be an element inside the innermost route that has changed
* [Nick Colley](https://www.slideshare.net/frontenders-valtech/accessible-client-side-routing-nick-colley) and [Daniel Nixon](https://github.com/oaf-project/oaf-svelte-routing) prefer finding some primary focus target, e.g. the `<h1>`
* [David Luhr](https://github.com/sveltejs/sapper/issues/1083) suggests resetting focus to the top of the DOM

I'm sure there are other suggestions too. Each comes from an accessibility expert and has a solid argument, but ultimately they are mutually exclusive. Any would be better than the status quo though.

My inclination in the short term is to simply focus `<body>` (i.e. David Luhr's suggestion) — it's easy to implement, and matches the behaviour of server-rendered apps. In the future perhaps we could investigate alternatives, particularly ones that put power in app developers' hands rather than making one-size-fits-all decisions for them.

The short term solution ought to be fairly straightforward — we just change this...

https://github.com/sveltejs/kit/blob/e09b6be95af81077e0fb0240ad94b1c0e41c9b3a/packages/kit/src/runtime/internal/router/index.js#L180-L182

...to this:

```js
document.body.setAttribute('tabindex', '-1');
document.body.focus();
```

(Better still, only set the `tabindex` once, when the router initialises.)

## 2. Announcing the page change

[...]

Issue #3091, opened 2021-12-22
Templating app.html


ebeloded:
### Describe the problem

Currently SvelteKit templating is fairly simple: it replaces `%svelte.head%` and `%svelte.body%` entries in `app.html`.

However, sometimes it is necessary to augment some parts of `app.html` that are outside of Svelte's control. 

For example, for i18n one needs to be able to change the `lang` attribute on `<html>` or set a class on `<body>` tag.

### Describe the proposed solution

It would be useful if SvelteKit provided some built-in templating solution for `app.html` that extends beyond replacing `%svelte.head%` and `%svelte.body%` and provides access other request parameters.

### Alternatives considered

It is possible to do such templating in `handle` hook. For example, we can define some logic in `app.html` in EJS format and then run the templating engine on `response.body` before returning it from the hook.

### Importance

would make my life easier

miohtama:
As a workaround please checl [svelte-body](https://www.npmjs.com/package/svelte-body) package. Though not sure if it can do `<HTML lang>`.

abdo643-HULK:
it can be done with <svelte:body> but with a lot of limitations right now, but there is an open issue for that: https://github.com/sveltejs/svelte/issues/3105. For the <html> you can't but any attributes that need to change in the app.html file for this to work. I use it like this in my project : 
```
<svelte:head>
	<html {lang} {dir} />
</svelte:head>
```

arichard-info:
> it can be done with svelte:body but with a lot of limitations right now, but there is an open issue for that: [sveltejs/svelte#3105](https://github.com/sveltejs/svelte/issues/3105). For the `<html>` you can't but any attributes beforehand in the app.html file that need to change for this to work. I use it like this in my project :
> 
> ```
> <svelte:head>
> 	<html {lang} {dir} />
> </svelte:head>
> ```

Wow thanks for the `<svelte:head>` trick. Didn't know that `<html>` was a valid tag. Is is documented somewhere ?

abdo643-HULK:
> > it can be done with svelte:body but with a lot of limitations right now, but there is an open issue for that: [sveltejs/svelte#3105](https://github.com/sveltejs/svelte/issues/3105). For the `<html>` you can't but any attributes beforehand in the app.html file that need to change for this to work. I use it like this in my project :
> > ```
> > <svelte:head>
> > 	<html {lang} {dir} />
> > </svelte:head>
> > ```
> 
[...]

PR #5856, opened 2022-08-08
Add accessibility documentation

geoffrich:
Add docs around SvelteKit's accessibility features and what developers need to do to play nicely with them.

I focused these docs on SvelteKit-specific concerns (route announcements, focus management) instead of general a11y practices (e.g. skip link, button vs link, etc). In the future, a "Guides" section on learn.svelte.dev may be a good place to go deeper into general a11y best practices, or integrated into the official SvelteKit tutorial.

I'm looking for feedback on prose, accuracy of information, and thoughts on anything else we want to include (or exclude).

Leaving in draft for now until the dust settles on the big routing changes.

Prior art:
- [Next.js](https://nextjs.org/docs/accessibility)
- [Remix](https://remix.run/docs/en/v1/guides/accessibility)
- [Ember](https://guides.emberjs.com/release/accessibility/application-considerations/)

Context around route announcement/focus management: https://github.com/sveltejs/kit/issues/307

---

### Tests

### Changesets

### DIFF
--- a/documentation/docs/19-accessibility.md
+++ b/documentation/docs/19-accessibility.md
@@ -1,0 +1,63 @@
+---
+title: Accessibility
+---
+
+SvelteKit strives to provide an accessible platform for your app by default. Svelte's [compile-time accessibility checks](https://svelte.dev/docs#accessibility-warnings) will also apply to any SvelteKit application you build. 
+
+Here's how SvelteKit's built-in accessibility features work and what you need to do to help these features to work as well as possible.
+
+We recognize that accessibility can be hard to get right. If you want to suggest improvements to how SvelteKit handles accessibility, please [open a GitHub issue](https://github.com/sveltejs/kit/issues).
+
+### Route announcements
+
+In traditional server-rendered applications, every navigation (e.g. clicking on an `<a>` tag) triggers a full page reload. When this happens, screen readers and other assistive technology will read out the new page's title so that users understand that the page has changed.
+
+Since navigation between pages in SvelteKit happens without reloading the page (known as [client-side routing](/docs/appendix#routing)), SvelteKit injects a [live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) onto the page that will read out the new page name after each navigation. This determines the page name to announce by inspecting the `<title>` element.
+
+Because of this behavior, every page in your app should have a unique, descriptive title. In SvelteKit, you can do this by placing a `<svelte:head>` element on each page:
+
+```svelte
+/// file: src/routes/index.svelte
+<svelte:head>
+	<title>Todo List</title>
+</svelte:head>
+```
+
+This will allow screen readers and other assistive technology to identify the new page after a navigation occurs. Providing a descriptive title is also important for [SEO](/docs/seo#manual-setup-title-and-meta).
+
+### Focus management
+
+In traditional server-rendered applications, every navigation will reset focus to the top of the page. This ensures that people browsing the web with a keyboard or screen reader will start interacting with the page from the beginning.
+
+To simulate this behavior during client-side routing, SvelteKit focuses the `<body>` element after each navigation. If you want to customize this behavior, you can implement custom focus management logic using the `afterNavigate` hook:
+
+```js
+/// <reference types="@sveltejs/kit" />
+// ---cut---
+import { afterNavigate } from '$app/navigation';
+
+afterNavigate(() => {
+	/** @type {HTMLElement | null} */
+	const to_focus = document.querySelector('.focus-me');
+	to_focus?.focus();
+});
+```
+
+You can also programmatically navigate to a different page using the [`goto`](/docs/modules#$app-navigation-goto) function. By default, this will have the same client-side routing behavior as clicking on a link. However, `goto` also accepts a `keepfocus` option that will preserve the currently-focused element instead of resetting focus. If you enable this option, make sure the currently-focused element still exists on the page after navigation.
+
+### Lang attribute
+
+By default, SvelteKit's page template sets the default language of the document to English. If your content is not in English, you should update the `<html>` element in `src/app.html` to have the correct [`lang`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang#accessibility) attribute. This will ensure that any assistive technology reading the document uses the correct pronunciation. For example, if your content is in German, you should update `app.html` to the following:
+
+```html
+/// file: src/app.html
+<html lang="de">
+```
+
+### Further reading
+
+For the most part, building an accessible SvelteKit app is the same as building an accessible web app. You should be able to apply information from the following general accessibility resources to any web experience you build:
+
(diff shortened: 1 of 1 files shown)


documentation/docs/19-accessibility.md
```diff
@@ -0,0 +1,63 @@
...
+afterNavigate(() => {
+	/** @type {HTMLElement | null} */
+	const to_focus = document.querySelector('.focus-me');
+	to_focus?.focus();
+});
+```
+
+You can also programmatically navigate to a different page using the [`goto`](/docs/modules#$app-navigation-goto) function. By default, this will have the same client-side routing behavior as clicking on a link. However, `goto` also accepts a `keepfocus` option that will preserve the currently-focused element instead of resetting focus. If you enable this option, make sure the currently-focused element still exists on the page after navigation.
+
+### Lang attribute
```
geoffrich, line 48: It might also be good to show how to dynamically update the `lang` attribute for multi-lang sites - see https://github.com/sveltejs/kit/issues/3091#issuecomment-1112589090

documentation/docs/19-accessibility.md
```diff
@@ -0,0 +1,63 @@
...
+import { afterNavigate } from '$app/navigation';
+
+afterNavigate(() => {
+	/** @type {HTMLElement | null} */
+	const to_focus = document.querySelector('.focus-me');
+	to_focus?.focus();
+});
+```
+
+You can also programmatically navigate to a different page using the [`goto`](/docs/modules#$app-navigation-goto) function. By default, this will have the same client-side routing behavior as clicking on a link. However, `goto` also accepts a `keepfocus` option that will preserve the currently-focused element instead of resetting focus. If you enable this option, make sure the currently-focused element still exists on the page after navigation.
```
MelSumner, line 46: [question, suggestion] Is there a fallback if the element no longer exists? Might be useful to let devs know what the outcome would be, either way. 👍

MelSumner (commented):
@geoffrich This is really useful information. It identifies what's available, what users need to do themselves, and reliable links for additional info. 👍

accuser:
@geoffrich this is a great start, and the snippets you have included are really helpful.

My only caution is that getting a11y right requires a lot more thought, and I wonder if this needs to be more explicit in the documentation? i.e., whilst Svelte/SvelteKit provides some really good helpers, on their own they don't make an a11y app. 

You're already thinking about this with the proposed 'Guides' section on learn.svelte.dev, but helping people be aware of the built-in capabilities as well as understand that these alone don't make an accessible app would be a really open approach to building a better AX.

documentation/docs/19-accessibility.md
accuser: NB: the example in [#3091 (comment)](https://github.com/sveltejs/kit/issues/3091#issuecomment-1112589090) refers to `transformPage()` but this was replaced with `transformPageChunk()` in #5657.

documentation/docs/19-accessibility.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Accessibility
+---
+
+SvelteKit strives to provide an accessible platform for your app by default. Svelte's [compile-time accessibility checks](https://svelte.dev/docs#accessibility-warnings) will also apply to any SvelteKit application you build. 
+
+Here's how SvelteKit's built-in accessibility features work and what you need to do to help these features to work as well as possible.
+
+We recognize that accessibility can be hard to get right. If you want to suggest improvements to how SvelteKit handles accessibility, please [open a GitHub issue](https://github.com/sveltejs/kit/issues).
+
```
benmccann, line 10: I'm not really sure we need this sentence. It could apply to any section as we want people to open issues about issues they encounter anywhere.
```suggestion
```

MelSumner: [comment] I thought this bit was thoughtful; it's signaling to the accessibility community that their feedback is welcome, which is a welcome exception in tech (by large, over the last 30 years). The sentence may not be a technical necessity but it might be a community necessity. Just a perspective to consider. 👍

documentation/docs/19-accessibility.md
geoffrich: There's no fallback element -- when `keepfocus` is set, SvelteKit won't call `focus` at all. So if the element is removed, focus will be lost just as if you removed the currently focused element in vanilla JS. I can try to clarify this part a bit. Maybe add something like:

> If the element no longer exists, the user's focus will be lost.

geoffrich:
> My only caution is that getting a11y right requires a lot more thought, and I wonder if this needs to be more explicit in the documentation? i.e., whilst Svelte/SvelteKit provides some really good helpers, on their own they don't make an a11y app.

@accuser would something to this effect in the intro be what you're looking for? "Keep in mind that while SvelteKit aims to provide an accessible foundation, you are still responsible for making sure your application code is accessible. If you're new to accessibility, see the "Further Reading" section of this guide for additional resources."

I had something similar to that in one of my drafts, but it got lost along the way.

documentation/docs/19-accessibility.md
geoffrich: I can go either way on this, but Melanie captures why I included it in the first place. It emphasizes that we want to be receptive to a11y suggestions and improve as needed.

dummdidumm: Agree with keeping it

geoffrich:
The big `+page` PR merged, so this is ready for review now.

documentation/docs/19-accessibility.md
```diff
@@ -0,0 +1,88 @@
...
+ * @returns {string}
+ */
+function get_lang(event) {
+	return 'en';
+}
+// ---cut---
+/** @type {import('@sveltejs/kit').Handle} */
+export function handle({ event, resolve }) {
+	return resolve(event, {
+		transformPageChunk: ({ html }) => html.replace('%lang', get_lang(event))
```
Rich-Harris, line 77:
```suggestion
		transformPageChunk: ({ html }) => html.replace('%lang%', get_lang(event))
```

documentation/docs/19-accessibility.md
```diff
@@ -0,0 +1,88 @@
...
+afterNavigate(() => {
+	/** @type {HTMLElement | null} */
+	const to_focus = document.querySelector('.focus-me');
+	to_focus?.focus();
+});
+```
+
+You can also programmatically navigate to a different page using the [`goto`](/docs/modules#$app-navigation-goto) function. By default, this will have the same client-side routing behavior as clicking on a link. However, `goto` also accepts a `keepfocus` option that will preserve the currently-focused element instead of resetting focus. If you enable this option, make sure the currently-focused element still exists on the page after navigation. If the element no longer exists, the user's focus will be lost, making for a confusing experience for assistive technology users.
+
+### Lang attribute
```
Rich-Harris, line 48: super minor aesthetic point but how about changing it to this, so that `lang` stays lowercase?

```suggestion
### The "lang" attribute
```

Rich-Harris (approved):
this is great! i added one very small suggestion, but either way this gets a hearty 👍 from me

documentation/docs/19-accessibility.md
geoffrich: Good with me 👍

Rich-Harris (approved):

PR #5882, opened 2022-08-13
Nested layout is in the subdirectory remark.

RaduTerec:
Minor documentation edit.

Added a remark for better understanding where the nested layout should be placed.


Rich-Harris:
Thank you — the file this PR changes is out of date due to #5778, and layouts are now discussed in https://kit.svelte.dev/docs/routing#layout-layout-svelte. I think the wording is sufficiently clear

PR #5861, opened 2022-08-10
[fix] support linked @sveltejs/kit package in Vite 3.0.3+

benmccann:

PR #5901, opened 2022-08-16
[chore] fix migration test linting

benmccann:
lint is failing on https://github.com/sveltejs/kit/pull/5861. this should fix it

### DIFF
--- a/packages/migrate/migrations/routes/migrate_page_server/index.spec.js
+++ b/packages/migrate/migrations/routes/migrate_page_server/index.spec.js
@@ -8,7 +8,7 @@ for (const sample of read_samples(import.meta.url)) {
 		if (!sample.filename) {
 			throw new Error('Missing filename for sample');
 		}
-		const actual = migrate_page_server(sample.before, sample.filename);
+		const actual = migrate_page_server(sample.before, sample.filename ?? '+page.server.js');
 		assert.equal(actual, sample.after);
 	});
 }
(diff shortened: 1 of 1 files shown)


packages/migrate/migrations/routes/migrate_page_server/index.spec.js
```diff
@@ -5,6 +5,9 @@ import { migrate_page_server } from './index.js';
 
 for (const sample of read_samples(import.meta.url)) {
 	test(sample.description, () => {
+		if (!sample.filename) {
+			throw new Error('Missing filename for sample');
+		}
```
Rich-Harris, line 10:
```suggestion
```

packages/migrate/migrations/routes/migrate_page_server/index.spec.js
```diff
@@ -5,6 +5,9 @@ import { migrate_page_server } from './index.js';
...
 		const actual = migrate_page_server(sample.before, sample.filename);
```
Rich-Harris, line 11:
```suggestion
		const actual = migrate_page_server(sample.before, sample.filename ?? '+page.server.js');
```

packages/migrate/migrations/routes/migrate_page_server/index.spec.js
```diff
@@ -5,7 +5,7 @@ import { migrate_page_server } from './index.js';
 
 for (const sample of read_samples(import.meta.url)) {
 	test(sample.description, () => {
-		const actual = migrate_page_server(sample.before, sample.filename);
+		const actual = migrate_page_server(sample.before, sample.filename ?? '+page.server.js');
```
benmccann, line 8: Should this defaulting be done inside the method? I.e. make the second parameter optional

Rich-Harris: this code is saying that all samples implicitly _do_ have a filename, `+page.server.js`, unless one is specified. making the parameter optional for the sake of the tests is the wrong solution

Rich-Harris (approved):

PR #5907, opened 2022-08-16
make generated type import path ends in `.js` 

### LINKED ISSUE #5899: generated .../$types.d.ts has imports with '.ts'-extension which causes the new +page.server.ts/+page.svelte data prop to have type 'any'
update: see https://github.com/sveltejs/kit/issues/5899#issuecomment-1215987727 for the actual issue

---

### Describe the bug

First off - maybe this is expected? But when I read:
https://kit.svelte.dev/docs/routing#$types
> Throughout the examples above, we've been importing types from a $types.d.ts file. This is a file SvelteKit creates for you in a hidden directory if you're using TypeScript (or JavaScript with JSDoc type annotations) to give you type safety when working with your root files.
> 
> For example, annotating export let data with PageData (or LayoutData, for a +layout.svelte file) tells TypeScript that the type of data is whatever was returned from load:

I thought that the `data` prop would be fully typed!

With the `+page.server.ts`:
```ts
import type { PageServerLoad } from './$types';

import { getEvent } from '$lib/api';

export const load: PageServerLoad = async ({ params }) => {
	const { event, eventResponses } = await getEvent(params.id);

	return {
		event: event,
		eventResponses: eventResponses
	};
};
```
and `+page.svelte`:
```svelte
<script lang="ts">
	import type { PageData } from './$types';

	export let data: PageData;
</script>
```
I expected `data` to have the type
```
data: {
    event: event,
    eventResponses: eventResponses
}
```
but instead, as noted earlier, I get `data: any` (hope you understand the notation used)

### Reproduction

see the description

can also clone or checkout my repo:
https://github.com/ollema/sersophane/tree/b1380f17998c430e7f91404f92d6fed21e37c001/src/routes/events/%5Bid%5D

I think this is not super hard to reproduce - either it should happen for everyone OR I might have just misinterpreted the $types chapter

### Logs
I get this "problem" from vscode when viewing `/home/s0001325/work/sersophane/tsconfig.json`:

```
[{
    "resource": "/home/s0001325/work/sersophane/tsconfig.json",
    "owner": "typescript",
    "severity": 8,
    "message": "Cannot write file '/home/s0001325/work/sersophane/src/routes/+page.js' because it would overwrite input file.",
    "source": "ts",
    "startLineNumber": 1,
    "startColumn": 1,
    "endLineNumber": 1,
    "endColumn": 2
}]
```
Tried restarting vscode, tried reinstalling packages etc. it is probably not related but it was not there before the migration. Who knows!

My `tsconfig.json` is very simple:
```
{
	"compilerOptions": {
		"allowJs": true,
		"checkJs": true,
		"esModuleInterop": true,
		"forceConsistentCasingInFileNames": true,
		"resolveJsonModule": true,
		"skipLibCheck": true,
		"sourceMap": true,
		"strict": true
	}
}
```

### System Info

```shell
System:
    CPU: (16) x64 Intel(R) Core(TM) i9-10885H CPU @ 2.40GHz
    Memory: 19.86 GB / 31.20 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.7.0 - ~/.nvm/versions/node/v18.7.0/bin/node
    npm: 8.15.0 - ~/.nvm/versions/node/v18.7.0/bin/npm
  Browsers:
    Chrome: 104.0.5112.79
  npmPackages:
    @sveltejs/adapter-node: next => 1.0.0-next.86 
    @sveltejs/kit: next => 1.0.0-next.407 
    svelte: ^3.49.0 => 3.49.0 
    vite: ^3.0.7 => 3.0.7
```

### Severity

annoyance

### Additional Information

_No response_

ignatiusmb:
Fixes #5899 

Is there any other possible extensions? Should we have a utility function to extract and replace the extension of a filename?

### Tests

### Changesets

### DIFF
--- a/.changeset/twelve-stingrays-cry.md
+++ b/.changeset/twelve-stingrays-cry.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/kit": patch
+---
+
+make generated type import path ends in `.js` 
--- a/packages/kit/src/core/sync/write_types.js
+++ b/packages/kit/src/core/sync/write_types.js
@@ -390,18 +390,13 @@ function process_node(ts, node, outdir, params, groups) {
 			const types = [];
 			for (const method of ['POST', 'PUT', 'PATCH']) {
 				if (proxy.exports.includes(method)) {
-					if (proxy.modified) {
-						types.push(`Kit.AwaitedErrors<typeof import('./proxy${basename}').${method}>`);
-					} else {
-						// If the file wasn't tweaked, we can use the return type of the original file.
-						// The advantage is that type updates are reflected without saving.
-						types.push(
-							`Kit.AwaitedErrors<typeof import("${path_to_original(
-								outdir,
-								node.server
-							)}").${method}>`
-						);
-					}
+					// If the file wasn't tweaked, we can use the return type of the original file.
+					// The advantage is that type updates are reflected without saving.
+					const from = proxy.modified
+						? `./proxy${basename}`
+						: path_to_original(outdir, node.server);
+
+					types.push(`Kit.AwaitedErrors<typeof import('${from}').${method}>`);
 				}
 			}
 			errors = types.length ? types.join(' | ') : 'null';
--- a/packages/kit/src/core/sync/write_types.js
+++ b/packages/kit/src/core/sync/write_types.js
@@ -393,7 +393,7 @@ function process_node(ts, node, outdir, params, groups) {
 					// If the file wasn't tweaked, we can use the return type of the original file.
 					// The advantage is that type updates are reflected without saving.
 					const from = proxy.modified
-						? `./proxy${basename}`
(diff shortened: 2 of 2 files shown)


juanvillacortac (approved):
I confirm that this solves this problem which is really annoying. I hope it can be merged soon. This is making migrating to the new API quite annoying if I have my project strongly typed.

dummdidumm (approved):
Ah, interesting, I had the same problem in another case but didn't realize this would also affect TS proxies. I consolidated the logic 👍

ollema:
I updated to 409 and it's still not working, how can I debug this? :thinking:

iceghost:
@igna Looks like this PR only add the extension to `AwaitedErrors`. `AwaitedProperties` fix is missing.

PR #5916, opened 2022-08-16
[fix] correctly provide server parent data

### LINKED ISSUE #5904: +layout.server.js not correctly passing data down during client-side hydration
### Describe the bug

After client side navigation, calling `await parent()` from within a +page.server.js exported load function returns `{"data":null}` instead of the value returned from the +layout.server.js exported load function

### Reproduction

- Load https://stackblitz.com/edit/sveltejs-kit-template-default-8az3ew?file=src/routes/+page.server.js
- Note that layout data, and page data are correct initially
- Click on 'Another page' link, to do a client side navigation to a new page
- Note that layout data is now incorrect (as is the layout data that was merged into page data)

### Logs

_No response_

### System Info

```shell
stackblitz default template fork
```

### Severity

### Additional Information

_No response_

dummdidumm:
Fixes #5904

- the data property was not unwrapped on the server, resulting in wrong results on client side navigations
- the root layout data was ommitted because of a wrong undefined check

### Tests

### Changesets


dummdidumm:
Tests fail because the selectors are queried too soon when the old value is still present in that text field. Turns out we need to adjust the way we write test (or rather, we did it wrong previously).
Instead of
```ts
expect(await page.textContent('h2')).toBe('2');
```
it should be
```ts
await expect(page.locator('h2')).toHaveText('2');
```
which automatically retries until the text is matched or it times out

packages/kit/src/runtime/server/index.js
```diff
@@ -250,7 +250,7 @@ export async function respond(request, options, state) {
 								try {
 									if (error) return;
 
-									const node = n ? await options.manifest._.nodes[n]() : undefined;
+									const node = n !== undefined ? await options.manifest._.nodes[n]() : undefined;
```
Rich-Harris, line 253:
```suggestion
									// == because it could be undefined (in dev) or null (in build, because of JSON.stringify)
									const node = n == undefined ? n : await options.manifest._.nodes[n]();
```

packages/kit/src/runtime/server/index.js
```diff
@@ -260,7 +261,11 @@ export async function respond(request, options, state) {
 												/** @type {import('types').JSONObject} */
 												const data = {};
 												for (let j = 0; j < i; j += 1) {
-													Object.assign(data, await promises[j]);
+													const parent = await promises[j];
+													if (!parent || parent instanceof HttpError || 'error' in parent) {
+														continue;
```
Rich-Harris, line 266: no point going any further, we can just bail at this point

```suggestion
														return;
```

packages/kit/src/runtime/server/index.js
```diff
@@ -260,7 +261,11 @@ export async function respond(request, options, state) {
 												/** @type {import('types').JSONObject} */
 												const data = {};
 												for (let j = 0; j < i; j += 1) {
-													Object.assign(data, await promises[j]);
+													const parent = await promises[j];
+													if (!parent || parent instanceof HttpError || 'error' in parent) {
+														return;
```
Conduitry, line 266: Is this correct? You just `return undefined`? You don't want to `break` or `return data`?

dummdidumm, line 266: To please TS
```suggestion
														return data;
```

dummdidumm: I just changed this to `return data`. I don't think it makes a difference in the end because everything below an error is discarded, but it makes TS shut up, so ..

Rich-Harris: actually you know what? we should be throwing in that case

Rich-Harris: hang on, we're making this way too complicated. the try-catch is in the wrong place. one sec...

Rich-Harris: cancel that, i confused myself — forgot we need to return `[{ data }, { data }, { error }]` to render the correct `+error.svelte` page

Rich-Harris (approved):

PR #5925, opened 2022-08-16
[chore] more robust tests

dummdidumm:
using
await expect(page.locator('h2')).toHaveText('2');
instead of
expect(await page.textContent('h2')).toBe('2');
makes the tests more robust because the latter retries until it times out or the text is present. The former will resolve as soon as the element is found, but it may still have an old value from before the tests action

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/apps/basics/src/routes/store/data/only-data-changes/+page.svelte
+++ b/packages/kit/test/apps/basics/src/routes/store/data/only-data-changes/+page.svelte
@@ -1,9 +1,8 @@
 <script>
-    import { page, session } from '$app/stores';
+    import { session } from '$app/stores';
     export let data;
 </script>
 
 <pre id="page-data">{JSON.stringify(data)}</pre>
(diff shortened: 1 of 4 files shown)


Rich-Harris:
Am conflicted about this. It would definitely make sense in the context of an app, but for us it's important that the testing framework isn't smoothing over any timing issues — it gives us less confidence that (for example) the DOM has been updated immediately after a specific promise resolves, etc

dummdidumm:
But how to tell the test framework that the thing you want to run finished? As mentioned in #5916, tests are flaky because the selector runs too soon after for example a button click. How to tell if this is a timing issue or the test framework selecting too soon?

Rich-Harris:
`clicknav` automatically calls `waitForNavigation`. For button clicks that involve network activity, we can deflake them by adding a `page.waitForNavigation` or `page.waitForLoadState('networkidle')` immediately after the click

dummdidumm:
Closing in favor of the other PRs that make this more robust

Issue #5176, opened 2022-06-07
Add warning for use of session store outside of a component


ghostdevv:
### Describe the problem

You are unable to use the `session` store outside of a component and currently you just get an error message along the lines of `Function called outside component initialisation`

### Describe the proposed solution

Talked in the discord about potentially adding something to svelte core to allow you to check whether you are in component initialisation or not and then libs can use that to throw more helpful error messages

### Importance

nice to have

rmunn:
This is probably moot now that https://github.com/sveltejs/kit/pull/5946 has removed the special `$session` store. (It can now be implemented quite easily by anyone who needs it, using `event.locals` and a `+layout.server.ts` file). So I'm closing the issue; if the decision to remove the session store gets reversed, we can reopen the issue if/when that happens.

Issue #1726, opened 2021-06-21
Built-in method of refreshing session from server periodically


rmunn:
In https://github.com/sveltejs/kit/issues/1165#issuecomment-825067344, Rich Harris wrote:

> I almost wonder if SvelteKit should have some built-in stuff for refreshing the session from the server periodically (including when you refocus a page, [swr](https://swr.vercel.app/) style). #46 included some thinking along those lines.

This seems like a good idea, and it also seems to be a different feature than #1165 or #46. So I'm opening this issue to track discussion of a session-refresh feature, so this idea doesn't get lost. (I don't have any specific ideas for what the feature should look like, I'm just opening this issue so there's somewhere to discuss the idea, and hoping that other people will have a better idea of what they would want in a session-refresh feature).

rmunn:
I've thought about this and the design that @Rich-Harris sketched out in https://github.com/sveltejs/kit/issues/46#issuecomment-714901886, and here's the design for session persistence and refreshing that I think would be best:

## Session persistence

* There's a `$session.persist()` method on the client-side session store
* `$session.persist()` has to be called explicitly, because there are plenty of use cases where session data would be updated in a way that shouldn't be persisted *yet*, and the page author wants to control the timing of session persistence
* If auto-persisting sessions is desired, it's as simple as subscribing to the session store with a function that calls `$session.persist()`.
* When `$session.persist()` is called, it sends a POST to `/__session` (that URL would be configurable in svelte.config.js)
    * I prefer `__session` (rather than `__session__` which @Rich-Harris originally proposed) for its similarity to `__layout.svelte` and `__error.svelte`
* The contents of the POST are a JSON representation of the session object, with anything non-serializable stripped out
    * In particular, functions would be stripped out by `$session.persist()`, so that Svelte-Kit users can add functions to their session stores if they need to, without having to worry about overriding `$session.persist()` if they do so
    * Dates would be serialized in the format returned by `Date.toJson` (e.g., '2021-06-15T12:34:56.789Z'), which should be obvious but is worth saying explicitly
* There is a hook in hooks.ts (or .js) called `persistSession()` which receives the POST contents (deserialized into a Javascript object)

## Session refreshing

* There would also be a `$session.refresh()` method available on the session store
* `$session.refresh()` would trigger a POST to `/__session_refresh` (URL configurable in svelte.config.js), passing it the same JSON that `persistSession()` would receive (e.g. functions stripped out before serializing)
[...]

PR #5946, opened 2022-08-16
Remove session

### LINKED ISSUE #4426: Navigating from prerendered page to non-prerendered page does not trigger hooks
### Describe the bug

When a site has a mix of prerendered and non-prerendered pages, if the first page a client lands on is a prerendered page, `getSession` will not be called when the client navigates to a non-prerendered page.

This means that if you have a site with a landing page (which is the same for all users) and personalized content (log in to save your todo lists), prerendering becomes a trap. Because if you use `getSession` to look up the users' data in the database and load their todo lists, that will break as soon as you decide to prerender your landing page.

### Reproduction

https://github.com/rmunn/prerender-and-hooks-repro

Run `pnpm run dev`, then go to `/` first, which is prerendered. Then click on the navigation link to `/userpage`, and notice that `getSession` is *not* called, so the user page would not be able to present its personalized content.

Navigating directly to `/userpage` works.

### Logs

```shell
Actual:
  Handle hook ran for URL /
  getSession was called for URL /

Expected:
  Handle hook ran for URL /
  getSession was called for URL /
  Handle hook ran for URL /userpage
  getSession was called for URL /userpage
```

### System Info

```shell
System:
    OS: Linux 5.4 Linux Mint 19.1 (Tessa)
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
    Memory: 6.51 GB / 62.58 GB
    Container: Yes
    Shell: 4.4.20 - /bin/bash
  Binaries:
    Node: 16.14.0 - /usr/local/bin/node
    Yarn: 1.22.17 - /usr/bin/yarn
    npm: 7.23.0 - ~/.local/bin/npm
  Browsers:
    Chrome: 74.0.3729.169
    Firefox: 96.0.3
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.33 
    @sveltejs/kit: next => 1.0.0-next.302 
    svelte: ^3.44.0 => 3.46.4
```

### Severity

### Additional Information

I can work around this by adding `router = false` to the landing page which I've marked as `prerender = true`. But the current behavior, where prerendering one page changes the behavior of a *different* page (whose `handle` and `getSession` hooks no longer run) seems wrong to me. The https://kit.svelte.dev/docs/hooks documentation is quite clear that hooks for prerendered pages run at build time, but I expected that that meant that `handle` and `getSession` hooks would only run for non-prerendered pages, not that they would never run at all if a prerendered page was the first page visited on the site.

So either this is a bug, or it's a case of unclear documentation.

Rich-Harris:
Implements #5883. This removes `event.session` from `load`, the `session` store, and `getSession`. It's easy to use `event.locals` in `handle` along with `+layout.server.js`/`+page.server.js` to provide equivalent functionality.

Closes #4426.

### Tests

### Changesets


benmccann (approved):

rmunn:
I've closed two issues as moot now that `session` is gone:

* #5176 
* #1726

A quick issue search on the word "session" turned up a lot of issues that mention that word, but only those two seemed to be specifically about the special `session` store that has just been removed. If this PR is reverted in the near future, we can reopen those issues.

MathiasWP:
> What about SPA? This was a really bad move

If you application is SPA only, then you should be able to achieve the same behaviour by creating your own store and calling it "session" 👍

Rich-Harris:
If you have access to a server, then you can return the equivalent of session data from a `+layout.server.js`. If you don't, then you couldn't return `session` anyway.

bjon:
@Rich-Harris How do I get the data from `+layout.server.js`, in `+layout.js`? `await parent()` is empty in `+layout.js`.

dummdidumm:
https://kit.svelte.dev/docs/load#input-properties-data

bjon:
@dummdidumm Thanks a lot! Didn't even know that data prop existed. 👍🏻

PR #5955, opened 2022-08-16
avoid double import of `$env/dynamic/[mode]` modules

### LINKED ISSUE #5941: Dynamic Public Environment Variables are undefined in +server.ts endpoints
### Describe the bug

Hello, I want to start off by thanking everyone involved for their hard work and clever ideas that have brought Svelte and Sveltekit to the points that the projects are at now 🥰

I updated to the latest version of Sveltekit today, and after going through all the migrations and fixing everything, I have one remaining problem in my project, that dynamic public environment variables are not available in +server.ts endpoints.

I made a simple example repository showing the issue. After you install and build it, you can see that the public dynamic is rendered correctly when you navigate to /

If you navigate to /static, the endpoint will return a public and a private static environment variable and display 'working foo'.

If you navigate to /dynamic, the endpoint will return a public and private dynamic environment variable, and should display 'working bar'. However, the public dynamic variable is undefined, so it displays 'undefined bar'.

### Reproduction

https://github.com/ssbm-oro/sveltekit-dynamic-public-bug

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.5
    CPU: (8) arm64 Apple M2
    Memory: 78.63 MB / 24.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.6.0 - /opt/homebrew/bin/node
    npm: 8.13.2 - /opt/homebrew/bin/npm
  Browsers:
    Firefox: 103.0.1
    Safari: 15.6
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.411 
    svelte: ^3.44.0 => 3.49.0 
    vite: ^3.0.0 => 3.0.8
```

### Severity

annoyance

### Additional Information

_No response_

Rich-Harris:
Attempt to fix #5941. No test because the bug doesn't happen inside the workspace

### Tests

### Changesets

### DIFF
--- a/.changeset/nasty-shoes-mate.md
+++ b/.changeset/nasty-shoes-mate.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Prevent double import of env modules
--- a/packages/kit/src/vite/dev/index.js
+++ b/packages/kit/src/vite/dev/index.js
@@ -286,27 +286,27 @@ export async function dev(vite, vite_config, svelte_config, illegal_imports) {
 		}
 	});
 
-	return async () => {
+	const runtime_base = runtime_directory.startsWith(process.cwd())
+		? `/${path.relative('.', runtime_directory)}`
+		: `/@fs${
+				// Windows/Linux separation - Windows starts with a drive letter, we need a / in front there
+				runtime_directory.startsWith('/') ? '' : '/'
+		  }${runtime_directory}`;
+
+	const { set_private_env } = await vite.ssrLoadModule(`${runtime_base}/env-private.js`);
+	const { set_public_env } = await vite.ssrLoadModule(`${runtime_base}/env-public.js`);
+
+	const env = get_env(vite_config.mode, svelte_config.kit.env.publicPrefix);
+	set_private_env(env.private);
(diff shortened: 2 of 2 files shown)


benmccann:
Will this prevent the user from being able to change the env files after server startup?

Rich-Harris:
Changing `.env` will cause the dev server to be restarted

benmccann (approved):
lgtm. although I haven't fully followed what's happening and why this fixes things. @bluwy and @dominikg might be understanding the situation more closely

bluwy (approved):
@benmccann it looks like the issue is that we're slightly tapping into Vite's internals to load `env-private.js` and `env-public.js`, where if Vite controls the entire import analysis flow, it would usually load them as `/node_modules/@sveltejs/kit/src/runtime/env-private.js`. Only imports outside of root would (usually?) use `/@fs`, which this PR changes to emulate this.

Ideally we would put these runtime files in `pkg.exports`, so we can `ssrLoadModule` them directly without resolving the path ourselves, but @Rich-Harris brought up that we want to keep this private API and intellisense-wise.

I think the solution now is fine, but if more issues appear later, it could be worth investigating a safer way.

PR #5972, opened 2022-08-17
[fix] manifest generation

### LINKED ISSUE #5910: Nested error pages do not render without a `+page.svelte`
### Describe the bug

After running migration, the old page `load` function that returned a `404` is transformed into an equivalent `+server.js` file. However, this only causes a `json` to be returned instead of rendering the nested error page.

I've tried returning a 404 from a `+page.js` load function but the result is the same as above.

Currently, I've only been successful by using a combination of throwing 404 from `+page.js` and adding an empty `+page.svelte` file.

Perhaps this requires a change of behaviour or a clarification in the docs.

See: https://kit.svelte.dev/docs/advanced-routing#rest-parameters
> ...the marx-brothers/+error.svelte file will not be rendered if you visit /marx-brothers/karl, because no route was matched. If you want to render the nested error page, you should create a route that matches any /marx-brothers/* request, and return a 404 from it:
> ```
> src/routes/
> ├ marx-brothers/
> | ├ [...path]/
> │ ├ chico/
> │ ├ harpo/
> │ ├ groucho/
> │ └ +error.svelte
> └ +error.svelte
> ```

### Reproduction

https://github.com/s3812497/sveltekit-nested-error-pages-bug

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.5
    CPU: (8) arm64 Apple M1 Pro
    Memory: 1.46 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
    npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm
  Browsers:
    Brave Browser: 104.1.42.95
    Chrome: 104.0.5112.79
    Firefox: 103.0.1
    Firefox Developer Edition: 103.0
    Safari: 15.6
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.408 
    svelte: ^3.44.0 => 3.49.0 
    vite: ^3.0.0 => 3.0.7
```

### Severity

annoyance

### Additional Information

_No response_

dummdidumm:
Render pages without a .svelte file in their proper layout/error files
Fixes #5910

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/sync/create_manifest_data/index.js
+++ b/packages/kit/src/core/sync/create_manifest_data/index.js
@@ -175,7 +175,7 @@ export default function create_manifest_data({
 				// This ensures that layouts and errors are set for pages that have no Svelte file
 				// and only redirect or throw an error, but are set to the Svelte file definition if it exists.
 				// This ensures the proper error page is used and rendered in the proper layout.
-				if (!(route.errors.length || route.layouts.length) || item.kind === 'component') {
+				if (item.kind === 'component' || route.layouts.length === 0) {
 					const { layouts, errors } = trace(
 						tree,
 						id,
--- a/packages/kit/src/core/sync/create_manifest_data/index.js
+++ b/packages/kit/src/core/sync/create_manifest_data/index.js
@@ -172,12 +172,22 @@ export default function create_manifest_data({
 			if (item.is_page) {
 				const route = /** @type {import('types').PageData} */ (route_map.get(id));
 
-				if (item.kind === 'component') {
-					route.leaf.component = project_relative;
-
-					const { layouts, errors } = trace(tree, id, item.uses_layout, project_relative);
+				// This ensures that layouts and errors are set for pages that have no Svelte file
+				// and only redirect or throw an error, but are set to the Svelte file definition if it exists.
+				// This ensures the proper error page is used and rendered in the proper layout.
+				if (!(route.errors.length || route.layouts.length) || item.kind === 'component') {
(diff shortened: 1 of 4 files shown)


packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -172,12 +172,22 @@ export default function create_manifest_data({
...
 				const route = /** @type {import('types').PageData} */ (route_map.get(id));
 
-				if (item.kind === 'component') {
-					route.leaf.component = project_relative;
-
-					const { layouts, errors } = trace(tree, id, item.uses_layout, project_relative);
+				// This ensures that layouts and errors are set for pages that have no Svelte file
+				// and only redirect or throw an error, but are set to the Svelte file definition if it exists.
+				// This ensures the proper error page is used and rendered in the proper layout.
+				if (!(route.errors.length || route.layouts.length) || item.kind === 'component') {
```
Rich-Harris, line 178: My brain can't make heads or tails of this condition — can we simplify it? Since `layouts.length` and `errors.length` will only ever be `0` at the same time, we only need to check one, I think?

```suggestion
				if (item.kind === 'component' || route.layouts.length === 0) {
```

dummdidumm: That's true. The root layout will always be added.

Rich-Harris (approved):

PR #5983, opened 2022-08-17
chore: give verbose names to form args "e"

josefaidt:

This update does not quite change or add any functionality, but aims at improving the readability of the `form.ts` file from the default create-svelte app. I noticed `e` was used as the event argument, and then later in the same handler as an error variable. To me this was a bit confusing to read and it does not appear to hold significant value with a shorter variable name. For newcomers this can be off-putting or confusing reading multiple variables with the same name in the same function (despite block scoping)

### Tests

### Changesets

### DIFF
--- a/.changeset/chilled-years-smash.md
+++ b/.changeset/chilled-years-smash.md
@@ -1,0 +1,5 @@
+---
+"create-svelte": patch
+---
+
+make variable names more descriptive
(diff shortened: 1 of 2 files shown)


Rich-Harris (approved):

Rich-Harris:
thanks 👍

PR #5987, opened 2022-08-17
check if data is serializable

Rich-Harris:
This throws an error if a `load` function in `+page.server.js` or `+layout.server.js` returns data that can't be serialized as JSON, so things like `/hello/` don't become `{}`.

Supersedes #5635.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -254,6 +254,7 @@ export async function respond(request, options, state) {
 									return {
 										// TODO return `uses`, so we can reuse server data effectively
 										data: await load_server_data({
+											dev: options.dev,
 											event,
 											node,
 											parent: async () => {
@@ -289,7 +290,7 @@ export async function respond(request, options, state) {
 								}
 							});
 
-							response = json({
+							return json({
 								type: 'data',
 								nodes: await Promise.all(promises)
 							});
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -290,7 +290,7 @@ export async function respond(request, options, state) {
 								}
 							});
 
-							return json({
+							response = json({
 								type: 'data',
 								nodes: await Promise.all(promises)
 							});
--- a/packages/kit/src/runtime/server/page/load_data.js
+++ b/packages/kit/src/runtime/server/page/load_data.js
@@ -3,12 +3,13 @@ import { LoadURL, PrerenderingURL } from '../../../utils/url.js';
 /**
  * Calls the user's `load` function.
  * @param {{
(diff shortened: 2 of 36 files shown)


packages/kit/src/runtime/server/page/load_data.js
```diff
@@ -79,3 +86,42 @@ async function unwrap_promises(object) {
...
+		if (Array.isArray(value)) {
+			value.forEach((child, i) => {
+				check_serializability(child, id, `${path}[${i}]`);
+			});
+			return;
+		}
+
+		// ...and objects
+		const tag = Object.prototype.toString.call(value);
+		if (tag === '[object Object]') {
```
gtm-nayan, line 118: This wouldn't work for custom classes. If they pass through they'll fail `instanceof` checks and their methods aren't iterated over by either `Object.keys` or `for..in`.

Rich-Harris, line 118: luckily i know where i can find a more robust check!

```suggestion
		// This simple check might potentially run into some weird edge cases
		// Refer to https://github.com/lodash/lodash/blob/2da024c3b4f9947a48517639de7560457cd4ec6c/isPlainObject.js?rgh-link-date=2022-07-20T12%3A48%3A07Z#L30
		// if that ever happens
		if (Object.getPrototypeOf(value) === Object.prototype) {
```

gtm-nayan: That one is essentially "if it's not a freshly constructed plain object then no bueno", 
I think it'll miss Object.defineProperties and other stuff, but if people are going to those lengths for the returned object, I wouldn't want to accomodate that.

gtm-nayan: Also, are we going to need these checks if #6008 is implemented?

Rich-Harris: No, in that case we'd have to rely on `devalue` throwing an error if stuff isn't serializable. Though it would be preferable to serialize the entire array in one go, which would make it tricker to track down the error... I haven't really thought about the best way to do that yet, but in any case #6008 probably won't happen for a while (feels like a post-1.0 thing)

packages/kit/src/runtime/server/index.js
```diff
@@ -289,7 +290,7 @@ export async function respond(request, options, state) {
 								}
 							});
 
-							response = json({
+							return json({
```
Rich-Harris, line 293: uh i think this change was made accidentally. would result in cookies not being set etc

```suggestion
							response = json({
```

2foo:
ObjectIds get blocked despite passing a `JSON.stringify({ a: new ObjectId() })`. Is this intended?

PR #5996, opened 2022-08-17
[chore] remove chokidar as dependency of kit

Conduitry:
Now that `svelte-kit package` is no longer a part of core SvelteKit, it no longer needs `chokidar` as a dependency.

### Tests

### Changesets

### DIFF
--- a/.changeset/twenty-numbers-breathe.md
+++ b/.changeset/twenty-numbers-breathe.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/kit": patch
+---
+
+[chore] remove chokidar as dependency of kit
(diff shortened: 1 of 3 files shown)


Conduitry:
The test failures look to be caused by flakiness on GitHub's side. The issue isn't in the action runner, but rather that the github.com URL that the build is trying to download the `sharp` binaries from is 500ing.

Rich-Harris:
Yeah, that happens periodically. Is there an alternative to vite-imagetools that doesn't use sharp? Would be great to use a JS-only dependency if a suitable one exists

Rich-Harris (approved):

PR #6028, opened 2022-08-18
[fix] Handle `throw error/redirect` in `+server.js`

### LINKED ISSUE #5993: Incorrect response `status code` when throwing error from `+server.ts`
### Describe the bug

When throwing an error from a `+server.ts` route, the `status code` in the `response header` is set to `500` instead of the status code specified in the `error()` helper. The nearest Error Boundary does render and the `error` property in the `$page` store does have the correct status code, however the response header doesn't match.

### **+server.ts**

```
import { error, redirect } from '@sveltejs/kit';
import type { RequestHandler } from './$types';

export const GET: RequestHandler = ({ params }) => {
	if (true) {
		throw error(404, 'Shortcode Not Found');
	}

	return redirect(301, 'https://kit.svelte.dev');
};
```

### **Response Header**

```
Status Code: 500 Internal Server Error
```

### **$page store**

```
{
  error: HttpError {
    name: 'HttpError',
    stack: undefined,
    status: 404,
    message: 'Shortcode Not Found'
  },
  params: { shortcode: 'test' },
  routeId: '[shortcode]',
  status: 500,
  url: URL {
    href: 'http://localhost:5173/test',
    origin: 'http://localhost:5173',
    protocol: 'http:',
    username: '',
    password: '',
    host: 'localhost:5173',
    hostname: 'localhost',
    port: '5173',
    pathname: '/test',
    search: '',
    searchParams: URLSearchParams {},
    hash: ''
  },
  data: {}
}
```

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 5.49 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.7.0 - /opt/homebrew/bin/node
    npm: 8.16.0 - /opt/homebrew/bin/npm
  Browsers:
    Brave Browser: 104.1.42.95
    Chrome: 104.0.5112.79
    Firefox Developer Edition: 104.0
    Safari: 15.6
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.417
    svelte: ^3.44.0 => 3.49.0
    vite: ^3.0.0 => 3.0.8
```

### Severity

### Additional Information

_No response_

dummdidumm:
Fixes #5993

This results in a (in my opinion correct) change in behavior: If you `throw` an `error`, you will no longer have the message rendered in the nearest `+error.svelte`, instead it's just blank text, which is what I'd expect from querying a regular endpoint.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/endpoint.js
+++ b/packages/kit/src/runtime/server/endpoint.js
@@ -1,3 +1,4 @@
+import { HttpError, Redirect } from '../../index/private.js';
 import { check_method_names, method_not_allowed } from './utils.js';
 
 /**
@@ -29,16 +30,29 @@ export async function render_endpoint(event, route) {
 		return method_not_allowed(mod, method);
 	}
 
-	const response = await handler(
-		/** @type {import('types').RequestEvent<Record<string, any>>} */ (event)
-	);
-
-	if (!(response instanceof Response)) {
-		return new Response(
-			`Invalid response from route ${event.url.pathname}: handler should return a Response object`,
-			{ status: 500 }
+	try {
+		const response = await handler(
+			/** @type {import('types').RequestEvent<Record<string, any>>} */ (event)
 		);
-	}
 
-	return response;
+		if (!(response instanceof Response)) {
+			return new Response(
+				`Invalid response from route ${event.url.pathname}: handler should return a Response object`,
+				{ status: 500 }
+			);
+		}
+
+		return response;
+	} catch (error) {
+		if (error instanceof HttpError) {
+			return new Response(error.message, { status: error.status });
+		} else if (error instanceof Redirect) {
+			return new Response(undefined, {
+				status: error.status,
+				headers: { Location: error.location }
+			});
+		} else {
+			throw error;
+		}
+	}
 }
--- a/.changeset/honest-parrots-compare.md
+++ b/.changeset/honest-parrots-compare.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Handle `throw error/redirect` in `+server.js`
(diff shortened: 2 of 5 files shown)


packages/kit/src/runtime/server/endpoint.js
```diff
@@ -29,16 +30,29 @@ export async function render_endpoint(event, route) {
...
+	} catch (error) {
+		if (error instanceof HttpError) {
+			return new Response(error.message, { status: error.status });
+		} else if (error instanceof Redirect) {
+			return new Response(undefined, {
+				status: error.status,
+				headers: { Location: error.location }
+			});
+		} else {
+			throw error;
```
dummdidumm, line 55: Should this be converted into a `Response`, too? If it isn't this means unexpected errors get a regular "page" treatment and they are rendered within possible `+error.svelte` pages which seems wrong.

Rich-Harris: I think so, yeah

Rich-Harris: though actually... i wonder if that should happen at the bottom of `respond`. We already have a `TODO` comment there asking this exact question.

that way we get all the `options.handle_error` and sanitization logic centralised

Rich-Harris: gah, ok that's not possible because we want to render an error page if user code errors in `handle` (at least, that's the decision we made at some point in the past).

i do feel like there's some opportunities for simplification around error handling generally. might noodle in a separate PR. will merge this in the meantime though, no point in delaying it

Rich-Harris (approved):

PR #6036, opened 2022-08-18
[docs] Change set-cookie to Set-Cookie

tbdrz:
Change `set-cookie` to `Set-Cookie`, according the mdn docs (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie)

### Tests

### Changesets

### DIFF
--- a/documentation/docs/03-routing.md
+++ b/documentation/docs/03-routing.md
@@ -164,7 +164,7 @@ export async function POST({ request, setHeaders, url }) {
 	}
 
 	setHeaders({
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Thank you, but headers are case-insensitive, and many systems normalize them to be lowercase (e.g. `req.headers` in a Node `http` server). There's nothing to be gained by making people hunt for the shift key, and in any case there are several other instances of `setHeaders` in the docs that all use lowercase

Issue #5994, opened 2022-08-17
Accumulate data into `export let data`


sserdyuk:
### Describe the bug

This is an "opinionated" take on the recent changes to routes and `load` function behavior. It's a big change, that I am not a fan of but can see the merit. Among other things it now requires more boilerplate code, which we all don't like to deal with. 

Here's the problem:  +page.svelte doesn't receive `data` unless there's +page.ts file with `load` function even though we might have `data` set by parent layouts. This forces one to create a bunch of +page.ts files with a the following boilerplate. Wouldn't be nicer to do that as a default for all pages, so we don't have to create +page.ts files unless we're actually doing something specific there? 

```
import type { PageLoad } from './$types';

export const load: PageLoad = async ({ parent }) => {
  return await parent();
};
```
 

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-f8ky2l?file=src/routes/+page.js

The +page.js file seems to be completely unnecessary but you can't remove it or +page receives no `data`.

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (4) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    Yarn: 1.22.10 - /bin/yarn
    npm: 7.17.0 - /bin/npm
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.64 
    @sveltejs/kit: next => 1.0.0-next.416 
    svelte: ^3.46.0 => 3.49.0 
    vite: ^3.0.0 => 3.0.8
```

### Severity

serious, but I can work around it

babichjacob:
This is less work now than it used to be.

You can [use `$page.data` to get data from ancestors and/or children](https://kit.svelte.dev/docs/load#shared-state) without writing a `load` function and thereby `+page.js` file at all. 

```svelte
<script>
  import { page } from "$app/stores";
</script>

<div>
{JSON.stringify($page.data)}
</div>
```

dummdidumm:
You can use the `$page.data` store which contains the accumulated data of the current page. Does this solve your use case?

sserdyuk:
@babichjacob Thank you for the idea. Unfortunately, `$page` doesn't pass the types from the parents.

benmccann:
[...]

PR #6041, opened 2022-08-18
[docs] note on parent and waterfalls

### LINKED ISSUE #6017: `$page.data` friction - allow parallel loads and "waterfall" data merges
### Describe the problem

Presumably, one of the main benefits to the new load architecture, specifically around the `await parent()` feature, is to allow all `load`s to run in parallel unless they explicitly don't need to. Also presumably, the most common source of "long" `load`s are network calls. If all of my loads are _blazingly fast_ -- as the kids like to say -- waterfalling isn't that big of a deal, but if I throw a few network calls in there, suddenly that whole "running in parallel" thing is pretty awesome.

Unfortunately, I'm finding that I often have to throw the "running in parallel" benefits out the window because I need to be able to deeply merge `data`. For example, if I have a `layout` key in data which itself is an object being listened to by the root `+layout.svelte` through `$page.data`:

```ts
// /routes/foo/+layout.ts
export const load: LayoutLoad = () => {
  return {
    layout: {
      showFooter: true,
    }
  }
} 

// /routes/foo/bar/+layout.ts
export const load: LayoutLoad = () => {
  const things = await some_stuff_we_dont_want_to_happen_serially();

  return {
    layout: {
      showPotato: true,
    }
  }
}

// /routes/+layout.svelte
<script lang="ts">
</script>

<Potato show={!!$page.data?.layout?.showPotato} />
<slot />
<Footer show={!!$page.data?.layout?.showFooter} />
```

See the footgun? `/routes/foo/bar/+layout.ts` will completely overwrite the `layout` key in `data` because it's shallow-ly merged. In order to fix it, we have to change it to:

```ts
// /routes/foo/bar/+layout.ts
export const load: LayoutLoad = async ({ parent }) => {
  const things = await some_stuff_we_dont_want_to_happen_serially();
  const { layout } = await parent();

  return {
    layout: {
      ...layout,
      showPotato: true,
    }
  }
}
```

But, ugh, now that load has to `await parent` when _all I want to do_ is deeply merge the objects! The load itself doesn't actually have any dependency upon the parent at all.

Thankfully, the solution is pretty simple. I can't take credit -- @Rich-Harris came up with it when we were whiteboarding in the maintainers' chat. He's smart like that. The solution is to allow data resolution to occur _after_ all `load`s have finished. Rendering doesn't start until all `load`s have resolved anyway, so there's no issue here. Essentially, we allow `load`s to run in parallel, but data resolution to run as a (partial or complete) waterfall afterwards (_if_ a developer needs it to):

```ts
// /routes/foo/bar/+layout.ts
export const load: LayoutLoad = () => {
  const things = await some_stuff_we_dont_want_to_happen_serially();

  // data, in this case, is the merged contents of all datas prior to this `load`, including any `load`s that also return this kind of function
  return (data) => ({
    layout: {
      ...data.layout,
      showPotato: true,
    }
  });
}
```

This allows the `load` to run in parallel, but for its returned _data_ to depend upon prior `load`s' data (which is returned, presumably, _after_ all of their time-consuming network activities have completed).

### Alternatives considered

Just get over it and deal with shallow-ly merged `$page.data` forever. Not a big fan. 😆 

### Importance

would make my life easier

### Additional Information

_No response_

dummdidumm:
Closes #6017

### Tests

### Changesets

### DIFF
--- a/documentation/docs/05-load.md
+++ b/documentation/docs/05-load.md
@@ -188,6 +188,25 @@ export async function load({ parent }) {
 
 In `+page.js` or `+layout.js` it will return data from `load` functions in parent `+layout.js` files. Implicitly, a missing `+layout.js` is treated as a `({ data }) => data` function, meaning that it will also return data from parent `+layout.server.js` files.
 
+Be careful not to introduce accidental waterfalls when using `await parent()`. If for example you only want to merge parent data into the returned output, call it _after_ fetching your other data.
+
+```diff
+/// file: src/routes/foo/+page.server.js
+// @filename: $types.d.ts
+export type PageLoad = import('@sveltejs/kit').Load<{}, null, { a: number, b: number }>;
+
+// @filename: index.js
+// ---cut---
+/** @type {import('./$types').PageLoad} */
+export async function load({ parent, fetch }) {
+-	const parentData = await parent();
+	const data = await fetch('./some-api');
++	const parentData = await parent();
+
+	return { ...data, ...parentData };
+}
+```
+
 #### setHeaders
 
 If you need to set headers for the response, you can do so using the `setHeaders` method. This is useful if you want the page to be cached, for example:
--- a/documentation/docs/05-load.md
+++ b/documentation/docs/05-load.md
@@ -203,7 +203,10 @@ export async function load({ parent, fetch }) {
 	const data = await fetch('./some-api');
 +	const parentData = await parent();
 
-	return { ...data, ...parentData };
+	return {
+		...data
+		meta: { ...parentData.meta, ...data.meta }
+	};
 }
 ```
 


documentation/docs/05-load.md
```diff
@@ -188,6 +188,25 @@ export async function load({ parent }) {
...
+
+// @filename: index.js
+// ---cut---
+/** @type {import('./$types').PageLoad} */
+export async function load({ parent, fetch }) {
+-	const parentData = await parent();
+	const data = await fetch('./some-api');
++	const parentData = await parent();
+
+	return { ...data, ...parentData };
```
Rich-Harris, line 206: This example will become a bit confusing if we implement #5994, this top-level merging will happen automatically (albeit with the precedence running in the opposite direction)

```suggestion
	return {
		...data
		meta: { ...parentData.meta, ...data.meta }
	};
```

Rich-Harris (approved):

PR #6050, opened 2022-08-18
Accumulate `data`

### LINKED ISSUE #5994: Accumulate data into `export let data`

Rich-Harris:
Closes #5994.

Today, if you have a page with a `load`...

```js
// src/routes/foo/+page.js
export function load() {
  return {
    pagedata: 'hello from page load'
  };
}
```

...then the corresponding `data` prop will contain `pagedata` and nothing else. If you have some data from a layout `load` function that you need in the page, you need to call `await parent()` in the page's `load`:

```diff
// src/routes/foo/+page.js
-export function load() {
+export async function load({ parent }) {
+  const { layoutdata } = await parent();
  return {
+    layoutdata,
    pagedata: 'hello from page load'
  };
}
```

As of this PR, that's no longer the case — layout data will automatically be inherited (with correct types!) so that you can reference `data.layoutdata` (or, more realistically, `data.user.name`) throughout your app without boilerplate.

### Tests

### Changesets


packages/kit/src/runtime/client/client.js
```diff
@@ -403,10 +403,10 @@ export function create_client({ target, base, trailing_slash }) {
 		let data = {};
 		let data_changed = false;
 		for (let i = 0; i < filtered.length; i += 1) {
-			Object.assign(data, filtered[i].data);
+			data = { ...data, ...filtered[i].data };
 			// Only set props if the node actually updated. This prevents needless rerenders.
 			if (!current.branch.some((node) => node === filtered[i])) {
-				result.props[`data_${i}`] = filtered[i].data;
+				result.props[`data_${i}`] = data;
```
dummdidumm, line 409: Is this logic still correct? If any parent updates, so does the data so we likely can no longer rely on "this node didn't change", but rather "everything up to and including this node didn't change"

packages/kit/src/core/sync/write_types.js
```diff
@@ -449,36 +449,43 @@ function process_node(ts, node, outdir, params, groups) {
...
+			node.shared,
+			`${parent_type} & ${server_data}`,
+			proxy
+		)}`;
+		load = `Kit.Load<${params}, ${server_data}, ${parent_type}, OutputData>`;
+	} else if (server_data === 'null') {
+		data = parent_type;
 	} else {
-		data = server_data;
+		data = `${parent_type} & ${server_data}`;
```
dummdidumm, line 470: This could give a type conflict if they set the same prop but to different types. It probably needs the same omit logic as the parent type function.

packages/kit/src/core/sync/write_types.js
```diff
@@ -449,36 +449,43 @@ function process_node(ts, node, outdir, params, groups) {
...
 	if (node.shared) {
 		const content = fs.readFileSync(node.shared, 'utf8');
 		const proxy = tweak_types(ts, content, shared_names);
 		if (proxy?.modified) {
 			written_proxies.push(write(`${outdir}/proxy${path.basename(node.shared)}`, proxy.code));
 		}
 
-		data = get_data_type(node.shared, 'load', server_data, proxy);
-		load = `Kit.Load<${params}, ${server_data}, ${get_parent_type('LayoutData')}, OutputData>`;
+		data = `${parent_type} & ${get_data_type(
```
dummdidumm, line 461: Same omit gotcha here

dummdidumm (commented):
Looks good! I left three things that probably need a change, also the docs should probably mention this

packages/kit/src/core/sync/write_types.js
Rich-Harris: good catch, added

packages/kit/src/runtime/client/client.js
Rich-Harris: oops, yep — nice save

Rich-Harris: https://github.com/sveltejs/kit/pull/6050/commits/f0823cb459f22d4c22549bb490680f0c6a165ade

dummdidumm (approved):

tsukhu:
Thank you for this update. I have deleted 3 page.ts files now :-)

rohanrajpal:
Love this!

Issue #5952, opened 2022-08-16
Redirects in `load` functions break when running client-side


WaltzingPenguin:
### Describe the bug

There appears to be no way to use the new redirect method and disable SSR globally.

### Reproduction

1. Create a new skeleton project
2. Create `src/routes/page.ts` with the contents:
```ts
import { redirect } from "@sveltejs/kit";

export function load() {
  throw redirect(307, "/about")
}
```
3. Create `src/hooks.ts` with the contents:
```ts
import type { Handle } from '@sveltejs/kit'

export const handle: Handle = async ({ event, resolve }) => {
  return resolve(event, { ssr: false })
}
```
4. Navigating to root now results in an error page, with the redirect class passed to it, instead of being redirected to "/about"

Repo: https://github.com/WaltzingPenguin/sveltekit-redirect-hooks

### System Info

```shell
System:
    OS: Windows 10 10.0.19043
    CPU: (12) x64 AMD Ryzen 5 1600 Six-Core Processor
    Memory: 7.50 GB / 15.95 GB
  Binaries:
    Node: 16.15.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.10 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 7.20.3 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 104.0.5112.81
    Edge: Spartan (44.19041.1266.0), Chromium (104.0.1293.54)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.64
    @sveltejs/kit: next => 1.0.0-next.413
    svelte: ^3.44.0 => 3.49.0
    vite: ^3.0.0 => 3.0.8
```

### Severity

blocking an upgrade

denovodavid:
As a simple workaround I've done the following:

```ts
const location = '/login';
if (browser) return await goto(location);
else throw redirect(302, location);
```

This worked for me in browser and SSR, but yes ideally `redirect()` should work client side.

fev4:
I'm experiencing the same situation without disabling SSR:

```ts
// file: src/routes/+layout.ts

import {
  envelopeGuard,
  envelopeGuardRedirect,
} from '$lib/utils/guards';
import { accounts } from '$lib/stores/accountsStore';
import { get } from 'svelte/store';

export async function load({ url }) {
  const { pathname } = url;
  const accountsArray = get(accounts);
[...]

PR #6057, opened 2022-08-18
fix: prevent optimizeDeps for @sveltejs/kit

dominikg:
part of solution for https://github.com/sveltejs/kit/issues/5952

see https://github.com/sveltejs/kit/issues/5952#issuecomment-1218844057, where some of the ids mentioned are in `.vite` and the `?v=hash` query is also a sign

### Tests

### Changesets

### DIFF
--- a/.changeset/famous-boxes-tap.md
+++ b/.changeset/famous-boxes-tap.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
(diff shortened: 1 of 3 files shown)


Rich-Harris:
It's definitely _part_ of the fix for #5952, but we need to figure out the query parameter stuff as well

dominikg:
is there another place where optimizeDeps.exclude needs to be set? Otherwise let's merge this and figure the other part separately

bluwy (approved):

PR #6084, opened 2022-08-19
make create-svelte tests a bit less look-at-me

Rich-Harris:
the `create-svelte` tests are absurdly verbose, and it makes it very hard to find actual problems when reviewing CI failures


dummdidumm (approved):
A look at the CI output shows this is way better 👍

PR #6100, opened 2022-08-19
keep browser stores in always-accessible singleton

### LINKED ISSUE #6043: Ability to access `$app/stores` at any time in the browser
### Describe the problem

There is sometimes code that I know will only ever be run on the browser (for example, if it's only called from an event handler), and having to jump through the same hoops as I do on the server in order to associate that with the appropriate (and only) instances of the stores is annoying.

Store values on the browser should live in a singleton module (or something similar) rather than in the Svelte context for the root component. `getStores()` should be able to be called at any time, not just during component initialization. If it wouldn't cause too much of a divergence between browser and server code, we could also remove the transparent calling of `getStores()` during the autosubscription to any of the individual stores.

### Alternatives considered

Right now, I'm using a fairly ugly workaround where I serialize the values I need to a `data-foo` attribute in my root layout, which I then read and deserialize elsewhere.

Another option presumably would be to have an `onMount` call in my root layout that sets the store(s) as a global on `window`.

### Importance

would make my life easier

### Additional Information

_No response_

Conduitry:
Resolves #6043.

This lets you use `$app/stores` everywhere on the browser, not just where you would be allowed to on the server. We don't need to worry about mixing data for different requests like we do on the server - there's only ever one set of stores at once - so we can be a lot more relaxed about this.

### Tests

### Changesets


dummdidumm (approved):
Love this!

PR #6105, opened 2022-08-19
use enumerable: false for temporary no-longer-exists getters

Conduitry:
There are a few objects in the codebase with getters that give more helpful errors messages that describe what features have replaced these removed/renamed features. However, these getters are also triggered when doing an object `{ ...spread }`, which isn't very helpful. This updates these to use `Object.definedProperties` so that we may specify `enumerable: false` on these getters.

### Tests

### Changesets

### DIFF
--- a/.changeset/ninety-walls-heal.md
+++ b/.changeset/ninety-walls-heal.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+use `enumerable: false` on "[x] no longer exists" getters so that they are triggered by spreading
--- a/.changeset/ninety-walls-heal.md
+++ b/.changeset/ninety-walls-heal.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-use `enumerable: false` on "[x] no longer exists" getters so that they are triggered by spreading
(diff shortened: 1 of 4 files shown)


benmccann (approved):

benmccann:
this is to provide a `load` wrapper with a `fetch` that provides a base URL? that sounds handy and like something I would use and I've seen other people ask for. what do you think about adding that to the FAQ?

Rich-Harris:
@benmccann for `fetch` specifically, the intent was to let server-only `load` functions use the same `fetch` as shared `load` functions, so that specific case would no longer be necessary.

There are other times it would be useful to do this, but right now we're stymied by `url` being a getter: https://github.com/sveltejs/kit/blob/b335f38a35acb3be2408f23cde27eb4a34e1ffda/packages/kit/src/runtime/client/client.js#L536-L539

In #6056 I plan to replace that with a `TrackedURL` object that only triggers the dependency tracking on accessing things like `pathname` and `toString` (not `origin` or `hostname`, because those can't change), instead of `url` itself.

(`parent` is too, but it actually doesn't need to be — we only need to care about invocation, not access.)

There's also `event.clientAddress`, which I think we'll need to change to `event.getClientAddress()` so that we can promote patterns that use object spread (and also use it ourselves).

PR #6117, opened 2022-08-20
override prompts with command line arguments

navtoj:
This adds support for automatically answering the interactive prompts in environments where manual entry is not possible; e.g. scripts, batch runs, etc.

### Before
`npm create svelte@latest project_directory`

### After
`npm create svelte@latest project_directory --template=skeleton --types=typescript --eslint=true --prettier=true --playwright=true --overwrite=true`

> ## All arguments are optional.

- **--template=** `default`, `libskeleton`, or `skeleton`
Screen Shot 2022-08-20 at 5 39 39 AM

- **--types=** `checkjs` or `typescript`
Screen Shot 2022-08-20 at 5 39 42 AM

- **--eslint=** `true` or `false`
Screen Shot 2022-08-20 at 5 39 45 AM

- **--prettier=** `true` or `false`
Screen Shot 2022-08-20 at 5 39 48 AM

- **--playwright=** `true` or `false`
Screen Shot 2022-08-20 at 5 39 51 AM

- **--overwrite=** `true` or `false`
Screen Shot 2022-08-20 at 5 44 22 AM

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -24,6 +24,7 @@ async function main() {
 	console.log(disclaimer);
 
 	let cwd = process.argv[2] || '.';
+	if (cwd.startsWith('--')) cwd = '.';
 
 	/**
 	 * @typedef {object} params
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -1,8 +1,10 @@
 #!/usr/bin/env node
 import fs from 'fs';
-import path from 'path';
 import { bold, cyan, gray, green, red } from 'kleur/colors';
+import path from 'path';
 import prompts from 'prompts';
+import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
 import { create } from './index.js';
 import { dist } from './utils.js';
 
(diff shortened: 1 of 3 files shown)


Conduitry:
Thank you for the PR. This has come up several times before, though, and is something we're not comfortable with for a couple reasons. If someone is running the CLI, they're likely going to always be pulling down the latest version, and we may occasionally add additional questions to the script. If they don't specify an option for one of them, what happens? Does the script suddenly become interactive again? If the user specifies at least one option, do we make assumptions about all of the other ones based on some default? If we reorganize some of the prompts, what do we do with CLI args that are now invalid? Any path we take here seems suboptimal.

As a compromise, the `create-svelte` package also exposes a programmatic API, which _will_ respect semver (once SvelteKit is in general release). Any new prompts that must be answered will be a major version bump. Anything that changes the output in a way that's not a clear backwards-compatible improvement will also be a major version bump. If you need automated access to bootstrapping new projects, you can write a tiny script that calls this API and depends on `create-svelte` like it would any other dependency, with all the benefits of semver. And the majority of people - those just calling the CLI - will continue to get the latest one (no matter what its semver major version number is) and can just response to the prompts.

JReinhold:
For anyone landing here, the Storybook team has released a community solution to this called [`create-svelte-with-args`](https://github.com/storybookjs/create-svelte-with-args). It is a small CLI wrapper around `create-svelte` that allows you to create SvelteKit projects without the interactive prompts.

@Conduitry I've tried to be very explicit in the documentation stating that `create-svelte-with-args` is not supported nor endorsed by the Svelte core team. The last thing we want is for this to become a burden for you as maintainers because you've stated some very valid arguments above as to why you don't want to support this officially. Let me know if you have any thoughts on this.

PR #6155, opened 2022-08-22
[fix] render error page without root layout if that failed to load

### LINKED ISSUE #3068: Thrown exceptions from __layout.svelte load result in plain text error responses

### LINKED ISSUE #4815: Returning non-200 status from `load` in `__layout.svelte` causes a `Uncaught (in Promise) Error`

dummdidumm:
Fixes #4815
Fixes #3068

### Tests

### Changesets

### DIFF
--- a/.changeset/silver-buses-suffer.md
+++ b/.changeset/silver-buses-suffer.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] render error page without root layout if that failed to load
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -818,14 +818,19 @@ export function create_client({ target, base, trailing_slash }) {
 		/** @type {Record<string, string>} */
 		const params = {}; // error page does not have params
 
-		const root_layout = await load_node({
-			node: await default_layout,
-			url,
-			params,
-			routeId,
-			parent: () => Promise.resolve({}),
-			server_data: null // TODO!!!!!
-		});
+		let root_layout;
+		try {
+			root_layout = await load_node({
+				node: await default_layout,
+				url,
+				params,
+				routeId,
+				parent: () => Promise.resolve({}),
+				server_data: null // TODO!!!!!
+			});
+		} catch (e) {
+			// Root layout threw an error, we have to ignore it at this point
+		}
 
 		const root_error = {
 			node: await default_error,
--- a/packages/kit/src/runtime/server/page/index.js
+++ b/packages/kit/src/runtime/server/page/index.js
@@ -242,42 +242,52 @@ export async function render_page(event, route, options, state, resolve_opts) {
 
 					const status = error instanceof HttpError ? error.status : 500;
 
-					while (i--) {
-						if (route.errors[i]) {
-							const index = /** @type {number} */ (route.errors[i]);
-							const node = await options.manifest._.nodes[index]();
+					/**
+					 * @param {number} idx
+					 * @param {Array<Loaded>} branch
+					 */
+					const render_error = async (idx, branch) => {
+						const index = /** @type {number} */ (route.errors[idx]);
+						const node = await options.manifest._.nodes[index]();
 
-							let j = i;
-							while (!branch[j]) j -= 1;
+						return await render_response({
+							event,
+							options,
+							state,
+							resolve_opts,
+							page_config: { router: true, hydrate: true },
+							status,
+							error,
+							branch: branch.concat({
+								node,
+								data: null,
+								server_data: null
+							}),
+							fetched,
+							cookies,
+							validation_errors: undefined
+						});
+					};
 
-							return await render_response({
-								event,
-								options,
-								state,
-								resolve_opts,
-								page_config: { router: true, hydrate: true },
-								status,
-								error,
-								branch: compact(branch.slice(0, j + 1)).concat({
-									node,
-									data: null,
-									server_data: null
-								}),
-								fetched,
-								cookies,
-								validation_errors: undefined
-							});
+					if (i === 0) {
+						// if the error is in the root layout, we render the error page only
+						return await render_error(0, []);
+					} else {
+						while (i--) {
+							if (route.errors[i]) {
+								let j = i;
+								while (!branch[j]) j -= 1;
+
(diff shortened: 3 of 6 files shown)


Rich-Harris:
I definitely don't think we should introduce this behaviour — it's effectively non-deterministic.

With the [new layouts proposal](https://github.com/sveltejs/kit/discussions/6124), it would be straightforward to have the fallback error page render inside a blank layout (or a layout that is guaranteed not to error, because it doesn't have an accompanying `load`).

That leaves the question of what to do if the root layout _does_ error. Provided we've given app developers the tools to avoid that scenario (which #6124 does), then a plain 500 response seems perfectly acceptable during SSR. If it happens on the client, the path of least resistance might be to simply reload the page.

dummdidumm:
How is this non-determenistic? If the root layout fails, it's not used - this is equivalent to a blank layout with the difference that the user doesn't have to create a group just to guarantee that his root layout never errors. If you talk about non-determinism on the client due to the try-catch - that could be adjusted so that the same logic as on the server applies: if we detect that the root layout errored, not use it, else use it.

Rich-Harris:
It's non-deterministic because failures are non-deterministic. As an app author I shouldn't have to design my error page such that it will work both inside and outside the layout depending on factors outside my control

dummdidumm:
But how likely is it that you designed your error page that way? It could only fail if it references `$page.data` in a way thinking the data from the layout is present. I argue this is less likely than being surprised by "why do I get a blank error message" followed by "ugh, I need to create an empty layout and therefore a needless group just for that?"

Rich-Harris:
It's very unlikely, and that's the problem!

dummdidumm:
It's unlikely that the root layout errors. It's even more unlikely that someone uses things from the root layout in their root error page. It's very likely that someone wants to show something better than blank text if the root layout fails. Therefore everyone who wants that is forced to use a `(group)` (under the new proposal) to get around that. I don't think it's good that you are forced to create a blank layout and a otherwise unnecessary `(group)` just to be safe when rendering the error page - this feels like "losing" one layout to safety.

Rich-Harris:
Totally agree that there should be something better than blank text, but what that thing _is_ absolutely has to be predictable. There will always be cases where we need to bail completely (even if layout `load` succeeds, the error page itself could reference `{data.nonexistent.property}`; if it fails, then even `data.existent.property` would blow up), and so if a plain text response is unacceptable then _that's_ the problem we need to tackle.

One possible idea: an optional `src/error.html` that can contain things like `%sveltekit.error.message%`. If unspecified, falls back to a default error page that's slightly nicer than the one we have now. Unlike `+error.svelte`, we can make a cast-iron guarantee that this will render successfully.

dummdidumm:
People have expressed the desire previously to have a error page break out of the complete layout stack (including the root layout). Can we satisfy both needs by introducing a special `+error@<something>.svelte` which means "this error page is rendered without any layout"? `error.html` would be just that but you couldn't reuse your Svelte components there.

Rich-Harris:
I think it's better to use the existing primitives. My reaction to 'but then you'll need to put everything inside a `(main)` group' is 'so what?' In practice it doesn't make it any harder to navigation around the codebase — it _feels_ superfluous in the case where you're only using the blank layout for error pages, but it avoids all these problems...

* what if I want to share some CSS between my error page and the rest of the app? it's easier to do that in a root layout than to duplicate it in my error page
* what if I want the analytics code in my root layout `afterNavigate` function to keep working as the user navigates to/from/between errors?
* what if I later discover another use for my blank layout?

...and that's _before_ we try to answer the design question (what is the `<something>`?), let alone the documentation and maintenance burden that comes from creating new scenario-solving primitives. 

`error.html` is solving a different problem, and I think it's worth considering on its own merits (but probably in a separate issue).

PR #6165, opened 2022-08-22
clarify empty env vars

### LINKED ISSUE #6160: Unable to use undeclared static environment variables
### Describe the bug

I try to import an environment variable from `$env/static/public` not present in any `.env` file or environment and get an error. I do not know how the internals work, but the exported variables seems to be collected only from the environment and not from usages.

I filed this issue because Svelte Kits behaviour differs from Vite's `loadEnv` function, which allows me to use unknown variable names.

### Reproduction

https://github.com/inta/sveltejs-kit-template-default-3djxav

Stackblitz does not load the `.env` file (but I include it here anyways):
https://stackblitz.com/edit/sveltejs-kit-template-default-3djxav?terminal=dev

### Logs

```shell
The error message differs between my local machine:

500
ambiguous indirect export: PUBLIC_NOT_WORKING


and Stackblitz:

500
The requested module '/@id/__x00__$env/static/public' does not provide an export named 'PUBLIC_NOT_WORKING'
SyntaxError: The requested module '/@id/__x00__$env/static/public' does not provide an export named 'PUBLIC_NOT_WORKING'
```

### System Info

```shell
System:
    OS: Linux 5.18 Debian GNU/Linux bookworm/sid
    CPU: (16) x64 AMD Ryzen 7 PRO 4750U with Radeon Graphics
    Memory: 13.89 GB / 30.59 GB
    Container: Yes
    Shell: 3.5.0 - /usr/bin/fish
  Binaries:
    Node: 18.7.0 - /usr/bin/node
    npm: 8.13.2 - ~/.npm/global-modules/bin/npm
  Browsers:
    Chrome: 104.0.5112.101
    Chromium: 104.0.5112.101
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.428 
    svelte: ^3.46.0 => 3.49.0 
    vite: ^3.0.8 => 3.0.9
```

### Severity

annoyance

### Additional Information

Some background info:

I was implementing feature flags for some components and wanted them to always fall back to a default value. If missing variables lead to an error, I always have to make sure that every environment is up to date and has all new feature flags before introducing them in the codebase, otherwise the build pipeline would be broken.

Rich-Harris:
closes #6160

### DIFF
--- a/packages/kit/scripts/special-types/$env+static+private.md
+++ b/packages/kit/scripts/special-types/$env+static+private.md
@@ -5,3 +5,15 @@ _Unlike_ [`$env/dynamic/private`](https://kit.svelte.dev/docs/modules#$env-dynam
 ```ts
 import { API_KEY } from '$env/static/private';
 ```
+
+Note that all environment variables referenced in your code should be declared (for example in an `.env` file), even if they don't have a value until the app is deployed:
+
+```
+MY_FEATURE_FLAG=""
+```
+
+You can override `.env` values from the command line like so:
+
+```bash
+MY_FEATURE_FLAG="enabled" npm run dev
+```


tcc-sejohnson:
Should we add the same (or a "see here") text to the static public file?

Rich-Harris:
I was counting on "Similar to `$env/static/private`" being enough

tcc-sejohnson:
Seems like we should be taking a different tact for documenting these:

# `$env`

Stuff about all env vars

## `$env/static/*`

Stuff about static env vars

## `$env/dynamic/*

Stuff about dynamic env vars

## $env/*/public

Stuff about public env vars

## $env/*/private

Stuff about private env vars

There would be no "see here because they're the same" and much more logical/less repetitive grouping of concepts...

Rich-Harris:
Yeah, you're not wrong, though that would involve a much more comprehensive update to how this part of the documentation is automatically generated from the types

PR #6174, opened 2022-08-23
Implement new layouts system

### LINKED ISSUE #2154: Layout Stack not Fully Reset for src/routes/__layout.svelte

### LINKED ISSUE #4940: __error not correctly displayed when default layout inherits from a named layout
### Describe the bug

Let us assume such a file structure
``` hash
-routes
  └login
      └index@init.svelte
    |-__error.svelte
    |-__layout-init.svelte
    |-__layout@init.svelte
    └index.svelte
```
`__layout-init.svelte` used to load initialization content such as shared styles and external js files for whole site, login page use just `__layout-init.svelte` cuz it's a additional page that don't need `__layout@init.svelte`, everything is ok so far, but, the `__error.svelte` just inherit from `__layout@init.svelte` and lost `__layout-init.svelte` that means error page can't load initialization content ( which should inherit from -> `__layout@init.svelte` -> `__layout-init.svelte`)

I think it's a bug or design to?

### Reproduction

[here is online demo](https://stackblitz.com/edit/sveltejs-kit-template-default-fmnrme?file=src/routes/index.svelte)

### Logs

_No response_

### System Info

```shell
Browsers:
    Chrome: 101.0.4951.64
    Edge: 101.0.1210.47
    Firefox: 100.0
    Safari: 15.4
  npmPackages:
    @sveltejs/adapter-static: ^1.0.0-next.30 => 1.0.0-next.30 
    @sveltejs/kit: next => 1.0.0-next.326 
    svelte: ^3.44.0 => 3.48.0
```

### Severity

### Additional Information

_No response_

### LINKED ISSUE #5311: "magic" layout for resetting
### Describe the problem

Ever since named layouts it's no longer possible to do `__layout.reset.svelte`.
A workaround is to define a top level `__layout-reset.svelte` with just an empty slot, then apply this reset as a named slot: `__layout@reset.svelte`

Have `reset` be a _magic_ named layout that does exactly the same without having to define this file.

Of course this would break for people that use the reset with something else but an empty slot, maybe SvelteKit can check if it is defined and if not fall back to an empty slot ?

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

### LINKED ISSUE #5763: Per-route hook or middleware
### Describe the problem

Occasionally, I have a need to load some data in an endpoint and then re-use some of that data in a child endpoint. There's currently no way to pass the parent data down to the child on the server-side other than indirectly via the rendering system, and in some cases that data shouldn't go to the client at all and / or would add unnecessary bloat to the page payload.

To paint a more concrete picture of a scenario, imagine a forum system ...

`/routes/topic/[id]/__layout.svelte` Renders the topic title, author, summary, date, reply count etc...
`/routes/topic/[id]/index.ts` Is the endpoint for the topic data
`/routes/topic/[id]/[slug]/index.svelte` Renders the paginated posts in the topic
`/routes/topic/[id]/[slug]/index.tx` Is the endpoint for the post data, it uses the page number from the url + a slice of the post references on the topic to fetch

How to re-use the data from the parent endpoint, in the child endpoint, without passing it through the page (which doesn't need it).

What I'd like to have is the ability to hook into the request at any level, to pass data down. Exactly like the current hooks, which only runs at the root app level. Or maybe it could work like the new `load` system is going to, where a child can await a parent ... but on the server side for endpoints (?).

### Alternatives considered

Right now I get around it by using a sequenced handler in the app-level hooks, to check for a particular `routeId` prefix, do the data fetch there, put it into `locals` and then have access to it in both places.

But it feels odd that I can't use the routing system to do something so routing-related, and I didn't see anything in the new routing proposal that would address it.

### Importance

nice to have

### Additional Information

It's not possible to add anything to locals from the parent endpoint AFAICT (I guess 'cause it's a child request at that point?)

It's a fairly normal NoSQL approach to embed references to a list of things in a parent as it saves on expensive queries (esp. with offsets when using paginated data).

Avoiding unnecessary reads of the parent data is also beneficial, it would only be 1 but it's additional cost when charged per-read and adds to overall request latency.

This kind of thing has been trivial to do when writing server APIs in Go, but hasn't translated well to a SvelteKit powered API.

Running and accessing a caching server on Google Cloud from Cloud Run is surprisingly complicated and expensive. Per-instance caching would be problematic on an auto-scaling platform (for cache invalidation) so reading per-request is the best compromise.

Related to: https://github.com/sveltejs/kit/issues/3881

### LINKED ISSUE #6196: Reuse default +layout.js and +layout.server.js in named layout.
### Describe the problem

If I want to have a second root layout, just a different "layout" (`+layout.svelte`) I should copy-paste `+layout.js` and `+layout.server.js`. 
In old routing, schema duplicate was only inside `__layout.js` `load` function. Even if I move `load` function to a separate file, I will duplicate two extra files for each named layout. 

To reduce boilerplate, the named layout should have a way to reuse the default layout client and server functions. 

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

Rich-Harris:
# How to migration from named layouts

Named layouts have been removed in favor of `(groups)`. See the documentation for an introduction: https://kit.svelte.dev/docs/advanced-routing#advanced-layouts

How to migrate highly depends on your use case:

### Use case 1: Rewinding to an earlier layout / "breaking out"

If you used named layouts to rewind the layout inheritance chain to a layout higher up the tree, you can reference the segment directly now:

```diff
|- marketing
-  |- +layout-marketing.svelte
-  |- +layout@marketing.svelte
+  |- +layout.svelte
   |- pricing
     |- +layout.svelte
     |- business
       |- +layout@marketing.svelte // reference the "marketing" segment (i.e. the folder with than name) in your routes
```

### Use case 2: Rewinding to a blank root layout

If you had `+layout-blank.svelte` in your root to reset to a blank layout, you need to adjust your actual root layout to be blank and move that functionality into another layout further below in the tree, either using an if-statement in your root layout or using the `(group)` feature:

```diff
- |- +layout-blank.svelte
  |- +layout.svelte // make this the blank layout
+ |- (root)
+   |- +layout.svelte // your old root layout.svelte
    |- marketing
      |- +layout.svelte
      |- pricing
        |- +layout@.svelte // @ without a string means "reset to the root layout"
```

### Use case 3: Different layouts for routes at the same nesting level

If you used named layouts to specify "these 2 pages get layout X and these other 2 next to them get layout Y", use groups instead.

Before:
```
|- +layout-authed.svelte
|- +layout-unauthed.svelte
|- admin
  |- +page@authed.svelte
|- user
  |- +page@authed.svelte
|- about
  |- +page@unauthed.svelte
|- home
  |- +page@unauthed.svelte
```
After
```
|- (authed)
  |- +layout.svelte
  |- admin
    |- +page.svelte
  |- user
    |- +page.svelte
|- (unauthed)
  |- +layout.svelte
  |- about
    |- +page.svelte
  |- home
    |- +page.svelte
```

### Use case 4: arbitrary inheritance chains

If you previously used named layouts to enhance other named layouts, which also enhance layouts, to mix and match them, then that's no longer possible this way. In cases like this - which from reading the layout chains probably got confusing anyway - we advice to use composition and/or if-statements instead. The new documentation has a section with an example for this as inspiration.

# PR description

#6124 

Closes #6196 (using `(groups)` and/or composition)
Closes #5763 (root layout guaranteed to always exist + `await parent()`)
Closes #5311 (`+page@.svelte`)
Closes #4940 (no longer possible to get into this situation)
Closes #2154 (only a single root layout now)

* [x] documentation
* [x] prevent conflicts
* [x] print useful error when encountering a `+layout-foo.svelte`
* [x] simplify the code that currently has to deal with named layouts
* [x] update `write_types`
* [x] don't reference root layout/error in `dictionary` in `client-manifest.js`
* [x] get it to work in windows

### Tests

### Changesets


dummdidumm (approved):
Looks great! I did some small tweaks but didn't find anything that's wrong.

One thing I noticed is that the `RouteData` interface has `leaf` and `page`, and they both are (un)set at the same time. I would prefer to merge them into one property `page`. Since I want to start on the `App.PageData` PR and therefore need the `write_types` changes, I'll merge either way and we can do that in a follow-up PR, if you agree with my thinking.

packages/kit/src/runtime/client/ambient.d.ts
```diff
@@ -7,11 +7,12 @@ declare module '__GENERATED__/client-manifest.js' {
...
 	/**
-	 * A map of `[routeId: string]: [errors, layouts, page]` tuples, which
+	 * A map of `[routeId: string]: [leaf, layouts, errors]` tuples, which
 	 * is parsed into an array of routes on startup. The numbers refer to the
-	 * indices in `nodes`.
+	 * indices in `nodes`. The route layout and error nodes are not referenced,
+	 * they are always number 0 and 1 and always apply.
 	 */
-	export const dictionary: Record<string, [number[], number[], number]>;
+	export const dictionary: Record<string, [number, number[]?, number[]?]>;
```
ignatiusmb, line 15: For clarity, I think we should use named tuples here

```suggestion
	export const dictionary: Record<string, [leaf: number, layouts?: number[], errors?: number[]]>;
```

packages/kit/src/runtime/client/parse.js
```diff
@@ -2,25 +2,30 @@ import { exec, parse_route_id } from '../../utils/routing.js';
 
 /**
  * @param {import('types').CSRPageNodeLoader[]} nodes
- * @param {Record<string, [number[], number[], number, 1?]>} dictionary
+ * @param {Record<string, [number, number[]?, number[]?, 1?]>} dictionary
```
ignatiusmb, line 5: From above, we could then reuse it here (looks a bit awkward) and it would give more informative hover types / autocompletion. The `1?` at the end would be removed, though it doesn't seem to be used anymore?

```suggestion
 * @param {typeof import('__GENERATED__/client-manifest.js').dictionary} dictionary
```

packages/kit/src/utils/array.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Removes falsy values from an array.
+ *
+ * @function
+ * @template T
+ * @param {Array<T>} arr
+ * @returns {Array<NonNullable<T>>}
```
ignatiusmb, line 7: We can simplify this a bit since the value is inferred from the return statement

```suggestion
 * @template T
 * @param {Array<T>} arr
```

packages/kit/src/utils/routing.js
```diff
@@ -79,6 +88,13 @@ export function parse_route_id(id) {
 	return { pattern, names, types };
 }
 
+/**
+ * @param {string} segment
+ */
```
ignatiusmb, line 93: Shouldn't we collapse this?
```suggestion
/** @param {string} segment */
```

packages/kit/src/utils/array.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Removes falsy values from an array.
```
dominikg, line 2:
```suggestion
 * Removes nullish values from an array.
```

.changeset/many-forks-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] implement new layout system
```
dominikg, line 5: Do you plan to update the PR description or add a link to the docs explaining the change? Users combing changelogs to see what they have to update would appreciate

packages/kit/src/utils/routing.js
dummdidumm: For types on variables I think that's prefereable, but for functions I'd like to keep it that way.

.changeset/many-forks-sniff.md
dummdidumm: Yes I will write up some migration comments in the PR description, which should probably noted in the description
```suggestion
[breaking] implement new layout system (see the PR for migration instructions)
```

packages/kit/src/runtime/client/parse.js
dummdidumm: That's a good idea, it connects the places where it's used, which is always great for navigating the code base. The 1 is obsolete, so another plus for connecting them.

documentation/docs/04-advanced-routing.md
```diff
@@ -126,75 +126,66 @@ assert.equal(
...
-│ │ └ three/+page.svelte
-│ └ +layout.svelte
-├ +layout.svelte
--└ +layout-root.svelte
-+└ +layout-root@default.svelte
-```
+#### When to use layout groups
 
-> `default` is a reserved name — in other words, you can't have a `+layout-default.svelte` file.
+Not all use cases are suited for layout grouping, nor should you feel compelled to use them. It might be that your use case would result in complex `(group)` nesting, or that you don't want to introduce a `(group)` for a single outlier. It's perfectly fine to use other means such as composition (reusable `load` functions or Svelte components) or if-statements to achieve what you want.
```
dominikg: It would be great to elaborate on the reusable layouts or conditionals inside a single layout to establish best practices. Maybe linking to an examples repo?

dummdidumm: I enhanced the section with a little code example. For more a repository or an entry in the recipes section on `learn.svelte.dev` would be good indeed - but we can do that later 😄

documentation/docs/04-advanced-routing.md
```diff
@@ -126,75 +126,88 @@ assert.equal(
...
-> `default` is a reserved name — in other words, you can't have a `+layout-default.svelte` file.
+```js
+/// file: src/routes/nested/route/+layout.js
+import { reusableLoad } from '$lib/reusable-load-function';
+
+export function load(event) {
+	// Add additional logic here, if needed
+	return reusableLoad(event);
+}
+```
```
Conduitry, line 213: It looks like `$lib/reusable-load-function` not being a real module that shiki-twoslash can find is preventing the docs from building.

dummdidumm: thanks, will update 😅

documentation/docs/04-advanced-routing.md
```diff
@@ -126,75 +126,94 @@ assert.equal(
...
+/// file: src/routes/nested/route/+layout.js
+// @filename: ambient.d.ts
+declare module "$lib/reusable-load-function" {
+	export function reusableLoad(event: import('@sveltejs/kit').LoadEvent): Promise<Record<string, any>>;
+}
+// @filename: index.js
+// ---cut---
+import { reusableLoad } from '$lib/reusable-load-function';
+
+export function load(event) {
```
Conduitry, line 215:
```suggestion
/** @type {import('./$types').PageLoad} */
export function load(event) {
```

I think? Right now TS is still complaining that `event` is implicitly of `any` type, and I think the best way to handle that is to type `load` itself, because that's what people are going to be doing in their own code.

documentation/docs/04-advanced-routing.md
Conduitry: Yay! It's deploying now - https://kit-svelte-9pu9lu6by-svelte.vercel.app/ - resolving.

maximedupre:
When will this be available? I'm asking because I did a bunch of modifications based on the docs, but now I just realized that the doc has been deployed before the package 🥹 

And I don't remember how to do the previous named layout method (and no more doc on this) 😅

Rich-Harris:
@maximedupre I just merged the release PR (#6217) so it should hit npm any minute now

swgn:
```
src/routes/
├ (app)
│ ├ admin/
│ │ └ +page.svelte
│ └ +layout.svelte
├ +page.svelte
└ +layout.svelte```

tsukhu:
Excellent change love it

oneezy:
This PR broke symlink routes

❌ symlinks break `1.0.0-next.432` 
✅ symlinks work `1.0.0-next.431`

Created issue #6303 with repro

PR #6179, opened 2022-08-23
feat: Move `hooks.{js|ts}` init to `Server.init` instead of `Server.respond`

### LINKED ISSUE #6147: A way to validate things on server initialization
### Describe the problem

Related to #1538.

In other server solutions, a common pattern is check for the presence of necessary environment variables on startup, for the purposes of showing a friendly error message.

For example:

```ts
const { DATABASE_URI } = process.env;

if (DATABASE_URI === undefined || DATABASE_URI === "") {
  console.error(
    'The "DATABASE_URI" environment variable is not set. This is necessary in order for the server to function.',
  );
  console.error(
    'If you are deploying to Node, did you forget to copy the ".env.example" file to ".env" and fill in the values?',
  );
  console.error(
    "If you are deploying to Vercel, did you forget to add this environment variable to your project settings page?",
  );
  process.exit(1);
}
```

Doing this kind of thing prevents end-users from having to troubleshoot more-complicated run-time errors. ("Why is the database logic failing? Is the app that I just downloaded bugged?")

In #1538, [Rich-Harris recommends using "hooks.js"](https://github.com/sveltejs/kit/issues/1538#issuecomment-1002106271), but this doesn't solve the problem, as the logic in that file will only be executed when a user actually surfs to the page, rather than on server initialization.

SvelteKit should expose an idiomatic way to perform server-start validation of this manner.

This would also an appropriate place to e.g. initiate necessary database connections, so that the app can e.g. fail early if the database password is wrong (rather than when the first user actually tries to use your product).

### Importance

nice to have

tcc-sejohnson:
Closes #6147.

- Moved initialization of the `hooks.js` module to `Server.init` instead of `Server.respond`.
- `await`ed `Server.init` in all adapters that use it (`adapter-static` doesn't).
- Added dynamic environment variable support to prerendering. Added tests to verify this support.

TODO:
- @Rich-Harris I left you a few letters in the comments of a couple adapters. They're not big deals, just something for you to be aware of and make a judgment on

### Tests

### Changesets


packages/adapter-netlify/src/serverless.js
```diff
@@ -9,11 +9,20 @@ import { split_headers } from './headers';
...
+	let init_promise = server.init({
 		env: process.env
 	});
 
 	return async (event, context) => {
+		if (init_promise !== null) {
+			// dear Rich,
+			// should we just await it indiscriminately?
+			// sincerely,
+			// me
```
Rich-Harris, line 21: I prefer this version! It's closer to how it currently works. Hopefully we can get rid of `init_promise` at some point soon, not sure what the timeline on netlify supporting ESM is

```suggestion
```

packages/adapter-node/src/handler.js
```diff
@@ -11,7 +11,9 @@ import { env } from './env.js';
 /* global ENV_PREFIX */
 
 const server = new Server(manifest);
-server.init({ env: process.env });
+// Dear Rich,
+// it says top-level await isn't available. Should it be?
+let init_promise = server.init({ env: process.env });
```
Rich-Harris, line 16: I think we should update the tsconfig to allow TLA and see if anyone complains. it's supported in all the node versions we care about

```suggestion
await server.init({ env: process.env });
```

packages/adapter-node/src/handler.js
```diff
@@ -21,6 +23,14 @@ const host_header = env('HOST_HEADER', 'host').toLowerCase();
...
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
+/** @type {import('polka').Middleware} */
+const init = async () => {
+	if (init_promise !== null) {
+		await init_promise;
+		init_promise = null;
+	}
+};
+
```
Rich-Harris, line 33: we can remove this after updating the tsconfig

```suggestion
```

packages/adapter-node/src/handler.js
```diff
@@ -132,6 +142,7 @@ function get_origin(headers) {
 
 export const handler = sequence(
 	[
+		init,
```
Rich-Harris, line 145:
```suggestion
```

packages/kit/src/vite/build/build_server.js
```diff
@@ -99,21 +99,21 @@ export class Server {
...
 		this.options.public_env = pub;
-	}
-
-	async respond(request, options = {}) {
-		if (!(request instanceof Request)) {
-			throw new Error('The first argument to server.respond must be a Request object. See https://github.com/sveltejs/kit/pull/3384 for details');
-		}
 
-		if (!this.options.hooks) {
+    if (!this.options.hooks) {
```
Rich-Harris, line 103: tabs

```suggestion
		if (!this.options.hooks) {
```

packages/kit/src/vite/dev/index.js
```diff
@@ -32,6 +32,11 @@ export async function dev(vite, vite_config, svelte_config, illegal_imports) {
 
 	sync.init(svelte_config, vite_config.mode);
 
+	/** @type {Partial<import('types').Hooks>} */
+	const user_hooks = resolve_entry(svelte_config.kit.files.hooks)
+		? await vite.ssrLoadModule(`/${svelte_config.kit.files.hooks}`)
+		: {};
+
```
Rich-Harris, line 39: i think this needs to happen after `set_private_env` and `set_public_env`, no? otherwise anything that references dynamic env vars in hooks would fail, i think

Rich-Harris (commented):
this feels way better. comments inline

Rich-Harris:
Moved the `user_hooks = ...` code back inside the middleware, because a) it was running before aliases were set up, and b) we need to do it per-request so that changes to `src/hooks.js` take effect without a server restart

Rich-Harris (approved):

Issue #5967, opened 2022-08-17
Generate automatic `<slot />` `+layout.svelte` when `+layout.ts` is present


tcc-sejohnson:
### Describe the problem

Sometimes, I just want to create "side effects" from visiting a route: 

```ts
// +layout.ts
import { crumbsStore } from '$lib/components/Breadcrumb';
import type { LayoutLoad } from './$types';

export const load: LayoutLoad = () => {
	crumbsStore.addCrumb({ path: '/data-analytics/operations', name: 'Operations', enabled: true });

	return {
		pageTitle: 'Operations',
	};
};
```

In this case, there doesn't need to be any `+layout.svelte`, because all I want to do is add my `pageTitle` to the rolled-up `data` object (which my root layout is listening to, setting the page's Title) and add a breadcrumb to a store (which is how my `Breadcrumb` component works). Right now, I have to create a `+layout.svelte` with only `<slot />` as the content, or nothing will be rendered on or under this route. That seems kind of silly.

### Describe the proposed solution

When a `+layout.ts` or a `+layout.server.ts` exists but there isn't any `+layout.svelte`, Kit should just inject a default `+layout.svelte` containing an empty slot.

Note: If we went this route, we'd have to make an update to how we're generating types, as currently `+layout.ts` and `+layout.server.ts` don't get their types generated if there's no `+layout.svelte`.

### Alternatives considered

Just live with the extra files 🤷🏻 

### Importance

nice to have

dummdidumm:
When implementing this we need to take into account named layouts. So far, the `@` sign is only allowed on `.svelte` files, which would need to be softened here.

AlbertMarashi:
Also bumped into this issue, was confused why `LayoutLoad` wasn't appearing in `import("./$types")`

LeoDog896:
Also encountered this issue with making a form action. This took me so long to realize, so an error message (at the very least) would be appreciated. (The file errors out when a +page.server.ts is present only, with the error: node.component is not a function)

ghostdevv:
Ran into this when I had a layout group without a `+layout.svelte` the error I got was `node.component is not a function`

UnlimitedBytes:
Ran into this while following the docs: https://kit.svelte.dev/docs/load#input-methods-parent and it was very annoying to find out what the hell was wrong because the docs list it as valid..

Issue #4287, opened 2022-03-10
Throw an error (or warn?) if using adapter-static with invalid routes


Rich-Harris:
### Describe the problem

`adapter-static` is a source of confusion for people who believe that it allows you to _somehow_ publish an app with dynamic server rendering and/or endpoints with POST/PATCH/DELETE/etc handlers to GitHub pages.

### Describe the proposed solution

I haven't thought deeply about the implementation, beyond this:

* If pages are left unprerendered, `adapter-static` should throw an error unless `fallback` was specified.
* If an endpoint has non-GET handlers, `adapter-static` should throw an error regardless of whether `fallback` was specified.

### Alternatives considered

More complete documentation, and spending more time dealing with issues

### Importance

would make my life easier

Mlocik97:
I'm for warn, but not error... Sometimes I want to have implemented endpoints with post/patch/delete or hooks and other dynamic server stuff, and use `adapter-static` for example if you want to implement website (using node resp. `adapter-node` or some server-less one), but at the same time use same source codes to generate static assets for for example electron/neutralino app, or nativescript mobile app. Then I have two adapters in config, and two build commands that build for website or app based on for example `process.env`. So error would "break" this. But there should be warn definitelly.

Rich-Harris:
Prerendering happens automatically, regardless of what adapter you're using. In the rare case that an adapter doesn't specify where the prerendered output should be copied to, it's accessible at `.svelte-kit/output/prerendered`, so I'm not sure there's I understand why there's a reason to run `adapter-static` alongside a separate adapter.

Spenhouet:
I'm a bit confused about what's supposed to work and what not.

I'm using the **page routing (via slugs in file names), matchers and page endpoints (only GET)**. It's working when starting the dev environment but not when building it with adapter-static and deploying it to gh-pages.
**Which of these three features should generally work with adapter-static?** Are there any caveats? 

_Not going into any specifics here. If you say all three features generally should work, I will dig deeper and eventually open an issue to sort it out._

Mlocik97:
> Prerendering happens automatically, regardless of what adapter you're using. In the rare case that an adapter doesn't specify where the prerendered output should be copied to, it's accessible at `.svelte-kit/output/prerendered`, so I'm not sure there's I understand why there's a reason to run `adapter-static` alongside a separate adapter.

sometimes I have reason to not prerender (for adapter-node).

PR #6197, opened 2022-08-23
[breaking] `ssr/hydrate/router/prerender.default` are now configurable in `+page(.server).js` and `+layout(.server).js`

### LINKED ISSUE #6356: Prerendering overhaul
### Describe the problem

There's a few improvements that need to happen to prerendering that, while separate, are probably best tackled in one go. The broader context for this is #6197, in which it will be possible to set `export const prerender` in a layout — e.g. the root layout — such that it will be inherited by all pages within.

* Instead of being a boolean, `prerender` should be an enum so that we can distinguish between pages that _can_ be prerendered from those that _must_ be prerendered. The latter can safely be excluded from the SSR manifest, whereas presently we can only exclude prerendered pages without dynamic parameters
* `+server.js` files should be able to declare their prerenderability (or otherwise) as well. `+server.js` files with `POST` etc cannot be prerendered.
* https://github.com/sveltejs/kit/issues/4287

Must-prerender pages could be declared with

```js
export const prerender = 'always';
```

I'm not sure what a great name is for the current `prerender = true` (i.e. can-prerender). One suggestion was this...

```js
export const prerender = 'maybe';
```

...though I would be interested in alternative suggestions. (This isn't something you'd use often — only in cases where, for example, you want to prerender your 1,000 most popular blog posts or whatever and let the long tail be SSR'd, though #661 would be a better solution in the longer term.)

Since `+server.js` files are standalone — i.e. they don't have a `+layout` to inherit from — the naive approach would be to make it necessary to declare the setting for every file individually, which is a pain. Instead, I think `+server.js` files should inherit their prerenderability from the pages that request them during prerendering (i.e. if an endpoint is only fetched by a must-prerender page, it can be excluded; if it is fetched by a can-prerender page or isn't fetched at all, it must be included). If necessary, it can be made explicit with `export const prerender = value`.

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

dummdidumm:
# How to migrate

## browser.hydrate / browser.router / prerender.default

These options are removed from the `svelte.config.js`. Set them inside the root layout instead, either your `+layout.js` or `+layout.server.js`:

```diff
const config = {
	kit: {
-		browser: { hydrate: false, router: false },
-		prerender: { default: true }
	}
};
```

```js
// src/routes/+layout.js
export const hydrate = false;
export const router = false;
export const prerender = true;
```

(For now, you will need a `+layout.svelte` alongside your `+layout[.server].js` — this [will be fixed soon](https://github.com/sveltejs/kit/issues/5967).)

You can also set these options at lower layouts or inside pages if you want.

For other changes around `prerender` see this PR: #6392

## handle: resolve(event, { ssr: ... })

The `ssr` option was removed from the `resolve` function that is passed into the `handle` hook. Set it at the appropriate `+layout.js` or `+layout.server.js` or `+page.js` or `+page.server.js`. The following example shows how to turn off SSR for all admin pages:

```diff
export function handle({ event, resolve }) {
   return resolve(event, { 
-      ssr: !event.url.pathname.startsWith('/admin')
   });
```

```js
// src/routes/admin/+layout.js
export const ssr = false;
```

# PR description

Closes #6356

Can't believe how little code the actual change is 😄 

During implementing this, I realized a shortcoming of the current implementation: If `+page.js` accesses a browser global, for example
```js
document;
export function load() {}
```
..then this will fail on the server with an error, even if you did set `ssr` in `handle`. To change this behavior we would need to lazy-load `shared` just like we do with `component` already - should we make this change? If yes, should I incorporate it into this PR?

### Tests

### Changesets


dummdidumm:
I created #6207 as the enhanced version of this. I'd say that PR is a worthwhile addition and I doubt that the performance overhead of an additional lazy load is noticeable.

Rich-Harris:
Replies to the conversation in #6207:

I _have_ used `router` on a per-page basis, though admittedly in a very niche scenario. The covid tracker banner on https://www.nytimes.com...

---
 
image

---

...is a block of HTML extracted from a statically rendered SvelteKit app and placed onto the homepage. Disabling the router for the entire embed means that links will always take you to a fresh page, instead of trying to navigate within the homepage, which makes authoring the banner much easier (particularly when it uses components that are common to the whole app).

> one of the use cases I'd most like to support is load shedding. If your servers are overwhelmed then you can change an environment variable to have your clients render the page instead of your servers for a period of time to give you time to figure out your server utilization

Why does that need to be a `resolve` option, rather than this?

```js
// src/lib/config.js
import { env } from '$env/dynamic/public';

export const ssr = !env.PUBLIC_DISABLE_SSR;
```

```js
// +page.js
export { ssr } from '$lib/config.js';
```

The argument against allowing layouts to set `ssr` is that the semantics get super fuzzy, for me at least. I see it as a flag that signals whether something _can_ be SSR'd (and that's how we treat it when it's a `resolve` option), so to have a layout opt out of SSR only for a page to opt back _in_ to SSR feels nonsensical. At the very least, it's ambiguous. Having it be a page-only option eliminates the ambiguity, and gives us the option to change the behaviour in future.

documentation/docs/12-page-options.md
```diff
@@ -13,23 +11,27 @@ SvelteKit includes a [client-side router](/docs/appendix#routing) that intercept
 In certain circumstances you might need to disable [client-side routing](/docs/appendix#routing) with the app-wide [`browser.router` config option](/docs/configuration#browser) or the page-level `router` export:
 
 ```js
-/// file: +page.js/+page.server.js
+/// file: +page.js/+page.js
```
Rich-Harris, line 14:
```suggestion
/// file: +page.js
```

documentation/docs/12-page-options.md
```diff
@@ -13,23 +11,27 @@ SvelteKit includes a [client-side router](/docs/appendix#routing) that intercept
...
 
+You can control this setting on a per-app (via `svelte.config.js`) or per-page (via `+page.js`) basis. If both are specified, per-page settings override per-app settings in case of conflicts.
+
 ### hydrate
 
 Ordinarily, SvelteKit [hydrates](/docs/appendix#hydration) your server-rendered HTML into an interactive page. Some pages don't require JavaScript at all — many blog posts and 'about' pages fall into this category. In these cases you can skip hydration when the app boots up with the app-wide [`browser.hydrate` config option](/docs/configuration#browser) or the page-level `hydrate` export:
 
 ```js
-/// file: +page.js/+page.server.js
+/// file: +page.js/+page.js
```
Rich-Harris, line 27:
```suggestion
/// file: +page.js
```

documentation/docs/12-page-options.md
```diff
@@ -69,3 +73,14 @@ Because prerendering writes to the filesystem, it isn't possible to have two end
 For that reason among others, it's recommended that you always include a file extension — `src/routes/foo.json/+server.js` and `src/routes/foo/bar.json/+server.js` would result in `foo.json` and `foo/bar.json` files living harmoniously side-by-side.
 
 For _pages_, we skirt around this problem by writing `foo/index.html` instead of `foo`.
+
+### ssr
+
+Normally, SvelteKit renders your page on the server first and sends that HTML to the client where it's hydrated. If you set `ssr` to `false`, it renders an empty 'shell' page instead. This is useful if your page accesses browser-only methods or objects, but in most situations it's not recommended ([see appendix](/docs/appendix#ssr)).
+
+```js
+/// file: +page.js/+page.js
```
Rich-Harris, line 82:
```suggestion
/// file: +page.js
```

benmccann:
```
// +page.js
export { ssr } from '$lib/config.js';
```
is quite painful because it needs to be added to every individual page in your application and you'd need a `+page.js` for every page. That being said, if we were to allow it in `+layout.js` I would be mostly fine with that solution. I think it'd be perfectly cromulent to say that `(main)/+layout.svelte` is set based on an environment variable and `(main)/admin/+layout.svelte` is set always to `false`. You would lose the ability to do something like say that it's based on an environment variable, but always `true` for Google, but that's perhaps a niche enough usecase that I can live without that.

dummdidumm:
The `ssr` option can be set in layouts - which, as stated in the other PR, got me thinking, why we don't allow this for all options.

Rich-Harris:
> You would lose the ability to do something like say that it's based on an environment variable, but always `true` for Google

I'm definitely okay with that — [Google agree it's a bad idea](https://developers.google.com/search/docs/advanced/javascript/dynamic-rendering)

dummdidumm:
> > The `ssr` option can be set in layouts
> 
> only because of this PR! I'm saying that I'm not totally convinced that's a good idea, as it leads to ambiguity. At the very least we need to talk about the semantics because I don't think 'child beats parent' makes sense for `ssr` (even though it probably _would_ make sense for other options)

I definitely want this to be possible to set in layouts as well. The main use case I see for this is to disable SSR for sections of your app (like the admin section), not for single pages. So what would be better semantics then? "As soon as it's explicitly disabled it cannot be enabled again, because if you disable it in a layout it means you express that the layout already could have client-only code"? I don't know, this feels more complicated to explain and you lose out on some options when you "know what you are doing". "Last one wins" has my vote because of this.

Rich-Harris:
'Last one wins' is definitely the option that's easiest to apply consistently across page options, so as long as we're not worried about a layout declaring `ssr = false` because it uses a browser-only component only to blow up because of a misbehaving child, then I'm okay with that. Especially if layout-level options are the price of removing the `resolve` option.

If we do implement layout-level options then we could presumably eliminate `config.kit.browser` and `config.kit.prerender.enabled` in favour of setting them in the root layout, which feels like a nice win. Should we try and do that (and `.server.js` options) in this PR, or separately? Advantage of doing it here would be that we don't need to add temporary complexity to the docs

dummdidumm:
Agree that we should do it as part of this PR then.

If we do remove those options, do people have all possibilities still to enable/disable it? In other words, do they have access to globals/variables they cannot access in their root layout?
There's also the drawback that some people might never need a layout and they are forced to create one then - is that a big "ugh" or just a small "meh"?

Rich-Harris:
> If we do remove those options, do people have all possibilities still to enable/disable it?

Apart from controlling `ssr` per-request, yeah, I think that covers everything

> some people might never need a root layout and they are forced to create one then

I'm not _too_ worried about this — I'd expect almost all apps to need one eventually, so I'd file it under 'meh' personally

dummdidumm:
I removed the `ssr` option from `handle` as well as the `browser.hydrate/router` options. Wasn't sure about the prerendering option, I left that in for now.

Rich-Harris:
This is a huge improvement, I'm really glad we landed on this.

I think it's worth making `prerender` the same as the others — it'll make this stuff easier to document consistently. The one place where that becomes a bit of a nuisance is in `adapter-static`, which fails if `prerender.default` isn't true, but that's really only a half-measure _anyway_. I think we need to implement https://github.com/sveltejs/kit/issues/4287 plus the prerender enums we were talking about yesterday (plus prerenderable `+server.js` endpoints), so that adapters can know whether there are any un-prerendered/un-prerenderable routes. I'll work on that in a separate PR

documentation/docs/03-routing.md
```diff
@@ -108,6 +111,8 @@ export async function load({ params }) {
 
 During client-side navigation, SvelteKit will load this data from the server, which means that the returned value must be serializable using [devalue](https://github.com/rich-harris/devalue).
 
+Like `+page.js`, `+page.server.js` can export [page options](/docs/page-options) — `prerender`, `hydrate`, `router` and `ssr`.
```
dummdidumm, line 114: I did the same mistake 😄 
```suggestion
Like `+page.js`, `+page.server.js` can export [page options](/docs/12-page-options) — `prerender`, `hydrate`, `router` and `ssr`.
```

Rich-Harris: nope, the numeric prefix is incorrect! reverted

dummdidumm: yeah just realized, sry for the noise

Rich-Harris (approved):

Rich-Harris:
We'll need to update the migration guide here after we implement #6436

PR #6200, opened 2022-08-23
[docs] clarify depends docs

benmccann:
Clarifies both where the dependency exists and that `depends` has a var args param. The example was initially unclear to me as it looked to be showing one URL depending on another.

### DIFF
--- a/documentation/docs/05-load.md
+++ b/documentation/docs/05-load.md
@@ -99,7 +99,7 @@ An instance of [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL), co
 
 #### depends
 
-This function declares that the `load` function has a _dependency_ on a variable number of specific URLs, which can subsequently be used with [`invalidate()`](/docs/modules#$app-navigation-invalidate) to cause `load` to rerun.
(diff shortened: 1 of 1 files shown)


documentation/docs/05-load.md
```diff
@@ -99,7 +99,7 @@ An instance of [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL), co
 
 #### depends
 
-This function declares a _dependency_ on specific URLs, which can subsequently be used with [`invalidate()`](/docs/modules#$app-navigation-invalidate) to cause `load` to rerun.
+This function declares that the `load` function has a _dependency_ on a variable number of specific URLs, which can subsequently be used with [`invalidate()`](/docs/modules#$app-navigation-invalidate) to cause `load` to rerun.
```
Rich-Harris, line 102:
```suggestion
This function declares that the `load` function has a _dependency_ on one or more URLs, which can subsequently be used with [`invalidate()`](/docs/modules#$app-navigation-invalidate) to cause `load` to rerun.
```

Rich-Harris (approved):

PR #6222, opened 2022-08-24
always create $types.d.ts for a route with layout, leaf or endpoint

Rich-Harris:
Introduced a bug in #6174 — a route without a `+page|layout[.server].js` or `+server.js` wouldn't have `$types.d.ts` created for it, even though that file would be referenced by a child route that inherited from it. This is the correct check.

No test added because it was already causing failures (at a loss as to how it snuck through CI previously, but 🤷)

### Tests

### Changesets

### DIFF
--- a/.changeset/small-apples-visit.md
+++ b/.changeset/small-apples-visit.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Always create \$types for a route with a layout, leaf or endpoint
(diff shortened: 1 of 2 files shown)


.changeset/small-apples-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Always create \$types for a route with a layout, leaf or endpoint
```
Rich-Harris, line 5:
```suggestion
Always create `$types` for a route with a layout, leaf or endpoint
```

dummdidumm (approved):

PR #6226, opened 2022-08-24
[feat] add App.PageData type

### LINKED ISSUE #5951: Introduce a feature for bottom-up page-to-layout communication
### Describe the problem

Prior to the recent API overhaul (#5748), I used to have a `Meta.svelte` component that was rendered by my root layout. This component would read from `$page.stuff` — which I had typed in `app.d.ts` — and set various page-level head tags such as `<title>`, `<meta name="description">` and so on based on those `stuff` values. Every page would then have to provide this information in its `load` function, like so:
```ts
export function load() {
  return {
    stuff: {
      meta: {
        title: 'Home page',
        titleSuffix: 'long',
        description: 'Blah blah blah'
      }
    }
  };
}
```

My `app.d.ts` file looked something like:
```ts
declare namespace App {
  interface Stuff {
    meta: {
      title: string;
      titleSuffix?: 'short' | 'long' | 'none';
      description?: string;
    };
  }
}
```

And the `<Meta>` component:
```svelte
<script lang="ts">

  const titleSuffixes = new Map<TitleSuffix, string>([
    ['short', ' | Foo'],
    ['long', ' | Foo Bar Buzz'],
    ['none', '']
  ]);

  $: ({
    title,
    titleSuffix = 'short',
    description,
  } = $page.stuff.meta);
</script>

<svelte:head>
  <title>{title + titleSuffixes.get(titleSuffix)}</title>

  {#if description}
    <meta name="description" content={description} />
  {/if}
</svelte:head>
```

Now, ever since the API has been re-designed and the `stuff` thing has been removed, I've been struggling to re-create this properly.
Asking questions on the Discord server and reading [this part](https://github.com/sveltejs/kit/discussions/5774#discussioncomment-3292693:~:text=The%20data%20from%20all%20your%20load%20functions%20(or%20%2Blayout.server.js/%2Bpage.server.js%20files)%20is%20merged%20into%20a%20single%20object%20that%20can%20be%20used%20anywhere%20in%20your%20app%20via%20%24page.data%2C%20which%20replaces%20%24page.stuff.) of the migration guide lead me to think that returning this information as part of the data returned by each load function is apparently the way things like this are meant to be handled now. But I find that this is an almost unacceptable alternative, for two major reasons + a more minor one:
1. No type-safety while providing the values in the load functions:
```ts
export const load: PageLoad = () => {
  return {
    meta: {
      title: true, // Nothing's preventing me from doing this
      titleSuffix: 'somenonsense', // or this
      descriptititoitnotintotn: 'Blah blah blah', // or this 
    }
  };
};
```
2. No type-safety while reading the values in the `<Meta>` component

`Meta.svelte`:
```svelte
<script lang="ts">
</script>

<svelte:head>
    <!-- Absolutely no type-safety here, `meta` is `any` -->
    <title>{$page.data.meta.title}</title>
</svelte:head>
```
3. These values will be visible from the `+page.svelte` file included as part of its `data` prop, which is weird since it's actually none of its business.

image
`meta` isn't really something that this page is concerned with, so this is slightly awkward.

This (particularly the lack of type-safety) clearly yields a drastically poorer developer experience compared to the previous API. I sense that not enough thought has been put into this particular requirement when re-designing the API, namely bottom-up communication between pages and layouts. Page data isn't really fit for this use case at all.

I don't have a clear-cut proposal here, the general idea is that perhaps you could consider introducing a new construct designed specifically for bottom-up page-to-layout communication, in order to properly accommodate use cases like this, which are by no means uncommon or unusual, or niche. It actually gets more complex in a real-world app where there are typically multiple levels of layouts, and each of them receives a different set of configurations/settings; which is why I believe this is something that deserves a dedicated feature, so to speak.

### Importance

crucial

### Additional Information

Please don't just hand wave this away, thanks. I know it's annoying that literally dozens of issues get created on this repo every single day and naturally it's almost impossible to give each and every single one of them a lot of attention, but this one is a real problem, I think, and one that is in need of a response from your side.

And tell me if I'm missing something, because I certainly raised an eyebrow while trying to do this but failing to find a satisfying solution, so it's possible that I'm unaware of something here.

dummdidumm:
Closes #5951

Marking as draft because the "layout data is required if one of the pages doesn't have a load function" check isn't implemented yet. If have an idea about storing the layout nodes in a map, which is updated by its pages, which also solves the O(n^2) problem, but my head needs a break so I'll revisit it later.
This all has gotten pretty sophisticated, which is why I'm also inclined to create some tests for it.

### Tests

### Changesets


Rich-Harris:
I was thinking we could add a `children: PageNode[]` property on `PageNode` and add children at the same time as we do this:

https://github.com/sveltejs/kit/blob/9c7a6a9be352049cf12caec3ba7be62f5c907e29/packages/kit/src/core/sync/create_manifest_data/index.js#L292

That would solve the O(n^2) problem and also ensure that the parent-child relationships were correct (which the current naive approach doesn't, since `src/routes/foo/+layout.svelte` treats `src/routes/foo/bar/+page@.svelte` as a child, incorrectly)

packages/kit/src/core/sync/write_types/index.js
```diff
@@ -51,7 +51,7 @@ export async function write_all_types(config, manifest_data) {
 
 	// For each directory, write $types.d.ts
 	for (const route of manifest_data.routes) {
-		update_types(config, manifest_data, route);
+		update_types(config, create_routes_map(manifest_data), route);
```
Rich-Harris, line 54: surely `create_routes_map` should happen outside the loop?

packages/kit/src/core/sync/write_types/index.js
```diff
@@ -155,48 +171,44 @@ function update_types(config, manifest_data, route) {
...
-				// TODO this is O(n^2), see if we need to speed it up
-				for (const name of other.names) {
+		route.layout.child_pages?.forEach((page) => {
+			const leaf = routes.get(page);
+			if (leaf) {
+				for (const name of leaf.names) {
 					layout_params.add(name);
 				}
 			}
+			if (!page?.server && !page?.shared) {
```
Rich-Harris, line 209: is the optional chaining necessary? haven't fully grokked the PR yet but at first glance it feels like `child_pages` shouldn't have any falsy values

```suggestion
			if (!page?.server && !page?.shared) {
```

packages/kit/src/core/sync/write_types/index.js
```diff
@@ -251,18 +250,24 @@ function update_types(config, manifest_data, route) {
...
+ * @param {boolean} is_page
+ * @param {boolean} [all_pages_have_load]
  */
-function process_node(node, outdir, params) {
-	let data;
-	let load;
-	let server_load;
-	let errors;
+function process_node(node, outdir, is_page, all_pages_have_load = true) {
+	const params = `${is_page ? 'Route' : 'Layout'}Params`;
```
Rich-Harris, line 257: this is a side-note, probably for a separate PR, but I've been wondering if we should just always spell out the params type, i.e.

```ts
PageLoad<{ id: string }>
```

rather than 

```ts
PageLoad<RouteParams>
```

It's obviously more repetitive, but it means that if you hover over the `params` object you see this...

image

...instead of this, which is less useful:

image

packages/kit/src/core/sync/write_types/index.js
```diff
@@ -274,33 +279,52 @@ function process_node(node, outdir, params) {
...
+			`export type ${prefix}ServerLoad<OutputData extends ${output_data_shape} = ${output_data_shape}> = Kit.ServerLoad<${params}, ${parent_type}, OutputData>;`
+		);
+
+		exports.push(`export type ${prefix}ServerLoadEvent = Parameters<${prefix}ServerLoad>[0];`);
+
+		if (is_page) {
+			let errors = 'unknown';
+			if (proxy) {
+				const types = [];
+				for (const method of ['POST', 'PUT', 'PATCH']) {
```
Rich-Harris, line 304: why isn't `DELETE` included here?

Rich-Harris: (i'm sure there's a reason, i just mention it because it wasn't clear — maybe a comment would help)

Rich-Harris (approved):
Just had some minor comments, but to the extent I was able to understand this (it's brain-bending stuff 😅) it looks awesome. Very excited for this one

packages/kit/src/core/sync/write_types/index.js
dummdidumm: that's a good idea yeah, let's do this in a separate PR.

PR #6247, opened 2022-08-25
increase timeout to a silly number

Rich-Harris:
seeing a lot of timeouts on the safari tests for some reason, possibly a combination of the machine being underpowered and the AMP validator being a steaming turd


Rich-Harris:
this seems to have done the trick so i'm going to merge it, so other PRs can take advantage

PR #6271, opened 2022-08-25
[fix] type tweaks

### LINKED ISSUE #6261: `Typescript` doesn't show an error when trying to return an array in `load` function
The return type of the `load` function has to be a `Record`.
`Typescript` doesn't show an error when you trying to return an array.

_I opened this issue because I first thought `SvelteKit` incorrectly converts arrays to an object. I have updated the issue_

# original issue

### Describe the bug

When returning an `Array` from a `load` function, the `+page.svelte`'s `data` prop receives it as an array like object instead of a real object.

_input_
```ts
['some', 'items']
```

_output_

```ts
{
   0: 'some',
   1: 'items'
}
```

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-q6unzn?file=src/routes/+page.svelte

### Logs

_No response_

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.437 
    svelte: ^3.46.0 => 3.49.0 
    vite: ^3.0.8 => 3.0.9
```

### Severity

blocking an upgrade

### Additional Information

_No response_

dummdidumm:
- T extends Record<string, any>  allows T to be an array, Record<string, known> doesn't, closes #6261
- RequestHandler generic is partial now to align with the other types and prevent type conflicts
- generate definite set of route params, this is more correct and makes hover info better

### Tests

### Changesets

### DIFF
--- a/packages/kit/types/index.d.ts
+++ b/packages/kit/types/index.d.ts
@@ -195,17 +195,17 @@ export interface HandleError {
  */
 export interface Load<
 	Params extends Partial<Record<string, string>> = Partial<Record<string, string>>,
-	InputData extends Record<string, any> | null = Record<string, any> | null,
-	ParentData extends Record<string, any> = Record<string, any>,
-	OutputData extends Record<string, any> | void = Record<string, any> | void
+	InputData extends Record<string, unknown> | null = Record<string, any> | null,
+	ParentData extends Record<string, unknown> = Record<string, any>,
+	OutputData extends Record<string, unknown> | void = Record<string, any> | void
 > {
 	(event: LoadEvent<Params, InputData, ParentData>): MaybePromise<OutputData>;
 }
 
 export interface LoadEvent<
 	Params extends Partial<Record<string, string>> = Partial<Record<string, string>>,
-	Data extends Record<string, any> | null = Record<string, any> | null,
-	ParentData extends Record<string, any> = Record<string, any>
+	Data extends Record<string, unknown> | null = Record<string, any> | null,
+	ParentData extends Record<string, unknown> = Record<string, any>
 > {
 	fetch(info: RequestInfo, init?: RequestInit): Promise<Response>;
 	params: Params;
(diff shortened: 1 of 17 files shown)


Rich-Harris (approved):

packages/kit/types/index.d.ts
```diff
@@ -253,7 +253,9 @@ export interface RequestEvent<
  *
  * It receives `Params` as the first generic argument, which you can skip by using [generated types](/docs/types#generated-types) instead.
  */
-export interface RequestHandler<Params extends Record<string, string> = Record<string, string>> {
+export interface RequestHandler<
```
ivanhofer, line 256: @dummdidumm shouldn't this be
```ts
	Params extends Record<string, string> = Record<string, string>
> {
	(event: RequestEvent<Partial<Params>>): MaybePromise<Response>;
}
```
?

dummdidumm: Why exactly?

ivanhofer: I assume you want to makr all `params` of the `event` as optional 

[Here is an example](https://www.typescriptlang.org/play?#code/PTAEGMFcCdoUwHYBdQEsC2AHANndikBDJVAewQFgAoOAD01OhVWTmgDNDw5QANARgA8ABULRC6AM6g6SRABNpopqkLZBAJTjhG8wZKTQWAcwA0oA0YTGAfDdABeUMpJrN23fsMnzlk3fsAb2pQUNAACkwxCUkALmdoqQBKeIA3UlR5agBfamodBANQWn54gUFA0CjxdFKLb2tzaokAJgB+eL9rUGz7JwRIbGxQQmlCBABPPKoS8Mrm2viAciWepNAQUABJCHHQUkx0VBQAAwWWk+pZwOz1zZ2EODh5UCRSKtHpM8T+S6ppzbsVC0ah0BhMNCsDhcHi8FqCEJhZQxGS0OQIRSgLQ6aB6LpmepWWyOLEeXFeIm+Bq2RGhIK0iILOIJFRuZFSGwpUDpTI5aYFIq0FpleHzH6dak9PqgAZDEZjSbTIVzD41OorNYbMA7cB7A5HU7nP7Km53bWgeTkJYoR7PV7vKKSaRM6ZAA) that demonstrates the issue:

dummdidumm: I see. No it's working as intended. What you pass for the generic should be applied as-is, i.e. required params should be passed and can be expected to exist. It changes to `Partial<..>` so that in case you don't type the generic you can't just access the park without checking if it exists (in strict mode)

ivanhofer: Ah, now I get it. Thanks for the explanation. Didn't knew this trick before.

PR #6288, opened 2022-08-25
fix route sorting with groups

### LINKED ISSUE #6235: Route prioritization should disregard (foo) segments
### Describe the bug

When the routes are being prioritized, they should disregard `(foo)` segments. Currently, something like `(foo)/[...catchAll]` is prioritized above `bar` because it has more segments. `(foo)` segments shouldn't be counted when doing this sorting.

### Reproduction

none, currently

### Logs

_No response_

### System Info

```shell
System:
    Memory: 10.36 GB / 12.32 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Yarn: 1.22.19
    npm: 8.15.0
```

### Severity

blocking an upgrade

### Additional Information

_No response_

Rich-Harris:
closes #6235

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/sync/create_manifest_data/index.spec.js
+++ b/packages/kit/src/core/sync/create_manifest_data/index.spec.js
@@ -213,7 +213,10 @@ test('sorts routes correctly', () => {
 		routes.map((p) => p.id),
 		[
 			'',
+			'(foo)',
+			'(foo)/(bar)',
 			'about',
+			'foo',
 			'post',
 			'post/bar',
 			'post/foo',
@@ -221,6 +224,7 @@ test('sorts routes correctly', () => {
 			'post/f[zz]',
 			'post/f[xx]',
 			'post/f[yy]',
+			'foo/[bar]',
 			'post/[id]',
 			'[endpoint]',
 			'[wildcard]',
@@ -228,6 +232,7 @@ test('sorts routes correctly', () => {
 			'[...rest]/deep/[...deep_rest]',
 			'[...rest]/abc',
 			'[...rest]/deep',
+			'(foo)/(bar)/[...all]',
 			'[...anotherrest]',
 			'[...rest]'
 		]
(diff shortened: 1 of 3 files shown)


packages/kit/src/core/sync/create_manifest_data/index.spec.js
```diff
@@ -507,36 +507,31 @@ test('creates routes with named layouts', () => {
...
 			pattern: '/^/a/a2/?$/',
 			page: { layouts: [0, 2], errors: [1, undefined], leaf: 9 }
 		},
 		{
-			id: 'b/c/c2',
-			pattern: '/^/b/c/c2/?$/',
-			page: { layouts: [0], errors: [1], leaf: 11 }
+			id: 'a/a1',
+			pattern: '/^/a/a1/?$/',
+			page: { layouts: [0, 4], errors: [1, undefined], leaf: 10 }
```
Conduitry, line 518: Is having this one _after_ `(special)/a/a2` correct? I suppose in this case it isn't going to matter, but do we need to worry about how we're handling URLs with params in them? Is `(foo)/[b=b]` going to come before `[a=a]` because of the `(`? Ordinarily `[a=a]` would come before `[b=b]`, and it seems weird to me if that changes because of a group that doesn't affect any of the URLs.

Rich-Harris: `[a=a]` and `[b=b]` are sorted lexically, and so are `(foo)/[b=b]` and `[a=a]`. I get what you're saying but this feels simpler to me than 'identical-for-the-purposes-of-ranking routes are sorted lexically but only after ignoring their groups'. I don't know that there's an objectively right answer here — I think that if you have this level of ambiguity in your routes then you're making a rod for your own back regardless of what we do

Conduitry: Yup, that's fair.

Conduitry (approved):

PR #6293, opened 2022-08-25
[fix] prevent unused types warning

### LINKED ISSUE #6290: 'PageLoad is declared but its value is never read' when running svelte-check
### Describe the bug

When running `svelte-check` against a new project with a `+page.ts` that uses the `PageLoad` type, the `PageLoad` type is not used in the generated `proxy+page.ts` in the `.svelte-kit` directory. This causes the hint `'PageLoad' is declared but its value is never read` to be thrown.

### Reproduction

Create a new Svelte Kit app:

```
npm create svelte@latest my-app

Need to install the following packages:
  create-svelte@latest
Ok to proceed? (y) y

create-svelte version 2.0.0-next.161

Welcome to SvelteKit!

This is beta software; expect bugs and missing features.

Problems? Open an issue on https://github.com/sveltejs/kit/issues if none exists already.

✔ Which Svelte app template? › SvelteKit demo app
✔ Add type checking with TypeScript? › Yes, using TypeScript syntax
✔ Add ESLint for code linting? Yes
✔ Add Prettier for code formatting? Yes
✔ Add Playwright for browser testing? Yes
```

Install dependencies:

```sh
cd my-app
npm i
```

Add the following code to the end of `src/routes/about/+page.ts`:

```ts

export const load: PageLoad = () => {
  console.log('loading about page');
  return {};
};
```

Run Svelte Check:

```
npm run check
```

### Logs

```shell
npm run check

> my-app@0.0.1 check
> svelte-kit sync && svelte-check --tsconfig ./tsconfig.json


====================================
Loading svelte-check in workspace: /Users/ccamargo/Projects/my-app
Getting Svelte diagnostics...

/Users/ccamargo/Projects/my-app/.svelte-kit/types/src/routes/about/proxy+page.ts:15:1
Hint: 'PageLoad' is declared but its value is never read. 

```

### System Info

```shell
System:
  OS: macOS 11.6.8
  Memory: 275.84 MB / 16.00 GB
  Shell: 5.8 - /bin/zsh
Binaries:
  Node: 16.14.2 - /usr/local/opt/node@16/bin/node
  npm: 8.5.0 - /usr/local/opt/node@16/bin/npm
Browsers:
  Chrome: 104.0.5112.101
  Edge: 104.0.1293.63
  Firefox: 101.0.1
  Safari: 15.6
npmPackages:
  @sveltejs/kit: next => 1.0.0-next.440 
  svelte: ^3.46.0 => 3.49.0 
  vite: ^3.0.8 => 3.0.9
```

### Severity

annoyance

### Additional Information

Here is the resulting section of the `proxy+page.ts` file in `.svelte-kit`:

```ts

export const load = () => {
  console.log('loading about page');
  return {};
};
```

Notice `PageLoad` is imported but not used.

dummdidumm:
Fixes #6290

### Tests

### Changesets

### DIFF
--- a/.changeset/neat-cameras-swim.md
+++ b/.changeset/neat-cameras-swim.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] prevent unused types warning
--- a/packages/kit/src/core/sync/write_types/index.js
+++ b/packages/kit/src/core/sync/write_types/index.js
@@ -559,7 +559,13 @@ export function tweak_types(content, names) {
 										arg.name.end,
 										`: Parameters<${type}>[0]` + (add_parens ? ')' : '')
 									);
+								} else {
+									// prevent "type X is imported but not used" when svelte-check runs
+									code.append(`;${type};`);
 								}
+							} else {
+								// prevent "type X is imported but not used" when svelte-check runs
+								code.append(`;${type};`);
 							}
 
 							modified = true;
--- a/packages/kit/src/core/sync/write_types/index.js
+++ b/packages/kit/src/core/sync/write_types/index.js
@@ -560,11 +560,11 @@ export function tweak_types(content, names) {
 										`: Parameters<${type}>[0]` + (add_parens ? ')' : '')
 									);
 								} else {
-									// prevent "type X is imported but not used" when svelte-check runs
+									// prevent "type X is imported but not used" (isn't silenced by @ts-nocheck) when svelte-check runs
 									code.append(`;${type};`);
 								}
 							} else {
-								// prevent "type X is imported but not used" when svelte-check runs
+								// prevent "type X is imported but not used" (isn't silenced by @ts-nocheck) when svelte-check runs
 								code.append(`;${type};`);
 							}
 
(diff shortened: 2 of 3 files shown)


dummdidumm:
Just realized it's also benefitial to add a ts-nocheck comment at the top of all of these, so in case there are type errors they don't show up twice

Rich-Harris (approved):

Rich-Harris:
hahaha this is an incredible solution — i was thinking about walking the AST to see if the `import type` declaration was unused so we could remove it altogether. real 'the russians used a pencil' energy

Rich-Harris (changes requested):
rescinding my approval, but only so i don't accidentally merge it before the no-check stuff gets added

benmccann:
This really does feel like we need type tests :smile:  I don't know that we need to hold this up for tests since there's so much on our plates right now, but it might be nice to do post-summit

dummdidumm:
I agree we need type tests, but this one isn't really related to that. It's a hidden file generated from the user's file, and there are already tests for it.

Rich-Harris (approved):

benmccann:
Ah, I missed that there's a test because it was in the `src` die and I'm so used to all our tests being integration tests. Thanks for correcting me!

Issue #4051, opened 2022-02-22
Consider how to pass options between handlers in sequence


f-elix:
### Describe the bug

When using the `sequence` helper to chain multiple handle functions, `transformPage` is only actually run in the last one. Not sure if it was intended or not, but it would be nice if we could use `transformPage` in all handles, regardless of order.

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-wg5zme?file=src%2Fhooks.js&terminal=dev

You can see in the reproduction that the html `lang` attribute is not changed. If we swap the two handle functions so that the one that uses `transformPage` comes last, the html is changed.

### System Info

```shell
System:
    OS: Windows 10 10.0.22000
    CPU: (8) x64 Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz
    Memory: 19.88 GB / 39.76 GB
  Binaries:
    Node: 16.13.1 - C:\Program Files\nodejs\node.EXE
    npm: 8.4.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (98.0.1108.43)
    Internet Explorer: 11.0.22000.120
  npmPackages:
    @sveltejs/adapter-vercel: 1.0.0-next.43 => 1.0.0-next.43
    @sveltejs/kit: ^1.0.0-next.278 => 1.0.0-next.278
    svelte: ^3.46.4 => 3.46.4
```

### Severity

serious, but I can work around it

PR #6297, opened 2022-08-25
reorganise codebase a bit

Rich-Harris:
This started out with me working on #4051 and immediately wanting to move `packages/kit/src/hooks.js` to `packages/kit/src/hooks/sequence.js` so that we could have a `sequence.spec.js` alongside it, but not wanting to do that because it would make `src` even more raggedy than it currently is.

As of this PR, the contents of `src` look like this...

```bash
src/
├ core/     # the dev/build logic
├ exports/  # stuff exposed as the `@sveltejs/kit` package
├ runtime/  # code that runs when people use a SvelteKit app
├ utils/    # self-explanatory
└ cli.js    # self-explanatory
```

...which I think makes more sense than `src/vite`, `src/node` etc. There's definitely more that could be done here (the `runtime` directory in particular has become a bit chaotic), but sorting out the top level feels like a good start, and would make me feel calmer when I'm working in the codebase.

No changeset because this change should be invisible to users. I used `git mv` to preserve history and minimise the chance of merge conflicts.


benmccann:
This seems pretty good overall.

The contributing guide would need to be updated. 

It feels a little weird to me that `core` doesn't contain the most important code paths. Should it be called `build` or something?

(the directory tree in the description gave me a chuckle. I think someone's been typing `src/routes` a few too many times :smile:)

dummdidumm (approved):
Test failures seem unrelated/due to flakyness. Will do one rerun to be sure.

Rich-Harris:
The `$env` thing is a red herring. We should probably make it less noisy. It's actually #6306

Rich-Harris:
The distinction between `core` and `runtime` is roughly this: `runtime` is the stuff that runs through Vite during `dev` and is included in the app once it's built, `core` is the stuff that runs in Node and only during dev/build time. It definitely wouldn't make sense for `sync` to be in `runtime`, for example.

PR #6299, opened 2022-08-25
clarify when load runs

### LINKED ISSUE #2301: Rethink or clarify when `load` runs

Rich-Harris:
closes #2301

### DIFF
--- a/documentation/docs/05-load.md
+++ b/documentation/docs/05-load.md
@@ -351,9 +351,15 @@ export function load({ locals }) {
 
 SvelteKit tracks the dependencies of each `load` function to avoid re-running it unnecessarily during navigation. For example, a `load` function in a root `+layout.js` doesn't need to re-run when you navigate from one page to another unless it references `url` or a member of `params` that changed since the last navigation.
 
-Using [`invalidate(url)`](/docs/modules#$app-navigation-invalidate), you can re-run any `load` functions that depend on the invalidated resource (either implicitly, via [`fetch`](#fetch)), or explicitly via [`depends`](#depends). You can also invalidate _all_ `load` functions by calling `invalidate()` without an argument.
+A `load` function will re-run in the following situations:
 
-If a `load` function is triggered to rerun, the page will not remount — instead, it will update with the new `data`.
+- It references a property of `params` whose value has changed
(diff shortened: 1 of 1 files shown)


benmccann (approved):
thanks!

PR #6311, opened 2022-08-26
[fix] do not call server and update page.data unnecessarily

### LINKED ISSUE #6201: Requests for __data.json should be skipped entirely when possible
### Describe the bug

Re-running server-side `load` on each navigation is supposed to be skipped now (unless the `load` function depends on `url` etc.), but this does not appear to be fully working. Even if a given page has no `+page.server.js` file, the browser is still making requests to `__data.json` on each navigation, which return things like `{"type":"data","nodes":[{"type":"skip"},null]}` - the root (layout) node is skipped, and the leaf (page) node has no data. Requests that would tell the server to skip all of the nodes for which there are no `.server.js` files should be skipped entirely.

### Reproduction

https://github.com/Conduitry-Repros/kit-6201

### Logs

_No response_

### System Info

```shell
System:
    Memory: 8.34 GB / 12.32 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Yarn: 1.22.19
    npm: 8.15.0
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.431 
    svelte: ^3.44.0 => 3.49.0 
    vite: ^3.0.4 => 3.0.9
```

### Severity

### Additional Information

This is the situation I referred to in https://github.com/sveltejs/kit/pull/6056#issuecomment-1220125221 - it seems like it was not correctly (or fully?) handled.

### LINKED ISSUE #6286: LayoutData is reset even when unchanged
### Describe the bug

The `/__data.json` file is re-requested on every route change, even if the data is unchanged and not `invalidate()`'d. The data prop is then re-assigned which triggers updates within the app.

Response:
```json
{"type":"data","nodes":[{"type":"skip"},null]}
```

It's less noticeable during dev, when things are local, but with a proper network request the latency can be noticeable - simulating a longer network request makes it apparent and it is also triggering a change on the Layout / Page data which is triggering store re-calcs and (I presume) potential re-renders.

### Reproduction

Attach an extra hook to highlight the latency issue:

```ts
import type { Handle } from '@sveltejs/kit'

  await delay(500)

  return resolve(event)
}

async function delay(ms: number) {
  await new Promise(resolve => setTimeout(resolve, ms))
}
```

Or, use that dev-tool network throttling option that I just remembered about while typing this ... 😊 

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.5.1
    CPU: (6) x64 Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
    Memory: 1.89 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.14.0 - /usr/local/bin/node
    npm: 8.3.1 - /usr/local/bin/npm
  Browsers:
    Brave Browser: 102.1.39.111
    Chrome: 104.0.5112.101
    Chrome Canary: 107.0.5260.0
    Firefox: 103.0.2
    Safari: 15.6.1
  npmPackages:
    @sveltejs/adapter-node: next => 1.0.0-next.87 
    @sveltejs/kit: next => 1.0.0-next.440 
    svelte: ^3.49.0 => 3.49.0 
    vite: ^3.0.9 => 3.0.9
```

### Severity

### Additional Information

Related to https://github.com/sveltejs/kit/issues/6201 (couldn't add a comment to it)

dummdidumm:
Fixes #6201
Fixes #6286

Marking this as a draft because the test fails, because the root `+layout` always reruns on URL changes. Not sure how we should go about that, I'd like to prevent creating a new test app. Should we introduce a top level group?

Also: Not sure if this is the best implementation-wise, or if some things could be made shorter here (saving bytes on the client), or if we can infer this somehow.

### Tests

### Changesets


Rich-Harris:
Yeah, I ran into the same problem with https://github.com/sveltejs/kit/pull/6275 and glossed over it. Wrapping everything in a `(group)` would certainly make it easier to test anything that relies on bypassing the current route layout.

Implementation-wise, don't we just need to update data if any nodes (server or shared) were invalid, and otherwise leave it be? I'm not sure I understand why we need to treat `leaf_uses_server_data` differently

dummdidumm:
If
- the page uses a layout which is already loaded and doesn't need to be reloaded
- the page itself doesn't have a load function

then `route.uses_server_data` is `true` and `invalid_server_nodes` will also have `true` as the last entry, because the `loader` is different, because it's a different page. Therefore we also need to separately check if the leaf has a server load function, so that if all nodes leading up to it can be reused, no backend call needs to be made.

Rich-Harris:
Ah — sounds like the problem is that the `invalid_server_nodes` reducer is appending `true` for all new nodes, regardless of whether they have a server-only module or not. In _some_ cases we only need to care about the leaf, but the leaf isn't special — you could equally have a situation where you navigated from `/a/b/c` to `/a/d/e`, and both the leaf (`c -> e`) and the layout (`b -> d`) changed. The layout changing would cause the same false positive.

So we have two choices: we can either load all the nodes in order to determine whether they have a server-only module (which causes a delay), or we can encode `uses_server_data` in the manifest for each node individually (which increases the manifest size). If we wanted to minimise the effect on the manifest size we could encode it in a bitmask, perhaps, though that would limit route depth (including groups) to 31.

dummdidumm:
Oh right, any layout along the way could also not have a load function. What if we use the `~` trick for all layouts and the leaf? That way each layout encodes "I have a server load function" and the leaf no longer needs to encode that for itself and all parent, rather only itself.

dummdidumm:
I adjusted the root layout so it no longer reruns, which made the tests I've written possible/correct. Wondering why the CI is so flaky recently though ... locally it ran fine.

Rich-Harris:
The `adapter-static` tests are failing for me locally — I would guess there's an infinite reload loop involved, though I haven't investigated yet

Rich-Harris (approved):

Issue #6305, opened 2022-08-26
Arrays returned by endpoints (page or layout) are transformed into objects


SiegfriedEhret:
### Describe the bug

I have this `+page.svelte`:

```svelte
<script lang="ts">
	import type { PageData } from './$types';
	export let data: PageData;
</script>

<pre>{JSON.stringify(data, null, 2)}</pre>
```

And this `+page.ts`:

```ts
/** @type {import('./$types').PageLoad} */
export function load() {
	return ['42', '9000'];
}
```

In my page, I get this:

<img width="144" alt="image" src="https://user-images.githubusercontent.com/537567/186864314-de6c1780-4f13-46b5-906a-079a8c3272b6.png">

Since layout data are merged into children pages data, I understand that we must work with objects, but I would love to see a note in the docs about this.

### Reproduction

https://github.com/SiegfriedEhret/svelte-request/tree/main/src/routes/test

### System Info

```shell
❯ npx envinfo --system --binaries --browsers --npmPackages "{svelte,@sveltejs/*,vite}"

  System:
    OS: macOS 12.5
    CPU: (8) x64 Apple M1
    Memory: 216.27 MB / 16.00 GB
    Shell: 3.2.2 - /usr/local/bin/fish
  Binaries:
    Node: 16.17.0 - ~/.local/share/nvm/v16.17.0/bin/node
    npm: 8.18.0 - ~/.local/share/nvm/v16.17.0/bin/npm
  Browsers:
    Chrome: 104.0.5112.101
    Firefox Developer Edition: 105.0
    Safari: 15.6
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.66
    @sveltejs/kit: next => 1.0.0-next.442
    svelte: ^3.46.0 => 3.49.0
    vite: ^3.0.8 => 3.0.9
```

### Severity

annoyance

IvankoB:
Me confirm it too. The array object returned by endpoint looses its 'length' property on client side thus considered as non-iterable Affecting to Svelte's #each is very nasty. 
The workaround as iteration over Object.keys() causes convinient index functionality of #each be unavailable.

dummdidumm:
This is the expected behavior. The data of parent layouts and the data of the whole page need to be collected into one object, which is why arrays won't work. I think the serialization checker should take this into account.

SiegfriedEhret:
[...]

PR #6313, opened 2022-08-26
[docs] clarify load return value shape

dummdidumm:
Part of #6305

Writing it, I thought if we should also mention that the output is merged with the other load functions so people can also understand _why_ this is necessary.

### Tests

### Changesets


documentation/docs/05-load.md
```diff
@@ -4,7 +4,7 @@ title: Loading data
 
 A [`+page.svelte`](/docs/routing#page-page-svelte) or [`+layout.svelte`](/docs/routing#layout-layout-svelte) gets its `data` from a `load` function.
 
-If the `load` function is defined in `+page.js` or `+layout.js` it will run both on the server and in the browser. If it's instead defined in `+page.server.js` or `+layout.server.js` it will only run on the server, in which case it can (for example) make database calls and access private [environment variables](/docs/modules#$env-static-private), but can only return data that can be serialized as JSON.
+If the `load` function is defined in `+page.js` or `+layout.js` it will run both on the server and in the browser. If it's instead defined in `+page.server.js` or `+layout.server.js` it will only run on the server, in which case it can (for example) make database calls and access private [environment variables](/docs/modules#$env-static-private), but can only return data that can be serialized as JSON. On both the client and the server the output needs to be a dictionary at the top level.
```
Conduitry, line 7: I'm not sure how universally understood the term "dictionary" is. It's a term I've more heard in the context of other programming languages. Maybe "key-value object" or something like that? I do think it would be good to mention that this is because we merge the data at different levels, and so this only makes sense if it's a regular object.

documentation/docs/05-load.md
```diff
@@ -256,7 +256,7 @@ export async function load({ setHeaders }) {
 
 ### Output
 
-Any promises on the returned `data` object will be resolved, if they are top-level properties. This makes it easy to return multiple promises without creating a waterfall:
+On the client, the only requirement for the return value is to be a dictionary at the top level. On the server, the output also needs to be JSON-serializable. On both the client and server any promises on the returned `data` object will be resolved, if they are top-level properties. This makes it easy to return multiple promises without creating a waterfall:
```
Rich-Harris, line 259: I think the output stuff should go _after_ the promise discussion, because it's weird to say 'it has to be JSON-serializable' immediately followed by 'the top-level properties can be promises'. I think we could change this sentence back, and specify the constraints after the code sample

```suggestion
Any promises on the returned `data` object will be resolved, if they are top-level properties. This makes it easy to return multiple promises without creating a waterfall:
```

dummdidumm: Yeah that's better.

documentation/docs/05-load.md
```diff
@@ -256,7 +256,7 @@ export async function load({ setHeaders }) {
 
 ### Output
 
-Any promises on the returned `data` object will be resolved, if they are top-level properties. This makes it easy to return multiple promises without creating a waterfall:
+On the client, the only requirement for the returned `data` is to be an object at the top level. On the server, the output also needs to be JSON-serializable. The exception to this rule are top-level promises, which will be resolved before returning. This makes it easy to return multiple promises without creating a waterfall:
```
Rich-Harris, line 259: I think we need to avoid 'client' terminology, since `+page.js#load` will run on both sides:

```suggestion
The returned `data`, if any, must be an object of values. For a server-only `load` function, these values must be JSON-serializable. Top-level promises will be awaited, which makes it easy to return multiple promises without creating a waterfall:
```

dummdidumm: Third time's the charm, this one's great 👍

Rich-Harris (approved):

PR #6318, opened 2022-08-26
Allow server-only `load` functions to return more than JSON

### LINKED ISSUE #6008: Make it possible to return more than JSON from server-only `load` functions
### Describe the problem

Right now, the output of `load` in `+layout.server.js` or `+page.server.js` has to be serialized with `JSON.stringify` so that it can be transported to the client.

Occasionally it's useful to transport values that _can't_ be serialized as JSON. `Date` springs to mind, along with things like this, where anything expecting `selected` to be a member of `options` post-deserialization will be disappointed:

```js
export function load() {
  const options = [
    { flavour: 'chocolate' },
    { flavour: 'strawberry' },
    { flavour: 'banana' }
  ];

  const selected = options[0];

  return { options, selected };
}
```

We have a proven technique for this: https://github.com/rich-harris/devalue. It's fast, and doesn't involve a deserialization step. (It doesn't currently support [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) though, we'd need to add that.)

To work with CSP, we can't `eval` the result (including by creating a `<script>`), we need to `import` it. Since we need to get fresh data when navigating back to a previously-visited page, we need to use a cachebusting mechanism; since modules live in a cache that we have no way of purging, we need to be careful to avoid memory leaks.

We can't therefore do this...

```js
export const data = {...};
```

...and must instead do this, so that the data can be garbage collected once it's been used:

```js
window.__data = {...};
```

On the client side, it would look something like this:

```
let uid = 1;

// later
await import(`${url.pathname}/__data.js`);
const data = window.__data;
delete window.__data;
```

### Alternatives considered

There's a variety of alternatives to devalue listed here: https://github.com/rich-harris/devalue#see-also, though I'm not aware of any reasons to favour them.

Or, we could just stick to JSON, which has [better performance characteristics](https://v8.dev/blog/cost-of-javascript-2019#json) if you have a huge amount of data.

### Importance

nice to have

### Additional Information

_No response_

### LINKED ISSUE #6357: Invalidatation should always query the invalid server nodes with `fetch({ cache: 'no-cache' })`
### Describe the bug

Solution:
I'll upload PR soon.

### Reproduction

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    Memory: 2.92 GB / 15.49 GB
  Binaries:
  Browsers:
  npmPackages:
```

### Severity

### Additional Information

_No response_

Rich-Harris:
closes #6008
fixes #6357

This allows `load` functions to return things like `Date` objects, regexes, `Map` and `Set`, `BigInt` and so on. It also allows repeated and cyclical references, for the times when they're useful.

TODO:

* [x] docs
* [x] finish updating tests

### Tests

### Changesets


Conduitry:
Did you have an opinion about a statically analyzable config for making a given page use JSON instead of Devalue? Do you think the performance impact is not big enough to warrant that?

Rich-Harris:
I think it's _probably_ fine — for all but the most complex objects, we're talking fractions of milliseconds. One thing devalue could do to reduce the overhead is, after it completes its initial walk of the object, return `JSON.stringify(data)` if it didn't encounter any non-JSON-able values (or repeated references of objects or arrays). Though the resulting data will be larger as a result, so it's probably a wash.

LukeHagar (commented):
This commit is hilarious

mostrecent:
I've another idea re custom classes within devalue (requires the class to be imported when deserialization happens):
```javascript
> (0, eval)("({ a: new ObjectId('507f191e810c19729de860ea')})") // returns
{ a: new ObjectId("507f191e810c19729de860ea") }
```
So, devalue would keep user-defined custom classes as strings as it does with Date. Does this make any sense? (apologies for not putting this idea into devalue's repo.)

Rich-Harris:
No, I'm afraid that's probably a non-starter. The code generated by `devalue` needs to know how to reference that class in the client. It would involve jumping through some absurd hoops. Just take care of it on the server

packages/kit/src/runtime/client/client.js
```diff
@@ -1363,3 +1354,27 @@ export function create_client({ target, base, trailing_slash }) {
...
+ * @param {boolean[]} invalid
+ * @returns {Promise<import('types').ServerData>}
+ */
+async function load_data(url, invalid) {
+	const data_url = new URL(url);
+	data_url.pathname += `${url.pathname.endsWith('/') ? '' : '/'}__data.js`;
+	data_url.searchParams.set('__invalid', invalid.map((x) => (x ? 'y' : 'n')).join(''));
+	data_url.searchParams.set('__id', String(data_id++));
+
+	await import(/* @vite-ignore */ data_url.href);
```
dummdidumm, line 1371: could this be more sensitive to race conditions? For example you go to page A, then B right away, B is loaded faster, so you end up with A here? It might be safer to write to `window.__sveltekit_data_<id>` if that's possible or check if another request is pending.
In general I think a small comment why we do `await import` and `window._something` here might be good for future readers.

packages/kit/src/runtime/server/page/render.js
```diff
@@ -174,6 +174,29 @@ export async function render_response({
...
 	const prefixed = (path) => (path.startsWith('/') ? path : `${assets}/${path}`);
 
+	const serialized = { data: '', errors: 'null' };
+
+	try {
+		serialized.data = devalue(branch.map(({ server_data }) => server_data));
+	} catch (e) {
+		// TODO if we wanted to get super fancy we could track down the origin of the `load`
+		// function, but it would mean passing more stuff around than we currently do
+		const error = /** @type {any} */ (e);
```
dummdidumm, line 184: Maybe a small comment what this error throws (I'm guessing something like "X cannot be devalued")

packages/kit/src/runtime/server/page/render.js
```diff
@@ -191,7 +214,9 @@ export async function render_response({
 				error: ${error && serialize_error(error, e => e.stack)},
 				node_ids: [${branch.map(({ node }) => node.index).join(', ')}],
 				params: ${devalue(event.params)},
-				routeId: ${s(event.routeId)}
+				routeId: ${s(event.routeId)},
+				data: ${devalue(branch.map(({ server_data }) => server_data))},
+				errors: ${validation_errors ? devalue(validation_errors) : 'null'}
```
dummdidumm, line 219: This should probably reuse the `serialized.data` etc that are written to above

packages/kit/src/runtime/server/page/index.js
```diff
@@ -102,7 +103,7 @@ export async function render_page(event, route, page, options, state, resolve_op
 		}
 
 		const should_prerender_data = nodes.some((node) => node?.server);
-		const data_pathname = `${event.url.pathname.replace(/\/$/, '')}/__data.json`;
+		const data_pathname = `${event.url.pathname.replace(/\/$/, '')}/__data.js`;
```
dummdidumm, line 106: This could reuse `DATA_SUFFIX`

packages/kit/src/runtime/client/client.js
```diff
@@ -1284,32 +1273,32 @@ export function create_client({ target, base, trailing_slash }) {
...
 				const branch_promises = node_ids.map(async (n, i) => {
+					const server_data_node = server_data_nodes[i];
+
+					if (server_data_node?.type === 'error') {
+						if (server_data_node.httperror) {
+							throw error(server_data_node.httperror.status, server_data_node.httperror.message);
+						}
+
+						throw server_data_node.error;
+					}
```
dummdidumm, line 1300: I'm not this has the desired outcome. If there's a server error from loading data we want to show the error page within the correct layout, and this would result in the root error page being shown, if I understand this code correctly.

dummdidumm (commented):
This looks very nice! Looking at the code I think it may even be possible to make this configurable to some extend. The window method stays but what kind of format is written to it could be configured by the user (JSON, superjson, whatever). Not for now but thought for later.

packages/kit/src/runtime/client/client.js
Rich-Harris: It's not vulnerable to race conditions, because we read and delete the data in the same tick that the promise resolves. Something would have to happen in between the `await import(...)` and the `const server_data = ...` for it to be racy.

We can't use a unique ID for the data because that would fail in cases where the data was prerendered.

I'll add some comments

Rich-Harris: ae423fd9abba9f6c04f0e718405b6f82b00f03aa

packages/kit/src/runtime/server/page/render.js
Rich-Harris: 3e3d05de1468f6541154c7f284d49649016d7041

packages/kit/src/runtime/server/page/render.js
Rich-Harris: whoops! fixed b970d5cfafd076d943b8e8ce3eb1d42f9cc490ab

packages/kit/src/runtime/server/page/index.js
Rich-Harris: 358d0ceb323f9c53a2bfc6ef1acd9f742eadf1b5

packages/kit/src/runtime/client/client.js
Rich-Harris: Ah yeah, you're right. Will fix this

Rich-Harris: I realised that errors won't be serialized anyway 🤦 Removed this block.

If a client-side error occurs during `load` then it will fall back to the root error page, which isn't 'correct' but would be finicky to solve. On the basis that this isn't a regression, I think we should gloss over it for now and merge this

dummdidumm (approved):

packages/kit/src/runtime/client/client.js
dummdidumm: I didn't mean that we should handle the "client side error occurs during hydration load" error in some way, only that the `throw` case would be wrong, since in case of an error you have the nodes leading up to an error page already and only want to hydrate that, not throw an error along the way. So all good now 👍

0gust1:
Oh. Yeah. :)

PR #6328, opened 2022-08-26
[docs] don't hardcode file extension in docs server

benmccann:
This does not expose any user-visible changes, but makes it easier to enable TypeScript support to the docs backend. allowing me to write code blocks in TypeScript

Screenshot from 2022-08-26 14-13-38

Using this throughout the docs would require a larger overhaul to render both JS and TS and add a toggle between them, so I'm not sure that's something I'll do in the short-term. However, this is helpful in allowing me to experiment with stuff like adding a TypeScript example to our homepage to demonstrate that we provide TypeScript type declarations.

### DIFF
--- a/sites/kit.svelte.dev/src/lib/docs/server/index.js
+++ b/sites/kit.svelte.dev/src/lib/docs/server/index.js
@@ -92,7 +92,7 @@ export async function read_file(dir, file) {
 				})
 				.replace(/\*\\\//g, '*/');
 
-			if (language === 'js' || language === 'ts') {
+			if (language === 'js') {
 				try {
 					if (source.includes('./$types') && !source.includes('@filename: $types.d.ts')) {
 						const params = parse_route_id(options.file || `+page.${language}`)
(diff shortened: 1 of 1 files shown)


benmccann:
It looks like https://kit.svelte.dev/docs/modules#$app-env is triggering an error:

> Cannot find module '$app/env' or its corresponding type declarations.

I wonder why it doesn't work on that page but does on https://kit.svelte.dev/docs/migrating#integrations-html-minifier

Rich-Harris:
Yeah, excluding TypeScript is deliberate, because those snippets are impossible to render otherwise

benmccann:
I don't understand why they would fail and https://kit.svelte.dev/docs/migrating#integrations-html-minifier which also uses `$app/env` works though

Longer-term we should probably find some other way to exclude those snippets I'd guess. But for now I updated it to continue skipping TypeScript but dynamically populate the file extension so that I only have to change one line when I want to temporarily enable TypeScript

Rich-Harris (approved):

PR #6350, opened 2022-08-27
[feat] support for other aliases in package command

BeeMargarida:
Related to https://github.com/sveltejs/kit/issues/1950

This PR includes support for other aliases besides `$lib` for the package command. It reads the alias information from `kit` object in `svelte.config.js`.

I'm not sure this is the correct approach, open to suggestions and improvements.

### Tests

### Changesets

### DIFF
--- a/.changeset/weak-hotels-hang.md
+++ b/.changeset/weak-hotels-hang.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/package': patch
+---
+
+Support for other aliases in package command
--- a/.changeset/weak-hotels-hang.md
+++ b/.changeset/weak-hotels-hang.md
@@ -2,4 +2,4 @@
 '@sveltejs/package': patch
 ---
 
-Support for other aliases in package command
+[feat] Support aliases set through `kit.alias`
--- a/packages/package/src/utils.js
+++ b/packages/package/src/utils.js
@@ -18,22 +18,31 @@ import { posixify, mkdirp, walk } from './filesystem.js';
 export function resolve_lib_alias(file, content, config) {
 	/**
 	 * @param {string} match
-	 * @param {string} _
 	 * @param {string} import_path
+	 * @param {string} alias
+	 * @param {string} value
 	 */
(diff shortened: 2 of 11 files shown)


packages/package/src/utils.js
```diff
@@ -18,22 +18,31 @@ import { posixify, mkdirp, walk } from './filesystem.js';
...
+
+	const aliases = { $lib: path.resolve(config.package.source), ...(config.kit?.alias ?? {}) };
+	for (const [alias, value] of Object.entries(aliases)) {
+		content = content.replace(/from\s+('|")([^"';,]+?)\1/g, (match, _, import_path) =>
+			replace_import_path(match, import_path, alias, value)
+		);
+		content = content.replace(/import\s*\(\s*('|")([^"';,]+?)\1\s*\)/g, (match, _, import_path) =>
+			replace_import_path(match, import_path, alias, value)
+		);
+	}
```
dummdidumm, line 45: This is propbably super edge-case-y, but this could theoretically replace import paths multiple times. Could we turn that around and do the for loop inside the replace functions, and check if a replacement was done by checking if output != match?

dummdidumm (changes requested):
This looks awesome, thank you! I left one minor comment; also, could you add a changeset (by running `npx changeset` in the root of the monorepo and following the prompts)?

packages/package/src/utils.js
BeeMargarida: You are right, there was that possibility. I moved the loop inside, but I did not understood what you meant by the last part of checking if output != match.

.changeset/weak-hotels-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/package': patch
+---
+
+Support for other aliases in package command
```
dummdidumm, line 5:
```suggestion
[feat] Support aliases set through `kit.alias`
```

dummdidumm (approved):
👍

martinbeentjes:
Awesome! Thanks for this 👌🏼

Issue #6354, opened 2022-08-28
`invalidate('...')` doesn't work after navigating to a page with parameters


aradalvand:
### Describe the bug

Repro: https://github.com/aradalvand/sveltekit-invalidate-with-params

It seems like having a route with params (e.g. `routes/whatever/[id]`) along with a load function that have `depends(...)` doesn't work as expected, when you navigate to that page with the client-side router, and then run `invalidate('...')`, it won't work. Very weird. I'm pretty confident this is a bug.

### Reproduction

- Clone https://github.com/aradalvand/sveltekit-invalidate-with-params
- Run and open the app (the home page) in the browser
- Open the console, run `invalidate('user')`. You'll see that the layout's load function gets re-run, as expected:
![image](https://user-images.githubusercontent.com/26527405/187087544-eee0938b-19a2-4e18-81aa-5beb0b12a9fb.png)
- Now, click on the link in the home page titled "Go to foo"
- After navigating to the `/foo/123` page, open the console and run the same command again. This time, it won't work, the layout's load function isn't re-run:
![image](https://user-images.githubusercontent.com/26527405/187087581-0f36b16e-2f3a-45b9-b4dc-01bcde127e62.png)
 

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 20.04.4 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
    Memory: 486.11 MB / 3.77 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 16.15.1 - /usr/bin/node
    npm: 8.11.0 - /usr/bin/npm
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.65
    @sveltejs/kit: next => 1.0.0-next.445
    svelte: ^3.44.0 => 3.49.0
    vite: ^3.0.4 => 3.0.9
```

### Severity

blocking all usage of SvelteKit

### Additional Information

The problem only exists for pages that have parameters, I have no idea why, but I think this information could be useful for diagnosing the problem.

dummdidumm:
This works as expected. The `invalidate`/`depends` function expects an URL / constructs one, so in this case `depends('user')` becomes `http://127.0.0.1:5173/user`, but `invalidate('user')` inside `foo` becomes `http://127.0.0.1:5173/foo/user` - this doesn't match, so nothing reruns.
[...]

Issue #6349, opened 2022-08-27
+layout.server.ts load doesn't run after invalidate()


CaptainCodeman:
### Describe the bug

The `invalidate()` function no longer triggers a reload of the root layout (which has changed data). It works on v443, but stopped working on v444+

### Reproduction

https://github.com/CaptainCodeman/sveltekit-example/tree/ssr

The `invalidate()` call happens after client-side auth state changes and `/session` is called to set or clear server-side cookie. `LayoutData` should reflect the `user` from that cookie.

Needs firebase emulator running to execute locally. I can make a simpler cutdown repro if needed, was just working on this example which made me notice it:

No data reloads with v444:

https://user-images.githubusercontent.com/304910/187050362-e0a2204c-d334-43dd-8f83-f4e6f4286728.mov

Same code with v443 showing layout data updates:

https://user-images.githubusercontent.com/304910/187050370-5d9d9692-2a9e-4e77-8089-4e9d33092d91.mov

### System Info

```shell
System:
    OS: macOS 12.5.1
    CPU: (6) x64 Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
    Memory: 1.97 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.14.0 - /usr/local/bin/node
    npm: 8.3.1 - /usr/local/bin/npm
  Browsers:
    Brave Browser: 102.1.39.111
    Chrome: 104.0.5112.101
    Chrome Canary: 107.0.5265.0
    Firefox: 103.0.2
    Safari: 15.6.1
    Safari Technology Preview: 16.0
  npmPackages:
    @sveltejs/adapter-node: next => 1.0.0-next.87 
    @sveltejs/kit: 1.0.0-next.444 => 1.0.0-next.444 
    svelte: ^3.49.0 => 3.49.0 
    vite: ^3.0.9 => 3.0.9
```

### Severity

serious, but I can work around it

bjon:
There is another broken thing with v444. `data` in `+layout.ts` is now sometimes null, even though `+layout.server.ts` returns an object.

CaptainCodeman:
Do you mean something that's part of the data prop when it's inside the svelte component, or in the http response from the `/__data.json` request?

bjon:
I mean the `load({data})` in `layout.ts`, that should hold the data from `layout.server.ts`.

luhagel:
[...]

PR #6358, opened 2022-08-28
[fix] skip `fetch` caching if some server node was invalidated

### LINKED ISSUE #6357: Invalidatation should always query the invalid server nodes with `fetch({ cache: 'no-cache' })`

Tal500:
fixes #6357

and maybe also #6354 and #6349

### Tests

### Changesets

### DIFF
--- a/.changeset/hip-forks-hunt.md
+++ b/.changeset/hip-forks-hunt.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Skip `fetch` caching if some server node was invalidated (fixes #6357)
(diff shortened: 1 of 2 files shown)


Rich-Harris:
thank you — this will be obsoleted by the changes in #6318, so I'll close this

PR #6327, opened 2022-08-26
Run server.init() for edge functions on Vercel

SG60:
Required since #6179.

This should close #6308.

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #6360, opened 2022-08-29
[adapter-vercel] don't use top level await in Vercel Edge Functions

Schniz:
This fixes a regression created in #6327, due to the usage of top level await, which isn't supported in Vercel Edge Functions right now.
Using a deferred promise and awaiting within the response it practically the same thing, but works.

### Tests

### Changesets

### DIFF
--- a/.changeset/shiny-sloths-juggle.md
+++ b/.changeset/shiny-sloths-juggle.md
@@ -1,5 +1,6 @@
 ---
 '@sveltejs/adapter-vercel': patch
+'@sveltejs/adapter-netlify': patch
 ---
 
 Don't use top-level-await, as it is not supported right now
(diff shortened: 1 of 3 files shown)


dummdidumm (approved):
Thank you!

evdama:
The problem still exists with `"@sveltejs/kit": "^2.5.18"` and `"@sveltejs/adapter-vercel": "^5.4.0"` using `runtime: edge`

```js
// svelte.config.js

import adapter from '@sveltejs/adapter-vercel'
import { vitePreprocess } from '@sveltejs/vite-plugin-svelte'

const config = {
  kit: {
    adapter: adapter( { runtime: 'edge' } ),
    alias: {
      $components: 'src/components',
    },
  },
  preprocess: vitePreprocess(),
  vitePlugin: {inspector: true}
}
```
export default config

PR #6367, opened 2022-08-29
[breaking] add error.html

### LINKED ISSUE #3068: Thrown exceptions from __layout.svelte load result in plain text error responses

dummdidumm:
This is a static error page that will be rendered by the server when everything else goes wrong
Closes #3068

Implementing this brought up some design questions around how dynamic the error page is. The places I added it previously showed the error message and possibly the stack trace in dev, which is no longer the case. Should we have something like an optional placeholder named `%sveltekit.error%` which is passed the stringified message, and in dev also the stack trace?

### Tests

### Changesets

### DIFF
--- a/documentation/docs/03-routing.md
+++ b/documentation/docs/03-routing.md
@@ -190,7 +190,7 @@ If an error occurs during `load`, SvelteKit will render a default error page. Yo
 <h1>{$page.status}: {$page.error.message}</h1>
 ```
 
-SvelteKit will 'walk up the tree' looking for the closest error boundary — if the file above didn't exist it would try `src/routes/blog/+error.svelte` and `src/routes/+error.svelte` before rendering the default error page.
+SvelteKit will 'walk up the tree' looking for the closest error boundary — if the file above didn't exist it would try `src/routes/blog/+error.svelte` and `src/routes/+error.svelte` before rendering the default error page. If that fails, the contents of a static error page (`src/error.html` by default) will be rendered instead.
 
 ### +layout
 
--- a/documentation/docs/15-configuration.md
+++ b/documentation/docs/15-configuration.md
@@ -40,7 +40,8 @@ const config = {
 			params: 'src/params',
 			routes: 'src/routes',
 			serviceWorker: 'src/service-worker',
-			template: 'src/app.html'
+			template: 'src/app.html',
+			errorPage: 'src/error.html'
 		},
 		inlineStyleThreshold: 0,
 		methodOverride: {
--- a/packages/create-svelte/templates/default/src/error.html
+++ b/packages/create-svelte/templates/default/src/error.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="utf-8" />
+		<title>Error</title>
+	</head>
+	<body>
+		<h1>Error</h1>
+		<p>An error occurred.</p>
+	</body>
+</html>
--- a/packages/kit/src/core/config/index.js
+++ b/packages/kit/src/core/config/index.js
@@ -38,6 +38,34 @@ export function load_template(cwd, config) {
 }
 
 /**
+ * Loads the error page (src/error.html by default) if it exists.
+ * Falls back to a generic error page content.
+ * @param {string} cwd
+ * @param {import('types').ValidatedConfig} config
+ */
+export function load_error_page(cwd, config) {
+	const { errorPage } = config.kit.files;
+	const relative = path.relative(cwd, errorPage);
+
+	if (fs.existsSync(errorPage)) {
+		return fs.readFileSync(errorPage, 'utf-8');
+	} else {
+		console.log(`${relative} does not exist. Using generic error page instead.`);
+		return `<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="utf-8">
+	<title>Error</title>
+</head>
+<body>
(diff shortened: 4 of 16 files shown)


Rich-Harris:
We definitely need to include `%sveltekit.status%` and `%sveltekit.message%` (though I just remembered that sanitizing error messages in production is still a TODO). I think we should leave `stack` out of it since it'll always be printed to the terminal (in fact I think the same applies to normal errors — our error handling could use a bit of a tidy-up, separately to this PR)

packages/kit/src/core/config/index.js
```diff
@@ -37,6 +37,34 @@ export function load_template(cwd, config) {
...
+ * @param {import('types').ValidatedConfig} config
+ */
+export function load_error_page(cwd, config) {
+	const { errorPage } = config.kit.files;
+	const relative = path.relative(cwd, errorPage);
+
+	if (fs.existsSync(errorPage)) {
+		return fs.readFileSync(errorPage, 'utf-8');
+	} else {
+		console.log(`${relative} does not exist. Using generic error page instead.`);
```
dummdidumm, line 53: I made it so that in case `src/error.html` (or whatever) doesn't exist we only output a log and fall back to a generic error page instead. This way not everyone needs to have this if they don't want to (or need to, in case of SPAs). Alternatives:
- be completely silent / no logging
- throw

dummdidumm (commented):
Made the error page dynamic with respect to status / message

packages/create-svelte/templates/default/src/error.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
```
Rich-Harris, line 1: I kinda feel like we should remove these files from the templates, because they make it seem like this (rather than `src/routes/+error.svelte`) is the proper place to define your error page, when in reality this is the absolute last resort, and a framework default is totally acceptable (if we just make it a bit nicer than it currently is). Thoughts?

documentation/docs/15-configuration.md
```diff
@@ -40,7 +40,8 @@ const config = {
 			params: 'src/params',
 			routes: 'src/routes',
 			serviceWorker: 'src/service-worker',
-			template: 'src/app.html'
+			template: 'src/app.html',
+			errorPage: 'src/error.html'
```
Rich-Harris, line 44: `template` and `errorPage` feel like weird siblings, especially now that `errorPage` is _also_ a template. I wonder if this should become `appTemplate` and `errorTemplate`? Wouldn't be a breaking change for most people since most people don't set these values anyway

packages/create-svelte/templates/default/src/error.html
dummdidumm: Sounds sensible. It could be a little overwhelming with all these files adding up.

documentation/docs/15-configuration.md
dummdidumm: Yeah I was wondering this too, appTemplate/errorTemplate feel more consistent.

documentation/docs/03-routing.md
```diff
@@ -190,7 +190,7 @@ If an error occurs during `load`, SvelteKit will render a default error page. Yo
 <h1>{$page.status}: {$page.error.message}</h1>
 ```
 
-SvelteKit will 'walk up the tree' looking for the closest error boundary — if the file above didn't exist it would try `src/routes/blog/+error.svelte` and `src/routes/+error.svelte` before rendering the default error page.
+SvelteKit will 'walk up the tree' looking for the closest error boundary — if the file above didn't exist it would try `src/routes/blog/+error.svelte` and `src/routes/+error.svelte` before rendering the default error page. If that fails, the contents of a static error page (`src/error.html` by default) will be rendered instead.
```
Rich-Harris, line 193:
```suggestion
SvelteKit will 'walk up the tree' looking for the closest error boundary — if the file above didn't exist it would try `src/routes/blog/+error.svelte` and `src/routes/+error.svelte` before rendering the default error page. If _that_ fails, SvelteKit will bail out and render a static fallback error page, which you can customise by creating a `src/error.html` file.
```

Rich-Harris (approved):

PR #6372, opened 2022-08-29
Bundle non-prod dependencies with esbuild

### LINKED ISSUE #3176: Provide ability to bundle dev dependencies for Node adapter

Rich-Harris:
This closes #3176, by always bundling dev dependencies using `esbuild`. Supersedes #6301.

### Tests

### Changesets

### DIFF
--- a/.changeset/big-trainers-love.md
+++ b/.changeset/big-trainers-love.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-node': patch
+---
+
+Bundle non-production dependencies with esbuild
--- a/packages/adapter-node/README.md
+++ b/packages/adapter-node/README.md
@@ -140,11 +140,13 @@ app.listen(3000, () => {
 
 ## Deploying
 
-You will need the output directory (`build` by default), the project's `package.json`, and the production dependencies in `node_modules` to run the application. Production dependencies can be generated with `npm ci --prod`, you can also skip this step if your app doesn't have any dependencies. You can then start your app with
+You will need the output directory (`build` by default), the project's `package.json`, and the production dependencies in `node_modules` to run the application. Production dependencies can be generated with `npm ci --prod` (you can skip this step if your app doesn't have any dependencies). You can then start your app with
 
 ```bash
 node build
 ```
+
+Development dependencies will be bundled into your app using `esbuild`. To control whether a given package is bundled or externalised, place it in `devDependencies` or `dependencies` respectively in your `package.json`.
 
 ## Changelog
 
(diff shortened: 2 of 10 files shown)


benmccann:
It's probably worth adding something to the `adapter-node` readme saying that `devDependencies` will be bundled and `dependencies` will not so that people know what to put where

gterras:
> It's probably worth adding something to the `adapter-node` readme saying that `devDependencies` will be bundled and `dependencies` will not so that people know what to put where

Also maybe a general note explaining the difference since it's less obvious than a common node project, like the Sapper docs did.

AlexRMU:
What about minification of the entire server part and bundling of prod dependencies?

Rich-Harris:
There's really nothing to be gained from minifying server-side code except making it harder to debug. As the README says, if you want a dependency to be bundled, move it to devDependencies

Conduitry (approved):

AlexRMU:
My build weighs 120 MB, which is 3 times more than node 16 alpine. It's sad. You can make minification optional.

PR #6382, opened 2022-08-29
[fix] remove vestigial JSON server_data

Conduitry:
When we started serializing server data with devalue, the old JSON serialization of it was left in. This removes it.

This also updates the types so that `server_data` is no longer a valid `type` to pass to `render_json_payload_script`. This required a slight adjustment to the unit tests.

### Tests

### Changesets

### DIFF
--- a/.changeset/quiet-poems-tease.md
+++ b/.changeset/quiet-poems-tease.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] remove unnecessary JSON serialization of server data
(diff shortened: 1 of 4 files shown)


Conduitry:
The difference with and without the extra `server_data` block can be seen on, for example https://kit.svelte.dev/docs/introduction and https://kit-svelte-di38rlap3-svelte.vercel.app/docs/introduction

Rich-Harris (approved):

Rich-Harris:
JFC these windows test timeouts are killing me. wtf is wrong with github actions

PR #6392, opened 2022-08-29
Prerendering overhaul

Rich-Harris:
# Migration

## General

`+server` pages can be prerendered, too now, by adding the `export const prerender = ..` option to them. By default they are not prerendered, but any `+server` page that is referenced from a prerendered page inherits the `prerender` option of that page unless explicitly set otherwise. This might lead to a situation where you now get a "Cannot prerender a +server file with POST, PATCH, PUT, or DELETE" error. To fix it, add `export const prerender = false;` to the `+server` file where it happens, or move the mutative methods in a different file.

## adapter-static

Instead of only checking the default for prerendering, the actual found routes are now checked for prerendering. If one of them is not prerenderable, an error is thrown unless you set the `fallback` option. If you get this error now, adjust your code so that 
- either the non-prerenderable routes are not crawled anymore
- or the non-prerenderable routes are made prerenderable
- or set the `fallback` option

# PR description

#6356.

* [x] differentiate between prerenderable and non-prerenderable endpoints
* [x] exclude routes with `prerender = true` from generated manifests
* [x] allow `'auto'` (prerender, but also include in manifests) as a value, alongside `true` (only prerender, exclude from manifests) and `false` (never prerender, include in manifests)
* [x] throw in adapter-static if any routes are `prerender = false` or `prerender = 'auto'` (or unknown, in the case of endpoints that are never fetched during prerendering)
* [x] docs 
* [x] idk, probably some tests and stuff

Not in scope for this PR: removing `config.kit.prerender.default` and using defaults from `+layout[.server].js` files. That can be added to #6197 once this is merged.

### Tests

### Changesets


documentation/docs/12-page-options.md
```diff
@@ -32,22 +32,29 @@ export const hydrate = false;
 
 ### prerender
 
-It's likely that at least some pages of your app can be represented as a simple HTML file generated at build time. These pages can be [_prerendered_](/docs/appendix#prerendering).
+It's likely that at least some routes of your app can be represented as a simple HTML file generated at build time. These routes can be [_prerendered_](/docs/appendix#prerendering).
 
-Prerendering happens automatically for any page with the `prerender` annotation:
+Prerendering happens automatically for any `+page` or `+server` file with the `prerender` annotation:
```
benmccann, line 37: should we add `JS` here to clarify that you can't add it to `+page.svelte`?
```suggestion
Prerendering happens automatically for any `+page` or `+server` JS file with the `prerender` annotation:
```

documentation/docs/12-page-options.md
```diff
@@ -32,22 +32,29 @@ export const hydrate = false;
...
-/// file: +page.js/+page.server.js
+/// file: +page.js/+page.server.js/+server.js
 export const prerender = false;
 ```
 
+Routes with `prerender = true` will be excluded from manifests used for dynamic SSR, making your server (or serverless/edge functions) smaller. In some cases you might want to prerender a route but also include it in the manifest (for example, you want to prerender your most recent/popular content but server-render the long tail) — for these cases, there's a third option, 'auto':
+
+```js
+/// file: +page.js/+page.server.js/+server.js
+export const prerender = 'auto';
```
benmccann, line 55: now that I've seen the docs and understand what this option does, I'll throw another name into the ring. how about `'partial'`?

documentation/docs/12-page-options.md
Rich-Harris: I was hoping the code sample would make it clear — adding JS muddies the waters a bit in the case where people are using TypeScript

documentation/docs/12-page-options.md
Rich-Harris: I think I prefer `auto` — `partial` doesn't have the right connotation to me, it sounds like the page is being partially prerendered

documentation/docs/12-page-options.md
dummdidumm: There are some other places where it's `+page/layout` with the `.js` ending. We should probably make that consistent, either adding the `.js` file always or never (except for when we introduce it).

packages/kit/types/private.d.ts
```diff
@@ -184,6 +184,10 @@ export interface PrerenderErrorHandler {
 
 export type PrerenderOnErrorValue = 'fail' | 'continue' | PrerenderErrorHandler;
 
+export type PrerenderOption = boolean | 'auto';
```
dummdidumm, line 187: This should be public API, people might want to do `export let prerender: PrerenderOption = ...` or `myHelperDefiningWhatPrerenderOptionToUse(): PreprenderOption {..}`

packages/kit/src/runtime/server/page/fetch.js
```diff
@@ -10,9 +10,10 @@ import { domain_matches, path_matches } from './cookie.js';
  *   options: import('types').SSROptions;
  *   state: import('types').SSRState;
  *   route: import('types').SSRRoute | import('types').SSRErrorPage;
+ *   prerender_default?: true | false | 'auto';
```
dummdidumm, line 13: This can reuse the PrerenderOption type

dummdidumm (approved):
Looks good! Will adjust the minor suggestions I made myself and merge to get the related PR in good state

documentation/docs/12-page-options.md
```diff
@@ -32,22 +32,29 @@ export const hydrate = false;
...
 
-It's likely that at least some pages of your app can be represented as a simple HTML file generated at build time. These pages can be [_prerendered_](/docs/appendix#prerendering).
+It's likely that at least some routes of your app can be represented as a simple HTML file generated at build time. These routes can be [_prerendered_](/docs/appendix#prerendering).
 
-Prerendering happens automatically for any page with the `prerender` annotation:
+Prerendering happens automatically for any `+page` or `+server` file with the `prerender` annotation:
 
 ```js
-/// file: +page.js/+page.server.js
+/// file: +page.js/+page.server.js/+server.js
```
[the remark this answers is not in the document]
dummdidumm: Skipping this for now, seing that it's like this for all others as well.

documentation/docs/12-page-options.md
```diff
@@ -32,22 +32,29 @@ export const hydrate = false;
...
 
 Alternatively, you can set [`config.kit.prerender.default`](/docs/configuration#prerender) to `true` and prerender everything except pages that are explicitly marked as _not_ prerenderable:
 
 ```js
-/// file: +page.js/+page.server.js
+/// file: +page.js/+page.server.js/+server.js
 export const prerender = false;
 ```
 
+Routes with `prerender = true` will be excluded from manifests used for dynamic SSR, making your server (or serverless/edge functions) smaller. In some cases you might want to prerender a route but also include it in the manifest (for example, you want to prerender your most recent/popular content but server-render the long tail) — for these cases, there's a third option, 'auto':
```
dummdidumm, line 51: It took me a little while to understand why this works - why can I prerender AND SSR a route - until I thought "oh yes with `[slug]`". I don't know how to express this better, but maybe we can come up with something later.

PR #6401, opened 2022-08-29
merge `resolve` options when using `sequence`

### LINKED ISSUE #4051: Consider how to pass options between handlers in sequence

Rich-Harris:
Closes #4051.

After this change, a `transformPageChunk` option passed to the first `handle` in a sequence will be respected by later calls to `resolve`

### Tests

### Changesets

### DIFF
--- a/.changeset/red-rice-bathe.md
+++ b/.changeset/red-rice-bathe.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] merge resolve options when using sequence helper
--- a/packages/kit/src/exports/hooks/sequence.js
+++ b/packages/kit/src/exports/hooks/sequence.js
@@ -7,19 +7,40 @@ export function sequence(...handlers) {
 	if (!length) return ({ event, resolve }) => resolve(event);
 
 	return ({ event, resolve }) => {
-		return apply_handle(0, event);
+		return apply_handle(0, event, {});
 
 		/**
 		 * @param {number} i
 		 * @param {import('types').RequestEvent} event
+		 * @param {import('types').ResolveOptions | undefined} parent_options
 		 * @returns {import('types').MaybePromise<Response>}
 		 */
-		function apply_handle(i, event) {
+		function apply_handle(i, event, parent_options) {
 			const handle = handlers[i];
 
 			return handle({
 				event,
-				resolve: i < length - 1 ? (event) => apply_handle(i + 1, event) : resolve
+				resolve: (event, options) => {
+					/** @param {{ html: string, done: boolean }} opts */
+					const transformPageChunk = async ({ html, done }) => {
+						if (options?.transformPageChunk) {
+							html = (await options.transformPageChunk({ html, done })) ?? '';
+						}
+
+						if (parent_options?.transformPageChunk) {
+							html = (await parent_options.transformPageChunk({ html, done })) ?? '';
+						}
+
+						return html;
+					};
+
+					// TODO remove post-https://github.com/sveltejs/kit/pull/6197
+					const ssr = options?.ssr ?? parent_options?.ssr;
+
+					return i < length - 1
+						? apply_handle(i + 1, event, { transformPageChunk, ssr })
+						: resolve(event, { transformPageChunk, ssr });
+				}
 			});
 		}
 	};
--- a/packages/kit/src/exports/hooks/sequence.js
+++ b/packages/kit/src/exports/hooks/sequence.js
@@ -1,13 +1,4 @@
 /**
- * Merge an array of multiple handlers into one. The `resolve` passed
- * to the first handler will call the second handler, and so on,
- * until the last handler calls the framework.
- *
- * If handlers use the `transformPageChunk` option, the functions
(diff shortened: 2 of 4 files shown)


benmccann:
are there docs that should be updated or added to?

Rich-Harris:
had a crack at that in https://github.com/sveltejs/kit/pull/6401/commits/3862c84e91c6a0fc54f6bfe1bfbaa7703336e7fc

benmccann:
How about https://kit.svelte.dev/docs/modules#sveltejs-kit-hooks ? I was guessing it would need to be updated

dummdidumm (commented):
It feels weird to me that `transformPageChunk` is not called in order (last is first instead of last). Is there a reason to do it like this?

Rich-Harris:
Ah, dangit — I thought adding the description to the source code would update the docs, forgot I needed to update `ambient.d.ts` instead. Would be great if the types came directly from the source code and we didn't need to `declare module` for `@sveltejs/kit/*`, not sure what it would look like to set that up.

> It feels weird to me that `transformPageChunk` is not called in order (last is first instead of last). Is there a reason to do it like this?

It's analogous to calling `add_a(add_b(add_c('x')))` — the result is `xcba` not `xabc`

dummdidumm:
> It's analogous to calling `add_a(add_b(add_c('x')))` — the result is `xcba` not `xabc`

Yes, but because it's an array I read this left-to-right/top-to-bottom, not the other way around. Is there a situation where switching this behavior is strange/wrong?
I would at least make it more clear in the JSDoc that the last one runs first.

Rich-Harris:
Is c7081b7950f5f2df6bd25bc822b48baff4583718 not clear enough? I'm not sure what more we can do.

> Is there a situation where switching this behavior is strange/wrong?

I mean... yeah! You'd literally be changing the semantics to `xabc`. I think it would be really weird if the `transformPageChunk` that's applied directly to the response coming from SvelteKit ran _after_ more distant transform functions. Think of it like this:

```js
// src/hooks.js
async function increment({ event, resolve }) {
  const response = await resolve(event, {
    transformPageChunk: ({ html }) => {
      return html.replace(/number: (\d+)/g, (_match, n) => `number: ${+n + 1}`);
    }
  });

  const number = response.headers.get('x-number');
  response.headers.set('x-number', String(+number + 1));

  return response;
}

async function double({ event, resolve }) {
  const response = await resolve(event, {
    transformPageChunk: ({ html }) => {
      return html.replace(/number: (\d+)/g, (_match, n) => `number: ${+n * 2}`);
    }
  });

  const number = response.headers.get('x-number');
  response.headers.set('x-number', String(+number * 2));

  return response;
}

export const handle = sequence(increment, double);
```

```js
// src/routes/+page.js
export function load({ setHeaders }) {
  const number = 100;

  setHeaders({ 'x-number': String(number) });

  return { number };
}
```

```svelte
<!-- src/routes/+page.svelte -->
<script>
  export let data;
</script>

<h1>number: {data.number}</h1>
```

The way it's currently implemented, `<h1>count: 100</h1>` would be transformed to `<h1>count: 201</h1>`, and `x-count` would be transformed from `100` to `201`, because `double` is applied to the response from SvelteKit followed by `increment`. 

If we reversed the order in which `transformPageChunk` was applied to make it `xabc` instead of `xcba`, you'd be left with an incorrect result — the header would say `201` but the HTML would say `202`. 

Unless, that is, we reversed the order altogether and did this...

```js
const handle = sequence(
  function a({ event, resolve }) {
    console.log('i am called last');
    return resolve(event); // calls SvelteKit
  },
  function b({ event, resolve }) {
    console.log('i am called in the middle');
    return resolve(event); // calls 'a'
  },
  function c({ event, resolve }) {
    console.log('i am called first');
    return resolve(event); // calls 'b'
  }
);
```

...but that seems deeply unintuitive.

dummdidumm (approved):
I see - yes this makes sense that way then. Thanks!

PR #6422, opened 2022-08-30
feat: Disallow import of user-designated server-only modules from the client

tcc-sejohnson:
Allow users to take advantage of the same dev- and run-time graph analysis that `$env/*/private` modules benefit from. Any module named `*.server.*` will be disallowed for client import. This does _not_ include modules inside `node_modules`.

TODO:
- [x] Stop re-resolving `lib` 
- [ ] Find a place to put the documentation (+ does it need more? Less? I wrote a draft down in the comments on this issue.)

### Tests

### Changesets


tcc-sejohnson:
Hah well, I've broken things. Will fix later.

tcc-sejohnson:
Oh shoot, I need to do docs. It's ready for review in a _technical_ sense.

tcc-sejohnson:
I thought about adding a section for this, but it seems like overkill, so I reverted the commit. Here's what I wrote. Not sure where we want to put it...

# Server-Only Modules

SvelteKit cares about security. When writing your backend and frontend in the same repository, it can be easy to accidentally import sensitive data into your front-end code (environment variables containing API keys, for example). SvelteKit provides a way to prevent this entirely: Server-Only Modules. We export a few of these ourselves: See the private [`$env`](modules#$env-dynamic-private) modules.

Thankfully, we don't keep all of this to ourselves. You can declare your module as server-only as well by placing it anywhere in your `$lib` directory and naming it `*.server.js` (or `ts`, etc.) Then, no matter how many hops you jump through, importing it into client code...

```js
// $lib/bad.server.js
export const shouldExplode = 'boom';

// $lib/hop-1.js
export { shouldExplode } from '$lib/bad.server.js';

// $lib/hop-2.js
export { shouldExplode } from '$lib/hop-1.js';

// $lib/hop-3.js
export { shouldExplode } from '$lib/hop-2.js';

// $lib/hop-4.js
export { shouldExplode } from '$lib/hop-3.js';

// src/routes/+page.svelte
<script>
	import { shouldExplode } from '$lib/hop-4.js';
</script>

<p>{shouldExplode}</p>
```

...will generate a helpful error.

```bash
[vite-plugin-svelte-kit] Cannot import $lib/bad.server.js into client-side code:
- .svelte-kit/generated/nodes/2.js
  - src/routes/+page.svelte
    - $lib/hop-4.js
      - $lib/hop-3.js
        - $lib/hop-2.js
          - $lib/hop-1.js
            - $lib/bad.server.js
```

Note: This feature works with simple dynamic imports (eg. `await import('$lib/bad.server.js')`), but it will not work with complex ones (``await import(`$lib/{myVar}`)``).

PatrickG:
Does this only work with the `$lib` folder?
I think with the new routing system that recommends using the `routes` folder for custom files, it would be good if it would work in the `routes` folder as well.

tcc-sejohnson:
Ah, good point.

dummdidumm:
I think I'd go one step further and say "everything in your app that is imported by client code and is not in node_modules should follow this rule". People might create other aliases than lib and it wouldn't be caught for them.

Rich-Harris:
I'd agree with that — `lib` isn't special in this regard

tcc-sejohnson:
Okay, I switched to "Is not in `node_modules` and matches the pattern", but I'm trying to think through edge cases... would we want to add a check to make sure the file is in the project directory, eg. "Is not in `node_modules` but _is_ somewhere inside the project directly _and_ matches the pattern"?

packages/kit/src/exports/vite/utils.js
```diff
@@ -336,8 +388,16 @@ function find_illegal_vite_imports(node, illegal_imports, module_types, seen = n
 		return [{ name, dynamic: false }];
 	}
 
+	console.log({ name, startsWith: name.startsWith(node_modules_dir) });
+
```
Rich-Harris, line 392: guessing this is left over?

```suggestion
```

Rich-Harris (approved):

Rich-Harris:
> would we want to add a check to make sure the file is in the project directory, eg. "Is not in `node_modules` but _is_ somewhere inside the project directly _and_ matches the pattern"?

implemented in https://github.com/sveltejs/kit/pull/6422/commits/2f16ba2054b82ae7679c43ef8d45eec3a734d7cd

packages/kit/src/exports/vite/utils.js
```diff
@@ -4,6 +4,27 @@ import { loadConfigFromFile, loadEnv, normalizePath } from 'vite';
...
+const illegal_imports = new Set([
+	'/@id/__x00__$env/dynamic/private', //dev
+	'\0$env/dynamic/private', // prod
+	'/@id/__x00__$env/static/private', // dev
+	'\0$env/static/private' // prod
+]);
+
+/** @param {string} id */
+function is_illegal(id) {
+	console.log({ id, cwd: process.cwd(), node_modules_dir });
```
Rich-Harris, line 16:
```suggestion
```

accuser:
👍🏼 Really useful addition to Svelte Kit. I wonder if this could be extended to disallow client import of modules from a configured path, e.g., `$lib/server/**`?

tcc-sejohnson:
> 👍🏼 Really useful addition to Svelte Kit. I wonder if this could be extended to disallow client import of modules from a configured path, e.g., `$lib/server/**`? 

I think if we were going to do that, it would have to be from one _specific_ location, i.e. `$lib/server`. Any other (more configurable) approach strikes me as becoming fairly complex fairly quickly.

accuser:
> I think if we were going to do that, it would have to be from one _specific_ location, i.e. `$lib/server`. Any other (more configurable) approach strikes me as becoming fairly complex fairly quickly.

Agree. I welcome the utility in the `*.server.*` naming, but this could become noisy. Having a single location under which I could organise modules that the client would be disallowed from importing would be extremely useful. 

I'm actually not sure if `$lib/server` is the right name, as others might prefer `$lib/services` or `$lib/private`.

dummdidumm:
If this becomes an option within the `files` object that could make it configurable while restricting it to not get out of hand.

tcc-sejohnson:
> If this becomes an option within the `files` object that could make it configurable while restricting it to not get out of hand.

From a configuration standpoint we could provide it in `config.kit.files`, but we'd probably want to assert that it's a subdirectory of `config.kit.files.lib`. That's technically more opinionated than we _have_ to be (we could allow folders anywhere in the project tree, as long as they're _in the project tree_), but it seems like the right kind of thing to be opinionated about...

tcc-sejohnson:
@accuser @dummdidumm Implemented in #6623

PR #6425, opened 2022-08-30
Fix `global is not defined` on vercel edge

### LINKED ISSUE #6424: **Adapter Vercel**: `global is not defined` when build for edge network
### Describe the bug

When a project depends on packages (like `@trpc/client`, for example) that reference the `global` variable and in the `adapter-vercel` config you pass the `edge: true` prop, throws a runtime error when a render is hit saying: `global is not defined`.

### Reproduction

Try to deploy to vercel targeting edge functions instead of lamdba a project that imports a lib that reference the nodejs `global` variable like `@trpc/client` at any serverside point, like `+layout.server.js` or `+layout.js`, and the app should crash.

### Logs

_No response_

### System Info

```shell
System:
    OS: Linux 5.17 undefined
    CPU: (4) x64 AMD A10-7300 Radeon R6, 10 Compute Cores 4C+6G
    Memory: 2.35 GB / 6.76 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 16.14.2 - ~/.nvm/versions/node/v16.14.2/bin/node
    npm: 8.5.0 - ~/.nvm/versions/node/v16.14.2/bin/npm
  Browsers:
    Firefox: 99.0
```

### Severity

### Additional Information

I will make a PR providing a fix for this bug in a late.

juanvillacortac:
Closes #6424

### Tests

### Changesets

### DIFF
--- a/packages/adapter-vercel/index.js
+++ b/packages/adapter-vercel/index.js
@@ -190,7 +190,7 @@ export default function ({ external = [], edge, split } = {}) {
 					format: 'esm',
 					external,
 					sourcemap: 'linked',
-					banner: { js: 'var global = globalThis;' }
+					banner: { js: 'globalThis.global = globalThis;' }
 				});
 
 				write(
--- a/packages/adapter-vercel/index.js
+++ b/packages/adapter-vercel/index.js
@@ -189,7 +189,8 @@ export default function ({ external = [], edge, split } = {}) {
 					platform: 'browser',
 					format: 'esm',
 					external,
-					sourcemap: 'linked'
+					sourcemap: 'linked',
+					banner: { js: 'var global = globalThis;' }
 				});
 
 				write(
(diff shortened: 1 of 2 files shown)


dummdidumm (approved):

packages/adapter-vercel/index.js
```diff
@@ -189,7 +189,8 @@ export default function ({ external = [], edge, split } = {}) {
 					platform: 'browser',
 					format: 'esm',
 					external,
-					sourcemap: 'linked'
+					sourcemap: 'linked',
+					banner: { js: 'var global = globalThis;' }
```
Rich-Harris, line 193: this is more future-proof

```suggestion
					banner: { js: 'globalThis.global = globalThis;' }
```

Rich-Harris:
Thanks — I don't love adding hacks like this to work around bugs in libraries, but since this bug is probably quite widespread we'll make an exception.

PR #6428, opened 2022-08-30
[chore] bump TS to 4.8

### LINKED ISSUE #6387: Upgrade to TypeScript 4.8
### Describe the problem

We're currently on TypeScript 4.7. We should probably update to TypeScript 4.8 while we're making a bunch of breaking changes

Run `pnpm -r update typescript`

I tried and running `pnpm check` on the kit project is causing some issue with the generated types that I don't understand saying that `await parent()` is returning `null`. I'm only a TypeScript beginner, so am hoping someone more knowledgeable might have some more insight

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

dummdidumm:
Fixes a breaking change introduced by TS: NonNullable<null> is no longer never when strict null checks are false: https://github.com/microsoft/TypeScript/issues/50519

Closes #6387
This isn't a breaking change because everything is backwards compatible and only new users of `create-svelte` get a higher version (but it works with the older one, too)

### Tests

### Changesets

### DIFF
--- a/.changeset/four-meals-learn.md
+++ b/.changeset/four-meals-learn.md
@@ -1,0 +1,6 @@
+---
+'create-svelte': patch
+'@sveltejs/kit': patch
+---
+
+[chore] bump ts version and ensure it works with latest changes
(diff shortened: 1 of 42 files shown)


benmccann (approved):
thanks for figuring this out!!

Issue #1931, opened 2021-07-17
Allow error pages with json bodies


Astavie:
### Describe the problem

I'm making a full-stack website in SvelteKit, and am now working on my API. This api is located under `/api/*`. When a user requests from an unknown URI I want to send a 404 message. As the API responds with json, I would like to respond in the same manner with the error message. However, to my knowledge, `__error.svelte` files only support html as an output.

### Describe the proposed solution

I can see two solutions

1. implementation of an `__error.json` file
2. an export in endpoints that get executed on all undefined http methods

I prefer number 2, as then I could also implement 405 Method Not Allowed errors.

### Alternatives considered

For now, I'm using an endpoint like this:

```javascript
function response() { return {
    status: 404,
    body: {
        "error": "notfound",
        "message": "Not Found"
    }
}}

// Most HTTP request methods
export const get = response;
export const post = response;
export const put = response;
export const patch = response;
export const del = response;
export const options = response;
export const head = response;
```

this is located under `src/routes/api/[...error].ts`

Although this works, it's pretty ugly imo and it doesn't cover all infinite custom http methods clients could be using.

### Importance

nice to have

Conduitry:
Something perhaps tidier that you can already do now is use a [`handle` hook](https://kit.svelte.dev/docs#hooks-handle) that returns the desired JSON instead of whatever `resolve()` returns.

I'm not sure how we'd want `__error.json` (or, more likely, `__error.js`) to behave. I'm not clear on what its API should be, what the file should `export`, or what it would be passed. As for what the function _returns_, presumably it would be the same as other endpoint functions (i.e., an object with `status`, `headers`, `body`, etc.) and if it returned falsy, rendering would fall back to `__error.svelte`. But it's not clear what should be passed to this function.

Astavie:
The handle hook practically solves my 404 issue (thanks for pointing me to it!)
For my other issue (giving a 405 error when a wrong http method is used) it would be the easiest if there was a generic `handle` function like the `handle` hook in endpoints that receives the request before passing it along to the function corresponding to the http method, kinda like this:

```typescript
export function get() {
    return {
        status: 200,
        body: { "hello": "world" }
    }
}

export async function handle({ request, resolve  }) {
    // Try to resolve the request like normal with one of the functions corresponding to an http method
    const response = await resolve(request)

    // If it returns falsy, it means we haven't given the http method a function
    if (!response) {
        const error = { "error": "badmethod", "message": "Method Not Allowed" }
[...]

PR #6434, opened 2022-08-30
[breaking] render raw response when unexpected error occurs in endpoint

### LINKED ISSUE #4801: Differing behavior of exception in shadow endpoint vs ssr
### Describe the bug

In hooks.js I wrap the `await resolve(event)` in a try/catch in order to do things like rollback a DB transaction.

This works fine when the exception is thrown from the shadow endpoint but the exception fails to bubble up to `handle()` when the same page is rendered on the server.

### Reproduction

```js
// hooks.js

export const handle = async ({ event, resolve }) => {
  let response
  try {
    response = await resolve(event);
  } catch (err) {
    console.log('@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@')
    console.log(err)
    throw err
  }
}
```

```js
// routes/endpoint.js

export const get = async () => {
  throw new Error("I should see some @'s")
}
```

### Logs

*From a full page load (SSR)*

```shell
I should see some @'s!
Error: I should see some @'s!
    at get (/private/tmp/my-app/src/routes/todos/index.ts:2:10)
    at load_shadow_data (file:///private/tmp/my-app/.svelte-kit/runtime/server/index.js:2530:25)
    at async load_node (file:///private/tmp/my-app/.svelte-kit/runtime/server/index.js:2127:5)
    at async respond$1 (file:///private/tmp/my-app/.svelte-kit/runtime/server/index.js:2804:15)
    at async resolve (file:///private/tmp/my-app/.svelte-kit/runtime/server/index.js:3309:11)
    at async Object.handle (/private/tmp/my-app/src/hooks.ts:5:13)
    at async respond (file:///private/tmp/my-app/.svelte-kit/runtime/server/index.js:3253:20)
    at async file:///private/tmp/my-app/node_modules/@sveltejs/kit/dist/chunks/index.js:294:24
```

*From a shadow endpoint fetch (`__data.json`)*

```
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Error: I should see some @'s!
    at get (/private/tmp/my-app/src/routes/todos/index.ts:4:9)
    at render_endpoint (file:///private/tmp/my-app/.svelte-kit/runtime/server/index.js:171:25)
    at resolve (file:///private/tmp/my-app/.svelte-kit/runtime/server/index.js:3286:24)
    at async Object.handle (//private/tmp/my-app/src/hooks.ts:6:16)
    at async respond (file:///private/tmp/my-app/.svelte-kit/runtime/server/index.js:3253:20)
    at async file:///private/tmp/my-app/node_modules/@sveltejs/kit/dist/chunks/index.js:294:24
```

### System Info

```shell
System:
    OS: macOS 12.2.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 240.88 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.6.0 - /opt/homebrew/bin/node
    Yarn: 1.22.18 - ~/.yarn/bin/yarn
    npm: 8.5.2 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 100.0.4896.127
    Safari: 15.3
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.324
    svelte: ^3.46.0 => 3.48.0
```

### Severity

### Additional Information

I'm not 100% sure how to work around it...

dummdidumm:
..instead of failing completely. Also catch possible fetch error (which would occur when network fails).
Closes #4801
Related to #1931 (maybe even closing it? Can't tell from reading the issue)

There are still some code paths which would bubble up to the handler, but these are of the kind "you use a wrong SvelteKit option" or during prerendering ("cannot use mutative methods .."). The other all look like they have a try-catch around code paths where errors could occur.

Not sure if a test is needed here. We could theoretically do one for the endpoint error.

This is a breaking change insofar as people could have relied on this behavior somehow.

### Tests

### Changesets


benmccann:
I've seen people asking to have their API endpoints return JSON when failing. I'm not sure we necessarily need to address that now, but thought I'd mention it since this seems like it could return HTML errors more often

packages/kit/src/runtime/server/endpoint.js
```diff
@@ -51,16 +54,30 @@ export async function render_endpoint(event, mod, state) {
...
+			negotiate(event.request.headers.get('accept') || 'text/html', [
+				'text/html',
+				'application/json'
+			]) === 'application/json';
+
+		return wants_json
+			? json(error_to_pojo(error, options.get_stack), {
+					status: error instanceof HttpError ? error.status : 500
+			  })
+			: new Response(error.message, { status: error instanceof HttpError ? error.status : 500 });
```
Rich-Harris, line 81: Is there a reason not to render the HTML error page here? Feels like if request doesn't specifically ask for JSON we should just do that.

An alternative could be to render the HTML response if `sec-fetch-dest` is `document`, otherwise JSON

packages/kit/src/runtime/server/index.js
```diff
@@ -315,7 +315,11 @@ export async function respond(request, options, state) {
 
 				// we can't load the endpoint from our own manifest,
 				// so we need to make an actual HTTP request
-				return await fetch(request);
+				try {
+					return await fetch(request);
+				} catch (e) {
+					return new Response(coalesce_to_error(e).message, { status: 500 });
+				}
```
Rich-Harris, line 322: By try-catching this, we prevent `handleError` from being called, which seems bad (though honestly I'm not sure how this _could_ fail — `fetch` only throws if there's a network (including CORS errors) which shouldn't really apply on the server. Is there a reason for this change?

dummdidumm: I was just being paranoid, looking at all code paths and seing if anything could theoretically fail, and fetch is such a code path. In the related issue it read "handle should _never_ fail" so I thought "ok let's try that then". Maybe it's easier to wrap the whole code into one more big try-catch.

packages/kit/src/runtime/server/endpoint.js
dummdidumm: Mhm true we can reuse the static error page here. I didn't know about `sec-fetch-dest`, I did what was done in a different but related code path where there's also a check to "is this a JSON request".

packages/kit/src/runtime/server/index.js
Rich-Harris: That's what I was thinking, yeah

Rich-Harris (approved):

PR #6442, opened 2022-08-30
Apply `data-sveltekit-*` attributes to children

### LINKED ISSUE #790: Add option to prefetch links by default
## Is your feature request related to a problem? Please describe.

There should be an option to prefetch links by default. In my opinion, page transitions should always be instantaneous if possible. And this is the default in other frameworks like Next.js and Gatsby.

Now you could just add `sveltekit:prefetch` to every single `<a>` tag, but there are some reasons why this isn't practical/possible.

- It's annoying
- You can easily forget some
- If you're parsing markdown, you may not be able to add the prefetch attribute
- When dealing with user-generated content (purely on the browser), again, it may not be possible to add the attribute
- If you're using 3rd-party components, you cannot add the attribute

And not only are faster transitions better user experience, in some industries (like eCommerce), the extra 150-500 milliseconds can mean millions of dollars of more revenue.

## Describe the solution you'd like
Add field to `svelte.config.cjs` to automatically prefetch pages. Perhaps something like this?

```
module.exports = {
  kit: {
    prefetchLinks: true,
  },
};
```

Now there are likely valid reasons to not prefetch certain links, in which case you also need a way to opt-out of prefetching. This example seems like intuitive syntax:

```svelte
<a sveltekit:prefetch={false} href="blog/what-is-sveltekit">What is SvelteKit?</a>
```

## Describe alternatives you've considered
As far as I know, the current best solution is to add `sveltekit:prefetch` to every single `<a>` tag.

Alternatively, you could call `prefetchRoutes(routes)` on every page, but this is wasteful (all routes will be prefetched) and may not work with dynamic content.

## How important is this feature to you?
I'll keep using Svelte/kit if this doesn't get added, but I'd be ecstatic if it gets added!

Rich-Harris:
Closes #790. With this PR, we can set `data-sveltekit-prefetch` to the entire app if we want (and opt out on a per-element or per-subtree basis).

### Tests

### Changesets


packages/kit/src/runtime/client/utils.js
```diff
@@ -24,17 +24,60 @@ export function scroll_state() {
...
-/** @param {HTMLAnchorElement | SVGAElement} node */
-export function get_href(node) {
-	return node instanceof SVGAElement
-		? new URL(node.href.baseVal, document.baseURI)
-		: new URL(node.href);
+/**
+ * @param {HTMLElement} element
+ * @param {string} name
+ */
+export function get_closest_attribute(element, name) {
```
mquandalle, line 71: I don't know if the native [element.closest() API](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) would work in this case?

Rich-Harris: Whoops, this function isn't actually used — removing it.

Do you know if `element.closest()` respects shadow DOM? We ended up having to use `event.composedPath()` because manual traversal bypasses shadow DOM nodes, annoyingly.

mquandalle: Haven't tried but apparently `element.closest()` [doesn't cross shadow DOM boundaries](https://gist.github.com/pinkhominid/ac09780a5dfd356f6a5831d61c09b626)

Rich-Harris: Yeah, just tried it and reached the same conclusion. Thanks for nothing once again, web component dweebs

packages/kit/src/runtime/client/client.js
```diff
@@ -1192,11 +1189,9 @@ export function create_client({ target, base, trailing_slash }) {
 				// 2. 'rel' attribute includes external
 				const rel = (a.getAttribute('rel') || '').split(/\s+/);
 
-				if (
-					a.hasAttribute('download') ||
-					rel.includes('external') ||
-					a.hasAttribute('data-sveltekit-reload')
-				) {
+				console.error(options);
+
```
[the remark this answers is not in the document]
Rich-Harris: seems like it did

documentation/docs/09-link-options.md
```diff
@@ -0,0 +1,66 @@
...
+Occasionally, we need to tell SvelteKit not to handle a link, but allow the browser to handle it. Adding a `data-sveltekit-reload` attribute to a link...
+
+```html
+<a data-sveltekit-reload href="/path">Path</a>
+```
+
+...will cause a full-page navigation when the link is clicked.
+
+Links with a `rel="external"` attribute will receive the same treatment. In addition, they will be ignored during [prerendering](/docs/page-options#prerender).
+
```
Conduitry, line 35: I know you say above that all of the `data-sveltekit-` attributes can also be specified on parent elements, but is it weird to only explicitly call it out for `-prefetch` and not `-reload` or `-noscroll`?

.changeset/fresh-lemons-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Move data-sveltekit-prefetch to <nav> element
```
Conduitry, line 5:
```suggestion
Move `data-sveltekit-prefetch` to `<nav>` element
```

documentation/docs/09-link-options.md
Conduitry: :+1: That's fair.

Conduitry (approved):

PR #6446, opened 2022-08-30
replace `router`/`hydrate` options with `csr`

### LINKED ISSUE #6436: Replace `hydrate` and `router` options with `csr`
### Describe the problem

At present, there are four 'page options' — `prerender`, `hydrate`, `router` and `ssr`. As of #6197, these can be set in `+layout[.server].js` files as well as `+page[.server].js` files, which allows us to get rid of the corresponding config options.

`router` is an anomaly, however. It's the only option that has a meaning after the initial render — if you navigate to a page with `router = false`, it will disable the client-side router altogether. There's no getting around it: this is a deeply weird option that basically no-one needs (to understand why it was introduced in the first place, see https://github.com/sveltejs/kit/pull/6197#issuecomment-1225726404).

As of https://github.com/sveltejs/kit/issues/790#issuecomment-1231052222, we'll have a much more sensible approach for disabling the router — `data-sveltekit-reload` applied to a top-level element. This would mean we can get rid of the `router` page option altogether.

This does leave a gap though — right now, if `hydrate` and `router` are both `false`, we don't send any JavaScript with the page at all.

We merge `hydrate` and `router` into a single `csr` option that controls whether client-side rendering/routing takes place. (It's convenient that the acronym works for both.) If `csr` is `false`, no JS is sent with the page.

Strictly speaking, this is less flexible than what we have now, but having one-but-not-both be `true` is both kinda useless and currently buggy. This approach would be much simpler, both in implementation and communication — `prerender`, `ssr` and `csr` are a much more logical trio.

### Alternatives considered

Keep the current semantics of `hydrate` and always include the client-side router.

### Importance

would make my life easier

### Additional Information

I'm raising this here as an issue rather than a discussion because I expect approximately three people to be affected by this change.

Rich-Harris:
closes #6436

TODO:

* [x] update `create-svelte` templates
* [x] docs
* [x] noisy error if a page has `hydrate` or `router` exports

### Tests

### Changesets


benmccann (approved):

MaximeMenotti:
Hi @Rich-Harris !
I use sveltekit in a micro-frontend like usage, injecting html and js generated by svelte-kit in another website. 
I used `router=false` option to avoid sveltekit to interfere with the site but keep hydration to activate some js interaction.  

I read in this PR that I should use this to keep same usage: 
> old: router = false / hydrate = true -> new: csr = true + set data-sveltekit-reload if you want native navigations

Unfortunately links like `<a href="/" data-sveltekit-reload>Homepage</a>` are still catch by sveltekit router. 
Is there any other way or is it planned to allow to disable only the router for all links as it was the case before? 
It was a very convenient feature for me, I hope I can find a solution. 
Thanks and sorry to reopen this topic.

lukaszpolowczyk:
@MaximeMenotti To me, this is simply a bug that needs to be reported separately.

PR #6461, opened 2022-08-31
[breaking] respect cache-control max-age on the client

### LINKED ISSUE #4625: Respect Cache-Control of serialized SSR fetch responses during subsequent client-side navigation
### Describe the problem

The fetch responses that are serialized as part of the SSR response are currently only used during the initial hydration of the app. If the same URLs are requested during subsequent client-side navigation, these requests are performed (at least) once more, regardless of the Cache-Control headers in the responses.

It would be nice if, even after the initial view is hydrated, the `fetch` function that's passed to `load` still checked with the serialized responses to see whether any of them are still valid. We should already have the information we need to make this decision, because the response headers for all of these requests are also serialized into the server-rendered HTML.

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

I have a fairly large chunk of data I want to expose throughout my app. When I request it in the root `__layout` `load` and expose it through `stuff`, this results in the data being requested again after navigating, regardless of the `Cache-Control` header in the response.

Currently, my workaround is to fetch the data during the `handle` hook and expose it through `session`. From here, the root `__layout` can then still expose it to the rest of the app with `stuff` without requiring widespread code changes.

dummdidumm:
...for initially fetched responses
Closes #4625
The proposed etag logic is not implemented because it sounds too complex and would add needless code for 99% of users

This is a breaking change insofar that people could have relied on the old behavior somehow

I implemented this after consulting the MDN docs and concluding that only the max-age and age headers are relevant for us in this case.

### Tests

### Changesets


Rich-Harris:
I think we can simplify the client-side logic a bit by doing more work on the server (i.e. adding a `data-ttl` attribute). Though in the process of looking into that I realised that we need to spruce up the serialization logic (e.g. it's sending `validation_errors` twice, once in the `hydrate` object and once as a `<script>`), so I'm going to do that in a separate PR first

Rich-Harris:
So this is interesting (and by 'interesting' I mean 'very annoying') — Chrome respects `max-age` and doesn't refetch until the resource is stale, but Firefox and Safari don't:

https://user-images.githubusercontent.com/1162160/187767331-f75c013e-53db-4241-a060-98c36b377ae1.mov

Which means that if we do this, we'll arguably be broken in Firefox and Safari.

Thoughts?

dummdidumm:
What the ... why don't they do that? Isn't that the most important part of the cache header?

Rich-Harris:
yeah, idk. very surprising to have different behaviour between browsers for something this fundamental

Rich-Harris:
Huh, I guess there's something else going on here. If I create this Node server...

```js
import { createServer } from 'http';

createServer((req, res) => {
	if (req.url === '/') {
		res.writeHead(200, { 'Content-Type': 'text/html' });
		res.end(`
			<!DOCTYPE html>
			<body>
				<h1>${new Date().toString()}</h1>
				<button>update</button>
				<script>
					document.querySelector('button').addEventListener('click', async () => {
						const res = await fetch('/time');
						document.querySelector('h1').innerText = await res.text();
					});
				</script>
			</body>
		`);
	} else if (req.url === '/time') {
		res.writeHead(200, {
			'cache-control': 'max-age=5'
		});
		res.end(new Date().toString());
	} else {
		res.writeHead(404);
		res.end('not found');
	}
}).listen(3000);
```

...then all three browsers behave the same way (i.e. respecting the cache header).

packages/adapter-static/test/apps/prerendered/vite.config.js.timestamp-1661975687543.mjs
```diff
@@ -0,0 +1,13 @@
...
...ata:application/json;base64,ewogICJ2ZXJzaW9uIjogMywKICAic291cmNlcyI6IFsidml0ZS5jb25maWcuanMiXSwKICAic291cmNlc0NvbnRlbnQiOiBbImNvbnN0IF9fdml0ZV9pbmplY3RlZF9vcmlnaW5hbF9kaXJuYW1lID0gXCIvVXNlcnMvcmljaC9EZXZlbG9wbWVudC9TVkVMVEUvS0lUL2tpdC9wYWNrYWdlcy9hZGFwdGVyLXN0YXRpYy90ZXN0L2FwcHMvcHJlcmVuZGVyZWRcIjtjb25zdCBfX3ZpdGVfaW5qZWN0ZWRfb3JpZ2luYWxfZmlsZW5hbWUgPSBcIi9Vc2Vycy9yaWNoL0RldmVsb3BtZW50L1NWRUxURS9LSVQva2l0L3BhY2thZ2VzL2FkYXB0ZXItc3RhdGljL3Rlc3QvYXBwcy9wcmVyZW5kZXJlZC92aXRlLmNvbmZpZy5qc1wiO2NvbnN0IF9fdml0ZV9pbmplY3RlZF9vcmlnaW5hbF9pbXBvcnRfbWV0YV91cmwgPSBcImZpbGU6Ly8vVXNlcnMvcmljaC9EZXZlbG9wbWVudC9TVkVMVEUvS0lUL2tpdC9wYWNrYWdlcy9hZGFwdGVyLXN0YXRpYy90ZXN0L2FwcHMvcHJlcmVuZGVyZWQvdml0ZS5jb25maWcuanNcIjtpbXBvcnQgeyBzdmVsdGVraXQgfSBmcm9tICdAc3ZlbHRlanMva2l0L3ZpdGUnO1xuXG4vKiogQHR5cGUge2ltcG9ydCgndml0ZScpLlVzZXJDb25maWd9ICovXG5jb25zdCBjb25maWcgPSB7XG5cdGJ1aWxkOiB7XG5cdFx0bWluaWZ5OiBmYWxzZVxuXHR9LFxuXHRwbHVnaW5zOiBbc3ZlbHRla2l0KCldXG59O1xuXG5leHBvcnQgZGVmYXVsdCBjb25maWc7XG4iXSwKICAibWFwcGluZ3MiOiAiO0FBQThhLFNBQVMsaUJBQWlCO0FBR3hjLElBQU0sU0FBUztBQUFBLEVBQ2QsT0FBTztBQUFBLElBQ04sUUFBUTtBQUFBLEVBQ1Q7QUFBQSxFQUNBLFNBQVMsQ0FBQyxVQUFVLENBQUM7QUFDdEI7QUFFQSxJQUFPLHNCQUFROyIsCiAgIm5hbWVzIjogW10KfQo=
```
Conduitry: I dunno why this happened again. Are you aborting tests midway and then committing temporary files? Should we add a gitignore for these to be safe?

Conduitry: Oh, you deleted it in a later commit. Still, what's creating these?

Rich-Harris: that's probably where they're coming from, yeah

Rich-Harris:
Ok, digging deeper: it seems that browsers invalidate the cache for a given URL when you POST to it, which seems like reasonable behaviour. I think this wasn't happening before in Chrome because the response was a 500 (because the `+server.js#POST` didn't return a `Response`). Now that that's fixed, the behaviour is the same across browsers.

But it means that _our_ behaviour is wrong — we also need to invalidate the cache if a non-GET request is made. It also means we'll need to split the test between two URLs

Rich-Harris:
Just realised this probably shouldn't apply during prerendering

Conduitry:
... I'm not sure what _should_ happen during prerendering. Do we want to permanently cache the responses from endpoints? If we're going to request them the second and later times we visit a page, why not the first time? I know there was an issue floating around about being able to prerender endpoints, and I don't know how that interacts with this.

Rich-Harris:
Endpoints _are_ prerendered, unless they're marked as unprerenderable, in which case prerendering fails. So a prerendered page wouldn't be able to benefit from the cache-control header on the data, but also the data wouldn't require recomputing anyway.

I don't think it makes sense to change this (i.e. allow a page to be prerendered without its data also being prerendered) because if the data can become stale between deploys, you shouldn't be prerendering the page in the first place.

Rich-Harris:
> Do we want to permanently cache the responses from endpoints? If we're going to request them the second and later times we visit a page, why not the first time?

The trick would be knowing when to invalidate it (i.e. when a deploy happened). I tend to think this is best left to CDNs — a 304 is better than a stale 200 from cache.

Rich-Harris (approved):

PR #6467, opened 2022-08-31
Remove sveltekit:start event

### LINKED ISSUE #6448: Remove `sveltekit:start` event
### Describe the problem

The only real reason we have [this](https://kit.svelte.dev/docs/events) is so that we can do some setup work in our test suite. It's a bit of a weird appendage — you can't really use it to do useful setup work, and it would probably be better to just use root layouts, even if it would add a tiny bit of repeated code to our tests.

Remove the event

### Alternatives considered

Leave the event

### Importance

nice to have

### Additional Information

_No response_

illuzan:
Closes #6448

### Tests

### Changesets

### DIFF
--- a/.changeset/perfect-cycles-smell.md
+++ b/.changeset/perfect-cycles-smell.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Remove sveltekit:start event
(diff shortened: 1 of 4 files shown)


Rich-Harris:
Thanks — this needed changes to the tests but I couldn't push them to this branch, so I'll close it in favour of #6484

PR #6468, opened 2022-08-31
tidy up data serialization

Rich-Harris:
Related to #6461 

* simplifies data serialization
* prevents `validation_errors` from being serialized twice (once as a `<script>`, once by `devalue`)
* uses `data-` attributes so that we generate compliant HTML

### Tests

### Changesets


packages/kit/src/runtime/client/fetcher.js
```diff
@@ -44,10 +44,10 @@ if (import.meta.env.DEV) {
 export function initial_fetch(resource, opts) {
 	const url = JSON.stringify(typeof resource === 'string' ? resource : resource.url);
 
-	let selector = `script[sveltekit\\:data-type="data"][sveltekit\\:data-url=${url}]`;
+	let selector = `script[data-sveltekit-fetched][data-url=${url}]`;
 
 	if (opts && typeof opts.body === 'string') {
-		selector += `[sveltekit\\:data-body="${hash(opts.body)}"]`;
+		selector += `[data-hash="${hash(opts.body)}"]`;
```
benmccann, line 50: should this be `data-sveltekit-hash` to match the `data-sveltekit-fetched` above?

benmccann:
one minor comment, but otherwise lgtm

packages/kit/src/runtime/client/fetcher.js
Rich-Harris: no, we only need the namespaced attribute for identification/deconflicting. once we've got the element, it's better to use `data-foo`

PR #6469, opened 2022-08-31
Form actions

### LINKED ISSUE #3533: `<Form>`
### Describe the problem

Best practices around form data are slightly annoying — we want to encourage the use of native `<form>` behaviour as far as possible (accessible, works without JS, etc) but the best UX involves optimistic updates, pending states, and client-controlled navigation.

So far our best attempt at squaring this circle is the `enhance` action [found in the project template](https://github.com/sveltejs/kit/blob/f2d64256aecc7419ed6a8d424ef00686dbf8b33a/packages/create-svelte/templates/default/src/lib/form.ts). It's a decent start, but aside from being tucked away in an example rather than being a documented part of the framework itself, actions are fundamentally limited in what they can do as they cannot affect SSR (which means that method overriding and CSRF protection will always be relegated to userland).

Ideally we would have a solution that

1. was a first-class part of the framework
2. enabled best practices _and_ best UX
3. worked with SSR

I propose adding a `<Form>` component. By default it would work just like a regular `<form>`...

```svelte
<script>
  import { Form } from '$app/navigation';
</script>

<Form action="/todos.json" method="post">
  <input name="description" aria-label="Add todo" placeholder="+ tap to add a todo">
</Form>
```

...but with additional features:

### Automatic invalidation

Using the [`invalidate`](https://kit.svelte.dev/docs#modules-$app-navigation) API, the form could automatically update the UI upon a successful response. In the example above, using the form with JS disabled would show the endpoint response, meaning the endpoint would typically do something like this:

```js
export async function post({ request, locals }) {
  const data = await request.formData();

  // write the todo to our database...
  await db.write('todos', locals.user, data.get('description'));

  // ...then redirect the user back to /todos so they see the updated page
  return {
    status: 303,
    headers: {
      location: '/todos'
    }
  };
}
```

If JS isn't disabled, `<Form>` would submit the result via `fetch`, meaning there's no need to redirect back to the page we're currently on. But we _do_ want the page to reflect the new data. Assuming (reasonably) that the page is showing the result of a `GET` request to `/todos.json`, `<Form>` can do this automatically:

```js
// (this is pseudo-code, glossing over some details)

const url = new URL(action, location);
url.searchParams.set(method_override.parameter, method);

const res = await fetch(url, {
  method: 'post',
  body: new FormData(form)
});

// if the request succeeded, we invalidate the URL, causing
// the page's `load` function to run again, updating the UI
if (res.ok) {
  invalidate(action);
}
```

### Optimistic UI/pending states

For some updates it's reasonable to wait for confirmation from the server. For others, it might be better to immediately update the UI, possibly with a pending state of some kind:

```svelte
<Form
  action="/todos.json"
  method="post"
  pending={({ data }) => {
    // we add a new todo object immediately — it will be destroyed
    // by one without `pending: true` as soon as the action is invalidated
    todos = [...todos, {
      done: false,
      text: data.get('text'),
      pending: true
    }];
  }}
  done={({ form }) => {
    // clear the form
    form.reset();
  }}
>
  <input name="description" aria-label="Add todo" placeholder="+ tap to add a todo">
</Form>

{#each todos as todo}
  <div class="todo" class:done={todo.done} class:pending={todo.pending}>
    <!-- todo goes here --> 
  </div>
{/each}

<style>
  .pending {
    opacity: 0.5;
  }
</style>
```

> this example glosses over one problem — typically you might generate a UUID on the server and use that in a keyed each block or something. ideally the pending todo would also have a UUID that the server accepted so that the invalidation didn't cause glitches. need to think on that some more

### Error handling

There's a few things that could go wrong when submitting a form — network error, 4xx error (e.g. invalid data) or 5xx error (the server blew up). These are currently handled a bit inconsistently. If the handler returns an explicit error code, the page just shows the returned `body`, whereas if an error is thrown, SvelteKit renders the error page. 

#3532 suggests a way we can improve error handling, by rendering a page with validation errors. For progressively enhanced submissions, this wouldn't quite work — invalidating the action would cause a `GET` request to happen, leaving the validation errors in limbo. We can pass them to an event handler easily enough...

```svelte
<Form
  action="/todos.json"
  method="post"
  pending={({ data }) => {...}}
  done={({ form }) => {...}}
  error={async ({ response }) => {
    ({ errors } = await response.json());
  }}
/>
```

...but we don't have a great way to _enforce_ correct error handling. Maybe we don't need to, as long as we provide the tools? Need to think on this some more.

### Method overriding

Since the component would have access to the [`methodOverride`](https://kit.svelte.dev/docs#routing-endpoints-http-method-overrides) config, it could override the method or error when a disallowed method is used:

```svelte
<Form action="/todos/{todo.uid}.json" method="patch">
  <input aria-label="Edit todo" name="text" value={todo.text} />
  <button class="save" aria-label="Save todo" />
</Form>
```

### CSRF

We still need to [draw the rest of the owl](#72):

image

I think `<Form>` has an important role to play here though. It could integrate with Kit's hypothetical CSRF config and automatically add a hidden input...

```svelte
<!-- <Form> implementation -->
<script>
  import { csrf } from '$app/env'; // or somewhere
</script>

<form {action} {method} on:submit|preventDefault={handler}>
  <input type="hidden" name={csrf.key} value={csrf.value}>
  <slot/>
</form>
```

...which we could then _somehow_ validate on the server. For example — this may be a bit magical, but bear with me — maybe we could intercept `request.formData()` and throw an error if CSRF checking (most likely using the double submit technique) fails? We could add some logic to our [existing response proxy](https://github.com/sveltejs/kit/blob/8a5f5de72c7b55a3ce230c50ac904b291096da77/packages/kit/src/runtime/server/page/load_node.js#L208):

```js
// pseudo-code
  get(response, key, _receiver) {
    if (key === 'formData') {
      const data = await response.formData();
      const cookies = cookie.parse(request.headers.get('cookie'));

      // check that the CSRF token the page was rendered with 
      // matches the cookie that was set alongside the page
      if (data.get(csrf.key) !== cookies[csrf.cookie]) {
        throw new Error('CSRF checking failed');
      }

      return data;
    }
  }

  // ...
});
```

This would protect a lot of users against CSRF attacks without app developers really needing to do anything at all. We _would_ need to discourage people from using `<Form>` on prerendered pages, of course, which is easy to do during SSR.

### Alternatives considered

The alternative is to leave it to userland. I don't _think_ I've presented anything here that requires hooks into the framework proper — everything is using public APIs (real or hypothetical). But I think there's a ton of value in having this be built in, so that using progressively-enhanced form submissions is seen as the right way to handle data.

This is a big proposal with a lot of moving parts, so there are probably a variety of things I haven't considered. Eager to hear people's thoughts.

### Importance

would make my life easier

### Additional Information

_No response_

dummdidumm:
# How to migrate

The `POST/PUT/PATCH/DELETE` methods on `+page.server.js` are replaced by an `actions` object. 

#### Step 1

If you have more than one form on that page, think about names that best describe it and add these names as keys to the new action object inside `+page.server.js`. Example:

```js
// +page.server.js
export const actions = {
  addTodo: () => {},
  editTodo: () => {}
};
```

If you only have one form, you don't need to give it a name, use `default`:

```js
// +page.server.js
export const actions = {
  default: () => {}
};
```

#### Step 2

Move your methods over to the respective names of the actions object. Adjust the return value accordingly:
- If you redirect, replace `return { location: 'X' }` with `throw redirect(303, 'X')`
- If you return errors replace `return { status: 400, errors: errors }` with `return invalid(400, errors)`
- In case of a success you can now also return arbitrary data if you want (it needs to be a JSON-stringifiable object)

```diff
// +page.server.js
+ import { redirect, invalid } from '@sveltejs/kit';

- export async function POST({ request }) {
+export const actions = {
+  default: async ({ request }) => {
      const values = await request.formData();
 
      const username = /** @type {string} */ (values.get('username'));
      const password = /** @type {string} */ (values.get('password'));
 
      const user = await db.findUser(username);
 
      if (!user) {
-        return {
-          status: 403,
-          errors: {
-            username: 'No user with this username'
-          }
-        };
+        return invalid(403, { username: 'No user with this username' });
      }
 
-      return {
-        location: '/'
-      };
+      throw redirect(303, '/');
    }
+};
```

#### Step 3

Update your `<form>` elements in `+page.svelte` (or components used by it), if you were previously using `PUT`/`PATCH`/`DELETE` and are now using named actions:

```diff
// +page.svelte
-<form method="POST" action="/todos?_method=PATCH">
+<form method="POST" action="?/editTodo">
```

The `/editTodo` is a normal query parameter. SvelteKit interprets a query parameter with a leading `/` as identifying a named action. `?/editTodo` will call the `editTodo` action in `export const actions = { editTodo: ..., ... }`

#### Step 4

Replace `export let errors` with `export let form` in `+page.svelte`:

```diff
// +page.svelte
-/** @type {import('./$types').Errors} */
-export let errors;
+/** @type {import('./$types').ActionData} */
+export let form;
```

This will contain the return value of the action that just ran (as long as it didn't error or redirect), allowing you to show a 'success' message or (in the case of validation errors) fill in previously submitted values and indicate which fields were invalid.

### If you used those methods for something else

Either adjust your urls and only make POST requests with `FormData` to it, ot move these methods over to a `+server.js` file. We will very soon make it possible to colocate a `+server.js` file next to a `+page.svelte` file. The return value needs to be a `Response` object; you can use the `json` helper method to make creation of that easier.

# PR description

Implements https://github.com/sveltejs/kit/discussions/5875
Closes #3533 

TODO
- [x] `export const actions`
- [x] remove `export let errors` in favor of `export let form`
- [x] migration hints
- [x] migration docs
- [x] docs

### Tests

### Changesets


packages/kit/src/runtime/control.js
```diff
@@ -31,3 +31,28 @@ export class Redirect {
...
 	}
 }
+
+export class ValidationError {
+	/**
+	 * @param {number} status
+	 * @param {Record<string, any>} errors
+	 * @param {FormData | Record<string, any> | null | undefined} [values]
+	 */
+	constructor(status, errors, values) {
```
Rich-Harris, line 41: I was thinking `(status, fields, errors)` would be preferable in the common case where you're just passing `fields` back, because this...

```js
throw invalid(400, fields, {
  password: 'Your password is too secure'
});
```

is more natural than

```js
throw invalid(400, {
  password: 'Your password is too secure'
}, fields);
```

which Prettier would turn into this monstrosity:

```js
throw invalid(
  400,
  {
    password: "Your password is too secure",
  },
  fields
);
```

packages/kit/src/exports/index.js
```diff
@@ -43,3 +43,13 @@ export function json(data, init) {
...
 	});
 }
+
+/**
+ * Generates a `ValidationError` object.
+ * @param {number} status
+ * @param {Record<string, any>} errors
+ * @param {FormData} [form]
+ */
+export function validation(status, errors, form) {
```
Rich-Harris, line 53: what about `throw invalid(...)`? I think that better communicates 'something went wrong'

```suggestion
export function invalid(status, errors, form) {
```

packages/kit/src/runtime/control.js
```diff
@@ -31,3 +31,28 @@ export class Redirect {
...
+		if (values instanceof FormData) {
+			/** @type {Record<string, string>} */
+			const converted = {};
+			for (const [key, value] of values.entries()) {
+				if (typeof value === 'string') {
+					converted[key] = value;
+				}
+			}
```
Rich-Harris, line 49: this is buggy — it won't handle this case:

```html
<form>
  <input name="x">
  <input name="x">
</form>
```

I don't think we should accept a `Record` here. It makes it seem like you can return nested fields, and fields that aren't strings, etc. Clarity over convenience, always

packages/kit/src/runtime/server/page/actions.js
```diff
@@ -0,0 +1,197 @@
...
+			});
+		} else {
+			return json({ status: 200, result });
+		}
+	} catch (e) {
+		const error = /** @type {Redirect | HttpError | ValidationError | Error} */ (e);
+
+		if (error instanceof Redirect) {
+			return json({ status: 303, location: error.location });
+		}
```
Rich-Harris, line 61: We can't send 303s in the JSON case, because then the client has no idea what the URL is. I think we should send 200 responses with the relevant details encoded in the payload, and reserve non-200 status codes for non-validation errors

dummdidumm: This is a json 200 response with payload status/location

packages/kit/src/exports/index.js
dummdidumm: yeah that's better

packages/kit/src/runtime/control.js
dummdidumm: ok, switching it

packages/kit/src/runtime/control.js
dummdidumm: We need to return JSON though since in the progressive enhancement case the values need to be transferred this way. Seperatley but related, I think it would be nicer if `$form.values` is a POJO and not `FormData`; `values.get('x')`, you can do `values.x` instead, which feels better to me.

packages/kit/src/runtime/server/page/actions.js
Rich-Harris: 🤦 this is what i get for reviewing code while tired

packages/kit/src/runtime/control.js
Rich-Harris: We need to return JSON, but it needs to include all the data, and since any field can be duplicated it means everything needs to be an array of strings even in the common case where there's a single value, and I think this...

```js
$form.values.get('thing')
```

...is probably nicer than this:

```js
$form.values.thing[0]
```

I've seen APIs that do this...

```js
for (const [key, value] of values) {
  if (converted[key]) {
    if (typeof converted[key] === 'string') {
      converted[key] = [converted[key], value];
    } else {
      converted[key].push(value);
    }
  } else {
    converted[key] = value;
  }
}
```

...and it's a disaster when you have the 'could be one, could be multiple' case. I don't know. There isn't a great answer here.

documentation/docs/06-form-actions.md
```diff
@@ -0,0 +1,254 @@
...
+First we need to ensure that the page is _not_ reloaded on submission. For this, we prevent the default behavior. Afterwards, we run our JavaScript code instead which does the form submission through `fetch` instead.
+
+```svelte
+/// file: src/routes/login/+page.svelte
+<script>
+	import { submitted } from '$app/stores';
+	import { invalidateAll, goto } from '$app/navigation';
+
+	async function login(event) {
+		event.preventDefault(); // prevent native form submission
```
ak4zh, line 161: Why are we calling `preventDefault` again here?
The form already has `on:submit|preventDefault={login}`

dummdidumm: good catch, this should be deduplicated - not sure which way though. `|preventDefault` is more Svelte-like, but as a reader you may not have seen that before, and with `event.preventDefault` we can put an explanatory comment next to it - so I probably lean towards the latter.

ak4zh: I feel `|preventDefault` looks clean and makes sense for docs to use more of svelte way.
But yeah we loose comment option.

packages/kit/src/runtime/control.js
dummdidumm: Since we decided to allow to return arbitrary data, this is no longer relevant - resolving.

packages/kit/test/apps/basics/test/server.test.js
```diff
@@ -272,15 +272,18 @@ test.describe('Routing', () => {
 });
 
 test.describe('Shadowed pages', () => {
-	test('Action can return undefined', async ({ request }) => {
+	// TODO this doesn't work anymore because we call formData on the user's behalf,
+	// and posting FORM data will trigger the CSFR 403, which also check for in other tests in this app
```
Rich-Harris, line 276: i don't totally understand this, so i'm leaving a comment as a note to self to dig into it

```suggestion
	// TODO this doesn't work anymore because we call formData on the user's behalf,
	// and posting FORM data will trigger the CSRF 403, which also check for in other tests in this app
```

packages/kit/types/ambient.d.ts
```diff
@@ -80,6 +80,56 @@ declare module '$app/environment' {
...
+			 * Called when submission was successful and no redirect happens. If not set, will update the `form` property and call `invalidateAll`
+			 */
+			result?: (input: { data: FormData; response: Success; form: HTMLFormElement }) => void;
+		}
+	): { destroy: () => void };
+
+	/**
+	 * This action updates the `form` property of the current page with the given data.
+	 */
+	export function updateForm<Data extends Record<string, unknown> | null = Record<string, any>>(
```
Rich-Harris, line 128: what if `updateForm` took a `MutationResult`, so that a) you could just call it with the JSON response from an action, and b) it updated `$page.status` as well?

Rich-Harris (commented):
this is bad-ass. comments inline

packages/kit/src/core/sync/write_types/test/layout/_expected/$types.d.ts
```diff
@@ -23,7 +23,7 @@ export type PageServerLoad<
 		| void
 > = Kit.ServerLoad<RouteParams, PageServerParentData, OutputData>;
 export type PageServerLoadEvent = Parameters<PageServerLoad>[0];
-export type Errors = null;
+export type FormData = unknown;
```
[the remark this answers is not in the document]
Rich-Harris: updated this. _slightly_ weird that `export let form` has type `ActionData` but definitely less confusing. open to alternatives

packages/kit/src/runtime/client/client.js
```diff
@@ -881,7 +884,9 @@ export function create_client({ target, base, trailing_slash }) {
 			branch,
 			status: 200,
 			error: null,
-			route
+			route,
+			// On invalidateAll/same site navigation, form prop stays, else it's reset
+			form: current.url?.href !== url.href ? null : undefined
```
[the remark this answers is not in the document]
Rich-Harris: changed this

packages/kit/src/runtime/client/client.js
```diff
@@ -1161,6 +1166,28 @@ export function create_client({ target, base, trailing_slash }) {
...
 		},
 
+		update_form: async (form) => {
+			if (import.meta.env.DEV) {
+				// Nasty hack to silence harmless warnings the user can do nothing about
+				const warn = console.warn;
+				console.warn = (...args) => {
+					if (
+						args.length !== 1 ||
+						!/<(Layout|Page)(_[\w$]+)?> was created with unknown prop '(data|errors)'/.test(args[0])
```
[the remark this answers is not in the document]
Rich-Harris: fixed this and also DRYed out the code for squelching these warnings

Rich-Harris:
Recapping comments above, since GitHub's world-class UI hides actual conversation in favour of this sort of thing:

---

image

---

I think we're really close. I think it probably makes sense if `updateForm` takes a `MutationResult` instead of some data, and

1. updates `$page.status`
2. renders the correct `+error` page (we can assume, if `updateForm` is being called, that the form in question belongs to the current page — on which note, the `enhance` default should skip `updateForm` if not)
3. (maybe?) redirects if `type === 'redirect'`

I think the `enhance` action can be simplified. The nicest API would be if it took a single callback, which returned a function that handled the response:

```svelte
<form
  use:enhance={({ form, data, cancel }) => {
    // add pending/optimistic UI

    return async (result: MutationResult) => {
      if (result.type === 'success') await invalidateAll();
      updateForm(result);

      // undo pending/optimistic UI
    };
  }}
>
```

If we're paranoid about making it future-proof then we could make it `use:enhance={{ submit }}` as I suggested above, but if `updateForm` is capable of causing a nested `+error` page to be shown then I daresay that's unnecessary.

> maybe `updateForm` is the wrong name, if it's doing this stuff — maybe `applyMutation` or `applyResult` or something like that would be better? also, if it's going to be able to trigger the nested `+error` then it will need to become an `async` function

I noticed that optimistic updates on the demo app aren't working — it seems that binding to a property of an object inside an array inside `data` doesn't work. Not sure if something changed in Svelte. Anyway, just making a note that we'll need an alternative approach (but we need to completely overhaul the demo app anyway, so shouldn't spent too much time on it.)

packages/kit/src/runtime/server/page/actions.js
```diff
@@ -0,0 +1,153 @@
...
+
+	if (!actions) {
+		maybe_throw_migration_error(server);
+		// TODO should this be a different error altogether?
+		return new Response('POST method not allowed. No actions exist for this page', {
+			status: 405,
+			headers: {
+				// https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405
+				// "The server must generate an Allow header field in a 405 status code response"
+				allow: 'GET'
```
dummdidumm, line 39: There's no GET request that accepts JSON under this url. Does it still count even if the GET only applies to a text/html request?

dummdidumm (commented):
Suggested changes to the enhance API sound good, I'll adjust this. About `updateForm` - I'm not sure yet. First, it wouldn't be `MutationResult` though, the JSON response type is `FormFetchResponse` - does that change anything for you? Second, `status` is set to 204 in case of a success without data, which would be weird to set on the page. Third, it might result in unexpected page store updates, although we can mitigate this by checking if the status would be changed and only set it then.

david-plugge:
In the [multiple forms section](https://kit-svelte-8i9nryp39-svelte.vercel.app/docs/form-actions#multiple-forms) there is missing 
```js
/** @type {import('./$types').FormData} */
export let form;
```

```html
<script>
  import { enhance } from '$app/forms';
</script>

<ul>
  {#each todos as todo}
    <li>
      <form method="post" use:enhance={{ invalid: () => todo.invalid = true }}>
        <span>{todo.text}</span>
        <!-- use form?.invalid for the native form submission, use todo.invalid for progressive enhancement -->
        {#if form?.invalid || todo.invalid}
          <span>Can't be deleted yet</span>
        {/if}
        <button>Remove</button>
      </form>
    <li>
  {/each}
<ul>
```

 Does this actually work? If js is disabled, the error would still show up on every todo. At least it's not clear to me why it would work.
My intuition would be to add the todo ID to the form data to identify the correct form.

dummdidumm:
You are right, the example is flawed and it could also be solved by looking for the correct ID. I was trying to come up with a simple enough example to show that you can't always use it, like when you have multiple concurrent submissions going, but this example obviously isn't enough (which is good, because it shows you can use `form` for pretty long until you need something else).

Rich-Harris:
Ah yep, I missed that `FormFetchResponse` and `MutationResult` are separate on first reading. I don't totally grok why that's the case — what if we had a single `ActionResult` type that did both jobs?

```ts
type ActionResult =
  | { type: 'success'; status: number; data?: Record<string, any> }
  | { type: 'invalid'; status: number; data?: Record<string, any> }
  | { type: 'redirect'; location: string }
  | { type: 'error', error: { message: string } } // https://github.com/sveltejs/kit/discussions/6499
```

dummdidumm:
`MutationResult` is just an interal helper type to not have three separate variables of which only one can be set at the same time in `runtime/server/page/index.js`. We could deduplicate it though and use the same type for both, although `renderer.js` only is really interested in `data`.

packages/kit/src/runtime/server/page/actions.js
```diff
@@ -0,0 +1,153 @@
...
+			}
+		});
+	}
+
+	try {
```
cdcarson, line 44: I get that the "success and stay on page" case and the "invalid form" case are similar from the framework standpoint (i.e., in this try/catch.) Thus, `return`ing the error rather than `throw`ing. But it's inconvenient on the userland side. It makes it less easy to split out logic into functions that `return` something on success and `throw` errors.  

For example, I'd really rather do this...

```ts
/**
 * @throws ValidationError 
 */
const validateSignInForm = (raw: Record<string, unknown>): SignInFormData => {
  // perform basic validation, return the shape
}
/**
 * @throws ValidationError 
 */
 const signIn = async (event: RequestEvent, data: SignInFormData): Promise<LoginResult> => {
  // check email and password in the db, read/set cookies, return result
}
export const actions = {
  default: async (event: RequestEvent) => {
    const data =  validateSignInForm(myFormDataToPojo(await event.request.formData()));
    const result = await signIn(event, data);
    setFlashMessage(event, `Welcome, ${result.user.profile.name}.`);
    return redirect(303, result.authRedirect)
  }
}

```
...than this (using the same helpers)...

```ts
export const actions = {
  default: async (event: RequestEvent) => {
    let data: SignInFormData;
    try {
      data =  validateSignInForm(myFormDataToPojo(await event.request.formData()));
    } catch (error) {
      if (error instanceof ValidationError) {
        return error
      }
      throw error;
    }
    let result: LoginResult
    try {
      result = await signIn(event, data);
      
      
    } catch (error) {
      if (error instanceof ValidationError) {
        return error;
      }
      throw error;
    }

    // I think this has to be here since we can't put the throw in the second catch above...
    setFlashMessage(event, `Welcome, ${result.user.profile.name}.`);
    throw redirect(303, result.authRedirect)
   
  }
}
```

On the face of it this can be fixed pretty easily here. I know there was some concern about type safety if ValidationError is thrown rather than returned, but I think it's worth figuring that out separately rather than making the developer jump. through hoops and going against the convention.

dummdidumm: The types can't be infered if you `throw invalid`, which means your type safety would be lost. Only infering them for the success case goes only half-way. I can understand the desire to `throw` instead though. There's not perfect solution that makes all sides happy, unfortunately.

Rich-Harris: We don't need to make `ValidationError` importable — it's all in userland. As long as both the action and the `validateSignInForm` and `signIn` functions are all referencing the same thing, it's fine. Just need to do this:

```diff
} catch (e) {
-  if (e instanceof ValidationError) {
-    return e;
+  if (e instanceof MyValidationError) {
+    return invalid(e.status, e.data);
  }
  throw e;
}
```

If this is something you need to do in multiple actions, it can easily be abstracted out into a function.

This might seem inconvenient in the example above, but there are other patterns that don't involve try-catch:

```js
const [errors, data] = validateSignInForm(myFormDataToPojo(await event.request.formData()));
if (errors) return invalid(400, errors);
```

In general we don't want to encourage overuse of `throw` as a control flow mechanism — it makes code harder to follow. Returning validation errors does make more sense from the framework standpoint, but I would also argue it makes more sense from the app standpoint — rather than having three types of errors...

* `throw error(...)` causes the `+error` page to render
* `throw new Error(...)` causes the `+error` page to render, but also triggers `handleError`
* `throw invalid(...)` renders the same page as would have been rendered if nothing was thrown

...we have a simple rule that applies to `load` and `actions` and handlers in `+server.js` — if you `throw` (either an error or a redirect), the page will not be rendered; if you `return`, it will. This is a case where I'd strongly argue for clarity over convenience.

cdcarson: Ok. Got it. Just my ingrained mantra of "throw error, return success." I can deal with it, either as is or by making a wrapper. Thanks.

packages/kit/src/exports/index.js
```diff
@@ -1,4 +1,4 @@
-import { HttpError, Redirect } from '../runtime/control.js';
+import { HttpError, Redirect, ValidationError } from '../runtime/control.js';
```
cdcarson, line 1: One minor request, based on a couple of weeks working with the new helpers: can we make the status code the final argument, with a sensible default? So... 

```ts
export const redirect = (location: string, status = 303) => {...}; // not quite sure what'd be  the correct default
export const error = (message: string, status = 500) => {...};
export const invalid<T> = (data: T, status = 400) => {...};
```
If not, fine.

Rich-Harris: Prettier will butcher `invalid(data, status)`:

image

It won't do that for strings, but in the interests of consistency (and also encouraging people not to blindly use the defaults — and as you say, there isn't a plausible default for `redirect`) it's better to have the status code first

fjdumont: I prefer the leading status code because it clearly communicates the intent. Having a "sane" default might also get on your way. "Explicitly over magic" or something along the lines. Sorry for butchering the messages I picked up in your comments, Rich. ;-)

Each response has to have a status code, not a body or message though. Having an optional parameter forces it to the tail of the parameter list, which closes the API to future changes/improvements.

Also, I usually don't care too much about the message if the status code clearly communicates what's going on. 

Just my two cents, thanks for listening to my Ted talk.

cdcarson: Ok. Got it. Valid points. Yet another chapter ripped out of  "Chris Carson's Big Book of Code Style, As Divinely Revealed to Chris Carson by Chris Carson" and used to light the barbecue, but I'll survive.

.changeset/spicy-pugs-applaud.md
```diff
@@ -0,0 +1,6 @@
+---
+'create-svelte': patch
+'@sveltejs/kit': patch
+---
+
+[breaking] Implement form actions in favor of POST/PUT/PATCH/DELETE page endpoint methods
```
Rich-Harris, line 6:
```suggestion
[breaking] Replace `POST`/`PUT`/`PATCH`/`DELETE` in `+page.server.js` with `export const actions`
```

packages/kit/src/runtime/app/forms.js
```diff
@@ -0,0 +1,85 @@
...
+function guard(name) {
+	return () => {
+		throw new Error(`Cannot call ${name}(...) on the server`);
+	};
+}
+
+const ssr = import.meta.env.SSR;
+
+/** @type {import('$app/forms').applyAction} */
+export const applyAction = ssr ? guard('updateForm') : client.apply_action;
```
Rich-Harris, line 16:
```suggestion
export const applyAction = ssr ? guard('applyAction') : client.apply_action;
```

packages/kit/src/runtime/client/client.js
```diff
@@ -1161,6 +1138,32 @@ export function create_client({ target, base, trailing_slash }) {
 			await Promise.all(promises);
 		},
 
+		apply_action: async (result) => {
+			if (result.type === 'error') {
+				const url = new URL(location.href);
+				const intent = get_navigation_intent(url, true);
+				forced_error = result.error;
```
Rich-Harris, line 1145: this feels racy — probably better if we can pass it as an argument to `update`

packages/kit/src/runtime/server/page/actions.js
```diff
@@ -0,0 +1,179 @@
...
+import { error_to_pojo } from '../utils.js';
+
+// Info: `enhance` action and `updateForm` live in `runtime/app/forms.js`
+
+/** @param {import('types').RequestEvent} event */
+export function is_action_json_request(event) {
+	const accept = negotiate(event.request.headers.get('accept') || 'text/html', [
+		'text/html',
+		'application/json'
+	]);
```
Rich-Harris, line 14: this should probably default to JSON, since `fetch` has `*/*` and a document request will always favour HTML

```suggestion
	const accept = negotiate(event.request.headers.get('accept') || '*/*', [
		'application/json',
		'text/html'
	]);
```

packages/kit/src/runtime/server/page/actions.js
Rich-Harris: based on my understanding of [RFC 9110](https://httpwg.org/specs/rfc9110.html#status.405), yes. it's not 100% clear but I'm pretty sure you're supposed to return a list of methods a resource supports regardless of `accept`

packages/kit/src/runtime/server/page/index.js
```diff
@@ -68,42 +50,22 @@ export async function render_page(event, route, page, options, state, resolve_op
 
 		let status = 200;
 
-		/** @type {HttpError | Error} */
-		let mutation_error;
+		/** @type {import('$app/forms').SubmissionResult | undefined} */
```
Rich-Harris, line 53: I think `ActionResult` is probably a clearer name

```suggestion
		/** @type {import('$app/forms').ActionResult | undefined} */
```

packages/kit/src/runtime/server/page/render.js
```diff
@@ -188,9 +190,9 @@ export async function render_response({
 		throw error;
 	}
 
-	if (validation_errors) {
+	if (form_value) {
 		try {
-			serialized.errors = devalue(validation_errors);
+			serialized.form = devalue(form_value);
```
Rich-Harris, line 195: I guess we need to enforce this being JSONable, not sure if that's already happening elsewhere

packages/kit/types/ambient.d.ts
```diff
@@ -80,6 +80,64 @@ declare module '$app/environment' {
 	export const prerendering: boolean;
 }
 
+/**
+ * ```ts
+ * import { enhance, updateForm } from '$app/forms';
```
Rich-Harris, line 85:
```suggestion
 * import { enhance, applyAction } from '$app/forms';
```

packages/kit/types/ambient.d.ts
```diff
@@ -209,7 +267,7 @@ declare module '$app/paths' {
  * In the browser, we don't need to worry about this, and stores can be accessed from anywhere. Code that will only ever run on the browser can refer to (or subscribe to) any of these stores at any time.
  */
 declare module '$app/stores' {
-	import { Readable } from 'svelte/store';
+	import { Readable, Writable } from 'svelte/store';
```
Rich-Harris, line 270:
```suggestion
	import { Readable } from 'svelte/store';
```

packages/kit/src/core/sync/write_types/test/simple-page-server-only/_expected/$types.d.ts
```diff
@@ -20,11 +20,15 @@ export type PageServerLoad<
 	OutputData extends OutputDataShape<PageServerParentData> = OutputDataShape<PageServerParentData>
 > = Kit.ServerLoad<RouteParams, PageServerParentData, OutputData>;
 export type PageServerLoadEvent = Parameters<PageServerLoad>[0];
-export type Errors = null;
+export type ActionData = Kit.AwaitedActions<
+	typeof import('../../../../../../../../+page.server.js').actions
```
Rich-Harris, line 24: i don't really understand why there are so many `../` segments, but if the tests pass then i'm gonna pretend i didn't see it 🙈

packages/kit/src/core/sync/write_types/test/simple-page-server-only/_expected/$types.d.ts
```diff
@@ -20,11 +20,15 @@ export type PageServerLoad<
...
+>;
 export type PageServerData = Kit.AwaitedProperties<
 	Awaited<ReturnType<typeof import('../../../../../../../../+page.server.js').load>>
 >;
 export type PageData = Omit<PageParentData, keyof PageServerData> & PageServerData;
 export type Action = Kit.Action<RouteParams>;
+export type Actions = Kit.Actions<RouteParams>;
+export type ActionEvent = Kit.RequestEvent<RouteParams>;
```
Rich-Harris, line 32: Do we need a new type here, given that it's just a `RequestEvent`? Feels like we could just the same `RequestEvent<Params>` we generate for `+server.js` files

packages/kit/src/runtime/app/forms.js
```diff
@@ -0,0 +1,85 @@
...
+	const fallback_callback = (result) => {
+		if (
+			(result.type === 'success' || result.type === 'invalid') &&
+			location.origin + location.pathname === form.action.split('?')[0]
+		) {
+			applyAction(result);
+		}
+
+		if (result.type === 'success') {
+			invalidateAll();
```
Rich-Harris, line 30: this should happen before `applyAction`, so that the page updates in one go

packages/kit/src/runtime/app/forms.js
```diff
@@ -0,0 +1,85 @@
...
+		} else if (result.type === 'redirect') {
+			goto(result.location);
+		} else if (result.type === 'error') {
+			console.error(result.error);
```
Rich-Harris, line 34: we should be calling `applyAction` here

packages/kit/src/runtime/app/forms.js
```diff
@@ -0,0 +1,85 @@
...
+		}
+	};
+
+	/** @type {unknown} */
+	let current_token;
```
Rich-Harris, line 39: this feels like something that should be left to userland — we shouldn't be preventing multiple submissions as a blanket policy

packages/kit/src/runtime/server/page/actions.js
```diff
@@ -0,0 +1,179 @@
...
+
+	return (
+		accept === 'application/json' &&
+		event.request.method !== 'GET' &&
+		event.request.method !== 'HEAD'
```
Rich-Harris, line 19: the joys of getting rid of method overrides

```suggestion
		event.request.method === 'POST'
```

packages/kit/src/runtime/server/page/actions.js
```diff
@@ -0,0 +1,179 @@
+import { error, json } from '../../../exports/index.js';
+import { normalize_error } from '../../../utils/error.js';
+import { negotiate } from '../../../utils/http.js';
+import { HttpError, Redirect, ValidationError } from '../../control.js';
+import { error_to_pojo } from '../utils.js';
+
+// Info: `enhance` action and `updateForm` live in `runtime/app/forms.js`
+
```
Rich-Harris, line 8: i don't think we need this 😀 

```suggestion
```

packages/kit/types/index.d.ts
```diff
@@ -324,15 +320,28 @@ export interface ServerLoadEvent<
...
 
+export type Actions<
+	Params extends Partial<Record<string, string>> = Partial<Record<string, string>>,
+	OutputData extends Record<string, any> | void = Record<string, any> | void
+> = Record<string, Action<Params, OutputData>>;
+
+/**
+ * When calling a form action via fetch, the response will be one of these shapes.
+ */
+export type FormFetchResponse<
```
Rich-Harris, line 337: still not sure I understand why this is distinct from `ActionResult`, since errors will (I think?) be serialized as `{ type: 'error', error: any }`

packages/kit/src/runtime/client/client.js
Rich-Harris: did a slightly different approach — rather than calling `update`, we skip ahead to creating a new `NavigationState` that includes the error

Rich-Harris (approved):
amazing work 🙇 

i would merge this, but i found it made sense to use the new `cookies` API in the docs rather than have another thing that will immediately need to be rewritten, so i figured both could be merged and released simultaneously

packages/kit/src/core/sync/write_types/test/simple-page-server-only/_expected/$types.d.ts
dummdidumm: It's because the actual file that is compared against is deeply nested; I didn't want to create 5 useless folders in `_expected` 😄

packages/kit/src/core/sync/write_types/test/simple-page-server-only/_expected/$types.d.ts
dummdidumm: True, oversight from when ActionEvent still was something more than RequestEvent

weepy:
trying to migrate... what is the final cookies API ?

weepy:
aha --- nice i'm getting an informative error
```
Use `event.cookie.set(name, value, options)` instead of `event.setHeaders` to set cookies
```

rowantrollope:
Tried the upgrade to 472 this morning got the following error. 

This is a large project with many routes. Also had problem with the last install script failing, but I was able to migrate successfully manually.  The problem here is that the install fails and rolls-back to the previous build.  Any suggestions? 
Do I need to go upgrade the method overrides and so on BEFORE installing 472? 

Error: Method overrides have been removed in favor of actions. See the PR for more information: https://github.com/sveltejs/kit/pull/6469
    at file:///Users/rtrollope/git/bookjournal/node_modules/.pnpm/@sveltejs+kit@1.0.0-next.472_tqsk4he7hsj5ptovw3sfpzsuyu/node_modules/@sveltejs/kit/src/core/config/options.js:487:10
    at file:///Users/rtrollope/git/bookjournal/node_modules/.pnpm/@sveltejs+kit@1.0.0-next.472_tqsk4he7hsj5ptovw3sfpzsuyu/node_modules/@sveltejs/kit/src/core/config/options.js:379:43
    at file:///Users/rtrollope/git/bookjournal/node_modules/.pnpm/@sveltejs+kit@1.0.0-next.472_tqsk4he7hsj5ptovw3sfpzsuyu/node_modules/@sveltejs/kit/src/core/config/options.js:365:18
    at file:///Users/rtrollope/git/bookjournal/node_modules/.pnpm/@sveltejs+kit@1.0.0-

dummdidumm:
How do you run the install? Just `npm install`? Do you have a `postinall` or another post-install npm hook in your `package.json` by any chance? If so, try removing that temporarily and try again.

rowantrollope:
I ran pnpm I @sveltejs/kit@latest... My Package.json looks as follows: 

{
  "name": "xy",
  "version": "0.1.0",
  "scripts": {
    "dev": "vite dev",
    "build": "vite build",
    "package": "vite package",
    "preview": "vite preview",
    "check": "svelte-check --tsconfig ./tsconfig.json",
    "check:watch": "svelte-check --tsconfig ./tsconfig.json --watch",
    "lint": "prettier --ignore-path .gitignore --check --plugin-search-dir=. .",
    "format": "prettier --ignore-path .gitignore --write --plugin-search-dir=. ."
  },
  "devDependencies": {
    "@rgossiaux/svelte-headlessui": "^1.0.2",
    "@rgossiaux/svelte-heroicons": "^0.1.2",
    "@sveltejs/adapter-auto": "^1.0.0-next.66",
    "@sveltejs/kit": "1.0.0-next.463",
    "@tailwindcss/typography": "^0.5.4",
    "@types/cookie": "^0.5.1",
    "@types/dotenv-flow": "^3.2.0",
    "@zerodevx/svelte-toast": "^0.7.2",
    "autoprefixer": "^10.4.8",
    "firebase": "^9.9.3",
    "picmo": "^5.6.1",
    "postcss": "^8.4.16",
    "prettier": "^2.7.1",
    "prettier-plugin-svelte": "^2.5.0",
    "svelte": "^3.50.0",
    "svelte-check": "^2.9.0",
    "svelte-loading-spinners": "^0.1.7",
    "svelte-preprocess": "^4.10.7",
    "tailwindcss": "^3.1.8",
    "tslib": "^2.3.1",
    "typescript": "^4.8.2",
    "vite": "3.1.0-beta.1"
  },
  "type": "module",
  "dependencies": {
    "@fontsource/fira-mono": "^4.5.9",
    "@lukeed/uuid": "^2.0.0",
    "@node-redis/search": "^1.0.5",
    "@popperjs/core": "^2.11.6",
    "@sveltejs/adapter-vercel": "1.0.0-next.72",
    "@tailwindcss/line-clamp": "^0.4.0",
    "amazon-paapi": "^1.0.6",
    "cloudinary-core": "^2.13.0",
    "colorthief": "^2.3.2",
    "cookie": "^0.5.0",
    "dayjs": "^1.11.5",
    "devalue": "3",
    "dotenv-flow": "^3.2.0",
    "firebase-admin": "^10.2.0",
    "redis": "^4.3.0",
    "sass": "^1.54.8",
    "svelte-gestures": "^1.4.1",
    "svelte-popperjs": "^1.3.1",
    "svelte-tags-input": "^3.0.0"
  }
}

twobitunicorn:
@dummdidumm Thanks! I'll get working through them.

twobitunicorn:
@rowantrollope In the post-install script for svelte-kit it seems that it checks all of your `+page.server.(js,ts)` for `load`s and if it finds any will give you your error.  The quick and dirty I did was comment all the code out in all of my `+page.server.(js,ts)`s, do the update, and got to work changing it all.

twobitunicorn:
@rowantrollope Check your `svelte.config.js` and make sure you are no longer adding the method overrides for `PATCH` and `DELETE`.  The snippet below is possibly what you have if you have an older project (as I did!).	

```
// Override http methods
methodOverride: {
    allowed: ['PATCH', 'DELETE']
}
```

rowantrollope:
FIXED! That was it.  Thank you VERY much

Rowan

Sent via Superhuman ( ***@***.*** )

tsukhu:
- Upgraded to form actions for the login screen . This is a really nice feature
From the form post to processing of the action is there a way or a hook to know the current state of the server (for example - form submitted, action is in progress etc.)
- Some way to show the button spinner icon to know something is going on.

dummdidumm:
You can provide your own callback to the enhance action and set a variable to pending during fetching. When the return callback function is called you know the fetch is done.

dimplast:
in docs 
[Anatomy of an action](https://kit.svelte.dev/docs/form-actions#anatomy-of-an-action)

in file:
**src/routes/login/+page.svelte**

`<script>`
`  export let data;`
`  export let form;`
`</script>`

`{#if form?.success}`
`  <p>Successfully logged in! Welcome back, {data.user.name}</p>`
`{/if}`

**_Should we write?_**

`{#if form?.success}`
`  <p>Successfully logged in! Welcome back, {form.user.name}</p>`
`{/if}`

`  export let data;` **_is necessary_**

fjdumont:
> ` export let data;` **_is necessary_**

Ooh, that likely explains why my experiments did not work as expected. I was wondering whether forms is typed automatically. Having seen the incredible typing magic by @dummdidumm I assume it is, and that's probably why we return validation errors instead of throwing them?

vipero07:
> @rowantrollope Check your `svelte.config.js` and make sure you are no longer adding the method overrides for `PATCH` and `DELETE`. The snippet below is possibly what you have if you have an older project (as I did!).
> 
> ```
> // Override http methods
> methodOverride: {
>     allowed: ['PATCH', 'DELETE']
> }
> ```

This should be mentioned in the migration guide.

rowantrollope:
Yes that was it, and that fixed it! Thanks

Sent via Superhuman iOS ( ***@***.*** )

ulvidamirli:
How to run fetch request to API inside action function?

franksplaudio:
So what about array-data inside my form?
Given a form to edit e.g. a product category with multiple names in different languages. The names are inside an input field inside an #each block.
How can the actions method get the different values form the input fields?

franksplaudio:
Ok, got it. FormData.getAll().

j2l:
Does ` <p>Successfully logged in! Welcome back, {data.user.name}</p>` work for you?
On success, `data={}` right?

dlamon1:
Step 2 should import fail, not invalid

feynmanliang:
May want to address JSON POST endpoints for the conversion guide? In my case I needed to use `devalue.uneval` to restore the original `response.json()` shape.

dummdidumm:
See #7494 for that, I added a note to the migration description

PR #6485, opened 2022-08-31
[ci] Cache Playwright browsers by their versions

### LINKED ISSUE #6464: [ci] cache playwright browser binaries according to their version, to avoid conflict on caching
### Describe the problem

While I was working on some PR here, I had accidentally upgraded Playwright version from 1.25.0 to 1.25.1.
Turns out, there is a different browser binaries for that version, and the caching in the CI force Playwright to use the browser binaries of 1.25.0., although I used 1.25.1. This is the part of the CI workflow causing it:
https://github.com/sveltejs/kit/blob/db506894463cd846ac9e8ddc7fa542630ed5ba27/.github/workflows/ci.yml#L53-L57

This can be problematic not only to me, but also for the main project CI, when you'll decide to upgrade playwright, unless you'll decide to clean cache every time.

Instead of caching to `node_modules`, use a different caching directory for the binaries, and use current playwright version (+ the runner OS details) as the caching key.

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

Tal500:
closes #6464

### Tests

### Changesets

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,12 +70,9 @@ jobs:
       - name: Set Environment Variable of Playwright Browsers Path (windows)
         if: runner.os == 'Windows'
         run: |
-          echo "PLAYWRIGHT_BROWSERS_PATH=$HOME\__playwight_browsers" >> $env:GITHUB_ENV
-      - name: Set Environment Variable of Playwright Browsers Path (windows) - Part 2 (Test only)
-        if: runner.os == 'Windows'
-        run: |
           $PLAYWRIGHT_VERSION = (pnpm playwright --version)
           echo "PLAYWRIGHT_VERSION=$PLAYWRIGHT_VERSION" >> $env:GITHUB_ENV
+          echo "PLAYWRIGHT_BROWSERS_PATH=$HOME\__playwight_browsers" >> $env:GITHUB_ENV
       - name: Cache Playwright
         uses: actions/cache@v3
         id: playwright_cache
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
         run: echo "PLAYWRIGHT_BROWSERS_PATH=$GITHUB_WORKSPACE/__playwight_browsers" >> $GITHUB_ENV
       - name: Set Environment Variable of Playwright Browsers Path (windows)
         if: runner.os == 'Windows'
-        run: echo "PLAYWRIGHT_BROWSERS_PATH=$GITHUB_WORKSPACE/__playwight_browsers" >> $env:GITHUB_ENV
+        run: echo "PLAYWRIGHT_BROWSERS_PATH=%GITHUB_WORKSPACE%/__playwight_browsers" >> $env:GITHUB_ENV
       - name: Cache Playwright
         uses: actions/cache@v3
         id: playwright_cache
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
     env:
       TURBO_CACHE_KEY: ${{ matrix.os }}-${{ matrix.node-version }}
       # Install playwright's binray under node_modules so it will be cached together
-      PLAYWRIGHT_BROWSERS_PATH: ${{ $GITHUB_WORKSPACE }}/__playwight_browsers
+      PLAYWRIGHT_BROWSERS_PATH: ${{ env.GITHUB_WORKSPACE }}/__playwight_browsers
       KIT_E2E_BROWSER: ${{matrix.e2e-browser}}
     steps:
       - run: git config --global core.autocrlf false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,10 +74,9 @@ jobs:
       - name: Set Environment Variable of Playwright Browsers Path (windows) - Part 2 (Test only)
         if: runner.os == 'Windows'
         run: |
-          pnpm playwright --version > __playwright_version.txt
+          $Variable = (pnpm playwright --version)
           echo %PLAYWRIGHT_VERSION%
-          set /P PLAYWRIGHT_VERSION= < __playwright_version.txt
-          echo %PLAYWRIGHT_VERSION%
+          echo $PLAYWRIGHT_VERSION
       - name: Cache Playwright
         uses: actions/cache@v3
         id: playwright_cache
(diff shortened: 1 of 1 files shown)


Tal500:
Seems like the CI isn't running, maybe because it waits for the requested review?

dominikg:
Hi, glad you noticed that the current way we try to cache playwright isn't working, but please check out https://github.com/vitejs/vite/blob/main/.github/workflows/ci.yml for a working solution

(what we have right now in kit is based on a previous version of vite's )

Especially look at these:
https://github.com/vitejs/vite/blob/main/.github/workflows/ci.yml#L8
https://github.com/vitejs/vite/blob/main/.github/workflows/ci.yml#L62-80

Note kit is running a matrix for browser-tests with firefox and safari too so you may need to do a bit of extra work

dominikg:
Also we used to do this separate playwright caching in the past, but had issues with the separate cache being slower than outright downloading it or even causing suite errors when the caching failed.

So if the current solution doesn't work, the initial step would be to disable automatic playwright downloading and always calling `pnpm playwright install` 

cc @pngwn who worked on the initial playwright caching solution irrc

Tal500:
> Also we used to do this separate playwright caching in the past, but had issues with the separate cache being slower than outright downloading it or even causing suite errors when the caching failed.
> 
> So if the current solution doesn't work, the initial step would be to disable automatic playwright downloading and always calling `pnpm playwright install`
> 
> cc @pngwn who worked on the initial playwright caching solution irrc

If I understand you correctly, what that you suggest is not to take Playwright version into the key - since anyway Playwright will check that the binaries are OK.

The problem is about the design of Github Action caching - if there is a cache stored with some key - there will be no another cache that will take its place, even if the run was successful and something in the path had changed.

This is why I'm in favor of caching only according to the playwright version. How much relevant playwright version exist simultaneously?

dominikg:
My strategy would be
1) disable automatic playwright downloading AND caching. Always use `pnpm playwright install browsername`. 
This downloads browsers on every test run, but the cache is also "downloaded" so the difference may not be large
2) If the setup from 1) is not fast enough, copy the vite setup i posted and try with that. It contains extra handling for windows and i trust that they figured out how to handle playwright browser version updates

Tal500:
> My strategy would be
> 
>     1. disable automatic playwright downloading AND caching. Always use `pnpm playwright install browsername`.
>        This downloads browsers on every test run, but the cache is also "downloaded" so the difference may not be large
> 
>     2. If the setup from 1) is not fast enough, copy the vite setup i posted and try with that. It contains extra handling for windows and i trust that they figured out how to handle playwright browser version updates
> 
> 
> edit: explaination for 1). The goal of the current ci.yml setup in this repo was to cache the playwright browsers in the same cache as the dependencies, to avoid extra downloads. Putting them in a separate cache voids that.

My problem with the method you're doing in vite, is that once Github have cache-hit, it wouldn't update the cache in the end of execution (i.e. no cache overriding, only set cache if it was cache-miss).

So it means that if your using Playwright version X, and later you'll use version X+1, the cache will always be for version X and won't be updated, until there will be finally cache timeout(I think usually it's after 30 days). This is my problem with the way of Vite CI.

Did I miss something?

dominikg:
you may be right https://github.com/vitejs/vite/runs/8118481912?check_suite_focus=true#step:25:2

but my remark was about the different handling of windows cache path too, see https://github.com/vitejs/vite/blob/main/.github/workflows/ci.yml#L62

Either way, all this complexity and the questionable gain of caching playwright browser versions at all makes me think it would be better to not do it.

So just adding `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: "1"` to prevent double downloads is enough. No caching of playwright browsers at all.

Tal500:
I think that now the workflow looks good overall.
Just think this can speed up the checking process and avoid silly caching problems with Playwright version.

Tal500:
After I made the workflow to only download the needed browser, we have a problem:

We have an issue on `create-svelte` testing: It's install + build the default package, but Playwright complains it doesn't found the chrome browser (since you need to install all relevant Playwright browsers before build your package). Details:
https://github.com/sveltejs/kit/runs/8123872539?check_suite_focus=true

Need somehow to set dynamically the preferred Playwright browser also for the installation of `create-svelte`.

Tal500:
Now the caching works well

.github/workflows/ci.yml
```diff
@@ -64,7 +62,24 @@ jobs:
...
+        run: |
+          $PLAYWRIGHT_VERSION = (pnpm playwright --version)
+          echo "PLAYWRIGHT_VERSION=$PLAYWRIGHT_VERSION" >> $env:GITHUB_ENV
+          echo "PLAYWRIGHT_BROWSERS_PATH=$HOME\__playwight_browsers" >> $env:GITHUB_ENV
+      - name: Cache Playwright
+        uses: actions/cache@v3
+        id: playwright_cache
+        with:
+          path: ${{ env.PLAYWRIGHT_BROWSERS_PATH }}
+          key: OS - ${{ runner.os }} - Node ${{ matrix.node-version }} - Browser ${{ matrix.e2e-browser }} - Playwright Version - ${{ env.PLAYWRIGHT_VERSION }}
```
dominikg, line 81: Is keying by node version required? i don't think playwright installs different browser versions depending on the node version used.

dominikg:
see #6502, that makes adapter-static tests use the configured browser and uses `webkit` as key to avoid the rename step.

Downloading playwright browser takes between 5 and 13s in that run, the caching here also seems to take 5s, So we are looking at less than 10s gains.

.github/workflows/ci.yml
dominikg: The only place where this could go wrong is when two matrix jobs try to update the cache with the same key at the same time. :thinking: 

They would update with the same content though so either first-one-wins of last-one-wins. the winner should be picked up by subsequent runs. It would be bad if the loser failed it's run though

Tal500: OK, done it

dominikg:
I did enable the same caching as here in #6502 to get an apples to apples comparison and it looks like we do gain few seconds, 

actions cache is consistently at ~5s to do a cache restore, while some playwright downloads take a bit longer, about 10s or so.

At the same time the 4 caches  linux-chromium, linux-firefox, windows-chromium, macos-safari take about 350MB in cache space - per playwright version. 
So we're looking at 350-700mb cache space used if we only update playwright infrequently.

By comparison our pnpm store cache is ca 55MB in size currently.

PR #6489, opened 2022-08-31
[doc] more information on `depends` and `invalidate`

icalvin102:
`depends` and `invalidate` are not documented enough. 
Related issues (most recently #6354) and a lot of questions on the Svelte discord server show that there is some confusion about their usage.

### Tests

### Changesets

### DIFF
--- a/documentation/docs/05-load.md
+++ b/documentation/docs/05-load.md
@@ -99,11 +99,13 @@ An instance of [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL), co
 
 #### depends
 
-This function declares that the `load` function has a _dependency_ on one or more URLs, which can subsequently be used with [`invalidate()`](/docs/modules#$app-navigation-invalidate) to cause `load` to rerun.
+This function declares that the `load` function has a _dependency_ on one or more URLs or custom identifiers, which can subsequently be used with [`invalidate()`](/docs/modules#$app-navigation-invalidate) to cause `load` to rerun.
 
 Most of the time you won't need this, as `fetch` calls `depends` on your behalf — it's only necessary if you're using a custom API client that bypasses `fetch`.
 
 URLs can be absolute or relative to the page being loaded, and must be [encoded](https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding).
+
+Custom identifiers have to be prefixed with one or more lowercase letters followed by a colon to conform to the [URI specification](https://www.rfc-editor.org/rfc/rfc3986.html)
 
 ```js
 // @filename: ambient.d.ts
@@ -121,7 +123,11 @@ import * as api from '$lib/api';
 
 /** @type {import('./$types').PageLoad} */
 export async function load({ depends }) {
-	depends(`${api.base}/foo`, `${api.base}/bar`);
+	depends(
+		`${api.base}/foo`,
+		`${api.base}/bar`
+		'page:foobar'
+	);
 
 	return {
 		foo: api.client.get('/foo'),
--- a/packages/kit/types/ambient.d.ts
+++ b/packages/kit/types/ambient.d.ts
@@ -113,10 +113,22 @@ declare module '$app/navigation' {
 		opts?: { replaceState?: boolean; noscroll?: boolean; keepfocus?: boolean; state?: any }
 	): Promise<void>;
 	/**
-	 * Causes any `load` functions belonging to the currently active page to re-run if they `fetch` the resource in question. If no argument is given, all resources will be invalidated. Returns a `Promise` that resolves when the page is subsequently updated.
+	 * Causes any `load` functions belonging to the currently active page to re-run if they registered it with `fetch` or `depends`. Returns a `Promise` that resolves when the page is subsequently updated.
+	 *
+	 * If no argument is given, all resources will be invalidated.
+	 *
+	 * The `string` argument has to match the URI that was passed to `fetch` or `depends` exactly (including query parameters).
+	 *
+	 * The `function` argument can be used define a custom predicate. It receives the full URI and causes `load` to rerun if `true` is returned.
+	 * This can be useful if you want to invalidate based on a pattern instead of a exact match.
+	 *
+	 * ```js
+	 * // Example: Match '/path' regardless of the query parameters
+	 * invalidate((url) => (new URL(url)).pathname === '/path')`
+	 * ```
 	 * @param dependency The invalidated resource
 	 */
-	export function invalidate(dependency?: string | ((href: string) => boolean)): Promise<void>;
+	export function invalidate(dependency?: string | ((url: string) => boolean)): Promise<void>;
 	/**
 	 * Programmatically prefetches the given page, which means
 	 *  1. ensuring that the code for the page is loaded, and


documentation/docs/05-load.md
```diff
@@ -99,12 +99,14 @@ An instance of [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL), co
...
 #### depends
 
-This function declares that the `load` function has a _dependency_ on one or more URLs, which can subsequently be used with [`invalidate()`](/docs/modules#$app-navigation-invalidate) to cause `load` to rerun.
+This function declares that the `load` function has a _dependency_ on one or more URLs or custom identifiers, which can subsequently be used with [`invalidate()`](/docs/modules#$app-navigation-invalidate) to cause `load` to rerun.
 
 Most of the time you won't need this, as `fetch` calls `depends` on your behalf — it's only necessary if you're using a custom API client that bypasses `fetch`.
 
 URLs can be absolute or relative to the page being loaded, and must be [encoded](https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding).
 
+Custom identifiers have to be prefixed with one or more lowercase letters followed by a colon to conform to the [URI specification](https://www.rfc-editor.org/rfc/rfc3986.html)
```
Rich-Harris, line 108: This is brilliant! Had been agonising over what sort of API would work for both URLs and non-URLs, and all this time you can just do `a:b`. This makes life much easier

documentation/docs/05-load.md
```diff
@@ -121,7 +123,11 @@ import * as api from '$lib/api';
 
 /** @type {import('./$types').PageLoad} */
 export async function load({ depends }) {
-	depends(`${api.base}/foo`, `${api.base}/bar`);
+	depends(
+		`${api.base}/foo`,
+		`${api.base}/bar`
+		'page:foobar'
```
Rich-Harris, line 129: I think something like this probably communicates 'put what you like here' better:

```suggestion
		'my-stuff:foo'
```

packages/kit/types/ambient.d.ts
```diff
@@ -113,10 +113,22 @@ declare module '$app/navigation' {
 		opts?: { replaceState?: boolean; noscroll?: boolean; keepfocus?: boolean; state?: any }
 	): Promise<void>;
 	/**
-	 * Causes any `load` functions belonging to the currently active page to re-run if they `fetch` the resource in question. If no argument is given, all resources will be invalidated. Returns a `Promise` that resolves when the page is subsequently updated.
+	 * Causes any `load` functions belonging to the currently active page to re-run if they registered it with `fetch` or `depends`. Returns a `Promise` that resolves when the page is subsequently updated.
+	 *
+	 * If no argument is given, all resources will be invalidated.
```
Rich-Harris, line 118: Separate to this PR but I think we should probably create an `invalidateAll()` function — makes the `invalidate()` interface a bit less cluttered, and will be beneficial if we need to add an options object (since `invalidate(undefined, options)` is weird)

packages/kit/types/ambient.d.ts
```diff
@@ -113,10 +113,22 @@ declare module '$app/navigation' {
...
+	 *
+	 * The `string` argument has to match the URI that was passed to `fetch` or `depends` exactly (including query parameters).
+	 *
+	 * The `function` argument can be used define a custom predicate. It receives the full URI and causes `load` to rerun if `true` is returned.
+	 * This can be useful if you want to invalidate based on a pattern instead of a exact match.
+	 *
+	 * ```js
+	 * // Example: Match '/path' regardless of the query parameters
+	 * invalidate((url) => (new URL(url)).pathname === '/path')`
```
Rich-Harris, line 127: This would be way nicer if `url` was already a `URL` object rather than a string. Might change that alongside adding `invalidateAll()`

packages/kit/types/ambient.d.ts
```diff
@@ -113,10 +113,22 @@ declare module '$app/navigation' {
 		opts?: { replaceState?: boolean; noscroll?: boolean; keepfocus?: boolean; state?: any }
 	): Promise<void>;
 	/**
-	 * Causes any `load` functions belonging to the currently active page to re-run if they `fetch` the resource in question. If no argument is given, all resources will be invalidated. Returns a `Promise` that resolves when the page is subsequently updated.
+	 * Causes any `load` functions belonging to the currently active page to re-run if they registered it with `fetch` or `depends`. Returns a `Promise` that resolves when the page is subsequently updated.
```
Rich-Harris, line 116: I think some words got lost in this sentence during editing 😛 

```suggestion
	 * Causes any `load` functions belonging to the currently active page to re-run if they depend on the resource in question, via `fetch` or `depends`. Returns a `Promise` that resolves when the page is subsequently updated.
```

Rich-Harris (changes requested):
thank you! couple of small comments, plus some notes about improvements to the `invalidate` API that i think would be worth sneaking in ahead of this

documentation/docs/05-load.md
icalvin102: It has been in the description of https://github.com/sveltejs/kit/pull/4536 all this time. But I should have documented it a bit clearer.

I can relate to your agony. It took me a lot of thinking and tinkering with the `URL` constructor to arrive at this seemingly obvious solution.

icalvin102:
@lukaszpolowczyk `invalidate` causes the entire `load` function to rerun. I does not perform any magic inside of the load function.

Automatic caching of individual resources in `load` would be really difficult because some resources may depend on each other.

However giving the user a way to determine which resource should be fetched again would theoretically be possible if `load` would provide something like `isInvalidating(url)`

```js
export async function load({ fetch, isInvalidating }) {
  const url = `https://cms.example.com/articles.json`;
  const url2 = `https://cms.example.com/articles2.json`;
  let response, response2;
  
  if(isInvalidating(url)) {
    response = await fetch(url);
    cache.set(url, response);
  } else {
    response = cache.get(url);
  }
  
  if(isInvalidating(url2)) {
    response2 = await fetch(url);
    cache.set(url2, response2);
  } else {
    response2 = cache.get(url2);
  }
 
  return {
    articles: await response.json(),
    articles2: await response2.json()
  };
}
```

But this feature does currently not exist. Might be worth a feature request.

packages/kit/types/ambient.d.ts
icalvin102: True. `invalidate((url) => url.pathname === '/path')` or `invalidate(({pathname}) => pathname === '/path')` is a lot more readable and getting full url as string would still be possible with `url.href`.

Sadly I dont have the time to implement this my self at the moment.

lukaszpolowczyk:
> But this feature does currently not exist. Might be worth a feature request.

@icalvin102 I added - https://github.com/sveltejs/kit/issues/6495

packages/kit/types/ambient.d.ts
Rich-Harris: implemented in #6493

packages/kit/types/ambient.d.ts
Rich-Harris: implemented in #6493

Rich-Harris:
incorporated this branch into #6493, so I'll close this — thanks!

PR #6497, opened 2022-08-31
prefer JSON error responses over HTML in fetch case

### LINKED ISSUE #1931: Allow error pages with json bodies

Rich-Harris:
closes #1931.

Today, if you have a route like this...

```js
// src/routes/error/+server.js
export function GET() {
  throw new Error('yikes');
}
```

...then visiting `/error` will show you the fallback error page:

image

If you `fetch` the page with an `application/json` header, you'll get a JSON representation of the error...

image

...but by default `fetch` will get HTML. In most cases that's probably not that helpful.

This PR reverses the precedence during content negotiation, so that if a request has an `accept: */*` header (which `fetch` requests do, by default) it will prefer JSON to HTML.

### Tests

### Changesets

### DIFF
--- a/.changeset/odd-months-grow.md
+++ b/.changeset/odd-months-grow.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Prefer JSON responses when returning errors if accept header is _/_
+Prefer JSON responses when returning errors if accept header is `*/*`
--- a/.changeset/odd-months-grow.md
+++ b/.changeset/odd-months-grow.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Prefer JSON responses when returning errors if accept header is _/_
--- a/packages/kit/src/runtime/server/utils.js
+++ b/packages/kit/src/runtime/server/utils.js
@@ -194,8 +194,8 @@ export function handle_fatal_error(event, options, error) {
 
 	// ideally we'd use sec-fetch-dest instead, but Safari — quelle surprise — doesn't support it
 	const type = negotiate(event.request.headers.get('accept') || 'text/html', [
-		'text/html',
-		'application/json'
+		'application/json',
+		'text/html'
 	]);
 
 	if (event.url.pathname.endsWith(DATA_SUFFIX) || type === 'application/json') {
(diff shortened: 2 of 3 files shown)


.changeset/odd-months-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Prefer JSON responses when returning errors if accept header is _/_
```
Conduitry, line 5:
```suggestion
Prefer JSON responses when returning errors if accept header is \*/\*
```

I think Prettier got its hands on this and changed `*/*` to `_/_`, because an emphasized forward slash is definitely what you wanted.

Rich-Harris: fucking hell computers are idiots

Rich-Harris:
```suggestion
Prefer JSON responses when returning errors if accept header is `*/*`
```

Rich-Harris:
Marked this as draft because https://github.com/sveltejs/kit/pull/6497/files#diff-964f525b2136e97820d0ba82ceb5c6ab8d7a38c365fe198cfe855910500972cfR182 is unforgivable. Need to come up with some unambiguous way to communicate error information in a way that's comprehensible to both SvelteKit and users

Rich-Harris:
Ok, I have an idea for what to do about errors more broadly, but I think it falls out of the scope of this PR, so I'm marking it as ready again

dummdidumm (approved):

PR #6507, opened 2022-09-01
fix server manifest generation

Rich-Harris:
I was seeing this in the CI logs:

image

It's happening because `JSON.stringify` turns `undefined` to `null` inside an array, unhelpfully. Bit weird that it isn't actually causing anything to fail, but I figured it was worth fixing anyway

### DIFF
--- a/packages/kit/src/core/generate_manifest/index.js
+++ b/packages/kit/src/core/generate_manifest/index.js
@@ -60,13 +60,13 @@ export function generate_manifest({ build_data, relative_path, routes, format = 
 					if (!route.page && !route.endpoint) return;
 
 					return `{
-						id: ${s(route.id)},
-						pattern: ${route.pattern},
-						names: ${s(route.names)},
-						types: ${s(route.types)},
-						page: ${s(route.page)},
-						endpoint: ${route.endpoint ? loader(`${relative_path}/${build_data.server.vite_manifest[route.endpoint.file].file}`) : 'null'}
-					}`;
+					id: ${s(route.id)},
+					pattern: ${route.pattern},
+					names: ${s(route.names)},
+					types: ${s(route.types)},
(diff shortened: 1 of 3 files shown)


Rich-Harris:
losing my hair trying to understand these test failures

packages/kit/src/core/generate_manifest/index.js
```diff
@@ -60,13 +60,13 @@ export function generate_manifest({ build_data, relative_path, routes, format =
...
-						names: ${s(route.names)},
-						types: ${s(route.types)},
-						page: ${s(route.page)},
-						endpoint: ${route.endpoint ? loader(`${relative_path}/${build_data.server.vite_manifest[route.endpoint.file].file}`) : 'null'}
-					}`;
+					id: ${s(route.id)},
+					pattern: ${route.pattern},
+					names: ${s(route.names)},
+					types: ${s(route.types)},
+					page: ${route.page ? `{ layouts: ${get_nodes(route.page.layouts)}, errors: ${get_nodes(route.page.errors)}, leaf: ${route.page.leaf} }` : 'null'},
```
dummdidumm, line 67: Since we add the trailing comma now, we can probably get rid of the corresponding logic on the client in `parse.js`

dummdidumm (approved):

packages/kit/src/core/generate_manifest/index.js
Rich-Harris: We definitely could do the same thing — at one point in this PR I actually removed that code, somehow thinking that the change to the server `manifest.js` generation would also affect `client-manifest.js`. It'd probably be rare that the resulting manifest was larger than the code removed from `parse.js`, especially once zipped, so overall I think it'd be win. Probably one for a separate PR though

PR #6510, opened 2022-09-01
Block cross-site form POSTs by default

### LINKED ISSUE #72: Some sort of CSRF protection
(draw the rest of the owl)

Rich-Harris:
Closes #72. Adds a `config.kit.csrf.checkOrigin` option that defaults to `true`

### Tests

### Changesets

### DIFF
--- a/.changeset/strange-apples-vanish.md
+++ b/.changeset/strange-apples-vanish.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-[breaking] block cross-site mutative requests by default. disable with config.kit.csrf.checkOrigin
+[breaking] block cross-site form POSTs by default. disable with config.kit.csrf.checkOrigin
--- a/.changeset/strange-apples-vanish.md
+++ b/.changeset/strange-apples-vanish.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] block cross-site mutative requests by default. disable with config.kit.csrf.checkOrigin
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -18,12 +18,19 @@ const default_transform = ({ html }) => html;
 export async function respond(request, options, state) {
 	let url = new URL(request.url);
 
-	if (
-		options.csrf.check_origin &&
-		request.method === 'POST' &&
-		request.headers.get('origin') !== url.origin
-	) {
-		return new Response(`Cross-site ${request.method} requests are forbidden`, { status: 403 });
+	if (options.csrf.check_origin) {
+		const type = request.headers.get('content-type');
+
+		const forbidden =
+			request.method === 'POST' &&
+			request.headers.get('origin') !== url.origin &&
+			(type === 'application/x-www-form-urlencoded' || type === 'multipart/form-data');
+
+		if (forbidden) {
+			return new Response(`Cross-site ${request.method} form submissions are forbidden`, {
+				status: 403
+			});
+		}
 	}
 
 	const { parameter, allowed } = options.method_override;
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -19,7 +19,7 @@ export async function respond(request, options, state) {
 	let url = new URL(request.url);
 
 	if (options.csrf.check_origin) {
-		const type = request.headers.get('content-type');
+		const type = request.headers.get('content-type')?.split(';')[0];
 
 		const forbidden =
 			request.method === 'POST' &&
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -14,9 +14,19 @@ import { DATA_SUFFIX } from '../../constants.js';
 /** @param {{ html: string }} opts */
 const default_transform = ({ html }) => html;
 
+const mutative_methods = new Set(['POST', 'PUT', 'PATCH', 'DELETE']);
+
 /** @type {import('types').Respond} */
 export async function respond(request, options, state) {
 	let url = new URL(request.url);
+
+	if (
+		options.csrf.check_origin &&
+		mutative_methods.has(request.method) &&
+		request.headers.get('origin') !== url.origin
+	) {
+		return new Response(`Cross-site ${request.method} requests are forbidden`, { status: 403 });
+	}
 
 	const { parameter, allowed } = options.method_override;
 	const method_override = url.searchParams.get(parameter)?.toUpperCase();
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -8,13 +8,12 @@ import { decode_params, disable_search, normalize_path } from '../../utils/url.j
 import { exec } from '../../utils/routing.js';
 import { render_data } from './data/index.js';
 import { DATA_SUFFIX } from '../../constants.js';
+import { is_mutative_method } from '../../utils/http.js';
 
 /* global __SVELTEKIT_ADAPTER_NAME__ */
 
 /** @param {{ html: string }} opts */
 const default_transform = ({ html }) => html;
-
-const mutative_methods = new Set(['POST', 'PUT', 'PATCH', 'DELETE']);
 
 /** @type {import('types').Respond} */
 export async function respond(request, options, state) {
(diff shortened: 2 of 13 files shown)


Conduitry:
I'm wondering whether we should only do this for POSTs, because those are the only ones that can be done with old-fashioned form submits. The other ones would already be blocked anyway, right, unless you specifically took action to deal with the preflight CORS stuff? And if there were a way to only do this check for old-fashioned form submits and not other cross-origin fetch requests, that also might be nice, but I don't know that that's something we can reliably determine.

If this is something people are going to have to turn off before they can even start to do the legwork to allow CORS requests, that seems like it's inviting someone to then forget to close the form submit hole.

Rich-Harris:
Good point. Changed it to just `POST`. We could reliably determine native form submissions if Safari and IE11 supported `sec-fetch-dest`, but... they don't

Conduitry:
Could we block it only if it's a POST and the Origin doesn't match and it has a form Content-Type? If someone wanted to expose a POST endpoint for legitimate use by other servers, they would likely write it to accept JSON, so that wouldn't get in their way. And I think it still provides the same amount of protection.

Rich-Harris:
Oh you know what? That should definitely work. D'oh

packages/kit/src/runtime/server/index.js
```diff
@@ -18,12 +18,19 @@ const default_transform = ({ html }) => html;
...
-		request.headers.get('origin') !== url.origin
-	) {
-		return new Response(`Cross-site ${request.method} requests are forbidden`, { status: 403 });
+	if (options.csrf.check_origin) {
+		const type = request.headers.get('content-type');
+
+		const forbidden =
+			request.method === 'POST' &&
+			request.headers.get('origin') !== url.origin &&
+			(type === 'application/x-www-form-urlencoded' || type === 'multipart/form-data');
```
Conduitry, line 27: I think these might have a `; charset=` in them that would throw off this check.

packages/kit/src/runtime/server/page/fetch.js
```diff
@@ -132,6 +132,12 @@ export function create_fetch({ event, options, state, route, prerender_default }
 				throw new Error('Request body must be a string');
 			}
 
+			// we need to set an origin header if it's a POST request, to avoid
+			// CSRF protection blocking this request
+			if (opts.method?.toUpperCase() === 'POST') {
+				opts.headers.set('origin', event.url.origin);
+			}
+
```
Conduitry, line 140: Is this still necessary?

packages/kit/test/apps/basics/test/server.test.js
```diff
@@ -91,10 +103,15 @@ test.describe('Endpoints', () => {
...
+	test('request body can be read slow', async ({ baseURL, request }) => {
 		const data = randomBytes(1024 * 256);
 		const digest = createHash('sha256').update(data).digest('base64url');
-		const response = await request.put('/endpoint-input/sha256', { data });
+		const response = await request.put('/endpoint-input/sha256', {
+			data,
+			headers: {
+				origin: new URL(baseURL).origin
+			}
+		});
```
Conduitry, line 114: I don't think this change is needed anymore.

packages/kit/src/runtime/server/page/fetch.js
Rich-Harris: it is not. also i guess i can revert some of the test changes

packages/kit/test/apps/basics/test/server.test.js
```diff
@@ -109,8 +126,12 @@ test.describe('Errors', () => {
...
 	});
 
-	test('unhandled http method', async ({ request }) => {
-		const response = await request.put('/errors/invalid-route-response');
+	test('unhandled http method', async ({ baseURL, request }) => {
+		const response = await request.put('/errors/invalid-route-response', {
+			headers: {
+				origin: new URL(baseURL).origin
+			}
+		});
```
Conduitry, line 134: I don't think this change is needed anymore.

packages/kit/test/apps/basics/test/server.test.js
```diff
@@ -257,10 +278,11 @@ test.describe('Routing', () => {
 });
 
 test.describe('Shadowed pages', () => {
-	test('Action can return undefined', async ({ request }) => {
+	test('Action can return undefined', async ({ baseURL, request }) => {
 		const response = await request.post('/shadowed/simple/post', {
 			headers: {
-				accept: 'application/json'
+				accept: 'application/json',
+				origin: new URL(baseURL).origin
```
Conduitry, line 285: I don't think this change is needed anymore.

.changeset/strange-apples-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] block cross-site mutative requests by default. disable with config.kit.csrf.checkOrigin
```
Conduitry, line 5:
```suggestion
[breaking] block cross-site form POSTs by default. disable with config.kit.csrf.checkOrigin
```

packages/kit/src/runtime/server/index.js
Rich-Harris: I've never seen a `charset=`, but we _do_ need to split on `;` anyway because `multipart-formdata` will always be accompanied by `; boundary=`

packages/kit/test/apps/basics/test/server.test.js
```diff
@@ -22,6 +23,17 @@ test.describe('Content-Type', () => {
...
+	test('Blocks requests with incorrect origin', async ({ baseURL }) => {
+		const res = await fetch(`${baseURL}/csrf`, {
+			method: 'POST'
+		});
+
+		expect(res.status).toBe(403);
+		expect(await res.text()).toBe('Cross-site POST form submissions are forbidden');
+	});
+});
+
```
Conduitry, line 36: We're going to have to be cleverer now with this test to make it look more like a form submission.

Rich-Harris: oh ffs, one sec

Conduitry (approved):

Prinzhorn:
> Could we block it only if it's a POST and the Origin doesn't match **and it has a form Content-Type**? If someone wanted to expose a POST endpoint for legitimate use by other servers, they would likely write it to accept JSON, so that wouldn't get in their way. And I think it still provides the same amount of protection.

Until ~6 years ago the following would've bypassed the current implementation in Chrome and Firefox:

```html
<script>
  navigator.sendBeacon(
    'http://localhost:5173/todos',
    new Blob([JSON.stringify({ text: 'pwned' })], { type: 'application/json' })
  );
</script>
```

`sendBeacon` was using `no-cors` mode by default. I'm not up-to-date if that's still possible in semi-recent other browsers.

ChrisOgden:
This broke my implementation of Auth0 as an openid-connect module (rough draft but working - https://github.com/starbasehq/sveltekit-openid-connect/issues/11). In order to use the POST callback from Auth0 I have to set this option to false to disable CSRF checks. Would it be possible to whitelist certain domains/urls and still have the rest of the CSRF protection?

Conduitry:
Perhaps eventually. But for now, until we figure out what that API looks like, it should be pretty simple to disable that setting in the configuration and then re-create the adjusted check in your `handle` hook.

ChrisOgden:
> Perhaps eventually. But for now, until we figure out what that API looks like, it should be pretty simple to disable that setting in the configuration and then re-create the adjusted check in your `handle` hook.

That solution works for now. Here is how I did it:

top of hooks.js
```
// Where CSRF_ALLOWED is pulled from environment (process.env) and is a comma separated list of allowed hostname
// Where AUTH0_DOMAIN is endpoint from auth0 for tenant
const csrfAllowed = [`https://${AUTH0_DOMAIN}`, ...(CSRF_ALLOWED || '').split(',').filter(Boolean)]
```

Inside handle
```
export async function handle ({ event, resolve }) {
	const { forbidden: forbidCSRF, response: responseCSRF } = checkCSRF(event.request)
	if (forbidCSRF) return responseCSRF
	// ... rest of handle function
}
```

function at bottom of hooks.js
```
function checkCSRF (request) {

	const url = new URL(request.url)
	const type = request.headers.get('content-type')?.split(';')[0]
	const forbidden =
		request.method === 'POST' &&
		!_.includes([url.origin, ...csrfAllowed], request.headers.get('origin')) &&
		(type === 'application/x-www-form-urlencoded' || type === 'multipart/form-data')

	if (forbidden) {
		const response = new Response(`Cross-site ${request.method} form submissions are forbidden`, {
			status: 403
		})
		return { forbidden, response }
	} else {
		return { forbidden, response: null }
	}
}
```

Issue #6463, opened 2022-08-31
silence errors coming from `data-sveltekit-prefetch`


ivanhofer:
### Describe the problem

Now that we have [the option to prefetch all links inside a certain region of the DOM](https://kit.svelte.dev/docs/link-options#data-sveltekit-prefetch), I want to enable for all links instead of passing it to each link individually. I have added `data-sveltekit-prefetch` to the `body` tag.

But now I get a lot of errors:
```
Error: Attempted to prefetch a URL that does not belong to this app
````
some of my links will open to an external ressource.

### Describe the proposed solution

It would be great if `data-sveltekit-prefetch` would not throw errors and instead could silently fail.

Maybe a second optional param could be added to the `prefetch` function to allow to control if an error should be thrown or if the function should just returns.

This would also make it possible to have the same feature when calling `prefetch` programmatically.

### Importance

would make my life easier

**Comment by flayks** [comment]
Just my opinion but I'm not certain that putting `data-sveltekit-prefetch` on the `<body>` would be the best idea as it obviously would catch external links and then throw this error?  
Since this should only be applied to internal links/existing routes, maybe specifying it to certains blocks where you know you'll have some might be a better option, like the default example site has been updated to.

On another hand, `data-sveltekit-prefetch` could maybe check to not prefetch external links? That's another thing tho 😄

```svelte
<nav data-sveltekit-prefetch>
  <ul>
    <li><a href="/">Homepage</a></li>
    <li><a href="/route">Another route</a></li>
  </ul>
</nav>
```

ivanhofer:
The example in the docs applies it on the `main` section, so it is set on a rather big scope.

And the second approach you mention is what I did propose in the issue 😅

flayks:
> The example in the docs applies it on the `main` section, so it is set on a rather big scope.

True! I see the attribute on the in the nav in the default example though, so mixed examples here and there I suppose.
https://github.com/sveltejs/kit/blob/master/packages/create-svelte/templates/default/src/lib/header/Header.svelte#L13

> And the second approach you mention is what I did propose in the issue 😅

I agree 👌

Greenheart:
[...]

PR #6518, opened 2022-09-01
fix: Silently skip prefetching external URLs instead of warning

Greenheart:
Fix https://github.com/sveltejs/kit/issues/6463 by automatically skipping prefetching of external URLs instead of warning app developers. This aligns with how Next.js solves the same problem.

### Tests

### Changesets

### DIFF
--- a/.changeset/nice-poets-taste.md
+++ b/.changeset/nice-poets-taste.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Silently skip prefetching of external URLs instead of warning
+Silently skip prefetching of external URLs when using `data-sveltekit-prefetch`. Warn like before when calling `prefetch()` for external URLs.
--- a/.changeset/nice-poets-taste.md
+++ b/.changeset/nice-poets-taste.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Silently skip prefetching of external URLs instead of warning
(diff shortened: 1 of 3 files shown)


Rich-Harris:
Thank you! I think the error is useful in the case where you're programmatically prefetching. What about changing the behaviour of `trigger_prefetch` to ignore any links that would normally be ignored? (Might need to abstract some logic out from the `click` handler to avoid repetition)

Greenheart:
Thanks for the feedback! That sounds like a good way forward. When app devs call `prefetch()` there's a clear intention and the warning would be helpful.

I'll explore this further in the coming days. This would be a fun contribution to make!

Greenheart:
## Update:
Prefetching never happens for external URLs.

The difference is that programmatic calls to `prefetch()` will now throw the same error as before, while `trigger_prefetch` that is activated by `data-sveltekit-prefetch` silently fails just like the previous commit.


## Remaning TODOs
- [ ] I get two failing tests locally which didn't fail before. Any ideas what might have caused this? Perhaps because we're now inconsistent and throwing errors in one case but not in other cases? Or this might be caused by test flakiness of some kind.

```
@sveltejs/kit:test:     [chromium-dev] › test/client.test.js:309:2 › Errors › Root error falls back to error.html ======
@sveltejs/kit:test:     [chromium-dev] › test/client.test.js:360:2 › Load › accessing url.hash from load errors and suggests using page store
```

dummdidumm (approved):
CI succeeded and I don't get these test failures locally, so I'd say we are good - thank you 👍

Issue #4447, opened 2022-03-24
Knowing if we are in a client side navigation or direct hit


jycouet:
### Describe the problem

Let's try to describe the **Feature Request** with a small graph:

```mermaid
flowchart TB
    b3-->a2
    subgraph Page Home
    a0[New visitor]-->a1[load server]-->a2[load browser]-->a3[Enjoy the home page]
    style a0 fill:#29B56A,color:#000
    style a3 fill:#254CA0,color:#000
    end
    subgraph Page About
    b0[New visitor]-->b1[load server]-->b2[load browser]-->b3[Enjoy the about page]
    style b0 fill:#29B56A,color:#000
    style b3 fill:#254CA0,color:#000
    end
```

As you can see, I have 2 routes: Home and About.
On a direct hit, the load function is executed 2 times once `!browser` once `browser` ✅
On navigation from About to Home, the load function of Home is executed 1-time `browser` ✅

The problem: how can I know in a load function if I'm in a client side navigation or direct hit?

### Describe the proposed solution

The best to my mind would be to have something similar to `browser` like:
```ts
import { clientNavigation } from '$app/navigation';

export async function load() {
    console.log(`clientNavigation`, clientNavigation);
    return {};
}
```

When it's a direct hit, the 2 load functions would have clientNavigation to `false`.
When it's a client side navigation, the load function would have clientNavigation to `true`.

### Workaround I use today:

I have a file `clientNavigation.ts`
```ts
export let clientNavigation = false;

export const setClientNavigation = () => {
	clientNavigation = true;
};
```

And in `__layout.svelte` I have
```svelte
<script lang="ts">
	// Todo: https://github.com/sveltejs/kit/issues/4048#issuecomment-1073577686
	import { afterNavigate } from '$app/navigation';
	import { setClientNavigation } from '$lib/clientNavigation';

	afterNavigate(() => setClientNavigation());
</script>
```

And now I can use `clientNavigation` value to know if I'm in a client side navigation or direct hit.

### Importance

would make my life easier

### Additional Information

This would allow nice scenarios for my library https://github.com/jycouet/kitql awaiting or not to fill stores depending on the situation.

I would be happy to contribute (very happy actually!), just let me know:
 - If I forgot some part of the use case, things to consider, ...
 - Where to start?

mrkishi:
I wonder how common a use-case this is? Since workarounds are possible and they aren't even that hairy, we should consider if it warrants a new feature.

[...]

PR #6522, opened 2022-09-02
treat invalidation as navigation

### LINKED ISSUE #4383: `beforeNavigate` fires when the page is reloaded

### LINKED ISSUE #6375: `beforeInvalidate` and `afterInvalidate` events
### Describe the problem

I have a horizontal loader that's shown on every navigation that happens client-side (much like the one in kit.svelte.dev). It's using the `beforeNavigate` and `afterNavigate` events to appear and then disappear.

Occasionally I need to call `invalidate()` to re-run all the load functions (because the authentication status has changed, for example). I would like to show the same loader for when the load functions are re-running as well — because when you think about it, running `invalidate()` is kind of life refreshing the page.

It would be nice if SvelteKIt provided two other lifecycle functions called `beforeInvalidate` and `afterInvalidate` as well, or perhaps integrated the invalidation process into the same `beforeNavigate` and `afterNavigate` functions and then adjust the naming.

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

### LINKED ISSUE #6498: has a popState
### Describe the problem

I'm porting a react webapp to svelte, and I have a need to know when a page was loaded by push/replace or because the user pressed the browser's back button or if window.history.go(-1) method was called.

For this, I'm using my own code (below), but I would like to know if sveltekit could officially inform that.

This is working for my use case, 
but I'm sure it doesn't work for all cases

```js
import { browser } from '$app/environment'
import { afterNavigate } from '$app/navigation'

let idx = 0

export function listenPopState() {
	afterNavigate(() => {
		idx = window.history.state['sveltekit:index']
	})
}

export function isPopState() {
	return browser && idx > window.history.state['sveltekit:index']
}
```

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

### LINKED ISSUE #6517: Navigation types
### Describe the problem

Relevant: #4447, #4383, #6498.

The `navigating` store value is a `{ from: URL, to: URL }` object, which is also passed to `beforeNavigate` (along with a `cancel` method) and `afterNavigate` callbacks. 

Often, it's useful to know what _type_ of navigation is happening — whether we're coming from or going to a different document, or we navigated within the site to a different route by clicking on a link, or via `goto`, or by a `popstate` event (i.e. traversing the `history`). If it's a history traversal, it's also useful to know the delta (or at least whether we're going forward or back).

Since the types are mutually exclusive, it probably makes sense to be an enum, rather than

```
{ load: true, goto: false, ... }
```

or something of that nature. Delta can be reported in a separate property that would only apply in the `popstate` case:

```js
beforeNavigate(({ from, to, type, delta }) => {
  if (delta) {
    console.log(`travelling ${delta < 0 ? 'backwards' : 'forwards'} through time`);
  } else switch (type) {
    case 'unload': console.log('bye for now'); break;
    case 'click': console.log('clickety clack'); break;
    case 'goto': console.log('as you wish'); break;
  }
});
```

(`load` isn't pictured because it would never be fired in the `beforeNavigate` case, just like `unload` would never appear in `afterNavigate`.)

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
Update: since #6537 was merged, the only change this PR makes is to treat invalidation is equivalent to navigation for the purposes of `beforeNavigate`, `afterNavigate` and `navigating`. Leaving it here until we make a final decision, but there's some valid scepticism about whether this is the right move.

Original PR message follows.

---

closes #6517
closes #4383
closes #6498
closes #6375 
possibly #4447 too?

This adds a `type` property to the `Navigation` object found in various places:

```js
import { beforeNavigate, afterNavigate } from '$app/navigation';
import { navigating } from '$app/stores';

beforeNavigate(({ from, to, type }) => {
  console.log('beforeNavigate', type);
});

afterNavigate(({ from, to, type }) => {
  console.log('afterNavigate', type);
});

$: console.log($navigation?.type);
```

The type is one of the following:

* `load` — we just got here for the first time. `from` is `null` in this case
* `link` — the user clicked a link
* `goto` — the app called `goto` programmatically
* `popstate` — the user hit the back/forward button, or we're traversing the history some other way
* `invalidation` — the app called `invalidate(...)` or `invalidateAll()`. **This is a breaking change** — invalidations weren't previously treated as navigations
* `unload` — the page is being unloaded because the user clicked an external link, or submitted a form, or reloaded the page, etc. `to` will be `null` in this case

In the `popstate` case, there's an additional `delta` property, where -1 means navigating back and +1 means navigating forwards (but this can be any non-zero integer).

Todo

* [ ] docs

### Tests

### Changesets


aradalvand:
Correct me if I'm wrong but I believe this also closes #6375

Rich-Harris:
Yep, you're right — nice

dummdidumm (approved):
All good implementation-wise, therefore approving, but not 100% sure about adding `invalidating` to it yet.

madeleineostoja:
I can't think of a good API for this so take with a grain of salt, but IMO the only acceptable way for this to become part of navigation is if it was opt-in somehow. Requiring users to guard against `type !== 'invalidating'` in all their routing calls is asking for a footgun and even if you knew about it would get tiring real fast.

Rather than making it a `type` could it be an extra property in `Navigation`? Not as clean but much safer, ie:

```ts
interface Navigation {
  from: NavigationTarget;
  to: NavigationTarget;
  type: NavigationType;
  invalidation: boolean
  delta?: number;
}
```

It's weird that `type` would be null when invalidating then, but it's also weird that (presumably) `from` and `to` would also be null so 🤷🏻‍♀️

packages/kit/src/runtime/client/client.js
```diff
@@ -144,11 +144,24 @@ export function create_client({ target, base, trailing_slash }) {
...
-				await update(url, []);
+				if (!cancelled) {
+					stores.navigating.set(navigation);
+					await update(url, []);
+					callbacks.after_navigate.forEach((cb) => cb(navigation));
+				}
 
 				invalidating = null;
 				force_invalidation = false;
+				stores.navigating.set(null);
```
PatrickG, line 164: Can this result in a wrong `navigation` state when the user clicks on a link while the invalidating is in progress?
Would probably only be noticeable on a slow connection.

aradalvand:
I agree with @madeleineostoja, so what about something similar to [@lukaszpolowczyk's suggestion](https://github.com/sveltejs/kit/issues/6517#issuecomment-1235358605) where `beforeNavigate` and `afterNavigate` receive an `options` argument that lets the developer decide whether they want invalidation to be treated as navigation or not?
```js
beforeNavigate(({ type }) => {
    // show loader
}, { includeInvalidation: true });
```
@Rich-Harris Is this not feasible?

madeleineostoja:
I think `$navigating` is less of an issue, since you're less likely to be using it to perform an action that you don't want an invalidation trigger.

I like that proposal. It's a little clunky, but much safer at the cost of a little verbosity. I'd name the option something shorter but that's neither here nor there.

Rich-Harris:
I'm going to close this since the consensus (and I think I agree) seems to be against this solution. `$invalidating`, `beforeInvalidate`, `includeInvalidation` or any other API approach we settle on can be added in a non-breaking way later

PR #6530, opened 2022-09-02
feat: Cloudflare Pages _routes.json specification (#6441)

### LINKED ISSUE #6441: Upcoming change to Cloudflare Pages projects
### Describe the problem

# 👋 

Hi there. John here from the Cloudflare Pages team.

When a SvelteKit project is deployed to Cloudflare Pages, the server-side code is supported through Pages Functions (currently in beta). Unfortunately, by using Functions, each request must go through the server-side Functions code even if it doesn't need to. For instance, an immutable asset request to `https://kit.svelte.dev/_app/immutable/assets/_layout-ab34ca4f.css` would first have to route through Functions.

This is problematic since the request would "count" as a request to Functions even though only a static asset was served. Further, there is a slight amount of added latency.

The Cloudflare Pages team has been rolling out support to customize how project Functions are mounted and thus can save money and latency by providing this metadata on build. The `@sveltejs/adapter-cloudflare` package should expose a set of includes and excludes based on static files generated. For the time being, I suggest we be conservative and include the Functions worker on all routes but exclude on `/_app/immutable/`.

If the SvelteKit user uses the _hooks_ functionality, they may need to tune these patterns manually on a case-by-case basis.

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

We (the Cloudflare Pages team) will be contributing the relevant changes.

jrf0110:
Closes #6441 

### Tests

### Changesets


jahands (approved):
Jacob from the Cloudflare Pages team here - I'm also working on this project.

LGTM

packages/adapter-cloudflare/index.js
```diff
@@ -28,13 +29,23 @@ export default function () {
 
 			const relativePath = posix.relative(tmp, builder.getServerDirectory());
 
-			writeFileSync(
-				`${tmp}/manifest.js`,
-				`export const manifest = ${builder.generateManifest({
-					relativePath
-				})};\n\nexport const prerendered = new Set(${JSON.stringify(builder.prerendered.paths)});\n`
+			builder.log.info(
+				`adapter-cloudfare is writing its own _headers file. If you have your own, you should duplicate the headers contained in: ${dest}/_headers`
```
Rich-Harris, line 33: can we automate this, and copy the contents of `${builder.config.kit.files.assets}/_headers` into `${dest}/_headers` if it exists?

packages/adapter-cloudflare/index.js
```diff
@@ -55,3 +66,43 @@ export default function () {
...
+				// We're being conservative by not excluding all assets in
+				// /static just yet. If there are any upstream auth rules to
+				// protect certain things (e.g. a PDF that requires auth),
+				// then we wouldn't want to prevent those requests from going
+				// to the user functions worker.
+				// We do want to show an example of a _routes.json that
+				// excludes more than just /_app/immutable/*, and favicons
+				// are a reasonable choice
+				.filter(({ file }) => file.includes('favicon'))
+				.map((asset) => asset.file)
```
Rich-Harris, line 91: Can you elaborate on 'upstream auth rules'? Where would those be expressed? All static files are considered to be outside SvelteKit's purview (unless we added some platform-agnostic 'middleware' concept in future), so it feels like we could just list all static assets here

packages/kit/src/core/adapt/builder.js
```diff
@@ -113,6 +113,10 @@ export function create_builder({ config, build_data, routes, prerendered, log })
 			});
 		},
 
+		getBuildData() {
+			return build_data;
+		},
```
Rich-Harris, line 118: I think we can do everything we want in this PR without introducing new API surface area. `writeClient(dest)` returns an array of all the written files, so we could use that in place of `build_data.manifest_data.assets`, and `app_dir` can be accessed as `builder.config.kit.appDir`. 

`BuildData` isn't a public type and it would be better to keep it that way, otherwise it becomes part of the contract and is thus subject to semver

Rich-Harris (changes requested):
thank you! comments/questions inline

packages/kit/src/core/adapt/builder.js
jrf0110: Roger that. I'll remove

packages/adapter-cloudflare/index.js
jrf0110: > Where would those be expressed?

This would be the case of a Cloudflare Pages user adding logic to a catch-all functions route (e.g. `functions/_middleware.ts` or `functions/[[path.ts]]` or simply having a `_worker.js` file). They might be adding headers to static assets with their functions or protecting a certain files to only be accessible by emails ending in `@example.com`. Either way, we don't want to be _too prescriptive_ on whether or not functions runs before the asset or not.

Do note that _omitting a file from `excludes`_ will never accidentally break a user's function. The reverse is not true; A misplaced exclusion rule could prevent a function from being run that the user intended to be run. For that reason, we're intentionally being conservative on which assets are explicitly being excluded.

Further, there is a size limit for these descriptions. We're maxing out the _total number of rules_ - include or exclude - to 100. I'm sure plenty of sites would max out that rule limit with just excludes of static paths

packages/adapter-cloudflare/index.js
jrf0110: We _could_, but I think that ought to be out-of-scope of this PR. We could optimistically concatenate the auto-generated headers to an existing `${dest}/_headers` file. However, I expect most users are running their build process and then copying their `src/_headers` to `${dest}/_headers` after the build completes. We still need to think about the best way to approach this scenario

packages/adapter-cloudflare/index.js
```diff
@@ -23,18 +23,32 @@ export default function () {
 			builder.rimraf(tmp);
 			builder.mkdirp(tmp);
 
-			builder.writeClient(dest);
+			const written_files = builder.writeClient(dest);
+			console.log('written_files', written_files);
```
jrf0110, line 27: Oops. Remove

jrf0110:
Hey @Rich-Harris aside from the the automerging `_headers` file thing, I've addressed your feedback. Mind taking another gander?

Rich-Harris (approved):

Rich-Harris:
thank you!

Issue #4750, opened 2022-04-26
[Feature Request] Support trusted domains for SSR fetch with credentials


MarcusCemes:
### Describe the problem

**TLDR**: SRR fetching during data loading will **not** forward cookies to the API, if the API is not on a subdomain relative to SSR, with no reasonable option to override this behaviour. This is a delicate problem, as one small misconfiguration and the SSR server may accidentally forward client credentials to an unknown third party, effectively leaking their session key/token.

Currently, SSR fetch will **only** pass through the `cookie` header if the destination hostname has a suffix matching the SSR application's hostnames, these lines are responsible: This may be partially related to https://github.com/sveltejs/kit/issues/1777.

https://github.com/sveltejs/kit/blob/e3e93732a5eda595af4e1c77bc8649b35f03fa0b/packages/kit/src/runtime/server/page/load_node.js#L230-L247

> If the cookie is set with a domain of "example.com", the SvelteKit SSR is on "web.example.com" and the API at "api.example.com", client side data fetching will work, however, SSR will fail. If SSR is running on "example.com", credentials, notably the `cookie` header, _will_ be passed on to the API.

I imagine that the main issue is that the **client does not share domain/SameSite cookie information with the server**, only the key/value payload itself, therefore SvelteKit not able to determine whether the cookie is allowed to be shared with the third party, even if it's on the same root domain. The only way to be sure is by checking if the fetch URL is on a subdomain of the received request, only then will the cookie's domain overlap with absolute certainty.

I noticed this when moving from a localhost environment to production, everything broke. Client-side data loading would work correctly, SSR data loading would be unauthenticated, creating confusing state conflicts and _many_ hours of debugging... [Someone else](https://github.com/node-fetch/node-fetch/issues/1113#issuecomment-805302543) had the same experience a year ago.

I understand the rationale behind this, cookies and sessions are hard to keep secure at the best of times, but here the SSR behaviour diverges from the client data-loading behaviour. It's also a very difficult problem to solve, as there is **no way to override this behaviour** to the best of my knowledge.

### Describe the proposed solution

Add an **extra fetch parameter or credentials enum value** to allow the credentials to be transmitted to the URL, such as `credentials: "force"`. This would effectively be a way to tell SvelteKit/fetch _hey, I can vouch for this request, you can send the credentials as well_. I'm not sure if `node-fetch` would allow this, but I don't see why it shouldn't, as long as SvelteKit replaces the key with `credentials: "include"` and adds copies over the cookie header.

[...]

PR #6541, opened 2022-09-03
only include specified headers in server-side fetch requests and serialized responses

### LINKED ISSUE #1971: Blacklist fetch response headers when serialised for initial fetch
### Describe the problem

when serialise the fetch response into `<script type="application/json" data-type="svelte-data">`, it includes the fetch response header, which includes headers that is always changed:

- Date

which makes the html content hash different, even though the api response body and svelte rendered content is the same

blacklist some of the response header when serialised

### Alternatives considered

use `serverFetch` hook to overwrite fetch behavior on server, and remove the `Date` response header

### Importance

would make my life easier

### Additional Information

_No response_

### LINKED ISSUE #5253: Route request headers are forwarding to backend fetch calls.
### Describe the bug

Request headers for the initial HTML page load are getting passed through to external fetch requests on the backend. For example, if `http://localhost:3000` is requested with the `Accept: text/html` header, that gets passed along to my API requests.

This occurs when using `@sveltejs/node-adapter@1.0.0-next.78`.  I've included a reproduction, but here are some more details:

I'm seeing this in my server logs within my API, however it is also observable in the `externalFetch` svelte-kit hook:

```javascript
// hooks.js
export function externalFetch(request) {
  console.log(request.headers) // <-- this has request headers from my svelte-kit route!
  return fetch(request)
}
```

This is unexpected and causes some wonky interactions with our API. We've hit this in two places:

1. A client passes through `Accept: text/html` to our API endpoint, which results in our JSON API responding with a `406 Not Acceptable` error. This is fine 🤷, but pretty weird and an annoying class of errors to filter through in error tracking software.
2. More frustrating: a client passes through a` If-Modified-Since` header, our API returns a 304 Not Modified response, which results in a user facing error when `fetch` on the backend considers the request a failure:

```javascript
let request = await fetch("https://myapi.com/thing.json")

console.log(request.status) // 304

if (request.ok) { // this is false server-side
  // ...
} 
```

We'd expect these headers to not get automatically passed through. Those headers are for the SvelteKit request, not our API.

As a temporary work around, we're removing these headers using the `externalFetch` hook, but I think that this should not be necessary.

### Reproduction

I have provided a reproducing project here:
https://github.com/nhunzaker/sveltekit-reproduction

You can reproduce this issue by running the project with:

```
npm install
npm run dev
```

Then cURL the request with the Accept header:

```
curl --header "Accept: text/html" http://localhost:3000
```

When running the example, you'll see the request come through with the `Accept` header sent along with the request for `http://localhost:3000`:

```
External Fetch:
https://www.gov.uk/bank-holidays.json
Accept Header: text/html
```

(UK Bank holidays are not related to this issue, they simply have a public JSON API)

### Logs

_No response_

### System Info

```shell
System:
    Memory: 3.99 GB / 7.72 GB
    Container: Yes
    Shell: 3.1.0 - /usr/bin/fish
  Binaries:
    Node: 16.13.0 - ~/.asdf/installs/nodejs/16.13.0/bin/node
    Yarn: 1.22.17 - ~/.asdf/installs/nodejs/16.13.0/.npm/bin/yarn
    npm: 8.1.0 - ~/.asdf/plugins/nodejs/shims/npm
  npmPackages:
    @sveltejs/adapter-node: 1.0.0-next.78 => 1.0.0-next.78
    @sveltejs/kit: 1.0.0-next.350 => 1.0.0-next.350
    svelte: ^3.44.1 => 3.48.0
```

### Severity

### Additional Information

_No response_

Rich-Harris:
Closes #5253
Closes #5195
Closes #1971

This is a topic that has been discussed a lot on this repo without an adequate resolution: when making server-side `fetch` calls within a `load` function, how which (if any) of the original request headers do we include?

In the interests of making things 'just work' as far as possible, we've been including almost all headers ever since https://github.com/sveltejs/kit/pull/3631. But this causes its own problems, such as https://github.com/sveltejs/kit/issues/5253. A PR from @johnnysprinkles (#5195) removes all header forwarding, but has sat unmerged because it depends on adding the `event` object to the `load` inputs, which we definitely don't want.

This PR takes a different approach (basically the one suggested by @nhunzaker in https://github.com/sveltejs/kit/issues/5253#issuecomment-1165499804) — it allows app developers to create an allowlist of headers that should be forwarded in server-side fetches.

At the same time, we add a list of headers that should be included when we serialize the response, which means that things like `Date` don't mess up content hashing (which prevents us from responding with 304s — #1971):

```js
// svelte.config.js
export default {
  kit: {
    fetch: {
      forwardedRequestHeaders: ['x-my-custom-auth-header'],
      serializedResponseHeaders: ['x-my-custom-response-header']
    }
  }
};
```

Questions:

* Are these the right names for the options?
* What should the defaults be? Right now I included `content-type` in `serializedResponseHeaders` to get the tests to pass, but presumably there are other sensible defaults. To _some_ degree it doesn't matter, but it would be nice if most people didn't need to touch this option
* Should we go even further and remove the special handling for `cookie` and `authorization` headers when `credentials` is something other than `omit`?

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/fetch.js
+++ b/packages/kit/src/runtime/server/page/fetch.js
@@ -49,17 +49,11 @@ export function create_fetch({ event, options, state, route, prerender_default }
 		opts.headers = new Headers(opts.headers);
 
 		// merge headers from request
-		for (const [key, value] of event.request.headers) {
-			if (
-				key !== 'authorization' &&
-				key !== 'connection' &&
-				key !== 'content-length' &&
-				key !== 'cookie' &&
-				key !== 'host' &&
-				key !== 'if-none-match' &&
-				!opts.headers.has(key)
-			) {
-				opts.headers.set(key, value);
+		for (const header of options.fetch.included_request_headers) {
+			const value = event.request.headers.get(header);
+
+			if (value !== null && !opts.headers.has(header)) {
+				opts.headers.set(header, value);
 			}
 		}
 
@@ -192,34 +186,15 @@ export function create_fetch({ event, options, state, route, prerender_default }
 				async function text() {
 					const body = await response.text();
 
-					// TODO just pass `response.headers`, for processing inside `serialize_data`
-					/** @type {import('types').ResponseHeaders} */
-					const headers = {};
-					for (const [key, value] of response.headers) {
-						// TODO skip others besides set-cookie and etag?
-						if (key !== 'set-cookie' && key !== 'etag') {
-							headers[key] = value;
-						}
-					}
-
 					if (!opts.body || typeof opts.body === 'string') {
-						const status_number = Number(response.status);
-						if (isNaN(status_number)) {
-							throw new Error(
-								`response.status is not a number. value: "${
-									response.status
-								}" type: ${typeof response.status}`
-							);
-						}
-
 						fetched.push({
 							url: requested,
 							method: opts.method || 'GET',
 							body: opts.body,
 							response: {
-								status: status_number,
+								status: response.status,
 								statusText: response.statusText,
-								headers,
+								headers: response.headers,
 								body
 							}
 						});
--- a/packages/kit/src/runtime/server/page/fetch.js
+++ b/packages/kit/src/runtime/server/page/fetch.js
@@ -49,7 +49,7 @@ export function create_fetch({ event, options, state, route, prerender_default }
 		opts.headers = new Headers(opts.headers);
 
 		// merge headers from request
-		for (const header of options.fetch.included_request_headers) {
+		for (const header of options.fetch.forwarded_request_headers) {
 			const value = event.request.headers.get(header);
 
 			if (value !== null && !opts.headers.has(header)) {
--- a/.changeset/kind-ducks-vanish.md
+++ b/.changeset/kind-ducks-vanish.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] only include specified headers in server-side fetch requests and serialized responses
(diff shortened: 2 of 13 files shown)


packages/kit/src/runtime/server/page/fetch.js
```diff
@@ -49,17 +49,11 @@ export function create_fetch({ event, options, state, route, prerender_default }
 		opts.headers = new Headers(opts.headers);
 
 		// merge headers from request
-		for (const [key, value] of event.request.headers) {
-			if (
-				key !== 'authorization' &&
```
benmccann, line 54: I wonder if it'd be worth having a test ensuring these aren't set by default just so that if we ever decide we need to go back to a blacklist we don't accidentally add these back or to make sure we don't accidentally add these to a default whitelist in the future

dummdidumm:
I'm wondering if a static list is enough. Could there be use cases for "I want this forwarded for urls from X but not those from Y", which would hint at making this a hook somehow (which I don't know if it's possible)

Rich-Harris:
Yeah, I think a hook is the way to go. @Conduitry had a suggestion I really like, which solves the 'I want this forwarded for urls from X but not those from Y' problem while also providing a solution to https://github.com/sveltejs/kit/issues/4750 — we get rid of `externalFetch` and add a new `handleFetch` hook that gets passed the default `fetch` implementation, like so:

```js
export function handleFetch({ event, request, fetch }) {
  // `event` is the underlying `RequestEvent`
  // `request` is a `Request` object the normalizes the arguments to `fetch` in `load`
  // `fetch` is our normal fetch logic — get data from our own routes, or delegate to undici/whatever
}
```

This would allow us to get rid of `forwardedRequestHeaders`, though we'd still need `serializedResponseHeaders`

Rich-Harris:
closing in favour of #6565 and #6569

PR #6548, opened 2022-09-03
make test less flaky, hopefully

Rich-Harris:
this one keeps failing in firefox


dummdidumm (approved):
Makes sense, in general `page.locator)).toHaveText` feels more robust in many cases

PR #6550, opened 2022-09-03
simulate CORS errors in server-side fetch

### LINKED ISSUE #5074: fetch() unexpected behavor when encountering a CORS issue on client side
### Describe the bug

If SvelteKit's `fetch()` is called against a URL that causes a CORS error in a web browser, the resulting behaviour might be unexpected for a developer. The behaviour is also naturally different on client-side and server-side rendering, where SSR `fetch()` does not suffer from CORS issues.

- A CORS error does not return a response in `fetch()` but throws an exception. Developers relying on `response.ok` pattern to check the validity of the `fetch` result are caught off guard.
- This is a tough nut to crack, because it is rooted deeply in the web browser security model.
- The behaviour is unexpected because you need to work around using `try ... catch` instead of `response.ok` attribute to check the error.
- Although something must "have gone wrong before" this occurs, the unexpected behaviour is common, because many times HTTP 404 or HTTP 500 responses lack CORS headers on them. If a developer naively checks for `response.status === 404` the check might fail in a web browser, assuming a low quality third party API. This can happen if a developer is using `fetch()` in `load()`, `get()` or similar with third-party APIs. This is especially bad, when SSR renders different pages (w/working error handling) than the client-side router.

### Reproduction

This issue cannot be demostrated on StackBlitz (node.new) because it is related to low-level network primitives.

Here is an example one can throw in their `load()` function:

```typescript

            try {
                // causes CORS error because lack of CORS headers in 404
                resp = await fetch("https://example.com/random");
            } catch(e) {
                // A web browser throws TypeError: Failed to fetch
                // but happens only on client-side.
                // The exception is hard to distinguish from
                // any other exception, because it lacks metadata
                // (class name, attributes).
                console.error("Received", e)
                // Temporary workaround
                resp = {ok: false, statusText: e.message};
            }

	let error, meta;
	if (resp.ok) {
		meta = await resp.json();
		error = null;
	} else {
		meta = {};
		error = resp.statusText;
	}
```

### Logs

You will see this in the web developer console:

```
Access to fetch at 'https://example.com/random' from origin 'http://localhost:3000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

However I am not sure if this information is accessible within client-side JavaScript.

### System Info

```shell
System:
    OS: macOS 12.3.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 902.50 MB / 64.00 GB
    Shell: 5.8.1 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 16.14.2 - ~/.nvm/versions/node/v16.14.2/bin/node
    Yarn: 1.22.15 - ~/.nvm/versions/node/v16.14.2/bin/yarn
    npm: 8.7.0 - ~/.nvm/versions/node/v16.14.2/bin/npm
  Browsers:
    Chrome: 101.0.4951.64
    Firefox: 100.0.2
    Safari: 15.4
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.34 
    @sveltejs/kit: next => 1.0.0-next.308 
    svelte: ^3.46.0 => 3.46.6
```

### Severity

annoyance

### Additional Information

I am not sure what would be the best way to address the issue. 

If this issue is not fixable on a framework level, I suggest this issue is highlighted in
- tutorial
- examples
- documentation
- etc. 

and developers are educated that `fetch()` may throw an exception, besides returning error responses.

There are likely to be other similar low-level network issues in a web browser that needs to be handled via catch in a `fetch()` caller.

Rich-Harris:
closes #5074, by increasing the fidelity of the server-side `fetch` implementation passed to `load`. If a request is made for an external resource, and it doesn't respond with an appropriate `access-control-allow-origin` header, it will fail the same way that the browser's `fetch` fails (albeit with a different error message, since there's no standardisation here). 

If the request was made with `no-cors`, we return a response with an empty body.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/fetch.js
+++ b/packages/kit/src/runtime/server/page/fetch.js
@@ -163,7 +163,7 @@ export function create_fetch({ event, options, state, route, prerender_default }
 			// - sub.my.domain.com WILL receive cookies
 			// ports do not affect the resolution
 			// leading dot prevents mydomain.com matching domain.com
-			if (`.$url.hostname}`.endsWith(`.${event.url.hostname}`) && opts.credentials !== 'omit') {
+			if (`.${url.hostname}`.endsWith(`.${event.url.hostname}`) && opts.credentials !== 'omit') {
 				const cookie = event.request.headers.get('cookie');
 				if (cookie) opts.headers.set('cookie', cookie);
 			}
--- a/packages/kit/src/runtime/server/page/fetch.js
+++ b/packages/kit/src/runtime/server/page/fetch.js
@@ -152,6 +152,8 @@ export function create_fetch({ event, options, state, route, prerender_default }
 				requested = event.url.protocol + requested;
 			}
 
+			const url = new URL(requested);
+
 			// external fetch
 			// allow cookie passthrough for "same-origin"
 			// if SvelteKit is serving my.domain.com:
@@ -161,10 +163,7 @@ export function create_fetch({ event, options, state, route, prerender_default }
 			// - sub.my.domain.com WILL receive cookies
 			// ports do not affect the resolution
 			// leading dot prevents mydomain.com matching domain.com
-			if (
-				`.${new URL(requested).hostname}`.endsWith(`.${event.url.hostname}`) &&
-				opts.credentials !== 'omit'
-			) {
+			if (`.$url.hostname}`.endsWith(`.${event.url.hostname}`) && opts.credentials !== 'omit') {
 				const cookie = event.request.headers.get('cookie');
 				if (cookie) opts.headers.set('cookie', cookie);
 			}
@@ -176,6 +175,25 @@ export function create_fetch({ event, options, state, route, prerender_default }
 
 			const external_request = new Request(requested, /** @type {RequestInit} */ (opts));
 			response = await options.hooks.externalFetch.call(null, external_request);
+
+			if (opts.mode === 'no-cors') {
+				response = new Response('', {
+					status: response.status,
+					statusText: response.statusText,
+					headers: response.headers
+				});
+			} else {
+				if (url.origin !== event.url.origin) {
+					const acao = response.headers.get('access-control-allow-origin');
+					if (!acao || (acao !== event.url.origin && acao !== '*')) {
+						throw new Error(
+							`CORS error: ${
+								acao ? 'Incorrect' : 'No'
+							} 'Access-Control-Allow-Origin' header is present on the requested resource`
+						);
+					}
+				}
+			}
 		}
 
 		const set_cookie = response.headers.get('set-cookie');
--- a/.changeset/red-tomatoes-leave.md
+++ b/.changeset/red-tomatoes-leave.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] simulate CORS errors in server-side fetch
--- a/packages/kit/test/apps/basics/src/routes/load/cors/+page.js
+++ b/packages/kit/test/apps/basics/src/routes/load/cors/+page.js
@@ -1,0 +1,8 @@
+/** @type {import('./$types').PageLoad} */
+export async function load({ fetch, url }) {
+	const res = await fetch(`http://localhost:${url.searchParams.get('port')}`);
+
+	return {
+		text: await res.text()
+	};
+}
--- a/packages/kit/test/apps/basics/src/routes/load/cors/+page.svelte
+++ b/packages/kit/test/apps/basics/src/routes/load/cors/+page.svelte
@@ -1,0 +1,6 @@
+<script>
+	/** @type {import('./$types').PageData} */
+	export let data;
+</script>
+
+<h1>{@html data.text}</h1>
--- a/packages/kit/test/apps/basics/src/routes/load/cors/no-cors/+page.js
+++ b/packages/kit/test/apps/basics/src/routes/load/cors/no-cors/+page.js
@@ -1,0 +1,10 @@
+/** @type {import('./$types').PageLoad} */
+export async function load({ fetch, url }) {
+	const res = await fetch(`http://localhost:${url.searchParams.get('port')}`, {
+		mode: 'no-cors'
+	});
+
+	return {
+		text: await res.text()
+	};
+}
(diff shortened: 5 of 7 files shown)


packages/kit/src/runtime/server/page/fetch.js
```diff
@@ -161,10 +163,7 @@ export function create_fetch({ event, options, state, route, prerender_default }
 			// - sub.my.domain.com WILL receive cookies
 			// ports do not affect the resolution
 			// leading dot prevents mydomain.com matching domain.com
-			if (
-				`.${new URL(requested).hostname}`.endsWith(`.${event.url.hostname}`) &&
-				opts.credentials !== 'omit'
-			) {
+			if (`.$url.hostname}`.endsWith(`.${event.url.hostname}`) && opts.credentials !== 'omit') {
```
dummdidumm, line 166: I think there's a missing `{` - which, if I'm correct , makes we wonder how tests are still all passing
```suggestion
			if (`.${url.hostname}`.endsWith(`.${event.url.hostname}`) && opts.credentials !== 'omit') {
```

Rich-Harris: whoops. i think this behaviour is just untested

dummdidumm (approved):

packages/kit/src/runtime/server/page/fetch.js
```diff
@@ -176,6 +175,25 @@ export function create_fetch({ event, options, state, route, prerender_default }
 
 			const external_request = new Request(requested, /** @type {RequestInit} */ (opts));
 			response = await options.hooks.externalFetch.call(null, external_request);
+
+			if (opts.mode === 'no-cors') {
```
dummdidumm, line 179: Note for later: If we add `fetch` to `load` of `+page/layout.server` at some point, this needs to be checked because on the server it's ok to fetch cross origin.

kevinrenskers:
Does SvelteKit send the `Origin` header along with any requests made using the supplied `fetch` function, or do we have to do that ourselves? I am running into the issue now where I am seeing a (nicely looking) "CORS error: No 'Access-Control-Allow-Origin' header is present on the requested resource" error page, even though my server is configured to return CORS headers. But it only does this if the `Origin` is set.

Tyrenn:
Following @kevinrenskers comment, it seems that the Origin header is correctly added when fetching with POST but not with GET method.
I've just ran into this error myself.

Fixing it by manually adding the Origin header but I guess GET requests should have it automatically too

packages/kit/src/runtime/server/page/fetch.js
```diff
@@ -176,6 +175,25 @@ export function create_fetch({ event, options, state, route, prerender_default }
...
+				response = new Response('', {
+					status: response.status,
+					statusText: response.statusText,
+					headers: response.headers
+				});
+			} else {
+				if (url.origin !== event.url.origin) {
```
Cluster2a, line 186: @Rich-Harris, if url.origin = 'http://api.app.domain.io and event.url.origin = 'http://app.domain.io' this will result in checking the cors, even if this is not considered as a cross-origin request. Or am I wrong?

mattpilott:
I've been able to get around this error, would someone be able to post example code?
```js
   const res = await fetch(url, {
      headers: {
         'Access-Control-Allow-Origin': '*',
      }
   })
   const json = await res.json()

   if (res.ok) return json

   throw error(404, 'Not found')
}
```

aradalvand:
I just updated SvelteKit and I'm getting a CORS error that didn't previously exist (even in the browser's fetch):
image

How should this be solved?

mattpilott:
@aradalvand i too tried in the browser in onMount and same error, i have the same code working on a deploy from awhile back, i'm interested to see the workaround code from either @Tyrenn @kevinrenskers

aradalvand:
> Following @kevinrenskers comment, it seems that the Origin header is correctly added when fetching with POST but not with GET method.

@Tyrenn That doesn't seem to be the case. I'm sending POST requests and the Origin header isn't added automatically either.

kevinrenskers:
> @aradalvand i'm interested to see the workaround code from either @Tyrenn @kevinrenskers

Just add the Origin header yourself.

```
const config: RequestInit = { method: "get", headers:  { origin: "https://www.your-website.com" }};
const request = new Request("https://www.example.com", config);
const response = await fetch(request);
```

mattpilott:
@kevinrenskers i tried this:

```js
   const config = { method: 'get', headers: { origin: 'http://localhost:5173' } }
   const request = new Request(url, config)
   const res = await fetch(request)
   const json = await res.json()

   if (res.ok) return json

   throw error(404, 'Not found')
}
```
but no dice 😔

kevinrenskers:
You need to make sure that the server actually returns the `Access-Control-Allow-Origin` header. Best to try out with an API tool like [paw](https://paw.cloud) or [postman](https://www.postman.com).

kevinrenskers:
Well, there's your problem :)

mattpilott:
Does that mean it cant work? Even though i have a build that works? (kit 287)

Cluster2a:
> Does that mean it cant work? Even though i have a build that works? (kit 287)

It means that the API you are using should return this header. Do you have control over your API?

mattpilott:
@Cluster2a no control over the api unfortunately

bothness:
This change is causing some of my fetch requests that worked on previous versions of Svelte Kit (and with browser fetch) to fail with 500 errors.

In one specific example I was fetching a Google sheet as CSV. It doesn't seem to have an explicit 'Access-Control-Allow-Origin' header, but works fine with browser fetch.

I obviously can't control what headers Google returns, so this change basically breaks my app.

BetaConnector:
> @Cluster2a no control over the api unfortunately

Did you try to add mode = 'no-cors' as an option to your fetch?
If you don't need any credentials, you might be fine.

ivands:
When testing sveltekit on localhost with an API on example.com this will now throw an error.
Is there a way stop simulating the error for localhost?

packages/kit/src/runtime/server/page/fetch.js
MrTango: I was wondering if it should work, to do a fetch in +layout.server.js or +page.server.js, because they are both server side and should not care about CORS.

PR #6552, opened 2022-09-03
augment `Navigation` type

### LINKED ISSUE #6551: More information in `Navigation`
### Describe the problem

I've occasionally wondered if the `navigating` store, and the `from` and `to` objects passed to `beforeNavigate` and `afterNavigate`, need more information than just the URL.

#6331 provides a good example of a use case. Inside `beforeNavigate` we need to know the before/after `routeId`.

Changing `Navigation` from this...

```ts
export interface Navigation {
  from: URL | null;
  to: URL | null;
  type: NavigationType;
  delta?: number;
}
```

...to this:

```ts
export interface NavigationTarget {
  params: Params;
  routeId: string;
  url: URL;
}

export interface Navigation {
  from: NavigationTarget | null;
  to: NavigationTarget | null;
  type: NavigationType;
  delta?: number;
}
```

### Alternatives considered

`fromRouteId`, `toRouteId`, `fromParams`, `toParams`, etc. Ugh.

### Importance

nice to have

### Additional Information

_No response_

Rich-Harris:
closes #6551 — gives `beforeNavigate` and `afterNavigate` more information about which route they are navigating from and to

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -145,7 +145,8 @@ export function create_client({ target, base, trailing_slash }) {
 			const url = new URL(location.href);
 
 			invalidating = Promise.resolve().then(async () => {
-				await update(url, []);
+				const intent = get_navigation_intent(url);
+				await update(intent, url, []);
 
 				invalidating = null;
 				force_invalidation = false;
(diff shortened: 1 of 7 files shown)


packages/kit/src/runtime/client/client.js
```diff
@@ -1434,3 +1458,35 @@ async function load_data(url, invalid) {
...
+
+/**
+ * @param {'from' | 'to'} type
+ * @param {import('types').NavigationTarget} target
+ */
+function add_url_properties(type, target) {
+	for (const prop of properties) {
+		Object.defineProperty(target, prop, {
+			get() {
+				throw new Error(`${type}.${prop} should now be ${type}.url.${prop}`);
```
dummdidumm, line 1486:
```suggestion
				throw new Error(
					`The navigation shape changed - ${type}.${prop} should now be ${type}.url.${prop}`
				);
```

dummdidumm (approved):

PR #6555, opened 2022-09-03
Feat/hydrated

jycouet:
Issue link: #4447

Adding: 
```
import { hydrated } from '$app/environment';
```
To be able in a `load` function to know if the app was already hydrated or not.

---

### Tests

### Changesets


Rich-Harris:
Thank you. I'm not sure we really need this though — it's very easy to do in userland:

```js
// src/lib/hydrated.js
export let hydrated = false

export function update() {
  hydrated = true;
}
```

```svelte
<!-- src/routes/+layout.svelte -->
<script>
  import { update } from '$lib/hydrated.js';

  onMount(update);
</script>
```

Thereafter, any `load` function can access the value. Given how straightforward this is, and the general principle that we should avoid adding any non-essential API surface area, it doesn't feel justified to me. Am I missing something?

jycouet:
## Feedback
That's very straightforward for userland, I like it.

As a lib author, that means that we can't rely on a `kit` thing, and we have to teach our users to: 
1. add a `src/routes/+layout.svelte`
2. `init` our lib _(and do the hydrated logic there to rely on it later)_
```html
<!-- src/routes/+layout.svelte -->
<script>
  import { init } from 'MyLib';

  onMount(init);
</script>
```
These manual user steps are not ideal.

## Our solution
Thx to @AlecAivazis vite plugin, we will be able to inject this bit of code without any manual action. 
So I agree, let's not augment the API surface area. 
Therefore I'll close this PR and the issue linked to it.

## _Side comments_
- Great job, contributing to kit is very straightforward 🧡
- Turborepo is amazingly fast! ⚡⚡⚡
- My code almost made it!!! 😅 ^^

lukaszpolowczyk:
@jycouet Could you show the code for this vite plugin that does exactly what Rich's code does with `src/lib/hydrated.js` and `src/routes/+layout.svelte`?

I need an example of such simple code to do something different, but which should initialize in the same way.

jycouet:
This exact code is not done yet.
But it will be in Houdini's vite plugin ([here](https://github.com/HoudiniGraphql/houdini/tree/main/src/vite))

Will you need this for a library? 
Should we build an "`injector vite plugin`" to help others?

lukaszpolowczyk:
I need this under a script that I only use during dev. Now I init it as in Rich's example.

I am always in favor of universal automation where possible.

Issue #6337, opened 2022-08-27
Resolve relative `fetch` from `+layout` relative to the layout's route, not the leaf's route


tcc-sejohnson:
### Describe the problem

Pardon the dense title. I'm sure someone else could word it better.

Consider the following file structure:

```
src/
└── routes/
    ├── +layout.svelte
    ├── data/
    │   └── +server.ts
    ├── foo/
    │   └── foo2/
    │       └── foo3/
    │           └── foo4/
    │               └── +page.svelte
    └── bar/
        └── bar2/
            └── bar3/
                └── bar4/
                    └── +page.svelte
```

Now let's say we make a relative `fetch` in `+layout.svelte`: `const resp = await fetch('data');`

When visiting `/foo/foo2/foo3/foo4`, where do you suppose that fetch goes?

Something like `/foo/foo2/foo3/data`. _That's_ confusing. The unfortunate thing is, this becomes more annoying if you have a `layout` further down the tree, as the only way to reliably fetch an adjacent resource is to specify the whole path from the root manually, whereas if you actually _wanted_ the above relative behavior, you could easily achieve it yourself: ``await fetch(`${$page.url.pathname}/data`)``. 

### Describe the proposed solution

Relative `fetch`es in `+layout` files should resolve relative to the layout's directory, not relative to whatever leaf happens to be rendering.

### Alternatives considered

Just be annoyed. 🤷🏻 

### Importance

would make my life easier

### Additional Information

It's like, almost 3AM. If I've said anything stupid, please correct me.

Rich-Harris:
This makes intuitive sense, though a surprisingly non-trivial question is 'relative to what'? In a case like this...

```
src/routes/
├ [project]/
│ ├ [...files]/
│ │ ├ actions/
│ │ │ ├ embed/
│ │ │ │ └ +page.svelte
│ │ └ +layout.svelte
```

...if you visit `/project/foo/bar/baz/actions/embed` how does the layout 'know' that it should fetch relative to `/project/foo/bar/baz`? Obviously it's solvable, but it would require some changes to how we do things internally that might be slightly tricky.

But should it resolve relative to `/project/foo/bar/baz`, or to `/project/foo/bar/baz/`? Those are two very different things. Should it be driven by the `trailingSlash`, and if so what if it's `ignore`? We need to be able to answer this question, and with a solid rationale.

Rich-Harris:
[...]

PR #6559, opened 2022-09-04
`fetch` relative to layouts

Rich-Harris:
#6337. Just a failing test for now, as I realised that there's no sense proceeding until #6541 is merged.

Couple of thoughts:

* I think the only real way to make this work is to include a regex with each layout
* In #6337 we concluded that a `fetch` in `src/routes/foo/+layout.js` should resolve relative to `/foo/`, not `/foo`. Since it would be very weird if `src/routes/foo/+page.js` then resolved relative to `/foo`, it probably makes sense for _all_ relative fetches to resolve as though the trailing slash were there.

### Tests

### Changesets


Rich-Harris:
we decided against this

PR #6565, opened 2022-09-04
replace externalFetch with handleFetch

### LINKED ISSUE #4750: [Feature Request] Support trusted domains for SSR fetch with credentials

### LINKED ISSUE #5253: Route request headers are forwarding to backend fetch calls.

Rich-Harris:
# Migration guide

`externalFetch` has been replaced with `handleFetch`. Whereas `externalFetch` only ran when your `load` function contained a `fetch` call for an external URL, `handleFetch` runs for _all_ `fetch` calls in `load` that run on the server. Instead of receiving a `request` argument, it receives `{ event, request, fetch }`, where `event` is the underlying `RequestEvent` and `fetch` is SvelteKit's [internal `fetch` implementation](https://kit.svelte.dev/docs/load#input-methods-fetch).

```diff
-/** @type {import('@sveltejs/kit').ExternalFetch} */
-export function externalFetch(request) {
+/** @type {import('@sveltejs/kit').HandleFetch} */
+export function handleFetch({ event, request, fetch }) {
+  const internal = request.url.startsWith(event.url.origin + '/');
+  if (internal) return fetch(request);
  // handle external request
}
```

---

# Original PR description

Supersedes part of #6541 
Closes #5253
Closes #5195
Closes #4750

This replaces `externalFetch` with a new `handleFetch` hook, which gives developers the ability to forward whichever headers they need on a case-by-case basis — so for example if your `load` function looks like this...

```js
  const res = await fetch('/whatever');
  // ...
}
```

...the `fetch` will be made without any non-default headers, but you can add custom headers inside `handleFetch`:

```js
// src/hooks.js
const forwarded = [
  'x-my-custom-auth-header'
];

export async function handleFetch({ event, request, fetch }) {
  assert.ok(request instanceof Request);
  assert.equal(request.url, event.url.origin + '/whatever');

  for (const header of forwarded) {
    const value = event.request.headers.has(header);
    if (value !== null && !request.headers.has(header)) {
      request.headers.set(header, value);
    }
  }

  return fetch(request);
}
```

If unimplemented, defaults to `({ request, fetch }) => fetch(request)`.

### Tests

### Changesets


Conduitry:
Do you think it makes sense to call out the cookie sibling subdomain thing explicitly in the docs? Can we do that without it being too confusing? We've heard from a few people that it took then a while to even understand why their app wasn't working, so this might also be a good opportunity to tell people how the default conservative cookie passing works.

Rich-Harris:
Excellent point. Does https://github.com/sveltejs/kit/pull/6565/commits/b96a5fd387b7319f647031cf021886fb91ab9d87 spark joy? (I will settle for 'it is coherent and correct')

Conduitry:
Maybe write the code to check `origin` instead of substrings of the URL? Other than that, sounds good to me. :+1:

Rich-Harris:
in the code or in the docs?

Conduitry:
Uh. I had meant in the code, but in the docs might suffice, as long as they are careful not to instantiate URLs from relative `request.url`s. What you have here now might be perfectly fine, actually, because it's not going to be ambiguous anyway. I'm good with it if you don't want to make it more complicated.

Rich-Harris:
> as long as they are careful not to instantiate URLs from relative `request.url`s

There's no such thing — if you create a `Request` on the server, the URL must be fully resolved or Undici will throw — that's what [this line](https://github.com/sveltejs/kit/pull/6565/files#diff-3f12c17057d74a7d75f121baea53823d2ebfefa033e5cb834c05f6edca1a847dR267) is for

packages/kit/src/runtime/server/page/fetch.js
```diff
@@ -23,178 +22,144 @@ export function create_fetch({ event, options, state, route, prerender_default }
...
-							`CORS error: ${
-								acao ? 'Incorrect' : 'No'
-							} 'Access-Control-Allow-Origin' header is present on the requested resource`
-						);
+					const cookie = Object.entries(combined_cookies)
+						.map(([name, value]) => `${name}=${value}`)
+						.join('; ');
+
+					if (cookie) {
+						request.headers.set('cookie', cookie);
```
Conduitry, line 132: If the user sets a cookie in `handleFetch` and then calls the built-in `fetch` that's passed to it, will this clobber whatever cookie header they'd set?

Rich-Harris: bugger. yes, it will. do we try and merge cookies, or if someone explicitly set `cookie` already do we just leave it be?

Conduitry: After a bit more time reading the code: It looks like the `cookies` array is to make sure that we copy over the cookies from `set-cookie` in responses to the cookies we send in subsequent `fetch`es as part of this same `load` - is that correct? I'd missed that the first couple of times around, and there's probably a better name for that. (`new_cookies`? I don't know.)

I _think_ it still makes sense to automatically overlay the `set-cookie` cookies on whatever underlying cookies we had. If someone's `handleFetch` calls `fetch` with a `cookie` header, we just use that as `initial_cookies`, rather than an all-or-nothing of the `cookie` header from the original incoming request. Does that sound reasonable?

Looking at the code, I haven't quite worked out what would happen if someone decided to use `globalThis.fetch` in `handleFetch`. Would it just skip the `cookie` and `authorization` stuff, but still create the `Response` proxy so that if the user's `load` uses it, it gets serialized into the HTML? If that all still works, then that sounds like a reasonable escape hatch, and I feel more comfortable saying that the `handleFetch` `fetch` should overlay the `set-cookie` cookies unconditionally without any way to control that, and it would just overlay them on top of whatever `cookie` header was passed to `fetch` (if any), or else on top of whatever `cookie` header was present on the initial incoming request (if the domains match by our old logic), or else on top of nothing.

Rich-Harris: > Would it just skip the cookie and authorization stuff, but still create the Response proxy so that if the user's load uses it, it gets serialized into the HTML?

Yep. How does https://github.com/sveltejs/kit/pull/6565/commits/f24ba0dc0fb9011648817617cffb58bd305ee4e6 look?

Conduitry: I'm misunderstanding the code, or else I don't think this looks right. If `url.origin !== event.url.origin`, it doesn't look like this will respect the `new_cookies`.

I _think_ everything else looks right, but it's a bit hard to tell because a lot of this code looks like it's more complicated than it needs to be. I think it could be a bit more DRY, but that doesn't have to happen now, and I don't have specific recommendations of what to change.

In the case where there are both automatically inherited cookies and an explicitly specified `cookie` header, my suggestion above had been to completely replace the automatically inherited cookies with the explicitly specified ones, and I see that this will instead merge them, but I don't have a strong opinion on which would be less confusing. I don't know whether we need to distinguish (somehow) between `cookie` headers that came from `handleFetch` versus those that came from the `fetch` call in `load`. I hope not. How does the real browser `fetch` behave when you manually pass it a `cookie` header? What does that do to the cookies it would have automatically sent over?

Rich-Harris: I _think_ I fixed that

> How does the real browser fetch behave when you manually pass it a cookie header? 

Nothing, it's a [forbidden header name](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name). It's silently ignored

Conduitry: Looks good to me!

nhunzaker (commented):
Reading over the documentation changes and code, I think this'll work great for my use case. Thank you for considering it and making changes to the API.

PR #6583, opened 2022-09-05
bump vite

Rich-Harris:
3.1.0 is released 🎉


.changeset/famous-terms-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
```
dominikg, line 2:
```suggestion
'create-svelte': patch
'@sveltejs/kit': patch
```

the templates in create-svelte are updated with `^3.1.0` too

PR #6586, opened 2022-09-05
[breaking] error handling rework

dummdidumm:
# How to migrate

### Step 1: Rename hooks file

There's now also a hooks file for the client. `kit.files.hooks` is therefore now `{ server: string; client: string }` instead of `string`. So if you have used that setting, you need to adjust your config:

```diff
// svelte.config.js
export default {
  kit: {
-    hooks: 'src/myhooks'
+   hooks: { server: 'src/myhooks' }
```

If you didn't use that setting, then by default the server hooks now live in `src/hooks.server`, so you need to rename the file:
```diff
|- src
-   |- hooks.js
+   |- hooks.server.js
```

### Step 2: Adjust types

The interface `HandleError` has been renamed to `HandleServerError`, to disambiguate it from the new interface `HandleClientError`. The error type has been changed to `unknown`. Previously it was typed `Error`, but we can't know if this is always the case, so we align with TypeScript's type in `catch` here. This likely means you need to do a type cast, either by checking if it indeed is an `Error`, or just trusting your (third party) code to always throw `Error`s:

```diff
- import type { HandleError } from '@sveltejs/kit';
+ import type { HandleServerError } from '@sveltejs/kit';

- export const handleError: HandleError = ({ error }) => {
+ export const handleError: HandleServerError = ({ error }) => {
+    if (error instanceof Error) {
         SuperCoolErrorService.log(error);
+   }      
};
```

### Step 3: Define the desired error shape

The shape of `$page.error` is now defined by the `handleError` hook, which exists on the server and the client. `$page.error` is of type `App.PageError`, which you can define yourself in `app.d.ts`:

```ts
declare namespace App {
  // ...
  interface PageError {
    message: string; // this property is always required, to provide a sensible fallback
    whateverYouWant: string;
  }
```

You then need to make sure that the `handleError` function in both `hooks.server.js` and `hooks.client.js` adhere to the type by returning that shape from the function:

```js
export function handleError({ error }) {
   // ...
   return {
     message: error.message,
     whateverYouWant: 'something else'
   };
}
```

If `handleError` is not defined or the return value is undefined, then SvelteKit falls back to `{ message: 'Internal Error' }`

# PR description

Implements #6499

- [x] `handleError` returns JSON
- [x] `App.PageData` type
- [x]  `handleError` client hook
- [x] docs
- [x] migration docs

### Tests

### Changesets


documentation/docs/05-load.md
```diff
@@ -308,7 +308,7 @@ export function load({ locals }) {
 }
 ```
 
-If an _unexpected_ error is thrown, SvelteKit will invoke [`handleError`](/docs/hooks#handleerror) and treat it as a 500 Internal Server Error.
+If an _unexpected_ error is thrown, SvelteKit will invoke [`handleError`](/docs/hooks#hooks-server-js-handleerror) and treat it as a 500 Internal Server Error.
```
Rich-Harris, line 311:
```suggestion
If an _unexpected_ error is thrown, SvelteKit will invoke [`handleError`](/docs/hooks#hooks-server-js-handleerror) and treat it as a 500 Internal Error.
```

documentation/docs/07-hooks.md
```diff
@@ -2,11 +2,47 @@
...
+An optional `src/hooks.client.js` (or `src/hooks.client.ts`, or `src/hooks.client/index.js`) file exports one functions, optional, that run on the client — `handleError`.
+
+> The location of this file can be [configured](/docs/configuration) as `config.kit.files.hooks.server`
+
+#### handleError
+
+If an error is thrown during loading or rendering, this function will be called with the `error`. This allows for two things:
+
+- you can send data to an error tracking service, for example
+- you can customise the formatting. The return value of the function becomes the value of `$page.error`. To make this type-safe, you can customize the expected shape through `App.PageError`. Note that `message` is always a required property, to guarantee a sensible fallback behavior.
```
Rich-Harris, line 18:
```suggestion
- you can customise the formatting. The return value of the function becomes the value of `$page.error`. To make this type-safe, you can customize the expected shape through `App.PageError`, which must include a `message: string` property.
```

documentation/docs/07-hooks.md
```diff
@@ -2,11 +2,47 @@
...
+// ---cut---
+/** @type {import('@sveltejs/kit').HandleClientError} */
+export function handleError({ error }) {
+	// example integration with https://sentry.io/
+	Sentry.captureException(error);
+	return { message: /** @type {Error} */ (error).message }
+}
+```
+
+If unimplemented or no error shape is returned, SvelteKit will use the default formatting `{ message: string }`.
```
Rich-Harris, line 35:
```suggestion
The default return value, if none is provided or `handleError` is not implemented, is `{ message: 'Internal Error' }`.
```

documentation/docs/07-hooks.md
```diff
@@ -79,47 +115,37 @@ export async function handle({ event, resolve }) {
...
+#### handleError
 
-If an error is thrown during loading or rendering, this function will be called with the `error` and the `event` that caused it. This allows you to send data to an error tracking service, or to customise the formatting before printing the error to the console.
-
-During development, if an error occurs because of a syntax error in your Svelte code, a `frame` property will be appended highlighting the location of the error.
+If an error is thrown during loading or rendering, this function will be called with the `error` and the `event` that caused it. This allows for two things:
 
-If unimplemented, SvelteKit will log the error with default formatting.
+- you can send data to an error tracking service, for example
+- you can customise the formatting. The return value of the function becomes the value of `$page.error`. To make this type-safe, you can customize the expected shape through `App.PageError`. Note that `message` is always a required property, to guarantee a sensible fallback behavior.
```
Rich-Harris, line 123:
```suggestion
- you can customise the formatting. The return value of the function becomes the value of `$page.error`. To make this type-safe, you can customize the expected shape through `App.PageError`, which must include a `message: string` property.
```

documentation/docs/07-hooks.md
```diff
@@ -79,47 +115,37 @@ export async function handle({ event, resolve }) {
...
 	// example integration with https://sentry.io/
 	Sentry.captureException(error, { event });
+	return { message: /** @type {Error} */(error).message }
 }
 ```
 
-> `handleError` is only called for _unexpected_ errors. It is not called for errors created with the [`error`](/docs/modules#sveltejs-kit-error) function imported from `@sveltejs/kit`, as these are _expected_ errors.
-
-### handleFetch
+If unimplemented or no error shape is returned, SvelteKit use the default formatting `{ message: string }`.
```
Rich-Harris, line 140:
```suggestion
The default return value, if none is provided or `handleError` is not implemented, is `{ message: 'Internal Error' }`.
```

documentation/docs/07-hooks.md
```diff
@@ -79,47 +115,37 @@ export async function handle({ event, resolve }) {
...
 
-This function allows you to modify (or replace) a `fetch` request that happens inside a `load` function that runs on the server (or during pre-rendering).
+During development, if an error occurs because of a syntax error in your Svelte code, a `frame` property will be appended highlighting the location of the error.
```
Rich-Harris, line 142: Is this still true? Feels like this stuff should be kept out of the error object

```suggestion
```

packages/kit/src/core/config/options.js
```diff
@@ -141,7 +141,19 @@ const options = object(
 
 			files: object({
 				assets: string('static'),
-				hooks: string(join('src', 'hooks')),
+				hooks: (input, keypath) => {
+					// TOD remove this for the 1.0 release
```
Rich-Harris, line 145:
```suggestion
					// TODO remove this for the 1.0 release
```

packages/kit/src/runtime/client/client.js
```diff
@@ -790,17 +785,29 @@ export function create_client({ target, base, trailing_slash }) {
...
 						};
 					}
 
-					const status = e instanceof HttpError ? e.status : 500;
+					let status = 500;
+					/** @type {App.PageError} */
+					let error;
+
+					if (/** @type {any} */ (err)?.type === 'error') {
+						// this is the server error rethrown above, reconstruct but don't invoke
```
Rich-Harris, line 801: do we know this for certain? what if someone does `throw { type: 'error' }`? edge case, i know

packages/kit/src/runtime/client/client.js
```diff
@@ -1498,6 +1500,15 @@ async function load_data(url, invalid) {
 	return server_data;
 }
 
+/**
+ * @param {unknown} error
+ * @param {import('types').ClientRequestEvent} event
```
Rich-Harris, line 1505: I wonder if the name of this should reflect `LoadEvent` somehow

packages/kit/src/runtime/control.js
```diff
@@ -1,23 +1,21 @@
...
 	 * @param {number} status
-	 * @param {string | undefined} message
+	 * @param {{message: string} extends App.PageError ? (App.PageError | string | undefined) : App.PageError} body
 	 */
-	constructor(status, message) {
+	constructor(status, body) {
 		this.status = status;
-		this.message = message ?? `Error: ${status}`;
+		if (typeof body === 'string') {
+			this.body = { message: body };
```
Rich-Harris, line 9: do we want to allow this, or should we require people to always pass the object?

Rich-Harris (commented):
had a few questions/thoughts/notes-to-self but this mostly looks good

documentation/docs/07-hooks.md
dummdidumm: I liked the previous version better because it also explains why.

documentation/docs/07-hooks.md
dummdidumm: I liked the previous version better because it also explains why.

documentation/docs/07-hooks.md
dummdidumm: Mhm not sure, but I feel like it should (hopefully?) still be true - only that it's passed in, but not necessarily appears in your user-facing error page.

packages/kit/src/runtime/client/client.js
dummdidumm: We don't, no. It could definetly happen that way. To be completely safe we should append something that is super unlikely to be present somewhere else, like `__kit_serverload_error`

packages/kit/src/runtime/client/client.js
dummdidumm: Yeah I felt the same, but didn't come up with a good name

packages/kit/src/runtime/control.js
dummdidumm: I think we should allow this for the common case where people don't really care about these things.

documentation/docs/07-hooks.md
Rich-Harris: Ah right, I mistook input for output. Sleep deprivation

packages/kit/src/runtime/server/index.js
```diff
@@ -309,7 +310,8 @@ export async function respond(request, options, state) {
 			// so we need to make an actual HTTP request
 			return await fetch(request);
 		} catch (e) {
-			const error = coalesce_to_error(e);
+			// HttpError can come from endpoint - TODO should it be handled there instead?
+			const error = e instanceof HttpError ? e : coalesce_to_error(e);
```
Rich-Harris, line 314: do we still need `coalesce_to_error`? i've always been a bit fuzzy on the reason for that function, feels like we can probably get rid of it once and for all in this PR

packages/kit/src/runtime/client/client.js
```diff
@@ -790,17 +785,29 @@ export function create_client({ target, base, trailing_slash }) {
...
+							location: err.location
 						};
 					}
 
-					const status = e instanceof HttpError ? e.status : 500;
+					let status = 500;
+					/** @type {App.PageError} */
+					let error;
+
+					if (/** @type {any} */ (err)?.type === 'error') {
```
Rich-Harris, line 800: i think this should work, actually? bit cheeky but does the job with a minimum of fuss

```suggestion
					if (server_data_nodes.includes(err)) {
```

packages/kit/src/runtime/server/index.js
dummdidumm: Good question honestly haven't looked up where else it's used.

packages/kit/src/runtime/client/client.js
Rich-Harris: i made this change, seems to work

dummdidumm: Ooh yeah this is smart 🤓

packages/kit/src/runtime/server/index.js
Rich-Harris: it's used in a few different places, but all of them are a bit weird. though honestly it might be easier to disentangle them separately

Rich-Harris:
Not a problem for now, but something weird is happening with the docs — perhaps related to the overload? `error` is included twice

image

Rich-Harris (approved):
this is so much cleaner. really happy with how this turned out 🍻

iolyd:
> Defining: hooks:{server: 'src/hooks/index.ts'},
> 
> is throwing error and forcing me to rename the file to: hooks.server.ts
> 
> What is the point of server: setting?

Is there a reason why you would not want to name your files like so?:

```js
const config = {
    // ...
    files: {
        hooks: {
            server: 'src/hooks/index.server',
            client: 'src/hooks/index.client',
        },
    }
}
```

I think the compiler needs the `.server`/ `.client` to know if your module belongs to the server-side or the client-side.

machak:
I only use server hooks and my configuration points to it. Either use configuration or convention, but here, we are mixing both. 
Right now, I am given a choice to configure it and use my own file name, 
but if I do, I get punished for it.

machak:
if compiler depend on the naming, lets drop using "server:" and "client:" and use array only:
  files: {
        hooks: ['src/hooks/index.server','src/hooks/index.client'],
   }

malalecherocks:
If handle() is not implemented in hooks.client.js ... how can I prevent the user navigating between pages? I used to check there if the user had a cookie, if not then redirect to login.

How it's suppouse to do this now? or best practice?

PR #6593, opened 2022-09-05
add new cookies API

### LINKED ISSUE #6540: Add a `cookies` interface to `RequestEvent` and `LoadServerEvent`
### Describe the problem

In #5748, we introduced the `setHeaders` API:

```js
export function load({ setHeaders }) {
  setHeaders({
    'set-cookie': 'answer=42'
  });
}
```

[Recent explorations](https://github.com/sveltejs/kit/discussions/5875#discussioncomment-3540263) (see the 'cookie handling' section in particular) suggest the API falls short. We need a way to set a cookie inside an action that can be read during the subsequent `load`, and while we _could_ make it work with `setHeaders` it would involve some very unexpected behaviour (we would basically need to intercept `request.headers.get('cookie')` and sub in a value that included the `set-cookie`, which is a bad amount of magic).

We add a new `cookies` interface:

```js
// src/routes/login/+page.server.js
export const actions = {
  default: ({ fields, cookies }) => {
    const user = await db.getUserFromEmail(fields.get('email');

    // ...some logic happens...

    cookies.set('sessionid', await db.createSession(user), {
        maxAge: 30 * 24 * 60 * 60 * 1000 // stay logged in for 30 days
    });
  }
};
```

> `export const actions` is described in https://github.com/sveltejs/kit/discussions/5875#discussioncomment-3540263

```js
// src/routes/+layout.server.js
export async function load({ cookies }) {
  // this value could come from `request.headers.get('cookie')` or the previous `cookies.set` call
  const sessionid = cookies.get('sessionid');
  const user = await db.getUserFromSessionId(sessionid);
  return { user };
}
```

There is an experimental standard [`CookieStore`](https://developer.mozilla.org/en-US/docs/Web/API/CookieStore) interface, but I don't think we should use it. It's designed for use in browsers, which means all the methods are asynchronous — needlessly, in our case — and there's a `change` event that doesn't really make sense in our context. More important, it doesn't expose an `httpOnly` option (because that wouldn't make sense for client-side JavaScript), whereas I think `httpOnly` is the only sensible default.

The proposed API:

```ts
interface Cookies {
  delete(name: string): void;

  get(name: string, opts: {
    decode: (input: string) => string
  }): string;

  set(name: string, value: string, opts?: {
    expires?: Date;
    maxAge?: number;
    domain?: string;
    path?: string;
    secure?: boolean;
    httpOnly: boolean;
    sameSite: 'strict' | 'lax' | 'none'; // can only be `none` with `secure: true`,
    encode: (input: string) => string
  }); void;
}
```

If you're familiar with the [cookie](https://www.npmjs.com/package/cookie) package, this API will seem very familiar — it's what we'd use under the hood.

### Alternatives considered

Relying on people populating `event.locals` in actions, which feels unnecessarily restrictive.

Intercepting (and changing) `request.headers.get('cookie')`, which feels like too much magic.

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
closes #6540

* [x] docs
* [x] update tests to read cookies with this API instead of from request headers

### Tests

### Changesets


packages/kit/src/runtime/server/cookie.js
```diff
@@ -0,0 +1,76 @@
...
+				if (
+					cookie.name === name &&
+					domain_matches(url.hostname, cookie.options.domain) &&
+					path_matches(url.pathname, cookie.options.path)
+				) {
+					return cookie.value;
+				}
+			}
+
+			return decode(initial_cookies[name]);
```
cdcarson, line 31: I think you want to check for the cookie's existence first, since `decodeURIComponent(undefined)` gives the string `'undefined'` not `undefined`.

```ts
return typeof initial_cookies[name] === 'string' ? decode(initial_cookies[name]) : undefined;
```

Rich-Harris: ah, good catch. very slightly leaner:

```suggestion
			return name in initial_cookies ? decode(initial_cookies[name]) : undefined;
```

packages/kit/src/runtime/server/index.js
```diff
@@ -144,15 +146,9 @@ export async function respond(request, options, state) {
...
-					for (const cookie of new_cookies) {
-						if (cookies.includes(cookie)) {
-							throw new Error(`"${key}" header already has cookie with same value`);
-						}
-
-						cookies.push(cookie);
-					}
+					throw new Error(
+						`Use \`event.cookie.set(name, value, options)\` instead of \`event.setHeaders\` to set cookies`
+					);
```
dummdidumm, line 151: Instead of throwing here, couldn't we invoke something like `cookies.set(cookie.parse(value))` here?

packages/kit/src/runtime/server/index.js
```diff
@@ -144,15 +146,9 @@ export async function respond(request, options, state) {
 				const value = new_headers[key];
 
 				if (lower === 'set-cookie') {
-					const new_cookies = /** @type {string[]} */ (Array.isArray(value) ? value : [value]);
-
-					for (const cookie of new_cookies) {
-						if (cookies.includes(cookie)) {
-							throw new Error(`"${key}" header already has cookie with same value`);
```
dummdidumm, line 151: I believe this check is no longer done in the new version - any drawbacks omitting it and we should add it back, or is it fine like this?

documentation/docs/05-load.md
```diff
@@ -240,25 +241,7 @@ export async function load({ fetch, setHeaders }) {
...
-	setHeaders({
-		'set-cookie': 'b=2; HttpOnly'
-	});
-
-	setHeaders({
-		'set-cookie': ['c=3; HttpOnly', 'd=4; HttpOnly']
-	});
-}
-```
+You cannot add a `set-cookie` header with `setHeaders` — use the [`cookies`](/docs/types#sveltejs-kit-cookies) API in a server-only `load` function instead.
```
dummdidumm, line 244: This makes me wonder if we should have a section "inputs exclusive to isomorphic load" and "inputs exlusive to server-only load" in here (can be done in a separate PR)

dummdidumm (approved):
Looking good, but still wondering if we should allow cookies in `setHeaders`. If you set some simple cookies, you don't need to reach for that, and it may have this "oh why can't I use the platform" feeling to it. More comments inline, but nothing blocking this from merging.

packages/kit/src/runtime/server/index.js
Rich-Harris: We _could_, but no-one wants to write this...

```js
import * as cookie from 'cookie';

export const actions = {
  default: async ({ setHeaders }) => {
    // ...
    setHeaders({
      'set-cookie': cookie.serialize('answer', 42, {
        httpOnly: true
      })
    });
   }
};
```

...when they could write this:

```js
export const actions = {
  default: async ({ cookies }) => {
    // ...
    cookies.set('answer', 42);
   }
};
```

I for one would be grateful for an error message that said 'you should delete a bunch of code, and you can probably remove the `cookie` dependency'.

It also feels very weird if you use `setHeaders` followed by `cookies.get` — it doesn't feel like it should work, and yet it does. Always better to have one way to do things.

documentation/docs/05-load.md
Rich-Harris: Probably yeah, since we haven't actually introduced `RequestEvent` by this stage

packages/kit/src/runtime/server/index.js
```diff
@@ -144,15 +146,9 @@ export async function respond(request, options, state) {
...
-						}
-
-						cookies.push(cookie);
-					}
+					throw new Error(
+						`Use \`event.cookie.set(name, value, options)\` instead of \`event.setHeaders\` to set cookies`
```
kevinrenskers, line 150: Should't this be `event.cookies.set`?

Rich-Harris: yep. fixed

martinbeentjes:
@Rich-Harris going to ask the question here because I think here is a valid place. I just fixed an issue I had because I didn't set the `path` option on a cookie for the whole app (session management). What is the reasoning to now explicitly require an user to set the `path` on a cookie? Is that security related? 

Should that also be read in the documentation? If so, I can certianly help out and make a small PR to fix that.

PR #6602, opened 2022-09-06
remove cookie from demo app

Rich-Harris:
demo app is due for a more comprehensive overhaul, but we can at least simplify stuff using the new `cookies` API in the meantime


martinbeentjes:
all those small improvements, sveltekit becomes more and more awesome to use.

dummdidumm (approved):

Issue #6563, opened 2022-09-04
Unavoidable warning when using kit.paths.base when running vite build.


talwat:
### Describe the bug

Vite normally recommends the base path to end with a slash, as denoted by this warning when running `npm run build`: `(!) "base" option should end with a slash.`

However, say we remove this, sveltekit will throw its own error: `Error: config.kit.paths.base option must either be the empty string or a root-relative path that starts but doesn't end with '/'. See https://kit.svelte.dev/docs/configuration#paths`

This means that you are unable to get rid of the vite warning.

### Reproduction

First, you need to use `adapter-static` and add `export const prerender = true` to the root layout.

Then, add the `config.kit.paths.base` option without the `/` at the end.

I have recreated the environment in [this](https://github.com/talwat/kit-base-path-error) repo.

Finally, if you try to build it, I ran `npm run build`, you will receive the warning.

### Logs

```txt
Log with / at the end:

error during build:
Error: config.kit.paths.base option must either be the empty string or a root-relative path that starts but doesn't end with '/'. See https://kit.svelte.dev/docs/configuration#paths
    at file:///Users/tal/Temporary%20Code/kit-base-path-error/node_modules/@sveltejs/kit/src/core/config/options.js:200:13
    at file:///Users/tal/Temporary%20Code/kit-base-path-error/node_modules/@sveltejs/kit/src/core/config/options.js:389:43
    at file:///Users/tal/Temporary%20Code/kit-base-path-error/node_modules/@sveltejs/kit/src/core/config/options.js:375:18
    at file:///Users/tal/Temporary%20Code/kit-base-path-error/node_modules/@sveltejs/kit/src/core/config/options.js:375:18
    at file:///Users/tal/Temporary%20Code/kit-base-path-error/node_modules/@sveltejs/kit/src/core/config/options.js:375:18
    at validate_config (file:///Users/tal/Temporary%20Code/kit-base-path-error/node_modules/@sveltejs/kit/src/core/config/index.js:104:9)
    at process_config (file:///Users/tal/Temporary%20Code/kit-base-path-error/node_modules/@sveltejs/kit/src/core/config/index.js:72:20)
    at load_config (file:///Users/tal/Temporary%20Code/kit-base-path-error/node_modules/@sveltejs/kit/src/core/config/index.js:64:9)
    at async config (file:///Users/tal/Temporary%20Code/kit-base-path-error/node_modules/@sveltejs/kit/src/exports/vite/index.js:200:20)
[...]

PR #6621, opened 2022-09-06
[fix] address Vite warning when using base or asset path

benmccann:
Fixes https://github.com/sveltejs/kit/issues/6563

SvelteKit expects paths don't end with `/` and Vite expects that they do


Rich-Harris (approved):

talwat:
Thank you!

PR #6623, opened 2022-09-06
feat: Extend user-defined server-only modules functionality to `$lib/server`

tcc-sejohnson:
Extend the "throw an error if you import this on the client" functionality to `$lib/server` so that you don't necessarily have to name every module you want to be server-only `something.server.ts`. 

I also refactored the recursive graph traversal to try to encapsulate concerns a bit more. I was getting frustrated with having to make the same change to the same logic in different places, so I factored the "illegal-checking" away from the "traverse the vite/rollup graph" part. It's technically more lines of code, but most of the addition is in the form of JSDoc comments. I added a bunch of them to make the return types of the functions explicit -- we previously had a few very-subtle bugs where functions were returning `undefined` instead of `false` or something like that. Nothing causing any issues yet, but better to be explicit when recursion is involved.

### Tests

### Changesets


Rich-Harris (approved):

Rich-Harris:
note-to-self that we need to document this at some point, though it can wait till after sweden

tcc-sejohnson:
> note-to-self that we need to document this at some point, though it can wait till after sweden

See big documentation comment I wrote on the prior PR. I wasn't sure where to stick it. I'm going to create an issue so that we don't forget.

accuser:
@tcc-sejohnson this is an amazingly responsive PR - thank you! 👏🏼 

I think this will have a positive impact on the Svelte Kit DX.

PR #6633, opened 2022-09-07
change use:enhance signature to support `<button formaction>`

### LINKED ISSUE #6630: `use:enhance` should work with `<button formaction>`
### Describe the problem

This should work...

```html
<form method="POST">
  <button formaction="?/hello" use:enhance>click me</button>
</form>
```

...but it doesn't, because `enhance` assumes its first argument is a `<form>`. The same goes for `<input>` elements with `type="submit"` and `type="image"`.

image

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
closes #6630. This makes two breaking changes to the `enhance` action:

1. the input `form` is now `element`, and refers to the element the action was added to. To type that, the function and the SubmissionFunction interface now both take three generic params instead of two.
2. The returned callback is a `({ element, form, data, result }) => {...}` function instead of just `(result) => {...}`. This makes it easier to do things like `use:enhance={() => whatever}`, where `whatever` is a function that (for example) calls `form.reset()`.

### Tests

### Changesets

### DIFF
--- a/packages/kit/types/ambient.d.ts
+++ b/packages/kit/types/ambient.d.ts
@@ -116,7 +116,7 @@ declare module '$app/forms' {
 		Success extends Record<string, unknown> | undefined = Record<string, any>,
 		Invalid extends Record<string, unknown> | undefined = Record<string, any>
 	>(
-		form: Element,
+		element: Element,
 		/**
 		 * Called upon submission with the given FormData.
 		 * If `cancel` is called, the form will not be submitted.
--- a/packages/kit/types/ambient.d.ts
+++ b/packages/kit/types/ambient.d.ts
@@ -89,17 +89,18 @@ declare module '$app/forms' {
 	import type { ActionResult } from '@sveltejs/kit';
 
 	export type SubmitFunction<
-		Success extends Record<string, unknown> | undefined,
-		Invalid extends Record<string, unknown> | undefined
+		Element extends HTMLFormElement | HTMLInputElement | HTMLButtonElement = HTMLFormElement,
+		Success extends Record<string, unknown> | undefined = Record<string, any>,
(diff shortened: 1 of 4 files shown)


dummdidumm:
Pushed up a change that should make this work (infering `reset` exists etc). Though I'm also not sure if this is a good move in general (the form thing, not the return callback params change). It feels like it makes the common case slightly harder or confusing than it needs to be.

david-plugge:
Is there any need to access the element if it´s a button?
I´d suggest using `element.closest('form')` to get the parent form if the action is not applied on the form directly.

dummdidumm:
This is a good point, we could make that part non-breaking by passing in both `element` and `form` to the callback function - that way only the return callback function change is breaking.

Rich-Harris:
I just sat down at my laptop to suggest doing exactly that — we are on the same wavelength

packages/kit/types/ambient.d.ts
```diff
@@ -89,24 +89,34 @@ declare module '$app/forms' {
...
 	 * @param form The form element
 	 * @param options Callbacks for different states of the form lifecycle
 	 */
 	export function enhance<
+		Element extends HTMLFormElement | HTMLInputElement | HTMLButtonElement = HTMLFormElement,
 		Success extends Record<string, unknown> | undefined = Record<string, any>,
 		Invalid extends Record<string, unknown> | undefined = Record<string, any>
 	>(
-		form: HTMLFormElement,
+		form: Element,
```
Rich-Harris, line 119: doesn't really matter but for the sake of persnickitiness

```suggestion
		element: Element,
```

dummdidumm (approved):

Issue #6204, opened 2022-08-23
dynamic env is an empty object in production when using `config.kit.env.dir`


ambiguous48:
### Describe the bug

`config.kit.env.dir`, introduced in https://github.com/sveltejs/kit/pull/6175, only works in dev.

In dev, correct env is logged to console. In prod, empty object is logged to console. Minimal reproduction below.

### Reproduction

https://github.com/ambiguous48/dynamic-env-missing

### System Info

```shell
System:
    OS: Linux 5.18 Pop!_OS 22.04 LTS
    CPU: (24) x64 12th Gen Intel(R) Core(TM) i9-12900K
    Memory: 16.68 GB / 31.13 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 18.3.0 - ~/.local/share/pnpm/node
    Yarn: 1.22.19 - ~/.local/share/pnpm/yarn
    npm: 8.11.0 - ~/.local/share/pnpm/npm
  Browsers:
    Chrome: 104.0.5112.79
    Firefox: 103.0
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.65 
    @sveltejs/kit: next => 1.0.0-next.431 
    svelte: ^3.44.0 => 3.49.0 
    vite: ^3.0.4 => 3.0.9
```

### Severity

annoyance

tcc-sejohnson:
Based on my testing, this should be fixed with #6179. The problem isn't even the new `config.kit.env.dir` option, it's that (1) we weren't initializing the variables correctly in `prerendering` and (2) something something, complicated initialization sequences. Either way, the repro works fine when I point it at that branch.

benmccann:
Closing this then since that PR has been merged. Note that it hasn't been released yet

ambiguous48:
> Note that it hasn't been released yet

@benmccann [This](https://github.com/sveltejs/kit/releases/tag/%40sveltejs%2Fkit%401.0.0-next.433) is the release that you're referring to, yes? If the answer is no feel free to disregard this.

If it is, then the issue seems to still exist even after updating to the latest sveltekit ([reproduction here](https://github.com/ambiguous48/dynamic-env-missing)).

benmccann:
that reproduction looks to be on the latest version and any changes from a week ago have certainly been released by now, so I'd expect it to work

ambiguous48:
[...]

PR #6635, opened 2022-09-07
[docs] Highlight different `$env/dynamic` behavior between `dev` and `prod`

ambiguous48:
It wasn't immediately clear to me, https://github.com/sveltejs/kit/issues/6204, that `$env/dynamic` might be loaded differently depending on `dev`/`prod`. Because `$env/dynamic` included `.env` during `dev`, I erroneously assumed it would do the same during `prod` (a la dotenv) when using `adapter-node`.

The second sentence could be clearer but I don't know enough about the topic to make it so.

### Tests

### Changesets


Rich-Harris (approved):

Rich-Harris:
thanks!

PR #6309, opened 2022-08-26
[fix] wrong reuse of server load data

dummdidumm:
Don't reuse server load data from previous page if current doesn't have a load function

Fixes #6300

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [x] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #6641, opened 2022-09-07
Reuse server data if the client load is invalidated

### LINKED ISSUE #6459: Data from +layout.server.ts is null after navigating
> I mean the `load({data})` in `layout.ts`, that should hold the data from `layout.server.ts`.

This still seems to be an issue, even if there is no need to rerun the server load, I'd still expect the data object to stay the same instead of becoming `null` after navigating

// +layout.server.ts
```ts
export const load: LayoutServerLoad = async ({ locals, url }) => {
	if (!locals.token) {
		throw redirect(302, `/login?${new URLSearchParams({ callBackUrl: url?.pathname ?? '/' }).toString()}`)
	}

	const jwt = decode<JWTPayload>(locals.token)
	const session = {
		user: {
			id: jwt['sub'],
			email: jwt['email']
		},
	}
	return {
		token: locals.token,
		settings: locals.appSettings,
		session: session
	}
}
```

//+layout.ts
```ts
export const load: LayoutLoad = async ({ data, url }) => {
	await loadTranslations(getDefaultLocale(), url.pathname) // keep this just before the `return`
	return {
		token: data.token,
		settings: data.settings,
		session: data.session
	}
}
```
Will error out after page change
Screenshot 2022-08-30 at 07 20 56

EDIT: Running on next.454

_Originally posted by @luhagel in https://github.com/sveltejs/kit/issues/6349#issuecomment-1231155148_

mquandalle:
Fixes #6459

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -656,10 +656,9 @@ export function create_client({ target, base, trailing_slash }) {
 
 	/**
 	 * @param {import('types').ServerDataNode | import('types').ServerDataSkippedNode | null} node
-	 * @param {import('./types').DataNode | null} [previous]
 	 * @returns {import('./types').DataNode | null}
 	 */
-	function create_data_node(node, previous) {
+	function create_data_node(node) {
 		if (node?.type === 'data') {
 			return {
 				type: 'data',
@@ -671,8 +670,6 @@ export function create_client({ target, base, trailing_slash }) {
 					url: !!node.uses.url
 				}
 			};
-		} else if (node?.type === 'skip') {
-			return previous ?? null;
 		}
 		return null;
 	}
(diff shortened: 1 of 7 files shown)


mquandalle:
I think this issue was caused by #6309. So far I couldn't come up with an implementation that would work with both test cases, but @dummdidumm you probably have a better understanding than me of the logic here.

packages/kit/src/runtime/client/client.js
```diff
@@ -759,6 +756,8 @@ export function create_client({ target, base, trailing_slash }) {
 				throw server_data_node;
 			}
 
+			const prev_server_data_node = (!changed?.url && previous?.server) || null;
```
mquandalle, line 759: I realize that `changed.url` is not the perfect condition, because we will want to reuse the server data of a layout load even if the url changes, but I'm not familiar enough with the code base to determine the correct condition.

PR #6642, opened 2022-09-07
enable suggestions for generated types `./$types`

s3812497:
Adds a path to the generated `tsconfig.json` file to resolve an import to `.svelte-kit/types/src/routes/$types` as `./$types`.

Adding this path fixes a slight issue I've had with generated types where Quick Fix does not include `'./$types'` as a suggestion.

Before:
CleanShot 2022-09-07 at 21 00 43@2x

After:
CleanShot 2022-09-07 at 21 03 29@2x

Sorry if this is unnecessary or implemented incorrectly. This is my first time submitting a PR.

Also, I tried to find a more variable that would lead to `.svelte-kit/types/src/routes/` but the best I could find was `config.outDir` for `.svelte-kit`. Would appreciate if anyone knew a better path variable to use.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/sync/write_tsconfig.js
+++ b/packages/kit/src/core/sync/write_tsconfig.js
@@ -67,7 +67,11 @@ export function write_tsconfig(config, cwd = process.cwd()) {
 				compilerOptions: {
 					// generated options
 					baseUrl: config_relative('.'),
-					paths: get_tsconfig_paths(config),
+					paths: {
+						// enables aliased import for generated types
+						"./$types": [".svelte-kit/types/src/routes/*"],
+						...get_tsconfig_paths(config)
+					},
 					rootDirs: [config_relative('.'), './types'],
 
 					// essential options
--- a/packages/kit/src/core/sync/write_tsconfig.js
+++ b/packages/kit/src/core/sync/write_tsconfig.js
@@ -69,7 +69,7 @@ export function write_tsconfig(config, cwd = process.cwd()) {
 					baseUrl: config_relative('.'),
 					paths: {
 						// enables import autocomplete for generated types aliased path
-						'./$types': [`${config.outDir}/types/src/routes/*`],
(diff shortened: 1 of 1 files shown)


packages/kit/src/core/sync/write_tsconfig.js
```diff
@@ -67,7 +67,11 @@ export function write_tsconfig(config, cwd = process.cwd()) {
 				compilerOptions: {
 					// generated options
 					baseUrl: config_relative('.'),
-					paths: get_tsconfig_paths(config),
+					paths: {
+						// enables import autocomplete for generated types aliased path
+						'./$types': [`${config.outDir}/types/src/routes/$types`],
```
dummdidumm, line 72: Not sure if this is correct, I'm not sure what implications this has on typescript resolving these paths, since `./$types` resolved to a different d.ts file for each page. It will probably help with intellisense though (as you have shown). Need to think on this more when I have time and try it out.

s3812497: Appreciate you looking into this. I'm still new to typescript so I don't know the implications either.

Rich-Harris:
What do I have to do to get this to work? It's not showing up as a Quick Fix option for me:

image

s3812497:
> What do I have to do to get this to work? It's not showing up as a Quick Fix option for me:
> 
> image

The original paths that I used when testing this were:
```
"./types": [".svelte-kit/types/src/routes/$types", ".svelte-kit/types/src/routes/*/$types"],
```

However, I've discovered that the generated types full path import does not show as a Quick Fix option in any files unless there's at least one import statement in a file referencing $types.

Is it possible to get TypeScript to recognise the generated types without having to this initial manual import?

dummdidumm:
Closing this, this was fixed inside TypeScript itself. I'm not sure which version this lands in, but either next month or in 4 months it'll work as expected.

PR #6643, opened 2022-09-07
Respect `autofocus` after a client-side navigation or enhanced form submit

### LINKED ISSUE #6629: Navigation and `use:enhance` should apply `autofocus`
### Describe the problem

This is arguably a bug, insofar as client-side navigation and `enhance` are supposed to simulate a full page reload as far as possible, but I'm on my phone so no repro will be forthcoming.

In a navigation, we (IIRC) focus the body. In an enhanced form submit, the submit button stays focused.

In both cases, we should focus the first element with `autofocus` if such there be. If such there not be, `enhance` should focus the body by default.

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
Closes #6629. This makes client-side navigation/submission behave more like their native counterparts

### Tests

### Changesets


packages/kit/src/runtime/client/client.js
```diff
@@ -328,9 +304,8 @@ export function create_client({ target, base, trailing_slash }) {
 					scrollTo(0, 0);
 				}
 			}
-		} else {
-			// in this case we're simply invalidating
-			await tick();
+
+			if (!keepfocus) reset_focus();
```
dummdidumm, line 308: I don't fully grasp the implications of this, so I'll ask: what difference does it make to switch the order of execution?

packages/kit/src/runtime/client/client.js
```diff
@@ -1207,6 +1184,8 @@ export function create_client({ target, base, trailing_slash }) {
 				const post_update = pre_update();
 				root.$set(props);
 				post_update();
+
+				tick().then(reset_focus);
```
dummdidumm, line 1188: I think we should only do this when the type is success. When input is invalid, it might be nice to keep focus on the currently selected thing instead.

Rich-Harris: The currently selected thing will be the button that was used to submit the form — by keeping the focus there the only thing we're really doing is making it more likely that the user will resubmit bad data, I think

packages/kit/src/runtime/client/client.js
Rich-Harris: Focus management needs to happen after the DOM has been updated — previously, it would reset while DOM updates were still pending. We got away with it because we were focusing the `<body>` which is unchanged, but that won't do any more

Rich-Harris: Actually I'm wrong — we _can_ focus immediately, it turns out. And we need to, because if we wait for `tick()` to resolve it will blow away any manual focus management. Gonna blame sleep deprivation again

packages/kit/src/runtime/client/client.js
dummdidumm: Just tried this out - if you press enter on an input (which isn't that uncommon I'd say) the focus stays on that input after submitting it. I would be in favor of keeping the focus on that input on an invalid response.

Rich-Harris:
gah, i can't reproduce the webkit test failures locally. might have to rethink some stuff

Rich-Harris:
Couldn't figure out the webkit stuff so I, err... skipped the test on webkit

Rich-Harris:
The tests pass in webkit locally (and it works as expected when testing manually in Safari) so I have no idea what's going on with the CI failures, but frankly I am well past the point of caring. Reverted 3d5a235

dummdidumm (approved):

Issue #6609, opened 2022-09-06
`cookies.delete` not working when using form `enhance`


s3812497:
### Describe the bug

Using `enhance` on a `<form>` and throwing `redirect` in the form action causes `cookies.delete` to not work in the page server load function.

### Reproduction

https://github.com/s3812497/sveltekit-enhance-bug

Reproduction:
1. Click the enhanced form button.
2. User is redirected.
3. Cookie is not deleted.

Expected:
1. Click the enhanced form button.
2. User is redirected.
3. Cookie is deleted in `+page.server`

### System Info

```shell
System:
    OS: macOS 12.5.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 1.56 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.0 - ~/.nvm/versions/node/v16.17.0/bin/node
    npm: 8.15.0 - ~/.nvm/versions/node/v16.17.0/bin/npm
  Browsers:
    Chrome: 104.0.5112.79
    Firefox: 104.0.1
    Firefox Developer Edition: 105.0
    Safari: 15.6.1
    Safari Technology Preview: 16.0
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.71 
    @sveltejs/kit: next => 1.0.0-next.472 
    svelte: ^3.44.0 => 3.50.0 
    vite: ^3.1.0 => 3.1.0
```

### Severity

serious, but I can work around it

dummdidumm:
I'm not sure if there's something we can do here besides documenting it. The `enhance` function can't know that a cookie was deleted, so this needs to be done manually on the client through the browser APIs.

s3812497:
> I'm not sure if there's something we can do here besides documenting it. The `enhance` function can't know that a cookie was deleted, so this needs to be done manually on the client through the browser APIs.

For additional context, in my app before the migration:

1. A user could login and a cookie would be set by the `POST` action (the form uses the previous `enhance`).
2. The user gets redirected to the dashboard, which has a logout button (a simple `<a href="/auth?logout=true">`)
3. Navigating to `/auth?logout=true` would delete the cookie by checking the `params` for `logout` in `+page.server.ts`

However, the new `cookies.delete` api does not seem to delete the cookie compared to how it was previously done with `setHeaders` and an expired cookie.

This makes me wonder if the new `enhance` is preventing an expired cookie from being set.

[...]

PR #6648, opened 2022-09-07
fix cookie issues, add unit and e2e tests for cookies api

cdcarson:
There are some issues with cookie deletion unresolved in this issue: https://github.com/sveltejs/kit/issues/6609

I believe the issue was that we were not setting the path on delete (and maybe on set.) Some tests are needed to confirm. 

~~Marking this as draft until then.~~

- [x] unit tests 
- [x] e2e tests (probably good to have more eventually, fix and unskip the flaky one)
- [x] expose `cookie` serialize options from the SvelteKit cookies API ~~(this should be a separate typescript issue)~~ this is fixed by directly annotating the `get`, `set` and `delete` with JSDoc params rather than relying on the imported type of `cookies`
- [x] document api changes, if any (e.g. setting `path` to other than '/') See [this comment](https://github.com/sveltejs/kit/pull/6648#issuecomment-1241215406) for suggested copy.

### Tests

### Changesets


cdcarson:
This is as far as I can take it, for now. I'd liked to have written more Playwright tests, but wasn't able to figure out fancier stuff, like testing enhanced forms. So the unresolved issue in #6609 I was only able to check by running the repro against this pr. But the perfect is the enemy of the good, and I'm confident that this is on the right track.

**Net**: The main issue folks were having stemmed from Kit not setting `path` and `sameSite` by default, for both `set` and `delete`. See also #6650. The defaults in this pr are exactly what I had been using to set/delete before the API, with no problems, so I'm pretty confident they are the right ones.

packages/kit/src/runtime/server/cookie.js
```diff
@@ -1,48 +1,66 @@
...
+				return c.value;
 			}
 
 			return name in initial_cookies ? decode(initial_cookies[name]) : undefined;
 		},
-		set(name, value, options = {}) {
-			new_cookies.push({
+
+		/**
+		 *
```
benmccann, line 38:
```suggestion
```

packages/kit/test/apps/basics/test/client.test.js
```diff
@@ -785,3 +785,37 @@ test.describe('data-sveltekit attributes', () => {
...
+		const span = page.locator('#cookie-value');
+		expect(await span.innerText()).toContain('undefined');
+	});
+
+	test('set a cookie', async ({ page }) => {
+		await page.goto('/cookies/set');
+		const span = page.locator('#cookie-value');
+		expect(await span.innerText()).toContain('teapot');
+	});
+	test('delete a cookie', async ({ page }) => {
```
benmccann, line 801:
```suggestion

	test('delete a cookie', async ({ page }) => {
```

packages/kit/test/apps/basics/test/server.test.js
```diff
@@ -345,3 +346,37 @@ test.describe('Miscellaneous', () => {
...
+		const span = page.locator('#cookie-value');
+		expect(await span.innerText()).toContain('undefined');
+	});
+
+	test('set a cookie', async ({ page }) => {
+		await page.goto('/cookies/set');
+		const span = page.locator('#cookie-value');
+		expect(await span.innerText()).toContain('teapot');
+	});
+	test('delete a cookie', async ({ page }) => {
```
benmccann, line 362:
```suggestion

	test('delete a cookie', async ({ page }) => {
```

packages/kit/src/runtime/server/cookie.spec.js
```diff
@@ -39,17 +39,68 @@ paths.negative.forEach(([path, constraint]) => {
...
+test('default values when set is called', () => {
+	const { cookies, new_cookies } = cookiesSetup();
+	cookies.set('a', 'b');
+	const entry = new_cookies.get('a')?.options;
+	assert.equal(entry?.secure, true);
+	assert.equal(entry?.httpOnly, true);
+	assert.equal(entry?.path, '/');
+	assert.equal(entry?.sameSite, 'lax');
+});
+test('overridden defaults when set is called', () => {
```
benmccann, line 69:
```suggestion

test('overridden defaults when set is called', () => {
```

packages/kit/src/runtime/server/cookie.spec.js
```diff
@@ -39,17 +39,68 @@ paths.negative.forEach(([path, constraint]) => {
...
+	const { cookies, new_cookies } = cookiesSetup();
+	cookies.delete('a');
+	const entry = new_cookies.get('a')?.options;
+	assert.equal(entry?.secure, true);
+	assert.equal(entry?.httpOnly, true);
+	assert.equal(entry?.path, '/');
+	assert.equal(entry?.sameSite, 'lax');
+	assert.equal(entry?.maxAge, 0);
+});
+test('overridden defaults when delete is called', () => {
```
benmccann, line 89:
```suggestion

test('overridden defaults when delete is called', () => {
```

packages/kit/src/runtime/server/cookie.spec.js
```diff
@@ -39,17 +39,68 @@ paths.negative.forEach(([path, constraint]) => {
...
+	const { cookies, new_cookies } = cookiesSetup();
+	cookies.delete('a', { secure: false, httpOnly: false, sameSite: 'strict', path: '/a/b/c' });
+	const entry = new_cookies.get('a')?.options;
+	assert.equal(entry?.secure, false);
+	assert.equal(entry?.httpOnly, false);
+	assert.equal(entry?.path, '/a/b/c');
+	assert.equal(entry?.sameSite, 'strict');
+	assert.equal(entry?.maxAge, 0);
+});
+test('cannot override maxAge on delete', () => {
```
benmccann, line 99:
```suggestion

test('cannot override maxAge on delete', () => {
```

benmccann:
It looks to me like you added the same tests to `client.test.js` and `server.test.js`. If you want to run a test on both the client and server you can add it to `test.js` in the same directory

benmccann:
I just glanced quickly, but the Playwright tests look to me like they could cause flakiness. They run in parallel and are checking the same cookie values

Rich-Harris:
Thanks @cdcarson — re the parallel-tests-causing-flakiness thing @benmccann referred to, you can set a group of tests to run serially like so:

https://github.com/sveltejs/kit/blob/e20425fefcfe570dc678d42f61d7e13483672fc0/packages/kit/test/apps/basics/test/client.test.js#L648

cdcarson:
I moved the needlessly duplicate tests out of server and client into test.js.

Side note:

> [The tests] run in parallel and are checking the same cookie values

I'm a playwright noob. Does this mean each test does not have its own browser?

packages/kit/src/runtime/server/cookie.js
```diff
@@ -1,48 +1,65 @@
...
+			const c = new_cookies.get(name);
+			if (
+				c &&
+				domain_matches(url.hostname, c.options.domain) &&
+				path_matches(url.pathname, c.options.path)
+			) {
+				return c.value;
 			}
 
 			return name in initial_cookies ? decode(initial_cookies[name]) : undefined;
```
cdcarson, line 34: There's a gotcha here. `initial_cookies` have already been decoded by `cookie.parse`, using the default `decodeURIComponent`.  Decoding twice is likely bad. Two options:

1. Parse the cookies each time inside `get`, using the passed `decode` option if present.
2. Or, we just say that you cannot provide your own `decode` and `encode` options.  That'd be my pref, though there's gonna be somebody who won't like it. 

Not sure what to do. I'll do "parse inside get" for now unless @benmccann @Rich-Harris have any thoughts.

cdcarson:
Limit playwright tests snippet:

```bash
npx playwright test -g "Cookies API"
```

packages/kit/test/apps/basics/test/test.js
```diff
@@ -1798,3 +1798,37 @@ test.describe('Actions', () => {
...
+		const span = page.locator('#cookie-value');
+		expect(await span.innerText()).toContain('undefined');
+	});
+
+	test('set a cookie', async ({ page }) => {
+		await page.goto('/cookies/set');
+		const span = page.locator('#cookie-value');
+		expect(await span.innerText()).toContain('teapot');
+	});
+	test('delete a cookie', async ({ page }) => {
```
benmccann, line 1814:
```suggestion

	test('delete a cookie', async ({ page }) => {
```

packages/kit/test/apps/basics/test/server.test.js
```diff
@@ -2,6 +2,7 @@ import { expect } from '@playwright/test';
 import { test } from '../../../utils.js';
 import { fetch } from 'undici';
 import { createHash, randomBytes } from 'node:crypto';
+import { parse } from 'cookie';
```
benmccann, line 5:
```suggestion
```

benmccann:
Ignore my previous comment about the tests running in parallel. It seems each test does get its own browser context: https://playwright.dev/docs/browser-contexts

packages/kit/test/apps/basics/test/test.js
cdcarson: Ayayay, am I not formatting? I do pnpm format, I think.

Or is this something else?

cdcarson: Oh, I see, you want a line between each test, probably fixed by prettier. I'll try to do that.

packages/kit/src/runtime/server/cookie.js
cdcarson: PR updated with "parse inside get".

Here's the test that would fail otherwise: https://github.com/sveltejs/kit/blob/a73cad0e36c944774c3c603fb1f5895d18c20962/packages/kit/test/apps/basics/test/test.js#L1849

packages/kit/test/apps/basics/src/routes/cookies/encoding/+page.svelte
```diff
@@ -0,0 +1,12 @@
+<script>
+
+	/** @type {import('./$types').PageData} */
+	export let data;
+
+
```
benmccann, line 6:
```suggestion
```

packages/kit/test/apps/basics/test/server.test.js
benmccann: looks like this line still needs to be removed

packages/kit/test/apps/basics/src/routes/cookies/encoding/+page.svelte
```diff
@@ -0,0 +1,12 @@
...
+</script>
+
+Cookie: <span id="cookie-value">
+  {data.a}
+</span>
```
benmccann, line 11: it's a little funny to me the start and end tags don't line up. i'd either put the start tag on a new line or just put everything on a single line
```suggestion
Cookie: <span id="cookie-value">{data.a}</span>
```

packages/kit/test/apps/basics/src/routes/cookies/set-in-layout/+page.svelte
```diff
@@ -0,0 +1,8 @@
+<script>
+	/** @type {import('./$types').PageData} */
+	export let data;
+</script>
+
+Cookie: <span id="cookie-value">
+	{data.a}
+</span>
```
benmccann, line 8:
```suggestion
Cookie: <span id="cookie-value">{data.a}</span>
```

cdcarson:
I'm happy that this now fixes the issues we know about, knock on wood. 

Remaining todos, not necessarily blocking.

- Decide whether to keep `encode` and `decode` as options, or just go with the `cookie` library defaults (`de/encodeURIComponent`.) Currently we are keeping them. The API surface is only minimally larger, since they are optional and most folks will ignore them. On the other hand, the advantages of disabling bespoke `encode` and `decode` are (1) we would only have to read the cookies once and (2) there's very little case for having a bespoke solution for de/encoding 3 disallowed characters.
- Documentation. IMO, basically the following. (Wasn't sure where this would go in the docs.)

### Cookies

Unlike all other response headers, which you set with `setHeaders`, `set-cookie` headers should be set and deleted with `cookies`, which is available as part of `RequestEvent`. 
```js
// example tktk
```

The `cookies` API is  built on the `cookie` library, which you may already be familiar with. SvelteKit provides opinionated, sensible defaults for cookie serialization:

```js
{
	httpOnly: true,
	secure: true,
	path: '/',
	sameSite: 'lax'
}
```
Also, implicitly, cookies are set to the session, so they will disappear after the user closes the browser. If you want a long-lasting cookie, set `maxAge` to the number of seconds you want it to last. Other defaults can be changed as necessary by passing your overrides to `cookie.set` and `cookie.delete`.

```js
// example TKTK, including maxAge
```

**Important:** If you do use your own values for `httpOnly`, `Secure`, `SameSite` or `path` to set a cookie, you should use the same values to delete it. Some browsers are picky about this.

**Safari and dev mode:** Safari does not send `Secure` cookies to `localhost` over `http://`. To get around this... (TKTK not sure how far down the rabbit hole we want to go here.)

DoisKoh:
I'm having issues setting cookies when they're being passed between server to server requests. Also when setting them in the load function (in +layout.server.ts). My code was working fine before switching to the new cookies API. Would you happen to know if this might fix those issues as well?

Thanks...

cdcarson:
@DoisKoh Probably. There are some bugs in the current implementation that this PR fixes.

@benmccann  Do I have to take care of the [failing mac tests in the CI](https://github.com/sveltejs/kit/actions/runs/3032137486/jobs/4910171336)? These pass on my mac, and I'm not sure what's going on. Perfectly willing to try to figure it out, but want to check if this is something that happens often, and if there's a trick.  Thanks.

cdcarson:
@benmccann Ignore my last qusetion about mac ci tests. I figured it out.

Rich-Harris:
Thank you @cdcarson, this is awesome. I tried to make some changes locally and push them but for some reason it's not letting me, so I opened a new PR with those changes — we can either continue working in #6811 or you could merge that branch into this one.

The only real question I had was around the `path` default — it feels like this might be an opinion too far. Is it necessary?

cdcarson:
> The only real question I had was around the path default — it feels like this might be an opinion too far. Is it necessary?

@Rich-Harris Pretty sure it is necessary. See https://github.com/sveltejs/kit/issues/6650. Basically a cookie set in a request to `/sign-in` will have that path. A subsequent request to `/profile` won't send that cookie. 

The only issue is if the site base path is `/foo`. We can probably address this in the docs.

Rich-Harris:
Merged #6811, so I'll close this. Thanks @cdcarson!

PR #6658, opened 2022-09-07
Use abort controller in enhance action

f-elix:
This PR adds an `AbortController` to the `fetch` request made by the `enhance` action. It is used to cancel the request if the form is resubmitted while it is still pending, preventing users from sending many requests if they submit the form multiple times too rapidly. This is also the native browser behavior.

I didn't write any test for this because I am unsure on how to test it.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/app/forms.js
+++ b/packages/kit/src/runtime/app/forms.js
@@ -40,6 +40,9 @@ export function enhance(element, submit = () => {}) {
 		element instanceof HTMLFormElement ? element : /** @type {HTMLFormElement} */ (element.form);
 	if (!form) throw new Error('Element is not associated with a form');
 
+	/** @type {AbortController} */
+	let abort_controller;
+
 	/** @param {SubmitEvent} event */
 	async function handle_submit(event) {
 		event.preventDefault();
@@ -60,6 +63,12 @@ export function enhance(element, submit = () => {}) {
 			}) ?? fallback_callback;
 		if (cancelled) return;
 
+		if (abort_controller) {
+			abort_controller.abort();
+		}
+
+		abort_controller = new AbortController();
+
 		/** @type {import('types').ActionResult} */
 		let result;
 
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Thank you. This is slightly more opinionated than I think we want to be — it _should_ be possible to mash an `increment` button, if you haven't explicitly prevented that (e.g. by disabling the button while the submission is pending), and also the behaviour gets a bit harder to reason about with an `AbortController` in the mix (was the previous request aborted before or after the action ran? impossible to say), so I think this is best left in the hands of users.

What if we instead provided the `controller` to the callback? 

```svelte
<script>
  let previous_controller;
</script>

<form
  method="POST"
  use:enhance={({ controller }) => {
    previous_controller?.abort();
    previous_controller = controller;
  }}
>
```

dummdidumm:
Like the idea - will adjust the PR and then merge this

f-elix:
Thanks for the feedback! I take your point regarding the behavior being potentially confusing. 

I'd just like to add that imo mashing an increment button is the kind of situation that requires this setup. Only the last value should be sent to the server. If every click sends a request, you can run into race conditions, because you are not garanteed that the last click value will be the last value sent, since an earlier request could reach the server later than the last one. 

I'd argue that this is the behavior we want in 99% of cases and is also the way the form would behave without js.

Rich-Harris:
> I'd just like to add that imo mashing an increment button is the kind of situation that requires this setup

ah — I think we're envisaging different scenarios. If the action is idempotent (i.e. you're incrementing from a specific number `n` contained in the form to `n + 1`) then yeah, that makes sense. But if the `increment` action just adds one to something, I'd want three clicks to mean three increments, not one or two depending on network conditions. Totally agree that there's scope for buggy app code here, but I think it has to be solved at a higher level than these primitives

f-elix:
Ah gotcha! Hadn't thought of it that way. I guess passing the controller to the callback is the right solution then.

dummdidumm:
Btw I created a different PR with the changes because I couldn't modify this one but I pulled in your changes to make sure you get attributed correctly

Rich-Harris:
closing in favour of #6662 — thanks @f-elix!

PR #6662, opened 2022-09-08
[breaking] only allow enhance on form; pass abort controller

### LINKED ISSUE #6660: `use:enhance` does not properly handle `<button formaction="...">`
### Describe the bug

Adding `use:enhance` to a form like:

```html
<script lang="ts">
	import { enhance } from '$app/forms';
</script>


<form method="POST" action="?/a" use:enhance>
	<button>Submit A</button>
	<button formaction="?/b">Submit B</button>
</form>
```

Always calls `?/a` action. Removing it calls `?/b` when clicking "Submit B".

### Reproduction

https://github.com/danawoodman/sveltekit-formaction-use-enhance-bug

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.5.1
    CPU: (8) x64 Apple M2
    Memory: 23.95 MB / 24.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.4.0 - ~/.asdf/installs/nodejs/18.4.0/bin/node
    npm: 8.12.1 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Brave Browser: 105.1.43.89
    Chrome: 105.0.5195.102
    Firefox: 103.0
    Safari: 15.6.1
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.71 
    @sveltejs/kit: next => 1.0.0-next.476 
    svelte: ^3.44.0 => 3.50.0 
    vite: ^3.1.0 => 3.1.0
```

### Severity

blocking an upgrade

dummdidumm:
This is a breaking change because the `element` property was removed from `enhance` and the action only is possible to apply to forms again; also the interface generic has been removed accordingly.

Fixes #6660
Superseedes/closes #6658 
Reverts #6633, because we can achieve the same with `event.submitter`

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/app/forms.js
+++ b/packages/kit/src/runtime/app/forms.js
@@ -37,24 +37,10 @@ export function enhance(form, submit = () => {}) {
 	async function handle_submit(event) {
 		event.preventDefault();
 
-		let action = form.action;
-		const element_action = /** @type {HTMLButtonElement | HTMLInputElement | null} */ (
-			event.submitter
-		)?.formAction;
-		// the browser will always set formAction - if not set, it defaults to the url;
-		// we therefore have to check that formAction is indeed set
-		if (action !== element_action && element_action) {
-			if (
-				// different urls
-				action.split('?')[0] !== element_action.split('?')[0] ||
-				// same url - formAction needs to be set to something or else we ignore it.
-				// we take advantage of the fact that there cannot be a default and named action
-				// at the same time.
-				element_action.includes('?')
-			) {
-				action = element_action;
-			}
-		}
+		// We can't do submitter.formAction directly because that property is always set
+		const action = event.submitter?.hasAttribute('formaction')
+			? /** @type {HTMLButtonElement | HTMLInputElement} */ (event.submitter).formAction
+			: form.action;
 		const data = new FormData(form);
 		const controller = new AbortController();
 
--- a/packages/kit/src/runtime/app/forms.js
+++ b/packages/kit/src/runtime/app/forms.js
@@ -16,35 +16,30 @@ const ssr = import.meta.env.SSR;
 export const applyAction = ssr ? guard('applyAction') : client.apply_action;
 
 /** @type {import('$app/forms').enhance} */
-export function enhance(element, submit = () => {}) {
+export function enhance(form, submit = () => {}) {
 	/**
 	 * @param {{
-	 *   element: HTMLFormElement | HTMLButtonElement | HTMLInputElement;
-	 *   form: HTMLFormElement;
+	 *   action: string;
 	 *   result: import('types').ActionResult;
 	 * }} opts
 	 */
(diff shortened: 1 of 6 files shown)


dummdidumm:
Questioned my sanity until I discovered that `event.submitter` always seems to have `formAction` set - and to the default action if you don't have set it which causes subtle bugs. My idea is to check the formAction and infer if it is the default value and if the form has a non-default value then take the form's one, because we can't have both a default and a named action. The check also needs to take into account that actions could point to different URLs

repsac-by:
Since we are passing an object to the result callback, then maybe it makes sense to also pass `applyAction`?

```svelte
<form
  method="POST"
  use:enhance={() => {
    return async ({ applyAction, result }) => {
	await applyAction(result);
    };
  }}
>
```

dummdidumm:
Ok so fixed the issue by applying the logic I described above. The abort controller behavior now is that it _not_ invokes the callback return function, which seems like the best thing to do. I didn't want to introduce another `{ type: 'aborted' }`, which you could also make an argument for.

packages/kit/src/runtime/app/forms.js
```diff
@@ -16,46 +16,57 @@ const ssr = import.meta.env.SSR;
...
+				// different urls
+				action.split('?')[0] !== element_action.split('?')[0] ||
+				// same url - formAction needs to be set to something or else we ignore it.
+				// we take advantage of the fact that there cannot be a default and named action
+				// at the same time.
+				element_action.includes('?')
+			) {
+				action = element_action;
+			}
+		}
```
Rich-Harris, line 57: what if we checked for the attribute instead?

```suggestion
		const action = event.submitter?.hasAttribute('formaction')
			? event.submitter.formAction
			: form.action;
```

dummdidumm: Ooh if that works we totally should do that instead

packages/kit/src/runtime/app/forms.js
```diff
@@ -16,46 +16,57 @@ const ssr = import.meta.env.SSR;
...
+		const controller = new AbortController();
 
 		let cancelled = false;
 		const cancel = () => (cancelled = true);
 
 		const callback =
 			submit({
-				element,
-				data,
+				action,
```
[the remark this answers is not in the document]
dummdidumm: Inside the fallback callback we are using the action name to know whether or not we should call `applyAction` . If people would want to do something similar they AFAIK have no way to do this without it, since the submit event with the potential formAction on it is gone by then. If they _have_ a way then we should omit this

packages/kit/src/runtime/app/forms.js
dummdidumm: Works beatifully!

Rich-Harris (approved):

Issue #6462, opened 2022-08-31
devalue cause crash on Netlify after update


wentallout:
### Describe the bug

I updated some sveltekit-related npm packages then `build` successfully on local/dev but when deployed it just crashed on Netlify

```
Error - No "exports" main defined in /var/task/node_modules/devalue/package.json
```

### Reproduction

package.json
```
{
	"name": "portfolio",
	"version": "0.0.1",
	"scripts": {
		"dev": "vite dev",
		"build": "vite build",
		"preview": "vite preview",
		"check": "svelte-check --tsconfig ./jsconfig.json",
		"check:watch": "svelte-check --tsconfig ./jsconfig.json --watch",
		"test": "playwright test",
		"lint": "prettier --check . && eslint .",
		"format": "prettier --write .",
		"ngrok": "ngrok http 4173"
	},
	"devDependencies": {
		"@iconify-json/ph": "^1.1.2",
		"@playwright/test": "^1.25.0",
		"@supabase/supabase-js": "^1.35.6",
		"@sveltejs/adapter-auto": "next",
		"@sveltejs/adapter-netlify": "^1.0.0-next.75",
		"@sveltejs/kit": "next",
		"atropos": "^1.0.2",
		"eslint": "^8.16.0",
		"eslint-config-prettier": "^8.3.0",
		"eslint-plugin-svelte3": "^4.0.0",
		"normalize.css": "^8.0.1",
		"nprogress": "^0.2.0",
		"prettier": "^2.6.2",
		"prettier-plugin-svelte": "^2.7.0",
		"svelte": "^3.49.0",
		"svelte-check": "^2.8.0",
		"typescript": "^4.7.2",
		"unplugin-icons": "^0.14.8",
		"vite": "^3.1.0-beta.1",
		"vite-imagetools": "^4.0.5"
	},
	"type": "module"
}
```

### Logs

```shell
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /var/task/node_modules/devalue/package.json
    at new NodeError (node:internal/errors:372:5)
    at throwExportsNotFound (node:internal/modules/esm/resolve:472:9)
    at packageExportsResolve (node:internal/modules/esm/resolve:693:7)
    at resolveExports (node:internal/modules/cjs/loader:482:36)
    at Function.Module._findPath (node:internal/modules/cjs/loader:522:31)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:919:27)
[...]

PR #6666, opened 2022-09-08
Generate ESM functions for Netlify

eduardoboucas:
Netlify currently transpiles serverless functions with ESM syntax down to CJS, which is causing problems such as https://github.com/sveltejs/kit/issues/6462.

We've started to address that in https://github.com/netlify/zip-it-and-ship-it/pull/1198, which emits ESM files for functions with a `.mjs` extension. Because this is technically a breaking change, we're rolling it out gradually, but we've created a configuration property that allows frameworks like SvelteKit to opt-in to this behaviour straight away. This is done with a per-function configuration file introduced in https://github.com/netlify/zip-it-and-ship-it/pull/1030.

With that in place, this PR has the following changes:

- The conditional logic that decided whether to create CJS or ESM functions was removed, and it now generates ESM only
- The generated function files use the `.mjs` extension
- There's an additional file generated for each function, containing the configuration property mentioned above

We're in the final stages of rolling out this functionality to all customers, but wanted to get the PR in front of you sooner rather than later to gather any feedback and improve the experience for SvelteKit users as soon as possible.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-netlify/index.js
+++ b/packages/adapter-netlify/index.js
@@ -1,7 +1,6 @@
 import { appendFileSync, existsSync, readFileSync, writeFileSync } from 'fs';
 import { dirname, join, resolve, posix } from 'path';
 import { fileURLToPath } from 'url';
-import glob from 'tiny-glob/sync.js';
 import esbuild from 'esbuild';
 import toml from '@iarna/toml';
 
@@ -66,10 +65,6 @@ export default function ({ split = false, edge = edge_set_in_env_var } = {}) {
 				`\n\n/${builder.config.kit.appDir}/immutable/*\n  cache-control: public\n  cache-control: immutable\n  cache-control: max-age=31536000\n`
 			);
 
-			// for esbuild, use ESM
-			// for zip-it-and-ship-it, use CJS until https://github.com/netlify/zip-it-and-ship-it/issues/750
-			const esm = netlify_config?.functions?.node_bundler === 'esbuild';
-
 			if (edge) {
 				if (split) {
 					throw new Error('Cannot use `split: true` alongside `edge: true`');
@@ -77,7 +72,7 @@ export default function ({ split = false, edge = edge_set_in_env_var } = {}) {
 
 				await generate_edge_functions({ builder });
 			} else {
-				await generate_lambda_functions({ builder, esm, split, publish });
+				await generate_lambda_functions({ builder, split, publish });
 			}
 		}
 	};
@@ -148,9 +143,8 @@ async function generate_edge_functions({ builder }) {
  * @param {import('@sveltejs/kit').Builder} params.builder
  * @param { string } params.publish
  * @param { boolean } params.split
- * @param { boolean } params.esm
  */
-async function generate_lambda_functions({ builder, publish, split, esm }) {
+async function generate_lambda_functions({ builder, publish, split }) {
 	builder.mkdirp('.netlify/functions-internal');
 
 	/** @type {string[]} */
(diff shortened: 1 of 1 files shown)


packages/adapter-netlify/index.js
```diff
@@ -1,7 +1,6 @@
 import { appendFileSync, existsSync, readFileSync, writeFileSync } from 'fs';
 import { dirname, join, resolve, posix } from 'path';
 import { fileURLToPath } from 'url';
-import glob from 'tiny-glob/sync.js';
```
benmccann, line 4: we can remove `tiny-glob` from `packages/adapter-netlify/package.json` and then run `pnpm install` to update the `pnpm-lock.yaml` file since `tiny-glob` is no longer used

packages/adapter-netlify/index.js
```diff
@@ -160,19 +154,11 @@ async function generate_lambda_functions({ builder, publish, split, esm }) {
...
-			writeFileSync(filepath, output);
-		});
 
-		builder.copy(`${files}/cjs`, '.netlify', { replace });
-		writeFileSync(join('.netlify', 'package.json'), JSON.stringify({ type: 'commonjs' }));
-	}
+	builder.copy(`${files}/esm`, '.netlify', { replace });
+
+	// Configuring the function to use ESM as the output format.
+	const fnConfig = JSON.stringify({ config: { nodeModuleFormat: 'esm' }, version: 1 });
```
benmccann, line 161: we use underscores rather than camelCase for internal variables
```suggestion
	const fn_config = JSON.stringify({ config: { nodeModuleFormat: 'esm' }, version: 1 });
```

packages/adapter-netlify/index.js
```diff
@@ -201,14 +187,13 @@ async function generate_lambda_functions({ builder, publish, split, esm }) {
...
 					});
 
-					const fn = esm
-						? `import { init } from '../serverless.js';\n\nexport const handler = init(${manifest});\n`
-						: `const { init } = require('../serverless.js');\n\nexports.handler = init(${manifest});\n`;
+					const fn = `import { init } from '../serverless.js';\n\nexport const handler = init(${manifest});\n`;
 
-					writeFileSync(`.netlify/functions-internal/${name}.js`, fn);
+					writeFileSync(`.netlify/functions-internal/${name}.mjs`, fn);
+					writeFileSync(`.netlify/functions-internal/${name}.json`, fnConfig);
```
benmccann, line 196:
```suggestion
					writeFileSync(`.netlify/functions-internal/${name}.json`, fn_config);
```

packages/adapter-netlify/index.js
```diff
@@ -220,14 +205,13 @@ async function generate_lambda_functions({ builder, publish, split, esm }) {
...
+			format: 'esm'
 		});
 
-		const fn = esm
-			? `import { init } from '../serverless.js';\n\nexport const handler = init(${manifest});\n`
-			: `const { init } = require('../serverless.js');\n\nexports.handler = init(${manifest});\n`;
+		const fn = `import { init } from '../serverless.js';\n\nexport const handler = init(${manifest});\n`;
 
-		writeFileSync('.netlify/functions-internal/render.js', fn);
+		writeFileSync(`.netlify/functions-internal/render.json`, fnConfig);
```
benmccann, line 213:
```suggestion
		writeFileSync(`.netlify/functions-internal/render.json`, fn_config);
```

benmccann (commented):
Thank you! Thank you! THANK YOU!

>The generated function files use the .mjs extension

Is it necessary to use the `.mjs` extension if a user has `"type": "module" in their `package.json`? (which all SvelteKit users will have)

>We're in the final stages of rolling out this functionality to all customers, but wanted to get the PR in front of you sooner rather than later to gather any feedback and improve the experience for SvelteKit users as soon as possible.

To clarify, would it be okay if we merge this PR and roll it out to our users now or do we need to wait for something on your side still?

packages/adapter-netlify/index.js
eduardoboucas: Ah, my bad. Thanks!

packages/adapter-netlify/index.js
eduardoboucas: Done in https://github.com/sveltejs/kit/pull/6666/commits/8ebd7f93c5d2a30afe236ae70196ad3d951793e0.

eduardoboucas:
> Is it necessary to use the `.mjs` extension if a user has `"type": "module"` in their `package.json`? (which all SvelteKit users will have)

For now, yes. We'll gradually roll out ESM support for `.js` files, but using `.mjs` allows you to opt-in to this behaviour straight away.

> To clarify, would it be okay if we merge this PR and roll it out to our users now or do we need to wait for something on your side still?

We need to update our CLI with this change, which we'll do today. I'll keep this thread updated and let you know once it's good to merge.

ascorbic:
Does this work properly if somebody has manually set their bundler to esbuild?

eduardoboucas:
> Does this work properly if somebody has manually set their bundler to esbuild?

Yes, esbuild will respect these settings.

benmccann (approved):

benmccann:
Awesome!! Thank you so much!

PR #6678, opened 2022-09-08
make action passed to enhance callback a URL

Rich-Harris:
Already had a use for the new `action` property passed to `enhance`, though I immediately wanted it to be a `URL` instead of a `string`, since the most likely thing you'll want to do with it is check `action.searchParams.has('/foo')` etc

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/app/forms.js
+++ b/packages/kit/src/runtime/app/forms.js
@@ -19,7 +19,7 @@ export const applyAction = ssr ? guard('applyAction') : client.apply_action;
 export function enhance(form, submit = () => {}) {
 	/**
 	 * @param {{
-	 *   action: string;
+	 *   action: URL;
 	 *   result: import('types').ActionResult;
 	 * }} opts
 	 */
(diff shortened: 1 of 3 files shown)


packages/kit/src/runtime/app/forms.js
```diff
@@ -28,7 +28,7 @@ export function enhance(form, submit = () => {}) {
 			await invalidateAll();
 		}
 
-		if (location.origin + location.pathname === action.split('?')[0]) {
+		if (location.origin + location.pathname === action.origin + action.pathname) {
```
s3812497, line 31: What's the reason for having this `if` statement?
Would removing it fix #6695 ?

dummdidumm (approved):

packages/kit/src/runtime/app/forms.js
dummdidumm: It would fix it, yes, but the if check is there deliberately. I'll share more details on the issue.

PR #6684, opened 2022-09-09
[breaking] limit adapter-node request size

benmccann:
closes https://github.com/sveltejs/kit/issues/6542

Issue #6542, opened 2022-09-03
Limit size of post requests


benmccann:
### Describe the problem

I believe a malicious user could bring down the server by issuing a very large post request. I think requests get buffered into memory, so you could exhaust the server memory.

### Describe the proposed solution

Next.js has a default 1mb limit: https://nextjs.org/docs/api-routes/request-helpers#custom-config

Express has a default 100kb limit: https://github.com/expressjs/body-parser#limit-3

Play Framework has an in memory limit as well as a disk limit:
https://www.playframework.com/documentation/2.8.x/JavaBodyParsers

I'm not sure if a server-wide limit is enough or if we also need to be able to set it on a per-endpoint basis or some other granularity.

It should be configurable 

### Importance

nice to have

Rich-Harris:
What would that mean beyond doing this?

```js
export function handle({ event, resolve }) {
  const length = +(event.request.headers.get('content-length') ?? '0');
  if (length > MAX_SIZE) {
    throw error(413);
  }

  return resolve(event);
}
```

Could also reject POST requests without a content-length, but you get the idea. (I'm assuming — haven't tested this — that it's not possible for the requester to lie about the content-length, i.e. declare a short content-length but then have a huge payload. I assume `request.formData()` or `request.body.getReader()` would blow up in that case.)

One wrinkle is that if you wanted to do this on a per-route basis, we would need to change the [proposed actions API](https://github.com/sveltejs/kit/discussions/5875#discussioncomment-3540263):

```diff
export const actions = {
  /** @type {import('./$types').Action} */
-  uploadPhoto: ({ fields, files }) => {
+  uploadPhoto: ({ request }) => {
+    const length = +event.request.headers.get('content-length');
+    if (length > MAX_SIZE) throw error(413);
+    const { fields, files } = await formData(request);
    const title = fields.get('title');
    const image = files.get('image');

    // ...
  }
}
```

Perhaps that's a good thing though. It would mean we could specify `handleFile` behaviour per-action, if necessary, and maybe is a better way to provide type safety than generic arguments to `Action`. In the common case where no files are involved, a developer would just do `await request.formData()`.

Mlocik97:
[...]

PR #6702, opened 2022-09-09
Allow streaming with body_size_limit

repsac-by:
Allow streaming with body_size_limit
Revert #6684

### Tests

### Changesets

### DIFF
--- a/packages/adapter-node/README.md
+++ b/packages/adapter-node/README.md
@@ -71,7 +71,7 @@ Instead, we read from the _right_, accounting for the number of trusted proxies.
 
 ### `BODY_SIZE_LIMIT`
 
-The maximum request body size to accept in bytes. Defaults to 512kb. This option does not allow streaming. You can disable this option with a value of 0 and implement a custom check in [`handle`](https://kit.svelte.dev/docs/hooks#server-hooks-handle) if you need something more advanced.
+The maximum request body size to accept in bytes. Defaults to 512kb. You can disable this option with a value of 0 and implement a custom check in [`handle`](https://kit.svelte.dev/docs/hooks#server-hooks-handle) if you need something more advanced.
 
 ## Options
 
--- a/packages/kit/src/exports/node/index.js
+++ b/packages/kit/src/exports/node/index.js
@@ -11,7 +11,7 @@ function get_raw_body(req, body_size_limit) {
 		return null;
 	}
 
-	const length = Number(h['content-length']);
+	let length = Number(h['content-length']);
 
 	// check if no request body
 	if (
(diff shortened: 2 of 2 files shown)


packages/kit/src/exports/node/index.js
```diff
@@ -59,7 +56,7 @@ function get_raw_body(req, body_size_limit) {
 
 				size += chunk.length;
 				if (size > length) {
-					controller.error(new Error('content-length exceeded'));
+					req.destroy(new Error('request body size exceeded'));
```
benmccann, line 59: you could probably do something like check whether `content-length` or `body_size_limit` is smaller outside this loop and provide an appropriate error message here which is more detailed as to which one was exceeded

benmccann (commented):
this works, but seems like a minor improvement as it only allows small streams

I think we could also allow larger streams as suggested here: https://github.com/sveltejs/kit/issues/6542#issuecomment-1240288159. @Rich-Harris pushed back against the idea, but I'm not sure if he understood what I was suggesting or if there was something I didn't understand about his response

packages/kit/src/exports/node/index.js
```diff
@@ -59,7 +58,13 @@ function get_raw_body(req, body_size_limit) {
 
 				size += chunk.length;
 				if (size > length) {
-					controller.error(new Error('content-length exceeded'));
+					req.destroy(
+						new Error(
+							`request body size exceeded ${
+								content_length ? "specified 'content-length'" : 'allowed by the server'
```
benmccann, line 64: this would read "request body size exceeded allowed by server" which isn't a sentence that makes sense. we could change it like this:
```suggestion
								content_length ? "'content-length'" : 'BODY_SIZE_LIMIT'
```

packages/adapter-node/README.md
```diff
@@ -71,7 +71,7 @@ Instead, we read from the _right_, accounting for the number of trusted proxies.
 
 ### `BODY_SIZE_LIMIT`
 
-The maximum request body size to accept in bytes. Defaults to 512kb. This option does not allow streaming. You can disable this option with a value of 0 and implement a custom check in [`handle`](https://kit.svelte.dev/docs/hooks#server-hooks-handle) if you need something more advanced.
+The maximum request body size to accept in bytes. Defaults to 512kb. You can disable this option with a value of 0 and implement a custom check in [`handle`](https://kit.svelte.dev/docs/hooks#server-hooks-handle) if you need something more advanced.
```
benmccann, line 74:
```suggestion
The maximum request body size to accept in bytes including while streaming. Defaults to 512kb. You can disable this option with a value of 0 and implement a custom check in [`handle`](https://kit.svelte.dev/docs/hooks#server-hooks-handle) if you need something more advanced.
```

packages/kit/src/exports/node/index.js
repsac-by: @benmccann thank you

packages/kit/src/exports/node/index.js
```diff
@@ -59,7 +58,13 @@ function get_raw_body(req, body_size_limit) {
 
 				size += chunk.length;
 				if (size > length) {
-					controller.error(new Error('content-length exceeded'));
+					req.destroy(
+						new Error(
+							`request body size exceeded ${
+								content_length ? "'content-length'" : 'BODY_SIZE_LIMIT'
+							}: ${length}`
```
benmccann, line 65: one last suggestion. it might be slightly unclear which length is being printed, so I think this small wording change would help
```suggestion
							} of ${length}`
```

benmccann (approved):

Rich-Harris:
thank you!

PR #6710, opened 2022-09-09
Improve performance by compressing in parallel

sanjaiyan-dev:
Improving the performance by running the independent async functions in parallel ✈️.

### DIFF
--- a/packages/adapter-node/index.js
+++ b/packages/adapter-node/index.js
@@ -32,7 +32,7 @@ export default function (opts = {}) {
 
 			if (precompress) {
 				builder.log.minor('Compressing assets');
-				await Promise.allSettled([builder.compress(`${out}/client`), builder.compress(`${out}/prerendered`)])
+				await Promise.allSettled([builder.compress(`${out}/client`), builder.compress(`${out}/prerendered`)]);
 			}
 
 			builder.log.minor('Building server');
--- a/packages/adapter-node/index.js
+++ b/packages/adapter-node/index.js
@@ -32,7 +32,10 @@ export default function (opts = {}) {
 
 			if (precompress) {
 				builder.log.minor('Compressing assets');
-				await Promise.allSettled([builder.compress(`${out}/client`), builder.compress(`${out}/prerendered`)]);
+				await Promise.allSettled([
+					builder.compress(`${out}/client`),
+					builder.compress(`${out}/prerendered`)
+				]);
 			}
 
 			builder.log.minor('Building server');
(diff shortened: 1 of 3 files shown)


benmccann:
You need to run `pnpm format` to get the Lint CI check to pass. Right now it's failing with:
> @sveltejs/adapter-static:lint: [warn] Code style issues found in the above file. Forgot to run Prettier?

sanjaiyan-dev:
> You need to run `pnpm format` to get the Lint CI check to pass. Right now it's failing with:
> 
> > @sveltejs/adapter-static:lint: [warn] Code style issues found in the above file. Forgot to run Prettier?

Extremely sorry, I used GitHub web based editor so I couldn't run `lint` command, Can you please help me with it :)

benmccann:
You can address it by checking out the code on your machine:
```
git clone git@github.com:sanjaiyan-dev/svelte-kit.git
cd svelte-kit
git checkout sanjaiyan-static-parallel
pnpm install
pnpm format
```

sanjaiyan-dev:
> You can address it by checking out the code on your machine:
> 
> ```
> git clone git@github.com:sanjaiyan-dev/svelte-kit.git
> cd svelte-kit
> git checkout sanjaiyan-static-parallel
> pnpm install
> pnpm format
> ```

Tks a lot

packages/adapter-node/index.js
```diff
@@ -32,8 +32,10 @@ export default function (opts = {}) {
 
 			if (precompress) {
 				builder.log.minor('Compressing assets');
-				await builder.compress(`${out}/client`);
-				await builder.compress(`${out}/prerendered`);
+				await Promise.allSettled([
```
benmccann, line 35: wouldn't it be better to use `Promise.all` so that if one were to fail for some reason then we'd get the failure immediately? or is there some benefit to using `allSettled`?

benmccann (commented):
I pushed the formatting change for you

packages/adapter-node/index.js
sanjaiyan-dev: Sorry, actually I added `allSettled` thinking that code needs to be executed even after any one of the promises failed. 

Yh if it is needed to thrown error preferred to change to `all` 🤝

sanjaiyan-dev:
> I pushed the formatting change for you

Tks 💪🙌

packages/adapter-node/index.js
sanjaiyan-dev: > wouldn't it be better to use `Promise.all` so that if one were to fail for some reason then we'd get the failure immediately? or is there some benefit to using `allSettled`?

Done :)

benmccann (approved):
Thanks!!

PR #6712, opened 2022-09-09
Preloading critical style-sheets for improved performance 🚤  

sanjaiyan-dev:
This is inspired from `Next.js`.
[Preload critical stylesheets to improve loading speed](https://web.dev/preload-critical-assets/).

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -248,6 +248,7 @@ export async function render_response({
 		}
 
 		attributes.unshift('rel="stylesheet"');
+		head += `\n\t<link rel="preload" href="${path}" as="style">`;
 		head += `\n\t<link href="${path}" ${attributes.join(' ')}>`;
 	}
 


packages/kit/src/runtime/server/page/render.js
```diff
@@ -248,6 +248,7 @@ export async function render_response({
 		}
 
 		attributes.unshift('rel="stylesheet"');
+		head += `\n\t<link rel="preload" href="${path}" as="style">`;
```
rcabralc, line 251: I think this preloading is already being done by the Link header. Maybe the header should be replaced with this (might benefit prerendered pages)? But only if it's not inlined, I guess.

sanjaiyan-dev: > I think this preloading is already being done by the Link header. Maybe the header should be replaced with this (might benefit prerendered pages)? But only if it's not inlined, I guess.

Does adding both have any drawbacks ?

benmccann: > Maybe the header should be replaced with this (might benefit prerendered pages)?

The header is better performance. We fallback to including in the page when prerendering: https://github.com/sveltejs/kit/blob/56cb62dcbad93b6317f98aa93c87f469f0b199eb/packages/kit/src/runtime/server/page/render.js#L260

benmccann:
This isn't really mergable and there isn't enough detail in the issue description to understand if you're actually improving anything. It seems like you're mostly just duplicating existing functionality. If there's some issue with the existing implementation I'd recommend first filing an issue so we can get a clear understanding of what the issue is

PR #6740, opened 2022-09-11
[chore] update adapters to esbuild 0.15

Conduitry:
I was wondering why my project's lockfile had gotten so much longer after a recent upgrade, and at least part of the reason is that Vite 3.1.0 is bringing in esbuild 0.15 and the adapters are bringing in 0.14. This updates all of the adapters to ^0.15.6, the same range as Vite currently uses.

While I was at it, I also updated some dev deps on Vite throughout this repo from the 3.1.0 beta to ^3.1.0, so we don't have two versions of Vite getting installed.

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/templates/default/package.json
+++ b/packages/create-svelte/templates/default/package.json
@@ -13,7 +13,7 @@
 		"svelte": "^3.48.0",
 		"svelte-preprocess": "^4.10.6",
 		"typescript": "^4.8.2",
-		"vite": "3.1.0-beta.2"
+		"vite": "^3.1.0"
 	},
 	"type": "module",
 	"dependencies": {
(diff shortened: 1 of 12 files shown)


packages/create-svelte/templates/default/package.json
```diff
@@ -13,7 +13,7 @@
 		"svelte": "^3.48.0",
 		"svelte-preprocess": "^4.10.6",
 		"typescript": "^4.8.2",
-		"vite": "3.1.0-beta.2"
```
benmccann, line 16: we should probably add `create-svelte` to the changelog as well

dummdidumm (approved):

packages/create-svelte/templates/default/package.json
Conduitry: This isn't the package.json that's used when people run the CLI. That's package.template.json, which was already updated.

Conduitry:
The change to the changeset in bbc4e4e59989bcc4807371f943ce065a7504cc7b was not correct. This doesn't update any user-visible dependencies to Vite 3.1.0, and doesn't affect what create-svelte creates. I've reverted it in b68f242806047a3430f79375926c88c3abda7ba1.

PR #6756, opened 2022-09-12
Added comment about hook files and initialization

### LINKED ISSUE #1753: Discussion: Documentation around singletons/runtime setup (db connections seem most common)
This is a request for discussion.

One problem brought up in the Discord server pretty often revolves around one-time setup that should happen at startup, where "startup" is either defined as "whenever the server starts" or, for `adapter-static`, "whenever the app runs on the user's browser". 

In the FAQs, we have a "How do I setup a database?" section. This section documents functionality that is otherwise not documented anywhere, specifically around `hooks.js`. It says "You can execute any one-time setup code in hooks.js and import your database helpers into any endpoint that needs them." I imagine that would look something like this:

```ts
let ready = false;
const executeSetupActions = async (): void => {
  const t1 = setupDb1();
  const t2 = setupDb2();
  const t3 = setupTelemetry();
  const t4 = setupAuth();
  await Promise.all([t1, t2, t3, t4]);
  ready = true;
}();

const handle: Handle = () => {
  // For server-side code, this shouldn't REALLY matter unless you receive a request in the short
  // amount of time AFTER the server has started but BEFORE the connections are initialized
  while (!ready) {
    await new Promise(resolve => setTimeout(resolve, 2000));
  }
}
```

1. In the Hooks section of the documentation, document that setup code can be run in hooks _outside_ of the two exported functions `getSession` and `handle`. (It might be nice to also document what this does if you're using `adapter-static`.) If the SvelteKit-approved place to run setup code is in `hooks.js`, it should be documented with the documentation about `hooks.js`.
2. Include something like the code examples above in the "setting up a database" FAQ.'
3. It might also be a good idea to just document how to set up singletons. Yes, you _should_ know this as a JavaScript developer, but it's such a common use case that it might fit in nicely at the bottom of the FAQs.

1. Just leave it alone and continue answering this question in the Discord... often.

Important enough that I'm willing to make said adjustments in a pull request if I get the go-ahead from a maintainer :)

### LINKED ISSUE #6183: Setup each page - init hook, `export const blocking`, etc.
### Describe the problem

For i18n or similar subsystems, we should make initialization before any language things are needed, and for such initialization, we usually use a session (user language) and information from the HTTP request. 
Before the new routing system was introduced, we were using the root __layout.svelte to do it, it's working well for SSR and CSR time. 
I should mention that what i18n should be inited on the server side and init on the client side as well (or state should be send as is). 

After migration to new routing, we lose one "transitional" place to init such thing because `+layout.js` can be called after `+page.js`, and you must do `await parent()` in each page to avoid such race condition.  

For server-side only init, we have a good place it's `handle` in `hooks.js` with `event.locals` it's an excellent way to init server only things like DB connection, but for things what should be a transition from SSR to CSR we have no place. 

I also have a similar issue for i18n implementation project what I use https://github.com/cibernox/svelte-intl-precompile/issues/55

Maybe it's possible to do it more straightforwardly, but at least we need a global hook for what will be called on SSR and on CSR, after `+layout.server.js` but before `+layout.js` and `+page.js`. 

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

sphinxc0re:
Closes #1753. I added a line of documentation that clarifies that code in hook files runs on startup.

### Tests

### Changesets

### DIFF
--- a/documentation/docs/07-hooks.md
+++ b/documentation/docs/07-hooks.md
@@ -9,6 +9,8 @@ There are two hooks files, both optional:
 - `src/hooks.server.js` — your app's server hooks
 - `src/hooks.client.js` — your app's client hooks
 
+Those files can also be used to run code that runs at the start of the application. This can be useful for running initialization code for database ORMs or various other libraries.
+
 > You can configure the location of these files with [`config.kit.files.hooks`](/docs/configuration#files).
 
 ### Server hooks
--- a/documentation/docs/07-hooks.md
+++ b/documentation/docs/07-hooks.md
@@ -9,7 +9,7 @@ There are two hooks files, both optional:
 - `src/hooks.server.js` — your app's server hooks
 - `src/hooks.client.js` — your app's client hooks
 
-Those files can also be used to run code that runs at the start of the application. This can be useful for running initialization code for database ORMs or various other libraries.
+Those files can also be used to run code that runs at the start of the application. This can be useful for running initialization code for database clients or various other libraries.
 
 > You can configure the location of these files with [`config.kit.files.hooks`](/docs/configuration#files).
 
--- a/documentation/docs/07-hooks.md
+++ b/documentation/docs/07-hooks.md
@@ -9,7 +9,7 @@ There are two hooks files, both optional:
 - `src/hooks.server.js` — your app's server hooks
 - `src/hooks.client.js` — your app's client hooks
 
-Those files can also be used to run code at the start of the application. This can be useful for running initialization code for database clients or various other libraries.
+Code in these modules will run when the application starts up, making them useful for initializing database clients and so on.
 
 > You can configure the location of these files with [`config.kit.files.hooks`](/docs/configuration#files).
 
--- a/documentation/docs/07-hooks.md
+++ b/documentation/docs/07-hooks.md
@@ -9,7 +9,7 @@ There are two hooks files, both optional:
 - `src/hooks.server.js` — your app's server hooks
 - `src/hooks.client.js` — your app's client hooks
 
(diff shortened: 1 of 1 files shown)


documentation/docs/07-hooks.md
```diff
@@ -9,6 +9,8 @@ There are two hooks files, both optional:
 - `src/hooks.server.js` — your app's server hooks
 - `src/hooks.client.js` — your app's client hooks
 
+Those files can also be used to run code that runs at the start of the application. This can be useful for running initialization code for database ORMs or various other libraries.
```
benmccann, line 12: If we want to add a comment like this I'd probably make it a bit more general to database clients in general and not just ORMs
```suggestion
Those files can also be used to run code that runs at the start of the application. This can be useful for running initialization code for database clients or various other libraries.
```

tcc-sejohnson:
I don't know that I'm comfortable with this closing #6183, as `export const blocking` isn't really solved by this at all -- it's _only_ a server-side initializer. It would close #1753, though.

documentation/docs/07-hooks.md
```diff
@@ -9,6 +9,8 @@ There are two hooks files, both optional:
 - `src/hooks.server.js` — your app's server hooks
 - `src/hooks.client.js` — your app's client hooks
 
+Those files can also be used to run code at the start of the application. This can be useful for running initialization code for database clients or various other libraries.
```
Rich-Harris, line 12:
```suggestion
Code in these modules will run when the application starts up, making them useful for initializing database clients and so on.
```

Rich-Harris (approved):

Rich-Harris:
thanks!

Issue #5123, opened 2022-05-31
Auto adapter should download other adapters on demand


Conduitry:
### Describe the problem

Currently, `adapter-auto` has the other official serverless adapters as prod dependencies, so it can delegate to the correct one at runtime based on the current environment variables. This used to be an acceptable inefficiency, but the recent addition of `@vercel/nft` to the Vercel adapter (and thus to the auto adapter) added a lot of dependencies.

In the barest SvelteKit starter template, two thirds of the dependencies that get installed are only used by the auto adapter.

### Describe the proposed solution

To make builds reproducible, each version of the auto adapter should pull in specific fixed versions of all of the other adapters, rather than pull in the latest one. Each time any other dependency has a new version cut, the auto adapter should also get a new release, which installs that new version.

It would be nice to be able to use pnpm/changesets to automate the release of the auto adapter whenever another adapter is getting released. We might be able to use peerdeps on the other adapters at version `workspace:*` and have this automatically handled, like `workspace:*` prod deps do. If we do this, we should use `peerDependenciesMeta` to make sure that installing the auto adapter doesn't install the other adapters, which would defeat the whole purpose here.

### Importance

nice to have

### Additional Information

If we are only installing these dependencies on-demand and as part of the build happening in the cloud provider's remote build servers, I'm not sure whether it matters a ton whether we use the same package manager as was used to install the rest of the project's dependencies. If you run `npm install something` in a package that had previously used pnpm, does that remove any of the previous dependencies?

Should the selected adapter be installed as a dependency of the SvelteKit app or of the auto adapter? Since the ESM analogues of the `require.resolve` API aren't all in place yet, would one make our job harder when trying to delegate to the appropriate underlying adapter?

dummdidumm:
How reliable would this be? Are there ways you can invoke the build that fail detection of what package manger is in use or where to install to? Would people notice that the package manager lockfile is changed during build and be confused about that?

Conduitry:
I don't really think many of those would be concerns. I believe we should be able to call npm with `--no-save --no-package-lock` (or something like that) to have it not update any of the files, if that was the goal. Even if they are changed, I don't think that would be something people would even notice if it did happen, since the adapter would only be kicking in when the site is built in its host's remote build server (because each one looks for specific environment variables being set). And if you're using your host's provided build service that runs on their end, I doubt you'd be able to have the control to split things into different phases.

[...]

PR #6761, opened 2022-09-13
[feat] support Azure Static Web Apps in adapter-auto

geoffrich:
This PR adds a check for [Azure Static Web Apps](https://docs.microsoft.com/en-us/azure/static-web-apps/overview) to adapter-auto. If it succeeds, it uses [svelte-adapter-azure-swa](https://github.com/geoffrich/svelte-adapter-azure-swa), which I maintain. This package is not automatically added as a dependency of adapter-auto, users need to install it first.

The check uses the `GITHUB_ACTION_REPOSITORY` [environment variable](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) to check if we are building using the [Azure SWA action](https://github.com/Azure/static-web-apps-deploy). Azure SWA builds on GitHub actions by default.

I tested this using patch-package on one of my SWA projects and it seemed to work.

### Tests

### Changesets

### DIFF
--- a/.changeset/eight-vans-exercise.md
+++ b/.changeset/eight-vans-exercise.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-auto': patch
+---
+
+[feat] support Azure SWA
--- a/packages/adapter-auto/README.md
+++ b/packages/adapter-auto/README.md
@@ -9,6 +9,7 @@ The following environments are supported out-of-the-box, meaning a newly created
 - [Cloudflare Pages](https://developers.cloudflare.com/pages/) via [adapter-cloudflare](../adapter-cloudflare)
 - [Netlify](https://netlify.com/) via [adapter-netlify](../adapter-netlify)
 - [Vercel](https://vercel.com/) via [adapter-vercel](../adapter-vercel)
+- [Azure Static Web Apps](https://docs.microsoft.com/en-us/azure/static-web-apps/) via [svelte-adapter-azure-swa](https://github.com/geoffrich/svelte-adapter-azure-swa)
 
 ## Community adapters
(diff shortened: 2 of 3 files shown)


benmccann:
What happens if the user doesn't have it installed?

I'm a bit wary of this as it seems like a slippery slope where we'll start getting PRs to include all the various adapters in `adapter-auto`

I think `adapter-auto` needs a rethink. It drives me up the wall because it pulls in all sorts of crazy dependencies into every project via vercel/nft. It should probably install the necessary dependency rather than including them all or we should just kill the package or something

Rich-Harris:
> What happens if the user doesn't have it installed?

This is very much how it's supposed to work — for non-'core' adapters, we tell you how to install the adapter but we don't do it for you. Not a slippery slope, the more the merrier (as long as we don't end up in a situation where there are multiple competing adapters for a given platform).

There's an issue for automatically installing other adapters, precisely because of the `@vercel/nft` thing, which I would love to get rid of: https://github.com/sveltejs/kit/issues/5123. If we do that, we could consider making _all_ adapters recognised by `adapter-auto` install automatically. We should definitely have that conversation but for now I'll merge this — thanks @geoffrich

Rich-Harris (approved):

Conduitry:
> we could consider making _all_ adapters recognised by `adapter-auto` install automatically

This worries me. I talk in #5123 about the concerns of reproducible builds, and I don't know how we achieve that with third-party adapters without having a release of `adapter-auto` for every single release of all of these third-part adapters, which sounds like a mess. Unless we want to just say that you don't get the same reproducible build guarantees when pulling in third-party adapters?

PR #6773, opened 2022-09-13
[feat] allow +server.js files next to +page files

### LINKED ISSUE #5896: Bring back support for content negotiation (allow +page.svelte & +server.ts to co-exist)
### Describe the problem

Previous versions of SvelteKit supported content negotiation where a route could serve HTML to browsers or JSON data if an `accept` header was used for a data request (as well as supporting data being posted to the same route).

Existing sites may already use this, it's been part of HTTP since forever. You might want to replace such a site with SvelteKit (or previously relied on SvelteKit supporting this) but changing the existing URL structure would be problematic and impactful (other systems using endpoints, pages already indexed and publicized etc...)

I think a router should be flexible enough to support existing URL structures, even if they are not the favoured default for the framework.

Support the `accept` header in the routing and allow `application/json` requests to an endpoint at the same route. Pages should only be concerned with GET requests with no `accept` header (or one for html content).

### Alternatives considered

Adding a load function to `+page.server.ts` allows a GET request for data, but makes it difficult to control the response and would limit things to JSON (maybe you want CSV, PDF, or XML). It also appears limited in terms of what control you have over the response (http headers etc...) and still leaves you unable to use other HTTP methods.

Update: it appears that I should be treating `load` as GET and the other methods are supported. However they are inconsistent in terms of signature and what response is returned.

### Importance

### Additional Information

_No response_

dummdidumm:
Closes #5896

Also added a related paragraph to "Form Actions" because I saw some people asking about this specifically.

Marking as draft because I don't know what to make of this TODO:
https://github.com/sveltejs/kit/blob/55493cb64f420e413b91154cef6ad54a03b96d1f/packages/kit/src/core/adapt/builder.js#L49

### Tests

### Changesets


documentation/docs/06-form-actions.md
```diff
@@ -349,3 +349,37 @@ We can also implement progressive enhancement ourselves, without `use:enhance`,
 	<!-- content -->
 </form>
 ```
+
+### Alternatives
+
+Form actions are specifically designed to also work without JavaScript, which means you need to use forms for it. If you don't need the page to work without JavaScript and/or you for example want to interact with an API through JSON, you can instead use `+server.js` files:
```
benmccann, line 355: @Rich-Harris shared a link somewhere in the forms discussion on Discord about all the ways you might end up without javascript (even if it's not disabled, you might lose a network connection, etc.). I think it'd be helpful to include that here because it had a lot of possibilities I hadn't given much thought to and was really educational and thought provoking

should be "them" instead of "it" since it's referring to "actions" which is plural. the "and/or you for example" portion is a bit tricky to read so here's an alternative for that as well
```suggestion
Form actions are specifically designed to also work without JavaScript, which means you need to use forms with them. If you don't need the page to work without JavaScript, you can instead use `+server.js` files to interact with an API through JSON:
```

dummdidumm: Like the rewording, maybe Rich can put the link in here, too, so we can add it (can't find it)

packages/kit/src/runtime/server/endpoint.js
```diff
@@ -64,3 +65,24 @@ export async function render_endpoint(event, mod, state) {
...
+				'application/json') ||
+		// POST with FormData is for actions
+		(event.request.method === 'POST' &&
+			!(event.request.headers.get('content-type') ?? '')
+				.split(';')
+				.some((part) =>
+					['multipart/form-data', 'application/x-www-form-urlencoded'].includes(part.trim())
+				));
+	return result;
+}
```
Rich-Harris, line 88: This function is quite hard to read — something like this might be simpler:

```suggestion
export function is_endpoint_request(event) {
	const { method } = event.request;

	if (method === 'PUT' || method === 'PATCH' || method === 'DELETE') {
		return true;
	}

	if (method === 'GET') {
		const accept = event.request.headers.get('accept') ?? '*/*';
		return negotiate(accept, ['application/json', 'text/html']) === 'application/json';
	}

	if (method === 'POST') {
		const type = (event.request.headers.get('content-type') ?? '').split(';')[0];
		return type !== 'multipart/form-data' && type !== 'application/x-www-form-urlencoded';
	}

	return false;
}
```

That said, I'm not 100% sure this is the right logic. It's totally possible that someone would want to post form data to a `+server.js` handler, for example. I think we should maybe make it simpler: for both `GET` and `POST`, if the request is for `text/html` use `+page.server.js`, otherwise use `+server.js`:

```js
return negotiate(accept, ['*', 'text/html']) !== 'text/html';
```

documentation/docs/06-form-actions.md
Rich-Harris: To be honest I'm not totally sure this belongs in this section? I think it would be enough to mention in the `+server.js` documentation that it can live alongside `+page` files, and explain the content negotiation rules

packages/kit/src/runtime/server/endpoint.js
dummdidumm: Not sure if this has the desired outcome in case of a progressive enhancement case, the request isn't for `text/html` in this case. The safest way to check this is probably to check if the `+page.server.js` contains `actions`, and if so, then formdata requests are routed to that.

documentation/docs/06-form-actions.md
dummdidumm: I think that's a good thing to do in _addition_ to this paragraph. I saw some people going "oh wait so is this the only way I can do that?" and "so why can't I use this with JSON?". Having this clearly laid out at the bottom of this section is a good thing IMO.

packages/kit/src/runtime/server/endpoint.js
Rich-Harris: I think the rule needs to be that if you have a `+server.js` file, you're handling all non-HTML requests yourself. If we make assumptions to the contrary we're getting in people's way — the people likely to use content negotiation are the same people who are likely to be upset if we decide to invoke an action rather than a handler on their behalf.

If people _do_ want to invoke the action from the handler, they could potentially do something like this (once we add `fetch` to `RequestEvent`) — it's not the prettiest code...

```js
// src/routes/foo/+server.js
import { actions } from './+page.server.js';

export function POST({ fetch, request }) {
  return fetch(request.url, {
    method: request.method,
    body: request.body,
    headers: {
      accept: 'text/html'
    }
  });
}
```

...but it's a hell of a lot nicer than the alternative where you need to invoke a handler from an action because SvelteKit made that choice on your behalf.

Ultimately, if you have `+page.server.js` and `+server.js` in the same directory, you are opting for more control (potentially at the cost of some convenience), and a rule that says 'you are now responsible for handling non-HTML requests' is simpler than 'you are now responsible for handling non-HTML requests, except in the case where it's a `POST` request whose `content-type` begins with either `multipart/form-data` or `application/x-www-form-urlencoded`.

dummdidumm: That makes sense. What if we build in some kind of escape hatch for this situation? Like "if you want to call the form action use the `x-sveltekit-formaction` header"?

packages/kit/src/core/adapt/builder.js
```diff
@@ -46,7 +46,7 @@ export function create_builder({ config, build_data, routes, prerendered, log })
 
 				return {
 					id: route.id,
-					type: route.page ? 'page' : 'endpoint', // TODO change this if support pages+endpoints
+					type: route.page && route.endpoint ? 'both' : route.page ? 'page' : 'endpoint',
```
dummdidumm, line 49: It looks like this was never used by one of our adapters. Adding a third option is non-breaking, so any community adapters that _might_ use this can adapt

packages/kit/types/private.d.ts
```diff
@@ -195,7 +195,7 @@ export interface RequestOptions {
 
 export interface RouteDefinition {
```
dummdidumm, line 196: aside: it's strange to me that this is private API - it's public because it's part of the public builder methods

packages/kit/src/runtime/server/endpoint.js
Rich-Harris: That's the sort of thing we could potentially add later if people asked for it. My hunch is that they won't

packages/kit/src/core/adapt/builder.js
Rich-Harris: I checked all the community adapters, not one of them is using this. I think we can safely remove it altogether

```suggestion
```

Rich-Harris (approved):

PR #6776, opened 2022-09-13
[docs] note form prop update default behavior

### LINKED ISSUE #6695: Form `enhance` does not work when `action` points to a different route
### Describe the bug

`enhance` without a callback will not apply its action if the form action references a different route. However, it works when manually calling `applyAction(result)` in the callback.
For example:

```html
// routes/+page.svelte
<form use:enhance method="POST" action="/login"> // this doesn't work
<form use:enhance method="POST"> // this works
```

### Reproduction

https://github.com/s3812497/sveltekit-enhance-bug-2

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 13.0
    CPU: (8) arm64 Apple M1 Pro
    Memory: 8.07 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.0 - ~/.nvm/versions/node/v16.17.0/bin/node
    npm: 8.15.0 - ~/.nvm/versions/node/v16.17.0/bin/npm
  Browsers:
    Chrome: 104.0.5112.79
    Firefox: 104.0.1
    Firefox Developer Edition: 105.0
    Safari: 16.1
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.479 
    svelte: ^3.50.1 => 3.50.1 
    vite: ^3.1.0 => 3.1.0
```

### Severity

### Additional Information

_No response_

dummdidumm:
Closes #6695

### Tests

### Changesets

### DIFF
--- a/documentation/docs/06-form-actions.md
+++ b/documentation/docs/06-form-actions.md
@@ -249,11 +249,13 @@ The easiest way to progressively enhance a form is to add the `use:enhance` acti
 
 Without an argument, `use:enhance` will emulate the browser-native behaviour, just without the full-page reloads. It will:
 
-- update the `form` property and invalidate all data on a successful response
-- update the `form` property on a invalid response
+- update the `form` property if the action belongs to the page and invalidate all data on a successful response
+- update the `form` property on a invalid response if the action belongs to the page
 - update `$page.status` on a successful or invalid response
 - call `goto` on a redirect response
 - render the nearest `+error` boundary if an error occurs
+
+> By default the `form` property is only updated for actions that belong to the page because in the native form submission case you would be redirected to the page the action is on
 
 To customise the behaviour, you can provide a function that runs immediately before the form is submitted, and (optionally) returns a callback that runs with the `ActionResult`.
 
--- a/documentation/docs/06-form-actions.md
+++ b/documentation/docs/06-form-actions.md
@@ -255,7 +255,7 @@ Without an argument, `use:enhance` will emulate the browser-native behaviour, ju
 - call `goto` on a redirect response
 - render the nearest `+error` boundary if an error occurs
 
-> By default the `form` property is only updated for actions that belong to the page because in the native form submission case you would be redirected to the page the action is on
+> By default the `form` property is only updated for actions that are in `+page.server.js` alongside `+page.svelte` because in the native form submission case you would be redirected to the page the action is on
 
 To customise the behaviour, you can provide a function that runs immediately before the form is submitted, and (optionally) returns a callback that runs with the `ActionResult`.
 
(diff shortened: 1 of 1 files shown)


documentation/docs/06-form-actions.md
```diff
@@ -249,12 +249,14 @@ The easiest way to progressively enhance a form is to add the `use:enhance` acti
 
 Without an argument, `use:enhance` will emulate the browser-native behaviour, just without the full-page reloads. It will:
 
-- update the `form` property and invalidate all data on a successful response
-- update the `form` property on a invalid response
+- update the `form` property if the action belongs to the page and invalidate all data on a successful response
+- update the `form` property on a invalid response if the action belongs to the page
```
benmccann, line 253: I think that including the warning below is probably enough. I think it makes it harder to understand these bullet points to include more detail here
```suggestion
- update the `form` property and invalidate all data on a successful response
- update the `form` property on a invalid response
```

documentation/docs/06-form-actions.md
```diff
@@ -249,12 +249,14 @@ The easiest way to progressively enhance a form is to add the `use:enhance` acti
...
 - update `$page.status` on a successful or invalid response
 - call `goto` on a redirect response
 - render the nearest `+error` boundary if an error occurs
 
+> By default the `form` property is only updated for actions that belong to the page because in the native form submission case you would be redirected to the page the action is on
```
benmccann, line 258: Maybe we could be more specific about what "actions that belong to the page" means?
```suggestion
> By default the `form` property is only updated for actions that are in a `+page.server.js` alongside the `+page.svelte` because in the native form submission case you would be redirected to the page the action is on
```

documentation/docs/06-form-actions.md
dummdidumm: You're probably right - thought the same when writing it, you confirming it makes me confident removing it again

documentation/docs/06-form-actions.md
```diff
@@ -255,6 +255,8 @@ Without an argument, `use:enhance` will emulate the browser-native behaviour, ju
 - call `goto` on a redirect response
 - render the nearest `+error` boundary if an error occurs
 
+> By default the `form` property is only updated for actions that are in `+page.server.js` alongside `+page.svelte` because in the native form submission case you would be redirected to the page the action is on
```
benmccann, line 258: missing some articles
```suggestion
> By default the `form` property is only updated for actions that are in a `+page.server.js` alongside the `+page.svelte` because in the native form submission case you would be redirected to the page the action is on
```

benmccann (approved):

PR #6786, opened 2022-09-14
Clarify handleFetch only works on +page.js load

danawoodman:
Since this does not work with `+page.server` or other files in SvelteKit, it's best to clarify exactly where this works. 

This came up on Discord today: https://discord.com/channels/457912077277855764/1019398087061557360

### Tests

### Changesets

### DIFF
--- a/documentation/docs/07-hooks.md
+++ b/documentation/docs/07-hooks.md
@@ -92,9 +92,9 @@ export async function handle({ event, resolve }) {
 
 #### handleFetch
 
-This function allows you to modify (or replace) a `fetch` request that happens inside a `load` function that runs on the server (or during pre-rendering).
+This function allows you to modify (or replace) a `fetch` request that happens inside a `+page.js` `load` function that only runs on the server (or during pre-rendering).
 
-Or your `load` function might make a request to a public URL like `https://api.yourapp.com` when the user performs a client-side navigation to the respective page, but during SSR it might make sense to hit the API directly (bypassing whatever proxies and load balancers sit between it and the public internet).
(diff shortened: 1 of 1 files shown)


documentation/docs/07-hooks.md
```diff
@@ -92,9 +92,9 @@ export async function handle({ event, resolve }) {
 
 #### handleFetch
 
-This function allows you to modify (or replace) a `fetch` request that happens inside a `load` function that runs on the server (or during pre-rendering).
+This function allows you to modify (or replace) a `fetch` request that happens inside a `+page.js` `load` function that only runs on the server (or during pre-rendering).
```
Conduitry, line 95: I don't think "that only runs on the server" is the right thing to say here. It specifically does _not_ affect `+page.server.js` (which only runs on the server), at least until we eventually start passing that its own `fetch` as well, which is planned eventually. It _does_ affect the `fetch` in `load`s that may run on the browser or the server, but only when it happens to be running on the server.

danawoodman: > It does affect the fetch in loads that may run on the browser or the server, but only when it happens to be running on the server.

So... it only run on the server in +page.js load? If so, how else should it be worded?

Rich-Harris: The current wording is correct — the 'only' implies that it only applies to `+page.server.js` files (since `load` in a `+page.js` runs both on the server and in the browser), and the addition of `+page.js` is only accurate because we haven't yet got round to adding it to the `RequestEvent` passed to `+page.server.js` and `+server.js`. The solution is to fix _that_, so I'll close this issue

Issue #6793, opened 2022-09-14
404 has the message "Internal Error"


kevinrenskers:
### Describe the bug

When you hit a route that doesn't exist, the error message is "Internal Error" instead of something like "Page not found". Pretty sure this used to be something like the latter though.

"Internal Error" is not something we should be showing to our users when it's just a simple 404 error.

### Reproduction

No custom code needed at all. Just create a new project `npm create svelte@latest sk-404-bug`, run it, and enter an unknown route in your browser.

### System Info

```shell
System:
    OS: macOS 12.6
    CPU: (10) arm64 Apple M1 Max
    Memory: 1.21 GB / 32.00 GB
    Shell: 3.5.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 16.17.0 - ~/Library/pnpm/node
    npm: 8.15.0 - ~/Library/pnpm/npm
  Browsers:
    Safari: 16.0
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.73
    @sveltejs/kit: next => 1.0.0-next.481
    svelte: ^3.44.0 => 3.50.1
    vite: ^3.1.0 => 3.1.0
```

### Severity

annoyance

dummdidumm:
This is the expected behavior. All unexpected errors (and 404 is such an unexpected one) are shown as "Internal Error" by default. You need to implement the `handleError` hooks on the server and client and return the desired error shape from it. In your case it's probably enough to do

```js
// hooks.server.js
export function handleError({error}) {
  return { message: error.message };
}
```
```js
// hooks.client.js
export function handleError({error}) {
  return { message: error.message };
}
```

It's done this way to protect your app from accidentally showing sensitive data that could appear from an unexpected error.

The docs have more info on this: https://kit.svelte.dev/docs/hooks#shared-hooks-handleerror

kevinrenskers:
It seems that the default behavior is kinda weird, forcing the normal happy path usage to implement two hooks or you get nonsensical error messages for even 404 pages. Why are those hooks not simply the default, and people can override it if needed?

dummdidumm:
Because your message could contain sensitive data like "an error in occurred using <secret api key>". Also, it's not ensured that the `error` passed in is actually an `Error` object - anything can be thrown.

kevinrenskers:
[...]

PR #6794, opened 2022-09-14
[docs] remove obsolete note

dummdidumm:
No longer true with new handleError

### Tests

### Changesets


Conduitry:
Hmmm. Should we instead change the default `handleError` behavior so that we can have this behavior? It sounds like it would be expected and useful. Is that practical?

dummdidumm:
Sounds somewhat related to #6793 where it was requested to change the behavior this in a slightly different way, too.

Rich-Harris (approved):

Issue #5849, opened 2022-08-08
Generate Production package.json when building


tylkomat:
### Describe the problem

It's not always clear which packages should go into dependencies and which ones into devDependencies. There are regular new questions about this topic. 

### Describe the proposed solution

The compiler has all the information about which packages can be inlined and which can't. It can create a minimal production package.json file which only inludes the necessary packages.

### Importance

would make my life easier

### Additional Information

I already implemented the changes in kit and adapter-node, but I'm waiting for the big update until doing the PR.

benmccann:
This sounds like a duplicate of https://github.com/sveltejs/kit/issues/3176

tylkomat:
@benmccann It's not the same. I don't propose bundling the dependencies as not all can be bundled. My proposal merely converts this (example from my current project):
```json
{
  ...
  "devDependencies": {
    "@felte/reporter-svelte": "1.1.2",
    "@felte/validator-vest": "1.0.9",
    "@playwright/test": "1.24.2",
    "@prisma/client": "4.1.1",
    "@sveltejs/adapter-node": "1.0.0-next.85",
    "@sveltejs/kit": "1.0.0-next.401",
    "@tailwindcss/forms": "0.5.2",
    "@tailwindcss/typography": "0.5.4",
    "@types/nodemailer": "6.4.4",
    "@typescript-eslint/eslint-plugin": "5.31.0",
    "@typescript-eslint/parser": "5.31.0",
    "autoprefixer": "10.4.8",
    "eslint": "8.21.0",
    "eslint-config-prettier": "8.5.0",
    "eslint-plugin-svelte3": "4.0.0",
    "felte": "1.2.2",
    "form-data": "4.0.0",
    "got": "12.3.0",
    "mdsvex": "0.10.6",
    "nodemailer": "6.7.7",
    "paseto": "3.1.0",
    "postcss": "8.4.14",
    "postcss-load-config": "4.0.1",
    "prettier": "2.7.1",
    "prettier-plugin-svelte": "2.7.0",
    "promise-retry": "2.0.1",
    "soap": "0.45.0",
    "svelte": "3.49.0",
    "svelte-check": "2.8.0",
    "svelte-preprocess": "4.10.7",
    "tailwindcss": "3.1.7",
    "tslib": "2.4.0",
    "typescript": "4.7.4",
    "vest": "4.5.0",
    "vite": "3.0.4"
  },
  ...
}
```
to this:
```json
{
  ...
  "dependencies": {
[...]

PR #6797, opened 2022-09-14
Make build folder of adapter-node self-sufficient

tylkomat:
Currently you can't just take the content of the build folder to run the server in production.
The server won't start without an existing `package.json` file which is configured with `type: "module"`.

This PR creates a `package.json` file in the build folder.

To avoid errors after starting the server because if dependencies which should be in `dependencies` instead of `devDependencies` of `package,json`, this PR also extends the build process to create a production ready `package.json` file.

**Can't you just use the project's package.json and lockfile? Why do you need to generate a new one?**
I believe the compiler is smarter than I am, so I let it decide which packages should be dependencies in a production package.json.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-node/index.js
+++ b/packages/adapter-node/index.js
@@ -1,4 +1,4 @@
-import { readFileSync, writeFileSync } from 'fs';
+import { writeFileSync } from 'fs';
 import { fileURLToPath } from 'url';
 import * as esbuild from 'esbuild';
 
@@ -45,13 +45,13 @@ export default function (opts = {}) {
 				`export const manifest = ${builder.generateManifest({ relativePath: './' })};`
 			);
 
-			const optimizedPackageJson = builder.generatePackageJson()
+			const optimized_package_json = builder.generatePackageJson();
 			writeFileSync(
 				`${out}/package.json`,
-				optimizedPackageJson,
+				optimized_package_json
 			);
 
-			const pkg = JSON.parse(optimizedPackageJson);
+			const pkg = JSON.parse(optimized_package_json);
 
 			await esbuild.build({
 				platform: 'node',
--- a/packages/adapter-node/index.js
+++ b/packages/adapter-node/index.js
@@ -46,10 +46,7 @@ export default function (opts = {}) {
 			);
 
 			const optimized_package_json = builder.generatePackageJson();
-			writeFileSync(
-				`${out}/package.json`,
-				optimized_package_json
-			);
+			writeFileSync(`${out}/package.json`, optimized_package_json);
 
 			const pkg = JSON.parse(optimized_package_json);
 
--- a/packages/adapter-node/index.js
+++ b/packages/adapter-node/index.js
@@ -45,7 +45,13 @@ export default function (opts = {}) {
 				`export const manifest = ${builder.generateManifest({ relativePath: './' })};`
 			);
 
-			const pkg = JSON.parse(readFileSync('package.json', 'utf8'));
+			const optimizedPackageJson = builder.generatePackageJson()
+			writeFileSync(
+				`${out}/package.json`,
+				optimizedPackageJson,
+			);
+
+			const pkg = JSON.parse(optimizedPackageJson);
 
 			await esbuild.build({
 				platform: 'node',
--- a/packages/kit/src/core/adapt/builder.js
+++ b/packages/kit/src/core/adapt/builder.js
@@ -118,7 +118,7 @@ export function create_builder({ config, build_data, routes, prerendered, log })
 			return generate_package_json({
 				build_data,
 				cwd: process.cwd()
-			})
+			});
 		},
 
 		getBuildDirectory(name) {
--- a/packages/kit/src/core/adapt/builder.js
+++ b/packages/kit/src/core/adapt/builder.js
@@ -5,6 +5,7 @@ import { pipeline } from 'stream';
 import { promisify } from 'util';
 import { copy, rimraf, mkdirp } from '../../utils/filesystem.js';
 import { generate_manifest } from '../generate_manifest/index.js';
+import { generate_package_json } from '../generate_package_json/index.js';
 
 const pipe = promisify(pipeline);
 
@@ -111,6 +112,13 @@ export function create_builder({ config, build_data, routes, prerendered, log })
 				routes,
 				format
 			});
+		},
+
+		generatePackageJson: () => {
+			return generate_package_json({
+				build_data,
+				cwd: process.cwd()
+			})
 		},
 
 		getBuildDirectory(name) {
--- a/packages/kit/src/core/generate_package_json/index.js
+++ b/packages/kit/src/core/generate_package_json/index.js
@@ -45,7 +45,7 @@ export function generate_package_json({build_data, cwd}) {
  */
 function filter_import_packages(chunks, dependencies_set) {
   /** @type {string[]} */
-  const import_names = chunks.reduce((deps, chunk) => {
+  const import_names = chunks.reduce((/** @type {string[]} */ deps, chunk) => {
     if(chunk.imports.length > 0) {
       for(let i = 0; i < chunk.imports.length; i++) {
         const importName = chunk.imports[i];
--- a/packages/kit/src/core/generate_package_json/index.js
+++ b/packages/kit/src/core/generate_package_json/index.js
@@ -1,0 +1,77 @@
+import fs from 'node:fs';
+import path from 'node:path';
+
+/**
+ * Generates a production package.json
+ * @param {{
+ *   build_data: import('types').BuildData;
+ *   cwd: string;
+ * }} opts
+ */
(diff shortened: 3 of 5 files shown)


Rich-Harris:
Thank you. This is very `adapter-node` specific, so I don't think we need to add a new `generatePackageJson` API, the logic should be contained within that adapter.

Having said that, this makes me nervous. Putting a generated `package.json` in the output folder without a lockfile means that the production build will end up using different versions of packages, which is a non-starter. But you can't just copy the lockfile, because it won't match the generated `package.json`.

I think it's better if people are responsible for including/excluding files themselves.

tylkomat:
One option could be to fix the currently installed versions in the generated package.json. Then there are no surprises. It will be the same versions installed in production that you had when you developed it.

Otherwise I would argue, that just copying the package.json from the project is the same as letting people include/exclude files themselves, since this is what most of them (from my experiences on the discord) do. They copy the package.json and then get surprised if everything explodes on the next update. At least it is consistent.

benmccann:
Can't you just use the project's `package.json` and lockfile? Why do you need to generate a new one? (The answers to questions like these would be good to include in the PR description and code comments)

tylkomat:
I believe the compiler is smarter than I am, so I let it decide which packages should be dependencies in a production `package.json`.

Rich-Harris:
> One option could be to fix the currently installed versions in the generated package.json. Then there are no surprises. It will be the same versions installed in production that you had when you developed it.

This works fine for your immediate dependencies. It doesn't work for _their_ dependencies — that's the problem lockfiles solve.

> Can't you just use the project's `package.json` and lockfile?

This seems like a good approach to me. I'm not sure if there are any drawbacks — obviously the file will include stuff that doesn't make any sense (like a `"dev"` script, etc) but if it's only being used to determine which dependencies to install then that's probably fine.

> I believe the compiler is smarter than I am, so I let it decide which packages should be dependencies in a production `package.json`.

You'd be surprised! Giving users the ability to control which dependencies are included in a production build (by putting them in `pkg.dependencies` instead of `pkg.devDependencies`) is easier to understand, more consistent with the logic that determines what stuff gets bundled, and offers an escape hatch for when someone inevitably hits a bizarre edge case.

Conduitry:
Using the same package.json and lockfile sounds like the right path to me. Sure it'll have stuff in it you don't need, but those will be just lines in a text file. You should be installing dependencies with `npm ci --prod` when you're preparing your prod build. If you're using Docker, for example, you should be using a multi stage build so that you install all dependencies to build the app and then just install prod dependencies in the final stage and copy over the contents of `build`.

tylkomat:
> > One option could be to fix the currently installed versions in the generated package.json. Then there are no surprises. It will be the same versions installed in production that you had when you developed it.
> 
> This works fine for your immediate dependencies. It doesn't work for _their_ dependencies — that's the problem lockfiles solve.
> 

true

> > Can't you just use the project's `package.json` and lockfile?
> 
> This seems like a good approach to me. I'm not sure if there are any drawbacks — obviously the file will include stuff that doesn't make any sense (like a `"dev"` script, etc) but if it's only being used to determine which dependencies to install then that's probably fine.
> 
> > I believe the compiler is smarter than I am, so I let it decide which packages should be dependencies in a production `package.json`.
> 
> You'd be surprised! Giving users the ability to control which dependencies are included in a production build (by putting them in `pkg.dependencies` instead of `pkg.devDependencies`) is easier to understand, more consistent with the logic that determines what stuff gets bundled, and offers an escape hatch for when someone inevitably hits a bizarre edge case.

If you don't know how to decide which goes where you build your app and try to run it. You hit the first missing dependency, you update package.json, then try again. Rinse and repeat. In my project I had 7 such dependencies. Then you copy everything on the server and try to run it. Then you hit the missing package.json issue. This is very frustrating.

How about making it optional via adapter configuaration parameter.
Another approach could be svelte-kit sync or svelte-check to hint that some packages should be moved.

tylkomat:
The reason why I generated the `package.json` file comes from the question which packages go where? Is a specific package a devDependency or is it a prod dependency? How do you decide? This question was asked a lot of times on discord. If you don't know you may put everything in devDependencies and sort it out when you want to deploy. 

If you don't know what to look for after building you may end up in the rinse and repeat cycle described earlier. Since you put everything into devDependencies (or you missed something in dependencies), some dependencies may not have been bundled (I also did an issue earlier [#5849](https://github.com/sveltejs/kit/issues/5849)) . 

If you know what to look for you can build once and then search for imports and find those which don't import from a relative path. Relative path import means the dependency has been bundled. Some dependencies can't be bundled so they are imported from node_modules.

Again, how do you decide which package is a prod dependency? One Answer was: "When you need it in production it is a prod dependency.". May work, but may not be completely true. You may not miss any packages, but you may add more packages into prod dependencies then necessary. Which is fine I guess.

From my observations everything that is not treeshakable is not bundled. So if it's esm, it is usually bundled.

But why should I spent my time on figuring that out? The compiler does the bundling, so it knows what can be bundled and what can't. I let it decide and generate the package.json accordingly.

Rich-Harris:
I'm sorry, I still don't follow. The logic is very simple: if something is a prod dependency, it will be excluded from the generated bundle, and must be installed alongside your deployment: https://github.com/sveltejs/kit/blob/9dbcce994161ec7191a8a8ea5a510437232dd1ce/packages/adapter-node/index.js#L52-L62

This is described in the README: https://github.com/sveltejs/kit/tree/master/packages/adapter-node#deploying

> some dependencies may not have been bundled

In what circumstance? If that's happening, it's a very strange bug and it needs a reproduction.

tylkomat:
The problem I'm trying to solve starts earlier. 
Let me give you an example. These are your dependencies. Which one is a prod dependency and which one is devDependency?

```json
"@felte/reporter-svelte": "1.1.2",
"@felte/validator-vest": "1.0.9",
"@playwright/test": "1.25.0",
"@prisma/client": "4.2.1",
"@sveltejs/adapter-node": "1.0.0-next.89",
"@sveltejs/kit": "1.0.0-next.480",
"@tailwindcss/forms": "0.5.2",
"@tailwindcss/typography": "0.5.4",
"@types/nodemailer": "6.4.5",
"@typescript-eslint/eslint-plugin": "5.33.0",
"@typescript-eslint/parser": "5.33.0",
"autoprefixer": "10.4.8",
"eslint": "8.21.0",
"eslint-config-prettier": "8.5.0",
"eslint-plugin-svelte3": "4.0.0",
"felte": "1.2.2",
"form-data": "4.0.0",
"got": "12.3.1",
"mdsvex": "0.10.6",
"nodemailer": "6.7.7",
"paseto": "3.1.0",
"postcss": "8.4.16",
"postcss-load-config": "4.0.1",
"prettier": "2.7.1",
"prettier-plugin-svelte": "2.7.0",
"promise-retry": "2.0.1",
"soap": "0.45.0",
"svelte": "3.49.0",
"svelte-check": "2.8.0",
"svelte-preprocess": "4.10.7",
"tailwindcss": "3.1.8",
"tslib": "2.4.0",
"typescript": "4.7.4",
"vest": "4.6.2",
"vite": "3.1.0"
```

benmccann:
@tylkomat just make everything a dev dependency and be done with it

tylkomat:
@benmccann This is exactly what I'm trying to explain. If you do that you'll end up in the rinse and repeat cycle I described earlier. At least you did before #6301 was merged. Nonetheless you'll end up with 7 dependencies that are not bundled.

I created a stripped down example of the problem [here](https://github.com/tylkomat/sveltekit-no-bundling), which just includes 3 of the dependencies.
It works fine in dev mode. When you build it the server starts fine.

Before #6301 the server wouldn't start because dependencies where missing, I also can provide an example for that. 
Now the server starts fine, but when you access the page it explodes, which is even worse since you now expect everything to be fine. Also the error doesn't help you much to identify what the issue could be.

The server is build as a sibling to the source folder to simulate copying the content of the build folder to a server.
In a second step you have to copy the package.json to the build folder (`../no-bundling-prod`), but **don't** run npm install, since we're expecting the dependencies to be bundled.

Rich-Harris:
The issue here isn't that dependencies aren't being bundled, it's that dependencies that _shouldn't_ be bundled (because they're not compatible with `esbuild` in `esm` mode) _are_ being bundled. Ultimately this is a bug in `esbuild` (https://github.com/evanw/esbuild/issues/1921) — there's no reason why `require('util')` should be turned into `__require('util')`, given that it's a built-in module — but it's something we have to deal with for now.

This PR (if we ignore the complexity around lockfiles etc) would change that by making everything external by default. But that's undesirable since it increases the size of deployments.

I think the best solution might be to work around https://github.com/evanw/esbuild/issues/1921, perhaps by generating CommonJS output instead of ESM. I've opened a PR that implements that idea: #6855

tylkomat:
Thanks @Rich-Harris. Looks like #6896 solved it.

Issue #6477, opened 2022-08-31
+page.server.js no longer uses/respects setHeaders


homerjam:
### Describe the bug

Since [this commit](https://github.com/sveltejs/kit/commit/c530d33793a228fac684c71ed7926e6217101a90) the `+page.server.js` endpoints return a js file rather than json. This js file response no longer uses the headers set with the `setHeaders` function. This means we can't use cache-control etc.

### Reproduction

To reproduce, hover over the 'TODOS' menu item and observer the '__data.js' response headers. There should be a header as defined in `routes/todos/+page.server.js`;

https://stackblitz.com/edit/sveltejs-kit-template-default-8tt4be?file=src/routes/todos/+page.server.js

<img width="1224" alt="Screenshot 2022-08-31 at 17 29 32" src="https://user-images.githubusercontent.com/1055769/187730727-2e8f5dd4-97df-4661-a24e-5d56ec9b95c0.png">

### System Info

```shell
System:
    OS: macOS 12.5
    CPU: (8) arm64 Apple M2
    Memory: 155.45 MB / 24.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm
  Browsers:
    Chrome: 104.0.5112.101
    Safari: 15.6
  npmPackages:
    @sveltejs/adapter-auto: 1.0.0-next.70 => 1.0.0-next.70 
    @sveltejs/kit: 1.0.0-next.456 => 1.0.0-next.456 
    svelte: ^3.49.0 => 3.49.0 
    vite: 3.1.0-beta.1 => 3.1.0-beta.1
```

### Severity

annoyance

TranscendentalRide:
I noticed this in particular when trying to overwrite some headers with `throw redirect(301, '/todos')`. In particular want the redirect to look like it is external not internal; no luck.

example

```ts
setHeaders({
    'sec-fetch-mode': 'navigate',
    'sec-fetch-dest': 'document',
})
...

throw redirect(301,'/todos')
```

Doesn't work, actual headers

```
    'sec-fetch-mode' => 'cors',
    'sec-fetch-dest' => 'script'
```

dummdidumm:
[...]

PR #6809, opened 2022-09-14
[fix] correct headers for __data.js requests

### LINKED ISSUE #6458: Cache problem with client-side navigation: the response is now __data.js instead of __data.json
### Describe the bug

I believe the recent change where client-side navigation responds with __data.js instead of __data.json can be problematic.

I have Cloudflare on my site and I just realized that the __data.js files that are returned when navigating between pages are being cached on the edge and possibly also in the browser cache. I don't have any custom page rules for .js files in Cloudflare, this is the default behavior.

In some cases, the "age" in the response headers is showing very high numbers, much higher than what I have in my s-max-age cache-control headers in hooks.js (which used to cache on Vercel edge only, not Cloudflare).

681e4b3be00bfa462433bc8450004e6b

This means that many users are potentially seeing very old and outdated info for pages that are supposed to be highly dynamic.

### Reproduction

I don't think I can set up a reproduction since it would require a custom domain and adding it to Cloudflare.

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.22000
    Memory: 50.08 GB / 63.91 GB
  Binaries:
    npm: 8.11.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (104.0.1293.70)
  npmPackages:
    @sveltejs/adapter-auto: 1.0.0-next.69 => 1.0.0-next.70
    @sveltejs/kit: 1.0.0-next.454 => 1.0.0-next.456
    svelte: ^3.49.0 => 3.49.0
    vite: 3.1.0-beta.1 => 3.1.0-beta.1
```

### Severity

Blocking an upgrade

### Additional Information

Update: I rolled back to 446 for now since the 447 update seems to also have introduced a bunch of weird issues with both Sentry and Plausible Analytics.

dummdidumm:
Add all headers except cache-control which is always no-store
Fixes #6458
Fixes (?) #6477 <- not sure how to best handle this, conflicting requirements here

### Tests

### Changesets


.changeset/cyan-humans-perform.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] correct headers for \_\_data.js requests
```
Rich-Harris, line 5:
```suggestion
[fix] correct headers for `__data.js` requests
```

Rich-Harris:
I'm not sure we want to add headers other than `set-cookie`? There's no way for the application to access them so they're useless at best (except `set-cookie` and `cache-control`, which have side-effects). 

Given that some platforms sometimes apparently apply default `cache-control` headers if none are provided, and given that the `__id` parameter is per-user, I think we need to always apply `private, no-store`. If we respect `max-age` and `s-maxage` it should probably be in the form of a `window.__sveltekit_data_ttl` value or similar, the same way we cache data inlined into the HTML response.

dummdidumm:
I'm not well-versed with all the possibilities of headers, but in  #6477 on comment asked for setting headers besides the cache-control - don't know if that's sensible.

homerjam:
> not rerunning them is ensured differently
So is there another way to ensure data requests are not re-run unnecessarily? Is `window.__sveltekit_data_ttl` documented somewhere?

Rich-Harris:
> on comment asked for setting headers besides the cache-control - don't know if that's sensible

Those `sec-fetch-mode` and `sec-fetch-dest` headers would have no effect

> Is `window.__sveltekit_data_ttl` documented somewhere?

No, it doesn't exist. If we do implement it, it'll be an internal detail, not something you'd need to be aware of

dummdidumm:
Closing as obsolete - the no-store part was implemented in another PR and the cache will need a different implementation.

Issue #6650, opened 2022-09-07
`cookies` API not working in some cases


j4w8n:
### Describe the bug

When either of the below cookie removal methods are used, Firefox warns the `Cookie <name> has been rejected because it is already expired` and it never removes the cookie. Chrome never removes the cookie either.

The workaround is to use `cookies.set()`, in the `DELETE` method, and pass both `path` and `sameSite` options to reflect values of the existing cookie.
I include `path: '/',` when creating the cookie in `POST`, because without it, the path would be set to `/api`.

### Reproduction

https://github.com/j4w8n/cookie-bug

In a `api/cookie/+server.js` file:
1. create a cookie in a POST request using `cookies.set()`, including options `path` and `sameSite`
2. delete a cookie in a DELETE request using `cookies.delete()`, or `cookies.set()` without the `path` and `sameSite` options.

General fyi, without passing the `sameSite` option with `cookies.set()` in step 2, Firefox will warn that the cookie is misusing the recommended SameSite attribute. Not that it's crucial, because we're trying to delete it anyway.

### System Info

```shell
System:
    OS: macOS 12.5.1
    CPU: (4) x64 Intel(R) Core(TM) i5-5250U CPU @ 1.60GHz
    Memory: 31.02 MB / 4.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.0 - /usr/local/bin/node
    Yarn: 1.22.5 - /usr/local/bin/yarn
    npm: 8.5.5 - /usr/local/bin/npm
  Browsers:
    Chrome: 105.0.5195.102
    Firefox: 104.0.1
    Safari: 15.6.1
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.71 
    @sveltejs/kit: next => 1.0.0-next.476 
    svelte: ^3.44.0 => 3.50.0 
    vite: ^3.1.0 => 3.1.0
```

### Severity

serious, but I can work around it

cdcarson:
@j4w8n I'm working [on a fix](https://github.com/sveltejs/kit/pull/6648) that I hope will address the "missing" keys issues that you describe.  Right now, I have the following as defaults:

```js
{
  httpOnly: true,
  secure: true,
   path: '/',
   sameSite: 'lax',
   ...options // i.e. supplied to set or delete, will override the above
}
```
Does that look right to you? (It was what I was using before the cookies API, and seemed good.) Let me know. 

[...]

PR #6811, opened 2022-09-14
cookie stuff

Rich-Harris:
#6648, plus changes that for some reason I couldn't push to the PR


.changeset/cold-shirts-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+fix cookies API, add unit and e2e tests for cookies
```
Rich-Harris, line 5: no need to mention internal stuff like tests in changesets

```suggestion
[fix] various `cookies` fixes and improvements
```

packages/kit/src/runtime/server/cookie.js
```diff
@@ -1,49 +1,77 @@
-import * as cookie from 'cookie';
+import { parse } from 'cookie';
+
+/** @type {import('cookie').CookieSerializeOptions} */
+const DEFAULT_SERIALIZE_OPTIONS = {
+	httpOnly: true,
+	secure: true,
+	path: '/',
```
Rich-Harris, line 7: @cdcarson what's the rationale for having a default `path` value? For `httpOnly` and `secure` the defaults add security, and `sameSite` makes sense because it's already a default, and specifying it explicitly squelches console warnings.

But defaulting `path` to `/` makes things _less_ secure. Even though it's behaviour you often want, should it be the default?

cdcarson: Woops -- responded in the other PR: https://github.com/sveltejs/kit/pull/6648#issuecomment-1247237974. Pasted:

@Rich-Harris Pretty sure it is necessary. See https://github.com/sveltejs/kit/issues/6650. Basically a cookie set in a request to /sign-in will have that path. A subsequent request to /profile won't send that cookie.

The only downside of having the default is if the site base path is /foo. We can probably address this in the docs.

Rich-Harris: > As I understand it, we need the `cookies` API because cookies may be written and read in the _same request_, in different functions/files, especially with `actions`

We're not really any worse off because of `cookies`; the same is true if you add a `set-cookie` header any other way. The only difference is that we now have the _opportunity_ to set a default, and it's sorely tempting to take it.

To tl;dr a conversation among the maintainers just now: 👍 to removing the default path. Throwing an error feels heavy-handed and would be a breaking change, but if people _do_ end up getting confused by this stuff then a dev-time warning could be appropriate in future.

Agree re docs, will add something brief to this PR. We needed docs anyway because of the `httpOnly` default.

cdcarson: > We're not really any worse off because of cookies; the same is true if you add a set-cookie header any other way.

I guess I was harking back to the old days, doing my own form enhancement. I don't remember having to read a cookie after writing it -- but I was probably doing a whole bunch of goofy inefficient stuff instead.

Agree throwing would be bad. A warning might be nice -- it's an easy mistake.

PR #6813, opened 2022-09-14
[fix] beforeNavigate external link

### LINKED ISSUE #6726: The `beforeNavigate`/`afterNavigate` `type` is problematic
### Describe the bug

If you click on a link to an external URL on the page, or if you do a `goto` to an external URL, two successive `beforeNavigate` events will fire, the first one will have a `type` of `link`/`goto`, and the second one `unload`:

image

This behavior is strange, to say the least.
First of all, `beforeNavigate` firing twice for the same navigation action is... well, awkward and unexpected.
And if you have a navigation loader that you want to show only during client-side routing, you can't simply do this:

```js
beforeNavigation(({ type } => {
    if (type == 'unload') return;
    // Show loader
});
```
Because then the loader would also appear for external, native browser navigations.
The behavior is inconsistent with the "promise" @Rich-Harris made [here](https://github.com/sveltejs/kit/issues/4383#issuecomment-1072594489):

> Yep, type === 'unload' and !client would be equivalent. **All other types would be client-side navigations**

(emphasis mine)

This is now not the case, an instance of `beforeNavigate` firing with a `type` of something other than `unload` (like `link` in the example I provided above) COULD be a case of non-client-side navigation.

Instead, the API should've probably been designed such that we had `unload` as a separate boolean flag, alongside the `type` property, and then the event would only fire once when unloading would happen:
```js
beforeNavigate({ unload, type } => {
    if (unload) return; // e.g. If the user clicked on an external link, here `unload` is `true` and `type` is `link`.
    // Show loader
});
```

Please feel free to let me know if I'm missing something. Thank you.

### Reproduction

Explained above

### Logs

_No response_

### System Info

```shell
System:
    Memory: 461.70 MB / 3.77 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 16.15.1 - /usr/bin/node
    Yarn: 1.22.15 - /usr/bin/yarn
    npm: 8.11.0 - /usr/bin/npm
  npmPackages:
    @sveltejs/adapter-node: ^1.0.0-next.86 => 1.0.0-next.86 
    svelte: ^3.44.0 => 3.49.0 
    vite: ^3.1.0-beta.1 => 3.1.0
```

### Severity

serious, but i can work around it

### Additional Information

_No response_

dummdidumm:
Call beforeNavigate once with type unload on external navigation

Fixes #6726

Given the reasons in https://github.com/sveltejs/kit/issues/6726#issuecomment-1247106719 I think this is the most sensible thing to do

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -10,6 +10,7 @@ import { nodes, server_loads, dictionary, matchers, hooks } from '__GENERATED__/
 import { HttpError, Redirect } from '../control.js';
 import { stores } from './singletons.js';
 import { DATA_SUFFIX } from '../../constants.js';
+import { get } from 'svelte/store';
 
 const SCROLL_KEY = 'sveltekit:scroll';
 const INDEX_KEY = 'sveltekit:index';
@@ -72,7 +73,7 @@ export function create_client({ target, base, trailing_slash }) {
 	/** @type {Array<((url: URL) => boolean)>} */
 	const invalidated = [];
 
-	/** @type {{id: string | null, promise: Promise<import('./types').NavigationResult | undefined> | null}} */
+	/** @type {{id: string | null, promise: Promise<import('./types').NavigationResult> | null}} */
 	const load_cache = {
 		id: null,
 		promise: null
@@ -145,7 +146,12 @@ export function create_client({ target, base, trailing_slash }) {
 
 			invalidating = Promise.resolve().then(async () => {
 				const intent = get_navigation_intent(url, true);
-				await update(intent, url, []);
+				await update(
+					// We invalidate the current route, this can't be a native navigation
+					/** @type {import('./types').NavigationIntent | undefined} */ (intent),
+					url,
+					[]
+				);
 
 				invalidating = null;
 				force_invalidation = false;
@@ -679,7 +685,7 @@ export function create_client({ target, base, trailing_slash }) {
 
 	/**
 	 * @param {import('./types').NavigationIntent} intent
-	 * @returns {Promise<import('./types').NavigationResult | undefined>}
+	 * @returns {Promise<import('./types').NavigationResult>}
 	 */
 	async function load_route({ id, invalidating, url, params, route }) {
 		if (load_cache.id === id && load_cache.promise) {
@@ -842,6 +848,7 @@ export function create_client({ target, base, trailing_slash }) {
 					// if we get here, it's because the root `load` function failed,
 					// and we need to fall back to the server
 					await native_navigation(url);
+					// @ts-expect-error - this is unreachable
 					return;
 				}
 			} else {
@@ -938,9 +945,10 @@ export function create_client({ target, base, trailing_slash }) {
 	/**
 	 * @param {URL} url
 	 * @param {boolean} invalidating
+	 * @returns false when external url, undefined when no intent found, else the intent
 	 */
 	function get_navigation_intent(url, invalidating) {
-		if (is_external_url(url)) return;
+		if (is_external_url(url)) return false;
 
 		const path = decodeURI(url.pathname.slice(base.length) || '/');
 
@@ -994,6 +1002,13 @@ export function create_client({ target, base, trailing_slash }) {
 		let should_block = false;
 
 		const intent = get_navigation_intent(url, false);
+		if (intent === false) {
+			// Skip straight to native navigation, which calls the `beforeunload` handler
+			// which in turn calls beforeNavigate with the correct type.
+			accepted(); // needs to be called, else navigation is started twice on link click
+			await native_navigation(url);
+			return; // unnecessary, but TypeScript prefers it this way
+		}
 
 		/** @type {import('types').Navigation} */
 		const navigation = {
@@ -1220,19 +1235,23 @@ export function create_client({ target, base, trailing_slash }) {
 			addEventListener('beforeunload', (e) => {
 				let should_block = false;
 
-				/** @type {import('types').Navigation & { cancel: () => void }} */
-				const navigation = {
-					from: add_url_properties('from', {
-						params: current.params,
-						routeId: current.route?.id ?? null,
-						url: current.url
-					}),
-					to: null,
-					type: 'unload',
-					cancel: () => (should_block = true)
-				};
+				if (get(stores.navigating) === null) {
+					// If we're navigating, beforeNavigate was already called. If we end up in here during navigation,
+					// it's due to a 404 or another error, for which we don't want to call the hook again.
+					/** @type {import('types').Navigation & { cancel: () => void }} */
+					const navigation = {
+						from: add_url_properties('from', {
+							params: current.params,
+							routeId: current.route?.id ?? null,
+							url: current.url
+						}),
+						to: null,
+						type: 'unload',
+						cancel: () => (should_block = true)
+					};
 
-				callbacks.before_navigate.forEach((fn) => fn(navigation));
+					callbacks.before_navigate.forEach((fn) => fn(navigation));
+				}
 
 				if (should_block) {
 					e.preventDefault();
--- a/.changeset/khaki-schools-trade.md
+++ b/.changeset/khaki-schools-trade.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] call beforeNavigate once with type unload on external navigation
--- a/packages/kit/test/apps/basics/src/routes/before-navigate/prevent-navigation/+page.svelte
+++ b/packages/kit/test/apps/basics/src/routes/before-navigate/prevent-navigation/+page.svelte
@@ -2,12 +2,15 @@
 	import { beforeNavigate } from '$app/navigation';
 
 	let triggered = false;
-	beforeNavigate(({ cancel }) => {
+	let navigation_type;
+	beforeNavigate(({ cancel, type }) => {
 		triggered = true;
+		navigation_type = type;
 		cancel();
 	});
 </script>
 
 <h1>prevent navigation</h1>
 <a href="/before-navigate/a">a</a>
-<pre>{triggered}</pre>
+<a href="https://google.de">external</a>
+<pre>{triggered} {navigation_type}</pre>
--- a/packages/kit/test/apps/basics/test/client.test.js
+++ b/packages/kit/test/apps/basics/test/client.test.js
@@ -8,21 +8,21 @@ test.skip(({ javaScriptEnabled }) => !javaScriptEnabled);
 test.describe.configure({ mode: 'parallel' });
 
 test.describe('beforeNavigate', () => {
-	test('prevents navigation triggered by link click', async ({ clicknav, page, baseURL }) => {
+	test('prevents navigation triggered by link click', async ({ page, baseURL }) => {
 		await page.goto('/before-navigate/prevent-navigation');
 
 		await page.click('[href="/before-navigate/a"]');
 		await page.waitForLoadState('networkidle');
 
 		expect(page.url()).toBe(baseURL + '/before-navigate/prevent-navigation');
-		expect(await page.innerHTML('pre')).toBe('true');
+		expect(await page.innerHTML('pre')).toBe('true link');
 	});
 
 	test('prevents navigation triggered by goto', async ({ page, app, baseURL }) => {
 		await page.goto('/before-navigate/prevent-navigation');
 		await app.goto('/before-navigate/a');
 		expect(page.url()).toBe(baseURL + '/before-navigate/prevent-navigation');
-		expect(await page.innerHTML('pre')).toBe('true');
+		expect(await page.innerHTML('pre')).toBe('true goto');
 	});
 
 	test('prevents navigation triggered by back button', async ({ page, app, baseURL }) => {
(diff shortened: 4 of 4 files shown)


packages/kit/src/runtime/client/client.js
```diff
@@ -1220,19 +1235,23 @@ export function create_client({ target, base, trailing_slash }) {
...
-						url: current.url
-					}),
-					to: null,
-					type: 'unload',
-					cancel: () => (should_block = true)
-				};
+				if (get(stores.navigating) === null) {
+					// If we're navigating, beforeNavigate was already called. If we end up in here during navigation,
+					// it's due to a 404 or another error, for which we don't want to call the hook again.
+					/** @type {import('types').Navigation & { cancel: () => void }} */
```
Rich-Harris, line 1241: _Is_ this the behaviour we want? I can't decide. If you redirect during a navigation, `beforeNavigate` gets called for the initial navigation then again for each redirect (with `type === 'goto'`, which is arguably incorrect), and this would then be inconsistent with that. You could definitely make the case that we should only skip `beforeNavigate` for external link clicks

Rich-Harris:
Aside from https://github.com/sveltejs/kit/pull/6813#pullrequestreview-1108148449 this looks fine, though I can't shake the feeling that the code is trying to tell us we're thinking about some of this stuff in slightly the wrong way (and I don't mean this PR, I mean the code in general, this PR is just resurfacing it). E.g. things would probably be simpler if `update` had to be passed a `NavigationIntent`, and the logic for handling the SPA 404 case lived in `goto` or in `start.js`, since that's the only time it can happen.

Returning `NavigationIntent | false | undefined` from `get_navigation_intent` seems like it will confuse future us. I don't know that we need to distinguish between URLs that didn't match because they're external and URLs that didn't match because they didn't match — if we can treat them the same way we'll thank ourselves one day.

dummdidumm:
I tried to refactor the code around `update`/`navigate` but failed to come up with something better than the current state. Passing a `navigation_result` instead of a `intent` to `update` looks promising at first, but we can't be sure that `invalidate` won't need the same checks that `navigate` needs (theoretically there could be a redirect triggered by the invalidation, etc). 

Does this all hint that the types are not set correctly semantically? Should `type: "unload"` only be fired when the user closes the browser tab by hand and not during a navigation? Should a navigation to something external be `type: "link"/"goto"` and external can be determined by looking at `to`? `params/routeId` is `null` in this case, although that's also the case for 404, but that could be distinguished by looking at the `url` which would have a different host.

Rich-Harris:
> Should `type: "unload"` only be fired when the user closes the browser tab by hand and not during a navigation?

Ya know... that makes a lot of sense. I'm struggling to think of any reason not to do that. I think `to.url.origin !== location.origin` is easy enough to do that we don't need to add an additional `external` property

aradalvand:
This is cool and all but it means that if you only need to handle client-side navigations you'll have to do this:
```js
beforeNavigate(({ type, to }) => {
    if (type == 'unload' || to.url.origin != location.origin) return;
});
```
Both of these conditions are necessary and neither would be sufficient on its own.

Too verbose in my opinion, which is why I suggested `unload` (or `client` or whatever) as a separate boolean property, which would allow you to do this:
```ts
beforeNavigate(({ client }) => {
    if (client) return;
});
```

aradalvand:
In the case of the user manually closing the tab for instance:
- `type`: `manual` (not a great word but bikesheddable)
- `unload`: `true`
- `to`: `null` (basically means unknown)

In the case of the user clicking on an external link:
- `type`: `link`
- `unload`: `true`
- `to`: `https://google.com/...`

In the case of the user clicking on an internal 404 link or one that has the `data-sveltekit-reload` attribute set:
- `type`: `link`
- `unload`: `true`
- `to`: `https://your-site.com/...`

dummdidumm:
Right now external and "reload this" links don't have a `to` property, it's `null` (because `unload` is triggered, which we would need to change). Internal links that aren't found have a `to` property with `params` and `routeId` being `null`.
- If we say that any links that cause a full page reload have their `to` property set to `null`, the distinction doesn't need two checks, you could just do `if (to?.url.origin != location.origin` and adding another boolean would just introduce two ways to do the same.
- If we say that having the `url` set for full page reload links, too, then we would need an additional property because we can't distinguish whether or not the event will cause a full page reload: `type: "link", to: { params: null, routeId: null, url: "https://same-origin-you-are-currently-on.com/.. }`

dummdidumm:
I adjusted the code not call "unload" during navigation. The way to check if a client-side navigation is happening is to check for `to?.routeId != null`

aradalvand:
That's better, but don't you still think `to?.routeId != null` is a little unidiomatic?
I don't think it'd be clear to someone new to the framework that that's how they're supposed to handle client-side navigations, unless they know in detail how these events work.

PatrickG:
> ditch the `beforeNavigate` behavior for unload events entirely

My guess is that most people that want to `cancel()` a navigation in `beforeNavigation` also want/need the current behavior of preventing the navigation to an external link/closing the tab.
They would need to reimplement that in userland.

dummdidumm:
Resiviting this:

The beforeUnload stuff is a thing for another PR (if at all)

Having a dedicated `unload` property: I'm torn. It could be valuable to have `routeId` and `params` set for a link pointing to the app but with `data-sveltekit-reload` on it; currently they are both `null`. I also have to admit that `routeId` being `null` is not idiomatic at first (but one could also argue that's SvelteKit language you'll get used to, which also used in other places; for example a 404 is also visible through `routeId` being `null`).

Either way, if we think the `data-sveltekit-reload` case is worthwhile, we _need_ another property, at which point I'm split on whether it makes sense to have `unload` as a top level property besides `to`, `type` and `from`.
- Argument in favor: easier to use: `beforeNavigate({ unload }) => { if (!unload) { // handle only client-side navs here..` vs `beforeNavigate({ to }) => { if (!to || to.unload) {..`
- Argument against: `unload` and `type: 'unload'` are somewhat duplicative

... which brings me to questioning my initial thought of using `type: 'link'` even if it's an external link or full-reload-link. One of the arguments of it was "the code is unreasonably hard writing it with `type: 'unload'`, which hints at a design flaw", which is no longer the case. This line https://github.com/sveltejs/kit/blob/9a8c95544c831edb3c5c103e5228b751f86d7cc5/packages/kit/src/runtime/client/client.js#L1343 just needs to be changed to `type: 'unload'`

aradalvand:
> Argument against: unload and type: 'unload' are somewhat duplicative

I don't get this. If we have a separate `unload` flag then `type` would never be `unload`, it would be `link`, `goto`, etc.

dummdidumm:
We have `type: 'unload'` for the case where a user closes the browser window - so, no navigation, just `unload`. So we can't remove that type, because it's its own category/case.

aradalvand:
> We have type: 'unload' for the case where a user closes the browser window - so, no navigation, just unload. So we can't remove that type, because it's its own category/case.

Well, then just don't call that type `unload` to avoid cofusion?
It should be `type`: `browser` (or whatever) and `unload`: `true` in that scenario, as I suggested [here](https://github.com/sveltejs/kit/pull/6813#:~:text=The%20user%20manually%20closes%20the%20tab%20or%20types%20something%20else%20in%20the%20search%20bar%20and%20hits%20enter%3A).

dummdidumm:
The name of the type is not the point, the point is that `type: 'unload'` has the same semantic meaning as a possible `unload` boolean, so that's duplicate information.

aradalvand:
I don't see how that's a problem.
The two values just happen to overlap in one case.
This sort of thing happens all the time.

dummdidumm:
After discussing this we decided to keep the types as-is and add `unload` as a top-level boolean. @Rich-Harris this is ready to review.

aradalvand:
Nice to hear, thank you.
But wouldn't the value `unload` for `type` be confusing when there's a boolean flag with the same name, as we discussed?
Are you keeping it just to avoid a breaking change?

dummdidumm:
As explained above, `type: 'unload'` is different kind of navigation, it's called when for example you close the browser. Having this part of beforeNavigate is nice for things like "you have unsaved changes"-dialogs.

aradalvand:
I'm not arguing for its removal, just changing the word `unload` to something else to avoid unnecessary confusion (keeping the same underlying semantics).

Although it's not a big deal.

Rich-Harris:
`manual` isn't a great alternative though. Almost all navigation is the result of some manual action! 

One option we considered is `close`, which is appropriate in the case that it's being called as the result of someone closing a tab or the whole browser. It makes less sense in the case where someone is traversing the history stack across a document boundary.

It's definitely bikesheddable, but so far I'm not sure there's a better option than `type === 'unload'`. All suggestions welcome

aradalvand:
> manual isn't a great alternative though. Almost all navigation is the result of some manual action!

Fair enough! I agree.

> One option we considered is close, which is appropriate in the case that it's being called as the result of someone closing a tab or the whole browser. It makes less sense in the case where someone is traversing the history stack across a document boundary.
It's definitely bikesheddable, but so far I'm not sure there's a better option than type === 'unload'. All suggestions welcome

How about `leave`? It's similar to `close` but it's a bit more generic in terms of meaning and thus doesn't have that issue.
With its counterpart for `afterNavigate` being `enter`, for instance.

Rich-Harris:
Yeah, `enter`/`leave` is a good suggestion. I like that. In a similar vein `start`/`end` could also work, though `enter` arguably makes more sense than `start` if `afterNavigate` is fired after you click the back button from a different document.

One more piece of bikeshedding: should the boolean be `unload` or `willUnload`? `navigation.willUnload` feels slightly nicer to me than `navigation.unload`, now that I think about it.

dummdidumm:
I like both suggestions, updated the PR

aradalvand:
> Yeah, enter/leave is a good suggestion. I like that. In a similar vein start/end could also work, though enter arguably makes more sense than start if afterNavigate is fired after you click the back button from a different document.

I'm fine with both, although `enter`/`leave` feels slightly more intuitive to me, than `start`/`end`.

> One more piece of bikeshedding: should the boolean be unload or willUnload? navigation.willUnload feels slightly nicer to me than navigation.unload, now that I think about it.

If this one's called `willUnload` then will its counterpart in `afterNavigate` be called `hasLoaded` or something?

aradalvand:
I might be getting into "too much bikeshedding" territory but wouldn't you consider renaming the `type` property to `cause`? And then have one type called `browser` or something for indicating navigations where the cause is some native browser action (closing the tab, etc.)

It seems redundant to have two values (`enter` and `leave`) to describe the same kind of thing essentially.

dummdidumm:
- `willUnload` only makes sense in the scope of `beforeNavigate` and `$navigating`, ideally the types would represent that and `willUnload` is not part of `afterNavigate`. `csr` sounds appealing at first because of the option that has the same name, but is closing the tab which triggers `type: 'leave'` some routing? I mean, it's not, so it's `false`, but the naming somewhat suggests that the opposite would be a full page reload navigation. `willUnload` therefore describes better to me that your app state is about to be destroyed
- having `hasLoaded` doesn't make sense to me, it would always be `true`, so this property has no meaning. The "ok app started up" is part of `afterNavigate` with type `enter`.
- I like `trigger`, but I'm not sure if it's _that_ much better than `type` to justify this kind of breaking change. On the other hand, it could make the other renames more clear, since it's easy to detect that.
- `enter` vs `leave` and unifying them: `enter` is only available for `afterNavigate` and `leave` is only available for `beforeNavigate`. I think semantically they much more describe what is about to happen, so I feel like we should keep them. But the types could probably be tweaked.

Rich-Harris:
I'm in favour of the current changes — given that `enter` only happens with `afterNavigate` and `leave` only happens with `beforeNavigate`, I'd argue it's much less confusing than having the same thing in both cases. It can be made into a type error that would guard against this sort of sleep-deprivation-induced code, for example:

```js
beforeNavigate((navigation) => {
  if (navigation.type === 'enter') {
    // this doesn't make any sense!
  }
});

afterNavigate((navigation) => {
  if (navigation.type === 'leave') {
    // nor does this
  }
});
```

If the type was `browser` instead I can definitely imagine myself writing something like that and being confused about why it didn't work, since my brain equates 'before' and 'after' with 'earlier' and 'later', whereas to capture the start and end of the session you need to start with 'after' and end with 'before'.

Agree with @dummdidumm about `hasLoaded`. The point of having a separate `willUnload` boolean is that it's orthogonal to the navigation type, whereas an equivalent-but-opposite boolean only applies to a single navigation type, `enter`.

packages/kit/types/index.d.ts
```diff
@@ -384,12 +384,40 @@ export interface NavigationTarget {
 	url: URL;
 }
 
-export type NavigationType = 'load' | 'unload' | 'link' | 'goto' | 'popstate';
+/**
+ * - `enter`: Initial navigation is complete
+ * - `leave`: The page is about to be unloaded, but not due to a navigation
+ * - `link`: The page is about to change due to a navigation link. If it causes the site to unload can be determined through `external`
+ * - `goto`: The page is about to change due to calling `goto` or due to a redirect. If it causes a page unload can be determined through `external`
+ * - `popstate`: The page is about to change due to a browser back/forward navigation
```
Rich-Harris, line 392: I think we can finesse some of the wording — e.g. 'about to change' only applies to `beforeNavigate`, not `afterNavigate`. I think it's better if we describe the navigation itself

```suggestion
 * - `enter`: The app has hydrated
 * - `leave`: The user is leaving the app by closing the tab or using the back/forward buttons to go to a different document
 * - `link`: Navigation was triggered by a link click
 * - `goto`: Navigation was triggered by a `goto(...)` call or a redirect
 * - `popstate`: Navigation was triggered by back/forward navigation
```

packages/kit/types/index.d.ts
```diff
@@ -384,12 +384,40 @@ export interface NavigationTarget {
...
+	 * Where navigation is going to/has gone to
+	 */
 	to: NavigationTarget | null;
+	/**
+	 * The type of navigation:
+	 * - `enter`: Initial navigation is complete
+	 * - `leave`: The page is about to be unloaded, but not due to a navigation
+	 * - `link`: The page is about to change due to a navigation link. If it causes the site to unload can be determined through `external`
+	 * - `goto`: The page is about to change due to calling `goto` or due to a redirect. If it causes the site to unload can be determined through `external`
+	 * - `popstate`: The page is about to change due to a browser back/forward navigation
```
Rich-Harris, line 411:
```suggestion
	 * - `enter`: The app has hydrated
	 * - `leave`: The user is leaving the app by closing the tab or using the back/forward buttons to go to a different document
	 * - `link`: Navigation was triggered by a link click
	 * - `goto`: Navigation was triggered by a `goto(...)` call or a redirect
	 * - `popstate`: Navigation was triggered by back/forward navigation
```

Rich-Harris (approved):

Rich-Harris:
The types stuff would be a nice bonus but I'm not 100% sure what the best way to do that is, so I'll merge this and open a fresh issue

PR #6818, opened 2022-09-15
fix missing `opts` type for `cookies` api

s3812497:
Currently, the `cookies` api is missing the forwarded types `CookieSerializeOptions` and `CookieParseOptions` from the `@types/cookie` package, so we don't get any typescript autocompletion.

Example:
In a new SvelteKit project, the argument `opts` is missing its type:
CleanShot 2022-09-15 at 07 57 48@2x

However, after manually installing `@types/cookie` in a new project, the type appears:
CleanShot 2022-09-15 at 07 59 10@2x

~I was hoping that moving the types package to a dependency will ensure it gets installed automatically, but I have a feeling this might be the wrong way to fix the issue...~
Added it to peerDependencies as per @mquandalle 's advice.

### Tests

### Changesets


packages/kit/package.json
```diff
@@ -11,6 +11,7 @@
 	"type": "module",
 	"dependencies": {
 		"@sveltejs/vite-plugin-svelte": "^1.0.5",
+		"@types/cookie": "^0.5.1",
```
mquandalle, line 14: Maybe `peerDependencies`?

dummdidumm:
Feels strange to me to have `cookies` in dependencies but the types for it in peerDependencies. I think moving both to dependencies makes more sense.

Rich-Harris (approved):

Rich-Harris:
thank you!

PR #6825, opened 2022-09-15
`@sveltejs/package` Allow disabling the default behavior of copying package.json to the output directory when packaging a Svelte library

### LINKED ISSUE #6824: `@sveltejs/package` Allow disabling the default behavior of copying `package.json` to the output directory when packaging a Svelte library
### Describe the problem

For context, I'm building a cross-framework component library so I've set up a standard package structure for each framework in my monorepo with npm workspaces.

I'm using Vite with `@sveltejs/package` to develop and build my components. However, `@sveltejs/package` breaks my toolchain because it creates a copy of `package.json` in my `dist` folder.

npm workspaces picks up both the original `package.json` and `dist/package.json`, causing other dependent packages to break due to the duplication.

I rather keep my project structure and manage a single copy of `package.json` manually, so I'd like an option to disable the copy behavior.

This should ideally be added to `svelte.config.js` as a `copyPackageJson` option.

### Alternatives considered

I've tried running a script to delete `dist/package.json` manually after each build, but this frequently breaks when Nx builds multiple packages concurrently due to the race condition.

### Importance

would make my life easier

### Additional Information

_No response_

bryanmylee:

This resolves #6824.

For context, I'm building a cross-framework component library so I've set up a standard package structure for each framework in my monorepo with npm workspaces.

I'm using Vite with `@sveltejs/package` to develop and build my components. However, `@sveltejs/package` breaks my toolchain because it creates a copy of `package.json` in my `dist` folder.

npm workspaces picks up both the original `package.json` and `dist/package.json`, causing other dependent packages to break due to the duplication.

This PR adds a `copyPackageJson` option to allow disabling the copy feature.

### Tests

All existing tests for `@sveltejs/package` pass and I added a test for the new config option.

### Changesets


david-plugge:
The package command also copied the readme and changelog. What about a more general solution like an array of files and folders to copy or a filter function?

dummdidumm:
Yes that could be a more general solution. I'm tempted to put this into `files` but that's currently used for project files only, not sure if that breaking change is worth it (especially since it's not easy to detect) and if that's even a good API.

bryanmylee:
I didn't notice the README or changelog because it wasn't affecting the monorepo, but perhaps a more general `metadata` field would be better?

We can make its API similar to `files`, except we make the distinction that `files` only relates to project code, while `metadata` refers to the extra things like `README`, `package.json`, and the changelog.

bryanmylee:
Closed this in favor of #6864.

PR #6828, opened 2022-09-15
[fix] apply redirect/error logic for use:enhance

dummdidumm:
Came up in some issue that was closed but a comment in there made this good point that this isn't working as expected.

### Tests

### Changesets


packages/kit/src/runtime/app/forms.js
```diff
@@ -28,7 +28,13 @@ export function enhance(form, submit = () => {}) {
 			await invalidateAll();
 		}
 
-		if (location.origin + location.pathname === action.origin + action.pathname) {
+		// Don't apply action in case of success or invalid,
+		// because that updates the form prop which we don't want
```
Rich-Harris, line 32:
```suggestion
		// For success/invalid results, only apply action if it belongs to the,
		// current page, otherwise `form` will be updated erroneously
```

Rich-Harris (approved):

PR #6833, opened 2022-09-15
[fix] forward cookies from fetch on redirect response

### LINKED ISSUE #6792: Can't set or delete cookies if you also throw an error or redirect
### Describe the bug

When you call an internal endpoint to set or delete a cookie, this doesn't work if you also throw an error or redirect afterwards.

### Reproduction

https://github.com/kevinrenskers/sk-cookie-problem.

When you first open the homepage there is a button to login, it just sets a cookie. Then refresh the website. The cookie _should_ be deleted and then it should redirect, but instead the cookie isn't deleted and that causes an endless redirect loop.

When you remove the `throw`, the cookie _does_ get deleted. It seems that for some reason the cookie isn't deleted if you throw something as well. Same goes if you throw an error, not only when you throw a redirect.

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.6
    CPU: (10) arm64 Apple M1 Max
    Memory: 1.34 GB / 32.00 GB
    Shell: 3.5.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 16.17.0 - ~/Library/pnpm/node
    npm: 8.15.0 - ~/Library/pnpm/npm
  Browsers:
    Safari: 16.0
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.73
    svelte: ^3.44.0 => 3.50.1
    vite: ^3.1.0 => 3.1.0
```

### Severity

blocking an upgrade

### Additional Information

This is blocking me from handling 401 errors. What I need to do is remove the token cookie and redirect the user back to the safety of the homepage, but I can either delete the cookie or redirect, not both.

dummdidumm:
Fixes #6792

### Tests

### Changesets


dummdidumm:
Oh wow, the types lie to me, this isn't the same, we actually seem to use two different things to (de)serialize cookies; the one in `fetch.js` is different. Is there a reason for this?

packages/kit/src/runtime/server/index.js
```diff
@@ -246,12 +245,7 @@ export async function respond(request, options, state) {
 					}
 				}
 
-				for (const new_cookie of Array.from(new_cookies.values())) {
-					response.headers.append(
-						'set-cookie',
-						cookie.serialize(new_cookie.name, new_cookie.value, new_cookie.options)
-					);
-				}
+				add_cookies_to_headers(response.headers, Array.from(new_cookies.values()));
```
dummdidumm, line 248: To clarify my previous comment, this is the buggy code, and TypeScript is ok with it because extraneous properties are OK, and all required properties are the same between these types. The difference is that this shape has the options in the property `options` while in the other cases it's part of the same object that also has name/value.

dummdidumm:
I harmonized the types across the server, should be good now

Rich-Harris (approved):

PR #6843, opened 2022-09-15
Fixes #6842: Routes were not found by id under group layouts

### LINKED ISSUE #6842: Types are not properly updated for routes under group layouts
### Describe the bug

When altering return values from actions / load functions in +page.server.(ts/js) , the corresponding types are *automagically* updated while running `npm run dev`.

This experience breaks for routes under layout groups.

Under layout groups, one seems to be required to run `npx svelte-kit sync` and restart the language server if running vscode to get the latest types.

When moving such routes with said issue, away from a layout group, the wonderful DX resumes.

### Reproduction

This is a reproduction of the issue, taking the base demo-todo-app and putting all routes under a shared (groups) layout group.

Running the project and updating any return values from e.g. the (groups)/login/+page.server.js does not trigger an update in the generated types.

Moving the same route out of the (groups) layout group fixes the issue.

https://github.com/williamviktorsson/layout-groups-typing-bug

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.18363
    CPU: (4) x64 Intel(R) Core(TM) i5-6300U CPU @ 2.40GHz
    Memory: 2.03 GB / 7.92 GB
  Binaries:
    Node: 16.17.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.18.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 105.0.5195.102
    Edge: Spartan (44.18362.1533.0)
    Internet Explorer: 11.0.18362.1
  npmPackages:
    svelte: ^3.46.0 => 3.50.1
    vite: ^3.1.0 => 3.1.1
```

### Severity

### Additional Information

Layout groups reduce boilerplate code amongst other things, improving DX.

However, losing the great typing experience is a larger loss in DX.

williamviktorsson:
Fixes #6842 - Types are not properly updated for routes under group layouts

Routes were not found by id under group layouts. replacing '\' with '/' fixes this for route id's.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/sync/write_types/index.js
+++ b/packages/kit/src/core/sync/write_types/index.js
@@ -69,8 +69,8 @@ export async function write_types(config, manifest_data, file) {
 	}
 
 	const id = path.relative(config.kit.files.routes, path.dirname(file));
-
-	const route = manifest_data.routes.find((route) => route.id === id);
+	//replace '\' with '/' to fix #6842, routes not being found under layout groups
+	const route = manifest_data.routes.find((route) => route.id === id || route.id === id.replace(/\\/g, '/'));
 	if (!route) return; // this shouldn't ever happen
 
 	update_types(config, create_routes_map(manifest_data), route);
--- a/packages/kit/src/core/sync/write_types/index.js
+++ b/packages/kit/src/core/sync/write_types/index.js
@@ -69,7 +69,8 @@ export async function write_types(config, manifest_data, file) {
 	}
 
 	const id = path.posix.relative(config.kit.files.routes, path.dirname(file));
-	const route = manifest_data.routes.find((route) => route.id === id || route.id === id.replace(/\\/g, '/'));
+
+	const route = manifest_data.routes.find((route) => route.id === id);
 	if (!route) return; // this shouldn't ever happen
 
 	update_types(config, create_routes_map(manifest_data), route);
--- a/packages/kit/src/core/sync/write_types/index.js
+++ b/packages/kit/src/core/sync/write_types/index.js
@@ -68,7 +68,7 @@ export async function write_types(config, manifest_data, file) {
 		return;
 	}
 
-	const id = path.relative(config.kit.files.routes, path.dirname(file));
+	const id = path.posix.relative(config.kit.files.routes, path.dirname(file));
 	const route = manifest_data.routes.find((route) => route.id === id || route.id === id.replace(/\\/g, '/'));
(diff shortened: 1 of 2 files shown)


williamviktorsson:
Someone with more insight might tell if this is the best fix or if there needs to be some other update to the route id pattern matchers.

Conduitry:
I assume this is a Windows-only issue because of the `/` and the `\`? To be in line with how this sort of thing is being handled elsewhere in the codebase, we should be normalizing `id` with the `posixify()` function being used elsewhere, and then continue to compare it with the path route IDs as we are now.

williamviktorsson:
> I assume this is a Windows-only issue because of the `/` and the `\`? To be in line with how this sort of thing is being handled elsewhere in the codebase, we should be normalizing `id` with the `posixify()` function being used elsewhere, and then continue to compare it with the path route IDs as we are now.

Correct, just verified that the issue does not occur on my MacBook.

packages/kit/src/core/sync/write_types/index.js
```diff
@@ -69,8 +69,7 @@ export async function write_types(config, manifest_data, file) {
 	}
 
 	const id = path.relative(config.kit.files.routes, path.dirname(file));
```
Rich-Harris, line 71:
```suggestion
	const id = path.posix.relative(config.kit.files.routes, path.dirname(file));
```

packages/kit/src/core/sync/write_types/index.js
```diff
@@ -69,8 +69,7 @@ export async function write_types(config, manifest_data, file) {
...
-
-	const route = manifest_data.routes.find((route) => route.id === id);
+	const route = manifest_data.routes.find((route) => route.id === id || route.id === id.replace(/\\/g, '/'));
```
Rich-Harris, line 72:
```suggestion

	const route = manifest_data.routes.find((route) => route.id === id);
```

Rich-Harris (approved):

Rich-Harris:
thank you!

williamviktorsson:
> thank you!

Thank you! We’re a team of 25 students all on windows just getting started with SK the last few months so happy to keep enjoying this great DX 🎉👌

williamviktorsson:
Umh. Actually this doesn't work on windows.

Can we do this instead? It actually works (on windows).

```ts
const id = posixify(path.relative(config.kit.files.routes, path.dirname(file)));
```

PR #6855, opened 2022-09-16
create commonjs bundle with adapter-node, since esm output has issues

Rich-Harris:
This is an alternative approach to #6797. By generating CommonJS instead of ESM, `esbuild` doesn't replace `require` calls for built-in modules with `__require` (which throws a 'Dynamic require of ... is not supported' error — https://github.com/evanw/esbuild/issues/1921). 

As such, it's possible to bundle libraries that have such calls, instead of having to move them to `dependencies` so that they're not bundled, which is onerous since it's very hard to figure out which are the offending packages.

The downside of this approach is that it's no longer possible to have ESM packages as external dependencies — they must always be bundled. I don't have a clear sense of which is the lesser of two evils.

### Tests

### Changesets

### DIFF
--- a/.changeset/fast-crews-pay.md
+++ b/.changeset/fast-crews-pay.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-node': patch
+---
+
+Create CommonJS bundle with adapter-node
--- a/packages/adapter-node/ambient.d.ts
+++ b/packages/adapter-node/ambient.d.ts
@@ -1,16 +1,16 @@
-declare module 'ENV' {
+declare module '0ENV' {
 	export function env(key: string, fallback?: any): string;
 }
 
-declare module 'HANDLER' {
+declare module '0HANDLER' {
 	export const handler: import('polka').Middleware;
 }
 
-declare module 'MANIFEST' {
+declare module '0MANIFEST' {
 	import { SSRManifest } from '@sveltejs/kit';
 	export const manifest: SSRManifest;
 }
 
-declare module 'SERVER' {
+declare module '0SERVER' {
 	export { Server } from '@sveltejs/kit';
 }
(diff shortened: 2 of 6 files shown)


Conduitry:
I have the feeling you're not going to like it, but: Would it make sense for this to be configurable?

Also, what file extension are these being written out as? It looks like still `.js` which would be a problem if the project's `pkg.type` was `module`.

Rich-Harris:
Would definitely prefer to avoid configuration if we can, just because knowing when to use which is such an in-the-weeds problem. I guess the question is whether or not there are cases where an ESM package _couldn't_ be bundled. I suppose you could have an ESM package that depended on a native module like `sharp`. Blurgh.

I guess an alternative would be some version of https://github.com/evanw/esbuild/issues/1921#issuecomment-1010490128 — e.g. rewrite all the `__require` calls back to `require` and add this to the top of each module:

```js
const require = createRequire(import.meta.url);
```

Conduitry:
I'm less worried about adding potentially regrettable APIs to adapters than I am to SK core. If we can't/don't want to do something like the `__require`->`require` hack mentioned above, then I don't see another great option now besides allowing people to bundle to either CJS or to ESM.

Having a nasty API wart isn't as bad as telling some people they are just SOL unless they want to fork the adapter. If we find a satisfactory workaround for the esbuild problem (or if esbuild addresses this itself), then we can remove the option and release a new major version of the adapter.

benmccann:
My solution, which isn't an immediate one, is to fix Vite so that we no longer need esbuild involved. I filed https://github.com/vitejs/vite/issues/9919 about this and if Vite worked the way patak described then we could remove esbuild. I'm not sure if it should be considered a bug that might be able to be fixed in Vite 3 that it doesn't behave as expected, but there's a chance we can't fix it until Vite 4 a couple of months from now. Either way, I'll be pushing to get that fixed and drop esbuild so that we can have ESM and avoid all the esbuild bugs

Rich-Harris:
Alternative approach here: https://github.com/sveltejs/kit/pull/6896

Rich-Harris:
closing as #6896 was merged

PR #6864, opened 2022-09-17
`@sveltejs/package` Customize behavior of copying `package.json` and other essential files to the output directory when packaging a Svelte library

### LINKED ISSUE #6824: `@sveltejs/package` Allow disabling the default behavior of copying `package.json` to the output directory when packaging a Svelte library

bryanmylee:

This resolves #6824.

For context, I'm building a cross-framework component library so I've set up a standard package structure for each framework in my monorepo with npm workspaces.

I'm using Vite with `@sveltejs/package` to develop and build my components. However, `@sveltejs/package` breaks my toolchain because it creates a copy of `package.json` in my `dist` folder.

npm workspaces picks up both the original `package.json` and `dist/package.json`, causing other dependent packages to break due to the duplication.

This PR adds a `metadata` filter function to allow disabling the copy feature for metadata files like `package.json` and other whitelisted essential files.

### Tests

All existing tests for `@sveltejs/package` pass and I added a test for the new config option.

### Changesets


dominikg:
fyi: the test fail in create-svelte is unrelated to this issue and a fix for that is pending in #6866

dummdidumm:
Not 100% sure about the API yet but it's definitely going in the right direction. In the meantime, could you remove some of the unnecessary files from the tests? We don't need to have multiple Svelte files in there, copying them over and generating types for them is already checked in other tests. It should be enough to just have one `index.js` file.

bryanmylee:
Any updates on this? I would love to have this merged into the package itself.

bryanmylee:
The failing test seems to be unrelated to `svelte-package`

Rich-Harris:
Thank you for the PR. We ended up going in a different direction (#8825) — closing in favour of #8922

PR #6866, opened 2022-09-17
fix: prettier not working for svelte files in create-svelte projects

### LINKED ISSUE #6804: Prettier is ignoring svelte files
### Describe the bug

I created a new sveltekit app but when running the lint or format script prettier is ignoring all svelte files. It only works when explicitly telling prettier to check for svelte files with for example `src/**/*.{ts,svelte}` instead of `.`.

### Reproduction

- create a new sveltekit app with `pnpm create svelte@latest` and make sure to add prettier for code formatting
- install the packages with `pnpm install`
- run `pnpm format`

> You won't see any svelte files getting formatted

### Logs

_No response_

### System Info

```shell
System:
  OS: Linux 5.10 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
  CPU: (24) x64 AMD Ryzen 9 3900X 12-Core Processor
  Memory: 23.82 GB / 24.99 GB
  Container: Yes
  Shell: 5.8.1 - /usr/bin/zsh
Binaries:
  Node: 16.17.0 - ~/.local/share/pnpm/node
  npm: 8.15.0 - ~/.local/share/pnpm/npm
npmPackages:
  @sveltejs/adapter-node: next => 1.0.0-next.90
  svelte: ^3.50.1 => 3.50.1
  vite: ^3.1.0 => 3.1.0
```

### Severity

annoyance

### Additional Information

_No response_

dominikg:
fixes #6804 

i'm not sure why this is needed again, it was previously removed in favor of the `pluginSearchDirs` setting in prettierrc. 
Now we use both.

Note: the tests for create-svelte received 2 updates
1) pin them to the same vite version kit uses to avoid type incompatibility errors with `pnpm check` during tests
2) updated list of tested commands to include `format`, remove `prepare` and sort them differently

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/shared/+prettier/.prettierrc
+++ b/packages/create-svelte/shared/+prettier/.prettierrc
@@ -3,6 +3,7 @@
 	"singleQuote": true,
 	"trailingComma": "none",
 	"printWidth": 100,
+	"plugins": ["prettier-plugin-svelte"],
 	"pluginSearchDirs": ["."],
 	"overrides": [{ "files": "*.svelte", "options": { "parser": "svelte" } }]
 }
--- a/.changeset/light-eagles-call.md
+++ b/.changeset/light-eagles-call.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+fix prettier not formatting svelte files
--- a/.changeset/light-eagles-call.md
+++ b/.changeset/light-eagles-call.md
@@ -2,4 +2,4 @@
 'create-svelte': patch
 ---
 
-fix prettier not formatting svelte files
+[fix] prettier not formatting svelte files
--- a/packages/create-svelte/shared/+eslint+prettier/package.json
+++ b/packages/create-svelte/shared/+eslint+prettier/package.json
@@ -3,7 +3,7 @@
 		"eslint-config-prettier": "^8.3.0"
 	},
 	"scripts": {
-		"lint": "prettier --check . && eslint .",
-		"format": "prettier --write ."
+		"lint": "prettier --plugin-search-dir . --check . && eslint .",
+		"format": "prettier --plugin-search-dir . --write ."
 	}
 }
--- a/packages/create-svelte/shared/-eslint+prettier/package.json
+++ b/packages/create-svelte/shared/-eslint+prettier/package.json
@@ -1,6 +1,6 @@
 {
 	"scripts": {
-		"lint": "prettier --check .",
-		"format": "prettier --write ."
+		"lint": "prettier --plugin-search-dir . --check .",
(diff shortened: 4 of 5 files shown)


packages/create-svelte/shared/+prettier/.prettierrc
```diff
@@ -3,6 +3,7 @@
 	"singleQuote": true,
 	"trailingComma": "none",
 	"printWidth": 100,
+	"plugins": ["prettier-plugin-svelte"],
 	"pluginSearchDirs": ["."],
```
bluwy, line 7: Maybe we should do 
```suggestion
	"plugins": ["./node_modules/prettier-plugin-svelte"],
```
instead? I'm using this hack for all my projects which works well when running `prettier` through the CLI or VSCode extension. The `--plugin-search-dir .` can also be removed too.

I made a comment at https://github.com/sveltejs/prettier-plugin-svelte/issues/155#issuecomment-812876576

dominikg: would that work with yarn? also the search dir setting is likely needed if the users add more prettier plugins.

dummdidumm: better safe than sorry - let's keep it as is

bluwy: Hmm yeah my suggestion doesn't seem to work with yarn pnp (pnp also messes up vscode prettier plugin with this PR change and before). I guess we can go with this PR change still to have it partially working for yarn pnp.

> also the search dir setting is likely needed if the users add more prettier plugins.

Since they'd have to still add them in `plugins` with this PR, e.g. `"plugins": ["prettier-plugin-svelte", "prettier-plugin-something"]`, I don't think the `plugin-search-dir` would make a difference, besides not needing the `./node_modules/` part of my suggestion.

dominikg: i prefer not having `node_modules` listed explicitly. I'd love to get rid of the `--plugin-search-dir .` on the cli though so it's a tough one. Ultimately this PR uses multiple ways to try and ensure the plugin is found. The only other option i know would be to use `prettierrc.cjs` and `require('prettier-plugin-svelte')`  to take resolving out of prettiers own hands. But thats also ugly.

I think the way this PR uses both json rc and cli option to ensure the plugin is found during api and cli use is working and has only one duplication, which is unfortunate but not that bad.

dummdidumm (approved):

packages/create-svelte/shared/+prettier/.prettierrc
dummdidumm: Let's go with this solution then

PR #6879, opened 2022-09-19
fix: return 404 instead of 200 for stale assets with `adapter-cloudflare`

raduab:
Hello, this is my first PR, so please bear with me.

After experiencing numerous problems when deploying new code to Cloudflare Pages with stale assets and cache settings (e.g. when  Argo or Early Hints services were enabled), I discovered that requesting a non-existent asset from the `_app` directory always returned a status code of 200 and the contents of the static `404.html` file if it existed, or nothing at all if it didn't. This caused a lot of issues for our users because they had to hard refresh our web page after a new deployment in order to invalidate the wrong cached content in their clients due to some bizarre Cloudflare Edge issues—this was needed even after purging the Cloudflare's cache.

Instead of constructing a new response object and returning it as if the asset was there, I updated the Cloudflare adapter to return early if the asset was not found. I also included a check before caching the server response to avoid caching if the response was not ok.

### Tests

### Changesets


Rich-Harris (approved):

Rich-Harris:
thank you, and congrats on your first PR! 🎉

PR #6880, opened 2022-09-19
fix: #6458 add timestamp to force revalidating data

chientrm:

### Tests

### Changesets


dummdidumm:
Wouldn't `"cache-control": "private no-store"` on the response achieve the same without additional JS code on the client? That's what's done in #6809

chientrm:
> Wouldn't `"cache-control": "private no-store"` on the response achieve the same without additional JS code on the client? That's what's done in #6809

When you setup Custom Domain using Cloudflare Pages, the data response headers is somehow replaced with `cache-control: max-age=14400`. 🤦 

No issue when using default subdomain `*.pages.dev`.

Can't reproduce since it require a custom domain. 🤕🤕🤕

Rich-Harris:
That sounds like an extremely serious bug in Cloudflare Pages. If that's the case, this would be only one manifestation of it. I don't think it would be appropriate to work around it here. 

At the very least, it _would_ need a reproduction (and if one of the steps is 'set up a custom domain on Cloudflare Pages' then so be it)

chientrm:
> That sounds like an extremely serious bug in Cloudflare Pages. If that's the case, this would be only one manifestation of it. I don't think it would be appropriate to work around it here.
> 
> At the very least, it _would_ need a reproduction (and if one of the steps is 'set up a custom domain on Cloudflare Pages' then so be it)

This repository: [test-sk-cf-custom-domain](https://github.com/chientrm/test-sk-cf-custom-domain)

The page with default cf subdomain: [test-sk-cf-custom-domain.pages.dev](https://test-sk-cf-custom-domain.pages.dev/)

The page with custom domain: [test-sk-cf-custom-domain.dstserver.cloud](https://test-sk-cf-custom-domain.dstserver.cloud/)

Step to reproduce:

1. Browse page
2. Click `Invalidate` several times.

Expected: Text `Refreshing: true` appear for 3 seconds then being replaced with `Refreshing: false`.

Actual: Sometimes `Invalidate` do load new data, sometimes it load from `disk cache`.

CF cache settings: all `default`.

Rich-Harris:
I note that the repro's lockfile specifies `next.489` as the version. `cache-control: private, no-store` was added in `next.490`. Can you upgrade to the latest Kit and see if the bug is still present?

chientrm:
> I note that the repro's lockfile specifies `next.489` as the version. `cache-control: private, no-store` was added in `next.490`. Can you upgrade to the latest Kit and see if the bug is still present?

As your suggestion, I have done these steps:
1. Upgrade SvelteKit to latest version `next.493`.
After that, CF build failed with this error: `Error: Failed to publish your Function. Got error: Error 8000057: Rules must start with '/' in _routes.json`.
  Remove `static/favicon.png` help fix the `_routes.json` related error. This looks like another bug since a new PR #6530  🤔 .
2. Purge all CF cache on site `dstserver.cloud`
3. Repeat the testing steps. I did tried reload pages several times,
4. Results
    Screen Shot 2022-09-22 at 7 02 02 AM
Work as intended 👍

PR #6888, opened 2022-09-19
add support for cookies and setHeaders in handle hook

### LINKED ISSUE #6735: Can't set cookies/headers inside `handle` hook after call to `resolve`
### Describe the bug

`handle` hook doesn't add cookies as expected

### Reproduction

https://stackblitz.com/edit/node-shcasv

step 1: create a skeleton project
step 2: add `hooks.server.js`:

```js
/** @type {import("@sveltejs/kit").Handle} */
    console.log('hook ran');
    event.cookies.set('some', 'value', { secure: false, httpOnly: false });
    return await resolve(event);
}
```

step 3: open the site
step 4: try deleting the cookie (if present) on the browser and refresh. no new cookie is added

### Logs

_No response_

### System Info

```shell
System:
    CPU: (8) x64 Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz
    Memory: 2.05 GB / 3.73 GB
    Container: Yes
    Shell: 3.5.0 - /usr/bin/fish
  Binaries:
    Node: 18.7.0 - ~/.local/share/nvm/v18.7.0/bin/node
    Yarn: 1.22.10 - /mnt/c/Users/_/AppData/Roaming/npm/yarn
    npm: 8.15.0 - ~/.local/share/nvm/v18.7.0/bin/npm
  Browsers:
    Chrome: 103.0.5060.114
  npmPackages:
    svelte: ^3.44.0 => 3.50.1 
    vite: ^3.1.0 => 3.1.0
```

### Severity

### Additional Information

_No response_

david-plugge:
Closes #6735

Enables usage of `cookies` and `setHeaders` after or without calling `resolve(event)`.
Without the ability to use these helpers in handle the developer has to manually serialize cookies which seems unnecessary.
If this is not going to be merged the docs should make it clear how the helpers can be used in the hook. Or both helpers should be excluded from the event passed to handle.

**What it does:**

Headers in `Load` or `RequestHandler` are applied upon `resolve` so they are available in handle using `response.headers`, cookies are not (so they can be overwritten).
Headers and cookies set in handle are always applied.

Not 100% sure if that´s the correct/wanted behaviour.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -184,7 +184,6 @@ export async function respond(request, options, state) {
 	 * @param {import('types').ResolveOptions} [opts]
 	 */
 	async function resolve(event, opts) {
-		let includeCookies = false;
 		try {
 			if (opts) {
 				// TODO remove for 1.0
@@ -242,9 +241,6 @@ export async function respond(request, options, state) {
 					// want to cache stuff erroneously etc
 					add_headers_to_response(response, headers);
 				}
-				// only mark the cookies to be included
-				// if we add them to the headers it´s harder to delete them inside the handle hook
-				includeCookies = true;
 
 				return response;
 			}
@@ -282,10 +278,6 @@ export async function respond(request, options, state) {
 		} finally {
 			// reset headers
 			headers = {};
-			// clear the cookies if no valid route is found
-			if (!includeCookies) {
-				new_cookies.clear();
-			}
 		}
 	}
 
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -97,7 +97,6 @@ export async function respond(request, options, state) {
 	let headers = {};
 
 	const { cookies, new_cookies } = get_cookies(request, url);
-	let includeCookies = false;
 
 	if (state.prerendering) disable_search(url);
 
@@ -179,15 +178,13 @@ export async function respond(request, options, state) {
 		filterSerializedResponseHeaders: default_filter
 	};
 
-	let resolveCalled = false;
-
 	/**
 	 *
 	 * @param {import('types').RequestEvent} event
 	 * @param {import('types').ResolveOptions} [opts]
 	 */
 	async function resolve(event, opts) {
-		resolveCalled = true;
+		let includeCookies = false;
 		try {
 			if (opts) {
 				// TODO remove for 1.0
@@ -246,7 +243,7 @@ export async function respond(request, options, state) {
 					add_headers_to_response(response, headers);
 				}
 				// only mark the cookies to be included
-				// if we add them to the headers it´s hard to delete them inside the handle hook
+				// if we add them to the headers it´s harder to delete them inside the handle hook
 				includeCookies = true;
 
 				return response;
@@ -285,6 +282,10 @@ export async function respond(request, options, state) {
 		} finally {
 			// reset headers
 			headers = {};
+			// clear the cookies if no valid route is found
+			if (!includeCookies) {
+				new_cookies.clear();
+			}
 		}
 	}
 
@@ -309,6 +310,7 @@ export async function respond(request, options, state) {
 
 		// allows for `cookies` and `setHeaders` beeing used after and without the use `resolve(event)`
 		add_headers_to_response(response, headers);
+		add_cookies_to_headers(response.headers, Array.from(new_cookies.values()));
 
 		// respond with 304 if etag matches
 		if (response.status === 200 && response.headers.has('etag')) {
@@ -335,10 +337,6 @@ export async function respond(request, options, state) {
 					headers
 				});
 			}
-		}
-
-		if (!resolveCalled || includeCookies) {
-			add_cookies_to_headers(response.headers, Array.from(new_cookies.values()));
 		}
 
 		return response;
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -94,9 +94,10 @@ export async function respond(request, options, state) {
 	}
 
 	/** @type {Record<string, string>} */
-	const headers = {};
+	let headers = {};
 
 	const { cookies, new_cookies } = get_cookies(request, url);
+	let includeCookies = false;
 
 	if (state.prerendering) disable_search(url);
 
@@ -178,12 +179,15 @@ export async function respond(request, options, state) {
 		filterSerializedResponseHeaders: default_filter
 	};
 
+	let resolveCalled = false;
+
 	/**
 	 *
 	 * @param {import('types').RequestEvent} event
 	 * @param {import('types').ResolveOptions} [opts]
 	 */
 	async function resolve(event, opts) {
+		resolveCalled = true;
 		try {
 			if (opts) {
 				// TODO remove for 1.0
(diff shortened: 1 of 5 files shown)


david-plugge:
I will give it a try

david-plugge:
> Is this from a previous PR?

Yes, i didn´t add that function

packages/kit/src/runtime/server/index.js
```diff
@@ -281,13 +280,12 @@ export async function respond(request, options, state) {
...
-
-			event.setHeaders = () => {
-				throw new Error('Cannot use `setHeaders(...)` after the response has been generated');
-			};
+			// reset headers
+			headers = {};
+			// clear the cookies if no valid route is found
+			if (!includeCookies) {
+				new_cookies.clear();
+			}
```
Rich-Harris, line 288: Why do we need to do this?

david-plugge: by @dummdidumm in https://github.com/sveltejs/kit/issues/6735#issuecomment-1248119554
> It certainly needs to be called upon resolve because people could rely on the headers already being present on the response object when resolve returns.

This way headers don´t get set multiple times.

I´m not sure about the cookie part though.
https://github.com/sveltejs/kit/blob/6234c074a34066d10360bc3ba92804862efcce1e/packages/kit/src/runtime/server/index.js#L223
Here we check if a valid route exists and only in that case the headers and cookies are applied. I just tried to mimic that behaviour.
I guess that there is no way cookies and headers are set if no valid routes is found right? Then we can get rid of
```ts
if (!includeCookies) {
  new_cookies.clear();
}
```

Rich-Harris: Ah, I see. In that case, perhaps `setHeaders` and `cookies.set` should update the response immediately after `resolve` returns?

```js
response = await resolve(event);
setHeaders({ 'x-foo', 'bar' });
console.log(response.headers.get('x-foo')); // should this be `'bar'` or `null`?
```

I think we should always include cookies, regardless of whether the request matched a route — in other words we can get rid of the `includeCookies` stuff. I think the fact that that didn't already happen is best categorised as a bug.

> Not very related but is the `setHeaders` function going to change?

No, we decided against changing it to use the `Headers` API — it doesn't make sense when you look at it closely.

david-plugge: > Ah, I see. In that case, perhaps `setHeaders` and `cookies.set` should update the response immediately after `resolve` returns?

Not sure about cookies:

1. `cookies.set(...)` in `load`
2. `resolve` sets the cookies
3. `cookies.set(...)` in `handle`
4. `handle` resolves, we set the cookies again

If we don´t clear the cookies in between they will be set twice. But if we clear the cookies we can´t read them anymore from the `new_cookies` Map. And removing a cookie from the header is another thing to worry about.

> I think we should always include cookies

👍

Rich-Harris: Can't we do something like this?

```js
const response = await options.hooks.handle({...});

const { set } = event.cookies;
event.cookies.set = (name, value, opts) => {
  set.call(event.cookies, name, value, opts);
  response.headers.append('set-cookie', serialize(name, value, opts));
};

event.setHeaders = (headers) => {
  for (const key in headers) {
    response.headers.set(key, headers[key]);
  }
};
```

It's incomplete (doesn't prevent you from setting headers multiple times etc) and would be nicer if it didn't involve monkey-patching, but you get the idea — `cookies.get` continues to work, but it updates `response.headers` immediately

david-plugge: But the handle hook is already resolved by then, so the monkey-patching must happen inside `resolve`.
```ts
export async function handle({ resolve, event }) {

	event.cookies.set('key', 'value');

	return new Response(response.body, {
		status: 200
	})
}
```
In this case the cookie would be added to the response returned by `resolve`.

I think of the `setHeader` and `cookies.set` as a request scoped api so I do think it makes sense to apply headers and cookies to the final response.
If `setHeaders` and `cookies.set` only modify the response returned by `resolve` you cant´t use them if you don´t call `resolve`.

This gets more complicated than i expected...

Rich-Harris: > the monkey-patching must happen inside `resolve`

d'oh, of course 🤦 



> I think of the `setHeader` and `cookies.set` as a request scoped api so I do think it makes sense to apply headers and cookies to the final response

Counterpoint: if `setHeaders` is called inside a route handler (whether `+page.server.js` or `+server.js`) then I actually think it would be rather weird if the return value from `resolve` were discarded but the headers were still applied to a newly created `Response` object.

> This gets more complicated than i expected...

Indeed! I think the way to simplify this is to state that `setHeaders` and `cookies.set` only apply to the `resolve` return value, and cannot be used after it returns. If you want to changes headers after that point, you can do so via `response.headers`, and if you want to go rogue by not using the `resolve` return value then that's on you.

david-plugge: > Counterpoint: if `setHeaders` is called inside a route handler (whether `+page.server.js` or `+server.js`) then I actually think it would be rather weird if the return value from `resolve` were discarded but the headers were still applied to a newly created `Response` object.

Totally agree, i was talking about headers set in the handle hook.

> Indeed! I think the way to simplify this is to state that `setHeaders` and `cookies.set` only apply to the `resolve` return value, and cannot be used after it returns. If you want to changes headers after that point, you can do so via `response.headers`, and if you want to go rogue by not using the `resolve` return value then that's on you.

Yeah i guess that´s the only way to keep it simple.
One thing i´d like to see though is a serialize function for cookies with the same defaults `cookies.set` uses so we don´t have to manually add `cookie` as a dependency since it´s already included.

david-plugge: So i think I´ll close this PR and open a new one that:

- fixes the cookie bug so they are always included
  > I think we should always include cookies, regardless of whether the request matched a route — in other words we can get rid of the `includeCookies` stuff. I think the fact that that didn't already happen is best categorised as a bug.
- exports a function `serializeCookie` with the same defaults as `cookies.set`

Is there anything I missed?

Rich-Harris: Sounds good, thank you — though maybe it would be good if it was `cookies.serialize` rather than a separate function? arguably more convenient, and makes it easier to document why you'd use it over `cookies.set` if it's in the same place

cdcarson: > maybe it would be good if it was cookies.serialize

Isn't the idea that folks would use this if they go rogue and do their own cookie thing? If so, we might want to protect innocent, law-abiding folks from having to think about `cookies.set` vs `cookies.serialize`. I'd vote for making `serialize` a separate export (that the api would also use.)

Rich-Harris: It just feels like having all the cookie-related functionality in one place is going to make things more discoverable, and clearer that the two things are related (i.e. use the same defaults, etc). E.g. if I call `cookies.set(...)` after `resolve` has returned and I see this error message...

> Cannot call `cookies.set(...)` after `resolve` has returned a response. Use `response.headers.append('set-cookie', cookies.serialize(...))` instead

...then it doesn't feel like I'm dealing with two disconnected systems.

cdcarson: > two disconnected systems

Got it. Just worried about people getting confused in the following way... 

```ts
cookies.set('foo', cookies.serialize('bar'), {path: '/'});
```
...which is a bug.

cdcarson: That could be ameliorated by changing the `value` arg to `unserializedValue`...

```ts
set(name: string, unserializedValue: string, opts?: import('cookie').CookieSerializeOptions): void;
```

I'm fine either way.

PR #6871, opened 2022-09-18
Correct HTML attribute parsing

kissge:
Fix #6870.

Handling all possible character reference is I suppose definetely something external library comes into play.
There are several npm packages on this topic, and I chose [entities](https://www.npmjs.com/package/entities) judging from popularity and performance.

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [x] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #6891, opened 2022-09-19
Decode HTML entities in `href` attributes when crawling

Rich-Harris:
This is an alternative to #6871. `entities` is quite a large dependency and has a lot of functionality we don't need, so I figured we could probably just do it ourselves. 

### Tests

### Changesets


.changeset/spotty-trainers-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Fix the problem HTML attribute parsing does not take HTML reference into account.
```
Rich-Harris, line 5:
```suggestion
Decode HTML entities in `href` attributes when crawling
```

dummdidumm (approved):

PR #6896, opened 2022-09-19
bundle with rollup instead of esbuild

Rich-Harris:
Alternative to #6855, which is itself an alternative to #6797.

With this PR, we bundle `adapter-node` apps with Rollup instead of esbuild. It's slower, but avoids https://github.com/evanw/esbuild/issues/1921 which doesn't look like it'll be fixed any time soon (even though there's a PR — https://github.com/evanw/esbuild/pull/2067), and means that we can deal with both ESM and CJS dependencies without adding more configuration.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-node/index.js
+++ b/packages/adapter-node/index.js
@@ -52,6 +52,9 @@ export default function (opts = {}) {
 
 			const pkg = JSON.parse(readFileSync('package.json', 'utf8'));
 
+			// we bundle the Vite output so that deployments only need
+			// their production dependencies. Anything in devDependencies
+			// will get included in the bundled code
 			const bundle = await rollup({
 				input: {
 					index: `${tmp}/index.js`,
--- a/packages/adapter-node/index.js
+++ b/packages/adapter-node/index.js
@@ -1,6 +1,10 @@
 import { readFileSync, writeFileSync } from 'fs';
 import { fileURLToPath } from 'url';
 import * as esbuild from 'esbuild';
+import { rollup } from 'rollup';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
+import commonjs from '@rollup/plugin-commonjs';
+import json from '@rollup/plugin-json';
 
 const files = fileURLToPath(new URL('./files', import.meta.url).href);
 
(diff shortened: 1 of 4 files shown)


packages/adapter-node/index.js
```diff
@@ -49,16 +53,20 @@ export default function (opts = {}) {
...
 			const pkg = JSON.parse(readFileSync('package.json', 'utf8'));
 
-			await esbuild.build({
-				platform: 'node',
-				sourcemap: 'linked',
-				target: 'es2022',
-				entryPoints: [`${tmp}/index.js`, `${tmp}/manifest.js`],
-				outdir: `${out}/server`,
-				splitting: true,
+			const bundle = await rollup({
```
benmccann, line 55: can we add a comment explaining why a second bundling step is needed

benmccann (commented):
this sounds like a better plan to me than https://github.com/sveltejs/kit/pull/6855. it will be slower but will remove lots of headaches

packages/adapter-node/index.js
Rich-Harris: done

packages/adapter-node/index.js
```diff
@@ -1,6 +1,9 @@
 import { readFileSync, writeFileSync } from 'fs';
 import { fileURLToPath } from 'url';
-import * as esbuild from 'esbuild';
+import { rollup } from 'rollup';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
+import commonjs from '@rollup/plugin-commonjs';
+import json from '@rollup/plugin-json';
```
benmccann, line 6: do we need the JSON plugin? will people be importing JSON that hasn't already been bundled by Vite?

Conduitry: They could be using a devdependency that includes JSON which isn't bundled yet but which we want to bundle when adapting.

packages/adapter-node/package.json
```diff
@@ -45,6 +45,7 @@
 		"uvu": "^0.5.3"
 	},
 	"dependencies": {
-		"esbuild": "^0.15.7"
+		"@rollup/plugin-commonjs": "^22.0.1",
```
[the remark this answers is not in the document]
Rich-Harris: oops, fixed

packages/adapter-node/index.js
Rich-Harris: yep, this needs to work with `require('./stuff.json')`

benmccann (commented):
lgtm

the only other thought I had is whether it might be better to use vite rather than rollup to do the bundling. vite will already be in the user's dependencies, but the rollup plugins will not be, so it could reduce the dependency footprint

Rich-Harris:
True. I feel like it adds some more moving parts though, since Vite does a lot of stuff on top of what you get with Rollup. With Rollup everything is very straightforward and predictable

PR #6897, opened 2022-09-19
[fix] 404 cleanup

### LINKED ISSUE #6213: redirect in a layout fails if there is no page for it.
### Describe the bug

If you redirect in a layout that does not have a page you get the error
```
Error: {"status":301,"location":"/gb"}
    at Module.coalesce_to_error (file:///home/projects/sveltejs-kit-template-default-zpo2qc/node_modules/@sveltejs/kit/src/utils/error.js:27:5)
    at Module.respond_with_error (file:///home/projects/sveltejs-kit-template-default-zpo2qc/node_modules/@sveltejs/kit/src/runtime/server/page/respond_with_error.js:100:33)
    at async resolve (file:///home/projects/sveltejs-kit-template-default-zpo2qc/node_modules/@sveltejs/kit/src/runtime/server/index.js:382:13)
    at async respond (file:///home/projects/sveltejs-kit-template-default-zpo2qc/node_modules/@sveltejs/kit/src/runtime/server/index.js:217:20)
    at async eval (file:///home/projects/sveltejs-kit-template-default-zpo2qc/node_modules/@sveltejs/kit/src/vite/dev/index.js:416:22)
```

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-zpo2qc?file=src%2Froutes%2F%2Blayout.server.ts

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.5.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 467.41 MB / 32.00 GB
    Shell: 5.9 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 18.7.0 - /opt/homebrew/bin/node
    npm: 8.18.0 - /opt/homebrew/bin/npm
  Browsers:
    Brave Browser: 104.1.42.97
    Safari: 15.6.1
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.429
    svelte: ^3.44.0 => 3.49.0
    vite: ^3.0.0 => 3.0.9
```

### Severity

annoyance

### Additional Information

_No response_

### LINKED ISSUE #6774: More information about error in HandleServerError
### Describe the problem

Currently, the `handleError` hook in the server hooks gets triggered on a 404 error. I'm a bit confused about:

1. why [`HandleServerError`](https://kit.svelte.dev/docs/types#sveltejs-kit-handleservererror) needs to handle 404s. 
2. how [`HandleServerError`](https://kit.svelte.dev/docs/types#sveltejs-kit-handleservererror) can reliably tell the true "source" of an error coming from Kit, such as a 404.

## 1. Why Does HandleServerError need to handle 404?

Suppose i have a routeID like `/blogs/[slug]`, and the user requests `/blogs/nonexistent`. What makes sense to me is that in my `/src/routes/blogs/[slug]/+page.server.js`, I determine that the slug "nonexistent" does not exist and then I do something like this....

```js
import { error } from "@sveltejs/kit"

export function load() {
    throw error(404, "not blog with that slug")
}
```

`HandleServerError` is NOT triggered b/c this is an "expected error". This is imo the intuitive outcome because a 404 is the "client's fault", it is expected, and  not what `HandleServerError` is meant to handle, at least according to my understanding of [this section](https://kit.svelte.dev/docs/hooks#shared-hooks-handleerror)

> This function is not called for expected errors

**However**, if the user requests some route which doesn't have a directory in `routes/` at all, like `/totallyhasnohandler`, then a capital "E" Error is thrown by the framework, and `HandleServerError` *is* triggered. That is surprising to me. I don't want 404's to get included with the rest of my error-monitoring infra, which brings me to my next point....

## 2. How to reliably know why Kit threw this error and allow my `handleError` to react appropriately.

I want to divert things like 404s *away* from where the rest of the truly unexpected errors are going. Currently my hook looks somethign like this

```js
export function handleError({ error }) {
  if (error instanceof Error && error.message.includes("Not found:")) {
    return { message: "Not found" }
  }
  logToMoreIntenseErrorMonitoring(error)
  return { message: "internal error" }
}
```

Interestingly, Kit still knows to respond with a 404 to the request, but that state isn't getting passed to my hook, so I can't react to that. As a workaround, I am matching against the error message string which is getting me frowns from teammates. It just seems like there should be a more reliable way to get the source of this error. 

404s either wouldn't trigger the error hook, or maybe 404s and other errors thrown by Kit would be more differentiated.... maybe the hook can have an option status code parameter if it has that? Maybe instead of the exception being a vanilla `Error` instance, it would be an `HttpError` instance which has a status?

### Alternatives considered

Matching against error message strings when I see them in my logs and I want to do different logic on them

### Importance

nice to have

### Additional Information

I have a little repro of this phenomenon: https://stackblitz.com/edit/sveltejs-kit-template-default-7jzfsn?file=src%2Fhooks.server.js&terminal=dev

I edited the nav to link to a non-existent route just for the purposes of triggering this case, but this is not indicative of a realistic scenario. In that case we would want to catch the broken link. But in the real world, clients and bots request all kinds of random URLs that trigger 404s.

### LINKED ISSUE #6793: 404 has the message "Internal Error"

### LINKED ISSUE #6799: Thrown expected errors get a different status code on non-existing routes
### Describe the bug

If you throw an error from `+layout.ts`, this gets rendered with the correct status code and message using the default error page.

```
import { error } from "@sveltejs/kit";
import type { LayoutLoad } from "./$types";

export const load: LayoutLoad = async () => {
  throw error(401, "Invalid token.");
};
```

However, if you go to a non-existing page, that error code now becomes 500.

### Reproduction

https://github.com/kevinrenskers/sk-error-page

Open http://127.0.0.1:5173 or http://127.0.0.1:5173/page, and you will see a 401 "Invalid token" error. But if you open http://127.0.0.1:5173/foo (which doesn't exist) you will see instead the same message but now with the status code 500.

I would expect to still see the same 401 code there, as I have not modified it anywhere. Or I would expect to see a 404? But definitely not a 500.

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.6
    CPU: (10) arm64 Apple M1 Max
    Memory: 717.94 MB / 32.00 GB
    Shell: 3.5.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 16.17.0 - ~/Library/pnpm/node
    npm: 8.15.0 - ~/Library/pnpm/npm
  Browsers:
    Safari: 16.0
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.73
    svelte: ^3.44.0 => 3.50.1
    vite: ^3.1.0 => 3.1.0
```

### Severity

annoyance

### Additional Information

_No response_

### LINKED ISSUE #6886: Redirection with +layout.server.js [server side] isn't working
### Describe the bug

I have some old url parameters of NextJs, and I have recently unpdated my app to SvelteKit, and it's in production. And I wanted to redirect my old urls to new ones and tried the instruction according to https://kit.svelte.dev/docs/load#redirects about tweaking `+layout.server.js` and using `throw`.

### Reproduction

The below code isn't working, `throw` is just logging some useless stuff.
```
    import { redirect } from "@sveltejs/kit";
    
    
    export function load({ url }) {
      if (url.pathname === '/old-url-param') {
        throw redirect(301, '/new-url-param');
      } else if (url.pathname === '/old-url-param-2') {
        throw redirect(301, '/new-url-param-2');
      }
    }
```

### Logs

```shell
{
  has: [ 1, 31 ],
  keys: [
    { open: '\x1B[1m', close: '\x1B[22m', rgx: /\x1b\[22m/g },
    { open: '\x1B[31m', close: '\x1B[39m', rgx: /\x1b\[39m/g }
  ],
  reset: [Function: bound ],
  bold: [Function: bound ],
  dim: [Function: bound ],
  italic: [Function: bound ],
  underline: [Function: bound ],
  inverse: [Function: bound ],
  hidden: [Function: bound ],
  strikethrough: [Function: bound ],
  black: [Function: bound ],
  red: [Function: bound ],
  green: [Function: bound ],
  yellow: [Function: bound ],
  blue: [Function: bound ],
  magenta: [Function: bound ],
  cyan: [Function: bound ],
  white: [Function: bound ],
  gray: [Function: bound ],
  grey: [Function: bound ],
  bgBlack: [Function: bound ],
  bgRed: [Function: bound ],
  bgGreen: [Function: bound ],
  bgYellow: [Function: bound ],
  bgBlue: [Function: bound ],
  bgMagenta: [Function: bound ],
  bgCyan: [Function: bound ],
  bgWhite: [Function: bound ]
}
Redirecting /generator/about-us-generator
{
  has: [ 1, 31 ],
  keys: [
    { open: '\x1B[1m', close: '\x1B[22m', rgx: /\x1b\[22m/g },
    { open: '\x1B[31m', close: '\x1B[39m', rgx: /\x1b\[39m/g }
  ],
  reset: [Function: bound ],
  bold: [Function: bound ],
  dim: [Function: bound ],
  italic: [Function: bound ],
  underline: [Function: bound ],
  inverse: [Function: bound ],
  hidden: [Function: bound ],
  strikethrough: [Function: bound ],
  black: [Function: bound ],
  red: [Function: bound ],
  green: [Function: bound ],
  yellow: [Function: bound ],
  blue: [Function: bound ],
  magenta: [Function: bound ],
  cyan: [Function: bound ],
  white: [Function: bound ],
  gray: [Function: bound ],
  grey: [Function: bound ],
  bgBlack: [Function: bound ],
  bgRed: [Function: bound ],
  bgGreen: [Function: bound ],
  bgYellow: [Function: bound ],
  bgBlue: [Function: bound ],
  bgMagenta: [Function: bound ],
  bgCyan: [Function: bound ],
  bgWhite: [Function: bound ]
}
```

### System Info

```shell
System:
    CPU: (8) x64 AMD Ryzen 5 3500U with Radeon Vega Mobile Gfx
    Memory: 593.53 MB / 9.60 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 16.17.0 - /usr/local/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 8.15.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 105.0.5195.52
    Firefox: 104.0
```

### Severity

### Additional Information

Just lemme know how can I redirect from server side within `+layout.js` or `+layout.server.js` or something equivalent.

dummdidumm:
Fixes #6799
Fixes #6793
Fixes #6886
Fixes #6213
Closes #6774

### Tests

### Changesets

### DIFF
--- a/documentation/docs/04-advanced-routing.md
+++ b/documentation/docs/04-advanced-routing.md
@@ -61,16 +61,6 @@ export function load(event) {
 }
 ```
 
-Alternatively, you could also `throw` a `redirect` to get back to an existing page.
-
-If you don't want to invoke `+error.svelte` in this case, you could instead create a `+page.svelte` file with your custom 404 content:
-
-```html
-/// file: src/routes/marx-brothers/[...path]/+page.svelte
-<h1>Not Found</h1>
-<p>This Marx brother never lived</p>
-```
-
 > If you don't handle 404 cases, they will appear in [`handleError`](/docs/hooks#shared-hooks-handleerror)
 
 ### Matching
--- a/documentation/docs/04-advanced-routing.md
+++ b/documentation/docs/04-advanced-routing.md
@@ -71,6 +71,8 @@ If you don't want to invoke `+error.svelte` in this case, you could instead crea
 <p>This Marx brother never lived</p>
 ```
 
+> If you don't handle 404 cases, they will appear in [`handleError`](/docs/hooks#shared-hooks-handleerror)
+
 ### Matching
 
 A route like `src/routes/archive/[page]` would match `/archive/3`, but it would also match `/archive/potato`. We don't want that. You can ensure that route parameters are well-formed by adding a _matcher_ — which takes the parameter string (`"3"` or `"potato"`) and returns `true` if it is valid — to your [`params`](/docs/configuration#files) directory...
--- a/documentation/docs/04-advanced-routing.md
+++ b/documentation/docs/04-advanced-routing.md
@@ -22,7 +22,11 @@ If the number of route segments is unknown, you can use rest syntax — for exam
 }
 ```
 
-This also allows you to render custom 404s. Given these routes...
+> `src/routes/a/[...rest]/z/+page.svelte` will match `/a/z` (i.e. there's no parameter at all) as well as `/a/b/z` and `/a/b/c/z` and so on. Make sure you check that the value of the rest parameter is valid, for example using a [matcher](#matching).
+
+#### 404 pages
+
+Rest parameters also allow you to render custom 404s. Given these routes...
 
 ```
 src/routes/
@@ -47,7 +51,25 @@ src/routes/
 └ +error.svelte
 ```
 
-> `src/routes/a/[...rest]/z/+page.svelte` will match `/a/z` (i.e. there's no parameter at all) as well as `/a/b/z` and `/a/b/c/z` and so on. Make sure you check that the value of the rest parameter is valid, for example using a [matcher](#matching).
+```js
+/// file: src/routes/marx-brothers/[...path]/+page.js
+import { error } from '@sveltejs/kit';
+
+/** @type {import('./$types').PageLoad} */
+export function load(event) {
+	throw error(404, 'Not Found');
+}
+```
+
+Alternatively, you could also `throw` a `redirect` to get back to an existing page.
+
+If you don't want to invoke `+error.svelte` in this case, you could instead create a `+page.svelte` file with your custom 404 content:
+
+```html
+/// file: src/routes/marx-brothers/[...path]/+page.svelte
+<h1>Not Found</h1>
+<p>This Marx brother never lived</p>
+```
 
 ### Matching
 
--- a/packages/create-svelte/templates/default/src/routes/[...notfound]/+page.svelte
+++ b/packages/create-svelte/templates/default/src/routes/[...notfound]/+page.svelte
@@ -1,0 +1,6 @@
+<!-- 
+This file is shown when a URL doesn't match any other routes.
+More info: https://kit.svelte.dev/docs/advanced-routing#rest-parameters-404-pages
+-->
+<h1>404</h1>
+<p>Page not found</p>
--- a/packages/create-svelte/templates/default/src/routes/[...notfound]/+page.svelte
+++ b/packages/create-svelte/templates/default/src/routes/[...notfound]/+page.svelte
@@ -1,6 +1,0 @@
-<!-- 
-This file is shown when a URL doesn't match any other routes.
-More info: https://kit.svelte.dev/docs/advanced-routing#rest-parameters-404-pages
--->
-<h1>404</h1>
-<p>Page not found</p>
--- a/packages/create-svelte/templates/skeleton/src/routes/[...notfound]/+page.svelte
+++ b/packages/create-svelte/templates/skeleton/src/routes/[...notfound]/+page.svelte
@@ -1,0 +1,6 @@
+<!-- 
+This file is shown when a URL doesn't match any other routes.
+More info: https://kit.svelte.dev/docs/advanced-routing#rest-parameters-404-pages
+-->
+<h1>404</h1>
+<p>Page not found</p>
--- a/packages/create-svelte/templates/skeleton/src/routes/[...notfound]/+page.svelte
+++ b/packages/create-svelte/templates/skeleton/src/routes/[...notfound]/+page.svelte
@@ -1,6 +1,0 @@
-<!-- 
-This file is shown when a URL doesn't match any other routes.
-More info: https://kit.svelte.dev/docs/advanced-routing#rest-parameters-404-pages
--->
-<h1>404</h1>
-<p>Page not found</p>
--- a/.changeset/silver-pugs-worry.md
+++ b/.changeset/silver-pugs-worry.md
@@ -1,6 +1,5 @@
 ---
 '@sveltejs/kit': patch
-'create-svelte': patch
 ---
 
 [fix] tighten up handling and documentation around 404
--- a/.changeset/silver-pugs-worry.md
+++ b/.changeset/silver-pugs-worry.md
@@ -1,0 +1,6 @@
+---
+'@sveltejs/kit': patch
+'create-svelte': patch
+---
+
+[fix] tighten up handling and documentation around 404
--- a/documentation/docs/07-hooks.md
+++ b/documentation/docs/07-hooks.md
@@ -136,7 +136,7 @@ The following can be added to `src/hooks.server.js` _and_ `src/hooks.client.js`:
 If an unexpected error is thrown during loading or rendering, this function will be called with the `error` and the `event`. This allows for two things:
 
 - you can log the error
-- you can generate a custom representation of the error that is safe to show to users, omitting sensitive details like messages and stack traces. The returned value, which defaults to `{ message: 'Internal Error' }`, becomes the value of `$page.error`. To make this type-safe, you can customize the expected shape by declaring an `App.PageError` interface (which must include `message: string`, to guarantee sensible fallback behavior).
+- you can generate a custom representation of the error that is safe to show to users, omitting sensitive details like messages and stack traces. The returned value becomes the value of `$page.error`. It defaults to `{ message: 'Not Found' }` in case of a 404 (you can detect them through `event.routeId` being `null`) and to `{ message: 'Internal Error' }` for everything else. To make this type-safe, you can customize the expected shape by declaring an `App.PageError` interface (which must include `message: string`, to guarantee sensible fallback behavior).
 
 The following code shows an example of typing the error shape as `{ message: string; code: string }` and returning it accordingly from the `handleError` functions:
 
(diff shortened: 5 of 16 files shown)


documentation/docs/04-advanced-routing.md
```diff
@@ -47,7 +51,25 @@ src/routes/
...
+
+Alternatively, you could also `throw` a `redirect` to get back to an existing page.
+
+If you don't want to invoke `+error.svelte` in this case, you could instead create a `+page.svelte` file with your custom 404 content:
+
+```html
+/// file: src/routes/marx-brothers/[...path]/+page.svelte
+<h1>Not Found</h1>
+<p>This Marx brother never lived</p>
+```
```
Rich-Harris, line 72: Redirecting in this situation is a bit of a bad practice, and it's very bad for SEO to return a 200 for a not found resource, so I don't think we should promote these patterns in the docs

```suggestion
```

packages/create-svelte/templates/default/src/routes/[...notfound]/+page.svelte
```diff
@@ -0,0 +1,6 @@
+<!-- 
+This file is shown when a URL doesn't match any other routes.
+More info: https://kit.svelte.dev/docs/advanced-routing#rest-parameters-404-pages
+-->
+<h1>404</h1>
+<p>Page not found</p>
```
Comment (Rich-Harris) [inline packages/create-svelte/templates/default/src/routes/[...notfound]/+page.svelte:6]: See above — we really shouldn't be returning a 200 response here

```suggestion
```

packages/create-svelte/templates/skeleton/src/routes/[...notfound]/+page.svelte
```diff
@@ -0,0 +1,6 @@
+<!-- 
+This file is shown when a URL doesn't match any other routes.
+More info: https://kit.svelte.dev/docs/advanced-routing#rest-parameters-404-pages
+-->
+<h1>404</h1>
+<p>Page not found</p>
```
Comment (Rich-Harris) [inline packages/create-svelte/templates/skeleton/src/routes/[...notfound]/+page.svelte:6]:
```suggestion
```

Rich-Harris (changes requested):
The code changes look good but we shouldn't be encouraging people to have `[...404]/+page.svelte` files

documentation/docs/04-advanced-routing.md
dummdidumm: That makes sense

packages/create-svelte/templates/default/src/routes/[...notfound]/+page.svelte
Comment (dummdidumm) [reply packages/create-svelte/templates/default/src/routes/[...notfound]/+page.svelte:6]: Should we just not care about this in the skeleton/default template then? I feel like people might want to have this pattern implemented almost everytime, so we would save them some keystrokes and also possibly confusion how to deal with this (when they haven't read the advanced docs section).

Comment (Rich-Harris) [reply packages/create-svelte/templates/default/src/routes/[...notfound]/+page.svelte:6]: I'm just not sure what the point of such a file would be? We already have `+error.svelte`. My vote would be to not worry about it

documentation/docs/04-advanced-routing.md
dummdidumm: On second thought - why is `redirect` bad practise? It's the only way to have it working for both client and server (doing it in `handle` only works for the server, SPAs are left out), and in case of a first page hit the browser will just get a redirect response just like it would from the `handle` hook.

Rich-Harris: It's just semantically incorrect. HTTP status codes have very specific meanings, and none of the [redirection codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#redirection_messages) (300, 301, 302, 303, 304, 307, 308) are appropriate when a resource can't be found

packages/create-svelte/templates/default/src/routes/[...notfound]/+page.svelte
Comment (dummdidumm) [reply packages/create-svelte/templates/default/src/routes/[...notfound]/+page.svelte:6]: Ok - and now since we fall back to `Not Found` the default behavior is OK, too. Will remove these files again and adjust the changeset.

documentation/docs/04-advanced-routing.md
dummdidumm: Internal SPAs don't care about semantics, but I agree that this is a good enough argument to not bless it in the docs - people will likely try out and see that works, too.

Rich-Harris (approved):

Issue #6608, opened 2022-09-06
Add `Origin` request header for the server-side fetch


aradalvand:
### Describe the bug

Related #6550
Basically what @kevinrenskers said [here](https://github.com/sveltejs/kit/pull/6550#issuecomment-1236366109).

The behavior isn't currently sound and can cause unexpected errors — my server is configured to return the right CORS headers, but I'm getting CORS errors using the server-side fetch that I wouldn't get using the browser's fetch — the server-side fetch must automatically send an `Origin` header along with the request in order for the CORS stuff to work properly, this is how the native browser `fetch` works.

### Reproduction

Explained above.

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 20.04.4 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
    Memory: 1.01 GB / 3.77 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 16.15.1 - /usr/bin/node
    Yarn: 1.22.15 - /usr/bin/yarn
    npm: 8.11.0 - /usr/bin/npm
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.64 
    @sveltejs/adapter-node: ^1.0.0-next.86 => 1.0.0-next.86 
    @sveltejs/kit: next => 1.0.0-next.472 
    svelte: ^3.44.0 => 3.49.0 
    vite: ^3.1.0-beta.1 => 3.1.0
```

### Severity

blocking an upgrade

itssumitrai:
Facing the same issue, `Origin` header is not being forwarded which creates issues with CORS

ponderingexistence:
Same here

andreasnuesslein:
same here; problem was introduced in https://github.com/sveltejs/kit/tree/%40sveltejs/kit%401.0.0-next.468

randyridge:
also hit this, can workaround by manually setting the origin header, but it's a little bit of a hassle as we have to configure it per-environment and we use fetch indirectly via urql...

rohanrajpal:
One thing to note is that (after this issue is fixed) it'd be great to still allow customizing the origin.

It would be useful for the folks who run sveltekit behind an nginx proxy, typically in a proxy the URL would be `http://local.dev` but request would be coming from `https://local.dev`

itssumitrai:
[...]

PR #6901, opened 2022-09-19
fix: (issue-6608) Add origin header to server-side fetch

bummzack:
Add the `Origin` header to server-side fetch. 
The problem has been introduced with #6550, where browser CORS behaviour was simulated for the server-side requests. For CORS requests to work properly, the `Origin` header is mandatory.

Fixes: #6608 

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/fetch.js
+++ b/packages/kit/src/runtime/server/page/fetch.js
@@ -68,6 +68,13 @@ export function create_fetch({ event, options, state, route, prerender_default, 
 
 				const url = new URL(request.url);
 
+				if (
+					!request.headers.get('origin') &&
+					(url.origin !== event.url.origin || request.method !== 'GET')
+				) {
+					request.headers.set('origin', event.url.origin);
+				}
+
 				if (url.origin !== event.url.origin) {
 					// allow cookie passthrough for "same-origin"
 					// if SvelteKit is serving my.domain.com:
@@ -83,11 +90,6 @@ export function create_fetch({ event, options, state, route, prerender_default, 
 					) {
 						const cookie = get_cookie_header(url, request.headers.get('cookie'));
 						if (cookie) request.headers.set('cookie', cookie);
-					}
-
-					// Add request origin header in case it is missing
-					if (!request.headers.get('origin')) {
-						request.headers.set('origin', event.url.origin);
 					}
 
 					let response = await fetch(request);
--- a/packages/kit/src/runtime/server/page/fetch.js
+++ b/packages/kit/src/runtime/server/page/fetch.js
@@ -85,6 +85,11 @@ export function create_fetch({ event, options, state, route, prerender_default, 
 						if (cookie) request.headers.set('cookie', cookie);
 					}
 
+					// Add request origin header in case it is missing
+					if (!request.headers.get('origin')) {
+						request.headers.set('origin', event.url.origin);
+					}
+
 					let response = await fetch(request);
 
 					if (request.mode === 'no-cors') {
--- a/packages/kit/src/runtime/server/page/fetch.js
+++ b/packages/kit/src/runtime/server/page/fetch.js
@@ -68,11 +68,17 @@ export function create_fetch({ event, options, state, route, prerender_default, 
 
 				const url = new URL(request.url);
 
-				if (
-					!request.headers.get('origin') &&
-					(url.origin !== event.url.origin || request.method !== 'GET')
-				) {
+				if (!request.headers.has('origin')) {
 					request.headers.set('origin', event.url.origin);
+				}
+
+				// Remove Origin, according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin#description
+				if (
+					(request.method === 'GET' || request.method === 'HEAD') &&
+					((request.mode === 'no-cors' && url.origin !== event.url.origin) ||
+						url.origin === event.url.origin)
+				) {
+					request.headers.delete('origin');
 				}
 
 				if (url.origin !== event.url.origin) {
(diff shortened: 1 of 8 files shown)


packages/kit/src/runtime/server/page/fetch.js
```diff
@@ -68,6 +68,13 @@ export function create_fetch({ event, options, state, route, prerender_default,
 
 				const url = new URL(request.url);
 
+				if (
+					!request.headers.get('origin') &&
```
dummdidumm, line 72: In https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin#description it says "if a cross-origin GET or HEAD request is made in no-cors mode, the Origin header will not be added" - should we add that check here, or is this irrelevant here?

dummdidumm, line 72: In https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin#description it says "if a cross-origin GET or HEAD request is made in no-cors mode, the Origin header will not be added" - should we add that check here, or is this irrelevant here?

```suggestion
					!request.headers.get('origin') &&
					!request.mode !== 'no-cors' &&
```

bummzack: Hm, I think it's a good idea to stick as close to the spec as possible if the goal is to simulate browser behaviour. In that case we'd also have to add the `Origin` to all *same-origin* requests that are **not** `GET` or `HEAD` and exclude from `GET` and `HEAD` if the request is "no-cors"

dummdidumm: That is already part of the if-condition

bummzack: Is it? It doesn't check `HEAD` (not sure if relevant) and the `no-cors` check would omit the `Origin` header for all no-cors requests (but should only omit for `GET` and `HEAD`).

bummzack: My proposition to solve this:
```js
if (!request.headers.has('origin')) {
	request.headers.set('origin', event.url.origin);
}

// Remove Origin, according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin#description
if (
	(request.method === 'GET' || request.method === 'HEAD') &&
	((request.mode === 'no-cors' && url.origin !== event.url.origin) || url.origin === event.url.origin)
) {
	request.headers.delete('origin');
}
```

What do you think @dummdidumm ? This way the origin is always added, but removed for the special cases outlined on MDN where the origin is omitted

bummzack: I have modified the code with the changes outlined above. This should ensure the same behaviour as is described on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin#description). Eg.
- Add `Origin` to all cross-origin requests
- Don't send the `Origin` header for **same-origin** `GET` and `HEAD` requests
- Don't send the `Origin` header for **cross-origin** `"no-cors"` `GET` and `HEAD` requests

packages/kit/src/runtime/server/page/fetch.js
```diff
@@ -68,6 +68,19 @@ export function create_fetch({ event, options, state, route, prerender_default,
...
+				}
+
+				// Remove Origin, according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin#description
+				if (
+					(request.method === 'GET' || request.method === 'HEAD') &&
+					((request.mode === 'no-cors' && url.origin !== event.url.origin) ||
+						url.origin === event.url.origin)
+				) {
+					request.headers.delete('origin');
+				}
```
dummdidumm, line 82: The docs say nothing about removing the header when it's already present. So I think probably the most spec-compliant way is to add it when it's not present, given the mentioned conditions:
```suggestion
				// add origin, according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin#description
				if (
					!request.headers.has('origin') &&
					request.mode !== 'no-cors' &&
					(url.origin !== event.url.origin ||
						(request.method !== 'GET' && request.method !== 'HEAD'))
				) {
					request.headers.set('origin', event.url.origin);
				}
```

bummzack: @dummdidumm My understanding was, that the goal is to replicate browser behaviour. The browser does in fact remove the "Origin" for the request where no `Origin` should be present. You can easily try this yourself by doing a same-origin GET request and adding the `Origin` header

dummdidumm: Oh interesting ... then forget that comment. The logic still seems a little off to me, `no-cors` on a POST would not remove the origin, but it should. I think the logic can be simplified to just doing "if `no-cors` present, remove origin" (since it's a non-simple header according to the spec).
```suggestion
				if (!request.headers.has('origin')) {
					request.headers.set('origin', event.url.origin);
				}

				// Remove Origin, according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin#description
				if (
					request.mode === 'no-cors' ||
					((request.method === 'GET' || request.method === 'HEAD') &&
						url.origin === event.url.origin)
				) {
					request.headers.delete('origin');
				}
```

bummzack: @dummdidumm That is not what the description on MDN says. Also, if you do a `no-cors` POST request, the browser will add the Origin header (for both same and cross-origin requests). You can easily test that with the browser console.

Rich-Harris (approved):

Rich-Harris:
thank you!

PR #6918, opened 2022-09-20
[fix] break out of 404 SPA loop

### LINKED ISSUE #6854: ssr false with +layout.server file cause loop page refresh on 404
### Describe the bug

Create a `+layout.server.js` file when SSR is `false` cause loop refresh of the page.

### Reproduction

```
npm create svelte
npm i
```

Create the following files:
```
/src/routes/+layout.js

export const ssr = false;
```

```
/src/routes/+layout.server.js (empty)
```

```
npm run dev
```

Now when navigating on a non-existent URL the page keeps refreshing indefinitely.

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.22000
    CPU: (16) x64 AMD Ryzen 9 4900H with Radeon Graphics
    Memory: 11.75 GB / 31.42 GB
  Binaries:
    Node: 18.7.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.15.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (105.0.1343.27)
  npmPackages:
    svelte: ^3.46.0 => 3.50.1
    vite: ^3.1.0 => 3.1.1
```

### Severity

annoyance

### Additional Information

_No response_

dummdidumm:
Fixes #6854, but not other cases - when the root +layout(.server).js fails, we also go back to the server for a reload, which would also result in a loop. This could be handled maybe through storing some state in session storage and checking that upon entering an error state again.

Opening this PR regardless to get some discussion going - if we want to explictitly only handle this case for now, or go with a more general solution.

### Tests

### Changesets


Rich-Harris:
I remember this being finicky when I last looked into it, but: could we handle the SPA initialisation case differently to how we handle hydration and navigation, and never fall back to the server in that case? That way we don't need to worry about infinite loops, or keeping track in `localStorage` or whatever (which has caveats, e.g. if you have the right privacy settings in Firefox you can't use it).

dummdidumm:
That could be another possible solution. We would need to make sure it works for cases where only part of your app is an SPA, too, not sure how that would look like - possibly something like checking that when doing a root layout reload in the error case that the root layout tells you this is csr only.

dummdidumm:
Good point. Hydrate is false, and the router running means SPA mode. So therefore before falling back to the page we would check if we reload the url we are already on, and if so, prevent that. Essentially the existing SPA check but in a different place.

dummdidumm:
pushed up a WIP, code is more convoluted than I'd like and the error is incorrect in case of a root error (404 instead of whatever it should be instead), will look at this again tomorrow

dummdidumm:
Turns out the code was correct already, but I got confused by wrong error messages - turns out, it was/is buggy both on the client and the server:
- on the client it always returned "Internal Error" -> I fixed that here
- on the server, failing data requests (`__data.js?..`) are treated as "Not Found" because `event.routeId` is `null` in this case. Not sure what the right fix is here. Should we add the routeId of the page you're on, and if yes, can we find that out? Else we would need a status code handed to `handleError` after all.

dummdidumm:
Should be good to go now. Besides handling the SPA loop this also fixed some bugs around error handling.

Rich-Harris (approved):

PR #6923, opened 2022-09-20
only forward cookies for internal fetches - 

### LINKED ISSUE #6604: Cookie leak
### Describe the bug

It seems that cookies, acquired via fetch, in serverside requests are added to the response.

### Reproduction

Run a svelte app AND a php app on the same website. Use `fetch` in your svelte app to get something from a php website. If this fetch runs server-side the PHPSESSID from the fetch response will be added to the svelte response to the browser. Overwriting the PHPSESSID in the browser.

Since my svelte kit app does not use cookies at present, this was easily fixed with this 

```javascript
    response.headers.delete('set-cookie');
    return response;
}
```

### Logs

_No response_

### System Info

```shell
System:
    OS: Linux 5.4 Ubuntu 20.04.5 LTS (Focal Fossa)
    CPU: (128) x64 AMD EPYC 7742 64-Core Processor
    Memory: 36.57 GB / 78.07 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 16.14.0 - ~/.asdf/installs/nodejs/16.14.0/bin/node
    npm: 8.19.0 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 105.0.5195.102
    Firefox: 104.0
```

### Severity

### Additional Information

_No response_

Rich-Harris:
closes #6604.

I'm not sure if we also need to handle the case where `www.domain.com` is requesting data from `domain.com`, or where the external origin sets a third party cookie?

### Tests

### Changesets


dummdidumm (approved):
Re `domain` vs `www.domain` - I vaguely remember that these are treated differently, so this should be correct.
Re third party cookies: How can these be detected? From a quick reading it sounds like they should be forwarded since they are not related to your domain.

Approving to get you unblocked.

oetiker:
thank you!

PR #6924, opened 2022-09-20
[fix] tighten up navigation and invalidation logic

### LINKED ISSUE #5305: Abort Stale Invalidation Request
### Describe the problem

Consider the scenario where several [invalidate requests](https://kit.svelte.dev/docs/modules#$app-navigation-invalidate) to the same dependency are triggered in rapid succession (say because of an updating query parameter, e.g. `?search=foo`). 

The order in which these requests return from the server is not guaranteed, which means it's possible that the last resolving revalidation will be stale.

I wonder if we allow an optional abort signal, where `AbortSignal` implements the canonical [Abort API](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal). 

```typescript
function invalidate(
  dependency: string | ((href: string) => boolean)
  options?: { signal?: AbortSignal } 
): Promise<void>;
```

### Alternatives considered

Sveltekit's invalidate function could manage this automatically, but I worry there are scenarios where one might need multiple inflight invalidation requests. 

### Importance

would make my life easier

### Additional Information

As always, thanks so much for all the great work! Sveltekit is such a gamechanger for NYTs!

### LINKED ISSUE #6844: Invalidate all + redirects not working as expected.
### Describe the bug

discord thread: https://discord.com/channels/457912077277855764/1020064864066093196

Hey all - I think i ran into a bug or unexpected behavior that i can't wrap my head around.

The app is structured so that the root `/` acts as a central "router" to determine current user's state and redirect them to appropriate landing pages.

The child pages also have their own redirection logic, and if user somehow got to the page directly, user is redirected back to the router `/`.

There is a notion of "profiles" and user can switch between the profiles using a switcher component (think AWS, switching between profiles with different permissions). 

This component should 
`set the profile id in cookie from client side` => `invalidate/rerender the current page (client sided, via invalidateAll())` => `if the current page has any redirection logic, redirect with new profile; if page can render with new profile, render with new data, etc.`

The problem I am running into is, I call `invalidateAll()`, invalidation/rerunning nested layout code runs as expected (with the new profile id value), but at the very end of redirects, the `profile id` is seen as the previous value (which causes the "redirect loop" error, even when there really isn't a loop). 

### Reproduction

https://github.com/smblee/sveltekit-invalidate-all-bug

I reproduced it as close to my current repo as possible.

Say there are the following pages `/` (router, doesn't actually render), `notset` (ONLY when no profile set),`signup` (ONLY when profile is not-onboarded),  `onboarded-home` (ONLY when profile is onboarded).

try going from `Set not onboarded profile` => `Set onboarded profile` or the other way around. (it will always fail on the first try, in one direction).

you will fail with the following log trace: (
it can seem a bit overwhelming, but it's pretty simple. I basically logged the location and current profile id value at that time in code. and any redirection that happens)

```
+layout |  not-onboarded-id
(student)/+layout |  not-onboarded-id
(student)/signup/+page |  not-onboarded-id
Selector.svelte:11 ---------------------------
Selector.svelte:12 SETTING PROFILE: START
Selector.svelte:13 prev: not-onboarded-id
Selector.svelte:14 expected: onboarded-id
Selector.svelte:17 SETTING PROFILE: SET COOKIE
Selector.svelte:18 cookie: profileId=onboarded-id
Selector.svelte:19 SETTING PROFILE: INVALIDATE ALL START
+layout |  onboarded-id
(student)/+layout |  onboarded-id
(student)/signup/+page |  onboarded-id
(student)/signup/+page | redirecting to /
Selector.svelte:21 SETTING PROFILE: INVALIDATE ALL ENDED
Selector.svelte:22 ---------------------------
+layout |  onboarded-id
+page |  onboarded-id
+page | redirecting to /onboarded-home
(student)/(onboarded)/+layout |  not-onboarded-id                   <------ should be `onboarded-id`
+(student)/(onboarded)/+layout | redirecting to /
+page |  not-onboarded-id                               <------ should be `onboarded-id`
+page | redirecting to /signup
+layout |  onboarded-id

Error: Redirect loop...
```

### Logs

_No response_

### System Info

```shell
System:
    OS: macOS 12.5
    CPU: (10) arm64 Apple M1 Pro
    Memory: 109.05 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.1 - ~/.nvm/versions/node/v16.15.1/bin/node
    npm: 8.11.0 - ~/.nvm/versions/node/v16.15.1/bin/npm
  Browsers:
    Chrome: 105.0.5195.125
    Safari: 15.6
  npmPackages:
    svelte: ^3.44.0 => 3.50.1
    vite: ^3.1.0 => 3.1.1
```

### Severity

blocking an upgrade

### Additional Information

_No response_

### LINKED ISSUE #6902: Invalidate doesn't work while prefetching
### Describe the bug

Calling `invalidate` or `invalidateAll` doesn't behave as expected when a prefetch is happening. `load_route` returns early in this case, swallowing the `invalidate` calls.

### Reproduction

This thread has more info: https://discord.com/channels/457912077277855764/1013838400601731152

### Logs

_No response_

### System Info

```shell
Irrelevant
```

### Severity

annoyance

### Additional Information

In general I think we need to have some more stateful logic around invalidation - how they interact with redirects etc.

dummdidumm:
- Fixes #6844, invalidation is now only reset after navigation settled
- Fixes #5305, newer invalidations are no longer swalloed by ongoing older ones
- Fixes #6902, the pending prefetch is reset upon invalidation

No tests yet because a) my brain is fried from thinking about all this b) want to make sure first these are the right fixes

### Tests

### Changesets


packages/kit/src/runtime/client/client.js
```diff
@@ -91,18 +91,13 @@ export function create_client({ target, base, trailing_slash }) {
...
 		error: null,
-		session_id: 0,
 		// @ts-ignore - we need the initial value to be null
 		url: null
 	};
 
 	let started = false;
 	let autoscroll = true;
 	let updating = false;
-	let session_id = 1;
```
dummdidumm, line 102: Not sure what this was for, seems unused so I removed it

packages/kit/src/runtime/client/client.js
```diff
@@ -208,7 +203,10 @@ export function create_client({ target, base, trailing_slash }) {
 	 * @param {() => void} [callback]
 	 */
 	async function update(intent, url, redirect_chain, opts, callback) {
-		const current_token = (token = {});
+		// Do not update the token during a redirect, it belongs to a ongoing navigation; if a new one
+		// is started, it will should take precedence. If we don't do this, a race condition could
```
Rich-Harris, line 207:
```suggestion
		// is started, it should take precedence. If we don't do this, a race condition could
```

packages/kit/src/runtime/client/client.js
Rich-Harris: Back in the day, updating `$session` would cause an invalidation of all `load` functions that used it. This was related to that, and is no longer necessary

Rich-Harris:
This seems like the right direction, yeah

packages/kit/src/runtime/client/client.js
```diff
@@ -241,8 +241,6 @@ export function create_client({ target, base, trailing_slash }) {
 		// abort if user navigated during update
 		if (token !== current_token) return false;
```
Rich-Harris, line 242: i think we can get rid of `current_token` now?

```suggestion
		if (token !== nav_token) return false;
```

Rich-Harris (approved):
minor suggestion inline but LGTM 👍

PR #6925, opened 2022-09-20
add cookies.serialize function and always set headers/cookies

david-plugge:
Successor to https://github.com/sveltejs/kit/pull/6888

Closes https://github.com/sveltejs/kit/issues/6735

- fixes a bug where cookies and headers were only set when the request matched a route
- implements `cookies.serialize`

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -236,17 +236,6 @@ export async function respond(request, options, state) {
 					throw new Error('This should never happen');
 				}
 
-				if (!is_data_request) {
-					// we only want to set cookies on __data.js requests, we don't
-					// want to cache stuff erroneously etc
-					for (const key in headers) {
-						const value = headers[key];
-						response.headers.set(key, /** @type {string} */ (value));
-					}
-				}
-
-				add_cookies_to_headers(response.headers, Array.from(new_cookies.values()));
-
 				return response;
 			}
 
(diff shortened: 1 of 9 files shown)


david-plugge:
While working on this I noticed here:
https://github.com/sveltejs/kit/blob/5a54eb4af542a4b0c3b4c589c3df3ea520f5ca00/packages/kit/src/runtime/server/cookie.js#L40
the cookie headers is parsed on every `cookies.get` call. I think the parsed cookies should be cached

packages/kit/src/runtime/server/index.js
```diff
@@ -294,7 +283,19 @@ export async function respond(request, options, state) {
 	try {
 		const response = await options.hooks.handle({
 			event,
-			resolve,
+			resolve: (event, opts) =>
+				resolve(event, opts).then((response) => {
```
dummdidumm, line 287: Just from looking at this code I don't quite understand what the difference between this and the previous version is - could we maybe add a small explanatory comment? Is it that `state.prerendering?.fallback` wouldn't get the correct cookies/headers?

david-plugge: Headers and cookies were only applied if there was a valid route. So if you used `setHeaders` or `cookies.set` in handle they had no effect

Rich-Harris: It's so that cookies and headers get applied regardless of whether the request happened to match a route (i.e. you call `cookies.set` in `handle` or `src/routes/+layout.server.js` and it still gets applied to a 404 page created with `respond_with_error`). It _will_ apply to fallback responses but that will have no effect, since the fallback response is only ever written out as HTML.

Added a comment

Rich-Harris: jinx

dummdidumm (approved):

Rich-Harris:
thank you!

PR #6930, opened 2022-09-20
Fix `changed.params` tracking

### LINKED ISSUE #5004: Load requires parameter to exist in order to be tracked
### Describe the bug

Sveltekit requires the url parameter to be defined in order for it's usage to be tracked on the load function

I am guessing because it sets a getter tracking the parameter usage for all _defined_ parameters, but if a parameter is not defined, it won't track the parameter being used, and fails to re-run if it eventually gets defined without any other parameter changing

this is not inherently an issue for normal pages, but for layouts this issue is apparent

### Reproduction

root \_\_layout logging when Load runs
have a route `[first]` and another route `[first]/[second]`
travel to route `[first]/[second]`
- In this scenario, changing the `second` parameter causes a re-run on the root \_\_layout

travel to route `[first]` and back to `[first]/[second]` (the `first` parameter must not change)
- now changing the `second` parameter won't cause a re-run on the root \_\_layout, because changing to `[first]` made it not track the `second` parameter anymore

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (8) x64 Intel(R) Core(TM) i5-9300H CPU @ 2.40GHz    
    Memory: 2.83 GB / 11.86 GB
  Binaries:
    Node: 16.14.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.3.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (101.0.1210.47)
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.42
    @sveltejs/kit: next => 1.0.0-next.330
    svelte: ^3.44.0 => 3.48.0
```

### Severity

annoyance

### Additional Information

_No response_

Rich-Harris:
Fixes #5004. Leaving this here until I have sufficient motivation to tidy it up and add tests

### Tests

### Changesets


packages/kit/src/runtime/client/client.js
```diff
@@ -707,6 +695,16 @@ export function create_client({ target, base, trailing_slash }) {
 
 		const invalid_server_nodes = loaders.reduce((acc, loader, i) => {
 			const previous = current.branch[i];
+
+			const changed = current.url && {
+				url: id !== current.url.pathname + current.url.search,
+				params: previous?.server?.uses.params
+					? Array.from(previous.server.uses.params).filter(
+							(key) => current.params[key] !== params[key]
+					  )
```
Rich-Harris, line 704: hang on, this is daft — this logic belongs inside `has_changed`

packages/kit/test/apps/basics/src/routes/load/invalidation/params/+layout.svelte
```diff
@@ -0,0 +1,14 @@
...
+	/** @type {import('./$types').LayoutData} */
+	export let data;
+</script>
+
+<slot />
+
+<pre>{JSON.stringify({
+		a: data.a,
+		b: data.b
+	})}</pre>
```
dummdidumm, line 11:
```suggestion
<pre>{JSON.stringify(data)}</pre>
```

dummdidumm (approved):
I love how as a side effect the code also got smaller

packages/kit/test/apps/basics/src/routes/load/invalidation/params/+layout.svelte
Rich-Harris: it includes a bunch of stuff from the root layout if we do that

dummdidumm: Oh right, I forgot about the merge thing 👍

PR #6887, opened 2022-09-19
[fix] don't strip type="application/.." tags

dummdidumm:
Fixes #6862

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [x] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #6932, opened 2022-09-21
Quick check for undefined s4

niktek:

### Tests

### Changesets

As commented on #6887 the stripping fails for our project https://github.com/Brain-Bones/skeleton. We are not doing anything with application/ that that ticket is trying to remedy, but I can't see a way to figure out which file is making this fail.  This is a simple fix that makes packaging work for us again.

### DIFF
--- a/.changeset/smart-forks-double.md
+++ b/.changeset/smart-forks-double.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/package': patch
+---
+
+fix check for undefined on application/ strip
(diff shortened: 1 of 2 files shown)


dummdidumm:
Thank you - sorry for the oversight in the original PR. Could you add a changeset (run `npx changeset` in the monorepo root and follow the prompts)? Then we're good to merge.

niktek:
Hi,

Actually - please don't merge this yet.  I think it might be affecting the resultant tarballs (is this even possible ?).

I'm getting ` EACCES  EACCES: permission denied, open '/Users/nik/Library/pnpm/store/v3/file+brainandbones-skeleton-0.37.32.tgz/tarball-integrity'` when trying to install any generated package now.

Investigating further.

dummdidumm (approved):
Alright, merging this - the tarball thing sounds really weird and maybe unrelated, I can't imagine why such a code change would create a broken tarball.

PR #6936, opened 2022-09-21
Respond with status code 413 if request body is too large

### LINKED ISSUE #6920: Respond with status code 413 if request body is too large
### Describe the bug

Yesterday a few people in the discord tried to help someone who got an error 400 upon requesting an endpoint.
We needed almost an hour to find the problem because he only saw "400 Bad request" and not the message of the thrown `Error` (idk why).

If the response status code had been 413, we would have found the issue much earlier.

### Reproduction

Can not reproduce on stackblitz
https://replit.com/@PehGeh/GreenMutedComma
- click on "Upload a large file"

### Logs

_No response_

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (8) x64 AMD EPYC 7B12
    Memory: 37.87 GB / 62.80 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.2.0 - /nix/store/q3hh5viyhr3gvfrn6yakbd59gyna0v05-nodejs-18.2.0/bin/node
    Yarn: 1.22.18 - /nix/store/n7d0ibna8751b6kfj6i4fkanbdix8h7a-yarn-1.22.18/bin/yarn
    npm: 8.9.0 - /nix/store/q3hh5viyhr3gvfrn6yakbd59gyna0v05-nodejs-18.2.0/bin/npm
```

### Severity

annoyance

### Additional Information

https://github.com/sveltejs/kit/blob/ba168b0ff57caeb45a7b3185a97892ce745abec4/packages/adapter-node/src/handler.js#L51-L61

It seems like it would be enough to set the `status` property on the thrown errors to 413.

repsac-by:
Fixes #6920

### Tests

### Changesets

### DIFF
--- a/packages/adapter-node/src/handler.js
+++ b/packages/adapter-node/src/handler.js
@@ -56,7 +56,7 @@ const ssr = async (req, res) => {
 		});
 	} catch (err) {
 		res.statusCode = err.status || 400;
-		res.end(err.reason || 'Invalid request body');
+		res.end(err.message || err.toString() || 'Invalid request body');
 		return;
 	}
 
--- a/packages/adapter-node/src/handler.js
+++ b/packages/adapter-node/src/handler.js
@@ -56,7 +56,7 @@ const ssr = async (req, res) => {
 		});
 	} catch (err) {
 		res.statusCode = err.status || 400;
-		res.end(err.message || err.toString() || 'Invalid request body');
+		res.end('Invalid request body');
 		return;
 	}
 
--- a/packages/adapter-vercel/files/serverless.js
+++ b/packages/adapter-vercel/files/serverless.js
@@ -23,7 +23,7 @@ export default async (req, res) => {
 		request = await getRequest({ base: `https://${req.headers.host}`, request: req });
 	} catch (err) {
 		res.statusCode = err.status || 400;
-		return res.end(err.reason || 'Invalid request body');
+		return res.end(err.reason || err.toString() || 'Invalid request body');
 	}
 
 	setResponse(
--- a/packages/adapter-vercel/files/serverless.js
+++ b/packages/adapter-vercel/files/serverless.js
@@ -23,7 +23,7 @@ export default async (req, res) => {
 		request = await getRequest({ base: `https://${req.headers.host}`, request: req });
 	} catch (err) {
 		res.statusCode = err.status || 400;
-		return res.end(err.reason || err.toString() || 'Invalid request body');
+		return res.end('Invalid request body');
 	}
 
 	setResponse(
--- a/packages/kit/src/exports/node/index.js
+++ b/packages/kit/src/exports/node/index.js
@@ -1,4 +1,5 @@
 import * as set_cookie_parser from 'set-cookie-parser';
+import { HttpError } from '../../runtime/control.js';
 
 /**
  * @param {import('http').IncomingMessage} req
@@ -27,7 +28,8 @@ function get_raw_body(req, body_size_limit) {
 		if (!length) {
 			length = body_size_limit;
 		} else if (length > body_size_limit) {
-			throw new Error(
+			throw new HttpError(
+				413,
 				`Received content-length of ${length}, but only accept up to ${body_size_limit} bytes.`
 			);
 		}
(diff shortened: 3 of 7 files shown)


packages/kit/src/exports/node/index.js
```diff
@@ -27,7 +28,8 @@ function get_raw_body(req, body_size_limit) {
 		if (!length) {
 			length = body_size_limit;
 		} else if (length > body_size_limit) {
-			throw new Error(
+			throw new HttpError(
```
dummdidumm, line 31: You can use the `error` method here instead (prefered way)

packages/kit/src/exports/node/index.js
```diff
@@ -58,8 +61,10 @@ function get_raw_body(req, body_size_limit) {
 
 				size += chunk.length;
 				if (size > length) {
-					req.destroy(
-						new Error(
+					cancelled = true;
+					controller.error(
```
dummdidumm, line 65: I'm not well-versed with the details of this code section - could you explain briefly why this change is necessary?

packages/adapter-vercel/files/serverless.js
```diff
@@ -23,7 +23,7 @@ export default async (req, res) => {
 		request = await getRequest({ base: `https://${req.headers.host}`, request: req });
 	} catch (err) {
 		res.statusCode = err.status || 400;
-		return res.end(err.reason || 'Invalid request body');
+		return res.end(err.reason || err.toString() || 'Invalid request body');
```
dummdidumm, line 26: In the other places this was changed to `err.reason`, why not here? Can we safely assume that `toString` as a method is always present on the error object?

In general (again, not well-versed, just asking questions): Which errors end up in here, which would have `reason` set? Is this just an oversight from previous iterations?

packages/kit/src/exports/node/index.js
repsac-by: This returns an error in the RequestHandler and from there the status code gets into the Response, otherwise the connection is simply terminated without a response. `cancelled = true` is needed so that `controller.close()` does not occur in the `req.on('end', () => ...)`  because calling `close()` on a closed controller causes an unhandled exception.

packages/adapter-vercel/files/serverless.js
repsac-by: I myself don't understand where this `reason` can come from, but I don't know anything about this adapter, so I decided to leave everything as it is. In my opinion, all objects have toString() except Object.create(null).  But nevertheless it is possible to do so `err.toString?.()`

 I think it's just an oversight of previous iterations.

packages/adapter-node/src/handler.js
```diff
@@ -56,7 +56,7 @@ const ssr = async (req, res) => {
 		});
 	} catch (err) {
 		res.statusCode = err.status || 400;
-		res.end(err.reason || 'Invalid request body');
+		res.end(err.message || err.toString() || 'Invalid request body');
```
Rich-Harris, line 59: per https://github.com/sveltejs/kit/pull/6936/files#r976600055

```suggestion
		res.end('Invalid request body');
```

packages/kit/src/exports/vite/dev/index.js
```diff
@@ -397,7 +397,7 @@ export async function dev(vite, vite_config, svelte_config) {
 					});
 				} catch (/** @type {any} */ err) {
 					res.statusCode = err.status || 400;
-					return res.end(err.message || 'Invalid request body');
+					return res.end(err.message || err.toString() || 'Invalid request body');
```
Rich-Harris, line 400:
```suggestion
					return res.end('Invalid request body');
```

packages/kit/src/exports/vite/preview/index.js
```diff
@@ -137,7 +137,7 @@ export async function preview(vite, vite_config, svelte_config) {
 				});
 			} catch (/** @type {any} */ err) {
 				res.statusCode = err.status || 400;
-				return res.end(err.message || 'Invalid request body');
+				return res.end(err.message || err.toString() || 'Invalid request body');
```
Rich-Harris, line 140:
```suggestion
				return res.end('Invalid request body');
```

packages/adapter-vercel/files/serverless.js
repsac-by: Done

PR #6944, opened 2022-09-21
[fix] allow missing routes folder

### LINKED ISSUE #6921: Allow the `routes` folder to be empty
### Describe the problem

When there are no routes present in the `routes` folder, the following error is thrown: 
```
error when starting dev server:
Error: The filesystem router API has changed, see https://github.com/sveltejs/kit/discussions/5774 for details
    at create_routes_and_nodes (file:///P:/weather-apps-api/node_modules/.pnpm/@sveltejs+kit@1.0.0-next.490_svelte@3.50.1+vite@3.1.3/node_modules/@sveltejs/kit/src/core/sync/create_manifest_data/index.js:229:11)
```

Allowing the `routes` folder to be empty (when the `static` folder has content in it?)

### Alternatives considered

 To work around this annoyance, I created an empty `+page.svelte` component in the routes folder.

### Importance

nice to have

### Additional Information

_No response_

dummdidumm:
Fixes #6921

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/sync/create_manifest_data/index.spec.js
+++ b/packages/kit/src/core/sync/create_manifest_data/index.spec.js
@@ -173,8 +173,16 @@ test('creates routes with layout', () => {
 
 test('succeeds when routes does not exist', () => {
 	const { nodes, routes } = create('samples/basic/routes');
-	assert.equal(nodes, []);
-	assert.equal(routes, []);
+	assert.equal(nodes.map(simplify_node), [
+		{ component: 'layout.svelte' },
+		{ component: 'error.svelte' }
+	]);
+	assert.equal(routes.map(simplify_route), [
+		{
+			id: '',
+			pattern: '/^$/'
+		}
+	]);
 });
 
 // TODO some characters will need to be URL-encoded in the filename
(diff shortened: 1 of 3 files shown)


Rich-Harris:
errr i guess i rescind my approval until the tests pass 😅

packages/kit/src/core/sync/create_manifest_data/index.spec.js
```diff
@@ -173,8 +173,16 @@ test('creates routes with layout', () => {
...
+	assert.equal(nodes.map(simplify_node), [
+		{ component: 'layout.svelte' },
+		{ component: 'error.svelte' }
+	]);
+	assert.equal(routes.map(simplify_route), [
+		{
+			id: '',
+			pattern: '/^$/'
+		}
+	]);
```
dummdidumm, line 185: The test shows that nodes for the root layout and error are created. This means that the running code later on can safely assume these to be always present. Before that, the manifest was generated without these, causing `dev` and `build` to fail with "cannot read properties of nodes[0].component" or sth like that. Adding the default nodes even in case of a missing routes folder seems like the better option, even if it seems somewhat strange at first. I'll add a comment to the manifest about this.

PR #6947, opened 2022-09-21
see if this makes test less flaky

Rich-Harris:
this is currently the flakiest test. not really sure why it's failing, but maybe `waitForResponse` will help

### DIFF
--- a/packages/kit/test/apps/basics/test/client.test.js
+++ b/packages/kit/test/apps/basics/test/client.test.js
@@ -904,18 +904,9 @@ test('+server.js next to +page.svelte works', async ({ page }) => {
 	await page.goto('/routing/endpoint-next-to-page');
 	expect(await page.textContent('p')).toBe('Hi');
 
-	await page.click('button:has-text("GET")');
-	await expect(page.locator('pre')).toHaveText('GET');
-
-	await page.click('button:has-text("PUT")');
-	await expect(page.locator('pre')).toHaveText('PUT');
-
-	await page.click('button:has-text("PATCH")');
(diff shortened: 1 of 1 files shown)


packages/kit/test/apps/basics/test/client.test.js
```diff
@@ -904,18 +904,9 @@ test('+server.js next to +page.svelte works', async ({ page }) => {
...
-
-	await page.click('button:has-text("POST")');
-	await expect(page.locator('pre')).toHaveText('POST');
-
-	await page.click('button:has-text("DELETE")');
-	await expect(page.locator('pre')).toHaveText('DELETE');
+	for (const method of ['GET', 'PUT', 'PATCH', 'POST', 'DELETE']) {
+		page.click(`button:has-text("${method}")`);
+		await page.waitForResponse('/routing/endpoint-next-to-page');
+		expect(await page.textContent('pre')).toBe(method);
```
dummdidumm, line 910: This should make it even less flaky, as the page locator will try again until the text matches - which is also why I'm really puzzled that this test is flaky.
```suggestion
		await expect(page.locator('pre')).toHaveText(method);
```

Rich-Harris:
Going to merge, even though the tests timed out on Firefox, as the test in question is passing much more reliably now

packages/kit/test/apps/basics/test/client.test.js
dummdidumm: I don't think you can fix the underlying race condition. The browser may not have been repainted when the check is done, the "network idle" request may happen a little too soon

PR #5771, opened 2022-07-31
[fix] update CWD before running postinstall sync

Conduitry:
Fixes #5769, hopefully. It's really hard to know this sort of thing for sure until we merge and publish.

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #6949, opened 2022-09-21
run svelte-kit sync in all workspace directories

### LINKED ISSUE #6617: The postinstall script for a SvelteKit project runs in the root of a monorepo resulting in failure to generate the .svelte-kit folder
### Describe the bug

My .svelte-kit folder wasn't getting generated on initial npm install from the root of my monorepo, so I dug in and saw that when the @sveltejs/kit postinstall script was running, it's trying to access the svelte.config.js from the root rather than the project that has the dependency.

### Reproduction

Run "npm install --foreground-scripts" from the root of the following repo to see the errors about a missing svelte.config.js:

https://github.com/seantimm/sveltekit-postinstall-problem

### Logs

```shell
> @sveltejs/kit@1.0.0-next.474 postinstall
> node svelte-kit.js sync

Missing /projects/postinstall-problem/svelte.config.js - skipping
```

### System Info

```shell
System:
    OS: macOS 12.5.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 2.74 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - ~/.volta/tools/image/node/16.16.0/bin/node
    Yarn: 1.22.18 - ~/.volta/tools/image/yarn/1.22.18/bin/yarn
    npm: 8.15.0 - ~/.volta/tools/image/npm/8.15.0/bin/npm
    Watchman: 2022.07.04.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 105.0.5195.102
    Firefox Developer Edition: 99.0
    Safari: 15.6.1
```

### Severity

annoyance

### Additional Information

_No response_

Rich-Harris:
Closes #6617. In the case where `@sveltejs/kit` is installed to an npm workspace root, we run `svelte-kit sync` in all the directories specified in `pkg.workspaces`, taking account of things like `./packages/*`. (Though I've just read https://docs.npmjs.com/cli/v7/configuring-npm/package-json#workspaces and it seems that more complex globbing patterns are allowed, so this probably needs to use a globbing library.)

Difficult to add tests for this, so I didn't.

### Tests

### Changesets


packages/kit/src/cli.js
```diff
@@ -41,10 +41,6 @@ prog
 			return;
 		}
 
-		if (event === 'postinstall' && process.env.INIT_CWD) {
-			process.chdir(process.env.INIT_CWD);
-		}
-
```
Conduitry, line 47: We don't need an analogue of this (from #5771) anymore?

Rich-Harris: err.... yes. yes we do

packages/kit/postinstall.js
```diff
@@ -0,0 +1,27 @@
...
+	}
+} else {
+	directories.push(cwd);
+}
+
+for (const cwd of directories) {
+	if (fs.existsSync(path.join(cwd, 'svelte.config.js'))) {
+		process.chdir(cwd);
+		const config = await load_config();
+		await sync.all(config, 'development');
```
dummdidumm, line 25: While we're at it: I noticed that `npm install ` fails when you have an invalid `svelte.config.js`. This makes it very inconvenient for (for example) people upgrading from previous versions where some configs are now obsolete - for example, you update to the new form actions and used method overrides previously, then you first need to remove that from the config before `npm install`. It would probably be better to only show a warning in this case and continue installing the package. The error will still prevent you from actually starting the app - which is good - but you get a successful install - which is also good.

```suggestion
		try {
			const config = await load_config();
			await sync.all(config, 'development');
		} catch (e) {
			console.log('Error while trying to sync SvelteKit config');
			console.log(e);
		}
```

dummdidumm (approved):
Looks good, but I have one related improvement suggestion

dominikg:
What about other monorepo/workspace solutions (pnpm, yarn, lerna?). Do they have similar issues? At least pnpm uses a different way to declare workspace packages

Does `sync` validate that it is called on a package using svelte-kit? If not the current implementation in this PR could lead to problems in a workspace that mix kit and other svelte projects.

PR #6962, opened 2022-09-22
[Fix] broken assets path

### LINKED ISSUE #6326: Relative asset paths broke usage of imagetools
### Describe the bug

The relative asset paths for the static adapter do not work when I use vite imagetools.

The resulting paths are always starting with `./_app/...`  instead of `/_app/...`. So the images are only working on the root path and don't load on any subpage.

I'm not sure whether this has to be adopted in image tools or if there should be an option in svelte kit to disable that behavior. Maybe both. In general, I think relative paths are great and would love to have this feature work with imagetools.

Breaking PR in Svelte Kit: https://github.com/sveltejs/kit/pull/4250
Coresponding Issue in imagetools: https://github.com/JonasKruckenberg/imagetools/issues/367

### Reproduction

1. clone https://github.com/Amerlander/image-import-test and npm i, npm run build:
2. Open /about-Page and check the source below the headline.

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (8) x64 Intel(R) Xeon(R) CPU E3-1231 v3 @ 3.40GHz    
    Memory: 11.75 GB / 31.93 GB
  Binaries:
    Node: 16.14.2 - E:\Program Files\nodejs\node.EXE
    Yarn: 1.9.4 - C:\Program Files (x86)\Yarn\bin\yarn.CMD    
    npm: 8.5.0 - E:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 104.0.5112.102
    Edge: Spartan (44.19041.1266.0), Chromium (103.0.1264.37) 
  npmPackages:
    @sveltejs/adapter-static: ^1.0.0-next.39 => 1.0.0-next.39 
    @sveltejs/kit: next => 1.0.0-next.442 
    @sveltejs/vite-plugin-svelte: ^1.0.2 => 1.0.2 
    svelte: ^3.44.0 => 3.49.0 
    vite: ^3.0.4 => 3.0.9
```

### Severity

blocking an upgrade

### Additional Information

_No response_

frankmayer:
Fixes #6326 

This fixes broken usage of imagetools and possibly other vite plugins as well - when asset paths are used with the static adapter.
The breaking PR was #4250.

This fix introduces a new config option ~~`rootRelativeAssets`~~ `relativeAssets`, which defaults to ~~`true`~~ `false`, restoring the way things worked prior to the breaking PR, => namely: non-relative asset paths (or a "root-based asset path").

EDIT: Changed option name and logic

This approach may or may not be the correct one and may be up for discussion or be rejected completely. For that reason, I have not written a test for that, yet. But I have tested the fix successfully on a project which suffered that issue.

If this is a viable fix, I'll gladly complete this PR with anything else needed.

...and thank you all for this amazing project :1st_place_medal: 

---

### Tests

### Changesets


benmccann (changes requested):
this doesn't seem like the right fix to me for a couple reasons

one is that we should make it work with relative assets

the other is that this seems to also be affecting laravel, so it's quite possibly an issue in vite-imagetools: https://github.com/JonasKruckenberg/imagetools/issues/396

benmccann:
Can you check if https://github.com/JonasKruckenberg/imagetools/pull/400 would fix the issue you're facing?

frankmayer:
Thank you for your input. You were right. 
I tested https://github.com/JonasKruckenberg/imagetools/pull/400 with adapter-static and it seems to work now.

PR #6963, opened 2022-09-22
rename `App.PageError` to `App.Error`

### LINKED ISSUE #6905: `App.PageError` should probably be renamed `App.Error`
### Describe the bug

...because it also applies to non-page requests (e.g. if you throw an error in `handle` or a `+server.js` file, it will go through `handleError`, and if it's a non-page request it will be turned into a JSON response).

Annoyingly, this _would_ be a breaking change

### Reproduction

N/A

### Logs

_No response_

### System Info

```shell
N/A
```

### Severity

annoyance

### Additional Information

_No response_

Rich-Harris:
closes #6905. This is a breaking change, but I expect it to be very limited in impact.

### Tests

### Changesets

### DIFF
--- a/.changeset/kind-adults-drum.md
+++ b/.changeset/kind-adults-drum.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] rename App.PageError to App.Error
--- a/.changeset/kind-adults-drum.md
+++ b/.changeset/kind-adults-drum.md
@@ -1,5 +1,6 @@
 ---
 '@sveltejs/kit': patch
+'create-svelte': patch
 ---
 
 [breaking] rename App.PageError to App.Error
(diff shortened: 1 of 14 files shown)


dummdidumm:
Do you want to implement the "try to rename that inside `src/app.d.ts` during sync" thing I proposed in the issue, or did you decide against it?

Rich-Harris:
I agree with Dominik that it's probably a slippery slope. I think throwing an error telling the user to change it could be good, it just felt like a lot of effort when considering that we don't enforce a specific location for the user to declare those namespaces

dummdidumm:
We don't but I guess that 90% of the 5% of people that use this have it inside `src/app.d.ts`, which is where the starter templates put this.

dummdidumm (approved):

asmajlovicmars:
Not sure if it matters much, but even if PageError is commented in app.d.ts/js, the error will be produced: 
error when starting dev server:
Error: App.PageError has been renamed to App.Error — please update your src/app.d.ts

` 
// app.d.ts
// interface PageError {}
`

PR #6969, opened 2022-09-22
chdir before attempting postinstall

Rich-Harris:
I messed this up — it's reading the wrong `package.json`. gonna merge this and try it out again real quick (basically impossible to test without releasing)


gaarf:
FWIW, this breaks my monorepo setup (yarn + turborepo)

PR #6972, opened 2022-09-22
[fix] unwrap promises for load returns on the client

gtm-nayan:

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/utils/promises.js
+++ b/packages/kit/src/utils/promises.js
@@ -1,0 +1,11 @@
+/** @param {Record<string, any>} object */
+export async function unwrap_promises(object) {
+	/** @type {Record<string, any>} */
+	const unwrapped = {};
+
+	for (const key in object) {
+		unwrapped[key] = await object[key];
+	}
+
+	return unwrapped;
+}
--- a/packages/kit/src/utils/promises.js
+++ b/packages/kit/src/utils/promises.js
@@ -1,8 +1,8 @@
-/** 
+/**
  * Given an object, return a new object where all top level values are awaited
- * 
+ *
  * @param {Record<string, any>} object
- * @returns {Promise<Record<string, any>>} 
+ * @returns {Promise<Record<string, any>>}
  */
 export async function unwrap_promises(object) {
 	return Object.fromEntries(
--- a/packages/kit/src/utils/promises.js
+++ b/packages/kit/src/utils/promises.js
@@ -1,11 +1,11 @@
-/** @param {Record<string, any>} object */
+/** 
+ * Given an object, return a new object where all top level values are awaited
+ * 
+ * @param {Record<string, any>} object
+ * @returns {Promise<Record<string, any>>} 
+ */
 export async function unwrap_promises(object) {
-	/** @type {Record<string, any>} */
-	const unwrapped = {};
-
-	for (const key in object) {
-		unwrapped[key] = await object[key];
-	}
-
-	return unwrapped;
+	return Object.fromEntries(
+		await Promise.all(Object.entries(object).map(async ([key, value]) => [key, await value]))
+	);
 }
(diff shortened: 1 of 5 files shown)


Conduitry:
What happens right now if any of these aren't promises? How much time do we waste in that (common) case?

What happens if one of the promises rejects? What sort of errors does the user see in their application?

gtm-nayan:
> What happens right now if any of these aren't promises?

Without the PR it's equivalent to copying them over to the same key in a new object, with the PR the behavior for objects that aren't promises remain unchanged whereas objects that are promises are now resolved before being passed on.

> How much time do we waste in that (common) case? What happens if one of the promises rejects? What sort of errors does the user see in their application?

These I don't have a clear answer to either. The behaviour is already there on the server side for better or for the worse, and the intent for this PR is to make it consistent on the client.
I did change the implementation of unwrap_promises a bit, [ctx](https://discord.com/channels/457912077277855764/571775594002513921/1022525955304083476) but that could be done in a separate PR to keep things tidy if needed.

Rich-Harris:
> What happens if one of the promises rejects?

These are equivalent:

```js
export function load() {
  return {
    foo: Promise.reject(new Error('nope'))
  };
}
```

```js
  return {
    foo: await Promise.reject(new Error('nope'))
  };
}
```

> How much time do we waste in that (common) case?

If we're concerned about that we could do this sort of thing:

```js
/**
 * Given an object, return a new object where all top level values are awaited
 *
 * @param {Record<string, any>} object
 * @returns {Promise<Record<string, any>>}
 */
export async function unwrap_promises(object) {
  const awaited = {};
  const promises = [];

  for (const key in object) {
    if (typeof object[key]?.then === 'function') {
      promises.push(async () => {
        awaited[key] = await object[key];
      });
    } else {
      awaited[key] = object[key];
    }
  }
  
  await Promise.all(promises);
  return awaited;
}
```

Or even this, which just returns the original object if no top-level properties were promises:

```js
/**
 * Given an object, return a new object where all top level values are awaited
 *
 * @param {Record<string, any>} object
 * @returns {Promise<Record<string, any>>}
 */
export async function unwrap_promises(object) {
  for (const key in object) {
    if (typeof object[key]?.then === 'function') {
      const awaited = {};
      const promises = [];

      for (const key in object) {
        if (typeof object[key]?.then === 'function') {
          promises.push(async () => {
            awaited[key] = await object[key];
          });
        } else {
          awaited[key] = object[key];
        }
      }
      
      await Promise.all(promises);
      return awaited;
    }
  }

  return object;
}
```

Rich-Harris (approved):

Rich-Harris:
I'm inclined to think it's an unnecessary micro-optimisation, so I've approved this PR, but shout if you feel differently

Conduitry:
I _would_ say that this whole feature is sort of unnecessary and I would like to get rid of it, but then I am confronted with the fact that it was originally implemented in the framework itself as an `await` inside a `for` loop, and, uh, yeah this is probably a footgun. The question is whether enough people will read the part of the docs that say you can return promises to take advantage of it. The fact that this was completely broken on the client for so long makes me tend to think that this is not a well-used feature, and so this is mostly just causing a microscopic slowdown to everyone else.

In conclusion: Uh, um.

dummdidumm:
I agree with that sentiment. If I could undo it, I probably would. And yes, it being broken probably means noone ever used that feature.

packages/kit/src/utils/promises.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Given an object, return a new object where all top level values are awaited
+ *
+ * @param {Record<string, any>} object
+ * @returns {Promise<Record<string, any>>}
+ */
+export async function unwrap_promises(object) {
+	return Object.fromEntries(
+		await Promise.all(Object.entries(object).map(async ([key, value]) => [key, await value]))
+	);
```
Rich-Harris, line 10: my other suggested optimisation had a subtle bug — keys would be out of order. this feels nice and simple

```suggestion
	for (const key in object) {
		if (typeof object[key]?.then === 'function') {
			return Object.fromEntries(
				await Promise.all(Object.entries(object).map(async ([key, value]) => [key, await value]))
			);
		}
	}

	return object;
```

Issue #6568, opened 2022-09-04
prerender = true in /+layout.server.ts breaks the route after deployment to netlify / vercel


StephenGunn:
### Describe the bug

I found this bug in one of my bigger projects after migrating to the new routing configuration. It took me a while to pinpoint that it was the `prerender = true` setting in `/+layout.server.ts` that was causing the bug. The project runs fine locally using pnpm dev || pnpm build || pnpm preview. The project also has no problem building once deployed to Netlify / Vercel. It's only after publishing that it breaks.

I created a new project from scratch and was able to recreate the bug when deployed to both netlify and vercel.

- The repo for the scratch project is here: https://github.com/StephenGunn/layoutprerender

- The deployment on Netlify is here: https://glittering-mermaid-5b82d2.netlify.app

- The deployment on Vercel is here: https://layoutprerender.vercel.app

You'll be able to see the 500 error on both Netlify and Vercel using the above links. I also have examples of the exact same code but with prerendering not set and the routes work as intended.

I tried to make the code and example as clear as possible, let me know if you have any questions.

EDIT: prerender = true works on normal /+page.server.ts powered routes from my testing

### Reproduction

- clone https://github.com/StephenGunn/layoutprerender and deploy to netlify or vercel

### Logs

```shell
Netlify doesn't seem to log any error in the render.js function logs.

Vercel logs this: [GET] /episodes-3/__data.js?__invalid=nyn&__id=1

The error on the frontend is this:
500
Failed to fetch dynamically imported module: https://layoutprerender.vercel.app/episodes-3/__data.js?__invalid=nyn&__id=1
TypeError: Failed to fetch dynamically imported module: https://layoutprerender.vercel.app/episodes-3/__data.js?__invalid=nyn&__id=1
```

### System Info

```shell
Netlify / Vercel
```

### Severity

annoyance

### Additional Information

Marked as an annoyance because I can just turn off prerendering for my /+layout.server.ts and the site works.

But I would like to be able to turn this feature back on since it does slow down certain routes and puts added stress on my backend.

s3812497:
Hi Stephen,

What does the `ssr` `noExternal` config do in `vite.config.js`? [Link here](https://github.com/StephenGunn/layoutprerender/blob/4040b1692af8f7eae4bc5b6c1ccc8718f0b1af28/vite.config.ts#L6-L10).

Could this be the cause of the problem? (I've no idea, but just curious as I was looking through the code and comparing it to a new skeleton project).

[...]

PR #6974, opened 2022-09-22
throw an error if prerenderable routes are not prerendered

Rich-Harris:
Part of #6568 (but not all — still need to regenerate the manifest so that `prerender = true` routes cannot be dynamically server rendered at all during `preview`)

### Tests

### Changesets


documentation/docs/13-page-options.md
```diff
@@ -70,6 +70,15 @@ For _pages_, we skirt around this problem by writing `foo/index.html` instead of
 
 Note that this will disable client-side routing for any navigation from this page, regardless of whether the router is already active.
 
+#### Troubleshooting
+
+If you encounter an error like 'Route ... is marked as prerenderable, but was not prerendered' it's because the route in question (or a parent layout, if it's a page) has `export const prerender = true` but the page wasn't actually prerendered.
```
dummdidumm, line 75: Out of scope for this PR, but I think a dedicated page that lists all errors thrown by Kit would be nice. The errors thrown by Kit would then also have a link pointing to that site where you can read more about that error.

packages/kit/src/core/prerender/prerender.js
```diff
@@ -410,6 +416,14 @@ export async function prerender() {
 
 	await q.done();
 
+	for (const [route_id, prerender] of prerender_map) {
+		if (prerender === true && !prerendered_routes.has(route_id)) {
+			throw new Error(
+				`Route "${route_id}" is marked as prerenderable, but was not prerendered. See https://kit.svelte.dev/docs/page-options#prerender-troubleshooting for more info`
```
dummdidumm, line 422: Should we wait with throwing this error until we checked all routes? I don't know how often that is, but maybe people could get annoyed seing this trickle in one by one, having to rerun their build each time.
```
The following routes were marked as prerenderable, but were not prerender:
- a
- b
- ..
See ... for more info
```

dummdidumm (approved):

packages/kit/src/core/prerender/prerender.js
Rich-Harris: Yeah, that's way better. Updated

documentation/docs/13-page-options.md
dummdidumm, line 75:
```suggestion
If you encounter an error like 'The following routes were marked as prerenderable, but were not prerendered' it's because the route in question (or a parent layout, if it's a page) has `export const prerender = true` but the page wasn't actually prerendered.
```

PR #6975, opened 2022-09-22
[chore] fix flaky test

benmccann:
It looks like the test is still flaking out. I think it's because you could get the response, check the value, and then the value is updated. Writing it this way avoids that possibility

### DIFF
--- a/packages/kit/test/apps/basics/test/client.test.js
+++ b/packages/kit/test/apps/basics/test/client.test.js
@@ -905,8 +905,10 @@ test('+server.js next to +page.svelte works', async ({ page }) => {
 	expect(await page.textContent('p')).toBe('Hi');
 
 	for (const method of ['GET', 'PUT', 'PATCH', 'POST', 'DELETE']) {
-		page.click(`button:has-text("${method}")`);
-		await page.waitForResponse('/routing/endpoint-next-to-page');
-		expect(await page.textContent('pre')).toBe(method);
+		await page.click(`button:has-text("${method}")`);
+		await page.waitForFunction(
+			(method) => document.querySelector('pre').textContent === method,
+			method
+		);
 	}
 });


packages/kit/test/apps/basics/test/client.test.js
```diff
@@ -905,8 +905,10 @@ test('+server.js next to +page.svelte works', async ({ page }) => {
...
 
 	for (const method of ['GET', 'PUT', 'PATCH', 'POST', 'DELETE']) {
-		page.click(`button:has-text("${method}")`);
-		await page.waitForResponse('/routing/endpoint-next-to-page');
-		expect(await page.textContent('pre')).toBe(method);
+		await page.click(`button:has-text("${method}")`);
+		await page.waitForFunction(
+			(method) => document.querySelector('pre').textContent === method,
+			method
+		);
```
Rich-Harris, line 912: I'm surprised this reduces flakiness when prior to https://github.com/sveltejs/kit/pull/6947 it was using this, which should in theory be equivalent:

```suggestion
		await expect(page.locator('pre')).toHaveText(method);
```

Rich-Harris: unbelievable — making that change reintroduced flakiness. the hell. https://github.com/sveltejs/kit/actions/runs/3107914512/jobs/5036541318#step:8:397

packages/kit/test/apps/basics/test/client.test.js
```diff
@@ -905,8 +905,7 @@ test('+server.js next to +page.svelte works', async ({ page }) => {
 	expect(await page.textContent('p')).toBe('Hi');
 
 	for (const method of ['GET', 'PUT', 'PATCH', 'POST', 'DELETE']) {
-		page.click(`button:has-text("${method}")`);
-		await page.waitForResponse('/routing/endpoint-next-to-page');
-		expect(await page.textContent('pre')).toBe(method);
+		await page.click(`button:has-text("${method}")`);
+		await expect(page.locator('pre')).toHaveText(method);
```
dummdidumm, line 909: This is basically how it looked like prior to Rich's change, which is why I'm so confused why it would be flaky in the first place
https://github.com/sveltejs/kit/commit/2af2a9dba148ee8403ef5c163ef8827eb27a4800

packages/kit/test/apps/basics/test/client.test.js
dummdidumm: Yeah it's super strange, this is basically the way to un-flake it. Playwright bug..?

benmccann: maybe we need to set the `polling` option? https://playwright.dev/docs/api/class-page#page-wait-for-function

otherwise it will wait until there's a `requestAnimationFrame`. I'm not that familiar with `requestAnimationFrame`. Is it something that will happen periodically in the browser without animations running or do we need to fake a tick?
```
export function tick(/** @type {import('playwright-core').Page} */ page) {
	return page.evaluate(() => new Promise(requestAnimationFrame));
}
```

benmccann: This is wild. The page actually says "GET", but playwright is getting "loading" eventhough it's not on the page. This really does feel like a Playwright bug. I guess I'll try to see if I can reproduce it in a simpler project

Screenshot from 2022-09-23 08-14-46

Rich-Harris:
gonna revert to https://github.com/sveltejs/kit/pull/6975/commits/6d780b4d2d7a0ec4b496cd15590da8868ca3213c for now

Rich-Harris (approved):

PR #6976, opened 2022-09-22
tweak xss tests

### LINKED ISSUE #2648: Incorrect XSS test in #2597
### Describe the bug

The test added along with the XSS fix in #2597 is not correct. Besides the `pwned`/`pnwed` spelling discrepancy noted in a comment on there, there appear to be other issues as well, as I'm not getting the tests to fail properly when I revert the rest of the XSS fix.

### Reproduction

Revert the fix part of #2597 and run the tests.

### Logs

_No response_

### System Info

```shell
n/a
```

### Severity

### Additional Information

_No response_

Rich-Harris:
closes #2648, i think?


Conduitry:
@Rich-Harris I was looking again at #2597 to see what was getting newly escaped way back then, and also looking at the current `packages/kit/src/runtime/server/page/render.js` to see what's changed (hint: a lot), and one thing I did notice is that it looks like we might have opened up a new XSS hole recently.

https://github.com/sveltejs/kit/blob/3936b561c9b9028d109ac2a0f6aefb9757f670dc/packages/kit/src/runtime/server/page/render.js#L198

Depending on various things, this error message might contain text that's coming from the user (like a URL), which might have an improperly escaped `</script>`, which, etc. etc. I don't know exactly how I'd construct a vulnerable app, but it does seem like a good idea to only use `s`/`JSON.stringify` for things we're absolutely certain contain no text from the user.

Is there anything here in this object that it wouldn't be simpler to just run through `devalue`?

dominikg:
iirc #2648 says that if xss happens, the tests don't reach the `window.pwned` expect. Instead playwright fails at page load because the xss escape isn' perfect. The injected script is executed but after that a global js error happens on the page.

This PR improves the tests in case we correctly prevent xss, but as far as i can tell does not improve the fail.

Maybe there is a way to trap global errors in playwright?

Rich-Harris:
> Is there anything here in this object that it wouldn't be simpler to just run through `devalue`?

Yeah we could use `devalue` here. This object is the output of `handleError`, which means it defaults to

```js
{ "message": "Internal Error" }
```

and you'd have to work quite hard to make it dangerous. But using `devalue` would make it impossible rather than just hard.

> Instead playwright fails at page load because the xss escape isn' perfect.

I'm not sure I understand why, or how we'd change things to accommodate that?

Rich-Harris:
@dominikg bump — what would have to change in this PR before we can merge it?

dominikg:
Nothing I guess if we are ok with the fact that we get a browser error instead of a proper testcase fail.

Maybe the payload needs to be checked if the switch to devalue means it has to be adapted to be dangerous again if not properly escaped?

Rich-Harris:
Cool, will merge — I think any kind of error is fine as long as it alerts us to a regression

PR #6979, opened 2022-09-22
Sverdle

Rich-Harris:
Playable here: https://kit-git-sverdle-svelte.vercel.app/sverdle

* Closes https://github.com/sveltejs/kit/issues/1564
* Closes https://github.com/sveltejs/kit/pull/4266
* Closes https://github.com/sveltejs/kit/issues/4264

The `/todos` app in the default template is a nice idea, but it's a bit misleading. People understandably imagine that the back end runs locally and then are surprised by the latency.

This PR replaces it with a Wordle clone that keeps all the game logic and data on the server, and persists user state using cookies.

It's not 100% finished — there's a subtle bug with the keyboard showing some letters as wrong instead of right-letter-wrong-place, and it needs some instructions and probably some style tweaks. I'd also love to get some help on the a11y front (cc @geoffrich!)

### Tests

### Changesets


Rich-Harris:
(Also: we might think this thing is too complex to have in a demo app, in which case I'm certainly open to alternatives. The code can probably get simplified a bit though, it was written under conference deadline conditions.)

Conduitry:
One of my initial suggestions would be whether we can stick the word lists in some external package. That's a lot of stuff to ship with everyone using `create-svelte` whether or not they opt to use the Sverdle example.

Rich-Harris:
counterpoint: it's easier to delete the entire `sverdle` directory than to do that _and_ hunt around in `package.json` for unused packages. case in point: i forgot to remove the `cookie` and `@lukeed/uuid` packages even though they are surplus to requirements

gtm-nayan:
> counterpoint: it's easier to delete the entire `sverdle` directory than to do that and hunt around in package.json for unused packages. case in point: i forgot to remove the cookie and @lukeed/uuid packages even though they are surplus to requirements

Is the demo app supposed to be something people modify and build their own app upon? I had thought it was just something to learn from and then they would use the skeleton template when creating a new app.

How about downloading the words list from `kit.svelte.dev/somewhere` or `raw.githubusercontent....` and then cache it on disk when it first runs? Could be done using a vite-plugin or in create-svelte only if the demo template is selected, that way consumers of other templates wouldn't be affected.

martinbeentjes:
The templates are now a skeleton, a library skeleton and the todo app. Maybe to have Sverdle as a 'template' is misleading as it is complex. However, this is an useful application to look at. It answers my questions I had when I started with SvelteKit. 

I want to propose an enhanced create-svelte-app. Maybe with a bit more flexibility on the components to import. Implicitly this is just a simplified general template. For example, with the directories, and possible a README there? (or does that make it bloated?) So you have the structure of the full fledged SvelteKit app, but the cleanness of the skeleton.

dummdidumm:
Having more than two versions would be too much to choose as a newcomer ("what do I pick?") and maintain in my opinion. To be clear the Sverdle would replace the TODO example.

For me the question comes down to finding a good middle ground for the app that
- showcases most of the basic features
- is easy to dig into and understand
- is shiny, i.e. you like to browse it a few minutes, it's not just a "here's what you can do, look at the code, the app is boring"

bato3:
The TODO application is so good that everyone understands its idea. 

_Sverdle - WTF?_ First I need to find out what it is. And then only check how it works in SK.

**I agree with one thing:** TODO in its current form should disappear.

BTW:
---
I miss an example of logging in, storing data in the session, connecting to the database, `(group)`, `[slug]`, REST-API ...


I suggest that there are 2 applications to choose from:
- **very simple** - where TODO would be replaced with the form: "Enter your name, enter 5 here" (to show error handling)
- **advanced** - showing all the possibilities of SK. (Eg.: shop: list products, guest favourite (cookie), login, add to chart)


   - We should "create" design patterns to avoid security issues  with login: #6315, #6996.
   - "Deleting" the session made a lot of confusion. Suffice it to say that this mechanism is called differently in SK (and show how to use it)
   - You have to show that the database connection can be done in `hooks` (Before I accidentally found this solution, I was using `Custom Server` + second repo for API)
   - There should be an example for `(group)`, `[slug]` and resetting layouts.
   - There should be an example for using server.js, and #6997

Rich-Harris:
> Is the demo app supposed to be something people modify and build their own app upon? I had thought it was just something to learn from and then they would use the skeleton template when creating a new app.

The latter, though a lot of people do use the 'default' template (which should probably be renamed to 'demo') when creating repros etc. (That's probably our fault, since we publish https://github.com/sveltejs/kit-template-default but not https://github.com/sveltejs/kit-template-skeleton.)

> _Sverdle - WTF?_ First I need to find out what it is. And then only check how it works in SK.

When it's finished it'll have instructions and a more self-explanatory UI, which will hopefully address that

> I miss an example of logging in, storing data in the session, connecting to the database, `(group)`, `[slug]`, REST-API ...

I hear you, for sure. The trouble is that any not-completely-unrealistic demonstration of those features requires a back end, and then we're back at the same problem we had with `/todo` — we can either provide one for you a la api.svelte.dev, or we add a bunch of hoops you have to jump through if you want to actually deploy the thing. I definitely think there's a place for those sorts of examples (somewhere on learn.svelte.dev ideally, eventually) but it's probably not the demo app

packages/create-svelte/templates/default/src/routes/sverdle/+page.svelte
```diff
@@ -0,0 +1,262 @@
...
+	export let data: PageData;
+
+	/** @type {import('./$types').ActionData} */
+	export let form: ActionData;
+
+	$: i = data.answers.length;
+	$: won = data.answers.at(-1) === 'xxxxx';
+
+	/** @type {Record<string, 'exact' | 'close' | 'missing'>}*/
+	let keys: Record<string, 'exact' | 'close' | 'missing'> = {};
```
dummdidumm, line 18: It feels a little inconcistent that the states are written out for the keys but the guesses use the shortform.

packages/create-svelte/templates/default/src/routes/sverdle/+page.svelte
```diff
@@ -0,0 +1,262 @@
...
+		}
+	}
+</script>
+
+<div class="game">
+	<a class="how-to-play" href="/sverdle/how-to-play">How to play</a>
+
+	<form id="game" method="POST" action="?/enter" use:enhance>
+		{#each Array(6) as _, r}
+			{@const current = r === i}
```
dummdidumm, line 58: `row` would be more descriptive

packages/create-svelte/templates/default/src/routes/sverdle/+page.svelte
```diff
@@ -0,0 +1,262 @@
...
+
+			<div class="row" class:current class:illegal={current && form?.illegal}>
+				{#each Array(5) as _, c}
+					{@const answer = data.answers[r]?.[c] ?? '-'}
```
dummdidumm, line 62: `column` would be more descriptive

dummdidumm (commented):
This mostly looks good (left some comments inline) 👍 One question about the words: Could we reduce the possible (allowed) words so the files aren't that huge? Also, what are the allowed words for? Why can't we just compare against the chosen word?

I can't shake the feeling though (and I'm sorry, because you really put a lot of work into this) that it _may_ be too much for a demo app. I think it's super impressive to see how to make this work without JavaScript, but actually understanding the code takes some time and much of it is unrelated to SvelteKit concepts - which may not be obvious and then people might think "oh SvelteKit is too complex for me".

I'm wondering if a revamped version of the TODO app might be better (boooring, I know) - for example putting some fake login in front (appears in a popup with JS, else redirect to the login page) and using the same cookie mechanism that is shown in the Sverdle to safe the login state and the last X todos. Maybe we could do both - a simple TODO app for a basic auth flow and something everybody knows already, and a Sverdle to drop your jaw?

packages/create-svelte/templates/default/src/routes/sverdle/+page.svelte
Rich-Harris: It's basically because `exact` and `close` and `missing` are used as classnames (and only as classnames). I've moved this logic into `Keyboard.svelte` to (hopefully) make that clearer

Rich-Harris:
> Could we reduce the possible (allowed) words so the files aren't that huge? Also, what are the allowed words for? Why can't we just compare against the chosen word?

Not without making the game less fun to play! The `allowed` words are less common English words that are valid submissions, but won't ever be chosen as the answer.

> I can't shake the feeling though (and I'm sorry, because you really put a lot of work into this) that it _may_ be too much for a demo app. I think it's super impressive to see how to make this work without JavaScript, but actually understanding the code takes some time and much of it is unrelated to SvelteKit concepts - which may not be obvious and then people might think "oh SvelteKit is too complex for me".

Yeah, I get that. I've rewritten the app so that there's a single component with a single form. Even though we'd normally break something up into separate components to improve readability, in this case it had the opposite effect because (for example) the `enter` key inside `Keyboard.svelte` referenced the `<form>` in `+page.svelte`, but you'd only know that if you understood what the `form` and `id` attributes do in that context.

The new version is much easier to follow, if I do say so myself. The page is basically one screen's worth of script, another screen's worth of markup, then a bunch of styles. It's not _quite_ as simple as `/todos` but it's a lot more fun/unique and the use of cookies as a datastore feels much more appropriate in this context, so I reckon we can get away with it!

Tal500:
I think that this game is nice for demo, but not for initial template(the dictionary file isn't suitable for example).

I suggest to have some classical game that the user plays against the server(a simple AI player), like tic-tac-toe or 4 in a line. It avoids the dictionary file issue, while I'm sure that a simple AI mechanism can be coded/copied.

Additionally, I also think that the simple TODO app (or anything alike) is essential for template, because it demonstrates the CRUD ability, so I suggest to like this page in the template, and only add new ones.
Maybe a good idea for solving the remote latency feeling issue is to let the user decode, in some switch button, whether to use a local storage(e.g. by allowing the server to store the state in a cookie or by a in-memory storage) or a remote one.

BTW: It's always nice to have an indication for the user, that it's loading the result from server, e.g. by a simple loading line.

Rich-Harris:
> the dictionary file isn't suitable for example

It's fine! It's completely trivial compared to (for example) the contents of `node_modules`

> tic-tac-toe or 4 in a line

the trouble is those games are really boring :)

> it demonstrates the CRUD ability

I'm not sure what this means to be honest — SvelteKit doesn't have 'CRUD ability', it just has arbitrary actions. If we demonstrate CRUD we need a back end, which brings back the issue this PR was designed to address. A switchable back end adds complexity and makes the example sufficiently unrealistic that it starts to lose value.

> always nice to have an indication for the user, that it's loading the result from server

yeah, we could add something. not sure what the best UI would be

Tal500:
> > the dictionary file isn't suitable for example
> 
> It's fine! It's completely trivial compared to (for example) the contents of `node_modules`

OK

> > tic-tac-toe or 4 in a line
> 
> the trouble is those games are really boring :)

🧐🤨🙄 OK, matter of test. I just thought it's easier and more classic for a startpoint, in the logic of that the normal user doesn't use svelte-create to generate a game he likes, he goes to official sveltekit demos for it😅. Not sure if we need a game at all for a startpoint templats though, although I really would like to see these stuffs on demos.

> > it demonstrates the CRUD ability
> 
> I'm not sure what this means to be honest — SvelteKit doesn't have 'CRUD ability', it just has arbitrary actions. If we demonstrate CRUD we need a back end, which brings back the issue this PR was designed to address. A switchable back end adds complexity and makes the example sufficiently unrealistic that it starts to lose value.

I'm not good at word choosing here, I meant by "CRUD abilities" the meaning of "CRUD startpoint template", which is far more useful to the user. For example, I used this "TODO" generated route, for checking how Kit team chose to break the API everytime🤣. I believe that it will be useful, at least for me, to see "only the code I need to see" in the TODO route for remembering again the suggested CRUD structure in Kit.

dummdidumm:
> > Could we reduce the possible (allowed) words so the files aren't that huge? Also, what are the allowed words for? Why can't we just compare against the chosen word?
> 
> Not without making the game less fun to play! The `allowed` words are less common English words that are valid submissions, but won't ever be chosen as the answer.

I still don't get it. Why check against a list of allowed words? Why not do "ok you didn't hit any of the characters in this word -> wiggle"? Is it because you could have a valid word without any shared characters? How often does this happen? Is it worth keeping this for the additional thousands of words we need for that?

> The new version is much easier to follow, if I do say so myself. The page is basically one screen's worth of script, another screen's worth of markup, then a bunch of styles. It's not _quite_ as simple as `/todos` but it's a lot more fun/unique and the use of cookies as a datastore feels much more appropriate in this context, so I reckon we can get away with it!

Yeah it reads much better now (although the question stays if we are just too familiar already if it by now). Btw the "how to play" page needs some love 😄 Maybe a section on how it works under the hood would also be good, so that you can already grasp from a high level how it works without looking a the code. Something like "Under the hood this uses regular forms and cookies to implement the game. This means it even works without JavaScript!"

geoffrich:
I still need to do an a11y pass on this, but I can make a follow-up issue for that.

packages/create-svelte/scripts/build-templates.js
```diff
@@ -8,11 +8,18 @@ import { mkdirp, rimraf } from '../utils.js';
...
 	});
 
-	return prettier.format(transformed.code, {
+	// sucrase leaves invalid class fields intact
+	code = code.replace(/^\s*[a-z]+;$/gm, '');
+
+	// Prettier strips 'unnecessary' parens from .ts files, we need to hack them back in
+	code = code.replace(/(\/\*\* @type.+? \*\/) (.+?) \/\*\*\*\//g, '$1($2)');
+
+	return prettier.format(code, {
```
dummdidumm, line 22: I remember having the same problem in `prettier-plugin-svelte` (the parentheses thing) when using the `typescript` parser (https://github.com/sveltejs/prettier-plugin-svelte/issues/218), using the `babel-ts` parser fixed it. I wonder though why it doesn't work in this case, because the formatter uses the `babel` parser, which I would have expected to be a JS parser which should keep the parentheses.

Rich-Harris:
> I still don't get it. Why check against a list of allowed words? Why not do "ok you didn't hit any of the characters in this word -> wiggle"?

The wiggle happens if you enter an _invalid_ word, not if you enter a _wrong_ word

PR #6982, opened 2022-09-23
create-svelte: Skip prompts by passing args to CLI

JReinhold:
This PR makes it possible to skip some or all prompts when creating a new SvelteKit project with the `create-svelte` package. eg.:

```bash
npm create svelte@latest my-app \
  --name=my-new-app \
  --template=default \
  --types=checkjs \
  --prettier  \
  --no-eslint \
  --no-playwright
```

## The problem

In Storybook we have a system for creating new projects with the different frameworks programmatically, to continuously ensure Storybook works out of the box with all the frameworks. But that's a bit hard for us to do with SvelteKit, because `create-svelte` is an interactive experience and thus not easy to automate.

Instead of hacking around it internally in Storybook I thought it would be better to solve it for everyone. This is also useful in tutorial/guide situations where an author wants to simplify the learning path by removing prompts that a beginner might not understand.

## The solution

As per [the example in the `prompts` docs](https://github.com/terkelg/prompts#override), [`yargs`](https://github.com/yargs/yargs) is used to implicitly override any of the prompts if they are passed as named arguments. The implicitness is important here, because it means that if we add or change the prompts later they will automatically be overwritable by args.

I have _not_ documented this feature in the official docs because there didn't seem to be a good place to do it. I'm open to add it somewhere though if you think it makes sense.
It is however documented in the package's readme similarly to how the programmatic API is documented.

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -22,12 +22,14 @@ const { version } = JSON.parse(fs.readFileSync(new URL('package.json', import.me
 async function main() {
 	console.log(gray(`\ncreate-svelte version ${version}`));
 	console.log(disclaimer);
-	const args = await yargs(hideBin(process.argv)).argv;
+
+	const args = await yargs(hideBin(process.argv)).coerce('types', (arg) =>
+		arg === 'none' ? null : arg
+	).argv;
 	prompts.override({
 		...args,
 		dir: args._[0]?.toString()
 	});
-	console.log(args);
 
 	const { dir } = await prompts([
 		{
--- a/packages/create-svelte/bin.js
+++ b/packages/create-svelte/bin.js
@@ -5,6 +5,8 @@ import { bold, cyan, gray, green, red } from 'kleur/colors';
 import prompts from 'prompts';
 import { create } from './index.js';
 import { dist } from './utils.js';
+import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
 
 // prettier-ignore
 const disclaimer = `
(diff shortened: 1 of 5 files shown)


packages/create-svelte/bin.js
```diff
@@ -21,32 +23,33 @@ async function main() {
...
-				name: 'dir',
-				message: 'Where should we create your project?\n  (leave blank to use current directory)'
-			}
-		]);
-
-		if (opts.dir) {
-			cwd = opts.dir;
+	const args = await yargs(hideBin(process.argv)).coerce('types', (arg) =>
+		arg === 'none' ? null : arg
+	).argv;
```
JReinhold, line 28: this makes it so that `--types=none` ends up as `{ types: null }` internally.

Alternatively we could make the arg pass one-to-one either by making the arg `null` instead of `"none"`, _or_ changing the internal `null` to `"none"`.

Conduitry:
Thank you, but this isn't something we want to add. See, most recently, https://github.com/sveltejs/kit/pull/6117#issuecomment-1221323822 for more information.

JReinhold:
Oh wow sorry for wasting your time. My searching came up empty.

Ni55aN:
@dummdidumm the problem is that then you need to install the package and you can not use it through `npx`

JReinhold:
@Ni55aN I've created a separate package for this that might suit your needs. I attempt to keep it up-to-date regularly. https://github.com/storybookjs/create-svelte-with-args

Ni55aN:
@JReinhold thanks, opted for this package ✔

https://github.com/retejs/rete-kit/blob/1f713d3431a994d8782c674133ad924d258c0402/src/app/stack/svelte/index.ts#L17-L23

PR #6986, opened 2022-09-23
Added form prop to page store

### LINKED ISSUE #6701: Expose `form` property in `page` store
### Describe the problem

The new actions and forms api is very good, but I was very surprised (probably misread the docs and migration guide) to find out that the `form` prop is not exposed in the `page` store. 

Since we can have both `export let data` and `$page.data`, I think it would make a lot of sense to have to same for `form`.

Moreover, if I want to access the `form` property outside of +page.svelte (like in a +layout.svelte or a component further down the tree), the current api requires me to do a lot of setup, like for example populating a store with the data, exposing the data through context or passing the data as a prop. For one page, this is not too bad, but having to do this for every route is very cumbersome.

I realize I can access the `result` prop in the `enhance` callback, but again, it requires more setup if I want to use this data to populate input values and errors, and it would not work without js.

Add the `form` prop to the `page` store.

### Alternatives considered

Do this in userland, like using a store, context or passing the data as prop. In every page.

### Importance

would make my life easier

### Additional Information

This issue brought the issue count back to 300, sorry for this :)

sphinxc0re:
Closes #6701 

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/sync/write_root.js
+++ b/packages/kit/src/core/sync/write_root.js
@@ -14,10 +14,8 @@ export function write_root(manifest_data, output) {
 		1
 	);
 
-	const levels = [];
-	for (let i = 0; i <= max_depth; i += 1) {
-		levels.push(i);
-	}
+	// Create an array with the numbers in the range of [0, max_depth]
+	const levels = Array.from(Array(max_depth + 1).keys());
 
 	let l = max_depth;
 
@@ -57,7 +55,7 @@ export function write_root(manifest_data, output) {
 					setContext('__svelte__', stores);
 				}
 
-				$: stores.page.set(page);
+				$: stores.page.set({...page, form});
 				afterUpdate(stores.page.notify);
 
 				let mounted = false;
--- a/packages/kit/src/core/sync/write_root.js
+++ b/packages/kit/src/core/sync/write_root.js
@@ -55,7 +55,7 @@ export function write_root(manifest_data, output) {
 					setContext('__svelte__', stores);
 				}
 
-				$: stores.page.set({...page, form});
+				$: stores.page.set(page);
 				afterUpdate(stores.page.notify);
 
 				let mounted = false;
--- a/packages/kit/src/core/sync/write_root.js
+++ b/packages/kit/src/core/sync/write_root.js
@@ -14,8 +14,10 @@ export function write_root(manifest_data, output) {
 		1
 	);
 
-	// Create an array with the numbers in the range of [0, max_depth]
-	const levels = Array.from(Array(max_depth + 1).keys());
+	const levels = [];
+	for (let i = 0; i <= max_depth; i += 1) {
+		levels.push(i);
+	}
 
 	let l = max_depth;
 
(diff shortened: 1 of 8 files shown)


packages/kit/src/core/sync/write_root.js
```diff
@@ -57,7 +55,7 @@ export function write_root(manifest_data, output) {
 					setContext('__svelte__', stores);
 				}
 
-				$: stores.page.set(page);
+				$: stores.page.set({...page, form});
```
dummdidumm, line 58: A change more in line with the how it's done for the other values of page (namely `data`) would be to adjust the creation of the `page` object. This would need adjustments to `runtime/server/page/render.js` and `runtime/client.js` where that object is created.

sphinxc0re: Is it even possible to automatically update `$page.form` automatically on the client. As I understand it, forms do not have dynamic behavior on the client by default. The only thing I could do here is to update the `enhance` function to also update the `form` field.

Could you maybe elaborate a bit more on what the client should do here?

packages/kit/src/runtime/client/client.js
```diff
@@ -459,6 +459,7 @@ export function create_client({ target, base, trailing_slash }) {
 				routeId: route && route.id,
 				status,
 				url,
+				form,
```
dummdidumm, line 462: The `page_changed` check above probably needs to be enhanced by appending `|| form !== undefined`

dummdidumm (commented):
Thank you! I didn't test it out yet, but this should probably do it. Could you add a test for this aswell? It's probably enough to enhance the `enhance/+page.svelte` with another `<pre>{JSON.stringify($page.form)}</pre>` and expand the corresponding test.

sphinxc0re:
Sure! I will absolutely add a test. I just wanted to await feedback before doing so

sphinxc0re:
I just realized that I have yet to add types for the `$page.form` property. The problem I have is that I can't wrap my head around getting an automatically generated type `ActionData` working with an Ambient type `Page`.

dummdidumm:
You can't, make it `any`, there's no better way 👍

dummdidumm (approved):
Thank you!

sphinxc0re:
You're welcome! Thanks for the help!

ciscoheat:
So what does this form property refer to? :)

dummdidumm:
What do you mean exactly?

ciscoheat:
How do we use it?

dummdidumm:
Use what? This PR added `$page.form` and you access it like you access other things on that page store. I'm still not sure what you exactly want clarification on.

ciscoheat:
I assume it's a property that you added with good reason, so when is it not `undefined`, and under what circumstances? How does it improve my SvelteKit experience?

dummdidumm:
See #6701 for motivation. An mention in the form actions docs is indeed missing, will open a PR for that.

ciscoheat:
Just to be certain, it relates to [Form Actions](https://kit.svelte.dev/docs/form-actions), right?

mustofa-id:
So with this addition, I can subscribe to `page` store in the root layout to listen to every form action response to notify the user what's going on, like saved, updated, deleted, etc. It's a cinch!

PR #6992, opened 2022-09-23
see if this reduces flakiness

Rich-Harris:


Rich-Harris:
zero flakiness for this test now. great success

PR #6997, opened 2022-09-23
Use custom header to distinguish between `+page.server.js` and `+server.js` requests in `enhance`

### LINKED ISSUE #6983: `use:enhance` always targeting `+server.{js|ts}` instead of `+page.server.{js|ts}`
### Describe the bug

When `+server.{js|ts}` file in the same directory  as `+page.server.{js|ts}` file, the `use:enhance` form action cannot hit form `actions` methods. This is does not following content negotiation [rules](https://kit.svelte.dev/docs/routing#server-content-negotiation) as `use:enhance` always using `accept: 'application/json'` request header.

### Reproduction

https://github.com/mustofa-id/sveltekit-form-enhance-issue

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (4) x64 Intel(R) Core(TM) i3-4160 CPU @ 3.60GHz     
    Memory: 2.21 GB / 7.90 GB
  Binaries:
    Node: 16.15.1 - C:\Program Files\nodejs\node.EXE
    npm: 8.12.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (105.0.1343.42)
  npmPackages:
    @sveltejs/adapter-node: 1.0.0-next.95 => 1.0.0-next.95   
    @sveltejs/kit: 1.0.0-next.499 => 1.0.0-next.499 
    svelte: ^3.49.0 => 3.49.0 
    vite: ^3.1.1 => 3.1.1
```

### Severity

### Additional Information

_No response_

Rich-Harris:
Closes #6983 

### Tests

### Changesets


dummdidumm (approved):

dummdidumm (changes requested):
Ah wait, should we give this a more generic name, so it also works to distinguish `GET` requests? Also, this should probably be mentioned in the docs.

packages/kit/src/runtime/server/endpoint.js
```diff
@@ -77,13 +77,16 @@ export async function render_endpoint(event, mod, state) {
...
-	const { method } = event.request;
+	const { method, headers } = event.request;
 
 	if (method === 'PUT' || method === 'PATCH' || method === 'DELETE') {
 		// These methods exist exclusively for endpoints
 		return true;
 	}
 
+	// use:enhance uses a custom header to disambiguate
+	if (headers.get('x-sveltekit-action') === 'true') return false;
```
Rich-Harris, line 88: i'm not sure what it would mean for this to apply to `GET` requests

```suggestion
	if (method === 'POST' && headers.get('x-sveltekit-action') === 'true') return false;
```

dummdidumm (approved):
Yeah on second thought, that doesn't make sense, you don't want to do a fetch-get getting the HTML string (and if you do, just do `text/html`)

PR #6998, opened 2022-09-23
only run postinstall if package.json exists

### LINKED ISSUE #6987: Postinstall breaks when `package.json` is missing (for instance on a `pnpm fetch`)
### Describe the bug

It appears #6949 broke `pnpm fetch` when no `package.json` is present.

In the context of a Docker build like the following:

```Dockerfile
FROM node:18 as builder
RUN corepack enable
WORKDIR /workspace
# Step 1
COPY pnpm-lock.yaml ./
RUN pnpm fetch
# Step 2
COPY package.json ./
RUN pnpm install --frozen-lockfile --offline
# [...]
```

where Step 1 can be cached as long as `pnpm-lock.yaml` stays the same, even if `package.json` changes (for instance on a config or script change that does not require package re-install).

But from what I understand it means when `postinstall` is run in `pnpm fetch`, it doesn't have access to a `package.json` file.

### Reproduction

In any SvelteKit repo, create the lock file by running install:

```bash
pnpm install
```

Remove the installed packages and the `package.json` file:

```bash
rm -r node_modules/ package.json
```

Now, running `pnpm fetch` will fail on versions `1.0.0-next.495` onward (Logs provided below).

### Logs

```shell
.../esbuild@0.15.8/node_modules/esbuild postinstall$ node install.js
.../esbuild@0.15.8/node_modules/esbuild postinstall: Done
.../node_modules/@sveltejs/kit postinstall$ node postinstall.js
.../node_modules/@sveltejs/kit postinstall: node:fs:599
.../node_modules/@sveltejs/kit postinstall:   handleErrorFromBinding(ctx);
.../node_modules/@sveltejs/kit postinstall:   ^
.../node_modules/@sveltejs/kit postinstall: Error: ENOENT: no such file or directory, open 'package.json'
.../node_modules/@sveltejs/kit postinstall:     at Object.openSync (node:fs:599:3)
.../node_modules/@sveltejs/kit postinstall:     at Object.readFileSync (node:fs:467:35)
.../node_modules/@sveltejs/kit postinstall:     at file:///workspace/node_modules/.pnpm/@sveltejs+kit@1.0.0-next.499_svelte@3.50.1+vite@3.1.3/node_modules/@sveltejs/kit/postinstall.js:10:27 {
.../node_modules/@sveltejs/kit postinstall:   errno: -2,
.../node_modules/@sveltejs/kit postinstall:   syscall: 'open',
.../node_modules/@sveltejs/kit postinstall:   code: 'ENOENT',
.../node_modules/@sveltejs/kit postinstall:   path: 'package.json'
.../node_modules/@sveltejs/kit postinstall: }
.../node_modules/@sveltejs/kit postinstall: Node.js v18.9.0
.../node_modules/@sveltejs/kit postinstall: Failed
 ELIFECYCLE  Command failed with exit code 1.
The command '/bin/sh -c pnpm fetch' returned a non-zero code: 1
```

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 20.04.5 LTS (Focal Fossa)
    CPU: (16) x64 AMD EPYC 7B13
    Memory: 26.52 GB / 62.80 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.8.0 - ~/.nvm/versions/node/v18.8.0/bin/node
    npm: 8.18.0 - ~/.nvm/versions/node/v18.8.0/bin/npm
  npmPackages:
    @sveltejs/adapter-node: 1.0.0-next.95 => 1.0.0-next.95 
    @sveltejs/kit: 1.0.0-next.499 => 1.0.0-next.499 
    svelte: ^3.50.1 => 3.50.1 
    vite: ^3.1.3 => 3.1.3
```

### Severity

annoyance

### Additional Information

_No response_

### LINKED ISSUE #7000: postinstall fails if workspaces uses object notation
### Describe the bug

A change introduced in https://github.com/sveltejs/kit/pull/6969 seemingly prevents installation if using a monorepo with a `package.json` that specifies `workspaces` as an object instead of an array of globs.

this matches the pattern described here: https://classic.yarnpkg.com/blog/2018/02/15/nohoist/

### Reproduction

have `@svelte/kit` be included in a monorepo with package.json like follows and run `yarn`

```
	"workspaces": {
		"packages": [
			"packages/*"
		]
	},
```

### Logs

```shell
yarn install v1.22.19
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
[4/5] Linking dependencies...
[5/5] Building fresh packages...
[1/2] ⠈ @sveltejs/kit
error /home/user/src/monorepo/node_modules/@sveltejs/kit: Command failed.
Exit code: 1
Command: node postinstall.js
Arguments: 
Directory: /home/user/src/monorepo/node_modules/@sveltejs/kit
Output:
file:///home/user/src/monorepo/node_modules/@sveltejs/kit/postinstall.js:15
	for (const directory of pkg.workspaces) {
	                            ^

TypeError: pkg.workspaces is not iterable
    at file:///home/user/src/monorepo/node_modules/@sveltejs/kit/postinstall.js:15:30
```

### System Info

```shell
Binaries:
    Node: 16.17.0 - ~/.nvm/versions/node/v16.17.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.17.0/bin/yarn
    npm: 8.19.1 - ~/.nvm/versions/node/v16.17.0/bin/npm
```

### Severity

### Additional Information

_No response_

Rich-Harris:
* closes #6987
* closes #7000

### Tests

### Changesets

### DIFF
--- a/packages/kit/postinstall.js
+++ b/packages/kit/postinstall.js
@@ -7,7 +7,7 @@ import * as sync from './src/core/sync/sync.js';
 const cwd = process.env.INIT_CWD ?? process.cwd();
 process.chdir(cwd);
 
-if (fs.readFileSync('package.json')) {
+if (fs.existsSync('package.json')) {
 	const pkg = JSON.parse(fs.readFileSync('package.json', 'utf8'));
 
 	const directories = [];
--- a/packages/kit/postinstall.js
+++ b/packages/kit/postinstall.js
@@ -7,32 +7,34 @@ import * as sync from './src/core/sync/sync.js';
 const cwd = process.env.INIT_CWD ?? process.cwd();
 process.chdir(cwd);
 
-const pkg = JSON.parse(fs.readFileSync('package.json', 'utf8'));
-
-const directories = [];
-
-if (pkg.workspaces) {
-	for (const directory of pkg.workspaces) {
-		directories.push(...glob(directory, { cwd }).map((dir) => path.resolve(cwd, dir)));
-	}
-} else {
-	directories.push(cwd);
-}
-
-for (const cwd of directories) {
-	process.chdir(cwd);
-
-	if (!fs.existsSync('package.json')) continue;
-	if (!fs.existsSync('svelte.config.js')) continue;
-
+if (fs.readFileSync('package.json')) {
 	const pkg = JSON.parse(fs.readFileSync('package.json', 'utf8'));
-	if (!pkg.dependencies?.['@sveltejs/kit'] && !pkg.devDependencies?.['@sveltejs/kit']) continue;
 
-	try {
-		const config = await load_config();
-		await sync.all(config, 'development');
-	} catch (e) {
-		console.log('Error while trying to sync SvelteKit config');
-		console.log(e);
+	const directories = [];
+
+	if (pkg.workspaces) {
+		for (const directory of pkg.workspaces) {
+			directories.push(...glob(directory, { cwd }).map((dir) => path.resolve(cwd, dir)));
+		}
+	} else {
+		directories.push(cwd);
+	}
+
+	for (const cwd of directories) {
+		process.chdir(cwd);
+
+		if (!fs.existsSync('package.json')) continue;
+		if (!fs.existsSync('svelte.config.js')) continue;
+
+		const pkg = JSON.parse(fs.readFileSync('package.json', 'utf8'));
+		if (!pkg.dependencies?.['@sveltejs/kit'] && !pkg.devDependencies?.['@sveltejs/kit']) continue;
+
+		try {
+			const config = await load_config();
+			await sync.all(config, 'development');
+		} catch (e) {
+			console.log('Error while trying to sync SvelteKit config');
+			console.log(e);
+		}
 	}
 }
--- a/packages/kit/postinstall.js
+++ b/packages/kit/postinstall.js
@@ -4,37 +4,41 @@ import glob from 'tiny-glob/sync.js';
 import { load_config } from './src/core/config/index.js';
 import * as sync from './src/core/sync/sync.js';
 
-const cwd = process.env.INIT_CWD ?? process.cwd();
-process.chdir(cwd);
+try {
+	const cwd = process.env.INIT_CWD ?? process.cwd();
+	process.chdir(cwd);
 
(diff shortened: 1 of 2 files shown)


Conduitry:
Should we have another bigger try-catch here in case there are other weird unexpected errors? Given what we're using the postinstall script for here, it doesn't seem to me there would ever be a situation where I would prefer the whole install to fail.

packages/kit/postinstall.js
```diff
@@ -7,32 +7,34 @@ import * as sync from './src/core/sync/sync.js';
 const cwd = process.env.INIT_CWD ?? process.cwd();
 process.chdir(cwd);
 
-const pkg = JSON.parse(fs.readFileSync('package.json', 'utf8'));
+if (fs.readFileSync('package.json')) {
```
dummdidumm, line 10:
```suggestion
if (fs.existsSync('package.json')) {
```
Also, to save some indentation, maybe do `if (!existsSync) return`?

Rich-Harris: oops, good catch. I don't think you can early return from a module though?

Rich-Harris:
In general I'm not a fan of all-encompassing try-catch blocks, because they prevent us from catching actual logic bugs. Like, the thing in https://github.com/sveltejs/kit/issues/6987#issuecomment-1256242360 with `workspaces` not being an array of strings — if we just wrapped the whole thing then we wouldn't encounter that situation, we'd just bail out. I guess we could print the error and still exit with code 0

packages/kit/postinstall.js
Rich-Harris: yeah, just checked — 'Illegal return statement'

Conduitry:
Logging the exception sounds fine. I'm just saying that we shouldn't fail the whole install, because that doesn't help anyone, especially for something that would be generated when they build the project anyway.

packages/kit/postinstall.js
dummdidumm: Oh right, forgot that this is top-level 😅

dummdidumm (approved):

Issue #733, opened 2021-03-28
trailing slash removed when navigating


wighawag:
**Describe the bug**
I would like my path to end up with a trailing slash.
Currently with these lines, svelte kit always remove it : https://github.com/sveltejs/kit/blob/84e9023c06dc3579972f93e0d5cf1f2481aa3be4/packages/kit/src/runtime/client/router.js#L185-L188

**Expected behavior**
This behavior should at least be configurable.

Having a trailing slash is important to me as ipfs gateways append it anyway and so without being able to keep the trailing slash, there is a different behaviour between SPA navigation and navigation through request.

Plus it cause issue to calculate relative path as without slash it the resource is now considered to be in the parent folder

I think trailing slash are more accurate in that regard since on a static website, the corresponding underlying index.html file is in its own folder.

**Severity**
This is a blocker for me that deploy website on ipfs

I am working on a ipfs adapter to support ipfs properly and I am currently stuck with this.

wighawag:
The adapater repo is here : https://github.com/wighawag/sveltejs-adapater-ipfs
See readme for what it actually does : https://github.com/wighawag/sveltejs-adapater-ipfs/blob/main/README.md

The test app is here :  (branch : adpater-ipfs) can be found here : https://github.com/wighawag/svelte-kit-ipfs-example/tree/adapter-ipfs

With the current modif it works but ideally I would not need to replace generated code as this is brittle

Rich-Harris:
previous discussion [here](https://github.com/sveltejs/kit/issues/192), which lead to #267. from there:

> Starting with no slashes for the beta and waiting until we get a use case for something else sounds like a sensible plan to me

Sounds like we have a use case!

What we don't have is a sense of how this should be configured. Is an app-wide setting sufficient, or does it need to be per-page? What sort of thing did you have in mind?

wighawag:
Hi @Rich-Harris thanks for the link to past discussion

> What we don't have is a sense of how this should be configured. Is an app-wide setting sufficient, or does it need to be per-page?  

For my use case (ipfs hosting), an app-wide config would be sufficient

ipfs gateway force the trailing slash when an index.html is detected and used.

The trailing slash make it easier to compute relative path too

mhatvan:
Another use case is SEO, the same route with slash and without would be indexed as two different pages by a crawlbot, therefore deciding to redirect to either or makes sense.

techniq:
[...]

PR #7003, opened 2022-09-23
maybe improve types documentation a bit

Rich-Harris:
Some of the documentation around types is a bit ropey. For example I don't love how [this](https://kit.svelte.dev/docs/types#sveltejs-kit-cookies) looks:

image

Also, the [configuration documentation](https://kit.svelte.dev/docs/configuration) is kind of all over the place — this stuff really belongs in `types/index.d.ts`, so that it appears inline when you're actually _writing_ the config, rather than being manually written in a separate documentation file where it's less useful (and manually maintained as we make changes).

In this PR I'm trying some ideas for how we might improve it a bit. Am open to all suggestions. Some of this might need custom CSS to make it look halfway decent.

### DIFF
--- a/sites/kit.svelte.dev/src/lib/docs/server/index.js
+++ b/sites/kit.svelte.dev/src/lib/docs/server/index.js
@@ -368,7 +368,7 @@ function parse({ body, file, slug, code, codespan }) {
 					slug
 				});
 			} else {
-				throw new Error(`Unexpected <h${level}> in ${file}`);
+				// throw new Error(`Unexpected <h${level}> in ${file}`);
 			}
 
 			return `<h${level} id="${slug}">${html}<a href="#${slug}" class="anchor"><span class="visually-hidden">permalink</span></a></h${level}>`;
--- a/documentation/docs/30-types-presentation.md
+++ b/documentation/docs/30-types-presentation.md
@@ -1,0 +1,171 @@
+---
+title: Types presentation
+---
+
+Brainstorming ideas for better presentation of types.
+
+Compare the (WIP) examples below with https://kit.svelte.dev/docs/types#sveltejs-kit-cookies, https://kit.svelte.dev/docs/types#sveltejs-kit-config, https://kit.svelte.dev/docs/types#sveltejs-kit-kitconfig and https://kit.svelte.dev/docs/configuration.
+
+The idea is that we'd generate this automatically from `types/index.d.ts` etc (including `@example` and `@default` tags), and configuration would be documented this way rather than via the hand-written https://kit.svelte.dev/docs/configuration.
+
+### @sveltejs/kit
+
+#### Cookies
+
+An interface for interacting with cookies during a request.
+
+```ts
+get(name, opts): string | undefined
+```
+
+* `name: string` The name of the cookie
+* `opts?: import('cookie').CookieParseOptions` An object containing an optional `decode` function
+
+Gets a cookie that was previously set with `cookies.set`, or from the request headers.
+
+```ts
+set(name, value, opts): void;
+```
+
+* `name: string` The name of the cookie
+* `value: string` The cookie value
+* `opts?: import('cookie').CookieSerializeOptions` TODO Cookie options
+
+Sets a cookie. This will add a `set-cookie` header to the response, but also make the cookie available via `cookies.get` during the current request.
+
+The `httpOnly` and `secure` options are `true` by default, and must be explicitly disabled if you want cookies to be readable by client-side JavaScript and/or transmitted over HTTP. The `sameSite` option defaults to `lax`.
+
+By default, the `path` of a cookie is the 'directory' of the current pathname. In most cases you should explicitly set `path: '/'` to make the cookie available throughout your app.
+
+
+```ts
+delete(name, opts): void;
+```
+
+* `name: string` The name of the cookie
+* `value: string` The cookie value
+* `opts?: import('cookie').CookieSerializeOptions` TODO Cookie options
+
+Deletes a cookie by setting its value to an empty string and setting the expiry date in the past.
+
+
+```ts
+serialize(name, value, opts): string;
+```
+
+* `name: string` the name for the cookie
+* `value: string` value to set the cookie to
+* `opts?: import('cookie').CookieSerializeOptions` TODO Cookie options
+
+Serialize a cookie name-value pair into a Set-Cookie header string.
+
+The `httpOnly` and `secure` options are `true` by default, and must be explicitly disabled if you want cookies to be readable by client-side JavaScript and/or transmitted over HTTP. The `sameSite` option defaults to `lax`.
+
+By default, the `path` of a cookie is the current pathname. In most cases you should explicitly set `path: '/'` to make the cookie available throughout your app.
+
+#### Config
+
+Your project's configuration, exported from svelte.config.js.
+
+```ts
+compilerOptions?: CompileOptions
+```
+
+Svelte compiler options.
+
+```ts
+extensions?: string[]
+```
+
+An array of file extensions that should be treated as Svelte components.
+
+Default value: `['.svelte']`
+
+```ts
+kit?: KitConfig
+```
+
+SvelteKit-specific configuration.
+
+```ts
+package?: {...}
+```
+
+* `source?: string` `"src/lib"` Path to the source folder.
+* `dir?: string` `"package"` Path to write the package to
+* `emitTypes?: boolean` `true` Whether to generate `.d.ts` files
+* `exports?: (filepath: string) => boolean` `() => true` Whether to include a file in `pkg.exports`
+* `files?: (filepath: string) => boolean` `() => true` Whether to include a file in the package
+
+Configuration for `svelte-package`
+
+```ts
+preprocess?: any
+```
+
+Preprocessing options.
+
+```ts
+[key: string]: any;
+```
+
+Any additional options required by other tooling.
+
+#### KitConfig
+
+Your SvelteKit-specific configuration.
+
+```ts
+adapter?: Adapter
+```
+
+Run when executing `vite build` and determines how the output is converted for different platforms. See [Adapters](/docs/adapters).
+
+Default value: `undefined`
+
+```ts
+alias?: Record<string, string>
+```
+
+An object containing zero or more aliases used to replace values in `import` statements. These aliases are automatically passed to Vite and TypeScript.
+
+```js
+/// file: svelte.config.js
+/** @type {import('@sveltejs/kit').Config} */
+const config = {
+	kit: {
+		alias: {
+			// this will match a file
+			'my-file': 'path/to/my-file.js',
+
+			// this will match a directory and its contents
+			// (`my-directory/x` resolves to `path/to/my-directory/x`)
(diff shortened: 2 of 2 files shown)


SrGeneroso:
Just learning TypeDoc while learning TS.
My main motivation to learn TS is in fact having docs for free*. This combo is making me fall in love with TS. I heard JsDocs was gaining some traction in the svelte community as the first step to TS without having to get married to it.
Should svelte/kit have both? It seems JsDoc and TsDoc is mostly interoperable (don't have to add the type in the params in TsDoc as one diference), don't know enough to comment on it, but I've seen some TypeDoc magic that I can't wait to see around here. I might be talking nonsense and all of this is already implemented or something else, I just wanted to bump this, I'd loved to see better documentation not only for final users but also for devs of svelte/kit.

dummdidumm (commented):
Some general thoughts aside from the inline comment:
- the "generated types" and "tsconfig" sections are related, but they feel a little lost down there on that big API reference page. It feels like they deserve its own section
- speaking of API reference: Maybe it should be named that instead. Yes it's types, but they at the same time represent the API of these things, moreso if we transform them to something more appealing like this. So maybe "generated types" and "tsconfig" stay where they are and the rest moves to a new API section
- for inspiration, this is how the API section of the Angular docs looks like: https://angular.io/api . They have an overview and then separate pages for each. Depending on how elaborate we become with all these types this may or may not make sense. Too many little pages scattered out feels wrong, one giant page also feels wrong, something inbetween would be nice

dummdidumm:
During the work on #7087 I felt hindered by having to write more detailed docs inside JSDoc. You can't use JSDoc within JSDoc for example (`/** @type {..} */` will close the outer JSDoc). Moreover, it could get overwhelming seing an ocean of docs when hovering over a function. In also complicates future enhancements like translation or JS+TS in the docs with a switch.

I therefore think we need a different approach - some kind of mix between manual docs and "include X from the types". Made-up example-MD file that would need to be preprocessed:

```
## cookies

### set

--definition-- // <- this would extract the type definition from types.d.ts into here
--summary-- // <- this would extract the JSDoc from types.d.ts into here in case you don't want to duplicate things

Lorem ipsum stuff
```

dummdidumm:
Reverting part of my previous comment - the API docs should be very concise and more on the short side. The cookie example is the biggest it should probably get. If there's more it should go into a different part of the docs, explaining things in more detail and in the context of usage. What remains is the question about code samples, which I don't have a good answer to.

dummdidumm:
I pushed a POC of a generator enhancement that transforms interfaces into better looking docs similar to what you proposed. I added a border to each of the snippets that present sub headline to better distinguish them from regular code snippets. I also kept a trimmed down version of the whole interface at the start of each section so people can get a quick overview of that type.

I think the result is much better-looking, but only if we have actually some docs in there - so I think if we change this, we should also get down to writing a lot more quick API docs here at the same time.

The final part of the picture (for me at least) would be to restructure the types and modules sections into one "API Reference" section, and we group the modules / interfaces logically into sub pages. Something like this (picture incoming)

Rich-Harris:
Pushed up some design tweaks that make type blocks visually distinct from code blocks, and make things more grouped visually.

This is definitely an improvement over the status quo, but there's a bunch of issues (inherited or otherwise) that we should try and address if we can:

* [ ] code block references value imports, not types

image

---

* [ ] this still looks overwhelming, not sure if there's a better way to present it

image

---

* [ ] unrelated to this work, but super confusing wording

image

---

* [ ] properties without descriptions look weird, especially at the end

We need to add loads more docs

image

---

* [ ] do I as a user need to see this? seems complicated and irrelevant

image

---

* [ ] repetition of `cancel()` looks kinda awkward in situations like this

What if the top was just `interface BeforeNavigate extends Navigation {...}` or something?

image

---

* [ ] this is duplicative

Everything in this block is described immediately afterward — it's not clear it serves any real value

image

---

* [ ] some types don't create links, like `ValidatedConfig` here

Also, a sequence of properties without descriptions looks super weird

image

---

* [ ] indentation is wacky. also, what does "`dest` null" mean?

image

---

* [ ] bullet points and text are way out of alignment

image

---

* [ ] these should all link back to the relevant docs sections

image

---

* [ ] why do some things get property descriptions and others don't?

Also, the configuration docs should be driven from the types. we need to be able to show default values somehow

image

---

* [ ] references to generics show up as 'any'

image

---

* [ ] wonky margins around code blocks

image

---

* [ ] double line breaks are unnecessary

(assuming we don't get rid of the contents of these interfaces altogether)

image

---

* [ ] nested properties don't look great

image

dummdidumm:
This is a very thorough analysis! I agree with most of that. I think the two biggest things we can do are
- doc this shit up: everything should have some jsdoc so it doesn't look that empty
- rethink modules and types sections: I think we should merge them to an API section. If we think it gets too big that way, we could split each into a separate page, which would also help with "what ends/starts where?"

sites/kit.svelte.dev/src/lib/docs/server/index.js
```diff
@@ -120,7 +120,9 @@ export async function read_file(file) {
 				version_class = ' js-version';
 			}
 
-			if (language === 'js' || language === 'ts') {
+			if (language === 'ts') {
+				html = renderCodeToHTML(source, 'ts', { twoslash: false }, {}, highlighter);
```
dummdidumm, line 124: What's the reason for this change? Doesn't this remove these hover hints for all TS blocks (which also affects all other doc pages when people toggle the typescript version of code snippets)?

Rich-Harris: Blurgh yes it does — we need to distinguish between (effectively) `.ts` and `.d.ts`. On the `.d.ts` blocks they're noisy and useless

packages/kit/types/index.d.ts
```diff
@@ -174,58 +176,256 @@ export interface Cookies {
...
+	 * Your [adapter](https://kit.svelte.dev/docs/adapters) is run when executing `vite build`. It determines how the output is converted for different platforms.
+	 * @default undefined
+	 */
 	adapter?: Adapter;
+	/**
+	 * An object containing zero or more aliases used to replace values in `import` statements. These aliases are automatically passed to Vite and TypeScript.
+	 *
+	 * ```js
+	 * /// file: svelte.config.js
+	 * /** @type {import('@sveltejs/kit').Config} *\/
```
dummdidumm, line 189: This makes the whole code snippet unreadable in VS Code. I think it would be better to either have something like 
```js
/// type: import('@sveltejs/kit').Config
```
above things like this which get transformed to `/** @type ... */`, or leave out jsdoc comments inside jsdoc completely (I already added some code [`read_d_ts_file`] so that it doesn't fail in this case)

Rich-Harris:
Another issue — probably shouldn't block merging this PR, but worth opening as follow-up — type variables are ignored when converting JSDoc to TypeScript:

image

image

dummdidumm:
Don't know if you already saw this, so I'm posting it again: IMO it makes sense to rethink modules and types sections in general: I think we should merge them to an API section. Right now things are split across these two sections (for example `$app/forms` appears in both), which is weird. If we think it gets too big that way, we could split each into a separate page, which would also help with "what ends/starts where?". I quite like Angular's approach here, with an overview page from which you can jump to the details of what you need https://angular.io/api

packages/kit/types/index.d.ts
Rich-Harris: good catch — done

Rich-Harris:
Yeah, I had that in the back of my mind as I was fixing up the smaller things. It's definitely an option, though for the `SubmitFunction` in particular I found it made more sense just to move it into `@sveltejs/kit` (notwithstanding the fact that it makes this a breaking change).

It _is_ a big page. I definitely wouldn't rule out us subdividing it further later like Angular does. But I think it's worth avoiding if we can — Angular's API is just monstrously large, and if SvelteKit ends up as unwieldy as that then I plan to retire. 

Until then I think the modules/types distinction makes sense, given that all the types are exported from `@sveltejs/kit` or `./$types`. I think combining them at this stage would reduce clarity

Rich-Harris:
There's definitely room to add more descriptions, but that doesn't need to hold up this PR any further. I think the big thing remaining is to sort out the configuration docs:

* [ ] they're still written as though we can't document nested properties — needs tweaking
* [ ] we have both https://kit-svelte-8bwewn5rw-svelte.vercel.app/docs/configuration and https://kit-svelte-8bwewn5rw-svelte.vercel.app/docs/types#sveltejs-kit-kitconfig. There Can Be Only One. I'm not sure how best to shuffle things around though
* [ ] defaults aren't present for all values yet. (Also, it might be nice if we had some iconography for params/defaults/returns or at least a slightly improved design, rather than bullet points)
* [ ] if we retire `/docs/configuration` then it would be good to generate hash links for properties, at least at the top level

dummdidumm:
I vote for keeping the configuration section. It can be autogenerated, but I think it's better to have it separate, since it's rather big and semantically doesn't fit into types/modules for me.

Rich-Harris:
Another thing to fix in a follow-up PR
 
image

packages/create-svelte/templates/default/src/app.d.ts
```diff
@@ -1,4 +1,9 @@
 // See https://kit.svelte.dev/docs/types#app
 // for information about these interfaces
 // and what to do when importing types
-declare namespace App {}
+declare namespace App {
```
dummdidumm, line 4: this would need a changeset for `create-svelte`

packages/kit/types/index.d.ts
```diff
@@ -349,13 +349,16 @@ export interface KitConfig {
...
 		hooks?: {
-			/** @default "src/hooks.client" */
+			/**
+			 * The location of your client [hooks](/docs/hooks).
+			 * @default "src/hooks.client"
+			 */
 			client?: string;
-			/** @default "src/hooks.server" */
+			/**
+			 * The location of your server [hooks](/docs/hooks).
```
dummdidumm, line 359: this (and the comment above) needs an absolute link, since it's inside a `d.ts` file

Rich-Harris:
The `create-svelte` tests are failing — something to do with moving `SubmitFunction` form `$app/forms` to `@sveltejs/kit` I assume. @dummdidumm any idea why this is happening?

dummdidumm:
No idea why this fails ... I need to investigate more thoroughly tomorrow. So far it looks like this only fails for `.svelte` files, so this might be some bug inside language tools.

dummdidumm (approved):
"maybe", "a bit" - turned out to be far more than that 😄

dummdidumm:
> No idea why this fails ... I need to investigate more thoroughly tomorrow. So far it looks like this only fails for `.svelte` files, so this might be some bug inside language tools.

I lied, this also happens inside TS files. Still no idea why, it seems related to the specific interface, because if I for example use the `BeforeNavigate` interface it works.

realvasko:
I was trying to configure the trailingSlash option, as introduced in #733, but found it's missing in the current config and found this to be the PR where it got removed. I couldn't really figure out why, so was wondering if it was intentional? If so, is the feature no longer supported?

realvasko:
Never mind, saw that it's now part of the page options.

PR #7018, opened 2022-09-24
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,9 +10,14 @@ env:
   TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
   TURBO_TEAM: svelte
 
+permissions: {}
 jobs:
   release:
     # prevents this action from running on forks
+    permissions:
+      contents: write # to create release (changesets/action)
+      pull-requests: write # to create pull request (changesets/action)
+
     if: github.repository == 'sveltejs/kit'
     name: Release
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,11 +14,10 @@ permissions: {}
 jobs:
   release:
     # prevents this action from running on forks
+    if: github.repository == 'sveltejs/kit'
     permissions:
       contents: write # to create release (changesets/action)
       pull-requests: write # to create pull request (changesets/action)
-
-    if: github.repository == 'sveltejs/kit'
     name: Release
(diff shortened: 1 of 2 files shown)


benmccann:
Do we have to create a new GitHub token to receive benefit from this? (Or actually maybe a new one for the CI and a new one for  the release job)

sashashura:
Sorry, I don't follow. You don't need to introduce new tokens. For example, in ci.yml it just instructs that the automatically generated temporary GITHUB_TOKEN doesn't need special privileges. In release.yml, that only two write permissions (out of multiple ones given by default) are needed.

.github/workflows/release.yml
```diff
@@ -10,9 +10,14 @@ env:
...
 
+permissions: {}
 jobs:
   release:
     # prevents this action from running on forks
+    permissions:
+      contents: write # to create release (changesets/action)
+      pull-requests: write # to create pull request (changesets/action)
+
     if: github.repository == 'sveltejs/kit'
```
dominikg, line 21: the comment belongs to the if: condition, lets keep it that way
```suggestion
    # prevents this action from running on forks
    if: github.repository == 'sveltejs/kit'
    permissions:
      contents: write # to create release (changesets/action)
      pull-requests: write # to create pull request (changesets/action)
```

dominikg:
tested the proposed changes in another repo that uses changesets, the release workflow works as expected.
https://github.com/svitejs/vite-plugin-env-config/actions/runs/3176787479

dominikg (approved):

PR #7022, opened 2022-09-25
[docs] add page example to `depends`

### LINKED ISSUE #6788: Clarify `goto` behaviour and what conditions `load` gets called after navigating
### Describe the problem

While trying to make use of `goto` and `invalidateAll`, I couldn't get my page to call its `load` function predictably. It turns out, my page's `load` method had no dependencies that would trigger it to change, even though it's a navigational change with different params.

---

<details>
<summary>Details for those curious</summary>

## My example project

The layout and page files are in this file structure:

- `src/routes/[branch]/+layout.svelte`
- `src/routes/[branch]/+layout.server.js`
- `src/routes/[branch]/thing/[id]/+page.svelte`
- `src/routes/[branch]/thing/[id]/+page.server.js`

The broken behaviour:

- **Given** the current page is `/main/thing/abc123`
- **When** the user clicks a button in a 'Switch branch' dropdown (kinda like GitHub branches)
- **Then** it calls `goto('/otherbranch/thing/abc123')`
- **And** the target `[branch]` param has changed, and `[id]` stayed the same.
- **But** the target `+page.server.js`'s `load` method was not called again

In `+page.server.js`, I have:

```js
/** @type {import('./$types').PageServerLoad} */
export async function load ({ params, locals }) {
  if (!params.id) {
    // No id
    throw error(400, 'No ID provided')
  }

  // get db connection passed in from `src/hooks.server.js@handle`
  // which actually makes use of `params.branch` but isn't explicitly utilized in this `load`
  const { db } = locals

  // ... do db stuff

  return { rows }
}
```

It calls `params.id`, but it does not call `params.branch`, so I assume SvelteKit is aggressively optimizing this `load` to not invalidate if `params.branch` ever changes; and I was unaware. I made this adjustment, and now `load` runs after I call `goto('/different-branch-from-before/thing/abc123')`:

```diff
+ export async function load ({ params, locals, depends }) {
+   depends(params.branch) // this is probably wrong and I'm sorry
```

</details>

---

I came across this insightful comment: https://github.com/sveltejs/kit/issues/1684#issuecomment-862472061

It describes how `goto` calls a page (or layout's) respective `load` under the following conditions:

> `goto()` will only run `load()` if `page.params`, `page.path`, `page.query`, `session` or `context` is used inside of `load` and one of those values has changed since the last run of the load function.

On the surface this makes sense, the `load` method needs to have some sort of changed params in order for `goto` to call it again. However, I could not find this behaviour in the docs, and I still have uncertainties:

1. Does `page.params` refer to the `$page` store used on that `load` method's respective `+page.svelte`? Or does it (more likely) refer to `event.params` used in `load`?
2. How do I tell my page and its parent layout to call `load` when `goto` navigates to that page?
3. Is there a way to tell a `load` to *always* refetch values if the user navigates to it? I understand it's optimal to make use of `depends`, but I don't understand its functionality, either. 
4. How do I make use of `depends` so that if a path param changes, that `load` should invalidate? I assume it's `depends(params.targetParam)`, but I don't know.

I'm happy to work with the team to add some details to the docs to better capture the behaviour of `goto` and when it would or would not call `load` on the target route, how to assign path params to `depends` within `load`, and any applicable side-effects of these features.

### Alternatives considered

I guess I could do without the `event.locals` trickery and be more explicit with my dependencies in `load`, but I wanted to avoid some boilerplate. 

### Importance

would make my life easier

### Additional Information

This comment was also very informative and parts of it should probably make it to the docs: https://github.com/sveltejs/kit/issues/2560#issuecomment-937389988

Specifically the detail on how hard copies of store values will get properly recognized as changed in SvelteKit's optimization logic.

theetrain:
Closes #6788 

- Fix `invalidateAll` reference
- Add a page example for `depends` to illustrate custom URI in `invalidate`

### Tests

### Changesets

### DIFF
--- a/documentation/docs/05-load.md
+++ b/documentation/docs/05-load.md
@@ -118,6 +118,7 @@ declare module '$lib/api' {
 }
 
 // @filename: index.js
+/// file: src/routes/+page.js
 // ---cut---
 import * as api from '$lib/api';
 
@@ -134,6 +135,24 @@ export async function load({ depends }) {
 		bar: api.client.get('/bar')
 	};
 }
+```
+
+```svelte
+/// file: src/routes/+page.svelte
+<script>
+	import { invalidate } from '$app/navigation'
+
+	/** @type {import('./$types').PageData} */
+  export let data
+
+	const pageRefresh = async () => {
+		await invalidate('my-stuff:foo')
+	}
+</script>
+
+<p>{data.foo}<p>
+<p>{data.bar}</p>
+<button on:click={pageRefresh}>Refresh my stuff</button>
 ```
 
 #### fetch
@@ -342,7 +361,7 @@ A `load` function will re-run in the following situations:
 - It references a property of `url` (such as `url.pathname` or `url.search`) whose value has changed
 - It calls `await parent()` and a parent `load` function re-ran
 - It declared a dependency on a specific URL via [`fetch`](#input-methods-fetch) or [`depends`](#input-methods-depends), and that URL was marked invalid with [`invalidate(url)`](/docs/modules#$app-navigation-invalidate)
-- All active `load` functions were forcibly re-run with [`invalidate()`](/docs/modules#$app-navigation-invalidate)
+- All active `load` functions were forcibly re-run with [`invalidateAll()`](/docs/modules#$app-navigation-invalidate)
 
 If a `load` function is triggered to re-run, the page will not remount — instead, it will update with the new `data`. This means that components' internal state is preserved. If this isn't want you want, you can reset whatever you need to reset inside an [`afterNavigate`](/docs/modules#$app-navigation-afternavigate) callback, and/or wrap your component in a [`{#key ...}`](https://svelte.dev/docs#template-syntax-key) block.
 


documentation/docs/05-load.md
```diff
@@ -136,6 +137,24 @@ export async function load({ depends }) {
...
+<script>
+	import { invalidate } from '$app/navigation'
+
+	/** @type {import('./$types').PageData} */
+  export let data
+
+	const pageRefresh = async () => {
+		await invalidate('my-stuff:foo')
+	}
+</script>
```
benmccann, line 151: if we add this, you would need semicolons
```suggestion
<script>
	import { invalidate } from '$app/navigation';

	/** @type {import('./$types').PageData} */
  export let data;

	const pageRefresh = async () => {
		await invalidate('my-stuff:foo');
	}
</script>
```

documentation/docs/05-load.md
```diff
@@ -342,7 +361,7 @@ A `load` function will re-run in the following situations:
 - It references a property of `url` (such as `url.pathname` or `url.search`) whose value has changed
 - It calls `await parent()` and a parent `load` function re-ran
 - It declared a dependency on a specific URL via [`fetch`](#input-methods-fetch) or [`depends`](#input-methods-depends), and that URL was marked invalid with [`invalidate(url)`](/docs/modules#$app-navigation-invalidate)
-- All active `load` functions were forcibly re-run with [`invalidate()`](/docs/modules#$app-navigation-invalidate)
+- All active `load` functions were forcibly re-run with [`invalidateAll()`](/docs/modules#$app-navigation-invalidate)
```
benmccann, line 364: if we change the link text you also need to change the URL
```suggestion
- All active `load` functions were forcibly re-run with [`invalidateAll()`](/docs/modules#$app-navigation-invalidateall)
```

dummdidumm (approved):

Rich-Harris (approved):

PR #7039, opened 2022-09-26
Fixes default template app

cbillinis:
This PR changes the userid cookie on the default app to non-secure. Many people will start with the example app and will want to experiment. The majority of people will be running the app on http://localhost and therefore secure cookies cannot be set.

I tried the example app for the first time and was not working for me until I changed the userid cookie to non-secure.

### Tests

### Changesets

### DIFF
--- a/packages/create-svelte/templates/default/src/hooks.server.ts
+++ b/packages/create-svelte/templates/default/src/hooks.server.ts
@@ -8,7 +8,7 @@ export const handle: Handle = async ({ event, resolve }) => {
 		// if this is the first time the user has visited this app,
 		// set a cookie so that we recognise them when they return
 		userid = crypto.randomUUID();
-		event.cookies.set('userid', userid, { path: '/' });
+		event.cookies.set('userid', userid, { path: '/', secure: false });
 	}
 
 	event.locals.userid = userid;


Conduitry:
Is this a Safari thing? I think Chrome and Firefox will allow you to set and get secure cookies on localhost.

cbillinis:
> Is this a Safari thing? I think Chrome and Firefox will allow you to set and get secure cookies on localhost.

Yes, tested in Safari

Rich-Harris:
Thanks for the PR. We definitely don't want to encourage people to explicitly set `secure: false`. What if the default behaviour was to set `secure: false` if `url.hostname === 'localhost'`?

dummdidumm:
Would any deployed server also have localhost internally somehow? Should we be extra careful and do "is localhost AND dev mode"? That would fuck up preview mode locally though .. mhm

Rich-Harris:
I think it's fine — on platforms where the `Request` object doesn't have the host/origin set for you (i.e. `adapter-node`), you need to do it yourself anyway

Rich-Harris:
Opened https://github.com/sveltejs/kit/pull/7090

cbillinis:
Closing this due to #7090 which will solve this!

PR #7051, opened 2022-09-27
[feat] implement optional route params

### LINKED ISSUE #5072: Optional route parameters
### Describe the problem

I'm trying to add multiple languages to a site with a leading `/[locale]/...` route parameter.

There are two things that I cannot do right now without much overhead:

1) Having a default locale that has NO `[locale]` prefix. This would currently involve copying everything under `/[locale]` to `/` as well. Using `/[...locale]` does not work either, because a `/[...locale]/[...slug]` would not match correctly for many cases.

2) When using `adapter-static` and pre-rendering all routes, only routes picked up through crawling links will be rendered behind `/[locale]`. Using `kit.prerender.entries: ["*", "/de/something"]` will only pre-render `/de/something` but nothing else.

1) Add optional parameters with e.g. `[locale?]` syntax

2) Allow more wildcards in `kit.prerender.entries`, like `/de/*`

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

dummdidumm:
Closes #5072

I chose the `[[optional]]` syntax for alignment with the wider ecosystem/how other frameworks do it.
Probably needs some more tests, also wondering if we should have a check that tells you that you have both a required and optional param matching the same, which is forbidden - update: decided against that, we don't do that for conflicting required params, too, and it's probably too hard to find out taking into account `+server.js` and differing sub routes.
Apart from that it's ... done? I don't know. Feels like I'm missing something, but my local playground tests succeeded - update: with more tests I'm now confident it's done.

### Tests

### Changesets

### DIFF
--- a/documentation/docs/04-advanced-routing.md
+++ b/documentation/docs/04-advanced-routing.md
@@ -67,7 +67,7 @@ export function load(event) {
 
 A route like `[lang]/home` contains a parameter named `lang` which is required. Sometimes it's beneficial to make these parameters optional, so that in this example both `home` and `en/home` point to the same page. You can do that by wrapping the parameter in another bracket pair: `[[lang]]/home`
 
-Note that an optional route parameter cannot follow a rest parameter (`[...rest]/[[optional]]`), since parameters are matched 'greedily' and the optional parameter will always be unused.
+Note that an optional route parameter cannot follow a rest parameter (`[...rest]/[[optional]]`), since parameters are matched 'greedily' and the optional parameter would always be unused.
 
 ### Matching
 


packages/kit/src/core/sync/create_manifest_data/index.spec.js
```diff
@@ -320,6 +320,53 @@ test('allows rest parameters inside segments', () => {
 	]);
 });
 
+test('optional parameters', () => {
```
benmccann, line 323: It could be worth adding a case like `/[[locale]]/[...slug] `. I'm assuming `locale` should be greedy, but it'd be good to verify that with a test. You'd probably want to use a matcher with it in any real situation

benmccann (commented):
cool! this will help a lot with i18n, which is maybe the biggest ask!

documentation/docs/04-advanced-routing.md
```diff
@@ -63,6 +63,12 @@ export function load(event) {
 
 > If you don't handle 404 cases, they will appear in [`handleError`](/docs/hooks#shared-hooks-handleerror)
 
+### Optional parameters
+
+A route like `[lang]/home` contains a parameter named `lang` which is required. Sometimes it's benefitial to make these parameters optional, so that in this example both `home` and `en/home` point to the same page. You can do that by wrapping the parameter in another bracket pair: `[[lang]]/home`
```
benmccann, line 68:
```suggestion
A route like `[lang]/home` contains a parameter named `lang` which is required. Sometimes it's beneficial to make these parameters optional, so that in this example both `home` and `en/home` point to the same page. You can do that by wrapping the parameter in another bracket pair: `[[lang]]/home`
```

documentation/docs/04-advanced-routing.md
```diff
@@ -112,7 +120,8 @@ SvelteKit needs to know which route is being requested. To do so, it sorts them
 src/routes/foo-abc/+page.svelte
 src/routes/foo-[c]/+page.svelte
 src/routes/[a]/+server.js
-src/routes/[b]/+page.svelte
+src/routes/[c]/+page.svelte
+src/routes/[[b]]/sub/+page.svelte
```
Rich-Harris, line 124: just thinking out loud to myself... is this what we want? intuitively i feel like `/sub` should match `[[b]]/sub` before it matches `[c]`. hmm

benmccann: I think I'd agree with Rich at least for that one example. I'm not sure if there's an example motivating the version documented here

documentation/docs/04-advanced-routing.md
```diff
@@ -63,6 +63,12 @@ export function load(event) {
 
 > If you don't handle 404 cases, they will appear in [`handleError`](/docs/hooks#shared-hooks-handleerror)
 
+### Optional parameters
+
+A route like `[lang]/home` contains a parameter named `lang` which is required. Sometimes it's beneficial to make these parameters optional, so that in this example both `home` and `en/home` point to the same page. You can do that by wrapping the parameter in another bracket pair: `[[lang]]/home`
+
+Note that an optional route parameter cannot follow a rest parameter (`[...rest]/[[optional]]`), as it's impossible to know which way to match the route.
```
Conduitry, line 70: Does `[[optional]]/[...rest]` work? I would find it weird if one worked but not the other. You'd run into the same ambiguities either way. I'd expect both `[[optional]]`s to be greedy, which would probably mean they only make sense if you are using a route matcher.

Conduitry: @Rich-Harris Do you have thoughts about this? I see your solution was "disallow both", which is certainly _a_ solution, and probably the simplest one.

Rich-Harris: It's the simplest, but it's also the most future-proof, since if someone presents a compelling use case we can consider allowing them next to each other in a future release in a non-breaking way. Possibly by (for example) only allowing `[[optional=somematcher]]`

Conduitry: Makes sense to me. I'm good with disallowing both usages.

dummdidumm: The real problem is that the regex is greedy, so one works but not the other. I'm not sure if this is a good argument to allow optional before rest.

dummdidumm:
```suggestion
All route matches are done greedy. That means that a route like `[[optional1]]/[[optional2]]` matched against `/foo` will _always_ result in `{ optional1: 'foo' }`. For this reason an optional route parameter cannot follow a rest parameter (`[...rest]/[[optional]]`), as the `optional` parameter will never be matched.
```

Conduitry: (I've updated "greedy" to "greedily" in your suggestion.)

But I'm not really sure it's accurate to say that we _can't_ do this - it's just true that we can't do this with our current method of generating regexes for the route manifest. There's nothing stopping us from using non-greedy regexes or (in the `[...foo]/[[bar=matcher]]` case) needing to rely on a matcher to know which route should match. It may well be that we make the decision that these routes aren't worth the hassle of supporting or that they would result in a performance hit because we can't rely on regexes as much - but we should be careful not to frame these in the documentation as being inherently impossible to support.

leojbianchi:
Hi everyone, any update about the required review?

f-elix:
Thanks so much for this feature, its great! 

But I'm pretty sure that `[[lang]]/[...rest]` is going to be asked eventually, why not tackle it now? Would it be very complex to implement if a param matcher was required for this pattern?

documentation/docs/04-advanced-routing.md
Rich-Harris: I changed the implementation (and the docs) to disallow ambiguity in both directions

benmccann:
Supporting `[[lang]]/[...rest]` from the beginning probably makes sense to me. I think otherwise we're basically disallowing the use of `[...rest]` almost entirely for folks who use `[[lang]]` as that's going to be the root directory and everything else will be contained within it.

It doesn't seem surprising to me that we'd support `[[lang]]/[..rest]`, but not `[...rest]/[[lang]]` with the understanding that these are greedy constructs.

leojbianchi:
Thanks to all the collaborators who are working on this project. We were just waiting for this for our new development of www.mulkano.com, when we can implement it, we hope to share it with you. Cheers, Leo :)

dummdidumm:
Added random sorting prior to the test, which makes it fail for me locally - the `x/[...rest]` vs `[...rest]/x` case seems to be flawed

packages/kit/src/core/sync/create_manifest_data/index.spec.js
```diff
@@ -214,37 +210,41 @@ test('encodes invalid characters', () => {
...
-			'[wildcard]',
-			'[...rest]/deep/[...deep_rest]/xyz',
-			'[...rest]/deep/[...deep_rest]',
-			'[...rest]/abc',
-			'[...rest]/deep',
-			'(foo)/(bar)/[...all]',
-			'[...anotherrest]',
-			'[...rest]'
-		]
+test.only('sorts routes correctly', () => {
```
Conduitry, line 213: One of the things ESLint was doing before it was removed was to prevent things like this from being pushed. uvu doesn't support conditionally disallowing `only`/`skip`, so I had gone another route. If we don't want ESLint, we need to find another way to prevent this.

packages/kit/src/core/sync/create_manifest_data/index.spec.js
```diff
@@ -222,6 +222,8 @@ test('sorts routes correctly', () => {
 		'c/b[x]',
 		'c/foo',
 		'd/e',
+		'd/[[optional=match]]/e',
+		'd/[[optional]]/e',
```
dummdidumm, line 226: Found one more case which didn't sort correctly, for which I had to bring back the "second pass"-logic from my other branch. It doesn't complicate the logic in the sense of explaining it, only the implementation.

dummdidumm: Ah yes, I didn't think this through enough, one would have to be required - yes we can revert it then, and finally merge.

documentation/docs/04-advanced-routing.md
```diff
@@ -63,6 +63,12 @@ export function load(event) {
 
 > If you don't handle 404 cases, they will appear in [`handleError`](/docs/hooks#shared-hooks-handleerror)
 
+### Optional parameters
+
+A route like `[lang]/home` contains a parameter named `lang` which is required. Sometimes it's beneficial to make these parameters optional, so that in this example both `home` and `en/home` point to the same page. You can do that by wrapping the parameter in another bracket pair: `[[lang]]/home`
+
+Note that an optional route parameter cannot follow a rest parameter (`[...rest]/[[optional]]`) or vice versa, since this creates ambiguity.
```
Rich-Harris, line 70:
```suggestion
Note that an optional route parameter cannot follow a rest parameter (`[...rest]/[[optional]]`), since parameters are matched 'greedily' and the optional parameter will always be unused.
```

Conduitry: Perhaps "would always be unused"? Because we're not warning people against doing this, we're saying it's disallowed, and so this is what _would_ happen if they were allowed to, not what _will_ happen if they do.

documentation/docs/04-advanced-routing.md
```diff
@@ -63,6 +63,12 @@ export function load(event) {
 
 > If you don't handle 404 cases, they will appear in [`handleError`](/docs/hooks#shared-hooks-handleerror)
 
+### Optional parameters
+
+A route like `[lang]/home` contains a parameter named `lang` which is required. Sometimes it's beneficial to make these parameters optional, so that in this example both `home` and `en/home` point to the same page. You can do that by wrapping the parameter in another bracket pair: `[[lang]]/home`
+
+Note that an optional route parameter cannot follow a rest parameter (`[...rest]/[[optional]]`), since parameters are matched 'greedily' and the optional parameter will always be unused.
```
Conduitry, line 70:
```suggestion
Note that an optional route parameter cannot follow a rest parameter (`[...rest]/[[optional]]`), since parameters are matched 'greedily' and the optional parameter would always be unused.
```

https://github.com/sveltejs/kit/pull/7051#discussion_r999476196

Rich-Harris (approved):

documentation/docs/04-advanced-routing.md
Conduitry: Justice prevails

PR #7083, opened 2022-09-29
[feat] provide `update` to `use:enhance` callback

### LINKED ISSUE #7072: Provide the default callback to `enhance`
### Describe the problem

I've seen people try to add a loading state via enhance a couple times. They do something like this:

```
<form use:enhance={() => {
  loading = true;
  return () => {
    loading = false;
  }
}}>
```

However, they don't realize that once you return a callback from the function passed to enhance, the default behavior after a form submission no longer applies and you have to do the following yourself:

- run `invalidateAll` if the result was a success
- run `applyAction` if the current location matches or the result was a redirect/error

This logic currently happens in the [default callback](https://github.com/sveltejs/kit/blob/fdb166cb80949c2f54093a7787bcd5eb43524666/packages/kit/src/runtime/app/forms.js#L26-L40).

This means if you want to extend `enhance`, e.g. to add a loading state, you have to duplicate that logic, even if your customizations are minimal:

```diff
<form use:enhance={() => {
  loading = true;
  return ({action, result}) => {
+    if (result.type === 'success') {
+      await invalidateAll();
+    }
+
+    if (
+      location.origin + location.pathname === action.origin + action.pathname ||
+      result.type === 'redirect' ||
+      result.type === 'error'
+    ) {
+      applyAction(result);
+    }

    loading = false;
  }
}}>
```

(You could remove the check around `applyAction` and always apply it, but I don't like how that diverges you from the default `enhance` behavior, perhaps without even realizing it).

You can abstract this into a custom action that wraps `enhance` to reduce the boilerplate, but you won't get any updates that are made to the default callback behavior, e.g. the changes in https://github.com/sveltejs/kit/pull/6828.

All you wanted to do was toggle a loading state, but now you have to maintain the default `enhance` behavior as well.

Of course, there are cases where you want to customize the invalidate or `applyAction` behavior. However, I suspect those cases are more the exception than the norm.

Provide an additional argument to the callback returned from enhance. Name is bikesheddable, but for the purposes of this issue I'll call it `defaultBehavior`. The implementation of this method is the same as [fallback_callback](https://github.com/sveltejs/kit/blob/fdb166cb80949c2f54093a7787bcd5eb43524666/packages/kit/src/runtime/app/forms.js#L26-L40). People who want to add some logic after a form submission _and_ preserve the default behavior could call this method instead of recreating the logic themselves. This way, if the default behavior is updated, custom actions calling this method get it for free.

The example above would then look like this:

```html
<form use:enhance={() => {
  loading = true;
  return ({action, result, defaultBehavior}) => {
    defaultBehavior();
    loading = false;
  }
}}>
```

If people want to write enhance callbacks that _always_ do the default callback behavior, then they could create their own action that calls `defaultBehavior` without having to track any updates to the behavior.

### Alternatives considered

One alternative is to keep the status quo and require the user to duplicate the default behavior if desired.

Another alternative is to call `fallback_callback` automatically after the returned callback runs, but also pass a `cancel` arg to the returned callback that would prevent that behavior if desired. 

```html
<form use:enhance={() => {
  loading = true;
  return ({action, result, cancel}) => {
    // invalidate & applyAction called automatically, unless cancel is called
    loading = false;
  }
}}>
```

This makes the default case simpler. However, this has a few downsides:

- it's a breaking change for people returning a callback that didn't want `invalidateAll` and `applyAction` called
- because we need to run the provided callback to find out if `cancel` was called, there isn't a good way to run code after invalidation & applying the action has completed. Supplying `defaultBehavior` gives the user control over when invalidation happens.

### Importance

would make my life easier

### Additional Information

_No response_

dummdidumm:
Closes #7072

### Tests

### Changesets

### DIFF
--- a/documentation/docs/06-form-actions.md
+++ b/documentation/docs/06-form-actions.md
@@ -257,7 +257,7 @@ Without an argument, `use:enhance` will emulate the browser-native behaviour, ju
 
 > By default the `form` property and `$page.form` is only updated for actions that are in a `+page.server.js` alongside the `+page.svelte` because in the native form submission case you would be redirected to the page the action is on
 
-To customise the behaviour, you can provide a function that runs immediately before the form is submitted, and (optionally) returns a callback that runs with the `ActionResult`. Note that if you return a callback, the default behavior mentioned above is not triggered. To get it back, call `defaultBehavior`.
+To customise the behaviour, you can provide a function that runs immediately before the form is submitted, and (optionally) returns a callback that runs with the `ActionResult`. Note that if you return a callback, the default behavior mentioned above is not triggered. To get it back, call `update`.
 
 ```svelte
 <form
@@ -268,9 +268,9 @@ To customise the behaviour, you can provide a function that runs immediately bef
 		// `action` is the URL to which the form is posted
 		// `cancel()` will prevent the submission
 
-		return async ({ result, defaultBehavior }) => {
+		return async ({ result, update }) => {
 			// `result` is an `ActionResult` object
-			// `defaultBehavior` is a function which triggers the logic that would be triggered if this callback wasn't set
+			// `update` is a function which triggers the logic that would be triggered if this callback wasn't set
 		};
 	}}
 >
@@ -280,7 +280,7 @@ You can use these functions to show and hide loading UI, and so on.
 
 #### applyAction
 
-If you provide our own callbacks, you may need to reproduce part of the default `use:enhance` behaviour, such as showing the nearest `+error` boundary. Most of the time, calling `defaultBehavior` passed to the callback is enough. If you need more customization you can do so with `applyAction`:
+If you provide our own callbacks, you may need to reproduce part of the default `use:enhance` behaviour, such as showing the nearest `+error` boundary. Most of the time, calling `update` passed to the callback is enough. If you need more customization you can do so with `applyAction`:
 
 ```diff
 <script>
@@ -310,7 +310,7 @@ If you provide our own callbacks, you may need to reproduce part of the default 
 
 The behaviour of `applyAction(result)` depends on `result.type`:
 
-- `success`, `invalid` — sets `$page.status` to `result.status` and updates `form` and `$page.form` to `result.data`
+- `success`, `invalid` — sets `$page.status` to `result.status` and updates `form` and `$page.form` to `result.data` (no exceptions in contrast to `update` from `enhance`)
 - `redirect` — calls `goto(result.location)`
 - `error` — renders the nearest `+error` boundary with `result.error`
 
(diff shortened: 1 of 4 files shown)


Rich-Harris:
What about `update` instead of `defaultBehavior`?

documentation/docs/06-form-actions.md
```diff
@@ -257,18 +257,20 @@ Without an argument, `use:enhance` will emulate the browser-native behaviour, ju
...
 > By default the `form` property and `$page.form` is only updated for actions that are in a `+page.server.js` alongside the `+page.svelte` because in the native form submission case you would be redirected to the page the action is on
 
-To customise the behaviour, you can provide a function that runs immediately before the form is submitted, and (optionally) returns a callback that runs with the `ActionResult`.
+To customise the behaviour, you can provide a function that runs immediately before the form is submitted, and (optionally) returns a callback that runs with the `ActionResult`. Note that if you return a callback, the default behavior mentioned above is not triggered. To get it back, call `update`.
 
 ```svelte
 <form
 	method="POST"
-	use:enhance={({ form, data, cancel }) => {
+	use:enhance={({ form, data, action cancel }) => {
```
geoffrich, line 265:
```suggestion
	use:enhance={({ form, data, action, cancel }) => {
```

documentation/docs/06-form-actions.md
```diff
@@ -278,7 +280,7 @@ You can use these functions to show and hide loading UI, and so on.
 
 #### applyAction
 
-If you provide your own callbacks, you may need to reproduce part of the default `use:enhance` behaviour, such as showing the nearest `+error` boundary. We can do this with `applyAction`:
+If you provide our own callbacks, you may need to reproduce part of the default `use:enhance` behaviour, such as showing the nearest `+error` boundary. Most of the time, calling `update` passed to the callback is enough. If you need more customization you can do so with `applyAction`:
```
geoffrich, line 283:
```suggestion
If you provide your own callbacks, you may need to reproduce part of the default `use:enhance` behaviour, such as showing the nearest `+error` boundary. Most of the time, calling `update` passed to the callback is enough. If you need more customization you can do so with `applyAction`:
```

documentation/docs/06-form-actions.md
```diff
@@ -307,7 +310,7 @@ If you provide your own callbacks, you may need to reproduce part of the default
 
 The behaviour of `applyAction(result)` depends on `result.type`:
 
-- `success`, `invalid` — sets `$page.status` to `result.status` and updates `form` and `$page.form` to `result.data`
+- `success`, `invalid` — sets `$page.status` to `result.status` and updates `form` and `$page.form` to `result.data` (no exceptions in contrast to `update` from `enhance`)
```
geoffrich, line 313: Can you expand on this a bit more? I assume you're referring to that `applyAction` will update `form` regardless of the page's location, in contrast to `update`

dummdidumm: I adjusted the defaults explanation in the previous section and enhance the sentence in this - hope this helps

geoffrich: Yeah I think that's better 👍

geoffrich (approved):

PR #7087, opened 2022-09-29
[docs] refine load docs

dummdidumm:
- split up the input properties into common, shared, server
- docs for RequestEvent
- code example for invalidation section
- better separate locals section in handle docs
- link back to load invalidation docs

### Tests

### Changesets


documentation/docs/05-load.md
```diff
@@ -264,6 +219,56 @@ Setting the same header multiple times (even in separate `load` functions) is an
...
+	console.log(data.a); // `undefined`, it wasn't passed through in +page.js
+	console.log(data.b); // `2`
+</script>
+```
+
+In other words `+page.server.js` passes `data` along to `+page.js`, which passes `data` along to `+page.svelte`.
+
+### Inputs exclusive to server-only load functions
+
+In addition to the common inputs, the server-only `load` function provides a few more arguments which it inherits from `RequestEvent`. Namely these are `getClientAddress`, `cookies`, `locals`, `platform` and `request`. See the `RequestEvent` docs for more info.
```
theetrain, line 270: Can we add a link to `RequestEvent`'s docs from here since it's called out?

documentation/docs/05-load.md
```diff
@@ -362,9 +367,28 @@ A `load` function will re-run in the following situations:
...
 - All active `load` functions were forcibly re-run with [`invalidateAll()`](/docs/modules#$app-navigation-invalidateall)
 
+```js
+/// file: +page.js
+// @filename: ambient.d.ts
+declare function doStuffWith(arg: any): void;
+
+// @filename: index.js
+// ---cut---
+import { error } from '@sveltejs/kit';
```
theetrain, line 380: I don't see `error` being used in the `load` below. Maybe this can be removed.

packages/kit/types/ambient.d.ts
```diff
@@ -174,6 +174,10 @@ declare module '$app/navigation' {
 	/**
 	 * Returns a Promise that resolves when SvelteKit navigates (or fails to navigate, in which case the promise rejects) to the specified `url`.
 	 *
+	 * Note that navigating to another page that reuses the current `+page.svelte` will not remount the component.
+	 * Also note that not all `load` functions might be run again.
+	 * See [the load docs on invalidation](https://kit.svelte.dev/docs/load#invalidation) for more information.
```
theetrain, line 179: I think this will be very helpful 👍

theetrain (commented):
Thanks for streamlining the docs! Once I re-read the final copy, I'll get a feel for how a developer can learn about all the loading and invalidation logic in order to build pages with predictable re-loads.

documentation/docs/05-load.md
Rich-Harris: Links are applied automatically

image

documentation/docs/05-load.md
```diff
@@ -264,6 +219,56 @@ Setting the same header multiple times (even in separate `load` functions) is an
...
+<script>
+	/** @type {import('./$types').PageData} */
+	export let data;
+
+	console.log(data.a); // `undefined`, it wasn't passed through in +page.js
+	console.log(data.b); // `2`
+</script>
+```
+
+In other words `+page.server.js` passes `data` along to `+page.js`, which passes `data` along to `+page.svelte`.
```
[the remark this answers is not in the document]
Rich-Harris: The current wording is correct. `+page.server.js` -> `+page.js` -> `+page.svelte`

documentation/docs/05-load.md
```diff
@@ -362,9 +367,28 @@ A `load` function will re-run in the following situations:
...
 
+```js
+/// file: +page.js
+// @filename: ambient.d.ts
+declare function doStuffWith(arg: any): void;
+
+// @filename: index.js
+// ---cut---
+import { error } from '@sveltejs/kit';
+
```
Rich-Harris, line 381:
```suggestion
```

documentation/docs/05-load.md
```diff
@@ -362,9 +367,26 @@ A `load` function will re-run in the following situations:
...
+// @filename: ambient.d.ts
+declare function doStuffWith(arg: any): void;
+
+// @filename: index.js
+// ---cut---
+/** @type {import('./$types').LayoutServerLoad} */
+export function load({ url, params, parent, fetch, depends }) {
+	const response = await fetch('https://some-api.com'); // load reruns when invalidate('https://some-api.com') is called
+	depends('custom:key'); // load reruns when invalidate('custom:key') is called
+	doStuffWith(url); // load reruns when URL changes
```
Rich-Harris, line 384: This isn't quite right, it's only individual properties of `url`

```suggestion
	doStuffWith(url.pathname); // load reruns when the URL changes
```

Rich-Harris:
Docs are failing to deploy because the server-only `load` function doesn't yet include `fetch`. We could update the PR or (better) we could add `fetch` to `RequestEvent`

documentation/docs/05-load.md
```diff
@@ -362,9 +367,26 @@ A `load` function will re-run in the following situations:
...
 - All active `load` functions were forcibly re-run with [`invalidateAll()`](/docs/modules#$app-navigation-invalidateall)
 
+```js
+/// file: +page.js
+// @filename: ambient.d.ts
+declare function doStuffWith(arg: any): void;
+
+// @filename: index.js
+// ---cut---
+/** @type {import('./$types').LayoutServerLoad} */
```
dummdidumm, line 380:
```suggestion
/** @type {import('./$types').PageLoad} */
```

dummdidumm:
I went the "adjust the example" route - the `fetch` thing would be nice, but feels too unrelated to this PR.

documentation/docs/05-load.md
```diff
@@ -362,9 +367,26 @@ A `load` function will re-run in the following situations:
...
 
+```js
+/// file: +page.js
+// @filename: ambient.d.ts
+declare function doStuffWith(arg: any): void;
+
+// @filename: index.js
+// ---cut---
+/** @type {import('./$types').PageLoad} */
+export function load({ url, params, parent, fetch, depends }) {
```
Rich-Harris, line 381: this is causing the docs to fail to build

```suggestion
export async function load({ url, params, parent, fetch, depends }) {
```

documentation/docs/05-load.md
```diff
@@ -362,9 +367,26 @@ A `load` function will re-run in the following situations:
...
+declare function doStuffWith(arg: any): void;
+
+// @filename: index.js
+// ---cut---
+/** @type {import('./$types').PageLoad} */
+export async function load({ url, params, parent, fetch, depends }) {
+	const response = await fetch('https://some-api.com'); // load reruns when invalidate('https://some-api.com') is called
+	depends('custom:key'); // load reruns when invalidate('custom:key') is called
+	doStuffWith(url.pathname); // load reruns when the URL changes
+	doStuffWith(params.foo); // load reruns when the foo parameter changes
```
Rich-Harris, line 385: another deploy failure

```suggestion
/// file: src/routes/items/[id]/+page.js
// @filename: ambient.d.ts
declare function doStuffWith(arg: any): void;

// @filename: index.js
// ---cut---
/** @type {import('./$types').PageLoad} */
export async function load({ url, params, parent, fetch, depends }) {
	const response = await fetch('https://some-api.com'); // load reruns when invalidate('https://some-api.com') is called
	depends('custom:key'); // load reruns when invalidate('custom:key') is called
	doStuffWith(url.pathname); // load reruns when the URL changes
	doStuffWith(params.id); // load reruns when the foo parameter changes
```

documentation/docs/05-load.md
```diff
@@ -362,9 +367,26 @@ A `load` function will re-run in the following situations:
...
+declare function doStuffWith(arg: any): void;
+
+// @filename: index.js
+// ---cut---
+/** @type {import('./$types').PageLoad} */
+export async function load({ url, params, parent, fetch, depends }) {
+	const response = await fetch('https://some-api.com'); // load reruns when invalidate('https://some-api.com') is called
+	depends('custom:key'); // load reruns when invalidate('custom:key') is called
+	doStuffWith(url.pathname); // load reruns when the URL changes
+	doStuffWith(params.id); // load reruns when the foo parameter changes
```
dummdidumm, line 385:
```suggestion
	doStuffWith(params.id); // load reruns when the id parameter changes
```

documentation/docs/05-load.md
```diff
@@ -264,6 +219,56 @@ Setting the same header multiple times (even in separate `load` functions) is an
 
 You cannot add a `set-cookie` header with `setHeaders` — use the [`cookies`](/docs/types#sveltejs-kit-cookies) API in a server-only `load` function instead.
 
+### Inputs exclusive to shared load functions
```
Rich-Harris, line 222: oof, these double decker list items

image


```suggestion
### Shared-only inputs
```

documentation/docs/05-load.md
```diff
@@ -264,6 +219,56 @@ Setting the same header multiple times (even in separate `load` functions) is an
...
+	export let data;
+
+	console.log(data.a); // `undefined`, it wasn't passed through in +page.js
+	console.log(data.b); // `2`
+</script>
+```
+
+In other words `+page.server.js` passes `data` along to `+page.js`, which passes `data` along to `+page.svelte`.
+
+### Inputs exclusive to server-only load functions
```
Rich-Harris, line 268:
```suggestion
### Server-only inputs
```

documentation/docs/05-load.md
```diff
@@ -264,6 +219,56 @@ Setting the same header multiple times (even in separate `load` functions) is an
 
 You cannot add a `set-cookie` header with `setHeaders` — use the [`cookies`](/docs/types#sveltejs-kit-cookies) API in a server-only `load` function instead.
 
+### Shared-only inputs
```
geoffrich, line 222: The term "shared-only" was confusing to me, since I don't think we use this term to refer to `+page.js`/`+layout.js` `load`s anywhere else. Is there something more descriptive?

documentation/docs/05-load.md
```diff
@@ -264,6 +219,56 @@ Setting the same header multiple times (even in separate `load` functions) is an
...
+
+In addition to the common inputs, a `load` function that runs on both the server and the client has the following inputs:
```
geoffrich, line 224: I guess this helps clarify the heading, but it may be worth calling out +page and +layout for further clarity

packages/kit/types/ambient.d.ts
```diff
@@ -174,6 +174,10 @@ declare module '$app/navigation' {
 	/**
 	 * Returns a Promise that resolves when SvelteKit navigates (or fails to navigate, in which case the promise rejects) to the specified `url`.
 	 *
+	 * Note that navigating to another page that reuses the current `+page.svelte` will not remount the component.
+	 * Also note that not all `load` functions might be run again.
```
geoffrich, line 178: Seems like slightly awkward phrasing - maybe this reads better?
```suggestion
	 * Also note that some `load` functions will not be re-run after a navigation.
```

packages/kit/types/index.d.ts
```diff
@@ -337,17 +337,52 @@ export interface ParamMatcher {
...
+	/**
+	 * The route id of the current page - e.g. for `src/routes/blog/[slug]`, it would be `blog/[slug]`.
+	 */
 	routeId: string | null;
+	/**
+	 * If you need to set headers for the response, you can do so using the `setHeaders` method. This is useful for example if you want the page to be cached.
+	 *
+	 * Setting the same header multiple times is an error — you can only set a given header once.
+	 *
+	 * You cannot add a `set-cookie` header with `setHeaders` — use the [`cookies`](/docs/types#sveltejs-kit-cookies) API instead.
```
geoffrich, line 380: We use a relative docs URL here but absolute URLs everywhere else. Should it be one or the other? I'd prefer relative URLs, but I'm not sure if we need to include the hostname for some other reason.

packages/kit/types/ambient.d.ts
dummdidumm: "will" sounds like it would be certain, which it isn't. I want that "maybe, maybe not"-notion be present in the sentence, but I agree it could be phrased better.

geoffrich: Maybe this then?

```suggestion
	 * Also note that some `load` functions might not be re-run after a navigation.
```

packages/kit/types/index.d.ts
dummdidumm: Oversight - it should be absolute, so you can click on it when hovering the docs in your IDE

documentation/docs/05-load.md
Rich-Harris: I like 'universal load' but 'universal inputs' sounds like it applies everywhere. Given that the only one is `data` I wonder if the better solution is to restructure this section slightly instead

dummdidumm:
One thing I realized which may make it so hard to structure this right is that we maybe approach this wrong in the first place. Right now most of this section reads like API reference, mixed with some context and explanations around it. I think it would be better to move much of this into the API section in a slimmed down version. This section should instead be crafted around the task people want to achieve - getting data into the page, how and when to update that data, when to use shared vs server load functions. We can probably reuse what's there already but not centered around the API, rather the problems and solutions.

dummdidumm:
Created #7174 which "implements" my thoughts from the previous comment

Rich-Harris:
closed via #7174

PR #7097, opened 2022-09-30
[fix] decode routeIds in headers for prerendering

### LINKED ISSUE #7074: Prerender fails when the route contains Non-ASCII Characters
### Describe the bug

Prerendering process fails during setting the response header of `x-sveltekit-routeid` when routeId contains Non-ASCII Characters. https://github.com/sveltejs/kit/blob/master/packages/kit/src/runtime/server/index.js#L301

```sh
TypeError: Cannot convert argument to a ByteString because the character atindex 0 has a value of 21021 which is greater than 255.
```

### Reproduction

https://github.com/kwchang0831/issue-sveltekit-prerender-unicode-path

To Reproduce:
1. Create a route with Non-ASCII Characters. eg: `src/routes/初めまして/+page.svelte`.
2. Set `export const prerender = true;` in `+layout.ts`.
3. Run `pnpm build`

### Logs

_No response_

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    CPU: (12) x64 Intel(R) Xeon(R) CPU D-1541 @ 2.10GHz
    Memory: 21.55 GB / 31.35 GB
    Container: Yes
    Shell: 3.5.1 - /usr/bin/fish
  Binaries:
    Node: 18.7.0 - ~/.asdf/installs/nodejs/18.7.0/bin/node
    Yarn: 1.22.19 - ~/.asdf/shims/yarn
    npm: 8.15.0 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Firefox: 105.0.1
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.80 
    @sveltejs/kit: next => 1.0.0-next.505 
    svelte: ^3.44.0 => 3.50.1 
    vite: ^3.1.0 => 3.1.4
```

### Severity

### Additional Information

What I have tried to work around the issue:

Encode the routeId

https://github.com/sveltejs/kit/blob/720735d59b8bca3ba7fcd37e07cc9e5ae004ae2d/packages/kit/src/runtime/server/index.js#L300-L302

```js
  response.headers.set('x-sveltekit-routeid', encodeURI(event.routeId));
```

https://github.com/sveltejs/kit/blob/720735d59b8bca3ba7fcd37e07cc9e5ae004ae2d/packages/kit/src/core/prerender/prerender.js#L422-L426

```js
  if (prerender === true && !prerendered_routes.has(encodeURI(route_id))) {
```

I am not sure if this is the correct way to do, and this assumes that the routeID was not previously encoded with `encodeURI`.

Also, while I was trying to figure out the issue, I found that the routing will fail if the route is already encoded. eg: `src/routes/%E5%88%9D%E3%82%81/+page.svelte`. ~~I guess I should probably sumbit another issue for that.~~ The issue has submitted at #7075.

dummdidumm:
Fixes #7074

From what I can see the proposed changes in the issue are correct, there's nothing encoding the routeId prior to the code where it is then encoded for the header

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/prerender/prerender.js
+++ b/packages/kit/src/core/prerender/prerender.js
@@ -244,10 +244,13 @@ export async function prerender() {
 
 			const prerender = headers['x-sveltekit-prerender'];
 			if (prerender) {
-				const route_id = headers['x-sveltekit-routeid'];
-				const existing_value = prerender_map.get(route_id);
-				if (existing_value !== 'auto') {
-					prerender_map.set(route_id, prerender === 'true' ? true : 'auto');
+				const encoded_route_id = headers['x-sveltekit-routeid'];
+				if (encoded_route_id != null) {
+					const route_id = decodeURI(encoded_route_id);
+					const existing_value = prerender_map.get(route_id);
+					if (existing_value !== 'auto') {
+						prerender_map.set(route_id, prerender === 'true' ? true : 'auto');
+					}
 				}
 			}
 
@@ -305,7 +308,8 @@ export async function prerender() {
 
 		if (written.has(file)) return;
 
-		const route_id = response.headers.get('x-sveltekit-routeid');
+		const encoded_route_id = response.headers.get('x-sveltekit-routeid');
+		const route_id = encoded_route_id != null ? decodeURI(encoded_route_id) : null;
 		if (route_id !== null) prerendered_routes.add(route_id);
 
 		if (response_type === REDIRECT) {
--- a/.changeset/pretty-pigs-give.md
+++ b/.changeset/pretty-pigs-give.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] decode routeIds in headers for prerendering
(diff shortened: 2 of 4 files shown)


packages/kit/src/core/prerender/prerender.js
```diff
@@ -244,10 +244,13 @@ export async function prerender() {
 
 			const prerender = headers['x-sveltekit-prerender'];
 			if (prerender) {
-				const route_id = headers['x-sveltekit-routeid'];
-				const existing_value = prerender_map.get(route_id);
-				if (existing_value !== 'auto') {
-					prerender_map.set(route_id, prerender === 'true' ? true : 'auto');
+				const encoded_route_id = headers['x-sveltekit-routeid'];
+				if (encoded_route_id != null) {
```
benmccann, line 248: I might have just checked for truthiness would be a bit simpler to write as well as more robust to undefined or empty string. same below on line 312
```suggestion
				if (encoded_route_id) {
```

benmccann (approved):

packages/kit/src/core/prerender/prerender.js
dummdidumm: That would be wrong because the root route is the empty string

Rich-Harris: I sometimes wonder if route IDs should start with `/` for that reason. I've been caught out by that once or twice. Would be a breaking change, but a fairly minor one

Issue #7029, opened 2022-09-25
Static adapter complains about non-existant dynamic routes 


brgrz:
### Describe the bug

This is related to https://github.com/sveltejs/kit/issues/4157 but that issue is marked post-1.0 and I believe this should hopefully be resolved before that.

With the default (new) setup of a SvelteKit app a static adapter fails when doing `npm run build` **even when there are no dynamic routes present at all, with the following error:**

```
 @sveltejs/adapter-static: all routes must be fully prerenderable (unless using the 'fallback' option — see https://github.com/sveltejs/kit/tree/master/packages/adapter-static#spa-mode). Try adding `export const prerender = true` to your root layout.js — see https://kit.svelte.dev/docs/page-options#prerender for more details
Error: Encountered dynamic routes
```

I have no dynamic routes set up, there is only the default `page.svelte` file.

Also why does it mention layout.js, what file would that be? seems to me this hasn't been updated since the latest routing changes

I went and added both `prerender` and `ssr` options to the page (not both at the same time) but the issue remains

```
<script lang="ts">
	export const prerender = true;
	// export const ssr = false;
</script>

<h1>Welcome to SvelteKit</h1>
<p>Visit <a href="https://kit.svelte.dev">kit.svelte.dev</a> to read the documentation</p>
```

this is my svelte.config.js

```
import adapter from '@sveltejs/adapter-static';
import preprocess from 'svelte-preprocess';

/** @type {import('@sveltejs/kit').Config} */
const config = {
	// Consult https://github.com/sveltejs/svelte-preprocess
	// for more information about preprocessors
	preprocess: preprocess(),

	kit: {
		adapter: adapter({
			// default options are shown. On some platforms
			// these options are set automatically — see below
			pages: 'build',
			assets: 'build',
			fallback: null,
			precompress: false
		  })
	}
};

export default config;
```

### Reproduction

Create a new SvelteKit app, add the static adapter into the mix, use it in svelte.config.js then try to build the app.

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (8) x64 Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
    Memory: 5.11 GB / 15.78 GB
  Binaries:
    Node: 16.17.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.10.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 105.0.5195.127
[...]

PR #7108, opened 2022-10-01
Clarified file name in error message as per #7029

jacoblukewood:

### Tests

### Changesets

As per #7029 the displayed error message suggesting the user enable prerendering in the +layout.js/+layout.ts file should include the plus symbol and let the user know that if they are using TS, they should make the change in the .ts file.


Conduitry:
I'm in favor of adding the `+` because what's there now is wrong. I'm hesitant about adding `(+layout.ts if your project is using TypeScript)` because a) this error message is already pretty long and b) mentioning this sort of thing everywhere we mention a JS file is a slippery slope that will start to bog down all of our documentation and messages.

As a compromise, I could probably live with `+layout.js/+layout.ts` or `+layout.js/.ts` - but I think we can expect TypeScript users to mostly internally automatically translate references to `.js` into `.ts` without us needing to call it out as a possibility every time.

dummdidumm:
+1 on just adding the +

jacoblukewood:
> I'm in favor of adding the `+` because what's there now is wrong. I'm hesitant about adding `(+layout.ts if your project is using TypeScript)` because a) this error message is already pretty long and b) mentioning this sort of thing everywhere we mention a JS file is a slippery slope that will start to bog down all of our documentation and messages.
> 
> As a compromise, I could probably live with `+layout.js/+layout.ts` or `+layout.js/.ts` - but I think we can expect TypeScript users to mostly internally automatically translate references to `.js` into `.ts` without us needing to call it out as a possibility every time.

I agree that it does make the message quite long. I think `+layout.js/.ts` is a suitable replacement. I have made the suggested changed to the PR.

Rich-Harris (approved):

Rich-Harris:
thank you

PR #7113, opened 2022-10-01
Add `fetch` to `RequestEvent`

### LINKED ISSUE #6320: Add fetch method to ServerLoadEvent similar to LoadEvent to address same issues
### Describe the problem

The implementation of fetch for LoadEvent has a number of benefits highlighted here - https://kit.svelte.dev/docs/load#input-methods-fetch. 

Would it make sense to have a similar implementation of fetch for ServerLoadEvent? There are many cases  where this would be beneficial where you have +page.server.ts functionality accessing an internal REST API:

1. Using relative URLs
2. Credentialed requests on the server - inherits the cookie and authorization headers from the original page request
3. Internal requests go direct to the handler function when running on the server, without the overhead of an HTTP call 

A fetch method for ServerLoadEvent that works exactly the same ways LoadEvent.

### Alternatives considered

_No response_

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
Closes #6320. This moves SvelteKit's `fetch` wrapper to `RequestEvent` so that it can be used by server-only `load` functions and `+server.js` handlers as well as shared `load` functions. (It can even be used in `handle`, or — somewhat pointlessly — inside `handleFetch`, as a consequence of where it now lives.)

Marking as draft because we're no longer using (or able to use) `GENERIC_ERROR`, and I'm not quite sure what the consequences of that are just yet.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -96,7 +96,7 @@ export async function respond(request, options, state) {
 	/** @type {Record<string, string>} */
 	const headers = {};
 
-	const { cookies, new_cookies } = get_cookies(request, url);
+	const { cookies, new_cookies, get_cookie_header } = get_cookies(request, url);
 
 	if (state.prerendering) disable_search(url);
 
@@ -215,7 +215,6 @@ export async function respond(request, options, state) {
 					error: null,
 					branch: [],
 					fetched: [],
-					cookies: [],
 					resolve_opts
 				});
 			}
@@ -229,7 +228,15 @@ export async function respond(request, options, state) {
 				} else if (route.endpoint && (!route.page || is_endpoint_request(event))) {
 					response = await render_endpoint(event, await route.endpoint(), state);
 				} else if (route.page) {
-					response = await render_page(event, route, route.page, options, state, resolve_opts);
+					response = await render_page(
+						event,
+						route,
+						route.page,
+						options,
+						state,
+						resolve_opts,
+						get_cookie_header
+					);
 				} else {
 					// a route will always have a page or an endpoint, but TypeScript
 					// doesn't know that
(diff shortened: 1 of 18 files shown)


Conduitry:
Will the `fetch` that's passed to `handleFetch` now be identical to the `event.fetch` that's passed? It'd be a breaking change, but would this be a reason to stop passing `fetch` to keep things tidy?

Rich-Harris:
> Will the `fetch` that's passed to `handleFetch` now be identical to the `event.fetch` that's passed?

No. `event.fetch` invokes `handleFetch({ event, request, fetch })`; if the `fetch` argument was equivalent to `event.fetch` then _it_ would call `handleFetch({ event, request, fetch })`, and you'd keep going round in circles.

packages/kit/src/runtime/server/index.js
```diff
@@ -295,7 +299,7 @@ export async function respond(request, options, state) {
 							response.headers.set(key, /** @type {string} */ (value));
 						}
 					}
-					add_cookies_to_headers(response.headers, Array.from(new_cookies.values()));
+					add_cookies_to_headers(response.headers, Array.from(Object.values(new_cookies)));
```
repsac-by, line 302:
```suggestion
					add_cookies_to_headers(response.headers, Object.values(new_cookies));
```
Since `Object.values()` returns `Array`

Rich-Harris: d'oh. thanks!

Conduitry:
I'm not sure where best to document this, but I do think we're probably overdue for a little bit of an explanation somewhere about how the different instances of `fetch` interact. I don't think it's enough to just add `fetch` as a method to the `RequestEvent` type. There are a number of nuances about browser vs server behavior, and about credential handling, and about calling `handle` instead of making a network request for same-origin `fetch`es, etc. etc.

Rich-Harris:
Agree, it could be clearer. Do you think that blocks this PR, or are we good to merge it in the meantime? #7087 is blocked on this one (though there is some discussion in that issue about rethinking docs as well)

Conduitry:
@Rich-Harris Nope, I'm fine foregoing docs updates for now if we're going to be rethinking these docs in the context for #7087 anyway.

PR #7115, opened 2022-10-02
[docs] error create note on route `+page/layout.server.js`

mashanz:
creating `+page.server.js` without `+page.svelte` or `+layout.server.js`  without `+layout.svelte` will throw error and it not mentioned on documentation. Mention it on docs will make easier to notify people when creating `*.server.js`

### Tests

### Changesets


Rich-Harris:
Per https://twitter.com/Rich_Harris/status/1576397214611447808, this is a bug that we need to fix, not something that should be documented

PR #7149, opened 2022-10-04
Fix route sorting

Rich-Harris:
This is a breaking change, though I expect its impact to be very minimal.

#7051 made me realise that our route sorting algorithm is flawed — because it works left-to-right, given these two routes...

```
/[a]/b/c
/a/[b]/[c]
```

...a pathname like `/a/b/c` would match `/a/[b]/[c]` first, even though `/[a]/b/c` is more specific (or rather, _feels_ more specific, since we don't have a clearly articulated theory of specificity). The optional params sorting introduces more confusion: https://github.com/sveltejs/kit/pull/7051#discussion_r984913063

This PR changes things by computing a 'score' based on the average 'value' of each segment/part. It results in what I think is a more helpful and intuitive sort order:

https://github.com/sveltejs/kit/blob/cc4422c7ab3f52b4112e59cbd47788f40ad9c5be/packages/kit/src/core/sync/create_manifest_data/index.spec.js#L223-L245

The one bit I'm not sure about is `[...rest]/[required]` coming after `[required]`, but a solution that changes that order without changing other things is eluding me. 

### Tests

### Changesets


benmccann:
I would actually expect `/a/[b]/[c]` to be matched before `/[a]/b/c` because `a` is more specific than `[a]` and I'd expect matching to happen left-to-right. That also seems easier to explain than a scoring algorithm

Rich-Harris:
Do you have a proposal for a sorting algorithm that works that way but also ranks `/[[optional]]/foo` above `/[required]`?

arxpoetica:
I went back and forth on `/a/[b]/[c]` vs `/[a]/b/c`, etc., and ultimately decided 🤷🏻‍♂️.

My vote is _it might not matter_, but the **ClearlyArticulatedTheoryOfSpecificity™️** could be published—scoring or sorting—somewhere to help people not trip up.

Rich-Harris:
> The one bit I'm not sure about is `[...rest]/[required]` coming after `[required]`, but a solution that changes that order without changing other things is eluding me.

Actually I take that part back. If you have do have both of those routes, then `[required]` _must_ take precedence otherwise it couldn't be matched.

Still trying to figure out what sort of algorithm satisfies all our intuitions about the correct order

dummdidumm:
A rundown of (maybe conflicting) thoughts we have:
- we want `[required]` to be matched _after_ `[[optional]]/foo` so that `foo` matches the optional route first
- similarly, we may want `[required]` to be matched _after_ `[...a]/foo` so that `foo` matches the rest route first
- we want `[required]` to be matched _before_ `[[optional]]` and both _before_ `[...rest]` because of specificity
- we want `[...a]` to be matched _after_ `[...b]/foo` so that `foo` matches the more specific rest route first. If we did it the other way around, `[...b]/foo` could _never_ be matched
- we want the sorting to be intuitive and easy to explain
- we want a sorting that requires as little folder depth consideration as possible, so it's easier for people to look at their directory tree (ideally only those at the same level) and say "oh yeah this matches first"

The first two seem like the most controversial (personally it feels wrong to me), the others seem sensible.

If the first two would be discarded, i.e. `[required]` _always_ preceedes `[[optional]]` which _always_ preceedes `[...rest]`, then matching could stay greedy: Compare the current segment. If it's different, order them `hardcoded` < `[required]` < `[[optional]]` < `[...rest]`. If they are the same, look at the next one (this also solves the `[...a]` vs `[...b]/foo` case).

What's left to figure out under this assumption is the specificity _within_ a segment. To me it seems the best to also sort them greedily:
1. `a-b-c`
2. `a-b-[[optional]]`
3. `a-[required]-b`
4. `[required]-b-c`
5. `[required]-[[optional]]-c`
6. `[[optional]]-b-c`

This may feel slightly weird for 5 vs 6 the same way it may seem weird for people preferring the first two requirements in the first section above, but it's easy to explain and understand and consistent with the segment ordering.

To conclude, to me personally the current system (which AFAIK works greedily) is good, and the new optional params should be sorted between required and rest.

Rich-Harris:
Closing in favour of #7278

PR #7159, opened 2022-10-05
[fix] clear out build directory when running `adapter-netlify`

benmccann:
We used to do this and somehow lost it, which has been causing me frustration. I'm pretty sure we do it in the other adapters though I didn't confirm that


Rich-Harris:
looks like we do indeed do this in most cases

Issue #7125, opened 2022-10-02
Can't cache a page load (setting "cache-control" with `setHeaders` doesn't seem to work)


maximedupre:
### Describe the bug

For example, setting the cache here has no effect. 

```
export const load: PageLoad = async ({ fetch, setHeaders }) => {
    const res = await fetch(`/api/users`);

    if (!res.ok) throw error(res.status);

    const users: UserV2[] = await res.json();

    setHeaders({
        'cache-control': 'max-age=86400',
    });

    return {
        users,
    };
};
```

Maybe it's because the initial request to the page contains "Cache-Control: max-age=0"?

So the current behavior is that even if I set the "cache-control" header, the request is always processed by the server (instead of being served by disk/memory cache).

The expected behaviour is that the request is served from cache after being cached (obviously)

### Reproduction

Unfortunately, I can't really debug "cache-control" using node.new 🤔 The network panel doesn't give me access to the "sandboxed" requests.

### Logs

Raw response headers

```shell
HTTP/1.1 304 Not Modified
Access-Control-Allow-Origin: *
cache-control: max-age=86400
etag: "1ons99h"
Date: Sun, 02 Oct 2022 23:00:59 GMT
Connection: keep-alive
Keep-Alive: timeout=5
```

Raw request headers

```
GET /top-secret HTTP/1.1
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.9
Cache-Control: max-age=0
Connection: keep-alive
Cookie: ...
Host: localhost:3000
If-None-Match: "1ons99h"
Referer: http://localhost:3000/top-secret
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: same-origin
Sec-GPC: 1
Upgrade-Insecure-Requests: 1
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/106.0.0.0 Safari/537.36
```

### System Info

```shell
System:
    OS: macOS 12.6
    CPU: (8) arm64 Apple M2
    Memory: 2.30 GB / 24.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.7.0 - ~/.nvm/versions/node/v18.7.0/bin/node
[...]

PR #7174, opened 2022-10-07
[docs] rewrite load docs

### LINKED ISSUE #7148: Documentation of Advanced Layout Features
### Describe the problem

I think the additional layout features aren't structured very apparent. I actually only stumbled across those by going through the migration guide issues and totally overlooked them in the documentation.

I think the simplest solution would be to atleast provide cross links from the actual "layout" section to the "advanced layout" section.

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

dummdidumm:
Superseedes #7087 
Closes #7148 

This rewrites the load docs in a way that is more grouped around use cases and problems you want to solve. It also adds API docs for the `load` functions.

While I was at it, I also added a link checker to check for broken links, which promptly found some outdated links which I also udpated.

### Tests

### Changesets


documentation/docs/05-load.md
```diff
@@ -93,85 +146,176 @@ An instance of [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL), co
...
+	- you access private environment variables
+	- the returned data is serializable
+- use shared `load` in `+page.js`/`+layout.js` when
+	- you want more control over the `load` function
+	- you want to avoid a network hop when the resource is on a different server
+	- you need to return non-serializable data
+	- you are building an SPA served from a static file server
+- use both `load` functions in case you have a combination of requirements
+
+### Making fetch requests
```
dummdidumm, line 260: @Conduitry in #7113 you mentioned you wanted some more details about the subtle difference of `fetch` and when to use it - this would be the section to add it to.

documentation/docs/05-load.md
```diff
@@ -2,65 +2,118 @@
...
-	console.log(data.a); // `undefined`, it wasn't passed through in +page.js
-	console.log(data.b); // `2`
+	console.log(data.a); // 'a'
+	console.log(data.b); // 'b'
+	console.log(data.c.value); // `Promise {...}`
 </script>
 ```
 
-In other words `+page.server.js` passes `data` along to `+page.js`, which passes `data` along to `+page.svelte`.
+The `load` functions related to a certain page run in parallel by default, avoiding a waterfall of requests. Once all `load` functions have returned, rendering starts. This allows layouts and pages to access data from upper layouts. In the above example, the `data` prop in `blog/+page.svelte` _also_ has access to the `postsCount` returned from `blog/+layout.server.js`. Additionally, you can access the merged data of _all_ `load` functions through `$page.data`.
```
geoffrich, line 93: Oh, I didn't realize you could get layout data from the `data` prop directly. Does this include all nested layouts as well for a page?

documentation/docs/05-load.md
```diff
@@ -93,85 +146,176 @@ An instance of [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL), co
...
 > In some environments this is derived from request headers during server-side rendering. If you're using [adapter-node](/docs/adapters#supported-environments-node-js), for example, you may need to configure the adapter in order for the URL to be correct.
 
-### Input methods
+### Which load function to use
 
-`LoadEvent` also has the following methods:
+The `load` function is available in both `+page.js`/`+layout.js` and `+page.server.js`/`+layout.server.js` files. They share the same name because they do the same - providing data to a layout or page. Their capabalities differ in some ways though, and it makes sense to use one or the other depending on the situation - sometimes even both.
 
-#### depends
+If the `load` function is defined in `+page.server.js` or `+layout.server.js` ("server-only `load` function") it will only run on the server, in which case it can (for example) make database calls and access private [environment variables](/docs/modules#$env-static-private), but can only return data that can be serialized with [devalue](https://github.com/rich-harris/devalue).
```
geoffrich, line 153: This sentence is a bit lengthy. Maybe:
```suggestion
If the `load` function is defined in `+page.server.js` or `+layout.server.js` ("server-only `load` function") it will only run on the server, in which case it can (for example) make database calls and access private [environment variables](/docs/modules#$env-static-private). However, it can only return data that can be serialized with [devalue](https://github.com/rich-harris/devalue).
```

documentation/docs/05-load.md
```diff
@@ -93,85 +146,176 @@ An instance of [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL), co
...
+- use server-only `load` in `+page.server.js`/`+layout.server.js` when
+	- you want to access your data base directly
+	- you access private environment variables
+	- the returned data is serializable
+- use shared `load` in `+page.js`/`+layout.js` when
+	- you want more control over the `load` function
+	- you want to avoid a network hop when the resource is on a different server
+	- you need to return non-serializable data
+	- you are building an SPA served from a static file server
+- use both `load` functions in case you have a combination of requirements
```
geoffrich, line 258: For consistency, maybe:
```suggestion
- use both `load` functions when you have a combination of requirements
```

documentation/docs/05-load.md
```diff
@@ -93,85 +146,176 @@ An instance of [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL), co
...
+
+### Making fetch requests
 
-`fetch` is equivalent to the [native `fetch` web API](https://developer.mozilla.org/en-US/docs/Web/API/fetch), with a few additional features:
+Inside server-only `load` functions you can directly query your data base. Inside shared `load` functions you can't, as they also may run on the client. Also sometimes your data base may not be on the same server as your app. For these reasons it is often necessary to make `fetch` requests.
```
geoffrich, line 262:
```suggestion
Inside server-only `load` functions you can directly query your database. Inside shared `load` functions you can't, as they also may run on the client. Also sometimes your data base may not be on the same server as your app. For these reasons it is often necessary to make `fetch` requests.
```

documentation/docs/05-load.md
```diff
@@ -93,85 +146,176 @@ An instance of [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL), co
...
+/// file: src/routes/blog/+page.server.js
+/** @type {import('./$types').PageServerLoad} */
+export async function load({ fetch, setHeaders }) {
+	const url = `https://cms.example.com/articles.json`;
+	const response = await fetch(url);
+
+	setHeaders({
+		age: response.headers.get('age'),
+		'cache-control': response.headers.get('cache-control')
+	});
```
geoffrich, line 290: This doesn't actually work right now, see #7125 and #6477

geoffrich (commented):
This is a big improvement, thanks!

documentation/docs/05-load.md
```diff
@@ -93,85 +146,176 @@ An instance of [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL), co
...
+	// Increase count each time the load function is run, but only in the browser
+	const returnedCount = count;
+	if (browser) {
+		count++;
 	}
+	return { count: returnedCount };
 }
+```
+
+Sometimes it's even benefitial to use both `load` functions. For example you could first return some data from your database, and use that to create an object that is nonserializable and therefore couldn't be returned from the server `load` function. To access the server data in the shared `load` function, use the `data` input:
```
cdcarson, line 203:
```suggestion
Sometimes it's even beneficial to use both `load` functions. For example you could first return some data from your database, and use that to create an object that is non-serializable (so therefore couldn't be returned directly from the server `load` function.) To access the server data in the shared `load` function, use the `data` input:
```

documentation/docs/05-load.md
```diff
@@ -335,37 +434,93 @@ If an _unexpected_ error is thrown, SvelteKit will invoke [`handleError`](/docs/
...
+SvelteKit tracks the dependencies of each `load` function to avoid re-running it unnecessarily during navigation. For example, a `load` function in a root `+layout.js` doesn't need to re-run when you navigate from one page to another unless it references `url` or a member of `params` that changed since the last navigation. `load` functions also take into account parent `load` functions, if you reference them through `await parent()` - in this case, if an upper `load` function is rerun, so is the `load` function that does `await parent()`.
+
+`load` functions do not only run when you navigate, they can also be triggered to rerun by `invalidate` or `invalidateAll` (which we get to in the next paragraph). `invalidate` is closely connected to `fetch` and `depends`, which are both methods passed to `load`. Making a `fetch` request automatically registers the fetched URL as a dependency of the load function. `depends` does the same in a manual way and registers the specified URL as a dependency. You can then use `invalidate` and pass one of the registered URLs (or a function, if you need to decide based on a pattern rather than the full URL) to it to rerun the `load` function:
```
cdcarson, line 464: About `depends`. It took a while this morning for me to figure out that you call `depends` in the `load` _you are depending on_, not elsewhere (like a `load` further down the tree.) Maybe this was just because it was my first time using it. I'm fine with the word `depends` now, but maybe we can point out that it means (AFIACT) "define a cache key for this `load` function."

documentation/docs/05-load.md
dummdidumm: Yes, everything above the layout/page is merged into it. Think if it like `$page.data` which stops at the layout/page

documentation/docs/05-load.md
dummdidumm: Yeah I took this snippet from the old docs and later realized that, too. Not sure how important it is to switch to another example given we want to support this eventually. cc @Rich-Harris

documentation/docs/05-load.md
geoffrich: I think this could use more clarification in the docs, then, since this was a surprise to me! I thought page.data was the only way to get all merged data.

Rich-Harris:
Cannot for the life of me figure out why this is failing to deploy. Builds fine locally

Rich-Harris:
Ah, I tell a lie. Running `build` inside `packages/kit` to update the type definitions causes `sites/kit.svelte.dev` to fail to build

Rich-Harris:
This is good stuff, but I can't shake the feeling that it doesn't belong in reference docs. It's a guide. Maybe we should move it to learn.svelte.dev?

dummdidumm:
For me learn.svelte.dev is the place for
- the tutorial (even that I'm not 100% sure on from time to time)
- recipes (how to write your own X)
- guides (how to do auth)

For me this is not a guide in that sense.

What I envision for kit.svelte.dev to have is
- API docs (the thing that's currently split up in types and modules, and which we probably should merge and enhance)
- introductionary docs (later with some cross links to API docs, tutorial, guides) -> this is this rewrite

One thing that I noticed from lurking in the questions channel on Discord is that people have a hard time to understand how to actually use the parts we give them together, and this rewrite tries to answer that for the load function. API or pure reference docs can't solve that, the tutorial can't either because it does things piece by piece instead of giving an overview in one reading, which is why I think a middle ground between tutorial and API docs is needed. The routing docs for example do this nicely because they clearly show the use case and how things fit together. I tried something similar with these docs. Previously the pieces on that page were too disconnected to make sense of them as a whole.

Rich-Harris:
Fair enough — in that case we should probably start thinking about what a better API section would look like.

I have some edits locally (since the GitHub UI is a bad place to make prose edits), will push them up tomorrow

documentation/docs/05-load.md
```diff
@@ -2,253 +2,282 @@
...
-		await invalidate('my-stuff:foo');
-	}
-</script>
+Shared `load` functions are useful when you need to `fetch` data from an external API and don't need private credentials, since SvelteKit can get the data directly from the API rather than going via your server. They are also useful when you need to return something that can't be serialized, such as a Svelte component constructor.
 
-<p>{data.foo}<p>
-<p>{data.bar}</p>
-<button on:click={pageRefresh}>Refresh my stuff</button>
-```
+In rare cases, you might need to use both together — for example, you might need to return an instance of a custom class that was initialised with data from your server.
```
dummdidumm, line 185: Wondering if this would be a more real-world use case that may also help people discover how to do this:
```suggestion
In rare cases, you might need to use both together — for example, you might need to forward an authentication token from the server for subsequent use in shared `load` functions.
```

documentation/docs/05-load.md
```diff
@@ -2,253 +2,282 @@
...
+```
 
-### Input methods
+Type information for `$page.data` is provided by `App.PageData`.
 
-`LoadEvent` also has the following methods:
+### Shared vs server
 
-#### depends
+As we've seen, there are two types of `load` function:
```
dummdidumm, line 158:
```suggestion
As we've seen, there are two types of `load` functions:
```

Rich-Harris: `function` singular is correct!

Conduitry: I've seen PRs before for the svelte.dev docs that wanted to make this sort of grammar change. I think the singular might just be intentional and be Rich's style. It also feels a little bit more formal/British to me.

documentation/docs/05-load.md
Rich-Harris: I don't think we'd want to encourage that pattern. If you have to go the server for an auth token _anyway_ then it makes sense to use it on the server as well — you could get all the data you need in a single hop rather than going to the server for the token then making fetches to the external API from the browser.

The thinking here was that it directly follows from the previous two paragraphs:

1. use server-only load to get stuff from your db
2. use shared load to return funky stuff
3. use both together to return funky stuff initialised with data from the db

documentation/docs/05-load.md
Conduitry: Whoops, thanks for not letting me know Rich had already commented, GitHub.

documentation/docs/05-load.md
dummdidumm: Let's keep it as-is then

documentation/docs/05-load.md
```diff
@@ -2,253 +2,282 @@
...
+
+	return { item };
 }
 ```
 
-In `+page.js` or `+layout.js` it will return data from `load` functions in parent `+layout.js` files. Implicitly, a missing `+layout.js` is treated as a `({ data }) => data` function, meaning that it will also return data from parent `+layout.server.js` files.
+> Cookies will only be passed through if the target host is the same as the SvelteKit application or a more specific subdomain of it.
 
-Be careful not to introduce accidental waterfalls when using `await parent()`. If for example you only want to merge parent data into the returned output, call it _after_ fetching your other data.
+### Cookies and headers
```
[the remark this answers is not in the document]
Rich-Harris: Basically that cookies _aren't_ just a special case of headers — `cookies.set(...)` causes a `set-cookie` header to be appended, but it also updates the cookie store that accompanies the `RequestEvent`. Meanwhile `cookies.get(...)` is nothing to do with `response.headers`, it reads from `request.headers` and the event's cookie store. Headers are the _mechanism_ for dealing with cookies, but I think it's helpful to treat them as somewhat conceptually different — most headers just affect the response, but `set-cookie` affects the user agent.

In general I think it's good to be able to refer back to things that were previously covered, so that you're layering knowledge. Otherwise someone reads about `setHeader`, thinks 'great, I can use this to set a cookie', _then_ learns that actually no, they have to learn this whole other thing.

Rich-Harris (approved):

Issue #6466, opened 2022-08-31
The new endpoint fetching via dynamic import leads to resource leakage, use alternative evaluation instead


Tal500:
### Describe the problem

After PR #6318, instead of fetching JSON response from the endpoint, Kit uses dynamic import directly to fetch server data.
Disclaimer: I'm on the party who thinks this decision is too unorthodox in general, but this isn't what I want to discuss here, since the deed is done, I rather discuss about some symptom of it and how can we fix this.

The problem starts when we wants to have an SPA experience which lasts for long time (e.g. in progressive web apps, but not only), so we don't expect the user to refresh after X number of interaction, at most.
By design, as far as I know, there is no way to de-allocate the resources of dynamic import of ES module. Probably the reason was because it was expected to have a limited number of modules you wish to load, and the dynamic support for import was added not for allowing you to import infinite amount of resources, but rather for allowing you to lazy/on-demand import your module for interactive performance balancing.

In the new method (via PR #6318), after every data loading, another module is imported, which means the browser keeps the module resource in the memory, and never releases it - and there is no way even to tell the browser to.

### Describe the proposed solution

Since the fetched JS code is (and should) be very basic, like `window.some_var = {}`, I'm suggesting two alternative for the resourceful import:
1. Fetch the code from the server by the classic `fetch()`, and use the JS function `eval()` to evaluate the code.
2. Use a self-destructing script loading (like a classical JQuery-era techniques), something like:
```js
new Promise((resolve, reject) => {
  const script = document.createElement('script');
  script.src = url;
  const remove = () => { document.head.removeChild(script) };
  script.onload = () => {
    remove();
    resolve();
  };
  script.onerror = () => {
    remove();
    reject(`Error while trying to inject <script src="${url}"> to the head element.`);
  };
  document.head.appendChild(script);
})
```

Option 1 for sure is best&secure memory-wise (and also simple for caching challenges), however I think that the evaluation of option 2 might be slightly faster.

### Importance

nice to have

### Additional Information

I chose the importance to be "nice to have", but I think it's much more important for stability. This issue can be thought between feature and bug.

dummdidumm:
AFAIK this was considered while designing this approach, and that the window method deallocates the module. Do you have concrete data that this happens?

Tal500:
> AFAIK this was considered while designing this approach, and that the window method deallocates the module. Do you have concrete data that this happens?

I saw no official technical claim about it, but I'm using only my logic here.
[...]

PR #7177, opened 2022-10-07
[fix] load data through regular json request

dummdidumm:
using devalue's stringify/parse methods
Part of #7125 and #6477

Not sure exactly what's left for the above issues to get closed besides allowing setting cache headers again.

If we change to this we could use the same methods for interacting with form actions, allowing them to also return `Date` objects for example.

Marked as draft, @Rich-Harris feel free to take over/close.

### Tests

### Changesets


Rich-Harris:
> If we change to this we could use the same methods for interacting with form actions, allowing them to also return `Date` objects for example.

The wrinkle here is that if someone posts to an action manually, they'll get devalue-encoded data in return. We could handle it in SSR and `use:enhance` but would presumably need to expose something like

```js
import { decode } from '$app/forms';
```

or, more directly,

```js
import { decode } from 'devalue'; // doesn't currently exist, but decode(obj) would === parse(JSON.stringify(obj))
```

I think it's fine to merge this as-is then handle headers in a separate PR, so I'm marking this ready

packages/kit/src/runtime/client/client.js
```diff
@@ -10,6 +16,7 @@ import { HttpError, Redirect } from '../control.js';
 import { stores } from './singletons.js';
 import { DATA_SUFFIX } from '../../constants.js';
 import { unwrap_promises } from '../../utils/promises.js';
+import * as devalue from 'devalue';
```
dummdidumm, line 19: will tree shaking remove the other stuff correctly if we write the import like this?

dummdidumm (commented):
Can't approve because it's my own PR, but consider this an approval in case the answer to my question in the code-comment is "yes".
Edit: Oh and this needs a changeset.

packages/kit/src/runtime/client/client.js
Rich-Harris: Yep, Rollup doesn't care either way

Rich-Harris (approved):

0gust1:
Thanks, I'll check if it fixes #7203 !

Tal500:
I'm glad to see this PR fixes a theoretical consideration on a closed issue of mine:
fixes #6466

PR #7182, opened 2022-10-07
[docs] add typescript code snippets

### LINKED ISSUE #4452: toggling between `JSDoc` and `TypeScript` examples in docs
### Describe the problem

As someone who prefers to write TypeScript code, having all examples on the docs page annotated with JSDoc makes e.g copy & pasting annoying because I have to convert it manually to TypeScript syntax. For me, a person that knows both syntaxes well, this might be a small problem.

But I already saw a few people struggle because they were new to JSDoc comments and didn't know how the equivalent should look in TypeScript code. Having a proper TypeScript example would help them a lot.

A toggle could be provided somewhere on the docs page that allows to globally choose between `JSDoc` and `TypeScript`

Similar to the [Vue.js documentation](https://vuejs.org/guide/introduction.html) that allows to toggle between API styles.

### Alternatives considered

JSDoc only examples

### Importance

nice to have

### Additional Information

_No response_

dummdidumm:
...by adding a toggle that shows snippets either in TS or JS code. The toggle is implemented using CSS and a blocking script which adds a class to the html tag, which makes the correct one of the two snippets appear. The TS versions are generated from the JS versions through a transformation step.
Closes #4452

Marked as draft because ~`$app/X` imports are not recognized in the TS version for some reason; and~ I don't know where to place the toggle. It's at the top of the navigation sidebar for testing, but that obviously isn't a good solution. I see basically two options:
1. Some fixed place on the page so you toggle JS/TS globally
2. Each snippet that has a TS version gets some toggle icon on the top right. It would also toggle all other snippets to TS so you don't have to do it for each snippet. One way to achieve this with the given setup would be to put both the JS and generated TS version into one code snippet, separating them through some special comment. Then when marked invokes the `code` method we separate them again, transform them independently and then wrap them inside some block which would have the toggle visible if JavaScript is enabled in the browser.

### Tests

### Changesets


sites/kit.svelte.dev/src/lib/docs/server/index.js
```diff
@@ -392,3 +408,153 @@ export function slugify(title) {
 		.replace(/^-/, '')
 		.replace(/-$/, '');
 }
+
+/**
+ * Appends a JS->TS / Svelte->Svelte-TS code block after each JS/Svelte code block.
+ * The language is `generated-js`/`generated-svelte` which can be used to detect this in later steps.
+ *
```
benmccann, line 415:
```suggestion
```

sites/kit.svelte.dev/src/lib/docs/server/index.js
```diff
@@ -392,3 +408,153 @@ export function slugify(title) {
...
+				code.replace(outer, `<script lang="ts">${ts}</script>`) +
+				'\n```'
+			);
+		});
+}
+
+/**
+ * Transforms a JS code block into a TS code block by turning JSDoc into type annotations.
+ * Due to pragmatism only the cases currently used in the docs are implemented.
+ *
```
benmccann, line 443:
```suggestion
```

benmccann:
This is really amazing!!

I only noticed one small thing. It looks like something may have gone wrong processing this section: https://kit-svelte-3l0mx9kuo-svelte.vercel.app/docs/web-standards#fetch-apis-headers. It might be nice if we could figure out how to get a unit test for the fix as I might be scared to touch the docs generation code too much currently

Rich-Harris:
This is badass. I think a global toggle makes more sense, since it's a global setting, and it's less noise than repeating the toggle for every snippet. Maybe something like this (made by editing the input/output toggle on the mobile view of https://svelte.dev/repl)?

image

Very minor point but I think it's probably more idiomatic to have the `import type` declarations at the end of the import block — is that easy to do?

dummdidumm:
Added a better-looking global toggle

TODO/questions:
- should move to top with nav if nav is hidden -> needs update to site-kit so we know when nav is hidden
- should the TS enabled state be orange? Is this too distracting? Should they both be grey?
- REPL uses the same toggle - should this go into site-kit?

benmccann:
Just FYI, I tried to bump `site-kit` to the latest version the other day, but was unable to. I haven't gotten to see what the issue is. It seems the latest release may have accidentally dropped a `.css` file from being included or something

Rich-Harris:
There's a ton of consolidation that ought to happen with site-kit. At some point I'd love to do a big spring clean of it, and redesign parts of the Kit docs at the same time — the design language has evolved somewhat over the years, and it would be great if everything looked more like https://learn.svelte.dev.

So yeah, the toggle probably should go into site-kit but I don't think we should worry about it right now.

> should move to top with nav if nav is hidden -> needs update to site-kit so we know when nav is hidden

Galaxy brain — move it to the bottom:

image

I think this is ready! Too excited not to go ahead and merge this

Rich-Harris (approved):

weavdale:
Thanks so much for this update!!! I love being able to toggle to Typescript. I didn't notice the toggle right away though since its at the bottom of the page with same background color as main nav.

PR #7184, opened 2022-10-08
Hoist AMP script preload for minor performance improvement 🛥️ 

sanjaiyan-dev:
Hoist AMP script preload for minor performance improvement.

### DIFF
--- a/packages/amp/index.js
+++ b/packages/amp/index.js
@@ -1,7 +1,7 @@
 // https://amp.dev/documentation/guides-and-tutorials/learn/spec/amp-boilerplate/
 const boilerplate = `
+    <link rel="preload" as="script" href="https://cdn.ampproject.org/v0.js" />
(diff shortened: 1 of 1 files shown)


packages/amp/index.js
```diff
@@ -1,7 +1,7 @@
 // https://amp.dev/documentation/guides-and-tutorials/learn/spec/amp-boilerplate/
 const boilerplate = `
+    <link rel="preload" as="script" href="https://cdn.ampproject.org/v0.js" />
```
benmccann, line 3: this should use a tab instead of spaces to align with the other lines

benmccann:
I can't believe this would make any difference. However, if we were going to make this change, wouldn't you also want to move up `<script async src="https://cdn.ampproject.org/v0.js"></script>`?

sanjaiyan-dev:
> I can't believe this would make any difference. However, if we were going to make this change, wouldn't you also want to move up `<script async src="https://cdn.ampproject.org/v0.js"></script>`?

Sorry, actually I thought this would have very minor performance improvement because it would have least small improvement than current implementation of adding `link:rel=preload` just before the script tag.

sanjaiyan-dev:
> It _does_ seem doubtful that it would make any difference, and it's very surprising to me that the `<link>` is worth the extra bytes when the `<script>` immediately follows it, but this _does_ in fact match the current recommended [AMP boilerplate](https://amp.dev/boilerplate/), which appears to have changed since this was first created.

Ok, I actually made these changes after viewing them on next js implementation of AMP.
They had it like-:
   1) link preload
   2) Boilerplate CSS 
   3) Script tag

Rich-Harris:
thanks!

sanjaiyan-dev:
> thanks!

🤝💪

PR #7195, opened 2022-10-09
[fix] parsing content-type header for actions

### LINKED ISSUE #7187: Incorrect content-type parsing in actions
### Describe the bug

When using a simple fetch to an endpoint like so:
fetch('/login', {method: "POST", body: new URLSearchParams('key', 'value')})

This is sent by Chrome and Safari as a POST request with the following content-type header:
Content-Type: application/x-www-form-urlencoded;charset=UTF-8

But https://github.com/sveltejs/kit/blob/master/packages/kit/src/runtime/server/page/actions.js checks the content-type value by splitting on "; " (note the space). 

Therefore I cannot use the above fetch call because the endpoint returns:
Actions expect form-encoded data (received application/x-www-form-urlencoded;charset=UTF-8)

I think this is a bug and that you should instead split on ";" (without space).

### Reproduction

Use a simple fetch to an endpoint (/login/+page.server.js) like so:
fetch('/login', {method: "POST", body: new URLSearchParams('key', 'value')})

### Logs

_No response_

### System Info

```shell
any system
```

### Severity

### Additional Information

_No response_

repsac-by:
Fixes #7187

### Tests

### Changesets

### DIFF
--- a/.changeset/hot-shoes-bow.md
+++ b/.changeset/hot-shoes-bow.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Fix parsing content-type header for actions
--- a/packages/kit/src/runtime/server/page/actions.js
+++ b/packages/kit/src/runtime/server/page/actions.js
@@ -180,7 +180,7 @@ export async function call_action(event, actions) {
 		throw new Error(`No action with name '${name}' found`);
 	}
 
-	const type = event.request.headers.get('content-type')?.split('; ')[0];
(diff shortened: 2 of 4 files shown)


tcc-sejohnson:
Did you see the note from @Conduitry about making sure we're doing this for CSP? I haven't looked yet to see if any changes are necessary -- just want to leave this not as a "TODO".

repsac-by:
I saw this note, but I don't see how CSRF can be related to incorrect header parsing.

dummdidumm (approved):
The joys of coding with the VS Code web extension without any intellisense 😄

PR #7198, opened 2022-10-09
Add alternative prefix for actions

### LINKED ISSUE #7098: Using / in query params for actions with AWS Lambda function URLs
### Describe the problem

SvelteKit uses `/` as a signal character in query parameter keys. AWS Lambda Function URLs will not handle requests with a `/` in the query params. They just respond with a 400. I have opened a support ticket with them as well and they say the reasoning is that according to section 3.4 in this rfc https://www.ietf.org/rfc/rfc2396.txt, `/` is a reserved character.

> 3.4. Query Component
> 
>    The query component is a string of information to be interpreted by
>    the resource.
> 
>       query         = *uric
> 
>    Within a query component, the characters ";", "/", "?", ":", "@",
>    "&", "=", "+", ",", and "$" are reserved.

I have let them know that many other services allow it and handle it just fine. Including some other AWS services like CloudFront. So now the ticket is being "escalated" to the lambda team. Probably escalated to the bottom of their backlog.

In the SvelteKit config you could designate a different character. But to not break using other adapters it would have to be some sort of middle man fix. 

Ideally, I would still want to use a `/` in my form action to not break compatibility with other adapters.
```html
<form method='post' action='?/someAction'>
</form>
```

But, I guess this would be in the compilation phase, SvelteKit would transform those to the alternative signal character. SvelteKit would also be looking for the alternative signal character on the server.

Okay, this sucks too, ~~but I can't think of a better way either~~. Maybe alternative 4.

### Alternatives considered

1. Change the signal character in SvelteKit to no longer be a `/` and instead be something that is not reserved. I don't necessarily think this is a great option though as everybody has already migrated and is using `/` and realistically this is an edge case. `/` makes for pretty natural looking urls when using actions which is a nice bonus.

2. The solution proposed by AWS is that I can url encode the / every time (%2F I believe). This sucks because then I have to go to all my forms and do this ugly mess

```html
<form method='post' action='?%2FsomeAction'>
</form>
```

And then make sure in my adapter to decode it before passing to SvelteKit. This is terrible for multiple reasons. The first being looking at that makes me want to puke and also would break using any other adapter.

3. Of course I could basically implement the proposed solution of using a different signal character just in an adapter. Basically same steps of alternative 2, but instead of decoding `%2F` to `/` before SvelteKit serves the request, translate `#` to `/` or something similar. I don't know if that is even valid, but just and example. This maybe sucks less than having to write `%2F` all over my code base, but I am still breaking compatibility with other adapters which sucks.

4. Just thought of this one. There could be a SvelteKit config option to, in the compilation phase, url encode the /. The server should already be able to handle that without any adapter changes. This may be the best option as there is quite a bit less magic then the proposed solution. The main downside is kind of ugly urls, but I can live with that, especially since this is somewhat of an edge case.

5. Yell loudly enough that AWS cares to fix this.

### Importance

would make my life easier

### Additional Information

I'll update this if anything comes back from the lambda team, but I figure it would be unhealthy to hold my breath.

repsac-by:
Closes #7098 

I propose to use a second alternative prefix character  `.` for actions along with `/`
`.` is not on the reserved list of legacy [rfc2396](https://www.ietf.org/rfc/rfc2396.html#section-3.4)
there are no special notes for it in [rfc3986](https://www.ietf.org/rfc/rfc3986.html#section-3.4) and does not require escaping
 It also looks like an [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) `action="?.login"`

```js
// '.login='
new URLSearchParams({'.login': ''}).toString();

// '%2Flogin='
new URLSearchParams({'/login': ''}).toString();
```

An alternative would be to specify the prefix in the configuration, but that seems a bit redundant.

### Tests

### Changesets

### DIFF
--- a/documentation/docs/06-form-actions.md
+++ b/documentation/docs/06-form-actions.md
@@ -63,7 +63,7 @@ export const actions = {
 };
 ```
 
-To invoke a named action, add a query parameter with the name prefixed by a `/` character:
+To invoke a named action, add a query parameter with the name prefixed by a `/` or `.` character:
 
 ```svelte
 /// file: src/routes/login/+page.svelte
--- a/packages/kit/src/runtime/server/page/actions.js
+++ b/packages/kit/src/runtime/server/page/actions.js
@@ -165,9 +165,9 @@ export async function call_action(event, actions) {
 	const url = new URL(event.request.url);
 
 	let name = 'default';
-	for (const param of url.searchParams) {
-		if (param[0].startsWith('/')) {
-			name = param[0].slice(1);
+	for (const param of url.searchParams.keys()) {
+		if (param[0] === '/' || param[0] === '.') {
+			name = param.slice(1);
 			if (name === 'default') {
 				throw new Error('Cannot use reserved action name "default"');
 			}
--- a/packages/kit/test/apps/basics/src/routes/actions/enhance/+page.svelte
+++ b/packages/kit/test/apps/basics/src/routes/actions/enhance/+page.svelte
@@ -16,6 +16,7 @@
 <form method="post" action="?/login" use:enhance>
 	<input name="username" type="text" />
 	<button class="form1">Submit</button>
+	<button class="form1-prefix" formAction="?.login">Submit</button>
 	<button class="form1-register" formAction="?/register">Submit</button>
 	<button class="form1-submitter" formAction="?/submitter" name="submitter" value="foo">Submit</button>
 </form>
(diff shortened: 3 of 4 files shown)


Rich-Harris:
Thank you for the PR, but I'm going to close this as it's a (somewhat alarming!) bug in Lambda that can be worked around by slapping API Gateway in front of it. Having multiple ways to do something is always bad unless there's an exceptionally good reason, and 'Lambda is buggy' feels like a bad one.

PR #7221, opened 2022-10-11
[fix] ensure serialized headers check is always applied

### LINKED ISSUE #7112: Content-Type for responses with sveltes fetch always text/plain on client
### Describe the bug

I am fetching ressources from a database on the same server (also happening on localhost) and after migrating to latest svelte-kit (before was version next.3xxx) i observe there is a difference in the headers between server and client fetch.

grafik

Unfortunately i need to check the headers and call res.json reps. res.text accordingly and this is not working anymore.

### Reproduction

i tried to recreate this on stackblitz but get CORS issues (see https://stackblitz.com/edit/sveltejs-kit-template-default-uyfzch?file=src/routes/test/+page.js)
however locally you can easily test it by doing this:

0) Create sveltekit from scratch, no changes.
1) Create a /test route with an empty +page.svelte
2) create a +page.js with
```js
export async function load({parent ,fetch}) {

    let res = await fetch("https://run.mocky.io/v3/2daed5d1-bfdd-4777-8e0b-a60e3ea8af52", {
        method: "GET",
        headers: {
            "Accept": "application/json"
        }
    });

    console.log(res.headers.get("content-type")); //Should be application/json; but is text/plain during client side rendering
    res = await res.json();
    console.log(res);
    
    return { };
}
```

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    Memory: 17.19 GB / 31.92 GB
  Binaries:
    Yarn: 1.22.5 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 8.11.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (105.0.1343.53)
  npmPackages:
    @sveltejs/adapter-node: ^1.0.0-next.96 => 1.0.0-next.96
    @sveltejs/kit: ^1.0.0-next.507 => 1.0.0-next.507
    svelte: ^3.50.1 => 3.50.1
    vite: ^3.1.4 => 3.1.4
```

### Severity

### Additional Information

_No response_

dummdidumm:
Closes #7112 - if we decide this is enough, and that we don't want some minimum whitelist.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/load_data.js
+++ b/packages/kit/src/runtime/server/page/load_data.js
@@ -129,23 +129,6 @@ export async function load_data({
 								response_body: body,
 								response: response
 							});
-
-							// ensure that excluded headers can't be read
-							const get = response.headers.get;
-							response.headers.get = (key) => {
-								const lower = key.toLowerCase();
-								const value = get.call(response.headers, lower);
-								if (value && !lower.startsWith('x-sveltekit-')) {
-									const included = resolve_opts.filterSerializedResponseHeaders(lower, value);
-									if (!included) {
-										throw new Error(
-											`Failed to get response header "${lower}" — it must be included by the \`filterSerializedResponseHeaders\` option: https://kit.svelte.dev/docs/hooks#handle`
-										);
-									}
-								}
-
-								return value;
-							};
 						}
 
 						if (dependency) {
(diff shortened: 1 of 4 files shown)


packages/kit/src/runtime/server/page/load_data.js
```diff
@@ -186,6 +169,23 @@ export async function load_data({
 				}
 			});
 
+			// ensure that excluded headers can't be read
+			const get = response.headers.get;
```
dummdidumm, line 173: I'm wondering why the proxy is used instead of monkey patching the other methods as well - convenience?

Also: Should we apply this check in DEV only?

Rich-Harris (approved):

packages/kit/src/runtime/server/page/load_data.js
Rich-Harris: It would be more convenient to monkey-patch the `Response` methods instead of using the proxy, but we get 'illegal invocation' errors. For whatever reason that doesn't apply to `Headers` so we may as well do the convenient thing there.

I think it's fine to keep the error in production, since it doesn't result in extra code shipped to the client and might help someone diagnose a bug

PR #7223, opened 2022-10-11
sync: account for relative route path

gbkwiatt:
Currently when using relative route path in `config.kit.files.routes` e.g. `../../project/src/routes` sync fails because it generates types outside of `.svelte-kit/types`  It should account for relative paths and generate them inside `.svelte-kit/types`

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/sync/write_types/index.js
+++ b/packages/kit/src/core/sync/write_types/index.js
@@ -66,7 +66,7 @@ export async function write_all_types(config, manifest_data) {
 	for (const route of manifest_data.routes) {
 		if (!route.leaf && !route.layout && !route.endpoint) continue; // nothing to do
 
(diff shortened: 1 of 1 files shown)


dummdidumm:
Thanks! Could you add a changeset?

gbkwiatt:
> Thanks! Could you add a changeset?
> 
> Also, to be sure: Does this really fix the mentioned issue completely? You talk about some manifest not being found there, and the changed code is only related to generating types.

The issue was, if I wasn't using relative path for routes, sync was passing but build was failing due to wrong paths. But if I want to change routes path  to match vite manifest (it always resolves to relative path), it was creating it in wrong place due to being relative and sync was filing and not creating `route_meta_data.json`. Fixing the sync solve it for that use case.

dummdidumm (approved):

PR #7225, opened 2022-10-11
add custom error message for `throw invalid()` in actions

### LINKED ISSUE #7224: Unhelpful error message for `throw invalid()`
### Describe the bug

Using `throw invalid()` instead of `return invalid()` in an action displays an unhelpful error message in the server console: `[object Object]`

My silly brain decided to spend over an hour trying to debug this without realising I was using `throw` where I needed to use `return` (despite using form actions since it came out).

It may be helpful to return a more useful error message so someone else doesn't fall into the same trap.

### Reproduction

Press the submit button.

Expected: action returns form values to page.

Actual: error message in console: [object Object]

https://stackblitz.com/edit/sveltejs-kit-template-default-q9n9ns?file=src/routes/+page.server.js

### Logs

```shell
[object Object]
```

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.81 
    @sveltejs/kit: next => 1.0.0-next.512 
    svelte: ^3.46.0 => 3.51.0 
    vite: ^3.1.0 => 3.1.7
```

### Severity

annoyance

### Additional Information

_No response_

s3812497:
fixes #7224 
Throws an error that suggests a fix instead of displaying `[object Object]`.

The new error message:
`Cannot "throw invalid()". Use "return invalid()"`

### Tests

### Changesets

### DIFF
--- a/.changeset/pretty-scissors-march.md
+++ b/.changeset/pretty-scissors-march.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+add helpful error message for `throw invalid()` in form actions
--- a/packages/kit/src/runtime/server/page/actions.js
+++ b/packages/kit/src/runtime/server/page/actions.js
@@ -62,10 +62,6 @@ export async function handle_action_json_request(event, options, server) {
 			});
 		}
 
-		if (!(error instanceof HttpError)) {
-			options.handle_error(error, event);
-		}
-
 		return action_json(
 			{
 				type: 'error',
--- a/packages/kit/src/runtime/server/page/actions.js
+++ b/packages/kit/src/runtime/server/page/actions.js
@@ -69,19 +69,22 @@ export async function handle_action_json_request(event, options, server) {
 		return action_json(
 			{
 				type: 'error',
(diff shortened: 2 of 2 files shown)


dummdidumm:
Thank you! Could add the same at ~line 57 and add a changeset?

s3812497:
> Thank you! Could add the same at ~line 57 and add a changeset? Also, just throwing there isn't quite correct, there is a return with type error at the end where you should add the logic. Soemthing like `return { type: 'error': error: error instanceof Invalid ? new Error('cannot ..') : error`

Thanks for the guidance! Are the new changes correct? I wasn't sure if creating a function was warranted for returning the same error message in both `handle_action_request` and `handle_action_json_request`

dummdidumm:
Extracting that logic into a small function in the same file sounds good to me - other than that this looks 👍

dummdidumm (approved):

Issue #7107, opened 2022-10-01
[feat] pass data from `locals` to `load`


benmccann:
### Describe the problem

My site queries an external API with the following pattern:
- `handle` extracts user ID from cookie
- `+layout.svelte` fetches user by user ID
- `+page.svelte` fetches profile by user ID

These fetches previously happened in parallel. With `$session` removed I have no way to write this code without introducing a waterfall

### Describe the proposed solution

I'm happy with any solution folks can think of that solves this problem

I think we could potentially solve this as part of https://github.com/sveltejs/kit/issues/7105

### Importance

i cannot use SvelteKit in a performant manner without it

pilcrowOnPaper:
Duplicate of #6714, no?

pilcrowOnPaper:
I store the locals value in a global variable after the initial load on the browser and so I can access it on subsequent load functions that run in the browser. Doesn't solve the issue on the initial server load though.

ZetiMente:
I also have this same problem using Supabase.

dummdidumm:
Copying my Discord-thoughts into here:

What became clear to me that getSession was used for multiple things at once.

1. pass data from server to client load functions
2. make cross-cutting data easily available across all client load functions
3. trigger a load function to rerun when data changes (e.g. $session.user = null)
4. a convenient writable store that is somewhat global

Because it had so many use cases in once, it became somewhat confusing what runs when, because people may only think about 4 and then wonder about 3. So I think removing it in its form was correct, but we need some features that bring back the mentioned points in a easy-to-use way. Right now we have this situation:
1. pass data ->  can do this now by having a +layout.server.js and passing whatever you want to the client
2. cross-cutting data -> use await parent() -> this introduces a waterfall because you can't tell "hey I'm only interested in the data from parent X, don't wait on the 3 other layout loads after that". Something like locals for the client side would help with this.
3. trigger rerun -> there's nothing built-in for this right now. Rich shared a pattern somewhere about creating a custom state which triggers a unique invalidate function under the hood. Maybe it's enough to make this a blessed pattern and/or tidy it up a little and provide a wrapper around it.
4. convenient writable store -> again, nothing built-in, this would be #7105
3+4. It's not clear to me how to achieve a combination of these two in a nice API. session made it possible because it was on both sides.

--> I feel like we need to add recipes and/or APIs to ease this

dummdidumm:
[...]

PR #7232, opened 2022-10-11
shared state

david-plugge:
This is mainly just an exploration on how to solve #7107. Please leave some feedback!

```ts
// src/app.d.ts
declare namespace App {
	interface Locals {
		user?: MyUserType;
	}
	interface Session {
		user?: MyUserType;
		value: number;
		myClientOnlyValue?: number;
	}
}

// src/hooks.server.ts
import type { GetServerSession, Handle } from '@sveltejs/kit';

export const handle: Handle = async ({ event }) => {
	event.locals.user = await getTheUser(event);
	return resolve(event);
}

export const getServerSession: GetServerSession = ({ locals }) => {
	return {
		user: locals.user,
		value: 0
	};
};

// --- override or initialize if you don´t have a server (adapter static)
// src/hooks.client.ts
import type { GetClientSession } from '@sveltejs/kit';

export const getClientSession: GetClientSession = ({ session }) => {
	return {
		...session,
		myClientOnlyValue: 123
	};
};

// src/routes/+page.ts

export const load: PageLoad = async ({ session }) => {
	console.log(session);
	if (session.user) {
		session.value++;
	}
};
```

```html
<script lang="ts">
	import { invalidateAll } from '$app/navigation';

	$: console.log('page', $session);

	function update() {
		$session.user = { username: 'RandomUser' };
		invalidateAll();
	}
</script>

<button on:click={update}>Update user</button>
```

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/sync/write_client_manifest.js
+++ b/packages/kit/src/core/sync/write_client_manifest.js
@@ -100,6 +100,9 @@ export function write_client_manifest(config, manifest_data, output) {
 				handleError: ${
 					hooks_file ? 'client_hooks.handleError || ' : ''
 				}(({ error }) => { console.error(error) }),
+				getClientSession: ${
+					hooks_file ? 'client_hooks.getClientSession || ' : ''
+				}(({ session }) => session),
 			};
 		`)
 	);
--- a/packages/kit/src/exports/vite/build/build_server.js
+++ b/packages/kit/src/exports/vite/build/build_server.js
@@ -121,6 +121,7 @@ export class Server {
 
 			this.options.hooks = {
 				handle: module.handle || (({ event, resolve }) => resolve(event)),
+				getSharedState: module.getSharedState || (() => ({})),
 				handleError: module.handleError || (({ error }) => console.error(error.stack)),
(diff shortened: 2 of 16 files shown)


CalebBassham:
This seems like a more complex version of $session which was removed. Also, it seems like it will be hard to securely update a shared state on the client side.

dummdidumm:
Thank you! This pretty much looks like what `getSession` looked like, so it inherits all it's advantages but also disadvantages. 

Things that bugged me personally:
- no way to init an initial state on the client side if you don't have a server

Things that confused users:
- having load functions rerun if something was changed in session, even if they don't depend on that part of the state in that load function

Not sure how to best solve this without falling into the same disadvantages again yet.

david-plugge:
> * no way to init an initial state on the client side if you don't have a server

In that case a `getSharedState` hook inside _src/hooks.client.ts_ could work.

> * having load functions rerun if something was changed in session, even if they don't depend on that part of the state in that load function

We could make the `state` object a proxy and track which properties are accessed. But that will get very complicated and probably doesn´t make a lot of sense. So the invalidation part should probably stay on the user side eg. `invalidate` and `depends`.

Issue #7166, opened 2022-10-06
External routes failing on `sync` or `build`


gbkwiatt:
### Describe the bug

I have monorepo, where I have template that has dynamic routing that feeds all my apps. 

so each app in svelte.config.js has `files.routes: '../../template/src/routes'`

Now with above config I get, after install on `svelte-kit  sync` 
`ENOENT: no such file or directory, open '/home/user/monorepo/apps/core/.svelte-kit/types/route_meta_data.json'` 

So I figured that't because the routes can't be not inside app directory so changed it to 
``files.routes: 'node_modules/template/src/routes'`` 
Since my `template` package is symlinked in node_modules inside app directory

So now sync worked, but  after migration to latest svelte-kit, build fails because in vite manifest those routes resolve to `../../template/src/routes/...` and when build tries to find it in vite manifest it searches for `node_modules/template/src/routes` therefore it can't find it. 

so going back to first example, build works, but sync fails. 

So... how can I make both work ? 

### Reproduction

Have an app with routes referenced from outside of the project

### System Info

```shell
System:
    OS: Linux 5.10 Ubuntu 20.04.4 LTS (Focal Fossa)
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
    Memory: 5.13 GB / 9.72 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.1.0 - ~/.nvm/versions/node/v18.1.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.1.0/bin/yarn
    npm: 8.8.0 - ~/.nvm/versions/node/v18.1.0/bin/npm
  npmPackages:
    @sveltejs/adapter-auto: 1.0.0-next.80 => 1.0.0-next.80 
    @sveltejs/adapter-node: 1.0.0-next.82 => 1.0.0-next.82 
    @sveltejs/adapter-static: 1.0.0-next.44 => 1.0.0-next.44 
    @sveltejs/kit: 1.0.0-next.508 => 1.0.0-next.508 
    svelte: ^3.50.1 => 3.50.1 
    vite: ^3.1.4 => 3.1.4
```

### Severity

blocking an upgrade

gbkwiatt:
Creating symlink to router directory results in same error 
`TypeError: Cannot read properties of undefined (reading 'file')`
[...]

PR #7252, opened 2022-10-12
fix: properly convert absolute paths to relative

gbkwiatt:
fixes: #7166

Continuation, because I've noticed that it does not correctly resolve absolute paths to relative - if routes are relative paths outside of `cmd`  Also I think it's more robust  way to get relative path rather than using splice.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/exports/vite/build/build_server.js
+++ b/packages/kit/src/exports/vite/build/build_server.js
@@ -354,7 +354,9 @@ function get_methods(cwd, output, manifest_data) {
 	const lookup = {};
 	output.forEach((chunk) => {
 		if (!chunk.facadeModuleId) return;
-		const id = chunk.facadeModuleId.slice(cwd.length + 1);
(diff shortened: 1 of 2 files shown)


packages/kit/src/exports/vite/build/build_server.js
```diff
@@ -354,7 +354,9 @@ function get_methods(cwd, output, manifest_data) {
 	const lookup = {};
 	output.forEach((chunk) => {
 		if (!chunk.facadeModuleId) return;
-		const id = chunk.facadeModuleId.slice(cwd.length + 1);
+		const relative = path.relative(cwd, chunk.facadeModuleId);
+		// in case it's on windows - replace backslash with forward slash
+		const id = relative.replace(/\\/g, '/');
```
ignatiusmb, line 359:
```suggestion
		const id = posixify(path.relative(cwd, chunk.facadeModuleId));
```

dummdidumm (approved):

PR #7270, opened 2022-10-15
Fixing the Skeleton Template's svelte.config.js

untold-titan:
There isn't an issue opened for this issue, I just noticed it when I created a new Svelte Project. 

When I created the project, and opened it in VSCode, it gave me an error saying that svelte.config.js couldn't find the `adapter` variable. I went and checked the node module for `@sveltejs/adapter-auto` and sure enough, the exported variable is `adapters` not `adapter` so, I updated the import on the config file, and the error went away. 

Not a super big issue, but just something I noticed.

### DIFF
--- a/packages/create-svelte/templates/skeleton/svelte.config.js
+++ b/packages/create-svelte/templates/skeleton/svelte.config.js
@@ -1,11 +1,11 @@
-import adapter from '@sveltejs/adapter-auto';
+import adapters from '@sveltejs/adapter-auto';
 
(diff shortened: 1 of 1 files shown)


Conduitry:
I don't know what issue you're trying to fix, but the change in this PR wouldn't affect anything. `import foo from '...'` imports a default export from the module, regardless of what name `foo` is used. This is in contrast to `import { foo } from '...'` which imports a named export, which isn't what's happening here.

PR #7278, opened 2022-10-16
Route sorting

Rich-Harris:
Follow-up to #7051, supersedes #7149.

This took a looooong time, but I finally think we have a sorting algorithm that delivers the desired outcomes while being relatively simple and can be explained in relatively human-understandable language.

To recap, [this innocuous comment](https://github.com/sveltejs/kit/pull/7051#discussion_r984913063) suggested that in a matchup between these two routes...

```
[[optional]]/static
[required]
```

...we want `[[optional]]/static` to be tested first, because if the pathname is `/static` then it's very unlikely that we want it to match `[required]`. The same goes for `[...rest]` parameters, since these are also optional.

@dominikg's great insight was that `[[optional]]/static` is equivalent to `static` + `[required]/static`, meaning it's actually a _three_ way matchup resulting in this order:

```
static
[required]
[required]/static
```

But since the variant without the dynamic parameter will always outrank the one with it, we don't actually need to try both, we can just try the one without. So in order to determine the sort order of a bunch of different routes, we can just remove all the `[[optional]]` and `[...rest]` parameters. 

There's just one catch — `static` should outrank `static/[...rest]`. We account for that by removing all `[[optional]]` or `[...rest]` parameters _unless_ they're the very last part of a route.

I'm pretty sure this covers all the bases. If not, I will be very unhappy, because I have tried dozens of approaches and my brain is completely melted.

### Tests

### Changesets


dummdidumm (commented):
Good things take a while! This makes a lot of sense to me. I can't think of any loopholes or drawbacks. Just to be sure, a test that `foo/bar` is sorted before `foo/[...baz]/bar` would be nice.

dummdidumm (approved):

Rich-Harris:
closing in favour of #7304

PR #7280, opened 2022-10-17
Direct user to publish `./package/` directory instead of `./`

brev:
After running `svelte-package` (via `create-svelte`), the user is supposed to
run: `npm publish package/`. However, it is easy to accidentally run 
`npm publish` on the wrong directory out of habit. With this change, 
when the user accidentally tries to publish the `./` directory, they will be 
directed to publish the `./package/` directory instead.

In case it is useful, thanks.

### Tests

### Changesets

### DIFF
--- a/.changeset/wild-bags-cheat.md
+++ b/.changeset/wild-bags-cheat.md
@@ -1,0 +1,9 @@
+---
+'create-svelte': patch
+---
+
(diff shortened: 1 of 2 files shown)


dummdidumm (approved):
Thank you!

PR #7281, opened 2022-10-17
Update routing docs so `+server` code example route has file extension.

brev:
Update routing docs so `+server` code example route has file extension.
Route now has a file with extension (`.xml`), instead of bare. It took me a while to figure this out, I don't think there are any other docs about it.
In case it's helpful, thanks.

### Tests

### Changesets


gtm-nayan:
An example for file extensions might be helpful but this is definitely not the right place for it. Just complicates an example and takes away from the point.

Adding it to this list at the top of the page should be sufficient:
image

dummdidumm:
Agree, adding it to that list at the top is better. The example is complicated by the file extension.

Rich-Harris:
Thank you for the PR but I'm inclined to agree that this example is unnecessarily complicated — it makes things _more_ confusing, not less.

> Adding it to this list at the top of the page should be sufficient:

I'm not sure about that. The list is there to illustrate the basics of filesystem-based routing, I don't think it makes sense to add a new entry.

To be honest I'm not sure additional documentation is warranted. This is the first time I've heard of someone getting stuck on this. I can see us adding an example when we create tutorial content, but I don't think it belongs in reference documentation so I'll close this.

Issue #6700, opened 2022-09-09
Support for tools like telefunc


magne4000:
### Describe the bug

[Telefunc](https://telefunc.com/) allows simple frontend to backend calls, avoiding direct call to `fetch`.

```ts
// The `CreateTodo.telefunc.js` file is not loaded:
// Telefunc transforms `*.telefunc.js` imports into a
// thin HTTP client.
import { onNewTodo } from './CreateTodo.telefunc.js'

...

await onNewTodo({ text })
```

Using telefunc + sveltekit can cause the following issues:
- `Cannot import $env/static/private into client-side code`
- `Cannot import $lib/server/xyz.telefunc.ts into client-side code`

Those errors happen at build time and in dev mode, but in dev mode, switching between pages can make it work again. See reproduction repo

### Reproduction

https://github.com/magne4000/sveltekit-telefunc-repro

`pnpm i && pnpm run dev`

The 3 links at the top have the 3 following cases:
- working
- failing because of $env/static/private import
- failing because of $lib/server

### Logs

```shell
[vite-plugin-svelte-kit] Cannot import $lib/server/test3.telefunc.ts into client-side code:
- .svelte-kit/generated/nodes/4.js
  - src/routes/failing/server-folder/+page.svelte
    - $lib/server/test3.telefunc.ts
error during build:
Error: Cannot import $lib/server/test3.telefunc.ts into client-side code:
- .svelte-kit/generated/nodes/4.js
  - src/routes/failing/server-folder/+page.svelte
    - $lib/server/test3.telefunc.ts
    at IllegalModuleGuard.assert_legal (file:///home/magne/workspace/sveltekit-telefunc/node_modules/.pnpm/@sveltejs+kit@1.0.0-next.480_svelte@3.50.1+vite@3.1.0/node_modules/@sveltejs/kit/src/exports/vite/utils.js:47:11)
    at IllegalModuleGuard.assert_legal (file:///home/magne/workspace/sveltekit-telefunc/node_modules/.pnpm/@sveltejs+kit@1.0.0-next.480_svelte@3.50.1+vite@3.1.0/node_modules/@sveltejs/kit/src/exports/vite/utils.js:49:9)
[...]

PR #7288, opened 2022-10-17
[feat] allow users to override IllegalModuleGuard default behaviour

magne4000:
Closes https://github.com/sveltejs/kit/issues/6700.

Allows libs like [telefunc](https://telefunc.com/) (which transforms server code when used client side) to tell sveltekit that it can safely import files otherwise considered as server side on client side.

### Tests

### Changesets


brillout:
Author of Telefunc here. Let me know if there is anything we can do on our side to make progress on this PR.

Very much looking forward to be able to use SvelteKit with Telefunc.

benmccann (changes requested):
this seems like a workaround for some other bug. I'd expect the telefunc vite plugin to remove the private stuff on the client and the sveltekit vite plugin to use the version provided by the telefunc plugin. I'm not sure where that's breaking down, but hopefully we can figure that out

magne4000:
Not sure this is entirely a bug, because:
- telefunc generates different code when `isSsr=false`
- sveltekit executes its dependency check when `isSsr=true`.

I understand now what [@tcc-sejohnson meant](https://github.com/sveltejs/kit/issues/6700#issuecomment-1264734783)
> SvelteKit has no way of knowing that you're "not really" importing it on the client

Indeed, sveltekit has no way to know that telefunc manipulates code when `isSsr=false`, so the error message is understandable.

I tried many [alternative solutions on Telefunc side](https://github.com/magne4000/sveltekit-telefunc-repro/pull/1), but none were satisfactory. It would require https://github.com/vitejs/vite/issues/6810 to be solved for me to test other solutions.

But I still think that a solution on sveltekit side is the way to go. Telefunc doesn't do anything that vite does not allow. On the other hand, sveltekit dependency checks makes assumptions (which are fine 99% of the time) (in this case the assumption is that we have the same dependency tree on server and on client). It seems reasonable to think that it should offer a way to circumvent this check for cases like this one.

PS: Searching for alternative solutions, I stumbled upon [Rollup Inter-plugin communication](https://rollupjs.org/guide/en/#custom-module-meta-data). This seems to be a better approach to handle this PR.

benmccann:
I agree telefunc is doing nothing wrong here. I'm just not sure about this solution

> sveltekit executes its dependency check when isSsr=true.

Yeah, I just realized this too. It seems problematic. I'd hope there's a way we can run the dependency check against the client-side code instead of against the server-side code since what we're worried about is secrets being imported into the client-side code

Rich-Harris:
#7487 does not fix this issue. Reopening

packages/kit/src/exports/vite/graph_analysis/index.js
```diff
@@ -255,11 +267,12 @@ export class ViteImportGraph {
  * @param {(id: string) => import('rollup').ModuleInfo | null} node_getter
  * @param {import('rollup').ModuleInfo} node
  * @param {string} lib_dir
+ * @param {IllegalModuleGuardOptions} [options]
```
benmccann, line 270: I believe the rollup check is working and it's only the vite check that's broken, so I would remove this and have it only affect `dev` and not `build`

benmccann (commented):
If we were going to move forward with this, the following two documentation files would need to be updated with the new option:
- documentation/docs/30-advanced/50-server-only-modules.md
- documentation/docs/50-api-reference/10-configuration.md

I'd also suggest a different name for the option

But while I was writing this up, Rich came up with possibly a better solution. I thought I'd leave my notes here in case we end up coming back to this solution, but for the meantime it's probably not worth addressing these items

brillout:
Also, I'm thinking an alternative would be some kind of internal list of exceptions. It's hackish but on the other hand it would be a practical solution that doesn't involve exposing a new config.

PR #7296, opened 2022-10-17
remove non-browser code from browser builds

Rich-Harris:
This is a small thing but I noticed we're including this code in production client builds, quite unnecessarily. There might be other cases, and after this PR we can just wrap them in a `if (!__SVELTEKIT_BROWSER__)` block to remove them from client builds.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/utils/url.spec.js
+++ b/packages/kit/src/utils/url.spec.js
@@ -2,7 +2,7 @@ import * as assert from 'uvu/assert';
 import { describe } from './unit_test.js';
 import { resolve, normalize_path, make_trackable, disable_search } from './url.js';
 
-// define global required in url.js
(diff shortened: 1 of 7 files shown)


packages/kit/src/utils/url.spec.js
```diff
@@ -2,6 +2,9 @@ import * as assert from 'uvu/assert';
 import { describe } from './unit_test.js';
 import { resolve, normalize_path, make_trackable, disable_search } from './url.js';
 
+// @ts-expect-error define global required in url.js
+globalThis.__SVELTEKIT_BROWSER__ = false;
```
dummdidumm, line 6: didn't find a better way to achieve this. Using uvu's `-r somefile.js` doesn't work because that somehow switches everything to using `require`, which fails.

PR #7303, opened 2022-10-18
[fix] more informative serialization error messages

dummdidumm:

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/actions.js
+++ b/packages/kit/src/runtime/server/page/actions.js
@@ -248,5 +248,9 @@ function check_serializability(value, id, path) {
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


packages/kit/src/runtime/server/page/actions.js
```diff
@@ -248,5 +248,9 @@ function check_serializability(value, id, path) {
 		}
 	}
 
-	throw new Error(`${path} returned from action in ${id} cannot be serialized as JSON`);
+	throw new Error(
+		`${path} returned from action in ${id} cannot be serialized as JSON without loosing its original type` +
```
geoffrich, line 252:
```suggestion
		`${path} returned from action in ${id} cannot be serialized as JSON without losing its original type` +
```

Rich-Harris (approved):

Issue #4656, opened 2022-04-19
Support suffixes for [...rest] endpoints


ngalaiko:
### Describe the problem

I have a project with a following pages routes structure: 

```
/routes/index.svelte
/routes/posts/index.svelte
/routes/posts/index.json.js
/routes/posts/category1/index.svelte
/routes/posts/category2/index.svelte
/routes/posts/[...catchall]/index.json.js
/routes/posts/[...catchall]/index.svelte
```

What I am doing there:
* `/routes/posts/index.{svelte,json.js}` is a list page with every post
* `/routes/posts/category*/**/*.svelte` are posts with just markup and some props
* `/routes/posts/[...catchall]/index.svelte` fires when no post if found, so it queries `/routes/posts/[...catchall]/index.json.js` to check if the post has an alias and redirects

What I would like to do is to add a generic `/likes.json` endpoint to all the posts, so:

```
...
/routes/posts/[...catchall]/likes.json.js
/routes/posts/[...catchall]/index.svelte
/routes/posts/[...catchall]/index.svelte
```

But the problem I am facing is that the request is routed to the `/routes/posts/[...catchall]/index.svelte` handler, even though according to [sorting rules](https://kit.svelte.dev/docs/routing#advanced-routing-sorting), `likes.json` must have a priority

### Describe the proposed solution

Support catchall suffixes 

### Alternatives considered

I could solve this without having `[...rest]` matcher of course, but that involves either copy-paste, or manual routing. 

### Importance

would make my life easier

### Additional Information

This seems like an improvement of an existing [feature](https://github.com/sveltejs/kit/issues/1714)

Rich-Harris:
There _is_ a bug here — because the client-side routing table doesn't include endpoints, the router doesn't 'see' that `/posts/[...catchall]/likes.json.js` is a better match for `/posts/a/b/c/likes.json` than `/posts/[...catchall]/index.svelte`. We might have to include endpoints in the routing table to fix that.

That said, how are you running into this bug? If you hit `/posts/a/b/c/likes.json` directly, or `fetch` it, the data will come directly from the server. If you're _linking_ to the JSON file for some reason, you can workaround the bug by putting [`sveltekit:reload`](https://kit.svelte.dev/docs/a-options#sveltekit-reload) on the link. Is there some other way you're being sent to the wrong route?

Conduitry:
[...]

PR #7304, opened 2022-10-18
Error on conflicting routes

Rich-Harris:
Yet another follow-up to #7051, which can be considered an alternative to #7278. It reverts to the simpler algorithm initially present in that PR, but adds a check for routes that are in conflict with each other (for example `x/[[y]]` and `[[y]]/x`, which can't meaningfully coexist) so that the algorithm doesn't need to consider the additional cases added later in the PR.

The one case it doesn't handle is `x/[...rest]` vs `[...rest]/x` (or `x/[[optional]]` vs `[...rest]/x`), which is the exception to the rule that a route ending with a rest/optional parameters comes after an otherwise-equivalent route. I'm hopeful that we can solve that case without introducing a lot more complexity, but I haven't nailed it yet.

Another test that fails (both in this PR and in #7278) is `foo/bar[...rest]` vs `foo/[...rest]bar`.

### Tests

### Changesets


dummdidumm:
`foo/bar[...rest] vs foo/[...rest]bar` -> is that even allowed? Can rest params appear as part of a segment?

(if they can't, there's still `foo/bar[[optional]] vs foo/[[optional]]bar`, which .. should probably be a route conflict?)

Rich-Harris:
> is that even allowed? Can rest params appear as part of a segment?

It is, yeah — it's come up in the past for things like `[...path].json` — https://github.com/sveltejs/kit/issues/4656. Less likely that you'd need a route like that since the big loading revamp a few months back, but still a legitimate use case.

`foo/bar[[optional]]` and `foo/[[optional]]bar` would be treated as a conflict. The way this PR detects conflicts is it normalizes routes by replacing `[param]` with `<*>` and `[param=x]` with `<x>` and `[[param]]` with `<?*>` and so on...

1. `foo/bar<?*>`
2. `foo/<?*>bar`

...then figures out all the permutations of those routes:

3. `foo/bar`
4. `foo/bar<*>`
5. `foo/<*>bar`
6. `foo/bar`

Since 3 (derived from 1) and 6 (derived from 2) are identical, those routes are identified as being incompatible.

dummdidumm (approved):
Looks good!

The only thing I'm not sure about - but relaxing this is always possible later in a non-breaking way - is the very strict conflict check.

Imagine you have `a/[[b]]/c/[[d]]` and `a/c` as routes. Under this algorithm, this is a conflict, but the user could say "yes I know that that's a conflict in one of these cases, but I don't want to duplicate routes into `a/[b]/c/[d]`, `a/[b]/c/` and `a/c/[d]` (because `a/[b]/c/[[d]]` and `a/[[b]]/c/[d]` would also be a conflict in one case).

I don't know how a algorithm that relaxes this and at the same time sorts these things correctly without bringing the complexity of that other PR back would look like (probably something like "if this optional route only has one nn-conflicting permutation left it's an error), so I'm giving my approval regardless of this wrinkle.

Rich-Harris:
Ah, that's an interesting case. It seems like a fairly unlikely one, but if it _did_ arise then it would be possible to work around, or we could relax the conflict detection as you say. Aside from that edge case I think having the conflict detection is useful regardless (e.g. it found a conflicting route in the test app that wasn't doing any work)

PR #7309, opened 2022-10-19
Null check server_data

ivanvanderbyl:
This resolves an error we're seeing in production:

```
TypeError undefined is not an object (evaluating 'S.type')
```

This is caused by `server_data` being null.

### Tests

### Changesets


dummdidumm:
Could you add a test that fails without this check?

Rich-Harris:
If the test is tricky to figure out then a repro would also be helpful — thanks!

Rich-Harris:
Closing this, because whatever the bug is (please open an issue with a reproduction), this won't fix it

PR #7314, opened 2022-10-19
chore: improve error message for prefetching invalid url

dominikg:
Happened to come across `Uncaught (in promise) Error: Attempted to prefetch a URL that does not belong to this app ` in an app today and didn't immediately find the cause.

The url in the message can help finding where in the application it happened, also added some guidance and docs link for good measure.

In my case it was a link to a page that hadn't been created yet. If thats the only way to reach this error, maybe the url is enough to not confuse users into disabling prefetching too eagerly?

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -199,9 +199,7 @@ export function create_client({ target, base, trailing_slash }) {
 		const intent = get_navigation_intent(url, false);
 
 		if (!intent) {
-			throw new Error(
-				`Attempted to prefetch "${url}", a URL that does not belong to this app. To disable prefetching add data-sveltekit-prefetch="off" to a parent. See https://kit.svelte.dev/docs/link-options#data-sveltekit-prefetch for more information.`
-			);
+			throw new Error(`Attempted to prefetch a URL that does not belong to this app: ${url}`);
 		}
 
 		load_cache = { id: intent.id, promise: load_route(intent) };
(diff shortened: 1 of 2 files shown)


packages/kit/src/runtime/client/client.js
```diff
@@ -199,7 +199,9 @@ export function create_client({ target, base, trailing_slash }) {
 		const intent = get_navigation_intent(url, false);
 
 		if (!intent) {
-			throw new Error('Attempted to prefetch a URL that does not belong to this app');
+			throw new Error(
+				`Attempted to prefetch "${url}", a URL that does not belong to this app. To disable prefetching add data-sveltekit-prefetch="off" to a parent. See https://kit.svelte.dev/docs/link-options#data-sveltekit-prefetch for more information.`
```
Rich-Harris, line 203: IIUC the error shouldn't happen if prefetching happened as a result of the user stopping over a link, but only if the app calls `prefetch(bad_url)`. If that's not currently the case then I'd argue it's a bug, and we should fix that instead

dominikg: In my case it was a newly added `href="/some/new/route"` . The error happened during dev after the href was added but before +page.svelte was created. So bad url. Maybe skip and warn instead during dev and error only in build? But adding the url to the message seems helpful.

Rich-Harris: right, but what i'm saying is that `data-sveltekit-prefetch` shouldn't cause that error, it should be reserved for when you do `prefetch('/some/new/route')`

dummdidumm: So .. we have a bug hiding somewhere?

dominikg: Ahh. But adding the url value to the error message helps that case as well. The stacktrace isn't helping because it's internal to kit, so you have to start debugging to get a clue. 

I have removed the extra information to avoid confusion.

PR #7331, opened 2022-10-20
[docs] nav rework

dummdidumm:
Waiting on #7174 until we shuffle around all the `.md` files

- restructure nav on the left: top level sections, only show page titles
- add "on this page" nav on the right on big screens

### Tests

### Changesets


benmccann:
The "on this page" styling is a bit distracting with the grey and orange. Maybe we could make it more understated by making it look more like how Vite does it? (e.g. on https://vitejs.dev/config/)

I also think the gap between the content and sidebar is a bit strange. I think right now you kind of had to do that because the sidebar is grey, but if you just leave it with a white background then I think you could move it so that it's next to the content

I also wonder if we can't center the docs instead of having them left-aligned. Vite has a grey left nav, but have still centered the docs in a way that looks pretty decent (can look at the same https://vitejs.dev/config/ page for an example)

@vedam may have some design ideas as well

dummdidumm:
I switched the styling to something less intrusive

vedam:
Hey @benmccann @dummdidumm,
I also tweaked a bit locally.
And I also tried it with color-highlighted .active state.  ;-)
In addition, I made font-weight to correspond with the left side-bar.
proposal
If that's too distracting we should choose a darker gray or black for the .active state. 

In a first draft for the original base-styles [in site-kit](https://github.com/sveltejs/site-kit/blob/master/base.css) very long ago, i had intended for links `<a...` that the underlines are animated, growing in from left to right which I was able to achieve with the border-bottom style. Since we don't use that anyway, we might as well basically revert to `text-decoration: underline`.
Only the distance of the underline I would like to set off a bit from the text with `text-underline-offset: 3px`
```css
/*
	https://github.com/sveltejs/site-kit/blob/master/base.css
	LINE 295
*/
/*  links ------------------------------------- */
a {
	color: var(--prime);
}

a:hover  {
	text-decoration: underline;
	text-decoration-thickness: 2px;
	text-decoration-color: currentColor;
	text-underline-offset: 3px;
	transition: color .2s, text-decoration .2s;
}

a.no-underline {
	text-decoration: none;
}

```
@benmccann I can make a PR, but I'm afraid you'll get my whole mess like last time.
Only 2 files changed. (3 with the link change)

Rich-Harris:
Updating the file structure precipitated a few more changes to how the data is loaded. I still think there's some room for improvement for our styles, but I started work on design tweaks and it became a bit of a rabbit hole, so I think we should focus on just the contents/on-this-page stuff for now

Rich-Harris (approved):

PR #7338, opened 2022-10-21
prefix all route IDs with `/`

### LINKED ISSUE #7332: Prefix all route IDs with `/`
### Describe the problem

I don't recall if there was a reason we decided that the `routeId` for `src/routes/foo/+page.svelte` should be `foo` instead of `/foo`, but I think it was a mistake. On a number of occasions I've found myself doing things like `if (routeId)` and getting caught out because it can sometimes be the empty string.

Since we expose `routeId` via `event`, app developers will inevitably run into the same issue.

It would be safer, and arguably more intuitive, if route IDs were always prefixed with `/`. 

### Alternatives considered

The problem of course is that this would be a breaking change, and we can't detect people doing `routeId === 'foo'`. For the people currently using `routeId` this would be annoying.

The question is whether causing breaking change for that population is more costly than reaching 1.0 with a known design flaw.

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
closes #7332

### Tests

### Changesets

### DIFF
--- a/.changeset/strange-crabs-shop.md
+++ b/.changeset/strange-crabs-shop.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] Prefix all route IDs with /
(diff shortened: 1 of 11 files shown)


dummdidumm (approved):
Looks good. Any thoughts on the warning I proposed in #7332?

Rich-Harris:
I think @Conduitry is right, it will be annoying to have that warning appear once everyone has already made the change

Issue #2958, opened 2021-12-01
Dev server does not apply base path correctly


andycarlberg:
### Describe the bug

I'm using a [code-server](https://github.com/cdr/code-server) cloud environment to develop a SvelteKit project. In order to access the dev server, a proxy is provided that uses a subpath to proxy the dev server port (see [docs](https://coder.com/docs/code-server/latest/guide#using-a-subpath). This leads me to need a base path to be set for the dev server. 

If I set `kit.paths.base` as stated in the documentation, the first page load works correctly but the `import` statements are relative to the root which returns a 404, correctly, because those files are not available without the proxy. If I run `build` and `preview` everything works as expected since the base path is built into the compiled project.

**Expected behavior**
All paths in the dev server reference the base path instead of being relative to the root domain.

### Reproduction

1. Create a new SvelteKit project
2. Set `kit.paths.base` to a non-root path
3. Run the dev server

**Expected result**
All `import` statements look for paths with the base path or relative to the base path

**Actual result**
Module `import` statements are relative to the root

**Example**
![image](https://user-images.githubusercontent.com/7405933/144281269-72b6ae18-a7ca-4e51-a684-f11c187fadc4.png)
![image](https://user-images.githubusercontent.com/7405933/144281372-087c2a6e-453a-4a5f-8ff7-6ac9c43c477c.png)

### Logs

```shell
GEThttp://domain.tld/.svelte-kit/dev/runtime/internal/start.js
[HTTP/1.1 404 Not Found 8ms]

Loading module from “http://domain.tld/.svelte-kit/dev/runtime/internal/start.js” was blocked because of a disallowed MIME type (“application/json”). 3000
Loading failed for the module with source “http://domain.tld/.svelte-kit/dev/runtime/internal/start.js”. 3000:13:1
GEThttp://domain.tld/favicon.png
[HTTP/1.1 404 Not Found 6ms]
```

### System Info

```shell
System:
    OS: Linux 5.10 Ubuntu 20.04.3 LTS (Focal Fossa)
    CPU: (12) x64 Intel(R) Core(TM) i5-10400 CPU @ 2.90GHz
    Memory: 283.55 MB / 15.32 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 14.18.2 - /usr/bin/node
    Yarn: 1.22.15 - /usr/bin/yarn
    npm: 6.14.15 - /usr/bin/npm
  npmPackages:
[...]

PR #7343, opened 2022-10-21
Use `req.originalUrl` for better Vite compatibility

benmccann:
This gets us pretty darn close to being able to set the `base` option in Vite correctly. Right now we hardcode it to `./`, which isn't correct and I suspect is one of the reasons the highly upvoted https://github.com/sveltejs/kit/issues/2958 doesn't work. There's one more thing we'll need to fix it which is https://github.com/vitejs/vite/issues/9236 and I'm working on getting that in as part of Vite 4

The reason we need this is that Vite rewrites the `url` to strip out the base path, which makes the `url` not what we might expect. I looked into changing that in Vite and discussed with the Vite team. While there's the possibility we might be able to do it, it's not easy and would affect the entire Vite ecosystem, so we'd need to get buy in from all the other frameworks. And actually, what Vite is doing is how the Express ecosystem works and is explicitly what `originalUrl` was provided for, so there's a good argument they shouldn't change anything on their side

### DIFF
--- a/packages/kit/src/exports/node/index.js
+++ b/packages/kit/src/exports/node/index.js
@@ -94,8 +94,7 @@ function get_raw_body(req, body_size_limit) {
 
 /** @type {import('@sveltejs/kit/node').getRequest} */
 export async function getRequest({ request, base, bodySizeLimit }) {
-	// @ts-expect-error - the request may be a connect request
-	return new Request(base + (request.originalUrl || request.url), {
+	return new Request(base + request.url, {
 		method: request.method,
 		headers: /** @type {Record<string, string>} */ (request.headers),
 		body: get_raw_body(request, bodySizeLimit)
--- a/packages/kit/src/exports/node/index.js
+++ b/packages/kit/src/exports/node/index.js
@@ -94,7 +94,8 @@ function get_raw_body(req, body_size_limit) {
 
 /** @type {import('@sveltejs/kit/node').getRequest} */
 export async function getRequest({ request, base, bodySizeLimit }) {
-	return new Request(base + request.url, {
+	// @ts-expect-error - the request may be a connect request
+	return new Request(base + (request.originalUrl || request.url), {
 		method: request.method,
 		headers: /** @type {Record<string, string>} */ (request.headers),
 		body: get_raw_body(request, bodySizeLimit)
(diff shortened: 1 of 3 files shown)


packages/kit/src/exports/node/index.js
```diff
@@ -94,7 +94,8 @@ function get_raw_body(req, body_size_limit) {
 
 /** @type {import('@sveltejs/kit/node').getRequest} */
 export async function getRequest({ request, base, bodySizeLimit }) {
-	return new Request(base + request.url, {
+	// @ts-expect-error - the request may be a connect request
+	return new Request(base + (request.originalUrl || request.url), {
```
Conduitry, line 98: Does `request.originalUrl || request.url` serve any purpose other than for Vite's dev mode? I'd prefer not to have any dev-specific code end up in production code built with (say) the Node adapter, which I think this would.

If, on the other hand, this is also helpful for supporting Node adapter output in custom server mode using Connect, that should be made more explicit somewhere. Does it belong in the Node adapter's readme maybe?

benmccann: It helps both dev and preview.

It seems pretty unlikely to me that the values of `url` and `originalUrl` would differ in `adapter-node`. They could if the user mounted the app under a base path themselves in a custom server implementation, but I don't think I'd expect that to work regardless of which value were used here. We could throw an exception if `originalUrl` is defined and does not equal `url` if we want to enforce that they're the same. That'd then give us the ability to explicitly support them being different in the future without it being a breaking change as I really don't know what that'd mean at this point.

Rich-Harris:
Could we do something like `req.url = req.originalUrl` in the middleware? This change makes me nervous, it feels like a workaround rather than a fix

benmccann:
Yeah, that works too and is even easier. Done.

Conduitry:
Was there something else with `vite preview` that you wanted to address that isn't covered by this?

benmccann:
I don't think we'd have to do it in `preview` because I don't think Vite's `base` middleware will be invoked there. E.g. we're responsible for serving static assets in `preview` and that will remain the case even with Vite 4: https://github.com/sveltejs/kit/blob/be01da3403d571d2451ccb5349f6b9ce1e9c3fd7/packages/kit/src/exports/vite/preview/index.js#L57

Rich-Harris (approved):

PR #7350, opened 2022-10-22
Always avoid caching form submission on enhanced forms

### LINKED ISSUE #6357: Invalidatation should always query the invalid server nodes with `fetch({ cache: 'no-cache' })`

Tal500:
Due to the return of regular JSON requests of #7177, issue #6357 is relevant back again. Although now I couldn't reproduce the behavior, it stills problematic and needed to be fixed.

Clearly, when you simulate regular form submission, you should check that the browser never used a cached result of previous POST request.

It is unclear how browsers choose to handle this situation, see [this stackoverflow question](https://webmasters.stackexchange.com/questions/111298/what-happens-if-you-dont-set-cache-control-header).
Since browsers seems to be more careful on programically `fetch` POST requests, I couldn't catch a browser really use caching for this "no cache-control header" responses.

Currently on form submission from enhanced forms, there is no cache control header from the server, no cache guideline on the `fetch` function, so I think that the best solution is to guide the browser in the `fetch` method of the submission, to not use previously cached results.

fixes #6357 (in some sense)

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/app/forms.js
+++ b/packages/kit/src/runtime/app/forms.js
@@ -86,6 +86,7 @@ export function enhance(form, submit = () => {}) {
 					accept: 'application/json',
 					'x-sveltekit-action': 'true'
 				},
+				cache: 'no-cache',
 				body: data,
 				signal: controller.signal
 			});
--- a/packages/kit/src/runtime/app/forms.js
+++ b/packages/kit/src/runtime/app/forms.js
@@ -86,7 +86,7 @@ export function enhance(form, submit = () => {}) {
 					accept: 'application/json',
 					'x-sveltekit-action': 'true'
 				},
-				cache: 'no-cache',
+				cache: 'no-store',
 				body: data,
 				signal: controller.signal
 			});
--- a/.changeset/smart-dolphins-mix.md
+++ b/.changeset/smart-dolphins-mix.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Always avoid caching form submission on enhanced forms


packages/kit/src/runtime/app/forms.js
```diff
@@ -86,6 +86,7 @@ export function enhance(form, submit = () => {}) {
 					accept: 'application/json',
 					'x-sveltekit-action': 'true'
 				},
+				cache: 'no-cache',
```
Rich-Harris, line 89: Is there a reason to use `no-cache` rather than `no-store`? My understanding is that `no-cache` (confusingly) means 'cache but don't use the cached result', while `no-store` means 'don't cache', which seems more... thorough. But if `no-cache` is preferred I will confess ignorance

```suggestion
				cache: 'no-cache',
```

Tal500: I'm also not sure what is the best practice, I also know only the shallow definitions.

One theoretical advantage of `no-cache` comparing to `no-store`, is that if we have the following story:
1. The user post a guess for sverdle, and it's not a valid word, so the server returns "bad guess"
2. The user post another guess for sverdle, and it's also not a valid word, so the server returns "bad guess" back again.

If the cache is stored, then the browser can use the `ETag` HTTP header to check if the server response will be different, and then avoid full download.

Clearly it makes no sense in this small response, but in other cases (like server data processing) it might be useful.
On the other side, in small responses, the cache will be so small, so "who cares".

What do you think? Should it be `no-store` by default, but let the user override the cache parameter if he wishes?

Rich-Harris: Ah, interesting — didn't consider the ETag case. Since the action would have to generate those ETags manually (it doesn't happen automatically, unlike with pages, though maybe we could change that? Unsure if worth it), I suspect this is more of a theoretical benefit than anything, whereas `no-store` feels a bit more belt-and-braces. I see you changed it to `no-store` in the meantime — let's roll with that for now. Thanks!

Rich-Harris (approved):

PR #7355, opened 2022-10-22
chore: update dependencies

dominikg:
It looks like renovate has not sent PRs for minor updates in a while, so this bumps them accross the board.

There is a test fail that seems to happen due to playwright waitForNavigation behaving differently than before, upon updating from 1.25.0 to 1.27 in vite-plugin-svelte I saw similar problems and worked around them by waiting for more specific events like the vite devserver connection.

Will try to investigate that tomorrow but wanted to see CI results.

### Tests

### Changesets


turbo.json
```diff
@@ -1,75 +1,131 @@
...
-			"outputs": [],
-			"outputMode": "new-only"
-		},
-		"dev": {
-			"cache": false,
-			"outputMode": "new-only"
-		}
-	},
-	"globalDependencies": [".eslintrc.json", ".prettierrc", "tsconfig.json"]
+  "$schema": "https://turborepo.org/schema.json",
```
dummdidumm, line 2: accidental formatting (now has spaces instead of tabs)? Anything that actually changed in here?

dominikg: this was done by turbos migration tool and format didn't bother so I kept it that way. The code change is the way env dependencies are declared

dominikg:
no clue why it fails on mac only (4 times so it doesn't seem to be a coincidence). If anyone with a mac is able to reproduce that fail it would help.

Rich-Harris:
I can reproduce the failure locally. The test is expecting a console warning to appear, and it never does, but only when `KIT_E2E_BROWSER=webkit`.

The thing is the warning _is_ visible in Safari devtools:

image

It looks like a new Playwright bug. Downgrading to 1.25 fixes it, at least locally — 🤞 for a full bank of green lights

benmccann:
@dominikg it looks like this PR needs to be rebased

benmccann:
It looks like Rich merged in my change to upgrade to Vite 3.2, but I'm not sure that we should since the Vite 3.2 types are borked so now `lint` won't pass on this PR

Rich-Harris:
I just merged master to resolve the conflicts. Are you saying we need to downgrade to Vite 3.1.something? Why are the types borked?

benmccann:
`master` isn't on Vite 3.2. It's on Vite 3.1. I sent https://github.com/sveltejs/kit/pull/7400 to upgrade to 3.2.

3.2.1 fixed the types. You can either merge https://github.com/sveltejs/kit/pull/7400 or upgrade this PR to 3.2.1

turbo.json
dominikg: i just checked what happens and it looks like prettier is never run from the root of the monorepo, https://github.com/sveltejs/kit/blob/edd815ed7bba7ab88461f9291f924c05a01d1328/package.json#L11  this recursively calls format scripts in packages, but there is no separate format script in root so prettier is never called on root files.

i have formatted it manually with `pnpm prettier --write turbo.json` to keep the diff smaller here, but we should look into fixing this in a separate PR

benmccann (approved):

PR #7370, opened 2022-10-24
[fix] get type gen working again

dummdidumm:
since the routeId has changed, the matching logic needs an update

### Tests

### Changesets


dominikg:
Is this new routeId leading slash a breaking change? Or to rephrase: do we consider the manifest format public api or is it internal?

dummdidumm:
that change was marked as breaking, since you also can access it at runtime through `event.routeId` in `load`

dominikg (approved):

PR #7375, opened 2022-10-24
detect missing ids when prerendering

### LINKED ISSUE #3955: Validate hash links when prerendering
### Describe the problem

This is a relatively small thing, but it would be nice if the crawler could warn you if you prerender a page with a link like this...

```html
<a href="/docs/routing#advanced">advanced routing</a>
```

...and the `/docs/routing` page was updated:

```html
<h2 id="advanced-routing">Advanced routing</h2>
```

When prerendering, keep track of which hash links have been requested, and whether they match `id` attributes on the destination page.

### Importance

nice to have

Rich-Harris:
Closes #3955. This throws an error in the case where a page links to `/a#b` and `/a` doesn't include an element with `id="b"`.

We probably want to make that behaviour configurable somehow, since some people might want a warning rather than an error, or even custom error handling. For normal prerendering errors this is enabled via `config.kit.prerender.onError`, but the API can't accommodate these errors. I'm not really sure what to do about that, so I'm just going to leave this in draft for now.

Also not pictured: tests of funky IDs that involve URL or HTML encoding. No idea if that works in this branch currently but either way it needs tests.

### Tests

### Changesets


Rich-Harris:
The deploy failure is because this is working — it's identified an incorrect link

Rich-Harris:
This replaces `config.kit.prerender.onError` with `config.kit.prerender.handleHttpError`, to distinguish it from the new `config.kit.prerender.handleMissingId`.

Remaining question is whether the default for `config.kit.prerender.handleMissingId` should be `fail` or `warn`. Thoughts?

packages/kit/src/core/prerender/crawl.js
```diff
@@ -125,7 +128,9 @@ export function crawl(html) {
 								let escaped = false;
 
 								while (i < html.length) {
-									if (!escaped) {
+									if (escaped) {
+										escaped = false;
```
dominikg, line 132: this looks like a bugfix for double escaping. does that need a separate changeset?

packages/kit/src/core/prerender/prerender.js
```diff
@@ -12,45 +12,31 @@ import { load_config } from '../config/index.js';
...
+		case 'fail':
 			return (details) => {
-				log.error(format_error(details, config));
+				throw new Error(format(details));
 			};
-		case 'fail':
+		case 'warn':
 			return (details) => {
-				throw new Error(format_error(details, config));
+				log.error(format(details));
```
dominikg, line 34: is there a log.warn or should there be one?

packages/kit/src/core/prerender/prerender.js
```diff
@@ -133,7 +119,29 @@ export async function prerender() {
...
 	await server.init({ env: JSON.parse(env) });
 
-	const error = normalise_error_handler(log, config);
+	const handle_http_error = normalise_error_handler(
+		log,
+		config.prerender.handleHttpError,
+		({ status, path, referrer, referenceType }) => {
+			const message =
+				status === 404 && !path.startsWith(config.paths.base)
+					? `${path} does not begin with \`base\`, which is configured in \`paths.base\` and can be imported from \`$app/paths\` - see https://kit.svelte.dev/docs/configuration#paths for more info`
```
dominikg, line 128:
```suggestion
					? `${path} does not begin with \`${config.paths.base}\`, which is configured in \`paths.base\` and can be imported from \`$app/paths\` - see https://kit.svelte.dev/docs/configuration#paths for more info`
```

packages/kit/src/core/prerender/prerender.js
```diff
@@ -381,6 +407,17 @@ export async function prerender() {
 
 	await q.done();
 
+	// handle invalid fragment links
+	for (const [key, referrers] of expected_deeplinks) {
+		const [path, id] = key.split('#');
```
dominikg, line 412: this could break if there is another hash like `/foo#bar#baz`

documentation/docs/50-api-reference/10-configuration.md
```diff
@@ -253,32 +254,19 @@ See [Prerendering](/docs/page-options#prerender). An object containing zero or m
...
-    			}
-    		}
-    	}
-    };
-
-    export default config;
-    ```
+  - `'ignore'` - silently ignore the failure and continue
+  - `'warn'` — continue, but print a warning
+  - `(details) => void` — a custom error handler that takes a `details` object with `status`, `path`, `referrer` and `referenceType` properties
```
dominikg, line 262: Should we mention that you have to throw an error from custom handler to fail the prerender process/ build?

packages/kit/src/core/prerender/prerender.js
```diff
@@ -231,18 +245,30 @@ export async function prerender() {
 		const headers = Object.fromEntries(response.headers);
 
 		if (config.prerender.crawl && headers['content-type'] === 'text/html') {
-			for (const href of crawl(body.toString())) {
+			const { ids, hrefs } = crawl(body.toString());
+
+			actual_deeplinks.set(decoded, ids);
+
+			for (const href of hrefs) {
 				if (href.startsWith('data:') || href.startsWith('#')) continue;
```
dummdidumm, line 253: Does this mean (edge case) that if you have hash links to the same page that are wrong that they wouldn't get caught by the missing id logic?

packages/kit/src/core/prerender/prerender.js
```diff
@@ -381,6 +407,17 @@ export async function prerender() {
...
 	await q.done();
 
+	// handle invalid fragment links
+	for (const [key, referrers] of expected_deeplinks) {
+		const [path, id] = key.split('#');
+
+		const actual_ids = actual_deeplinks.get(path);
+
+		if (actual_ids?.includes(id)) {
+			handle_missing_id({ id, path, referrers: Array.from(referrers) });
```
dummdidumm, line 417: printing those missing ids in a loop like this means that it throws after the first error by default, and you don't see the other wrong links. I think it would be better to print all out in one go.
Aside: It would be good to also add something like "you can adjust the handling of missing ids through -config option-" so people have a clearer way out in case they don't want to change the hash.

dummdidumm (changes requested):
Some small suggestions, but mostly looks good

packages/kit/src/core/prerender/crawl.js
Rich-Harris: done

packages/kit/src/core/prerender/prerender.js
Rich-Harris: there is; the only difference is it prints yellow rather than red. we're already using it for informational purposes when prerendering (to indicate redirects), so i think it's better to use `log.error` here to differentiate

packages/kit/src/core/prerender/prerender.js
Rich-Harris: fixed

packages/kit/src/core/prerender/prerender.js
dummdidumm: "foo does not begin with `bar`, which is configured in `paths.base` and ..." sounds better to me than "foo does not begin with `base`, which is configured in `paths.base` and ..."

Rich-Harris: The trick is to steer people away from seeing `"/bar/"` in the message and just hardcoding that prefix

documentation/docs/50-api-reference/10-configuration.md
```diff
@@ -253,32 +254,19 @@ See [Prerendering](/docs/page-options#prerender). An object containing zero or m
...
+
+- `handleMissingId`
+
+  - `'fail'` — (default) fails the build when a prerendered page links to another prerendered page with a `#` fragment that doesn't correspond to an `id`
+  - `'ignore'` - silently ignore the failure and continue
+  - `'warn'` — continue, but print a warning
+  - `(details) => void` — a custom error handler that takes a `details` object with `path`, `id` and `referrers` properties
```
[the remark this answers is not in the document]
Rich-Harris:
```suggestion
  - `(details) => void` — a custom error handler that takes a `details` object with `path`, `id`, `referrers` and `message` properties. If you `throw` from this function, the build will fail
```

packages/kit/src/core/prerender/crawl.js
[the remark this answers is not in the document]
Rich-Harris: line 141

packages/kit/src/core/prerender/prerender.js
Rich-Harris: It does, yeah. Will fix

packages/kit/src/core/prerender/prerender.js
Rich-Harris: Had an internal debate with myself on this — initially I could have gone either way, but I think I lean towards keeping it as-is:

* we fail eagerly for HTTP errors so it might be odd not to do the same thing here
* the custom function signature gets a bit trickier, and we'd need to call it `handleMissingIds`. it's a bit unusual for an error handler to handle multiple errors in one go
* given that the default behaviour is to fail, it would be unlikely for multiple missing IDs to be introduced in one go (the exception being the day this feature ships and previously hidden bad links surface en masse)
* if someone does need to see all the missing IDs together, they could do `handleMissingId: warn` and run a build

packages/kit/src/core/prerender/prerender.js
dominikg: If you don't want to include the actual value, maybe completely rephrasing it can help transport that and steer users in the right direction? I got confused by the static `base` that doesn't make much sense when reading the error message in isolation.

packages/kit/src/core/prerender/crawl.js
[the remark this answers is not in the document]
dummdidumm: wtf I could've sworn I did a text search for occurences - on GitHub, maybe that's the problem 😄

packages/kit/src/core/prerender/prerender.js
```diff
@@ -155,6 +163,12 @@ export async function prerender() {
 	const seen = new Set();
 	const written = new Set();
 
+	/** @type {Map<string, Set<string>>} */
+	const expected_deeplinks = new Map();
+
+	/** @type {Map<string, string[]>} */
+	const actual_deeplinks = new Map();
```
dummdidumm, line 170: I was confused for a moment because I understand a regular link to not just the home page by "deeplink", I think something like "hashlink" would be better.

dummdidumm (approved):
Apart from a minor naming (which I'm ok with if you turn it down) thing this looks good!

packages/kit/src/core/prerender/prerender.js
Rich-Harris: agree — fixed

PR #7377, opened 2022-10-24
fix search

Rich-Harris:
i broke it in #7331


Rich-Harris:
going to self-merge as this is too important to stay this broken

PR #7386, opened 2022-10-25
[fix] show more descriptive error if data returned from `load` is a non-POJO

### LINKED ISSUE #7262: More informative error if someone returns a `Response` from `load`
### Describe the problem

Per https://github.com/sveltejs/kit/discussions/7259 — if someone does this...

```js
import { json } from '@sveltejs/kit';

export function load() {
  return json({ answer: 42 });
}
```

...they will get a cryptic error:

> DevalueError: Cannot stringify arbitrary non-POJOs

It would be easy to catch this case and provide a more specific error:

> `load` must return a plain object rather than a `Response`

For bonus points, it should identify the `load` function in question.

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

errmayank:
Closes #7262

- Tested it out with return `json`

Screenshot 2022-10-25 at 17 44 12
Screenshot 2022-10-25 at 17 42 52

- with `new Error`

Screenshot 2022-10-25 at 17 43 58
Screenshot 2022-10-25 at 17 43 15

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -182,6 +182,19 @@ export async function render_response({
 				`Data returned from \`load\` while rendering ${event.routeId} is not serializable: ${error.message} (data.${match[2]})`
 			);
 		}
+
+		const nonPojoError = /pojo/i.exec(error.message);
+
+		if (nonPojoError) {
+			const constructorName = branch[0]?.server_data?.data?.constructor?.name;
+
+			throw new Error(
+				`Data returned from \`load\` must be a plain object${
+					constructorName ? ` rather than ${constructorName} constructor` : ''
+				}`
+			);
+		}
+
 		throw error;
 	}
 
--- a/packages/kit/src/runtime/server/page/render.js
+++ b/packages/kit/src/runtime/server/page/render.js
@@ -186,11 +186,12 @@ export async function render_response({
 		const nonPojoError = /pojo/i.exec(error.message);
 
 		if (nonPojoError) {
-			const constructorName = branch[0]?.server_data?.data?.constructor?.name;
+			const constructorName = branch.find(({ server_data }) => server_data?.data?.constructor?.name)
+				?.server_data?.data?.constructor?.name;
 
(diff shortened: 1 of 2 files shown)


packages/kit/src/runtime/server/page/render.js
```diff
@@ -182,6 +182,19 @@ export async function render_response({
 				`Data returned from \`load\` while rendering ${event.routeId} is not serializable: ${error.message} (data.${match[2]})`
 			);
 		}
+
+		const nonPojoError = /pojo/i.exec(error.message);
+
+		if (nonPojoError) {
+			const constructorName = branch[0]?.server_data?.data?.constructor?.name;
```
dummdidumm, line 189: This could be from server_data deeper in the load tree, so I think it should be 
```suggestion
			const constructorName = branch.find(({ server_data }) => server_data?.data?.constructor?.name)?.server_data.data.constructor.name;
```

packages/kit/src/runtime/server/page/render.js
```diff
@@ -182,6 +182,19 @@ export async function render_response({
...
+
+			throw new Error(
+				`Data returned from \`load\` must be a plain object${
+					constructorName ? ` rather than ${constructorName} constructor` : ''
+				}`
```
dummdidumm, line 194:
```suggestion
				`Data returned from \`load\` (while rendering ${event.routeId}) must be a plain object${
					constructorName ? ` rather than an instance of ${constructorName}` : ''
				}`
```

dummdidumm (commented):
Thanks, looks mostly good, added some small comments

errmayank:
Thanks @dummdidumm, i've updated it.

errmayank:
Thanks @dummdidumm, updated.

dummdidumm:
Could you add a changeset by running `npx changeset` in the root and following the prompts (patch release)? Then we're good to go 👍

errmayank:
Added changeset

dummdidumm (approved):
Thank you!

Issue #2612, opened 2021-10-15
Prebundling of libraries with Svelte components


benmccann:
### Describe the bug

See https://github.com/illright/attractions/issues/311#issuecomment-944701061 for the original report

It was loading all of `svelte-feather-icons` when you use just a few icons. Vite is supposed to do dependency pre-bundling to avoid this: https://vitejs.dev/guide/dep-pre-bundling.html#the-why. However, Vite's pre-bundling uses esbuild, so we disabled it on Svelte components because it doesn't know what to do with `.svelte` files.

[Evan has proposed changing this](https://github.com/vitejs/vite/discussions/4921). I'm finally a convert to the idea.

### Reproduction

Try to use `svelte-feather-icons`

### Logs

```shell
https://github.com/illright/attractions/ was making 485 requests for each page in dev mode taking 10s to load
```

### System Info

```shell
Vite 2.6.7
```

### Severity

serious, but I can work around it

### Additional Information

You can work around it by doing deep imports

antony:
This was the same in Sapper. Reloads were taking 60 seconds. I made a reduced version of the library including only the icons we required, and reloads started taking 20 seconds. Game changer.

I suppose what this means is that, the library isn't properly tree-shakeable.

dominikg:
Until vite can pre-bundle svelte libraries it's best to use deep imports to avoid paying the tax of loading all exported svelte components. 

Icon libraries are usually the worst in this context, luckily https://github.com/antfu/unplugin-icons does not suffer from this.

benmccann:
Wow, that's a great solution! Yeah, I wonder if we could tweak that library and use it generally. I've asked the author in https://github.com/carbon-design-system/carbon-preprocess-svelte/issues/20

bluwy:
I think we discussed something similar too about transforming the entrypoint with only the named imports, so it works more generically. The preprocessor technique is interesting though, I suppose it can be worked around by checking the `exports` field and naively work with that.

Anywho, I've been working today on actually prebundling Svelte libraries. It seems to be working on [this branch](https://github.com/sveltejs/vite-plugin-svelte/tree/prebundle-svelte-libraries). Wish I thought about it sooner :smile:  The caveats now are that:

1. Prebundling would take a while (6s for `carbon-components-svelte`), but that's negligible in the long run.
[...]

PR #7388, opened 2022-10-25
[chore] upgrade to `vite-plugin-svelte` 1.1.0 and enable `prebundleSvelteLibraries`

benmccann:
We'll turn `prebundleSvelteLibraries` on by default in `vite-plugin-svelte` 2.0 when Vite 4 comes out. For now we turn it on here. The feedback thus far has been good

Ref https://github.com/sveltejs/kit/issues/2612

### DIFF
--- a/.changeset/itchy-ants-explode.md
+++ b/.changeset/itchy-ants-explode.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[chore] upgrade to vite-plugin-svelte 1.1.0 and enable prebundleSvelteLibraries
--- a/packages/kit/package.json
+++ b/packages/kit/package.json
@@ -10,7 +10,7 @@
 	"homepage": "https://kit.svelte.dev",
 	"type": "module",
 	"dependencies": {
-		"@sveltejs/vite-plugin-svelte": "^1.0.5",
+		"@sveltejs/vite-plugin-svelte": "^1.1.0",
 		"@types/cookie": "^0.5.1",
 		"cookie": "^0.5.0",
 		"devalue": "^4.0.0",
--- a/packages/kit/src/exports/vite/index.js
+++ b/packages/kit/src/exports/vite/index.js
@@ -61,7 +61,7 @@ const enforced_config = {
 
 /** @return {import('vite').Plugin[]} */
 export function sveltekit() {
-	return [...svelte(), kit()];
+	return [...svelte({ prebundleSvelteLibraries: true }), kit()];
 }
 
 /**
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -266,7 +266,7 @@ importers:
   packages/kit:
     specifiers:
       '@playwright/test': ^1.25.0
-      '@sveltejs/vite-plugin-svelte': ^1.0.5
+      '@sveltejs/vite-plugin-svelte': ^1.1.0
       '@types/connect': ^3.4.35
       '@types/cookie': ^0.5.1
       '@types/marked': ^4.0.7
@@ -292,7 +292,7 @@ importers:
       uvu: ^0.5.3
       vite: ^3.1.1
     dependencies:
-      '@sveltejs/vite-plugin-svelte': 1.0.5_svelte@3.48.0+vite@3.1.1
+      '@sveltejs/vite-plugin-svelte': 1.1.0_svelte@3.48.0+vite@3.1.1
       '@types/cookie': 0.5.1
       cookie: 0.5.0
       devalue: 4.0.0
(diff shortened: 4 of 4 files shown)


Rich-Harris (approved):

schmidtk:
Enabling `prebundleSvelteLibraries` is causing the below error building my application with vite. I flipped it to false to confirm that specifically caused the error, but I'm unsure how to resolve it. Are there vite/svelte/etc config changes necessary to accommodate this change? Thanks for your help, happy to provide more information if needed.

```
vite v3.1.4 building SSR bundle for development...
✘ [ERROR] No loader is configured for ".svelte" files: ../../node_modules/@sveltejs/kit/src/runtime/components/error.svelte

[vite-plugin-svelte-kit] Build failed with 1 error:
error: No loader is configured for ".svelte" files: ../../node_modules/@sveltejs/kit/src/runtime/components/error.svelte
error during build:
Error: Build failed with 1 error:
error: No loader is configured for ".svelte" files: ../../node_modules/@sveltejs/kit/src/runtime/components/error.svelte
    at failureErrorWithLog (../../node_modules/esbuild/lib/main.js:1566:15)
    at ../../node_modules/esbuild/lib/main.js:1024:28
    at runOnEndCallbacks (../../node_modules/esbuild/lib/main.js:1438:61)
    at buildResponseToResult (../../node_modules/esbuild/lib/main.js:1022:7)
    at ../../node_modules/esbuild/lib/main.js:1134:14
    at responseCallbacks.<computed> (../../node_modules/esbuild/lib/main.js:671:9)
    at handleIncomingPacket (../../node_modules/esbuild/lib/main.js:726:9)
    at Socket.readFromStdout (../../node_modules/esbuild/lib/main.js:647:7)
    at Socket.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:315:12)
```

benmccann:
@schmidtk can you file an issue with a way to reproduce this? I haven't see this error before and am unsure what to advise without some way to be able to reproduce and debug it

schmidtk:
@benmccann No problem, I was hoping it would ring a bell. 😄 I'll see if I can reproduce with a simple example, because I'm currently working in a reasonably large monorepo.

schmidtk:
@benmccann Ok narrowed it down further. The error is a combination of enabling `legacy.buildSsrCjsExternalHeuristics` in our vite config, and not having an error route in the app. When that flag is enabled and Sveltekit uses its default error page, we see the error I posted above.

I'll check with my team on why we're using that flag and see if we can stop doing that, but adding our own `+error.svelte` will prevent the build error as a workaround. FYI in case `buildSsrCjsExternalHeuristics` and `prebundleSvelteLibraries` should be working together.

jacob-8:
Enabling `prebundleSvelteLibraries` causes an error in my application which imports a Svelte component from a dependency that tries to use `import { PUBLIC_FIREBASE_CONFIG } from '$env/static/public';`. This worked fine in `1.0.0-next.525` but not in `1.0.0-next.526`. I'll try to add a minimal repro later.

bluwy:
@jacob-8 One workaround we discussed is to set `$env/static/public` to `optimizeDeps.exclude`. This could be the default in the future, but on the other hand Svelte packages should try to avoid SvelteKit features unless necessary. In this case, the package could have a function that receives a parameter for `PUBLIC_FIREBASE_CONFIG`

jacob-8:
> @jacob-8 One workaround we discussed is to set `$env/static/public` to `optimizeDeps.exclude`. 

Perfect! I'll try that.

> on the other hand Svelte packages should try to avoid SvelteKit features unless necessary.

Yeah, I used to pass in a config to the library and would like to return to doing this but with all the changes affecting initialization of a library on both client and server side it was just to painful to make one single solution work across all my projects without awaiting parent in each load. I'm waiting until the dust settles a bit more.

0gust1:
In our app we use [byteMD](https://github.com/bytedance/bytemd), and it stopped working with 526, but **only in dev mode** (`npm run preview` is fine). 

The error message in console is like `SyntaxError: The requested module '/node_modules/word-count/index.js?v=4f0918c6' does not provide an export named 'default`. 

As this message hints to vite dev mode module loading, and this PR was linked from 526 changelog, I post here. 
Will try some circumvention, but it's pretty annoying for us.

0gust1:
I hadn't much luck fiddling with config files, so I setup a blank sveltekit project to reproduce: 
TLDR; I reproduce the problem (reminder: dev mode only) with:
- sveltekit >v225
- `export const ssr = false;` in master (topmost) +layout.server.ts (our app is a SPA)

If I downgrade SK version to v225 or lower, it works again.

0gust1:
@benmccann reproductible repository here: https://github.com/0gust1/sveltekit-bug-repro-526
Should I open an issue ?

PR #7395, opened 2022-10-26
use locators over direct selectors

ignatiusmb:
Using https://playwright.dev/docs/locators which are stricter can help us prevent flakiness, they also break loudly and gives a much nicer error message, it would also complain when there's multiple elements that match the same selector

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/utils.js
+++ b/packages/kit/test/utils.js
@@ -56,10 +56,11 @@ export const test = base.extend({
 		 * @param {{ timeout: number }} options
 		 */
 		async function clicknav(selector, options) {
+			const element = page.locator(selector);
 			if (javaScriptEnabled) {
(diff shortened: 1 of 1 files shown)


dummdidumm:
The strictness is nice, but I'm not sure how it would help with flakyness, could you elaborate on that?

ignatiusmb:
I couldn't seem to find where I read it first, here's a supporting [article](https://cathalmacdonnacha.com/playwright-e2e-testing-tips-and-best-practices#heading-2-use-locators-over-selectors) that I found through a quick search for now.

The docs also mention locators as the recommended/alternative way to reduce flakiness (https://playwright.dev/docs/api/class-page#page-eval-on-selector).

Issue #1042, opened 2021-04-15
preload does not work and downloaded twice on firefox


xhebox:
**Describe the bug**

1. Here is the firefox bug of module preload: https://bugzilla.mozilla.org/show_bug.cgi?id=1425310, i.e. firefox does not support modulepreload. There is a similar issue for sapper, https://github.com/sveltejs/sapper/issues/1387.

2. The new vite already has a builtin preload plugin https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/importAnalysisBuild.ts#L156-L162 on the client side. It will insert preload links despite sveltejs/kit also does that: https://github.com/sveltejs/kit/blob/1007f6730600b3f770c5062d385d539661d6d0c6/packages/kit/src/runtime/server/page/render.js#L104-L106.

Links from svelte are like `/./_app/pages/$layout.svelte-3b4aa0d5.js`, but vite checks `/_app/pages/$layout.svelte-3b4aa0d5.js`. So vite does not think the inserted link is duplicated. One fix is:

```
--- a/node_modules/@sveltejs/kit/dist/chunks/index5.js
+++ b/node_modules/@sveltejs/kit/dist/chunks/index5.js
@@ -229,8 +229,8 @@ async function build_server(
 
                find_deps(file, js_deps, css_deps);
 
-               const js = Array.from(js_deps).map((url) => prefix + url);
-               const css = Array.from(css_deps).map((url) => prefix + url);
+               const js = Array.from(js_deps).map((url) => path.join(prefix, url));
+               const css = Array.from(css_deps).map((url) => path.join(prefix, url));
 
                const styles = config.kit.amp
                        ? Array.from(css_deps).map((url) => {
```

3. Vite inserted links like `<link rel="preload" as="script" crossorigin="" href="/_app/pages/$layout.svelte-a23c4e9e.js">`, but it does not take effect somehow. So when sveltekit import scripts, resource will be downloaded again. 

**To Reproduce**

Create an app with layout, build & start, and you should be able to see the twice downloading on the latest firefox.

**Expected behavior**

Despite of preloading or not, I only want one time downloading.

**Information about your SvelteKit Installation:**
<details>
  <summary>Diagnostics</summary>
-

```
  System:
    OS: Linux 5.8 Noname Linux
    CPU: (16) x64 AMD Ryzen 7 4800H with Radeon Graphics
    Memory: 20.62 GB / 30.36 GB
    Container: Yes
[...]

PR #7400, opened 2022-10-26
Upgrade to Vite 3.2

benmccann:
I wasn't sure if I should include a changeset and didn't for now. This mostly affects tests. `create-svelte` does use a newer version, but you'd have already gotten the latest upon using `npm install` previously, so I'm not sure there's really a user-visible change here


Conduitry:
Does 3.2 provide any new features we're going to want to make use of? Before we hit SvelteKit 1.0 is the time to be doing any bumping of peerdeps.

benmccann:
There's not much in 3.2 that affects us. Only thing I see in the changelog is https://github.com/vitejs/vite/pull/9970, which is a bug fix for https://github.com/sveltejs/kit/issues/1042

dominikg:
just triggered a rerun of tests on https://github.com/sveltejs/kit/pull/7355  that should go in first otherwise it gets breaking changes from this here. ofc. we can also just redo it after, but would rather not

benmccann:
We can't merge this anyway. `lint` is failing because of https://github.com/vitejs/vite/issues/10656

dominikg:
someone mentioned adding types/picomatch to devDependencies as a workaround. We can either do that or wait until fixed in vite. Also checking if we can enable out lint in ecosystem-ci to catch these earlier.

dominikg:
fix has already been merged in vite, so better to wait here and bump to 3.2.1

Rich-Harris:
i bumped esbuild across the board so that we don't end up with two versions in the lockfile

benmccann:
Sounds good

Rich-Harris (approved):

PR #7401, opened 2022-10-26
Drop CJS version of adapter-netlify

benmccann:
The CJS version should be unused since https://github.com/sveltejs/kit/pull/6666, so this should not be a user visible change. I still added it to the changelog though so that we can cut a new release

If this works well, a followup might be to see if we can get rid of the Rollup bundling step in this project since it seems the other adapters don't need it

### DIFF
--- a/.changeset/bright-countries-jump.md
+++ b/.changeset/bright-countries-jump.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-netlify': patch
+---
+
+Drop CJS version of adapter-netlify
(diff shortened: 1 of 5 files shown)


dummdidumm (approved):

Rich-Harris:
I _think_ we'll still need the bundling step, because otherwise apps will need to declare dependencies on things like `set-cookie-parser`. Same applies to `adapter-node` with its dependencies on e.g. `sirv` and `polka`

benmccann:
> I think we'll still need the bundling step, because otherwise apps will need to declare dependencies on things like set-cookie-parser. Same applies to adapter-node with its dependencies on e.g. sirv and polka

Presumably the adapter dependencies would just get installed when you run `npm run install` and then the adapter would bundle them. I don't think the user would need to declare any dependencies. But anyway, not really a priority for me at the moment

PR #7407, opened 2022-10-27
[feat] add invalidateAll option to goto

dummdidumm:
Useful after form submissions for example, which will now use it in case of a redirect response

### Tests

### Changesets


geoffrich:
Is it possible to document what the `invalidateAll` option does? (mostly self explanatory, but still good to specify). Though it seems like the `goto` options [aren't currently documented](https://kit.svelte.dev/docs/modules#$app-navigation-goto) anyway.

dummdidumm:
Existing params are documented, but in a weird way in the jsdoc. Will adjust and add doc for invalidateAll

Rich-Harris (approved):

PR #7412, opened 2022-10-27
[fix] support and pin undici 5.12.0

### LINKED ISSUE #7409: +server.ts API endpoint fails to parse body
### Describe the bug

The only file new file in the blank svelte-kit project is the `+server.ts` located in `src/routes/api/+server.ts`. It just echoes the incoming JSON. Hitting `http://localhost:5173/api` (with `content-type` and `accept` headers set to `application/json` and body `{ "sample": "test"}` ) will return `HTTP 400 Bad request` -> and the request never hits the API endpoint.

If I remove the content-type and accept headers it will enter the API endpoint but it would error out on `await request.json()` with:

```
Unexpected end of JSON input
SyntaxError: Unexpected end of JSON input
```

It might be a Windows issue but honestly I am not sure and can't test atm.

### Reproduction

https://github.com/callmeberzerker/body-not-working

Run `pnpm install` and then `pnpm dev`.

### Logs

```Shell
Sadly there are none.
```

### System Info

```Shell
System:
    OS: Windows 10 10.0.22621
    CPU: (16) x64 Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
    Memory: 16.30 GB / 31.92 GB
  Binaries:
    Node: 19.0.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.19.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.674.0), Chromium (105.0.1343.42)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    vite: ^3.2.0 => 3.2.0
```

### Severity

### Additional Information

N/A

repsac-by:
Fixes #7409 

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/exports/node/index.js
+++ b/packages/kit/src/exports/node/index.js
@@ -95,6 +95,8 @@ function get_raw_body(req, body_size_limit) {
 /** @type {import('@sveltejs/kit/node').getRequest} */
 export async function getRequest({ request, base, bodySizeLimit }) {
 	return new Request(base + request.url, {
+		// @ts-expect-error
+		duplex: 'half',
 		method: request.method,
 		headers: /** @type {Record<string, string>} */ (request.headers),
 		body: get_raw_body(request, bodySizeLimit)
--- a/.changeset/gold-moons-poke.md
+++ b/.changeset/gold-moons-poke.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] support undici 5.12.0 and pin it
--- a/packages/kit/package.json
+++ b/packages/kit/package.json
@@ -21,7 +21,7 @@
 		"set-cookie-parser": "^2.4.8",
 		"sirv": "^2.0.2",
 		"tiny-glob": "^0.2.9",
-		"undici": "^5.11.0"
+		"undici": "5.12.0"
 	},
 	"devDependencies": {
 		"@playwright/test": "^1.25.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -288,7 +288,7 @@ importers:
       svelte-preprocess: ^4.10.6
       tiny-glob: ^0.2.9
       typescript: ^4.8.2
-      undici: ^5.11.0
+      undici: 5.12.0
       uvu: ^0.5.3
       vite: ^3.1.1
     dependencies:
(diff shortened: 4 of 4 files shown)


jalavosus:
Not to knock your fix (it's a good one) but wouldn't a more stable fix be to pin to the last known good version of undici (5.11.0)?

repsac-by:
There are many fixes and improvements in 5.12.0.
I don't see any good reason to stay with 5.11.0 other than the addition of the `duplex` property.

N00nDay:
Worked like a charm for me. Thanks repsac-by

benmccann:
The Undici team indicated this is the correct fix

>This is part of the spec now. When you have a ReadableStream body, the duplex option is required.

packages/kit/src/exports/node/index.js
```diff
@@ -95,6 +95,8 @@ function get_raw_body(req, body_size_limit) {
 /** @type {import('@sveltejs/kit/node').getRequest} */
 export async function getRequest({ request, base, bodySizeLimit }) {
 	return new Request(base + request.url, {
+		// @ts-expect-error
```
benmccann, line 98: I would expect we shouldn't need this. Is there some detail we can add here about how/when we might be able to remove it. Are the types provided by Typescript or Undici? Is there and open issue we can link to or can we file a new one?

Conduitry:
The undici PR included updates to the types here (https://github.com/nodejs/undici/pull/1681/files#diff-cfe294cd3fb885afcce564fae32dd1524399460a9a04f141e07bb45858a47f92), but it looks like we're not using those. I tries this out locally (and updated the Kit repo to undici 5.12.0), but the types for `Request` appeared to be coming from TypeScript (`lib.dom.d.ts`) and not from undici. This `@ts-expect-error` is probably the best we can do for now. Once the types are added to TS and we upgrade, we'll start getting an error here, so at least we'll remember to remove this.

This change looks good to me. Can we get a changeset?

packages/kit/src/exports/node/index.js
Conduitry: These types appear to be coming from TS, not from undici. Locally, even when I updated the Kit repo to undici 5.12.0, TS still complained about the `duplex`.

benmccann:
We should probably pin the Undici version as well as they've indicated that it's experimental and may have breaking changes in minor/patch versions: https://github.com/nodejs/undici/pull/1681#issuecomment-1294178466

Conduitry:
That sounds good to me as well.

benmccann (approved):

PR #7419, opened 2022-10-28
DRY out data suffix handling

Rich-Harris:
A thought I had while reviewing #7416 — our data suffix handling is a bit raggedy, and would maybe be nicer if we encapsulated the repetitious logic into helper functions.

No changeset or tests needed for this change

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/utils/url.js
+++ b/packages/kit/src/utils/url.js
@@ -142,3 +142,20 @@ export function disable_search(url) {
 		});
 	}
 }
+
+const DATA_SUFFIX = '/__data.json';
+
+/** @param {string} pathname */
+export function has_data_suffix(pathname) {
+	return pathname.endsWith(DATA_SUFFIX);
+}
+
+/** @param {string} pathname */
+export function add_data_suffix(pathname) {
+	return pathname.replace(/\/$/, '') + DATA_SUFFIX;
+}
+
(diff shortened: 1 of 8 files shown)


packages/kit/src/utils/url.js
```diff
@@ -142,3 +142,20 @@ export function disable_search(url) {
...
+}
+
+/** @param {string} pathname */
+export function add_data_suffix(pathname) {
+	return pathname.replace(/\/$/, '') + DATA_SUFFIX;
+}
+
+/** @param {string} pathname */
+export function strip_data_suffix(pathname) {
+	return pathname.slice(0, -DATA_SUFFIX.length);
```
dummdidumm, line 160: should we use `has_data_suffix` here and make this a noop in case it's not the suffix? Would reduce code at one call site.

dummdidumm (approved):
One suggestion, but I'm fine with it either way

packages/kit/src/utils/url.js
Rich-Harris: I'm not a huge fan of functions that are 'forgiving' about their inputs — it adds modest convenience (albeit at a single call site in our case, IOW it's just moving the logic around rather than reducing repetition) at the cost of greater uncertainty and wasted computation (however negligible)

PR #7424, opened 2022-10-28
reduce test flakiness, maybe

Rich-Harris:
seen these tests cause a lot of flakiness. i think this should fix it


dummdidumm (approved):
This makes sense either way. I just hope that the repaint happens soon enough (`await tick()` after the navigation should hopefully ensure that)

Rich-Harris:
IIUC the promise won't resolve until the DOM has updated

PR #7427, opened 2022-10-29
Fixed typo in in the SSR glossary

jacoblukewood:


jacoblukewood (commented):
Rendering on the client side would cause your app to be 'in'accessible if JavaScript fails or is disabled.

jacoblukewood:
I read this five times and thought it was wrong.

I made this pull request, read it again, and realised I read it wrong. Apologies.

Issue #6167, opened 2022-08-22
Generated enum type for `routeId`


Conduitry:
### Describe the problem

`$page.routeId` is just a string currently, which means you can have typos that won't match the ID of any route. It would be nice if this were typed, so that typos will be caught as type errors, and so that any renamed pages or moved routes will also result in type errors if their corresponding `$page.routeId`s aren't also updated.

### Describe the proposed solution

We know the route manifest at build time. We should be able to construct a `'foo' | 'bar' | 'baz'` type that the `routeId` would adhere to. The question would be where this generated type ought to live and how it ought to be accessed by the app code. Ideally, people could just keep using the ambient `$app/stores` types, and these would pull in the appropriate enum from `.svelte-kit`.

### Importance

nice to have

benmccann:
One thing to think about when deciding whether/how to implement this - if we added the ability to import routes from another project I'm not quite as sure how to support that

minht11:
Currently `RouteId` and `LayoutIds` have generated types, but are not exported
<img width="300" alt="Image" src="https://github.com/user-attachments/assets/65d11766-585c-49b8-a68b-8f734bcb98f9" />
you can access it somewhat with `type RouteId = LayoutLoadEvent['route']['id']` would exporting it directly be possible?

In my case I am using route id to have different view transitions on different routes, I changed routes names but forgot to change route ids, having explicit types would have caught that.

Nic-Polumeyv:
Since #13864 there's a generated `RouteId` union importable from `$app/types`, and `public.d.ts` now types `page.route.id` and the load events with it. That's the generated type this asked for. I think this can be closed.

PR #7450, opened 2022-10-31
[breaking] replace routeId with route.id

### LINKED ISSUE #6980: `routeId` not tracked as dependency to layout `load`
### Describe the bug

When you access `event.routeId` in a layout `load` function, the `load` function does not re-run when `routeId` changes. Since routeId is based on the URL, we should track dependencies to it similar to [url and params](https://kit.svelte.dev/docs/load#invalidation).

This probably only makes sense in a `layout` load. On `page` loads, routeId should stay the same.

### Reproduction

[StackBlitz](https://stackblitz.com/edit/sveltejs-kit-template-default-pfr37v?file=src/routes/+layout.server.js)

1. Load page. Notice routeId for the current page is displayed
2. Navigate to another page. The routeId does not change until a page reload

In this case you could work around the issue by retrieving it from the `$page` store, but if you're doing something else with `routeId`, that would not be re-run.

```js
// +layout.server.js
export const load = ({routeId}) => {
  return {
    routeId
  }
}
```

### Logs

```shell
n/a
```

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.80 
    @sveltejs/kit: next => 1.0.0-next.499 
    svelte: ^3.46.0 => 3.50.1 
    vite: ^3.1.0 => 3.1.3
```

### Severity

### Additional Information

_No response_

dummdidumm:
Closes #6980

Draft, because the tracking part is missing, and I'm not sure yet if we want to replace routeId in _all_ places (like `$page.routeId`)

### Tests

### Changesets


Conduitry:
Is there anything else we foresee living in `route`? `params` could get moved in there maybe?

Whatever we do, for consistency I'd be in favor of having `$page` be the same shape to the extent possible.

dummdidumm:
`params` could be moved into it, but that would be too big of a breaking change at this point, and it also results in worse DX, because you'd have to destructure the input twice then.

Rich-Harris:
The site is failing to deploy because a code snippet in the docs references `route.id` instead of `route?.id`. It'd be easy to fix, but I wonder if instead the types should reflect the fact that `route` always exists in the context of `load` or a `+server.js` handler?

Rich-Harris:
I made that change for `LoadEvent` and `ServerLoadEvent`, trying to figure out the best approach for `RequestHandler`. If I make this change...

```diff
export interface RequestHandler<
	Params extends Partial<Record<string, string>> = Partial<Record<string, string>>
> {
-	(event: RequestEvent<Params>): MaybePromise<Response>;
+	(event: RequestEvent<Params> & { route: { id: string } }): MaybePromise<Response>;
}
```

...then it works, but this is what I get if I hover over `route`:

image

Not totally ideal. Do we need to come up with a new type that extends `RequestEvent`? Or should the route ID be a generic argument? (That way we could make it an actual value for the generated types, or a list of possible values for layouts, etc)

dummdidumm:
I thought this, too, but in some cases the root layout is loaded but without a route - for example when an infinite redirect loop is detected, or in case of a 404 - many _ugh_-vibes on this PR 😅

Rich-Harris:
Ah crap, you're right — will revert the last commit.

What if we went the generic route though, and made `route` always exist but with `id` defaulting to `string | null`?

```ts
// .svelte-kit/types/src/routes/blog/$types.d.ts
export type PageLoad<...> = Kit.Load<..., OutputData, 'blog'>;
export type LayoutLoad<...> = Kit.Load<..., OutputData, 'blog' | 'blog/[slug]'>;
```

Or is it misleading for the route object to exist when no route matched?

dummdidumm:
Funny how we go through the same thought process independently. I asked myself the same question, and came to the conclusion that it's weird. Then again, all of this is weird, so we need to choose something that is the least weird while being easy enough to use. After having implemented this, part of me thinks `getRouteId()` would be better after all - unless you imagine any other things ending up in that object later on.
The generic is interesting, although I'm wondering if that typing maintenance is worth it.
What are people using `routeId` for anyway? Surely it's only in layouts?

Rich-Harris:
And just as I'd come around to `route.id`! It's definitely more future-proof, but I can't imagine what it would need to be future-proof against (i.e. I can't think of any properties we'd need beyond `id`). Then again that's sort of the point of being future-proof — you can't anticipate everything.

I think parity with `$page` is a reasonably convincing argument. We _could_ have `event.getRouteId()` and `$page.routeId` but it would definitely feel a bit more natural to have `event.route.id` and `$page.route.id`.

> What are people using routeId for anyway? Surely it's only in layouts?

Yeah, I think the main use case is where you want some special case logic in a layout ('require the user to be logged in, unless the current `routeId` is `/login`', or whatever).

dummdidumm:
I like `route.id` more for future-proofness and for how it looks when reading it. I'm not sure how I feel about `route: { id: string | null }` vs `route: { id: string } | null`. The latter feels more correct, but it's easier to produce runtime nullpointers with it. So ... yeah, maybe go with `route: { id: string | null }`

packages/kit/types/index.d.ts
```diff
@@ -627,15 +635,17 @@ export interface ServerLoadEvent<
 
 export interface Action<
 	Params extends Partial<Record<string, string>> = Partial<Record<string, string>>,
-	OutputData extends Record<string, any> | void = Record<string, any> | void
+	OutputData extends Record<string, any> | void = Record<string, any> | void,
+	RouteId extends string | null = string | null
```
dummdidumm, line 639: do we need the overhead of that interface on the action interfaces? Actions can only be used on pages, so routeId can only be one thing, so it's useless to use it, so probably noone does that, so we probably can just type this as `string` on `RequestEvent`

Rich-Harris:
This autocompletion is kinda neat:

https://user-images.githubusercontent.com/1162160/199323074-7657d332-5e74-40fb-8681-8597f35ee530.mov

Also, this typechecking 😍 

image

packages/kit/types/index.d.ts
Rich-Harris: Is there any real drawback? It's contrived but you could imagine someone doing this sort of thing, where `some_shared_logic` is picky about the second argument:


```js
export const actions = {
  blah: ({ request, route }) => {
    some_shared_logic(request, route.id);
  }
};
```

Conduitry:
Does that mean this also closes #6167?

dummdidumm:
no, because that relates to `$page.routeId`, but you are probably right that this can also be infered from that.

Rich-Harris:
oops, didn't add the tracking part yet. ignore that i marked this PR ready

Conduitry:
FYI you can mark PRs as draft again. It's part of the "Reviewers" section in the upper right ("Still in progress? Convert to draft") which yes is an insane place for this to be.

Rich-Harris:
Interesting wrinkle I found while adding a test: this...

```js
export function load({ route }) {
  return { route };
}
```

...doesn't count as 'using' `route.id` even if the page renders it:

```svelte
<script>
  /** @type {import('./$types').LayoutData} */
  export let data;
</script>

<h1>route.id: {data.route.id}</h1>
<slot />
```

That's because the `uses` object is updated too late (after rendering/serialization). Need to thunkify it or something

dummdidumm:
Doesn't the same apply to params and url as well? Not sure if we need to optimize for this.

Rich-Harris:
Ok that was a bit of a faff but it will now correctly identify the dependency on `route.id` if you have this...

```js
export function load({ route }) {
  return { route };
}
```

.,..and the same goes for `url` and `params`. Surprised we didn't get bitten by that bug sooner

Rich-Harris (approved):

PR #7453, opened 2022-10-31
Read mode/credentials from `init` object, not `request`

Rich-Harris:
This is an alternative to #7192. Rather than checking for a `cf` property, it reads `mode` and `credentials` from the second argument passed to `fetch`, unless `fetch` was called with a `Request` object (which is a somewhat unusual case; code that does `fetch(request)` would have to be refactored to `fetch(url, init)` on Cloudflare). 

I'm slightly ambivalent about which approach is better. This gets rid of the Cloudflare-specific `cf` stuff, and in theory is more resilient to other runtimes doing the same thing, but it would force the aforementioned refactoring in a very small minority of cases.

### Tests

### Changesets


dummdidumm (approved):
I'd say we go with this first and then see how often it comes up afterwards? We can always add the other PR on top/instead later on.

PR #7461, opened 2022-11-01
[feat] add service-worker to demo app

### LINKED ISSUE #2246: Default Service Worker (in demo app or/and docs)
### Describe the problem

Let think about default configs and options and some features for svelte-create. I think, now configuring svelte takes a little more steps than is necesary. (mainly point 3 and 4 below)

Today we have:

- [ ] ~~1. Demo app as default template, I think Skeleton one would be better (as is more often used)~~
- [ ] ~~2. app.html as default template html file, why it's not index.html (index.html is "standard")~~
- [ ] 3. what about adding another option to CLI svelte-create "create service-worker"? It could probably use modified sapper service-worker file - https://gist.github.com/Mlocik97/cf0f4c39e46d45404fda8808a25244c8
- [x] 4. it would be probably nice and faster to add option to CLI create-svelte to choose from some adapters: like "use node adapter", "use static adapter", "I will choose adapter later" options. (when using node adapter, it can also add `"start": "node ./build"` to package.json scripts. #2459 

### Alternatives considered

Let it be as it is now.

### Importance

nice to have

### Additional Information

_No response_

dummdidumm:
Also more docs
Closes #2246

### Tests

### Changesets

### DIFF
--- a/documentation/docs/30-advanced/40-service-workers.md
+++ b/documentation/docs/30-advanced/40-service-workers.md
@@ -4,10 +4,73 @@ title: Service workers
 
 Service workers act as proxy servers that handle network requests inside your app. This makes it possible to make your app work offline, but even if you don't need offline support (or can't realistically implement it because of the type of app you're building), it's often worth using service workers to speed up navigation by precaching your built JS and CSS.
 
-In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker.ts`, or `src/service-worker/index.js`, etc) it will be built with Vite and automatically registered. You can disable automatic registration if you need to register the service worker with your own logic (e.g. prompt user for update, configure periodic updates, use `workbox`, etc).
+In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker.ts`, or `src/service-worker/index.js`, etc) it will be built with Vite and automatically registered. You can change the [location of your service worker](/docs/configuration#files) if you need to.
 
-> You can change the [location of your service worker](/docs/configuration#files) and [disable automatic registration](/docs/configuration#serviceworker) in your project configuration.
+You can [disable automatic registration](/docs/configuration#serviceworker) if you need to register the service worker with your own logic (e.g. prompt user for update, configure periodic updates, use `workbox`, etc). The default registration looks something like this:
 
-Inside the service worker you have access to the [`$service-worker` module](/docs/modules#$service-worker). If your Vite config specifies `define`, this will be applied to service workers as well as your server/client builds.
+```js
+if ('serviceWorker' in navigator) {
+	addEventListener('load', function () {
+		navigator.serviceWorker.register('./path/to/service-worker.js');
+	});
+}
+```
 
-Because it needs to be bundled (since browsers don't yet support `import` in this context), and depends on the client-side app's build manifest, **service workers only work in the production build, not in development**. To test it locally, use `vite preview`.
+Inside the service worker you have access to the [`$service-worker` module](/docs/modules#$service-worker), which provides you with the paths to all static assets, build files and prerendered pages. You're also provided with an app version string which you can use for creating a unique cache name. If your Vite config specifies `define` (used for global variable replacements), this will be applied to service workers as well as your server/client builds.
+
+The following example caches all files from the build eagerly and also caches all other requests occuring later on. This would make each page work offline once visited.
+
+```js
+// @ts-nocheck Official TS Service Worker typings are still a work in progress.
+import { build, files, version } from '$service-worker';
+
+// Give the current service worker instance a unique cache name
+const CACHE_NAME = `cache-${version}`;
+// In this simple example, we just cache all files
+const to_cache = build.concat(files);
+
+self.addEventListener('install', (event) => {
+	// Create new cache and add all files to it
+	async function addFilesToCache() {
+		const cache = await caches.open(CACHE_NAME);
+		await cache.addAll(to_cache);
+	}
+
+	event.waitUntil(addFilesToCache());
+});
+
+self.addEventListener('activate', (event) => {
+	// Remove previous cached data from disk
+	async function deleteOldCaches() {
+		const keyList = await caches.keys();
+		const cachesToDelete = keyList.filter((key) => key !== CACHE_NAME);
+		await Promise.all(cachesToDelete.map((key) => caches.delete(key)));
+	}
+
+	event.waitUntil(deleteOldCaches());
+});
+
+self.addEventListener('fetch', (event) => {
+	// Try to get the response from the cache, add to cache if not found
+	async function addToCache(request, response) {
+		const cache = await caches.open(CACHE_NAME);
+		await cache.put(request, response);
+	}
+
+	async function cacheFirst(request) {
+		const responseFromCache = await caches.match(request);
+		if (responseFromCache) {
+			return responseFromCache;
+		}
+		const response = await fetch(request);
+		addToCache(request, response.clone());
+		return response;
+	}
+
+	event.respondWith(cacheFirst(event.request));
+});
+```
+
+> Careful with caching too much for too long! Browsers have a limit on the amount they cache, and not everything should be cached, for example requests that contain dynamic data that change over time
+
+Because the service worker needs to be bundled (since browsers don't yet support `import` in this context), and depends on the client-side app's build manifest, **service workers only work in the production build, not in development**. To test it locally, use `vite preview` after running a build.
--- a/.changeset/four-llamas-travel.md
+++ b/.changeset/four-llamas-travel.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+[feat] add service-worker to demo app
--- a/packages/create-svelte/templates/default/src/service-worker.js
+++ b/packages/create-svelte/templates/default/src/service-worker.js
@@ -1,0 +1,55 @@
+// @ts-nocheck Official TS Service Worker typings are still a work in progress.
+import { build, files, version } from '$service-worker';
+
+// This service worker eagerly loads and caches all static files and assets of the application.
+// It also caches visited pages, so that once you visited a page, it can be accessed offline.
+// The exception is the Sverdle game as that's highly dynamic and doesn't make sense to be cached.
+
+// Give the current service worker instance a unique cache name
+const CACHE_NAME = `cache-${version}`;
+// In this simple example, we just cache all files
+const to_cache = build.concat(files);
+
+self.addEventListener('install', (event) => {
+	// Create new cache and add all files to it
+	async function addFilesToCache() {
+		const cache = await caches.open(CACHE_NAME);
+		await cache.addAll(to_cache);
+	}
+
+	event.waitUntil(addFilesToCache());
+});
+
+self.addEventListener('activate', (event) => {
+	// Remove previous cached data from disk
+	async function deleteOldCaches() {
+		const keyList = await caches.keys();
+		const cachesToDelete = keyList.filter((key) => key !== CACHE_NAME);
+		await Promise.all(cachesToDelete.map((key) => caches.delete(key)));
+	}
+
+	event.waitUntil(deleteOldCaches());
+});
+
+self.addEventListener('fetch', (event) => {
+	// Try to get the response from the cache, add to cache if not found
+	async function addToCache(request, response) {
+		const cache = await caches.open(CACHE_NAME);
+		await cache.put(request, response);
+	}
+
+	async function cacheFirst(request) {
+		const responseFromCache = await caches.match(request);
+		if (responseFromCache) {
+			return responseFromCache;
+		}
+		const response = await fetch(request);
+		// Don't cache Sverdle game interactions, they are dynamic
+		if (!request.url.includes('sverdle')) {
+			addToCache(request, response.clone());
+		}
+		return response;
+	}
+
+	event.respondWith(cacheFirst(event.request));
+});


benmccann:
I'm not sure how I feel about this. Service workers stick around in the browser so if you run one app with a service worker and then run a different app the service worker from the previous one is still there so it's a very strong form of caching. It makes debugging so much more annoying because I'm constantly having to remove service workers from reproduction reports and sometimes forget it's needed and it confuses users who don't know it will be happening. Personally I'm not sure I would add a service worker at all, but if we do I'd at least argue for it to be an option with a default value of false.

documentation/docs/30-advanced/40-service-workers.md
```diff
@@ -4,10 +4,73 @@ title: Service workers
...
 
-In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker.ts`, or `src/service-worker/index.js`, etc) it will be built with Vite and automatically registered. You can disable automatic registration if you need to register the service worker with your own logic (e.g. prompt user for update, configure periodic updates, use `workbox`, etc).
+In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker.ts`, or `src/service-worker/index.js`, etc) it will be built with Vite and automatically registered. You can change the [location of your service worker](/docs/configuration#files) if you need to.
 
-> You can change the [location of your service worker](/docs/configuration#files) and [disable automatic registration](/docs/configuration#serviceworker) in your project configuration.
+You can [disable automatic registration](/docs/configuration#serviceworker) if you need to register the service worker with your own logic (e.g. prompt user for update, configure periodic updates, use `workbox`, etc). The default registration looks something like this:
```
benmccann, line 9:
```suggestion
You can [disable automatic registration](/docs/configuration#serviceworker) if you need to register the service worker with your own logic (e.g. prompt user for update, configure periodic updates, [use `workbox`](https://vite-pwa-org.netlify.app/frameworks/sveltekit.html), etc). The default registration looks something like this:
```

dummdidumm: I don't think just adding a link here is enough. From what I understand, that plugin works completely different from SvelteKit and has zero overlap (you can't use them together). So I think it would be better to have a separate paragraph on that.

dummdidumm:
The service worker hanging around resulting in confusion is a good point - but don't you have that problem anyway as soon as you use any form of service-worker? Also, it's only for the demo app. What if we prefix the file with an underscore and add a comment to the file and the about page that you need to "opt in" to the service worker, and why we do that?

benmccann:
Yes, you'd probably have that problem with any form of service worker. Making it unused would be fine. Perhaps it would be clearer to call it something like `disabled-service-worker.js` than an underscore prefix?

documentation/docs/30-advanced/40-service-workers.md
benmccann: You can use it with SvelteKit, but I don't think you'd use it with SvelteKit's service worker. This was in a sentence about disabling SvelteKit's service worker, so I was assuming that that would be disabled first before adding the plugin, but maybe it could be made clearer with something more explicit like:

```suggestion
You can [disable automatic registration](/docs/configuration#serviceworker) if you need to register the service worker with your own logic. Examples of this would include prompting the user for update, configuring periodic updates, using `workbox`, etc. You may find the [Vite PWA plugin](https://vite-pwa-org.netlify.app/frameworks/sveltekit.html) helpful in these cases as it provides these functionalities and has built-in SvelteKit support. The default registration looks something like this:
```

dummdidumm: I expressed myself poorly. What I mean is that you probably can't use the `$service-worker` imports with that plugin (although I haven't tested that, but from their documentation it sounds like it isn't possible), so you can either use that plugin or use SvelteKit's service worker things, but not both.

benmccann: Oh yeah. I don't think they support it. Possibly they could add it at the cost of breaking dev mode support, but at least for now they don't have any support for it

dummdidumm: Added a paragraph highlighting alternatives and moves the Vite PWA plugin mention there.

dummdidumm:
Removed the service worker from the demo, more details in the PR description

documentation/docs/30-advanced/40-service-workers.md
```diff
@@ -4,10 +4,75 @@ title: Service workers
 
 Service workers act as proxy servers that handle network requests inside your app. This makes it possible to make your app work offline, but even if you don't need offline support (or can't realistically implement it because of the type of app you're building), it's often worth using service workers to speed up navigation by precaching your built JS and CSS.
 
-In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker.ts`, or `src/service-worker/index.js`, etc) it will be built with Vite and automatically registered. You can disable automatic registration if you need to register the service worker with your own logic (e.g. prompt user for update, configure periodic updates, use `workbox`, etc).
+In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker.ts`, or `src/service-worker/index.js`, etc) it will be built with Vite and automatically registered. You can change the [location of your service worker](/docs/configuration#files) if you need to.
```
benmccann, line 7: Vite doesn't really support service workers. I'd hide this as an implementation detail or else people may try to configure things in `vite.config.js` which I'm not sure we'd want.
```suggestion
In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker.ts`, `src/service-worker/index.js`, etc.) it will be bundled and automatically registered. You can change the [location of your service worker](/docs/configuration#files) if you need to.
```

documentation/docs/30-advanced/40-service-workers.md
```diff
@@ -4,10 +4,75 @@ title: Service workers
...
+
+	event.respondWith(cacheFirst(event.request));
+});
+```
+
+> Careful with caching too much for too long! Browsers have a limit on the amount they cache, and not everything should be cached, for example requests that contain dynamic data that change over time
+
+Because service workers need to be bundled (since browsers don't yet support `import` in this context), **SvelteKit's service workers only work in the production build, not in development**, since it depends on the client-side app's build manifest. To test it locally, use `vite preview` after running a build.
+
+SvelteKit's service worker implementation is deliberately low-level. If you need a more full-flegded but also more opinionated solution, we recommend looking at solutions like [Workbox](https://web.dev/learn/pwa/workbox) or [Vite PWA plugin](https://vite-pwa-org.netlify.app/frameworks/sveltekit.html). For more general information on service workers, we recommend [the MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers).
```
benmccann, line 78: Vite PWA plugin uses Workbox. I'm not sure I'd suggest people to try setting up workbox directly as it'd be quite complicated. Maybe we could say something like:

```
[Vite PWA plugin](https://vite-pwa-org.netlify.app/frameworks/sveltekit.html), which uses [Workbox](https://web.dev/learn/pwa/workbox)
```

documentation/docs/30-advanced/40-service-workers.md
```diff
@@ -4,10 +4,75 @@ title: Service workers
...
 
-In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker.ts`, or `src/service-worker/index.js`, etc) it will be built with Vite and automatically registered. You can disable automatic registration if you need to register the service worker with your own logic (e.g. prompt user for update, configure periodic updates, use `workbox`, etc).
+In SvelteKit, if you have a `src/service-worker.js` file (or `src/service-worker.ts`, or `src/service-worker/index.js`, etc) it will be built with Vite and automatically registered. You can change the [location of your service worker](/docs/configuration#files) if you need to.
 
-> You can change the [location of your service worker](/docs/configuration#files) and [disable automatic registration](/docs/configuration#serviceworker) in your project configuration.
+You can [disable automatic registration](/docs/configuration#serviceworker) if you need to register the service worker with your own logic (e.g. prompt user for update, configure periodic updates, use another solution like `workbox`, etc). The default registration looks something like this:
```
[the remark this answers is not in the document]
dummdidumm: what about
```suggestion
You can [disable automatic registration](/docs/configuration#serviceworker) if you need to register the service worker with your own logic (e.g. to prompt the user for update, configure periodic updates) or use another solution like `workbox` or an external plugin. The default registration looks something like this:
```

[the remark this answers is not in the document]
benmccann: I don't know. It's still implying you have to implement it yourself if you want to prompt the user for update or have periodic updates. Maybe we should just remove the list of things you might want to do from this section?

```suggestion
You can [disable automatic registration](/docs/configuration#serviceworker) if you need to register the service worker with your own logic or use another solution. The default registration looks something like this:
```

documentation/docs/30-advanced/40-service-workers.md
```diff
@@ -4,10 +4,75 @@ title: Service workers
...
+
+	event.respondWith(cacheFirst(event.request));
+});
+```
+
+> Careful with caching too much for too long! Browsers have a limit on the amount they cache, and not everything should be cached, for example requests that contain dynamic data that change over time
+
+Because service workers need to be bundled (since browsers don't yet support `import` in this context), **SvelteKit's service workers only work in the production build, not in development**, since it depends on the client-side app's build manifest. To test it locally, use `vite preview` after running a build.
+
+SvelteKit's service worker implementation is deliberately low-level. If you need a more full-flegded but also more opinionated solution, we recommend looking at solutions like [Vite PWA plugin](https://vite-pwa-org.netlify.app/frameworks/sveltekit.html), which uses [Workbox](https://web.dev/learn/pwa/workbox). For more general information on service workers, we recommend [the MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers).
```
benmccann, line 78: we could add a heading here to make it linkable from above

```suggestion
## External solutions

SvelteKit's service worker implementation is deliberately low-level. If you need a more full-flegded but also more opinionated solution, we recommend looking at solutions like [Vite PWA plugin](https://vite-pwa-org.netlify.app/frameworks/sveltekit.html), which uses [Workbox](https://web.dev/learn/pwa/workbox). For more general information on service workers, we recommend [the MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers).
```

dummdidumm: Given that the section isn't very big I'm not sure we need this. "On this page" would look rather empty, and the only entry being "External Solutions" feels kinda weird. So if we want to, we would need another heading somewhere.

benmccann: yeah, that's a fair point. other ideas for headings could be:
- creation
- registration (probably a sub-heading of the previous one)
- example

documentation/docs/30-advanced/40-service-workers.md
```diff
@@ -4,10 +4,75 @@ title: Service workers
...
+		return response;
+	}
+
+	event.respondWith(cacheFirst(event.request));
+});
+```
+
+> Careful with caching too much for too long! Browsers have a limit on the amount they cache, and not everything should be cached, for example requests that contain dynamic data that change over time
+
+Because service workers need to be bundled (since browsers don't yet support `import` in this context), **SvelteKit's service workers only work in the production build, not in development**, since it depends on the client-side app's build manifest. To test it locally, use `vite preview` after running a build.
```
benmccann, line 76: Should we use the `>` syntax here instead of bold to match the style of call-outs elsewhere? I might also move this sentence to come after the one starting with "Inside the service worker you have access to the [`$service-worker` module](/docs/modules#$service-worker)" since it's more related to accessing modules than the example service worker. And if we added more headings I think it'd make more sense with those headings

documentation/docs/30-advanced/40-service-workers.md
userquin: I'll try to release the pwa plugin for kit next week. I need to merge a few PR and publish a New pwa plugin version before releasing pwa plugin integrations.
With https://github.com/vite-pwa/vite-plugin-pwa/pull/390 we can use something similar to kit sw, just check the sw in the PR example.

documentation/docs/30-advanced/40-service-workers.md
dummdidumm: The boldness is obsolete with the dev service worker changes

documentation/docs/30-advanced/40-service-workers.md
```diff
@@ -4,11 +4,76 @@ title: Service workers
...
+	event.waitUntil(addFilesToCache());
+});
+
+self.addEventListener('activate', (event) => {
+	// Remove previous cached data from disk
+	async function deleteOldCaches() {
+		const keyList = await caches.keys();
+		const cachesToDelete = keyList.filter((key) => key !== CACHE_NAME);
+		await Promise.all(cachesToDelete.map((key) => caches.delete(key)));
+	}
```
Rich-Harris, line 50: Unless something has gone wrong, we should only ever need to delete a single old cache (and in any case, it should be a very fast operation) so i think we can use this theoretically-slower-but-in-practice-identical code which I think is a little easier to parse

```suggestion
	async function deleteOldCaches() {
		for (const key of await caches.keys()) {
			if (key !== CACHE_NAME) await caches.delete(key);
		}
	}
```

documentation/docs/30-advanced/40-service-workers.md
```diff
@@ -4,11 +4,76 @@ title: Service workers
...
+		const response = await fetch(request);
+		addToCache(request, response.clone());
+		return response;
+	}
+
+	event.respondWith(cacheFirst(event.request));
+});
+```
+
+> Careful with caching too much for too long! Browsers have a limit on the amount they cache, and not everything should be cached, for example requests that contain dynamic data that change over time
```
Rich-Harris, line 76: Since people are likely to just copy whatever we put in the docs, I'm not sure we should use a cache-first strategy here — it's a footgun. Cache-first makes sense for `build` and `files`, but for pages we should probably have a network-first strategy instead (and I think we should only fall back to the cache for GET requests?)

Rich-Harris: https://github.com/sveltejs/kit/pull/7832

userquin: Nuxt uses this workbox preset/tenplate: https://github.com/vite-pwa/vite-plugin-pwa/pull/390/files#diff-601038314cf235badafad57b9aca2c222c80d89c6a5153ee0708158a1efdb0d3R27

Rich-Harris (approved):

PR #6129, opened 2022-08-20
fix `git mv` commands when migrating files with `$` characters

Rich-Harris:
This should fix #6113 — tested with the case in that issue, works fine, though I'd appreciate a sanity check from others

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #7462, opened 2022-11-01
[feat] install adapters on demand

### LINKED ISSUE #5123: Auto adapter should download other adapters on demand

dummdidumm:
Closes #5123

Marking as draft because I still need to test this on one of the providers, with npm/pnpm/yarn

### Tests

### Changesets


packages/adapter-auto/index.js
```diff
@@ -10,31 +13,39 @@ for (const candidate of adapters) {
...
 			if (
 				error.code === 'ERR_MODULE_NOT_FOUND' &&
 				error.message.startsWith(`Cannot find package '${candidate.module}'`)
 			) {
-				throw new Error(
-					`It looks like ${candidate.module} is not installed. Please install it and try building your project again.`
-				);
+				try {
+					execSync(
+						`echo "Installing ${candidate.module}" && npm install ${candidate.module} --no-save --no-package-lock`,
```
Rich-Harris, line 23: TIL `--no-package-lock` — if that allows this to work across package managers, that'd be pretty sweet. 

Should we add a message along the lines of 'you should add this to your package.json for faster build times in future?'

packages/adapter-auto/index.js
```diff
@@ -10,31 +13,39 @@ for (const candidate of adapters) {
 
 		try {
 			module = await import(candidate.module);
```
Rich-Harris, line 15: am suddenly wondering if we should actually be importing the module from the app directory rather than the adapter-auto directory. in most cases it should Just Work, but you can imagine a situation where `adapter-auto` is installed in the workspace root while `adapter-foo` is installed inside the app

dummdidumm: Doesn't `candidate.module` just contain the name of the package, and the resolution algorithm should start at the `adapter-auto` directory and look for node_modules there, and if it's there, look for that module there? At least I (try to) use this fact by installing the package into the `adapter-auto` directory.

Rich-Harris: If the adapter is a dependency of the app (which it would be, if you'd already installed it) then surely resolution should start from there? (Until `import.meta.resolve` is stable, this would need [import-meta-resolve](https://www.npmjs.com/package/import-meta-resolve))

packages/adapter-auto/index.js
Conduitry: That'd only help people who look at the build logs, right? I wonder how many people that is.

Also, is there a reason for the `echo` to be part of the `exec` as well, instead of just a regular `console.log`?

If we get rid of that, can we do something with `execFileSync` like what we're doing [here](https://github.com/sveltejs/kit/blob/1798dad0b962b0aaca56c7aba9ae6a30b196d4d9/packages/migrate/migrations/routes/utils.js#L49) so we can completely avoid the shell's processing and escaping?

packages/adapter-auto/index.js
dummdidumm: I feel like this overcomplicates things for no good reason. In what world are you using adapter-auto, but have installed a more specific adapter in another place? You either have them in the same place or switched to the one you actually want to use a long time ago.

packages/adapter-auto/index.js
dummdidumm: What's the advantage to that? The current approach forwards the output to the main shell, so it _could_ help in case something goes wrong.

Rich-Harris: > That'd only help people who look at the build logs, right? I wonder how many people that is.

a small number, but larger than zero

> What's the advantage to that? The current approach forwards the output to the main shell

We'd still pipe `stderr` and `stdout`. The advantage is that there's no caveats around `candidate.module` being some weird string that we need to deal with — npm package names _should_ be fine, but for an example of the category of bugs it prevents, see https://github.com/sveltejs/kit/pull/6129/files.

Also `execFileSync` should be slightly faster as it doesn't involve a shell

dummdidumm:
The approach here seems to work when it comes to installing, but for some reason it
- says "cannot use import outside a module" when installing the adapter inside the auto-adapter's node_modules
- deletes some needed modules from node_modules when installing it in the top level modules

will continue to tinker  around

packages/adapter-auto/index.js
Rich-Harris: It's not so hard to imagine a situation where shared dependencies are installed in the workspace root but someone installs a one-off dependency in a package — this very repo used to work that way, until we decided to move all dependencies into packages. I also wonder if 'dependencies of `x` can import all other dependencies of `x`' is guaranteed across all package managers now and in the future. It's a side-effect of the resolution algorithm plus the way package managers populate `node_modules`, but it's the sort of thing that feels changeable, the same way pnpm prevents the npm 3+ behaviour of allowing `x` to directly import all indirect dependencies

packages/adapter-auto/index.js
dummdidumm: tried switching to it, can't switch because windows can't execute npm without the `.cmd` ending in this case

Rich-Harris:
😬 

this might be easier said than done

image

dummdidumm:
goddammit, why is this so unreliable to test?

Another idea:
modify the user's package.json and swap out adapter-auto with adapter-whatever-matches, then run npm/yarn/pnpm install again

Rich-Harris:
I don't love solving problems by adding dependencies, but we could experiment with https://github.com/antfu/ni

Conduitry:
I'm all for finding a better way to do this, but I think that can happen incrementally, and for right now, pretty much anything would be better than utterly flabbergasting dependency tree that `@vercel/nft` brings in.

I see that `ni` has no prod dependencies of its own, which is nice.

Some other options might be to `npm install <adapter>` in some temporary directory (maybe with a dummy `package.json` file?) and then move those files over into the user's `node_modules` folder. Although, yes, it certainly seems mysterious that the `npm install` we're doing here would be deleting the other dependencies in the first place.

Rich-Harris:
I thought about the 'install in some other directory option' but I think with npm you'd end up with a flat directory containing the adapter and all its dependencies, and then you wouldn't be able to move them into the real `node_modules` without conflicts with existing dependencies, which is fine if the versions match but somewhat hairy otherwise

dummdidumm:
The solution right now already install things in a separate node_modules folder, namely that inside of `adapter-auto`, and from Rich's comment it seems that doesn't work - so I'll give `ni` a shot.

dummdidumm:
> 😬
> 
> this might be easier said than done
> 
> image

@Rich-Harris how did you test this PR? Did you use `patch-package`? I settled for testing it out differently; by adding the modified package in a sibling folder and pointing the package.json to it: `"@sveltejs/adapter-auto": "./adapter-auto`

dummdidumm:
I'm completely lost as why running an install command from `execSync` (or `execFileSync`, or using a third party command line package that claims to have better support in Windows) just deletes all other packages. If I just run `npm install` all packages are deleted - it's as if the package.json's (dev)dependencies are seen as empty.

Rich-Harris:
> how did you test this PR? Did you use `patch-package`?

I'm just looking at the normal build logs. I didn't think there'd be any need to do anything funky since we're not making any changes to the package we're trying to lazily install

dummdidumm:
I still don't know why using `execSync` (or any of the related commands) removes all other node_modules. It seems that doing the "on-the-fly"-install inside the `postinstall` hook works better for this. But looking at the CI on Vercel that still seems to have problems, because for some reason the `prepare` hook of TypeScript is executed, which wants to start a gulp script, which fails. And I'm not sure installing anything outside of the `adapter-auto` directory is a good idea - we simply don't know where the install command was executed. We can only guess by looking for a lock file.

Rich-Harris:
I took a run at this tonight and I'm absolutely flummoxed. I tried replacing `adapter-auto/index.js` with just this...

```js
import { execSync } from 'child_process';

execSync(`pnpm install`, {
  stdio: 'inherit'
});

export default () => ({
  name: '@sveltejs/adapter-auto',
  adapt: (builder) => {}
});
```

...and tried `pnpm build` inside my `kit-sandbox` app (which is a sibling of the `kit` monorepo, and has the `kit` and `adapter-auto` packages linked), and the same thing happens — all my dependencies get nuked.

That's weird, but what elevates it to truly batshit is that if I invoke that code directly by doing this...

```
node svelte.config.js
```

...then I get the expected 'Already up to date' message; `pnpm install` works correctly. The files only get deleted when that code is being invoked via `vite build`.

I tried running `pnpm install` inside `vite-plugin-svelte`, right before `svelte.config.js` gets imported...

```diff
// node_modules/@sveltejs/vite-plugin-svelte/dist/index.js
+execSync('pnpm install', { stdio: 'inherit' });

const result = await dynamicImportDefault(
  pathToFileURL(configFile).href,
  fs2.statSync(configFile).mtimeMs
);
```

...and it nukes everything. Something very, very weird is happening. The `package.json` and `pnpm-lock.yaml` files are untouched when the code runs. I am very confused and a little bit scared.

Rich-Harris:
Oooohhhhhhhhhhh

It's because Vite is setting `process.env.NODE_ENV = 'production'`, which apparently causes `pnpm install` to nuke `devDependencies`

Rich-Harris:
`pnpm install --dev` _only_ installs dev dependencies, and `pnpm install --dev --prod` apparently doesn't install both 🙄 . The only thing we can do is reset `process.env.NODE_ENV` for the duration of the installation.

Will carry on with this tomorrow now that I'm unblocked

packages/adapter-auto/index.js
```diff
@@ -10,31 +13,55 @@ for (const candidate of adapters) {
...
+					console.log(`Installing ${candidate.module} on the fly...`);
+					execSync(
+						`${process.platform === 'win32' ? 'set' : ''} NODE_ENV=ignore_me && npm install ${
+							candidate.module
+						} --no-save --omit=dev --no-package-lock`,
+						{
+							stdio: 'inherit',
+							cwd: dirname(fileURLToPath(import.meta.url))
+						}
+					);
```
Conduitry, line 33: We should be able to manually pass env vars with the `env` option https://nodejs.org/dist/latest-v19.x/docs/api/child_process.html#child_processexecsynccommand-options

If we do that, we can also avoid involving the shell at all, and we can use `execFileSync` here instead.

packages/adapter-auto/index.js
Conduitry: Can we have `process.platform === 'win32' ? 'npm.cmd' : 'npm'` as the first argument of `execFileSync`? Does that work?

dummdidumm:
Still don't understand why running `npm install` inside the `adapter-auto` folder doesn't work. Running `npm install` from the root without `NODE_ENV` set works. Now the question is if we feel confident enough to always know where to run that command from. `process.cwd()` isn't necessarily the same directory from which `pnpm/npm/yarn install` ran.

packages/adapter-auto/index.js
dummdidumm: `execFileSync` needs `.cmd` endings for the `npm/pnpm/yarn` command on windows last time I tested this. So we either have to prepend that on windows or keep using `execSync`

dummdidumm:
I think so, too, but this wouldn't help you in case of a monorepo, where you run `pnpm install` at the root but `pnpm build` inside a sub directory (like the Kit monorepo). Likelyhood of this situation is probably small given that the adapters are probably run scoped to the current package, but who knows.

packages/adapter-auto/index.js
Conduitry: Yeah I saw your comment [in that thread](https://github.com/sveltejs/kit/pull/7462#discussion_r1011831808) after I commented here. I responded over on that thread. Even if we can't make `execFileSync` work, it would be nice to avoid the manual setting of the env var using shell commands.

I'm actually not even sure that `FOO=bar && do-something` would call `do-something` with `FOO` set to `bar`. I don't think environment variables automatically inherit that way in shell scripts without an `export`.

benmccann (approved):

PR #7484, opened 2022-11-03
[breaking] consistent casing in `goto`

### LINKED ISSUE #7460: Make `goto` options casing more consistent?
### Describe the problem

`goto` has this signature:
```ts
function goto(
  url: string | URL,
  opts?: {
    replaceState?: boolean;
    noscroll?: boolean;
    keepfocus?: boolean;
    state?: any;
  }
): Promise<void>;
```

`replaceState` is camel-cased while `noscroll` and `keepfocus` are not.

Rename:
- `noscroll` -> `noScroll`
- `keepfocus` -> `keepFocus`

(This also would make spell checkers happier.)

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

aantusahaa:
Closes #7460

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -183,6 +183,14 @@ export function create_client({ target, base, trailing_slash }) {
 			url = new URL(url, get_base_uri(document));
 		}
 
+		if (arguments[1].hasOwnProperty('keepfocus')) {
+			throw new Error('keepfocus has been renamed to keepFocus (note the difference in casing)');
+		}
+
+		if (arguments[1].hasOwnProperty('noscroll')) {
+			throw new Error('noscroll has been renamed to noScroll (note the difference in casing)');
+		}
+
 		return navigate({
 			url,
 			scroll: noScroll ? scroll_state() : null,
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -1150,13 +1150,14 @@ export function create_client({ target, base, trailing_slash }) {
 		},
 
 		goto: (href, opts = {}) => {
-			if (opts.hasOwnProperty('keepfocus')) {
+			// TODO remove for 1.0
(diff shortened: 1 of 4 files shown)


dummdidumm:
Thank you! Could you add checks to the public API call sites to check if people are calling it with the old casing, and throwing an error in that case? Something along the lines of "X has been renamed to Y (note the difference in casing)".

aantusahaa:
`keepfocus` could also be renamed for `update` and `navigate` methods as well. Should that be done?

dummdidumm:
Good question. It would safe a couple of characters in the final output because you can then use shorthands, but it conflicts with the internal coding style. I'm a little towards changing them as well, I'd wait for other opinions before making a decision.

dummdidumm:
Yes, this is what I meant, although it's probably better to add this to the `goto` method on line ~1152 (`goto: (href, opts = ..`) since that's the public interface

packages/kit/src/runtime/client/client.js
```diff
@@ -1149,7 +1149,21 @@ export function create_client({ target, base, trailing_slash }) {
...
 
-		goto: (href, opts = {}) => goto(href, opts, []),
+		goto: (href, opts = {}) => {
+			if (opts.hasOwnProperty('keepfocus')) {
+				throw new Error(
+					'`keepfocus` has been renamed to `keepFocus` (note the difference in casing)'
+				);
+			}
+
+			if (opts.hasOwnProperty('noscroll')) {
```
Rich-Harris, line 1159: `property in opts` is equivalent for all intents and purposes but is more byte-efficient

```suggestion
			// TODO remove for 1.0
			if ('keepfocus' in opts) {
				throw new Error(
					'`keepfocus` has been renamed to `keepFocus` (note the difference in casing)'
				);
			}

			if ('noscroll' in opts) {
```

Rich-Harris:
thank you!

PR #7487, opened 2022-11-03
[fix] check for private var usage in client code instead of server code

benmccann:
Fixes https://github.com/sveltejs/kit/pull/7288

Don't put the module into the graph with `ssrLoadModule` which will add the server-side code, but instead use `ensureEntryFromUrl` leaving the `ssr` parameter `undefined`

### DIFF
--- a/packages/kit/src/exports/vite/dev/index.js
+++ b/packages/kit/src/exports/vite/dev/index.js
@@ -117,6 +117,7 @@ export async function dev(vite, vite_config, svelte_config) {
 								extensions
 							);
 
+							// wait to load the module until after we check for illegal imports
 							result.shared = await vite.ssrLoadModule(url);
 						}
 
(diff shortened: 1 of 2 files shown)


bluwy (approved):
LGTM. Make sense to move `ssrLoadModule` only after checking the illegal imports.

dummdidumm:
Could we add some comments explaining why we use this other method, and use `resolveSSRModule` only after checking the graph? Someone without context could later come back to this code and think "oh let's move that duplicated code into that resolve function"

benmccann:
ok. I added a couple comments

packages/kit/src/exports/vite/dev/index.js
```diff
@@ -100,27 +100,29 @@ export async function dev(vite, vite_config, svelte_config) {
...
 
-							result.shared = module;
-
 							prevent_illegal_vite_imports(
 								module_node,
 								normalizePath(svelte_config.kit.files.lib),
 								extensions
 							);
+
+							result.shared = await vite.ssrLoadModule(url);
```
dummdidumm, line 120:
```suggestion
							// wait to load the module until after we check for illegal imports
							result.shared = await vite.ssrLoadModule(url);
```

dummdidumm (approved):

PR #7489, opened 2022-11-03
Deprecate svelte field in package.json

benmccann:
The `svelte` field was originally intended to allow package authors to be able to provide uncompiled `.svelte` files as well as a compiled `.js` version. We have decided if packages would like to provide a compiled version that a better way to do this is to create sub-package such as `compiled`.

The way the `svelte` field is currently implemented in `vite-plugin-svelte` breaks Vite's dependency optimization and we've decided removing support for the `svelte` field is the best path forward


dummdidumm:
If we do that we need to make sure the rollup and webpack plugins still work correctly. How can we ensure that?

benmccann:
The `svelte` field is optional in both plugins today

Here's a test showing that the Rollup plugin works without the `svelte` field: https://github.com/sveltejs/rollup-plugin-svelte/blob/master/test/node_modules/esm-no-pkg-export/package.json

Here's some documentation showing which fields `svelte-loader` resolves in order: https://github.com/sveltejs/svelte-loader#usage

`vite-plugin-svelte` does more complicated stuff with the field, but I believe that functionality is just an optimization and can be removed. I sent https://github.com/sveltejs/vite-plugin-svelte/pull/483 for that

dominikg:
related https://github.com/sveltejs/vite-plugin-svelte/pull/482

That fixes the situation mostly with vps, but i'm still testing it. Perf can be a problem and also what happens with ssr.optimizeDeps.

dominikg:
> The `svelte` field is optional in both plugins today
> 
> Here's a test showing that the Rollup plugin works without the `svelte` field: https://github.com/sveltejs/rollup-plugin-svelte/blob/master/test/node_modules/esm-no-pkg-export/package.json

there are no .svelte files in that test directory, sure that's testing svelte resolve? i think it may just be fore the package.json export warning.

benmccann:
>there are no .svelte files in that test directory, sure that's testing svelte resolve?

You're right. It wasn't testing it. However, it does work. I just made a sample Rollup project using a node module with the following `package.json` to confirm:
```
{
	"main": "./src/Component.svelte"
}
```

PR #7494, opened 2022-11-04
[breaking] use devalue to (de)serialize action data

### LINKED ISSUE #7488: Use devalue with form actions
### Describe the problem

Some objects, like Date, can be persisted on the client using the `load()` in a `+page.server.js`. The same is not true of Actions in the same file. 

This can be quite confusing. Because they are in the same file, even though load and Actions are not the same, many people consider them to work basically the same because of how similar they are. Not to mention how really similar they were before form actions.

It is also quite inconvenient for the developer. You can easily imagine a situation where in the load function you are sending the client a list of objects. One of the keys in the object is a Date. Let's say `lastModified`. Now, on the same page you fill out a form that creates a new record. You want to use progressive enhancement, so you create an enhance function where you are going to receive the new record, so you can display it alongside the existing ones. But wait, the returned record is not quite the same as the records you got from the initial load because you did not receive a Date. Instead you got some sort of string representing it. Now you have to deal with parsing it. Or, you go back to the load function and make it also just give a string, so you don't have to worry about the different behavior. 

Both solutions make some sense, but now that you have two ways of doing something, both are probably going to be used throughout the code base. Making it harder to maintain overall.

Use devalue to persist some objects from the server to the client when using Actions in the same way that is already done using `load()`

### Alternatives considered

Leave as is and make devs deal with the poor dx. 🤷🏻 

### Importance

would make my life easier

### Additional Information

Example: https://github.com/CalebBassham/repro-sk-actions-devalue/blob/master/src/routes/%2Bpage.server.js

dummdidumm:
This is a breaking change for everyone who fetches the actions directly and doesn't go through `use:enhance` (so hopefully close to noone).

```diff
+ import { deserialize } from '$app/forms';

const response = await fetch(action, {
	method: 'POST',
	headers: {
		accept: 'application/json',
		'x-sveltekit-action': 'true'
	},
	cache: 'no-store',
	body: data,
});

- result = await response.json();
+ result = deserialize(await response.text());
```

Allows for Date objects and more be part of the response
Closes #7488

Draft because we first need to decide if we really want this, and if we do, I need to update the types. The public type should stay the same, but we need to adjust the usages and accomodate for the fact that `data` is a `string` until parsed.

### Tests

### Changesets


Conduitry:
Should SK expose devalue somewhere (or maybe just its parser function)? That would make sure people wouldn't be on a different version, duplicating it in their project. It would mean we could have changes in the serialization format without it being a breaking change. And it would just be one fewer dependency for people to install if they were doing this.

That said, I don't have particularly strong feelings about whether we want to be switching to devalue in the first place.

Rich-Harris:
If we go down this road then yes, definitely — probably via `$app/forms` since that's the only place you'd need this. We could just re-export the `unflatten` function from `devalue` (perhaps with a different name if we don't like that one?)

dummdidumm:
Thinking ahead, with the current system in place we can adjust the code so that it's possible for people to provide their own (de)serializer, therefore an import location that is more generic would be better. If we agree on adding something like this in the future (or reserving the right to do so), then we probably need `$app/something-new` because it doesn't fit in any of the existing ones.
- `$app/serde`
- `$app/data`
- ?

CalebBassham:
> provide their own (de)serializer

Do you mean make it so that devalue can be replaced entirely by the user or that the user can add their own (de)serializers to devalue from sk. Either way, I can see uses for, but I don't think either are super necessary in the short-term.

Rich-Harris:
I'd vote for `$app/data` over `$app/serde`, since a lot of people aren't familiar with that term, and `data` is more generic and therefore future-proof.

> Do you mean make it so that devalue can be replaced entirely by the user or that the user can add their own (de)serializers to devalue from sk

Would probably mean adding custom (de)serialisation to devalue. Need to establish whether there's sufficient demand first

dummdidumm:
I added `$app/data` - but I realized in the process that it's not that easy usage-wise. To deserialize the action response, you have to do `const result = await result.json()` and then `result.data = devalue.parse(result.data)` - similar things for the `load` response from the server. Furthermore, I'm now wondering where outside of form actions you would need to explicitly deserialize stuff - for `load`, that's all handled under the hood. This is not possible to encapsulate in a generic manner in a `(de)serialize` function. Also, when would you want to serialize something yourself? At this point you're doing custom stuff anyway.
So an export from `$app/forms` is probably better after all? -> switching the implementation to do that instead (we can revert that commit if we want a more general location).

packages/kit/src/runtime/app/forms.js
```diff
@@ -91,7 +101,7 @@ export function enhance(form, submit = () => {}) {
 				signal: controller.signal
 			});
 
-			result = await response.json();
+			result = deserialize(await response.text());
```
dummdidumm, line 104: We could also make it `deserialize(await response.json())` but I somehow felt that passing in the raw text is better.

Rich-Harris (approved):

Rich-Harris:
[the hell is windows doing](https://github.com/sveltejs/kit/actions/runs/3465824504/jobs/5790166635)

image

dummdidumm:
seems this worked, but now there's an error with the sharp package, maybe due to the bumping of the patch dependencies..? Redeploying without cache works, so it's likely a missing postinstall script running or sth

PR #7507, opened 2022-11-04
prevent loading of illegal modules in the browser

### LINKED ISSUE #6700: Support for tools like telefunc

Rich-Harris:
Closes #6700. Closes #7288. Right now, we prevent people from importing illegal modules (`$env/[type]/private`, `$lib/server/**` and `**/*.server.{ext}`) by analysing the module graph during SSR. This is a flawed approach for a variety of reasons:

* It doesn't account for systems like telefunc, which transform modules differently between SSR and non-SSR modes such that the dangerous `import` declarations are removed
* it means you only get the error when you hit a page directly, not when you soft-navigate to a page. (I was surprised to realise that this morning!)
* it applies even to pages that have `csr = false`

A much simpler approach is to simply prevent the modules in question from being imported in the browser at dev time. We can do this by adding a guard in a `load` hook that throws if the relevant conditions are met. (At build time we keep the existing graph analysis.)

There's a couple of minor drawbacks: we don't get the import chain, just the module ID, and the error overlay is dismissable. I think these trade-offs are well worth it.

A nice bonus is that we can get rid of the code for traversing the Vite module graph. Since we now only need to worry about the Rollup module graph, this also means we can get rid of a layer of indirection (sorry @tcc-sejohnson, it was good code! but no longer needed). That part is a WIP.

### Tests

### Changesets


tcc-sejohnson:
As long as we're retaining the same graph analysis in build as we were before I see no problem with this -- I wanted to do this when we originally implemented the feature, but just couldn't figure out how. Please ping me when this is ready-er for merge -- happy to review 😁

Rich-Harris:
Ok I think this is ready now, marking it as such and crossing my fingers that the tests pass. One wrinkle: if the Vite error overlay appears before hydration, it will get nuked if `%sveltekit.body%` isn't wrapped in an element below the `<body>`, so I had to make that change to the test app.

I almost wonder if we should enforce that, since it's a bit of a footgun otherwise.

@tcc-sejohnson would be grateful if you do get a chance to review! The build stuff should be basically identical

packages/kit/src/exports/vite/graph_analysis/index.spec.js
```diff
@@ -1,191 +0,0 @@
-import { describe } from '../../../utils/unit_test.js';
```
tcc-sejohnson, line 1: I was excited about being able to simplify the graph analysis, but this kinda killed all of my excitement. It seems like a step backwards -- we're removing our tests for making sure that, given an illegally-named module, we're going to throw with our graph analysis. Getting the same level of coverage with integration tests is going to be tricky, and it's also going to make it harder in the future -- if someone discovers an edge case we're missing, we suddenly have to find some contrived way to test it through integration tests rather than just adding an illegal chain here.

packages/kit/src/exports/vite/graph_analysis/index.js
```diff
@@ -1,277 +1,81 @@
...
-		this.#seen.add(this.id);
-		for (const child of this.#module_info.importedModules) {
-			if (!this.#module_types.has(path.extname(this.id))) {
-				continue;
-			}
-			yield ViteImportGraph.#new_internal(this.#module_types, child, this.#seen);
+	return {
+		/** @param {string} id */
+		check: (id) => {
+			follow(id, []);
```
tcc-sejohnson, line 70: default is already empty array

```suggestion
			follow(id);
```

tcc-sejohnson (changes requested):

tcc-sejohnson:
> One wrinkle: if the Vite error overlay appears before hydration, it will get nuked if %sveltekit.body% isn't wrapped in an element below the <body>, so I had to make that change to the test app.

Anecdotally, I've seen a good number of people in the Discord that remove the `div` wrapper for "easier" styling, so... might be a good idea to try to find a way.

packages/kit/src/exports/vite/graph_analysis/index.spec.js
Rich-Harris: yeah, this was me being lazy — i couldn't really reuse much of the existing test code, so I cheated and deleted the file. reinstated tests, and added checks that the messages are being generated correctly

Rich-Harris:
alright, things finally seem to be working again. @tcc-sejohnson if you have the bandwidth to take another look since you requested changes, that would be great — looking forward to getting this one merged 🙏

lukaszpolowczyk:
> Anecdotally, I've seen a good number of people in the Discord that remove the div wrapper for "easier" styling, so... might be a good idea to try to find a way.

@tcc-sejohnson I in one project added to this `div` a `style="display: contents;"`, and it was enough for me.

I don't know if that can cause a problem, but I haven't noticed any.

packages/kit/src/exports/vite/graph_analysis/index.js
```diff
@@ -1,277 +1,107 @@
...
-	 * @returns {boolean}
+	 * @param {string} id
+	 * @param {Array<{ id: string, dynamic: boolean }>} chain
 	 */
-	#is_kit_illegal(module_id) {
-		return ILLEGAL_IMPORTS.has(module_id);
-	}
+	function follow(id, chain) {
+		if (seen.has(id)) return;
+		seen.add(id);
```
tcc-sejohnson, line 44: nothing is wrong here; I just wanted you to see what the heck GitHub thinks these lines correlate to in the original diff

tcc-sejohnson (approved):
LGTM! This should be a great improvement for the dev-time experience. Thank you!

tcc-sejohnson:
> > Anecdotally, I've seen a good number of people in the Discord that remove the div wrapper for "easier" styling, so... might be a good idea to try to find a way.
> 
> @tcc-sejohnson I in one project added to this `div` a `style="display: contents;"`, and it was enough for me.
> 
> I don't know if that can cause a problem, but I haven't noticed any.

That won't work for everyone. To be clear, I don't think there's really any logical reason to remove the wrapper div -- it's easy enough to set it up however you want to -- but some people do it and we don't want to break their apps if possible 😁

stephane-vanraes:
> That won't work for everyone. To be clear, I don't think there's really any logical reason to remove the wrapper div -- it's easy enough to set it up however you want to -- but some people do it and we don't want to break their apps if possible 😁

I am one of those removing the wrapper div, so maybe shedding some insights on that 🙂

The simple reason is that this div servers no purpose whatsoever in the _final product_, and it makes styling harder to a degree (you always have to remember that the body of your page is not `body` for the css).

As far as I understand this div is a crutch for shortcoming of a tool used in _development_ only and imho that is enough reason to try to find a way to make things work without, because utilities meant for development have no place in a production build.

Rich-Harris:
> besides potentially stopping hydration from removing extensions there injected markup

this is the key — we've had multiple reports of exactly this sort of thing happening. it's very much _not_ about preventing dev-time issues — it's the sort of bug that's likely to manifest in deeply confusing ways in production and that you'll be lucky to even find out about. we added a note to the docs...

image

...but i'm not sure that's enough, because people will always look at it and assume they can safely remove it. my personal inclination is to require `%sveltekit.body%` to be inside an element other than `<body>`, but i'm going to open a discussion to see how large a trade-off that would be

Rich-Harris:
https://github.com/sveltejs/kit/discussions/7585

PR #7511, opened 2022-11-05
Warn of cookie paths only when applicable

### LINKED ISSUE #7510: `cookies.get` incorrectly logs a warning when a cookie was removed from the browser instead of `cookies.delete`
### Describe the bug

Context provided by @dummdidumm in Discord:
> Some context about this warning and how the logic is to surface it:
Why this warning: Cookies have a path. This means they are only available at this path and below. So a path of /foo means your cookie is available at /foo, /foo/bar etc, but not / or /sthelse. Not setting the path value means the cookie's path is set to where the cookie is set on. This is a source of confusion ("why is my cookie not deleted/set/available?") so we added that warning. That said, that warning could probably use some love, at least a link to some more in-depth explanation what this is about.
> 
> Logic to surface it: Whenever you call cookies.set in dev mode, that cookie's name and its path are stored in a map, and whenever you call cookies.get and it turns up nothing, it prints a warning if a cookie with that name exists in that map. False positives could occur if you remove a cookie through another method - and writing this I just realized that session cookies are probably also false positives: If you set a cookie, close the browser tab and open it again, a session cookie is removed from the browser, but not from the map. Maybe that's something that happens here?

It seems that these proposed false positives are correct, when a cookie is removed via exterior means but still expected to be there via `cookies.get` this warning is thrown. However, the warning says "`did you mean to set its 'path' to '/'?`" even when the path is set correctly. 

I propose that this warning is only logged when the path isn't `/`, and print the path(s) associated with the cookie. If it is already set to `/`, then these cookies should be removed from the map to align with the lack of them in the browser. I'll attempt to submit a PR to fix this myself shortly, but I'm not super familiar with the inner workings of the project, so it might not be perfect.

### Reproduction

A simple reproduction can be found in [CookiesGetWarningRepro](https://github.com/ptlthg/CookiesGetWarningRepro), but it's a simple thing to check. Just remove a cookie from your browser in dev tools, then refresh or try to access it via `cookies.get` afterwards. The warning should then be logged.

### Logs

```Shell
Cookie with name 'session-cookie' was not found, but a cookie with that name exists at a sub path. Did you mean to set its 'path' to '/'?
Cookie with name 'normal-cookie' was not found, but a cookie with that name exists at a sub path. Did you mean to set its 'path' to '/'? 
Cookie with name 'expires-cookie' was not found, but a cookie with that name exists at a sub path. Did you mean to set its 'path' to '/'?
```

### System Info

```Shell
System:
    OS: Windows 10 10.0.19044
    CPU: (16) x64 AMD Ryzen 7 5800H with Radeon Graphics
    Memory: 45.67 GB / 59.86 GB
  Binaries:
    Node: 18.8.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.18.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (107.0.1418.35)
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.535
    svelte: ^3.44.0 => 3.52.0
    vite: ^3.1.0 => 3.2.2
```

### Severity

annoyance

### Additional Information

_No response_

ptlthg:
Fixes #7510 

Replaces the dev mode cookie warning with a more detailed one that doesn't get printed when a cookie was removed on the browser.

Essentially, only prints the warning when the cookie paths didn't include the request path.

### Tests

### Changesets

### DIFF
--- a/.changeset/green-foxes-wait.md
+++ b/.changeset/green-foxes-wait.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Prevents cookie not found warning when cookie was removed in browser.
--- a/.changeset/green-foxes-wait.md
+++ b/.changeset/green-foxes-wait.md
@@ -1,5 +1,0 @@
----
-'@sveltejs/kit': patch
----
-
-Prevents cookie not found warning when cookie was removed in browser.
--- a/packages/kit/src/runtime/server/cookie.js
+++ b/packages/kit/src/runtime/server/cookie.js
@@ -58,10 +58,10 @@ export function get_cookies(request, url, options) {
 			}
 
 			// Warn about cookies that are set with other paths in dev mode
-			if (cookie_paths[name]?.has(url.pathname)) {
+			if (cookie_paths[name]?.has('/')) {
 				// Cookie was set on this path, but removed elsewhere
 				// Reflect these changes and remove the path from the set
-				cookie_paths[name].delete(url.pathname);
+				cookie_paths[name].delete('/');
 			} else if (cookie_paths[name]?.size > 0) {
 				// Warn about cookie existence on other paths
 				const paths = `'${Array.from(cookie_paths[name]).join("', '")}'`;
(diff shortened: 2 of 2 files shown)


dummdidumm:
Thank you, but as pointed out the issue, the fix is incomplete. We should check the incoming cookies instead - I'll open a PR with that solution.

PR #7516, opened 2022-11-05
[fix] don't clean build and output directories when using `--watch`

samuelabreu:
When watching, static directory is missing on outDir because the vite svelte-kit plugin clean them

How to reproduce:

Create a clean svelte-kit project (choose skeleton and no to all questions)
- npm create svelte@latest watch_bug
- cd watch_bug
- npm install
- touch static/file.txt
- npx vite build -w
- Check files on .svelte-kit/output/client/, is missing file.txt and favicon.png

Couldn't devote too much time to do a better solution, so my PR only avoid cleaning directories when using --watch

### Tests

### Changesets

### DIFF
--- a/.changeset/large-starfishes-relate.md
+++ b/.changeset/large-starfishes-relate.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': minor
 ---
 
-Avoid cleaning build_dir and output_dir when using --watch
+[fix] don't clean build and output directories when using `--watch`
--- a/.changeset/large-starfishes-relate.md
+++ b/.changeset/large-starfishes-relate.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': minor
+---
+
+Avoid cleaning build_dir and output_dir when using --watch
--- a/packages/kit/src/exports/vite/index.js
+++ b/packages/kit/src/exports/vite/index.js
@@ -330,10 +330,10 @@ function kit() {
 			completed_build = false;
 
 			if (is_build) {
-        if (!vite_config.build.watch) {
-				  rimraf(paths.build_dir);
-				  rimraf(paths.output_dir);
-        }
+					if (!vite_config.build.watch) {
+						rimraf(paths.build_dir);
+						rimraf(paths.output_dir);
+					}
 				mkdirp(paths.build_dir);
 				mkdirp(paths.output_dir);
 			}
--- a/packages/kit/src/exports/vite/index.js
+++ b/packages/kit/src/exports/vite/index.js
@@ -330,10 +330,11 @@ function kit() {
 			completed_build = false;
 
 			if (is_build) {
-				rimraf(paths.build_dir);
+        if (!vite_config.build.watch) {
+				  rimraf(paths.build_dir);
+				  rimraf(paths.output_dir);
+        }
 				mkdirp(paths.build_dir);
-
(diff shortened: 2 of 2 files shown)


.changeset/large-starfishes-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': minor
+---
+
+Avoid cleaning build_dir and output_dir when using --watch
```
benmccann, line 5:
```suggestion
[fix] don't clean build and output directories when using `--watch`
```

benmccann (approved):

packages/kit/src/exports/vite/index.js
```diff
@@ -330,10 +330,11 @@ function kit() {
 			completed_build = false;
 
 			if (is_build) {
-				rimraf(paths.build_dir);
+        if (!vite_config.build.watch) {
+				  rimraf(paths.build_dir);
+				  rimraf(paths.output_dir);
+        }
```
benmccann, line 336: lint is failing. you need to use tabs instead of spaces
```suggestion
					if (!vite_config.build.watch) {
						rimraf(paths.build_dir);
						rimraf(paths.output_dir);
					}
```

packages/kit/src/exports/vite/index.js
```diff
@@ -330,10 +330,11 @@ function kit() {
 			completed_build = false;
 
 			if (is_build) {
-				rimraf(paths.build_dir);
+					if (!vite_config.build.watch) {
+						rimraf(paths.build_dir);
+						rimraf(paths.output_dir);
+					}
```
benmccann, line 336:
```suggestion
				if (!vite_config.build.watch) {
					rimraf(paths.build_dir);
					rimraf(paths.output_dir);
				}
```

dummdidumm:
I'm not sure if this works for all cases. Does Vite remove files that are obsolete? Especially with regards to prerendering, I'm imagining this leaving old files lying around, so we should probably at least delete those folder related to that.

samuelabreu:
I guess you are right, on this pr files could stay between builds. It fits my needs, so for now im using my fork. My use case is building an app using svelte+neutralinojs

benmccann:
> Does Vite remove files that are obsolete?

It does: https://vitejs.dev/config/build-options.html#build-emptyoutdir. However, that's probably not enough because Vite writes to `output/client` and what we're trying to clear here are `build` and `output`.

dummdidumm:
Given that it didn't work at all before, and may work for some more cases now, makes me inclined to merge this, even if it's somewhat incomplete and some gotchas could arise.
Rich probably has the inverse stance on this - better clearly broken than working with some gotchas.

samuelabreu:
Maybe another solution is use this code instead of skip clean those dirs:
`
import { mkdirp, posixify, resolve_entry, rimraf, copy } from '../../utils/filesystem.js';
...        
if (vite_config.build.watch) {
          copy(vite_config.publicDir, `${paths.output_dir}/client`);
}
`

Rich-Harris:
> Rich probably has the inverse stance on this - better clearly broken than working with some gotchas.

I think my stance in this case is that since it doesn't look like we're introducing any new bugs, there's no reason to let perfect be the enemy of the good. If there's a more complete fix needed in future we can cross that bridge when we come to it. Thanks @samuelabreu!

PR #7525, opened 2022-11-06
[docs] delete a sentence with the same meaning

RyanZhiNie:
The deleted sentence had the same meaning as its latter sentence, so I kept the latter one.

### Tests

### Changesets

### DIFF
--- a/documentation/docs/20-core-concepts/40-page-options.md
+++ b/documentation/docs/20-core-concepts/40-page-options.md
@@ -83,14 +83,7 @@ Since these routes cannot be dynamically server-rendered, this will cause errors
 
 ### ssr
 
-Normally, SvelteKit renders your page on the server first and sends that HTML to the client where it's hydrated. If you set `ssr` to `false`, it renders an empty 'shell' page instead. This is useful if your page is unable to be rendered on the server (because you use browser-only globals like `document` for example), but in most situations it's not recommended ([see appendix](/docs/glossary#ssr)). If you put this into your root `+layout.js`, you create an SPA (all pages are client-side rendered only).
-
-```js
-/// file: +page.js
-export const ssr = false;
-```
-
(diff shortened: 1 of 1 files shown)


documentation/docs/20-core-concepts/40-page-options.md
```diff
@@ -83,14 +83,7 @@ Since these routes cannot be dynamically server-rendered, this will cause errors
...
-
-```js
-/// file: +page.js
-export const ssr = false;
-```
-
-If you add `export const ssr = false` to your root `+layout.js`, your entire app will only be rendered on the client — which essentially means you turn your app into an SPA.
+Normally, SvelteKit renders your page on the server first and sends that HTML to the client where it's hydrated. If you set `ssr` to `false`, it renders an empty 'shell' page instead. This is useful if your page is unable to be rendered on the server (because you use browser-only globals like `document` for example), but in most situations it's not recommended ([see appendix](/docs/glossary#ssr)). If you add `export const ssr = false` to your root `+layout.js`, your entire app will only be rendered on the client — which essentially means you turn your app into an SPA.
```
dummdidumm, line 86:
```suggestion
Normally, SvelteKit renders your page on the server first and sends that HTML to the client where it's hydrated. If you set `ssr` to `false`, it renders an empty 'shell' page instead. This is useful if your page is unable to be rendered on the server (because you use browser-only globals like `document` for example), but in most situations it's not recommended ([see appendix](/docs/glossary#ssr)). If you add `export const ssr = false` to your root `+layout.js`, your entire app will only be rendered on the client — which essentially means you turn your app into an SPA.

```js
/// file: +page.js
export const ssr = false;
```
```

dummdidumm (approved):
Thank you! I adjusted the PR slightly, which means it ends up with just that one sentence deleted.

PR #7531, opened 2022-11-07
[fix] take into account Request input when serializing fetch data

### LINKED ISSUE #7514: Using fetch from PageEvent results in the wrong content being cached in certain circumstances
### Describe the bug

When fetching from f.ex a graphql endpoint, most of the time the URL is the same but the body (query) is different. This causes the fetch provided by the PageLoad event to use the wrong cached data during hydration.

### Reproduction

Here's an example and description of the problem.
https://github.com/tedsteen/reproducables/pull/1

### Logs

_No response_

### System Info

```Shell
System:
    OS: macOS 13.0
    CPU: (10) arm64 Apple M1 Pro
    Memory: 1.08 GB / 32.00 GB
    Shell: 5.9 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 18.11.0 - /opt/homebrew/bin/node
    npm: 8.19.2 - /opt/homebrew/bin/npm
  Browsers:
    Brave Browser: 107.1.45.116
    Safari: 16.1
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.535
    svelte: ^3.44.0 => 3.52.0
    vite: ^3.1.0 => 3.2.2
```

### Severity

### Additional Information

I fixed this by creating a hash of the body sent to fetch and add it as a query parameter. Not sure if I'm missing something or this is a viable way to fix this.

dummdidumm:
Fixes #7514

I'm not sure if this is the best approach. I'm a little nervous because cloning the whole Request body might become costly, and since we can't read the body only when absolutely needed (because the ReadableStream is already consumed by then) we have to do it unconditionally. I would hope though that people calling fetch like this are in the minority.

A different solution would be to somehow give the request another kind of ID, like a call index, but that could be messed up if the user does something with `if (browser) fetch(..)`.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/load_data.js
+++ b/packages/kit/src/runtime/server/page/load_data.js
@@ -103,6 +103,7 @@ export async function load_data({
 		data: server_data_node?.data ?? null,
 		route: event.route,
 		fetch: async (input, init) => {
+			const cloned_body = input instanceof Request && input.body ? input.clone().body : null;
 			const response = await event.fetch(input, init);
 
 			const url = new URL(input instanceof Request ? input.url : input, event.url);
@@ -149,7 +150,11 @@ export async function load_data({
 							fetched.push({
 								url: same_origin ? url.href.slice(event.url.origin.length) : url.href,
 								method: event.request.method,
-								request_body: /** @type {string | ArrayBufferView | undefined} */ (init?.body),
+								request_body: /** @type {string | ArrayBufferView | undefined} */ (
+									input instanceof Request && cloned_body
+										? await streamToString(cloned_body)
+										: init?.body
+								),
 								response_body: body,
 								response: response
 							});
@@ -232,4 +237,21 @@ export async function load_data({
 	const data = await node.shared.load.call(null, load_event);
 
 	return data ? unwrap_promises(data) : null;
+}
+
+/**
+ * @param {ReadableStream<Uint8Array>} stream
+ */
+async function streamToString(stream) {
+	let result = '';
+	const reader = stream.getReader();
+	const decoder = new TextDecoder();
+	while (true) {
+		const { done, value } = await reader.read();
+		if (done) {
+			break;
+		}
+		result += decoder.decode(value);
+	}
+	return result;
 }
--- a/.changeset/strange-walls-carry.md
+++ b/.changeset/strange-walls-carry.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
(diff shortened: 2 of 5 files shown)


Rich-Harris:
I see two possibilities:

1. Clone the body in the `fetch(request)` case (as is currently the case in this PR)
2. Bail in the `fetch(request)` case where `request.method !== 'GET'` (i.e. the request is repeated for real in the client)

Option 2 is 'safer', in that it doesn't matter how big the request body is. But in reality no-one is going to be making requests with huge bodies in the context of a `load` function, because GraphQL is the only realistic use case for POST requests in the first place. So I think my vote goes for option 1, despite it feeling somewhat unclean

(Aside: the fact that GraphQL queries can be large enough that we're forced to use POST requests in the first place is really just a symptom of how daft the whole thing is.)

packages/kit/src/runtime/server/page/load_data.js
```diff
@@ -149,7 +150,11 @@ export async function load_data({
 							fetched.push({
 								url: same_origin ? url.href.slice(event.url.origin.length) : url.href,
 								method: event.request.method,
-								request_body: /** @type {string | ArrayBufferView | undefined} */ (init?.body),
+								request_body: /** @type {string | ArrayBufferView | undefined} */ (
+									input instanceof Request && cloned_body
+										? await streamToString(cloned_body)
+										: init?.body
```
Rich-Harris, line 156: one thought — if `request_body` could be a `ReadableStream`...

```suggestion
									init?.body ?? input.clone?.().body
```

...then later, in `render.js`, we could do something like this:

```diff
	if (page_config.ssr && page_config.csr) {
+		for (const request of fetched) {
+			if (fetched.request_body instanceof ReadableStream) {
+				fetched.request_body = await to_string(fetched.request_body);
+			}
+		}
		body += `\n\t${fetched
			.map((item) =>
				serialize_data(item, resolve_opts.filterSerializedResponseHeaders, !!state.prerendering)
			)
			.join('\n\t')}`;
	}
```

Though it may actually be cheaper to keep the entire request clone and then do

```js
fetched.request_body = await fetched.clone.arrayBuffer();
```

than to implement our own stream to string function

dummdidumm: That doesn't work, because cloning the request at that point is too late, the body's `ReadableStream` is already consumed by then and you can't get the value.

Rich-Harris: oh, of course 😞

Rich-Harris (approved):

PR #7537, opened 2022-11-07
Upgrade to Vite 4

benmccann:
Just for testing at the moment. Still quite a ways from a final release


Rich-Harris:
looks like the green in the diff is mostly caused by multiple esbuild versions

PR #7541, opened 2022-11-07
remove nodes for prerendered routes from manifest

### LINKED ISSUE #6568: prerender = true in /+layout.server.ts breaks the route after deployment to netlify / vercel

Rich-Harris:
closes #6568

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/generate_manifest/index.js
+++ b/packages/kit/src/core/generate_manifest/index.js
@@ -40,6 +40,16 @@ export function generate_manifest({ build_data, relative_path, routes, format = 
 
 	const matchers = new Set();
 
+	const used_nodes = new Set();
+
+	for (const route of routes) {
+		if (route.page) {
+			for (const i of route.page.layouts) used_nodes.add(i);
+			for (const i of route.page.errors) used_nodes.add(i);
+			used_nodes.add(route.page.leaf);
+		}
+	}
+
 	// prettier-ignore
 	// String representation of
 	/** @type {import('types').SSRManifest} */
@@ -51,7 +61,7 @@ export function generate_manifest({ build_data, relative_path, routes, format = 
 		_: {
 			entry: ${s(build_data.client.entry)},
 			nodes: [
-				${Array.from(bundled_nodes.values()).map(node => loader(node.path)).join(',\n\t\t\t\t')}
+				${Array.from(bundled_nodes.values()).filter(node => used_nodes.has(node.index)).map(node => loader(node.path)).join(',\n\t\t\t\t')}
 			],
 			routes: [
 				${routes.map(route => {
--- a/packages/kit/src/exports/vite/index.js
+++ b/packages/kit/src/exports/vite/index.js
@@ -408,7 +408,7 @@ function kit() {
 					server
 				};
 
-				const manifest_path = `${paths.output_dir}/server/manifest.js`;
+				const manifest_path = `${paths.output_dir}/server/manifest-full.js`;
 				fs.writeFileSync(
 					manifest_path,
 					`export const manifest = ${generate_manifest({
@@ -431,6 +431,7 @@ function kit() {
 						script,
 						[
 							vite_config.build.outDir,
+							manifest_path,
 							results_path,
 							'' + verbose,
 							JSON.stringify({ ...env.private, ...env.public })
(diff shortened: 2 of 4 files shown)


Rich-Harris:
Noteworthy: in addition to preventing `vite preview` from rendering routes that should have been prerendered, this will reduce the size of serverless deployments in many cases

packages/kit/src/core/generate_manifest/index.js
```diff
@@ -51,7 +61,7 @@ export function generate_manifest({ build_data, relative_path, routes, format =
 		_: {
 			entry: ${s(build_data.client.entry)},
 			nodes: [
-				${Array.from(bundled_nodes.values()).map(node => loader(node.path)).join(',\n\t\t\t\t')}
+				${Array.from(bundled_nodes.values()).filter(node => used_nodes.has(node.index)).map(node => loader(node.path)).join(',\n\t\t\t\t')}
```
dummdidumm, line 64: `bundled_nodes` is only used here, so instead of filtering afterwards, I would switch it around and do the `used_nodes` first and while constructing `bundled_nodes` use that. Also a little comment on the `if`-block why we do this, because from reading it might look duplicative ("isn't nodes and routes the same?")

packages/kit/src/exports/vite/index.js
```diff
@@ -459,6 +460,16 @@ function kit() {
 					});
 				});
 
+				// generate a new manifest that doesn't include prerendered pages
+				fs.writeFileSync(
+					`${paths.output_dir}/server/manifest.js`,
+					`export const manifest = ${generate_manifest({
+						build_data,
+						relative_path: '.',
+						routes: manifest_data.routes.filter((route) => prerender_map.get(route.id) !== true)
```
dummdidumm, line 469: Does `prerender_map` leave out routes which have `export const prerender = 'auto';` (e.g. the "some blog posts static, the long tail dynamic" case)?

packages/kit/src/core/generate_manifest/index.js
Rich-Harris: yeah, it's broken, am in the process of fixing

packages/kit/src/exports/vite/index.js
Rich-Harris: it's a map of every route ID to `true`, `false` or `'auto'`. this filter will keep `'auto'` routes

dummdidumm (approved):

PR #7543, opened 2022-11-07
Upgrade to Vite 4

benmccann:
Shouldn't affect SvelteKit or its users much other than bumping a version. One nice change is that it fixes the value of `NODE_ENV` to be more inline with what one would expect, which I think should help us clean some stuff up

To migrate, simply bump your Vite and SvelteKit versions to the latest versions. See the full Vite 4 [migration guide](https://vitejs.dev/guide/migration.html) and [announcement](https://vitejs.dev/blog/announcing-vite4.html) for more details


benmccann:
I sent a fix for the failing CI: https://github.com/vitejs/vite/pull/10901

.changeset/old-singers-cough.md
```diff
@@ -0,0 +1,9 @@
+---
+'@sveltejs/adapter-cloudflare-workers': patch
+'@sveltejs/adapter-netlify': patch
+'@sveltejs/adapter-node': patch
```
benmccann, line 4:
```suggestion
'@sveltejs/adapter-cloudflare': patch
'@sveltejs/adapter-cloudflare-workers': patch
'@sveltejs/adapter-netlify': patch
```

.changeset/smooth-doors-fly.md
```diff
@@ -0,0 +1,11 @@
+---
+'@sveltejs/adapter-cloudflare': patch
+'@sveltejs/adapter-cloudflare-workers': patch
+'@sveltejs/adapter-netlify': patch
+'@sveltejs/adapter-node': patch
+'@sveltejs/adapter-vercel': patch
```
benmccann, line 6:
```suggestion
```

.changeset/smooth-doors-fly.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sveltejs/kit': patch
+'create-svelte': patch
+---
+
+Upgrade to Vite 4
```
dummdidumm, line 6: isn't this also `[breaking]` because it's a peer dependency for existing projects?

dummdidumm (commented):
I think we also need a "how to migrate" section in the PR description so people know what to do

packages/kit/src/exports/vite/index.js
```diff
@@ -561,7 +561,7 @@ function kit() {
 
 function check_vite_version() {
 	// TODO parse from kit peer deps and maybe do a full semver compare if we ever require feature releases a min
-	const min_required_vite_major = 3;
+	const min_required_vite_major = 4;
```
dummdidumm, line 564: is this function a good place to link to this PR so people get to proper migration instructions faster?

.changeset/smooth-doors-fly.md
benmccann, line 6:
```suggestion
[breaking] upgrade to Vite 4
```

dominikg:
> I think we also need a "how to migrate" section in the PR description so people know what to do

i think in this case the how to migrate is self explanatory, just update the required peer dependencies, there is no code change or config change needed as far as i am aware

dummdidumm:
Oh, that's a pleasant surprise, nice!

.changeset/smooth-doors-fly.md
benmccann: done

packages/kit/src/exports/vite/index.js
benmccann: don't think it's necessary as all they need to do is bump the version number

packages/kit/test/apps/options/test/test.js
```diff
@@ -7,9 +7,11 @@ test.describe.configure({ mode: 'parallel' });
 
 test.describe('base path', () => {
 	test('serves a useful 404 when visiting unprefixed path', async ({ request }) => {
-		const response = await request.get('/');
+		const response = await request.get('/slash/', { headers: { Accept: 'text/html' } });
```
Rich-Harris, line 10: Why did this need to be changed? Is there anything different about how `/` works in this scenario now?

packages/kit/src/exports/vite/dev/index.js
```diff
@@ -248,44 +244,6 @@ export async function dev(vite, vite_config, svelte_config) {
 	}
 
 	const assets = svelte_config.kit.paths.assets ? SVELTE_KIT_ASSETS : svelte_config.kit.paths.base;
-	const asset_server = sirv(svelte_config.kit.files.assets, {
```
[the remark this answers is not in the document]
Rich-Harris: (actually never mind, this is tested in the `options` app so hopefully would have failed if it didn't)

packages/kit/test/apps/options/test/test.js
Rich-Harris: ah, okay — seems it redirects to `base` automatically in the `/` case

Rich-Harris (changes requested):
i think this broke `paths.assets` handling. investigating

packages/kit/src/exports/vite/dev/index.js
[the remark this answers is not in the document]
Rich-Harris: i lied. the tests didn't cover this

packages/kit/test/apps/options/test/test.js
benmccann: The dev server will automatically redirect you to the base path rather than giving the error message

packages/kit/src/exports/vite/preview/index.js
```diff
@@ -73,7 +73,11 @@ export async function preview(vite, vite_config, svelte_config) {
 				next();
 			} else {
 				res.statusCode = 404;
-				res.end(`Not found (did you mean ${base + pathname}?)`);
+				res.end(
+					`The server is configured with a public base URL of ${base} - did you mean to visit <a href="${
+						base + pathname
+					}">${base + pathname}</a> instead?`
+				);
```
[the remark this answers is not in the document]
Rich-Harris: tried it: we get a plain 404 response, which isn't as helpful as what we had before. The HTML response is nicer (though slightly misleading — it lops off the trailing slash) but it makes it harder to figure out why e.g. your `fetch` is failing.

I wonder if we should continue to handle this ourselves?

## before

```html
<!-- with accept: text/html -->
Not found (did you mean /path-base/nope?)

<!-- with accept: * -->
Not found (did you mean /path-base/nope?)
```

## after

```html
<!-- with accept: text/html -->
The server is configured with a public base URL of /path-base - did you mean to visit <a href="/path-base/nope">/path-base/nope</a> instead?

<!-- with accept: * -->
Not found
```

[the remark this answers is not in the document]
benmccann: I could check with Vite if they'd be open to adding that in a 4.0.1. I think either way, whether we want to do it ourselves or implement it in Vite, we could probably do it in a follow up

packages/kit/test/apps/options/test/test.js
benmccann: did you want to remove the `Accept` header here?

packages/kit/test/apps/options/test/test.js
```diff
@@ -10,7 +10,7 @@ test.describe('base path', () => {
 		const response = await request.get('/slash/', { headers: { Accept: 'text/html' } });
 		expect(response.status()).toBe(404);
 		expect(await response.text()).toBe(
-			'The server is configured with a public base URL of /path-base - did you mean to visit <a href="/path-base/slash/">/path-base/slash/</a> instead?'
+			'Not found (did you mean <a href="/path-base/slash/">/path-base/slash/</a>?)'
```
benmccann, line 13: I probably would have changed the message to align with Vite instead as it'd make it easier to eventually switch over to their handling of things

packages/kit/test/apps/options/test/test.js
benmccann: did you want to remove the `Accept` header here?
```suggestion
		const response = await request.get('/slash/');
```

packages/kit/test/apps/options/test/test.js
Rich-Harris: The language doesn't quite mesh with our usage though — if we wanted to be similarly verbose we'd probably have something like

> The app is configured with a `paths.base` of /path-base/ — did you mean to visit ...

so that it refers to the config value as exposed to Kit users, and includes the trailing slash. I think it's better to do our own thing here

packages/kit/test/apps/options/test/test.js
Rich-Harris: That would change the response, so no. Though I guess we could test both HTML and plain responses

Rich-Harris (approved):

Bandit:
I'm probably stupid, but I went into package.json and set

```
"@sveltejs/kit": "^1.0.0-next.582"
"vite": "^4.0.0"
```

And then `npm i` and get `Conflicting peer dependency: vite@4.0.0`. No idea why?

benmccann:
I haven't seen that. Can you share a project that reproduces the issue?

Bandit:
Just tried to recreate the issue in StackBlitz but it seems to work fine, so guessing it's something specific to this (private) repo.

I tried renaming `node_modules` before running `npm i` but it gives me basically the same error: 

```
npm ERR! Found: vite@3.2.5
npm ERR! node_modules/vite
npm ERR!   dev vite@"^3.0.4" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer vite@"^4.0.0" from @sveltejs/kit@1.0.0-next.582
npm ERR! node_modules/@sveltejs/kit
npm ERR!   dev @sveltejs/kit@"^1.0.0-next.582" from the root project
```

I don't really know where to even start debugging the issue since I don't know a lot about NPM's inner workings. I could use `--force` or `--legacy-peer-deps` but that sounds scary

benmccann:
>npm ERR! Found: vite@3.2.5
npm ERR! node_modules/vite
npm ERR!   dev vite@"^3.0.4" from the root project

That sounds like you still have Vite 3 in a `package.json` somewhere.

Another thing you could try is `rm -rf package-lock.json node_modules && npm install`

Bandit:
That fixed it. Really appreciate the help - realise you're not tech support! Not sure what was wrong with my setup 😅

PR #7549, opened 2022-11-08
[docs] clarify greedy matching behavior

### LINKED ISSUE #7548: Optional route with matcher followed by rest parameters
### Describe the bug

Given a `[[lang=lang]]/[...path]` route, kit will fail to catch non-matched URLs with the rest parameters route.

The same doesn't happen if the optional route is followed by a static one: `[[lang=lang]]/t/[...path]`.

```js
// params/lang.js
export const match = (param) => {
  return ['fr', 'de'].includes(param);
};
```

I would expect:
```
/home -> { lang: undefined, path: 'home' } -> gives 404
/de/home -> { lang: 'de', path: 'home' } -> correct
```

In the `[[lang=lang]]/t/[...path]` scenario, it works as expected:
```
/t/home -> { lang: undefined, path: 'home' } -> correct
/de/t/home -> { lang: 'de', path: 'home' } -> correct
```

A common use case for this would be an i18n site with dynamic pages being served by a CMS.

### Reproduction

Here's a stackblitz reproducing the issue:
https://stackblitz.com/edit/sveltejs-kit-template-default-4srbff?file=src/routes/+layout.svelte

And here's another one showing the correct expected behaviour by placing a static route after the optional one:
https://stackblitz.com/edit/sveltejs-kit-template-default-a3vmdd?file=src/routes/+layout.svelte

### Logs

_No response_

### System Info

```Shell
System:
    OS: macOS 13.0
    CPU: (8) arm64 Apple M2
    Memory: 50.08 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.10.0 - ~/.nvm/versions/node/v18.10.0/bin/node
    npm: 8.19.2 - ~/.nvm/versions/node/v18.10.0/bin/npm
  Browsers:
    Brave Browser: 107.1.45.116
    Chrome: 107.0.5304.87
    Firefox: 106.0.5
    Safari: 16.1
  npmPackages:
    @sveltejs/kit: 1.0.0-next.538  => 1.0.0-next.538
    svelte: ^3.44.0 => 3.52.0
    vite: ^3.1.0 => 3.2.3
```

### Severity

### Additional Information

_No response_

dummdidumm:
Document a gotcha around optional parameter with matcher followed by rest parameter
As explained on the issue, I don't think we can (and should) change this, but we should document it.
Closes #7548

### Tests

### Changesets

### DIFF
--- a/documentation/docs/30-advanced/10-advanced-routing.md
+++ b/documentation/docs/30-advanced/10-advanced-routing.md
@@ -67,7 +67,7 @@ export function load(event) {
 
 A route like `[lang]/home` contains a parameter named `lang` which is required. Sometimes it's beneficial to make these parameters optional, so that in this example both `home` and `en/home` point to the same page. You can do that by wrapping the parameter in another bracket pair: `[[lang]]/home`
 
-Note that an optional route parameter cannot follow a rest parameter (`[...rest]/[[optional]]`), since parameters are matched 'greedily' and the optional parameter would always be unused.
+Note that an optional route parameter cannot follow a rest parameter (`[...rest]/[[optional]]`), since parameters are matched 'greedily' and the optional parameter would always be unused. For the same reason (greedy matching), a route like `[[optional=matcher]]/[...rest]` will _always_ assign the first route segment to the optional parameter - so if the matcher says the route doesn't fit, it's discarded instead of assigning the route segment to the `[...rest]` parameter.
 
 ### Matching
 


Rich-Harris:
Will probably end up approving this PR, but I kinda want to sit with it for a bit and see if we can come up with a way to make `[[lang=lang]]/[...rest]` work without adding overwhelming complexity (since I still have PTSD from the last battle with routing logic). It feels like something that _should_ work, and the 'regexes are greedy, so it doesn't' feels like a defeat

Rich-Harris:
merged #7753; closing

PR #7550, opened 2022-11-08
[fix] prevent double decoding of path segment

dmkret:
Segment is already decoded here, so another decoding is not needed

https://github.com/sveltejs/kit/blob/eab9680966812f8755b6d1fe37a310610ec43777/packages/kit/src/utils/routing.js#L24

### Tests

### Changesets


Rich-Harris (approved):

Rich-Harris:
thank you!

PR #7563, opened 2022-11-09
[fix] avoid decoding %25 in path segments

### LINKED ISSUE #7554: Cannot have encoded `%` in URL outside of parameter
### Describe the bug

https://github.com/sveltejs/kit/pull/7521 will not allow you to have a URL like this:

Screenshot from 2022-11-08 14-27-39

Probably the way to fix this is to replace the parameters with placeholders, decodeURI the altered URL, match with regex, and then decodeURIComponent on the parameters

### Reproduction

https://github.com/benmccann/kit/tree/encoded-percent-test

### Logs

_No response_

### System Info

```Shell
`master`
```

### Severity

annoyance

### Additional Information

_No response_

dmkret:
Closes #7554

This is a draft because new test causes 500 error on hydration step.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/utils/routing.js
+++ b/packages/kit/src/utils/routing.js
@@ -21,7 +21,8 @@ export function parse_route_id(id) {
 			: new RegExp(
 					`^${get_route_segments(id)
 						.map((segment, i, segments) => {
-							const decoded_segment = decodeURIComponent(segment);
+							// exclude %25 from decoding to match decoding of url.pathname
+							const decoded_segment = segment.split('%25').map(decodeURIComponent).join('%25');
 							// special case — /[...rest]/ could contain zero segments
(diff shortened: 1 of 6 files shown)


packages/kit/src/utils/routing.js
```diff
@@ -21,7 +21,8 @@ export function parse_route_id(id) {
 			: new RegExp(
 					`^${get_route_segments(id)
 						.map((segment, i, segments) => {
-							const decoded_segment = decodeURIComponent(segment);
+							// exclude %25 from decoding to match decoding of url.pathname
+							const decoded_segment = segment.split('%25').map(decodeURIComponent).join('%25');
```
benmccann, line 25: this seems like it's probably broken. `decode_pathname` uses `decodeURI`, which I think it correct. we shouldn't be doing it a different way here

dmkret: Yes, it seems that avoiding %25 doesn't solve all problems.

PR #7566, opened 2022-11-09
[fix] error docs + handle expected error

### LINKED ISSUE #7272: Have `error` helper redirect to error page when thrown in `handle` hook
### Describe the problem

Similar to #6738 

The documentation for `+error.svelte` and `Loading data > errors` is clear in covering behaviours of the `error` helper function in that it directs users to the nearest `+error.svelte` page when thrown in any `load` method, or returns an error Response object when thrown in an endpoint. 

Since the `handle` hook gets called for every request, whether it's an endpoint or page `load`, throwing `error` in `handle` always results in an error response. For pages, this leads to a generic 500-error page.

Demonstration: https://stackblitz.com/edit/sveltekit-handle-error?file=README.md,src%2Fhooks.server.js&terminal=dev

My specific context is I have a `src/routes/[branch]/+layout.svelte` file whose `handle` method connects to a database based on the `branch` param. If I cannot connect to the database with this `branch` value, the DB connection throws an error, and I want that to translate to a 404 error in SvelteKit.

More personal context for posterity not directly related to this issue: https://github.com/sveltejs/kit/issues/6788#issue-1372268840

Have the behaviour of `throw error()` correspond to the request's context when called in `handle`. This would mean:

1. When the request is a page `load`, then `throw error()` routes to the nearest `+error.svelte` with an error message when thrown in `handle`.
2. When the request is an endpoint, then `throw error()` returns its error message to the requester when thrown in `handle`.

### Alternatives considered

This isn't an earnest ask since I already have a decent workaround as suggested here: https://github.com/sveltejs/kit/issues/6738#issuecomment-1249399240.

Given the file structure `src/routes/[branch]/someotherpage`:

1. Add a `src/routes/+error.svelte` page to catch and display errors thrown in `src/routes/[branch]` or deeper.
2. Pass an error message to `event.locals` in the handle hook:

    ```js
    import { error } from '@sveltejs/kit';
    
      try {
        event.locals.db = await dbConnect(branch)
      } catch (err) {
        // add error to be later thrown in the layout `load` 
        event.locals.error = [404, `Branch "${branch}" does not exist.`]
        const response = await resolve(event)
        return response
      }
    }
    ```

3. Handle error in `src/routes/[branch]/+layout.server.js`

    ```js
    import { error } from '@sveltejs/kit'
    
    export async function load ({ locals }) {
      if (locals.error?.length > 0) {
        // this uses the spread operator since `locals.error`
        // is an array of parameters
        throw error(...locals.error)
      }
      // ...
    }
    ```

Demonstration of workaround: https://stackblitz.com/edit/sveltekit-handle-error-workaround?file=src%2Fhooks.server.js,src%2Froutes%2F[branch]%2F%2Bpage.server.js&terminal=dev

### Importance

nice to have

### Additional Information

If this proposal sounds favourable, then great. If it's something that won't be implemented, then I'd like to at least help clarify the behaviour of `error` somewhere in the docs when it comes to being thrown in `handle`, `load`, or a server endpoint. In the latter case, I hope the alternative explained above can help others in the future. 

As always, thank you SvelteKit maintainers for your time and efforts!

dummdidumm:
- handle expected errors thrown in handle hook correctly
- remove wrong catch of thrown invalid class (only viable inside actions; and it should be returned)
- document behavior around errors thrown in handle/load/endpoint

Closes #7272

### Tests

### Changesets

### DIFF
--- a/documentation/docs/20-core-concepts/10-routing.md
+++ b/documentation/docs/20-core-concepts/10-routing.md
@@ -129,6 +129,8 @@ If an error occurs during `load`, SvelteKit will render a default error page. Yo
 
 SvelteKit will 'walk up the tree' looking for the closest error boundary — if the file above didn't exist it would try `src/routes/blog/+error.svelte` and `src/routes/+error.svelte` before rendering the default error page. If _that_ fails (or if the error was thrown from the `load` function of the root `+layout`, which sits 'above' the root `+error`), SvelteKit will bail out and render a static fallback error page, which you can customise by creating a `src/error.html` file.
 
+> `+error.svelte` is _not_ used when an error occurs inside the [handle hook](/docs/hooks#server-hooks-handle) or an [endpoint](#server).
+
 ### +layout
 
 So far, we've treated pages as entirely standalone components — upon navigation, the existing `+page.svelte` component will be destroyed, and a new one will take its place.
@@ -267,7 +269,7 @@ export function GET({ url }) {
 
 The first argument to `Response` can be a [`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream), making it possible to stream large amounts of data or create server-sent events (unless deploying to platforms that buffer responses, like AWS Lambda).
 
-You can use the `error`, `redirect` and `json` methods from `@sveltejs/kit` for convenience (but you don't have to). Note that `throw error(..)` only returns a plain text error response.
+You can use the `error`, `redirect` and `json` methods from `@sveltejs/kit` for convenience (but you don't have to). Note that `throw error(..)` (or an unexpected error) only returns a plain text error response, [+error.svelte](#error) is not used in this case as an endpoint doesn't live in the context of a page.
 
 #### Receiving data
 
--- a/documentation/docs/20-core-concepts/10-routing.md
+++ b/documentation/docs/20-core-concepts/10-routing.md
@@ -129,7 +129,7 @@ If an error occurs during `load`, SvelteKit will render a default error page. Yo
 
 SvelteKit will 'walk up the tree' looking for the closest error boundary — if the file above didn't exist it would try `src/routes/blog/+error.svelte` and `src/routes/+error.svelte` before rendering the default error page. If _that_ fails (or if the error was thrown from the `load` function of the root `+layout`, which sits 'above' the root `+error`), SvelteKit will bail out and render a static fallback error page, which you can customise by creating a `src/error.html` file.
 
-> `+error.svelte` is _not_ used when an error occurs inside the [handle hook](/docs/hooks#server-hooks-handle) or an [endpoint](#server).
+> `+error.svelte` is _not_ used when an error occurs inside [`handle`](/docs/hooks#server-hooks-handle) or a [+server.js](#server) request handler.
 
 ### +layout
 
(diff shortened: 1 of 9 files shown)


documentation/docs/20-core-concepts/10-routing.md
```diff
@@ -267,7 +269,7 @@ export function GET({ url }) {
 
 The first argument to `Response` can be a [`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream), making it possible to stream large amounts of data or create server-sent events (unless deploying to platforms that buffer responses, like AWS Lambda).
 
-You can use the `error`, `redirect` and `json` methods from `@sveltejs/kit` for convenience (but you don't have to). Note that `throw error(..)` only returns a plain text error response.
+You can use the `error`, `redirect` and `json` methods from `@sveltejs/kit` for convenience (but you don't have to). Note that `throw error(..)` (or an unexpected error) only returns a plain text error response, [+error.svelte](#error) is not used in this case as an endpoint doesn't live in the context of a page.
```
Rich-Harris, line 272: 'Plain text' is probably a bit misleading here — it's an HTML response that uses `src/error.html`, unless the `accept` header prefers JSON

```suggestion
You can use the `error`, `redirect` and `json` methods from `@sveltejs/kit` for convenience (but you don't have to).

If an error is thrown (either `throw error(...)` or an unexpected error), the response will be a JSON representation of the error or a fallback error page — which can be customised via `src/error.html` — depending on the `Accept` header. The [`+error.svelte`](#error) component will not be rendered.
```

I actually think we _might_ need a dedicated section on error handling, the `handleError` docs only get us so far — we gloss over `src/error.html` a bit. Need to think about how to do that without being duplicative

documentation/docs/30-advanced/20-hooks.md
```diff
@@ -86,6 +86,8 @@ export async function handle({ event, resolve }) {
 }
 ```
 
+If an error is thrown during `handle`, it's treated as a fatal error. Depending on the requested response type either a JSON error response or the static fallback `src/error.html` is returned.
```
Rich-Harris, line 89: More evidence that we need a dedicated error handling page — we talk about `src/error.html` but it isn't included in the templates, and we don't explain what can go in it

```suggestion
If an error is thrown during `handle`, it is treated as fatal, and SvelteKit will respond with a JSON representation of the error or a fallback error page — which can be customised via `src/error.html` — depending on the `Accept` header. (Note that `resolve(...)` will never throw an error, it will always return a `Promise<Response>` with the appropriate status code).
```

documentation/docs/20-core-concepts/10-routing.md
```diff
@@ -129,6 +129,8 @@ If an error occurs during `load`, SvelteKit will render a default error page. Yo
 
 SvelteKit will 'walk up the tree' looking for the closest error boundary — if the file above didn't exist it would try `src/routes/blog/+error.svelte` and `src/routes/+error.svelte` before rendering the default error page. If _that_ fails (or if the error was thrown from the `load` function of the root `+layout`, which sits 'above' the root `+error`), SvelteKit will bail out and render a static fallback error page, which you can customise by creating a `src/error.html` file.
 
+> `+error.svelte` is _not_ used when an error occurs inside the [handle hook](/docs/hooks#server-hooks-handle) or an [endpoint](#server).
```
Rich-Harris, line 132:
```suggestion
> `+error.svelte` is _not_ used when an error occurs inside [`handle`](/docs/hooks#server-hooks-handle) or a [+server.js](#server) request handler.
```

Rich-Harris (commented):
I might experiment with an 'Error handling' section under advanced

benmccann:
I see a few different code changes here. Can we add tests for them? Error handling has been particularly tricky to get right

dummdidumm:
> I see a few different code changes here. Can we add tests for them? Error handling has been particularly tricky to get right

one was deleting some obsolete/oversight code which doesn't need a test, the other one I added a test for now 👍

dummdidumm:
We should probably merge #7575 first so that we can link to it from the other places.

Rich-Harris (approved):

PR #7578, opened 2022-11-09
fix flaky test

Rich-Harris:
this one has been particularly troublesome

### DIFF
--- a/packages/kit/test/apps/basics/src/routes/load/invalidation/depends/+page.svelte
+++ b/packages/kit/test/apps/basics/src/routes/load/invalidation/depends/+page.svelte
@@ -9,7 +9,7 @@
 <button
 	type="button"
 	class="shared"
-	on:click={() => invalidate('invalidate-depends:shared')}
+	on:click={() => (window.promise = invalidate('invalidate-depends:shared'))}
 >
 	invalidate shared
 </button>
@@ -18,7 +18,7 @@
 <button
 	type="button"
 	class="server"
-	on:click={() => invalidate('invalidate-depends:server')}
+	on:click={() => (window.promise = invalidate('invalidate-depends:server'))}
 >
 	invalidate server
 </button>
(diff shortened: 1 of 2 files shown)


Conduitry:
This doesn't feel like the Playwright-y way to go about this. Rather than waiting for the promise and then getting the text content, I think we should be doing something like `await expect(page.locator('p.server')).toHaveText(server)` and `await expect(page.locator('p.shared')).not.toHaveText(shared)` which will wait for the expectations to become true.

I have my doubts about whether `window.promise` is even a serializable value that can be transmitted from the browser to the Playwright process. The code here is probably effectively only waiting for `window.promise` to _evaluate_ in the browser, not waiting for it to resolve.

Rich-Harris:
Agree that locator-based tests are more idiomatic — if I were writing tests for an app, that's probably what I'd do. But in the context of tests for SvelteKit itself I think this approach makes more sense:

* the test is stricter — we're not just testing that clicking the button will _eventually_ cause the page to update in a particular way, we're testing that the DOM will be updated by the time the promise returned from `invalidate(...)` resolves. If we refactored the invalidation code and missed an `await` somewhere then we could easily break something and miss it if the test was less strict (granted, we could have a dedicated test for the timing aspect, but that's arguably duplicative and we'd ideally have tests for all the different forms of invalidation). I think this basic principle — test the framework behaviour as precisely as possible, rather than allowing Playwright's anti-flake mechanisms to gloss over any inconsistencies — applies broadly
* if the DOM updates incorrectly, the test will fail immediately rather than after a timeout — the more we can avoid tests timing out, the better
* it stands to reason that the tests will be faster if we're waiting on a promise rather than polling

Rich-Harris:
> I have my doubts about whether `window.promise` is even a serializable value that can be transmitted from the browser to the Playwright process.

From the [docs](https://playwright.dev/docs/evaluating):

> If the result is a Promise or if the function is asynchronous evaluate will automatically wait until it's resolved

dummdidumm (approved):

PR #7588, opened 2022-11-10
[fix] don't run beforeNavigate during redirects

### LINKED ISSUE #7584: `beforeNavigate` runs twice if redirecting in the navigation targets `load`
### Describe the bug

Don't know if this is a bug or a feature... But if I have a before navigate in a route. And then I redirect in the load function on the route we are navigating to. The beforeNavigate runs twice in the original route.

### Reproduction

It's reproduced here:
https://stackblitz.com/edit/sveltejs-kit-template-default-5ed6za?file=src/app.html

I've made this example of the bug occuring.
If we are at the `/` route (Home) and then go to `/about` (About) "BEFORE" is only logged once.
But if we go back to `/` and then go to `/test` (Test) we are correctly redirected to `/about` but "BEFORE" is logged twice.
The console looks like this for the latter case:
```
BEFORE
LOAD IN TEST
BEFORE
```

### Logs

_No response_

### System Info

```Shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.539 
    svelte: ^3.46.0 => 3.52.0 
    vite: ^3.1.0 => 3.2.3
```

### Severity

### Additional Information

This is quite a serious bug for how we use `beforeNavigate` in our application. But can be worked around

dummdidumm:
Fixes #7584

### Tests

### Changesets

### DIFF
--- a/.changeset/hungry-tips-obey.md
+++ b/.changeset/hungry-tips-obey.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-[fix] don't run beforeNavigate during redirects
+[breaking] don't run beforeNavigate during redirects
--- a/.changeset/hungry-tips-obey.md
+++ b/.changeset/hungry-tips-obey.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] don't run beforeNavigate during redirects
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -1066,7 +1066,10 @@ export function create_client({ target, base, trailing_slash }) {
 			}
 		};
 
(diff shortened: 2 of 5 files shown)


.changeset/hungry-tips-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] don't run beforeNavigate during redirects
```
Rich-Harris, line 5: we should probably slap a 'breaking' label on this, just to be transparent

```suggestion
[breaking] don't run beforeNavigate during redirects
```

Rich-Harris (approved):

PR #7591, opened 2022-11-10
clarify that pages with actions cannot be prerendered

### LINKED ISSUE #7586: Cannot prerender pages with actions
### Describe the bug

I have a simple form that once it submits, the default action consoles log the data and returns {success: true}. In production, it will take the data, send an e-mail and return success: true.

According to the docs: "When not to prerender", "The basic rule is this: for a page to be prerenderable, any two users hitting it directly must get the same content from the server." and "Not all pages are suitable for prerendering. Any content that is prerendered will be seen by all users."

In my case, both requirements are true, the submitted form data does not change the content of the page.

If I create a +page.server.ts, and include the default action, then when I hit the page in dev or production, it errors out with "Cannot prerender pages with actions".

Alternatively, I can try the API route with +server.ts, but docs recommend against it, in the section "Receiving data": "In general, form actions are a better way to submit data from the browser to the server."

Something has to give, either this error is unintended behavior, or the docs have to be changed, to include the pages with actions as not prerenderable and that the API route with +server.ts can be the only way to prerender pages with form elements.

Also an explanation in the docs regarding why the pages with actions cannot be prerendered, even if they do not change the content of the page, would be ideal.

Thanks in advance.

### Reproduction

+page.svelte
```
<form
			class="contact"
			method="POST"
			autocomplete="off"
			use:enhance={({ form, data, action, cancel }) => {
				contactFormStore.validateAll()
				if ($contactFormStore.fields.some((x) => x.errors[$langStore])) cancel()
				console.log('form', form)
				console.log('data', data)
				console.log('action', action)
			}}
		>
			<TextInput label={f.name.label} name="name" placeholder={f.name.placeholder} store={contactFormStore} />
			<ActionButton type="submit">{f.submission.send}</ActionButton>
</form>
```

+page.server.ts
```
import type { Actions } from '@sveltejs/kit'

	default: async ({ request }) => {
		const data = await request.formData()
		console.log(data)
		return { success: true `}`
	},
}
```

svelte.config.js
```
import adapter from '@sveltejs/adapter-node'
import preprocess from 'svelte-preprocess'

const config = {
	// Consult https://github.com/sveltejs/svelte-preprocess
	// for more information about preprocessors
	preprocess: preprocess({
		scss: {
			prependData: "@import './src/styles/variables.scss';",
		},
	}),

	kit: {
		adapter: adapter({
			out: 'build',
			precompress: true,
			envPrefix: '',
		}),
		alias: {
			$components: 'src/lib/components',
			'$components/*': 'src/lib/components/*',
			$stores: 'src/lib/stores',
			'$stores/*': 'src/lib/stores/*',
			$types: 'src/lib/types',
			'$types/*': 'src/lib/types/*',
			$utils: 'src/lib/utils',
			'$utils/*': 'src/lib/utils/*',
			$icons: 'src/lib/icons',
			'$icons/*': 'src/lib/icons/*',
		},
	},
}

export default config
```

### Logs

```Shell
Cannot prerender pages with actions
Error: Cannot prerender pages with actions
    at render_page (file:///home/bt/counselor-website-2/node_modules/@sveltejs/kit/src/runtime/server/page/index.js:88:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async resolve (file:///home/bt/counselor-website-2/node_modules/@sveltejs/kit/src/runtime/server/index.js:235:17)
    at async respond (file:///home/bt/counselor-website-2/node_modules/@sveltejs/kit/src/runtime/server/index.js:287:20)
    at async file:///home/bt/counselor-website-2/node_modules/@sveltejs/kit/src/exports/vite/dev/index.js:410:22
```

### System Info

```Shell
System:
    OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
    Memory: 27.27 GB / 30.99 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 18.12.0 - ~/.nvm/versions/node/v18.12.0/bin/node
    npm: 8.19.2 - ~/.nvm/versions/node/v18.12.0/bin/npm
  Browsers:
    Chrome: 107.0.5304.110
    Firefox: 106.0.5
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.84 
    @sveltejs/adapter-node: ^1.0.0-next.100 => 1.0.0-next.100 
    @sveltejs/adapter-static: ^1.0.0-next.46 => 1.0.0-next.46 
    @sveltejs/kit: next => 1.0.0-next.525 
    svelte: ^3.44.0 => 3.52.0 
    vite: ^3.1.0 => 3.1.8
```

### Severity

annoyance

### Additional Information

I have used both adapter-static and adapter-node, same error.

Rich-Harris:
closes #7586

### DIFF
--- a/documentation/docs/20-core-concepts/40-page-options.md
+++ b/documentation/docs/20-core-concepts/40-page-options.md
@@ -62,6 +62,8 @@ Note that you can still prerender pages that load data based on the page's param
 
 Accessing [`url.searchParams`](/docs/load#using-url-data-url) during prerendering is forbidden. If you need to use it, ensure you are only doing so in the browser (for example in `onMount`).
 
+Pages with [actions](/docs/form-actions) cannot be prerendered, because a server must be able to handle the action `POST` requests.
+
 #### Route conflicts
 
 Because prerendering writes to the filesystem, it isn't possible to have two endpoints that would cause a directory and a file to have the same name. For example, `src/routes/foo/+server.js` and `src/routes/foo/bar/+server.js` would try to create `foo` and `foo/bar`, which is impossible.
--- a/documentation/docs/60-appendix/30-glossary.md
+++ b/documentation/docs/60-appendix/30-glossary.md
@@ -16,7 +16,7 @@ Client-side rendering (CSR) is the generation of the page contents in the web br
 
 Prerendering means computing the contents of a page at build time and saving the HTML for display. This approach has the same benefits as traditional server-rendered pages, but avoids recomputing the page for each visitor and so scales nearly for free as the number of visitors increases. The tradeoff is that the build process is more expensive and prerendered content can only be updated by building and deploying a new version of the application.
 
-Not all pages can be prerendered. The basic rule is this: for content to be prerenderable, any two users hitting it directly must get the same content from the server. Note that you can still prerender content that is loaded based on the page's parameters as long as all users will be seeing the same prerendered content.
+Not all pages can be prerendered. The basic rule is this: for content to be prerenderable, any two users hitting it directly must get the same content from the server, and the page must not contain [actions](/docs/form-actions). Note that you can still prerender content that is loaded based on the page's parameters as long as all users will be seeing the same prerendered content.
 
 Pre-rendered pages are not limited to static content. You can build personalized pages if user-specific data is fetched and rendered client-side. This is subject to the caveat that you will experience the downsides of not doing SSR for that content as discussed above.
 


documentation/docs/20-core-concepts/40-page-options.md
```diff
@@ -62,6 +62,8 @@ Note that you can still prerender pages that load data based on the page's param
 
 Accessing [`url.searchParams`](/docs/load#using-url-data-url) during prerendering is forbidden. If you need to use it, ensure you are only doing so in the browser (for example in `onMount`).
 
+Pages with [actions](/docs/form-actions) cannot be prerendered, because a server must be able to handle the action `POST` requests.
```
dummdidumm, line 65: Is this enough explanation? People could go "yeah so why can't it handle that?" - then again, I'm not sure how to explain this without explaining a thousand other things.

dummdidumm (approved):
Giving my approval because, while I feel this could be explained a bit more, I don't know how to do that without going on an endless detour - better something in there than nothing.

documentation/docs/20-core-concepts/40-page-options.md
Rich-Harris: if people come back with further questions we can try and answer them, but i think the important thing is to clarify 'you can't do this' rather than get into the weeds of _why_ you can't do this

PR #7596, opened 2022-11-10
fix: Add support linking to for <a name="hash"> tags.

bfanger:
Fixes the following error:
```
The following pages contain links to /#hash, but no element with id="hash" exists on /
```
when using `<a name="hash">` anchors.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/prerender/crawl.js
+++ b/packages/kit/src/core/prerender/crawl.js
@@ -160,6 +160,8 @@ export function crawl(html) {
 								href = value;
 							} else if (name === 'id') {
 								ids.push(value);
+							} else if (name === 'name' && tag === 'A') {
+								ids.push(value);
 							} else if (name === 'rel') {
 								rel = value;
 							} else if (name === 'src') {
--- a/.changeset/empty-vans-draw.md
+++ b/.changeset/empty-vans-draw.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Add support for linking to <a name="hash"> tags


Conduitry:
Can we know the name of the element we're currently in, so we can avoid false negatives for things other than `<a>`?

Rich-Harris:
Yes, there's a `tag` variable which would be `A` here

packages/kit/src/core/prerender/crawl.js
```diff
@@ -160,6 +160,8 @@ export function crawl(html) {
 								href = value;
 							} else if (name === 'id') {
 								ids.push(value);
+							} else if (name === 'name') {
+								ids.push(value);
```
Rich-Harris, line 164: i think this should do the trick?

```suggestion
								if (tag === 'A') ids.push(value);
```

Conduitry: We should still check that `name === 'name'` as well, no?

bfanger: Added the `tag === 'A'` check

dummdidumm (approved):

dummdidumm:
lint fails are unrelated, so I'm merging this (might be fixed in master already?)

packages/kit/src/core/prerender/crawl.js
```diff
@@ -160,6 +160,8 @@ export function crawl(html) {
 								href = value;
 							} else if (name === 'id') {
 								ids.push(value);
+							} else if (name === 'name' && tag === 'A') {
+								ids.push(value);
```
Rich-Harris, line 164: nit: the previous suggestion allows us bail out of the `if-else` chain sooner — if `name === 'name'` then we don't want to also test for `name === 'rel'` etc regardless of whether `tag === 'A'`

```suggestion
							} else if (name === 'name') {
								if (tag === 'A') ids.push(value);
```

Rich-Harris:
yeah i just pushed a fix for the lint failures, figured they didn't need a PR

packages/kit/src/core/prerender/crawl.js
Rich-Harris: since this was merged i'm just gonna push this change direct to master, don't yell at me

PR #7597, opened 2022-11-10
dev time service workers

### LINKED ISSUE #5479: Service Worker Compilation during Development
### Describe the problem

Testing non-trivial service-worker features in a Sveltekit app is currently pretty painful. The most success I've had is using `npm run build && npm preview`  after every code change, losing the utility of dev mode when testing features that rely on the service-worker.

As far as I've been able to find, Sveltekit only supports building the service-worker in a production build.

Add a config option to enable building the service-worker in development mode, i.e. when running `npm run dev`. This should skip pre-caching (or provide a way for the user to skip serving cached files) and only load additional functionality from the user's service-worker.

### Alternatives considered

[vite-plugin-pwa](https://vite-plugin-pwa.netlify.app/frameworks/sveltekit.html) appears to support [building in development](https://vite-plugin-pwa.netlify.app/guide/development.html) but not for Sveltekit (at least by default, maybe there is a way to configure it to do so?) and also looks like it would add additional configuration to maintain to get it work as well as the built-in SW support.

### Importance

would make my life easier

### Additional Information

I am currently building/maintaining an app that makes extensive use of the service-worker to enable offline functionality. It hosts a small graphql server at /graphql as a single source of truth for pages to get data from. It is not proxying a "real" graphql server and instead fetches data from multiple places to store in IndexedDB, then queries IDB in the resolvers. It also handles an offline-capable file upload system, uploading files to the server when network service is available.

As a result. most of the app does not work well in dev. Sometimes I can use build-then-preview to get localhost to load the service worker, then stop preview and start dev and the page will start live-loading changes to non-service-worker code but this is flaky; if the SW gets unloaded the app needs to be (possibly rebuilt,) previewed, and swapped back to dev again.

If anyone has an alternative to get SWs working in dev I'd be glad to hear it.

Rich-Harris:
Closes #5479. In development, this intercepts requests for `/service-worker.js` and imports the `src/service-worker.js` module, and creates a virtual `$service-worker` module that includes all the usual exports. `build` and `prerendered` are empty arrays.

I haven't added tests for this because I'm not really sure what they should look like.

### Tests

### Changesets


dummdidumm (approved):

AlbertMarashi:
@Rich-Harris @dummdidumm  I believe firefox does not support `import`s inside of service workers, and this is creating errors during development in firefox. Any idea what we should do?

dummdidumm:
As stated in the documentation, this needs support from the browser, so you'd need to use Chromium-based browsers during dev to test this. The alternative is to build and check the result in Preview mode.

PR #7599, opened 2022-11-10
Workaround for the DOM clobbering for use:enhance

### LINKED ISSUE #7593: Form Actions throw error if one if the form field name is also `action`
### Describe the bug

When we have one of the form field name as `action` it is conflicting with the form's action and throw following error 

image

### Reproduction

create a form with one of the field  name as `action`
```html
<form class="space-y-6" method="POST" action="?/save" use:enhance>
      <input type="text" name="action"  id="action"    value={policy.action} />
      <button>Save</button>
</form>
```

### Logs

```Shell
forms.js?v=360cc9a6:50 Uncaught (in promise) TypeError: Failed to construct 'URL': Invalid URL
    at HTMLFormElement.handle_submit (forms.js?v=360cc9a6:50:18)
    at HTMLFormElement.sentryWrapped (helpers.ts:98:1)
```

### System Info

```Shell
System:
    OS: macOS 13.0
    CPU: (10) arm64 Apple M1 Max
    Memory: 30.77 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.11.0 - /opt/homebrew/bin/node
    npm: 8.19.2 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 107.0.5304.110
    Safari: 16.1
  npmPackages:
    @sveltejs/adapter-auto: 1.0.0-next.87 => 1.0.0-next.87 
    @sveltejs/adapter-node: 1.0.0-next.100 => 1.0.0-next.100 
    @sveltejs/adapter-static: 1.0.0-next.48 => 1.0.0-next.48 
    @sveltejs/kit: 1.0.0-next.538 => 1.0.0-next.538 
    svelte: 3.52.0 => 3.52.0 
    vite: 3.2.3 => 3.2.3
```

### Severity

annoyance

### Additional Information

I have this form before migrating to the new `Form Actions` to create/update my form data. I was using `svelte-forms-lib` and using fetch to post data before and was working. now when I am trying Form Actions, I noticed this bug

```ts
	save: async ({ request }) => {
		const formData = await request.formData();
		console.log('save action formData:', formData);
		return {
			success: true
		};
	}
};
```

repsac-by:
Fixes #7593

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/app/forms.js
+++ b/packages/kit/src/runtime/app/forms.js
@@ -50,10 +50,10 @@ export function enhance(form, submit = () => {}) {
 
 		const action = new URL(
 			// We can't do submitter.formAction directly because that property is always set
-			// We do form.cloneNode() for avoid DOM clobbering - https://github.com/sveltejs/kit/issues/7593
+			// We do cloneNode for avoid DOM clobbering - https://github.com/sveltejs/kit/issues/7593
 			event.submitter?.hasAttribute('formaction')
 				? /** @type {HTMLButtonElement | HTMLInputElement} */ (event.submitter).formAction
-				: /** @type {HTMLFormElement} */ (form.cloneNode()).action
+				: /** @type {HTMLFormElement} */ (HTMLFormElement.prototype.cloneNode.call(form)).action
 		);
 
 		const data = new FormData(form);
@@ -119,11 +119,13 @@ export function enhance(form, submit = () => {}) {
 		});
 	}
 
-	form.addEventListener('submit', handle_submit);
+	// @ts-expect-error
+	HTMLFormElement.prototype.addEventListener.call(form, 'submit', handle_submit);
 
 	return {
 		destroy() {
-			form.removeEventListener('submit', handle_submit);
+			// @ts-expect-error
+			HTMLFormElement.prototype.removeEventListener.call(form, 'submit', handle_submit);
 		}
 	};
 }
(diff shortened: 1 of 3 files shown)


dummdidumm:
How in the world does that fix the error? What kind of weird browser behavior is this?

repsac-by:
Because cloning is not deep, but only the form element itself without children.

dummdidumm:
I see, makes sense!

packages/kit/src/runtime/app/forms.js
```diff
@@ -49,9 +50,10 @@ export function enhance(form, submit = () => {}) {
 
 		const action = new URL(
 			// We can't do submitter.formAction directly because that property is always set
+			// We do form.cloneNode() for avoid DOM clobbering - https://github.com/sveltejs/kit/issues/7593
 			event.submitter?.hasAttribute('formaction')
 				? /** @type {HTMLButtonElement | HTMLInputElement} */ (event.submitter).formAction
-				: form.action
+				: /** @type {HTMLFormElement} */ (form.cloneNode()).action
```
Conduitry, line 56: Someone could conceivably have an `<input name="cloneNode">` which would still break this, no?

What are the performance implications of this clone? I'm hoping not huge, because it's not a deep clone.

repsac-by: > Someone could conceivably have an <input name="cloneNode"> which would still break this, no?

That's a good question. We'll call it through the prototype as well.

> What are the performance implications of this clone? I'm hoping not huge, because it's not a deep clone.

Hopefully this will not affect performance too much, because as noted above it is not a deep cloning.

Rich-Harris: Not remotely worried about the performance impact — if it was more than a fraction of a millisecond I'd be astonished (tried it just now with some `console.time(...)` calls and yep, it's basically free)

Rich-Harris (approved):

Rich-Harris:
thank you!

PR #7573, opened 2022-11-09
[fix] prevent usage of unsafe encoded characters in path segments

dmkret:
Closes #7567
Closes #7554
Closes #7570

This PR prevents using path segments (folder names) to contain uri-decodable characters like `[]%` and so on.

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [x] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

Issue #7567, opened 2022-11-09
URI decoding of routes does not work except for select few characters


dmkret:
### Describe the bug

When there is %-encoded character in pathname decodable by `decodeURI`, server sends 200 response for pages and 404 for static files (+page.js, +page.svelte etc.)

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-s8cge5

1. Click on `goto /%5b`
2. Expected: `this is %5b`. Received: glitch, `this is %5b` changes to `500 internal error`.

Inside network inspector you can see that `GET /%5b/+page.svelte` fails with 404.

### Logs

```Shell
Not found: /src/routes/%5b/+page.svelte
Error: Not found: /src/routes/%5b/+page.svelte
    at resolve (file:///home/projects/sveltejs-kit-template-default-s8cge5/node_modules/@sveltejs/kit/src/runtime/server/index.js:281:13)
    at resolve (file:///home/projects/sveltejs-kit-template-default-s8cge5/node_modules/@sveltejs/kit/src/runtime/server/index.js:312:5)
    at Object.handle (file:///home/projects/sveltejs-kit-template-default-s8cge5/node_modules/@sveltejs/kit/src/exports/vite/dev/index.js:360:66)
    at respond (file:///home/projects/sveltejs-kit-template-default-s8cge5/node_modules/@sveltejs/kit/src/runtime/server/index.js:309:40)
    at async eval (file:///home/projects/sveltejs-kit-template-default-s8cge5/node_modules/@sveltejs/kit/src/exports/vite/dev/index.js:430:22)
```

### System Info

```Shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 7.17.0 - /usr/local/bin/npm
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.87 
    @sveltejs/kit: next => 1.0.0-next.538 
    svelte: ^3.46.0 => 3.52.0 
    vite: ^3.1.0 => 3.2.3
```

### Severity

annoyance

dmkret:
Should %-encoded characters, decodable by `decodeURI` be completely rejected in folder names?

[...]

Issue #7577, opened 2022-11-09
Route encoding scheme needs a rethink


benmccann:
### Describe the bug

URLs come from the browser in a semi-encoded fashion. Our router has generally used the decoded versions of things on the file system to make for nice looking files. We then have to decode the URLs in order to match what's on the file system (i.e. `decodeURI`). However, we have historically run `decodeURIComponent` on the parameters. But it turns out the way we were doing that didn't work in all cases and in fact there's no way to actually handle the param parts differently from the non-param parts because we don't know what the param parts are until after the matching has been run which requires decoding first

I looked to see how Next handles this and apparently they don't as it's not documented and is one of the top bugs in their repo: https://github.com/vercel/next.js/issues/10084

There are some fundamental limitations in mapping between a URL and file system since they have different limitations. Here are the various possible combinations:
- Case 1: legal on FS. not encoded. (e.g. `A`)
- Case 2: illegal on FS. not encoded. (e.g. `*`)
- Case 3: legal on FS. sent encoded (e.g. `苗`)
- Case 4: illegal on FS. sent encoded (e.g. `<`)
- Case 5: kit special character (e.g. `[`)
- Case 6: URL encoding character (i.e. `%`)

With the characters that are illegal to put on the file system or are kit special characters, we have two choices: make them encoded on the FS or only allow them in parameters. Let's choose the latter as it will constrain our design less, make foreign characters much nicer, and be less breaking than forcing URLs to be encoded.

Now the big question is how can we `decodeURIComponent` URL parameters?

For reference, the list of characters replaced by `decodeURIComponent` but not `decodeURI` is:
- `/` - illegal on FS. must be sent encoded
- `?` - illegal on FS. must be sent encoded
- `:` - can be illegal on FS
- `#` - must be sent encoded. this needs to be made illegal in routes eventhough the FS accepts it
- `;`
- `,`
- `@`
- `&`
- `=`
- `+`
- `$`

We can't just make these characters illegal, because here we're talking about URL parameters and we need to be able to represent anything. We can't simply do a `decodeURIComponent` upfront and then match via regex because of stuff like `/` (imagine the route `/a/[b]/c` with the URL `/a/b%2fc`). I'll also note here that special casing stuff like `/` or `%2f` is the road to hell and is how we're in our current predicament (https://github.com/sveltejs/kit/pull/7521 tried to special case `%25`), so we very much don't want to do that either.

### Suggested fix

[...]

PR #7604, opened 2022-11-11
[breaking] remove support for encoded directories

benmccann:
This is technically breaking, but only barely. Encoded directories mostly didn't work save for 11 characters. It's not clear to me it's possible to support more than that and those characters can generally be represented in a much nicer way in decoded format anyway. There's really no reason a user should have been writing `%40` instead of `@` in their directory name in the first place. I think `#`, `?` (only on Linux and Mac), and `:` (only on certain Windows systems) are the only ones that might be a bit less convenient to write as you'd now have to use a parameter matcher now. This seems like a good trade-off for making the code and docs much easier to understand

I removed the `$SVLT` test simply because it's already covered by other tests such as the `@svelte` test. The others were removed because they were testing encoded characters in directories which we no longer support

Closes https://github.com/sveltejs/kit/issues/7554
Closes https://github.com/sveltejs/kit/issues/7567
Closes https://github.com/sveltejs/kit/pull/7573
Closes https://github.com/sveltejs/kit/pull/7563

Probably solves https://github.com/sveltejs/kit/issues/7577 although I'm going to leave that open for a bit longer as I'd like to check on a couple more things first


Conduitry:
We initially had this because we wanted to be able to support routes with arbitrary special characters in them. Without this, there's no way of having a route with literal characters in it that would normally have special meaning to SK other than to create a dummy param with a matcher that matches that one specific string. This works, but is ugly and wasteful. I don't think we want to completely remove support for encoded directory names.

benmccann:
Yes, however, that has never actually worked. I tried dozens of versions going back more than a year to at least Vite 2.5 and none of them supported putting `%5b` (an encoded `[`) in the route except as a parameter. So this doesn't change anything in that regard. And more importantly, as far as I can tell, I don't think it's something Vite supports and thus would be impossible for us to implement even if we wanted to.

If we make this change now it puts us in a safe place for 1.0 where what we claim to support and actually support match. It would be backwards compatible to add support for encoded routes in the future if anyone else can figure out how or there's a new release of Vite that begins supporting it. In the meantime, it's such an uncommon character to have in a URL and no one has ever filed an issue about it, so I don't think it's a very common use case that we need to implement before 1.0.

dummdidumm:
I haven't followed the (de)coding madness in the past, so how big of a breaking change is this? Were people previously forced to do `%40[username]` and now they are forced to do `@[username]`, or did `@[username]` always work, so people had two ways of doing the same? How often do people want to have `#` or `?` in their filenames (any examples of that in the wild)? Does this have any effect on the characters encoded _inside_ parameters (`[blog]` where `blog` is `does-this-work%3F`; `%3F` is `?`)?

benmccann:
`@` always worked, so this is a very minor breaking change. There are only three characters that are somewhat affected per the PR description. I don't believe you were ever allowed a `%` inside brackets as part of a parameter name

dummdidumm:
I mean the potential impact to people who have written down their paths in an encoded manner at the moment (`hello/world%3F/+page.svelte`) - from what I understand, this would no longer be possible, which might be a bigger breaking change than we think.

benmccann:
There were only 11 characters that were even possible to write in encoded format previously. I've honestly never seen one in the wild. `hello/world%3F` is a pretty ugly URL, so people don't tend to hard code symbols like that in URLs but only use them in dynamic contexts.

packages/kit/src/utils/routing.js
```diff
@@ -21,17 +21,26 @@ export function parse_route_id(id) {
...
+							if (segment.includes('%')) {
+								throw new Error(`Routes ${id} must be provided in decoded form.`);
+							}
+							if (/[#\/\?]/.test(segment)) {
+								// #, ?, and / will not be decoded by decodeURI so cannot be matched
+								// '/' is illegal on all file systems so we don't bother warning about it
+								throw new Error(
+									`Route ${id} should not contain '#' or '?'. Please use a parameter.`
+								);
+							}
```
Rich-Harris, line 33: this should happen sooner, in `create_manifest_data`. if it's here then this code will be sent to the client, needlessly

Rich-Harris:
Been noodling around on this stuff locally and while I was able to get everything basically working for SSR and in a built app, it fails when JS kicks in because Vite can't load modules with certain characters in them. I think it's probably a Vite bug, but I also think that using URL-encoded characters in filenames is always going to be fraught.

So I agree that we should remove support for URL-encoded characters in directory names. I _don't_ agree that we should prevent people from using those characters in their routes though. Asking people to do `foo[x=colon]bar` for `/foo:bar` — creating a useless `x` param and forcing the creation of a `colon` matcher — is user-hostile in my view.

So we need a way of encoding certain characters that Vite can understand and that works on all filesystems. Ideally, it would be familiar to web developers. Luckily, just such a method exists: HTML entities. The example above would become `foo&colon;bar`.

Ideally, the decoding would happen as early in the process as possible. I think we should be able to decode everything except the characters that that are meaningful to SvelteKit: `[ ] ( )`. (Aside: we've talked about this before without resolution — do we need to reserve any other characters, in case we want to implement something like Next's proposed [dynamic parallel routes](https://nextjs.org/blog/layouts-rfc#dynamic-parallel-routes)? Or are we okay with something like that requiring a major version bump?)

PR #7610, opened 2022-11-11
[fix] don't cache prefetch errors

### LINKED ISSUE #6709: Load function should re-run if it was prefetched with an error
### Describe the problem

I am having a situation where I am calling an API in my load functions and storing an access token and a refresh token in http cookies. When an endpoint returns 401 I redirect to a refresh token endpoint, update the cookies and redirect back to the original request. Anyways the problem happens when multiple requests are sent at the same time. Because the refresh token can only be used once. If multiple requests are sent with an expired access token, some of them will return 401 because the refresh token will no longer be valid. I can get around this in load functions by calling 1 endpoint first and the followed by others:
```
const res1 = await fetch(...);
const [res2, res3] = await fetch();
```

But what I can't control is when having links with prefetch enabled and a user hovers over multiple of them at the same time. If the token is expired, some of the load functions will fail (401). But the token will actually be refreshed in another request. So now when I open a prefetched link that failed, I get the 401 error although the token was refreshed. Does it make sense that a load function re-runs if an error was thrown during prefetching?

If an error was thrown in a load function during prefetching, it should re-run wen visiting the actual page instead of showing the error.

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

dummdidumm:
Fixes #6709

### Tests

### Changesets


packages/kit/test/apps/basics/test/client.test.js
```diff
@@ -698,28 +698,36 @@ test.describe('Prefetching', () => {
 		page
 	}) => {
 		await page.goto('/routing/a');
-		await app.prefetch('/routing/prefetched/hash-route#please-dont-show-me');
-		await app.goto('/routing/prefetched/hash-route');
+		await app.prefetch('/routing/prefetching/hash-route#please-dont-show-me');
```
Rich-Harris, line 701: i might just not be fully caffeinated yet but the diff shows a bunch of files that were deleted but are still referenced in these tests? somewhat confused as to how they're passing!

dummdidumm: oh yeah sorry, forgot to mention that. While working on the test I noticed that there was essentially some duplicate pages for the tests. I removed them and changed the urls of the affected tests (the contents of those deleted files were the same as the ones they point to now).

Rich-Harris: oooohhhhhh huh. probably my bad then — oops!

Rich-Harris (approved):

PR #7613, opened 2022-11-11
[docs] clarify load/action connection in example

dummdidumm:
From a Discord question

### Tests

### Changesets

### DIFF
--- a/documentation/docs/20-core-concepts/30-form-actions.md
+++ b/documentation/docs/20-core-concepts/30-form-actions.md
@@ -98,8 +98,9 @@ Each action receives a `RequestEvent` object, allowing you to read the data with
 // @errors: 2339 2304
 /// file: src/routes/login/+page.server.js
 /** @type {import('./$types').PageServerLoad} */
-export function load({ cookies }) {
-	return db.getUserFromSession(cookies.get('sessionid'));
+export async function load({ cookies }) {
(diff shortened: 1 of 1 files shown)


documentation/docs/20-core-concepts/30-form-actions.md
```diff
@@ -97,6 +97,11 @@ Each action receives a `RequestEvent` object, allowing you to read the data with
 ```js
 // @errors: 2339 2304
 /// file: src/routes/login/+page.server.js
+/** @type {import('./$types').PageServerLoad} */
+export function load({ cookies }) {
+	return db.getUserFromSession(cookies.get('sessionid'));
+}
```
Rich-Harris, line 103: Realistically, this code wouldn't be in the `/login` route, it'd be in the root layout. Is it worth having a separate snippet or is that overkill? Either way, we probably don't want to return it as a top-level value

```suggestion
export async function load({ cookies }) {
	const user = await db.getUserFromSession(cookies.get('sessionid'));
	return { user };
}
```

dummdidumm: For simplicity's sake I would keep it here. I was imagining `getUserFromSession` returning a `{ user } ` object, but your snippet is better, makes it more obvious.

Rich-Harris (approved):

PR #7615, opened 2022-11-11
narrow down possible status codes for redirects to 300-308

### LINKED ISSUE #7614: DX improvement: narrow down possible status codes for the `redirect` helper
### Describe the problem

The current implementation throws an error if the statuscode passed to `redirect` is smaller than `300` or bigger than `399` which is already a good indicator what kind of status codes are supported.

But actually the range is way smaller: from `300` to `308`
https://www.rfc-editor.org/rfc/rfc9110#name-redirection-3xx

With all other status codes the browser will not know what to do.

Another smaller issue is, that the `TypeScript` type is set to `number`, which is correct, but it could be narrowed down to the actual values that are supported.
_Background: a while ago someone threw a 401 redirect in our codebase and we didn't spot the issue immediately._

Narrowing down the amount of status code allowed for the `redirect` helper function to the official ones ranging from 300 to 308
- for runtime code
- and also at the TypeScript level

While testing a few things it seems that 305 and 306 are ignored by firefox and chrome. 300 gets ignored by chrome. So maybe also removing status codes that will not work makes sense? (306 is officialy marked as "unused")

Opening this issue for discussion reasons. I can also create the PR if it helps you :)

### Alternatives considered

Leaving as it is

### Importance

would make my life easier

### Additional Information

_No response_

ivanhofer:
closes #7614

### Tests

### Changesets


packages/kit/types/index.d.ts
```diff
@@ -727,11 +727,13 @@ export function error(
...
 ): HttpError;
 
+type HttpRedirectStatusCode = 300 | 301 | 302 | 303 | 304 | 305 | 306 | 307 | 308
+
 /**
  * Creates a `Redirect` object. If thrown during request handling, SvelteKit will
  * return a redirect response.
  */
-export function redirect(status: number, location: string): Redirect;
+export function redirect(status: HttpRedirectStatusCode, location: string): Redirect;
```
dummdidumm, line 736: Personally I would inline these, as the type is only used in one place (and I wouldn't want to export it for now).

packages/kit/src/runtime/control.js
```diff
@@ -55,14 +55,16 @@ export function error(status, message) {
 	return new HttpError(status, message);
 }
 
+/** @typedef { 300 | 301 | 302 | 303 | 304 | 305 | 306 | 307 | 308 } HttpRedirectStatusCode }
+
 /**
  * Creates a `Redirect` object. If thrown during request handling, SvelteKit will
  * return a redirect response.
- * @param {number} status
+ * @param {HttpRedirectStatusCode} status
```
dummdidumm, line 63: see other comment about inlining

packages/kit/test/apps/basics/test/test.js
```diff
@@ -1283,6 +1283,18 @@ test.describe('Redirects', () => {
 		);
 	});
 
+	test('errors on invalid 3xx status', async ({ baseURL, page, clicknav }) => {
```
dummdidumm, line 1286: doing an e2e test for this is kind of overkill, a unittest would be enough. But since that unit test tests the same logic we have inside the function, it's kind of pointless to have a unit test anyway in my opion 😄

ivanhofer: I just looked for `redirect` in the tests folder and copied an existing one. But I can get rid of it

packages/kit/types/index.d.ts
ivanhofer: I knew this commend was coming ^^. Sure!

packages/kit/types/index.d.ts
```diff
@@ -731,7 +731,7 @@ export function error(
  * Creates a `Redirect` object. If thrown during request handling, SvelteKit will
  * return a redirect response.
  */
-export function redirect(status: number, location: string): Redirect;
+export function redirect(status: 300 | 301 | 302 | 303 | 304 | 305 | 306 | 307 | 308, location: string): Redirect;
```
benmccann, line 734: you need to run `pnpm format` in the `kit` project. `lint` is failing here

ivanhofer: done

DanielHritcu:
Can't we make the type more generic ?

```typescript

type RangeStartArray<
	LENGTH extends number,
	ACC extends unknown[] = [],
> = ACC['length'] extends LENGTH
	? ACC
	: RangeStartArray<LENGTH, [...ACC, 1]>

type NumericRangeBuilder<
	START_ARR extends number[],
	END extends number,
	ACC extends number = never>
	= START_ARR['length'] extends END
	? ACC | END
	: NumericRangeBuilder<[...START_ARR, 1], END, ACC | START_ARR['length']>

/**
 * A type that represents a range of numbers 'FROM' to `END` (inclusive)
 * 
 * @example
 * type Range = NumericRange<100, 200> // 100 | 101 ... 199 | 200
 */
export type NumericRange<FROM extends number, TO extends number> = NumericRangeBuilder<RangeStartArray<FROM>, TO>

/**
 * 
 * @property {number} informational - 1xx
 * @property {number} success - 2xx
 * @property {number} redirection - 3xx
 * @property {number} clientError - 4xx
 * @property {number} serverError - 5xx
 * 
 * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
 */
export type HttpStatusCode = {
	informational: NumericRange<1, 199>;
	success: NumericRange<200, 299>;
	redirect: NumericRange<300, 399>;
	clientError: NumericRange<400, 499>;
	serverError: NumericRange<500, 599>;
}
```

```javascript
/**
 * Creates a `Redirect` object. If thrown during request handling, SvelteKit will
 * return a redirect response.
 * @param status The HTTP redirect status code (300-399)
 * @param location The location to redirect to
 */
export function redirect(status: HttpStatusCode['redirect'], location: string): Redirect;
```

ivanhofer:
> Can't we make the type more generic ?

If it would be more performant.. then probably yes.
I haven't seen any performant `NumberRange` implementation yet. This one takes a few minutes on my machine to tell me that `500` is not allowed.

dummdidumm (approved):
Number range generator is something for another PR, if ever - I'd rather have more simple types. In this case it's ok because it's only a handful of numbers, which can be hardcoded.

Rich-Harris:
thank you!

Issue #2527, opened 2021-09-30
slots are being called twice from __layout.svelte when dealing with $page store and the logic {#if block


opensas:
### Describe the bug

I have a _layout which should get the `<slot />` content inside a container depending on the path.

So I use $page.path to detect when there's a change in the path, and an `{#if path === 'xxx'}` block to handle each case.

At start everything works fine, layout is initialized one and the component is initialized once too

From there on, every time I navigate to another page the component is initialized twice

### Reproduction

Repo at gihub: https://github.com/opensas/sveltekit-slots-called-twice-error

Open the repo in gitpod: https://gitpod.io#github.com/opensas/sveltekit-slots-called-twice-error

in terminal

```shell
$ npm i -g pnpm
$ pnpm i
$ pnpm dev

click open browser (allows popups)

- first run, everything is ok

![image](https://user-images.githubusercontent.com/481687/135449515-964a3b49-7395-4a19-98b0-6b555f7d8ca9.png)

```
init __layout.svelte
Object { path: "/" }
about to display index.svelte from __layout.svelte
init root
```

- Click on MyComp, MyComp is initialized twice

![image](https://user-images.githubusercontent.com/481687/135449738-e76ce100-06cc-45bc-b1ec-d9d74edd843c.png)

```
init MyComp
Object { path: "/MyComp" }
about to display MyComp from __layout.svelte
init MyComp
```

### Logs

_No response_

### System Info

```shell
(as executed on gitpod console)

$ npx envinfo --system --binaries --browsers --npmPackages "{svelte,@sveltejs/*,vite}"
npx: installed 1 in 1.301s

  System:
    OS: Linux 5.4 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (16) x64 Intel(R) Xeon(R) CPU @ 2.80GHz
    Memory: 919.66 MB / 62.81 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 14.17.6 - ~/.nvm/versions/node/v14.17.6/bin/node
    Yarn: 1.22.11 - ~/.nvm/versions/node/v14.17.6/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v14.17.6/bin/npm
  npmPackages:
    @sveltejs/kit: next => 1.0.0-next.176 
    svelte: ^3.42.6 => 3.43.0
```

### Severity

serious, but I can work around it

baseballyama:
Hi.

[...]

PR #7617, opened 2022-11-11
[docs] explain how to break out of root layout

dummdidumm:
This comes up so often on Discord etc that it warrants its own section with some explanatory content.

### Tests

### Changesets

### DIFF
--- a/documentation/docs/30-advanced/10-advanced-routing.md
+++ b/documentation/docs/30-advanced/10-advanced-routing.md
@@ -229,7 +229,7 @@ src/routes/
 
 There is no way to break out of the root layout. You can be sure that it's always present in your app and for example put app-wide UI or behavior in it - and if you don't define one, it falls back to a simple `<slot />` component. If you want some pages to _not_ have the same layout as the rest, then you have two options:
 
-Option 1: Create a `(group)` at the root, in which you create a `+layout.svelte` which is used for each page below it. Every page that should inherit a blank root layout instead is put outside of that root `(group)`. In the following example, the login page inherits a blank root layout, while the other pages inherit the layout inside `(group)`:
+Option 1: Create a `(group)` at the root and a `+layout.svelte` within it, which is used for each page below it. Every page that should inherit a blank root layout instead is put outside of that root `(group)`. In the following example, the login page inherits a blank root layout, while the other pages inherit the layout inside `(group)`:
 
 ```
 src/routes/
--- a/documentation/docs/30-advanced/10-advanced-routing.md
+++ b/documentation/docs/30-advanced/10-advanced-routing.md
@@ -227,7 +227,7 @@ src/routes/
 
 #### Breaking out of the root layout
 
-There is no way to break out of the root layout. You can be sure that it's always present in your app and for example put app-wide UI or behavior in it - and if you don't define one, it falls back to a simple `<slot />` component. But what to do, if you want some pages to _not_ have the same layout as the rest? You have two options:
+There is no way to break out of the root layout. You can be sure that it's always present in your app and for example put app-wide UI or behavior in it - and if you don't define one, it falls back to a simple `<slot />` component. If you want some pages to _not_ have the same layout as the rest, then you have two options:
 
 Option 1: Create a `(group)` at the root, in which you create a `+layout.svelte` which is used for each page below it. Every page that should inherit a blank root layout instead is put outside of that root `(group)`. In the following example, the login page inherits a blank root layout, while the other pages inherit the layout inside `(group)`:
 
--- a/documentation/docs/30-advanced/10-advanced-routing.md
+++ b/documentation/docs/30-advanced/10-advanced-routing.md
@@ -207,8 +207,6 @@ src/routes/
 └ +layout.svelte
 ```
 
-There is no way to break out of the root layout. You can be sure that it's always present in your app and for example put app-wide UI or behavior in it.
-
 #### +layout@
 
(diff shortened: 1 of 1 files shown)


documentation/docs/30-advanced/10-advanced-routing.md
```diff
@@ -227,9 +225,41 @@ src/routes/
 └ +layout.svelte
 ```
 
+#### Breaking out of the root layout
+
+There is no way to break out of the root layout. You can be sure that it's always present in your app and for example put app-wide UI or behavior in it - and if you don't define one, it falls back to a simple `<slot />` component. But what to do, if you want some pages to _not_ have the same layout as the rest? You have two options:
```
gtm-nayan, line 230:
```suggestion
There is no way to break out of the root layout. You can be sure that it's always present in your app and for example put app-wide UI or behavior in it - and if you don't define one, it falls back to a simple `<slot />` component. If you want some pages to _not_ have the same layout as the rest, then you have two options:
```

benmccann: I think this is probably a good suggestion. Though I might make it "But if" instead of just "If"

documentation/docs/30-advanced/10-advanced-routing.md
```diff
@@ -227,9 +225,41 @@ src/routes/
...
+
+Option 1: Create a `(group)` at the root, in which you create a `+layout.svelte` which is used for each page below it. Every page that should inherit a blank root layout instead is put outside of that root `(group)`. In the following example, the login page inherits a blank root layout, while the other pages inherit the layout inside `(group)`:
```
benmccann, line 232:
```suggestion
Option 1: Create a `(group)` at the root and a `+layout.svelte` within it, which is used for each page below it. Every page that should inherit a blank root layout instead is put outside of that root `(group)`. In the following example, the login page inherits a blank root layout, while the other pages inherit the layout inside `(group)`:
```

documentation/docs/30-advanced/10-advanced-routing.md
```diff
@@ -227,9 +225,41 @@ src/routes/
 └ +layout.svelte
 ```
 
+#### Breaking out of the root layout
+
+There is no way to break out of the root layout. You can be sure that it's always present in your app and for example put app-wide UI or behavior in it - and if you don't define one, it falls back to a simple `<slot />` component. If you want some pages to _not_ have the same layout as the rest, then you have two options:
```
Rich-Harris, line 230: We need to phrase this differently — it's framed as a limitation rather than a deliberate design choice, and it's weird to have a 'how to do X' section that begins 'you can't'

```suggestion
#### Ignoring all layouts

The root layout applies to every page of your app. If you want some pages to _not_ have the same layout as the rest, then you have two options:
```

documentation/docs/30-advanced/10-advanced-routing.md
```diff
@@ -227,9 +225,41 @@ src/routes/
...
+	import { page } from '$app/stores';
+</script>
+
+{#if $page.url.pathname !== '/login'}
+	<slot />
+{:else}
+	<div>Surrounding UI</div>
+	<slot />
+{/if}
+```
```
Rich-Harris, line 258: I feel like this is a pattern we don't really want to promote — it's the sort of thing that can get out of hand quite quickly, and it introduces sort of a mismatch between the UI code (which has `if` statements) and data fetching code (which doesn't). It also defeats code-splitting — in a more realistic example, one of the branches would use a component, which we'd be importing in both cases

Rich-Harris (commented):
This honestly feels rather duplicative to me — can't we add remarks to the existing sections? The `(group)` section already includes an `/admin` route which would be a perfect example of this, and the `page@` section is all about breaking out of the layout hierarchy, so it feels like a more appropriate venue for this

documentation/docs/30-advanced/10-advanced-routing.md
dummdidumm: I phrased it specifically that way because on Discord people phrase their questions more around "breaking out of" than "ignoring" - but I agree that it's somewhat clunky the way I've written it.

documentation/docs/30-advanced/10-advanced-routing.md
dummdidumm: For simplicity's sake I think this solution is actually perfectly fine for a number of use cases. I agree though that this gets out of hand at some point, so we should point that out.

dummdidumm:
> This honestly feels rather duplicative to me — can't we add remarks to the existing sections? The `(group)` section already includes an `/admin` route which would be a perfect example of this, and the `page@` section is all about breaking out of the layout hierarchy, so it feels like a more appropriate venue for this

The `page@` section already had a sentence on that previously, and it seems people just glanced over it. That's why I decided to give this special case more room, since it's such a commonly asked thing. The placement of the section also has a somewhat nice segway to the "alternatives" section. I would be ok with moving it inside the `page@` section though, but with much more text and code snippets than what it has now.

documentation/docs/30-advanced/10-advanced-routing.md
Rich-Harris: Just remembered this bug: https://github.com/sveltejs/kit/issues/2527

We absolutely don't want to encourage this pattern

Rich-Harris:
I had a crack at a shorter version in #7651

dummdidumm:
Closing in favor of #7651

PR #7621, opened 2022-11-11
Html entities

Rich-Harris:
Follow-up to https://github.com/sveltejs/kit/pull/7604#issuecomment-1312163698

### Tests

### Changesets


documentation/docs/30-advanced/10-advanced-routing.md
```diff
@@ -123,27 +123,26 @@ src/routes/[...catchall]/+page.svelte
...
-);
-```
-
-To express a `%` character, use `%25`, otherwise the result will be malformed.
+Some characters can't be used on the filesystem — `/` on Linux and Mac, `\ / : * ? " < > |` on Windows. The `#` character has special meaning in URLs, and the `[ ] ( )` characters have special meaning to SvelteKit, so these also can't be used directly as part of your route.
+
+To use these characters in your routes, you can use HTML entities:
+
+- `\` — `&bsol;`
+- `/` — `&sol;`
```
benmccann, line 131: we cannot support `/`, `#`, or `?`. you must use a parameter for these. see the comment here https://github.com/sveltejs/kit/pull/7621/files#diff-a1a15b06db23e414ab4202232607550bcac6cfe2b0809777468e8dabb922b276R28

when we match we decode the URL with `decodeURI` and if you send an encoded version of these characters they will not be decoded, so you can't match them. they can only be sent via parameter

packages/kit/test/apps/basics/test/test.js
```diff
@@ -323,16 +323,16 @@ test.describe('Encoded paths', () => {
...
+	test('allows non-ASCII character in parameterized route segment', async ({ page, clicknav }) => {
 		await page.goto('/encoded');
-		await clicknav('[href="/encoded/$SVLT"]');
-		expect(await page.textContent('h1')).toBe('$SVLT');
+		await clicknav('[href="/encoded/@svelte"]');
+		expect(await page.textContent('h1')).toBe('@svelte');
 	});
 
-	test('allows %-encoded characters in filenames', async ({ page, clicknav }) => {
+	test('allows characters to be represented as HTML entities', async ({ page, clicknav }) => {
```
benmccann, line 332: some other characters worth testing as having unique properties:
- `#`
- `?`
- `*`
- `苗` (as `&#x82D7;`)
- `<`

you could probably just shove them all in this same test

Rich-Harris: Added tests for all but `&#x82D7;` which is... tricky. We don't let you include a `#` character in a filename, because it breaks the URL as far as Vite is concerned.

So we either need to accept the limitation that only named entities can be used (which covers all the exceptions that we actually care about — the characters that can't be used in a Windows filename plus the characters that have special meaning to SvelteKit), or we need to come up with an alternative way of encoding characters. URL-encoding is a no-go, and `\u1234`-style encoding won't work because you can't have a `\` character on Windows

Rich-Harris: we could always do something wacky like `src/routes/{u+1234}/+page.svelte`, if we reserved the `{ }` characters

Rich-Harris: I actually quite like that, the more I think of it — maybe we use that scheme instead of HTML entities, so that there's one consistent approach? Much simpler implementation-wise.

Any other characters we should be reserving for future use?

Conduitry: I'm not sure how I feel about that particular syntax versus other potential ones - but I do think it's a good move to support only ASCII/Unicode code points here, and not all possible named HTML entities.

Rich-Harris: What other syntaxes should we consider? I think the `u+1234` bit makes sense, as that's in fairly widespread use already, so it's really just the delimiters we need to worry about. If we want symmetrical delimiters then the options are `{u+1234}`, `[u+1234]` or `(u+1234)`. I think `( )` is probably a bit weird since elsewhere it means 'ignore this'. I don't hate `[ ]` — it's not breaking, since `[u+1234]` is currently disallowed, and avoids using up another character pair we might put to some other use in future.

We could also do `|u+1234|` or whatever though that's probably a bit weird.

Conduitry: `|` is disallowed in path names in Windows (and would likely cause confusion with shell escaping anyway, because of stdout/stdin piping).

`[u+1234]` not using up another delimiter just for this seems a bit compelling.

PR #7629, opened 2022-11-12
[fix] add optional Promise return type to `enhance` action

### LINKED ISSUE #7628: `SubmitFunction` type error when using `async` callback in `use:enhance`
### Describe the bug

Follow up to #7608, `use:enhance` throws a type error when using an `async` callback because the `SubmitFunction` interface does not expect any Promise-like return value.

### Reproduction

Use an `async` function as the callback for any `use:enhance` action.

### Logs

```
Type '({ cancel, data }: { action: URL; data: FormData; form: HTMLFormElement; controller: AbortController; cancel(): void; }) => Promise<void | (({ update, result }: { form: HTMLFormElement; action: URL; result: ActionResult<...>; update(options?: { ...; } | undefined): Promise<...>; }) => Promise<...>)>' is not assignable to type 'SubmitFunction<Record<string, any>, Record<string, any>>'.
  Type 'Promise<void | (({ update, result }: { form: HTMLFormElement; action: URL; result: ActionResult<Record<string, any>, Record<string, any>>; update(options?: { ...; } | undefined): Promise<...>; }) => Promise<...>)>' is not assignable to type 'void | ((opts: { form: HTMLFormElement; action: URL; result: ActionResult<Record<string, any>, Record<string, any>>; update(options?: { ...; } | undefined): Promise<...>; }) => void)'.
```

### System Info

```Shell
System:
    OS: Windows 10 10.0.22621
    CPU: (12) x64 Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
    Memory: 15.90 GB / 31.75 GB
  Binaries:
    Node: 16.17.1 - C:\Program Files\nodejs\node.EXE
    npm: 8.9.0 - ~\AppData\Roaming\npm\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.891.0), Chromium (107.0.1418.42)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @sveltejs/adapter-vercel: 1.0.0-next.81 => 1.0.0-next.81
    @sveltejs/kit: 1.0.0-next.544 => 1.0.0-next.544
    svelte: ^3.53.1 => 3.53.1
    vite: ^3.2.3 => 3.2.3
```

### Severity

annoyance

### Additional Information

_No response_

s3812497:
fixes #7628

My bad for forgetting to update the type!

Not sure if this is the best way to update it to optionally return a Promise.

### Tests

### Changesets

### DIFF
--- a/.changeset/honest-eyes-thank.md
+++ b/.changeset/honest-eyes-thank.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] add Promise return type to the `enhance` action
--- a/packages/kit/types/ambient.d.ts
+++ b/packages/kit/types/ambient.d.ts
@@ -95,6 +95,20 @@ declare module '$app/environment' {
 declare module '$app/forms' {
 	import type { ActionResult } from '@sveltejs/kit';
 
+	export type ActionCallback<
+		Success extends Record<string, unknown> | undefined = Record<string, any>,
+		Invalid extends Record<string, unknown> | undefined = Record<string, any>
+	> = (opts: {
+		form: HTMLFormElement;
+		action: URL;
+		result: ActionResult<Success, Invalid>;
+		/**
+		 * Call this to get the default behavior of a form submission response.
+		 * @param options Set `reset: false` if you don't want the `<form>` values to be reset after a successful submission.
+		 */
+		update(options?: { reset: boolean }): Promise<void>;
+	}) => void;
+
 	export type SubmitFunction<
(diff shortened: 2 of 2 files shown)


Rich-Harris:
Thanks! We can simplify this with a `MaybePromise<T>` type — we use this in a bunch of places already. Unfortunately you can't reference private types in an ambient declaration AFAIK, so we have to recreate it, but it still nets out to less code. (Unless @dummdidumm has any smarter ideas?)

dummdidumm (approved):
Looks good! If we need `MaybePromise` in more places in ambient in the future, we should just make it public.

PR #7521, opened 2022-11-06
[fix] prevent double decoding of params

dmkret:
Closes #7515 

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [x] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #7644, opened 2022-11-14
Use escape sequences to denote special characters

### LINKED ISSUE #7554: Cannot have encoded `%` in URL outside of parameter

### LINKED ISSUE #7567: URI decoding of routes does not work except for select few characters

### LINKED ISSUE #7577: Route encoding scheme needs a rethink

Rich-Harris:
The latest chapter in the encoding saga

closes #7621 
closes #7604 
closes #7554
closes #7567
closes #7573
closes #7563
closes #7577

TODO:

* [x] documentation
* [x] allow escape sequence next to a parameter
* [x] validate escape sequences
* [ ] ~~test surrogate pairs~~ on second thoughts we don't need to use surrogate pairs, we can just support `[u+10ffff]` directly

### Tests

### Changesets


packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -102,14 +102,45 @@ function create_routes_and_nodes(cwd, config, fallback) {
...
 				throw new Error(`Invalid route ${id} — parameters must be separated`);
 			}
 
 			if (count_occurrences('[', id) !== count_occurrences(']', id)) {
 				throw new Error(`Invalid route ${id} — brackets are unbalanced`);
 			}
 
+			if (/#/.test(segment)) {
+				// Vite will barf on files with # in them
+				throw new Error(`Route ${id} should be renamed ${id.replace(/#/g, '&num;')}`);
```
dummdidumm, line 141: needs to be updated

Rich-Harris: done

documentation/docs/30-advanced/10-advanced-routing.md
```diff
@@ -123,27 +123,33 @@ src/routes/[...catchall]/+page.svelte
...
+Some characters can't be used on the filesystem — `/` on Linux and Mac, `\ / : * ? " < > |` on Windows. The `#` character has special meaning in URLs, and the `[ ] ( )` characters have special meaning to SvelteKit, so these also can't be used directly as part of your route.
 
-```js
-// @filename: ambient.d.ts
-declare global {
-	const assert: {
-		equal: (a: any, b: any) => boolean;
-	};
-}
+To use these characters in your routes, you can use hexadecimal escape sequences:
```
dummdidumm, line 128: I would add a small note here that the format is `[x/u+code]`, and after the list possibly link to some code table where you can look up more of them

packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -102,14 +102,45 @@ function create_routes_and_nodes(cwd, config, fallback) {
 		 * @param {import('types').RouteData | null} parent
 		 */
 		const walk = (depth, id, segment, parent) => {
-			if (/\]\[/.test(id)) {
+			const unescaped = id.replace(/\[([ux])\+([^\]]+)\]/i, (match, type, code) => {
```
dummdidumm, line 105: there could be more than one
```suggestion
			const unescaped = id.replace(/\[([ux])\+([^\]]+)\]/gi, (match, type, code) => {
```

documentation/docs/30-advanced/10-advanced-routing.md
```diff
@@ -123,27 +123,33 @@ src/routes/[...catchall]/+page.svelte
...
+- `)` — `[x+29]`
 
-// @filename: index.js
-// ---cut---
-assert.equal(
-	decodeURIComponent('%40[username]'),
-	'@[username]'
-);
-```
+For example, to create a `/:-)` route, you would create a `src/routes/[x+3a]-[x+29]/+page.svelte` file.
```
benmccann, line 145: I think there are four character in the non-escaped version and three in the escaped version, so one is probably missing

packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -102,14 +102,45 @@ function create_routes_and_nodes(cwd, config, fallback) {
...
 				throw new Error(`Invalid route ${id} — parameters must be separated`);
 			}
 
 			if (count_occurrences('[', id) !== count_occurrences(']', id)) {
 				throw new Error(`Invalid route ${id} — brackets are unbalanced`);
 			}
 
+			if (/#/.test(segment)) {
+				// Vite will barf on files with # in them
+				throw new Error(`Route ${id} should be renamed ${id.replace(/#/g, '[x+23]')}`);
```
benmccann, line 141:
```suggestion
				throw new Error(`Route ${id} should be renamed to ${id.replace(/#/g, '[x+23]')}`);
```

documentation/docs/30-advanced/10-advanced-routing.md
Rich-Harris: I don't follow? The path segment is `:-)`, which is represented by `[x+3a]-[x+29]`

packages/kit/src/utils/routing.js
```diff
@@ -1,6 +1,8 @@
 const param_pattern = /^(\[)?(\.\.\.)?(\w+)(?:=(\w+))?(\])?$/;
 
-/** @param {string} id */
+/**
+ * @param {string} id
```
benmccann, line 4: wouldn't hurt to have a jsdoc on this one
```suggestion
 * Creates the regex pattern, extracts parameter names, and generates types for a route
 * @param {string} id
```

benmccann:
For the pages like `escape-sequences/[u+82d7]/+page.svelte`, could we put the unescaped character like `苗` in the contents of the page? That would help me a lot in reading the code

documentation/docs/30-advanced/10-advanced-routing.md
benmccann: oh, I thought `/` was part of the route ID (I think it was in the last PR). Maybe we could change it then:
```suggestion
For example, to create a `:-)` route, you would create a `src/routes/[x+3a]-[x+29]/+page.svelte` file.
```

.changeset/smooth-years-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] allow hex/unicode escape sequences in routes
```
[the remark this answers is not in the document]
Rich-Harris: the trouble with automigrating something like this is that maybe someone wants a literal `%23` in their route... we wouldn't be able to tell which ones were intentional and which ones weren't. migration guide makes sense though

documentation/docs/30-advanced/10-advanced-routing.md
Rich-Harris: `/` _is_ part of the route ID — all route IDs are prefixed with `/`, we're fairly consistent about that 
image

maybe if it was `/smileys/:-)` or `/emoticons/:-)`?

documentation/docs/30-advanced/10-advanced-routing.md
Rich-Harris: added some docs

documentation/docs/30-advanced/10-advanced-routing.md
Rich-Harris: I changed it to `/smileys/:-)`

documentation/docs/30-advanced/10-advanced-routing.md
```diff
@@ -123,27 +123,41 @@ src/routes/[...catchall]/+page.svelte
...
-declare global {
-	const assert: {
-		equal: (a: any, b: any) => boolean;
-	};
-}
+':'.charCodeAt(0).toString(16); // '3a', hence '[x+3a]'
+```
 
-export {};
+You can also use Unicode escape sequences. Generally you won't need to as you can use the unencoded character directly, but if — for some reason — you can't use emoji in filenames, for example, then you can use the escaped characters. In other words, these are equivalent:
```
Conduitry, line 153: The wording of this tripped me up for a second. "Emoji" is probably an _acceptable_ plural of "emoji" but I don't think that's universal enough to phrase this this way.

packages/kit/src/utils/routing.js
```diff
@@ -143,3 +149,20 @@ export function exec(match, { names, types, optional }, matchers) {
...
+
+/** @param {string} str */
+function escape(str) {
+	return (
+		str
+			.normalize()
+			// escape [ and ] before escaping other characters, since they are used in the replacements
+			.replace(/[[\]]/g, '\\$&')
+			// replace %, /, ? and # with their encoded versions
+			.replace(/%/g, '%25')
```
benmccann, line 161: I'm a bit suspicious about this line and suspect perhaps it should be deleted. Can we create a test that checks `%` as a route id?

documentation/docs/30-advanced/10-advanced-routing.md
```diff
@@ -123,27 +123,41 @@ src/routes/[...catchall]/+page.svelte
...
-	decodeURIComponent('%40[username]'),
-	'@[username]'
-);
+```
+src/routes/[u+d83e][u+dd2a]/+page.svelte
+src/routes/🤪/+page.svelte
 ```
 
-To express a `%` character, use `%25`, otherwise the result will be malformed.
+The format for a Unicode escape sequence is `[u+nnnn]` where `nnnn` is a valid value between `0000` and `10ffff`. (Unlike JavaScript string escaping, there's no need to use surrogate pairs to represent code points above `ffff`.) To learn more about Unicode encodings, consult [Programming with Unicode](https://unicodebook.readthedocs.io/unicode_encodings.html).
```
Conduitry, line 160: Glossing over the `[u+nnnn]`/`[u+nnnnn]`/`[u+nnnnnn]` thing here definitely makes this easier to read, but is this correct enough? Perfectly willing to be overruled (especially since I don't have a specific alternate suggestion) if we think this is the right balance of accuracy and readability, when combined with linking to this external resource.

packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -102,14 +102,45 @@ function create_routes_and_nodes(cwd, config, fallback) {
...
+				} else {
+					if (code.length < 4 || code.length > 6) {
+						throw new Error(
+							`Unicode escape sequence in ${id} should be between four and six characters`
+						);
+					}
+
+					return String.fromCharCode(parseInt(code, 16));
+				}
+			});
```
Conduitry, line 129: "should" sounds like what we'd say in a warning, not an error. I'd suggest using "must" instead, like we are doing in the error just below this.

documentation/docs/30-advanced/10-advanced-routing.md
Rich-Harris: Huh. Wikipedia says this...

> An emoji ([/ɪˈmoʊdʒiː/](https://en.wikipedia.org/wiki/Help:IPA/English) [i-MOH-jee](https://en.wikipedia.org/wiki/Help:Pronunciation_respelling_key); plural emoji or emojis[[1]](https://en.wikipedia.org/wiki/Emoji#cite_note-1)) is a [pictogram](https://en.wikipedia.org/wiki/Pictogram), [logogram](https://en.wikipedia.org/wiki/Logogram), [ideogram](https://en.wikipedia.org/wiki/Ideogram) or [smiley](https://en.wikipedia.org/wiki/Smiley) embedded in text and used in electronic messages and [web pages](https://en.wikipedia.org/wiki/Web_page). The primary function of **emoji** is to fill in emotional cues otherwise missing from typed conversation.[[2]](https://en.wikipedia.org/wiki/Emoji#cite_note-2) Examples of **emoji** are 😂, 😃, 🧘🏻‍♂️, 🌍, 🌦️, 🍞, 🚗, 📞, 🎉, ❤️, 🍆, 🍑 and 🏁. **Emoji** exist in various genres, including facial expressions, common objects, places and types of weather, and animals.

...but Buzzfeed says this:

image

I guess this is just one of those cases where I will just have to clench my fist, smile, and pretend I don't know any better, like when people order 'a cannoli' or 'a panini'

Conduitry: You could rework the sentence to skirt the first-clenching. "but if - for some reason - you can't use a filename with an emoji in it" or something similar.

documentation/docs/30-advanced/10-advanced-routing.md
Rich-Harris: Even if we changed that it would still be incorrect, because not every integer between 0 and 0x10ffff is a valid code point. (For that inaccuracy to matter, someone would have to be making up code points at random rather than trying to represent a specific character.) We have to draw the line somewhere (would be a length digression if we were to be completely rigorous about all this), and I think this is _probably_ a reasonable place to draw it, but I am certainly open to persuasion here

documentation/docs/30-advanced/10-advanced-routing.md
Conduitry: Also is your implication here that people should be saying "an emojo"

packages/kit/src/utils/routing.js
Rich-Harris: it's definitely necessary. added a test, and also added it to the list of documented escape codes which was an oversight previously

packages/kit/src/core/sync/create_manifest_data/index.js
Rich-Harris: fixed

documentation/docs/30-advanced/10-advanced-routing.md
Conduitry: Nope, that's all fair, just wanted to bring it up. Resolving.

packages/kit/src/utils/routing.js
benmccann: thanks. I guess I'll trust the fact that the test is passing, but I must admit I'm confused by this line still. The three lines after this make sense to me because `decodeURI` will leave those escape sequences untouched. However, it will turn `%25` into `%` which we'll then match against. So if we're matching against just `%` then it shouldn't be `%25` in the regex. Where did I go wrong? I'm wondering if there's something special with `%` that deserves an extra comment here

benmccann: ohhhh. I just figured out what is going on here. https://github.com/sveltejs/kit/pull/7521 special cased `%25`: https://github.com/sveltejs/kit/blob/7a7abbe34713813d3771aac21adb5ae8d1e66336/packages/kit/src/utils/url.js#L62

that's a pretty weird implementation. I suspect it might break `%2525`. If we leave it like this then we should probably add a test for that case. But I would personally instead change that implementation in `url.js` to something like this:
```
/**
 * @param {string} pathname
 */
export function decode_pathname(pathname) {
	// replace / with ? because / will screw up regex
	// there can't be a ? in the path so use it as a safe character and swap it back in decode_params
	return pathname.split('/').map(segment => decodeURIComponent(segment).replace(/\//g, '?')).join('/');
}

/** @param {Record<string, string>} params */
export function decode_params(params) {
	for (const key in params) {
		params[key] = params[key].replace(/\?/g, '/');
	}
	return params;
}
```

benmccann: hmm. maybe the current implementation is fine actually. but i'll suggest an update to the comment
```

File: packages/kit/src/utils/routing.js [hunk 12]
```diff
@@ -143,3 +149,20 @@ export function exec(match, { names, types, optional }, matchers) {
...
 }
+
+/** @param {string} str */
+function escape(str) {
+	return (
+		str
+			.normalize()
+			// escape [ and ] before escaping other characters, since they are used in the replacements
+			.replace(/[[\]]/g, '\\$&')
+			// replace %, /, ? and # with their encoded versions
```
benmccann, line 160:
```suggestion
			// replace %, /, ? and # with their encoded versions because decode_pathname leaves them untouched
```

benmccann: also, would you mind updating the JSDoc for `decode_pathname`. it says something like "to prevent double decoding", but should really just say "to prevent decoding"

Rich-Harris: hang on, do we need `decode_pathname`? can't we just do `decodeURI`?

benmccann: oh, yeah, I think you're probably right :laughing:

benmccann: probably it didn't work because of Vite

Rich-Harris: it _does_ prevent double decoding though — it stops `/%2523` from turning into `/#` instead of `%23`, which is what would happen if we did `decodeURI` followed by `decode_params` instead of `decode_pathname` followed by `decode_params`

PR #7647, opened 2022-11-14
[fix] remove global variable from test to fix flakiness

benmccann:
one down. several to go...


packages/kit/test/apps/basics/src/routes/errors/error-html/make-root-fail/+server.js
```diff
@@ -1,9 +1,7 @@
-import { set_should_fail } from '../../../+layout.server';
-
 /**
  * @type {import('./$types').RequestHandler} param0
  */
-export function GET({ url }) {
-	set_should_fail(url.searchParams.get('type'));
+export function GET({ cookies, url }) {
+	cookies.set('fail-type', url.searchParams.get('type'));
```
dummdidumm, line 5: Haven't tested, but we probably need to set the path attribute here
```suggestion
	cookies.set('fail-type', url.searchParams.get('type'), { path: '/' });
```

dummdidumm (commented):
How does this reduce flakyness due to race conditions. What's different to a global variable and a cookie? Is it that we assume the browser instances are isolated so that their cookie values don't conflate?

dominikg:
> How does this reduce flakyness due to race conditions. What's different to a global variable and a cookie? Is it that we assume the browser instances are isolated so that their cookie values don't conflate?

playwright isolates page in different tests https://playwright.dev/docs/browser-contexts

Conduitry:
Yeah, what Dominik said. Each test gets its own browser context (and thus its own cookies). By avoiding this global state on the server, we should be able to have these tests run in parallel against a single instance of the app being tested.

packages/kit/test/apps/basics/test/client.test.js
```diff
@@ -774,10 +774,11 @@ test.describe('Routing', () => {
 	});
 
 	test('does not normalize external path', async ({ page }) => {
+		/** @type {string[] } */
 		const urls = [];
 
 		const { port, close } = await start_server((req, res) => {
-			if (req.url !== '/favicon.ico') urls.push(req.url);
+			if (req.url && req.url !== '/favicon.ico') urls.push(req.url);
```
Rich-Harris, line 781: why do we need this? i don't believe `req.url` can ever be undefined, but even if it was then `undefined !== '/favicon.ico` is perfectly fine

packages/kit/test/apps/basics/src/routes/errors/error-html/make-root-fail/+server.js
Rich-Harris: I suspect it will work because implicitly, the path is `/errors/error-html` which is the path we visit when the cookie gets read, but there's no harm in making it apply globally

packages/kit/test/apps/basics/src/routes/+layout.server.js
```diff
@@ -1,24 +1,15 @@
...
-export function set_should_fail(value) {
-	should_fail = value;
-}
-
 /** @type {import('./$types').LayoutServerLoad} */
-export async function load() {
+export async function load({ cookies }) {
+	const should_fail = cookies.get('fail-type');
 	if (should_fail) {
+		cookies.delete('fail-type');
```
Conduitry, line 7: Should we have a `path: '/'` here as well if we're creating the cookie with one?

Rich-Harris: yes

```suggestion
		cookies.delete('fail-type', { path: '/' });
```

packages/kit/test/apps/basics/test/client.test.js
benmccann: it was to make `urls` have type `string[]`. alternatively, we could change `urls` to `Array<string | undefined>`

packages/kit/test/apps/basics/test/client.test.js
```diff
@@ -774,10 +774,11 @@ test.describe('Routing', () => {
 	});
 
 	test('does not normalize external path', async ({ page }) => {
+		/** @type {string[] } */
```
benmccann, line 777:
```suggestion
		/** @type {string[]} */
```

packages/kit/test/apps/basics/test/client.test.js
benmccann: I made that change to use `Array<string | undefined>` instead

Rich-Harris (approved):

Issue #7641, opened 2022-11-14
Adapter-node creates double slash in manifest.js file paths when building


xpertekShaun:
### Describe the bug

https://github.com/sveltejs/kit/blob/master/packages/adapter-node/index.js#L50 sets the relative path to `'./'` which is then prepended to the list of manifest.js files

![image](https://user-images.githubusercontent.com/78720069/201675358-3d4ce04d-5223-4699-9e23-5fc7b06c4afc.png)

This leads to errors during building like the following

![image](https://user-images.githubusercontent.com/78720069/201675569-518a9461-c85d-4a19-8829-74cfa3def332.png)

I have seen weird behaviour, as building on windows appears to be ok, but linux and github CI both crash 

### Reproduction

1. `npm create svelte@latest my-app`
2. choose all defaults
3. `cd my-app`
4. `npm i -D @sveltejs/adapter-node`
5. `npm install`
6. Then change svelte.config.js from `adapter-auto` to `adapter-node`
7. `npm run build`
8. inspect `.svelte-kit/adapter-node/manifest.js`

### Logs

```Shell
> Using @sveltejs/adapter-node
error during build:
Error: Could not resolve './/entries/matchers/Actor.js' from .svelte-kit/adapter-node/manifest.js
    at error (file:///SolutionFolder/node_modules/rollup/dist/es/shared/rollup.js:1858:30)
    at ModuleLoader.handleResolveId (file:///SolutionFolder/node_modules/rollup/dist/es/shared/rollup.js:22156:24)
    at ModuleLoader.resolveDynamicImport (file:///SolutionFolder/node_modules/rollup/dist/es/shared/rollup.js:22210:120)
    at async file:///SolutionFolder/node_modules/rollup/dist/es/shared/rollup.js:22105:32
The command '/bin/sh -c npm run build' returned a non-zero code: 1

Error: Process completed with exit code 1.
```

### System Info

```Shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 AMD Ryzen 5 5600U with Radeon Graphics
    Memory: 3.14 GB / 15.31 GB
  Binaries:
    Node: 16.15.1 - ~\.nvm\versions\node\v16.15.1\bin\node.EXE
    npm: 8.11.0 - ~\.nvm\versions\node\v16.15.1\bin\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (107.0.1418.42)
[...]

PR #7653, opened 2022-11-14
[fix] safely join url segments in manifest

benmccann:
fixes https://github.com/sveltejs/kit/issues/7641

### DIFF
--- a/packages/kit/src/core/generate_manifest/index.js
+++ b/packages/kit/src/core/generate_manifest/index.js
@@ -37,7 +37,7 @@ export function generate_manifest({ build_data, relative_path, routes, format = 
 		build_data.manifest_data.nodes.map((_, i) => {
 			if (used_nodes.has(i)) {
 				reindexed.set(i, reindexed.size);
-				return `${relative_path}/nodes/${i}.js`;
+				return joinUrlSegments(relative_path, `/nodes/${i}.js`);
 			}
 		})
 	);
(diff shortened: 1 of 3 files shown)


packages/kit/src/core/generate_manifest/index.js
```diff
@@ -111,14 +111,32 @@ export function generate_manifest({ build_data, relative_path, routes, format =
...
 		}
 	}`.replace(/^\t/gm, '');
 }
+
+/**
+ * @param {string} a
+ * @param {string} b
+ * @returns
+ */
+export function joinUrlSegments(a, b) {
```
Conduitry, line 131: This is an internal API, so it should use snake case. I also don't think it needs to be exported from this module.

benmccann: yep, thanks. I swiped this from Vite and forgot to update it :smile:

Rich-Harris:
Can't we just change `relativePath: './'` to `relativePath: '.'`? I'm not a fan of loosey goosey functions that let you be careless about inputs

Rich-Harris:
https://github.com/sveltejs/kit/pull/7655

benmccann:
I tend to think along the lines of be liberal in what you accept and conservative in what you send: https://en.wikipedia.org/wiki/Robustness_principle

That being said, either one works, so I don't feel strongly

dummdidumm (approved):

PR #7660, opened 2022-11-15
Ignore Vite timestamp files by default in `create-svelte` templates (added to `.gitignore`)

Tal500:
Because why new SvelteKit users need to figure this out by themselves?

### Tests

### Changesets

### DIFF
--- a/.changeset/clean-grapes-think.md
+++ b/.changeset/clean-grapes-think.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Ignore Vite timestamp files by default in `create-svelte` templates (added to `.gitignore`)
--- a/packages/create-svelte/templates/default/.gitignore
+++ b/packages/create-svelte/templates/default/.gitignore
@@ -8,3 +8,5 @@ node_modules
 !.env.example
 .vercel
 .output
+vite.config.js.timestamp-*
+vite.config.ts.timestamp-*
--- a/packages/create-svelte/templates/skeleton/.gitignore
+++ b/packages/create-svelte/templates/skeleton/.gitignore
@@ -6,3 +6,5 @@ node_modules
 .env
 .env.*
 !.env.example
+vite.config.js.timestamp-*
+vite.config.ts.timestamp-*
--- a/packages/create-svelte/templates/skeletonlib/.gitignore
+++ b/packages/create-svelte/templates/skeletonlib/.gitignore
@@ -6,3 +6,5 @@ node_modules
 .env
 .env.*
 !.env.example
+vite.config.js.timestamp-*
+vite.config.ts.timestamp-*


dummdidumm (commented):
I see these from time to time - what are these files about, anyway? FWIW they aren't ignored by the Vite starter templates either, so I'm hesitant to put them in there.

bluwy:
Vite prebundles the config files with esbuild by default and writes them to disk to `import()` it. Usually it's removed after it's evaluated so it's only temporary, so I don't think we need to gitignore them (?)

dominikg:
vite creates a micro bundle of vite config to load it. thats it. they can sometimes be left as residue if the process crashes  / terminates at the wrong moment. ignoring them is the right call and should be done in vite too cc @bluwy

Tal500:
> Vite bundles the config files with esbuild by default and writes them to disk to `import()` it. Usually it's removed after it's evaluated so it's only temporary, so I don't think we need to gitignore them (?)

Sometimes you cancel vite build process, and then they stuck there, and you may commit them by mistake (it happened also here in this repo!)
Both SvelteKit repo and my personal repos decided to ignore this files in `.gitignore`, so I just thought it is a consensus to ignore them.

Additional context of Vite: https://github.com/vitejs/vite/issues/9470#issuecomment-1200451057

dominikg:
alternatively vite could instead put that file into `node_modules/.vite/config.timestamp-xxx.mjs` that is already ignored and the dir is used by vite for other generated things

bluwy:
Writing to a different directory would require rewriting the import paths. I think Vite should make sure the config files are removed instead on process exit to make it transparent.

dominikg:
messing with process.onExit at least used to have implications for signal handling and it only lessens the impact. you could still end up with it staged when you run `git add .` at the wrong point in time.  If moving it to an ignored directory is not an option, these patterns here are the next best solution :thinking:

bluwy:
The issue is that we don't want to leak an implementation detail into a user file though, i think deleting on process exit is probably good enough in most case to prevent another point of configuration.

Rich-Harris:
I guess the question is whether this change is likely to land in Vite any time soon? I've definitely been bitten by this in the past, adding a couple of lines to a `.gitignore` feels fairly harmless, but if we expect it to make it into a patch release in the next few days then it's probably not worth it

Rich-Harris:
IIUC the timestamped file is the _output_ of esbuild rather than the input — it's the thing that Vite imports.

I'm personally of the view that esbuild shouldn't be involved at all here — just require `vite.config.js` to be a normal module that adheres to normal module rules, avoiding this whole song and dance (and preventing the spurious warnings that appear when you first start a SvelteKit project before the types have been generated), but I don't expect the Vite team to agree with me :)

Rich-Harris:
Added a note to the Vite 4 discussion proposing removing the config bundling step https://github.com/vitejs/vite/discussions/10570#discussioncomment-4148194

bluwy:
I tried to make a fix to clean the files but nodejs makes it incredibly hard to detect process exit without caveats ☹️ I would also prefer not bundling them in most case, but there will be tradeoffs and it would be a big change that we might want to defer for Vite 5. We'll be discussing this more in the next Vite team meeting though.

> Can't you emulate them and pass the content of a virtual file to ESBuild instead of writing to the disk? Is this possible in ESBuild?

We are writing the result of the esbuild output to disk for nodejs to evaluate it.

dummdidumm (approved):
If the change is non-trivial then I'm ok with adding these to ignore

Issue #7640, opened 2022-11-14
adapter-cloudflare: only including `_app` content in `_routes.json`


userquin:
### Describe the bug

The adapter should include any assets inside the `client` output folder instead `_app` only (`kit.appDir`).

Latest changes in kit allow plugins to emit/generate assets in the client folder: `@vite-pwa/sveltekit`, `sitemap`...

The problem is here: https://github.com/sveltejs/kit/blob/master/packages/adapter-cloudflare/index.js#L42 (should include the client folder)

### Reproduction

~~no, sorry, just reported from Discord here: https://discord.com/channels/937808017016119440/937973377883336704/1041084393239891988~~

https://github.com/hazre/neosvr-mod-manager/tree/fix/pwa

### System Info

```Shell
NA
```

### Severity

blocking all usage of SvelteKit

### Additional Information

The app is using the new `@vite-pwa/sveltekit` plugin to add the PWA, cloudflare returns 404 for the service worker and webmanifest assets.

userquin:
Related kit changes:
- https://github.com/sveltejs/kit/pull/5620
- https://github.com/sveltejs/kit/pull/5601#issuecomment-1188391404

hazre:
Here's a list of the cloudflare folder. Assets are definitely there and they work fine locally with `pnpm run preview` but when I deploy it to cloudflare pages, I get a 404s, which is weird.

https://gist.github.com/hazre/1ebf72e8f43f0252822076aa6c33b3c6

benmccann:
@jrf0110 tagging you in case you or anyone else from Cloudflare might be interested in giving this one a look. We no longer have any Svelte maintainers employed at Cloudflare and I don't have a Cloudflare account so couldn't be very confident in any attempt I made

tedsteen:
When deploying to Cloudflare with kit `1.0.0-next.544` I don't get any functions deployed. do I understand correctly that this issue related to that?

jrf0110:
@benmccann I'll get some 👀 on this today! Thanks for the ping

Rich-Harris:
@tedsteen that sounds like a separate issue

Rich-Harris:
started looking at this, wrote up notes in https://github.com/sveltejs/kit/pull/7667

jrf0110:
Just to clarify, the `_routes.json` file is an optimization.  Missing entries _should not_ cause any assets to 404

userquin:
[...]

PR #7667, opened 2022-11-15
pass asset list to adapters

Rich-Harris:
had a quick look at #7640 to see if there's anything needed from the framework itself. i'm a little unclear on why requests are 404ing — it seems that the worker should be able to serve the assets, even if doing so is wasteful.

regardless, if we're not going to add all of `written_files` to the `exclude` array, then we probably need to pass the `assets` list (which contains all the files in the `static` folder) so that we can figure out which files were generated? not sure about any of this, so just going to leave this here in draft form for now


Rich-Harris:
going to close this until there's more clarity on requirements

Issue #7374, opened 2022-10-24
Add ability to disable Node polyfills


pzuraq:
### Describe the problem

We've been using Mock Service Worker to mock out requests to our backend. We use the Node interceptor for it so that the mocks work in both SSR and on the client. However, when we recently started updating our SvelteKit version, we noticed that this stopped working. We think that this is due to the Undici polyfill, which appears to completely avoid standard global Node modules like `http` and `https`.

In addition, the error logs for these mocked libraries and _significantly_ harder to read and debug, as they use a few tricks for scheduling that make the original call site difficult to find.

### Describe the proposed solution

Modern versions of Node (e.g. Node 18) have all of the mocked libraries available as globals by default, so there is no need to mock them. I propose either adding an option to disable the polyfills manually, or ignoring the polyfill if a global version exists already.

### Alternatives considered

None

### Importance

would make my life easier

Rich-Harris:
Re-opening — #7668 caused #7673

Rich-Harris:
(we should add a multipart test before attempting this again)

Rich-Harris:
After discussion, we agreed to add a `polyfill: false` option (or similar) to `adapter-node`, so that the polyfills can be excluded from server code altogether where applicable.

Rich-Harris:
Removing from the 1.0 milestone since this update won't affect Kit itself

bluwy:
> For reference, Node 18 has been available since April, 2022. However, [AWS Lambda does not yet support Node 18](https://github.com/aws/aws-lambda-base-images/issues/47). Based on past history, it usually takes them about a year to add support.

Looks like AWS Lambda now [supports node 18](https://aws.amazon.com/about-aws/whats-new/2022/11/aws-lambda-support-node-js-18/)

benmccann:
I just tried Netlify. It still uses Node 16 by default, but does support Node 18.

chronicIntrovert:
> After discussion, we agreed to add a `polyfill: false` option (or similar) to `adapter-node`, so that the polyfills can be excluded from server code altogether where applicable.

We are using `adapter-cloudflare-workers` where this is happening - can we also have an option to disable polyfills there?

Rich-Harris:
`adapter-cloudflare-workers` doesn't add polyfills. You must be experiencing something else

PR #7668, opened 2022-11-15
only apply polyfills where necessary

Rich-Harris:
ref #7374. i think we tried this once before but had to undo it because of problems with Node 18, but let's give it another go and see what happens with the tests

### Tests

### Changesets

### DIFF
--- a/.changeset/many-rivers-hide.md
+++ b/.changeset/many-rivers-hide.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Only apply polyfills where necessary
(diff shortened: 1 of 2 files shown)


benmccann (approved):
looks like we originally added it here: https://github.com/sveltejs/kit/commit/7ed3c95e0d0160f0b53469ec7c187ca528c0ed88#diff-8cae4ccc562be888fea664ae92b47321887907a9d0839c25809da043d6359457

and then removed it here https://github.com/sveltejs/kit/pull/5292 because of https://github.com/sveltejs/kit/pull/5292#issuecomment-1170388162

since all the checks are passing I don't see a reason not to merge this

Issue #7654, opened 2022-11-14
Upgrade to TypeScript 4.9


benmccann:
### Describe the problem

TypeScript 4.9 comes out tomorrow

### Describe the proposed solution

Upgrade before 1.0

### Importance

nice to have

PR #7669, opened 2022-11-15
bump typescript

Rich-Harris:
#7654. We also need to

* [ ] update the TS conversion for the docs to use `satisfies` (in this PR, or in a follow-up?)
* [ ] use `satisfies` in the template

### Tests

### Changesets

### DIFF
--- a/.changeset/cool-seahorses-cross.md
+++ b/.changeset/cool-seahorses-cross.md
@@ -2,4 +2,4 @@
 'create-svelte': patch
 ---
 
-Bump TypeScript version
+Bump TypeScript version and use `satisfies` operator
--- a/.changeset/cool-seahorses-cross.md
+++ b/.changeset/cool-seahorses-cross.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Bump TypeScript version
--- a/packages/adapter-vercel/index.js
+++ b/packages/adapter-vercel/index.js
@@ -82,7 +82,11 @@ const redirects = {
 };
 
 /** @type {import('.').default} **/
-export default function ({ external = [], edge, split } = {}) {
+export default function ({
+	external = /** @type {string[]} */ ([]),
+	edge = false,
+	split = false
+} = {}) {
 	return {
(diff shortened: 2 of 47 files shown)


.changeset/cool-seahorses-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Bump TypeScript version
```
Rich-Harris, line 5:
```suggestion
Bump TypeScript version and use `satisfies` operator
```

benmccann:
I filed an issue in the sucrase repo: https://github.com/alangpierce/sucrase/issues/763

dummdidumm:
Re "use satisfies in the docs" - I would maybe wait 1-2 months before doing that, so people can familiarize themselves more with that. Also, VS Code is likely not on TS 4.9 yet so everyone not turning on "use TS workspace version" would see type errors and be confused.

dummdidumm:
Another blocker: `prettier` doesn't support that syntax yet

dummdidumm:
Bummer, `function foo () {} satisfies Bar` is invalid syntax, so we need to keep using `const foo = (() => {}) satisfies Bar` instead.

benmccann:
It looks like prettier has already merged support and it will be included in the upcoming 2.8 release (https://github.com/prettier/prettier/issues/13792)

There's an open issue for `typescript-eslint`: https://github.com/typescript-eslint/typescript-eslint/issues/5688

packages/adapter-vercel/index.js
```diff
@@ -82,7 +82,11 @@ const redirects = {
 };
 
 /** @type {import('.').default} **/
-export default function ({ external = [], edge, split } = {}) {
+export default function ({
```
benmccann, line 85: don't need to duplicate the types here. just declare the function as a `const` and then export it later

benmccann:
`master` is now using TypeScript 4.9, so most of this PR is obsolete. I'm going to close it in favor of https://github.com/sveltejs/kit/pull/7861, which extracts out the parts that haven't already been committed to master

Issue #7673, opened 2022-11-16
All POSTs with enhanced forms throw error 'multipart/form-data not supported'


hanszoons:
### Describe the bug

Latest sveltekit update breaks all enhanced forms. All POSTs with enhanced forms results in a 500 error.

Downgrading to @sveltejs/kit@1.0.0-next.546 reverses this problem.

Maybe something to do with #7494 ?

### Reproduction

* Install Demo app via npm create svelte@latest
* run dev, try to post in Sverdle

### Logs

```Shell
multipart/form-data not supported
NotSupportedError: multipart/form-data not supported
    at Request.formData (node:internal/deps/undici/undici:2383:19)
    at enter (/src/routes/sverdle/+page.server.js:57:30)
    at call_action (file:///Users/hans/dev/test-sveltekit-bug/node_modules/@sveltejs/kit/src/runtime/server/page/actions.js:204:9)
    at handle_action_json_request (file:///Users/hans/dev/test-sveltekit-bug/node_modules/@sveltejs/kit/src/runtime/server/page/actions.js:42:22)
    at render_page (file:///Users/hans/dev/test-sveltekit-bug/node_modules/@sveltejs/kit/src/runtime/server/page/index.js:44:11)
    at async resolve (file:///Users/hans/dev/test-sveltekit-bug/node_modules/@sveltejs/kit/src/runtime/server/index.js:241:17)
    at async respond (file:///Users/hans/dev/test-sveltekit-bug/node_modules/@sveltejs/kit/src/runtime/server/index.js:292:20)
    at async file:///Users/hans/dev/test-sveltekit-bug/node_modules/@sveltejs/kit/src/exports/vite/dev/index.js:413:22
```

### System Info

```Shell
System:
    OS: macOS 13.0.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 2.04 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.9.0 - ~/.nvm/versions/node/v18.9.0/bin/node
    npm: 8.19.1 - ~/.nvm/versions/node/v18.9.0/bin/npm
  Browsers:
    Chrome: 107.0.5304.110
    Firefox: 105.0.2
    Safari: 16.1
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.88 
    @sveltejs/kit: next => 1.0.0-next.548 
    svelte: ^3.46.0 => 3.53.1 
[...]

PR #7674, opened 2022-11-16
Revert "only apply polyfills where necessary"

Rich-Harris:
Reverts sveltejs/kit#7668, as it caused #7673 (since Node 18 apparently has an old version of undici)


Rich-Harris:
how do you undelete a file (i.e. the changeset) in a PR? do we just have to recreate it?

dummdidumm:
yeah, recreating is the best option. And then another changeset for the revert 😄

PR #7678, opened 2022-11-16
Add double check for 'keepFocus' and 'noScroll' in goto to allow for backcompatibity

paoloricciuti:
It's a very simple change. It avoids throwing an error when keepfocus or noscroll it's in the opts of goto if there are also the correct case inside the opts. This should allow for back compatibility for library writers because they can pass both.

I wasn't able to run test for this.

---

### Tests

### Changesets

### DIFF
--- a/.changeset/hot-keys-lick.md
+++ b/.changeset/hot-keys-lick.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/kit": patch
+---
+
+Ignore presence of `keepfocus` and `noscroll` in `goto` options if correctly-cased options are also present


dummdidumm:
Which library depends on this and wants to be able to use both? Why doesn't the library do a major version bump (or a minor if in prerelease mode)?

Conduitry:
This change doesn't make sense to me. Why only throw an error if both the old and new form are set? Wouldn't this mean that someone who still only has `nofocus` set would no longer get an error but also their code would silently not work they way they were expecting?

paoloricciuti:
> I don't know whether you did something with line endings, but every single line appears as changed in Git.

My bad it was line endings, i fixed it now

> Which library depends on this and wants to be able to use both? Why doesn't the library do a major version bump (or a minor if in prerelease mode)?

I have a library that i've already fixed with the new case and bumped the version. But this way if an user uses my library (or some other library that want to mantain backcompatibility) they can use it either if they use a newer version of kit or an older one. And it also seems much more reasonable to check if both case exist instead of just one.

paoloricciuti:
> This change doesn't make sense to me. Why only throw an error if both the old and new form are set? Wouldn't this mean that someone who still only has `nofocus` set would no longer get an error but also their code would silently not work they way they were expecting?

It'll throw if there's no keepFocus AND there's keepfocus instead of throwing only if there's keepfocus. Basically if you have both it will not throw. If you have keepfocus without keepFocus it will throw.

Conduitry:
Oh, I missed the `!`. That's not the right syntax, though. `!` is higher precedence than `in`, so you need some parentheses. Have you actually tried this change?

packages/kit/src/runtime/client/client.js
```diff
@@ -1202,13 +1202,13 @@ export function create_client({ target, base, trailing_slash }) {
...
 			// TODO remove for 1.0
-			if ('keepfocus' in opts) {
+			if (!'keepFocus' in opts && 'keepfocus' in opts) {
 				throw new Error(
 					'`keepfocus` has been renamed to `keepFocus` (note the difference in casing)'
 				);
 			}
 
-			if ('noscroll' in opts) {
+			if (!'noScroll' in opts && 'noscroll' in opts) {
```
Rich-Harris:
```suggestion
			if ('noscroll' in opts && !('noScroll' in opts)) {
```

Rich-Harris (changes requested):
this addresses @Conduitry's response and fixes the typechecking failure

packages/kit/src/runtime/client/client.js
paoloricciuti: Yeah i was already fixing this (while also screwing line endings again) -.-

paoloricciuti:
> please fix the indentation. the diff should be 2 lines, not 1,726!

I finally did it...sorry.

packages/kit/src/runtime/client/client.js
```diff
@@ -1202,13 +1202,13 @@ export function create_client({ target, base, trailing_slash }) {
 
 		goto: (href, opts = {}) => {
 			// TODO remove for 1.0
-			if ('keepfocus' in opts) {
+			if (!('keepFocus' in opts) && ('keepfocus' in opts)) {
```
Rich-Harris, line 1205:
```suggestion
			if ('keepfocus' in opts && !('keepFocus' in opts)) {
```

packages/kit/src/runtime/client/client.js
```diff
@@ -1202,13 +1202,13 @@ export function create_client({ target, base, trailing_slash }) {
...
 				throw new Error(
 					'`keepfocus` has been renamed to `keepFocus` (note the difference in casing)'
 				);
 			}
 
-			if ('noscroll' in opts) {
+			if (!('noScroll' in opts) && ('noscroll' in opts)) {
```
Rich-Harris, line 1211:
```suggestion
			if ('noscroll' in opts && !('noScroll' in opts)) {
```

Rich-Harris (commented):
it's failing CI because you didn't run `pnpm format`

Rich-Harris (approved):

Rich-Harris:
we got there in the end — thanks!

paoloricciuti:
> we got there in the end — thanks!

Thank for your patience...did this rapidly with GitHub.dev but it got me in more trouble than cloning apparently lol

PR #7681, opened 2022-11-16
clarify merging semantics

### LINKED ISSUE #7677: Clarify 'last one wins' in load section of the documentation
### Describe the problem

In the load section of the [documentation](https://kit.svelte.dev/docs/load#layout-data) in the appendix it is stated that
> If multiple load functions return data with the same key, the last one 'wins'.

Please clarify what does that mean. It can either be the deepest load function in routing hierarchy or the last load function to resolve (Promises). This phrasing creates unnecessary ambiguity and it can be crucial for some use cases to know this clarification.

Include clarification in the documentation.

### Alternatives considered

None

### Importance

would make my life easier

### Additional Information

None

Rich-Harris:
closes #7677

### DIFF
--- a/documentation/docs/20-core-concepts/20-load.md
+++ b/documentation/docs/20-core-concepts/20-load.md
@@ -132,7 +132,7 @@ Data returned from layout `load` functions is available to child `+layout.svelte
 +{/if}
 ```
 
-> If multiple `load` functions return data with the same key, the last one 'wins'.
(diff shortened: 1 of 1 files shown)


Conduitry:
Do we want to talk about how multiple nested layouts inherit/override in this same way?

Rich-Harris:
I feel like it's pretty obvious once you have one example, no?

Conduitry:
I'll defer to you on that. I thought what we had before this PR was fine too.

Rich-Harris:
So did I

benmccann:
we don't necessarily need an example we could just replace "last" with something more specific like:

> the child overrides the parent (e.g. a page would override a layout)

Rich-Harris:
An example makes it concrete (for example what is meant by 'key'? with the example it's very obvious)

dummdidumm (approved):

PR #7682, opened 2022-11-16
allow .d.ts files to have + prefix

### LINKED ISSUE #7662: Router picks up .d.ts files
### Describe the bug

When using tooling such as graphql-code-generator 

### Reproduction

Create a file anywhere under `routes` called `+layout.server.d.ts`.

### Logs

_No response_

### System Info

```Shell
System:
    OS: macOS 12.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 171.77 MB / 32.00 GB
    Shell: 5.8 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 18.10.0 - /nix/store/m3za3mblwg9az7vpkkvlfm04z66mcr6y-nodejs-18.10.0/bin/node
    npm: 8.19.2 - /nix/store/m3za3mblwg9az7vpkkvlfm04z66mcr6y-nodejs-18.10.0/bin/npm
  Browsers:
    Brave Browser: 106.1.44.101
    Chrome: 107.0.5304.110
    Chrome Canary: 99.0.4828.0
    Firefox: 106.0.5
    Safari: 15.2
```

### Severity

annoyance

### Additional Information

I can think of four solutions:

1. The router should ignore `.d.ts` files by default. As long as TypeScript is considered a first-class citizen in SvelteKit this doesn't seem like an unreasonable solution to me.
  a. Another consideration: only allow specific extensions to be picked up by the router, maybe configured some way in Vite.
3. Configurable ignore for the router. This is more complex, but opens up end users to be able to exclude any generated files of various extensions.
4. Don't generate files with `+` at the start of their name. Unfortunately, I couldn't find a way to configure graphql code generator to do this, but I could either place all generated code in its own `.d.ts` file outside of `routes`, or, as suggested on Discord, I can place all queries in `queries.ts` files which would result in files such as `__generated__/queries.d.ts`, which should not be picked up by the router.

I'd be willing to put up a PR for the first solution if the SvelteKit team would be open to that solution.

Rich-Harris:
closes #7662

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/sync/create_manifest_data/index.js
+++ b/packages/kit/src/core/sync/create_manifest_data/index.js
@@ -192,7 +192,7 @@ function create_routes_and_nodes(cwd, config, fallback) {
 			// process files first
 			for (const file of files) {
 				if (file.is_dir) continue;
-				if (!file.name.startsWith('+')) continue;
+				if (!file.name.startsWith('+') || file.name.endsWith('.d.ts')) continue;
 				if (!valid_extensions.find((ext) => file.name.endsWith(ext))) continue;
 
 				const project_relative = posixify(path.relative(cwd, path.join(dir, file.name)));
--- a/packages/kit/src/core/sync/create_manifest_data/index.js
+++ b/packages/kit/src/core/sync/create_manifest_data/index.js
@@ -192,8 +192,22 @@ function create_routes_and_nodes(cwd, config, fallback) {
 			// process files first
 			for (const file of files) {
 				if (file.is_dir) continue;
-				if (!file.name.startsWith('+') || file.name.endsWith('.d.ts')) continue;
+				if (!file.name.startsWith('+')) continue;
 				if (!valid_extensions.find((ext) => file.name.endsWith(ext))) continue;
+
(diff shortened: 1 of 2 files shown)


dummdidumm (approved):
I'm good with this change, given that this isn't a runtime file

packages/kit/src/core/sync/create_manifest_data/index.js
```diff
@@ -192,7 +192,7 @@ function create_routes_and_nodes(cwd, config, fallback) {
 			// process files first
 			for (const file of files) {
 				if (file.is_dir) continue;
-				if (!file.name.startsWith('+')) continue;
+				if (!file.name.startsWith('+') || file.name.endsWith('.d.ts')) continue;
```
dummdidumm, line 195: mhhhhm or should we scope it to "file ends with `.d.ts` but has to have one of the known file names"? So that you can't do `+yolo.d.ts`

Rich-Harris: blurgh i guess we probably should

knpwrs:
Thank you for this!

PR #7685, opened 2022-11-16
[feat] warn if svelte not found in dependencies or peerDependencies

benmccann:
This allows `vite-plugin-svelte` to detect which libraries are Svelte libraries

### DIFF
--- a/.changeset/gentle-rocks-run.md
+++ b/.changeset/gentle-rocks-run.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/package': patch
+---
+
+[feat] warn if svelte not found in dependencies or peerDependencies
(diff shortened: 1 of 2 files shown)


Conduitry (changes requested):
We should be able to just do `!pkg.dependencies?.svelte && !pkg.peerDependencies?.svelte`, right? We don't need to iterate through all of the keys in the objects.

benmccann:
duh. thank you!

dummdidumm:
wow Github playing tough, @Conduitry you need to approve as well since you requested changes, else we can't merge 😅

benmccann:
doesn't stop you if you're an admin I guess :laughing:

PR #7690, opened 2022-11-16
[chore] bump pnpm

benmccann:
see if this clears the dumb cache


Rich-Harris:
wheeee that looks to have done the trick

Issue #7042, opened 2022-09-26
Custom Server ( express ) throwing CORS error


bobthered:
### Describe the bug

When using a custom node server, CORS errors are returned from form submissions to actions endpoint.

### Reproduction

[@bobthered/sveltekit-issue-custom-server-cors](https://github.com/bobthered/sveltekit-issue-custom-server-cors)

### System Info

```shell
System:
    OS: macOS Mojave 10.14.6
    CPU: (4) x64 Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz
    Memory: 996.74 MB / 8.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 16.15.1 - /usr/local/bin/node
    npm: 8.19.2 - /usr/local/bin/npm
  Browsers:
    Chrome: 105.0.5195.125
    Safari: 12.1.2
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.80 
    @sveltejs/adapter-node: ^1.0.0-next.96 => 1.0.0-next.96 
    @sveltejs/kit: next => 1.0.0-next.504 
    svelte: ^3.44.0 => 3.50.1 
    vite: ^3.1.0 => 3.1.3
```

### Severity

blocking all usage of SvelteKit

MCole-K9:
Don't have experience with custom server, nor am I understanding the use case in sveltekit(I would love learn tho ).  That aside, have you tried passing cors some configuration?

AiHKhaled:
did you try installing the cors package?

Rich-Harris:
I assume you're referring to this?

> Cross-site POST form submissions are forbidden

I suspect this is happening because you don't have an `ORIGIN` set: https://github.com/sveltejs/kit/tree/master/packages/adapter-node#environment-variables

PR #7693, opened 2022-11-16
warn if recommended env vars are missing

Rich-Harris:
#7042. When using `adapter-node`, if the `ORIGIN` header is missing, and `PROTOCOL_HEADER` and `HOST_HEADER` aren't both set, then request URLs are very likely to be incorrect. That will result in the CSRF check misfiring, for example.

I don't 100% know if this solution is appropriate but it's the best I've got

### Tests

### Changesets

### DIFF
--- a/.changeset/chilled-toes-report.md
+++ b/.changeset/chilled-toes-report.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/adapter-node': patch
+---
+
+Warn if ORIGIN/PROTOCOL_HEADER/HOST_HEADER are unset
--- a/packages/adapter-node/src/handler.js
+++ b/packages/adapter-node/src/handler.js
@@ -21,6 +21,12 @@ const body_size_limit = parseInt(env('BODY_SIZE_LIMIT', '524288'));
 
 const dir = path.dirname(fileURLToPath(import.meta.url));
 
+if (!origin && (!address_header || !host_header)) {
+	console.warn(
+		`You should specify either ${ENV_PREFIX}ORIGIN or both ${ENV_PREFIX}PROTOCOL_HEADER and ${ENV_PREFIX}HOST_HEADER so that adapter-node can reliably determine the request URL. See https://github.com/sveltejs/kit/tree/master/packages/adapter-node#environment-variables for more information`
+	);
+}
+
 /**
  * @param {string} path
  * @param {boolean} client


Conduitry:
Hm. In the absence of all of these, the adapter assumes that the origin is `https://` followed by whatever the `Host` header container. Do we feel like that's often not going to be good enough? I don't know that we want to be making people worry about setting these when (presumably) that's the case they're most likely to be in in production.

At work, I have a multitenant app (with an external API) and there I _can't_ set an fixed explicit `ORIGIN` env var, and so this sort of a warning message would encourage me to either (a) ignore it, or (b) jump through some hoops to find out what HTTP header might already be present and happen to equal `https` so that I can ultimately achieve the exact same result.

In this application, the only time I need to set `ORIGIN` is when running e2e tests, because they don't have an HTTPS proxy sitting in front of them then.

I would tend towards "document this more explicitly", but I don't know where to do that other than in https://github.com/sveltejs/kit/tree/master/packages/adapter-node#origin-protocol_header-and-host_header which I know a lot of people aren't going to look at.

Rich-Harris:
The other thought I had was to document it at the point of need, i.e. when the cross-site-but-not-really POST fails. But it only actually affects `adapter-node` users. I agree with all your points though

benmccann:
I don't quite understand why this would happen. It doesn't seem like it should be necessary to me because doesn't it just use the `host` header by default? 
https://github.com/sveltejs/kit/blob/172e792db86d481be397211d8236c700d458e320/packages/adapter-node/src/handler.js#L19

I guess you could also do it inside `adapter-node`'s `get_origin` function (probably just once for the app) so that it only logs it if you actually do something using the origin. I'm not sure if SvelteKit will always call that internally though making that equivalent to always doing it

Conduitry:
It does use the `Host` header by default, and it assumes HTTPS, which I think are good defaults. The question is whether we'd rather push people away from just using these defaults to try to avoid them getting bitten. I'd argue no. I think this warning would cause more chaos in the world than it removes. If we spruce up the Node adapter readme a tiny, we will at least have something to point to to tell people that this is the default and that's why their [whatever] is breaking in production.

benmccann (changes requested):
I'd agree this seems like it'd cause more trouble than it'd save. I think updating the readme to describe when you need to override the default would be better

Rich-Harris:
closing in favour of #7744

PR #7703, opened 2022-11-17
allow invalidate to be called with multiple resources

### LINKED ISSUE #7349: Minor improvement: allow passing an array of urls (strings, `URL`s or functions) to `invalidate`
### Describe the problem

I'm finding myself doing this often in my `enhance` "result" functions:

```ts
await Promise.all([
  invalidate('page:plans'),
  invalidate('layout:plans'),
  invalidate('layout:dashboard')
]);
```

This is fine, as long as I remember to wrap the `invalidate` calls in `Promise.all` rather than `await`ing each one, thus allowing the `__data.json` requests to run concurrently. 

I'm pretty sure there wouldn't be a huge performance hit in most cases, but still it's a point of uncertainty. I'd much rather be able to do this:

```ts
await invalidate([
  'page:plans',
  'layout:plans',
  'layout:dashboard'
]);
```

Currently:
```ts
export function invalidate(url: string | URL | ((url: URL) => boolean)): Promise<void>;
```

Improved:
```ts
// InvalidateArg for clarity...
type InvalidateArg = string | URL | ((url: URL) => boolean);
export function invalidate(url: InvalidateArg | InvalidateArg[]): Promise<void>;
```

I'm pretty sure this is doable here:

https://github.com/sveltejs/kit/blob/87b3037bb6fcf574d5d50c91f42cc91d3b97286c/packages/kit/src/runtime/client/client.js#L1145-L1161

### Alternatives considered

1. Write a wrapper:

```ts
type InvalidateArg = string | URL | ((url: URL) => boolean);
export const wrapInvalidate = async (arg: InvalidateArg | InvalidateArg[]): Promise<void> => {
  const args: InvalidateArg[] = Array.isArray(arg) ? arg : [arg];
  await Promise.all(args.map((a) => invalidate(a)));
};
```

2. Express the cache dependencies somehow in the page load function, therefore, on the client, only having to invalidate the page data rather than the page plus whatever layout data may have been affected. I have no idea whether this is possible, and it seems like it could be a mess/inefficient:

  - only `invalidate('page:plans')` on the client
  - `load` in `+page.server.ts` somehow knows to load & merge the parent layout data?

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
closes #7349 

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -1217,19 +1217,21 @@ export function create_client({ target, base, trailing_slash }) {
 			return goto(href, opts, []);
 		},
 
-		invalidate: (resource) => {
-			if (resource === undefined) {
+		invalidate: (...resources) => {
+			if (resources.length === 0) {
 				// TODO remove for 1.0
 				throw new Error(
 					'`invalidate()` (with no arguments) has been replaced by `invalidateAll()`'
 				);
 			}
 
-			if (typeof resource === 'function') {
-				invalidated.push(resource);
-			} else {
-				const { href } = new URL(resource, location.href);
-				invalidated.push((url) => url.href === href);
+			for (const resource of resources) {
+				if (typeof resource === 'function') {
+					invalidated.push(resource);
+				} else {
+					const { href } = new URL(resource, location.href);
+					invalidated.push((url) => url.href === href);
+				}
 			}
 
 			return invalidate();
--- a/.changeset/nervous-wombats-know.md
+++ b/.changeset/nervous-wombats-know.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Allow `invalidate(...)` to be called with multiple resources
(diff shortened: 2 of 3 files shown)


packages/kit/src/runtime/client/client.js
```diff
@@ -1217,19 +1217,21 @@ export function create_client({ target, base, trailing_slash }) {
 			return goto(href, opts, []);
 		},
 
-		invalidate: (resource) => {
-			if (resource === undefined) {
+		invalidate: (...resources) => {
```
dummdidumm, line 1220: With that we lose the ability to add an options object down the road, which was IIRC was one of the reasons for spreading out `invalidate` and `invalidateAll` - are we ok with that?

Rich-Harris: excellent point. let's revisit later if enough time goes by that we're confident we won't need to add options. i guess another option would be to allow arrays, but that would be at odds with `depends(...urls)` and generally less nice.

will add a note to the issue

PR #7704, opened 2022-11-17
[fix] don't preload fonts by default

dummdidumm:
Too many false positives when people use fallback fonts or load fonts depending on unicode ranges Also some explanation in the docs about what preload is called with

### Tests

### Changesets


benmccann:
seems like a good idea to me. I'd love for users to have the ability to preload assets as well

Rich-Harris (approved):

PR #7707, opened 2022-11-17
Made `config.kit.outDir` able to be located in `node_modules`

tinnedtea:
Previously, if someone tried to make `config.kit.outDir` something like `./node_modules/.svelte-kit`, build and dev processes would execute, but they would quickly start to error like this:
```
X [ERROR] Could not resolve "..\\..\\..\\@sveltejs\\kit\\src\\runtime\\components\\layout.svelte"

    node_modules/.kit/generated/nodes/0.js:1:37:
      1 │ export { default as component } from "..\\..\\..\\@sveltejs\\kit\\src\\runtime\\components\\layout.svelte";
        │                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        ╵                                      "./..\\..\\..\\@sveltejs\\kit\\src\\runtime\\components\\layout.svelte"

  Use the relative path "./..\\..\\..\\@sveltejs\\kit\\src\\runtime\\components\\layout.svelte" to
  reference the file "node_modules/@sveltejs/kit/src/runtime/components/layout.svelte". Without the
  leading "./", the path "..\\..\\..\\@sveltejs\\kit\\src\\runtime\\components\\layout.svelte" is
  being interpreted as a package path instead.
```
The error message already explains the underlying issue: Certain imports are being wrongly interpreted. To combat this I've added a `relative_path` function, which fixes this behaviour.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/sync/utils.js
+++ b/packages/kit/src/core/sync/utils.js
@@ -34,7 +34,7 @@ export function trim(str) {
 
 /**
  * This function generates relative paths, whoose `from` argument can safely reside
- * inside of `node_modules`. Otherwise paths returned by `path.relative` would be interpreted
+ * inside of `node_modules`. Otherwise paths returned by `path.relative` could be interpreted
  * as package paths.
  *
  * @param { string } from
--- a/packages/kit/src/core/sync/utils.js
+++ b/packages/kit/src/core/sync/utils.js
@@ -31,3 +31,16 @@ export function trim(str) {
 	const pattern = new RegExp(`^${indentation}`, 'gm');
 	return str.replace(pattern, '').trim();
 }
+
+/**
+ * This function generates relative paths, whoose `from` argument can safely reside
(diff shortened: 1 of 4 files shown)


packages/kit/src/core/sync/utils.js
```diff
@@ -31,3 +31,16 @@ export function trim(str) {
...
+/**
+ * This function generates relative paths, whoose `from` argument can safely reside
+ * inside of `node_modules`. Otherwise paths returned by `path.relative` could be interpreted
+ * as package paths.
+ *
+ * @param { string } from
+ * @param { string } to
+ * @returns { string }
+ */
+export function relative_path(from, to) {
```
dummdidumm, line 44: Could you move this function to `utils/filesystem.js`?

dummdidumm (approved):

Rich-Harris:
thanks!

Issue #7298, opened 2022-10-17
adapter-cloudflare: serve prerendered pages and static assets using `_routes.json`


Mexican-Man:
### Describe the problem

From what I've seen and tested (please correct me), sites deployed via Cloudflare Pages still process prerendered pages (`export const prerender = true;`) and static assets through `_worker.js`. This means that even requests to static resources contribute to the host's "function calls" budget.

### Describe the proposed solution

[This recently updated Cloudflare doc](https://developers.cloudflare.com/pages/platform/functions/billing/#will-requests-to-assets-be-charged) indicates that `_routes.json` can be used to bypass `_worker.js`, which typically handles all requests. By restructuring the build output folder (to serve static assets as they are normally routed in-app), we could take advantage of this feature. This would make adapter-cloudflare infinitely more viable for high traffic sites.

### Alternatives considered

- Using adapter-static, but that only works if *all* of your site is prerendered.
- Stricter rate limiting, but that's not scalable, reliable, or customizable (4 rules max on Cloudflare).
- Maybe surprise outages are OK ¯\\\_(ツ)\_/¯

### Importance

i cannot use SvelteKit without it

### Additional Information

I understand that traditionally, the contents of the build output folder is not meant to be exposed, and that `_worker.js` is meant to handle *every* request, even for static assets. However, I think that falling back onto Cloudflare Page's built-in file routing would be well worth it if it meant saving *actual* money on hosting costs (or in my case, having my site stay live for the whole day).

MedLeon:
This is a blocking issue for me as well.

I do not think that the `_redirect ` file is the way to go. The whole concept of a monolith _worker.js file is just for migration and should be abondend at some point. 
The "new" way would be to deploy a `[[path]].js` file. I would recommend the Cloudflare adapter for Astro. There you got two modes and one of them produces a  `[[path]].js` file.

Mexican-Man:
> I would recommend the Cloudflare adapter for Astro. There you got two modes and one of them produces a `[[path]].js` file.

That's good to know, I'll keep that in mind.

> I do not think that the `_redirect ` file is the way to go. The whole concept of a monolith _worker.js file is just for migration and should be abondend at some point. The "new" way would be to deploy a `[[path]].js` file.

[...]

PR #7708, opened 2022-11-17
Prerender routes json

Mexican-Man:
### Problem

Prerendered pages on `adapter-cloudflare` should not require a worker request. See issue for more info #7298 

### Solution

Prerender pages as `/foo/index.html` instead of `/foo.html`, then exclude `/foo` in `_routes.json`. This will bypass the worker entirely, and still serve the prerendered page.

I ran the test suite, and tested it on my site with `vite dev` and deployed to Cloudflare Pages. I got the feeling that the change to `packages/kit/src/core/prerender/prerender.js` would cause more friction that it seemed to.


Mexican-Man:
Y'know, I suppose it would break the prerender tests, wouldn't it? haha

Mexican-Man:
Not sure about that last test. I can't seem to get `goto()` to react to anything.

Rich-Harris:
Thanks you but we can't just change `about.html` to `about/index.html` willy-nilly. One means `/about`, the other means `/about/`.

I'm also not sure this approach scales very far. There's quote a low limit to the number of entries you can have in your `_routes.json` (100 for `include` and `exclude` combined, IIUC) which many sites would immediately exceed.

Mexican-Man:
> Thanks you but we can't just change `about.html` to `about/index.html` willy-nilly. One means `/about`, the other means `/about/`.

That's what I've been alluding to in #7298 . I've got it working to Cloudflare's spec using wrangler, and `vite dev` the behavior *seems* to function identically, but I *don't* know what the exact expected behavior to start with (if different). `/about` still resolves correctly, and `/about/index.html` still 404s (when I was testing). Although I can't speak to `/about` vs `/about/`, Chrome usually changes that automatically 😅

The one failed test says something is wrong, but I can't seem to get it to work no matter what I do (I'm not too familiar with playwright, unfortunately). Aside from that, it seems to work fine for me "in practice".

> I'm also not sure this approach scales very far. There's quote a low limit to the number of entries you can have in your `_routes.json`

~100 prerendered pages? How about adding a cap, or a switch + warning if over 100 entries. Then the worst case would be what's currently happening.

Rich-Harris:
I merged #8422, which adds all assets and prerendered pages (up to the 100-rule limit), so I'll close this — thanks.

In general I think the automated approach is preferable to manually specifying exclusions, though in the rare cases that it _is_ necessary it could be done by modifying the `_routes.json` after it's been created.

Mexican-Man:
Bit late, but just wanted to say thanks. I just took a look at last month's bill, and we're down 85% on billed requests 😊(just from prerendered pages, alone). HUGE; thanks Rich!

PR #7715, opened 2022-11-18
docs: add relevant variables to form example

theetrain:
* Show `$page` store and `form` variable in invalid form example. 

I was a bit stumped by how to to display the status code passed into `return invalid()`, so I hyper-focused on the "Validation errors" section on this page and couldn't find information despite it being mentioned:

> The invalid function lets you return an <ins>HTTP status code</ins> (typically 400 or 422, in the case of validation errors) along with the data

`$page` and `$page.status` are mentioned several times on that document already, so hopefully this small change to the example will help others.

### Tests

### Changesets

### DIFF
--- a/documentation/docs/20-core-concepts/30-form-actions.md
+++ b/documentation/docs/20-core-concepts/30-form-actions.md
@@ -178,8 +178,17 @@ export const actions = {
 
 ```diff
 /// file: src/routes/login/+page.svelte
+<script>
+	import { page } from '$app/stores'
+
+	export let form
+</script>
+
 <form method="POST" action="?/login">
 -	<input name="email" type="email">
++	{#if form?.missing || form?.incorrect}
++		<p>{$page.status} There are errors in your form submission. See below</p>
++	{/if}
 +	{#if form?.missing}<p class="error">The email field is required</p>{/if}
 +	{#if form?.incorrect}<p class="error">Invalid credentials!</p>{/if}
 +	<input name="email" type="email" value={form?.email ?? ''}>
--- a/documentation/docs/20-core-concepts/30-form-actions.md
+++ b/documentation/docs/20-core-concepts/30-form-actions.md
@@ -181,13 +181,16 @@ export const actions = {
 <script>
 	import { page } from '$app/stores'
 
+	/** @type {import('./$types').ActionData} */
 	export let form
 </script>
 
 <form method="POST" action="?/login">
 -	<input name="email" type="email">
-+	{#if form?.missing || form?.incorrect}
-+		<p>{$page.status} There are errors in your form submission. See below</p>
++	{#if form?.incorrect && $page.status === 400}
++		<p>There are errors in your form submission. See below</p>
++	{:else if form?.incorrect && $page.status === 409}
++		<p>Account already exists, please log in or register with a different email.</p>
 +	{/if}
 +	{#if form?.missing}<p class="error">The email field is required</p>{/if}
 +	{#if form?.incorrect}<p class="error">Invalid credentials!</p>{/if}
--- a/documentation/docs/20-core-concepts/30-form-actions.md
+++ b/documentation/docs/20-core-concepts/30-form-actions.md
@@ -140,7 +140,7 @@ export const actions = {
 
 #### Validation errors
 
-If the request couldn't be processed because of invalid data, you can return validation errors — along with the previously submitted form values — back to the user so that they can try again. The `invalid` function lets you return an HTTP status code (typically 400 or 422, in the case of validation errors) along with the data:
+If the request couldn't be processed because of invalid data, you can return validation errors — along with the previously submitted form values — back to the user so that they can try again. The `invalid` function lets you return an HTTP status code in `$page.status` (typically 400 or 422, in the case of validation errors) along with the data in `form`:
 
 ```diff
 // @errors: 2339 2304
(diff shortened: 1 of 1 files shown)


documentation/docs/20-core-concepts/30-form-actions.md
```diff
@@ -178,8 +178,17 @@ export const actions = {
 
 ```diff
 /// file: src/routes/login/+page.svelte
+<script>
+	import { page } from '$app/stores'
+
+	export let form
```
Rich-Harris, line 184:
```suggestion
  /** @type {import('./$types').ActionData} */
	export let form;
```

documentation/docs/20-core-concepts/30-form-actions.md
```diff
@@ -178,8 +178,17 @@ export const actions = {
...
+</script>
+
 <form method="POST" action="?/login">
 -	<input name="email" type="email">
++	{#if form?.missing || form?.incorrect}
++		<p>{$page.status} There are errors in your form submission. See below</p>
++	{/if}
```
Rich-Harris, line 191: do these three lines add much? The goal is to keep these samples as direct and concise as possible

```suggestion
```

documentation/docs/20-core-concepts/30-form-actions.md
```diff
@@ -178,8 +178,17 @@ export const actions = {
 
 ```diff
 /// file: src/routes/login/+page.svelte
+<script>
+	import { page } from '$app/stores'
+
```
Rich-Harris, line 183: if the `if` block is removed, these lines can be too

```suggestion
```

Rich-Harris (commented):
thanks! comments inline

theetrain:
Thanks for the feedback! I would like to keep the `{#if}` block in order to illustrate `$page.status` and `form` information being used together. 

While I was hastily developing with Form Actions, I wrongfully assumed the information returned by `return invalid(409, { success: false })` would all be available in the `form` variable since I was looking at this specific example in the docs at the time. With this proposed change, it should be clearer that the returned status code is retrieved in `$page.status`, where it can always be found in Kit. 

When read from top to bottom, `$page.status` is certainly brought up and explained well, but I hope there's some room to contain this bit of helpful repetition that would have saved me. 

I modified my example to be less contrived:

https://github.com/sveltejs/kit/blob/909fc42409e6b5c286a4ba6db25ae7b77072d6f6/documentation/docs/20-core-concepts/30-form-actions.md?plain=1#L190-L194

documentation/docs/20-core-concepts/30-form-actions.md
```diff
@@ -178,8 +178,20 @@ export const actions = {
...
+	/** @type {import('./$types').ActionData} */
+	export let form
+</script>
+
 <form method="POST" action="?/login">
 -	<input name="email" type="email">
++	{#if form?.incorrect && $page.status === 400}
++		<p>There are errors in your form submission. See below</p>
++	{:else if form?.incorrect && $page.status === 409}
++		<p>Account already exists, please log in or register with a different email.</p>
```
dummdidumm, line 193: While the example is less contrived now, it's still not ideal since
- you can make that distinction through returning a special boolean from `return invalid(..)`, which would save some keystrokes
- it's insecure (we had such a version previously, where it was pointed out that telling the user that the email exists can be used for further attacks)

I think the best way to handle this is probably to keep this out of the code example and instead note about this in a separate paragraph below. Something like "The status code returned from `invalid` is available through `$page.status`"

documentation/docs/20-core-concepts/30-form-actions.md
```diff
@@ -140,7 +140,7 @@ export const actions = {
 
 #### Validation errors
 
-If the request couldn't be processed because of invalid data, you can return validation errors — along with the previously submitted form values — back to the user so that they can try again. The `invalid` function lets you return an HTTP status code (typically 400 or 422, in the case of validation errors) along with the data:
+If the request couldn't be processed because of invalid data, you can return validation errors — along with the previously submitted form values — back to the user so that they can try again. The `invalid` function lets you return an HTTP status code in `$page.status` (typically 400 or 422, in the case of validation errors) along with the data in `form`:
```
dummdidumm, line 143:
```suggestion
If the request couldn't be processed because of invalid data, you can return validation errors — along with the previously submitted form values — back to the user so that they can try again. The `invalid` function lets you return an HTTP status code (typically 400 or 422, in the case of validation errors) along with the data. The status code is available through `$page.status`, the data through `form`:
```

theetrain: Wow you're quick; I've applied this change, but added 'and' to the final list item since I feel it reads bit more naturally.

documentation/docs/20-core-concepts/30-form-actions.md
```diff
@@ -140,7 +140,7 @@ export const actions = {
 
 #### Validation errors
 
-If the request couldn't be processed because of invalid data, you can return validation errors — along with the previously submitted form values — back to the user so that they can try again. The `invalid` function lets you return an HTTP status code (typically 400 or 422, in the case of validation errors) along with the data:
+If the request couldn't be processed because of invalid data, you can return validation errors — along with the previously submitted form values — back to the user so that they can try again. The `invalid` function lets you return an HTTP status code (typically 400 or 422, in the case of validation errors) along with the data. The status code is available through `$page.status`, and the data through `form`:
```
dummdidumm, line 143: yeah don't know why I added a comma there, makes more sense without one 👍 
```suggestion
If the request couldn't be processed because of invalid data, you can return validation errors — along with the previously submitted form values — back to the user so that they can try again. The `invalid` function lets you return an HTTP status code (typically 400 or 422, in the case of validation errors) along with the data. The status code is available through `$page.status` and the data through `form`:
```

dummdidumm (approved):

documentation/docs/20-core-concepts/30-form-actions.md
theetrain: No Oxford comma in Kit docs, eh? I see that other sentences don't use the comma, so mine shouldn't as well 😃.

PR #7719, opened 2022-11-18
Make `trailingSlash` a page/endpoint option, like `prerender`

### LINKED ISSUE #7302: trailingSlash override functionality
### Issue: Looking for a way to override trailingSlash from `always` -> `never` in a route sub-path

Is it possible to override the default configuration for 'always' or 'never' for a childless leaf node route?

If I'm not mistaken, before the August breaking changes, this was possible to override at the folder/page level with something like:
```<script context="module"> export const trailingSlash = 'never'; </script>```

Trying to export this from a `+page.server.ts` does not seem to convey to the current build.

It's nice to use always/never instead of using `ignore` to ensure all urls have a single canonical, but there are use cases where it's nice to enforce `always` as the default, and then apply `never` to childless node pages.

For example:
/topic/ - trailing slash
/topic/subtopic/ - trailing slash
/topic/subtopic/leaf-content-with-no-children - Would like to use `trailingSlash = never` for content here?

Asked on discord, was advised this was intentional but possibly worth creating an Issue to confirm this is a permanent, intentional design decision.

This would only make sense in the context of using `always` as the site-wide default, then allowing over-rides in subfolders.

The config option would need to be overridable server-side in a `+layout.server.ts` or `+page.server.ts` in a sub-folder. 

In this example, a `+page.server.ts` in `/topic/subtopic/` would be applied to flip `always` to `never` for children of `subtopic`.

Routes matching a regex like `/topic/subtopic/[slug]` would not receive a `trailingSlash`.

### Alternatives considered

Using ignore and building the enforcement manually in JS or enforce with rewriterules at the LB layer.

Definitely possible, but feels undesirable. 

### Importance

nice to have

### Additional Information

It feels like a fairly common canonicalization practice has been that folders get a trailing slash, child-most content pages do not.

If you are in a situation with a legacy site that behaves this way, you must make some tough, permanent decisions about lots of 301's if they have used a structure where **only** child-most content does not have a trailing slash.

Without a built-in method to replicate, it increases the level of difficulty to migrate existing sites. 

Again, all possible, but it would feel smoother if this was possible without adding in nginx/custom logic/etc :)

Rich-Harris:
Closes #7302. This moves `trailingSlash` configuration from `svelte.config.js` and into pages. Benefits:

* Moving stuff out of config and into code is always good
* We can now have e.g. `/blog/` and `/blog/my-post`
* We can now control trailing slashes on endpoints (niche, but will probably be useful one day)

It has very limited impact on client-side bundle size or data transfer.

The downside is that on the server, we need to load code for the route before we can return a redirect response. In practice the impact is negligible (since we don't need to _render_ anything, just import/evaluate it) and in any case it's not the most important thing to optimise.

TODO:

* [x] documentation
* [x] delete `packages/kit/test/prerendering/trailing-slash` etc — no longer necessary, the tests can be moved into other apps

### Tests

### Changesets


dummdidumm (approved):
Looks good code-wise. Question I have though: How likely is it that people are pissed off by this because they now have to add `export const trailingSlash = ".."` to all of their `+server.js` files? Previously, you had it to set it only once in the config, and it applied to endpoints as well.

Rich-Harris:
> Previously, you had it to set it only once in the config, and it applied to endpoints as well.

Not quite — trailing slashes were completely ignored for endpoints. Now they'll be treated as `never` by default, which I think is what people will want in the overwhelming majority of cases (it's only really pages where a trailing slash ever makes sense), but in the exceptional cases where they _do_ want to enforce a trailing slash they can now do so much more easily than before, where they had to implement the redirect logic themselves

PR #7721, opened 2022-11-19
ignore popstate events from outside the router

### LINKED ISSUE #6914: Disable router for all links
### Describe the problem

I use sveltekit in a micro-frontend like usage, injecting html and js generated by svelte-kit in another website.
I used `router=false` option to avoid sveltekit to interfere with the router of the receiving site but keep hydration to activate some js interaction.

Since [this PR](https://github.com/sveltejs/kit/pull/6446) was merged I should use this to keep same usage:

> old: router = false / hydrate = true -> new: csr = true + set data-sveltekit-reload if you want native navigations

Unfortunately receiving site links like `<a href="/">Homepage</a>` are still caught by sveltekit router even if I set `data-sveltekit-reload` in `<html>` tag of the receiving site.

Bring back another configuration field `router` set to true by default. 
As `csr` and `ssr` fields add a way to configure this new field in `page.js` or `layout.js`

### Alternatives considered

_No response_

### Importance

### Additional Information

I don't have all the history of the changes of `router` field, I hope this issue was not already studied. 
It was a very convenient feature for me, I hope I can find a solution.
I can make the PR with proposed solution if needed.

Rich-Harris:
fixes #6914

### Tests

### Changesets

### DIFF
--- a/.changeset/ninety-colts-protect.md
+++ b/.changeset/ninety-colts-protect.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Ignore popstate events from outside the router
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -1458,7 +1458,7 @@ export function create_client({ target, base, trailing_slash }) {
 			});
 
 			addEventListener('popstate', (event) => {
-				if (event.state) {
+				if (event.state?.[INDEX_KEY]) {
 					// if a popstate-driven navigation is cancelled, we need to counteract it
 					// with history.go, which means we end up back here, hence this check
 					if (event.state[INDEX_KEY] === current_history_index) return;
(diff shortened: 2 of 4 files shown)


benmccann:
Will the scroll get restored properly in this scenario? Will it just fall-through to the browser and the browser will handle it correctly?

Rich-Harris:
> Will the scroll get restored properly in this scenario?

That's entirely up to whatever code was responsible for the non-SvelteKit history entry. If it's also listening for `popstate` events and handling them appropriately then yes, scroll will be properly restored; if not, it's not our problem.

dummdidumm (approved):
For future readers - the reason why this fixes the issue is because the OP's use case is embedding SvelteKit as a microapp inside another bigger site, which can also have navigation interactions, which previously interfered with the SvelteKit app. This change makes it so all popstate events not leading to states within the SvelteKit app are ignored.

PR #7722, opened 2022-11-19
always set accept/accept-language headers when making a self-request

### LINKED ISSUE #7720: internal `fetch` request is not initialized correctly
### Describe the bug

When we call `fetch` to an internal endpoint the request is not initialized the same way it would be in the browser or using `undici`:

https://github.com/sveltejs/kit/blob/0d4596340389233031588a77e478f9fb34a26d4e/packages/kit/src/runtime/server/fetch.js#L131

Here's the spec: https://fetch.spec.whatwg.org/#fetching

In particular, I noticed that we're not setting the `Accept` header, but it looks like there's other stuff we're not doing either like setting `Accept-Language`.

### Reproduction

Call a `+server.js` from a `load` function in a `+page.js`

See https://github.com/sveltejs/sites/pull/398

### Logs

_No response_

### System Info

```Shell
`master`
```

### Severity

### Additional Information

_No response_

Rich-Harris:
closes #7720. We previously decided not to automatically include request headers, but I think these exceptions are valid. Given that `accept` is used within Kit for content negotiation purposes, it makes sense to default it to `*/*` if unspecified. Similarly, if `accept-language` were to be used for i18n purposes, it makes sense for it to match the initial page request.

### Tests

### Changesets


benmccann (approved):
awesome. thanks!

dummdidumm (approved):

PR #7724, opened 2022-11-19
[chore] don't override printUrls

benmccann:
See https://github.com/vitejs/vite/issues/9378#issuecomment-1229771480 for an explanation

Should this close https://github.com/vitejs/vite/issues/9378?

### DIFF
--- a/.changeset/green-timers-develop.md
+++ b/.changeset/green-timers-develop.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Prevent Vite from nuking logs on startup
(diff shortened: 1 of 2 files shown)


Rich-Harris:
> Should this close [vitejs/vite#9378](https://github.com/vitejs/vite/issues/9378)?

No. The current behaviour still makes no sense. Many things could log info/warnings/errors, and not all of them have access to `config.logger`. Nor does it make sense that we have to store `deferred_warning` until `configResolved` just so that _we're_ able to use `config.logger`.

However, this new information is useful:

> It seems Vite already does this as long as `config.logger.warn`/`warnOnce`/`error` is used.

All we need to do, it seems, is add a `config.logger.warn('')` inside `configResolved`, and we will effectively trick Vite into not nuking logs. Updating the PR.

benmccann:
the changes lgtm

PR #7735, opened 2022-11-21
stronger dynamic env types

ignatiusmb:
Prefixes the dynamic type index signature with `config.env.publicPrefix`

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/core/env.js
+++ b/packages/kit/src/core/env.js
@@ -2,6 +2,15 @@ import { GENERATED_COMMENT } from '../constants.js';
 import { runtime_base } from './utils.js';
 
 /**
+ * @typedef {'public' | 'private'} EnvType
+ * @typedef {{
+ * 	public: Record<string, string>;
+ * 	private: Record<string, string>;
+ * 	prefix: string;
+ * }} EnvData
+ */
+
+/**
  * @param {string} id
  * @param {Record<string, string>} env
  * @returns {string}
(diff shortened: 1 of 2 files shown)


packages/kit/src/core/env.js
```diff
@@ -25,46 +34,53 @@ export function create_static_module(id, env) {
...
  */
 export function create_dynamic_types(id, env) {
-	const properties = Object.keys(env)
+	const properties = Object.keys(env[id])
 		.filter((k) => valid_identifier.test(k))
 		.map((k) => `\t\t${k}: string;`);
 
-	properties.push(`\t\t[key: string]: string | undefined;`);
+	properties.push(
+		`\t\t[key: \`${env.prefix}\${string}\`]: ${id === 'public' ? 'string | undefined' : 'never'}`
```
dummdidumm, line 75: instead of pushing `never`, shouldn't we just leave out this key/value pair?

dummdidumm (commented):
Does this mean that so far the public dynamic types were all wrong?

packages/kit/src/core/env.js
ignatiusmb: Agree with not using `never`, but we should still include this because if we leave this out, the interface will still have the `[key: string]: string | undefined` signature added below. In that case, accessing variables starting with `env.publicPrefix` from the `/dynamic/private` module will return the `string | undefined` union, which is not as nice compared to only (currently `never`, changing to) `undefined` as this will get caught on compile/lint.

ignatiusmb:
It's not wrong per se, but not narrow enough in that the public dynamic types allow all variables by using `[key: string]`. We can generate a stricter interface where only keys that starts with the prefix will exists, e.g., ``[key: `PUBLIC_${string}`]`` and will error out on typos, or other variables that does not start with `PUBLIC_`, or even when the user changes their `env.publicPrefix` option.

packages/kit/src/core/env.js
```diff
@@ -25,46 +34,55 @@ export function create_static_module(id, env) {
...
-	properties.push(`\t\t[key: string]: string | undefined;`);
+	properties.push(
+		`\t\t[key: \`${env.prefix}\${string}\`]: ${
+			id === 'public' ? 'string | undefined' : 'undefined'
+		};`
+	);
+
+	if (id === 'private') {
+		properties.push(`\t\t[key: string]: string | undefined;`);
+	}
```
dummdidumm, line 82: I think it might make more sense to combine this into

```js
if (id === 'private') {
  // push [key: string] signature
} else {
  // push [`PUBLICPREFIX_${string}`: string] signature
}
```
(whats the minimum required TS version for this btw? I think it's in our range, right?)

ignatiusmb: I can see that, but it'll look more like this

```js
if (id === 'private') {
  // push { [`PUBLICPREFIX_${string}`: string]: undefined } signature
  // push { [key: string]: string | undefined } signature
} else {
  // push { [`PUBLICPREFIX_${string}`: string]: string | undefined } signature
}
```

I believe it is within our range, minimum version for this was 4.4 where [template literal in index signature](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#symbol-template-signatures) was added

dummdidumm: Mhm for some reason I thought that private dynamic also includes the public variables for convenience - nevermind then.

dummdidumm (approved):

PR #7736, opened 2022-11-21
external fetch calls: ensure serialized cookie values are url-encoded

ximus:
### TLDR

before: external fetch cookie was sent plain

```
browser
     |
(request with url-encoded cookie) (cookie=foo%2Bbar)
     |
     v
sveltekit node app
     |
(external fetch request made with url-DEcoded cookie) (cookie=foo+bar)
     |
     v
backend service
```

after: external fetch cookie is url-encoded

```
browser
     |
(request with url-encoded cookie) (cookie=foo%2Bbar)
     |
     v
sveltekit node app
     |
(external fetch request made with url-ENcoded cookie) (cookie=foo%2Bbar)
     |
     v
backend service
```

### Description

My kit app was making external `fetch` requests with cookies that were different from the the cookies that were part of the original request to the app.

My request coming from the browser had a session cookie that was url-encoded. My sveltekit server code would then use the kit-`fetch` to make a request to a backend api server that uses that same session token in the cookie. But when the request to the api left the kit app, the cookie was not url-encoded anymore. This caused problems in my api server because it would parse the
cookies by doing a url-decode on the values. Given my session token could contain characters that could further url-decode, it would get garbled when it was url-decoded by the api, because it was not url-encoded anymore.

From my superficial web readings, there seems to be no hard standard on the encoding of cookies. But the industry trend seems to be to url-encode cookie values.

### Conclusion

Given that:

1. The industry trend seems to be to url-encode cookie values. 
2. This cookie.js code already url-decodes cookies on the way in. 
3. "Forwarded" input cookies are different on the way out than on the way in

I assumed the attached fix is reasonable, that cookies should just always be url-encoded.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/cookie.spec.js
+++ b/packages/kit/src/runtime/server/cookie.spec.js
@@ -42,12 +42,13 @@ paths.negative.forEach(([path, constraint]) => {
 	});
 });
 
-/** @param {string} href */
-const cookies_setup = (href = 'https://example.com') => {
-	const url = new URL(href);
+/** @param {{ href?: string, headers?: Record<string, string> }} href */
+const cookies_setup = ({ href, headers } = {}) => {
+	const url = new URL(href || 'https://example.com');
 	const request = new Request(url, {
 		headers: new Headers({
-			cookie: 'a=b;'
+			cookie: 'a=b;',
+			...headers
 		})
 	});
 	return get_cookies(request, url, { dev: false, trailing_slash: 'ignore' });
@@ -72,7 +73,7 @@ test('default values when set is called', () => {
 });
 
 test('default values when set is called on sub path', () => {
-	const { cookies, new_cookies } = cookies_setup('https://example.com/foo/bar');
+	const { cookies, new_cookies } = cookies_setup({ href: 'https://example.com/foo/bar' });
 	cookies.set('a', 'b');
 	const opts = new_cookies['a']?.options;
 	assert.equal(opts?.secure, true);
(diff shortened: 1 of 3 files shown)


ximus:
if this change makes sense, then I assume there needs to be a changeset as this is a change in behavior.

dummdidumm (changes requested):
I think the change makes sense. We need to more the encoding step into the previous for loops though. Theoretically, someone could have invoked `cookies.set` with a different encoding option, which we need to use (line 177 something like `.. = (cookie.options.encode || encodeURIComponent)(cookie.value);`)

packages/kit/src/runtime/server/cookie.spec.js
```diff
@@ -146,4 +147,18 @@ test('cookie names are case sensitive', () => {
...
 
+test('serialized cookie header should be url-encoded', () => {
+	const href = 'https://example.com';
+	const { cookies, get_cookie_header } = cookies_setup({
+		href,
+		headers: {
+			cookie: 'a=f%C3%BC' // a=fü
+		}
+	});
+	// not that one should do this, but we follow the spec...
```
benmccann, line 158: I don't understand what this comment is trying to suggest. Why shouldn't one use `ö` in a cookie value? Maybe better to just remove the comment...
```suggestion
```

benmccann:
lgtm. stackoverflow suggests this is the right solution: https://stackoverflow.com/questions/1969232/what-are-allowed-characters-in-cookies

packages/kit/src/runtime/server/cookie.spec.js
ximus: woops sorry, this is a leftover from the previous test case that I copy pasted initially. removing ...

ximus:
Thanks @dummdidumm, refactored to honour any `encode` option passed to `set()`.

Also created two top level `encode`/`decode` constants to make more consistent and clear the default encoding functions, and referenced them wherever there is a parse or serialisation.

Added a changeset.

kept as separate commits for ease of PR review but you can squash on merge or I can squash myself, let me know.

dummdidumm (approved):
Thank you!

PR #7739, opened 2022-11-21
adapter-auto: defer adapter installation until required

gtm-nayan:

### Tests

### Changesets

### DIFF
--- a/packages/adapter-auto/index.js
+++ b/packages/adapter-auto/index.js
@@ -5,9 +5,6 @@ import { adapters } from './adapters.js';
 import { dirname, join } from 'path';
 import { existsSync } from 'fs';
 
-/** @type {import('./index').default} */
-let fn;
-
 /** @type {Record<string, (name: string) => string>} */
 const commands = {
 	npm: (name) => `npm install -D ${name}`,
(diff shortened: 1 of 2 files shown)


dummdidumm (approved):
Makes sense to me - leaving this open so others can give their opinions on this

Rich-Harris:
If we add new capabilities to adapters in future — for example, if they're allowed to control aspects of the dev environment — then this won't work

Conduitry:
Probably we don't want the auto adapter to be affecting dev mode? It wouldn't have any of the env vars it looks at anyway. Presumably people should explicitly install their target adapter if they want to have it impact dev mode too.

gtm-nayan:
To provide platform specific features in dev mode you'd have to know what platform you'd be deploying to, and at that point adapter-auto can be swapped out manually.

I thought the idea was that adapter-auto would serve as a minimal deployable setup for new users unfamiliar with `adapters` and users requiring more features can use specific ones for their platform. There's a warning in adapter-auto which reflects that sentiment: `If you plan on staying on this deployment platform, consider replacing @sveltejs/adapter-auto with ${match.module}. This will give you faster and more robust installs, and more control over deployment configuration.`

Rich-Harris:
That's a fair point! I rescind my earlier comments. That said, I'm not 100% sure I understand what problem this solves?

packages/adapter-auto/index.js
```diff
@@ -46,72 +43,78 @@ async function import_from_cwd(name) {
...
-					);
-				}
-			} else {
-				throw error;
+/** @typedef {import('@sveltejs/kit').Adapter} Adapter */
+
+/**
+ * @returns {Promise<Adapter | undefined>} The corresponding adapter for the current environment if found otherwise undefined
+ */
+async function get_adapter() {
```
benmccann, line 51: I'd make it clearer from the name that this is installing the adapter as `get` sounds side-effect free
```suggestion
async function install_adapter() {
```

benmccann:
> I'm not 100% sure I understand what problem this solves?

Right now it's installing the adapter up-front, which seems to be doing the install in the middle of the Vite bundling causing two versions of SvelteKit to end up in the final bundle. By waiting until the bundle is written, hopefully that's fixed.

I'm not 100% sure how to test this other than merging it and updating the sites repo. There's probably a way, but I'd be just as happy to merge this even if it doesn't work as I think it's a nice way to structure the code, so I'd say we merge this and see if it works

packages/adapter-auto/index.js
Rich-Harris: In some cases it _won't_ install anything though. We could make it `get_or_install_adapter` or whatever but i think it's clear enough from context

PR #7747, opened 2022-11-21
write redirects for prerendered pages with trailing slashes

### LINKED ISSUE #7725: Prerendered pages with trailing slash 404 when directly loaded
### Describe the bug

This only appears to happen when deployed (I've tested Vercel, but not other deployment options) and works as expected with a local `npm run build && npm run preview`.

With `export const prerender = true;` in `src/routes/+layout.ts` and `trailingSlash: "always"` in the config, rendered pages 404 when directly loaded.

The page loads as expected when navigating from another page (i.e. client-side navigation), but when loaded directly (or when the link is clicked before code implementing client-side navigation is loaded), the URL returns a 404. This isn't a problem with `trailingSlash: "never"`.

### `trailingSlash: "always"`
- [Demo](https://repro-prerender-404-k64onxso1-drzax.vercel.app/)
- [Code](https://github.com/drzax/repro-prerender-404)

There's a PR in that repo where the only change is to set `trailingSlash: "never"`:
- [Demo](https://repro-prerender-404-6hgfib8uc-drzax.vercel.app/)
- [Code](https://github.com/drzax/repro-prerender-404/pull/1/files)

This doesn't appear to affect the root path (presumably because it always has a trailing slash), but all other paths appear to be affected. 

### Logs

_No response_

### System Info

```Shell
System:
    OS: macOS 12.6
    Memory: 51.70 MB / 32.00 GB
    Shell: 5.1.8 - /usr/local/bin/bash
  Binaries:
    Node: 18.8.0 - /usr/local/bin/node
    npm: 8.19.2 - /usr/local/bin/npm
  Browsers:
    Chrome: 109.0.5414.10
    Edge: 107.0.1418.52
    Firefox: 106.0.5
    Firefox Developer Edition: 107.0
    Safari: 16.0
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.88
    @sveltejs/adapter-vercel: ^1.0.0-next.81 => 1.0.0-next.81
    @sveltejs/kit: next => 1.0.0-next.551
    svelte: ^3.44.0 => 3.53.1
    vite: ^3.1.0 => 3.2.4
```

### Severity

### Additional Information

In my use case I can set trailingSlash to 'never', but can easily imagine a scenario where that might not be possible or desirable. 

Vercel build output where trailing slash is 'always'
image

Vercel build output where trailing slash is 'never'
image

This may be related to (or even the same bug as) #6236, but I couldn't tell for sure so thought I'd submit this anyway — apologies if it's a duplicate.

Rich-Harris:
fixes #7725. The prerendered page wasn't being served because the `overrides` config in `.vercel/output/config.json` can't have a `path` value with a trailing slash. To get the correct behaviour we have to remove the trailing slash, and also add two entries to the `routes` config for each prerendered page:

```json
{ "src": "/the-path/", "dest": "/the-path" },
{ "src": "/the-path", "status": 308, "headers": { "Location": "/the-path/" } }
```

### Tests

### Changesets

### DIFF
--- a/packages/adapter-vercel/index.js
+++ b/packages/adapter-vercel/index.js
@@ -182,18 +182,6 @@ export default function ({ external = [], edge, split } = {}) {
 
 			builder.log.minor('Writing routes...');
 
-			console.log(
-				JSON.stringify(
-					{
-						version: 3,
-						routes,
-						overrides
-					},
-					null,
-					'  '
-				)
-			);
-
 			write(
 				`${dir}/config.json`,
 				JSON.stringify(
--- a/packages/adapter-vercel/index.js
+++ b/packages/adapter-vercel/index.js
@@ -180,13 +180,29 @@ export default function ({ external = [], edge, split } = {}) {
 
 			builder.log.minor('Writing routes...');
 
+			console.log(
+				JSON.stringify(
+					{
+						version: 3,
+						routes,
+						overrides
+					},
+					null,
+					'  '
+				)
+			);
+
 			write(
 				`${dir}/config.json`,
-				JSON.stringify({
-					version: 3,
-					routes,
-					overrides
-				})
+				JSON.stringify(
+					{
+						version: 3,
+						routes,
+						overrides
+					},
+					null,
+					'  '
+				)
 			);
 		}
 	};
--- a/packages/adapter-vercel/index.js
+++ b/packages/adapter-vercel/index.js
@@ -13,9 +13,9 @@ export default function ({ external = [], edge, split } = {}) {
 			const node_version = get_node_version();
 
 			const dir = '.vercel/output';
-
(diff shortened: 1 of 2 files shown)


packages/adapter-vercel/index.js
```diff
@@ -164,19 +182,29 @@ export default function ({ external = [], edge, split } = {}) {
...
+				JSON.stringify(
+					{
+						version: 3,
+						routes,
+						overrides
+					},
+					null,
+					'  '
+				)
+			);
```
Conduitry, line 195: Is it intended to also be logging this to stdout, or was this just for debugging? Will that make things too noisy?

Rich-Harris: it is not intended. oops

dummdidumm (approved):

PR #7749, opened 2022-11-21
[fix] preserve Vite CLI opts

benmccann:
@dominikg noticed this while trying to implement some new logging and discovering he couldn't configure the log level

### DIFF
--- a/packages/kit/src/exports/vite/utils.js
+++ b/packages/kit/src/exports/vite/utils.js
@@ -25,7 +25,7 @@ export async function get_vite_config(config, config_env) {
 		mode: config_env.mode,
 		logLevel: config.logLevel,
 		clearScreen: config.clearScreen,
-		optimizeDeps: { force: config.optimizeDeps.force },
+		optimizeDeps: { force: config.optimizeDeps.force }
 	};
 }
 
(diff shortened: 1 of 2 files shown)


Rich-Harris (approved):

packages/kit/src/exports/vite/utils.js
```diff
@@ -19,7 +19,14 @@ export async function get_vite_config(config, config_env) {
...
 		throw new Error('Could not load Vite config');
 	}
-	return { ...loaded.config, mode: config_env.mode };
+	return {
+		...loaded.config,
+		// CLI opts
+		mode: config_env.mode,
+		logLevel: config.logLevel,
+		clearScreen: config.clearScreen,
+		optimizeDeps: { force: config.optimizeDeps.force }
```
dominikg, line 28: is this enough or do you need to clone the whole optimizeDeps config? right now it is disabled in build by default but in a future version or user config it may be ebabled?

dominikg: ahh, that's only the cli values and merge happen s later

benmccann: It's only `--force` that you can specify via the CLI. I copied this from the Vite codebase

dominikg (approved):

PR #7753, opened 2022-11-22
Allow chained optional parameters

### LINKED ISSUE #7548: Optional route with matcher followed by rest parameters

Rich-Harris:
Closes #7548, alternative to #7549.

This is late night code — it's not exactly self-explanatory, but it works. I think. It might be possible to neaten it up a bit, but not now 😴 

It solves the `/[[lang=lang]]/[...rest]` case, where you want `/some-path` to result in `{ rest: 'some-path' }`. Currently, that won't match anything, because the `[[lang=lang]]` param 'eats' the first segment even if it doesn't match, which violates intuitions about how this stuff should work.

The solution is more general than just that case, however — any time an optional parameter with a matcher fails to match, we try it again with the next one by essentially shifting the regex matches rightwards.

### Tests

### Changesets

### DIFF
--- a/.changeset/forty-icons-switch.md
+++ b/.changeset/forty-icons-switch.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Allow chained optional parameters
--- a/.changeset/forty-icons-switch.md
+++ b/.changeset/forty-icons-switch.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Allow chained optional parameters
+Roll over non-matching optional parameters instead of 404ing
--- a/packages/kit/src/utils/routing.js
+++ b/packages/kit/src/utils/routing.js
@@ -125,33 +125,33 @@ export function exec(match, params, matchers) {
 	const result = {};
 
 	let m = 1;
-	let prev = '';
+	let buffered = '';
 
 	for (const param of params) {
-		let value = (param.chained && prev) || match[m++];
+		let value = (param.chained && buffered) || match[m++];
 
-		if (param.rest && param.chained && prev && match[m]) {
+		if (param.rest && param.chained && buffered && match[m]) {
 			value += `/${match[m++]}`;
 		}
 
-		if (!value) {
-			if (param.rest) result[param.name] = '';
-			continue;
-		}
-
-		if (param.matcher) {
-			const matcher = matchers[param.matcher];
-			if (!matcher(value)) {
-				if (param.optional && param.chained) {
-					prev = value;
-					continue;
+		if (value) {
+			if (param.matcher) {
+				const matcher = matchers[param.matcher];
+				if (!matcher(value)) {
+					if (param.optional && param.chained) {
+						buffered = value;
+						continue;
+					}
+					return;
 				}
-				return;
 			}
+
+			result[param.name] = value;
+		} else {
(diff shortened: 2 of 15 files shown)


Rich-Harris:
Note to self - replace the integration test with unit tests

.changeset/forty-icons-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Allow chained optional parameters
```
Rich-Harris, line 5:
```suggestion
Roll over non-matching optional parameters instead of 404ing
```

packages/kit/src/utils/routing.js
```diff
@@ -119,35 +117,65 @@ export function get_route_segments(route) {
...
+
+		if (value === undefined) {
+			// if `value` is undefined, it means this is
+			// an optional or rest parameter
+			if (param.rest) result[param.name] = '';
+		} else {
+			if (param.matcher && !matchers[param.matcher](value)) {
+				// in the `/[[a=b]]/[[c=d]]` case, if the value didn't satisfy the `b` matcher,
+				// try again with the next segment by shifting values rightwards
+				if (param.optional && param.chained) {
```
dummdidumm, line 151: What about `/[[a=b]]/[[c=d]]/e`? `chained` is only `true` for `rest` in that case right now

dummdidumm (approved):

PR #7762, opened 2022-11-22
rename `prerendering` to `building`, `remove config.kit.prerender.enabled`

Rich-Harris:
Follow-up to #7716. This PR makes it possible to add route-level configuration in future that can be used inside adapters to enable things like granular control over things like runtime and region.

It's a breaking change, because it means we need to remove `config.kit.prerender.enabled` to ensure that it's possible to run the app at build time.

It also renames `prerendering` to `building`:

```diff
-import { prerendering } from '$app/environment';
+import { building } from '$app/environment';
```

If any code should _not_ run during `build` (for example, connecting to a production database), it can be wrapped in an if block:

```js
import { building } from '$app/environment';

if (!building) {
  do_some_setup_work();
}
```

### Tests

### Changesets


packages/kit/types/ambient.d.ts
```diff
@@ -77,17 +77,17 @@ declare module '$app/environment' {
 	export const browser: boolean;
 
 	/**
-	 * Whether the dev server is running. This is not guaranteed to correspond to `NODE_ENV` or `MODE`.
+	 * SvelteKit analyses your app during the `build` step by running it. During this process, `building` is `true`.
```
dummdidumm, line 80:
```suggestion
	 * SvelteKit analyses your app during the `build` step by running it. During this process, `building` is `true`.
	 * You can use this to wrap any code that should not be executed during build in a \`if (!building) {...}\` block.
```
Related: It might also be good to note in the page options docs that `building` will be `true` while prerendering. I also think it makes sense to explain somewhere how exactly the semantics of `building` are - for example, that this means top level code is run just because things are loaded, but that functions like `load` are not executed, unless at a later stage during prerendering.

Rich-Harris: agree, added some stuff to the docs

dummdidumm:
How many people have disabled that setting? I guess not many, because it works flawlessly even if you don't prerender anything, except when you do eager db setups, in which case you are probably already doing `if (prerendering) {..}`? Is there value in having a similar setting so that that you can opt out of build time analysis?

Conduitry:
I had disabled it in apps (that I was first porting from Sapper to SvelteKit a year ago), but I don't remember what specific problem I was avoiding by doing so. There aren't any eager database connections I had to avoid. But in any case, it looks like I don't need this configuration anymore, because the apps work fine without disabling it.

Rich-Harris:
I posted a link to the [discussion](https://github.com/sveltejs/kit/discussions/7716) in the two (well-populated) issue threads where this had come up in the past to get maximum visibility. Lots of 👍 reactions — tl;dr is I think we're in the clear. If we need to add some opt-out in future we can consider it but we shouldn't do it defensively

dummdidumm (approved):

sdekna:
This change is a breaking change that broke a lot of apps, but no explanatory migration procedure is presented, is it possible to have one? I couldn't find my way through this migration.

dummdidumm:
I updated the PR description with a more descriptive migration guide.

sdekna:
Many thanks man!
I think as the framework is approaching 1.0, it would be helpful to host a releases page somewhere on the website where every new release is announced and any breaking changes or notes explained... This might help.
Thanks again.

PR #7766, opened 2022-11-22
client-side router ignores links outside app target

### LINKED ISSUE #7763: Router should ignore clicks from outside the app container
### Describe the problem

I had a follow-up thought about https://github.com/sveltejs/kit/issues/6914. In cases where people are using SvelteKit to generate embeds that sit in a larger application, it's necessary to come up with elaborate workarounds to prevent clicks in the main app nav (for example) being handled by SvelteKit's router.

If we instead only listened for click events coming from inside the part of the app SvelteKit controls, it would be very easy to embed SvelteKit apps elsewhere.

I'm pretty sure we would just need to replace https://github.com/sveltejs/kit/blob/93712f09e955c1d9c8d8d7f838fa56d98290251a/packages/kit/src/runtime/client/client.js#L1393 and other calls to `addEventListener(...)` with `target.addEventListener(...)`.

The only downside I can see is that links inside `src/app.html` outside `%sveltekit.body%` wouldn't trigger the client-side router, but a) you can work around it if you really need to with `goto`, and b) i would be astonished if anyone was doing that. Technically a breaking change, of course, so we should do this quickly if we intend to do it at all.

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

Rich-Harris:
closes #7763

### Tests

### Changesets

### DIFF
--- a/.changeset/strange-roses-sell.md
+++ b/.changeset/strange-roses-sell.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Client-side router ignores links outside app target
+[breaking] Make client-side router ignore links outside %sveltekit.body%
--- a/.changeset/strange-roses-sell.md
+++ b/.changeset/strange-roses-sell.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Client-side router ignores links outside app target
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -75,7 +75,7 @@ function check_for_removed_attributes() {
 
 /**
  * @param {{
- *   target: Element;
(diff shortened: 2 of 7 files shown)


dummdidumm (approved):
Love it when fixes are that easy

.changeset/strange-roses-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Client-side router ignores links outside app target
```
Conduitry, line 5:
```suggestion
[breaking] Make client-side router ignore links outside app target
```

Since `target` isn't something that's in the app's config anymore, I don't know whether we ought to use a different word here. Maybe:

```suggestion
[breaking] Make client-side router ignore links outside %sveltekit.body%
```

?

PR #7767, opened 2022-11-22
disallow error status codes outside 400-599 range

### LINKED ISSUE #7765: `throw error(204)` returns invalid response
### Describe the bug

Undici is fails to `fetch` an endpoint which returns `throw error(204)` because it expects there to be no body. Users cannot currently save in the REPL as a result

### Reproduction

https://github.com/sveltejs/sites/blob/9c15c79b4bcdf043166b4b3974576e86bb783289/sites/svelte.dev/src/routes/(authed)/repl/%5Bid%5D.json/%2Bserver.js#L98

### Logs

_No response_

### System Info

```Shell
`master`
```

### Severity

blocking an upgrade

### Additional Information

_No response_

Rich-Harris:
closes #7765. Also I found some unused code 

### Tests

### Changesets


.changeset/four-lamps-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Disallow status codes outside 400-599
```
Conduitry, line 5:
```suggestion
[breaking] Disallow status codes outside 400-599
```

Do we want to label this as breaking? `throw error(204)` _was_ being used by one of our own codebases, so I don't think it's unreasonable to assume that someone else would have done it too.

packages/kit/src/exports/index.js
```diff
@@ -8,6 +8,10 @@ import { HttpError, Redirect, ValidationError } from '../runtime/control.js';
  * @param {any} message
  */
 export function error(status, message) {
+	if ((!__SVELTEKIT_BROWSER__ || __SVELTEKIT_DEV__) && (status < 400 || status > 599)) {
```
dummdidumm, line 11: I assume you did this to safe some bytes in the client bundle - should we do the same as part of this PR for `redirect`?

Rich-Harris: good catch, done

dummdidumm (approved):

PR #7780, opened 2022-11-23
[feat] allow handleError to return a promise

### LINKED ISSUE #7756: Make `handleError` `async`
### Describe the problem

I want to use the handleError hook of SvelteKit, but I realized that it currently has to be a synchronous function. But unfortunately, our error mapper is an asynchronous function. Since it is not possible in JavaScript to synchronously await an asynchronous function (like in C#), it seems we cannot use the handleError hook currently.

Make the return type of `HandleServerError` and `HandleClientError` a `MaybePromise`, like the handle-hook.

```ts
export interface HandleServerError {
  (input: { error: unknown; event: RequestEvent }): MaybePromise<void | App.Error>;
}

export interface HandleClientError {
  (input: { error: unknown; event: NavigationEvent }): MaybePromise<void | App.Error>;
}
```

### Alternatives considered

Continue not to use `handleError`.

### Importance

would make my life easier

### Additional Information

_No response_

dummdidumm:
Closes #7756

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/utils.js
+++ b/packages/kit/src/runtime/server/utils.js
@@ -113,10 +113,10 @@ export function static_error_page(options, status, message) {
  * @param {import('types').SSROptions} options
  * @param {unknown} error
  */
-export function handle_fatal_error(event, options, error) {
+export async function handle_fatal_error(event, options, error) {
 	error = error instanceof HttpError ? error : coalesce_to_error(error);
 	const status = error instanceof HttpError ? error.status : 500;
-	const body = handle_error_and_jsonify(event, options, error);
+	const body = await handle_error_and_jsonify(event, options, error);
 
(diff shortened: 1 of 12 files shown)


packages/kit/src/runtime/server/utils.js
```diff
@@ -138,7 +138,7 @@ export function handle_fatal_error(event, options, error) {
  * @param {import('types').RequestEvent} event
  * @param {import('types').SSROptions} options
  * @param {any} error
- * @returns {App.Error}
+ * @returns {import('../../../types/private.js').MaybePromise<App.Error>}
```
Rich-Harris, line 141:
```suggestion
 * @returns {import('types').MaybePromise<App.Error>}
```

Rich-Harris (approved):

PR #7797, opened 2022-11-24
[fix] handle redirects in handle hook while processing data request

### LINKED ISSUE #7787: Clientside link navigation doesn't respect redirects
### Describe the bug

When using SvelteKit's clientside link navigation, redirects are not being respected unless the link has the `data-sveltekit-reload` attribute. If a page returns a redirect, Kit just throws a 500 Internal Error for the user (no message on the server). You can verify that the responses are getting sent correctly and have 3xx status codes by checking devtools.

This happens both when redirecting from `hooks.server.ts` and `+page.server.ts`.

When using the `data-sveltekit-reload` or manually refreshing the page, the redirect works fine.

It doesn't seem to matter how the redirect is sent: Kit's `redirect()` error, `Response(null, { status: 302, headers: { Location: '/login' } })`, and `Response.redirect(event.url.origin + '/login', 302);` all give the same result.

I couldn't find any docs on this, so I assume this is either unintended behavior or I'm doing something wrong.

### Reproduction

https://github.com/styxpilled/sveltekit-navigation-redirect-issue

### Logs

_No response_

### System Info

```Shell
System:
    OS: Windows 10 10.0.19044
    CPU: (4) x64 Intel(R) Core(TM) i3-9100F CPU @ 3.60GHz
    Memory: 453.96 MB / 15.94 GB
  Binaries:
    Node: 18.11.0 - ~\AppData\Local\pnpm\node.EXE
    npm: 8.19.2 - ~\AppData\Local\pnpm\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (107.0.1418.56)
```

### Severity

annoyance

### Additional Information

_No response_

dummdidumm:
Fixes #7787

During fixing this, I wondered why the tests also pass without the fix, and discovered that this specific behavior will only occur if there's no root `+layout.server.js` (because then it wouldn't try to rerequest the data, failing again, resulting in a fallback to the server where the redirect would be handled correctly), which we have in the tests. I'm not sure if it's worth creating another test suite for this, so I left it like this.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -13,7 +13,7 @@ import {
 	strip_data_suffix
 } from '../../utils/url.js';
 import { exec } from '../../utils/routing.js';
-import { INVALIDATED_HEADER, render_data } from './data/index.js';
+import { INVALIDATED_HEADER, redirect_json_response, render_data } from './data/index.js';
 import { add_cookies_to_headers, get_cookies } from './cookie.js';
 import { create_fetch } from './fetch.js';
 import { Redirect } from '../control.js';
@@ -289,10 +289,23 @@ export async function respond(request, options, state) {
 			}
 		}
 
+		// Edge case: If user does `return Response(30x)` in handle hook while processing a data request,
+		// we need to transform the redirect response to a corresponding JSON response.
+		if (is_data_request && response.status >= 300 && response.status < 309) {
+			const location = response.headers.get('location');
+			if (location) {
+				return redirect_json_response(new Redirect(/** @type {any} */ (response.status), location));
+			}
+		}
+
 		return response;
 	} catch (error) {
 		if (error instanceof Redirect) {
-			return redirect_response(error.status, error.location);
+			if (is_data_request) {
+				return redirect_json_response(error);
+			} else {
+				return redirect_response(error.status, error.location);
+			}
 		}
 		return handle_fatal_error(event, options, error);
 	}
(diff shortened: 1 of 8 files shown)


styxpilled:
Reading this surprised me, because I found this issue in a project that had `+layout.server.js`.
After some tests I here's what I found:
- if you redirect in `hooks.server.js` and you don't have `+layout.server.js`, kit will load that route. If you have `+page.server.js` in that route, it will fail with a 500 error.
- if you have `+layout.server.js`, the redirect will work.
- but then if you don't have `+page.server.js`, the redirect won't work. 
  - if you have a `+page.js` that fetches some stuff from an endpoint (which you may also be blocking with hooks) that fails, the page will fail with a 500 error.
  - if you don't have that, the page will display as normal, even though it shouldn't.

Rich-Harris (approved):

packages/kit/src/runtime/server/index.js
```diff
@@ -289,10 +289,23 @@ export async function respond(request, options, state) {
 			}
 		}
 
+		// Edge case: If user does `return Response(30x)` in handle hook while processing a data request,
+		// we need to transform the redirect response to a corresponding JSON response.
+		if (is_data_request && response.status >= 300 && response.status < 309) {
```
Rich-Harris, line 294: nit

```suggestion
		if (is_data_request && response.status >= 300 && response.status <= 308) {
```

Issue #587, opened 2021-03-21
If `files.assets === 'public'`, Vite copies the directory's contents


Rich-Harris:
This potentially conflicts with our own static asset handling

benmccann:
Not sure if it's the same issue, but I saw @GrygrFlzr filed https://github.com/vitejs/vite/issues/2587 about Vite's public directory as well

jthegedus:
Hi, this issue is the cause for an annoying workflow/bug with the [Firebase adapter](https://github.com/jthegedus/svelte-adapter-firebase/issues/22).

The adapter reads the user's `firebase.json` configuration file to identify which dir contains static assets to push to the Firebase CDN, in Firebase this defaults to `public` and is almost never changed by the user. This results in the following process:

- `svelte-kit build` outputs `.svelte/output/client/_app/*`
- Firebase adapter then executes `copy_static_files` and `copy_client_files` which copies the `static` dir and `.svelte/output/client/_app/*` dir contents to `public`
- on subsequent `svelte-kit build` executions Vite then copies `public` contents into `.svelte/output/client/_app/*`, then outputs the new build as well, resulting in .svelte/output/client/_app/*` and .svelte/output/client/_app/_app/*`. This then gets copied by the adapter `copy_static_files` and `copy_client_files` into `public`.
- each re-run of `svelte-kit build` produces a further nested `_app` dir

Is it possible for my adapter to specifically disable this Vite config or should we disable it within SvelteKit, because it is causing a conflict for my use case.

jthegedus:
The current workaround for the [Firebase adapter is documentation](https://github.com/jthegedus/svelte-adapter-firebase#quickstart) to users to add

```
"build": "npx -y rimraf <dir used in firebase.json.hosting.public> && svelte-kit build --verbose",
```

to their `package.json`. Again, the default value is `public`, but ideally this would be something these users didn't have to worry about.

theodorejb:
I am also running into this issue. Is there an official workaround or fix in the works?

theodorejb:
I was able to work around this issue by setting the following in `svelte.config.js`:

```js
const config = {
    // ...
    kit: {
        // ...
        vite: {
            publicDir: false,
        },
    },
};
```

Would it make sense for SvelteKit to set `publicDir` to `false` by default to avoid the current unexpected behavior?

bluepuma77:
Shouldn't this be resolved in latest `SvelteKit v1.0.0-next.295` from npm?

[...]

Issue #7800, opened 2022-11-24
Remove `getStaticDirectory()`


Rich-Harris:
### Describe the problem

As with #7798, `getStaticDirectory()` is unused by official adapters. Static assets are included in `getClientDirectory()`, and the adapter has a list of files in `static`, so it's not necessary.

It's quite likely that we'll want to support `static: ['static', process.env.THEME]` or something like that in future, and `getStaticDirectory()` would make that somewhat tricky.

### Describe the proposed solution

Check community adapters aren't using `getStaticDirectory()`, and remove it.

### Importance

would make my life easier

Rich-Harris:
closed via #7809

PR #7809, opened 2022-11-25
Removed getStaticDirectory() 

mjh316:

Removes getStaticDirectory() because of no more usage, as stated in #7800.

### Tests

### Changesets

### DIFF
--- a/.changeset/grumpy-maps-complain.md
+++ b/.changeset/grumpy-maps-complain.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/kit": patch
+---
+
+[breaking] remove `getStaticDirectory()` from builder API
--- a/packages/kit/src/core/adapt/builder.js
+++ b/packages/kit/src/core/adapt/builder.js
@@ -126,10 +126,6 @@ export function create_builder({ config, build_data, routes, prerendered, log })
 			return `${config.kit.outDir}/output/server`;
 		},
 
-		getStaticDirectory() {
-			return config.kit.files.assets;
-		},
-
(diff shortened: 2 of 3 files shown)


mjh316:
It seems [svelte-adapter-firebase](https://github.com/jthegedus/svelte-adapter-firebase) has [one usage](https://github.com/jthegedus/svelte-adapter-firebase/blob/main/src/index.js).
Other than that, all the adapters on https://sveltesociety.dev/components#adapters do not use it.

dummdidumm:
Thank you! We should ask them for what exactly they need it and if they can use something else to achieve what they want before merging this.

benmccann:
@jthegedus it would be interesting to get your input on this PR. thanks!

jthegedus:
We are using this in the Firebase adapter to resolve https://github.com/sveltejs/kit/issues/587 in our own way, but I believe we can work around it regardless. I need to do a big update/rewrite of the adapter now a lot of things have changed in Firebase and Kit.

Rich-Harris (approved):

Rich-Harris:
thank you!

benmccann:
@jthegedus I was just reading your comment (https://github.com/sveltejs/kit/issues/587#issuecomment-831585946) on that issue to make sure we didn't cause you any trouble by merging this. What I would recommend is copying the assets from the output directory. The output directory will contain both the generated application files and assets from the `static` directory which should probably all be served by the CDN as they'll all static assets. Let me know if that would work or if there's anything I can do otherwise to help you get things running on the latest version of SvelteKit

PR #7811, opened 2022-11-25
Make console.warn named function

wtachau:
We have some code that relies on introspecting the `name` of `console.warn`— some metaprogramming to provide dev-only logs. This update to make `console.warn` an anonymous function means it no longer has a name, which breaks that assumption.

Normally:
```js
console.warn.name
> 'warn'
```
In SvelteKit dev:
```js
console.warn.name
> ''
```

### DIFF
--- a/.changeset/itchy-beers-suffer.md
+++ b/.changeset/itchy-beers-suffer.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Make console.warn wrapper named rather than anonymous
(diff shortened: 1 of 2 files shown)


dummdidumm:
What's the reason for this? Stack trace?

wtachau:
> What's the reason for this? Stack trace?

@dummdidumm Sorry for the delay! I created this PR in github.dev and it didn't prompt me for a description. Just updated the PR.

Rich-Harris (approved):

Rich-Harris:
thank you!

babhishek21:
BTW this eats up the call trace because `console.warn` is not bound to `console` anymore. 
image

Shall I make a PR to fix this? Or does this need an issue for discussion first?

PR #7812, opened 2022-11-25
[chore] don't hardcode port in test

benmccann:
just a minor cleanup I noticed while looking at the test that's failing the latest Playwright version


Rich-Harris:
should this use `baseURL`?

Rich-Harris (approved):

PR #7828, opened 2022-11-27
Client-side routing for `<form method="GET">`

### LINKED ISSUE #7251: `enhance` should progressively enhance `<form method="GET">`
### Describe the problem

Currently, `enhance` only works on forms where `method="POST"`. I would also like to progressively enhance a form where `method="GET"`, e.g. a search form.

For of an example of the kind of form where this would be useful, see the [REPL search form](https://github.com/sveltejs/sites/blob/e7567164fa5dc1a098d5e575ab2cfab7a5e27473/sites/svelte.dev/src/routes/(authed)/apps/%2Bpage.svelte#L85-L98) on svelte.dev. We currently have to write the `submit` callback ourselves.

It would be nice if we could `use:enhance` here as well to get the same behavior. This would also let us encourage people to further #useThePlatform by teaching progressive enhancement of `<form method="GET">` and `<form method="POST">`.

```html
<form
	on:submit|preventDefault={(e) => {
		const search = new FormData(e.target).get('search');
		goto(search ? `/apps?search=${encodeURIComponent(search)}` : '/apps');
	}}
>
	<input
		type="search"
		placeholder="Search"
		aria-label="Search"
		name="search"
		value={data.search}
	/>
</form>
```

Instead of calling `fetch`, when `method === 'GET'` the action would construct the query string from the form data and call `goto(action)`, i.e. call the `load` function for the given action. Since this is a navigation and no data is returned, `applyAction` would not be called and the `form` property would not be updated.

So on submit, the following form would call `goto('/search?q=cats')`, the same as the browser would do if JS was not available. We call any `+page.js` or `+page.server.js` `load` function and navigate to the page at that route (what `goto` normally does).

```
<form method="GET" action="/search" use:enhance>
	<input name="q" value="cats">
	<button>Submit</button>
</form>
```

`enhance` should also be updated to take `<button formmethod="GET">` into account, likely using `event.submitter`.

I don't think a post-submit callback makes sense in this case, since it would trigger navigation. Any customizations could be implemented with `$navigating`, `beforeNavigate`, and `afterNavigate`.

One open question is do we allow customizing the options passed to `goto`? For instance, if the user wants to keep focus on the form with `goto(action, { keepfocus: true })`. Or do we have them implement the action themselves in that case?

### Alternatives considered

Do nothing and have people implement it themselves.

### Importance

nice to have

### Additional Information

_No response_

Rich-Harris:
Closes #7251. Just a failing test for now, because we need to refactor the `find_anchor` stuff to make this work, and there's no point embarking on that until #7776 is merged.

TODO:

* [x] add a `submit` handler (roughly described [here](https://github.com/sveltejs/kit/issues/7251#issuecomment-1324512643), except it needs an `event.defaultPrevented` check)
* [x] add a `form` navigation type (or `submit`? not sure), that is applied regardless of `willUnload`
* [x] docs

### Tests

### Changesets


packages/kit/src/runtime/client/client.js
```diff
@@ -1513,6 +1528,54 @@ export function create_client({ target, base }) {
...
+
+				const form = /** @type {HTMLFormElement} */ (
+					HTMLFormElement.prototype.cloneNode.call(event.target)
+				);
+
+				const submitter = /** @type {HTMLButtonElement | HTMLInputElement} */ (event.submitter);
+
+				const method = event.submitter?.hasAttribute('formmethod')
+					? submitter.formMethod
+					: form.method;
```
tcc-sejohnson, line 1542:
```suggestion
				const method = submitter?.formMethod ?? form.method;
```

would this work?

packages/kit/src/runtime/client/client.js
```diff
@@ -1513,6 +1528,54 @@ export function create_client({ target, base }) {
...
 			});
 
+			target.addEventListener('submit', (event) => {
+				if (event.defaultPrevented) return;
+
+				const form = /** @type {HTMLFormElement} */ (
+					HTMLFormElement.prototype.cloneNode.call(event.target)
+				);
+
+				const submitter = /** @type {HTMLButtonElement | HTMLInputElement} */ (event.submitter);
```
tcc-sejohnson, line 1538:
```suggestion
				const submitter = /** @type {HTMLButtonElement | HTMLInputElement | null} */ (event.submitter);
```

danger, will robinson

packages/kit/src/runtime/client/client.js
```diff
@@ -1513,6 +1528,54 @@ export function create_client({ target, base }) {
...
+
+				const method = event.submitter?.hasAttribute('formmethod')
+					? submitter.formMethod
+					: form.method;
+
+				if (method !== 'get') return;
+
+				const url = new URL(
+					event.submitter?.hasAttribute('formaction') ? submitter.formAction : form.action
```
tcc-sejohnson, line 1547:
```suggestion
					submitter?.formAction ?? form.action
```

packages/kit/src/runtime/client/client.js
```diff
@@ -1513,6 +1528,54 @@ export function create_client({ target, base }) {
...
+				);
+
+				if (is_external_url(url, base)) return;
+
+				const { noscroll, reload } = get_router_options(
+					/** @type {HTMLFormElement} */ (event.target)
```
tcc-sejohnson, line 1553:
```suggestion
					/** @type {HTMLFormElement | null} */ (event.target)
```

again, probably shouldn't pretend nullable things aren't possibly null and should just handle null in the function or return early

packages/kit/src/runtime/client/client.js
```diff
@@ -1513,6 +1528,54 @@ export function create_client({ target, base }) {
...
+				);
+				if (reload) return;
+
+				event.preventDefault();
+				event.stopPropagation();
+
+				// @ts-ignore `URLSearchParams(fd)` is kosher, but typescript doesn't know that
```
tcc-sejohnson, line 1560:
```suggestion
				// @ts-expect-error `URLSearchParams(fd)` is kosher, but typescript doesn't know that
```

in case TypeScript gets its shit together in the future and we can remove the comment

tcc-sejohnson (commented):
Also opening another (very small) PR against this for type improvements

tcc-sejohnson:
Opened #7883 for smol type improvements

packages/kit/src/runtime/client/client.js
Rich-Harris: almost. the reason it's like this is that `submitter.formAction` is always defined whether or not there's an `action` attribute; i thought the same thing would be true for `formMethod`. Turns out it's the empty string if the attribute is unspecified. So we should be able to do this (`||` instead of `??`):

```suggestion
				const method = submitter.formMethod || form.method;
```

packages/kit/src/runtime/client/client.js
Rich-Harris: Sadly we can't do this for the reason given above

packages/kit/src/runtime/client/client.js
dummdidumm: Don't we also need the `cloneNode` thing due to https://github.com/sveltejs/kit/issues/7593? And if we can simplify that, we should also simplify that in `runtime/app/forms.js`

packages/kit/src/runtime/client/client.js
dummdidumm: Won't work, see the code inside `runtime/app/forms.js` (the web platform is bonkers)

packages/kit/src/runtime/client/client.js
Rich-Harris: this is just TypeScript being dumb though, no? Could `event.target` ever actually be `null`?

dummdidumm: It's not null in this case though, we know it exists. The types don't, that's why they play it safe.

packages/kit/src/runtime/client/client.js
dummdidumm: We know this always be defined

packages/kit/src/runtime/client/client.js
```diff
@@ -1236,9 +1249,10 @@ export function create_client({ target, base }) {
 			observer.disconnect();
 
 			for (const a of target.querySelectorAll('a')) {
-				const { url, external, options } = find_anchor(a, base);
+				const { url, external } = get_link_info(a, base);
+				const options = get_router_options(a);
```
PatrickG, line 1253:
```suggestion
				if (external) continue;

				const options = get_router_options(a);
				if (options.reload) continue;
```

We should not get the router options if the link is external anyway (same logic as in the `preload` function.).

packages/kit/src/runtime/client/utils.js
```diff
@@ -80,12 +80,59 @@ const levels = {
...
+ */
+export function find_anchor(element, target) {
+	while (element !== target) {
+		if (element.nodeName.toUpperCase() === 'A') {
+			return /** @type {HTMLAnchorElement | SVGAElement} */ (element);
+		}
+
+		element = /** @type {Element} */ (parent_element(element));
+	}
+}
```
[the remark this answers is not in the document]
Rich-Harris: I've no doubt that's true, but unfortunately `closest()` doesn't work, because it can't find elements in parent shadow DOM — yet another reason adding it to the web platform was a mistake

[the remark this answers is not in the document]
Rich-Harris: (incidentally, `element.closest(selector)` will return `element` if it matches `selector`, so if this _did_ work — if we decided not to bother with shadow DOM, on the basis that the wounds are self-inflicted — we could dispense with the first `if` block)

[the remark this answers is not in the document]
PatrickG: > I've no doubt that's true, but unfortunately `closest()` doesn't work, because it can't find elements in parent shadow DOM — yet another reason adding it to the web platform was a mistake

Oh, makes sense. I didn't thought about that.


> (incidentally, `element.closest(selector)` will return `element` if it matches `selector`, so if this _did_ work — if we decided not to bother with shadow DOM, on the basis that the wounds are self-inflicted — we could dispense with the first `if` block)

I included this `element.nodeName.toUpperCase() !== 'A'` check because it was so much faster than `.clostest('a')` on an `HTMLAnchorElement` in my tests.

packages/kit/src/runtime/client/client.js
```diff
@@ -1236,10 +1249,14 @@ export function create_client({ target, base }) {
...
 			for (const a of target.querySelectorAll('a')) {
-				const { url, external, options } = find_anchor(a, base);
-
+				const { url, external } = get_link_info(a, base);
 				if (external) continue;
 
+				const options = get_router_options(a);
+				if (options.reload) continue;
+
+				if (external || options.reload) continue;
```
PatrickG, line 1258:
```suggestion
```

This is a remnant of my previous suggestion. Github didn't let me include this line in the selection because there was a deleted line in between.

ignatiusmb: Thanks, we also need to delete some of the empty line or lint would fail, removed in a separate commit.

PR #7831, opened 2022-11-27
another weirdly delayed typechecking failure

Rich-Harris:


Rich-Harris:
gonna do a cheeky self-merge since it's wreaking havoc on other PRs

PR #7841, opened 2022-11-28
feat: Typing improvements to preload PR

tcc-sejohnson:
Please consider these changes as 100% a suggestion -- I figured this was easier than making some super-complex GitHub comment. This should be safe to merge into the branch without additional changes.

This more explicitly states what typings we're going for in this utils file. As with most situations where we provide stricter types, this just helps us not shoot ourselves in the foot later by somehow introducing `undefined` where we expect null, adding a new value to `data-sveltekit-blah` without updating it everywhere it occurs, etc. The main improvement is just in the parsing and validation when we pull `data-sveltekit-{x}` values from DOM elements. Some mapped typing there ensures that we're both validating the value and that the data attribute we're trying to retrieve provides the values we want it to.


packages/kit/src/runtime/client/utils.js
```diff
@@ -26,34 +26,62 @@ export function scroll_state() {
...
 /**
  * @param {Element} element
- * @param {string} name
+ * @param {T} attributeName
+ * @returns {typeof anchor_attribute_valid_values[T][number] | null}
+ * @template {AnchorAttribute} T
+ */
+function getValidatedAttribute(element, attributeName) {
+	const value = element.getAttribute(attributeName);
+	return validateAttributeValue(element, attributeName, value);
```
gtm-nayan, line 52: Does this file be bundled by esbuild or rollup at build time? The latter [doesn't treeshake](https://rollupjs.org/repl/?version=3.5.0&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtb2R1bGVfMS5qcyUyMiUyQyUyMmNvZGUlMjIlM0ElMjJmdW5jdGlvbiUyMGZvbyh4KSUyMCU3QiU1Q24lMjAlMjAlMjBpZiUyMChmYWxzZSklMjAlN0IlNUNuJTIwJTIwJTIwJTIwJTIweCUyMCUyQiUzRCUyMDElM0IlNUNuJTIwJTIwJTIwJTdEJTVDbiUyMCUyMHJldHVybiUyMHglM0IlNUNuJTdEJTVDbiU1Q25leHBvcnQlMjBjb25zdCUyMHklMjAlM0QlMjBmb28oMiklM0IlMjIlN0QlMkMlN0IlMjJuYW1lJTIyJTNBJTIybW9kdWxlXzIuanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmVzJTIyJTJDJTIybmFtZSUyMiUzQSUyMm15QnVuZGxlJTIyJTJDJTIyYW1kJTIyJTNBJTdCJTIyaWQlMjIlM0ElMjIlMjIlN0QlMkMlMjJnbG9iYWxzJTIyJTNBJTdCJTdEJTdEJTJDJTIyZXhhbXBsZSUyMiUzQW51bGwlN0Q=) identity function calls so doing it this way might save some bytes.

```suggestion
	return __SVELTEKIT_DEV__ ? validateAttributeValue(element, attributeName, value) : value;
```

and the function below can skip the check.

tcc-sejohnson: Good call -- I'm not sure it'd be treeshaken either way, but this makes more sense regardless.

packages/kit/src/runtime/client/utils.js
```diff
@@ -26,34 +26,62 @@ export function scroll_state() {
...
-/** @type {Record<string, number>} */
 const levels = {
-	tap: PRIORITY_TAP,
-	hover: PRIORITY_HOVER,
-	viewport: PRIORITY_VIEWPORT,
-	page: PRIORITY_PAGE,
-	'': PRIORITY_HOVER,
-	off: -1
+	...PRELOAD_PRIORITIES,
+	'': PRELOAD_PRIORITIES['hover']
```
dummdidumm, line 84: Come style suggestion: use `.hover` instead

sites/kit.svelte.dev/src/lib/docs/server/type-info.js
```diff
@@ -0,0 +1,596 @@
+/* This file is generated by running `node scripts/extract-types.js`
+   in the packages/kit directory — do not edit it */
+export const modules = [
```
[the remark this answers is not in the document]
Rich-Harris: d'oh

PR #7852, opened 2022-11-28
Upgrade dependencies

benmccann:
I think only `adapter-auto` and `create-svelte` need an updated changeset


packages/adapter-auto/package.json
```diff
@@ -30,10 +30,11 @@
 		"check": "tsc"
 	},
 	"devDependencies": {
-		"@types/node": "^16.11.68",
-		"typescript": "^4.8.4"
+		"@sveltejs/kit": "workspace:*",
```
Rich-Harris, line 33: this dependency is new, is it needed? it's referenced in `index.d.ts` but that was previously taken care of by the `"paths"` in `tsconfig.json`. maybe this way is better, idk, but presumably we'd want to update that config?

packages/adapter-cloudflare-workers/package.json
```diff
@@ -29,13 +29,13 @@
 		"check": "tsc"
 	},
 	"dependencies": {
-		"@cloudflare/workers-types": "^4.0.0",
+		"@cloudflare/workers-types": "^4.20221111.1",
```
Rich-Harris, line 32: holy shit really? 🤔

Rich-Harris (commented):
just had one minor query

packages/adapter-auto/package.json
benmccann: Here's what I get without it:

Screenshot from 2022-11-28 12-50-42

I just removed the `paths` config instead

Rich-Harris (approved):

PR #7861, opened 2022-11-28
Use `satifies` operator

benmccann:
sucrase, prettier, and eslint all support TypeScript 4.9 now. However, it looks like VS Code is still on 4.8.4

### DIFF
--- a/.changeset/cool-seahorses-cross.md
+++ b/.changeset/cool-seahorses-cross.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+Use `satisfies` operator
--- a/documentation/docs/20-core-concepts/20-load.md
+++ b/documentation/docs/20-core-concepts/20-load.md
@@ -271,7 +271,7 @@ export async function load({ cookies }) {
 Both server-only and shared `load` functions have access to a `setHeaders` function that, when running on the server, can set headers for the response. (When running in the browser, `setHeaders` has no effect.) This is useful if you want the page to be cached, for example:
 
 ```js
-// @errors: 2322
+// @errors: 2322 1360
 /// file: src/routes/products/+page.js
 /** @type {import('./$types').PageLoad} */
 export async function load({ fetch, setHeaders }) {
--- a/documentation/docs/30-advanced/20-hooks.md
+++ b/documentation/docs/30-advanced/20-hooks.md
@@ -157,7 +157,7 @@ declare namespace App {
 
 ```js
 /// file: src/hooks.server.js
-// @errors: 2322 2571 2339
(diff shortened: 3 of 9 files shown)


Rich-Harris:
`Game` was moved into a separate file the other day so that it could be unit-tested. Looks like it's now duplicated?

benmccann:
Yeah, I caught that right as you did. I've updated this. Going to mark it as draft until VS Code supports TypeScript 4.9, but I think it's otherwise ready to go

benmccann:
VS Code now supports TypeScript 4.9 and `satisfies`: https://code.visualstudio.com/updates/v1_74#_typescript-49

Mine just updated to include support for it

Rich-Harris:
The docs still convert this...

image

...into this when you opt into TypeScript mode:

image

Ideally it would be `export async function load(...) {...} satisifes PageServerLoad`. Is it worth trying to get that working before merging this, so that we're singing from the same hymn sheet everywhere? @dummdidumm how complex would that be?

dummdidumm:
Not that complex, but it's not going to be `export async function load(...) {...} satisifes PageServerLoad` (that's sadly not supported yet for some reason), it's going to be `export const load = (async  (...) => {...}) satisifes PageServerLoad`

Rich-Harris:
ah yeah, turned out it was actually very simple: https://github.com/sveltejs/kit/pull/8001. unfortunately the site doesn't build with that change, hence separate branch rather than a commit to this one

dummdidumm (approved):
Giving this my approval, but a little hesitant to merge it just now. I'm not sure how fast/how many devs will have the latest VS Code version, so a good percentage could be confused by that.

Rich-Harris:
I feel like we ought to wait until the docs are in sync anyway

dummdidumm:
They are now, I merged your PR into this one

Rich-Harris:
oh! i totally missed that. nice. how long should we wait? I feel like VSCode updates probably propagate quite quickly, though I'm not sure if there's a way to check. Part of me thinks that a couple of questions in Discord/etc is a reasonable price to pay for smoother typechecking for people checking out 1.0 for the first time

benmccann:
I think I would lean towards trying to get this in now because a bunch of people will be trying to write tutorials, post/answer StackOverflow questions, etc. and all that content will be using outdated syntax if we release 1.0 without this. Merging this now will cause more questions on Discord from people who haven't updated VS Code yet, but I think it's probably the lesser of two evils and just a temporary hiccup

524c:
I'm watching this final race towards 1.0 🚀
I noticed that there is a typo in the title..

benmccann:
updated the title. luckily I got the changeset right first time around

PR #7862, opened 2022-11-28
fix handleError docs

Rich-Harris:
i think the TypeScript 4.9 bump broke this

### DIFF
--- a/documentation/docs/30-advanced/25-errors.md
+++ b/documentation/docs/30-advanced/25-errors.md
@@ -81,7 +81,7 @@ Unexpected errors will go through the [`handleError`](/docs/hooks#shared-hooks-h
 
 ```js
 /// file: src/hooks.server.js
-// @errors: 2322 2571 18046 2339
+// @errors: 2322 2571 2339
 // @filename: ambient.d.ts
 const Sentry: any;
 
--- a/documentation/docs/30-advanced/25-errors.md
+++ b/documentation/docs/30-advanced/25-errors.md
@@ -77,11 +77,11 @@ By default, unexpected errors are printed to the console (or, in production, you
 { "message": "Internal Error" }
 ```
 
-Unexpected errors will go through the [`handleError`](/docs/hooks#shared-hooks-handleerror) hook, where you can add your own error handling — for example, sending errors to a reporting service, or returning a custom error object. 
+Unexpected errors will go through the [`handleError`](/docs/hooks#shared-hooks-handleerror) hook, where you can add your own error handling — for example, sending errors to a reporting service, or returning a custom error object.
 
 ```js
 /// file: src/hooks.server.js
-// @errors: 2322 2571
+// @errors: 2322 2571 18046 2339
 // @filename: ambient.d.ts
 const Sentry: any;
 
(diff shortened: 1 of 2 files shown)


documentation/docs/30-advanced/25-errors.md
```diff
@@ -77,11 +77,11 @@ By default, unexpected errors are printed to the console (or, in production, you
 { "message": "Internal Error" }
 ```
 
-Unexpected errors will go through the [`handleError`](/docs/hooks#shared-hooks-handleerror) hook, where you can add your own error handling — for example, sending errors to a reporting service, or returning a custom error object. 
+Unexpected errors will go through the [`handleError`](/docs/hooks#shared-hooks-handleerror) hook, where you can add your own error handling — for example, sending errors to a reporting service, or returning a custom error object.
 
 ```js
 /// file: src/hooks.server.js
-// @errors: 2322 2571
+// @errors: 2322 2571 18046 2339
```
benmccann, line 84: we probably don't need 18046 and 2339. I think adding 18046 is probably better than using `?.` as I don't think error should really be `undefined`

Rich-Harris: it's `unknown`, which I think is as it should be

benmccann: I was thinking the user would define the type of `error` using `src/app.d.ts`: https://kit.svelte.dev/docs/errors#type-safety

If the user does that then I think the type would no longer be `unknown` and you don't need the `?.` or 2339

If we do want to keep the `?.` then I think you could probably remove 18046 as it's not included above

Rich-Harris: `App.Error` is the _output_ of `handleError`, not the input. It's `unknown` because TypeScript has no real way of knowing that some library code somewhere doesn't do `throw 'potato'`

documentation/docs/30-advanced/25-errors.md
```diff
@@ -94,7 +94,7 @@ export function handleError({ error, event }) {
 
 	return {
 		message: 'Whoops!',
-		code: error.code ?? 'UNKNOWN'
+		code: error?.code ?? 'UNKNOWN'
```
benmccann, line 97: I just created a new project, upgraded to TypeScript 4.9, and created a hook with this code. It's giving an error on this line telling me `error` is `unknown`

benmccann: It's stupid, but the TypeScript folks intend for this not to work: https://github.com/microsoft/TypeScript/issues/35799

benmccann: I can only get this to work is either changing the type to `any`:
```
/** @type {import('@sveltejs/kit').HandleServerError} */
	/** @type {any} */
	const err = error;
	return {
		message: 'Whoops!',
		code: err?.code ?? 'UNKNOWN'
	};
}
```

Or checking if the type is an `Error` and saying it might have a `code`:
```
/** @type {import('@sveltejs/kit').HandleServerError} */
	if (!(error instanceof Error)) {
		throw new Error('Invalid error type ' + error);
	}
	/** @type {Error & {code?: string}} */
	const err = error;
	return {
		message: 'Whoops!',
		code: err?.code ?? 'UNKNOWN'
	};
}
```

PR #7878, opened 2022-11-29
Disallow unknown exports

### LINKED ISSUE #7845: Reserve `config` export
### Describe the problem

We've alluded to a vague plan to support route-level deployment config in the future — something like this...

```js
export const config = {
  runtime: 'edge'
};
```

...where `config` is a freeform (but serializable) object that gets passed to adapters so that they can use it to customise the deployment.

At the moment, you can export anything from a `+(layout|page)(.server)?.js` file. This means that adding a new _known_ export would be a breaking change, as someone might already be doing `export const config = {...}` for some other reason.

Reserve the `config` name, i.e. throw an error if a route file exports a binding called `config`.

### Alternatives considered

Prevent _any_ unknown exports. This is safer, but might feel unduly restrictive.

### Importance

would make my life easier

### Additional Information

_No response_

Rich-Harris:
Closes #7845, by disallowing all unknown exports. For a minority of apps, this will be a breaking change, which is solved easily by moving the offending exports into a non-route file and importing them.

A related breaking change is that prerendering will fail (whether or not you're actively using prerendering) if modules cannot be imported for analysis at build time (e.g. because a file references a missing variable or something). This is a good thing, and something we should have done before, since it prevents surprising runtime errors.

### Tests

### Changesets


.changeset/friendly-jobs-perform.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] disallow unknown exports
```
Conduitry, line 5: This should be clearer about where unknown exports are disallowed. Is it all `+*.js` files?

Rich-Harris:
```suggestion
[breaking] disallow unknown exports from `+(layout|page)(.server)?.js` and `+server.js` files
```

.changeset/friendly-jobs-perform.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] disallow unknown exports from `+(layout|page)(.server)?.js` and `+server.js` files
```
Conduitry, line 5: Do we now want to call out the `_` exception in the changeset?

Either way, now that there's this exception, it should definitely be documented somewhere, and perhaps we could also make the error message nicer. ('If you're exposing this to facilitate testing, consider doing...')

dummdidumm, line 5:
```suggestion
[breaking] disallow unknown exports (except when starting with an underscore) from `+(layout|page)(.server)?.js` and `+server.js` files
```

dummdidumm (approved):

PR #7879, opened 2022-11-29
fix undefined comparison

524c:
I tested it in version 1.0.0-next.560 and 1.0.0-next.567 and there is an error in comparison with undefined.
In version 1.0.0-next.559 this does not happen.

### Tests

### Changesets

https://user-images.githubusercontent.com/527641/204590283-eb608052-e927-4368-a82b-99b0229fdc71.mp4

### DIFF
--- a/packages/kit/src/runtime/client/utils.js
+++ b/packages/kit/src/runtime/client/utils.js
@@ -98,7 +98,7 @@ export function find_anchor(element, base) {
 	/** @type {typeof valid_link_options['reload'][number] | null} */
 	let reload = null;
 
-	while (element !== document.documentElement) {
+	while (element && element !== document.documentElement) {
 		if (!a && element.nodeName.toUpperCase() === 'A') {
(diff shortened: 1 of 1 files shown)


packages/kit/src/runtime/client/utils.js
```diff
@@ -98,7 +98,7 @@ export function find_anchor(element, base) {
 	/** @type {typeof valid_link_options['reload'][number] | null} */
 	let reload = null;
 
-	while (element !== document.documentElement) {
+	while (element && element !== document.documentElement) {
```
benmccann, line 101: do you really need to make this change? are you getting `null` or `undefined` for `documentElement`?

dummdidumm: no, but if element is `null`, it wouldn't stop at this point

benmccann (approved):
This needs a changeset. Possibly a test. But the code lgtm

PR #7065, opened 2022-09-28
Allow fetching ArrayBufferView request body types

repsac-by:
- Fixes #5680 
- Fixes #5710 

### Please don't delete this checklist! Before submitting the PR, please make sure you do the following:
- [x] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [x] This message body should clearly illustrate what problems it solves.
- [x] Ideally, include a test that fails without this PR but passes with it.

### Tests
- [x] Run the tests with `pnpm test` and lint the project with `pnpm lint` and `pnpm check`

### Changesets
- [ ] If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. All changesets should be `patch` until SvelteKit 1.0

PR #7897, opened 2022-12-01
FormData or URLSearchParams as body in fetch in load function

WAKayser:
At the moment the only blocker for supporting FormData or URLSearchParams as the body of a fetch request, is the hashing function. This updates the hashing function to create a hash for these types of bodies. 

### Tests

### Changesets

### DIFF
--- a/.changeset/fuzzy-buses-call.md
+++ b/.changeset/fuzzy-buses-call.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Allow formdata and urlsearchparams as body for fetch in load functions.
--- a/packages/kit/src/runtime/hash.js
+++ b/packages/kit/src/runtime/hash.js
@@ -12,8 +12,19 @@ export function hash(value) {
 		const buffer = new Uint8Array(value.buffer, value.byteOffset, value.byteLength);
 		let i = buffer.length;
 		while (i) hash = (hash * 33) ^ buffer[--i];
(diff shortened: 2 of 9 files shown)


dummdidumm (approved):

dummdidumm (changes requested):
Reverting my approval - in #7065 `FormData` was applied, too, but was removed due to concerns of shipping unneeded JS for 99,9% of people.

Could you give a use case that needs support for this?

Trying to square the circle, I'm wondering if it would be an acceptable tradeoff to allow other kinds of data with a warning, but in such a way that their value can't be reused in hydration and need to rerun again. (in fact, this may happen here, too, since `build_selector` doesn't handle FormData/URLSearchParams currently)

WAKayser:
If it increases the bundle size,  this can also be removed in favor of a warning or something similar. 
Maybe a nicer solution would be to ignore these types of data for hashing and allow developers to invalidate these manually?

dummdidumm:
Could you clarify what your use case is? E.g. why do you need to do a POST request inside a `load` function?

WAKayser:
I need a post request inside the load function so that I can call an external API with some formdata. With this I can set the required filtering attributes for showing the correct data. Ideally this would not be done with a POST request, as it does not change any state. But unfortunately you cant always change API behavior.

WAKayser:
Ill close this for now. As there are enough workarounds possible

PR #7907, opened 2022-12-02
[docs] update reference to `getClientAddress()`

aradalvand:
The adapter-node README file mentions `RequestEvent.clientAddress`, but that's now a function: `RequestEvent.getClientAddress()`.

### Tests

### Changesets


benmccann:
https://github.com/sveltejs/kit/pull/6237 for reference

benmccann (approved):
I added a changeset so the docs will be fixed in the published package. I feel like we should probably separate docs from publishing, but I guess that's for another time

PR #7911, opened 2022-12-02
[fix] correct Vite config merging with force option

benmccann:
two suggestions from @dominikg:
- code was broken when `force` was used as we'd drop the other `optimizeDeps` options
- we don't need our own config merge function as Vite has one that's exported

### DIFF
--- a/packages/kit/src/exports/vite/utils.js
+++ b/packages/kit/src/exports/vite/utils.js
@@ -1,5 +1,5 @@
 import path from 'path';
-import { loadConfigFromFile, loadEnv } from 'vite';
+import { loadConfigFromFile, loadEnv, mergeConfig } from 'vite';
 import { runtime_directory } from '../../core/utils.js';
 import { posixify } from '../../utils/filesystem.js';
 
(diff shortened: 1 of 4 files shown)


bluwy (approved):

packages/kit/src/exports/vite/utils.js
```diff
@@ -61,16 +44,6 @@ export function deep_merge(...objects) {
...
 }
 
-/**
- * normalize kit.vite.resolve.alias as an array
- * @param {import('vite').AliasOptions} o
- * @returns {import('vite').Alias[]}
- */
-function normalize_alias(o) {
-	if (Array.isArray(o)) return o;
-	return Object.entries(o).map(([find, replacement]) => ({ find, replacement }));
```
dummdidumm, line 71: anyone knows why we did this previously?

dominikg: vite alias config has a variable type, can be an array or a map https://vitejs.dev/config/shared-options.html#resolve-alias

mergeConfig export of vite takes care of this internally

PR #7928, opened 2022-12-03
a11y improvements to kit.svelte.dev

geoffrich:
Some minor a11y improvements to the docs. 

The largest change was fixing the accessibility of the JS/TS toggle. Instead of a checkbox, I refactored it to use `<button aria-pressed>` instead. Using a checkbox isn't inaccessible, but a [button is a better fit for this usecase](https://adrianroselli.com/2019/08/under-engineered-toggles-too.html#Button):

>  If flipping the toggle immediately performs an action on the page (such as updating a setting), then a `<button>` is a better fit. Unlike a checkbox, which only changes its state, a `<button>` conveys to a user that something will happen. 

The label on the control also wasn't clear (it was "JavaScript TypeScript"), so I instead gave it the label "TypeScript code examples" (visually hidden). Combined with the state of `aria-pressed`, it will convey whether TS examples are on or off.

I also removed the `outline: none` so we have focus styles.

screenshot of toggle focus styles

Other changes:
- Give each `<nav>` a unique label (resolves [landmarks are unique](https://dequeuniversity.com/rules/axe/4.4/landmark-unique) Axe issue)
- Label search input
- Add `<h1>` to search page

With these changes, the only automated axe issues left are related to color contrast.

### Tests

### Changesets

### DIFF
--- a/sites/kit.svelte.dev/src/lib/docs/Contents.svelte
+++ b/sites/kit.svelte.dev/src/lib/docs/Contents.svelte
@@ -5,7 +5,7 @@
 	export let contents = [];
 </script>
 
-<nav>
+<nav aria-label="Docs">
 	<ul class="sidebar">
 		{#each contents as section}
 			<li>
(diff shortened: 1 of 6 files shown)


Rich-Harris (approved):

Rich-Harris:
🙌

Issue #6259, opened 2022-08-25
`import.meta.env` undefined after upgrade from 405 to 432


sidharthv96:
### Describe the bug

`import.meta.env` is undefined after upgrading to v432, causing vitest to fail. 

<img width="695" alt="image" src="https://user-images.githubusercontent.com/10703445/186577076-5555b221-0e80-4ab7-b009-139341664f4e.png">

### Reproduction

PR of the upgrade branch. Only contains changes for the upgrade.

- https://github.com/mermaid-js/mermaid-live-editor/pull/983 (Closed with @brev's fix)
- https://github.com/mermaid-js/mermaid-live-editor/pull/994 
### Logs

```shell
❯ node_modules/@sveltejs/kit/src/runtime/app/env.js:4:41
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Suites 1 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  src/lib/util/serde.test.ts [ src/lib/util/serde.test.ts ]
      2|  * @type {import('$app/env').browser}
      3|  */
      4| export const browser = !import.meta.env.SSR;
       |                                         ^
      5| 
      6| /**
 ❯ async /Users/sidv/dev/mermaid-live-editor/src/lib/util/stats.ts:1:256
 ❯ async /Users/sidv/dev/mermaid-live-editor/src/lib/util/state.ts:5:31
 ❯ async /Users/sidv/dev/mermaid-live-editor/src/lib/util/serde.test.ts:5:31

TypeError: CTest Files  1 failed | 1 passed (2)
     Tests  1 passed (1)
  Start at  10:12:58
  Duration  1.53s (transform 974ms, setup 110ms, collect 795ms, tests 15ms)

annot read properties of undefined (reading 'SSR')
[...]

PR #7933, opened 2022-12-04
fix: add   and  to optimizeDeps.exclude

dominikg:
to allow prebundling for libraries using them

see https://discord.com/channels/457912077277855764/1024975960564580383

note that optimizeDeps.exclude automatically excludes subpaths too, implementation in vite looks like this:  `i === exclude || i.startsWith(exclude+'/')`

cc @bluwy @babichjacob 

### Tests

### Changesets

### DIFF
--- a/.changeset/healthy-walls-move.md
+++ b/.changeset/healthy-walls-move.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+add $app and $env to optimizeDeps.exclude so that libraries using these work correctly when prebundled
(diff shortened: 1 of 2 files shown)


dominikg:
this PR may improve the situation for vitest unit tests using $app or $env too

bluwy (approved):

dummdidumm:
It doesn't fix #6259, but I think this solves #1485 (not completely but) for people who want to consume libraries using SvelteKit-specific stuff.

PR #7938, opened 2022-12-04
[docs] fix SearchBox and Nav styles

s3812497:
Fixes styling issues for:
* SearchBox invisible placeholder text.
* SearchBox item colour in light mode when hovering or focused.

Before:
Incorrect text colour for light mode search box styles
After:
CleanShot 2022-12-05 at 01 46 26@2x

Not the most elegant fix by adding more CSS variables :/ but I couldn't find any existing ones to match the original hover colours for the SearchBox. 

* Nav with overlapping media queries at 800px viewport.

Before:
Mobile menu is not hidden when adjusting viewport width

After:
CleanShot 2022-12-05 at 09 08 59@2x

### Tests

### Changesets

### DIFF
--- a/sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
+++ b/sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
@@ -229,7 +229,8 @@
 	}
 
 	input::placeholder {
-		color: rgba(0, 0, 0, 0.3);
+		color: var(--sk-text-3);
+		opacity: 0.3;
 	}
 
 	input:focus-visible {
@@ -330,18 +331,6 @@
 		padding: 1rem;
 	}
 
-	:root {
-		--sk-search-small-focus: hsla(0, 0%, 100%, 0.6);
-		--sk-search-strong-focus: hsl(0, 0%, 100%);
-	}
-
-	@media (prefers-color-scheme: dark) {
-		:root {
-			--sk-search-small-focus: var(--sk-text-2);
-			--sk-search-strong-focus: var(--sk-text-1);
-		}
-	}
-
 	a:hover {
 		background: rgba(0, 0, 0, 0.05);
 	}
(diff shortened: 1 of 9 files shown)


sites/kit.svelte.dev/src/lib/search/SearchBox.svelte
```diff
@@ -329,8 +325,22 @@
 		padding: 1rem;
 	}
 
+	:root {
+		--sk-search-bg-hover: hsl(0, 0%, 93%);
+		--sk-search-small-focus: hsla(0, 0%, 100%, 0.6);
+		--sk-search-strong-focus: hsl(0, 0%, 100%);
```
dummdidumm, line 331: I think we should try to reuse the existing tokens as much as possible. The third var for example exists as `--sk-back-1` for light mode already. It doesn't have to be exactly like the previous version if it means we can reuse existing tokens.

s3812497: Do you mean something like this?
```
:root {
    --sk-search-small-focus: var(--sk-back-4);
    --sk-search-strong-focus: var(--sk-back-2);
}
```
There aren't any tokens that are white in both light and dark mode. Should there be? Or is it ok to use a raw colour value?

sites/site-kit/src/lib/components/Nav.svelte
```diff
@@ -150,7 +150,7 @@
 		line-height: 1;
 	}
 
-	@media (max-width: 800px) {
+	@media (max-width: 799px) {
```
Rich-Harris, line 153: this needs to be 800px otherwise this happens 

image

```suggestion
	@media (max-width: 800px) {
```

Rich-Harris:
thank you!

sites/site-kit/src/lib/components/Nav.svelte
s3812497: Thank you Rich!
However, there's still one niggling problem that only happens in dark mode.

The mobile menu colour is different from the navbar and still sticks at 800px only.
SvelteKit different coloured navbar in dark mode

Is there a quick way to add this fix to remove the mobile menu background colour or should I open another PR?
remove mobile background colour from Nav component

Issue #7925, opened 2022-12-03
Routing conflicts occur when multiple optional parameters


shuaixr:
### Describe the bug

Routing conflicts occur when the routes folder structure is as follows:
```
.
├── +page.svelte
└── path_a
    └── [[parameter_b]]
        └── [[parameter_c]]
            ├── +page.svelte
            └── +page.ts

```

### Reproduction

1.Create a sveltekit project

2.Let the routes folder structure look like this:
```
.
├── +page.svelte
└── path_a
    └── [[parameter_b]]
        └── [[parameter_c]]
            ├── +page.svelte
            └── +page.ts

```

3.Run `npm dev`

### Logs

```Shell
error when starting dev server:
Error: The "/path_a/[[parameter_b]]/[[parameter_c]]" and "/path_a/[[parameter_b]]/[[parameter_c]]" routes conflict with each other
    at prevent_conflicts (file:///home/username/svelte_project/node_modules/.pnpm/@sveltejs+kit@1.0.0-next.571_svelte@3.53.1+vite@3.2.4/node_modules/@sveltejs/kit/src/core/sync/create_manifest_data/index.js:504:11)
    at create_routes_and_nodes (file:///home/username/svelte_project/node_modules/.pnpm/@sveltejs+kit@1.0.0-next.571_svelte@3.53.1+vite@3.2.4/node_modules/@sveltejs/kit/src/core/sync/create_manifest_data/index.js:292:2)
    at create_manifest_data (file:///home/username/svelte_project/node_modules/.pnpm/@sveltejs+kit@1.0.0-next.571_svelte@3.53.1+vite@3.2.4/node_modules/@sveltejs/kit/src/core/sync/create_manifest_data/index.js:25:28)
    at Module.create (file:///home/username/svelte_project/node_modules/.pnpm/@sveltejs+kit@1.0.0-next.571_svelte@3.53.1+vite@3.2.4/node_modules/@sveltejs/kit/src/core/sync/sync.js:44:24)
    at update_manifest (file:///home/username/svelte_project/node_modules/.pnpm/@sveltejs+kit@1.0.0-next.571_svelte@3.53.1+vite@3.2.4/node_modules/@sveltejs/kit/src/exports/vite/dev/index.js:57:35)
[...]

PR #7939, opened 2022-12-04
[fix]Continuous optionals should not throw conflict error

aolose:
fix issue [7925](https://github.com/sveltejs/kit/issues/7925)

folder structure [[a]]/[[b]]/+page will throw routes conflict error

the error from the function `prevent_conflicts`

here:
```
permutations = [
...permutations.map((x) => x + next),
...permutations.map((x) => x + `<${matcher}>${next}`)
];
```
will generate permutations [ '//', '/<*>/', '//<*>', '/<*>/<*>' ]
but  actually, it' impossible match route  /empty/[[b]] . so '//<*>' should be remove from result.

### Tests

### Changesets


dummdidumm (commented):
Thank you! Could you add a changeset and add a test in the `index.spec.js` file next to the changed file?

aolose:
> Thank you! Could you add a changeset and add a test in the `index.spec.js` file next to the changed file?

okay, done.

dummdidumm:
thanks for the test, however for this it's sufficient to create a unit test - you would create another unit test similar to the existing one in `create_manifest_data/index.spec.js` at line 718.

aolose:
> thanks for the test, however for this it's sufficient to create a unit test - you would create another unit test similar to the existing one in `create_manifest_data/index.spec.js` at line 718.

thank you, I'll try it.

dummdidumm (approved):
Thank you!

PR #7950, opened 2022-12-05
[fix] allow SvelteKit to be used without bundling

benmccann:
Banish `import.meta.env` from our codebase so that `@sveltejs/kit` works without bundling

Fixes https://github.com/vitest-dev/vitest/issues/2298 and a ton of other issues I've seen like this around the ecosystem

We should probably print a big loud warning when running `svelte-package` if anyone tries to use `import { browser, dev } from '$app/environment'` and tell them to use `esm-env` unless they intend for their package to only be used in SvelteKit (https://github.com/sveltejs/kit/issues/8033)


dominikg:
this is neat!  can we rely on `browser` and `development` conditions in exports maps to always be correctly resolved.
 As long as it happens in the vite build step i think that is the case, but what about edge runtimes, do they always get output of vite or another bundler? or would there be a point where we have to rely on them to do the right thing with these conditions?

dominikg:
using `!browser` for ssr a good step for compatibility :+1:   otherwise node vs deno vs x would be playing catchup.

benmccann:
> what about edge runtimes, do they always get output of vite or another bundler?

Yes. If they weren't then the code currently in `main` wouldn't work since those environments wouldn't know how to handle `import.meta.env` either

packages/environment/package.json
```diff
@@ -0,0 +1,25 @@
...
+			"browser": {
+				"development": "./dev-browser.js",
+				"default": "./prod-browser.js"
+			},
+			"default": {
+				"development": "./dev-ssr.js",
+				"default": "./prod-ssr.js"
+			}
+		},
+		"./package.json": "./package.json"
```
dominikg, line 23: do we need the package.json export? in a discussion in vite-plugin-svelte repo it was brought up that this isn't strictly needed.

benmccann: `rollup-plugin-svelte` still requires it at the moment, but we should work towards removing that. I figured it's good to keep it here until then as it doesn't hurt anything to have it

packages/environment/package.json
```diff
@@ -0,0 +1,25 @@
+{
+	"name": "@sveltejs/environment",
+	"version": "1.0.0-next.1",
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/sveltejs/kit",
+		"directory": "packages/environment"
+	},
+	"license": "MIT",
+	"homepage": "https://kit.svelte.dev",
```
dominikg, line 10: create readme in this package and point homepage there? esp. if we want the community to use this package standalone instead of via kit.

benmccann: done

packages/environment/README.md
```diff
@@ -0,0 +1,19 @@
+# @sveltejs/adapter-node
```
geoffrich, line 1:
```suggestion
# @sveltejs/environment
```

benmccann:
Thanks! I wasn't aware of `__SVELTEKIT_DEV__`. I've replaced it as well. However, doing so causes two tests to fail that I believe is due to a bug that Rich discovered in Vite 3 (https://github.com/vitejs/vite/issues/9203). I fixed it for Vite 4, so after we upgrade, I'll update this branch against `master` and I imagine those tests should start passing. I'll mark this as a draft until then

dominikg (approved):

PR #7953, opened 2022-12-05
[docs] fix iOS search box hidden by nav

### LINKED ISSUE #7947: Issues with search field in SvelteKit documentation (on iOS).
### Describe the bug

Some minor annoyances with the SvelteKit documentation on iOS:

1. It is impossible to tap the 'x' on the right that clears the search input. Instead, the hamburger menu always "steals" the tap. Workaround is to select the text and delete. Or keep tapping keyboard backspace until text is gone.
2. Perhaps related to the issue above, the search field seems to be covered by the navigation header. As you can see in the screen shots below, the input cursor overlaps the Svelte logo.
3. In dark mode, you cannot see the text you entered. This seems to only happen on iOS in dark mode. (Also tried Windows 11 Chrome and Android Chrome.) Happens in both Safari and Chrome. In the image below, you can see I entered 'test' into the search field, but you cannot see the search input text, only the input cursor.

image

Normally the search input is visible in light mode, but when in auto dark mode and forcing light mode via the iOS control center, it still invisible. Perhaps in normal light mode, the search input was low enough to see the text and 'x':
image

### Reproduction

Search documentation on https://kit.svelte.dev/ 

### Logs

```Shell
n/a
```

### System Info

```Shell
- iOS version 16.0.3
- iPhone SE (2020)
```

### Severity

annoyance

### Additional Information

_No response_

s3812497:
fixes #7947 
Moves the `<SearchBox>` component out of the `<main>` element because iOS has an issue with `z-index`.
The `<nav>` had a higher z-index than the container of the SearchBox so the search box gets hidden. It ignores the z-index of the search box.

### Tests

### Changesets

### DIFF
--- a/.changeset/modern-radios-develop.md
+++ b/.changeset/modern-radios-develop.md
@@ -1,0 +1,5 @@
+---
+'kit.svelte.dev': patch
+---
+
+fix iOS site searchbar overlapping nav
--- a/.changeset/modern-radios-develop.md
+++ b/.changeset/modern-radios-develop.md
@@ -1,5 +1,0 @@
----
-'kit.svelte.dev': patch
----
-
-fix iOS site searchbar overlapping nav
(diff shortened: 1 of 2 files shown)


Rich-Harris (approved):

Rich-Harris:
thank you!

PR #7958, opened 2022-12-06
[fix] handle missing server file for an action request

### LINKED ISSUE #7957: form `enhance` incorrectly parses response text as JSON
### Describe the bug

The form `enhance` returns an unhelpful error message whenever the server returns an error response. This is because it tries to parse the response as JSON while SvelteKit's error responses are text only. As a result, the original error message is obscured behind the failed JSON parsing.

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-qg7vng?file=src%2Froutes%2F%2Bpage.svelte&terminal=dev

### Logs

```Shell
500
Unexpected token 'C', "Cross-site"... is not valid JSON
SyntaxError: Unexpected token 'C', "Cross-site"... is not valid JSON
    at JSON.parse (<anonymous>)
    at deserialize (https://sveltejskittemplatedefaultqg7vng-gfaq--5173.local-credentialless.webcontainer.io/node_modules/@sveltejs/kit/src/runtime/app/forms.js?v=32fe10c2:21:22)
    at HTMLFormElement.handle_submit (https://sveltejskittemplatedefaultqg7vng-gfaq--5173.local-credentialless.webcontainer.io/node_modules/@sveltejs/kit/src/runtime/app/forms.js?v=32fe10c2:106:13)
untitled page
```

### System Info

```Shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.90 
    @sveltejs/kit: next => 1.0.0-next.571 
    svelte: ^3.53.1 => 3.53.1 
    vite: ^3.2.4 => 3.2.4
```

### Severity

annoyance

### Additional Information

_No response_

### LINKED ISSUE #7965: Unhandled form action when `+page.server` does not exist
### Describe the bug

> The offending code is inside `actions.js` and `runtime/server/page/index.js` (line 41, the "is action request but no `+page.server.js` exists" case is unhandled).

_Originally posted by @dummdidumm in https://github.com/sveltejs/kit/issues/7958#issuecomment-1339110939_

A fetch request to a form action is being handled incorrectly when the server file does not exist . This returns a `200` response with the HTML of the page itself. Ideally, a `405` no action exists error would be returned.

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-bk2z1s?file=svelte.config.js,src%2Froutes%2F%2Bpage.svelte

### Logs

_No response_

### System Info

```Shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.90 
    @sveltejs/kit: next => 1.0.0-next.571 
    svelte: ^3.53.1 => 3.53.1 
    vite: ^3.2.4 => 3.2.4
```

### Severity

annoyance

### Additional Information

_No response_

s3812497:
fixes #7965 
fixes #7957 

* Returns a `405` response for an action request when no server file exists for a page.
* Adds a JSON body for the "cross-origin POST" and "no actions exist" error responses.

### Tests

### Changesets

### DIFF
--- a/.changeset/witty-spies-occur.md
+++ b/.changeset/witty-spies-occur.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-fix `enhance` error message when form action doesn't exist
+fix `enhance` error message when form action doesn't exist or csrf is enabled
--- a/.changeset/witty-spies-occur.md
+++ b/.changeset/witty-spies-occur.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+fix `enhance` error message when form action doesn't exist
--- a/packages/kit/src/runtime/app/forms.js
+++ b/packages/kit/src/runtime/app/forms.js
@@ -1,4 +1,5 @@
 import * as devalue from 'devalue';
+import { error } from '../../exports/index.js';
 import { client } from '../client/singletons.js';
 import { invalidateAll } from './navigation.js';
 
@@ -103,7 +104,13 @@ export function enhance(form, submit = () => {}) {
 				signal: controller.signal
 			});
 
-			result = deserialize(await response.text());
+			const response_text = await response.text();
+
+			if (response.headers.get('content-type') !== 'application/json') {
+				throw error(response.status, response_text);
+			}
+
+			result = deserialize(response_text);
 		} catch (error) {
 			if (/** @type {any} */ (error)?.name === 'AbortError') return;
 			result = { type: 'error', error };
--- a/packages/kit/src/runtime/app/forms.js
+++ b/packages/kit/src/runtime/app/forms.js
@@ -104,13 +104,7 @@ export function enhance(form, submit = () => {}) {
 				signal: controller.signal
 			});
 
-			const response_text = await response.text();
-
-			if (response.headers.get('content-type') !== 'application/json') {
-				throw error(response.status, response_text);
-			}
-
-			result = deserialize(response_text);
+			result = deserialize(await response.text());
 		} catch (error) {
 			if (/** @type {any} */ (error)?.name === 'AbortError') return;
 			result = { type: 'error', error };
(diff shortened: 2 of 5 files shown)


dummdidumm:
Thank you - I'm not sure if this is the right fix though. Looking at the linked issue, that error response should have been JSON instead - the "Accept" header clearly states that. So we should fix this on the server instead. Are there other situations besides the reproducible in the issue where you experienced this bug?

dummdidumm:
Interesting - thanks for investigating! The Accept header should be honored in these cases, so in the progressive enhancement case where JSON is expected for the response, it should be sent as such.
The offending code is inside `actions.js` and `runtime/server/page/index.js` (line 41, the "is action request but no `+page.server.js` exists" case is unhandled).

s3812497:
Thanks for the guidance @dummdidumm ! I've added handling for a missing server file using the same "405 POST method not allowed. No actions exist for this page" response. Hopefully I've implemented this correctly.

dummdidumm:
Thank you! Ok so this brings up another design question: Right now the response is still not JSONified, which it should be, but if we do that, it ideally should honor the `App.Error` interface. So this should probably use `handle_error_and_jsonify` and return a JSON response from that (similar to what's happeneing further below in that method).

s3812497:
Yup, looks much better now. What about the plain text response from the CSRF check?
Should a JSON response also be made available to prevent the parsing error?
```js
	if (options.csrf.check_origin) {
		const forbidden =
			request.method === 'POST' &&
			request.headers.get('origin') !== url.origin &&
			is_form_content_type(request);

		if (forbidden) {
			const csrf_error = error(403, `Cross-site ${request.method} form submissions are forbidden`);
			// respond with json
			if (request.headers.get('accept') === 'application/json') {
				return json(csrf_error.body, { status: csrf_error.status });
			}
			return new Response(csrf_error.body.message, { status: csrf_error.status });
		}
	}
```

dummdidumm:
yeah that makes sense 👍

dummdidumm (approved):
Thank you!

PR #7960, opened 2022-12-06
Improve Sverdle a11y

geoffrich:
This PR makes some a11y improvements to the Sverdle demo.

**use hidden inputs** - previously, each letter guess was its own unlabeled input. There wasn't a great way to label them, and they weren't inputs you could directly interact with, so I changed them to hidden inputs instead and used `<div>`s for the tiles. This resolves 30 Axe errors.

**respect reduced motion** for confetti and jiggle animation.

**add visually hidden headings**. This resolves an Axe error (no h1) and also gives landmarks so assistive tech users can easily navigate between rows.

**add visually hidden text for letter state** - previously, whether a letter was a correct guess or not was solely conveyed visually. This adds visually hidden text to identify a letter's state to non-sighted users. The description is taken from the NYT Wordle's hidden text (present/absent/correct/empty)

**remove aria-selected** - this ARIA attribute is only valid [on gridcell, option, row and tab roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected) and was not being used on elements with those roles, so I removed it.

**increase color contrast** on selected tile - possibly controversial. The previous color used for a "correct" letter had [really poor contrast](https://contrast-ratio.com/#white-on-%2340b3ff) (2.3). The new color gets to 4.6, and I think it looks okay? The variable I changed is only used for Sverdle right now.

screenshot of sverdle game with new dark blue color

### Tests

### Changesets

### DIFF
--- a/.changeset/many-spoons-enjoy.md
+++ b/.changeset/many-spoons-enjoy.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+fix: improve Sverdle a11y
--- a/packages/create-svelte/templates/default/src/routes/styles.css
+++ b/packages/create-svelte/templates/default/src/routes/styles.css
@@ -8,7 +8,7 @@
 	--color-bg-1: hsl(209, 36%, 86%);
 	--color-bg-2: hsl(224, 44%, 95%);
 	--color-theme-1: #ff3e00;
-	--color-theme-2: #40b3ff;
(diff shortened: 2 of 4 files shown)


dummdidumm (approved):
Thank you! To me the new color is actually more pleasant to the eyes, so +1 on that change from me

Rich-Harris (approved):

PR #7962, opened 2022-12-06
fix: ActionData type - disallow `undefined`, allow `null`

### LINKED ISSUE #7961: ActionData is defined as `inferred | undefined` but can actually be `inferred | undefined | null`
### Describe the bug

The `ActionData` type automatically adds `undefined` to itself for obvious reasons, but in practice, when the page is rendered not-in-reponse-to a form action, the form is `null`. I opened a PR fixing the specific definition of `ActionData`, but I noticed that, while the `form` property `packages/kit/src/runtime/server/page` is strictly `something | null`, the `form` properties in `packages/kit/src/runtime/client` are _not_, and range between `something | undefined`, `something | undefined | null`, and `something | null`. I just want to reach a consensus on what's going on here before merging my simple PR -- if the types internally are futzed up, we're going to end up accidentally introducing `undefined` as a value for `form` at some point.

### Reproduction

https://stackblitz.com/edit/sveltejs-kit-template-default-7rhsve?file=src/routes/+page.svelte

### Logs

_No response_

### System Info

```Shell
System:
    OS: macOS 12.6
    CPU: (10) arm64 Apple M1 Pro
    Memory: 226.42 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.14.2 - ~/Library/Caches/fnm_multishells/13881_1670305350623/bin/node
    npm: 8.19.2 - ~/Library/Caches/fnm_multishells/13881_1670305350623/bin/npm
  Browsers:
    Chrome: 107.0.5304.121
    Safari: 15.6.1
  npmPackages:
    svelte: ^3.53.1 => 3.53.1
```

### Severity

### Additional Information

_No response_

tcc-sejohnson:
Closes #7961. Prior art: #6994.

### Tests

### Changesets

### DIFF
--- a/.changeset/beige-kangaroos-appear.md
+++ b/.changeset/beige-kangaroos-appear.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] adjust ActionData type
--- a/packages/kit/src/core/sync/write_types/index.js
+++ b/packages/kit/src/core/sync/write_types/index.js
@@ -401,7 +401,7 @@ function process_node(node, outdir, is_page, proxies, all_pages_have_load = true
 						? `./proxy${replace_ext_with_js(basename)}`
 						: path_to_original(outdir, node.server);
 
-					type = `Expand<Kit.AwaitedActions<typeof import('${from}').actions>> | undefined`;
+					type = `Expand<Kit.AwaitedActions<typeof import('${from}').actions>> | null`;
 				}
 			}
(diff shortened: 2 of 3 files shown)


dummdidumm (approved):

dummdidumm:
Thanks for pointing out the inconsistency about undefined/null - there was indeed one situation where undefined could sneak in, where it shouldn't.

dummdidumm:
almost forgot, this also needs a changeset

PR #7969, opened 2022-12-06
[breaking] add embedded option, turned off by default

### LINKED ISSUE #7940: GET form triggers an entire page reload if rendered outside of SvelteKit's root
### Describe the bug

If a form is rendered outside of SvelteKit's content, and triggers a navigation, the entire page reloads.

This might seem like an absurd thing to care about, but a search form inside of a modal, rendered in a portal is a reasonably common occurrence, and exactly how I discovered this. 

Incidentally, form *actions* appear to work fine. You can post from a form rendered in a portal, and the invalidation appears to work correctly, without the full reload 

### Reproduction

https://github.com/arackaf/svelte-kit-form-nav-repro

First textbox is rendered inside of SvelteKit, and works correctly. 

Second one is rendered in a portal, and triggers a full page reload.

### Logs

```Shell
n/a
```

### System Info

```Shell
System:
    OS: macOS 13.0.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 208.96 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.11.1 - /usr/local/Cellar/nvm/0.38.0/versions/node/v16.11.1/bin/node
    Yarn: 1.22.17 - /usr/local/Cellar/nvm/0.38.0/versions/node/v16.11.1/bin/yarn
    npm: 8.18.0 - ~/Documents/node_modules/.bin/npm
  Browsers:
    Chrome: 108.0.5359.94
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.90
    @sveltejs/kit: next => 1.0.0-next.571
    svelte: ^3.53.1 => 3.53.1
    vite: ^3.2.4 => 3.2.4
```

### Severity

### Additional Information

_No response_

dummdidumm:
## Migration

If you relied on the new behavior introduced in #7766, you can get that back by adding `embedded: true` to your `svelte.config.js`:

```diff
export default {
  kit: {
+    embedded: true
  }
}
```

## PR description

Closes #7940
Closes #4935
Adds a new option which defaults to false. If true, events related to navigation etc are listened to on the target element rather the html root

~Open questions~ solved: 
- ~approach sensible? I figured using the `define` feature for this is the most straightforward and treeshakes nicely~ it is
- ~using `data-sveltekit-root` instead (as discussed in the issue)? Advantage: no config option, more control ("I use popper.js with embedded Kit, and it adds itself to some tag higher than the target). Disadvantage: feels clunky, a little more startup code that might not be as reliable.~ option is the way to go

### Tests

### Changesets

### DIFF
--- a/.changeset/popular-crabs-shop.md
+++ b/.changeset/popular-crabs-shop.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] add embedded option, turned off by default
--- a/packages/kit/src/core/config/index.spec.js
+++ b/packages/kit/src/core/config/index.spec.js
@@ -76,6 +76,7 @@ const get_defaults = (prefix = '') => ({
 			checkOrigin: true
 		},
 		endpointExtensions: undefined,
+		embedded: false,
 		env: {
 			dir: process.cwd(),
 			publicPrefix: 'PUBLIC_'
--- a/packages/kit/src/core/config/options.js
+++ b/packages/kit/src/core/config/options.js
@@ -129,6 +129,8 @@ const options = object(
 				checkOrigin: boolean(true)
 			}),
 
+			embedded: boolean(false),
+
 			// TODO: remove this for the 1.0 release
 			endpointExtensions: error(
 				(keypath) => `${keypath} has been renamed to config.kit.moduleExtensions`
--- a/packages/kit/src/exports/vite/build/build_server.js
+++ b/packages/kit/src/exports/vite/build/build_server.js
@@ -66,6 +66,7 @@ export class Server {
 				check_origin: ${s(config.kit.csrf.checkOrigin)},
 			},
 			dev: false,
+			embedded: ${config.kit.embedded},
 			handle_error: (error, event) => {
 				return this.options.hooks.handleError({
 					error,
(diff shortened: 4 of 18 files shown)


arackaf:
🎉🥳🎊🍾

benmccann (changes requested):
we need a docs change still for this one

Rich-Harris (approved):

Rich-Harris:
I'm not sure how best to add docs for this beyond https://kit-svelte-e9gz8bfaa-svelte.vercel.app/docs/configuration#embedded? (Though I guess it should probably mention the `params` stuff. Oh, and we can skip sending `params` from the server unless `embedded: true`. Will update in a sec)

image

This is a niche enough thing that it probably doesn't warrant a new section of the docs — I think it's probably enough that people can find it via the config docs, and we can point people to it as and when needs arise.

Conduitry:
I'm in favor of not documenting this further than this note in the configuration docs, especially since the old default behavior was (it would seem) solely for the benefit of the NY Times and I'm sure they have other out-of-band ways of complaining to you.

arackaf:
Should we investigate which irresponsible individual at the NYT was shipping things to production using a pre-beta application framework?

johnnysprinkles:
Awesome, I just got SvelteKit 1.0.0-next.584 and can confirm #4935 is resolved. By the way, this is for a SvelteKit app related to the recently announced Apple Music Sing feature.

patriciatiffany:
I came here from the docs, trying to figure out what `embedded: true` does, exactly. I have an inkling it will be useful for my use cases, but that short phrase in the docs isn't quite enough for me to understand. Could I get some context?

PR #7979, opened 2022-12-07
[breaking] strip __data.json from url

### LINKED ISSUE #7926: hide `/__data.json` from the `url` object inside `handle`
### Describe the problem

I find it weird that the `url` object inside `handle` in `hooks.server.ts` differs from the `url` object passed to the route files.

In the end, the `/__data.json` information from the route is just for the `resolve` function to distinguish what it should return.

Use-cases I see problematic with the current approach:
 - redirecting in case of an unauthenticated request and saving the current information as a "redirectTo" parameter
   This works fine as long as you directly access `/dashboard/project1` in the browser, but if you would get redirected via a client-side navigation, the "redirectTo" parameter will be `/dashboard/project1/__data.json`
 - logging: when I want to track how many requests were made to `/dashboard/project1` I would also need to look for `/dashboard/project1/__data.json` entries (assuming I log all accesses with its url from within `hooks.server.ts`)

It is easy to make a workaround for those usecases and manually remove the `/data.json` at the end of the url, but you first have to discover that this "issue" exists. I see this as a hurdle for new or inexperienced users.

And I also see it problematic if e.g. for some reasons `__data.json` gets renamed to `__data.json5` or `__data.xml`  in the future to support other content formats. Then the workarounds will no longer work.

> Here is a small project demonstrating a problematic logging issue: https://stackblitz.com/edit/sveltejs-kit-template-default-sllc4q?file=src/hooks.server.ts
> 1. Click on the "page 1" link
> 2. Reload the page
> 3. See that both requests resulted in different log outputs

Use the same `url` object within `hooks.server.ts` (without `/data.json` at the end) and add a new attribute to the `event` object to distinguish between regular requests and data requests.

This would mean a breaking change, but the migration process should be fairly simple.

### Alternatives considered

Leave it as it is

### Importance

would make my life easier

### Additional Information

_No response_

dummdidumm:
This is breaking change for people who relied on the `__data.json` suffix being present on `event.url` which is now no longer the case. Use the new `isDataRequest` boolean instead.

```diff
export function handle({ event, resolve }) {
-  if (event.url.pathname.endsWith('__data.json') {
+  if (event.isDataRequest) {
    // ..
  }
  return resolve(event);
}
```

closes #7926

Open question: Should we also remove those things from the original `request` object? I vote for "no".

### Tests

### Changesets

### DIFF
--- a/packages/kit/test/apps/basics/src/hooks.server.js
+++ b/packages/kit/test/apps/basics/src/hooks.server.js
@@ -45,6 +45,17 @@ export const handle = sequence(
 		return resolve(event);
 	},
 	({ event, resolve }) => {
+		if (
+			event.request.url.includes('__data.json') &&
+			(event.url.pathname.endsWith('__data.json') || !event.isDataRequest)
+		) {
+			throw new Error(
+				'__data.json requests should have the suffix stripped from the URL and isDataRequest set to true'
+			);
+		}
+		return resolve(event);
+	},
+	({ event, resolve }) => {
 		if (event.url.pathname.includes('fetch-credentialed')) {
 			// Only get the cookie at the test where we know it's set to avoid polluting our logs with (correct) warnings
 			event.locals.name = /** @type {string} */ (event.cookies.get('name'));
--- a/.changeset/stupid-panthers-learn.md
+++ b/.changeset/stupid-panthers-learn.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-[breaking] strip \_\_data.json from url
+[breaking] strip `__data.json` from url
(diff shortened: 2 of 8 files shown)


packages/kit/test/apps/basics/src/hooks.server.js
```diff
@@ -44,6 +44,17 @@ export const handle = sequence(
 		event.locals.answer = 42;
 		return resolve(event);
 	},
+	({ event, resolve }) => {
+		if (
+			event.request.url.includes('__data.json') &&
```
Rich-Harris, line 49: this would prevent contrived-but-allowed things like `/__data.json/wheee`, no? Don't we just want to check the end of the pathname?

Conduitry: This is just part of a test, not the code that users are going to run.

Rich-Harris: i definitely knew that and wasn't just being absent-minded

PR #7983, opened 2022-12-07
[docs] give integrations their own page

benmccann:
Integrations are important enough they should probably show up in the lefthand sidebar. I also updated the content a bit


documentation/docs/60-appendix/05-integrations.md
```diff
@@ -0,0 +1,28 @@
+---
+title: Integrations
+---
+
+## Preprocessors
+
+Preprocessors transform your `.svelte` files before passing them to the compiler. E.g. if your `.svelte` file uses TypeScript and PostCSS, it must first be transformed into JavaScript and CSS so that the Svelte compiler can handle it. There are many [available preprocessors](https://sveltesociety.dev/tools#preprocessors).
```
Rich-Harris, line 7:
```suggestion
Preprocessors transform your `.svelte` files before passing them to the compiler. For example, if your `.svelte` file uses TypeScript and PostCSS, it must first be transformed into JavaScript and CSS so that the Svelte compiler can handle it. There are many [available preprocessors](https://sveltesociety.dev/tools#preprocessors).
```

Rich-Harris (approved):

PR #7986, opened 2022-12-07
[breaking] update esbuild to 0.16

dominikg:
vite-4 is going to be released with esbuild 0.16 too 

please see the https://github.com/evanw/esbuild/blob/main/CHANGELOG.md

there are some breaking changes in 0.16, marking this as breaking for the adapters even though output may ultimately be the same/very similar

### Tests

### Changesets

### DIFF
--- a/.changeset/mighty-houses-sing.md
+++ b/.changeset/mighty-houses-sing.md
@@ -5,4 +5,4 @@
 '@sveltejs/adapter-vercel': patch
 ---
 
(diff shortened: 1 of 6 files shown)


benmccann:
I was doing this as part of https://github.com/sveltejs/kit/pull/7543 to keep the versions in sync. I'm also upgrading Rollup. I'll go ahead and close this one so that we can keep everything together

PR #7990, opened 2022-12-08
[feat] split Vite plugin in two

benmccann:
The Storybook folks are currently working on Storybook 7 which has two nice improvements: you can use a Vite builder without installing Webpack at all and it has first-class support for SvelteKit! A wrinkle in the SvelteKit support is that SvelteKit runs in the same Vite instance and installs a middleware that ends up serving all the requests breaking the Storybook server. This splits the SvelteKit Vite plugin into two so that they can remove the middleware plugin - that way they still get the aliases and other things our plugin provides without taking over the whole server and breaking everything else.

This is arguably a breaking change if anyone is looking for `vite-plugin-svelte-kit`, but hopefully no one is doing that. It's a bit of an implementation detail that we've never publicly documented or claimed would be stable

### DIFF
--- a/packages/kit/src/exports/vite/index.js
+++ b/packages/kit/src/exports/vite/index.js
@@ -183,383 +183,387 @@ function kit() {
 	// TODO remove this for 1.0
 	check_vite_version();
 
-	return [{
-		name: 'vite-plugin-sveltekit-build',
+	return [
+		{
+			name: 'vite-plugin-sveltekit-build',
 
-		/**
-		 * Build the SvelteKit-provided Vite config to be merged with the user's vite.config.js file.
-		 * @see https://vitejs.dev/guide/api-plugin.html#config
-		 */
-		async config(config, config_env) {
-			vite_config_env = config_env;
-			svelte_config = await load_config();
+			/**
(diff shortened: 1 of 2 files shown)


dummdidumm (commented):
What do you think about moving each array entry into a separate function? Would decrease the indentation level which could help readability.

benmccann:
I can't easily move them into separate functions, but I was able to declare them as variables first and then create the array which avoided the extra level of indentation, so hopefully that works

packages/kit/src/exports/vite/index.js
```diff
@@ -74,7 +74,7 @@ export function sveltekit() {
  * - https://rollupjs.org/guide/en/#build-hooks
  * - https://rollupjs.org/guide/en/#output-generation-hooks
  *
- * @return {import('vite').Plugin}
+ * @return {Array<import('vite').Plugin>}
```
Rich-Harris, line 77:
```suggestion
 * @return {import('vite').Plugin[]}
```

Rich-Harris (approved):
just one minor comment but otherwise this LGTM

PR #7994, opened 2022-12-08
[breaking] make `sveltekit()` return a promise

bluwy:
In vite-plugin-svelte, it currently helps SvelteKit set the [fallback hydrate value](https://github.com/sveltejs/vite-plugin-svelte/blob/18a17bebe5a6804ab04cb1039b190dcdec2fd5a4/packages/vite-plugin-svelte/src/utils/options.ts#L304-L308). This PR moves the logic into kit itself so we can remove the hardcoded logic.

Returning a promise is supported since Vite 3 so no peer dep bump is needed. In practice this shouldn't break most SvelteKit apps, unless someone is calling `sveltekit()` programmatically and accessing the plugins individually.

This will likely conflict with https://github.com/sveltejs/kit/pull/7990

### Tests

### Changesets


packages/kit/src/exports/vite/index.js
```diff
@@ -74,12 +92,11 @@ export function sveltekit() {
  * - https://rollupjs.org/guide/en/#build-hooks
  * - https://rollupjs.org/guide/en/#output-generation-hooks
  *
+ * @param {{ svelte_config: import('types').ValidatedConfig }} options
+ *
```
benmccann, line 96:
```suggestion
```

packages/kit/src/exports/vite/index.js
```diff
@@ -59,9 +59,27 @@ const enforced_config = {
...
-/** @return {import('vite').Plugin[]} */
-export function sveltekit() {
-	return [...svelte(), kit()];
+/** @return {Promise<import('vite').Plugin[]>} */
+export async function sveltekit() {
+	const svelte_config = await load_config();
+	/** @type {import('@sveltejs/vite-plugin-svelte').Options} */
+	const vite_plugin_svelte_options = svelte_config.vitePlugin || {};
+
+	// Default hydratable to true
```
benmccann, line 68: should `hydratable: true` go before `...vite_plugin_svelte_options.compilerOptions`? If this is correct as is, then the comment should probably be changed from "Default" to "force" and we should warn if we override the user's config just as we do with `enforced_config`

bluwy: It shouldn't be overriding it since it does a `vite_plugin_svelte_options.compilerOptions?.hydratable == null` check before. We could skip the check and make it go before the `...` and it should work the same too.

bluwy: updated to put `hydratable: true` before `...`. I also found that I was doing the whole flow wrong 🙈

benmccann (approved):

PR #7997, opened 2022-12-08
`sizes` needs to be on the `source`-tag

PatrickG:
The `sizes` attribute needs to be on the `source`-tag.
See [source](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source#attr-sizes) vs [picture](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture#attributes) attributes.

### Tests

### Changesets

### DIFF
--- a/sites/kit.svelte.dev/src/routes/home/Hero.svelte
+++ b/sites/kit.svelte.dev/src/routes/home/Hero.svelte
@@ -11,9 +11,10 @@
 			<a class="cta" href="/docs/introduction">read the docs</a>
 		</div>
 
-		<picture class="hero-image" sizes="(min-width: 480px) 800px, (min-width: 1024px) 480px, 600px">
+		<picture class="hero-image">
 			{#each Object.entries(background.sources) as [format, images]}
 				<source
+					sizes="(min-width: 480px) 800px, (min-width: 1024px) 480px, 600px"
 					srcset={images.map((i) => `${i.src} ${i.w}w`).join(', ')}
 					type={'image/' + format}
 				/>


Rich-Harris:
Thank you. The sizes themselves were based on an old layout, so I've updated those. For the life of me though, I can't get the browser to actually load the 960px-wide image. It loads the 1440px-wide one every time, even on tiny screens. Any idea what's going on?

s3812497:
> For the life of me though, I can't get the browser to actually load the 960px-wide image. It loads the 1440px-wide one every time, even on tiny screens. Any idea what's going on?

Could it be related to retina screen? This would cause the browser to load the higher res images.

Rich-Harris:
That was my thought too, but I tried crazy breakpoints like 3000px and still nothing :/

sites/kit.svelte.dev/src/routes/home/Hero.svelte
```diff
@@ -11,11 +11,12 @@
 			<a class="cta" href="/docs/introduction">read the docs</a>
 		</div>
 
-		<picture class="hero-image" sizes="(min-width: 480px) 800px, (min-width: 1024px) 480px, 600px">
+		<picture class="hero-image">
 			{#each Object.entries(background.sources) as [format, images]}
 				<source
+					sizes="(min-width: 768) 1440px, 960px"
```
s3812497, line 17:
```diff
- sizes="(min-width: 768) 1440px, 960px"
+ sizes="(min-width: 768px) 1440px, 960px"
```
After adding the missing `px` I've observed that Chrome, Safari, and Firefox all have different behaviours for this (tested on MacOS).
1. Firefox is the only browser that correctly implements this.
* Respects the `sizes` attribute.
* Changes the image when resizing the window.
2. Safari.
* Respects the `sizes` attribute.
* Does **not** change the image when resizing the window (it only loads the correct image on initial page load).
3. Chrome
* Does **not** respect the `sizes` attribute (always loads the largest image).
* Removing the `sizes` attribute enables `srcset` to work correctly (loads the next largest image on resize).

Would be great if someone else could verify these behaviours too.

PatrickG: I can not test on Safari, but it seems Firefox behaves similarly to Chrome for me. It always loads the largest if the `sizes` attribute is present.

> Removing the sizes attribute enables srcset to work correctly (loads the next largest image on resize).

I would not say "correctly", but at least something happens.
Without the `sizes` attribute, Chrome switches to the large image when the screen width is >= 588px (DPR: 2.0) while Firefox switches to the large image when the screen width is >= 481 (DPR: 2.0)

Rich-Harris:
Thank you @PatrickG and @s3812497 for the thorough investigation — agree that removing `sizes` is probably the best solution.

I shouldn't complain — I'm old enough to remember IE6 — but the web platform continues to be a cruel prank. And they wonder why people turn to JavaScript instead of uSinG tHe PlAtfOrM

PR #8001, opened 2022-12-08
generate TS with `satisfies`

Rich-Harris:
would be nice to get this change in as it reduces the need for proxy types. the site won't build though, because (I think) shiki-twoslash is using an older TypeScript and doesn't know how to deal with the newfangled syntax. A cursory glance didn't show me how to fix it so I'm just going to leave this here for a moment


benmccann:
It looks like `shiki/twoslash` does not bundle typescript, so should just use the version we have. But perhaps we will need to ask @orta to add support for the `satisfies` keyword?

orta:
I can't see the deploy error messages, but to my knowledge there shouldn't be any issues with twoslash accepting satisfies as a keyword - twoslash doesn't put a constraint on TS I think, so it should just accept whatever version is in the repo

dummdidumm (approved):
Had nothing to do with Shiki, just some syntax error and then new errors popping up that needed silencing

PR #8003, opened 2022-12-08
[breaking] throw an error if load return value isn't a POJO

### LINKED ISSUE #7998: throw an error if `load` return value isn't a POJO
### Describe the problem

This sort of works but you really shouldn't do it:

```js
export function load() {
  return [];
}
```

Among other things, it conflicts with `App.PageData`, and falls apart when layout and page data are merged.

Throw an error if the value returned from `load` isn't a POJO (at least in dev). Technically a breaking change, so deserves a 1.0 milestone.

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

dummdidumm:
closes #7998

will do tests tomorrow

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/client/client.js
+++ b/packages/kit/src/runtime/client/client.js
@@ -677,6 +677,18 @@ export function create_client({ target, base }) {
 				try {
 					lock_fetch();
 					data = (await node.shared.load.call(null, load_input)) ?? null;
+					const not_object = typeof data !== 'object' && data != null;
+					if (not_object || Array.isArray(data) || data instanceof Response) {
+						throw new Error(
+							`a load function related to route '${route.id}' returned ${
+								not_object
+									? `a ${typeof data}`
+									: data instanceof Response
+									? 'a Response object'
+									: 'an array'
+							}, but must return a plain object (e.g '{ data: ...}')`
+						);
+					}
 				} finally {
 					unlock_fetch();
 				}
--- a/packages/kit/src/runtime/server/page/load_data.js
+++ b/packages/kit/src/runtime/server/page/load_data.js
@@ -59,6 +59,7 @@ export async function load_server_data({ event, state, node, parent }) {
 	});
 
 	const data = result ? await unwrap_promises(result) : null;
+	validate_load_response(data, /** @type {string} */ (event.route.id));
 
 	return {
 		type: 'data',
@@ -236,9 +237,10 @@ export async function load_data({
 		}
 	});
 
-	const data = await node.shared.load.call(null, load_event);
-
-	return data ? unwrap_promises(data) : null;
+	const result = await node.shared.load.call(null, load_event);
+	const data = result ? unwrap_promises(result) : null;
+	validate_load_response(data, /** @type {string} */ (event.route.id));
+	return data;
 }
 
 /**
@@ -256,4 +258,23 @@ async function stream_to_string(stream) {
 		result += decoder.decode(value);
 	}
 	return result;
+}
+
+/**
+ * @param {any} data
+ * @param {string} [routeId]
+ */
+function validate_load_response(data, routeId) {
+	const not_object = typeof data !== 'object' && data != null;
+	if (not_object || Array.isArray(data) || data instanceof Response) {
+		throw new Error(
+			`a load function related to route '${routeId}' returned ${
+				not_object
+					? `a ${typeof data}`
+					: data instanceof Response
+					? 'a Response object'
+					: 'an array'
+			}, but must return a plain object (e.g '{ data: ...}')`
+		);
+	}
 }
--- a/.changeset/light-lamps-crash.md
+++ b/.changeset/light-lamps-crash.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[breaking] throw an error on invalid load response


packages/kit/src/runtime/client/client.js
```diff
@@ -677,6 +677,18 @@ export function create_client({ target, base }) {
...
+						throw new Error(
+							`a load function related to route '${route.id}' returned ${
+								not_object
+									? `a ${typeof data}`
+									: data instanceof Response
+									? 'a Response object'
+									: 'an array'
+							}, but must return a plain object (e.g '{ data: ...}')`
+						);
+					}
```
Conduitry, line 691: Can we do this check only in dev mode?

Separately, can we be stricter about enforcing that this is a POJO? Something like `Object.getPrototypeOf(data) === Object.prototype`. (This would also cover the array and `Response` cases.) (Would that break any proxy stuff? I'm not sure whether we're doing any here.) We could also do this on the server.

packages/kit/src/runtime/client/client.js
```diff
@@ -677,6 +677,18 @@ export function create_client({ target, base }) {
...
+					const not_object = typeof data !== 'object' && data != null;
+					if (not_object || Array.isArray(data) || data instanceof Response) {
+						throw new Error(
+							`a load function related to route '${route.id}' returned ${
+								not_object
+									? `a ${typeof data}`
+									: data instanceof Response
+									? 'a Response object'
+									: 'an array'
+							}, but must return a plain object (e.g '{ data: ...}')`
```
geoffrich, line 689: A bit of a nitpick, but this would result in the data being accessing in the page via `data.data`, which we shouldn't encourage. Maybe suggest a non-data word instead?

```suggestion
							}, but must return a plain object (e.g '{ values: ... }')`
```

Conduitry: `values` makes sense as a suggestion if what they were returning was an array, but I'm not sure it does otherwise.

geoffrich: Ah yeah, I guess I'm thinking about this in the array context since that's been a common source of confusion.

Maybe ` { stuff: ... }` ?

dummdidumm: I went with `{ something: .. }`

packages/kit/src/runtime/client/client.js
dummdidumm: tested this out, a proxy still satisfies this check, so we should be able to use this check instead

packages/kit/src/runtime/client/client.js
```diff
@@ -673,10 +673,23 @@ export function create_client({ target, base }) {
...
+						throw new Error(
+							`a load function related to route '${route.id}' returned ${
+								typeof data !== 'object'
+									? `a ${typeof data}`
+									: data instanceof Response
+									? 'a Response object'
+									: Array.isArray(data)
+									? 'an array'
+									: 'a non-plain object'
+							}, but must return a plain object at the top level (e.g '{ something: ...}')`
```
Rich-Harris, line 690: what if we don't have a key at all?

```suggestion
							}, but must return a plain object at the top level (i.e. \`return {...}\`)`
```

packages/kit/src/runtime/server/page/load_data.js
```diff
@@ -257,3 +262,23 @@ async function stream_to_string(stream) {
...
+		throw new Error(
+			`a load function related to route '${routeId}' returned ${
+				typeof data !== 'object'
+					? `a ${typeof data}`
+					: data instanceof Response
+					? 'a Response object'
+					: Array.isArray(data)
+					? 'an array'
+					: 'a non-plain object'
+			}, but must return a plain object at the top level (e.g '{ something: ...}')`
```
Rich-Harris, line 281:
```suggestion
			}, but must return a plain object at the top level (i.e. \`return {...}\`)`
```

packages/kit/test/apps/basics/test/test.js
```diff
@@ -421,6 +421,38 @@ test.describe('Errors', () => {
...
+			if (javaScriptEnabled) {
+				await page.goto('/');
+				await app.goto('/errors/invalid-load-response');
+			} else {
+				await page.goto('/errors/invalid-load-response');
+			}
+
+			expect(await page.textContent('footer')).toBe('Custom layout');
+			expect(await page.textContent('#message')).toBe(
+				"This is your custom error page saying: \"a load function related to route '/errors/invalid-load-response' returned an array, but must return a plain object at the top level (e.g '{ something: ...}')\""
```
Rich-Harris, line 435:
```suggestion
				"This is your custom error page saying: \"a load function related to route '/errors/invalid-load-response' returned an array, but must return a plain object at the top level (i.e. `return {...}`)\""
```

packages/kit/test/apps/basics/test/test.js
```diff
@@ -421,6 +421,38 @@ test.describe('Errors', () => {
...
+			if (javaScriptEnabled) {
+				await page.goto('/');
+				await app.goto('/errors/invalid-server-load-response');
+			} else {
+				await page.goto('/errors/invalid-server-load-response');
+			}
+
+			expect(await page.textContent('footer')).toBe('Custom layout');
+			expect(await page.textContent('#message')).toBe(
+				"This is your custom error page saying: \"a load function related to route '/errors/invalid-server-load-response' returned an array, but must return a plain object at the top level (e.g '{ something: ...}')\""
```
Rich-Harris, line 453:
```suggestion
				"This is your custom error page saying: \"a load function related to route '/errors/invalid-server-load-response' returned an array, but must return a plain object at the top level (i.e. `return {...}`)\""
```

packages/kit/src/runtime/client/client.js
```diff
@@ -673,10 +673,23 @@ export function create_client({ target, base }) {
 				}
 			});
 
-			if (import.meta.env.DEV) {
+			if (__SVELTEKIT_DEV__) {
```
dummdidumm, line 676: btw as a drive-by I replaced the `import.meta.DEV` occurences withour own global

PR #8005, opened 2022-12-08
[site] optimize showcase images

benmccann:


sites/kit.svelte.dev/src/lib/Image.svelte
```diff
@@ -0,0 +1,17 @@
+<script>
+	export let src;
+	export let alt;
+	export let width = undefined;
+	export let height = undefined;
```
Rich-Harris, line 5: we should add types here

sites/kit.svelte.dev/vite.config.js
```diff
@@ -2,13 +2,29 @@ import { sveltekit } from '@sveltejs/kit/vite';
 import * as path from 'path';
 import { imagetools } from 'vite-imagetools';
 
+const supportedExtensions = ['png', 'jpg', 'jpeg'];
```
Rich-Harris, line 5:
```suggestion
import * as path from 'path';
import { imagetools } from 'vite-imagetools';

const supportedExtensions = ['.png', '.jpg', '.jpeg'];
```

sites/kit.svelte.dev/vite.config.js
```diff
@@ -2,13 +2,29 @@ import { sveltekit } from '@sveltejs/kit/vite';
...
 const config = {
 	assetsInclude: ['**/*.vtt'],
 
 	logLevel: 'info',
 
-	plugins: [imagetools(), sveltekit()],
+	plugins: [
+		imagetools({
+			defaultDirectives: (url) => {
+				const extension = url.pathname.substring(url.pathname.lastIndexOf('.') + 1);
```
Rich-Harris, line 16:
```suggestion
				const extension = path.extname(url.pathname);
```

Rich-Harris (changes requested):
if we do this then we probably want to have pngs rather than jpgs for the original screenshots, otherwise we're sacrificing quality needlessly

sites/kit.svelte.dev/src/lib/Image.svelte
benmccann: I've updated this as well as I can. I'm going to have to expose the `Picture` type from `vite-imagetools` before I can fully do it, but don't want to block on that

sites/kit.svelte.dev/vite.config.js
```diff
@@ -2,13 +2,29 @@ import { sveltekit } from '@sveltejs/kit/vite';
...
 	logLevel: 'info',
 
-	plugins: [imagetools(), sveltekit()],
+	plugins: [
+		imagetools({
+			defaultDirectives: (url) => {
+				const extension = path.extname(url.pathname);
+				if (supportedExtensions.includes(extension)) {
+					return new URLSearchParams({
+						format: 'avif;webp;' + extension,
```
Rich-Harris, line 19: ah crap, i guess this should be this now? though it seems to work anyway...

```suggestion
						format: 'avif;webp;' + extension.slice(1),
```

Rich-Harris: gonna leave it as-is, if it ain't broke etc

Rich-Harris (approved):

PR #8018, opened 2022-12-09
feat: Environment variable FAQ for Vercel adapter

tcc-sejohnson:
cc @Rich-Harris, discussed internally

A draft of the Vercel environment variables FYI in advance of removing documentation of their `VITE_` prefixed variants.

### Tests

### Changesets

### DIFF
--- a/packages/adapter-vercel/README.md
+++ b/packages/adapter-vercel/README.md
@@ -34,6 +34,34 @@ export default {
 };
 ```
 
+## Environment Variables
+
+Vercel makes a set of [non-prefixed deployment-specific environment variables](https://vercel.com/docs/concepts/projects/environment-variables#system-environment-variables) available. This makes them private environment variables, accessible from `$env/static/private` and `$env/dynamic/private` (sometimes -- more on that later), and inaccessible from their public counterparts. To access one of these variables from the client:
+
+```js
+// +layout.server.js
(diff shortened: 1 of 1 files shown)


packages/adapter-vercel/README.md
```diff
@@ -34,6 +34,34 @@ export default {
...
+<!-- +layout.svelte -->
+<script>
+  /** @type {import('./$types').LayoutServerData} */
+	export let data;
+</script>
+
+<p>This staging environment was deployed from {data.deploymentGitBranch}.</p>
+```
+
+Since all of these variables are available at buildtime on Vercel, we recommend using their `$env/static/private` varients for static replacement -- and you'll _have_ to if you're deploying to edge functions, as Vercel currenly will not include them dynamically at runtime.
```
Rich-Harris, line 63:
```suggestion
Since all of these variables are unchanged between build time and run time when building on Vercel, we recommend using `$env/static/private` — which will statically replace the variables, enabling optimisations like dead code elimination — rather than `$env/dynamic/private`. If you're deploying with `edge: true` you _must_ use `$env/static/private`, as `$env/dynamic/private` and `$env/dynamic/public` are not currently populated in edge functions on Vercel.
```

Rich-Harris (commented):
couple of suggested tweaks

PR #8046, opened 2022-12-09
[fix] set Vite base URL

benmccann:
Setting Vite's base URL makes the 404 message a bit different in dev vs prod, but that seems like a very minor issue. Setting it will make a number of things easier like probably allowing us to fix https://github.com/sveltejs/kit/issues/2958, fix Vite PWA trying to read the base path from Vite, etc.

There were a few places I added trailing slashes. This avoids unnecessary redirects and makes the tests run faster and be less flaky. That probably could have been a separate PR, but I found it while working on this, so just crammed them in here

### DIFF
--- a/packages/kit/test/apps/options/test/test.js
+++ b/packages/kit/test/apps/options/test/test.js
@@ -9,21 +9,15 @@ test.describe('base path', () => {
 	test('serves a useful 404 when visiting unprefixed path', async ({ request }) => {
 		const html = await request.get('/slash/', { headers: { Accept: 'text/html' } });
 		expect(html.status()).toBe(404);
-		if (process.env.DEV) {
-			// Vite's message
-			expect(await html.text()).toBe(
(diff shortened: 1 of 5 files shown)


packages/kit/test/apps/options/test/test.js
```diff
@@ -9,9 +9,15 @@ test.describe('base path', () => {
...
-			'Not found (did you mean <a href="/path-base/slash/">/path-base/slash/</a>?)'
-		);
+		if (process.env.DEV) {
+			// Vite's message
+			expect(await html.text()).toBe(
+				'The server is configured with a public base URL of /path-base - did you mean to visit <a href=\"/path-base/slash/\">/path-base/slash/</a> instead?'
+			);
+		} else {
+			// SvelteKit's message
+			expect(await html.text()).toBe('Not found (did you mean <a href=\"/path-base/slash/\">/path-base/slash/</a>?)');
```
Rich-Harris: i think if we're going to do this we should just use the same message as Vite

benmccann: done

Rich-Harris (approved):

PR #8053, opened 2022-12-10
add type declaration for vitePreprocess reexport

gtm-nayan:
aside: Using 'declare module' for these exports kind of feels like a hack to me, unless it's needed for the docs we should maybe be using "exports.types" alongside the js files listed in package.json but out of scope for this PR.

### Tests

### Changesets

### DIFF
--- a/.changeset/violet-donuts-share.md
+++ b/.changeset/violet-donuts-share.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+add declaration for vitePreprocess reexport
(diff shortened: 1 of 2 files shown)


bluwy (approved):

Rich-Harris:
> Using 'declare module' for these exports kind of feels like a hack to me

Agree with that sentiment, I can't remember if there's a reason we're doing in this way rather than generating `.d.ts` files in `prepublishOnly`. (Though writing them by hand is useful for one very specific reason — we can rename a function like `invalid` to `fail` in a way that will cause a red squiggly when you import the old function but that will _also_ allow us to export a function with the old name that errors at runtime with a helpful error.)

PR #8055, opened 2022-12-10
[fix] improve split between Vite plugins

benmccann:
Splitting the Vite plugin helped Storybook in dev mode. However, it's still broken in production mode because we set the input and output overriding Storybook's config. This PR moves that so that setting those properties only occurs in the "build" plugin and not the "setup" plugin


packages/kit/src/exports/vite/build/utils.js
```diff
@@ -170,6 +152,57 @@ export function get_default_build_config({ config, input, ssr, outDir }) {
...
+				output: {
+					format: 'esm',
+					entryFileNames: ssr ? '[name].js' : `${prefix}/[name]-[hash].js`,
+					chunkFileNames: ssr ? 'chunks/[name].js' : `${prefix}/chunks/[name]-[hash].js`,
+					assetFileNames: `${prefix}/assets/[name]-[hash][extname]`,
+					hoistTransitiveImports: false
+				},
+				preserveEntrySignatures: 'strict'
+			},
+			target: ssr ? 'node14.8' : undefined
```
Rich-Harris, line 186: i think this is out of date

```suggestion
			target: ssr ? 'node16.14' : undefined
```

packages/kit/src/exports/vite/index.js
```diff
@@ -634,8 +684,17 @@ function warn_overridden_config(config, resolved_config) {
  * @param {string[]} out used locally to compute the return value
  */
 function find_overridden_config(config, resolved_config, enforced_config, path, out) {
+	if (
+		typeof config === 'undefined' ||
+		config === null ||
+		typeof resolved_config === 'undefined' ||
+		resolved_config === null
+	) {
```
Rich-Harris, line 692: can you explain this logic change?

aside from the bigger picture, not sure why we're doing `typeof` checks here. can we just do this?

```suggestion
	if (config == null || resolved_config == null) {
```

packages/kit/src/exports/vite/dev/index.js
```diff
@@ -284,11 +284,6 @@ export async function dev(vite, vite_config, svelte_config) {
 		}
 	});
 
-	// set `import { version } from '$app/environment'`
-	(await vite.ssrLoadModule(`${runtime_prefix}/env.js`)).set_version(
-		svelte_config.kit.version.name
-	);
```
Rich-Harris, line 290: why don't we need this any more?

Rich-Harris: ah nvm, i see it's now in `configureServer`

packages/kit/src/exports/vite/index.js
benmccann: We now have two plugins with a `config` method which each provide a partial config, so some of the config values may be `undefined` as they're defined in the other plugin

Conduitry: We can probably still do `== null` for each though. The only time I think you'd need to do `typeof foo === 'undefined'` is if `foo` may or may not be a global variable and you were trying to avoid throwing an exception.

benmccann: ok, yeah, that's a good change

man, JS is weird. I guess I failed that question on the JS quiz. I thought that would match `''` or `0` if it matches both `null` and `undefined`, but I guess now

packages/kit/src/exports/vite/index.js
```diff
@@ -554,12 +547,69 @@ function kit({ svelte_config }) {
...
+				assets
+			});
 		}
 	};
 
 	/** @type {import('vite').Plugin} */
-	const plugin_middleware = {
-		name: 'vite-plugin-sveltekit-middleware',
+	const plugin_build = {
+		name: 'vite-plugin-sveltekit-build',
```
Rich-Harris, line 574: this naming feels kinda confusing to me — this applies in dev too, no?

benmccann: yeah. renamed it to bundle. maybe that's better?

Rich-Harris: not really — no bundling happens in `dev` — though i don't have a better suggestion off the top of my head just yet

benmccann: some stuff gets bundled (e.g. stuff in `node_modules`), but yeah I guess not everything does. how about `compile`? all the `.svelte` files are compiled even in dev. I just pushed a change to rename it

Rich-Harris (approved):

packages/kit/src/exports/vite/index.js
Rich-Harris: yeah, seems fine

PR #8034, opened 2022-12-09
make `<pre>` elements in docs scrollable

Rich-Harris:
closes #8032

Issue #8032, opened 2022-12-09
Docs code example sidescroll on mobile


joeldrake:
### Describe the bug

When browsing the sveltekit docs on mobile I can not sidescroll the code example blocks.

### Reproduction

Go to https://kit.svelte.dev/docs/accessibility#focus-management on an iphone. Try to sidescroll a code example.

### System Info

```Shell
Ios 16.1.1
```

### Severity

annoyance

benmccann:
Even https://kit.svelte.dev/docs/modules#$app-navigation-goto on desktop overflows and can't be read

PR #8056, opened 2022-12-10
[fix] make `<p> <code>` elements in docs scrollable

joeldrake:
This is a continuation of #8034

Basically I found another thing that did not sidescroll, here:
https://kit.svelte.dev/docs/advanced-routing

Can be seen on both mobile and desktop:
Screenshot 2022-12-10 at 07 57 00

Instead of doing an issue like before (#8032) I thought I would give it a go to fix it myself this time.

note:
The inline-flex and padding is to give the scrollable element the same appearance as before. I have tested my change on desktop chrome (osx) and on mobile safari (iphone ios) to great success.

### Tests

### Changesets

### DIFF
--- a/sites/site-kit/src/lib/styles/text.css
+++ b/sites/site-kit/src/lib/styles/text.css
@@ -56,6 +56,14 @@
 	background: transparent;
 }
 
+.text p code {
+	max-width: 100%;
+	display: inline-flex;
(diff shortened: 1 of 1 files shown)


Rich-Harris (approved):

Rich-Harris:
thank you!

PR #8063, opened 2022-12-10
Make vitePreprocess vs svelte-preprocess difference a bit clearer

kaisermann:
As commented on Discord, if we are specifying that there are two maintained official preprocessors, maybe we could provide an explanation about when to use one or the other. This uses `vite-plugin-svelte` docs as a reference and moves the differentiation part to after both preprocessors
image


benmccann:
oops. I didn't see this and just sent https://github.com/sveltejs/kit/pull/8065 too

PR #8069, opened 2022-12-10
[docs] add captions toggle to homepage video

geoffrich:
This PR enables the video captions by default and adds a toggle to turn them on and off. We don't have a ton of styling flexibility, but I did what I could. The visual "enabled" state of the toggle is just a text underline.

I applied the same "unused" styling as the sound toggle, to make it more obvious that it's available.

I wasn't sure where to get an icon for the CC toggle so it's just text for now (which seems to cause some alignment issues with the sound toggle in firefox). I think it looks okay, but happy to swap out for an icon if someone can provide an SVG.

We could disable the captions by default if necessary, but imo it's nice having them on by default so people can understand the video without needing to turn on sound or hunt for the CC button. Especially since we autoplay the video and there's no way to seek at the moment - this helps people not miss the beginning.

<details>
<summary>Screenshots of captions on different browsers</summary>

## Chrome

image

## Safari

image

## Firefox

image

## Mobile screen width (Chrome)

image

</details>

### Tests

### Changesets


sites/kit.svelte.dev/src/routes/home/Video.svelte
```diff
@@ -46,6 +53,19 @@
 			observer.disconnect();
 		};
 	});
+
+	/**
+	 * @param {HTMLTrackElement} node
+	 */
+	function set_cue_position(node) {
+		const cues = node.track.cues;
+		if (!cues) return;
```
Rich-Harris, line 62: I suspect this works because the .vtt file is loaded by the time the JS runs (basically the inverse of the problem in `onMount` here), but it _could_ presumably happen in the opposite order. Is there an event we can listen for if `cues` doesn't exist yet?

benmccann:
nice! the captions are really great to have. it's nice that I can get the value out of the video while it's muted and I also think it will grab the attention of visitors who might otherwise skip over it

sites/kit.svelte.dev/src/routes/home/Video.svelte
geoffrich: Good call - I was able to listen to the `<track>`'s `load` event if we don't have cues yet. I tested it by wrapping the `<track>` in an `{#if browser}` block and it seemed to work. https://github.com/sveltejs/kit/pull/8069/commits/922677fdf7712786f20ee8be8ad5f18d67f30736

Rich-Harris (approved):

Rich-Harris:
* added a pair of icons
* removed the `opacity: 1` on focus, because we now have the outline, and having the `opacity: 1` persist after you've interacted with the button next to a button that _doesn't_ have `opacity: 1` looks weird
* synced the `unused` class between the two buttons — as soon as you interact with one, they both fade into the background (the reason they begin unfaded is so that you know there _are_ buttons, but that's no longer necessary once you've interacted with one of them)

amazing work getting the subtitles to look good, i was very much resigned to them not doing so. i think having them visible when you first hit the page is a huge win.

PR #8095, opened 2022-12-12
more robust recovery from manifest errors

Rich-Harris:
follow-up to #8093:

* clears the error overlay once the error is fixed (by doing a full page reload — not sure if there's a better way?)
* removes the 'failed to update manifest' message — that's internal jargon we don't really use in a public-facing way
* removes the stack trace — it's all internal stuff that's no help to the user
* responds with an error page if the page is reloaded, instead of a) stumbling along with a stale manifest or b) in the case that the manifest was invalid from the start, failing with a cryptic internal error

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/exports/vite/dev/index.js
+++ b/packages/kit/src/exports/vite/dev/index.js
@@ -470,7 +470,9 @@ export async function dev(vite, vite_config, svelte_config) {
 					res.writeHead(500, {
 						'Content-Type': 'text/html; charset=utf-8'
 					});
-					res.end(error_template({ status: 500, message: manifest_error.message ?? 'Invalid routes' }));
+					res.end(
+						error_template({ status: 500, message: manifest_error.message ?? 'Invalid routes' })
+					);
 
 					return;
 				}
--- a/packages/kit/src/exports/vite/dev/index.js
+++ b/packages/kit/src/exports/vite/dev/index.js
@@ -36,6 +36,9 @@ export async function dev(vite, vite_config, svelte_config) {
 	/** @type {import('types').SSRManifest} */
 	let manifest;
 
+	/** @type {Error | null} */
+	let manifest_error = null;
+
 	/** @param {string} id */
 	async function resolve(id) {
 		const url = id.startsWith('..') ? `/@fs${path.posix.resolve(id)}` : `/${id}`;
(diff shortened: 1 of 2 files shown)


packages/kit/src/exports/vite/dev/index.js
```diff
@@ -51,18 +54,24 @@ export async function dev(vite, vite_config, svelte_config) {
...
+
+			if (manifest_error) {
+				manifest_error = null;
+				vite.ws.send({ type: 'full-reload' });
+			}
 		} catch (error) {
-			console.error(colors.bold().red('Failed to update manifest'));
+			manifest_error = /** @type {Error} */ (error);
+
+			console.error(colors.bold().red('Invalid routes'));
```
dummdidumm, line 65: This could be more than invalid routes (matchers for example) - I chose "manifest" because it was a more generic while not being completely opaque to the user

packages/kit/src/exports/vite/dev/index.js
```diff
@@ -447,6 +456,25 @@ export async function dev(vite, vite_config, svelte_config) {
...
+				};
+
+				if (manifest_error) {
+					console.error(colors.bold().red('Invalid routes'));
+					console.error(manifest_error);
+
+					res.writeHead(500, {
+						'Content-Type': 'text/html; charset=utf-8'
+					});
+					res.end(error_template({ status: 500, message: 'Invalid routes' }));
```
dummdidumm, line 473:
```suggestion
					res.end(error_template({ status: 500, message: manifest_error.message ?? 'Invalid routes' }));
```

packages/kit/src/exports/vite/dev/index.js
```diff
@@ -51,18 +54,24 @@ export async function dev(vite, vite_config, svelte_config) {
...
+			console.error(colors.bold().red('Invalid routes'));
 			console.error(error);
 			vite.ws.send({
 				type: 'error',
 				err: {
-					message: `Failed to udpate manifest: ${
-						/** @type {Error} */ (error)?.message ?? 'Unknown error'
-					}`,
-					stack: /** @type {Error} */ (error)?.stack ?? ''
+					message: manifest_error?.message ?? 'Invalid routes',
```
dummdidumm, line 70:
```suggestion
					message: manifest_error.message ?? 'Invalid routes',
```

packages/kit/src/exports/vite/dev/index.js
Rich-Harris: I think it's still okay — matchers _are_ part of your routes

dummdidumm (approved):

PR #8096, opened 2022-12-12
[fix] load errorTemplate from correct location

dummdidumm:
The alternative would be to move the "set a fallback error.html path if not set" logic into load_error_page, which @Rich-Harris added [here](https://github.com/sveltejs/kit/pull/6367/commits/d5c986bf28e889dddbcef4ab3af8ca5a8083f508)

### Tests

### Changesets

### DIFF
--- a/.changeset/four-ligers-help.md
+++ b/.changeset/four-ligers-help.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] load errorTemplate from correct location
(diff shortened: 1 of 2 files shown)


Rich-Harris:
I think the other option is preferable — `vite/dev/index.js` is already an unwieldy monster, so anything to keep logic out of there is a good thing, and the less we need to restart the server the better

Rich-Harris (approved):

PR #8113, opened 2022-12-12
[fix] don't de/encode initial and explicit string headers

### LINKED ISSUE #7929: Don't enforce url-encoding of cookies
### Describe the problem

With #7736 we moved from being opinionated that cookies should _not_ be url-encoded to being opinionated that they _should_ be url-encoded.  This means that authentication will now fail if we forward a cookie to a backend service that requires that cookies are _not_ url-encoded.  Url-encoding cookies may be very common but it is not universal so it makes sense to allow for either case.

We currently respect the `encode` option passed when calling `cookies.set()` so it's possible to set a cookie that is not url-encoded, but if that cookie later comes back from the client and we 'forward' it to a backend service using `fetch` it will get url-encoded.

Example flow, with a backend service that insists on non-url-encoded cookies:
```
backend service
     |
(response with non-url-encoded cookie) (cookie=foo+bar)
     |
     v
sveltekit node app
     |
(cookie set using cookies.set() with { encode: (value) => value }, i.e. no encoding) (cookie=foo+bar)
     |
     v
browser
     |
(request, returning cookie just as it was when the browser received it) (cookie=foo+bar)
     |
     v
sveltekit node app
     |
(external fetch request, url-encoding of cookie is enforced) (cookie=foo%2Bbar)
     |
     v
backend service declines authentication as it expects non-url-encoded cookie
```

Be unopinionated about whether cookies should be url-encoded.  For cookies sent by the browser don't decode them when we parse them and don't encode them later - just leave them as they are.  For explicit header cookies let people specify their own encoder function if they want to, just as they can when adding to `new_cookies` by calling `cookies.set()`.

```diff
// kit/packages/kit/src/runtime/server/cookie.js
export function get_cookies(request, url, dev, trailing_slash) {
	const header = request.headers.get('cookie') ?? '';
-	const initial_cookies = parse(header, { decode });
+	const initial_cookies = parse(header, { decode: (value) => value });  // i.e. don't decode
```
and

```diff
// kit/packages/kit/src/runtime/server/cookie.js
	/**
	 * @param {URL} destination
	 * @param {string | null} header
+	 * @param {function} header_encode
	 */
-	function get_cookie_header(destination, header) {
+	function get_cookie_header(destination, header, header_encode) {
		/** @type {Record<string, string>} */
		const combined_cookies = {};

		// cookies sent by the user agent have lowest precedence
		for (const name in initial_cookies) {
-			combined_cookies[name] = encode(initial_cookies[name]);
+			combined_cookies[name] = initial_cookies[name]; 
		}

		// cookies previous set during this event with cookies.set have higher precedence
		for (const key in new_cookies) {
			const cookie = new_cookies[key];
			if (!domain_matches(destination.hostname, cookie.options.domain)) continue;
			if (!path_matches(destination.pathname, cookie.options.path)) continue;

			const encoder = cookie.options.encode || encode;
			combined_cookies[cookie.name] = encoder(cookie.value);
		}

		// explicit header has highest precedence
		if (header) {
			const parsed = parse(header, { decode });
+			const encoder = header_encode || encode;
			for (const name in parsed) {
-				combined_cookies[name] = encode(parsed[name]);
+				combined_cookies[name] = encoder(parsed[name]);
			}
		}

		return Object.entries(combined_cookies)
			.map(([name, value]) => `${name}=${value}`)
			.join('; ');
	}

```
The `'serialized cookie header should be url-encoded'` test in kit/packages/kit/src/runtime/server/cookie.spec.js would need to be removed.

### Alternatives considered

Not doing any decoding/encoding of cookies by default, but if url-encoding cookies is very common it probably makes sense to leave that as the default behavior for `cookies.set()` and for explicit header cookies.

### Importance

would make my life easier

### Additional Information

If this change makes sense I'm willing to work on a PR.

dummdidumm:
fixes #7929

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/cookie.js
+++ b/packages/kit/src/runtime/server/cookie.js
@@ -8,10 +8,6 @@ import { normalize_path } from '../../utils/url.js';
  * @type {Record<string, Set<string>>} */
 const cookie_paths = {};
 
-// default encoding functions for header cookie values
-const encode = encodeURIComponent;
-const decode = decodeURIComponent;
-
 /**
  * @param {Request} request
  * @param {URL} url
@@ -20,23 +16,25 @@ const decode = decodeURIComponent;
  */
 export function get_cookies(request, url, dev, trailing_slash) {
 	const header = request.headers.get('cookie') ?? '';
-	const initial_cookies = parse(header, { decode });
+	const initial_cookies = parse(header, { decode: (value) => value });
 
 	const normalized_url = normalize_path(url.pathname, trailing_slash);
 	// Emulate browser-behavior: if the cookie is set at '/foo/bar', its path is '/foo'
 	const default_path = normalized_url.split('/').slice(0, -1).join('/') || '/';
 
 	if (dev) {
+		// TODO this could theoretically be wrong if the cookie was set unencoded?
+		const initial_decoded_cookies = parse(header, { decode: decodeURIComponent });
 		// Remove all cookies that no longer exist according to the request
 		for (const name of Object.keys(cookie_paths)) {
 			cookie_paths[name] = new Set(
 				[...cookie_paths[name]].filter(
-					(path) => !path_matches(normalized_url, path) || name in initial_cookies
+					(path) => !path_matches(normalized_url, path) || name in initial_decoded_cookies
 				)
 			);
 		}
 		// Add all new cookies we might not have seen before
-		for (const name in initial_cookies) {
+		for (const name in initial_decoded_cookies) {
 			cookie_paths[name] = cookie_paths[name] ?? new Set();
 			if (![...cookie_paths[name]].some((path) => path_matches(normalized_url, path))) {
 				cookie_paths[name].add(default_path);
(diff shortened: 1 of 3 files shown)


packages/kit/src/runtime/server/cookie.js
```diff
@@ -20,23 +16,25 @@ const decode = decodeURIComponent;
...
 	const header = request.headers.get('cookie') ?? '';
-	const initial_cookies = parse(header, { decode });
+	const initial_cookies = parse(header, { decode: (value) => value });
 
 	const normalized_url = normalize_path(url.pathname, trailing_slash);
 	// Emulate browser-behavior: if the cookie is set at '/foo/bar', its path is '/foo'
 	const default_path = normalized_url.split('/').slice(0, -1).join('/') || '/';
 
 	if (dev) {
+		// TODO this could theoretically be wrong if the cookie was set unencoded?
```
benmccann, line 26: I'm not sure it makes sense to add a todo because I don't know how we'd ever fix it. Maybe change the comment like yhis
```suggestion
		// This could be wrong if the cookie was set unencoded
		// but since it's just for printing a debug message we can live with that
```

dummdidumm: I have some nebulous thoughts how this could be done, for example through never decoding it and then updating that object with an unencoded version of the new cookies, but want to look into that later - so I'd like to keep the TODO for now (but you may very well be right!).

PR #8114, opened 2022-12-12
remove `site-kit` fork

Rich-Harris:
I moved the `site-kit` package back into its own repo — https://github.com/sveltejs/site-kit — with the changes from the fork in this repo, which we no longer need


sites/kit.svelte.dev/src/routes/+layout.svelte
```diff
@@ -2,7 +2,12 @@
 	import '@sveltejs/site-kit/styles/index.css';
 	import { browser } from '$app/environment';
 	import { page, navigating } from '$app/stores';
-	import { Icon, Icons, Nav, NavItem, PreloadingIndicator, SkipLink } from '@sveltejs/site-kit';
+	import Icon from '@sveltejs/site-kit/components/Icon.svelte';
+	import Icons from '@sveltejs/site-kit/components/Icons.svelte';
+	import Nav from '@sveltejs/site-kit/components/Nav.svelte';
+	import NavItem from '@sveltejs/site-kit/components/NavItem.svelte';
+	import PreloadingIndicator from '@sveltejs/site-kit/components/PreloadingIndicator.svelte';
+	import SkipLink from '@sveltejs/site-kit/components/SkipLink.svelte';
```
dummdidumm, line 10: Why did you chose to use deep imports here? I liked the previous version better, much cleaner/less import clutter. We would need to add a `index.js` in `lib` in site-kit and it should work.

dummdidumm: What kind of exports do you imagine not being treeshakeable? Feels like a premature optimization, I'd rather have that non-treeshakeable thing not be exported from `@sveltejs/site-kit` then, if such a thing ever shows up.

Rich-Harris: E.g. if we had a `<Docs>` component that had `import './doc-styles.css'`, stuff like that. Very easy to accidentally slip in

dummdidumm: But this case would be fine, wouldn't it? The Docs component would be treeshaken out of the page; and if it's about "styles should not leak into other parts of the page" then that's going to happen either way once the Docs component is loaded (those styles wouldn't be unloaded when imported like this).

Rich-Harris: What I'm saying is that a site that _didn't_ use `<Docs>` would still end up bundling `doc-styles.css`, because they're transitively imported and `.css` files are side-effectful

Rich-Harris: Also it's better to only load those styles when you actually visit the `/docs` page, which is what would happen with individual imports. With bundle imports all CSS loaded by _any_ component is necessarily loaded on any page that includes _one_ component

dummdidumm: Shouldn't tree shaking be smart enough to split the chunks so that only the needed components and its css are loaded for a page? Basically seing "Docs is unused, let's remove that component and therefore everything that it loads from that bundle"?

Rich-Harris: No, it doesn't work that way. Side-effects in transitively imported modules are applied, otherwise you're no longer optimising, you're changing the semantics of the code.

dummdidumm: Interesting, I thought it would behave differently (like eager tree shaking). TIL! I'm good then.

PR #8083, opened 2022-12-12
[fix] set platform to neutral for cloudflare build

benmccann:
closes https://github.com/sveltejs/kit/issues/8082

PR #8122, opened 2022-12-13
Revert "[fix] set platform to neutral when calling esbuild"

Rich-Harris:
Reverts sveltejs/kit#8083. Turns out this was a mistake, since it breaks packages like `cross-fetch` which assume that if you're not resolving `browser` modules, you want the Node version instead.

Since we bundle `esm-env`, the original reason for #8083 no longer applies.

### DIFF
--- a/packages/adapter-cloudflare-workers/index.js
+++ b/packages/adapter-cloudflare-workers/index.js
@@ -63,6 +63,7 @@ export default function ({ config = 'wrangler.toml' } = {}) {
 
 			await esbuild.build({
 				platform: 'browser',
+				conditions: ['worker', 'browser'],
 				sourcemap: 'linked',
 				target: 'es2020',
 				entryPoints: [`${tmp}/entry.js`],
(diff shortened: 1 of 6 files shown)


packages/adapter-cloudflare-workers/index.js
```diff
@@ -62,15 +62,14 @@ export default function ({ config = 'wrangler.toml' } = {}) {
 			);
 
 			await esbuild.build({
-				platform: 'neutral',
-				mainFields: ['module', 'main'],
-				conditions: ['worker'],
+				platform: 'browser',
```
dummdidumm, line 65: should we add `conditions: ['worker', 'browser']` as they do in their wrangler config (and to Vercel Edge functions as well)?

Rich-Harris: seems reasonable in the cloudflare case. not sure about vercel — someone might be doing something cloudflare-specific? the runtimes aren't 100% identical, might be safer to just play dumb for now

dummdidumm (approved):

PR #8131, opened 2022-12-13
[breaking] prerender shells when ssr false and prerender not false

### LINKED ISSUE #3966: Non-SSR'd pages should be prerendered by default
### Describe the problem

While we recommend that people use SSR as widely as possible, some apps don't. In these cases, there's no obvious reason not to prerender the shell page.

If a page is rendered with `ssr: false`, don't bail out during prerendering here, unless `export const prerender` is explicitly `false` (to provide an escape hatch when the app is doing something funky in `handle`): https://github.com/sveltejs/kit/blob/b5cf676d30515b2168238064169d754276a7d5bc/packages/kit/src/runtime/server/page/respond.js#L70-L76

### Alternatives considered

The alternative is to continue dynamically server-rendering empty pages. Totally fine, since rendering an empty page is cheap, but wasteful nonetheless.

### Importance

nice to have

### Additional Information

_No response_

dummdidumm:
Closes #3966

This is a breaking change insofar that pages that have `export const ssr = false` and not explicitly set `export const prerender = false` or `prerender.entries: []` will have their `handle` hooks run, which could be unexpected and cause unintended side effects. In that's the case for you, add `export const prerender = false` next to wherever you have `export const ssr = false`.

### Tests

### Changesets


packages/kit/src/runtime/server/page/index.js
```diff
@@ -79,13 +79,38 @@ export async function render_page(event, route, page, options, state, resolve_op
...
 			}
 		} else if (state.prerendering) {
+			// Try to render the shell when ssr is false.
+			// People can opt out of this behavior by explicitly setting prerender to false
+			if (
+				should_prerender !== false &&
+				get_option(nodes, 'ssr') === false &&
+				!leaf_node.server?.actions
+			) {
+				return await render_response({
```
dummdidumm, line 97: duplicated to code a little further below, but this way it's more readable (you don't need to hold all these conditions in your head until you arrive at the other condition)

packages/kit/src/runtime/server/page/index.js
```diff
@@ -79,13 +79,38 @@ export async function render_page(event, route, page, options, state, resolve_op
...
 			if (mod && mod.actions) {
 				throw new Error('Cannot prerender pages with actions');
 			}
 		} else if (state.prerendering) {
+			// Try to render the shell when ssr is false.
+			// People can opt out of this behavior by explicitly setting prerender to false
+			if (
+				should_prerender !== false &&
+				get_option(nodes, 'ssr') === false &&
+				!leaf_node.server?.actions
```
dummdidumm, line 95: These checks are duplicated to the checks in `prerendering.js`. Is there a way to deduplicate this? I don't see how.

packages/kit/src/core/prerender/prerender.js
```diff
@@ -401,7 +401,14 @@ export async function prerender() {
...
 
-			const prerender = get_option(nodes, 'prerender') ?? false;
+			const should_prerender = get_option(nodes, 'prerender');
+			const prerender =
+				// Try prerendering if ssr is false and no server needed.
+				// People can opt out of this behavior by explicitly setting prerender to false
+				(should_prerender !== false &&
+					get_option(nodes, 'ssr') === false &&
+					!page?.server?.actions) ||
+				(should_prerender ?? false);
```
dummdidumm, line 411: mhhhhm I originally had `nodes.some(node => node.server?.load` as a condition in there, but removed it because that is  not called and prerendered - BUT, I think the server manifest will not include this route then anymore. Is my hunch correct? What can we do about that besides "don't prerender if there's any server load function"? Is this also true for shared load functions? Need to test this ..

dummdidumm: confirmed, it doesn't work for nodes that have any server load functions, because they are left out of the server manifest, which results in a 404.

packages/kit/src/runtime/server/page/index.js
dummdidumm: Code is no longer the same

packages/kit/src/core/prerender/prerender.js
dummdidumm: fixed through setting it to `'auto'` in this case

dummdidumm:
Ha, this uncovered an unrelated bug in `write_types`. I removed the `check` for now to get CI green. When I open the PR to fix it I'll reenable it.

Rich-Harris (approved):

Issue #5397, opened 2022-07-06
Storybook support


benmccann:
### Describe the problem

Storybook mostly works now, but there's still some things that can be cleaned up. Opening this as a tracking issue for the various pending PRs

### Describe the proposed solution

- https://github.com/intuit/design-systems-cli/pull/701
- https://github.com/storybookjs/addon-svelte-csf/pull/61
- https://github.com/storybookjs/addon-svelte-csf/pull/62
- https://github.com/storybookjs/addon-svelte-csf/pull/63
- https://github.com/storybookjs/addon-svelte-csf/pull/66
- https://github.com/storybookjs/builder-vite/pull/427
- https://github.com/storybookjs/builder-vite/pull/428
- https://github.com/storybookjs/builder-vite/pull/444
- https://github.com/storybookjs/builder-vite/pull/445
- https://github.com/storybookjs/storybook/pull/18645
- https://github.com/storybookjs/storybook/pull/18651
- https://github.com/svelte-add/svelte-add/pull/220
- https://github.com/storybookjs/storybook/pull/19338
- https://github.com/storybookjs/storybook/issues/19409
- fix `npx sb init`
  - https://github.com/storybookjs/storybook/pull/18648
  - https://github.com/storybookjs/storybook/pull/18705
  - update `viteFinal` to match [michaelwooley/storybook-experimental-vite](https://github.com/michaelwooley/storybook-experimental-vite) when SvelteKit is present
  - reference:
    - [svelte generator](https://github.com/storybookjs/storybook/blob/next/lib/cli/src/generators/SVELTE/index.ts)
    - [base generator](https://github.com/storybookjs/storybook/blob/next/lib/cli/src/generators/baseGenerator.ts)
    - [example files](https://github.com/storybookjs/storybook/tree/next/lib/cli/src/frameworks/svelte)
- check potential alias issues on [michaelwooley/storybook-experimental-vite](https://github.com/michaelwooley/storybook-experimental-vite#readme) and public env vars issue https://github.com/sveltejs/kit/issues/6225#issuecomment-1226055441
 

### Importance

nice to have

GauBen:
I ran into this very issue this morning, and after a bit of trial and error, I managed to have a working configuration: https://github.com/windicss/windicss/issues/267#issuecomment-1175919481

I'll try to contribute to svelte-add.

benmccann:
I think rather than creating a new Svelte adder it may make more sense to fix `npx sb init`, which does the same thing

[...]

PR #8138, opened 2022-12-13
[chore] move writeBundle and closeBundle to compile plugin

benmccann:
these are really more about compiling than setup. It also fixes Storybook `build` (the original version only worked with `dev` as I hadn't tested against `build`)

closes https://github.com/sveltejs/kit/issues/5397


Rich-Harris (approved):
taking your word for it 😆

Issue #8108, opened 2022-12-12
Errors not correctly handled when it occurs in +layout.server


raythurnevoid:
### Describe the bug

The error incoming from the _data.json request is not correctly handled here:
https://github.com/sveltejs/kit/blob/96751b0636f1cd511498de1de4d70990079348ea/packages/kit/src/runtime/client/client.js#L1751

The current behavior results in receiving an Error with message [object object] in hooks.client.

![image](https://user-images.githubusercontent.com/53383860/207137460-f40e55db-2775-4656-a0f4-90fc22a1ceee.png)
![image](https://user-images.githubusercontent.com/53383860/207137629-5babd7fa-b19e-4180-894a-3ae988cb975a.png)

### Reproduction

run `throw error(500, 'test');` in +layout.server

### System Info

```Shell
System:
    OS: Windows 10 10.0.22621
    CPU: (16) x64 AMD Ryzen 9 4900H with Radeon Graphics
    Memory: 13.83 GB / 31.42 GB
  Binaries:
    Node: 18.10.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.19.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.819.0), Chromium (108.0.1462.46)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @sveltejs/adapter-node: 1.0.0-next.102 => 1.0.0-next.102 
    @sveltejs/kit: 1.0.0-next.571 => 1.0.0-next.571 
    svelte: ^3.53.1 => 3.53.1
```

### Severity

annoyance

raythurnevoid:
looks like it's not happening in the latest version of svelte kit (1.0.0-next.588)

Issue #7914, opened 2022-12-02
Parsing errors and dev errors - info in documentation


lukaszpolowczyk:
### Describe the problem

Should there be information about parsing errors and dev stage errors in the Errors - https://kit.svelte.dev/docs/errors documentation?

E.g. HMR window contains internal server data, but it is not a security risk for internal server data.

Or as there are parsing errors in hooks, internal errors are displayed on the page, containing absolute paths from the server.

This data only displays at the dev stage, so it is safe.
But after reading https://kit.svelte.dev/docs/errors, you might think that this data should not display and that it is a security threat to the server's internal data.

I think it is worth adding information to the documentation, about the errors that display on the site only in dev mode and that they are safe.

---

If I don't understand something, please explain.

### Describe the proposed solution

As above.

### Importance

nice to have

dummdidumm:
I'm not totally sure I understand what you mean by "dev time only". In general, if an error is thrown, you should not just print out the stack as that would be security risk. You could implement something yourself in the `handleError` hooks like "in development, also show the stack" - is that what you mean by adding?

lukaszpolowczyk:
> I'm not totally sure I understand what you mean by "dev time only". 

@dummdidumm It is about an unexpected error that is not picked up by `handleError`, nor will `src/error.html` be displayed, such as the code:

```javascript
// src/hooks.server.js

/** @type {import('@sveltejs/kit').HandleServerError} */
export function handleError({ error, event }) {
  throw "zz";
}
/** @type {import('@sveltejs/kit').Handle} */
export async function handle({ event, resolve }) {
  throw "oo";
  const response = await resolve(event);
  return response;
}
```
will return a white content page in the browser:
```
ReferenceError: ddf is not defined
    at //home/projects/sveltejs-kit-template-default-kqwysb/src/hooks.server:2:1
    at instantiateModule (file://file:///home/projects/sveltejs-kit-template-default-kqwysb/node_modules/vite/dist/node/chunks/dep-ed9cb113.js:53295:15)
```
![obraz](https://user-images.githubusercontent.com/16800535/207122551-277dfad6-2464-4542-854c-a8b095da607a.png)


Stackblitz version:
https://stackblitz.com/edit/sveltejs-kit-template-default-kqwysb?file=src%2Fhooks.server.js&terminal=dev

The same when to make simple:
```javascript
// src/hooks.server.js
ddf;
```

it also returns a white page with errors:
[...]

PR #8142, opened 2022-12-14
feat: support throwing error in handle

### LINKED ISSUE #7966: support for throwing `error` in `handle` function
### Describe the problem

When doing a client-side `__data.json` request, errors thrown with the `error` helper function inside `hooks.server.ts` , will not return the correct error object.

It works as expected when `error` gets thrown inside a `+server.*` file.

Here is a demonstration of the issue; https://stackblitz.com/edit/sveltejs-kit-template-default-djcepj?file=src/hooks.server.ts

1. start the application
2. click on the "test hooks.server.ts" link
3. a 401 error gets thrown inside `hooks.servert.ts`
4. the page shows a 500 error
5. reload the page
6. the 401 error gets correctly shown

related: https://github.com/sveltejs/kit/issues/7609
The issue was already described as a sidenote there.

`SvelteKit` should support throwing `error` objects also for `__data.json` requests inside `hooks.server.ts`.
This makes the usage of the `error` helper function more consistent.

### Alternatives considered

Leave it as it is

### Importance

would make my life easier

### Additional Information

_No response_

ivanhofer:
Closes #7966
Heavily inspired by #7612; no new tests for the same reason described there.

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/index.js
+++ b/packages/kit/src/runtime/server/index.js
@@ -3,7 +3,7 @@ import { render_page } from './page/index.js';
 import { render_response } from './page/render.js';
 import { respond_with_error } from './page/respond_with_error.js';
 import { is_form_content_type } from '../../utils/http.js';
-import { GENERIC_ERROR, handle_fatal_error, redirect_response } from './utils.js';
+import { GENERIC_ERROR, handle_fatal_error, redirect_response, static_error_page } from './utils.js';
 import {
 	decode_pathname,
 	decode_params,
@@ -16,7 +16,7 @@ import { exec } from '../../utils/routing.js';
 import { INVALIDATED_HEADER, render_data } from './data/index.js';
 import { add_cookies_to_headers, get_cookies } from './cookie.js';
 import { create_fetch } from './fetch.js';
-import { Redirect } from '../control.js';
+import { HttpError, Redirect } from '../control.js';
 
 /* global __SVELTEKIT_ADAPTER_NAME__ */
 
@@ -363,6 +363,9 @@ export async function respond(request, options, state) {
 	} catch (error) {
 		if (error instanceof Redirect) {
 			return redirect_response(error.status, error.location);
+		}
+		if (error instanceof HttpError) {
+			return static_error_page(options, error.status, error.body.message);
 		}
 		return handle_fatal_error(event, options, error);
 	}
--- a/.changeset/selfish-berries-pump.md
+++ b/.changeset/selfish-berries-pump.md
@@ -1,5 +1,0 @@
----
-'@sveltejs/kit': patch
----
-
-[feat] support throwing error in handle
--- a/.changeset/selfish-berries-pump.md
+++ b/.changeset/selfish-berries-pump.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[feat] support throwing error in handle


ivanhofer:
My git cli says I don't have any changes to pull. Not sure how to bring this branch up2date.

packages/kit/src/runtime/server/index.js
```diff
@@ -364,6 +364,9 @@ export async function respond(request, options, state) {
 		if (error instanceof Redirect) {
 			return redirect_response(error.status, error.location);
 		}
+		if (error instanceof HttpError) {
+			return static_error_page(options, error.status, error.body.message);
```
dummdidumm, line 368: I'm confused why this is necessary. `handle_fata_error` wraps this functionality, so the outcome should be the same

ivanhofer: It seems some behaviour has changed since last week. I have updated the example: https://stackblitz.com/edit/sveltejs-kit-template-default-djcepj?file=src/hooks.server.ts

When I opened the issue last week, I got a 500 error when **hovering** over the link.
With the latest `SvelteKit` version hovering works, but **clicks** result in a 500 error shown.

ivanhofer:
> did you pull from the master branch or your error-handle branch?

I pulled from master, and I also updated my fork to contain the latest changes from the original repo. Pulling directly from the original does not work either. Not sure what is broken.

packages/kit/src/runtime/server/index.js
dummdidumm: sounds related to #8108

dummdidumm:
Tried a quick stab at this, it's trickier than I thought. Ideally we fall back to the static error page, but I'm not sure how to do that from a data response yet.

packages/kit/src/runtime/server/index.js
ivanhofer: I think those are two different issues. I could not reproduce the issue described in #8108

ivanhofer:
> Tried a quick stab at this, it's trickier than I thought. Ideally we fall back to the static error page, but I'm not sure how to do that from a data response yet.
Is the problem that it is a client-side navigation, the data request responds with json and can't be used to deliver the error page because it would be html? And we can't render the `+error.svelte` page because the `hooks.server.js` has no information about the rendering process because it never calls `resolve(event)`?

lukaszpolowczyk:
@ivanhofer @dummdidumm Is this somehow related?
https://github.com/sveltejs/kit/issues/7914#issuecomment-1347002959

gtm-nayan:
I was confused by the changes here, turns out the server is returning the correct response, Simon's right on handle_fatal_error being able to deal with this already, the problem is on the client side.

[Here](https://github.com/sveltejs/kit/blob/c5b43af17cd0d07eccbf6a941900f3170d49136b/packages/kit/src/runtime/client/client.js#L1697) load_data creates an error object, which is caught and passed in [here](https://github.com/sveltejs/kit/blob/c5b43af17cd0d07eccbf6a941900f3170d49136b/packages/kit/src/runtime/client/client.js#L749) but there is no status information in that catch clause so it just shows a generic 500.

Also related and probably can be solved at the same time, because of the same "just thrown a native Error" mechanic, the HttpError from the server is never actually instantiated into an HttpError in the browser.

ivanhofer:
@gtm-nayan thanks for your investigation.
Haven't had the time to follow up on this issue. But it seems my assumption was wrong and this PR does not make sense at all. I will close it.

PR #8162, opened 2022-12-14
chore(create-svelte): address flakiness in playwright test

mxschmitt:
Congrats to the version release! Before no web-first assertion was used for the assertion, which means that it did not retry and just failed when the text was not matching. Now its using [toHaveText](https://playwright.dev/docs/test-assertions#locator-assertions-to-have-text) which is recommended and not flaky.

There were more such places inside the sveltekit codebase, seems not related, since its not customer facing, but might still be worth updating.

### Tests

### Changesets

### DIFF
--- a/.changeset/rare-frogs-crash.md
+++ b/.changeset/rare-frogs-crash.md
@@ -1,0 +1,5 @@
+---
+"create-svelte": patch
+---
+
+[fix] address flakiness in playwright test
--- a/packages/create-svelte/shared/+playwright+default/tests/test.ts
+++ b/packages/create-svelte/shared/+playwright+default/tests/test.ts
@@ -2,5 +2,5 @@ import { expect, test } from '@playwright/test';
 
 test('about page has expected h1', async ({ page }) => {
 	await page.goto('/about');
-	expect(await page.textContent('h1')).toBe('About this app');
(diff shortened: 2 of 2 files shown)


benmccann:
> There were more such places inside the sveltekit codebase, seems not related, since its not customer facing, but might still be worth updating.

Yes, please! PRs welcome to reduce flakiness

benmccann (approved):

.changeset/rare-frogs-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-svelte": patch
```
dominikg, line 2: changes in tests don't require changesets. please remove this file

dominikg:
general note on `await expect`: 

i think in this instance it's ok to use it, but in general this is not what we want. esp within framework tests we want to expect acutal state at a specific point, not eventual state after a specific point.
this is both faster and more correct

.changeset/rare-frogs-crash.md
gtm-nayan: It changes a file that is shipped with create-svelte, so the version should be bumped in this case.

dominikg: oh, my bad. thanks for checking @gtm-nayan

PR #8166, opened 2022-12-14
deploy to the edge

Rich-Harris:


sites/kit.svelte.dev/package.json
```diff
@@ -12,6 +12,7 @@
 	"devDependencies": {
 		"@sveltejs/adapter-auto": "workspace:^",
 		"@sveltejs/adapter-static": "workspace:^",
```
Conduitry, line 14: Do we still need the auto (or static) adapters here?

Rich-Harris: nope

PR #8176, opened 2022-12-15
use prerendered endpoint to get content for search

### LINKED ISSUE #8165: `https://kit.svelte.dev/search` does not work
### Describe the bug

...as I discovered in slightly embarrassing fashion. It's trying to read from the filesystem, which isn't available. At one point it was loading a prerendered `content.json`, which allowed it to work, but at some point I evidently broke that.

The search function still works, but there's no JS-less fallback.

### Reproduction

Visit https://kit.svelte.dev/search.

### Logs

_No response_

### System Info

```Shell
N/A
```

### Severity

annoyance

### Additional Information

_No response_

gtm-nayan:
closes #8165 

A better solution might be something like a custom vite plugin for the docs, so that all the files are resolved by the bundler instead of a prerender call.

### Tests

### Changesets

### DIFF
--- a/sites/kit.svelte.dev/src/routes/search/+page.server.js
+++ b/sites/kit.svelte.dev/src/routes/search/+page.server.js
@@ -10,7 +10,6 @@ export async function load({ url, fetch }) {
 		init(blocks);
 	}
 
-
 	const query = url.searchParams.get('q');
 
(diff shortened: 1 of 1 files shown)


dummdidumm (approved):
👍

PR #8186, opened 2022-12-15
chore: add ignore list to changesets config

dominikg:
uses a negated micromatch pattern to only include @sveltejs packages and published non-org packages.

This skips over the plethora of test apps and other non-published things like the templates.

Con: adding new non-org packages does require extending the negated pattern
Pro: adding new test apps or other non-published packages does not require changes

As the second case is arguably happening more often and first publish of a new package is something we have to plan for either way, i prefer this approach over a positive pattern

### Tests

### Changesets

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -5,5 +5,6 @@
 	"linked": [],
 	"access": "public",
 	"baseBranch": "master",
-	"bumpVersionsWithWorkspaceProtocolOnly": true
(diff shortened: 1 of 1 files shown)


Conduitry:
I think doing it in this direction is a good idea. We might someday add another non-org package, but I can't foresee any specific ones now.

Conduitry (approved):

PR #8192, opened 2022-12-16
[docs] remove confusion about next vs previous layout

yousufiqbal:
Next should be previous. (I think)

### Tests

### Changesets

### DIFF
--- a/documentation/docs/20-core-concepts/10-routing.md
+++ b/documentation/docs/20-core-concepts/10-routing.md
@@ -201,7 +201,7 @@ We can create a layout that only applies to pages below `/settings` (while inher
 <slot></slot>
 ```
 
-By default, each layout inherits the next layout above it. Sometimes that isn't what you want - in this case, [advanced layouts](/docs/advanced-routing#advanced-layouts) can help you.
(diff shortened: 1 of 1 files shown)


documentation/docs/20-core-concepts/10-routing.md
```diff
@@ -201,7 +201,7 @@ We can create a layout that only applies to pages below `/settings` (while inher
 <slot></slot>
 ```
 
-By default, each layout inherits the next layout above it. Sometimes that isn't what you want - in this case, [advanced layouts](/docs/advanced-routing#advanced-layouts) can help you.
+By default, each layout inherits the previous layout above it. Sometimes that isn't what you want - in this case, [advanced layouts](/docs/advanced-routing#advanced-layouts) can help you.
```
dummdidumm, line 204: I think both are right, but I understand this could be confusing. Maybe we should just remove that word altogether?
```suggestion
By default, each layout inherits the layout above it. Sometimes that isn't what you want - in this case, [advanced layouts](/docs/advanced-routing#advanced-layouts) can help you.
```

yousufiqbal: Yes. That's much better.

PR #8193, opened 2022-12-16
[docs] clarify that server load should be used in example and isn't optional

yousufiqbal:
Should is better word here.

### Tests

### Changesets

### DIFF
--- a/documentation/docs/20-core-concepts/20-load.md
+++ b/documentation/docs/20-core-concepts/20-load.md
@@ -34,7 +34,7 @@ export function load({ params }) {
 
 Thanks to the generated `$types` module, we get full type safety.
 
-A `load` function in a `+page.js` file runs both on the server and in the browser. If your `load` function should _always_ run on the server (because it uses private environment variables, for example, or accesses a database) then you should put it in a `+page.server.js` instead.
+A `load` function in a `+page.js` file runs both on the server and in the browser. If your `load` function should _always_ run on the server (because it uses private environment variables, for example, or accesses a database) then it would go in a `+page.server.js` instead.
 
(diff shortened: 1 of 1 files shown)


documentation/docs/20-core-concepts/20-load.md
```diff
@@ -34,7 +34,7 @@ export function load({ params }) {
 
 Thanks to the generated `$types` module, we get full type safety.
 
-A `load` function in a `+page.js` file runs both on the server and in the browser. If your `load` function should _always_ run on the server (because it uses private environment variables, for example, or accesses a database) then you can put it in a `+page.server.js` instead.
+A `load` function in a `+page.js` file runs both on the server and in the browser. If your `load` function should _always_ run on the server (because it uses private environment variables, for example, or accesses a database) then you should put it in a `+page.server.js` instead.
```
dummdidumm, line 37: A shorter version
```suggestion
A `load` function in a `+page.js` file runs both on the server and in the browser. If your `load` function should _always_ run on the server (because it uses private environment variables, for example, or accesses a database) then put it in a `+page.server.js` instead.
```

benmccann, line 37: the sentence sounds funny with "should" used twice
```suggestion
A `load` function in a `+page.js` file runs both on the server and in the browser. If your `load` function should _always_ run on the server (because it uses private environment variables, for example, or accesses a database) then it would go in a `+page.server.js` instead.
```

benmccann:
this is a very marginal PR. I think it was probably fine before. This is arguably an improvement, but it's not clear. I'll go ahead and merge this, but we really prefer PRs that are more substantial or fixing something that's more clearly wrong

PR #8225, opened 2022-12-19
[chore] fix some test flakiness

benmccann:
`waitForTimeout` is inherently flaky and we should avoid using it as much as possible
`page.textContent` is discouraged: https://playwright.dev/docs/api/class-page#page-text-content
as is `page.click`: https://playwright.dev/docs/api/class-page#page-click

The `window.now` cache invalidation basically didn't work at all. We were just accidentally triggering cache invalidation a lot of the time because the tests ran slowly. If you had a bigger cache invalidation window the test would fail somewhat consistently

### DIFF
--- a/packages/kit/src/exports/node/index.js
+++ b/packages/kit/src/exports/node/index.js
@@ -104,7 +104,11 @@ export async function getRequest({ request, base, bodySizeLimit }) {
 }
 
 /** @type {import('@sveltejs/kit/node').setResponse} */
-export async function setResponse(res, response) {
+export async function setResponse(res, response, opt_req) {
+	if (process.env.DEBUG && opt_req.url === '/load/cache-control/increment') {
+		console.log('setting response');
+	}
+
 	const headers = Object.fromEntries(response.headers);
 
 	if (response.headers.has('set-cookie')) {
@@ -116,13 +120,24 @@ export async function setResponse(res, response) {
 	}
 
 	res.writeHead(response.status, headers);
+	if (process.env.DEBUG && opt_req.url === '/load/cache-control/increment') {
+		console.log('wrote headers');
+	}
 
 	if (!response.body) {
+		if (process.env.DEBUG && opt_req.url === '/load/cache-control/increment') {
+			console.log('NO BODY');
+		}
+
 		res.end();
 		return;
 	}
 
 	if (response.body.locked) {
+		if (process.env.DEBUG && opt_req.url === '/load/cache-control/increment') {
+			console.log('BODY LOCKED');
+		}
+
 		res.write(
 			'Fatal error: Response body is locked. ' +
 				`This can happen when the response was already read (for example through 'response.json()' or 'response.text()').`
@@ -134,6 +149,9 @@ export async function setResponse(res, response) {
 	const reader = response.body.getReader();
 
 	if (res.destroyed) {
+		if (process.env.DEBUG && opt_req.url === '/load/cache-control/increment') {
+			console.log('RESPONSE DESTROYED');
+		}
 		reader.cancel();
 		return;
 	}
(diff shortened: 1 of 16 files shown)


Rich-Harris:
Agree that `waitForTimeout` is flaky and should be avoided, but I'm less convinced about `page.textContent`. The Playwright docs recommend using `page.locator` because it automatically waits for the locator to exist, which is helpful when you're testing an app but not when you're testing a _framework_ — we need to be able to make assertions about timing that an app doesn't. By using `page.locator`, we're being loosey-goosey when we need to be on top of our shit. In most tests it won't make a difference, but adopting it as a default makes me nervous.

(I'm not sure about `page.click` — I thought it behaved the same as `page.textContent` but it seems it _also_ waits for the selector to match? In which case I'm not sure why they recommend `page.locator` over it.)

packages/kit/test/apps/basics/test/test.js
```diff
@@ -1622,10 +1621,7 @@ test.describe('Routing', () => {
 
 		try {
 			await page.goto(`/routing?port=${port}`);
-			await Promise.all([
-				page.click(`[href="http://localhost:${port}"]`),
-				page.waitForURL(`http://localhost:${port}/`)
-			]);
+			await page.locator(`[href="http://localhost:${port}"]`).click();
```
dummdidumm, line 1624: What is this even testing (unrelated to the changes)? I don't see an assertion here.

dummdidumm (approved):
Your explanation makes sense

packages/kit/test/apps/basics/test/test.js
benmccann: Good question. Looks like that test was added in https://github.com/sveltejs/kit/pull/1487. I haven't gotten a chance to take a look at it yet and figure out if it might actually be testing something

benmccann: I've changed this back to the prior code and added a comment describing the test

packages/kit/src/exports/node/index.js
```diff
@@ -104,7 +104,14 @@ export async function getRequest({ request, base, bodySizeLimit }) {
...
 
 /** @type {import('@sveltejs/kit/node').setResponse} */
-export async function setResponse(res, response) {
+export async function setResponse(res, response, opt_req) {
+	const DEBUG =
+		opt_req?.url === '/load/cache-control/count' ||
+		opt_req?.url === '/load/cache-control/increment';
+	if (DEBUG) {
+		console.log('-! setting response');
+	}
```
Rich-Harris, line 113: I assume the plan is to remove this stuff before marking the PR ready?

Rich-Harris (changes requested):
countering @dummdidumm's approval pending the removal of all the DEBUG stuff

benmccann:
There's something funny going on that I haven't been able to figure out yet. I put this in draft and added tons of debug to reproduce it on the CI. I wouldn't bother taking a look at it yet. If I can figure out the issue I might just send the fix in a new PR as I'm making a bit of a mess of this one

packages/kit/test/apps/basics/test/client.test.js
```diff
@@ -1094,10 +1086,10 @@ test.describe.serial('Invalidation', () => {
 
 	test('$page.url can safely be mutated', async ({ page }) => {
 		await page.goto('/load/mutated-url?q=initial');
-		expect(await page.textContent('h1')).toBe('initial');
+		expect(page.getByText('initial')).toBeVisible();
```
Rich-Harris, line 1089: surely this needs an `await` otherwise it's not testing anything at all?

```suggestion
		await expect(page.getByText('initial')).toBeVisible();
```

packages/kit/test/apps/basics/test/client.test.js
```diff
@@ -1094,10 +1086,10 @@ test.describe.serial('Invalidation', () => {
...
 
-		await page.click('button');
-		expect(await page.textContent('h1')).toBe('updated');
+		await page.locator('button').click();
+		expect(page.getByText('updated')).toBeVisible();
```
Rich-Harris, line 1092: ditto

```suggestion
		await expect(page.getByText('updated')).toBeVisible();
```

PR #8226, opened 2022-12-19
[chore] remove unused dependency

benmccann:
I don't see anywhere in the code where this is referenced

I don't notice any issues on the homepage from this being removed: https://kit-svelte-7ioul50vz-svelte.vercel.app/


bluwy (approved):

benmccann:
Looks like it was used in one of the homepage prototypes and that was changed, but the `package.json` wasn't updated to remove it, so this should be safe. https://github.com/sveltejs/kit/commit/b41fc7b962a6f75e59471fd3589e02c0b209128d#diff-59bf054b276c64dca6c10c9c7d4364861318e643480a6ef181e451400f615fd2

Issue #8230, opened 2022-12-20
Platform is undefined in dev mode


chrskerr:
### Describe the bug

In the docs ([here](https://kit.svelte.dev/docs/types#app)) the default type for Platform is `{}`, but in dev it is coming through as undefined.

This makes type safety harder, since TS doesn't enforce an is-object check before attempting to access properties. 

My preference would like to default Platform to an empty object, rather than updating the type to allow undefined. 

This is an issue for us since adapter-node will only apply during production builds and as such during dev the Platform object is undefined.

### Reproduction

Open repo https://github.com/chrskerr/sveltekit-platform-demo, and start in dev mode

Open the app in any browser, and you'll see 'undefined' printed into the terminal as the result of `console.log(typeof event.platform)` from hooks.server.ts

VSCode believes the type of event.platform to be Readonly<Platform> or Readonly<{}>, meaning the below fails TS because 'something' doesn't exist on event.platform (correct) but it doesn't catch that event.platform is maybe undefined 

<img width="898" alt="Screenshot 2022-12-20 at 11 50 20 am" src="https://user-images.githubusercontent.com/10432093/208562388-f9d8768f-2c97-4284-b7bf-ac10489df103.png">

### System Info

```Shell
System:
    OS: macOS 13.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 266.11 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.18.1 - ~/.nvm/versions/node/v16.18.1/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v16.18.1/bin/npm
  Browsers:
    Chrome: 108.0.5359.124
    Firefox: 107.0
    Safari: 16.2
  npmPackages:
    @sveltejs/adapter-auto: ^1.0.0 => 1.0.0 
    @sveltejs/adapter-node: ^1.0.0 => 1.0.0 
    @sveltejs/kit: ^1.0.0 => 1.0.1 
    svelte: ^3.54.0 => 3.55.0 
    vite: ^4.0.0 => 4.0.2
```

### Severity

annoyance

paukuman:
alternative, using a `npm run pages:preview`

PR #8232, opened 2022-12-20
Update request type definitions to make platform possibly undefined

chrskerr:
This PR is to resolve issue https://github.com/sveltejs/kit/issues/8230 in which the platform object does not match it's types.

`event.platform` is typed as an empty object by default so I am changing the server response to fallback to this rather than allowing it to be undefined.

An alternative could be to update the types to make platform optional within event.

### Tests

### Changesets

### DIFF
--- a/.changeset/shy-monkeys-raise.md
+++ b/.changeset/shy-monkeys-raise.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-Update typings of event.platform to be possibly undefined
+Update typings of `event.platform` to be possibly undefined
--- a/.changeset/shy-monkeys-raise.md
+++ b/.changeset/shy-monkeys-raise.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-event.platform falls back to empty object when not defined
+Update typings of event.platform to be possibly undefined
--- a/.changeset/shy-monkeys-raise.md
+++ b/.changeset/shy-monkeys-raise.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': minor
+---
+
+event.platform falls back to empty object when not defined
--- a/.changeset/shy-monkeys-raise.md
+++ b/.changeset/shy-monkeys-raise.md
@@ -1,5 +1,5 @@
 ---
-'@sveltejs/kit': minor
+'@sveltejs/kit': patch
 ---
 
 event.platform falls back to empty object when not defined
--- a/packages/kit/types/index.d.ts
+++ b/packages/kit/types/index.d.ts
@@ -865,7 +865,7 @@ export interface RequestEvent<
 	/**
 	 * Additional data made available through the adapter.
 	 */
-	platform: Readonly<App.Platform>;
+	platform: Readonly<App.Platform> | undefined;
 	/**
 	 * The original request object
 	 */
--- a/packages/kit/types/index.d.ts
+++ b/packages/kit/types/index.d.ts
@@ -865,7 +865,7 @@ export interface RequestEvent<
 	/**
 	 * Additional data made available through the adapter.
 	 */
-	platform: Readonly<App.Platform> | undefined;
+	platform: Readonly<App.Platform> | void;
 	/**
 	 * The original request object
 	 */
--- a/packages/kit/types/index.d.ts
+++ b/packages/kit/types/index.d.ts
@@ -865,7 +865,7 @@ export interface RequestEvent<
 	/**
 	 * Additional data made available through the adapter.
 	 */
-	platform: Readonly<App.Platform> | void;
+	platform: Readonly<App.Platform> | undefined;
 	/**
 	 * The original request object
 	 */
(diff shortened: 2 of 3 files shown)


.changeset/shy-monkeys-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': minor
```
dummdidumm, line 2:
```suggestion
'@sveltejs/kit': patch
```

dummdidumm (approved):

Rich-Harris:
Thanks for the PR! I think the proposed alternative (making `event.platform` optional) is a more 'honest' approach — "if `platform` exists then it takes this shape".

Obviously the _real_ solution is to populate `platform` in `dev` as well, but that's a can of worms to open later...

chrskerr:
No problems :) 

I had hoped that this PR would ensure that `platform` was always an empty object, but I think given the runtime differences between `dev` and `prod` builds (due to the invocation of the adapter) updating the types will help avoid mistakes.

Something like the below?

The goal being to allow platform itself to be undefined (this pr) and also to ensure that adapter specific keys are definitely optional.

```ts
// packages/kit/types/index.d.ts

	platform: Readonly<App.Platform>;
```

to:
```ts
// packages/kit/types/index.d.ts

	platform: Partial<Readonly<App.Platform>> | undefined;
```

dummdidumm:
I think `Partial<..>` isn't needed here - either the whole platform is defined (in prod), or it isn't (in dev), there's no "something exists, something else doesn't". Each adapter can define what exactly `App.Platform` contains.

chrskerr:
Yeah that makes sense to me, I imagine that I should also revert my change to `server/index.js` as well? 

Which would mean that if the platform object is undefined (not supplied by adapter or dev mode), or it is exactly as described by the adapter.

If happy with this approach, then I can update the PR today :)

packages/kit/types/index.d.ts
```diff
@@ -865,7 +865,7 @@ export interface RequestEvent<
 	/**
 	 * Additional data made available through the adapter.
 	 */
-	platform: Readonly<App.Platform>;
+	platform: Readonly<App.Platform> | undefined;
```
Rich-Harris, line 868: I think you're 'supposed' to use `void` rather than `undefined` (i.e. the type rather than the value):

```suggestion
	platform: Readonly<App.Platform> | void;
```

Rich-Harris (approved):

Rich-Harris:
thank you!

.changeset/shy-monkeys-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Update typings of event.platform to be possibly undefined
```
Conduitry, line 5:
```suggestion
Update typings of `event.platform` to be possibly undefined
```

packages/kit/types/index.d.ts
```diff
@@ -865,7 +865,7 @@ export interface RequestEvent<
 	/**
 	 * Additional data made available through the adapter.
 	 */
-	platform: Readonly<App.Platform>;
+	platform: Readonly<App.Platform> | void;
```
dummdidumm, line 868: I think this will be super annoying for type safety, AFAIK you can't use the `?` operator with it. `void` means something like "do not use this return value, could be anything"

Rich-Harris: Not sure I follow? [This](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgAoBs5hgeygW2QG8AoZc5fCAZ2rgHMIAuZasKUegbhIF8SSoSLEQoAShACOAVxpgAogDcI4YmQoAHTNjz4WEuABMcIdAE8APBiy4CAPmQAfZIpzBDPfiRjSQCMMAmyIwKyuAAFACU+lKybEoqYGoUyFAQYNJQIMkpmtq2esi+hhAwoBCG6uS8ngIIJmzIKAC8wekJEZE8JPUg1DjoEAB06Dj04cNaNroA-ENUtAwQXUA) seems to work fine:

```ts
interface Platform {
    message: string;
}

interface RequestEvent {
    platform: Readonly<Platform> | void;
}

function getEvent(): RequestEvent {
    return {
        platform: undefined
    };
}

const e = getEvent();

console.log(e.platform?.message);
```

dummdidumm: Mhm might have mixed up something, nevermind then (though I still feel `undefined` is the better type)

Rich-Harris: Is there a reason it's better? I feel confident that I've run into cases before where `undefined` caused problems that were solved by using `void` instead, though I can't recall the specifics

chrskerr: My understanding is that `void` actually means "won't be used used", so in the context of a function return it implies that nothing will be explicitly returned but doesn't enforce that the returned value will be `undefined`, which means that the following code is okay with TS.

```ts
const v: () => void = () => 1
``` 

In this context here, a property assignment, I believe that `void` and `undefined` have the same outcome, but that `undefined` is semantically more typical.

Rich-Harris: fair enough, I'll change it back

```suggestion
	platform: Readonly<App.Platform> | undefined;
```

Issue #6868, opened 2022-09-17
extending a tsconfig in .svelte-kit/tsconfig.json


Alexandre-Fernandez:
### Describe the bug

I have a monorepo which shares a base tsconfig and I want sveltekit to extend it.
However the root sveltekit tsconfig extends `.svelte-kit/tsconfig.json` and I cannot modify `.svelte-kit/tsconfig.json` to extend my  base tsconfig because it gets overridden by the auto-generation.
So from what I understand my only choice if I want to extend my base tsconfig is to lose the autogenerated one ?

### Reproduction

Create a turborepo app and create a sveltekit app inside the app folder.

### System Info

```shell
Linux
```

### Severity

annoyance

### Additional Information

I created this package to be able to extend multiple tsconfigs: 
https://github.com/Alexandre-Fernandez/tsconfig-multiple-extends

bluwy:
Looks like there has been some work for array support for `extends`: https://github.com/microsoft/TypeScript/pull/50403

PR #8237, opened 2022-12-20
feat: enable extending user tsconfig

robBowes:
Closes https://github.com/sveltejs/kit/issues/6868 by allowing the user to specify a root tsconfig to extend. This is useful for monorepos where the is a base tsconfig file at the root that needs to be extended.

### Tests

### Changesets


Rich-Harris:
Thank you. `extends` probably isn't the right name — it should be something tsconfig-specific.

I wonder if a more future-proof approach might be to allow the generated config to be augmented in arbitrary ways? Something like this:

```js
// svelte.config.js
import path from 'path';

export const config = {
  kit: {
    tsconfig: (tsconfig) => {
      tsconfig.extends = path.resolve('../tsconfig.json');
    }
  }
};
```

Or this:

```js
// svelte.config.js
import path from 'path';

export const config = {
  kit: {
    tsconfig: (tsconfig) => {
      return {
        ...tsconfig,
        extends: path.resolve('../tsconfig.json')
      };
    }
  }
};
```

I don't know if `tsconfig` is the best name for the option.

dummdidumm:
I like the second version of Rich's proposal. I think `tsconfig` is the right name.

.changeset/warm-pumas-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Allow sveltekit tsconfig to be extended by adding extends option to svelte config
```
benmccann, line 5:
```suggestion
'@sveltejs/kit': minor
---

feat: allow SvelteKit tsconfig to be extended by adding extends option to svelte config
```

Rich-Harris:
going to close this in favour of #8606 which implements the idea above — thanks!

Issue #8180, opened 2022-12-15
Unable to perform Vitest integrations tests using $env/dynamic/private 


kyllerss:
### Describe the bug

I finally got around to migrating away from `import.meta.env.*` for all of my environment variables and onto the SvelteKit-recommended `$end/dynamic|static/public|private`. I had a number of integration tests I had written in `vitest` that are now breaking with the following error:

```
FAIL  tests/integration/hooks.test.ts [ tests/integration/hooks.test.ts ]
Error: Cannot import $env/dynamic/private into client-side code
 ❯ Context.load node_modules/@sveltejs/kit/src/exports/vite/index.js:304:12
 ❯ Object.load node_modules/vite/dist/node/chunks/dep-67e7f8ab.js:40201:46
 ❯ loadAndTransform node_modules/vite/dist/node/chunks/dep-67e7f8ab.js:36553:24
```
My integration tests import a number of back-end modules (eg. database access code, etc...).

I am surprised by the fact that the integration tests are being treated as client code that is tripping up the server-client serialization safety mechanism.

I am at a loss as to what to do here. Any recommendations on how to address this issue? Is there any way I can turn off this safety mechanism for my integration tests? Is there something entirely different I should be doing to do database integration tests?

PS. My reproduction project is using the stock configuration that comes from a new project. It calls the `vitest` tests *unit* tests. I am not sure if I should be reading too much into that. If `vitest` is not the right place for writing these types of tests, what alternatives should I look into?

### Reproduction

The reproduction test illustrates the problem when importing a class that makes a reference to `$env/dynamic/private`.

https://github.com/kyllerss/sveltekit-issue-8180

### System Info

```Shell
System:
    OS: Linux 5.15 KDE neon 5.26 5.26
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
    Memory: 9.68 GB / 15.34 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 16.18.0 - ~/apps/node-v16.18.0-linux-x64/bin/node
    npm: 8.19.2 - ~/apps/node-v16.18.0-linux-x64/bin/npm
  Browsers:
    Brave Browser: 108.1.46.144
    Chrome: 108.0.5359.124
    Firefox: 108.0
  npmPackages:
    @sveltejs/adapter-auto: ^1.0.0 => 1.0.0 
[...]

PR #8250, opened 2022-12-22
Ignore param matchers that match .{spec,test}.{js,ts}

### LINKED ISSUE #7583: Allow unit test files in params directory
### Describe the problem

We try to write unit tests for every piece of code. Therefore we also want to write unit tests for matchers. We put the unit test files in the same directory as the code file. So because we have a `wellknown.ts` in the params directory, we also want to have a `wellknown.spec.ts` file.

But if we put such a file there, SvelteKit (`svelte-kit sync`) throws an error:
```
Matcher names can only have underscores and alphanumeric characters — "wellknown.spec.ts" is invalid
```

It would be great if SvelteKit allows unit test files inside the `params` directory.

I think this would include these files:
- `*.spec.ts`
- `*.test.ts`
- `*.spec.js`
- `*.test.js`

### Alternatives considered

An alternative would be to move the unit test files to another place. But this would not be consistent anymore with every other unit test file in our projects.

### Importance

would make my life easier

### Additional Information

_No response_

markjaquith:
This allows for param matcher tests to be collocated with their matchers

fixes #7583

Collocation of tests is a common practice, and the current param matcher scanning will throw an error if there are `.{test,spec}.{js,ts}` files in the `src/params` directory, as it sees them as malformed param names.

This is a simple change that just skips over those files instead of throwing an exception.

### Tests

### Changesets

### DIFF
--- a/.changeset/healthy-lizards-drum.md
+++ b/.changeset/healthy-lizards-drum.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Ignore `*.test.js` and `*.spec.js` files in `params` directory
--- a/documentation/docs/30-advanced/10-advanced-routing.md
+++ b/documentation/docs/30-advanced/10-advanced-routing.md
@@ -90,6 +90,8 @@ export function match(param) {
 
 If the pathname doesn't match, SvelteKit will try to match other routes (using the sort order specified below), before eventually returning a 404.
 
+Each module in the `params` directory corresponds to a matcher, with the exception of `*.test.js` and `*.spec.js` files which may be used to unit test your matchers.
+
 > Matchers run both on the server and in the browser.
 
 ## Sorting
--- a/packages/kit/src/core/sync/create_manifest_data/index.js
+++ b/packages/kit/src/core/sync/create_manifest_data/index.js
@@ -76,6 +76,9 @@ function create_matchers(config, cwd) {
 					matchers[type] = matcher_file;
 				}
 			} else {
+				// Allow for matcher test collocation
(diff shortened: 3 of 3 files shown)


benmccann:
There are other places we might need special handling for test extensions. E.g. https://github.com/sveltejs/kit/issues/8180 may need it. Perhaps we should create a list of extensions that can be referenced other places where it may be needed

markjaquith:
> Perhaps we should create a list of extensions that can be referenced other places where it may be needed

Makes sense. Don't know if it makes sense to abstract out the `js` and `ts` parts (I really thought CoffeeScript was gonna be the last "transpiles to JS" language but I was wrong):

```ts
const langExtensions = ['js', 'ts']

const testFlags = ['spec', 'test']

const testExtensions = langExtensions.map(lang => testFlags.map(flag => `.${flag}.${lang}`)).flat()

function isTestFile(file: string) {
	return testExtensions.some(ext => file.endsWith(ext))
}
```

benmccann:
There's a setting for `js` and `ts` part: https://kit.svelte.dev/docs/configuration#moduleextensions

Rich-Harris:
> Perhaps we should create a list of extensions that can be referenced other places where it may be needed

Are you saying there'd be a config option that allows you do to e.g. `+page.test.js`?

benmccann:
No, just that we could refactor `.test` and `.spec` into a variable holding an array that can be referenced elsewhere

Rich-Harris (approved):

Rich-Harris:
I looked into #8180 and the solution is unrelated, so I'm going to file the test extensions array under YAGNI

Rich-Harris:
thank you!

Issue #8257, opened 2022-12-25
Sveltekit generated project using typescript has vite.config.js file instead of ts


thebspin:
### Describe the bug

When generating a typescript project (skeleton) it gets me a vite.config.js file instead of .ts like it used to do in the past.
Playwright for example does has it's .ts file for it's config.

### Reproduction

Generate a new typescript skeleton project, with all the options on yes (playwright, vitest, prettier and eslint).

### System Info

```Shell
System:
    OS: macOS 12.6
    CPU: (10) arm64 Apple M1 Pro
    Memory: 2.98 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.1 - ~/Library/Caches/fnm_multishells/7782_1671956111960/bin/node
    npm: 8.15.0 - ~/Library/Caches/fnm_multishells/7782_1671956111960/bin/npm
  Browsers:
    Chrome: 108.0.5359.124
    Firefox Developer Edition: 106.0
    Safari: 16.0
  npmPackages:
    @sveltejs/adapter-auto: ^1.0.0 => 1.0.0 
    @sveltejs/kit: ^1.0.0 => 1.0.1 
    svelte: ^3.54.0 => 3.55.0 
    vite: ^4.0.0 => 4.0.3
```

### Severity

annoyance

tcc-sejohnson:
Is `vite.config.ts` even supported? I can't find reference to it in their docs.

dominikg:
yes it is, https://vitejs.dev/config/ 

> Vite also directly supports TS config files. You can use vite.config.ts with the defineConfig helper as well.

tcc-sejohnson:
So weird -- I even went to that page and Ctrl+F-ed for `vite.config.ts`! Thanks... 🙃

aradalvand:
Doesn't #8258 implement this? Shouldn't it be closed?

dummdidumm:
Thank you for noticing, closing

PR #8258, opened 2022-12-25
create-svelte TypeScript project should use vite.config.ts instead of vite.config.js

AmFlint:
## Description

Hello !

I opened this PR regarding the issue https://github.com/sveltejs/kit/issues/8257, for TypeScript projects, it makes sense to generate vite.config.ts files. This PR is a very tiny quality of life improvement for TypeScript users but doesn't change anything in a functional way.

By default, the vite command will look for vite.config.js or vite.config.ts file.

I took this issue even though it's probaby not "required" in the project right now, but it allows me to get started with a simple task first on the SvelteKit Repository.

### Tests

### Changesets


AmFlint:
Test failures on macos and webkit seems unrelated to this PR

benmccann:
I think @dominikg had commented previously that this may be something we'd want to avoid. I'm not sure if there's a previous discussion of it on a ticket or PR somewhere

bluwy:
I think it makes sense to use `vite.config.ts` if they chose TypeScript. There isn't a big difference with how Vite handles it if it's using `.js` instead.

dominikg:
for vite config it's a matter of taste, as vite bundles it on the fly either way (not great imho, but it is what it is). svelte config otoh always is .js, so having one config in js and others in ts isn't great either....

at the end of the day users can easily switch to the style they prefer and while I would opt for vite.config.js, defaulting to ts is ok, if that's what the majority prefers.

benmccann (approved):

AmFlint:
I honestly don't mind either, changing the extension from .js to .ts takes 2 seconds, and less maintenance (see my PR I had to move the same file to different places to fit the logic of the svelte-create package, which means if we wish to update it in the future, we'll have to update it in multiple places + test and assert it works in these different templates).

Rich-Harris:
If we do this — and I personally think that bundling config files is a [costly mistake](https://github.com/vitejs/vite/discussions/10570#discussioncomment-4148194) and I'd love to see a future version of Vite skip the bundling step for `vite.config.js` and discourage the use of `vite.config.ts` — then we should generate both `.js` and `.ts` files from the same `.ts` source, like we do for other modules. I've updated the PR

.changeset/famous-emus-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-svelte': patch
+---
+
+For TypeScript project, generate a vite.config.ts file instead of vite.config.js
```
Rich-Harris, line 5:
```suggestion
---
'create-svelte': minor
---

Create vite.config.ts for TypeScript projects
```

PR #8311, opened 2023-01-02
Fix route formatting in error message

jtescher:
This patch uses the route id in the error message rather than the route object `... (at [object Object])`

### Tests

### Changesets

### DIFF
--- a/packages/kit/src/runtime/server/page/load_data.js
+++ b/packages/kit/src/runtime/server/page/load_data.js
@@ -211,7 +211,7 @@ export async function load_data({
 						const included = resolve_opts.filterSerializedResponseHeaders(lower, value);
 						if (!included) {
 							throw new Error(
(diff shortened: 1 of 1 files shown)


benmccann:
Thanks! This certainly looks like an improvement. Though I'm a bit confused as to what the route ID in the error message is supposed to indicate. I first read it as saying that's where the setting needed to be set. I guess it's just saying which route was loaded, but I'm not sure why it's saying that instead of saying which URL was loaded (which we probably wouldn't need to do anyway as it should probably be done when the exception is caught). I'm just not really sure how the route is involved here.

Rich-Harris (approved):

Rich-Harris:
Thanks!

@benmccann including the route ID just makes it slightly easier to find where the offending header is being read

Rich-Harris:
ah crap, forgot a changeset

PR #8315, opened 2023-01-02
[docs] remove unused async

benmccann:
I'm not an expert at promises, but I believe this `async` is unnecessary since we don't do any `await` inside the function


Conduitry:
Not necessary, no, but I'm wondering whether it might be pedagogically helpful. When users write this function, it returns a promise, and so they can make it `async` if they want to. I'm not sure.

benmccann:
Yeah, that's true. At the same time, I'm not sure what you would do where you'd actually want to add an `await` to this function

Rich-Harris (approved):

Rich-Harris:
I'm in favour of this change — I think it's easy enough to discover that you can make the function `async`, either by experimentation or looking at the `HandleFetch` type

Issue #8302, opened 2022-12-31
load functions runs on both server and client


chrisseto:
### Describe the bug

Despite the what is stated by the [documentation](https://kit.svelte.dev/docs/load) svelte kit apps will re-fetch data via the `load` function unless specific methods of the response object are accessed.

### Reproduction

This can be easily reproduced by implementing your own JSON unmarshaling, not that you would want to do that. I stumbled across this bug because I was using a [GRPC Web Client](https://www.npmjs.com/package/@protobuf-ts/grpcweb-transport) which decodes the payload from `body` directly.

```javascript
// Place in any +page.ts/js file

export const load = (async ({ fetch }) => {
    const resp = await fetch('/some-endpoint');

  // You wouldn't ever want to do this. What's important is that we're not accessing .text or .json
  const chunks: Array<any> = [];
  for await (let chunk of resp.body) {
      chunks.push(chunk)
  }

  return JSON.parse(Buffer.concat(chunks).toString('utf-8'));
}) satisfies PageLoad;

```

### Logs

```Shell
N/A
```

### System Info

```Shell
System:
    OS: macOS 13.0.1
    CPU: (8) arm64 Apple M1
    Memory: 349.42 MB / 16.00 GB
    Shell: 3.5.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 18.11.0 - /opt/homebrew/bin/node
    Yarn: 1.22.17 - /opt/homebrew/bin/yarn
    npm: 8.19.2 - /opt/homebrew/bin/npm
  Browsers:
    Brave Browser: 107.1.45.133
    Safari: 16.1
  npmPackages:
    @sveltejs/adapter-auto: next => 1.0.0-next.90
    @sveltejs/adapter-node: ^1.0.0 => 1.0.0
    @sveltejs/adapter-static: ^1.0.0-next.48 => 1.0.0-next.48
    @sveltejs/kit: next => 1.0.0-next.571
    svelte: ^3.53.1 => 3.53.1
    vite: ^3.2.4 => 3.2.4
```

### Severity

serious, but I can work around it

### Additional Information

What's been most frustrating about tracking this down is the phrasing of the documentation. It heavily implies, if not outright states, that the serialized data is what is _returned from `load`_. However, `load` will always be called on both the client and the server. Svelte's `fetch` function is what memoizes the requests but only under certain circumstances. If my understanding is correct, I'm happy to update the documentation myself.

[...]

PR #8317, opened 2023-01-02
improve fetch response recording and serialization

chrisseto:
Previously, calling the sveltekit supplied fetch method would return a proxy around fetch that overrode certain methods. Accessing these methods would push the response into the fetched array which would then be used on the client side to avoid re-fetching data. This method required overriding every method on fetch that could read the body and could leave developers stumped if they happened to read .body from a different method. This commit opts to attach a transformer to the responses original body and pushes into fetched once the body has been completely read. This ensures that all methods of consuming .body will be appropriately recorded.

Notably, this strategy requires encoding the body as a base64'd array of bytes rather than raw text which introduces a minor overhead for serialization and deserialization.

### Tests

### Changesets

### DIFF
--- a/.changeset/thin-cheetahs-rescue.md
+++ b/.changeset/thin-cheetahs-rescue.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+Add support for using all .body methods in universal load functions
--- a/packages/kit/src/runtime/client/fetcher.js
+++ b/packages/kit/src/runtime/client/fetcher.js
@@ -77,7 +77,7 @@ export function initial_fetch(resource, opts) {
 		const ttl = script.getAttribute('data-ttl');
 		if (ttl) cache.set(selector, { body, init, ttl: 1000 * Number(ttl) });
 
-		return Promise.resolve(new Response(body, init));
+		return Promise.resolve(new Response(atob(body), init));
 	}
 
 	return native_fetch(resource, opts);
--- a/packages/kit/src/runtime/server/page/load_data.js
+++ b/packages/kit/src/runtime/server/page/load_data.js
@@ -137,69 +137,14 @@ export async function load_data({
 				}
 			}
 
-			const proxy = new Proxy(response, {
-				get(response, key, _receiver) {
-					async function text() {
-						const body = await response.text();
-
-						if (!body || typeof body === 'string') {
-							const status_number = Number(response.status);
-							if (isNaN(status_number)) {
-								throw new Error(
-									`response.status is not a number. value: "${
-										response.status
-									}" type: ${typeof response.status}`
-								);
-							}
-
-							fetched.push({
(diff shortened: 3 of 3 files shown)


benmccann:
I'll let Rich review this, but a couple notes:
- the tests are failing and would need to be fixed before anything could be merged
- the PR description was extremely hard to read because it was rendered with a horizontal scrollbar due to spaces present before each line of text. Please avoid that method of formatting the text in the future. I've edited the description to fix the formatting on this one

chrisseto:
@benmccann would you mind if I hold off on fixing the test until an initial review? I'd like to get a rough thumbs up on the approach here before combing through the failures. I'm new to the world of svelte and don't write that much JS, so I want to make sure I'm not accidentally breaking something I don't know about.

> the PR description was extremely hard to read because it was rendered with a horizontal scrollbar due to spaces present before each line of text.

Apologies, that was GitHub's default rending of the commit message. I'll track down my description formatting script and update it to strip leading spaces (or just join the text).

Rich-Harris:
Thank you but we definitely shouldn't be base64 encoding data by default. It's not a 'minor' overhead, it increases payload size by a third and makes it uncompressible.

I'm going to close this PR as we can't merge it in its current state — please open an issue (describing what you're trying to do and why the current approach is insufficient) instead so that we can discuss possible solutions.

chrisseto:
There is an issue for this PR https://github.com/sveltejs/kit/issues/8302. I must have amended the commit message but forgot to update the PR description.

I think I tunnel visioned on supporting all cases. It should be fine to encode the payload as utf-8 text ([fetch's .text _always_ decodes as utf-8](https://developer.mozilla.org/en-US/docs/Web/API/Response/text)) and decode the utf-8 text back to bytes. That should match the current implementation's behavior and may even "just work" for binary if reading the body directly but  I'll have to experiment to see what happens. With that solution, there will be no size overhead for all currently supported payloads and there will only be the slightest overhead on the client side to decode and re-encode text. That overhead can likely be worked around with some simple heuristics or reading the content type header.

Happy to discuss more here or on the aforementioned issue.

ximus:
My app primarly makes `grpc-web` calls that I want inlined in html page responses. I’d be happy to help talk about and implement something rather than maintain my fork.

I don't really see a way around some form of inflating encoding. If arbitrary data needs to be embedded in either JSON or HTML, it's going to need to be base-something right?

Not all inlined fetch responses would need to be encoded.

Callers of `response.text()` and `response.json()` are easy to deal with, responses can be inlined as is (as is currently the case). This doesn’t have to change.
Callers of `response.body` or `response.arrayBuffer()` however could be fetching data that is incompatible with with html. Short of being able to peek at that data and determine if encoding is required, we can only consider blindly encoding responses from those two calls.

Now some people might be calling `body` or `arrayBuffer()` on data that doesn’t need to be encoded, and would lose out by getting their stuff encoded. Does anyone have any sense on how likely that is? If we can afford to just encode those two response calles by default?

PR #8321, opened 2023-01-03
[docs] add an FAQ regarding separate backend

benmccann:
I also moved the middleware FAQ down a bit as it discusses a less common need

### DIFF
--- a/documentation/faq/80-integrations.md
+++ b/documentation/faq/80-integrations.md
@@ -73,13 +73,13 @@ onMount(() => {
 });
 ```
 
-### How do I use a backend on another server possibly written in another language?
+### How do I use a different backend API server?
 
-You must first devise a method of determining whether a request should be served by your website or API. Some possibilities include utilizing the subdomain, path, or `Accept` header.
+You may either send requests directly to the API server or you may proxy them to the API server.
 
-Utilizing the path is probably the easiest method. For this method, you may wish to utilize Vite's [`server.proxy`](https://vitejs.dev/config/server-options.html#server-proxy) option in order to proxy any requests under a certain path such as `/api` to your API server.
+If you send requests directly to the API server, be aware that you would need to deal with [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), which will result in complications such as generally requiring requests to be preflighted resulting in higher latency. Requests to a separate subdomain may also increase latency due to an additional DNS lookup, TLS setup, etc. If you wish to use this method, you may find [`handleFetch`](/docs/hooks#server-hooks-handlefetch) helpful.
 
-If you would like to utilize the subdomain method, you may use [`handleFetch`](/docs/hooks#server-hooks-handlefetch) to rewrite the URL of any requests. E.g. if you had a site hosted on `www.mysite.com` and made a call to `fetch('/profile')` that would result in a request to `https://www.mysite.com/profile`, which you could change to `https://api.mysite.com/profile` with the `handleFetch` hook. However, be aware that you would need to deal with [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), which will result in complications such as generally requiring requests to be preflighted resulting in higher latency. Requests to a separate subdomain may also increase latency due to an additional DNS lookup, TLS setup, etc.
+In order to avoid CORS complications, it is probably easiest to proxy requests to the API server. In production, you would configure a load balancer to route a path like `/api` to the API server. For local development, you can utilize Vite's [`server.proxy`](https://vitejs.dev/config/server-options.html#server-proxy) option.
 
 ### How do I use middleware?
 
--- a/documentation/faq/80-integrations.md
+++ b/documentation/faq/80-integrations.md
@@ -75,11 +75,20 @@ onMount(() => {
 
 ### How do I use a different backend API server?
 
-You may either send requests directly to the API server or you may proxy them to the API server.
+You can use [`event.fetch`](/docs/load#making-fetch-requests) to request data from an external API server, but be aware that you would need to deal with [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), which will result in complications such as generally requiring requests to be preflighted resulting in higher latency. Requests to a separate subdomain may also increase latency due to an additional DNS lookup, TLS setup, etc. If you wish to use this method, you may find [`handleFetch`](/docs/hooks#server-hooks-handlefetch) helpful.
 
-If you send requests directly to the API server, be aware that you would need to deal with [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), which will result in complications such as generally requiring requests to be preflighted resulting in higher latency. Requests to a separate subdomain may also increase latency due to an additional DNS lookup, TLS setup, etc. If you wish to use this method, you may find [`handleFetch`](/docs/hooks#server-hooks-handlefetch) helpful.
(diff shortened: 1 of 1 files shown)


documentation/faq/80-integrations.md
```diff
@@ -106,6 +73,47 @@ onMount(() => {
 });
 ```
 
+### How do I use a backend on another server possibly written in another language?
+
+You must first devise a method of determining whether a request should be served by your website or API. Some possibilities include utilizing the subdomain, path, or `Accept` header.
```
Rich-Harris, line 78:
```suggestion
You must first determine whether a request should be served by your website or API. Some possibilities include utilizing the subdomain, path, or `Accept` header.
```

Rich-Harris (changes requested):
I think I approved this too hastily — I'm re-reading it and to be honest I find the whole thing a bit confusing. Why is the answer to 'how do I use a backend on another server possibly written in another language' not just 'use `event.fetch`'? Why are we talking about Vite options? 

I think this needs to be way more straightforward, and that begins with figuring out what questions we're actually trying to answer here

benmccann:
> Why is the answer to 'how do I use a backend on another server possibly written in another language' not just `'use event.fetch'`?

It's not exactly clear to me what you have in mind. But let's say SvelteKit is running on http://localhost:5173/ and the backend on http://localhost:9000/. If you do `fetch('http://localhost:9000/profile')` then you have to deal with CORS. If you do `fetch('/api/profile')` and set Vite's `server.proxy` option then you no longer have to deal with CORS which greatly simplifies things

Conduitry:
Is the idea that, in production, you'll have some other API gateway sitting in front of both your SvelteKit app and your API, and which is responsible for sending requests off to the appropriate one? So, from the browser, you want to be able to just send same-origin fetches, but you also need to make that work from the SvelteKit app during SSR?

Is this what's attempted to be solved here? So you set up a dev mode Vite middleware that proxies the appropriate requests out to another (local, remote, whatever) instance of the external API server?

Rich-Harris:
I mean sure, you have to deal with CORS but that's just a fact about web development — as far as same-origin policy is concerned, http://localhost:5173 requesting data from http://localhost:9000 is no different than https://example.com requesting data from https://api.example.com or https://external-api.com, no? So addressing it here feels like it adds unnecessary confusion, as does talking about `Accept` headers.

I think it would probably make sense to have a separate question (not in `integrations`) along the lines of 'how do I proxy a path like `/api` to an external back-end?' that covers the necessary combination of `handleFetch` and `server.proxy`

Conduitry:
I think the part of the point is that, in production, for requests made from the browser, you _don't_ have to worry about CORS. You have an API gateway that sends requests out to either your API server or your SvelteKit app, and everything is fine. Making things work for local development with that setup sounds better than making your API server implement CORS that it doesn't actually want in production. There should be a proxy to it during local dev that approximates what's going to be happening in production. And there should be some straightforward way for the server-side SvelteKit app to access it (both during local dev and in production) during SSR.

benmccann:
I'm fine with moving the question into a different file if that's a better place to put it.

I agree with everything @Conduitry said.

I'm not quite sure what the best title for the question is. I see lots of people asking "how do I use SvelteKit with a different backend?" and things like that. I think it's helpful for the FAQ to reflect the question in a way that's similar to what users ask. A beginner might not know to look for the term "proxy" or know what "CORS" is. I think it's helpful if we can guide users facing this common challenge in a way that lets them get an answer without prior knowledge of those terms

benmccann:
I rewrote the question and answer. Hopefully it's a bit clearer now

documentation/faq/80-integrations.md
```diff
@@ -106,6 +73,47 @@ onMount(() => {
 });
 ```
 
+### How do I use a different backend API server?
+
+You may either send requests directly to the API server or you may proxy them to the API server.
+
+If you send requests directly to the API server, be aware that you would need to deal with [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), which will result in complications such as generally requiring requests to be preflighted resulting in higher latency. Requests to a separate subdomain may also increase latency due to an additional DNS lookup, TLS setup, etc. If you wish to use this method, you may find [`handleFetch`](/docs/hooks#server-hooks-handlefetch) helpful.
+
+In order to avoid CORS complications, it is probably easiest to proxy requests to the API server. In production, you would configure a load balancer to route a path like `/api` to the API server. For local development, you can utilize Vite's [`server.proxy`](https://vitejs.dev/config/server-options.html#server-proxy) option.
```
Rich-Harris, line 82: I think we need an option for people who can't configure a load balancer (i.e. anyone using a PaaS)

````suggestion
You can use `event.fetch` to request data from an external API server, but be aware that you would need to deal with [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), which will result in complications such as generally requiring requests to be preflighted resulting in higher latency. Requests to a separate subdomain may also increase latency due to an additional DNS lookup, TLS setup, etc. If you wish to use this method, you may find [`handleFetch`](/docs/hooks#server-hooks-handlefetch) helpful.

Another approach is to set up a proxy to bypass CORS headaches. In production, you would configure a load balancer to route a path like `/api` to the API server; for local development, use Vite's [`server.proxy`](https://vitejs.dev/config/server-options.html#server-proxy) option.

If configuring a load balancer isn't an option, you could alternatively add an [API route](https://kit.svelte.dev/docs/routing#server):

```js
// src/routes/api/[...path]/+server.js
export function GET({ params, url }) {
	return fetch(`https://my-api-server.com/${params.path + url.search}`);
}
```

(Note that you may also need to proxy `POST`/`PATCH` etc requests, and forward `request.headers`, depending on your needs.)
````

Comment (benmccann) [reply documentation/faq/80-integrations.md:82]: I think you can pretty much always do it without an API route, but the terminology might be a bit different. E.g. it looks like Vercel and Netlify call them "rewrites":
https://vercel.com/docs/project-configuration#project-configuration/rewrites
https://docs.netlify.com/routing/redirects/rewrites-proxies/

Maybe we should mention those either in addition to or in place of the API route option?

Comment (Rich-Harris) [reply documentation/faq/80-integrations.md:82]: I'm not sure if `vercel.json` will work, since we're using the Build Output API. I'm pretty sure there are some platforms where it's not possible (Cloudflare Workers, for example). Maybe we just don't bring up load balancers?

````suggestion
You can use [`event.fetch`](/docs/load#making-fetch-requests) to request data from an external API server, but be aware that you would need to deal with [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), which will result in complications such as generally requiring requests to be preflighted resulting in higher latency. Requests to a separate subdomain may also increase latency due to an additional DNS lookup, TLS setup, etc. If you wish to use this method, you may find [`handleFetch`](/docs/hooks#server-hooks-handlefetch) helpful.

Another approach is to set up a proxy to bypass CORS headaches. In production, you would rewrite a path like `/api` to the API server; for local development, use Vite's [`server.proxy`](https://vitejs.dev/config/server-options.html#server-proxy) option.

How to setup rewrites will depend on your deployment platform. If rewrites aren't an option, you could alternatively add an [API route](https://kit.svelte.dev/docs/routing#server):

```js
// src/routes/api/[...path]/+server.js
export function GET({ params, url }) {
	return fetch(`https://my-api-server.com/${params.path + url.search}`);
}
```

(Note that you may also need to proxy `POST`/`PATCH` etc requests, and forward `request.headers`, depending on your needs.)
````

Rich-Harris (approved):

PR #8327, opened 2023-01-03
[docs] update FAQ about library inclusion and packaging

benmccann:
There was a lot of outdated information in this FAQ


documentation/faq/70-packages.md
```diff
@@ -2,15 +2,15 @@
...
+Libraries work best with Vite when they distribute an ESM version, especially if those libraries are dependencies of a Svelte component, and you may wish to suggest to library authors that they provide an ESM version. However, CommonJS (CJS) dependencies should work as well since, by default, [`vite-plugin-svelte` will ask Vite to pre-bundle them](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#what-is-going-on-with-vite-and-pre-bundling-dependencies) using `esbuild` to convert them to ESM.
```
dominikg, line 14:
```suggestion
Libraries work best with Vite when they distribute an ESM version, especially if they are dependencies of a Svelte component library. You may wish to suggest to library authors that they provide an ESM version. However, CommonJS (CJS) dependencies should work as well since, by default, [`vite-plugin-svelte` will ask Vite to pre-bundle them](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#what-is-going-on-with-vite-and-pre-bundling-dependencies) using `esbuild` to convert them to ESM.
```

dominikg (approved):

documentation/faq/70-packages.md
```diff
@@ -2,15 +2,15 @@
 title: How do I fix the error I'm getting trying to include a package?
 ---
 
-Vite's SSR support has become fairly stable since Vite 2.7. Most issues related to including a library are due to incorrect packaging.
+Most issues related to including a library are due to incorrect packaging. You can check if a library has packaging issues by entering it into [the publint website](https://publint.bjornlu.com/).
```
bluwy, line 5:
```suggestion
Most issues related to including a library are due to incorrect packaging. You can check if a library has packaging issues by entering it into [the publint website](https://publint.dev/).
```

I bought a new domain recently 😬

documentation/faq/70-packages.md
```diff
@@ -2,15 +2,15 @@
...
 
-Libraries work best with Vite when they distribute an ESM version and you may wish to suggest this to library authors. Here are a few things to keep in mind when checking if a library is packaged correctly:
+Here are a few things to keep in mind when checking if a library is packaged correctly:
 
 - `exports` takes precedence over the other entry point fields such as `main` and `module`. Adding an `exports` field may not be backwards-compatible as it prevents deep imports.
 - ESM files should end with `.mjs` unless `"type": "module"` is set in which any case CommonJS files should end with `.cjs`.
 - `main` should be defined if `exports` is not. It should be either a CommonJS or ESM file and adhere to the previous bullet. If a `module` field is defined, it should refer to an ESM file.
-- Svelte components should be distributed entirely as ESM and have a `svelte` field defining the entry point.
+- Svelte components should be distributed as uncompiled `.svelte` files with any JS in the package written as ESM only.
```
bluwy, line 12: Maybe we should also note that they need to be preprocessed.
```suggestion
- Svelte components should be distributed as uncompiled `.svelte` files with any JS in the package written as ESM only. Custom script and style languages, like TypeScript and SCSS, should be preprocessed as vanilla JS and CSS respectively.
```

benmccann: Yes, that seems nice to mention. Do they need to be preprocessed or is it optional?

bluwy: It's good practice to always preprocess it. But I think it has the same exception as internal-facing packages, where raw `.ts` etc are published, so `.svelte` files could go unpreprocessed too.

Though I think we can simply say they always need to be preprocessed to keep the suggestion simple.

dominikg: I believe we should postulate that they must be preprocessed. Esp. if we want to use the svelte exports condition, that should have a defined baseline, not "maybe you need mdsvex too"

dominikg: If someone wants to paint outside these lines in their internal monorepo, cool, be our guest. But anything for general public consumption? .svelte inside an npm package should be vanilla.

documentation/faq/70-packages.md
```diff
@@ -2,15 +2,15 @@
...
 
-Libraries work best with Vite when they distribute an ESM version and you may wish to suggest this to library authors. Here are a few things to keep in mind when checking if a library is packaged correctly:
+Here are a few things to keep in mind when checking if a library is packaged correctly:
 
 - `exports` takes precedence over the other entry point fields such as `main` and `module`. Adding an `exports` field may not be backwards-compatible as it prevents deep imports.
 - ESM files should end with `.mjs` unless `"type": "module"` is set in which any case CommonJS files should end with `.cjs`.
 - `main` should be defined if `exports` is not. It should be either a CommonJS or ESM file and adhere to the previous bullet. If a `module` field is defined, it should refer to an ESM file.
-- Svelte components should be distributed entirely as ESM and have a `svelte` field defining the entry point.
+- Svelte components should be distributed as uncompiled `.svelte` files with any JS in the package written as ESM only. Custom script and style languages, like TypeScript and SCSS, should be preprocessed as vanilla JS and CSS respectively.
```
benmccann, line 12:
```suggestion
- Svelte components should be distributed as uncompiled `.svelte` files with any JS in the package written as ESM only. Custom script and style languages, like TypeScript and SCSS, should be preprocessed as vanilla JS and CSS respectively. We recommend using [`svelte-package`](/docs/packaging) for packaging Svelte libraries, which will do this for you.
```

Issue #8324, opened 2023-01-03
Sveltekit support for Deno in dev


Jakob5358:
### Describe the problem

A couple of releases ago, deno realesed npm support, so you can use svelte + vite in deno. Sveltekit however gives an error as following when you visit the page in the browser and the page won't load. 
![image](https://user-images.githubusercontent.com/75086936/210345095-78d5d1be-2f00-4185-ac57-7dec5e8347d3.png)

As you can see below, a deno maintainer found that the issue was that sveltekit was polyfilling apis here from undici. They don't work quite the same as those apis in deno causing the error. Polyfills are injected here: https://github.com/sveltejs/kit/blob/e7bc0be2b25aff5ac151e3d83b771ad80cac1ab8/packages/kit/src/exports/node/polyfills.js#L6-L28 (By commenting out these lines i was able to get sveltekit running properly)

More details are in his issue in the deno repo: https://github.com/denoland/deno/issues/17248

### Describe the proposed solution

My suggestion is to add some kind of option to avoid these polyfills all together. I did see https://github.com/sveltejs/kit/pull/7668 which led to https://github.com/sveltejs/kit/issues/7673 which led to it being reverted in https://github.com/sveltejs/kit/pull/7675 . As I understand it this would not inject them when an api was already in globalThis, but what I'm instead suggesting is a manual option so that we can choose not to override the globals, probably in svelte.config.js.

It could for example be
const config = {
  injectGlobals: false
}

### Alternatives considered

As I understand it currently the node adapter is used for dev, perhaps there is some way to switch this out for a official deno adapter which unlike the one on npm also runs in dev but currently I don't see the need for this as this, making a whole new adapter just for changing one file.

### Importance

i cannot use SvelteKit without it

### Additional Information

The importance more means I cannot use sveltekit in deno without it.

benmccann:
Rather than introducing new config I wonder if we could do this automatically. E.g. we could check for `globalThis.Deno` or check for Node with `String(globalThis.fetch).includes("lazyUndici")` (the latter courtesy of @dominikg)

dominikg:
The above is a more explicit check if the local fetch is undici and it's a bit brittle. a more generic node runtime check would be `typeof process !== "undefined" && !!process?.versions?.node` , maybe even parsing the node version to determine if the provided undici is sufficient.

if we want to detect deno at buildtime, it could be done with
```
import {DENO} from 'node-conditions';
if(DENO) {
  ...
}
```
[...]

PR #8338, opened 2023-01-05
[fix] don't polyfill undici if using Deno or Bun

benmccann:
Fixes https://github.com/sveltejs/kit/issues/8324

### DIFF
--- a/.changeset/wise-adults-pull.md
+++ b/.changeset/wise-adults-pull.md
@@ -2,4 +2,4 @@
 '@sveltejs/kit': patch
 ---
 
-[fix] conditionally polyfill but better to support Deno
+[fix] don't polyfill undici if using Deno or Bun
--- a/.changeset/wise-adults-pull.md
+++ b/.changeset/wise-adults-pull.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] conditionally polyfill but better to support Deno
--- a/packages/kit/src/exports/node/polyfills.js
+++ b/packages/kit/src/exports/node/polyfills.js
@@ -18,11 +18,6 @@ const globals = {
 // exported for dev/preview and node environments
 // TODO: remove this once we only support Node 18.11+ (the version multipart/form-data was added)
 export function installPolyfills() {
-	// don't polyfill on non-Node platforms like Deno
-	if (!globalThis.process?.versions?.node) {
-		return;
-	}
-
 	for (const name in globals) {
 		Object.defineProperty(globalThis, name, {
 			enumerable: true,
--- a/packages/kit/src/exports/node/polyfills.js
+++ b/packages/kit/src/exports/node/polyfills.js
@@ -17,6 +17,20 @@ const globals = {
 
 // exported for dev/preview and node environments
 export function installPolyfills() {
+	const version_string = globalThis.process?.versions?.node;
+
+	// don't polyfill on non-Node platforms like Deno
+	if (!version_string) {
+		return;
+	}
+
+	const version = version_string.split('.');
+	// multipart/form-data was added in Undici 5.11: https://github.com/nodejs/undici/releases/tag/v5.11.0
+	// Node 18.11 upgraded to Unidi 5.11: https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md#2022-10-13-version-18110-current-danielleadams
+	// We don't need to polyfill if it's already available
+	if (parseInt(version[0]) > 18 || parseInt(version[1]) >= 11) {
+		return;
+	}
 	for (const name in globals) {
 		Object.defineProperty(globalThis, name, {
 			enumerable: true,
--- a/packages/kit/src/exports/node/polyfills.js
+++ b/packages/kit/src/exports/node/polyfills.js
@@ -17,10 +17,8 @@ const globals = {
 
 // exported for dev/preview and node environments
 export function installPolyfills() {
-	const version_string = globalThis.process?.versions?.node;
-
 	// don't polyfill on non-Node platforms like Deno
-	if (!version_string) {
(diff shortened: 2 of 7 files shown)


Conduitry:
I find the version sniffing a little unpleasant. And, especially since we were already pinning to a specific Undici version before because of problems we were running into, it might potentially cause breakages in people's apps if we deferred to whatever version were built into their version of Node. What if we just checked for `process.version.node` and installed our polyfill if that is present (i.e., for all versions of Node) and skipped it if not (i.e., for Deno, Bun, etc.)?

benmccann:
That could work too. I don't have strong feelings about one vs the other, so went ahead and changed it as you suggested

packages/kit/src/exports/node/polyfills.js
```diff
@@ -16,7 +16,13 @@ const globals = {
 };
 
 // exported for dev/preview and node environments
+// TODO: remove this once we only support Node 18.11+ (the version multipart/form-data was added)
 export function installPolyfills() {
+	// don't polyfill on non-Node platforms like Deno
+	if (!globalThis.process?.versions?.node) {
+		return;
+	}
```
bartlomieju, line 24: Hey folks, Bartek from the Deno team here. This fix won't work - when executing code from npm packages we are injecting globals like `process` or `setImmediate` behind the scenes, if we didn't do that most of packages wouldn't work as they often rely on `process` global. I suggest to do `typeof globalThis.Deno !== "undefined"` check instead

bartlomieju: The error reported in https://github.com/sveltejs/kit/pull/8338#issuecomment-1372057526 is caused by this check

Jarred-Sumner:
> I find the version sniffing a little unpleasant. And, especially since we were already pinning to a specific Undici version before because of problems we were running into, it might potentially cause breakages in people's apps if we deferred to whatever version were built into their version of Node. What if we just checked for `process.version.node` and installed our polyfill if that is present (i.e., for all versions of Node) and skipped it if not (i.e., for Deno, Bun, etc.)?

For Bun, either `typeof globalThis.Bun !== “undefined”` or `typeof process.versions.bun !== “undefined”`. 

`process.versions.node` is set in Bun because many npm packages will fail to load or do something strange if it’s not set

benmccann:
Thanks @Jarred-Sumner. I'm curious, what do you set the Node version to? Would it be something similar to Node like 18.11.0 or more like a fake value like 100.0.0 or 1.0.0 or something?

packages/kit/src/exports/node/polyfills.js
benmccann: @bartlomieju thanks! I had tested `process.versions.node` directly and not through an npm package. I'm curious, what do you set `process.versions.node` to? I wonder if my earlier solution of polyfilling only when the version is less than 18.11.0 might work for Deno

dummdidumm:
Sounds like we should have a blacklist (check who shouldn't get the polyfill, e.g. bun and deno) instead of a whitelist (check for node) either way.

Rich-Harris:
Seems to me the environment detection should happen at the relevant `installPolyfills()` callsites (in `dev`/`preview`/`prerender` etc) rather than inside the function, no? We don't want this cruft in production

benmccann:
> Seems to me the environment detection should happen at the relevant installPolyfills() callsites (in dev/preview/prerender etc) rather than inside the function, no? We don't want this cruft in production

I had debated that. I did it this way to reduce the amount of code required and also because it'd probably allow `adapter-node` to be used by Deno and Bun instead of needing a new adapter for each environment. There are already Deno and Bun adapters, but some users may not discover them and if `adapter-node` just worked out of the box that'd be cool (e.g. I think the Deno folks weren't aware of `adapter-deno`). The runtime cost in production would be pretty minimal and it's always nice to keep dev and production working similarly

Rich-Harris:
People using `adapter-node` for Deno/Bun deployments would be a terrible outcome!

As this PR stands we'd be adding the detection code to Netlify and Vercel apps (plus any community adapters that use the polyfills), quite unnecessarily. It doesn't belong inside the function

.changeset/wise-adults-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/kit': patch
+---
+
+[fix] conditionally polyfill but better to support Deno
```
benmccann, line 5:
```suggestion
[fix] don't polyfill undici if using Deno or Bun
```

packages/kit/src/utils/platform.js
```diff
@@ -0,0 +1,3 @@
+export function non_node_runtime() {
+	return typeof globalThis.Deno !== 'undefined' || typeof globalThis.Bun !== 'undefined';
+}
```
Rich-Harris, line 3: what about this? I always try to avoid negatives in function names, particularly if the return value needs to be negated, resulting in a double negation. Also we can add a check for Node 18.11 if we give it a more generic name

```suggestion
export function should_polyfill() {
  if (typeof Deno !== 'undefined') return false;
  if (typeof Bun !== 'undefined') return false;
  const [major, minor] = process.versions.node.split('.');
  return +major < 18 || +major === 18 && +minor < 11;
}
```

benmccann: I'm happy with that as long as we're okay with it being unclear what version of undici users will end up with per @Conduitry's point: https://github.com/sveltejs/kit/pull/8338#issuecomment-1371572551

Rich-Harris: Yeah, would be equally happy with this

```suggestion
export function should_polyfill() {
	if (typeof Deno !== 'undefined') return false;
	if (typeof Bun !== 'undefined') return false;
	return true;
}
```

mglikesbikes:
I think this has broken the cloudflare-workers adapter, `crypto` is no longer being polyfilled it would seem.

PR #8343, opened 2023-01-05
[docs] add `@sentry/svelte` import for clarity

oekazuma:
This addition makes it easier to see that Sentry is providing Svelte SDK.

### DIFF
--- a/documentation/docs/30-advanced/20-hooks.md
+++ b/documentation/docs/30-advanced/20-hooks.md
@@ -164,6 +164,7 @@ const Sentry: any;
 
 // @filename: index.js
 // ---cut---
+import * as Sentry from '@sentry/svelte';
 import crypto from 'crypto';
 
 /** @type {import('@sveltejs/kit').HandleServerError} */
@@ -187,6 +188,8 @@ const Sentry: any;
 
 // @filename: index.js
 // ---cut---
+import * as Sentry from '@sentry/svelte';
+
 /** @type {import('@sveltejs/kit').HandleClientError} */
 export function handleError({ error, event }) {
 	const errorId = crypto.randomUUID();
--- a/documentation/docs/30-advanced/25-errors.md
+++ b/documentation/docs/30-advanced/25-errors.md
@@ -87,6 +87,8 @@ const Sentry: any;
 
 // @filename: index.js
 // ---cut---
+import * as Sentry from '@sentry/svelte';
+
 /** @type {import('@sveltejs/kit').HandleServerError} */
 export function handleError({ error, event }) {
 	// example integration with https://sentry.io/


Rich-Harris:
Thank you! I'm not sure this is the right package to use though — it looks like `@sentry/svelte` re-exports `captureException` from `@sentry/browser`, which I assume we don't want to be using in server hooks. I don't know my way around Sentry well enough to know if that will break, or if there's a server-side equivalent that we should use instead (there's `@sentry/node`, but presumably that's Node-specific).

documentation/docs/30-advanced/20-hooks.md
```diff
@@ -164,6 +164,7 @@ const Sentry: any;
 
 // @filename: index.js
 // ---cut---
+import * as Sentry from '@sentry/svelte';
```
Lms24, line 167: For the server hook, we should use `@sentry/node`, as `@sentry/svelte` only works on the browser-side.

```suggestion
import * as Sentry from '@sentry/node';
```

documentation/docs/30-advanced/20-hooks.md
```diff
@@ -187,6 +188,8 @@ const Sentry: any;
...
+
 /** @type {import('@sveltejs/kit').HandleClientError} */
 export function handleError({ error, event }) {
```
Lms24, line 194: In both files (client and server), we also have to initialize the Sentry SDK so that errors are caught. 

(btw, take this exact suggestion with a grain of salt, I'll leave it up to you how detailed you want to document the `Sentry.init` call)

```suggestion
import * as Sentry from '@sentry/svelte';

Sentry.init({...})

/** @type {import('@sveltejs/kit').HandleClientError} */
export function handleError({ error, event }) {
```

documentation/docs/30-advanced/25-errors.md
```diff
@@ -87,6 +87,8 @@ const Sentry: any;
 
 // @filename: index.js
 // ---cut---
+import * as Sentry from '@sentry/svelte';
```
Lms24, line 90: Since this is example is for the server, we need to import `@sentry/node` here as well and also initialize the SDK
```suggestion
import * as Sentry from '@sentry/node';

Sentry.init({...})
```

Lms24 (commented):
Hi, I'm working on the Sentry JavaScript SDKs. We really appreciate the shout out on the SvelteKit docs and I think it's great that you want to add more detailed instructions how to use Sentry. 

Just a couple of notes:
* As correctly pointed out by @Rich-Harris, `@sentry/svelte` is only meant for Svelte apps running in the browser (i.e. not the server part of SvelteKit apps). Instead, you can use `@sentry/node` on the server side. 
* Just calling `Sentry.captureException` isn't enough, both Sentry SDKs need to be initialized by calling `Sentry.init`. 

Coincidentally, I recently started looking into using Sentry in SvelteKit and posted some primary instructions in our [GH discussion about SvelteKit support](https://github.com/getsentry/sentry-javascript/discussions/5838#discussioncomment-4591389). We'll eventually create a proper SvelteKit SDK but these instructions can be seen as a starting point what people can already do with currently available Sentry SDKs. Feel free to take a look for more details. Also we always appreciate feedback and input ;) 

Lmk if you have additional questions, happy to help!

oekazuma:
From my research, it seems that Sentry support for SvelteKit is currently under trial.
https://github.com/getsentry/sentry-javascript/discussions/5838#discussioncomment-4591389

The server side seems to work on node by using `@sentry/node`, but not all features seem to work perfectly.

I thought about adding this to the documentation so as not to confuse users when using Sentry with SvelteKit, but I don't know what to do...

oekazuma:
@Rich-Harris @Lms24 
Thank you!

I've applied the suggested changes!

documentation/docs/30-advanced/20-hooks.md
```diff
@@ -164,6 +164,7 @@ const Sentry: any;
 
 // @filename: index.js
 // ---cut---
+import * as Sentry from '@sentry/node';
```
Lms24, line 167: Sorry, forgot to add the init call here as well:

```suggestion
import * as Sentry from '@sentry/node';

Sentry.init({...})
```

Rich-Harris:
Thank you @Lms24 — just to be sure, will `@sentry/node` work in non-Node server environments like Cloudflare Workers, Deno, Vercel Edge Functions and so on?

Lms24:
>will @sentry/node work in non-Node server environments like Cloudflare Workers, Deno, Vercel Edge Functions and so on?

@Rich-Harris unfortunately not yet. We're working on compatibility with these environments (https://github.com/getsentry/sentry-javascript/issues/5611) but no concrete time frame yet.

Rich-Harris (approved):

PR #8346, opened 2023-01-05
Revert "Improved example code"

Rich-Harris:
Reverts sveltejs/kit#8194. I disagree with that change — in SvelteKit layouts, slots are 'void' in that they can't have fallback content. This change makes it look like you _can_ put stuff inside slots, and almost looks like you're _expected_ to put something there.

### DIFF
--- a/documentation/docs/20-core-concepts/20-load.md
+++ b/documentation/docs/20-core-concepts/20-load.md
@@ -90,7 +90,7 @@ export async function load() {
 </script>
 
 <main>
-	<!-- +page.svelte is rendered here -->
+	<!-- +page.svelte is rendered in this <slot> -->
 	<slot />
 </main>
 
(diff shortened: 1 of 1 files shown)


Rich-Harris:
(cc @yousufiqbal, so that this isn't just reverted silently with no explanation)

documentation/docs/20-core-concepts/20-load.md
```diff
@@ -90,9 +90,8 @@ export async function load() {
 </script>
 
 <main>
-	<slot>
-		<!-- +page.svelte is rendered here -->
-	</slot>
+	<!-- +page.svelte is rendered here -->
```
benmccann, line 93: you could arguably say "in this slot" instead of "here" in case that helps with the original confusion at all

benmccann (approved):

documentation/docs/20-core-concepts/20-load.md
Rich-Harris:
```suggestion
	<!-- +page.svelte is rendered in this <slot> -->
```

PR #8365, opened 2023-01-06
disable illegal import detection when running tests

Rich-Harris:
This is a naive but effective fix for #8180. When running Vitest, both `process.env.VITEST` and `process.env.TEST` are `true`, meaning we can simply ignore illegal imports during unit testing.

I opted for `TEST` over `VITEST` because it will then work with any unit testing framework that works with Vite, as long as they adhere to this convention.

I think it's totally fine to skip detection during tests, since it'd be vanishingly rare that something was explicitly testing for that behaviour. The error will show up soon enough in dev, and the build will fail.

One gotcha is that if you do `TEST=true npm run dev` it will disable illegal import detection during dev (not build), which I think is probably a fine trade-off.

### Tests

### Changesets


Conduitry:
I think we should list this in the docs somewhere. Unless this `TEST=true` convention is very widespread and somehow everyone's heard about it but me?

Rich-Harris:
added some docs

dummdidumm (approved):

PR #8368, opened 2023-01-06
add package script, change behaviour of build script

### LINKED ISSUE #8240: [create-svelte] packaging shouldn't override the build command for skeletonlib
### Describe the problem

I've seen a few instances ([recent](https://discord.com/channels/457912077277855764/1054974231483523134)) of people being confused by the `build` task for the skeletonlib template.

I'm assuming it's carrying over the conventions of `build` being used to create the dist files in projects that use tsc, babel etc.? Opening an issue instead of a direct PR in case my assumption was wrong and there is some benefit I missed.

The `app` part of the skeletonlib template is often used for as a testbed and for documentation sites. For the former use case, having a build command allows package authors to verify that they don't get any weird differences between dev and build mode due to bundlers, and a build command would be needed regardless for co-located documentation sites.

move packaging into a "package" task, and reinstate the "build" task

### Alternatives considered

_No response_

### Importance

nice to have

### Additional Information

_No response_

Rich-Harris:
closes #8240

### Tests

### Changesets

### DIFF
--- a/.changeset/honest-eagles-beg.md
+++ b/.changeset/honest-eagles-beg.md
@@ -1,0 +1,5 @@
+---
+'create-svelte': major
+---
+
+Make `build` script build the app, and add a new `package` script
--- a/packages/create-svelte/templates/skeletonlib/package.template.json
+++ b/packages/create-svelte/templates/skeletonlib/package.template.json
@@ -3,7 +3,8 @@
 	"version": "0.0.1",
 	"scripts": {
 		"dev": "vite dev",
-		"build": "svelte-kit sync && svelte-package",
+		"build": "vite build && npm run package",
+		"package": "svelte-kit sync && svelte-package",
 		"prepublishOnly": "echo 'Did you mean to publish `./package/`, instead of `./`?' && exit 1"
 	},
 	"devDependencies": {


.changeset/honest-eagles-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-svelte': major
+---
+
+Make `build` script build the app, and add a new `package` script
```
Rich-Harris, line 5:
```suggestion
[breaking] Make `build` script build the app, and add a new `package` script
```

arxpoetica (approved):

packages/create-svelte/templates/skeletonlib/package.template.json
```diff
@@ -3,7 +3,8 @@
 	"version": "0.0.1",
 	"scripts": {
 		"dev": "vite dev",
-		"build": "svelte-kit sync && svelte-package",
+		"build": "vite build && npm run package",
+		"package": "svelte-kit sync && svelte-package",
```
gtm-nayan, line 7: IIRC svelte-package doesn't do checking and svelte-check isn't included in the library template. svelte-check and a check command should probably be added in as well esp with #8484 merged.

Rich-Harris: Good point. I wonder if we should try and make `svelte-check` only apply to packaged files rather than the entire app — would be annoying if you couldn't publish because of a finicky type error in your docs site that you don't really care about

dummdidumm: For that we could create a `tsconfig.lib.json` which only `include`s the files inside `lib` and reference that from `svelte-check`

Rich-Harris: feels a bit messy to create a new file, and if `config.kit.files.lib` changed the developer would also need to change `tsconfig.lib.json`... perhaps there's a way to integrate checking into the `svelte-package` command instead?

dummdidumm:
Closing in favor of #8922 (which will also make this change)

PR #8370, opened 2023-01-06
error in dev mode if global fetch is used with relative URL

### LINKED ISSUE #7976: Attempt to error better when wrong fetch is used on an endpoint
### Describe the problem

I've now twice gotten burned by using the "wrong" fetch in my loaders. Running

fetch("/api/my-endpoint")

with the global fetch that's in scope will puke, since "/api/my-endpoint" has no way of resolving. Obviously you have to destructure the fetch that's passed to your loader. Simple enough, but it's an easy thing to miss when you're new (at SvelteKit), and it's *maddening* to debug.

Would it be possible for the global fetch to detect when an (otherwise valid) api endpoint is being requested, and error with something clear about needing to use the provided fetch to hit endpoints?

I know this woulnd't be foolproof, since shared loaders are (I assume?) using the browser's provided fetch, which would be hard for you all to wrap, but on the server you're already providing your own fetch, so could this check be added to that (presumably for non-production builds)? 

### Alternatives considered

n/a

### Importance

nice to have

### Additional Information

_No response_

Rich-Harris:
closes #7976

### Tests

### Changesets


Conduitry (changes requested):
What happens if someone uses a protocol-relative URL like `//foo/bar`? This check would fail to catch it and it would (I think) then still fail with the global `fetch`.

Can we instead just check something simpler like that the URL matches `/^\w+:\/\//`? Or heck, just hard code `http://` and `https://` since I don't think anything else would succeed anyway?

packages/kit/src/exports/vite/dev/index.js
```diff
@@ -29,6 +29,19 @@ export async function dev(vite, vite_config, svelte_config) {
...
 
+	const fetch = globalThis.fetch;
+	globalThis.fetch = (info, init) => {
+		if (typeof info === 'string') {
+			if (new URL(info, 'fake://fake').protocol === 'fake:') {
+				throw new Error(
+					`Cannot use relative URL (${info}) with global fetch — use \`event.fetch\` instead: https://kit.svelte.dev/docs/web-standards#fetch-apis`
+				);
+			}
+		}
```
Rich-Harris, line 40: probably simpler

```suggestion
		if (typeof info === 'string' && !/^\w+:\/\//.test(info)) {
			throw new Error(
				`Cannot use relative URL (${info}) with global fetch — use \`event.fetch\` instead: https://kit.svelte.dev/docs/web-standards#fetch-apis`
			);
		}
```

Rich-Harris:
The check would work — `new URL('//foo.com', 'fake://fake').protocol === 'fake:'`. But yeah, a regex is probably simpler. I opened a suggestion

Conduitry:
Oh, whoops, not sure what I was thinking there anymore. I do still like the regex though.

dummdidumm (approved):
Approving the regex approach. Hardcoding http(s) would make things like tauri apps etc impossible to use the global fetch

